PR #114, opened 2019-11-12
misc fixes to paths to get AL pxrUsdTranslators tests running

elrond79:
- $USD_MAYA_ROOT not defined / needed (should always be part of this project)
- made sure that the pixar maya plugin's python dir is added to the
  pythonpath before the main USD python dir; this is necessary because the
  pixar maya plugin also defines a pxr/__init__.py, which needs to come
  before the main USD pxr/__init__.py
- created / used explicit names in place of some $# variables (ie, $3 / $4),
  for clarity


elrond79:
I believe so, as part of our vset export pipe.

kxl-adsk:
@elrond79 can you switch this PR to merge into dev branch?

elrond79:
Done!

PR #115, opened 2019-11-12
[pxr] fix for building with PXR_VALIDATE_GENERATED_CODE

elrond79:
Handy when doing releases...


kxl-adsk:
What's the reasoning for opening this PR for master?

elrond79:
Because it's pertinent to master's purpose (stable release builds), and because it's easy to merge from master to dev, but we don't want to merge the other way right now.

Having said that, it's also not crucial, so if you'd prefer it be against dev, I can change the target.

kxl-adsk:
Please change to be against dev.

elrond79:
Done.

PR #124, opened 2019-11-19
enable setting of USDMAYA_PROXYSHAPEBASE debug symbol via env var

elrond79:
also, get rid of PXRUSDMAYA_TRANSLATORS debug symbol, because it's
not used (and also doesn't have an env symbol defined)


kxl-adsk:
PXRUSDMAYA_TRANSLATORS debug symbol will come in the next drop from refactoring_sandbox...but not needed for now.

PR #128, opened 2019-11-22
only register VP2.0 fragments when in interactive sessions of Maya

mattyjams:
Registering the fragments fails when in non-interactive sessions of Maya. This mainly prevents errors from being emitted during tests when plugins are initialized.

### DIFF
--- a/lib/render/vp2ShaderFragments/shaderFragments.cpp
+++ b/lib/render/vp2ShaderFragments/shaderFragments.cpp
@@ -110,6 +110,11 @@ namespace
 // Fragment registration
 MStatus HdVP2ShaderFragments::registerFragments()
 {
+    if (MGlobal::mayaState() != MGlobal::kInteractive) {
+        // Only register fragments in interactive Maya sessions.
+        return MS::kSuccess;
+    }
+
     MHWRender::MRenderer* theRenderer = MHWRender::MRenderer::theRenderer();
     if (!theRenderer) {
         return MS::kFailure;
@@ -176,6 +181,11 @@ MStatus HdVP2ShaderFragments::registerFragments()
 // Fragment deregistration
 MStatus HdVP2ShaderFragments::deregisterFragments()
 {
+    if (MGlobal::mayaState() != MGlobal::kInteractive) {
+        // Only deregister fragments in interactive Maya sessions.
+        return MS::kSuccess;
+    }
+
     MHWRender::MRenderer* theRenderer = MHWRender::MRenderer::theRenderer();
     if (!theRenderer) {
         return MS::kFailure;


HdC-adsk:
@mattyjams We still need to register these shader fragments for non-interactive maya session so that batch render (command line render) can generate the correct images. Could you show me the errors? Maybe we can delay the registration somewhere to fix those errors.

mattyjams:
Hmm, interesting. Maybe this has something to do with our Maya configuration then? I'm seeing something like this at the top of the stderr output for basically every headless test that loads our proxy shape plugin:
```
This plugin does not support createPlatformOpenGLContext!
Warning: This file is from an older version of Maya.  If saved, it will not be readable by previous versions.
File read in  0.18 seconds.
VP2 Error : Failed to initialize graphics device.

API error detected in maya_open_source/maya-usd/lib/nodes/proxyShapePlugin.cpp at line 86
: (kFailure): Unexpected Internal Failure

API error detected in maya_open_source/maya-usd/plugin/pxr/maya/plugin/pxrUsd/plugin.cpp at line 73
```

The `VP2 Error` is likely not a good start, and then the first API error is on the `CHECK_MSTATUS()` call after attempting to register the fragments, and that error status then propagates up to the pxrUsd plugin causing the second API error.

I'm not seeing any output about fragments or fragment graphs failing to register, so that suggests that either getting the MRenderer or getting the MFragmentManager must be failing in `HdVP2ShaderFragments::registerFragments`. I can dig into it a bit more tomorrow and see if I can figure anything else out.

HdC-adsk:
Ok, so it is the headless config that makes thing different. To be compatible with headless config, we should not really initialize Viewport 2. I will comment in the code to propose a change and please help to test to see if it can pass the headless tests. Thanks.

lib/render/vp2ShaderFragments/shaderFragments.cpp
```diff
@@ -110,6 +110,11 @@ namespace
 // Fragment registration
 MStatus HdVP2ShaderFragments::registerFragments()
 {
+    if (MGlobal::mayaState() != MGlobal::kInteractive) {
+        // Only register fragments in interactive Maya sessions.
+        return MS::kSuccess;
+    }
+
     MHWRender::MRenderer* theRenderer = MHWRender::MRenderer::theRenderer();
```
HdC-adsk: As I commented in the conversation, we should not reject registration/deregistration. For headless config, we should not really ask Viewport 2.0 to initialize. Therefore, the line can be changed to:

// We don't want to force Viewport 2.0 to be initialized.
MHWRender::MRenderer* theRenderer = MHWRender::MRenderer::theRenderer(false);

lib/render/vp2ShaderFragments/shaderFragments.cpp
```diff
@@ -176,6 +181,11 @@ MStatus HdVP2ShaderFragments::registerFragments()
 // Fragment deregistration
 MStatus HdVP2ShaderFragments::deregisterFragments()
 {
+    if (MGlobal::mayaState() != MGlobal::kInteractive) {
+        // Only deregister fragments in interactive Maya sessions.
+        return MS::kSuccess;
+    }
+
     MHWRender::MRenderer* theRenderer = MHWRender::MRenderer::theRenderer();
```
HdC-adsk: Same as registration.

lib/render/vp2ShaderFragments/shaderFragments.cpp
HdC-adsk: With this change, `theRenderer->getFragmentManager()` and `theRenderer->getShaderManager()` will return `nullptr`. We should not treat them as failure as they are expected for headless config. So, we should adjust the code structure a bit and only do the work when the fragment manager and the shader manager (i.e. move the major body of function inside a `if (fragmentManager)` clause.

mattyjams:
Great, thanks @HdC-adsk! That makes sense.

For clarity, I'm going to close out this PR and open up a new one with the different approach to solving this that you suggested. We're doing something similar in the pxrUsdPreviewSurface plugin of checking the interactive-ness of the session before registering/de-registering fragments, so I'll make a similar fix there as well:

https://github.com/Autodesk/maya-usd/blob/c567a1661152ec2450359767fa2b0ca8cac4dc83/plugin/pxr/maya/plugin/pxrUsdPreviewSurface/plugin.cpp#L217

Thanks again!

HdC-adsk:
Thanks @mattyjams . You are helping us out!

mattyjams:
Replaced this with #130. Thanks!

PR #130, opened 2019-11-22
skip fragment registration/de-registration when MRenderer or MFragmentManager are not available

mattyjams:
This resolves an issue where headless contexts of Maya would emit API errors during plugin registration because the renderer cannot be initialized. We now no longer force the renderer to initialize, and we bypass fragment registration/de-registration and return success if we cannot get a handle to the MRenderer or MFragmentManager.

More detail is available in the abandoned PR #128.


HdC-adsk:
@kxl-adsk I've locally built and tested the change. It can work for both interactive maya and command line render. I believe @mattyjams has covered the tests on headless configuration.

Therefore please merge the PR when you see fit. Thanks.

PR #133, opened 2019-11-26
Update to Pixar USD source version to latest dev sha

kxl-adsk:


kxl-adsk:
@mattyjams we are updating our internal artifacts and it looks like dev branch should build and run on the latest Pixar USD dev commit (2ef943b). Can you confirm by approving this PR?

PR #134, opened 2019-11-26
Merge dev to refactoring sandbox 6

kxl-adsk:
This brings the latest changes from dev to refactoring_sandbox. One new issue which is related to the latest USD dev branch (required to build this branch) is that hdMaya no longer has basic supports materials. We will have to resolve this when backporting mtoh to the dev branch.


kxl-adsk:
Logged #135 to keep track of issues with materials in hdMaya and new USD versions (after refactoring of resource registry).

Issue #126, opened 2019-11-21
ProxyShape plays only once for animated USD mesh


BigRoy:
### Issue

When playing through time with an animated geometry the first playback through time seems smooth with the attached USD pighead. It gives *no visible errors* as to why it stops playing back. Additionally I've had it happen that when selecting the mesh the wireframe would actually play correctly, but then after the first cycle there also stop playing altogether. And then, when going to another frame and reselecting the mesh shows the updated "wireframe selection overlay". This seems to be a display update bug in Viewport 2.0.

[pighead_mesh.zip](https://github.com/Autodesk/maya-usd/files/3875737/pighead_mesh.zip)

_The attached USD scene is an export from Houdini._

It works fine however when playing back other animated production `.abc` files I had laying around, with those I can *not* reproduce it.

### To reproduce

This is easily visible when loaded through `AL_USDMayaPlugin` with the menu item but the same occurs through the `mayaUsdProxyShape` with this:
```python
from maya import cmds

path = "path/to/pighead_mesh.usd"
node = cmds.createNode("mayaUsdProxyShape")
cmds.setAttr(node + ".filePath", path, type="string")

# Make sure to connect time
cmds.connectAttr("time1.outTime", node + ".time")
```

It seems to fail as soon as you "go back in frames" and especially whenever you've played this file until the end of the frame samples, e.g. play all frames past frame `50` and then restart. Then you'll see no motion whatsoever in the Viewport.

**What I've tried:**

- Running `ogs -reset` does resolve it, until it reaches the end again.
- I have Maya 2019 timeline caching disabled. Enabling it makes no difference in behavior.
- It happens in both parallel and DG evaluation mode.
- Like the USD file my Maya scene is 24 FPS, that matches fine. With other FPS settings in Maya the issue is also there.
- When enabling debugging features from the AL USD menu there's little new information I can see that might cause it. With `USD_VALUE_RESOLUTION` it does show it consistently trying to resolve the time samples as I step through time. Like:
```
RESOLVE: reading field /usdexport4/mesh_0.points:timeSamples from @p:/Projects/test1/assets/usd_assets/work/model/houdini/OUTPUT/usd/pighead/pighead_model_full.usd@, with requested time = 18.000 (local time = 18.000) reading from sample 18.000 
[...]

PR #139, opened 2019-11-28
Fix regression due to avoiding repepetive MPxSubSceneOverride::setGeometryForRenderItem() call

HdC-adsk:
https://github.com/Autodesk/maya-usd/issues/126 ProxyShape plays only once for animated USD mesh

The performance optimization was to avoid repetitive calls into
MPxSubSceneOverride::setGeometryForRenderItem() but unfortunately
it brought regression because the API call has a side effect to
trigger consolidation/instancing update.


kxl-adsk:
@mattyjams , @robthebloke is this something you would like to review as well?

HdC-adsk:
This is to fix the issue: #126 (ProxyShape plays only once for animated USD mesh)

PR #140, opened 2019-11-29
Replacing direct use of 'python' executable with variable.

seando-adsk:


kxl-adsk:
This change is required in the dev branch as well. How about we put it there first and then merge into refactoring sandbox? (otherwise, we will have duplicated commits and merge conflicts)

PR #141, opened 2019-11-29
Replacing direct use of 'python' executable with variable.

seando-adsk:


kxl-adsk:
@murphyeoin can you please review this change?

murphyeoin:
Hi @seando-adsk code itself looks perfect. Where is the variable going to be defined?

seando-adsk:
@murphyeoin It gets defined by FindPythonInterp.cmake which is called here: https://github.com/Autodesk/maya-usd/blob/dev/cmake/python.cmake#L59

seando-adsk:
Note: I had previously made this same type of change a while ago. I guess some new instances of 'python' crept back in.

PR #149, opened 2019-12-03
[al] make pushToPrim default to On

elrond79:
...a recent AL change (55e01cf9e52d14069f86763bf447a9c6ee7650a9) added
the ability to control the setting for pushToPrim for newly created
nodes via an optionVar, which is useful, but it also changed the default
from on to off.

In addition to being breaking behavior, this is also highly
non-intuitive for artists, who will be confused when they try to
transform the usd object, and it does nothing.  I would also argue that
the ability to pushToPrim is the primary thing which distinguishes this
plugin from the Pixar plugin, and as such, is probably why most people
are using it in the first place.

NOTE: I haven't yet run the tests for this, as the tests are still not fully functional on the dev branch.  Assuming we agree that this change should be made, we may wish to delay merging it in until the tests work, at which time we can ensure that this modification doesn't require making changes to any tests.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/PluginRegister.h
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/PluginRegister.h
@@ -211,7 +211,7 @@ MStatus registerPlugin(AFnPlugin& plugin)
 
   if(!MGlobal::optionVarExists("AL_usdmaya_pushToPrim"))
   {
-    MGlobal::setOptionVarValue("AL_usdmaya_pushToPrim", false);
+    MGlobal::setOptionVarValue("AL_usdmaya_pushToPrim", true);
   }
 
   MStatus status;


cfmoore007:
@elrond79  - This also initially tripped us up ( and readAnimatedAttributes ).

We resorted to setting the default optionVar ourselves to work-around the issue.  

If we're going for this one, why not the other?

elrond79:
Hmm, good point - I guess I wasn't aware the default for that had changed as well.

And yeah, for now we're also setting the default ourselves.  Just think it might trip up newcomers.

kxl-adsk:
Based on the conversation, it's unclear to me whether or not there is still something to do in this PR? To be precise, should `readAnimatedAttributes ` have a matching default value? 

Also, I see @robthebloke reviewed it. Do we need anyone else to approve it? (@murphyeoin)

elrond79:
@robthebloke or @murphyeoin - thoughts on also setting readAnimatedValues default to true? I guess the default for this changed slightly further back - in 97d3ff3e.

elrond79:
I decided to make readAnimatedValues it's own PR, since I could potentially see AL objecting to that, but not to this.

elrond79:
Oops, forgot to link it: https://github.com/Autodesk/maya-usd/pull/226

kxl-adsk:
OK. Just double-checking - @murphyeoin you are good to make this change?

elrond79:
Thanks!

PR #151, opened 2019-12-04
Fix #144: Quote path to allow arguments to parse correctly

### LINKED ISSUE #144: Build fails due to set_tests_properties called with incorrect number of arguments.
# Issue

In some case building fails on [this line in `plugin/al/usdtransaction/AL/usd/transaction/tests/CMakeLists.txt`](https://github.com/Autodesk/maya-usd/blob/f952d26bd76d59b141e3944add16c7c946620ea7/plugin/al/usdtransaction/AL/usd/transaction/tests/CMakeLists.txt#L52) with the following error:

```
-- Found GTest: D:/git_pipeline/rez-packages/maya_usd/build/platform-windows/arch==AMD64/maya-usd/googletest-install/lib/gtest.lib
CMake Error at plugin/al/usdtransaction/AL/usd/transaction/tests/CMakeLists.txt:50 (set_tests_properties):
  set_tests_properties called with incorrect number of arguments.
```

This makes the build fail.

A [working patch was mentioned here](https://github.com/Autodesk/maya-usd/issues/86#issuecomment-561011247). It [fixed my build problems here](https://github.com/Autodesk/maya-usd/issues/126#issuecomment-561052142).

Likely it fails to correctly parse the arguments if there's a path or variable resolved there that contains a path with spaces (or maybe backslashes?), that I didn't figure out yet.

---

Should this be patched up with a PR?

BigRoy:
This fixes the issue described in #144


BigRoy:
I can seem to find little information related to it too. I did find [this post mentioning quotes are essential for a case with labels](https://stackoverflow.com/a/24500217) that are set with spaces so it seems. Also the GoogleTest documentation [shows an example with double quotes too](https://cmake.org/cmake/help/v3.9/prop_test/PASS_REGULAR_EXPRESSION.html?highlight=set_tests_properties). Not sure if that helps, but at least it shows the quotes aren't wrong syntax-wise I suppose.

HamedSabri-adsk:
I agree the syntax is very weird and not so intuitive. However, it is OK for now. 

@BigRoy is this issue platform specific?

I have ran into similar issues when I was adding UFE tests. In that case, I used set_property routine.

https://github.com/Autodesk/mayausd/blob/b77aa8fdf8f5ed684503cfea9ea8be35b0bc83a9/test/lib/ufe/CMakeLists.txt#L108

HamedSabri-adsk:
@BigRoy "shows an example with double quotes too" doesn't seem to link correctly to the GoogleTest example.

BigRoy:
> "shows an example with double quotes too" doesn't seem to link correctly to the GoogleTest example.

It doesn't? It works for me. Does [this](https://cmake.org/cmake/help/latest/prop_test/PASS_REGULAR_EXPRESSION.html) work?
Funnily enough for me `CMakeList` link you posted doesn't work, hehe.

Anyway, let me know if you need more information from my end.

elrond79:
So, I was playing with this and learned something - I thought the reason you quotes args in cmake was to avoid problems with a "var with spaces" getting translated into multiple args.  Turns out it has nothing to do with spaces, just lists (ie, semicolons).

To see an example, check this out: https://gist.github.com/elrond79/ac6941b7337c607b10521a57cac85b70

You can download it and run with `cmake -P cmake_quoting_test.cmake`.

So, this PR would help in that it would preserve a windows-style PYTHONPATH which has ";" separators (and would therefore otherwise get treated as a list).  However, I'm concerned that it would still not be correct on windows, because it would be inserting a linux-style ":" separators, which would not be treated "special" on windows.

Unless CMAKE itself does something tricky, and always converts ":" characters to ";" for the ENVIRONMENT property on windows?

HamedSabri-adsk:
@BigRoy Hmm, weird!  
Try this:
https://github.com/Autodesk/maya-usd/blob/b77aa8fdf8f5ed684503cfea9ea8be35b0bc83a9/test/lib/ufe/CMakeLists.txt#L108

How does your link is related to GoogleTest?? What you are showing me points at Cmake doc PASS_REGULAR_EXPRESSION. 

@elrond79 
Yes, I also noticed the linux-style ":" separators which could be problematic on Windows but it looks like things has worked on Windows according to @BigRoy . It could be that set_tests_properties has magically converted the ":" to ";".  

I have seen some weird behaviours with some of these CMake property calls before.

Example:

On Windows, there seems to be a bug in CMake's separate_arguments command. This command is supposed to parses a space-separated string <args> into a list of items, and stores this list in semicolon-separated form.

See doc:
https://cmake.org/cmake/help/latest/command/separate_arguments.html

The problem is if your original list args have space between them ( e.g  Program Files ), then this command doesn't respect it and concatenates them which may not be desired.

```
set(mylist 
    "c:/Program Files/potato-yellow"
    "c:/basket/potato-reds"
    "c:/basket/potatoes"
    "c:/my basket/potatoes"
)
```

`separate_arguments(mylist WINDOWS_COMMAND "${mylist}")
`
```
c:/ProgramFiles/potato-yellow;
c:/basket/potato-reds;
c:/basket/potatoes;
c:/mybasket/potatoes;

```

elrond79:
Actually, separate_arguments is working as intended in your example.  The problem is the weird way that cmake itself handles lists: a "list" in cmake is just a semi-colon separated string.  Further, when you feed multiple arguments to many arguments (such as set), it implicitly converts them to a list.  So in your example, the set is actually inserting ";" between your four paths.

`separate_arguments`, however, assumes that the separators you "want" ARE spaces (it's a windows command), and that any ";" are NOT supposed to be interpreted as special (they might be path-separators, for example).  So, it ESCAPES any ";", and then converts any spaces into un-escaped ";", to turn it into a cmake-list.

This is all demonstrated by this script (again, run with cmake -P):

```
set(mylist 
    "first item"
    "second-item"
)
message("mylist: ${mylist}")
# mylist: first item;second-item

separate_arguments(mylist_sep WINDOWS_COMMAND "${mylist}")
message("mylist_sep: ${mylist_sep}")

# mylist_sep: first;item\;second-item
message("mylist_sep (not quoted): " ${mylist_sep})

# mylist_sep (not quoted): firstitem;second-item
```

This "everything is a string" stuff is one of the reasons I really dislike cmake as a language.  Even bash actually has a true array type (though, admittedly, with a very awkward syntax!)

elrond79:
Also - the mere fact that the build now works doesn't mean that the set_properties is working as intended.  Even if the tests pass, it could be because they're setting some env vars "themselves".

BigRoy:
> Also - the mere fact that the build now works doesn't mean that the set_properties is working as intended

That's also what I'm scared of too - that this just passes because it fails to run something else as it should. What's the best way to confirm it does what it needs to do?

HamedSabri-adsk:
@elrond79 fair enough. In my case, I am more fed up with dealing with ";" ":" "spaces", etc... on different platforms.

HamedSabri-adsk:
I propose that we do a similar thing that was done with UFE tests as I mentioned in the link ^

This is IMO a better solution for 2 reasons:

1- it is easy to add  a new path to the pythonPath list.
2- it is very clear that list entries need to be separated either by ":" or ";"  depending on the platform.

```
set(pythonPath
    "${CMAKE_BINARY_DIR}/${INSTALL_DIR_SUFFIX}/usdtransaction"
    "${CMAKE_BINARY_DIR}/src"
    "$ENV{PYTHONPATH}"
)
if(IS_WINDOWS)
    string(REPLACE ";" "\;" pythonPath "${pythonPath}")
else()
    separate_arguments(pythonPath NATIVE_COMMAND "${pythonPath}")
    string(REPLACE "\;" ":" pythonPath "${pythonPath}")
endif()
set_tests_properties(${USDTRANSACTION_PYTHON_TEST_NAME} PROPERTIES
ENVIRONMENT "PYTHONPATH=${pythonPath}"
)
```

BigRoy:
With [this commit](https://github.com/Autodesk/maya-usd/pull/151/commits/f46907fd333599a70fd89beea4bc54574b6b6887) I changed it to what @HamedSabri-adsk proposed. It compiles fine for me, but then again... I'm not actually sure it does what it needs to do. So, definitely **test** this if you know what it should do.

HamedSabri-adsk:
@BigRoy Thanks for the change.  
Fyi, I've created a new PR ( https://github.com/Autodesk/maya-usd/pull/167 ) that would fix bunch of related issues with MacOSx as well as two of AL's python unit tests(testTransactionManager, testTransaction).These two tests are failing on Windows. I have the fix for them but I am away for a few days. Once PR167 is reviewed, tested, and merged, I will close this one.

BigRoy:
@HamedSabri-adsk should we close this PR in favor of #167 ?

HamedSabri-adsk:
@BigRoy  Thanks. I am closing this one. 

Regarding #167, I am going to close that PR as well later :)

We are going to have a new PR very soon that will bring in a lot of fixes and improvements to Animal Logic unit tests. This new PR will have the fixes from #167 already in it.

Issue #95, opened 2019-10-31
refactoring_sandbox : UFE : 'invalid' warnings for SetTranslate on an incompatible xformable.


BigRoy:
### Issue

When trying out the [example Pixar Kitchen set USD file](http://graphics.pixar.com/usd/downloads.html) I'm getting the following errors when selecting lots of small objects at the same time, e.g. marque selecting over the bowl of cereals on the kitchen table, and then translating those things around.

```
Warning: in SetTranslate at line 701 of D:\dev\usd\USD\pxr\usd\lib\usdGeom\xformCommonAPI.cpp -- XformCommonAPI: Attempted to SetTranslate on an incompatible xformable </Kitchen_set/Props_grp/DiningTable_grp/TableTop_grp/CerealBowl_grp/Cheerios_grp/CheerioB_2/Geom/Cheerio>.
Warning: in SetTranslate at line 701 of D:\dev\usd\USD\pxr\usd\lib\usdGeom\xformCommonAPI.cpp -- XformCommonAPI: Attempted to SetTranslate on an incompatible xformable </Kitchen_set/Props_grp/DiningTable_grp/TableTop_grp/CerealBowl_grp/Cheerios_grp/CheerioB_20/Geom/Cheerio>.
Warning: in SetTranslate at line 701 of D:\dev\usd\USD\pxr\usd\lib\usdGeom\xformCommonAPI.cpp -- XformCommonAPI: Attempted to SetTranslate on an incompatible xformable </Kitchen_set/Props_grp/DiningTable_grp/TableTop_grp/CerealBowl_grp/Cheerios_grp/CheerioB_1/Geom/Cheerio>.
Warning: in SetTranslate at line 701 of D:\dev\usd\USD\pxr\usd\lib\usdGeom\xformCommonAPI.cpp -- XformCommonAPI: Attempted to SetTranslate on an incompatible xformable </Kitchen_set/Props_grp/DiningTable_grp/TableTop_grp/DinnerMat_1/Geom/DinnerMat>.
Warning: in SetTranslate at line 701 of D:\dev\usd\USD\pxr\usd\lib\usdGeom\xformCommonAPI.cpp -- XformCommonAPI: Attempted to SetTranslate on an incompatible xformable </Kitchen_set/Props_grp/DiningTable_grp/TableTop_grp/CerealBowl_grp/Cheerios_grp/CheerioB_43/Geom/Cheerio>.
[...]

Issue #76, opened 2019-10-25
[MAYA-103876] refactoring_sandbox : UFE : Translating Prim in objectSpace does not take into account its local rotation changes


BigRoy:
### Issue

Whenever the translate tool is in "object space" it will totally ignore any local scene rotations that were performed on the Prims in the USD scene. It will always move in the original local space direction that was stored in the USD scene prior to any local scene rotation changes to that Prim.

1. Load USD proxy shape.
2. Select a transform in it.
3. Rotate the objects around.
4. Translate along a single axis in objectSpace.

Note that now the translation is happening as if Step 3 never occurred.

### Specs

- Windows 10
- Maya 2019.2
- refactoring_sandbox ( `d0bb624` ) 
- Pixar USD dev ( `b85ddac` )

---

Mentioned in https://github.com/Autodesk/maya-usd/issues/51

cfmoore007:
One of our layout artists here also ran into this problem.  They use object space manipulation all of the time.

santosg87:
this is now fixed internally and should be available on the next PR :)

cfmoore007:
This appears to be addressed in PR115 with our tests here.

@BigRoy is it also working for you?

kxl-adsk:
This is already fixed for a while. @BigRoy please reopen if you find we missed something.

PR #158, opened 2019-12-05
Added class to provide an API to work with maya transform nodes types 

murphyeoin:
Within AL_USDUtils, there is a new MayaTransformAPI class that will allow you to easily work with maya transforms in USD. Includes Python bindings

@kxl-adsk this is Rob's API that we discussed. He is currently integrating with UFE

This is the first PR I'm submitting since we got our tests working, so I can fairly much guarantee that this passes the AL test suite on a VFX Platform 2019 compliant CentOS 7.4/USD 0.19.11/maya-2019 configuration


kxl-adsk:
This is a core functionality that would be great to have in the core library. If I'm not mistaken, this is as well where you will need it in order to integrate with the UFE plugin. 

Any reasons to keep it in the AL plugin only?

robthebloke:
We put it into the usdutils library because it's code that doesn't have any dependencies on Maya. Having it available to other parts of our pipeline (without needing to pull in the maya libs) seemed like the way to go.

kxl-adsk:
@robthebloke Exactly, this is why for example having it in https://github.com/Autodesk/maya-usd/tree/dev/lib/utils would make it available to all plugins.

robthebloke:
@kxl-adsk I think you're misunderstanding me. Look here: https://github.com/Autodesk/maya-usd/blob/dev/lib/utils/utilFileSystem.cpp#L5  Those utils clearly depend on the Maya API.

We can't really be in a situation where our Nuke plugin has to link against the Maya API to use a single class that has no dependencies on the Maya API. That's not workable for us.

elrond79:
So, to muddy the waters slightly - when Eoin first mentioned that this PR might be coming, I said we also had a similar PR, which sounded like it covered very similar ground.  On a quick glance over this, it seems like it indeed does.  It was first proposed [here](https://github.com/AnimalLogic/AL_USDMaya/pull/43), and actually merged, before it was backed out due to some merge conflicts, at which point [this PR](https://github.com/AnimalLogic/AL_USDMaya/pull/64) was made.  It got held up over concerns over linking against a library in the Pixar plugin... and there wasn't a clear place to put "common" components at the time.  The current incarnation, against maya_usd, can be found [here.](https://github.com/Autodesk/maya-usd/compare/dev...LumaPictures:pr/al/use_UsdMayaXformStack)

@kxl-adsk you mentioned you'd like to move this functionality to a higher level, since the general concept is fairly generic.  That's precisely what [UsdMayaXformStack](https://github.com/LumaPictures/maya-usd/blob/dev/plugin/pxr/maya/lib/usdMaya/xformStack.h) does - it defines the general concept of an "xform stack", which is simply a defined set of xformops in a certain order.  Maya's canonical xform stack is one example, the XformCommonAPI is another.  It then defines methods for checking whether a given set of xform ops is compatible with a stack, etc.

Actually, the concepts behind UsdMayaXformStack are general enough that I've considered trying to make it into a PR against core USD... however, it seems the other main plugins (houdini + katana) don't have a fixed xform stack in the same way that Maya does, so I wasn't sure how much utility there would be for it outside of Maya.

Regardless - if we're attempting to unify stuff, then at some point we'll need to unify this code with UsdMayaXformStack, already in  use within the Pixar plugin.  I'll try to do a more in-depth comparison of these two branches - where they overlap, what unique functionality they have, and ideally, how they might be merged - but likely won't be able to get to until after addressing the outstanding Maya-to-Hydra work.

elrond79:
One feature that I think UsdMayaXformStack handles nicely, that a quick look through this PR doesn't show, is dealing with split / non-split pivots.

ie, stacks like the XformCommonAPI define a single pivot, used for both rotate and scale.  Maya's API defines separate rotate + scale pivots.  A simple attempt to translate an XformCommonAPI stack into a Maya stack might then conclude that they're incompatible; however, an XformCommonAPI stack can be translated to a Maya stack by simply setting the rotate + scale pivots to be the same.

Also, since the AL plugin maintains a "live" link to the USD xform op list, it doesn't split the single pivot op into two until required (ie, the rotate + scale pivots are set to different values), ensuring that any USD-level changes to the single pivot will still get reflected in the maya object.

kxl-adsk:
Thanks for bringing this up @elrond79. It sounds like a topic for the TSC meeting. 

I wasn't aware of UsdMayaXformStack and it does sound like we should dig into it more and understand the potential for reconciliation (having comparison will help). One thing I will add to this, we also have to make sure this helps us solve issues we are seeing reported around UFE manipulation limitations (caused by restricting it to XformCommonAPI). @cfmoore007 would be great to hear from you regarding that.

Going back to the requirement to build a higher level, common components that all plugins can build on top of. I see your point @robthebloke about having the possibility to build components that are reusable in the pipeline and not dependent on Maya, but this doesn't mean we can't create a new lib in /lib folder.

BigRoy:
> One thing I will add to this, we also have to make sure this helps us solve issues we are seeing reported around UFE manipulation limitations (caused by restricting it to XformCommonAPI). @cfmoore007 would be great to hear from you regarding that.

To be sure, are you referring to #95 and #76 - or other cases?

cfmoore007:
We are currently investigating a double scaling issue with multipleXform ops (outside the UFE context) and pushToPrim.

Our setup is as follows :
"
vector3f xformOp:scale:modelScale = (0.5, 0.5, 0.5)
matrix4d xformOp:transform:model = ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) )
token[] xformOpOrder = ["xformOp:transform:model","xformOp:scale:modelScale"]
"

With UFE, #95 is definitely in the mix.  We are using opSuffixes quite a bit ( as you can see above ).

Actually, looking at this closer (with UFE), it's a slightly different error :
"
Unable to get attribute associated with the xformOp 'xformOp:scale:modelScale', on the prim at path </blah>. Skipping xformOp in the computation of the local transformation at prim.
"
More digging to do!  This particular flavor of xform issue has not been logged yet ...

subing85:
Hi，
How to add renderman Object Statements to pxr.UsdGeom.Mesh?

I added the render man Object Statements in to mesh for example* Trace Auto
Bias, Trace Bais and Trace Displacements* on to the mesh and assign the the
renderman shader to that respective mesh in maya, while export the usd I
would like to export  render man Object Statements ( *Trace Auto Bias*) as
well, How to do that?.

I want to bring this USD in Katana

In Maya render man Object Statements  show as a custom attributes, If I
export custom attribute
with USD, The U

Any Examples or help  would be greatly appreciated!

Regads,
Subin

elrond79:
So, I just had a chat with the Animal Logic folks... and we decided that for now we should just move ahead with their implementation (this PR).  I'd like to circle back and unify both our efforts at some point... but I don't think I have time to do that now, and in the meantime, they apparently have a bunch of other PRs which rely on this.

Also, our work was done in an attempt to unify the AL and Pixar code, and eliminate some bugs we were seeing... we haven't really built much extra infrastructure on top of it.  So we can probably use AL's work until we can get around to merging everything again

kxl-adsk:
Sounds good. Thx for the update.

murphyeoin:
Thanks @elrond79! @kxl-adsk as discussed we will move the code to the core utilities lib, and split out later if we have enough functionality that doesn't depend on Maya

kxl-adsk:
Now with #288 you should have a place to put this in.

kxl-adsk:
@murphyeoin any update on this PR?

lib/CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ target_link_libraries(${PROJECT_NAME}
         usdSkel
         usdUtils
         vt
+        mayaUsdUtils
```
kxl-adsk: I can't find why we need this dependency?

lib/usd/utils/tests/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
+find_package(GTest REQUIRED)
```
kxl-adsk: Tests are living in a separate `test` folder where we match the folder structure from source directory, for example: 
https://github.com/Autodesk/maya-usd/tree/dev/test/lib/usd/utils
https://github.com/Autodesk/maya-usd/tree/dev/test/lib/ufe

Please move all the tests following pattern above.

lib/usd/utils/MayaTransformAPI.h
```diff
@@ -0,0 +1,263 @@
...
+    { return bool(m_prim); }
+
+  /// \brief  returns the API this prim is compatible with
+  /// \return the transform API that was found
+  TransformAPI api() const
+    { return m_api; }
+
+protected:
+  void convertMatrixOpToComponentOps(const UsdGeomXformOp& op);
+  bool matchesMayaTrasformProfile(const std::vector<UsdGeomXformOp>& orderedOps);
```
ppt-adsk: Typo: "matchesMayaTrasformProfile" --> "matchesMayaTransformProfile".

Also awkward that this is not const.

lib/usd/utils/MayaTransformAPI.h
```diff
@@ -0,0 +1,263 @@
...
+  void insertScalePivotOp();  
+  void insertRotateAxisOp();  
+  void insertScalePivotTranslateOp(); 
+  void insertRotatePivotOp();  
+  void insertRotatePivotTranslateOp();
+private:
+  GfVec3f _extractScaleFromMatrix(const UsdTimeCode time) const;
+  GfVec3f _extractRotateFromMatrix(const UsdTimeCode time) const;
+  GfVec3d _extractTranslateFromMatrix(const UsdTimeCode time) const;
+  bool _matchesMayaTrasformProfile(const std::vector<UsdGeomXformOp>& orderedOps);
```
ppt-adsk: Same as line 225.

lib/usd/utils/module.cpp
```diff
@@ -0,0 +1,8 @@
+#include "pxr/base/tf/pyModule.h"
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+TF_WRAP_MODULE
+{
+    TF_WRAP(MayaTransformAPI);
```
ppt-adsk: For my own education, does this ensure that the MayaTransformAPI wrapper ends up in the mayaUsd.lib package?

lib/usd/utils/tests/data/common_api.usda
```diff
@@ -0,0 +1,10 @@
+#usda 1.0
+
+def Xform "pCube1"
+{
+    float3 xformOp:rotateXYZ = (1, 2, 3)
+    float3 xformOp:scale = (4, 5, 6)
+    double3 xformOp:translate = (7, 8, 9)
+    float3 xformOp:translate:pivot = (10, 11, 12)
+    uniform token[] xformOpOrder = ["xformOp:translate", "xformOp:translate:pivot", "xformOp:rotateXYZ", "xformOp:scale", "!invert!xformOp:translate:pivot"]
+}
```
ppt-adsk: Nit-pick: do we care about no newline at end of file?

lib/usd/utils/tests/test_MayaTransformAPI.cpp
```diff
@@ -0,0 +1,1196 @@
...
+        "xformOp:translate:scalePivot",
+        "xformOp:transform:shear",
+        "xformOp:scale", 
+        "xformOp:translate:scalePivotINV"]
+}
+)";
+
+TEST(MayaTransformAPI, orderCheckIsValid)
+{
+  #if AL_SUPPORT_LEGACY_NAMES
```
ppt-adsk: A little confused as to why this is testing the AL_SUPPORT_LEGACY_NAMES.  The tests look very similar to the ones starting at line 153.

lib/usd/utils/tests/test_MayaTransformAPI.cpp
```diff
@@ -0,0 +1,1196 @@
...
+}
+
+const char* getDataPath(const char* const str)
+{
+  static char buffer[512];
+  sprintf(buffer, "%s/%s.usda", AL_EXTRAS_TEST_DATA, str);
+  return buffer;
+}
+
+/*
```
ppt-adsk: Commented out code?

lib/usd/utils/tests/test_MayaTransformAPI.cpp
```diff
@@ -0,0 +1,1196 @@
+#include <gtest/gtest.h>
```
ppt-adsk: Thanks for the nice test.  Did not look at this line by line, just looked at the test intent for each routine, which seems to cover things.  Hopefully a code coverage tool would say the same!

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
```diff
@@ -1854,6 +1854,36 @@ Ufe::Path ProxyShape::ufePath() const
 }
 #endif
 
+//----------------------------------------------------------------------------------------------------------------------
+SdfPathVector ProxyShape::getTranslatedPrimPaths() const
+{
+  SdfPathVector paths(m_requiredPaths.size());
```
ppt-adsk: Isn't this just
SdfPathVector paths;
paths.reserve(m_requiredPaths.size());
std::copy(m_requiredPaths.begin(), m_requiredPaths.end(), std::back_inserter(paths));
?

lib/usd/utils/MayaTransformAPI.cpp
```diff
@@ -0,0 +1,1046 @@
...
+  bool result = _matchesMayaTrasformProfile(orderedOps);
+  if(!result)
+  {
+    m_api = TransformAPI::kFallback;
+  }
+  return result;
+}
+
+//----------------------------------------------------------------------------------------------------------------------
+bool MayaTransformAPI::_matchesMayaTrasformProfile(const std::vector<UsdGeomXformOp>& orderedOps)
```
ppt-adsk: Seems like this could be const.

lib/usd/utils/MayaTransformAPI.cpp
```diff
@@ -0,0 +1,1046 @@
...
+const TfToken scaleOld("xformOp:scale:scale");
+#define XFORM_NAME_CHECK(newName) (name == (newName) || name == (newName ## Old))
+#else
+#define XFORM_NAME_CHECK(newName) (name == (newName))
+#endif
+}
+}
+
+//----------------------------------------------------------------------------------------------------------------------
+bool MayaTransformAPI::matchesMayaTrasformProfile(const std::vector<UsdGeomXformOp>& orderedOps)
```
ppt-adsk: Awkward that this isn't const.  Could m_api be mutable, or does this pose a problem for object invariants and semantics?

lib/usd/utils/MayaTransformAPI.cpp
```diff
@@ -0,0 +1,1046 @@
...
+{
+  bool reset = false;
+  std::vector<UsdGeomXformOp> ops = UsdGeomXformable(prim).GetOrderedXformOps(&reset);
+  if(matchesMayaTrasformProfile(ops))
+  {
+    // only assign if we can process the data
+    m_prim = prim;
+  }
+
+  // if the prim is not valid, check to see if we have a transform Op.
```
ppt-adsk: Why only if the prim is invalid?

lib/usd/utils/MayaTransformAPI.cpp
```diff
@@ -0,0 +1,1046 @@
...
+      m_prim = prim;
+      this->convertMatrixOpToComponentOps(ops[0]);
+    }
+  }
+}
+
+//----------------------------------------------------------------------------------------------------------------------
+void MayaTransformAPI::inheritsTransform(const bool inherit)
+{
+  if(m_prim)
```
ppt-adsk: In some sense you've chosen silent failure for invalid prims (here and elsewhere in your implementation).  Have you thought about raising an exception?

lib/usd/utils/MayaTransformAPI.cpp
```diff
@@ -0,0 +1,1046 @@
...
+//----------------------------------------------------------------------------------------------------------------------
+RotationOrder MayaTransformAPI::rotateOrder() const
+{
+  return m_order;
+}
+
+//----------------------------------------------------------------------------------------------------------------------
+void MayaTransformAPI::translate(const GfVec3d& value, const UsdTimeCode& time)
+{
+  if(m_api != TransformAPI::kFallback && m_prim)
```
ppt-adsk: Again, we'll get silent failure if m_api != TransformAPI::kFallback.

lib/usd/utils/tests/test_MayaTransformAPI.cpp
robthebloke: Yeah, that was the MEL used to generate the unit tests. Thought it made sense to leave it there for someone in the future, but can remove.

cmake/test.cmake
```diff
@@ -129,6 +130,7 @@ finally:
         MAYA_PLUG_IN_PATH
         MAYA_SCRIPT_PATH
         PXR_PLUGINPATH_NAME
+        LD_LIBRARY_PATH
```
kxl-adsk, line 133: Changes from #464 have leaked into this PR. Please remove.

cmake/test.cmake
```diff
@@ -178,6 +180,14 @@ finally:
          "${CMAKE_INSTALL_PREFIX}/plugin/al/lib/usd")
     list(APPEND mayaUsd_varname_PXR_PLUGINPATH_NAME
          "${CMAKE_INSTALL_PREFIX}/plugin/al/plugin")
+    if (IS_LINUX AND CMAKE_SKIP_RPATH)
```
kxl-adsk, line 183: Same here - part of #464 - please remove.

lib/mayaUsd/CMakeLists.txt
```diff
@@ -176,4 +176,5 @@ add_subdirectory(utils)
 
 if(BUILD_HDMAYA)
     add_subdirectory(render/mayaToHydra)
+    add_dependencies(mtoh hdMaya)
```
kxl-adsk, line 179: Why? How is that connected to the new API you are adding?

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.h
```diff
@@ -603,12 +603,27 @@ class ProxyShape
...
   MString getMayaPathFromUsdPrim(const UsdPrim& usdPrim) const;
 
+  /// \brief  A method that attempts to abstract away the setting of the pushToPrim attribute on a transform node. 
+  ///         Once support for parenting maya nodes to UFE objects has been added, refactor this method to provide
+  ///         similar read only support. 
+  /// \param  usdPath path to the prim you want to enable/disable pushToPrim
+  /// \param  pushToPrimState enable/disable pushToPrim
+  /// \return true if the node is valid
+  AL_USDMAYA_PUBLIC
+  bool enablePushToPrim(const SdfPath& usdPath, bool pushToPrimState) const;
```
kxl-adsk, line 625: What's the connection to this PR?

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.h
```diff
@@ -603,12 +603,27 @@ class ProxyShape
   MString recordUsdPrimToMayaPath(const UsdPrim &usdPrim,
                                   const MObject &mayaObject);
 
+  /// \brief  returns the stored maya node path for a prim
+  /// \param  usdPrim a prim that has been brought into maya
+  /// \return  a dag path to the maya object
+  AL_USDMAYA_PUBLIC
+  SdfPathVector getTranslatedPrimPaths() const;
```
kxl-adsk, line 610: What's the connection to this PR?

plugin/al/plugin/AL_USDMayaTestPlugin/plugin.cpp
```diff
@@ -24,17 +24,25 @@
...
 AL_USDMAYATEST_PLUGIN_PUBLIC
 MStatus initializePlugin(MObject obj)
 {
   MFnPlugin plugin(obj, "Animal Logic", "1.0", "Any");
   AL_REGISTER_COMMAND(plugin, UnitTestHarness);
-  return AL::usdmaya::registerPlugin(plugin);
+  auto status = AL::usdmaya::registerPlugin(plugin);
+  // make sure lockPrims are enabled prior to running tests. 
+  // Store the value as a global so it can be restored when the plugin unloads. 
+  g_AL_usdmaya_ignoreLockPrims = MGlobal::optionVarIntValue("AL_usdmaya_ignoreLockPrims");
```
kxl-adsk, line 37: Again, please explain why it's part of this PR

cmake/test.cmake
robthebloke: I can't.

robthebloke: I can't.

lib/mayaUsd/CMakeLists.txt
robthebloke: It isn't. It's connected to the api you've added, and a fix I've had to add to fix it?

robthebloke: It isn't. It's connected to the api you've added, and a fix I've had to add to fix it?

cmake/test.cmake
robthebloke: You're asking me to submit a PR with it being unable to pass our internal integration tests. At this point, I'd rather submit a PR that removes mayaUsd/ufe from maya-usd so that I can actually do some fucking work.

cmake/test.cmake
HamedSabri-adsk: @robthebloke I looked at all the cmake changes in this PR and they make sense to me but as it was mentioned before, there was an agreement around handling LD_LIBRARY_PATH inside AL's code base rather than test.cmake

Please see Eoin's comment: 
https://github.com/Autodesk/maya-usd/pull/464#issuecomment-620939015

lib/mayaUsd/CMakeLists.txt
HamedSabri-adsk: @robthebloke  `add_dependencies(mtoh hdMaya)` is not really needed. Once you sync your branch with latest dev, you should get this change automatically:

https://github.com/Autodesk/maya-usd/commit/5eab9f4cc57f9563e07f176624bd8dba421e709c

robthebloke:
I give up. Let's fork as a way out of the impasse.

PR #161, opened 2019-12-06
make USD_VERSION_NUM available project-wide

mattyjams:
I'm adding some new conditionals in hdMaya that use USD_VERSION_NUM and found that it was only being added as a compiler definition for hdMaya and AL_USDMaya. One file I changed was being included by mayaToHydra which is built through a different CMakeLists.txt that did not add the compiler definition, which caused that library to build incorrectly. Instead of adding the definition to mayaToHydra as well, it made more sense to just apply the compiler definition across the whole project.

The FindUSD.cmake module is what computes the value and sets USD_VERSION_NUM as a CMake variable. The top-level CMakeLists.txt includes CXXDefaults.cmake after it calls `find_package(USD REQUIRED)`, so the CMake variable should be defined by the time we want to add a compiler definition from it in CXXDefaults.cmake.


HamedSabri-adsk:
Hi Matt, lgtm but I dont see USD_VERSION_NUM being set in FindUSD.cmake.

HamedSabri-adsk:
Never mind. I do see it. Reviewing code on phone is not a good idea.🙂

kxl-adsk:
@mattyjams can you make this change in `dev` branch? Then we can merge it downstream to `refactoring_sandbox`.

mattyjams:
@kxl-adsk: No problem. I just rebased the change to dev.

Note though that this means we've lost the change to `lib/usd/hdMaya/CMakeLists.txt`, since hdMaya is only present in refactoring_sandbox. We'll want to remove `USD_VERSION_NUM` from that file when this gets merged to refactoring_sandbox.

kxl-adsk:
Roger that. I will make the change and put you on the review.

Issue #69, opened 2019-10-25
refactoring_sandbox : VP2 Delegate : Duplicating ProxyShape Crash


cfmoore007:
**Issue :**

Maya crashes when duplicating a proxyShape node when the VP2_RENDER_DELEGATE_PROXY is enabled.

**Steps to re-produce :**

1) open maya ( UFE does not need to be set, just the envar for VP2_RENDER_DELEGATE_PROXY )

**2) import an AL proxyShape :**

```
import maya.cmds as cmds
cmds.loadPlugin ("AL_USDMayaPlugin")

# import proxyShape
cmds.AL_usdmaya_ProxyShapeImport(file="3_Spheres.usda")

# select and duplicate
cmds.select("AL_usdmaya_Proxy",r=True)
cmds.duplicate()
```

( USD Debug Information )
```
ProxyShape::ProxyShape
ProxyShape::postConstructor
ProxyShapeSelection::onSelectionChanged 0
ProxyShapeSelection::onSelectionChanged 0
ProxyShape::getUsdStage
ProxyShape::compute AL_usdmaya_Proxy1Shape.outStageData
```

**3) Crash**

Stack Trace :
```
------------------------ 'maya.bin' is dying ------------------------
maya.bin crashed. FATAL ERROR: Used null prim
in Usd_IssueFatalPrimAccessError at line 243 of /pxr/usd/lib/usd/primData.cpp

The stack can be found in /usr/tmp/st_maya.bin.32238
done.
------------------------------------------------------------------

```

**Testing Environment :**
- (Linux) Maya 2019.2 
- refactoring_sandbox ( d0bb624 ) and Pixar USD dev ( b85ddac2 )

- ( Windows ) beta ( latest PR )
- refactoring_sandbox ( w/ PR#69 ) and Pixar USD-rc3 

( I can't seem to upload any files here, but the test was with a sphere usd file )

Side note : the crash did not occur w/ the Autodesk proxy shape

cfmoore007:
[3_Spheres.zip](https://github.com/Autodesk/maya-usd/files/3773888/3_Spheres.zip)
Added file.

fowlertADSK:
Thanks for reporting this.  I am able to reproduce it and we're investigating now.

fowlertADSK:
Internally we're tracking this as MAYA-101680.

cfmoore007:
[...]

PR #162, opened 2019-12-06
Copy AL proxy shape internal data on duplication.

ppt-adsk:


lib/ufe/ProxyShapeHandler.cpp
```diff
@@ -61,12 +61,8 @@ std::vector<std::string> ProxyShapeHandler::getAllNames()
 /*static*/
 UsdStageWeakPtr ProxyShapeHandler::dagPathToStage(const std::string& dagPath)
 {
-    UsdStageWeakPtr stagePtr;
-
-    if (UsdPrim usdPrim = UsdMayaQuery::GetPrim(dagPath))
-        stagePtr = usdPrim.GetStage();
-
-    return stagePtr;
+	auto prim = UsdMayaQuery::GetPrim(dagPath);
```
ppt-adsk: No change in behavior, was just in my commit, so I kept it.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.h
```diff
@@ -1005,6 +1005,7 @@ class ProxyShape
   MStatus preEvaluation(const MDGContext & context, const MEvaluationNode& evaluationNode) override;
   void CacheEmptyBoundingBox(MBoundingBox&) override;
   UsdTimeCode GetOutputTime(MDataBlock) const override;
+  void copyInternalData(MPxNode* srcNode) override;
```
ppt-adsk: Missing call to hook for duplicating Maya node private data.

kxl-adsk:
@cfmoore007 this is a complete fix for #69 , i.e. including a proper way to duplicate proxy shape private data.

cfmoore007:
@kxl-adsk  - excellent, will try it out now, and update the other issue accordingly.

cfmoore007:
1) Confirmed, this works ( compiled this PR against latest dev )  :

duplicateNoCrash

No crash.

2) Edits to the original stage also are preserved when duplicated :

duplicateEditsPreserved

3) However, there's a bug, where if you move the sphere ( I will log this separately ) :
duplicateMovesSameTime

ppt-adsk:
Unit test added.  Without the fix, test fails, with the fix, test passes.

Issue #135, opened 2019-11-26
refactoring_sandbox : mtoh : no material support in hdMaya with USD 20.2


kxl-adsk:
Found when merging dev branch to refactoring_sandbox which brings USD 20.2 (dev branch).

PR #169, opened 2019-12-11
First pass of hdMaya fixes to support core USD commit 62f65d0 (>19.11)

mattyjams:
This PR is a first step towards addressing #135. It adapts hdMaya to build against recent changes in core USD, in particular https://github.com/PixarAnimationStudios/USD/commit/12ed52f5fcb8b308d92731da479ab116df50c582 which removed a handful of Hydra API methods.

Note that while this change allows maya-usd to build against that core USD commit and also gets unconnected materials working again in mayaToHydra, materials connected to file texture nodes will still not display the texture in the viewport. There have been significant changes in Hydra and Storm, and further work is needed in hdMaya to get textures working again.


kxl-adsk:
Great job @mattyjams and thank you @elrond79 for the review.

I wanted to merge it, but it looks like it does not compile on Windows. In delegateDebugCodes.h you used preprocessor directive in an invocation of a macro which results in undefined behavior.

lib/usd/hdMaya/adapters/materialAdapter.cpp
```diff
@@ -243,6 +198,37 @@ const std::string& HdMayaMaterialAdapter::GetPreviewDisplacementSource() {
     return _PreviewShader().displacementCode;
 }
 
+// Specialized version of :
+// https://en.cppreference.com/w/cpp/algorithm/lower_bound
```
robthebloke: What's been specialised about it? _(A comment explaining the changes to the algorithm might be useful in future!)_

lib/usd/hdMaya/adapters/materialAdapter.cpp
```diff
@@ -243,6 +198,37 @@ const std::string& HdMayaMaterialAdapter::GetPreviewDisplacementSource() {
...
+            first = ++it;
+            count -= step + 1;
+        } else {
+            count = step;
+        }
+    }
+    return first != previewShaderParams.cend()
+               ? (first->param.name == id ? first
+                                               : previewShaderParams.cend())
+               : first;
```
robthebloke: Can't this just be done with std::lower_bound? or am I missing something? :)
```c++
auto first = previewShaderParams.cbegin();
auto end = previewShaderParams.cend();
auto cmp = [] (const HdMayaShaderParam& a, const TfToken& b) { return a.param.name < b; };
it = std::lower_bound(first, end, id, cmp);
return (it != end) ? (it->param.name == id ? it : end) : end;
```

lib/usd/hdMaya/adapters/materialAdapter.cpp
```diff
@@ -666,6 +659,10 @@ class HdMayaShadingEngineAdapter : public HdMayaMaterialAdapter {
 
         HdMaterialNetworkMap materialNetworkMap;
         materialNetworkMap.map[HdMaterialTerminalTokens->surface] = materialNetwork;
+#if USD_VERSION_NUM >= 1911
+        materialNetworkMap.terminals.push_back(
+            materialNetwork.nodes.back().path);
```
robthebloke: Do we know that ```materialNetwork.nodes``` will never be empty here?

lib/usd/hdMaya/adapters/materialAdapter.cpp
mattyjams: Good question! I'm actually not terribly familiar with this code. I haven't modified this function at all with this change. I just moved it from higher up in the file to down here closer to its first use so that it was inside the `#if USD_VERSION_NUM <= 1911` conditional. Hopefully it can just go away in the not too distant future if/when we drop support for older versions of USD?

There's more work to be done in this library to get textures working again, so we should revisit this as part of that work.

lib/usd/hdMaya/adapters/materialAdapter.cpp
mattyjams: I think as it's currently written, if `GetMaterial()` above returns success, we *must* have added nodes to the network. It's a good idea to be a little paranoid though in case that changes at some point. Good catch!

mattyjams: Addressed this in 8654c4a.

mattyjams:
> Great job @mattyjams and thank you @elrond79 for the review.
> 
> I wanted to merge it, but it looks like it does not compile on Windows. In delegateDebugCodes.h you used preprocessor directive in an invocation of a macro which results in undefined behavior.

Ahh, sorry about that @kxl-adsk. I don't have easy access to a Windows box, so I hadn't tried building on that platform.

I just pushed c7d096d that should hopefully address the Windows build breakage. Let me know if that looks good on your end now. Thanks!

fowlertADSK:
I just synced to your branch @mattyjams and was able to build on Windows.  Is that confirmation all we were waiting for here?

kxl-adsk:
Looks good, thank you.

PR #170, opened 2019-12-11
fix an error on loading plugin, that mayaUsd_ShaderFragments is a folder

elrond79:
happens because the plugin is marked as a library, instead of a resource,
and specifies a LibraryPath (which it tries to load)

### DIFF
--- a/lib/render/vp2ShaderFragments/plugInfo.json
+++ b/lib/render/vp2ShaderFragments/plugInfo.json
@@ -3,10 +3,9 @@
         {
             "Name": "mayaUsd_ShaderFragments", 
             "Info": {}, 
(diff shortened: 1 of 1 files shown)


HdC-adsk:
LGTM.

@elrond79 Thanks for fixing this error. Just for my own education, how do you figure out the fix? We found no detailed doc about how to write the json file properly.

elrond79:
Well, in this case I'd actually already seen + fixed a similar problem before.  However, the first time I ran across it... I think I was just inferring from context, reading the docs/comments in pxr/base/plugin.h, and comparing to existing plugInfo.json files in the codebase.

HdC-adsk:
Thanks for the info. Happy to see you fixed it.

PR #175, opened 2019-12-12
unset the executable bit for AL CMakeLists.txt and code files (no code changes)

mattyjams:
This has been causing some minor headache for us as we shuttle files between version controls systems, so I thought I'd clean it up.

I can't think of a reason why any of these files should be executable, but let me know if I've got that wrong.


kxl-adsk:
I just discovered that https://github.com/Autodesk/maya-usd/pull/179/files broke Windows build. I can't validate & merge this PR before we get the build fixed.

kxl-adsk:
Quick update - it was only my windows build env affected - not sure where jinja was coming from before, but after installing it, all works again. False alarm.

Issue #127, opened 2019-11-21
Maya crashes in AL's plugin profiling code when opening file


pilarmolinalopez:
The following assert consistently fails when opening one of our scene files:
```
void Profiler::pushTime(const AL::usdmaya::ProfilerSectionTag* entry) {
      assert(MAX_TIMESTAMP_STACK_SIZE > m_stackPos);
```
This function is in AL's profiling code in the AL/usdmaya/CodeTimings.cpp file. MAX_TIMESTAMP_STACK_SIZE is equal to 16. The timestamp stack in AL's profiling code keeps overflowing.  Profiler::pushTime() is being called from ProxyShape::loadStage(). The scene causing the crash has multiple rigged references and proxyShapes. We have not been able to isolate the issue into a more simple scene yet.

This doesn't happen when we disable the UFE build.
**Disabling the UFE part of the build fixes the issue (-DCMAKE_WANT_UFE_BUILD=0)**

A few more notes about our environment:
- environment - recent dev branch or refactoring_sandbox
- AL plugin compiled with CMAKE_WANT_UFE_BUILD=1
- environment variable MAYA_WANT_UFE_SELECTION being set is irrelevant ( occurs with it enabled and also disabled )

kxl-adsk:
Overflow can cause many issues in random places... @murphyeoin - should this stack size be increased? What makes the number of pushed start times go so high? (is it safe to run in parallel?)

pilarmolinalopez:
Forgot to say, I increased the size up to 50 and it would still crash.
What I don't understand yet is why this only happens when UFE build is enabled. @kxl-adsk - Do you have any clues about that?

kxl-adsk:
Does it mean that without UFE we are not overflowing this stack?

pilarmolinalopez:
Correct. As I said in the description, if we compile with CMAKE_WANT_UFE_BUILD=0 the stack does not get overflowed. It overflows when we compile with CMAKE_WANT_UFE_BUILD=1. This happens regardless MAYA_WANT_UFE_SELECTION being set or not.

kxl-adsk:
Thank you for clarification. We will need a way to reproduce it...and in the meantime, can you paste the full callstack when MAX_TIMESTAMP_STACK_SIZE assert is hit? (I expect to see some recursive calls in there)

pilarmolinalopez:
[...]

PR #178, opened 2019-12-12
make sure stage is set in datablock in ProxyShape::LoadStage

pilarmolinalopez:
This PR fixes issue #127 

It just makes sure that the stage is set in the data block (in ProxyShape::loadStage()) before sending any notifications or causing side effects.

Inside ProxyShape::loadStage() this call `m_stage->SetEditTarget(m_stage->GetSessionLayer());` causes other events to be triggered. If the stage data is not set in the datablock before this call we experience the error reported in issue #127

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
@@ -1484,7 +1484,7 @@ void ProxyShape::loadStage()
           MayaUsdStageData* usdStageData = createData<MayaUsdStageData>(MayaUsdStageData::mayaTypeId, data);
           usdStageData->stage = m_stage;
           usdStageData->primPath = m_path;
-          MStatus status = outputDataValue(dataBlock, outStageData(), usdStageData);
+          outputDataValue(dataBlock, outStageData(), usdStageData);
           
           // Set the edit target to the session layer so any user interaction will wind up there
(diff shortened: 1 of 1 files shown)


kxl-adsk:
@robthebloke can you help us in reviewing this change?

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
```diff
@@ -1479,6 +1479,13 @@ void ProxyShape::loadStage()
           stageId = StageCache::Get().Insert(m_stage);
           outputInt32Value(dataBlock, m_stageCacheId, stageId.ToLongInt());
 
+          // Set the stage in datablock so it's ready in case it needs to be accessed
+          MObject data;
+          MayaUsdStageData* usdStageData = createData<MayaUsdStageData>(MayaUsdStageData::mayaTypeId, data);
+          usdStageData->stage = m_stage;
+          usdStageData->primPath = m_path;
+          MStatus status = outputDataValue(dataBlock, outStageData(), usdStageData);
```
kxl-adsk: /maya-usd/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp:1465:19: error: unused variable 'status' [-Werror,-Wunused-variable]
          MStatus status = outputDataValue(dataBlock, outStageData(), usdStageData);

pilarmolinalopez: that should be fixed now. thanks!

PR #179, opened 2019-12-13
use the FindPython CMake module instead of FindPythonInterp

mattyjams:
The FindPythonInterp module became deprecated in CMake 3.12. With that now being the minimum required version of CMake for maya-usd, we should be using the supported FindPython module instead.

A redundant find call was also removed from plugin/al/CMakeLists.txt.


CMakeLists.txt
```diff
@@ -50,11 +50,11 @@ if (APPLE)
    set(CMAKE_OSX_ARCHITECTURES "x86_64")
 endif()
 
-if (DEFINED PYTHON_INCLUDE_DIR AND DEFINED PYTHON_LIBRARIES AND DEFINED PYTHON_EXECUTABLE)
+if (DEFINED PYTHON_INCLUDE_DIR AND DEFINED PYTHON_LIBRARIES AND DEFINED Python_EXECUTABLE)
     SET(PYTHON_INCLUDE_DIRS "${PYTHON_INCLUDE_DIR}")
     SET(PYTHONLIBS_FOUND TRUE)
-    find_package(PythonInterp 2.7 REQUIRED)
-    if(NOT PYTHONINTERP_FOUND)
+    find_package(Python 2.7 EXACT REQUIRED COMPONENTS Interpreter)
```
mattyjams: I opted for `Python 2.7 EXACT` here to make it explicit that Python 3 is not (yet) supported. This should accept any patch version of 2.7.X though. But I used `Python` rather than `Python2` in hopes of making it easier for when we do want to transition to Python 3. The `Python_EXECUTABLE` and `Python_VERSION` usages shouldn't need to change again.

PR #181, opened 2019-12-16
add LD_LIBRARY_PATH entries for proper linux support

pilarmolinalopez:
The current .mod file contains various entries for /lib.  However, it seems it sets the PATH environment variable (used by Windows), and not LD_LIBRARY_PATH (which is needed by linux)

### DIFF
--- a/mayaUSD.mod.template
+++ b/mayaUSD.mod.template
@@ -2,10 +2,12 @@
 PYTHONPATH+:=lib/python
 PATH+:=bin
 PATH+:=lib
+LD_LIBRARY_PATH+:=lib
 
 + PXR_USDMAYA ${USD_VERSION} ${CMAKE_INSTALL_PREFIX}/plugin/pxr
 PYTHONPATH+:=lib/python
 PATH+:=maya/lib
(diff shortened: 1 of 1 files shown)


HamedSabri-adsk:
Hi @pilarmolinalopez,  

We really shouldn't need to set any LD_LIBRARY_PATH.  rpaths are handled automatically for both MacOSX and Linux in the cmake.  

We handle all the rpath gymnastics by using these 3 routines: mayaUsd_init_rpath, mayaUsd_add_rpath, mayaUsd_install_rpath 

On MaxOSX, there is one issue related to rpath + LibGlew that I am aware of which I need to fix. Other than that, I haven't seen any recent issues with rpath on either Linux or MacOSX.

HamedSabri-adsk:
Hi @pilarmolinalopez  can we close this issue?

Thanks.

cfmoore007:
The reason we deployed this fix was actually for the issue raised, here :  https://github.com/Autodesk/maya-usd/pull/192.
We'll close this out.
Thanks!

PR #185, opened 2019-12-20
USDGeomScope support in  Proxy Shape 

murphyeoin:
## Description

Currently any Prim of type USDGeomScope will get translated to an AL_USDMaya_Transform on any relevant Proxy Shape related operation e.g;
+ AL_usdmaya_ProxyShapeImportAllTransforms
+ AL_usdmaya_TranslatePrim
+ transiently on selection
etc.

We want to avail of 2 characteristics of Scopes when translating to maya:
1. Their maya transforms are locked 
2. They cannot be animated and therefore don't need a time input

We now translate those to a new MPxTransform derived node called "Scope"

What was done?
- created Scope MPxNode
- created BasicTransformationMatrix which is used by the Scope MPxNode
- have the original Transform and TransformationMatrix inherit from them as they are relatively simple and lightweight
- updated any code which makes assumptions about AL_usdmaya_transform being the only transform type (most of this is ProxyShape/Selection related) and ideally use the base Scope class pointer
- update code to create Scope nodes where appropriate (i.e mostly when translating from USDGeomScope nodes)
- refactored the transform node creation in ProxyShapeSelection - there were 2 methods performing more or less the same node creation
- add/update some tests etc

Doing this work made me remember we have a weird hack in AL_USDMaya which allows creation of arbitrary transform nodes. This works by tagging an Xform in USD with the
"al_usdmaya_transformType" metadatum... It allows you to create custom transform types without needing to add a proper factory/plugin system or create an explicit dependency on the type - basically a poor man's factory. How this stuff interacts with AL_usdmaya transforms and scopes is anyone's guess really. At least the code is mostly in one place now

## Future work

+ add  standard import/export functionality for Scopes - wasn't needed by us, but should be done to make the feature complete
+ as mentioned above the effort could be made to refactor the code to make transforms a plugin - currently the code for supporting AL_USDMaya_Transform, AL_USDMaya_Scope, and user-defined types is smeared a bit all over the codebase - although this refactor has made it a bit more general (most code should work with any type that derives from Scope). 
+ Consider having an interface for the custom Transform types (currently Scope, Transform) and matrices (BasicTransformationMatrix,  TransformationMatrix) rather than having one inherit from the other


kxl-adsk:
@pilarmolinalopez can you help review this change?

fowlertADSK:
Hey Eoin.  I started writing up some comments/questions as part of a review and I think the direction most of my thoughts were going were towards things like how general the transform classes are and if some of this might belong in a common area instead of just being in the AL plugin (similar to the discussion in PR #158).  I’ve been slowed down by a bunch of other stuff, but maybe you can answer that question before I spend too much time on it.  Do we think that there is functionality here that would be nice to have down in the core lib next to or as part of the proxy base class that is shared with the other plugins?  

I think the answer is “Yes, eventually”, but right now it’s only the AL plugin that would use it.

murphyeoin:
Hi Tim, in this case the functionality is very AL Proxy Shape live interaction workflow specific - there's not even a general maya->USD Scope importer/exporter..

kxl-adsk:
Totally makes sense, Eoin.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
```diff
@@ -1637,7 +1637,9 @@ bool ProxyShape::lockTransformAttribute(const SdfPath& path, const bool lock)
   r.setLocked(lock);
   s.setLocked(lock);
 
-  if (lock && MFnDependencyNode(lockObject).typeId() == AL_USDMAYA_TRANSFORM)
+  MFnDependencyNode fn(lockObject);
+  Scope* transformNode = static_cast<Scope*>(fn.userNode());
```
fowlertADSK: Should this one and the one on line 2144 be dynamic_cast?

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Scope.cpp
```diff
@@ -0,0 +1,296 @@
...
+      AL_MAYA_CHECK_ERROR(attributeAffects(inAttr, shear), errorString);
+      AL_MAYA_CHECK_ERROR(attributeAffects(inAttr, rotatePivot), errorString);
+      AL_MAYA_CHECK_ERROR(attributeAffects(inAttr, rotatePivotTranslate), errorString);
+      AL_MAYA_CHECK_ERROR(attributeAffects(inAttr, scalePivot), errorString);
+      AL_MAYA_CHECK_ERROR(attributeAffects(inAttr, scalePivotTranslate), errorString);
+      // Maya 2018 (checked 2018.2 and 2018.3) has a bug where, if any loaded plugin has an MPxTransform subclass
+      // that has ANY attribute that connected to rotateAxis, it will cause the rotateAxis to evaluate INCORRECTLY,
+      // even on the BASE transform class!
+      // See this gist for full reproduction details:
+      //   https://gist.github.com/elrond79/f9ddb277da3eab2948d27ddb1f84aba0
```
fowlertADSK: Do you happen to know if we have an internal bug for this?  If not we should get one logged.  It does sound familiar.

fowlertADSK:
Appreciate the added tests for the new class, have you only been able to run them on Linux?  If not, I'll build it and run them locally tomorrow morning.

murphyeoin:
Hi Tim, yes have only been able to run on Linux. This code has been in production for a while though, so I think it's mostly error free

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Scope.cpp
murphyeoin: Not sure...(If that was addressed to me)

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Transform.cpp
```diff
@@ -105,10 +104,11 @@ MStatus Transform::initialise()
   try
   {
     setNodeType(kTypeName);
+    inheritAttributesFrom("AL_usdmaya_Scope");
 
     addFrame("USD Prim Information");
-    m_primPath = addStringAttr("primPath", "pp", kReadable | kWritable | kStorable | kConnectable | kAffectsWorldSpace, true);
-    m_inStageData = addDataAttr("inStageData", "isd", MayaUsdStageData::mayaTypeId, kWritable | kStorable | kConnectable | kHidden | kAffectsWorldSpace);
+    addInheritedAttr("primPath");
+    addInheritedAttr("inStageData");
```
fowlertADSK: FYI I am spending some time today trying to get it built and see if the tests pass.  This is a line you'll have problems with.  Paul removed this function and replaced it with something slightly different.  See commit 476826f1

fowlertADSK:
I managed to pull your changes into the branch where Hamed and I were working on getting the tests to run/pass on Windows and I think I got a successful run (after changing a couple lines where the functions you were using had been removed).  You'll have to update those but then I'd be ready to close this one.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
murphyeoin: I think you're right -It would be OK to use static_cast if the data had been pre-filtered to only allow Scopes or classes derived from it,  or we were doing some kind of checking earlier on in the method, but as "lockObject" can be any maya object  we should either be doing those checks or using a dynamic_cast to check that it's a valid Scope at run -time... (add some minor performance cost). Will update. Thanks!

murphyeoin:
@fowlertADSK @kxl-adsk addressed comments, merged latest dev

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/TransformationMatrix.h
```diff
@@ -294,6 +294,78 @@ class TransformationMatrix
...
+
+  /// \brief  Applies a local space translation offset to the computed matrix. Useful for positioning objects on a
+  ///         table.
+  /// \param  localTranslateOffset the local space offset to apply to the transform.
+  inline void setLocalTranslationOffset(const MVector& localTranslateOffset)
+    { m_localTranslateOffset = localTranslateOffset; }
+
+  /// \brief  return the prim this transform matrix is attached to
+  /// \return the prim this transform matrix is controlling
+  inline const UsdPrim& prim() const
```
kxl-adsk: /git/usd/maya-usd/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/TransformationMatrix.h:310:25: error: 'prim' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
  inline const UsdPrim& prim() const
                        ^
/git/usd/maya-usd/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/BasicTransformationMatrix.h:73:26: note: overridden virtual function is here
  virtual const UsdPrim& prim() const

kxl-adsk: In addition, there is no value in making the virtual method inlined.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Scope.h
```diff
@@ -0,0 +1,138 @@
...
+  MStatus connectionBroken(const MPlug& plug, const MPlug& otherPlug, bool asSrc) override;
+  bool isBounded() const override
+    { return true; }
+  bool treatAsTransform() const override
+    { return false; }
+
+  //--------------------------------------------------------------------------------------------------------------------
+  /// Data members
+  //--------------------------------------------------------------------------------------------------------------------
+  bool updateTransformInProgress = false;
```
kxl-adsk: /git/usd/maya-usd/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Scope.h:110:8: error: private field 'updateTransformInProgress' is not used [-Werror,-Wunused-private-field]
  bool updateTransformInProgress = false;

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/nodes/test_ScopeMatrix.cpp
```diff
@@ -0,0 +1,171 @@
...
+      MFnDependencyNode fn(obj);
+      MFnTransform xform(obj);
+
+      MString str;
+      AL::usdmaya::nodes::Scope* ptr = (AL::usdmaya::nodes::Scope*)fn.userNode();
+
+      if (ptr->typeId()==AL::usdmaya::nodes::Transform::kTypeId)
+      {
+        //It's a transform
+        AL::usdmaya::nodes::BasicTransformationMatrix* matrix = ptr->transform();
```
kxl-adsk: /git/usd/maya-usd/plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/nodes/test_ScopeMatrix.cpp:130:56: error: unused variable 'matrix' [-Werror,-Wunused-variable]
        AL::usdmaya::nodes::BasicTransformationMatrix* matrix = ptr->transform();

kxl-adsk:
@murphyeoin attaching patch I used to fix compilation errors.
[strict_mode_patch.txt](https://github.com/Autodesk/maya-usd/files/4122920/strict_mode_patch.txt)

murphyeoin:
Thanks @kxl-adsk can you confirm that  running the compiler in strict mode is part of your standard testing process? If so should we just make it part of the build? (perhaps as an additional target? )

kxl-adsk:
Yes, we run the compiler in strict mode on all platforms. You can pass it to `build.py` like this `--build-args=",-DPXR_STRICT_BUILD_MODE=1"`

murphyeoin:
Hi @kxl-adsk I fixed them in our fork and have updated. I will add the strict mode compile to our build script as the aim is that we're all testing the same thing. Any other flags in your build-args that we should know about? Thanks

kxl-adsk:
At the moment, no, there is nothing else.

PR #154, opened 2019-12-04
Autodesk/ppt/refactoring sandbox to dev 00

ppt-adsk:
This is the second large pull request to move code from the Pixar plugin to the mayaUsd core library and translator plugins.  Impact on the AL plugin is minimal.  All Pixar and mayaUsd core tests pass on my CentOS 7 Linux box (albeit with some difficult to reproduce random failures).  All AL tests using the AL test plugin pass when run inside an interactive Maya.  When running the test stage of the build script, the following tests fail, but these are apparently known to fail in the dev branch at this time:
	 74 - AL_MayaUtilsTests (Failed)
	 77 - TestAL_USDMaya (Failed)
	 78 - TestUSDMayaPython (Failed)
	 80 - TestAdditionalTranslators (Failed)
	 81 - TestPxrUsdTranslators (Failed)

PR #191, opened 2020-01-03
modernize and clean up mayaUsd and mayaUsdPlugin targets!

HamedSabri-adsk:


CMakeLists.txt
```diff
@@ -13,10 +12,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# =======================================================================
 #
 
-cmake_minimum_required(VERSION 3.12.0)
+cmake_minimum_required(VERSION 3.13...3.17)
```
HamedSabri-adsk: cmake_policy(VERSION <min>[...<max>]).  This syntax was introduced in CMake V12.4.

I've set the min policy to 3.13, although I think 3.15 is more suitable.

cmake/utils.cmake
```diff
@@ -148,18 +148,37 @@ function(mayaUsd_install_rpath rpathRef NAME)
 endfunction()
 
 function(mayaUsd_promoteMayaUsdHeader)
-    set(srcFile ${CMAKE_CURRENT_SOURCE_DIR}/base/mayaUsd.h.src)
+    set(srcFile ${CMAKE_CURRENT_SOURCE_DIR}/mayaUsd.h.src)
```
HamedSabri-adsk: Get rid off hard-coded `base` directory.

cmake/utils.cmake
```diff
@@ -203,7 +222,7 @@ function(mayaUsd_copyFiles target)
         message(FATAL_ERROR "DESTINATION keyword is not specified.")
     endif()
```
HamedSabri-adsk: mayaUsd_promoteHeaderList function now takes a TARGET, DIRNAME, and HEADERS which are needed to be passed explicitly.

lib/CMakeLists.txt
```diff
@@ -311,134 +132,22 @@ install(TARGETS ${LIBRARY_NAME}
...
-            )
-        endforeach()
-    endif()
-
-endif()
-
-if(IS_MACOSX)
-    set(_macDef OSMac_)
+add_subdirectory(ufe)
 endif()
```
HamedSabri-adsk: Cleaned up the top level CMake. This file was getting really large and was hard to follow up what is happening there.

All the sub-directories (base, listeners, nodes, utils,render,ufe) now have their own CMakeLists.  As of CMakeV12.4, 
The target_* directive calls don't need to be part of the same CMakeLists.txt where add_executable or add_library directives are called.

lib/base/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+# -----------------------------------------------------------------------------
+# sources
+# -----------------------------------------------------------------------------
+target_sources(${TARGET_NAME} 
+    PRIVATE
+        debugCodes.cpp
+)
```
HamedSabri-adsk: Always use target_sources!

cmake/utils.cmake
```diff
@@ -148,18 +148,37 @@ function(mayaUsd_install_rpath rpathRef NAME)
...
         message(STATUS "promoting: " ${srcFile})
     endif()
     configure_file(${srcFile} ${dstFile})
 endfunction()
 
-function(mayaUsd_promoteHeaderList)
-    foreach(header ${ARGV})
+function(mayaUsd_promoteHeaderList target)
+    cmake_parse_arguments(PREFIX
+        "TARGET"
```
seando-adsk: From the doc, the first arg is the options. "target" is already a arg passed to the function. I had already made this change in my local branch like so:
```cmake
    cmake_parse_arguments(PREFIX 
        ""          # options
        "SUBDIR"    # one_value keywords
        "FILES"     # multi_value keywords
        ${ARGN}
    )
```
Why is DIRNAME mandatory? 

I had also fixed the copyDirectory below.

lib/ufe/CMakeLists.txt
```diff
@@ -0,0 +1,207 @@
+# -----------------------------------------------------------------------------
```
seando-adsk: I'm assuming that this block was just copied from the other cmake file and nothing has really changed in it. If something did change can you please point it out?

plugin/adsk/plugin/CMakeLists.txt
```diff
@@ -1,50 +1,99 @@
...
+target_compile_definitions(${TARGET_NAME}
     PRIVATE
         MAYAUSD_PLUGIN_EXPORT
 )
 
-target_include_directories(${PLUGIN_PACKAGE}
+if(UFE_FOUND)
+    target_compile_definitions(${TARGET_NAME} 
+        PRIVATE
+            -DWANT_UFE_BUILD
```
seando-adsk: I'm thinking that rather than have this in a separate if, it could be done using a "generator expression". Something like this just after line 50:

```cmake
$<UFE_FOUND:WANT_UFE_BUILD>
```

Note: you also don't need the -D as it will be removed (as per the docs).

Note: You can also use the same syntax for the target_include_directories below.

HamedSabri-adsk: @seando-adsk  Cool! Thanks for pointing it out. Even better! will change it.

ppt-adsk:
Hey Hamed, are you sure this doesn't conflict with
https://github.com/Autodesk/maya-usd/tree/Autodesk/ppt/refactoring_sandbox_to_dev_00
?  See
https://github.com/Autodesk/maya-usd/pull/154
I'm wondering about lib/CMakeLists.txt, for example.

lib/ufe/CMakeLists.txt
HamedSabri-adsk: Correct. Nothing really changed here, expect the sub-directory prefix that is gone now:

https://github.com/Autodesk/maya-usd/blob/334926c3a6e965fcea048fb76f3980cf50e3f002/lib/CMakeLists.txt#L80

cmake/utils.cmake
HamedSabri-adsk: @seando-adsk 
Right, I didn't really have to pass the target and was solely added for debugging purposes.  Target is not really used inside the function.

Both DIRNAME  and HEADERS keywords are mandatory. DIRNAME is needed because we no longer have the header files with the subdir prefix ( e.g `base/api.h`, `base/debugCodes.h`, etc... ). Instead the subdir name now needs to be passed to mayaUsd_promoteHeaderList function as a separate argument.

```
e.g
set(headers
    api.h
    debugCodes.h
)

I am going to change get rid of the target and rename DIRNAME to SUBDIR. 
e.g
mayaUsd_promoteHeaderList(HEADERS ${headers} SUBDIR base)
mayaUsd_promoteHeaderList(HEADERS ${headers} SUBDIR nodes)
mayaUsd_promoteHeaderList(HEADERS ${headers} SUBDIR listeners)
```

Did you remove target from mayaUsd_copyDirectory? I believe target was needed to be passed to `add_custom_command` command?

seando-adsk: @HamedSabri-adsk Okay I see for both those keywords being mandatory now. No I didn't remove TARGET from any of the functions.

HamedSabri-adsk:
> 
> 
> Hey Hamed, are you sure this doesn't conflict with
> https://github.com/Autodesk/maya-usd/tree/Autodesk/ppt/refactoring_sandbox_to_dev_00
> ? See
> #154
> I'm wondering about lib/CMakeLists.txt, for example.

Hey Pierre, yes indeed. I was aware of this and wanted to discuss with you. 
I was thinking to wait for #154 to be merged into dev first. Then I would rebase my branch in this PR (obviously I need to add more new changes) and then have this branch merge to dev. 

What do you think?

kxl-adsk:
@HamedSabri-adsk there are couple of other PRs which will wait for #154. This one should wait as well.

plugin/adsk/plugin/CMakeLists.txt
```diff
@@ -50,12 +50,10 @@ target_compile_definitions(${TARGET_NAME}
...
 
-if(UFE_FOUND)
-    target_compile_definitions(${TARGET_NAME} 
-        PRIVATE
-            -DWANT_UFE_BUILD
-    )
-endif()
+target_compile_definitions(${TARGET_NAME} 
+    PRIVATE
+        $<$<BOOL:${UFE_FOUND}>:WANT_UFE_BUILD>
```
HamedSabri-adsk: @seando-adsk I used generator expression $<BOOL:...> to convert FALSE or TRUE to 0 or 1

`$<UFE_FOUND:WANT_UFE_BUILD> --> this would cause an error because ${UFE_FOUND} isn't 1 or 0`

`$<$<BOOL:${UFE_FOUND}>:WANT_UFE_BUILD>   --> BOOL:${UFE_FOUND} converts TRUE or FALSE to 1 or 0`

cmake/utils.cmake
```diff
@@ -178,7 +178,7 @@ function(mayaUsd_promoteHeaderList target)
 
     foreach(header ${headerFiles})
         set(srcFile ${CMAKE_CURRENT_SOURCE_DIR}/${header})
-        set(dstFile ${CMAKE_BINARY_DIR}/include/mayaUsd/${dirName}/${header})
+        set(dstFile ${CMAKE_BINARY_DIR}/include/mayaUsd/${subDir}/${header})
 
         set(content "#pragma once\n#include \"${srcFile}\"\n")
```
HamedSabri-adsk: @seando-adsk  target is gone and I renamed DIRNAME to SUBDIR. Thanks.

plugin/adsk/plugin/CMakeLists.txt
seando-adsk: So the whole point was to merge this into the one above at line 48. Just move line 55 to just after 50. Then you don't need two target_compile_definitions. Same for the target_include_directories below.

HamedSabri-adsk: Ugh d’accord ! je suis aveugle...

seando-adsk: Excellent thanks.

HamedSabri-adsk:
Closing this PR. I am internally working on this since there is going to be more new changes coming in. Will create a separate PR once it's ready.

PR #192, opened 2020-01-04
add a test to ensure that the AL.usdmaya python lib links correctly

elrond79:
also added a test to ensure that this module can be imported, even without
loading the plugin or even instantiating maya (mostly just to check that
the linking is set up correctly)

### DIFF
--- a/plugin/al/plugin/AL_USDMayaTestPlugin/py/CMakeLists.txt
+++ b/plugin/al/plugin/AL_USDMayaTestPlugin/py/CMakeLists.txt
@@ -20,7 +20,7 @@ separate_argument_list(pythonPath)
 separate_argument_list(mayaPluginPath)
 
 add_test(
-    NAME ${TEST_NAME}
+    NAME TestUSDMayaPython
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     COMMAND ${MAYA_PY_EXECUTABLE} -c "import sys;from unittest import main;import maya.standalone; \
                                       maya.standalone.initialize(name='python'); \
@@ -35,7 +35,25 @@ add_test(
                                       "
 )
 
-set_property(TEST ${TEST_NAME} APPEND PROPERTY ENVIRONMENT 
+set_property(TEST TestUSDMayaPython APPEND PROPERTY ENVIRONMENT 
+    "PYTHONPATH=${pythonPath}"
+    "PATH=${path}"
+    "MAYA_PLUG_IN_PATH=${mayaPluginPath}"
+    "PXR_PLUGINPATH_NAME=${CMAKE_INSTALL_PREFIX}/lib/usd"
+    "MAYA_NO_STANDALONE_ATEXIT=1"
+)
+
+add_test(
+    NAME TestUSDMayaPythonModuleOnly
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    # Note that, since this test does NOT call maya.standalone.initialize, we can
+    # just let unittest.main call sys.exit "normally"
+    COMMAND ${MAYA_PY_EXECUTABLE} -c "import sys;from unittest import main; \
+                                      import testUSDMayaPythonModuleOnly; testProg = main(module=testUSDMayaPythonModuleOnly); \
+                                      "
+)
+
+set_property(TEST TestUSDMayaPythonModuleOnly APPEND PROPERTY ENVIRONMENT 
     "PYTHONPATH=${pythonPath}"
     "PATH=${path}"
     "MAYA_PLUG_IN_PATH=${mayaPluginPath}"
--- a/plugin/al/plugin/AL_USDMayaTestPlugin/py/CMakeLists.txt
+++ b/plugin/al/plugin/AL_USDMayaTestPlugin/py/CMakeLists.txt
@@ -1,4 +1,4 @@
-set(TEST_NAME TestUSDMayaPython)
+set(MAIN_TEST_NAME TestUSDMayaPython)
 
 set(path 
     "${CMAKE_INSTALL_PREFIX}/lib"
@@ -20,7 +20,7 @@ separate_argument_list(pythonPath)
 separate_argument_list(mayaPluginPath)
 
 add_test(
-    NAME TestUSDMayaPython
+    NAME ${MAIN_TEST_NAME}
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     COMMAND ${MAYA_PY_EXECUTABLE} -c "import sys;from unittest import main;import maya.standalone; \
                                       maya.standalone.initialize(name='python'); \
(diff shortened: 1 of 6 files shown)


HamedSabri-adsk:
@elrond79 looks good but this test is not really portable! There is really no need to have a shell script to setup bunch of env variables. What if one wants to run this test from Windows command line?

For setting env variables for ctests, one can simply use `set_property(TEST ${target} APPEND PROPERTY ENVIRONMENT)`

We have a PR for AL unit tests that is going to come up soon that addresses all these issues.

elrond79:
Oh, cool - which PR / branch is it? I can mimic the style you're using there...

HamedSabri-adsk:
The branch is in our internal repo. We have almost all of AL units tests passing on Windows. We just need to clean up a few things but hopefully by next Tuesday we should have something to show. Would that work?

A good example to mimic the style can be found in the ufe tests:

https://github.com/Autodesk/maya-usd/blob/dev/test/lib/ufe/CMakeLists.txt

Hope this helps!

HamedSabri-adsk:
New style can be found here: https://github.com/Autodesk/maya-usd/pull/198

elrond79:
Given that there's still some issues to be sorted out on how to handle multiple python tests in this test module (https://github.com/Autodesk/maya-usd/pull/198#discussion_r368238806), I'm going to put this on hold until #198 is merged in...

elrond79:
Updated against dev now that #198 has been merged.

Now all this does is add the test (the actual rpath fix was also in #198).

Will update the PR name to reflect...

plugin/al/plugin/AL_USDMayaTestPlugin/py/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ separate_argument_list(pythonPath)
 separate_argument_list(mayaPluginPath)
 
 add_test(
-    NAME ${TEST_NAME}
+    NAME TestUSDMayaPython
```
kxl-adsk: No need to revert this part of #198 where we switched to use the variable `TEST_NAME`.

plugin/al/plugin/AL_USDMayaTestPlugin/py/CMakeLists.txt
```diff
@@ -35,10 +35,28 @@ add_test(
                                       "
 )
 
-set_property(TEST ${TEST_NAME} APPEND PROPERTY ENVIRONMENT 
+set_property(TEST TestUSDMayaPython APPEND PROPERTY ENVIRONMENT
```
kxl-adsk: Same here, keep variable `TEST_NAME`

plugin/al/plugin/AL_USDMayaTestPlugin/py/CMakeLists.txt
```diff
@@ -35,10 +35,28 @@ add_test(
...
 )
+
+add_test(
+    NAME TestUSDMayaPythonModuleOnly
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    # Note that, since this test does NOT call maya.standalone.initialize, we can
+    # just let unittest.main call sys.exit "normally"
+    COMMAND ${MAYA_PY_EXECUTABLE} -c "import sys;from unittest import main; \
+                                      import testUSDMayaPythonModuleOnly; testProg = main(module=testUSDMayaPythonModuleOnly); \
+                                      "
```
kxl-adsk: Make exit with proper status reflecting test results. 
```
sys.exit(not testProg.result.wasSuccessful()); \
```

plugin/al/plugin/AL_USDMayaTestPlugin/py/testUSDMayaPythonModuleOnly.py
```diff
@@ -0,0 +1,48 @@
+'''Test that basic functionality of the python module works without maya.
```
kxl-adsk: Missing license note - see https://github.com/Autodesk/maya-usd/blob/dev/test/lib/ufe/ufeTestPlugins/ufeTestCmdsPlugin.py#L4

plugin/al/plugin/AL_USDMayaTestPlugin/py/CMakeLists.txt
elrond79: I reverted because I now need 2 tests (among other reasons, we can't bundle the new test in because it needs maya.initialize to NOT be called), and I wasn't sure how to handle having two different TEST_NAME variables.  ie, I could do:

```
set(MAIN_TESTS_NAME, TestUSDMayaPython)
set(MODULE_ONLY_TESTS_NAME, TestUSDMayaPythonModuleOnly)

add_test( NAME ${MAIN_TESTS_NAME})
...
add_test( NAME ${MODULE_ONLY_TESTS_NAME})
```

...but at that point, I wasn't sure if we were gaining much by using variables instead of the test names directly.

However, if you want to go that route, I don't have a strong preference.

kxl-adsk: Two different variables sound good to me.

plugin/al/plugin/AL_USDMayaTestPlugin/py/CMakeLists.txt
elrond79: unittest.main automatically does precisely what you wrote.

The only reason we needed to break it up - ie, use `main(exit=False)`, capture the testProg, get it's result and then call `sys.exit` explicitly, was because we needed to call some code inbetween the test finishing, and calling `sys.exit` - namely, `maya.standalone.uninitialize`.

The comment I added, about not calling maya.standalone.initialize, was meant to explain that.  I can expand that a bit, if you'd like, stating more explicitly what I've just explained.

Or, if you still prefer making the boilerplate be as similar as possible, I can add in the sys.exit / testProg / exit=False stuff, though it's not strictly necessary.

plugin/al/plugin/AL_USDMayaTestPlugin/py/CMakeLists.txt
elrond79: K, will do.

plugin/al/plugin/AL_USDMayaTestPlugin/py/testUSDMayaPythonModuleOnly.py
elrond79: Will do...

plugin/al/plugin/AL_USDMayaTestPlugin/py/CMakeLists.txt
elrond79: Oops - I guess, if we're going the simpler approach, I should also change

```
testProg = main(...)
```

to be just

```
main(...)
```

kxl-adsk: Didn't notice that one! Making it main(...) sounds like the right thing to do.

plugin/al/plugin/AL_USDMayaTestPlugin/py/testUSDMayaPythonModuleOnly.py
elrond79: Done - and also added the copyright to other py tests in the dir!

elrond79:
Think all the requested changes have been addressed...

kxl-adsk:
@elrond79 For unknown reason I don't see this new test on MacOS. Do you have MacOS to validate?

PR #193, opened 2020-01-04
[pxr] when loading the pixar plugin, trigger an immediate fail

elrond79:
...if we find an operation failed, instead of printing a warning and continuing

I originally added this because to catch an error on an internal luma branch, but it has also exposed an error in the AL test TestPxrUsdTranslators (fixed in https://github.com/Autodesk/maya-usd/commit/2c78161eac9f5bd4baabb3f8abff6564f9ec0400)


kxl-adsk:
@elrond79 FYI. We will wait for #154 to be merged first to avoid conflicts that would force us to revalidate & retest much bigger PR.

kxl-adsk:
@elrond79 Please resolve conflicts.

elrond79:
Done...

PR #194, opened 2020-01-06
[al] set the rpath for testMayaSchemas bin on MacOS / Linux

elrond79:
...this allows the test to complete successfully.

Also, this change makes sure that RUNPATH is used instead of RPATH for
all shared libs and executables on Linux; RUNPATH is preferrable because
it is overrideable by LD_LIBRARY_PATH.

Note that this was effectively already always used for all AL shared libs,
due to the use of PARENT_SCOPE in al/translators/CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,11 @@ if(NOT WIN32)
 endif()
 string(REPLACE ";" " " CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 
+if(IS_LINUX)
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--enable-new-dtags")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--enable-new-dtags")
+endif()
+
 
 #==============================================================================
 # Tests
--- a/plugin/al/schemas/AL/usd/schemas/maya/tests/CMakeLists.txt
+++ b/plugin/al/schemas/AL/usd/schemas/maya/tests/CMakeLists.txt
@@ -28,6 +28,12 @@ target_link_libraries(testMayaSchemas
     ${Boost_PYTHON_LIBRARY}
     )
 
+if(IS_MACOSX OR IS_LINUX)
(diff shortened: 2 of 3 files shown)


HamedSabri-adsk:
@elrond79 As we discussed in #192, we already have these fixes in an internal branch. There are a lot more fixes that we have done and I am planning to bring these changes to the public repo sometimes tomorrow morning.

elrond79:
Ah, the implication I picked up in #192 was that those changes were mostly windows-related. I didn't realize you'd also made Linux / rpath fixes in your private branch as well.
Looking forward to seeing all your test fixes!

TrevorThomson:
@HamedSabri-adsk, I just wanted to check: are you saying your fixes in an internal branch going to switch to using RUNPATH rather than RPATH?
I'm asking 'cause we build a variant of USD for each Renderman version in use (to support prman delegates) and then, due to the usage of RPATH, we'll have to build variants of maya_usd for each USD/Renderman combination.  If we could use LD_LIBRARY_PATH at runtime, however, we could reduce the number of variants...
Thanks!

HamedSabri-adsk:
Hi @TrevorThomson, Yes correct. As @elrond79 also mentioned above, we prefer to use RUNPATH over RPATH.

If RPATH is used, then we cannot override it during execution since LD_LIBRARY_PATH has lower precedence. The order of precedence for search paths on Linux are:

1- RPATH
2- LD_LIBRARY_PATH
3- RUNPATH

TrevorThomson:
@HamedSabri-adsk, that sounds good--thanks!

elrond79:
@TrevorThomson - Just wanted to quickly add that you don't need to wait for a PR - just add this to your cmake command, if you're invoking it directly:

```
-DCMAKE_EXE_LINKER_FLAGS=-Wl,--enable-new-dtags -DCMAKE_SHARED_LINKER_FLAGS=-Wl,--enable-new-dtags
```

If you're using the build.py... I'm not sure who to set that flag.  Normally you'd use "--build-flags=", but I think that handles "," character specially, and I'm not sure you can escape it...

HamedSabri-adsk:
This PR can be closed. Please see https://github.com/Autodesk/maya-usd/pull/198

elrond79:
Confirmed that these changes are all in #198... thanks!

PR #196, opened 2020-01-10
MAYA-102389 - MayaUsd.ufe python bindings don't work in Maya Debug (Win)

seando-adsk:
#### MAYA-102389 - MayaUsd.ufe python bindings don't work with Maya Debug on Windows

* Full python debug support.
* Fix python cmake variable names (that come from find_package).
* Must build boost with python debugging flag.
* On Windows when using debug python the libraries must have _d in name.


build.py
```diff
@@ -341,6 +291,13 @@ def BuildAndInstall(context, buildArgs, stages):
             extraArgs.append('-DMAYA_DEVKIT_LOCATION="{devkitLocation}"'
                              .format(devkitLocation=context.devkitLocation))
 
+        # Many people on Windows may not have python with the 
+        # debugging symbol (python27_d.lib) installed, this is the common case.
+        if context.buildDebug and context.debugPython:
+            extraArgs.append('-DMAYAUSD_DEFINE_BOOST_DEBUG_PYTHON_FLAG=ON')
+        else:
+            extraArgs.append('-DMAYAUSD_DEFINE_BOOST_DEBUG_PYTHON_FLAG=OFF')
+
```
seando-adsk: New cmake flag to use when we compiling with a debug python.

build.py
```diff
@@ -402,13 +359,16 @@ def RunTests(context,extraArgs):
...
-                    help="Build in Release mode")
+                    help="Build in Release mode (default: %(default)s)")
 
-parser.add_argument("--build-relwithdebug", dest="build_relwithdebug", action="store_true",
-                    help="Build in RelWithDebInfo mode")
+parser.add_argument("--build-relwithdebug", dest="build_relwithdebug", action="store_true", default=True,
+                    help="Build in RelWithDebInfo mode (default: %(default)s)")
+
+parser.add_argument("--debug-python", dest="debug_python", action="store_true",
+                      help="Define Boost Python Debug if your Python library comes with Debugging symbols (default: %(default)s).")
```
seando-adsk: New build flag which means you are using a debug build of python (which we do internally for our Maya debug build). You would point to that python by setting the defines: 
PYTHON_INCLUDE_DIR, Python_EXECUTABLE, PYTHON_LIBRARIES
in your `--build-args` flag.

seando-adsk: Note: it also requires that you have that debug python in your path and set the PYTHONPATH for it.

cmake/defaults/CXXDefaults.cmake
```diff
@@ -71,4 +71,4 @@ endif()
 # https://docs.python.org/2/c-api/intro.html#include-files
 #
 # The long term plan is to adhere to the required behavior.
-include_directories(SYSTEM ${PYTHON_INCLUDE_DIR})
+include_directories(SYSTEM ${PYTHON_INCLUDE_DIRS})
```
seando-adsk: To go along with other changes that Matt made, the proper variable that FindPackage(Python) sets is DIR**S**

cmake/defaults/msvcdefaults.cmake
```diff
@@ -78,6 +78,16 @@ if (NOT Boost_USE_STATIC_LIBS)
...
+if(${MAYAUSD_DEFINE_BOOST_DEBUG_PYTHON_FLAG})
+    # We should not have to add this define manually as it is added by the
+    # compiler when you specify the /MTd or /MDd option. But it is not - needs
+    # to be investigated.
+    _add_define(_DEBUG)
+
+    _add_define(BOOST_DEBUG_PYTHON)
+    _add_define(BOOST_LINKING_PYTHON)
+endif()
+
```
seando-adsk: We need all these extra defines with building with a debug python. I still need to figure out why _DEBUG is not being added automatically as it should be. It is needed when including the python headers (on Windows) so that the proper #import statement is used.

plugin/pxr/cmake/macros/Private.cmake
```diff
@@ -68,9 +68,13 @@ endfunction() # _copy_headers
...
     # Library names are either something like tf.so for shared libraries
-    # or _tf.so for Python module libraries. We want to strip the leading
-    # "_" off.
-    string(REPLACE "_" "" LIBNAME ${LIBRARY_FILENAME})
+    # or _tf.pyd/_tf_d.pyd for Python module libraries.
+    # We want to strip off the leading "_" and the trailing "_d".
+    set(LIBNAME ${LIBRARY_FILENAME})
+    if(IS_WINDOWS AND ${MAYAUSD_DEFINE_BOOST_DEBUG_PYTHON_FLAG})
+        string(REGEX REPLACE "_d$" "" LIBNAME ${LIBNAME})
+    endif()
```
seando-adsk: This code makes sure to strip off the trailing _d from the library name.

plugin/pxr/cmake/macros/Private.cmake
```diff
@@ -68,9 +68,13 @@ endfunction() # _copy_headers
...
+    string(REGEX REPLACE "^_" "" LIBNAME ${LIBNAME})
```
elrond79: Good change, if the intent of the old code was simply to remove "_" prefix (as the comment claims), this is definitely a better solution...

build.py
```diff
@@ -98,56 +98,6 @@ def IsVisualStudio2017OrGreater():
...
-        # Ensure that the paths are absolute, since depending on the version of
-        # Python being run and the path used to invoke it, we may have gotten a
-        # relative path from distutils.sysconfig.PREFIX.
-        return (
-            os.path.abspath(pythonExecPath),
-            os.path.abspath(pythonLibPath),
-            os.path.abspath(pythonIncludeDir))
-
-    return None
-
```
mattyjams: Thanks for removing this! This is only relevant for the core USD build to make sure it uses Maya's Python, not when building maya-usd:
https://github.com/PixarAnimationStudios/USD/blob/4b1162957b2ad219e1635c81de8343be490e41fc/build_scripts/build_usd.py#L125

build.py
```diff
@@ -402,13 +359,16 @@ def RunTests(context,extraArgs):
...
+                    help="Build in Debug mode (default: %(default)s)")
 
 parser.add_argument("--build-release", dest="build_release", action="store_true",
-                    help="Build in Release mode")
+                    help="Build in Release mode (default: %(default)s)")
 
-parser.add_argument("--build-relwithdebug", dest="build_relwithdebug", action="store_true",
-                    help="Build in RelWithDebInfo mode")
+parser.add_argument("--build-relwithdebug", dest="build_relwithdebug", action="store_true", default=True,
+                    help="Build in RelWithDebInfo mode (default: %(default)s)")
```
mattyjams: Maybe these should be in a mutually_exclusive_group?

cmake/defaults/CXXDefaults.cmake
mattyjams: Ahh, thanks for catching that!

lib/CMakeLists.txt
```diff
@@ -31,6 +31,9 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 if(NOT DEFINED BOOST_ROOT)
     set(BOOST_ROOT ${PXR_USD_LOCATION})
 endif()
+if(WIN32 AND MAYAUSD_DEFINE_BOOST_DEBUG_PYTHON_FLAG)
+    set(Boost_USE_DEBUG_PYTHON ON)
+endif()
```
mattyjams: Sorry, I'm not super-familiar with this, but do we explicitly *not* want to set Boost_USE_DEBUG_PYTHON on Linux or Mac? A similar snippet was added in plugin/al/CMakeLists.txt. Maybe this should be done in the top-level CMakeLists.txt? We don't/can't disable building Python bindings in maya-usd, so I think we know we're going to need to find Boost, right?

elrond79: Alternatively... if we DO only want it on windows, maybe we could have this in the msvcdefaults.cmake?  That way it would all be in one place.  Also - again, not really familiar with the details, but do we ever want to `_add_define(BOOST_DEBUG_PYTHON)` but not `set(Boost_USE_DEBUG_PYTHON ON)`?  If they should always be a pair, makes even more sense to have them grouped.

elrond79:
Well, haven't been able to test... but reading the code over, it looks good!

cmake/defaults/msvcdefaults.cmake
elrond79: I get that _DEBUG is a msvc thing... but are BOOST_DEBUG_PYTHON and BOOST_LINKING_PYTHON as well?

seando-adsk: Ah you are right, they are not. I'll move those two into the CXXDefaults.cmake file.

lib/CMakeLists.txt
seando-adsk: Yeah I guess it shouldn't have the WIN32. The reason this is here is because it is actually a variable used in the FindBoost.cmake. So I wanted it to be set directly above where we are calling find_package(Boost).

plugin/pxr/cmake/macros/Private.cmake
seando-adsk: Yeah I also had to make this change in our USD build as it is the same code.

build.py
seando-adsk: Yes that is a good suggestion. I've done. it.

PR #199, opened 2020-01-14
reintroduce migrated UsdMaya Python bindings back into the pxr.UsdMaya namespace

mattyjams:
Pull request #154 relocates a bunch of Python bindings from pxr.UsdMaya to mayaUsd.lib. To ease the transition for Python code that uses the bindings from the old package namespace, this change reintroduces the migrated names back into the pxr.UsdMaya namespace by importing them from mayaUsd.lib. It is wrapped in a try/except so that it will work both before and after PR #154 is merged.


kxl-adsk:
This makes sense. Thank you, Matt.

PR #202, opened 2020-01-14
Disable deprecated warning on windows (matching gcc and clang)

kxl-adsk:
This allows us now to build maya-usd in strict mode on all platforms.

### DIFF
--- a/cmake/defaults/msvcdefaults.cmake
+++ b/cmake/defaults/msvcdefaults.cmake
@@ -56,6 +56,9 @@ _disable_warning("4180")
 # tbb/enumerable_thread_specific.h
 _disable_warning("4334")
 
+# suppress deprecation warning.
+_disable_warning("4996")
+
 # Disable warning C4996 regarding fopen(), strcpy(), etc.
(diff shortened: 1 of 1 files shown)


cmake/defaults/msvcdefaults.cmake
```diff
@@ -56,6 +56,9 @@ _disable_warning("4180")
 # tbb/enumerable_thread_specific.h
 _disable_warning("4334")
 
+# suppress deprecation warning.
+_disable_warning("4996")
```
mattyjams: Does disabling the warning this way also cover the warnings caused by the following two defines, or do we still need those?

ppt-adsk: It does (see
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=vs-2017
), but I would rather leave the defines in the code base, as these are more granular.  When we remove the
_disable_warning("4996")
in the future, I would like us to revert back to the minimal set of deprecation warning disabling provided by _CRT_SECURE_NO_WARNINGS and _SCL_SCCURE_NO_WARNINGS.

PR #204, opened 2020-01-15
Implement ufe attribute notify dev

ppt-adsk:


lib/ufe/StagesSubject.cpp
```diff
@@ -207,6 +224,20 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
...
+        // isRelationalAttributePath() considers only relational attributes
+        if (changedPath.IsPrimPropertyPath()) {
+            if (inAttributeChangedNotificationGuard()) {
+                pendingAttributeChangedNotifications[ufePath] =
+                    changedPath.GetName();
+            }
+            else {
+                Ufe::Attributes::notify(ufePath, changedPath.GetName());
+            }
+        }
```
ppt-adsk: See comments in UsdAttribute.cpp on why this is needed.  Essentially, USD sends two USD notifications for a single set attribute call, so we collapse this down to a single UFE notification.  This is done through use of a guard that restricts the scope of notification collapsing to a single USD set attribute, and a map that does the notification collapsing.

lib/ufe/StagesSubject.cpp
```diff
@@ -223,5 +254,39 @@ void StagesSubject::onStageSet(const UsdMayaProxyStageSetNotice& notice)
...
+
+    if (attributeChangedNotificationGuardCount < 0) {
+        TF_CODING_ERROR("Corrupt attribute changed notification guard.");
+    }
+
+    if (attributeChangedNotificationGuardCount > 0 ) {
+        return;
+    }
+
+    for (const auto& notificationInfo : pendingAttributeChangedNotifications) {
```
ppt-adsk: On guard expiration send out collected notifications.  With current use of the guard, this will be a loop of a single element, but we are ready for any more general use.

lib/ufe/UsdAttribute.cpp
```diff
@@ -47,6 +48,36 @@ static constexpr char kErrorMsgEnumNoValue[] = "Enum string attribute has no val
...
+    // likely because of the warning given in the SdfChangeBlock documentation:
+    // 
+    // https://graphics.pixar.com/usd/docs/api/class_sdf_change_block.html
+    //
+    // which stages that "it is not safe to use [...] [a] downstream API [such
+    // as Usd] while a changeblock is open [...]".
+    //
+    // Therefore, we have implemented an attribute change block notification of
+    // our own in the StagesSubject, which we invoke here, so that only a
+    // single UFE attribute changed notification is generated.
```
ppt-adsk: Would love to get Pixar feedback on this.  It feels like a single call to UsdAttribute Set() should cause a single USD notification, but we get two, as per the comment, and thus the guard workaround.

test/lib/ufe/test-samples/ballset/StandaloneScene/top_layer.ma
```diff
@@ -113,7 +113,7 @@ createNode script -n "uiConfigurationScriptNode";
...
+		+ "            -strokes 1\n            -motionTrails 1\n            -clipGhosts 1\n            -greasePencils 1\n            -shadows 0\n            -captureSequenceNumber -1\n            -width 1125\n            -height 825\n            -sceneRenderFilter 0\n            $editorName;\n        modelEditor -e -viewSelected 0 $editorName;\n\t\tif (!$useSceneConfig) {\n\t\t\tpanel -e -l $label $panelName;\n\t\t}\n\t}\n\n\n\t$panelName = `sceneUIReplacement -getNextPanel \"outlinerPanel\" (localizedPanelLabel(\"ToggledOutliner\")) `;\n\tif (\"\" != $panelName) {\n\t\t$label = `panel -q -label $panelName`;\n\t\toutlinerPanel -edit -l (localizedPanelLabel(\"ToggledOutliner\")) -mbv $menusOkayInPanels  $panelName;\n\t\t$editorName = $panelName;\n        outlinerEditor -e \n            -docTag \"isolOutln_fromSeln\" \n            -showShapes 1\n            -showAssignedMaterials 0\n            -showTimeEditor 1\n            -showReferenceNodes 0\n            -showReferenceMembers 0\n            -showAttributes 0\n            -showConnected 0\n            -showAnimCurvesOnly 0\n            -showMuteInfo 0\n"
```
ppt-adsk: Minor change to the test file because opening it interactively in Maya always turned off shaped display in the Outliner, thus driving me crazy.

test/lib/ufe/testAttribute.py
```diff
@@ -51,6 +71,23 @@ def assertColorAlmostEqual(self, ufeColor, usdColor):
         for va, vb in zip(ufeColor.color, usdColor):
             self.assertAlmostEqual(va, vb, places=6)
 
+    def runUndoRedo(self, attr, newVal, decimalPlaces=None):
```
ppt-adsk: Add undo / redo testing for all attribute types.

test/lib/ufe/testAttribute.py
```diff
@@ -342,3 +405,148 @@ def testAttributeDouble3(self):
 
         # Then make sure that new UFE value matches what it in USD.
         self.assertVectorAlmostEqual(ufeAttr.get(), usdAttr.Get())
+
+        self.runUndoRedo(ufeAttr,
+                         ufe.Vector3d(vec.x()-1.0, vec.y()-2.0, vec.z()-3.0))
+
+    def testObservation(self):
```
ppt-adsk: Add observation testing.  Asserting on the number of notifications received is how I found the duplicate USD attribute Set() notification problem.

test/lib/ufe/ufeTestUtils/mayaUtils.py
```diff
@@ -64,10 +62,14 @@ def isMayaUsdPluginLoaded():
...
-        successLoad = successLoad and loadPlugin(plugin)
-    return successLoad
+    # Load the mayaUsdPlugin first.
+    if not loadPlugin("mayaUsdPlugin"):
+        return False
+
+    # Load the UFE support plugin, for ufeSelectCmd support.  If this plugin
+    # isn't included in the distribution of Maya (e.g. Maya 2019 or 2020), use
+    # fallback test plugin.
+    return loadPlugin("ufeSupport") or loadPlugin("ufeTestCmdsPlugin")
```
ppt-adsk: Next version of Maya has a built-in Python ufeSelectCmd, Maya 2019 and 2020 do not.  Use fallback in those cases.

lib/ufe/StagesSubject.cpp
```diff
@@ -223,5 +254,39 @@ void StagesSubject::onStageSet(const UsdMayaProxyStageSetNotice& notice)
...
+
+    ++attributeChangedNotificationGuardCount;
+
+}
+
+AttributeChangedNotificationGuard::~AttributeChangedNotificationGuard()
+{
+    --attributeChangedNotificationGuardCount;
+
+    if (attributeChangedNotificationGuardCount < 0) {
```
seando-adsk: If you want, these two lines can be condensed to a single line: 
    if (--attributeChangedNotificationGuardCount < 0) {

lib/ufe/UsdAttribute.cpp
mattyjams: @ppt-adsk: Something seems awry here. Your breakdown of the call sequence looks correct, and for the initial creation of an attribute, there should indeed be two notifications: one for the creation of the attribute spec, and then a second for the "info only" change in its value. Subsequent calls to `Set()` on the same attribute should only generate a single notification though. They should end up bailing out of `_CreatePropertySpecForEditing()` early, before an SdfChangeBlock is created or any new property spec is created:
https://github.com/PixarAnimationStudios/USD/blob/4b1162957b2ad219e1635c81de8343be490e41fc/pxr/usd/lib/usd/stage.cpp#L1337

This should be getting tested in core USD here:
https://github.com/PixarAnimationStudios/USD/blob/4b1162957b2ad219e1635c81de8343be490e41fc/pxr/usd/lib/usd/testenv/testUsdNotices.py#L82

Though that tests authoring metadata, I amended the test script locally and verified that the behavior is the same for regular attributes:
```
            attr = s.GetPrimAtPath("/Foo").CreateAttribute(
                "attr", Sdf.ValueTypeNames.Int)
            self.assertEqual(self._changeCount, 4)
            self.assertEqual(self._contentsCount, 2)
            self.assertEqual(self._objectsCount, 2)
            self.assertEqual(self._editTargetsCount, 0)

            self._ResetCounters()
            attr.Set(42)
            self.assertEqual(self._changeCount, 2)
            self.assertEqual(self._contentsCount, 1)
            self.assertEqual(self._objectsCount, 1)
            self.assertEqual(self._editTargetsCount, 0)
```

ppt-adsk: Re-reading your post, and focusing on one of the aspects: "for the initial creation of an attribute, there should indeed be two notifications: one for the creation of the attribute spec, and then a second for the "info only" change in its value."  Does the initial creation of the attribute mean creation of the attribute spec in the edit target layer?  If so, can it be argued that this specific to the USD data model, and should be abstracted away by UFE, which should provide only a single notification of the attribute value being set?  A client interested in USD details should subscribe to USD notifications, and not go through UFE notifications.

I realize this is not answering the question of why subsequent calls to USD set attribute aren't generating a single notification, but I'm wondering whether we need the guard around USD set attribute regardless, because the initial set attribute is expected to generate two USD notifications, which we want to collapse down to a single UFE notification.

kxl-adsk: Indeed, the goal for UFE is to abstract the data model. We don't want to mimic any specific behaviors one data model would have.

lib/ufe/StagesSubject.cpp
```diff
@@ -25,12 +25,32 @@
 #include <ufe/scene.h>
 #include <ufe/sceneNotification.h>
 #include <ufe/transform3d.h>
+#include <ufe/attributes.h>
```
mattyjams: This should go at the top of this group, though `path.h` is unfortunately in the wrong spot too.

lib/ufe/UsdAttribute.cpp
ppt-adsk: I agree with you.  Hopefully we'll be able to gain insight into this through other work and come back to it with a clearer picture.

lib/ufe/StagesSubject.cpp
```diff
@@ -25,12 +25,32 @@
...
+
+bool inAttributeChangedNotificationGuard()
+{
+    return attributeChangedNotificationGuardCount > 0;
+}
+
+std::unordered_map<Ufe::Path, std::string> pendingAttributeChangedNotifications;
+
+}
+#endif
```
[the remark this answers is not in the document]
ppt-adsk: I could put the members in AttributeChangedNotificationGuard, but that would require them being exposed in the header, whereas now they're an implementation detail in the .cpp file.

As to multi-threaded authoring, not a concern for us currently or in the foreseeable future, from what I understand.

lib/ufe/StagesSubject.cpp
ppt-adsk: Done.

test/lib/ufe/testAttribute.py
```diff
@@ -21,8 +21,23 @@
...
+    def __init__(self):
+        super(TestObserver, self).__init__()
+        self.changed = 0
+
+    def __call__(self, notification):
+        if isinstance(notification, ufe.AttributeChanged):
+            self.changed += 1
+
+    def notifications(self):
+        return self.changed
```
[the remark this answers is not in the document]
ppt-adsk: Done.

ppt-adsk:
Thanks Matt for the review!

PR #205, opened 2020-01-15
Backport package stage from refactoring_sandbox

kxl-adsk:


kxl-adsk:
@murphyeoin, @mattyjams this is something that you are not using, but feel free to add review.

build.py
```diff
@@ -274,6 +274,49 @@ def RunCTest(context, extraArgs=None):
...
+        shutil.make_archive(pkgName, 'zip', installDir)
+
+        # copy zip file to package directory
+        try:
+            if os.path.isdir(pkgDir):
+                testFile = os.path.join(pkgDir, "canwrite")
+                open(testFile, "w").close()
+                os.remove(testFile)
+            else:
+                os.makedirs(pkgDir)
```
mattyjams: Why not just let the `shutil.copy()` below raise a permissions error and truncate this to just create the directory if it doesn't exist?

build.py
```diff
@@ -274,6 +274,49 @@ def RunCTest(context, extraArgs=None):
...
+
+            for file in os.listdir(buildDir):
+                if file.endswith(".zip"):
+                    zipFile = os.path.join(buildDir, file)
+                    shutil.copy(zipFile, pkgDir)
+        except Exception as e:
+            PrintError("Could not write to directory {dir}. Change permissions "
+                       "or choose a different location to install to."
+                       .format(dir=dir))
```
mattyjams: Maybe include the exception message in the error message for some additional context?

HamedSabri-adsk: Good point!  sure, I will have a change based on these feedbacks.

PR #211, opened 2020-01-21
Misc build fixes against latest dev

elrond79:
Fixes two issues I encountered when trying to build on linux:

- cmake was erroring when ${MAYAUSD_DEFINE_BOOST_DEBUG_PYTHON_FLAG} was not defined
- PRIu64 was not defined


HamedSabri-adsk:
@elrond79 lgtm! 

Was this the error you were getting?

```
CMake Error at lib/CMakeLists.txt:376 (if):
  if given arguments:
    "IS_WINDOWS" "AND"
  Unknown arguments specified
```

@cfmoore007 I believe this is a better fix to the issues that you mentioned earlier on slack. Can you give it a try and see if it solves your issues?

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
```diff
@@ -13,6 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+#include <cinttypes>
+
```
HamedSabri-adsk: Can this be wrapped around some #ifdef? Not sure if this applies to other platforms.

ppt-adsk: Should be fine, it's a C++ 11 header.

HamedSabri-adsk: sounds good!

elrond79:
> Was this the error you were getting?
> 
> ```
> CMake Error at lib/CMakeLists.txt:376 (if):
>   if given arguments:
>     "IS_WINDOWS" "AND"
>   Unknown arguments specified
> ```

Yup, that was it exactly!  Just FYI, you could also fix by adding quotes - "${MAYAUSD_DEFINE_BOOST_DEBUG_PYTHON_FLAG}" - but stylistically I guess prefer stripping the variable deref.

HamedSabri-adsk:
Good to know, I prefer the style you already chose!

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
kxl-adsk: OK, I understand that fixes done to remove compilation errors on OSX when building in strict mode created this issue - https://github.com/Autodesk/maya-usd/commit/c55c976fcefc07bc44f98c4e808e4089fcb60372#diff-94b4ec7e80f873f5e91609163e535682
Correct?

seando-adsk:
Sorry about the mistake with ${MAYAUSD_DEFINE_BOOST_DEBUG_PYTHON_FLAG} - I don't know why I did that. There is one more in "cmake/defaults/CXXDefaults.cmake" that I think should be changed as well.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
ppt-adsk: Yes, and depending on which of inttypes.h or cinttypes is included first, you will get different results --- see
https://stackoverflow.com/questions/14535556/why-doesnt-priu64-work-in-this-code/30851225

elrond79:
Good catch on the other uses of ${MAYAUSD_DEFINE_BOOST_DEBUG_PYTHON_FLAG}, I pushed a updated fix.  And no worries - I've made the same mistake before (as has probably anyone who's written cmake files at some point).

elrond79:
Thanks for the fast turnaround on this!

PR #216, opened 2020-01-28
MAYA-102918 Adding a basic Export Translator to the MayaUSD plugin

fowlertADSK:
Adding a new Export Translator for the MayaUSD plugin.  Contains a minimal set of export options plus the UI to set them.  For the most part this is a direct copy of the Pixar options with only minor adjustments to how they show up in the export options window, which was easy enough to do since we have based the current read/write jobs in the core library on the Pixar plugin.

### DIFF
--- a/plugin/adsk/plugin/exportTranslator.cpp
+++ b/plugin/adsk/plugin/exportTranslator.cpp
@@ -59,7 +59,7 @@ UsdMayaExportTranslator::writer(const MFileObject &file,
     
     MStringArray filteredTypes;
     // Get the options 
-    if ( optionsString.length() > 0 ) {
+    if (optionsString.length() > 0) {
         MStringArray optionList;
         MStringArray theOption;
         optionsString.split(';', optionList);
--- a/plugin/adsk/plugin/exportTranslator.cpp
+++ b/plugin/adsk/plugin/exportTranslator.cpp
@@ -1,0 +1,211 @@
+//
+// Copyright 2016 Pixar
+// Copyright 2019 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "exportTranslator.h"
+
+#include <mayaUsd/fileio/jobs/jobArgs.h>
+#include <mayaUsd/fileio/shading/shadingModeRegistry.h>
+#include <mayaUsd/fileio/jobs/writeJob.h>
+#include <mayaUsd/fileio/utils/writeUtil.h>
+
+#include <maya/MFileObject.h>
+#include <maya/MGlobal.h>
+#include <maya/MSelectionList.h>
+#include <maya/MString.h>
+
+#include <string>
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+MAYAUSD_NS_DEF {
+
+const MString UsdMayaExportTranslator::translatorName("USD Export");
+
+void* UsdMayaExportTranslator::creator() {
+    return new UsdMayaExportTranslator();
+}
+
+UsdMayaExportTranslator::UsdMayaExportTranslator() :
+        MPxFileTranslator() {
+}
+
+UsdMayaExportTranslator::~UsdMayaExportTranslator() {
+}
+
+MStatus
+UsdMayaExportTranslator::writer(const MFileObject &file, 
+                 const MString &optionsString,
+                 MPxFileTranslator::FileAccessMode mode ) {
+
+    std::string fileName(file.fullName().asChar(), file.fullName().length());
+    VtDictionary userArgs;
+    bool exportAnimation = false;
+    GfInterval timeInterval(1.0, 1.0);
+    double frameStride = 1.0;
+    bool append=false;
+    
+    MStringArray filteredTypes;
+    // Get the options 
+    if ( optionsString.length() > 0 ) {
+        MStringArray optionList;
+        MStringArray theOption;
+        optionsString.split(';', optionList);
+        for(int i=0; i<(int)optionList.length(); ++i) {
+            theOption.clear();
+            optionList[i].split('=', theOption);
+            if (theOption.length() != 2) {
+                continue;
+            }
+
+            std::string argName(theOption[0].asChar());
+            if (argName == "animation") {
+                exportAnimation = (theOption[1].asInt() != 0);
+            }
+            else if (argName == "startTime") {
+                timeInterval.SetMin(theOption[1].asDouble());
+            }
+            else if (argName == "endTime") {
+                timeInterval.SetMax(theOption[1].asDouble());
+            }
+            else if (argName == "frameStride") {
+                frameStride = theOption[1].asDouble();
+            }
+            else if (argName == "filterTypes") {
+                theOption[1].split(',', filteredTypes);
+            }
+            else {
+                userArgs[argName] = UsdMayaUtil::ParseArgumentValue(
+                    argName, theOption[1].asChar(),
+                    PXR_NS::UsdMayaJobExportArgs::GetDefaultDictionary());
+            }
+        }
+    }
+
+    // Now resync start and end frame based on export time interval.
+    if (exportAnimation) {
+        if (timeInterval.IsEmpty()) {
+            // If the user accidentally set start > end, resync to the closed
+            // interval with the single start point.
+            timeInterval = GfInterval(timeInterval.GetMin());
+        }
+    }
+    else {
+        // No animation, so empty interval.
+        timeInterval = GfInterval();
+    }
+
+    MSelectionList objSelList;
+    if(mode == MPxFileTranslator::kExportActiveAccessMode) {
+        // Get selected objects
+        MGlobal::getActiveSelectionList(objSelList);
+    } else if(mode == MPxFileTranslator::kExportAccessMode) {
+        // Get all objects at DAG root
+        objSelList.add("|*", true);
+    }
+
+    // Convert selection list to jobArgs dagPaths
+    UsdMayaUtil::MDagPathSet dagPaths;
+    for (unsigned int i=0; i < objSelList.length(); i++) {
+        MDagPath dagPath;
+        if (objSelList.getDagPath(i, dagPath) == MS::kSuccess) {
+            dagPaths.insert(dagPath);
+        }
+    }
+    
+    if (dagPaths.empty()) {
+        TF_WARN("No DAG nodes to export. Skipping.");
+        return MS::kSuccess;
+    }
+
+    const std::vector<double> timeSamples = UsdMayaWriteUtil::GetTimeSamples(
+            timeInterval, std::set<double>(), frameStride);
+    PXR_NS::UsdMayaJobExportArgs jobArgs = PXR_NS::UsdMayaJobExportArgs::CreateFromDictionary(
+            userArgs, dagPaths, timeSamples);
+    for (unsigned int i=0; i < filteredTypes.length(); ++i) {
+        jobArgs.AddFilteredTypeName(filteredTypes[i].asChar());
+    }
+
+    UsdMaya_WriteJob writeJob(jobArgs);
+    if (!writeJob.Write(fileName, append)) {
+        return MS::kFailure;
+    }
+    
+    return MS::kSuccess;
+}
+
+MPxFileTranslator::MFileKind
+UsdMayaExportTranslator::identifyFile(
+        const MFileObject& file,
+        const char*  /*buffer*/,
+        short  /*size*/) const
+{
+    MFileKind retValue = kNotMyFileType;
+    const MString fileName = file.fullName();
+    const int lastIndex = fileName.length() - 1;
+
+    const int periodIndex = fileName.rindex('.');
+    if (periodIndex < 0 || periodIndex >= lastIndex) {
+        return retValue;
+    }
+
+    const MString fileExtension = fileName.substring(periodIndex + 1, lastIndex);
+
+    if (fileExtension ==
+            UsdMayaTranslatorTokens->UsdFileExtensionDefault.GetText() || 
+        fileExtension ==
+            UsdMayaTranslatorTokens->UsdFileExtensionASCII.GetText() || 
+        fileExtension ==
+            UsdMayaTranslatorTokens->UsdFileExtensionCrate.GetText() ||
+        fileExtension ==
+            UsdMayaTranslatorTokens->UsdFileExtensionPackage.GetText()) {
+        retValue = kIsMyFileType;
+    }
+
+    return retValue;
+}
+
+/* static */
+const std::string&
+UsdMayaExportTranslator::GetDefaultOptions()
+{
+    static std::string defaultOptions;
+    static std::once_flag once;
+    std::call_once(once, []() {
+        std::vector<std::string> entries;
+        for (const std::pair<std::string, VtValue> keyValue :
+                PXR_NS::UsdMayaJobExportArgs::GetDefaultDictionary()) {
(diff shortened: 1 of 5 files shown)


plugin/adsk/plugin/plugin.cpp
```diff
@@ -59,6 +60,17 @@ MStatus initializePlugin(MObject obj)
         status.perror("mayaUsdPlugin: unable to register import translator.");
     }
 
+    status = plugin.registerFileTranslator(
+        MAYAUSD_NS::UsdMayaExportTranslator::translatorName,
```
seando-adsk: You should just use "MayaUsd" instead of "MAYAUSD_NS"

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -15,20 +15,92 @@
 // limitations under the License.
 //
 
+global proc mayausdTranslatorExport_AnimationCB() {
```
seando-adsk: To be consistent with other places it would be better to use "mayaUsd"

seando-adsk: Such as the import where it has: mayaUsdTranslatorImport_XXX

plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -0,0 +1,211 @@
...
+    std::string fileName(file.fullName().asChar(), file.fullName().length());
+    VtDictionary userArgs;
+    bool exportAnimation = false;
+    GfInterval timeInterval(1.0, 1.0);
+    double frameStride = 1.0;
+    bool append=false;
+    
+    MStringArray filteredTypes;
+    // Get the options 
+    if ( optionsString.length() > 0 ) {
```
pilarmolinalopez: why the spaces between the brackets and the content? I don't see this style in the rest of this file

plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include <mayaUsd/fileio/utils/writeUtil.h>
+
+#include <maya/MFileObject.h>
+#include <maya/MGlobal.h>
+#include <maya/MSelectionList.h>
+#include <maya/MString.h>
+
+#include <string>
+
+PXR_NAMESPACE_USING_DIRECTIVE
```
pilarmolinalopez: not super familiar with this but why are we using PXR_NAMESPACE_USING_DIRECTIVE here while the importer uses PXR_NAMESPACE_OPEN_SCOPE ?

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -37,20 +109,138 @@ global proc int mayaUsdTranslatorExport (string $parent,
 //            resultCallback ( string $optionsString )
 //
 //    Returns:
-//        1 if successful.
+//        1 if successfull.
```
pilarmolinalopez: I believe this is wrong

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -15,20 +15,92 @@
...
-//    This script posts the mayaUsd export file translator options.
+//    This script posts the USD Export file translator options.
 //    The optionsString is of the form:
 //    varName1=value1;varName2=value2;...
 //
 //  Parameters:
 //    $parent - the elf parent layout for this options layout. It is
 //            always a scrollLayout.
-//    $action - the action that is to be performed with this invocation of this proc.
+//    $action - the action that is to be performed with this invokation of this proc.
```
pilarmolinalopez: ?

fowlertADSK: Hmm...I didn't even notice these two "typo" fixes.  I didn't change those myself but I'll revert them both.

plugin/adsk/plugin/exportTranslator.cpp
fowlertADSK: I believe the importer is incorrect.  So far both import and export are basically copied from the Pixar plugin and I think the source for the ADSK importer was copy/pasted from Pixar and then edited a bit but the Pixar namespace wasn't changed.  PXR_NAMESPACE_OPEN_SCOPE puts the class into the pxr namespace but the ADSK plugin has it’s own.  I also copy/pasted a lot of what the Pixar plugin has in its export code, but when I went with the MayaUSD namespace for the exporter I had to either add a “pxr::” to the start of some of the functions or just use that macro to add a “using pxr”.

plugin/adsk/plugin/exportTranslator.cpp
fowlertADSK: I think it was only because I did most of the work in one editor that was automatically enforcing a style, and the flipped to another for a bit.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
fowlertADSK: Sure.  I'm also planning on removing the "global" from most of these functions since it isn't needed.

plugin/adsk/plugin/exportTranslator.cpp
fowlertADSK: Actually...it was like this because that's also how it is in the Pixar plugin code I copied it from.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
fowlertADSK: And just like the spacing above, these two typos actually came over from the equivalent code in the Pixar plugin.

plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -0,0 +1,211 @@
...
+                entries.push_back(TfStringPrintf("%s=%s",
+                        keyValue.first.c_str(),
+                        keyValue.second.Get<std::string>().c_str()));
+            }
+        }
+        entries.push_back("animation=0");
+        entries.push_back("startTime=1");
+        entries.push_back("endTime=1");
+        entries.push_back("frameStride=1.0");
+        defaultOptions = TfStringJoin(entries, ";");
```
robthebloke: Lordy loo! Talk about convoluted! Just use a ```std::ostringstream```, and build the string in one go. Better yet, use our code generators that handle all of this boiler plate stuff for you ;)

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -41,16 +108,134 @@ global proc int mayaUsdTranslatorExport (string $parent,
...
+        $currentOptions = mayaUsdTranslatorExport_AppendFromCheckbox($currentOptions, "animation", "animationCheckBox");
+        $currentOptions = mayaUsdTranslatorExport_AppendFromCheckbox($currentOptions, "eulerFilter", "eulerFilterCheckBox");
+        $currentOptions = mayaUsdTranslatorExport_AppendFrameRange($currentOptions);
+        $currentOptions = mayaUsdTranslatorExport_AppendFromFloatField($currentOptions, "frameStride", "frameStrideField");
+        $currentOptions = mayaUsdTranslatorExport_AppendFromTextField($currentOptions, "parentScope", "parentScopeField");
+        $currentOptions = mayaUsdTranslatorExport_AppendFromCheckbox($currentOptions, "exportInstances", "exportInstancesCheckBox");
+        $currentOptions = mayaUsdTranslatorExport_AppendFromCheckbox($currentOptions, "exportVisibility", "exportVisibilityCheckBox");
+        $currentOptions = mayaUsdTranslatorExport_AppendFromCheckbox($currentOptions, "mergeTransformAndShape", "mergeTransformAndShapeCheckBox");
+        $currentOptions = mayaUsdTranslatorExport_AppendFromCheckbox($currentOptions, "stripNamespaces", "stripNamespacesCheckBox");
```
robthebloke: FWIW, we've banned people writing boiler plate MEL code here. It takes too much dev effort, it's bug prone, and it's usually trivial to auto generate the code. For reference, this is the method that generates our import/export dialogs: 

https://github.com/Autodesk/maya-usd/blob/dev/plugin/al/mayautils/AL/maya/utils/PluginTranslatorOptions.cpp#L712

And here's an example of how we're specifying the options:
https://github.com/Autodesk/maya-usd/blob/dev/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/ExportTranslator.h#L74 

All of our AE templates, menus, command dialogs, export dialogs are all autogenerated. The code generators are all within our mayautils lib here:
https://github.com/Autodesk/maya-usd/tree/dev/plugin/al/mayautils/AL/maya/utils

plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -0,0 +1,211 @@
...
+    if (dagPaths.empty()) {
+        TF_WARN("No DAG nodes to export. Skipping.");
+        return MS::kSuccess;
+    }
+
+    const std::vector<double> timeSamples = UsdMayaWriteUtil::GetTimeSamples(
+            timeInterval, std::set<double>(), frameStride);
+    PXR_NS::UsdMayaJobExportArgs jobArgs = PXR_NS::UsdMayaJobExportArgs::CreateFromDictionary(
+            userArgs, dagPaths, timeSamples);
+    for (unsigned int i=0; i < filteredTypes.length(); ++i) {
```
robthebloke: Loop condition depends on a DSO exported function. Take a local copy.
```c++
for (unsigned int i=0, filteredTypes.length(); i < n; ++i)
```

plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -0,0 +1,211 @@
...
+        // Get selected objects
+        MGlobal::getActiveSelectionList(objSelList);
+    } else if(mode == MPxFileTranslator::kExportAccessMode) {
+        // Get all objects at DAG root
+        objSelList.add("|*", true);
+    }
+
+    // Convert selection list to jobArgs dagPaths
+    UsdMayaUtil::MDagPathSet dagPaths;
+    for (unsigned int i=0; i < objSelList.length(); i++) {
```
robthebloke: Loop condition depends on a DSO exported function. Take a local copy.
```c++
 for (unsigned int i=0, n = objSelList.length(); i < n; i++)
```

plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -0,0 +1,211 @@
...
+        MStringArray theOption;
+        optionsString.split(';', optionList);
+        for(int i=0; i<(int)optionList.length(); ++i) {
+            theOption.clear();
+            optionList[i].split('=', theOption);
+            if (theOption.length() != 2) {
+                continue;
+            }
+
+            std::string argName(theOption[0].asChar());
```
robthebloke: Why are you taking a copy? MString already has support for == "textString".

plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -0,0 +1,211 @@
...
+    double frameStride = 1.0;
+    bool append=false;
+    
+    MStringArray filteredTypes;
+    // Get the options 
+    if (optionsString.length() > 0) {
+        MStringArray optionList;
+        MStringArray theOption;
+        optionsString.split(';', optionList);
+        for(int i=0; i<(int)optionList.length(); ++i) {
```
robthebloke: Loop condition depends on a DSO exported function. Take a local copy of the length.

plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+UsdMayaExportTranslator::~UsdMayaExportTranslator() {
+}
+
+MStatus
+UsdMayaExportTranslator::writer(const MFileObject &file, 
+                 const MString &optionsString,
+                 MPxFileTranslator::FileAccessMode mode ) {
+
+    std::string fileName(file.fullName().asChar(), file.fullName().length());
```
robthebloke: const?

plugin/adsk/plugin/exportTranslator.cpp
fowlertADSK: Looks like it was originally done this way because it's used with the VtDictionary in the final 'else' case, which needs the std::string.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
fowlertADSK: Boilerplate is when you write/copy the exact same chunk of code over and over again.  Extracting a chunk of code into its own function and then calling that function in multiple places is the opposite of “boilerplate code” is it not?  That is what’s going on here, each of the checkbox’s or float fields call the exact same function with their option/control names passed in.

I’ve actually had to debug a bit of the C++ auto-generated MEL in the AL plugin and I’ll admit that I wasn’t a fan of that experience.  I’m not the original author of most of this so I don’t have a really strong opinion about a lot of it (almost everything here comes from the Pixar plugin), but I’m definitely not anti-MEL and wouldn’t want to move all of this into compiled C++ for no real benefit.  E.g. everything in PluginTranslatorOptions.cpp is actually MEL…a developer had to write all of that in MEL, debug all of that in MEL, before copying it into C++, adding a ton of casts, making sure they got all of their strings quoted correctly, before compiling it.  I’m not sure I’d agree with that approach being easier and less bug prone.  I’m sure that once it’s done it’s pretty much never going to change again, but I think it’s the same for the functions here.  

The one thing I would like to change is to move some of this into a shared file because I think there are exact copies of all of this in the import .mel file.


Now I have to say Rob that it’s quite ironic that you and I are talking about MEL today because I just spent an hour looking at your web page.  A Maya user had hacked together a script based on some snippets they grabbed from your MEL tutorials and couldn’t figure out how to make it all work so they contacted our Support for help.  At least I’m assuming it’s your page…I can’t imagine there being more than one Maya expert named RobTheBloke 😊

plugin/adsk/plugin/exportTranslator.cpp
fowlertADSK: Yeah, I guess instead of a vector of strings with a join at the end we could use ostringstream.  I’ll take a look at simplifying this one.

One general comment to make about using any of the AL utility code…  At this point we have moved a lot of the Pixar translator code down into the core lib, so pretty much all of the options defined down there are still exactly how Pixar has had them, and this Default Dictionary stuff is all Pixar right now, but not much of the AL code has been moved there yet.  I don’t think anyone is avoiding using AL code but a lot it isn’t available just yet in the shared area.

plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -16,6 +16,10 @@
 //
 #include "exportTranslator.h"
 
+#include <string>
+#include <set>
+#include <sstream>
+
```
fowlertADSK: Following our new convention of putting these at the top.

plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -57,6 +61,8 @@ UsdMayaExportTranslator::writer(const MFileObject &file,
     double frameStride = 1.0;
     bool append=false;
     
+    std::set<double> frameSamples;
```
fowlertADSK: A default/empty set<double> was always passed in down below, now parsing the input options to see if the user set them in the UI.

plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -184,25 +203,23 @@ UsdMayaExportTranslator::GetDefaultOptions()
     static std::string defaultOptions;
     static std::once_flag once;
     std::call_once(once, []() {
-        std::vector<std::string> entries;
+        std::ostringstream optionsStream;
```
fowlertADSK: Changing to an ostringstream as suggested.

fowlertADSK:
Apologies for leaving this one open for so long.  It was originally intended for a Maya PR but there were issues that paused the integration and I temporarily moved on to other things.  I think I have made all requested changes now, but in the meantime I did add one new option + made some changes to align with the design spec which you will see here.

The new option is a text field line to allow users to type in multi-sample #’s, which is an “advanced” option that people are probably not used to setting in any UI’s right now.  Once we add actual commands for Import/Export and people can script this behavior (which I think is how people have been doing it with the Pixar plugin) then we might come back to this UI, but for now we wanted something in to allow people to test the behavior.

Can you take another look @pilarmolinalopez and @robthebloke  ?  Thx!

PR #217, opened 2020-01-28
apply maya.standalone.uninitialize() fix from #198 to mayaUsd tests

mattyjams:
This prevents `unittest.main()` from exiting on test failures and ensures that `maya.standalone.uninitialize()` gets called before returning success or failure.

@HamedSabri-adsk previously made this fix for the UFE tests in #198.

### DIFF
--- a/test/lib/CMakeLists.txt
+++ b/test/lib/CMakeLists.txt
@@ -55,12 +55,15 @@ foreach(script ${test_script_files})
     add_test(
         NAME ${target}
         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
-        COMMAND ${MAYA_PY_EXECUTABLE} -c "from unittest import main; \
+        COMMAND ${MAYA_PY_EXECUTABLE} -c "import sys; \
+                                          from unittest import main; \
                                           import maya.standalone; \
                                           maya.standalone.initialize(name='python'); \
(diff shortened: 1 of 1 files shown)


test/lib/CMakeLists.txt
```diff
@@ -55,12 +55,15 @@ foreach(script ${test_script_files})
...
                                           import ${target}; \
-                                          main(module=${target}); \
-                                          maya.standalone.uninitialize()"
+                                          testProg = main(module=${target}, exit=False); \
+                                          maya.standalone.uninitialize(); \
+                                          sys.exit(not testProg.result.wasSuccessful());
+                                          "
     )
     set_property(TEST ${target} APPEND PROPERTY ENVIRONMENT
         "PATH=${path}"
```
HamedSabri-adsk: @mattyjams, looks good to me.  Thank you.

On another note, we now have a new utility function "separate_argument_list" that was recently added to utils.cmake and deals with those crazy ";" ":" stuff on different platforms.
  
Please replace line 42 to 51 with these two call:

```
separate_argument_list(path)
separate_argument_list(pythonPath)
```

mattyjams: Ahh, good call! I opened #218 for that change.

HamedSabri-adsk: Thanks.

PR #219, opened 2020-01-29
Fix compilation error in strict mode on macOS

kxl-adsk:


lib/ufe/UsdAttributes.cpp
```diff
@@ -27,9 +27,6 @@
 //		the one in USD.
 // PXR_NAMESPACE_USING_DIRECTIVE
 
-static constexpr char kErrorMsgUnknown[] = "Unknown UFE attribute type encountered";
```
ppt-adsk: Should be
#ifdef UFE_ENABLE_ASSERTS
[lines 30 and 31]
#endif

kxl-adsk: Good catch!

PR #223, opened 2020-01-29
Update maya-to-hydra on refactoring sandbox

elrond79:
This makes it compatible with usd-20.02 (and 19.07, and 19.11), and restores proper translation of basic textures (ie, the maya "file" node).

@mattyjams - I know this is a PR against refactoring sandbox, where we generally don't require as much scrutiny - but I'd like you to look over  feedback on these two commits, if you can:

https://github.com/Autodesk/maya-usd/pull/223/commits/7924a8eb449cba3cbe11f42b5987f862e207a3f5
https://github.com/Autodesk/maya-usd/commit/ca1aaf7f719af1a13b0364f029aaab20df8d70ae

One thing in particular I'm uncertain of (in ca1aaf7f719af1a13b0364f029aaab20df8d70ae) is how I'm overriding HdSceneDelegate::GetTextureResourceID... or, to be more precise, how I'm NOT overriding it in usd-0.20.02.

It seemed like the way that HdStorm was using this was it was only returning a value (other than -1) if it was for a texture buffer, not for on-file textures, read in as part of the material network.  In which case, my scene delegate just always returns -1 (since I don't make any texture buffers outside of those used directly by materials).  Is this correct?


elrond79:
Oops, I realized the branch I pushed was slightly out of date. Update the commit ids in my previous message.

lib/usd/hdMaya/utils.cpp
```diff
@@ -0,0 +1,177 @@
...
+                                     MayaAttrs::file::fileTextureName, true)
+                                 .asString()
+                                 .asChar()}
+                   : ret;
+    }
+}
+
+std::tuple<HdWrap, HdWrap> GetFileTextureWrappingParams(
+    const MObject& fileObj) {
+    constexpr std::tuple<HdWrap, HdWrap> def{HdWrapClamp, HdWrapClamp};
```
kxl-adsk: This line fails to compile on macOS (not in strict mode), details

/git/usd/maya-usd/lib/usd/hdMaya/utils.cpp:110:42: error: constexpr variable 'def' must be initialized by a constant expression
    constexpr std::tuple<HdWrap, HdWrap> def{HdWrapClamp, HdWrapClamp};
                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/git/usd/maya-usd/lib/usd/hdMaya/utils.cpp:110:42: note: non-constexpr constructor 'tuple<pxrInternal_v0_20__pxrReserved__::HdWrap, pxrInternal_v0_20__pxrReserved__::HdWrap, false, false>' cannot be used in a constant expression
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/tuple:723:9: note: declared here
        tuple(_Up&&... __u)

elrond79:
I added another commit, which adds an option to enable / disable building of maya-to-hydra, as  well as consolidates the hdMaya changes more, to facilitate extraction to dev

lib/usd/hdMaya/utils.cpp
elrond79: Hmm, guess this is a difference between how clang and gcc handle constexpr? I'll take a look...

kxl-adsk:
I like the cleanup you are doing. Any chance you are planning on bringing tests as well?

elrond79:
Hah, I actually hadn't noticed yet that the tests never made the transition from the maya-to-hydra repo.

Unfortunately, I probably won't have time in this push to migrate those.  If you want, we can set the default for BUILD_HDMAYA to false, until we do...

kxl-adsk:
Let's keep it on, we will catch this way any potential breaking changes in USD.

lib/usd/hdMaya/utils.cpp
elrond79: So, apparently, tuples are strictly only constexpr for c++14, not c++11.  Apparently gcc still allows / makes them constexpr, though.

Regardless, compiling in optimized mode should treat const and constexpr in this situation identically:

https://godbolt.org/z/C6wx5U

...so I'll just change to const.

elrond79: Fixed!

elrond79:
More updates - this corresponds to the the state pushed to dev in #231, https://github.com/Autodesk/maya-usd/pull/231/commits/e459338caa5765f56748fb9511db0fa35c3869d0.

lib/usd/hdMaya/adapters/materialAdapter.cpp
```diff
@@ -552,6 +559,52 @@ class HdMayaShadingEngineAdapter : public HdMayaMaterialAdapter {
...
+
+
+    HdTextureResourceSharedPtr GetTextureResource(
+        const TfToken& paramName) override {
+        TF_DEBUG(HDMAYA_ADAPTER_MATERIALS)
+            .Msg("HdMayaShadingEngineAdapter::GetTextureResource(%s): %s\n",
+                    paramName.GetText(), GetID().GetText());
+        auto fileObj = GetConnectedFileNode(_surfaceShader, paramName);
+        if (fileObj == MObject::kNullObj) { return {}; }
+        if(GetDelegate()->IsHdSt()) {
```
mattyjams: Maybe move this check before we try to get `fileObj` so we can bail out earlier if it fails?

lib/usd/hdMaya/CMakeLists.txt
```diff
@@ -1,28 +1,35 @@
...
 
+if(CMAKE_WANT_UFE_BUILD)
+    if(UFE_FOUND)
+        message(STATUS "Building with UFE ${UFE_VERSION} features enabled.")
+        include_directories(${UFE_INCLUDE_DIR})
+        add_definitions(-DWANT_UFE_BUILD)
+    else()
+        message(STATUS "UFE not found. UFE features will be disabled.")
+    endif()
+endif()
```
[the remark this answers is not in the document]
elrond79: re: redundancy - I had sort of the same thoughts.  However, this is sort of an established pattern at this point (search for "Building with UFE" to see what I mean), so my vote would be to address it in a unified way, across the project, as a separate PR.

If it still bothers you, though, I don't mind changing it here.

Also, FYI - we've already bumped the minimum CMake version. We're now on 3.12.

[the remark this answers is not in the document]
mattyjams: Yeah, that's cool with me. A targeted clean-up in a later PR makes more sense.

And thanks for the reminder about the current minimum version! I think I remember seeing 3.1.1 in a non-top-level CMakeLists.txt somewhere, so I couldn't remember what we actually landed on.

lib/usd/hdMaya/adapters/materialAdapter.cpp
elrond79: Addressed in the dev-branch PR:

https://github.com/Autodesk/maya-usd/pull/231/commits/6549209f9b47035641d63dd4afaedfbf9a84d46b

I'll migrate (or merge) all those changes over to this branch once #231 is merged - assuming we still need refactoring_sandbox at that point!

kxl-adsk: We have one more PR to do - MayaReference schema move. After that, would be good to have ability to confirm that nothing else was forgotten.

So yes, please merge changes back once #231 is merged (I will let you pick the best strategy for it).

elrond79:
This merges in the latest dev - including pr/hdmaya/maya_to_hydra_dev - into refactoring sandbox.

For reference, these merges were made:

- merge the dev commit, immediately before pr/hdmaya/maya_to_hydra_dev was merged, into here
- merge the base commit (73730b7f00dae508b3508df1958ba310df2d43de) from pr/maya/maya_to_hydra_dev - because I had already merged dev, and everything from this commit was already in this branch, I knew I could make this as a do-nothing merge (ie, `git merge -s ours`)
- merge pr/hdmaya/maya_to_hydra_dev - this brought over all the code-review commits, made since the last update to this branch
- merge dev (after the merge of  pr/hdmaya/maya_to_hydra_dev) - this resulted in no changes, and is basically just book-keeping, to make sure latest dev is merged into refactoring_sandbox

elrond79:
After doing these merges, I then compared the diffs of dev and this branch, as a sanity check - the only maya-to-hydra diff was a tabs-vs-space-tabs issue, fixed in https://github.com/Autodesk/maya-usd/pull/223/commits/6d46c8effdce42e5b668c20ffb7d206fb1639882

PR #224, opened 2020-01-29
[util] make MayaNodeNameToSdfPath deal with underworld nodes

elrond79:
This opens the door for exporting / importing imagePlane nodes in the future.

### DIFF
--- a/lib/utils/util.cpp
+++ b/lib/utils/util.cpp
@@ -69,6 +69,7 @@
 
 #include <boost/functional/hash.hpp>
 
+#include <algorithm>
 #include <sstream>
 #include <string>
 #include <unordered_map>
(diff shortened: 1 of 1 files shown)


kxl-adsk:
It's unclear to me what is the problem that this PR is solving? Once that's cleared out, we will need a unit test for it.

elrond79:
Without this change, it will error when it tries to translate underworld nodes, like:

`|foo|myMesh|myMeshShape->|underworldXform`

or, more commonly:

`|foo|myCamera|myCameraShape->|imagePlane1|imagePlaneShape1`

This came out of my maya-to-hydra refactoring, where I was removing the luma-specific code.  The removed code adds support for imagePlanes, but requires a corresponding branch on USD, which we haven't published yet.

This was the only bit that wasn't directly related to image planes - I wasn't sure how common underworld-node use was, outside of image planes, so I thought I'd make the PR.  (I only know of one other case where underworld nodes are created automatically - some sort of nurbs-surface projection node, I think.) If you don't think anyone will care, I can just make it part of our internal (imagePlane-specific) branch, and I'll close this PR.

And yeah, if you think it would be useful, I can absolutely make a unittest for it!

kxl-adsk:
Thank you for the explanation. It makes sense to add and you have a very valid case. Please add a unit test for it.

elrond79:
I couldn't find an easy way to create an import / export error that this would solve, without also adding an ImagePlane translator, so I'm just going to close this PR for now. Will re-open if I come up with a better test case.

PR #225, opened 2020-01-30
linux/rpath: fix relative path finding for symlinked dirs

elrond79:
This fixes linking issues if you happen to have a symlink as part of your build path.

Previously, mayaUsd_add_rpath would use "realpaths" when comparing to origin, but mayaUsd_init_rpath would ONLY use a realpath if the argument was originally a relative path.  Thus, if you fed in an absolute path with a symlink, you would end up comparing a symlink path to a realpath.

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -83,8 +83,10 @@ function(mayaUsd_init_rpath rpathRef origin)
         else()
             set(origin "${CMAKE_INSTALL_PREFIX}/${origin}")
         endif()
-        get_filename_component(origin "${origin}" REALPATH)
     endif()
+    # mayaUsd_add_rpath uses REALPATH, so we must make sure we always
+    # do so here too, to get the right relative path
+    get_filename_component(origin "${origin}" REALPATH)
     set(${rpathRef} "${origin}" PARENT_SCOPE)
 endfunction()
 


elrond79:
This addresses the note mentioned here:

https://github.com/Autodesk/maya-usd/pull/198#pullrequestreview-348126946
https://github.com/Autodesk/maya-usd/pull/198#issuecomment-579050095

kxl-adsk:
@elrond79 what's the simplest way to reproduce the issue using our build script?

elrond79:
This was on Linux - add args (ie, --pxrusd-location, etc) as needed to find your dependencies.

```
# The problem seems to crop up on rebuilds - to simulate, we pre-create this dir
mkdir -p /tmp/dir1/dir2/dir3/workspace/install/RelWithDebInfo/plugin/al/lib
ln -s /tmp/dir1/dir2/dir3 /tmp/symlinkDir
python $MAYA_USD_SRC/build.py /tmp/symlinkDir/workspace -v 3 --stages configure,build,install
```

The above build should appear to complete successfully - however, if you inspect some of the libs, you'll see they have linking problems - ie:

```ldd /tmp/symlinkDir/workspace/install/RelWithDebInfo/plugin/al/lib/python/AL/usd/transaction/_AL_USDTransaction.so | grep AL_USDTransaction.so
```

...which should show:

```
libAL_USDTransaction.so => not found
```

You can also inspect the RUNPATH directly, to see that it's going up the wrong number of directories:

```
readelf -a /tmp/symlinkDir/workspace/install/RelWithDebInfo/plugin/al/lib/python/AL/usd/transaction/_AL_USDTransaction.so | grep RUNPATH
```

PR #226, opened 2020-01-30
[al] make readAnimatedValues default to On

elrond79:
...a recent AL change (97d3ff3e) added the ability to control the setting
for pushToPrim for newly created nodes via an optionVar, which is useful,
but it also changed the default from on to off.

In addition to being breaking behavior, this is also highly
non-intuitive for artists.

This is the sister PR to https://github.com/Autodesk/maya-usd/pull/149 - but I decided to break it into it's own PR, since the performance implications are potentially bigger, so I wanted to give others a chance to object.


kxl-adsk:
@murphyeoin We don't really have preferences for this particular PR. How should we proceed?
@cfmoore007 Are you internally using this PR? Any concerns for merging?

cfmoore007:
We set this value on by default, but do not currently use this PR to do so.  No concerns for merging here.

kxl-adsk:
Closing as per offline conversation with @elrond79

PR #229, opened 2020-01-31
Write out HEAD sha/date in the build.log for the sake of troubleshooting & diagnostics

HamedSabri-adsk:


build.py
```diff
@@ -64,6 +64,17 @@ def GetCommandOutput(command):
         pass
     return None
 
+def GetGitHeadInfo(context):
+    """write out HEAD commitID and Date."""
+    try:
+        with CurrentWorkingDirectory(context.mayaUsdSrcDir):
+            commitSha = subprocess.check_output('git rev-parse HEAD').decode()
+            commitDate = subprocess.check_output('git log -1 --pretty=format:"%ad"'.split()).decode()
```
mattyjams: We just grabbed the commit in the previous command, so maybe slightly cleaner to get the date this way?:
```
git show <commitSha> --pretty=format:"%ad"
```
Either works though. :)

HamedSabri-adsk: Thanks @mattyjams  good suggestion.  Just one minor thing, `git show <commitSha>` shows diff information which we don't want. passing `-s` did the trick.

Also instead of passing commitSHA, I simply passed HEAD.  Either of these cosmetic works.

`git show -s HEAD --format="%ad"`

HamedSabri-adsk:
> 
> 
> @HamedSabri-adsk Can you test it on MacOS - I'm getting `ERROR: Failed to run git commands : [Errno 2] No such file or directory`

Hey @kxl-adsk , should be fixed now.

PR #230, opened 2020-01-31
[vp2RenderDelegate] fix error/warning unused GetListAdjustment

elrond79:
GetListAdjustment is used iff WANT_UFE_BUILD is on, so added a guard

also, I noticed that the _globalListAdjustment is only used if WANT_UFE_BUILD
is on, so I also added a guard for that

now that PXR_STRICT_BUILD_MODE is on by default, this will cause the build to fail without this fix

### DIFF
--- a/lib/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -67,6 +67,7 @@ namespace
     //! Representation selector for selection update
     const HdReprSelector kSelectionReprSelector(HdVP2ReprTokens->selection);
 
+#if defined(WANT_UFE_BUILD)
     MGlobal::ListAdjustment GetListAdjustment()
     {
         // Keyboard modifiers can be queried from QApplication::keyboardModifiers()
@@ -96,6 +97,7 @@ namespace
 
         return listAdjustment;
     }
+#endif // defined(WANT_UFE_BUILD)
 
     //! \brief  Configure repr descriptions
     void _ConfigureReprs()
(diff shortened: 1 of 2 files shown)


kxl-adsk:
Hmmm...change makes sense but for some reason, I can't reproduce compilation error (both on Windows and OSX).

HdC-adsk:
LGTM.
@kxl-adsk Have you compiled the code change with strict mode? If not, I will run it.

kxl-adsk:
@HdC-adsk I will test it before merging. Please approve if everything is ok.

elrond79:
@kxl-adsk The compilation error will only happen if `PXR_STRICT_BUILD_MODE` is on (which it now is by default), `WANT_UFE_BUILD` is off, and `MAYA_ENABLE_UPDATE_FOR_SELECTION` is off.  That last one, it seems, is on iff maya version is >= 2020 - so you'll only see the error for maya <= 2019.

Currently, though, if `WANT_UFE_BUILD` is off, but `MAYA_ENABLE_UPDATE_FOR_SELECTION` is on, the warning won't trigger, but it probably should - because though `GetListAdjustment` is used, it's only in a trivial way - ie, it it stores it's value in `_globalListAdjustment`, which is then never used.  That's the other half of what this PR is doing - making `_globalListAdjustment` only be defined / set if `WANT_UFE_BUILD` is on.

kxl-adsk:
Now it makes sense, thank you @elrond79

elrond79:
I do have a Mac laptop I can use, but the USD build is out of date.  Also, when I tried to compile, I ran into this issue: https://github.com/Autodesk/maya-usd/pull/234

Do you still want me to confirm I see the test once my USD build finishes?

kxl-adsk:
No, I'm good.

elrond79:
Ok, cool - and thanks!

PR #232, opened 2020-01-31
Improvements to USDGeomScope Functionality 

murphyeoin:
+ File Import now tries to create maya nodes which have an al_usdmaya_transformType metadatum
+ fixed a bug in TransformationMatrix inheritance

This is internal AL#1013


kxl-adsk:
@fowlertADSK, @pilarmolinalopez - since you have done the initial USDGeomScope review, I hope you don't mind if I send this one your way too.

PR #237, opened 2020-01-31
Tremblp/maya 102361/rename fixes

ppt-adsk:


lib/ufe/ProxyShapeHierarchy.cpp
```diff
@@ -48,11 +51,22 @@ ProxyShapeHierarchy::~ProxyShapeHierarchy()
...
 
 /*static*/
-ProxyShapeHierarchy::Ptr ProxyShapeHierarchy::create(Ufe::HierarchyHandler::Ptr mayaHierarchyHandler)
+ProxyShapeHierarchy::Ptr ProxyShapeHierarchy::create(const Ufe::HierarchyHandler::Ptr& mayaHierarchyHandler)
 {
 	return std::make_shared<ProxyShapeHierarchy>(mayaHierarchyHandler);
 }
 
+/*static*/
+ProxyShapeHierarchy::Ptr ProxyShapeHierarchy::create(
```
ppt-adsk: Convenience method.

lib/ufe/ProxyShapeHierarchyHandler.cpp
```diff
@@ -17,15 +17,15 @@
...
+#include <ufe/ProxyShapeHierarchy.h>
+
 MAYAUSD_NS_DEF {
 namespace ufe {
 
 ProxyShapeHierarchyHandler::ProxyShapeHierarchyHandler(Ufe::HierarchyHandler::Ptr mayaHierarchyHandler)
 	: Ufe::HierarchyHandler()
 	, fMayaHierarchyHandler(mayaHierarchyHandler)
-{
-	fProxyShapeHierarchy = ProxyShapeHierarchy::create(mayaHierarchyHandler);
```
ppt-adsk: ProxyShapeHierarchy stores a UFE scene item, so best create it as needed, otherwise we'd need observation to keep the scene item's path up to date, which gratuitous complexity.

lib/ufe/ProxyShapeHierarchyHandler.cpp
```diff
@@ -45,8 +45,7 @@ Ufe::Hierarchy::Ptr ProxyShapeHierarchyHandler::hierarchy(const Ufe::SceneItem::
 {
 	if (isAGatewayType(item->nodeType()))
 	{
-		fProxyShapeHierarchy->setItem(item);
-		return fProxyShapeHierarchy;
+		return ProxyShapeHierarchy::create(fMayaHierarchyHandler, item);
```
ppt-adsk: Create the ProxyShapeHierarchy on demand.

lib/ufe/UsdStageMap.h
```diff
@@ -22,22 +22,37 @@
...
-	Map of AL_usdmaya_ProxyShape UFE path to corresponding stage.
-
-	Map of stage to corresponding AL_usdmaya_ProxyShape UFE path.  Ideally, we
-	would support dynamically computing the path for the AL_usdmaya_ProxyShape
-	node, but we assume here it will not be reparented.  We will also assume that
-	a USD stage will not be instanced (even though nothing in the data model
-	prevents it).
+	Two-way map of proxy shape UFE path to corresponding stage.
+
+	We will assume that	a USD proxy shape will not be instanced (even though
```
ppt-adsk: Previous USD stage map was path-based, which was impossible to maintain in the presence of path changes through renaming.  Simply use the proxy shape MObject instead.

lib/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -62,32 +64,56 @@ UsdSceneItem::Ptr UsdUndoRenameCommand::renamedItem() const
 	return fUfeDstItem;
 }
 
-bool UsdUndoRenameCommand::rename(SdfLayerHandle layer, const Ufe::Path& ufeSrcPath, const SdfPath& usdSrcPath, const SdfPath& usdDstPath)
+bool UsdUndoRenameCommand::renameRedo()
```
ppt-adsk: Re-implemented USD renaming.  We now SdfCopySpec the source to the destination, in the same layer, and inactivate the source, on do / redo.  On undo, we RemovePrim the destination, and activate the source.  This is arguably a naïve implementation, but it's a start.

Unfortunately, undo fails, because RemovePrim fails, not clear why at this point, will continue investigation.

test/lib/ufe/testRename.py
```diff
@@ -0,0 +1,155 @@
...
+        # MAYA-92350: should not need to re-bind hierarchy interface objects
+        # with their item.
+        propsHierarchy = ufe.Hierarchy.hierarchy(propsItem)
+        propsChildren = propsHierarchy.children()
+
+        self.assertEqual(len(propsChildren), len(propsChildrenPre))
+        self.assertIn(ball35RenItem, propsChildren)
+
+        # Rename undo / redo does not work yet.  PPT, 28-Jan-2020.
+        return
```
ppt-adsk: Not testing undo / redo at this point, because it fails as per comments above.

lib/ufe/Global.cpp
```diff
@@ -22,7 +22,6 @@
 #include "UsdTransform3dHandler.h"
 #include "UsdSceneItemOpsHandler.h"
 
-#include <ufe/rtid.h>
```
mattyjams: Looks like we're actually adding a usage of `Ufe::Rtid` at line 130, so I'm not sure why we're taking this include out? I know it was added to the header, but since there's usage in both files, I'd prefer to see the include in both files.

lib/ufe/ProxyShapeHierarchy.cpp
```diff
@@ -95,8 +109,10 @@ Ufe::SceneItem::Ptr ProxyShapeHierarchy::sceneItem() const
 bool ProxyShapeHierarchy::hasChildren() const
 {
 	const UsdPrim& rootPrim = getUsdRootPrim();
-	if (!rootPrim.IsValid())
+	if (!rootPrim.IsValid()) {
+        UFE_LOG("invalid root prim in ProxyShapeHierarchy::hasChildren()");
 		return false;
+    }
```
mattyjams: Looks like maybe a space missing before the brace?

lib/ufe/ProxyShapeHierarchy.h
```diff
@@ -68,9 +72,9 @@ class MAYAUSD_CORE_PUBLIC ProxyShapeHierarchy : public Ufe::Hierarchy
 	const UsdPrim& getUsdRootPrim() const;
 
 private:
-	Ufe::SceneItem::Ptr fItem;
-	Hierarchy::Ptr fMayaHierarchy;
-	Ufe::HierarchyHandler::Ptr fMayaHierarchyHandler;
+	Ufe::SceneItem::Ptr fItem{nullptr};
+	Hierarchy::Ptr fMayaHierarchy{nullptr};
+	Ufe::HierarchyHandler::Ptr fMayaHierarchyHandler{nullptr};
```
mattyjams: Is the explicit `nullptr` initialization here needed for some reason? It looks like these are `std::shared_ptr` so shouldn't they be empty by default?

lib/ufe/UsdStageMap.cpp
```diff
@@ -15,6 +15,43 @@
...
+    auto dagPath = MayaUsd::ufe::nameToDagPath(noWorld);
+    MObjectHandle handle(dagPath.node());
+    if (!handle.isValid()) {
+        TF_CODING_ERROR("'%s' is not a path to a proxy shape node.",
+                        noWorld.c_str());
+    }
+    return handle;
+}
+
+// Assuming proxy shape nodes cannot be instanced, simply return the first path.
```
mattyjams: I'll admit I'm not sure I fully understand the context here, but is this a safe assumption? In our pipeline, we definitely do instance `pxrUsdReferenceAssembly`/`pxrUsdProxyShape` nodes using either nParticles or MASH.

lib/ufe/UsdStageMap.cpp
```diff
@@ -31,32 +68,52 @@ UsdStageMap g_StageMap;
...
+    fObjectToStage[proxyShape] = stage;
+    fStageToObject[stage] = proxyShape;
 }
 
 UsdStageWeakPtr UsdStageMap::stage(const Ufe::Path& path) const
 {
+    auto proxyShape = proxyShapeHandle(path);
+    if (!proxyShape.isValid()) {
+        return nullptr;
+    }
```
mattyjams: Looks like maybe a leading space missing from this block to align it with the code below.

lib/ufe/UsdStageMap.h
```diff
@@ -22,22 +22,37 @@
...
 #include <unordered_map>
 
+// Allow for use of MObjectHandle with std::unordered_map.
+namespace std {
+template <> struct hash<MObjectHandle> {
+    std::size_t operator()(const MObjectHandle& handle) const {
+      return static_cast<std::size_t>(handle.hashCode());
+    }
+};
+}
```
mattyjams: This would be a great candidate for the kind of Maya-as-the-only-dependency utilities library that Animal Logic has been advocating for. We actually had something for this in usdMaya (now migrated to mayaUsd) as well:
https://github.com/Autodesk/maya-usd/blob/31e6f34afd6df370eed6bf4daab77ad420b98c54/lib/utils/util.h#L105

There's nothing USD or UFE related about either implementation.

lib/ufe/UsdStageMap.h
mattyjams: Again, slightly concerned about the no instancing assumption, but maybe I'm misunderstanding.

lib/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -62,32 +64,62 @@ UsdSceneItem::Ptr UsdUndoRenameCommand::renamedItem() const
...
+
+bool UsdUndoRenameCommand::renameUndo()
+{
+    // Regardless of where the edit target is currently set, switch to the
+    // layer where we copied the source prim into the destination, then
+    // restore the edit target.
+    auto editTarget = fStage->GetEditTarget();
+    fStage->SetEditTarget(fLayer);
+    bool status = fStage->RemovePrim(fUsdDstPath);
+    fStage->SetEditTarget(editTarget);
```
mattyjams: Using a `UsdEditContext` might help simplify this a bit:
https://github.com/PixarAnimationStudios/USD/blob/be1a80f8cb91133ac75e1fc2a2e1832cd10d91c8/pxr/usd/usd/editContext.h#L40

lib/ufe/wrapUtils.cpp
```diff
@@ -18,13 +18,42 @@
...
+        }
+    }
+    else {
+        names.push_back(src);
+    }
+    return names;
+}
+
+}
+
```
mattyjams: Could we not just use `TfStringTokenize()` below instead?:
https://github.com/PixarAnimationStudios/USD/blob/be1a80f8cb91133ac75e1fc2a2e1832cd10d91c8/pxr/base/tf/stringUtils.h#L424

lib/ufe/UsdUndoRenameCommand.cpp
ppt-adsk: Thanks to @mattyjams , undo / redo works!

test/lib/ufe/testRename.py
ppt-adsk: Undo / redo now works.

lib/ufe/UsdStageMap.cpp
mattyjams: For some additional context, the scene file for the `testPxrUsdMayaGLInstancerDraw` unit test is an example of an instancer setup:
https://github.com/Autodesk/maya-usd/blob/dev/plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/InstancerDrawTest/InstancerDrawTest.ma
https://github.com/Autodesk/maya-usd/blob/dev/plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testPxrUsdMayaGLInstancerDraw.py

ppt-adsk: Created an issue describing the problem, and some areas that would need to be worked on:
https://github.com/Autodesk/maya-usd/issues/256

lib/ufe/Global.cpp
ppt-adsk: Not a huge deal, to be sure, but isn't that violating Don't Repeat Yourself?  If maintenance removes the use of Ufe::Rtid, I'd now have two places to remove the include of ufe/rtid.h.

lib/ufe/UsdStageMap.h
ppt-adsk: No, you're not misunderstanding...  As per other comment entered as
https://github.com/Autodesk/maya-usd/issues/256

lib/ufe/ProxyShapeHierarchy.cpp
ppt-adsk: Tabs vs spaced, tabified in
https://github.com/Autodesk/maya-usd/pull/237/commits/f53f29c7258bc1d99897d843dbfef2e0316dd702

lib/ufe/ProxyShapeHierarchy.h
ppt-adsk: Reverted in
https://github.com/Autodesk/maya-usd/pull/237/commits/f53f29c7258bc1d99897d843dbfef2e0316dd702

lib/ufe/UsdStageMap.cpp
ppt-adsk: Tabified in
https://github.com/Autodesk/maya-usd/pull/237/commits/f53f29c7258bc1d99897d843dbfef2e0316dd702

lib/ufe/UsdUndoRenameCommand.cpp
ppt-adsk: Excellent suggestion, and even better, exception-safe (if RemovePrim() throws).  Fixed in
https://github.com/Autodesk/maya-usd/pull/237/commits/f53f29c7258bc1d99897d843dbfef2e0316dd702

lib/ufe/wrapUtils.cpp
ppt-adsk: Great suggestion, done in
https://github.com/Autodesk/maya-usd/pull/237/commits/f53f29c7258bc1d99897d843dbfef2e0316dd702

kxl-adsk:
@mattyjams you already approved the change, but there are new commits implementing your code review suggestions. Can you re-approve once reviewed (or simply let me know that you are done with the review).

HamedSabri-adsk:
PS: `UFE 2.5.6 or later` is required for this branch.

lib/ufe/UsdStageMap.cpp
```diff
@@ -31,32 +68,52 @@ UsdStageMap g_StageMap;
 
 void UsdStageMap::addItem(const Ufe::Path& path, UsdStageWeakPtr stage)
 {
-	fPathToStage[path] = stage;
-	fStageToPath[stage] = path;
+	// We expect a path to the proxy shape node, therefore a single segment.
+	auto nbSegments = path.nbSegments();
```
HamedSabri-adsk: One minor thing that I just noticed, It doesn't look like we are guarding UsdStageMap.cpp against UFE1.0.  

See https://github.com/Autodesk/maya-usd/blob/f53f29c7258bc1d99897d843dbfef2e0316dd702/lib/CMakeLists.txt#L214

ppt-adsk: Yup, I think that's fine: UsdStageMap is required for proper UFE v1 use, and it doesn't depend on any UFE v2 features, so it should only be protected for UFE compilation, which it currently is, not against UFE v2.

HamedSabri-adsk: Makes sense. Thanks

ppt-adsk: As discussed, completely missed your point, great catch!  Fixed in
https://github.com/Autodesk/maya-usd/pull/237/commits/e508bdb626a80dcf58d08e17c85867514a36dd5b

lib/ufe/UsdStageMap.cpp
```diff
@@ -69,7 +69,12 @@ UsdStageMap g_StageMap;
 void UsdStageMap::addItem(const Ufe::Path& path, UsdStageWeakPtr stage)
 {
 	// We expect a path to the proxy shape node, therefore a single segment.
-	auto nbSegments = path.nbSegments();
+	auto nbSegments = 
+#ifdef UFE_V2_FEATURES_AVAILABLE
```
HamedSabri-adsk: lgtm. Thanks Pierre.

PR #240, opened 2020-01-31
created a test suite for all the test cases in TestUSDMayaPython and …

HamedSabri-adsk:
…should make it easier for future tests to be added in there.


elrond79:
Hi Hamed - apologies, it seems like we may have just done some overlapping work.  While writing a new test, I decided to try and reduce the boilerplate necessary, and ended up with a slightly different solution that looks like this:

https://github.com/LumaPictures/maya-usd/blob/pr/base/pytest_cmake_function/plugin/al/plugin/AL_USDMayaTestPlugin/py/CMakeLists.txt

ie, I made it simpler to add python tests, so I just broke up the combined tests into 3 separate tests.

HamedSabri-adsk:
closing this. PR242 covers this.

PR #243, opened 2020-02-01
[build] missing space meant CMAKE_MACOSX_RPATH wasn't being properly set

elrond79:
Title says it all...

### DIFF
--- a/build.py
+++ b/build.py
@@ -250,7 +250,7 @@ def RunCMake(context, extraArgs=None, stages=None):
                 'cmake '
                 '-DCMAKE_INSTALL_PREFIX="{instDir}" '
                 '-DCMAKE_BUILD_TYPE={variant} '
-                '-DCMAKE_EXPORT_COMPILE_COMMANDS=ON'
+                '-DCMAKE_EXPORT_COMPILE_COMMANDS=ON '
                 '{osx_rpath} '
                 '{generator} '
                 '{extraArgs} '
--- a/build.py
+++ b/build.py
@@ -232,11 +232,6 @@ def RunCMake(context, extraArgs=None, stages=None):
     if generator is not None:
         generator = '-G "{gen}"'.format(gen=generator)
 
-    # On MacOS, enable the use of @rpath for relocatable builds.
-    osx_rpath = None
-    if MacOS():
-        osx_rpath = "-DCMAKE_MACOSX_RPATH=ON"
-
     # get build variant 
     variant= BuildVariant(context)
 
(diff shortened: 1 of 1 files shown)


HamedSabri-adsk:
Not really sure if I understand the relationships here.

kxl-adsk:
@elrond79 We haven't experienced any rpath issues on mac. How can the issue be reproduced?

elrond79:
Well, I don’t know if there were rpath issues or not.  All I could verify
is that the cmake flag wasn’t being set properly, because it was
concatenated with the previous option, because it was missing a following
space.

Perhaps the option isn’t needed at all?

You should be able to reproduce on any Mac build. Did you look at the diff?

kxl-adsk:
There is something weird going on here. The `CMAKE_MACOSX_RPATH` is needed to have rpaths working on MacOS and they are even without this change. 

The information provided in your [comment](https://github.com/Autodesk/maya-usd/pull/243#issuecomment-581104371) was important since it was easy to assume you tested this on MaxOS and that's how you found the problem. For future PRs - please include more information about tests you did and repro steps.

elrond79:
Well, now that I look at it again, I agree with the general sentiment that the title + description here were not descriptive enough... so I apologize that it created confusion.  I have a tendency sometimes dash off very quick / not descriptive PRs for typo fixes like this.

We can make it the policy to always state whether or not tests have been done... however, I do feel that there needs to be the possibility for submitting some PRs based on code review alone - particularly for one-character typos like this one.

Now, re: this particular issue - in this case, I DID notice the problem as the result of attempting an actual MacOS build.  I happened to notice it in the print-out of my cmake command line, and have since verified that things are messed up by inspecting my CMakeCache.txt:

```
CMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON-DCMAKE_MACOSX_RPATH=ON
```

(I also verified that `CMAKE_MACOSX_RPATH` is not being set.)  I wasn't able to verify whether there were rpath issues - because I haven't yet gotten to the point of having a completed, error-free build on MacOSX.  (Which is an unrelated issue... I think it's somehow picking up an older version of USD, but haven't had the chance to really dig in yet.)

I'm not familiar with CMAKE_MACOSX_RPATH, but looking at the docs, my best guess is that it's no longer needed since the minimum cmake version got bumped.  According to this:

https://cmake.org/cmake/help/latest/policy/CMP0042.html#policy:CMP0042

...now that we're using 3.0, it should be on by default.

elrond79:
Shall I amend the PR to simply drop the option altogether?

HamedSabri-adsk:
Yes we can drop the option, now that cmake finally has decided to support this flag out of the box.

elrond79:
Ok, updated to remove the osx_rpath stuff altogether.

elrond79:
@kxl-adsk - oops, noticed I pushed this within seconds of you adding your approval - just wanted to make sure you saw this!

kxl-adsk:
Yep, I saw.

PR #245, opened 2020-02-04
[al] Performance improvements when processing prim meta data

robthebloke:
Previously we a mishmash of code buried within the ProxyShape that would handle 3 separate meta data tags we could apply to a prim within a scene. These are:

## al_usdmaya_excludeFromProxyShape 
If this boolean value is true, the prim would be added to an excluded geometry list that we then pass to UsdImaging to disable the rendering of these prims within Hydra. 

## al_usdmaya_lock
This meta data tag affects the transform chains created when importing maya nodes via either the translator framework, or the translate prim command. It controls whether the transform attributes _(scale, rotate, etc)_ are locked after creation. There are 3 possible values for this meta data:
* inherited : This is the default. If set, the transform will be locked or unlocked based on the lock status of it's parent nodes. If the prim has no parent, then the default state is unlocked. 
* transform : A poor choice of name, a better name would be "locked". If this is set, the transform _(and any child transforms which are set to inherited)_ will have their transform attributes locked. 
* unlocked : If set, the transform attributes of this node _(and any child transforms with the 'inherited' lock status)_ will be unlocked and free to be modified. 

## al_usdmaya_selectability
This meta data tag controls whether it's possible to select a prim _(and it's children)_ within the Maya viewport. We use this at AL to say, prevent animators from accidentally selecting the environments whilst animating. This can have two values:
* selectable : This is the default. If set, the prim can be selected in the maya viewport. 
* unselectable : If set, the prim _(and all child prims)_ will be unselectable in the viewport.

## Problems this PR addreses

The motivating factor for the PR was that we found the previous code used to handle the meta data did not scale to very large set sizes. This PR primarily addresses those performance problems,  has fixed a few bugs and inconsistencies along the way, and additional tests have been added. In general, these are the main aims:

* Removed the performance bottlenecks that occurred after a variant switch when the lock and selectability states needed to be updated. 
* The TranslatePrim command did not respect the meta data tags. 
* Ensured that changes to the meta data on a prim within a proxy shape, will correctly be reflected within Maya. 

## Code changes 

* Previously the code used to build up and modify the internal meta data lists the was generalised behind the HierarchyLogics interface. Sadly there are subtle differences in the way each of those meta data items need to be handled, and so generalising the code simply lead to a fairly inefficient process needed to update and maintain the internal databases. We've removed the base class, and have split the three meta data tags into their own classes _(with an eye on maximising performance of each one)_
* Added the optionVar **AL_usdmaya_ignoreLockPrims**, to disable all of the above features _(If you are not using the above functionality, I'd suggest enabling the optionvar - it somewhat improves performance!)_
* Added a new LockManager to maintain the locked/unlocked status of prims. We went through a number of iterations of this class, eventually settling on a version that stores two SdfPath arrays: _locked_ and _unlocked_. These lists are sorted internally so that we can quickly determine any lock prim state quickly with 2x std::lower_bound calls, and then a simple check to see which path is the longest _(i.e. the nearest parent)_. 
* Added a 3rd argument for the TransformIterator which allows it to stop iterating into an instance _(we can't select prims within a master prim, only the highest instance level, so there isn't any point us recursing further down the tree)_. By default the iterator will iterate into instances _(the same behaviour as before)_
* Changed the way that paths were being stored within the SelectabilityDB to improve the performance of querying and modifying the list of paths. 
* Added extra tests to cover some of the edge cases we found during testing.


plugin/al/lib/AL_USDMaya/AL/usdmaya/SelectabilityDB.h
```diff
@@ -28,23 +29,60 @@ namespace usdmaya {
...
+  ///-------------------------------------------------------------------------------------------------------------------
+  /// \brief  Determines whether there is an internal entry for the path specified (and only this path!). If you wish 
+  ///         to determine selectability, call isPathUnselectable instead.
+  /// \param  path the path to check to see if exists
+  /// \return true if the path is contained, false if not.
+  ///-------------------------------------------------------------------------------------------------------------------
+  bool containsPath(const SdfPath& path) const
+  {
+    auto foundPathEntry = std::lower_bound(m_unselectablePaths.begin(), m_unselectablePaths.end(), path);
+    if(foundPathEntry != m_unselectablePaths.end() && path == *foundPathEntry)
```
ppt-adsk: Nit-pick: could just return the result of the boolean expression and avoid the conditional altogether.

return (foundPathEntry != m_unselectablePaths.end() && path == *foundPathEntry);

Fine as it is, of course.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
```diff
@@ -77,6 +77,16 @@ typedef void (*proxy_function_prototype)(void* userData, AL::usdmaya::nodes::Pro
 
 const char* ProxyShape::s_selectionMaskName = "al_ProxyShape";
 
+//----------------------------------------------------------------------------------------------------------------------
+MDagPath ProxyShape::parentTransform()
```
ppt-adsk: You sure you really need this?  You're inheriting from MayaUsdProxyShapeBase, which has the exact same implementation of parentTransform():

https://github.com/Autodesk/maya-usd/blob/dev/lib/nodes/proxyShapeBase.cpp#L921

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.h
```diff
@@ -1061,19 +1018,26 @@ class ProxyShape
   std::string generateTranslatorId(UsdPrim prim) override
    { return m_translatorManufacture.generateTranslatorId(prim); }
 
+public:
+  bool isLockPrimFeatureActive() const
+  {
+    bool ignoreLockPrims = MGlobal::optionVarIntValue("AL_usdmaya_ignoreLockPrims");
+    //The lock Prim functionality is a UI thing - no need to have it on in batch mode
+    //However, this also causes the tests to fail which is bad
```
ppt-adsk: Why do the tests fail?  What's the interaction?

plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/TransformIterator.h
```diff
@@ -128,14 +128,15 @@ class TransformIterator
 
   std::vector<StackRef> m_primStack;
   UsdStageRefPtr m_stage;
+  size_t m_currentItem;
```
kxl-adsk: /git/usd/maya-usd/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/TransformIterator.h:131:10: error: private field 'm_currentItem' is not used [-Werror,-Wunused-private-field]
  size_t m_currentItem;

As well, for variables like this (and m_stopOnInstance) below, use in-class member initializers instead (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-default)

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.h
robthebloke: When we run the tests on our build server, ````MGlobal::mayaState() == MGlobal::kInteractive``` will return false, which would end up disabling the meta-data processing entirely. Not overly useful when you've written tests to test said functionality.

kxl-adsk:
@robthebloke please resolve conflicts.

robthebloke:
It is not required to build the plug in, but it is required to load the plug in successfully.

kxl-adsk:
It’s not something we observed, nor we have any issues logged for it. Additionally, most of Pixar tests rely on this plugin, so I really don’t see how we could have missed it. 

Given that this change doesn’t belong to PR you have open, let’s not sneak changes in hidden behind “merge from dev”.

kxl-adsk:
@robthebloke Let's get this PR in by removing the unrelated change which causes compilation problems. Then please open a new PR with only this new change and description of the problem it fixes.

PR #246, opened 2020-02-04
[al] Check transform values prior to writing to USD. (#993)

robthebloke:
This PR aims to address two issues when writing data from the Transform nodes to USD (after a TRS tool, or attribute change). 

1. If the incoming transform values match the original values in USD, do not write the values to USD. This is to avoid accidentally inserting overs to layers accidentally (such as the session layer). 
2. Prevent the transform values being written to UsdTimeCode::Default, if the attribute in question has time samples. This is to prevent cases where accidentally writing a default value to the session layer _(for example)_, would create an over that hides all of the underlying animation data.


plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/TransformationMatrix.cpp
```diff
@@ -43,6 +43,21 @@ MPxTransformationMatrix* TransformationMatrix::creator()
   return new TransformationMatrix;
 }
 
+//----------------------------------------------------------------------------------------------------------------------
+bool hasEmptyDefaultValue(const UsdGeomXformOp& op, UsdTimeCode time)
```
fowlertADSK: I'm not sure about the actual c++ rule but I always prefer to be explicit and wrap these local functions in an unnamed namespace{}.

Was also thinking that we could just make _hasEmptyDefaultValue_ do everything you have below.  It's only used in that exact same chunk of code in the 5 or 6 places.  Why not just have it also check "timeCode.IsDefault() && op.GetNumTimeSamples()"

PR #248, opened 2020-02-04
[al] added generate unique key method for translators (#1005)

robthebloke:
Typically when using a custom translator, the following action occurs on a variant switch:

If a prim that had previously been translated into Maya, remains in the stage after a variant switch, and that prim supports update, then the translator's update method will be called. In some cases this update step may be entirely superfluous _(because none of the data changed)_. 

For some of the translators we use internally, there are cases where the cost of updating can be very large, so this PR adds a way to optionally generate a hash for a translated prim, which can be used again later on to determine whether update should be called or not. 

Within a translator, you would override the generateUniqueKey method, to return a hashed value for a specific prim.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.cpp
@@ -19,6 +19,8 @@
 #include "maya/MSelectionList.h"
 #include "maya/MFnDagNode.h"
 
+#include <string>
+
 namespace AL {
 namespace usdmaya {
 namespace fileio {
@@ -390,6 +392,10 @@ MString TranslatorContext::serialise() const
     {
       oss << "," << getNodeName(it.createdNodes()[i].object());
     }
+    if (it.uniqueKey())
+    {
+      oss << ",uniquekey:" << it.uniqueKey();
+    }
     oss << ";";
   }
   return MString(oss.str().c_str());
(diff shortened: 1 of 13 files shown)


plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.cpp
```diff
@@ -419,8 +425,27 @@ void TranslatorContext::deserialise(const MString& string)
 
     PrimLookup lookup(SdfPath(strings2[0].asChar()), strings3[0].asChar(), obj);
 
+    static const MString uniqueKeyPrefix("uniquekey:");
+
     for(uint32_t j = 2; j < strings3.length(); ++j)
     {
+      if (strings3[j].substring(0, 10) == uniqueKeyPrefix)
+      {
+        auto keyStr(strings3[j].substring(10, strings3[j].length()));
```
pilarmolinalopez: it may be a good idea to have 10 be defined as a variable with a descriptive name (maybe uniqueKeyPrefixLength) instead of being hardcoded twice here

plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.cpp
```diff
@@ -534,6 +559,49 @@ void TranslatorContext::removeEntries(const SdfPathVector& itemsToRemove)
   AL_MAYA_CHECK_ERROR2(status, "failed to remove translator prims.");
 }
 
+//----------------------------------------------------------------------------------------------------------------------
+void TranslatorContext::updateUniqueKeys()
+{
+  auto stage = getUsdStage();
```
pilarmolinalopez: maybe check that stage != nullptr ?

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/proxy/PrimFilter.cpp
```diff
@@ -64,25 +64,52 @@ PrimFilter::PrimFilter(
       // if the type remains the same, and the type supports update
       if(existingTranslatorId == newTranslatorId)
       {
-        if(supportsUpdate)
+        // locate the path and delete from the removed set (we do not want to delete this prim!
```
pilarmolinalopez: close parenthesis at the end of this comment

plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/wrapTranslatorBase.cpp
```diff
@@ -96,6 +97,25 @@ class TranslatorBaseWrapper
...
 
+  std::size_t generateUniqueKey(const UsdPrim& prim) const override
+  {
+    if (Override o = GetOverride("generateUniqueKey"))
+    {
+      auto res = std::function<boost::python::object (const UsdPrim&)>(TfPyCall<boost::python::object>(o))(prim);
+      if (!res) {
+        return 0;
+      }
+      TfPyLock pyLock;
```
pilarmolinalopez: Sorry, I'm not very familiar with TfPyLock... are we using this lock here?

PR #251, opened 2020-02-04
[al] clean up boundary of BasicTransformationMatrix and TransformationMatrix

elrond79:
Similar to the recent scope fixes (with shadowing of m_prim), I noticed that there were some oddities in the division between BasicTransformationMatrix and TransformationMatrix.

Specifically:

- `m_scope` never seemed to be used anywhere
- `m_transformNode` seemed to be unnecessarily shadowed; unshadowing also meant that `setMObject` didn't need to be virtual / repeated
- there seemed to be no need for `Scope::transform()` to be virtual - it's never overridden, and `Transform::getTransMatrix()` is already defined to get a `TransformationMatrix*`


elrond79:
(pushed a small update - made `Scope::transform` inline, similar to `Transform::getTransMatrix()`

elrond79:
No worries - you never catch everything, and code is always evolving.

Thanks for the quick review + merge!

PR #252, opened 2020-02-05
GPLTech experiments

gplsteph:
- Added namespace support for referencing USD files. (not 100% sure it's fully working)
- Added option to control if the geometry should be exported or not, allowing for example to export USD files with just uvs.


kxl-adsk:
@gplsteph Please create a separate PR for each change (you can remove the namespace change from this PR and open a new one; make sure to document the reason why you made the change and what you think is remaining). Each PR will as well need a test.

kxl-adsk:
As per offline conversation, closing this PR since it's not a complete implementation.

Issue #32, opened 2019-08-28
where did the AL MayaReference go?


DanEnglesson:
Hi,

`from AL.usd.schemas.maya import MayaReference` does not work anymore and looking at all the `lib/python` libraries that is included in the `PYTHONPATH` it's not included there. It seems to have been moved to `maya-usd/lib` which is not added to the PYTHONPATH; I might be missing something here but from what I can see the MayaReference is never added to the PYTHONPATH envar in the mayaUSD.mod file. 

Cheers,
Dan

DanEnglesson:
I should mentioned this is for the refactoring_sandbox. The master branch has the MayaReference in the AL plugin still.

kxl-adsk:
Correct, it was moved as part of plugin merge. Now every plugin should be able to import USD file with Maya reference prims. But as you have discovered, we haven't yet finished python binding part of it. We are now in process of moving python bindings out of Pixar plugin into the core library and then we will add missing Maya Reference.

kxl-adsk:
Logged as  MAYA-100476

DanEnglesson:
OK, good to know. Thanks for the info.

rhaleblian:
Yes, thanks for that info on intent. My shop was using that API, it's a current showstopper for migration to maya-usd away from the AL and Pixar sources.

kxl-adsk:
You can still migrate to the dev branch since the move of Maya reference schema was only done in the refactoring_sandbox branch.

DanEnglesson:
Yes, as Krystian said, this is only affecting the refactoring_sandbox branch. We have successfully switched to maya-usd and are currently using the master branch. You should be able to switch to either the master or dev branch. 

Thanks for clarifying this Krystian.

rhaleblian:
You guys were clear about the affected branch, i am just fixated on `refactoring_sandbox` as I've used that branch for cherry-picking our site changes. Getting those changes onto `master` instead may be a better way to go.

BigRoy:
Just for full clarity. The intended goal is that the `refactor_sandbox` branch at some point will become the new `master`. As it's the intermediate work of the full merge of the separare plug-ins. Correct?

As such, would he AL and Pixar side of plug-in nodes then get deprecated in favor of `maya-usd`'s own loader and proxy shapes? Or what exactly is the goal?

kxl-adsk:
[...]

PR #255, opened 2020-02-06
Backport MayaReference schema refactoring from refactoring_sandbox

kxl-adsk:
This is the last change which we have done in refactoring_sandbox. It moves MayaReference schema out of the AnimalLogic plugin and makes it part of our core modules. It also adds a reader for MayaReference allowing us to import files with MayaReference schema.

This change contains the original backported code from refactoring_sandbox + number of fixes, python bindings (issue #32) and tests that were not present in refactoring_sandbox.


kxl-adsk:
@murphyeoin There is one part of this change I still don't like and would like to hear your opinion. I made a very limited amount of renaming here...but not sure if this is the right way to go. At the end, I would like to rename the schema class to be MayaReference instead of ALMayaReference. How should we proceed?

murphyeoin:
Hi @kxl-adsk renaming the ALMayaReference -> MayaReference is undoubtedly the right thing to do. We have a copy of the schema in a separate library which we release separately so our USD clients don't all have to depend on AL_USDMaya. Therefore I think all we would need is an #ifdef  anywhere the USD Schema is referred to by name to allow us to keep compatibility (it would be very bad if we weren't able to read our existing USD scenes). We could phase this out over time. Hopefully this might just be in the line where the translator is associated with the specific schema? (I haven't had time to go through the code changes yet so can't point out the specific line(s) )

kxl-adsk:
@murphyeoin as per our offline conversation, I took a stab at trying solution explained here https://groups.google.com/forum/#!msg/usd-interest/Pn5nkKfswN4/vnbnLocjEQAJ.

Inheritance of schemas works, but this doesn't make our life easier for translators. Pixar translation framework (now part of mayaUsd lib) allows me to register the same translator for more than one type. Unfortunately, I can't do that with AL translators, meaning I will have to code the same translator twice.

I can't convince myself that code duplication is where we want to go. So we are left for now with keeping the old name. What others think, is there any other option I'm not seeing? (@mattyjams, @elrond79, @cfmoore007)

lib/fileio/translators/translatorMayaReference.cpp
```diff
@@ -0,0 +1,489 @@
...
+  // Given a function set attached to a DAG node, create a string from the
+  // node's full path by replacing all the path dividers (|) by underscores
+  // and stripping off the first character so the name doesn't start with an
+  // underscore.  (We use this for creating unique names for Maya reference
+  // nodes created from ALUSD proxy nodes.)
+  //
+  MString refNameFromPath(const MFnDagNode &nodeFn)
+  {
+    MString name = nodeFn.fullPathName();
+    name.substitute("|", "_");
```
robthebloke: You are replacing single characters. Calling a method designed for strings is overkill if you ask me. 
```c++
std::replace(name.asChar(), name.asChar() + name.length(), '|', '_');
```

lib/fileio/translators/translatorMayaReference.cpp
```diff
@@ -0,0 +1,489 @@
...
+}
+
+MStatus 
+UsdMayaTranslatorMayaReference::UnloadMayaReference(const MObject& parent){
+    TF_DEBUG(PXRUSDMAYA_TRANSLATORS).Msg("MayaReferenceLogic::UnloadMayaReference\n");
+    MStatus status;
+    MFnDependencyNode fnParent(parent, &status);
+    if (status)
+    {
+        MPlug messagePlug = fnParent.findPlug("message", true, &status);
```
robthebloke: Do we need to look this attribute up by name? Maybe use an MNodeClass upfront to grab the MObject for the attribute, and initialise the MPlug directly (thereby avoiding the need for fnParent completely)

lib/fileio/translators/translatorMayaReference.cpp
```diff
@@ -0,0 +1,489 @@
...
+        if (status)
+        {
+            MPlugArray referencePlugs;
+            messagePlug.connectedTo(referencePlugs, false, true);
+
+            // Unload the connected references.
+            for (uint32_t i = 0; i < referencePlugs.length(); ++i)
```
robthebloke: ```referencePlugs.length()``` is DSO exported, take a local copy. 
```c++
for (uint32_t i = 0, n = referencePlugs.length(); i < n; ++i)
```

lib/fileio/translators/translatorMayaReference.cpp
```diff
@@ -0,0 +1,489 @@
...
+    /*
+       From the Maya docs:
+       > This message attribute is used to connect specific nodes that may be
+       > associated with this reference (i.e. group, locator, annotation). Use of
+       > this connection indicates that the associated nodes have the same
+       > lifespan as the reference, and will be deleted along with the reference
+       > if it is removed.
+     */
+    MStatus result;
+    MPlug destArrayPlug = refNode.findPlug("associatedNode", true);
```
robthebloke: extract the MObject handle for the attribute upfront, to avoid the need for the binary search of string compares. e.g. upfront:

```c++
MNodeClass nc("reference");
UsdMayaTranslatorMayaReference::associatedNodeHandle = nc.attribute("associatedNode");
UsdMayaTranslatorMayaReference::messageHandle = nc.attribute("message");
```
Then later:
```c++
MPlug destArrayPlug(refNode.node(), associatedNodeHandle);
```

mattyjams:
I'm not really familiar with the AL translator code, but my gut instinct is definitely to use inheritance for this as described in that usd-interest thread, where the current `ALMayaReference` schema is moved/renamed to `MayaReference` and a "new" `ALMayaReference` takes its place and simply inherits from `MayaReference`.
Maybe a naive/dumb question, but what's the thing forcing the duplication of the code per schema, and/or could you point me at the code that would need to be duplicated? I would have thought that the bulk of the translator code could be factored out and amended to operate on `MayaReference` schema prims, which would then allow it to work with both schemas, since `ALMayaReference` `IsA()` `MayaReference`.

kxl-adsk:
@mattyjams Created temporary branch with my test - see commit b20949aff9a79. Interesting places are https://github.com/Autodesk/maya-usd/blob/b20949aff9a79d4c0b22729449724cca57ebcfc1/plugin/al/translators/MayaReference.cpp#L50 and https://github.com/Autodesk/maya-usd/blob/b20949aff9a79d4c0b22729449724cca57ebcfc1/lib/usd/translators/mayaReferenceReader.cpp#L58

kxl-adsk:
@robthebloke Thank you for your comments. This is all refactored code (not a new code), but I'm happy to improve the code once we figure out the naming issue.

elrond79:
> @murphyeoin as per our offline conversation, I took a stab at trying solution explained here https://groups.google.com/forum/#!msg/usd-interest/Pn5nkKfswN4/vnbnLocjEQAJ.
> 
> Inheritance of schemas works, but this doesn't make our life easier for translators. Pixar translation framework (now part of mayaUsd lib) allows me to register the same translator for more than one type. Unfortunately, I can't do that with AL translators, meaning I will have to code the same translator twice.
> 
> I can't convince myself that code duplication is where we want to go. So we are left for now with keeping the old name. What others think, is there any other option I'm not seeing? (@mattyjams, @elrond79, @cfmoore007)

Could we essentially do the same trick that we did with the schema for the translator itself? ie, make a another C++ class which inherits from the desired translator, and just changes some methods / attributes to return a different node name?

If that doesn't make sense - apologies, I haven't actually looked at the relevant code yet, I'm just spitballing here...

elrond79:
Hmm - yeah, so what happens if in the .h we just do:

```
class ALMayaReference : public MayaReference
{
public:
  AL_USDMAYA_DECLARE_TRANSLATOR(ALMayaReference);
}
```

and in the .cpp, just:

```
AL_USDMAYA_DEFINE_TRANSLATOR(ALMayaReference, MayaUsd_SchemasALMayaReference)
```

kxl-adsk:
My hope was that we won't have to code against two schemas, at least this is how I interpreted 

> That allowed all code to immediately be changed over to talking about Material, because any authored Look prim will identify itself as a Material via UsdPrim::IsA()

There are ways to reduce the amount of code duplication (I haven't tried what you suggested yet, @elrond79)...but that doesn't change the fact that we may run into cases (like with translator registry) where we will have to code things for both schemas to get them working. Is this the price to pay for rename or there is something we are not seeing?

kxl-adsk:
OK. No matter what we will do, one day we will have to rename the core schema...so I decided to do the rename now AND go with the inheritance for both AL schema and AL translator. See ab21d54ef.

I would love to have @robthebloke or @murphyeoin review this PR (we have some details pointed out, but since they are an improvement to existing code, I first look for feedback about the move and rename). If anyone else can give it a try, please go ahead.

FWIW. Once this change is in, we are done with the refactoring_sandbox branch.

cfmoore007:
I'm getting c-test failures with the latest dev branch + this PR ( using USD 20.02 )  Windows 10, 
The following tests FAILED:
          3 - testMayaUsdSchemasMayaReference (Failed)
         84 - AL_USDMayaTestPlugin (Failed)
         88 - TestAdditionalTranslators (Failed)
         89 - TestPxrUsdTranslators (Failed)
Errors while running CTest

3)
test 3
    Start 3: testMayaUsdSchemasMayaReference

3: Test command: D:\APPLICATIONS\Autodesk\Maya2021\bin\mayapy.exe "-c" "from unittest import main;                                           import maya.standalone;                                           maya.standalone.initialize(name='python');                                           import testMayaUsdSchemasMayaReference;                                           main(module=testMayaUsdSchemasMayaReference);                                           maya.standalone.uninitialize()"
3: Environment variables:
3:  PATH=D:/APPLICATIONS/MayaUSD/dev-990974a-pr255-rman23.1/lib;D:\APPLICATIONS\Autodesk\Maya2021/bin;C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\bin\HostX86\x86;C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCPackages;C:\Program Files (x86)\Microsoft SDKs\TypeScript\3.1;C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\CommonExtensions\Microsoft\TestWindow;C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\CommonExtensions\Microsoft\TeamFoundation\Team Explorer;C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\bin\Roslyn;C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Team Tools\Performance Tools;C:\Program Files (x86)\Microsoft Visual Studio\Shared\Common\VSPerfCollectionTools\;C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6.1 Tools\;C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x86;C:\Program Files (x86)\Windows Kits\10\bin\x86;C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\\MSBuild\15.0\bin;C:\WINDOWS\Microsoft.NET\Framework\v4.0.30319;C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\;C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\Tools\;C:\Program Files (x86)\Common Files\Oracle\Java\javapath;C:\WINDOWS\system32;C:\WINDOWS;C:\WINDOWS\System32\Wbem;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;C:\Program Files (x86)\GtkSharp\2.12\bin;C:\WINDOWS\System32\OpenSSH\;D:\APPLICATIONS\Python\Python27;D:\APPLICATIONS\Python\Python27\Scripts;D:\APPLICATIONS\CMake\bin;C:\Users\g8kee\AppData\Local\bin\NASM;C:\Program Files\Microsoft SQL Server\130\Tools\Binn\;D:\APPLICATIONS\QuickTime\QTSystem\;C:\Program Files\dotnet\;C:\Program Files\Git\cmd;C:\Users\g8kee\.dnx\bin;C:\Program Files\Microsoft DNX\Dnvm\;D:\APPLICATIONS\USD\v20.02-rman23\bin;D:\APPLICATIONS\USD\v20.02-rman23\lib;;C:\Program Files\NVIDIA Corporation\NVIDIA NvDLISR;C:\Program Files (x86)\NVIDIA Corporation\PhysX\Common;C:\Users\g8kee\.dnx\bin;C:\Users\g8kee\AppData\Local\Microsoft\WindowsApps;C:\Program Files\Pixar\RenderManProServer-23.1\\bin;C:\Users\g8kee\AppData\Local\Microsoft\WindowsApps;C:\Users\g8kee\AppData\Local\GitHubDesktop\bin;D:\APPLICATIONS\JetBrains\PyCharm Community Edition 2019.3.1\bin;;C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin;C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\Ninja;D:\APPLICATIONS\USD\v20.02-rman23.1\bin;D:\APPLICATIONS\USD\v20.02-rman23.1\lib
3:  PYTHONPATH=D:/APPLICATIONS/MayaUSD/dev-990974a-pr255-rman23.1/lib/python;D:\APPLICATIONS\USD\v20.02-rman23.1\lib\python;D:\APPLICATIONS\USD\v20.02-rman23.1\lib\python
3:  MAYA_NO_STANDALONE_ATEXIT=1
3: Test timeout computed to be: 10000000
3: Warning: At initialization event does not appear in the list of available events
3: Warning: At initialization event does not appear in the list of available events
3: Warning: At initialization event does not appear in the list of available events
3: Warning: At initialization event does not appear in the list of available events
3: E
3: ======================================================================
3: ERROR: testMayaReferenceAttributes (testMayaUsdSchemasMayaReference.MayaUsdSchemasMayaReferenceTestCase)
3: ----------------------------------------------------------------------
3: Traceback (most recent call last):
3:   File "testMayaUsdSchemasMayaReference.py", line 30, in testMayaReferenceAttributes
3:     layer = Sdf.Layer.CreateAnonymous()
3: ErrorException:
3:      Error in 'pxrInternal_v0_20__pxrReserved__::TfType::AddAlias' at line 1166 in file D:\Repos\USD\pxr\base\tf\type.cpp : 'Cannot set alias 'ALMayaReference' under 'UsdSchemaBase', because it is already set to 'MayaUsd_SchemasALMayaReference', not 'AL_usd_MayaReference'.'
3:
3: ----------------------------------------------------------------------
3: Ran 1 test in 0.006s
3:
3: FAILED (errors=1)
1/1 Test #3: testMayaUsdSchemasMayaReference ...***Failed    4.78 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =   4.94 sec

The following tests FAILED:
          3 - testMayaUsdSchemasMayaReference (Failed)
Errors while running CTest

kxl-adsk:
I run tests on both Windows and MacOS before pushing latest commits - all were passing. Can you:
- run tests on dev branch only; if still failing
- double check if you have mod file properly configured

cfmoore007:
Just compiled only the latest dev.  All tests passed.

cfmoore007:
Ok, re-ran the tests here in win10 and it worked.
Opening a new shell to confirm.  
All good here from a windows pov.
Also tested at work on linux and using the AL test : https://github.com/Autodesk/maya-usd/blob/dev/plugin/al/translators/tests/testMayaRefUnloadable.usda   
... worked as expected.

cfmoore007:
I also just ran the AL unit test on linux, and it passed.

Ran a variant switch test - https://github.com/Autodesk/maya-usd/blob/dev/plugin/al/translators/tests/testMayaRefUnloadable.usda  in the UI, and that also succeeded.

From a QA perspective, looks good.

kxl-adsk:
Thank you for reviews @murphyeoin and validation @cfmoore007 !

@robthebloke Your suggestions are valid improvements for moved code, but I see them as a separate PR. This PR was about moving the schema, renaming and keeping backward compatibility. I will open a new PR with fixes shortly.

PR #260, opened 2020-02-07
add MProfiler/trace performance tracing when using Pixar's batch renderer

mattyjams:
This takes a cue from recent work in the VP2.0 render delegate and adds some performance tracking to Pixar's batch renderer Hydra integration. This is proving to be helpful as we work to deprecate our usage of the legacy viewport, and also as we look towards eventually adopting the VP2.0 render delegate.

The MProfilerScopes added are good for tracking overall time spent in the batch renderer when using Maya's Profiler. The trace performance tracking is good for looking at the time spent in batch renderer setup and overhead versus the time spent in underlying Hydra and USD libraries, since those are also instrumented using trace.

The one change here that affects users **not** using the batch renderer is the addition of a `TRACE_FUNCTION()` call in `MayaUsdProxyShapeBase::boundingBox()`. A lot of time is currently spent generating bounding boxes (serially) using the batch renderer, so that may also be a current bottleneck for the VP2.0 render delegate.

### DIFF
--- a/lib/nodes/proxyShapeBase.cpp
+++ b/lib/nodes/proxyShapeBase.cpp
@@ -34,6 +34,7 @@
 #include "pxr/base/tf/staticTokens.h"
 #include "pxr/base/tf/stringUtils.h"
 #include "pxr/base/tf/token.h"
+#include "pxr/base/trace/trace.h"
 
 #include "pxr/usd/ar/resolver.h"
 #include "pxr/usd/sdf/layer.h"
@@ -620,6 +621,8 @@ MayaUsdProxyShapeBase::GetOutputTime(MDataBlock dataBlock) const
 MBoundingBox
 MayaUsdProxyShapeBase::boundingBox() const
 {
+    TRACE_FUNCTION();
+
     MStatus status;
 
     // Make sure outStage is up to date
--- a/lib/render/pxrUsdMayaGL/batchRenderer.cpp
+++ b/lib/render/pxrUsdMayaGL/batchRenderer.cpp
@@ -94,6 +94,14 @@ TF_DEFINE_PRIVATE_TOKENS(
 
 TF_INSTANTIATE_SINGLETON(UsdMayaGLBatchRenderer);
 
+const int UsdMayaGLBatchRenderer::ProfilerCategory = MProfiler::addCategory(
+#if MAYA_API_VERSION >= 20190000
+    "UsdMayaGLBatchRenderer", "UsdMayaGLBatchRenderer"
+#else
+    "UsdMayaGLBatchRenderer"
+#endif
+);
+
 /* static */
 void
 UsdMayaGLBatchRenderer::Init()
--- a/lib/render/pxrUsdMayaGL/batchRenderer.cpp
+++ b/lib/render/pxrUsdMayaGL/batchRenderer.cpp
@@ -39,6 +39,7 @@
 #include "pxr/base/tf/stringUtils.h"
 #include "pxr/base/tf/stl.h"
 #include "pxr/base/tf/token.h"
+#include "pxr/base/trace/trace.h"
 #include "pxr/base/vt/types.h"
 #include "pxr/base/vt/value.h"
 #include "pxr/imaging/glf/contextCaps.h"
@@ -66,6 +67,7 @@
 #include <maya/MMatrix.h>
 #include <maya/MObject.h>
 #include <maya/MObjectHandle.h>
+#include <maya/MProfiler.h>
 #include <maya/MSceneMessage.h>
 #include <maya/MSelectInfo.h>
 #include <maya/MSelectionContext.h>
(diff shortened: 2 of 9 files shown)


lib/render/pxrUsdMayaGL/batchRenderer.cpp
```diff
@@ -130,6 +140,13 @@ UsdMayaGLBatchRenderer::GetDelegatePrefix(const bool isViewport2) const
 bool
 UsdMayaGLBatchRenderer::AddShapeAdapter(PxrMayaHdShapeAdapter* shapeAdapter)
 {
+    TRACE_FUNCTION();
+
+    MProfilingScope profilingScope(
+        ProfilerCategory,
+        MProfiler::kColorG_L3,
```
mattyjams: Is there a common convention in the colors/levels used with the Profiler? I picked `G` fairly arbitrarily just because it was the last one. Then for levels, I used `1` for the "main" Maya API callback entry points in the `MPxSurfaceShapeUI`/`MHWRender::MPxDrawOverride` overrides, `2` for the proxy shape adapter as a kind of middle tier, and then `3` for the batch renderer itself at the lowest level.

Let me know if there's a better pattern for that!

kxl-adsk: This should help - https://knowledge.autodesk.com/support/maya/learn-explore/caas/CloudHelp/cloudhelp/2019/ENU/Maya-Customizing/files/GUID-54095CD4-7C0E-4031-8981-A45131348182-htm.html

Shades can be used to make the reading of stacked events easier, but we didn't enforce any rule for it.

kxl-adsk: In general - blue should be reserved for rendering tasks, green for compute, and purple for dirty propagation.

lib/nodes/proxyShapeBase.cpp
```diff
@@ -620,6 +621,8 @@ MayaUsdProxyShapeBase::GetOutputTime(MDataBlock dataBlock) const
 MBoundingBox
 MayaUsdProxyShapeBase::boundingBox() const
 {
+    TRACE_FUNCTION();
```
kxl-adsk: How can this trace log be used?

lib/render/pxrUsdMayaGL/batchRenderer.cpp
```diff
@@ -130,6 +140,13 @@ UsdMayaGLBatchRenderer::GetDelegatePrefix(const bool isViewport2) const
...
+        "Batch Renderer Adding Shape Adapter");
```
kxl-adsk: Since you are in the batch renderer category, what's the advantage of adding `Batch Renderer` to every event name? It makes names very long.

lib/render/pxrUsdMayaGL/batchRenderer.cpp
mattyjams: Ahh, thanks @kxl-adsk! This linked page had some helpful examples too:
https://help.autodesk.com/view/MAYAUL/2019/ENU/?guid=GUID-0FF1BA78-0493-463F-8E66-8014FB0F47BC

It took a bit of trial and error though to figure out the mapping of `ProfilingColor` enum value to display color though:
```
kColorA = Purple
kColorB = Green
kColorC = Blue
kColorD = Orange
kColorE = Yellow
kColorG = Gray
```

I'm about to push some new commits with a more "compliant" color scheme with yellows for render-related compute and drawing prep and blues for rendering.

lib/render/pxrUsdMayaGL/batchRenderer.cpp
mattyjams: I was finding the full event name useful when looking at the CPU or thread views in the Profiler vs. the category view. That way I could see batch renderer performance in the context of a full scene render and not have to turn off any other categories.

lib/nodes/proxyShapeBase.cpp
mattyjams: Yeah, it's not that easily discoverable, but the docs for the trace library cover this a little bit under the "Recording and Reporting" section on this page:
https://graphics.pixar.com/usd/docs/api/trace_page_front.html

Here's a little incantation for the script editor that captures trace library calls:
```
from pxr import Trace
collector = Trace.Collector()
collector.enabled = True

# Do some stuff here, like a viewport draw, for example.
cmds.refresh()

# Dump the results to a file.
Trace.Reporter.globalReporter.Report('trace.txt')
```

lib/render/pxrUsdMayaGL/batchRenderer.cpp
kxl-adsk: Some additional tricks related to profiler and category view. Explore the option to collapse events. This reduces the noise by collapsing all events to starting one on a thread. What will remain is a view of the current concurrency level.

You can find this option in the right-click menu, acting on selected events.

lib/nodes/proxyShapeBase.cpp
kxl-adsk: Very useful, thank you!

PR #206, opened 2020-01-16
New Import UI feature - WIP

seando-adsk:
#### MAYA-101565 - Add VariantSets to import dialog and allow selection of a Variant

* Added new build flag --qt-location with corresponding define QT_LOCATION.
* Building of Qt UI is optional based on 'CMAKE_WANT_QT_BUILD' (default ON) and QT_LOCATION being passed in.
* Added optional SUBDIR arg when promoting files.
* Created a USD ImportUI dialog from internal design.
* Added usd/ui folder with new ImportUI dialog.
* From the mayaUsd plugin register a new command to open the ImportUI dialog.
* Added the mayaUsd translator import options based on design specs.

* New import data, which supports:
** Root prim path.
** Stage population mask for importing partial usd file.
** Stage load set.
** Variant sets associated with root prim and invididual prims.
** Added clearData flag to usdImportDialog command.
** Merged variants into single column.
** Added ItemDelegate for customized widget creation for Variants. Use persistent editor with label/combobox for variant selection.
** Moved tree item creation logic from factory to tree item.
** Tree checking & stage population mask.
** New tree header with checkbox for load column - not hooked up yet.
** Commented out all the usd import options (for now) as none of them are implemented yet.

PR #272, opened 2020-02-10
identify Qt 5.12 as the minimum required version for Qt-dependent features

mattyjams:
The newly added import dialog functionality makes use of `QOverload` which was introduced in Qt 5.7, and `QSortFilterProxyModel::setRecursiveFilteringEnabled()` which was introduced in Qt 5.10. Some users may still be using VFX platform 2018 which specifies Qt 5.6.1 as the required version, so they may not be able to build the Qt-dependent features until they switch to VFX platform 2019+ which specifies Qt 5.12 instead.

This change adds CMake enforcement of Qt 5.10+ when building with Qt enabled, and adds a table of version numbers for the optional dependencies, of which Qt is currently the only one.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,7 +91,7 @@ endif()
 
 if(DEFINED QT_LOCATION)
     set(CMAKE_PREFIX_PATH "${QT_LOCATION}")
-    find_package(Qt5 COMPONENTS Core Gui Widgets REQUIRED)
+    find_package(Qt5 5.12 COMPONENTS Core Gui Widgets REQUIRED)
     if(Qt5_FOUND)
         message(STATUS "Building with Qt features enabled.")
     endif()
--- a/doc/build.md
+++ b/doc/build.md
@@ -17,7 +17,13 @@ Before building the project, consult the following table to ensure you use the r
 |    Python Packages    | PyYAML, PySide, PyOpenGL, Jinja2        | PyYAML, PySide2, PyOpenGL, Jinja2              |      PyYAML, PySide, PyOpenGL, Jinja2             |
 |    Build generator    | Visual Studio, Ninja (Recommended)    |  XCode, Ninja(Recommended)                       |    Ninja(Recommended)       |
 |    Command processor  | Visual Studio X64 2015 or 2017 command prompt  |                     bash                |             bash            |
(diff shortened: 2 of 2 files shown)


seando-adsk:
@mattyjams Sorry my fault. When implementing this new dialog I wasn't thinking about older versions of Maya. Ideally the Qt used to compile the plugin should be the same one as inside the Maya version you are targeting. So for Maya 2018/2019 that would be Qt 5.6.1 and for Maya 2020 it is currently 5.12.6. This is a must because the mayaUsdUI.dll will get the Qt shared libs from within Maya. 

I would suggest for now you set the minimum at 5.12 and say that (like the compiler) it must match the Qt version in Maya. Then in the build.md say something to the effect that the new Qt features are only supported in Maya 2020.

I will look at how to support this dialog in Qt 5.6.1.

mattyjams:
No problem @seando-adsk. I code reviewed #206 but obviously didn't try building it internally, so I only just caught this yesterday.

I'll update the CMakeLists.txt to require 5.12 (being a little flexible on the patch release version?) and add a note in build.md.

This got me wondering though whether it might eventually be possible to build against Maya's Qt directly rather than force clients to build their own version that matches. Maya already ships with the libraries, so maybe it's just the headers that are missing? In our Maya installation, I see a `qt-5.6.1-include.tar.gz` file in the `include` directory which looks like it has everything we need. Seems like it would be cleaner/safer to make `FindMaya.cmake` generate a `Maya::Qt` imported target from the Qt stuff in the Maya installation if that's an option. Then it wouldn't be necessary to build your own and point `build.py` at it.

mattyjams:
Rebased and amended the commit based on @seando-adsk's feedback. CMake now enforces Qt 5.12, and I added a note about Maya 2020 being the only currently supported version for the Qt features in build.md.

seando-adsk:
@mattyjams Clients should not have to build Qt. The version from Maya is included with the devkit (as an optional download). However I know it requires some steps to get it working, such as extracting those qt-XXX-tar.gz files. And in the case of the 5.12.6 with Maya 2020 devkit the Qt cmake files reference the debug libs which are not shipped with the devkit. So some manual changes are needed. I've brought this issue up internally with the Maya team.

PR #209, opened 2020-01-18
build ufe Python bindings module consistently with other modules

mattyjams:
This change is similar to #207 but applied to the ufe module. Instead of building the module as `mayaUsd/ufe.so` directly, we instead build it as `mayaUsd/ufe/_ufe.so` and bring its contents into the ufe namespace with an `__init__.py` that uses `Tf.PrepareModule()`.

Also similarly to #207, this adds a `TfScriptModuleLoader` registry function that ensures that when the ufe library is loaded as a result of importing its Python module, the core USD libraries that it depends on also get loaded.

Without the changes in this PR, the test that was added fails with an error that looks like this:
```
======================================================================
ERROR: testImportModule (testUfePythonImport.UfePythonImportTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "testUfePythonImport.py", line 35, in testImportModule
    invalidPrim = mayaUsdUfe.getPrimFromRawItem(0)
TypeError: No to_python (by-value) converter found for C++ type: pxrInternal_v0_20__pxrReserved__::UsdPrim
```

PR #273, opened 2020-02-11
AL #1011: recreate mapping table used by getMayaPathFromUsdPrim after scene reload

murphyeoin:
Calling getMayaPathFromUsdPrim() works fine when you create a proxy shape and call it right away, but fails if you saved and reloaded the scene - The mapping table it uses is not being recreated on scene load/reload and it returns an empty string for any queries.

This change introduces an addition to the ABI of the AL_USDMayaUtils library which in the end we're not using directly, but it may be useful so I left it in


murphyeoin:
@kxl-adsk  valid request - I'll get around to it this week. Thanks

murphyeoin:
@kxl-adsk looks like we had those tests all along - they were disactivated

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
```diff
@@ -2138,24 +2141,20 @@ void ProxyShape::deserialiseTransformRefs()
         {
           MFnDependencyNode fn(node);
           Scope* transformNode = dynamic_cast<Scope*>(fn.userNode());
+          const uint32_t required = tstrs[2].asUnsigned();
+          const uint32_t selected = tstrs[3].asUnsigned();
+          const uint32_t refCounts = tstrs[4].asUnsigned();
+          SdfPath path(tstrs[1].asChar());
+          m_requiredPaths.emplace(path, TransformReference(node, transformNode, required, selected, refCounts));          
           if(transformNode)
```
kxl-adsk: Can you confirm that this is the core change - before you were calling `recordUsdPrimToMayaPath` for everything that WAS NOT a transformNode and now you do it only for transformNode's.

murphyeoin: Not exactly - recordUsdPrimToMayaPath was not previously being called at all, which is why we got the wrong answer on scene reload. Now it's called for every valid prim/Usdmaya scope combination

murphyeoin:
> Re-enabled tests are failing on Windows and MacOS
> 
> ```
> 82: ------------------------ 'Python' is dying ------------------------
> 82: Python crashed. FATAL ERROR: attempted member lookup on NULL TfRefPtr<UsdStage>
> ```
> 
> Both tests require similar changes to these [77fc0a7](https://github.com/Autodesk/maya-usd/commit/77fc0a781ae96e0f832de8fd0267f02bad795ca6) (notice close before we try to open the file)

I think I know why this might need to work the way you describe in windows (can't write to an already open file?) but one of the changes looks a bit weird
```
_tmpMayafile = tempfile.NamedTemporaryFile(delete=True, suffix=".ma") 
_tmpMayafile.close()
```
Isn't this going to delete the file on disk after it's been closed? What is the point of these pair of lines  in that case? (maybe something I don't understand?)

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
kxl-adsk: Right...not sure what I was seeing before. Thx!

kxl-adsk:
When you open a file for writing in python, depending on the Python interpreter implementation, you can't be sure that your modifications have been flushed to disk. Closing the handle guarantees it on any Python interpreter.

Close doesn't delete the file.

murphyeoin:
> When you open a file for writing in python, depending on the Python interpreter implementation, you can't be sure that your modifications have been flushed to disk. Closing the handle guarantees it on any Python interpreter.
> 
> Close doesn't delete the file.

The part that's confusing me is the "delete=True". According to [python docs](https://docs.python.org/2/library/tempfile.html) "If delete is true (the default), the file is deleted as soon as it is closed." - in the original code the file is only closed once it goes out of scope, with Hamed's branch, the file is closed immediately after being created - so I'm not sure how the rest of the code is working, unless I've misinterpreted what's happening

fowlertADSK:
Should be "delete=False".  My original change to fix some of the tests on Windows worked that way, but assuming @HamedSabri-adsk followed that pattern and just missed that.

HamedSabri-adsk:
If I recall correctly, there were some issues with "with statement" on Windows:

I replaced:
`with tempfile.NamedTemporaryFile(delete=True, suffix=".ma") as _tmpMayafile:`

with
```
_tmpMayafile = tempfile.NamedTemporaryFile(delete=True, suffix=".ma") 
_tmpMayafile.close()
```

And kept the passed arguments the same. 

We do have the same pattern in these files already:

https://github.com/Autodesk/maya-usd/blob/bd6aa9133846ca238cdb7958e7f83fd3f8061377/plugin/al/plugin/AL_USDMayaTestPlugin/py/testTranslators.py#L355

https://github.com/Autodesk/maya-usd/blob/528c04147c78405716159e8060c0843270893bd7/plugin/al/plugin/AL_USDMayaTestPlugin/py/testProxyShape.py#L47

murphyeoin:
@HamedSabri-adsk I'm totally happy to replace the "with" statements in the interest of cross-platform compatibility... after looking at your additional examples from the code, I think I see what's happening - when we call NamedTemporaryFile(delete=true) it should work without problems if we're just using it as as a temp-file-path-generator and not relying on the file itself to exist (i.e we export to that filepath only).

murphyeoin:
@kxl-adsk that is strange - obviously the USD file referenced in one of the proxy shapes is either non-existent or valid? Would you be able to add  a few print statements in there to help show the problem, and send me the output? Or do you know (from your maya experience, why the assertion in line 260 passes in linux, but not in other platforms? We're a bit stuck debugging this any further due to our lack of access to windows...

kxl-adsk:
@murphyeoin We are able to reproduce both problems (dying python and asserts in tests) on all platforms (Windows, Linux, macOS). I also tried on Maya 2019 and the same thing. I would love to understand why you can't reproduce it?

kxl-adsk:
We discovered that #209 was no longer failing and we were able to merge it. It's unclear to me what fix it for #209 (I'm suspecting recent refactoring to how we add tests), but it would be worth retesting this PR as well.

@murphyeoin Can you resolve conflicts?

murphyeoin:
Apologies @kxl-adsk just got back to this now - all tests (including the 2 re-activated tests which tests this functionality) run on linux in our internal OSS docker build, test code has been updated to be multi-platform following the pattern in the other tests.

kxl-adsk:
Same failures as reported on Feb 26th/27th.

murphyeoin:
Argh! I'll have a poke around and try and debug what might be the problem, or at least as much as I can do from Linux. I did add some more debug logging at some point, so I might get you to re-run with that on - but I'll review our previous conversations first

murphyeoin:
My bad - I had tested on the wrong branch

kxl-adsk:
It was already part of the https://github.com/Autodesk/maya-usd/pull/273#pullrequestreview-364766653 comment - just removing assemblies check is not enough. We are still failing with
```
93: ------------------------ 'Python' is dying ------------------------
93: Python crashed. FATAL ERROR: attempted member lookup on NULL TfRefPtr<UsdStage>
93: in operator-> at line 978 of /Users/kxl/Dev/git/usd/_workspace/Pixar_USD/include/pxr/base/tf/refPtr.h
93: 
93: The stack can be found in DBNPC0VFHXR.ads.autodesk.com:/var/folders/l1/dml401512fv70lrtkh57djqr0000gq/T//st_Python.26053
93: done.
93: ------------------------------------------------------------------
1/1 Test #93: TestALUSDMayaPython_ProxyShape ...Child aborted***Exception:   7.86 sec
```

Tests have to pass before we can accept this PR.

kxl-adsk:
Closing this PR since there was no activity for a few months.

PR #276, opened 2020-02-11
MAYA-103070 - Minimal Import UI for PR112

seando-adsk:
The import dialog UI is only available when compiling with the optional Qt features. So we need to protect the Import translator UI when this dialog/command is missing.

### DIFF
--- a/plugin/adsk/scripts/mayaUsdTranslatorImport.mel
+++ b/plugin/adsk/scripts/mayaUsdTranslatorImport.mel
@@ -51,15 +51,17 @@ global proc int mayaUsdTranslatorImport (string $parent,
 
         columnLayout -adj true -rs 5 mayaUsdTranslator_OptsCol;
 
-        rowLayout -numberOfColumns 3 -cw 1 $cw1 -cat 1 "right" 0;
-            text -label "Scope and Variants: " -al "right" -ann "Select a USD file and click Hierarchy View to build the scope of your import and switch variants" ;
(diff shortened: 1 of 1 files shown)


plugin/adsk/scripts/mayaUsdTranslatorImport.mel
```diff
@@ -51,15 +51,17 @@ global proc int mayaUsdTranslatorImport (string $parent,
...
-        rowLayout -numberOfColumns 3 -cw 1 $cw1 -cat 1 "right" 0;
-            text -label "Scope and Variants: " -al "right" -ann "Select a USD file and click Hierarchy View to build the scope of your import and switch variants" ;
-            textField -ed false -nbg true -text "Please select a file to enable this option" -w $cw2 mayaUsdTranslator_SelectFileField;
-            button -label "Hierarchy View" -c "string $usdFile = currentFileDialogSelection(); usdImportDialog $usdFile;" mayaUsdTranslator_ViewHierBtn;
-
-            // Hide the button by default.
-            button -e -visible false mayaUsdTranslator_ViewHierBtn;
-        setParent ..;
-
+        if (`exists usdImportDialog`)
```
seando-adsk: The command usdImportDialog will only be available when compiling with the optional Qt features enabled (by using --qt-location). So when not available we need to protect the import translator UI by not showing the "Hierarchy View" button.

seando-adsk:
@mattyjams During some internal testing we noticed a problem with our new import translator options when compiling **without** Qt. Can you have a quick look at these changes?

PR #283, opened 2020-02-13
Fix Maya2018 + UFE build failures (MAYA-102911)

HamedSabri-adsk:


CMakeLists.txt
```diff
@@ -80,6 +80,14 @@ find_package(USD REQUIRED)
...
 
+string(SUBSTRING ${MAYA_API_VERSION} "0" "4" MAYA_MAJOR_VERSION)
+if(${MAYA_MAJOR_VERSION} STRLESS "2019")
+    set(CMAKE_WANT_UFE_BUILD OFF)
+    message(AUTHOR_WARNING "========================================================================================= \
+                            UFE is not available in Maya${MAYA_MAJOR_VERSION}. Maya2019 and later are supported only. \
+                            ========================================================================================= ")
+endif()
+
 if(CMAKE_WANT_UFE_BUILD)
```
HamedSabri-adsk: I set the CMAKE_WANT_UFE_BUILD to OFF if Maya's Major version in less than 2019.

CMakeLists.txt
```diff
@@ -80,6 +80,14 @@ find_package(USD REQUIRED)
 include(cmake/usd.cmake)
 include(${USD_CONFIG_FILE})
 
+string(SUBSTRING ${MAYA_API_VERSION} "0" "4" MAYA_MAJOR_VERSION)
```
HamedSabri-adsk: I also wonder if instead of doing this here, I add these two version variables in FindMaya.cmake:

MAYA_MAJOR_VERSION
MAYA_MINOR_VERSION

seando-adsk: Yes I think this should be in FindMaya.cmake. But what I would suggest is to look at the section where it gets "MAYA_API_VERSION" from MTypes.h. I would suggest adding a similar regex to get the **MAYA_APP_VERSION**. That variable was added to MTypes.h starting in Maya 2019. So if you find it (in MTypes.h) you can use it directly. Otherwise you can use your logic above to read the MAYA_APP_VERSION from the first 4 chars of the API version.

CMakeLists.txt
```diff
@@ -80,6 +80,14 @@ find_package(USD REQUIRED)
...
+if(${MAYA_MAJOR_VERSION} STRLESS "2019")
+    set(CMAKE_WANT_UFE_BUILD OFF)
+    message(AUTHOR_WARNING "========================================================================================= \
+                            UFE is not available in Maya${MAYA_MAJOR_VERSION}. Maya2019 and later are supported only. \
+                            ========================================================================================= ")
+endif()
+
```
seando-adsk: If you end up moving your stuff to FindMaya.cmake I would suggest moving this up to directly under line 33.

HamedSabri-adsk: Hey @seando-adsk,  See 766736c 

> If you end up moving your stuff to FindMaya.cmake I would suggest moving this up to directly under line 33.

This won't work because we want to have this check after `find_package(Maya REQUIRED)` is called.

seando-adsk: As yes you are correct.

CMakeLists.txt
HdC-adsk: Is there any reason why we cannot test with MAYA_API_VERSION directly?

HamedSabri-adsk: @HdC-adsk it just easier to think about the MAYA_APP_VERSION and not worry about the zeros or any numbers that goes after:) 

e.g
#define MAYA_API_VERSION 20180000
#define MAYA_API_VERSION 20210000
vs
#define MAYA_APP_VERSION 2018
#define MAYA_APP_VERSION 2019

HdC-adsk: I think it is easy to test with MAYA_API_VERSION < 20190000. I don't have strong preference, so I guess we will need to remove the "MAYA_APP_VERSION not allowed" item from coding standard. (I added it only because MAYA_API_VERSION is consistent with all recent releases.) What do you think?

CMakeLists.txt
```diff
@@ -80,6 +80,13 @@ find_package(USD REQUIRED)
 include(cmake/usd.cmake)
 include(${USD_CONFIG_FILE})
 
+if(${MAYA_APP_VERSION} STRLESS "2019")
```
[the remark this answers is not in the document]
HamedSabri-adsk: Thanks @mattyjams. Good point. See d1434a2

As for the version, I also agree with earlier comment from @HdC-adsk and you.
We recommend using MAYA_API_VERSION in the *.h,*.cpp for back port compatibility since MAYA_APP_VERSION was introduced since Maya 2019.

Issue #263, opened 2020-02-07
Making a ProxyShape a live surface is broken in VP2.0


dj-mcg:
**Describe the bug**
No geometry is created when trying to use a create tool with a ProxyShape that is made live when using Viewport 2.0.

**Steps to reproduce**
1. Run the testProxyShapeLiveSurface. Observe the failed assert

**Expected behavior**
The test passes with no errors.

**Additional context**
This test is part of a suite of tests that are run internally at Pixar. They are not included run as part of the external Pixar usdMaya plugin tests since they are making assumptions about the runtime environment that may not be valid for all users. However, moving forward, these tests should be run internally within Autodesk to validate correctness when testing any changes or fixes to the usdMaya plugin.

kxl-adsk:
This is because of https://github.com/Autodesk/maya-usd/blob/dev/lib/nodes/proxyShapeBase.cpp#L1013

@mattyjams Is code in https://github.com/Autodesk/maya-usd/blob/81aff08bec82e7323a4a2e6d69b2d6420306f3d5/lib/render/pxrUsdMayaGL/batchRenderer.cpp#L1105 dependent on a specific render delegate or can be shared by any? Maybe we could bring this down to the base class?

santosd:
I was able to repro the issue in my own scene using Maya USD it currently doesn't support making a USD element a live surface so we will need to prioritize this one.

mattyjams:
Hey @kxl-adsk, sorry for the extremely delayed response here, but I'm just starting to dig into this to try to unblock some internal user testing of the VP2.0 render delegate.

I could definitely bring down and/or refactor the closest point delegate into something that should work for both the Pixar batch renderer and the VP2.0 render delegate, but there's a tricky hitch for the latter.

Ultimately, we'll be executing a Hydra pick task using an `HdEngine` and the `HdRenderIndex` that the VP2.0 render delegate/`MPxSubSceneOverride` for the shape currently owns. As far as I can tell, the render delegate is not setting root paths on its `HdRprimCollection`, so we might get away without having to query it for that, and we can instead just use the default absolute root path. Whatever handles the `closestPoint()` callback could probably manage its own `HdEngine`, but I don't think we can get around needing the specific `HdRenderIndex` that the `ProxyShapeDelegate` currently owns.

[...]

PR #242, opened 2020-02-01
[tests] add cmake functions for adding tests

elrond79:
When trying to add a new python test recently, I noticed the large amount of boilerplate code necessary to create a simple test.  I decided to see if I could do something to reduce that, and it turned out it wasn't too bad - this PR makes adding most python test modules a single line.  See, ie:

https://github.com/LumaPictures/maya-usd/blob/pr/base/pytest_cmake_function/plugin/al/translators/tests/CMakeLists.txt

It does this by introducing three new utility cmake functions:

```
mayaUsd_add_pymodule_test
mayaUsd_add_py_test
mayaUsd_set_test_env
```

PR #286, opened 2020-02-14
fix for Pixar testProxyShapeLiveSurface test

mattyjams:
I don't think anyone other than us is running this test, but this addresses an issue where we were seeing intermittent test failures. Execution of the test would sometimes barrel on through the assembly load and attempt to create new geometry using the proxy as a live surface before the geometry for the proxy was ready.

This adds a cmds.refresh() call immediately after creating the test window/viewport to ensure that the geometry that we're making live is present and will work as a live surface.

Note that this does **not** address #263. When the VP2.0 render delegate is enabled, the live surface functionality seems to be broken and this test fails.

### DIFF
--- a/plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeLiveSurface.py
+++ b/plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeLiveSurface.py
@@ -80,6 +80,13 @@ class testProxyShapeLiveSurface(unittest.TestCase):
                 edit=True, displayAppearance='smoothShaded', rnm='vp2Renderer')
         cmds.showWindow(window)
 
+        # Force all views to re-draw. This causes us to block until the
+        # geometry we're making live has actually been evaluated and is present
+        # in the viewport. Otherwise execution of the test may continue and the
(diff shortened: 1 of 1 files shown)


kxl-adsk:
How can one run these tests?

mattyjams:
Yeah, good question @kxl-adsk! :)

We never had these pxrUsdMayaGL tests running as part of the USD repo, but we really should get them going here now. I was waiting to see how #242 shakes out before I tried putting something together for this.

Internally, we have a pretty simple Python script that sets up the environment and then runs the test by launching Maya with the -command option. For the maya-usd repo, I think it would look something like this:

```
import subprocess

moduleEnv = "MAYA_MODULE_PATH=<maya-usd build dir>/install/RelWithDebInfo"

mayaExecutable = "/path/to/maya/binary" # (e.g. "MAYA_EXECUTABLE")

test_file_path = "<maya-usd repo root dir>/plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeLiveSurface.py"

melExecCommand = """python("execfile(\\\"{test_file_path}\\\")")""".format(
    test_file_path=test_file_path)

subprocess.check_call(["env", moduleEnv, mayaExecutable, '-command', melExecCommand])
```

Issue #122, opened 2019-11-18
AL's SIMD operations are disabled


HamedSabri-adsk:
### Issue

AL's SIMD operations are wrapped by AL_UTILS_ENABLE_SIMD but I can't find a #define directive for it anywhere in the code base. 

### Specs

- All platforms ( Linux, MacOSX, Windows )
- affected branches: master, dev, refactoring_sandbox

murphyeoin:
Hi @HamedSabri-adsk as discussed, this is probably just an omission - we can certaintly try and switch on again

kxl-adsk:
Indeed it can be easily switched on...but then it has to support all platforms.

vlasovi:
The issue seems obsolete, closing.

PR #288, opened 2020-02-18
This change brings AL_USDUtils shared library under lib + relevant Unit Tests.

HamedSabri-adsk:
This PR brings `AL_USDUtils` library and relevant unit tests under the `maya-usd/lib`

### Library name:
Library name has changed from `AL_USDUtils.(so,dll,dylib)` to `MayaUsdUtils.(so,dll,dylib)`

### Namespace changes:
Namespace has changed from `AL::usd::utils::` to `MayaUsdUtils::`

### Header inclusion:
All the header files are installed under "include\MayaUsdUtils" and included using angle brackets by the targets that link against MayaUsdUtils library.
```
#include "AL/usd/utils/ForwardDeclares.h"  -----> #include <MayaUsdUtils/ForwardDeclares.h>
#include "AL/usd/utils/SIMD.h"             -----> #include <MayaUsdUtils/SIMD.h>
```

### UnitTest:
`test_DiffCore.cpp` unit test has now moved into `maya-usd/test/lib/usd/utils` . DiffCore tests can be run by running below command:

`ctest -R DiffCore`


HamedSabri-adsk:
Hi @robthebloke , I will add comments on important changes on github sometimes today or tomorrow to make it easier for you to review things.

Thanks again.

lib/usd/utils/SIMD.h
```diff
@@ -59,7 +60,8 @@
 # define ENABLE_SOME_AVX_ROUTINES 1
 #endif
 
-namespace AL {
+namespace MayaUsd {
```
murphyeoin: When we make API breaking changes - can we please have a synopsis in the PR description? This stuff is important to know for anyone using these APIs

lib/usd/utils/DiffCore.h
```diff
@@ -13,15 +13,15 @@
...
 // limitations under the License.
 //
-#pragma once
+#ifndef UTILS_DIFFCORE_H
+#define UTILS_DIFFCORE_H
 
-#include "AL/usd/utils/Api.h"
-#include "AL/usd/utils/ALHalf.h"
+#include "Api.h"
+#include "ALHalf.h"
```
murphyeoin: again - can we not change this until coding standards have been agreed upon?

plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.h
```diff
@@ -32,7 +32,7 @@
 
 #include <vector>
 #include <string>
-#include "AL/usd/utils/ForwardDeclares.h"
+#include <usdUtils/ForwardDeclares.h>
```
murphyeoin: This seems inconsistent with the other include convention which would look like
```
#include "ForwardDeclares.h"
```
again - I would argue for keeping the existing convention pending discussion

plugin/al/lib/AL_USDMaya/CMakeLists.txt
```diff
@@ -194,7 +194,7 @@ target_link_libraries(${LIBRARY_NAME}
     AL_EventSystem
     AL_USDMayaUtils
     AL_MayaUtils
-    AL_USDUtils
+    USDUtils
```
murphyeoin: Following on from comment above about API breaking changes being noted in the PR description - a change to the library name merits the same treatment

lib/usd/utils/DiffCore.cpp
```diff
@@ -13,14 +13,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-#include "AL/usd/utils/DiffCore.h"
-#include "AL/usd/utils/SIMD.h"
+#include "DiffCore.h"
+#include "SIMD.h"
```
robthebloke: Prefer absolute paths.

lib/usd/utils/ForwardDeclares.h
```diff
@@ -13,7 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-#pragma once
+#ifndef UTILS_FORWARDDECLARES_H
+#define UTILS_FORWARDDECLARES_H
```
robthebloke: ```#pragma once ``` is preferable to include guards. Also, this choice of include guard name does not conform to the provisional coding standards.

plugin/al/lib/AL_USDMaya/AL/usdmaya/Global.h
```diff
@@ -16,10 +16,10 @@
 #pragma once
 
 #include "AL/usdmaya/Api.h"
-
-#include "AL/usd/utils/ForwardDeclares.h"
 #include "AL/event/EventHandler.h"
 
+#include <usdUtils/ForwardDeclares.h>
```
robthebloke: Why are you forcing the compiler to search the -isystem paths first? Paths specified with -I should use "", not <>. This will just slow down the build.

plugin/al/usdmayautils/AL/usdmaya/utils/DgNodeHelper.cpp
```diff
@@ -2187,7 +2187,8 @@ MStatus DgNodeHelper::getVec3Array(MObject node, MObject attribute, GfHalf* cons
     v[0] = elem.child(0).asFloat();
     v[1] = elem.child(1).asFloat();
     v[2] = elem.child(2).asFloat();
-    AL::usd::utils::float2half_4f(v, h);
+    v[3] = 0;
```
robthebloke: This is a de-optimisation. I honestly could not care less about -strict compiler warnings on this line of code. The code change is clearly worse. Remove this assignment.

murphyeoin:
Hi @HamedSabri-adsk thanks for making these changes so quickly, including the unit tests in C++ which was a bit of a blocker for us.
A note that as soon as this is merged, I will probably updated it to contain the AL_USDTransaction library, which is currently a separate DSO - I don't think it's necessary to have 2  libs with the same dependencies (i.e mostly USD only)

plugin/al/usdmayautils/AL/usdmaya/utils/DgNodeHelper.cpp
kxl-adsk: @robthebloke Use-before-set errors are tricky to find and debug, so I wouldn't recommend ignoring compiler warnings. Looking at the performance side, it as well tells you something. You are making a copy of 4 elements with the cast where the last one is not initialized. This means you don't even care about the last element. Given that SIMD is disabled (#122), the currently compiled code could be much more efficient if you would have only operated on 3 channels. This would as well avoid the compiler warning.

To close on it, we can't merge changes that don't compile in strict mode.

lib/usd/utils/DiffCore.h
kxl-adsk: To be sure, your ask is to change it to:
```
#include "UsdUtils\Api.h"
#include "UsdUtils\ALHalf.h"
```

plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.h
kxl-adsk: Not sure I understand, ForwardDeclares.h is now moved to UsdUtils

lib/usd/utils/DiffCore.cpp
kxl-adsk: As in here - https://github.com/Autodesk/maya-usd/pull/288#discussion_r381584257 ?

plugin/al/lib/AL_USDMaya/AL/usdmaya/Global.h
kxl-adsk: > By default, the preprocessor looks for header files included by the quote form of the directive #include "file" first relative to the directory of the current file, and then in a preconfigured list of standard system directories. For example, if /usr/include/sys/stat.h contains #include "types.h", GCC looks for types.h first in /usr/include/sys, then in its usual search path.

For the angle-bracket form #include <file>, the preprocessor’s default behavior is to look only in the standard system directories.

Given that this is a separate library, what exactly will be slowed down in this case? ForwardDeclares.h won't be in the same folder so preprocessor will go and look in the pre-configured list of directories.

lib/usd/utils/DiffCore.cpp
HamedSabri-adsk: @robthebloke 

I don't understand the significance behind absolute path inside the shared library. This is what I see has been done in your code base prior to this PR and believe is what you referring to:

1-  In the top level AL's CMakeLists.txt, USDUTILS_INCLUDE_LOCATION is set to usdutils sub-directory:
`set(USDUTILS_INCLUDE_LOCATION ${CMAKE_CURRENT_LIST_DIR}/usdutils)`
2- Inside usdutils directory we go 3 folders deep(AL/usd/utils) and in the CMakeLists.txt inside you add ${USDUTILS_INCLUDE_LOCATION} to target_include_directories
```
target_include_directories(${USDUTILS_LIBRARY_NAME} PUBLIC
    ${USDUTILS_INCLUDE_LOCATION} 
)
```
3- Hence, you can include things like so: #include "AL/usd/utils/DebugCodes.h"

AFAIK, there is no significant difference inside the shared library itself to go with either "AL/usd/utils/DebugCodes.h" or "DebugCodes.h". 

However, I think I understand what you mean by absolute path if one wants to use these header files outside and inclusions like this: #include "AL/usd/utils/ForwardDeclares.h" makes sense to me.

In this PR, I have already installed the headers file inside "usdUtils" directory and inclusion to those header files is done like so:
e.g
#include <usdUtils/ForwardDeclares.h>
#include <usdUtils/SIMD.h>
#include <usdUtils/ALHalf.h>

I am not happy with the usdUtils name. Perhaps we could change `usdUtils` to something like these:

```
#include <mayaUsd/usd/utils/ForwardDeclares.h>
or 
#include <lib/usd/utils/ForwardDeclares.h>
```

HamedSabri-adsk:
@murphyeoin Great, thank you as well.

plugin/al/usdmayautils/AL/usdmaya/utils/DiffPrimVar.cpp
```diff
@@ -14,7 +14,8 @@
 // limitations under the License.
 //
 #include "AL/usdmaya/utils/DiffPrimVar.h"
-#include "AL/usd/utils/SIMD.h"
+#include <MayaUsdUtils/SIMD.h>
+#include <MayaUsdUtils/DiffCore.h>
```
kxl-adsk: Why this new include is needed?

plugin/al/usdmayautils/AL/usdmaya/utils/DiffPrimVar.cpp
```diff
@@ -24,6 +25,8 @@
 
 PXR_NAMESPACE_USING_DIRECTIVE
 
+using namespace MayaUsdUtils;
```
kxl-adsk: Why do you need this `using`?

plugin/al/usdmayautils/AL/usdmaya/utils/MeshUtils.cpp
```diff
@@ -13,11 +13,14 @@
...
+
+#include "AL/usdmaya/utils/MeshUtils.h"
 #include "AL/maya/utils/Utils.h"
 #include "AL/usdmaya/utils/DiffPrimVar.h"
-#include "AL/usdmaya/utils/MeshUtils.h"
 #include "AL/usdmaya/utils/Utils.h"
-#include "AL/usd/utils/DebugCodes.h"
+
+#include <MayaUsdUtils/DebugCodes.h>
+#include <MayaUsdUtils/DiffCore.h>
```
kxl-adsk: Again, why we need DiffCore now and wasn't needed before?

plugin/al/usdmayautils/AL/usdmaya/utils/Utils.h
```diff
@@ -17,11 +17,10 @@
 #pragma once
 
 #include "AL/usdmaya/utils/Api.h"
-
+#include "AL/maya/utils/ForwardDeclares.h"
```
kxl-adsk: We haven't yet agreed on coding standards, but keeping includes ordered in a consistent way is a good practice either way.

cmake/utils.cmake
```diff
@@ -209,6 +209,49 @@ function(mayaUsd_get_unittest_target unittest_target unittest_basename)
     set(${unittest_target} "${unittest_name}" PARENT_SCOPE)
 endfunction()
 
+function(mayaUsd_promoteHeaderListWithSubdir)
```
kxl-adsk: Weird, this was added as part of #255

HamedSabri-adsk: If I recall correctly, at the time your branch `kxl-adsk/MAYA-102925/backport_mayareference_schema_refactoring` wasn't merged in and was still in review. I took mayaUsd_promoteHeaderListWithSubdir from your branch for sake of saving time and added it as part of this PR.

plugin/al/usdmayautils/AL/usdmaya/utils/DiffPrimVar.cpp
HamedSabri-adsk: There are calls to "vec3AreAllTheSame" and "compareArray" in this file from DiffCore and this header is indeed needed.

Previously DiffCore.h was getting included twice both in the *.h and *.cpp. Hence, including DiffPrimVar.h unintentionally would bring in DiffCore.h which is not good.

plugin/al/usdmayautils/AL/usdmaya/utils/MeshUtils.cpp
HamedSabri-adsk: see above comment.

plugin/al/usdmayautils/AL/usdmaya/utils/DiffPrimVar.cpp
HamedSabri-adsk: I added the using namespace instead of explicitly calling namespace with scope operator because there are so many variable and function calls to SIMD operations and I didn't want to make a bigger change than needed. 

loadu4f
f128
movemask4f
cmpne4f
loadu4d
.....

I rather get rid of the using and have the MayaUsdUtil:: for all them. 

@murphyeoin can you please comment on this?

cmake/utils.cmake
kxl-adsk: GitHub tricked me, I thought you already merged dev here since I didn't see any conflicts. All good, thx!

HamedSabri-adsk:
> Following tests are failing with this change:
> 79 - AL_MayaUtilsTests (Failed)
> 82 - AL_USDMayaTestPlugin (Failed)
> 83 - TestUSDMayaPython (Failed)
> 84 - TestUSDMayaPythonModuleOnly (Failed)
> 86 - TestAdditionalTranslators (Failed)
> 87 - TestPxrUsdTranslators (Failed)

@kxl-adsk  See 6609dfb1. do these tests pass for you on MacOS?

PR #291, opened 2020-02-19
save and restore uniform buffer bindings during batch renderer selection

mattyjams:
This addresses an issue when running the [testProxyShapeSelectionPerformance](https://github.com/Autodesk/maya-usd/blob/dev/plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeSelectionPerformance.py) test where only the first image would be drawn correctly and all images drawn after the first selection operation would be completely black.

This workaround was already being used during rendering, but this change applies it during selection as well.

### DIFF
--- a/lib/render/px_vp20/utils.cpp
+++ b/lib/render/px_vp20/utils.cpp
@@ -27,6 +27,7 @@
 #include "pxr/base/gf/vec3f.h"
 #include "pxr/base/gf/vec4f.h"
 #include "pxr/base/tf/stringUtils.h"
+#include "pxr/imaging/garch/gl.h"
 #include "pxr/imaging/glf/simpleLight.h"
 #include "pxr/imaging/glf/simpleLightingContext.h"
 #include "pxr/imaging/glf/simpleMaterial.h"
@@ -1077,6 +1078,27 @@ px_vp20Utils::OutputDisplayStatusToStream(
         case MHWRender::kNoStatus:
             stream << "kNoStatus";
             break;
+    }
+}
+
+
+UBOBindingsSaver::UBOBindingsSaver()
+{
+    for (size_t i = 0u; i < _uniformBufferBindings.size(); ++i) {
+        glGetIntegeri_v(
+            GL_UNIFORM_BUFFER_BINDING,
+            static_cast<GLuint>(i),
+            &_uniformBufferBindings[i]);
+    }
+}
+
+UBOBindingsSaver::~UBOBindingsSaver()
+{
+    for (size_t i = 0u; i < _uniformBufferBindings.size(); ++i) {
+        glBindBufferBase(
+            GL_UNIFORM_BUFFER,
+            static_cast<GLuint>(i),
+            _uniformBufferBindings[i]);
     }
 }
 
--- a/lib/render/px_vp20/utils.cpp
+++ b/lib/render/px_vp20/utils.cpp
@@ -1082,7 +1082,7 @@ px_vp20Utils::OutputDisplayStatusToStream(
 }
 
 
-UBOBindingsSaver::UBOBindingsSaver()
+GLUniformBufferBindingsSaver::GLUniformBufferBindingsSaver()
 {
     for (size_t i = 0u; i < _uniformBufferBindings.size(); ++i) {
         glGetIntegeri_v(
@@ -1092,7 +1092,7 @@ UBOBindingsSaver::UBOBindingsSaver()
     }
 }
 
-UBOBindingsSaver::~UBOBindingsSaver()
+GLUniformBufferBindingsSaver::~GLUniformBufferBindingsSaver()
 {
     for (size_t i = 0u; i < _uniformBufferBindings.size(); ++i) {
         glBindBufferBase(
--- a/lib/render/px_vp20/utils.h
+++ b/lib/render/px_vp20/utils.h
@@ -24,6 +24,7 @@
 #include "pxr/base/gf/matrix4d.h"
 #include "pxr/base/gf/matrix4f.h"
 #include "pxr/base/gf/vec4f.h"
+#include "pxr/imaging/garch/gl.h"
 #include "pxr/imaging/glf/simpleLightingContext.h"
 
 #include <maya/M3dView.h>
@@ -33,6 +34,7 @@
 #include <maya/MMatrix.h>
 #include <maya/MSelectionContext.h>
 
+#include <array>
 #include <ostream>
 
 
@@ -112,6 +114,40 @@ class px_vp20Utils
     private:
         px_vp20Utils() = delete;
         ~px_vp20Utils() = delete;
+};
+
+
+/// Simple RAII class to save uniform buffer bindings, to deal with a Maya
+/// issue.
+///
+/// XXX: When Maya is using OpenGL Core Profile as the rendering engine (in
+/// either compatibility or strict mode), batch renders like those done in the
+/// "Render View" window or through the ogsRender command do not properly track
+/// uniform buffer binding state. This was causing issues where the first batch
+/// render performed would look correct, but then all subsequent renders done
+/// in that Maya session would be completely black (no alpha), even if the
+/// frame contained only Maya-native geometry or if a new scene was
+/// created/opened.
+///
+/// To avoid this problem, this object can be used to save and restore Maya's
+/// uniform buffer bindings across Hydra/OpenGL calls. We try not to bog down
+/// performance by saving and restoring *all* GL_MAX_UNIFORM_BUFFER_BINDINGS
+/// possible bindings, so instead we only do just enough to avoid issues.
+/// Empirically, the problematic binding has been the material binding at
+/// index 4.
+class UBOBindingsSaver
+{
+    public:
+        static constexpr size_t UNIFORM_BINDINGS_TO_SAVE = 5u;
+
+        MAYAUSD_CORE_PUBLIC
+        UBOBindingsSaver();
+
+        MAYAUSD_CORE_PUBLIC
+        ~UBOBindingsSaver();
+
+    private:
+        std::array<GLint, UNIFORM_BINDINGS_TO_SAVE> _uniformBufferBindings;
 };
 
 
--- a/lib/render/px_vp20/utils.h
+++ b/lib/render/px_vp20/utils.h
@@ -135,16 +135,16 @@ class px_vp20Utils
 /// possible bindings, so instead we only do just enough to avoid issues.
 /// Empirically, the problematic binding has been the material binding at
 /// index 4.
-class UBOBindingsSaver
+class GLUniformBufferBindingsSaver
 {
     public:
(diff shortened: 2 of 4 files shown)


mattyjams:
Actually, I just saw that this was turned into an RAII class in mayaToHydra:
https://github.com/Autodesk/maya-usd/blob/dev/lib/render/mayaToHydra/renderOverride.cpp#L110

Maybe it'd be better if I moved that to a public place ([px_vp20/utils.h](https://github.com/Autodesk/maya-usd/blob/dev/lib/render/px_vp20/utils.h)?) and used it from there instead? I'll try that and update the branch.

lib/render/pxrUsdMayaGL/batchRenderer.cpp
```diff
@@ -1129,6 +1136,27 @@ UsdMayaGLBatchRenderer::_TestIntersection(
                  GL_TEXTURE_BIT |
                  GL_POLYGON_BIT);
 
+    // XXX: When Maya is using OpenGL Core Profile as the rendering engine (in
```
HdC-adsk: To help me understand better, when is _TestIntersection() called?

Is UsdMayaGLBatchRenderer being called from the draw override solution from Pixar?

lib/render/pxrUsdMayaGL/batchRenderer.cpp
```diff
@@ -1129,6 +1136,27 @@ UsdMayaGLBatchRenderer::_TestIntersection(
...
+    // where the first batch render performed would look correct, but then all
+    // subsequent renders done in that Maya session would be completely black
+    // (no alpha), even if the frame contained only Maya-native geometry or if
+    // a new scene was created/opened.
+    //
+    // To avoid this problem, we need to save and restore Maya's bindings
+    // across Hydra calls. We try not to bog down performance by saving and
+    // restoring *all* GL_MAX_UNIFORM_BUFFER_BINDINGS possible bindings, so
+    // instead we only do just enough to avoid issues. Empirically, the
+    // problematic binding has been the material binding at index 4.
```
HdC-adsk: Looks like this would only affect selection performance? Any other workflow that would be impacted?

lib/render/pxrUsdMayaGL/batchRenderer.cpp
mattyjams: Yeah, `UsdMayaGLBatchRenderer` is used to image `pxrUsdProxyShape` and `mayaUsdProxyShape` when `VP2_RENDER_DELEGATE_PROXY` is disabled.

The `MPxDrawOverride::userSelect()` override for proxy shapes is what ends up eventually calling `_TestIntersection()` to do viewport-based picking:
https://github.com/Autodesk/maya-usd/blob/c75e69638c560a82163f449cbcc3cbfbd8077e27/lib/render/pxrUsdMayaGL/proxyDrawOverride.cpp#L259

lib/render/pxrUsdMayaGL/batchRenderer.cpp
mattyjams: The one other workflow that might be affected is when live surface is used. We end up using Hydra to do the intersection testing for that as well:
https://github.com/Autodesk/maya-usd/blob/c75e69638c560a82163f449cbcc3cbfbd8077e27/lib/render/pxrUsdMayaGL/proxyShapeDelegate.cpp#L96

mattyjams:
I just pushed a re-worked version of this where the `UBOBindingsSaver` was moved to `px_vp20/utils` and both mayaToHydra and the Pixar batch renderer use it from there.

lib/render/pxrUsdMayaGL/batchRenderer.cpp
```diff
@@ -1196,6 +1198,8 @@ UsdMayaGLBatchRenderer::_ComputeSelection(
 
     _selectResults.clear();
 
+    UBOBindingsSaver bindingsSaver;
```
mattyjams: This version is slightly different from the previous version in that the instantiation of the `UBOBindingsSaver` happens **outside** of `_TestIntersection()`. In relatively infrequent cases, the loop below may go through more than one iteration, so this avoids repeatedly saving and restoring the bindings on each iteration.

This is also why a `UBOBindingsSaver` was needed in `TestIntersectionCustomPrimFilter()` above as well.

The original version of this PR is available for comparison in commit 37d2924fb875037539ac38a13034d052606d0ff1.

mattyjams:
Here's the simple test scene I was using for this:
[CubeModelProxy.zip](https://github.com/Autodesk/maya-usd/files/4228004/CubeModelProxy.zip)

Extract that archive and launch Maya on the ma file with `VP2_RENDER_DELEGATE_PROXY=0` (**disabled**).

Then I had this incantation in the script editor (distilled from the unit test script mentioned in the description):
```
import os

testDir = os.path.abspath('.')

cmds.workspace(testDir, o=True)


def _WriteViewportImage(outputImageName):
    cmds.setAttr('defaultRenderGlobals.currentRenderer', 'mayaHardware2', type='string')
    # Set the image format to PNG.
    cmds.setAttr('defaultRenderGlobals.imageFormat', 32)
    # Set the render mode to shaded and textured.
    cmds.setAttr('hardwareRenderingGlobals.renderMode', 4)
    # Specify the output image prefix. The path to it is built from the
    # workspace directory.
    cmds.setAttr('defaultRenderGlobals.imageFilePrefix', outputImageName, type='string')
    # Apply the viewer's color transform to the rendered image, otherwise
    # it comes out too dark.
    cmds.setAttr("defaultColorMgtGlobals.outputTransformEnabled", 1)
    # Do the render.
    cmds.ogsRender(camera='persp', currentFrame=True, width=960, height=540)

_WriteViewportImage('testImage1')

# Click on the cube in the viewport to select it here

_WriteViewportImage('testImage2')
```

Without these changes, I was seeing testImage2.png coming out completely black. With the changes, it shows the selected cube instead.

mattyjams:
For what it's worth, it looks like the underlying issue is something that Animal Logic ended up tripping over as well:
https://github.com/Autodesk/maya-usd/blob/c75e69638c560a82163f449cbcc3cbfbd8077e27/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyDrawOverride.cpp#L464

lib/render/px_vp20/utils.h
```diff
@@ -115,6 +117,38 @@ class px_vp20Utils
...
+/// frame contained only Maya-native geometry or if a new scene was
+/// created/opened.
+///
+/// To avoid this problem, this object can be used to save and restore Maya's
+/// uniform buffer bindings across Hydra/OpenGL calls. We try not to bog down
+/// performance by saving and restoring *all* GL_MAX_UNIFORM_BUFFER_BINDINGS
+/// possible bindings, so instead we only do just enough to avoid issues.
+/// Empirically, the problematic binding has been the material binding at
+/// index 4.
+class UBOBindingsSaver
```
HdC-adsk: Does it make sense to add a GL post fix to the class name? I think this makes it clearer that the class is GL-specific.

mattyjams: Yeah, makes sense. The name is a little opaque at a quick glance.

How does `GLUniformBufferBindingsSaver` sound?

HdC-adsk:
@mattyjams Thanks for making this into a utilities class that can be used elsewhere.

lib/render/px_vp20/utils.cpp
```diff
@@ -1081,4 +1082,25 @@ px_vp20Utils::OutputDisplayStatusToStream(
 }
 
 
+GLUniformBufferBindingsSaver::GLUniformBufferBindingsSaver()
```
kxl-adsk: Compilation error on Windows (missing export)
```
   Creating library lib\render\mayaToHydra\mtoh.lib and object lib\render\mayaToHydra\mtoh.exp
renderOverride.cpp.obj : error LNK2019: unresolved external symbol "public: __cdecl pxrInternal_v0_20__pxrReserved__::GLUniformBufferBindingsSaver::GLUniformBufferBindingsSaver(void)" (??0GLUniformBufferBindingsSaver@pxrInternal_v0_20__pxrReserved__@@QEAA@XZ) referenced in function "public: class Autodesk::Maya::OpenMaya20200000::MStatus __cdecl pxrInternal_v0_20__pxrReserved__::MtohRenderOverride::Render(class Autodesk::Maya::OpenMaya20200000::MHWRender::MDrawContext const &)" (?Render@MtohRenderOverride@pxrInternal_v0_20__pxrReserved__@@QEAA?AVMStatus@OpenMaya20200000@Maya@Autodesk@@AEBVMDrawContext@MHWRender@456@@Z)
renderOverride.cpp.obj : error LNK2019: unresolved external symbol "public: __cdecl pxrInternal_v0_20__pxrReserved__::GLUniformBufferBindingsSaver::~GLUniformBufferBindingsSaver(void)" (??1GLUniformBufferBindingsSaver@pxrInternal_v0_20__pxrReserved__@@QEAA@XZ) referenced in function "public: class Autodesk::Maya::OpenMaya20200000::MStatus __cdecl pxrInternal_v0_20__pxrReserved__::MtohRenderOverride::Render(class Autodesk::Maya::OpenMaya20200000::MHWRender::MDrawContext const &)" (?Render@MtohRenderOverride@pxrInternal_v0_20__pxrReserved__@@QEAA?AVMStatus@OpenMaya20200000@Maya@Autodesk@@AEBVMDrawContext@MHWRender@456@@Z)
lib\render\mayaToHydra\mtoh.mll : fatal error LNK1120: 2 unresolved externals
```

mattyjams: Ahh, sorry I missed that. Should be fixed now.

kxl-adsk:
Which commit fixed it?

mattyjams:
I made the fix as part of an interactive rebase, so the export macros were added in commit 32def62ac2e484b0d1c1fab390e10312540c56f9.

lib/render/px_vp20/utils.h
```diff
@@ -115,6 +117,40 @@ class px_vp20Utils
...
+/// To avoid this problem, this object can be used to save and restore Maya's
+/// uniform buffer bindings across Hydra/OpenGL calls. We try not to bog down
+/// performance by saving and restoring *all* GL_MAX_UNIFORM_BUFFER_BINDINGS
+/// possible bindings, so instead we only do just enough to avoid issues.
+/// Empirically, the problematic binding has been the material binding at
+/// index 4.
+class GLUniformBufferBindingsSaver
+{
+    public:
+        static constexpr size_t UNIFORM_BINDINGS_TO_SAVE = 5u;
```
HdC-adsk: This looks like an implementation detail and should be hidden in the private section?

mattyjams: Yup, makes sense! Done in commit 5c448b5bdb5e23bfe3aaba7a230c49cadc723e74.

Issue #279, opened 2020-02-12
Build Issue in windows10 and maya 2018


arunpillaii:
The USD 19.11 with maya 2018 in windows 10 is failing with the below error.
 python build.py --maya-location "C:\Program Files\Autodesk\Maya2018" --pxrusd-location "C:\USD\19.11" "C:\USD\maya_usd\2018"

**Build log**
  test_usdmaya_AttributeType.cpp
  Unknown compiler version - please run the configure tests and report the results
     Creating library C:/USD/maya_usd/2018/build/RelWithDebInfo/plugin/al/plugin/AL_USDMayaTestPlugin/RelWithDebInfo/AL_USDMayaTestPlugin.lib and object C:/USD/maya_usd/2018/build/RelWithDebInfo/plugin/al/plugin/AL_USDMayaTestPlugin/RelWithDebInfo/AL_USDMayaTestPlugin.exp
  AL_USDMayaTestPlugin.vcxproj -> C:\USD\maya_usd\2018\build\RelWithDebInfo\plugin\al\plugin\AL_USDMayaTestPlugin\RelWithDebInfo\AL_USDMayaTestPlugin.mll

Traceback (most recent call last):
  File "build.py", line 584, in <module>
    BuildAndInstall(context, context.buildArgs, context.stagesArgs)
  File "build.py", line 354, in BuildAndInstall
    RunCMake(context, extraArgs, stagesArgs)
  File "build.py", line 266, in RunCMake
    multiproc=FormatMultiProcs(context.numJobs, generator)))
  File "build.py", line 156, in Run
    .format(cmd=cmd, log=os.path.abspath(context.logFileLocation)))
RuntimeError: Failed to run 'cmake --build . --config RelWithDebInfo --target install -- /M:8'
See C:\USD\maya_usd\2018\build\RelWithDebInfo\build_log.txt for more details.
[build_log.txt](https://github.com/Autodesk/maya-usd/files/4192550/build_log.txt)
**Specs:**
 - OS & version [e.g. Windows 10]
 - Maya version 2018

fowlertADSK:
I'm not 100% sure about the "Unknown compiler version" error you're getting, but you need to use the compiler version that matches the version of Maya you are building against.  I believe Maya 2018 and 2019 both used Visual Studio 2015, not 2017, so I think you're going to have to switch (the Build docs were updated recently to hopefully make this more clear).  Give that shot and see if it helps.

HamedSabri-adsk:
[...]

Issue #233, opened 2020-01-31
WIndows : Issue while compiling the dev branch


fdevic:
Hello!

I tried to compiling with latest code in the dev branch.

Steps :
`set PATH=%PATH%;c:\Python27;c:\Python27\Scripts
`

In the x64 native prompt of Visual2017
Git
```
git clone https://github.com/PixarAnimationStudios/USD
cd USD
git checkout v19.11
```
compiling:
`python USD\build_scripts\build_usd.py --build-args=boost,"--with-date_time --with-thread --with-system --with-filesystem" --force-all --no-maya --python --no-houdini "e:\myproject\usd_19_11_boost_no_alembic"`

With the repo as is and modifying the D:\myproject\usd_19_11_boost_no_alembic\cmake\pxrTargets.cmake file
with : (see issue #86 : https://github.com/Autodesk/maya-usd/issues/86 )

```
# Loop over all imported files and verify that they actually existforeach(target ${_IMPORT_CHECK_TARGETS} )  foreach(file ${_IMPORT_CHECK_FILES_FOR_${target}} )#    if(NOT EXISTS "${file}" )#      message(FATAL_ERROR "The imported target \"${target}\" references the file   \"${file}\"# =============================================================================================    get_filename_component(barename "${file}" NAME_WE)    # HACK - the following are dynamic libs in Windows (no .lib) - these will be dll's only    set(skip_dynamic_libs "hdStream" "usdObj" "usdShaders")    list(FIND skip_dynamic_libs ${barename} _index)    if(NOT EXISTS "${file}" AND NOT ((${_index} GREATER -1) AND WIN32))# =============================================================================================      message(FATAL_ERROR "The imported target \"${target}\" references the file         \"${file}\"but this file does not exist.  Possible reasons include:* The file was deleted, renamed, or moved to another location.* An install or uninstall procedure did not complete successfully.* The installation package was faulty and contained   \"${CMAKE_CURRENT_LIST_FILE}\"but not all the files it references.")    endif()  endforeach()  unset(_IMPORT_CHECK_FILES_FOR_${target})endforeach()
--

```

Then compile maya-usd with :
`python build.py --maya-location="c:\Program Files\autodesk\Maya2020" --pxrusd-location="e:\myproject\usd_19_11_boost_no_alembic" --devkit-location="d:\myproject\external\Maya\sdk\2020\devkitbase" e:\\maya_usd_19_11`

And I have errors :
[...]

PR #293, opened 2020-02-20
MAYA-103572 Disable a couple warnings

fowlertADSK:
Disable a couple warnings before including a couple boost headers.  Avoids "warning treated as errors" in a couple specific cases.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/LayerManager.h
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/LayerManager.h
@@ -30,7 +30,19 @@
 
 #include <map>
 #include <set>
-#include <boost/thread.hpp>
+
+// On Windows, against certain versions of Maya and with strict compiler
+// settings on, we are getting warning-as-error problems with a couple
+// boost includes.  Disabling those warnings for the specific includes
+// for now instead of disabling the strict settings at a higher level.
+#ifdef _MSC_VER
+#pragma warning( push )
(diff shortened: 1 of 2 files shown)


kxl-adsk:
This was causing compilation errors on Windows with Maya 2018 and 2019 (#279 #233).
@robthebloke please review.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/LayerManager.h
```diff
@@ -30,7 +30,19 @@
...
+// boost includes.  Disabling those warnings for the specific includes
+// for now instead of disabling the strict settings at a higher level.
+#ifdef _MSC_VER
+#pragma warning( push )
+#pragma warning( disable : 4002 )
+#endif
+  #include <boost/thread.hpp>
+#ifdef _MSC_VER
+#pragma warning( pop )
+#endif
```
robthebloke: No real problem against doing this. If this happens on a frequent occurrence, then a macro such as:
```c++
#define PUSH_DISABLE_WARNING_MSVC4002 \
  __pragma(warning(push)) \
  __pragma(warning(disable:4002)) 

#define PUSH_DISABLE_WARNING_MSVC4275 \
  __pragma(warning(push)) \
  __pragma(warning(disable:4275)) 

#define POP_DISABLE_WARNINGS \
  __pragma(warning(pop)) 
```
Might be an idea, but no real issue with this as is.

PR #294, opened 2020-02-20
A piece of PR183 - approximate equality checks

elenadBlueSky:
Since the full pr/animationMissingAfterMayaFileSaveReopen was too controversial, breaking out some pieces. This piece has the part that everyone agreed was safe to merge in - approximate equality checks of new values against currently set values to determine whether or not new values need to be saved.  Plus a couple of changes that address what looked like small oversights.
(This change alone fixes the problem where we bring in a usd file with a sphere, turn on the Move To manipulator, and then, via python calls, select the sphere, set the sphere's rotateZ to 45 degrees, unselect the sphere, reselect the sphere and the rotation (and usd "over" that sets it) is inexplicably reset to 0,0,0.)

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/TransformationMatrix.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/TransformationMatrix.cpp
@@ -1119,6 +1119,7 @@ void TransformationMatrix::initialiseToPrim(bool readFromPrim, Scope* transformN
             MPlug(transformNode->thisMObject(), MPxTransform::shearXZ).setValue(tempShear.y);
             MPlug(transformNode->thisMObject(), MPxTransform::shearYZ).setValue(tempShear.z);
             m_shearTweak[0] = m_shearTweak[1] = m_shearTweak[2] = 0;
+            m_shearFromUsd = tempShear;
           }
         }
       }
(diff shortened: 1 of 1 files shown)


kxl-adsk:
Added @elrond79 and @murphyeoin as reviewers since they were providing feedback on #183

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/TransformationMatrix.cpp
```diff
@@ -1119,6 +1119,7 @@ void TransformationMatrix::initialiseToPrim(bool readFromPrim, Scope* transformN
             MPlug(transformNode->thisMObject(), MPxTransform::shearXZ).setValue(tempShear.y);
             MPlug(transformNode->thisMObject(), MPxTransform::shearYZ).setValue(tempShear.z);
             m_shearTweak[0] = m_shearTweak[1] = m_shearTweak[2] = 0;
+            m_shearFromUsd = tempShear;
```
elrond79: Ideally, this would be on it's own PR or at least commit - both because it's obviously different than the rest of the changes, and because it seems like an important enough bug fix to deserve splitting out.

Having said that - this does seem like a needed fix, and you've already been good enough to split things out into two separate commits, so I'm not going to hold things up over this.

PR #295, opened 2020-02-20
A piece of PR183 - disabling push to prim to avoid callbacks while updating data

elenadBlueSky:
Since the full pr/animationMissingAfterMayaFileSaveReopen was too controversial, breaking out some pieces.  This piece disables push to prim inside the initialisePrim method while data is being updated in order that overs don't accidentally get generated during this operation.  At the end of the method, push to prim is reenabled.
(This change alone fixes the problem where we bring in a usd file with an animated camera, save this to a .ma file, read in the .ma file and the animation is gone because there is now an "over" in the current layer that covers up the animation (which is in a deeper layer).)


kxl-adsk:
Added @elrond79 and @murphyeoin as reviewers since they were reviewers on #183

elrond79:
Well, as I [said in 183](https://github.com/Autodesk/maya-usd/pull/183#discussion_r377846427), I'd still prefer a basic RAII object to handle this:

> This works, but now there's unnecessary code duplication. Plus, it will make the code in the calling function slightly cleaner - in addition to moving the cleanup code out of initialiseToPrim, we also move the pushOriginallyEnabled variable, and also avoid the need for the non-standard (non-)indentation inside of the try-finally block (which I assume was done to avoid creating a large diff).
> Creating an extra class for this might feel a bit awkward at first, but it really is the standard way to deal with these situations in C++.

Perhaps we should get another opinion? @murphyeoin or @robthebloke - either of you want to weight in?

elenadBlueSky:
@kxl-adsk - I'll add the RAll object as you requested and resubmit.

elenadBlueSky:
@kxl-adsk - Note that at the end of initialiseToPrim there is an operation on m_flags potentially unsetting push to prim.  Because of this, I can't just let the disableNow variable go out of scope when the method is done - I have to create a new scope so it goes out of scope before this last m_flags operation.  Otherwise that operation may be undermined by the disableNow destructor being called after it.

kxl-adsk:
My brain may be shutting down now, but all I see is bitwise NOT which will clear all the bits with exception to kPushToPrimEnabled. This means the scoped object won't make any difference in this case.

kxl-adsk:
Wait..my bad.

kxl-adsk:
Simply add a scope that will cause the destruction of scoped disabler earlier. Sorry, that's how it is when you code during meetings.

elenadBlueSky:
I already added the scope.  Want me to indent everything in the middle?  It'll be a large diff, but you know what's happening now.

kxl-adsk:
Yes, please.

kxl-adsk:
@elenadBlueSky please resolve merge conflicts

PR #299, opened 2020-02-24
Fix up UFE unit test failures, and improved TRS tool behaviour

robthebloke:
So I've been hammering away at a quick prototype to improve the UFE/USD integration, and have run into a few issues. This is a PR that I'm opening simply to get some visibility, and a bit of help/advice to fix up the issues. This is not a PR that's ready for the primetime. 

**!!! PLEASE NO CODE REVIEWS ON THIS PR, JUST HELP WITH THE QUESTIONS BELOW !!!**

This PR is an exploratory test to see how far we can get with UFE within our current pipeline. It aims to do the following:

1. Utilises the maya transform API within UFE to add support for scale/rotate pivots. If the profile cannot be matched, the prim reverts to effectively being read only (can read from the evaluated matrix, but won't modify any UsdGeomXformOps). It supports the geom xform ops that the AL plugin have been generating for a while now, and the UsdGeomXformCommonAPI. If it's not supported, UFE transforms become read only (any won't write to USD).

2. UsdTimeCode::Default v.s. a value from the timeline. At the moment this PR reads/writes time sampled values based on the time returned from the proxy shape _(previously it would only read/write the value at UsdTimeCode::Default(), so animated prims could not be selected)_. We need to come up with a set of business rules to govern how this works in practice _(i.e. when do we modify sampled values, when do we edit default values, do we look at the current edit target, do we ignore it?)_. This is an area where bluesky and AL are quite keen on doing some iteration on, because it's an area that will bite us hard if we get it wrong. 

3. I've tried to fix up the AL unit tests when the VP2 render delegate and UFE is enabled. Primarily this has focused on trying to get the AL_usdmaya_ProxyShapeSelect command to play nicely with UFE. Within most of our tools built on top of AL_USDMaya, we tend to select and deselect prims using this command anyway, so we figured it would be a good place to start abstracting the old v.s. new system. There are 2 unit tests I've effectively disabled, simply because they worry about the transform node behaviour, and we aren't that worried because we should be able to ditch the transform node once UFE works. There is however one test that is failing, and I don't know how to make it work... So here comes my questions.... 

```AL_maya_test_UnitTestHarness -ktf -f "ProxyShapeSelect.pseudoRootSelect";```

At the moment, the above unit test fails. The test effectively just wants to select the root prim path, ```SdfPath("/");```. In the old system, this would simply end up selecting the ProxyShape node itself within maya. 

Replicating that behaviour within UFE is causing me a few issues.... 

1. It doesn't seem to be possible to create a path in UFE for the root prim '/' ?
2. I attempted to copy the previous behaviour, namely: ```when '/' is selected, select the ProxyShape node using standard maya selection```. This doesn't seem possible with UFE? If you attempt to select "/", UFE says the path is invalid, and wont let you select it. If you modify the selection list _(i.e. add the proxy shape to an MSelectionList, and then call MGlobal::setActiveSelectionList)_, the call to setActiveSelectionList immediately clears the UFE selection, **_and_** inserts the UFE path "AL_usdmaya_ProxyShape1". However I cannot insert that item into UFE myself _(The following paths are both invalid: "AL_usdmaya_ProxyShape1", "|world|transform1|AL_usdmaya_ProxyShape1/", so what is the correct path to use???)

Question 1: **(Question for Autodesk)**
If i want to select the prim path "/root/foo", UFE will create that item as ```"|world|transform1|AL_usdmaya_ProxyShape1/root/foo"```. Cool. If i want to select the path "/", ufe will simply report an invalid path, and throw an error. Ok, so I guess I need to special case that, and simply select the proxy shape via MGlobal::setActiveSelectionList.  Except when I do that, UFE now reports "AL_usdmaya_ProxyShape1" as a selected item, but has wiped the rest of the selection list clean. If I attempt to insert the UFE item "AL_usdmaya_ProxyShape1", I can't, because it's not a valid path. There's a chicken & egg thing going on here, and I'm not sure how we are supposed to handle it? So how do we add maya nodes to the UFE selection list, via UFE? or, how do we represent "/" in UFE? 

Question 2: **(Question for Bluesky, and other interested parties...)**
Currently the behaviour of this PR is to pull the time value directly from the proxy shape, and use that time value for each get/set of each transform op. This kinda ruins any ability to be able to get/set the default value, but does at least mean that transform manips follow animated transforms. We need to start deciding the business logic for how we are going to handle this in a way that doesn't lead to a catalogue of confusion. 

Sorry for waffling too much, any help would be appreciated :)

### CODE REVIEW

**Comment by kxl-adsk** [comment]
I haven't looked at the changes yet, but I love the idea of collecting feedback this way. FYI. GitHub has a draft option which makes it crystal clear that PR is not a final change (and can't be merged) - https://github.blog/2019-02-14-introducing-draft-pull-requests/

**Comment by robthebloke** [comment]
One step at a time. Apologies for this PR being very lazy. It should probably be thought of as AL's sandbox_refactor branch. It's not clean. It's not tidy. It probably features a load of other sandbox changes that aren't yet ready for the prime time. I'm stuck, so threw it up to get help! :) 

I'll do a code review on my own code, and highlight the issues. Hopefully that will help target people towards the actual code problems (and I'll clean up at a later date into some tidy sane PR's, one track able fix at a time)

File: lib/ufe/Global.cpp [hunk 1]
```diff
@@ -69,6 +69,14 @@ StagesSubject::Ptr g_StagesSubject;
 
 bool InPathChange::fInPathChange = false;
 
+void refreshStages()
+{
+	if(g_StagesSubject)
+	{
+		g_StagesSubject->afterOpen();
```
robthebloke: So... When you change the file path on a proxy shape, UFE waits for a notice from usd, which only seems to be sent once maya is idle. This works fine for VP2 selection in the UI, but breaks our unit tests (the stage map is only set up AFTER the tests have finished running).

lib/ufe/UsdRotatePivotTranslateUndoableCommand.cpp
```diff
@@ -16,42 +16,40 @@
...
 #include "UsdRotatePivotTranslateUndoableCommand.h"
 #include "private/Utils.h"
+#include "AL/usd/utils/MayaTransformAPI.h"
+#include "../base/debugCodes.h"
 
 MAYAUSD_NS_DEF {
 namespace ufe {
 
-UsdRotatePivotTranslateUndoableCommand::UsdRotatePivotTranslateUndoableCommand(const UsdPrim& prim, const Ufe::Path& ufePath, const Ufe::SceneItem::Ptr& item)
+UsdRotatePivotTranslateUndoableCommand::UsdRotatePivotTranslateUndoableCommand(const UsdPrim& prim, const Ufe::Path& ufePath, const Ufe::SceneItem::Ptr& item, const UsdTimeCode& timeCode)
```
robthebloke: You'll notice a change here, that's repeated a few times. I need to get a time code from *somewhere*, so at the moment, it just pulls the time value from the current time. input to the proxy shape.

I think the code change here is the correct one to make, but that boots the question of where that time value comes from up to the business logic. At the moment that comes from the time line.

Previously it would be based on the push2prim and readanimatedvalues attrs, and the behaviour could change based on a couple of option vars. Fuck knows if I can fully explain the logic, uhm

Business logic point #1, where does the time come from, when is it the default, when can we read/write from the default, and how do we going about fixing it when we realise we've made the wrong choice(s)?

lib/ufe/UsdRotatePivotTranslateUndoableCommand.cpp
```diff
@@ -16,42 +16,40 @@
...
+	return std::make_shared<UsdRotatePivotTranslateUndoableCommand>(prim, ufePath, item, timeCode);
 }
 
 void UsdRotatePivotTranslateUndoableCommand::undo()
 {
-	fPivotAttrib.Set(fPrevPivotValue);
+	TF_DEBUG(MAYAUSD_UFE_MANIPULATORS).Msg("UsdRotatePivotTranslateUndoableCommand::undo %s (%lf, %lf, %lf) @%lf\n",
+	  fPath.string().c_str(), fPrevPivotValue[0], fPrevPivotValue[1], fPrevPivotValue[2], fTimeCode.GetValue());
+	AL::usd::utils::MayaTransformAPI api(fPrim);
+	api.rotatePivot(GfVec3f(fPrevPivotValue), fTimeCode);
```
robthebloke: And so this if where the maya transform api comes into play. I want to provide to TDs a python (or c++) api they can just set a value, with a time, and it just works.

Again, it comes back to business rules. Where does fTimeCode come from?

lib/ufe/private/Utils.cpp
```diff
@@ -126,89 +126,5 @@ UsdGeomXformCommonAPI convertToCompatibleCommonAPI(const UsdPrim& prim)
...
-		}
-		catch (const std::exception& e) {
-			std::string err = TfStringPrintf("Failed to set pivot for prim %s - %s",
-				path.string().c_str(), e.what());
-			UFE_LOG(err.c_str());
-			throw;
-		}
-	}
-}
-
```
robthebloke: Removing because the interface is not verbose enough for our needs.

ppt-adsk:
Hi Rob, I definitely will have a look this week, but today and tomorrow are bad because of internal deadlines here.

ppt-adsk:
Hi Rob, sorry for the delay in answering, Preview Release 113 got in the way, then I had to take a couple of sick days, then I was off on vacation for two days for our spring break.  Answers inline below.  Hope it's still relevant.

> [...]
> `AL_maya_test_UnitTestHarness -ktf -f "ProxyShapeSelect.pseudoRootSelect";`
> 
> At the moment, the above unit test fails. The test effectively just wants to select the root prim path, `SdfPath("/");`. In the old system, this would simply end up selecting the ProxyShape node itself within maya.
> 
> Replicating that behaviour within UFE is causing me a few issues....
> 
>     1. It doesn't seem to be possible to create a path in UFE for the root prim '/' ?

Correct.  There is no such UFE node (and in fact there is no such node in USD either, it's a pseudo-root), therefore there is no such UFE path.  Your previous strategy of selecting the proxy shape node itself is correct, those are the semantics we intend and use: the proxy shape is the "root" of the remaining USD tree.

>     2. I attempted to copy the previous behaviour, namely: `when '/' is selected, select the ProxyShape node using standard maya selection`. This doesn't seem possible with UFE? If you attempt to select "/", UFE says the path is invalid, and wont let you select it. If you modify the selection list _(i.e. add the proxy shape to an MSelectionList, and then call MGlobal::setActiveSelectionList)_, the call to setActiveSelectionList immediately clears the UFE selection, **_and_** inserts the UFE path "AL_usdmaya_ProxyShape1". However I cannot insert that item into UFE myself _(The following paths are both invalid: "AL_usdmaya_ProxyShape1", "|world|transform1|AL_usdmaya_ProxyShape1/", so what is the correct path to use???)

"|world|transform|AL_usdmaya_ProxyShape1" would be the proper path.

> Question 1: **(Question for Autodesk)**
> If i want to select the prim path "/root/foo", UFE will create that item as `"|world|transform1|AL_usdmaya_ProxyShape1/root/foo"`. Cool. If i want to select the path "/", ufe will simply report an invalid path, and throw an error. Ok, so I guess I need to special case that, and simply select the proxy shape via MGlobal::setActiveSelectionList. Except when I do that, UFE now reports "AL_usdmaya_ProxyShape1" as a selected item, but has wiped the rest of the selection list clean. If I attempt to insert the UFE item "AL_usdmaya_ProxyShape1", I can't, because it's not a valid path. There's a chicken & egg thing going on here, and I'm not sure how we are supposed to handle it? So how do we add maya nodes to the UFE selection list, via UFE? or, how do we represent "/" in UFE?

Don't use the Maya selection to make UFE selections.  The UFE selection is a superset of the Maya selection, and we make sure that any Maya selection (that doesn't have components, i.e. points, faces, edges, etc.) is properly reflected in UFE.  However, that also means that if you use the Maya API to make a selection, then the UFE selection must be changed to match, and as you saw this will wipe the UFE selection.  So using the Python API, it would be something like:
import ufe
psPathSegment = ufe.PathSegment("|world|transform1|AL_usdmaya_ProxyShape1", 1, "|")
usdPathSegment = ufe.PathSegment("/foo"), 2, "/")
path = ufe.Path([psPathSegment, usdPathSegment])
item = ufe.Hierarchy.createItem(path)
ufe.GlobalSelection.get().append(item)

> Question 2: **(Question for Bluesky, and other interested parties...)**
> Currently the behaviour of this PR is to pull the time value directly from the proxy shape, and use that time value for each get/set of each transform op. This kinda ruins any ability to be able to get/set the default value, but does at least mean that transform manips follow animated transforms. We need to start deciding the business logic for how we are going to handle this in a way that doesn't lead to a catalogue of confusion.
> 
> Sorry for waffling too much, any help would be appreciated :)

Not sure how to set the default value.  Using the time value directly from the proxy shape sounds good to me, but I can see that authoring the default value is a requirement.  Discussion to be continued, no doubt.

kxl-adsk:
Connecting to open PR #357 which makes the reading part of UFE interface aligned with what you see in the viewport and frame selection. Writing to USD still continues to use default time - here we have some more thinking to do.

kxl-adsk:
Closing since questions were answered and remaining changes seems to be covered by #158 and #371

PR #300, opened 2020-02-24
Improve performance based on suggestions in PR255

kxl-adsk:
Separate PR to address following suggestions from #255 
https://github.com/Autodesk/maya-usd/pull/255#discussion_r378591553
https://github.com/Autodesk/maya-usd/pull/255#discussion_r378591806
https://github.com/Autodesk/maya-usd/pull/255#discussion_r378592716

This PR doesn't address following suggestion, explanation will be added in comments
https://github.com/Autodesk/maya-usd/pull/255#discussion_r378589393

### DIFF
--- a/lib/fileio/translators/translatorMayaReference.cpp
+++ b/lib/fileio/translators/translatorMayaReference.cpp
@@ -130,6 +130,22 @@ namespace {
     name = MString(name.asChar() + 1);
     return name;
   }
+
+  // Cache "associatedNode" attribute to avoid look-up cost. Since UsdMayaTranslatorMayaReference doesn't have initialization method
+  // we use method with static variable to defere the search when everything is initialized.
+  const MObject getAssociatedNodeAttr()
+  {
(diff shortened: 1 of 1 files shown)


lib/fileio/translators/translatorMayaReference.cpp
```diff
@@ -130,6 +130,22 @@ namespace {
     name = MString(name.asChar() + 1);
```
kxl-adsk: Can't leave the comment at line 128...but what was suggested:
```
std::replace(name.asChar(), name.asChar() + name.length(), '|', '_');
```

Mutating data like thing under an object is not recommended. You never know how data is managed and what other data may live alongside.

kxl-adsk:
@robthebloke please review

lib/fileio/translators/translatorMayaReference.cpp
robthebloke: Single Responsibility Principle. If your string class is doing more than simply being a string, it's probably not the best design. It might be worth in that case adding an overload for MString::replace that takes char instead of char* ?

PR #278, opened 2020-02-12
apply updates based on recent core USD dev branch commits

mattyjams:
In particular, this PR absorbs the following core USD commits:
https://github.com/PixarAnimationStudios/USD/commit/055b8532b86f4e327fdca68536bf686d13b88ff3
https://github.com/PixarAnimationStudios/USD/commit/340759c4ef0f3c1588718acf48809e6f5dc44a57

I updated build.md to reflect that the most recent core USD commit supported by maya-usd is the one currently on the tip of the core USD dev branch (https://github.com/PixarAnimationStudios/USD/commit/d0d449aa6cb898f3bc3cfe9eda234d4122dcb993)

PR #301, opened 2020-02-25
Fix copy which becomes illegal in later USD versions.

marsupial:
Somewhere in 22.02 or 22.05 with gcc-6.3 this statement becomes un-compilable

@lgritz


mattyjams:
Thanks @marsupial! This is actually one of the things addressed by #278.

A bunch of new commits were pushed to the core USD dev branch last week, so I have a few additional changes that should go with those. I'm still waiting on one particular core commit to get pushed though, and I'll likely close out #278 in favor of a new one.

marsupial:
Ok, good to know.
Even better as I was just about to start adding the driver support!

PR #304, opened 2020-02-25
MAYA-103070 - Minimal Import UI for PR113

seando-adsk:
#### MAYA-103070 - Minimal Import UI for PR113
* Added checkbox images used in import dialog.
* Added query flag '-primPath' to command 'usdImportDialog'. Use this from import options the primpath.
* When reopening the Import dialog it is now restored with any existing data which matches current file.
* Added extra input flags to test application 'testMayaUsdUI' used to set rootPrimPath and primVarSelection (single one).
* New checkbox behavior. Modified as per design specs.
* From MayaUsd import translator after a sucessful import, clear the import data as we don't want to re-use it on a subsequent import. 
* Added new interface IMayaMQtUtil to deal with running the dialog from both within Maya and standalone. When running standalone we cannot use MQtUtil for things like loading QPixmap or dpi scaling. Created two derived classes to deal with both Maya and standalone.


lib/usd/ui/cmds/USDImportDialogCmd.cpp
```diff
@@ -67,45 +70,65 @@ MStatus USDImportDialogCmd::doIt(const MArgList& args)
 {
 	MStatus st;
 	MArgParser argData(syntax(), args, &st);
+	if (!st) return st;
 
-	if(argData.isFlagSet(kClearDataFlag))
+	if (argData.isQuery())
```
seando-adsk: Note: this diff looks bad, but all I really did was add the isQuery() section and then put the code below into the else (so it all got indented).

lib/usd/ui/views/TreeItem.cpp
```diff
@@ -41,7 +50,36 @@ int TreeItem::type() const
...
 }
 
-void TreeItem::setCheckState(Qt::CheckState st)
+const QPixmap& TreeItem::checkImage() const
+{
+	if (fsCheckBoxOn == nullptr)
+	{
+		const TreeModel* treeModel = qobject_cast<const TreeModel*>(model());
+		if (treeModel != nullptr)
+		{
```
mattyjams: Feels like maybe there should be at least a `TF_VERIFY` (or its equivalent) here? If for whatever reason `treeModel` is invalid here, it looks like we could end up derefencing `nullptr` below?

lib/usd/ui/views/USDImportDialog.cpp
```diff
@@ -36,14 +37,29 @@ USDImportDialog::USDImportDialog(const std::string& filename, QWidget* parent)
 {
 	if (!fStage)
 		throw std::invalid_argument("Invalid filename passed to USD Import Dialog");
+	if (mayaQtUtil == nullptr)
+		throw std::invalid_argument("Invalid IMayaMQtUtil passed to USD Import Dialog");
```
mattyjams: @kxl-adsk Do we have anything in the coding guidelines on using exceptions (or **not** using them)?

I'm not sure what, if anything, catches these?

lib/usd/ui/views/USDQtUtil.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+float USDQtUtil::dpiScale(float size) const
+{
+	return MQtUtil::dpiScale(size);
+}
+
+QPixmap* USDQtUtil::createPixmap(const std::string& imageName) const
+{
+	return MQtUtil::createPixmap(imageName.c_str());
+}
```
seando-adsk: Maya version of the IMayaMQtUtil interface which uses MQtUtil.

lib/usd/ui/views/TreeModel.cpp
```diff
@@ -228,12 +221,77 @@ void TreeModel::openPersistentEditors(QTreeView* tv, const QModelIndex& parent)
...
+		QVector<int> roles;
+		roles << Qt::DecorationRole;
+		emit dataChanged(modelIndex, modelIndex, roles);
+
+		// Then check-disable all the children of the clicked item.
+		setChildCheckState(modelIndex, TreeItem::CheckState::kChecked_Disabled);
+	}
+}
+
+void TreeModel::onItemClicked(TreeItem* item)
```
seando-adsk: Called by the dialog in response to clicking on an item. We use it to toggle the checkstate.

lib/usd/ui/views/TreeModel.cpp
```diff
@@ -228,12 +221,77 @@ void TreeModel::openPersistentEditors(QTreeView* tv, const QModelIndex& parent)
...
+	}
+}
+
+void TreeModel::uncheckEnableTree()
+{
+	// When unchecking any item we uncheck-enable the entire tree.
+	setChildCheckState(QModelIndex(), TreeItem::CheckState::kUnchecked);
+}
+
+void TreeModel::checkEnableItem(TreeItem* item)
```
seando-adsk: New spec from design on checkbox behavior.

lib/usd/ui/views/TreeModel.cpp
```diff
@@ -228,12 +221,77 @@ void TreeModel::openPersistentEditors(QTreeView* tv, const QModelIndex& parent)
...
+		// Note: only the load column (0) has children, so we use it when looking for children.
+		QModelIndex childIndex = this->index(r, kTreeColumn_Load, parent);
 		if (hasChildren(childIndex))
 		{
 			openPersistentEditors(tv, childIndex);
 		}
 	}
 }
 
+void TreeModel::setRootPrimPath(const std::string& path)
```
seando-adsk: Used to initialize the tree from previous import data.

lib/usd/ui/views/ItemDelegate.cpp
```diff
@@ -43,7 +45,26 @@ QWidget* ItemDelegate::createEditor(QWidget* parent, const QStyleOptionViewItem&
 		assert(treeItem != nullptr);
 		if (nullptr != treeItem)
 		{
-			VariantsEditorWidget* editor = new VariantsEditorWidget(parent, this, treeItem->prim());
+			// Look in the import data (from the tree model) to see if it has a prim variant selection
```
seando-adsk: When creating the variant set selection combo boxes, we'll look at import data for matching items to set from.

plugin/adsk/plugin/importTranslator.cpp
```diff
@@ -114,12 +125,15 @@ UsdMayaImportTranslator::reader(
...
-    UsdMaya_ReadJob mUsdReadJob(
-        (fileName != MayaUsd::ImportData::cinstance().filename()) ? importData : MayaUsd::ImportData::cinstance(),
-        jobArgs);
+    UsdMaya_ReadJob mUsdReadJob(importData, jobArgs);
     std::vector<MDagPath> addedDagPaths;
     bool success = mUsdReadJob.Read(&addedDagPaths);
+
+    // After a successful import we clear the import data as we don't want to
+    // re-use it on a subsequent import.
+    if (success)
```
seando-adsk: New behavior from design. This will fix the problem where a user imports from dialog and then runs manual command. Since the import data will be cleared out it won't be re-used for next manual command.

lib/usd/ui/views/USDImportDialog.cpp
```diff
@@ -36,14 +37,29 @@ USDImportDialog::USDImportDialog(const std::string& filename, QWidget* parent)
...
 
 	fUI->setupUi(this);
 
+	// If we were given some import data we will only use it if it matches our
+	// input filename. In the case where the user opened dialog, clicked Apply and
+	// then reopens the dialog we want to reset the dialog to the previous state.
+	const ImportData* matchingImportData = nullptr;
```
seando-adsk: This happens when a user opens the import dialog, sets some variants, changes prim (by checking) to import and clicks "Apply" to close dialog. Then before importing they reopen dialog (clicking Hierarchy View). We will use the import data to set the dialog to it's previous state.

lib/usd/ui/views/USDImportDialog.cpp
kxl-adsk: We have a mention in C++ core guidelines which talks about exceptions and alternative solution. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i10-use-exceptions-to-signal-a-failure-to-perform-a-required-task

kxl-adsk: An additional guideline would be about noexcept  https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-noexcept

lib/usd/ui/cmds/USDImportDialogCmd.cpp
seando-adsk: Yeah good suggestion, I do that.

lib/usd/ui/demo/testMayaUsdUI.cpp
```diff
@@ -26,11 +27,44 @@
 #include <iostream>
 #include <string>
 
+class TestUIQtUtil : public MayaUsd::IMayaMQtUtil
+{
+public:
+	~TestUIQtUtil() override;
```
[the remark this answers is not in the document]
seando-adsk: Yes. I did it for the USDQtUtil version as well.

PR #306, opened 2020-02-25
MAYA-103409 Minimal Create USD Stage

luceric-adsk:
A user currently can't create a USD stage using the UI. This UI will allow the user to create a stage using an existing USD layer.

In Maya's top level menu, under Create, a new entry 'Universal Scene Description (USD)' is underneath 'Scene Management' category.  Please have this in the What's New green highlight style.  On hover, the submenu appears with the following entry:
"Stage from Existing Layer..."
Please have this in the What's New green highlight style.   On click, the following file dialogue appears[...]

### DIFF
--- a/lib/nodes/proxyShapeBase.cpp
+++ b/lib/nodes/proxyShapeBase.cpp
@@ -111,6 +111,7 @@ const MString MayaUsdProxyShapeBase::displayFilterLabel("USD Proxies");
 MObject MayaUsdProxyShapeBase::filePathAttr;
 MObject MayaUsdProxyShapeBase::primPathAttr;
 MObject MayaUsdProxyShapeBase::excludePrimPathsAttr;
+MObject MayaUsdProxyShapeBase::loadPayloadsAttr;
 MObject MayaUsdProxyShapeBase::timeAttr;
 MObject MayaUsdProxyShapeBase::complexityAttr;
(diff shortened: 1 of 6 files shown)


lib/nodes/proxyShapeBase.cpp
```diff
@@ -177,6 +178,20 @@ MayaUsdProxyShapeBase::initialize()
     retValue = addAttribute(excludePrimPathsAttr);
     CHECK_MSTATUS_AND_RETURN_IT(retValue);
 
+    loadPayloadsAttr = numericAttrFn.create(
+        "loadPayloads",
+        "lpl",
+        MFnNumericData::kBoolean,
+        0.0,
```
mattyjams: Shouldn't this attribute be turned on by default? That would preserve existing behavior of loading payloads when opening the stage.

luceric-adsk: let me fix that by doing in the definition line instead

luceric-adsk:
added support for case when user is using native dialog

lib/nodes/proxyShapeBase.cpp
mattyjams: Oh, sorry! I read right past that `setDefault(true)` call.

But yeah, looks a bit cleaner to just have it in the initial definition. Thanks!

PR #311, opened 2020-02-26
[hdMaya] remove MtohGetDefaultRenderer

elrond79:
wasn't used anywhere - believe it was a holdover when we had a single maya renderOverride, and an option to switch between the various hydra renderDelegates; now that there is a 1-to-1 relationship between renderOverrides and renderDelegates, no need for this

Putting do-not-merge-yet on this, because it will create merge conflicts with https://github.com/Autodesk/maya-usd/pull/292, and I'd like to get that merged first.  I'll resolve conflicts here after that is merged - just wanted to create this now, as a bookmark for myself.

### DIFF
--- a/lib/render/mayaToHydra/utils.cpp
+++ b/lib/render/mayaToHydra/utils.cpp
@@ -201,29 +201,5 @@ const MtohRendererSettings& MtohGetRendererSettings() {
     return MtohInitializeRenderPlugins().second;
 }
 
-TfToken MtohGetDefaultRenderer() {
-    const MtohRendererDescriptionVector& plugins = MtohGetRendererDescriptions();
-    if (plugins.empty())
-        return {};
-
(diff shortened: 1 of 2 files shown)


kxl-adsk:
@elrond79 please resolve conflicts

elrond79:
Updated to resolve merge conflicts - sorry for the delay!

elrond79:
Good catch @kxl-adsk ! Pushed a fix.

However, I was surprised that I didn't see this warning/error myself, as I build with `PXR_STRICT_BUILD_MODE=ON`.  Looking closer at this, I learned that for gcc `-Wall` does NOT add `-Wunused-const-variable`.  Should we add this to the list of flags that PXR_STRICT_BUILD_MODE enables?

kxl-adsk:
It makes sense to add it - can you build it first on Linux and open a separate PR for it?

PR #312, opened 2020-02-27
[hdMaya] Build menu UI from TfEnum

marsupial:
Requires:
  https://github.com/PixarAnimationStudios/USD/pull/1122

Likely will conflict with: https://github.com/Autodesk/maya-usd/pull/292, but proof of concept that support outside usdview is trivial / worth it.

### DIFF
--- a/lib/render/mayaToHydra/renderGlobals.cpp
+++ b/lib/render/mayaToHydra/renderGlobals.cpp
@@ -79,6 +79,14 @@ void _CreateEnumAttribute(
     node.addAttribute(o);
 }
 
+void _CreateEnumAttribute(MFnDependencyNode& node, const TfToken& attrName,
+    const TfEnum& defValue) {
+    std::vector<std::string> names = TfEnum::GetAllNames(defValue);
+    TfTokenVector tokens(names.begin(), names.end());
+    return _CreateEnumAttribute(node, attrName, tokens,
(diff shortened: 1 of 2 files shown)


elrond79:
Well, we obviously need to wait for the upstream USD PR to get merged first, but otherwise, looks good to me!

kxl-adsk:
@marsupial please resolve conflicts

marsupial:
This change will actually work all the way back to 19.07.
It's just introducing the convention that a TfEnum stored in a VtValue can be used to build a menu.

PixarAnimationStudios/USD#1122 is just the same convention pushed through a few layers to reach Python & **usdview**

kxl-adsk:
@marsupial Probably when I will read the code it will be obvious, but for now I'm confused. Why did you mention then that this PR requires PixarAnimationStudios/USD#1122 ?

marsupial:
Probably just doing too many things so chose a bad phrasing...

Either way having this -sanctioned- by USD itself is obviously a bigger win then just doing it for Maya alone.

Is it worth opening up a topic on usd-interest?
Do you guys have plans on representing string/int choices in the settings window?

kxl-adsk:
No plans currently on our end. As for usd-interest - I think you have PR open, this should start the discussion with Pixar.

Issue #262, opened 2020-02-07
Selection of a ProxyShape will select individual prims in the hierarchy in VP2.0


dj-mcg:
**Describe the bug**
When selecting a ProxyShape object in Viewport 2.0, a subset of prims in the ProxyShape hierarchy are often selected rather than the entire ProxyShape itself. While there may be workflows where this behavior is helpful, the default behavior should be that the entire Proxy Shape is selected.

**Steps to reproduce**
1. Run the testPorxyShapeSelectionPerformance test. Observe the failed assert

**Expected behavior**
The test passes with no errors.

**Additional context**
This test is part of a suite of tests that are run internally at Pixar. They are not included run as part of the external Pixar usdMaya plugin tests since they are making assumptions about the runtime environment that may not be valid for all users. However, moving forward, these tests should be run internally within Autodesk to validate correctness when testing any changes or fixes to the usdMaya plugin.

kxl-adsk:
The described bug is actually expected behavior. We enable native workflows on USD data and selection is the foundation for many of them.

It sounds like for your workflows, you would like to have control over selection behavior, i.e USD prim vs DG Shape.

kxl-adsk:
@mattyjams assigned this issue to you since you have open PR for it.

santosg87:
Hi @dj-mcg can this one be closed, since the PR was merged on #315

dj-mcg:
I'm fine with it as long as @mattyjams <https://github.com/mattyjams> approves
- thanks!

mattyjams:
Yup, we addressed this with the changes in #315. Ok to close, thanks @santosg87!

PR #315, opened 2020-02-28
add a way for pxrUsdProxyShape to disable subpath selection with the VP2.0 render delegate

mattyjams:
I wanted to float this change as a possible solution to #262. As we're testing the VP2.0 render delegate internally at Pixar, we're trying to ensure that the selection behavior of our proxy shapes is consistent when switching back and forth with our current batch renderer viewport integration. Our pxrUsdProxyShape node (and pxrUsdReferenceAssembly nodes which are the primary way our proxies are introduced) aren't really designed to support subpatch selection, so workflows aiming to expose more of the guts of a USD stage should favor one of the other proxy shape types.

### DIFF
--- a/lib/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -506,6 +506,14 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
     MDagPath& dagPath) const
 {
 #if defined(WANT_UFE_BUILD)
+    if (_proxyShape == nullptr) {
+        return false;
+    }
+
+    if (!_proxyShape->isUfeSelectionEnabled()) {
+        return false;
+    }
+
     // When point snapping, only the point position matters, so return false
     // to use the DAG path from the default implementation and avoid the UFE
     // global selection list to be updated.
@@ -514,9 +522,6 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
 
     auto handler = Ufe::RunTimeMgr::instance().hierarchyHandler(USD_UFE_RUNTIME_ID);
     if (handler == nullptr)
-        return false;
-
-    if (_proxyShape == nullptr)
         return false;
 
     // Extract id of the owner Rprim. A SdfPath directly created from the render
--- a/lib/nodes/proxyShapeBase.cpp
+++ b/lib/nodes/proxyShapeBase.cpp
@@ -955,8 +955,9 @@ MayaUsdProxyShapeBase::parentTransform()
     return proxyTransformPath;
 }
 
-MayaUsdProxyShapeBase::MayaUsdProxyShapeBase() :
-    MPxSurfaceShape()
+MayaUsdProxyShapeBase::MayaUsdProxyShapeBase(const bool enableUfeSelection) :
+        MPxSurfaceShape(),
+        _isUfeSelectionEnabled(enableUfeSelection)
 {
     TfRegistryManager::GetInstance().SubscribeTo<MayaUsdProxyShapeBase>();
 }
--- a/lib/nodes/proxyShapeBase.h
+++ b/lib/nodes/proxyShapeBase.h
@@ -218,9 +218,25 @@ class MayaUsdProxyShapeBase : public MPxSurfaceShape,
         Ufe::Path ufePath() const;
 #endif
 
+        /// Returns whether the proxy shape allows subpaths within its
+        /// hierarchy to be selected independently when using the Viewport 2.0
+        /// render delegate.
+        ///
+        /// UFE/subpath selection must be enabled or disabled when constructing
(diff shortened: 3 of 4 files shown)


kxl-adsk:
I think it makes sense...but I'm getting concerned where it will bring us if we continue to add virtual methods for things like this on the base class. Maybe it's a good time to introduce flags?

mattyjams:
Famous last words, I know, but I'm hoping that this is a fairly temporary thing, and that we're not going to have a huge proliferation of new virtuals we'll want to add. Or at least not for the purposes of the Pixar proxy shape anyway.

Our current aim is to get feature (and performance) parity when using `pxrUsdProxyShape` between the Pixar batch renderer and the VP2.0 render delegate. Once we get there, the next goal would be to start phasing out `pxrUsdProxyShape` and `pxrUsdReferenceAssembly`. The Pixar proxy shape won't really offer new functionality at that point, and assemblies are dead tech, so we'll be working towards adopting `mayaUsdProxyShape` and UFE instead, and we'll want to steer others that way as well.

I'm not sure what you have in mind for flags, but for this particular issue it can't really be a build-time or env var thing. We'll want to be using `pxrUsdProxyShape` **and** testing `mayaUsdProxyShape` both while imaging with the VP2.0 render delegate. The render delegate will need to know how to deal with selection depending on which node type it's working with.

HdC-adsk:
@mattyjams I am not sure whether I understand why you mean by "temporary thing". Does it make sense to add a TODO to remove this code at some point?

The only suggestion is, can you rename the new method to enableUfeSelection(), which is the term we usually use.

HdC-adsk:
I think what @kxl-adsk meant by flags is to introduce a flag in the ProxyRenderDelegate that can be checked in getInstancedSelectionPath(). And it makes sense to me.

kxl-adsk:
Actually, I was trying to find this pattern we use internally but didn't yet have time. What I had in mind is a simple bitfield which you can set during the construction of an object (either from a base class or derived). In this bitfield, you can encode information about different characteristics of your object, like disable_ufe_selection. This will remove the need for a virtual method.

After writing this, I convinced myself that it's too early to do this. So ignore me.

HdC-adsk:
But the performance concern still makes sense. There seems to be no need to call into this virtual method for every selection hit. Thus I would suggest to query the flag [here ](https://github.com/Autodesk/maya-usd/blob/dev/lib/render/vp2RenderDelegate/proxyRenderDelegate.cpp#L395) and use it in selection interpretation.

kxl-adsk:
Right, good point @HdC-adsk !

mattyjams:
Thanks @kxl-adsk and @HdC-adsk for the great feedback here.

I just pushed a new version of this in commit 5c8528e1c38ec3771de42f1054990919a06fa6f6. I kept the virtual function, but I renamed it to `enableUfeSelection()` as @HdC-adsk suggested. It also now gets called and its return value is stored when the `ProxyRenderDelegate` is constructed. We can then just check the member variable during `getInstancedSelectionPath()` and avoid the virtual dispatch. It's **not** expected to vary over the course of a Maya session (or frankly at all), so we don't need to re-query it on every execute.

And in terms of the temporary-ness of this, I think it is temporary insofar as the `pxrUsdProxyShape` node is temporary. We don't really intend to support UFE workflows with `pxrUsdProxyShape` and instead are happy to see its usage gradually replaced with `mayaUsdProxyShape` instead. That led me to think it didn't really warrant a TODO comment, but let me know if you think otherwise.

HdC-adsk:
@mattyjams Makes sense. Thanks.

mattyjams:
@kxl-adsk Ok, I just posted a new version of this in commit a879ac268c63d1e84b761db4e29c0583e56f537c that removes the virtual entirely and adds a parameter to the `MayaUsdProxyShapeBase` constructor instead.

You can compare this to the previous iteration in commit 5c8528e1c38ec3771de42f1054990919a06fa6f6.

lib/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -196,6 +196,10 @@ ProxyRenderDelegate::ProxyRenderDelegate(const MObject& obj)
 
     const MFnDependencyNode fnDepNode(obj);
     _proxyShape = static_cast<MayaUsdProxyShapeBase*>(fnDepNode.userNode());
+
+    _isUfeSelectionEnabled = (_proxyShape != nullptr) ?
```
kxl-adsk: @mattyjams What's the remaining value of this variable given that we have inline access to it on proxy shape?

mattyjams: Good point. None really, I suppose. We can just rely on the compiler to inline away the function call.

Rebased and updated to remove this member variable in commit dab4b46f5900c4a3bffd7d64a1dcb91fc0b949f3.

PR #317, opened 2020-02-28
MAYA-103673 update Create USD Stage menu

fowlertADSK:
Adding optionVars to store Create USD Stage options
Adding a proper Option Box menu item.

### DIFF
--- a/plugin/adsk/scripts/mayaUsdMenu.mel
+++ b/plugin/adsk/scripts/mayaUsdMenu.mel
@@ -65,8 +65,17 @@ proc initRuntimeCommands() {
             -category   "Menu items.Maya USD"
             -command    "mayaUsd_createStagesFromExistingLayer"
             mayaUsdCreateStageFromExistingLayer;
-        source "mayaUsd_createStagesFromExistingLayer.mel";
     }
+
+    if (!`runTimeCommand -exists mayaUsdCreateStageFromExistingLayerOptions`) {
+        runTimeCommand -default true
+            -annotation "Create a USD Stage from an existing USD layer options"
+            -category   "Menu items.Maya USD"
+            -command    "mayaUsd_createStagesFromExistingLayerOptions"
+            mayaUsdCreateStageFromExistingLayerOptions;
+    }    
+
+    source "mayaUsd_createStagesFromExistingLayer.mel";
 }
 
 ///////////////////////////////////////////////////////////////////////////////
@@ -95,6 +104,7 @@ global proc mayaUsdMenu_createMenuCallback() {
                 -annotation "Create a USD stage" 
                 -version $mayaVersion`;
             menuItem -runTimeCommand mayaUsdCreateStageFromExistingLayer;
+            menuItem -runTimeCommand mayaUsdCreateStageFromExistingLayerOptions -optionBox true;
         } else {
             error "Could not create mayaUSD create menu";
         }
--- a/plugin/adsk/scripts/mayaUsd_createStagesFromExistingLayer.mel
+++ b/plugin/adsk/scripts/mayaUsd_createStagesFromExistingLayer.mel
@@ -15,17 +15,17 @@
 
 proc setOptionVars (int $forceFactorySettings)
 {
-	if ($forceFactorySettings || !`optionVar -exists stageFromExistingLayer_primPath`) {
-		optionVar -stringValue stageFromExistingLayer_primPath "";
-	}
+    if ($forceFactorySettings || !`optionVar -exists stageFromExistingLayer_primPath`) {
+        optionVar -stringValue stageFromExistingLayer_primPath "";
+    }
 
-	if ($forceFactorySettings || !`optionVar -exists stageFromExistingLayer_excludePrimPath`) {
-		optionVar -stringValue stageFromExistingLayer_excludePrimPath "";
-	}
+    if ($forceFactorySettings || !`optionVar -exists stageFromExistingLayer_excludePrimPath`) {
+        optionVar -stringValue stageFromExistingLayer_excludePrimPath "";
+    }
 
-	if ($forceFactorySettings || !`optionVar -exists stageFromExistingLayer_loadPayloads`) {
-		optionVar -intValue stageFromExistingLayer_loadPayloads 1;
-	}        
+    if ($forceFactorySettings || !`optionVar -exists stageFromExistingLayer_loadPayloads`) {
+        optionVar -intValue stageFromExistingLayer_loadPayloads 1;
+    }
 }
 
 // stageFromExistingLayer_UISetup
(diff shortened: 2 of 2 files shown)


plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -65,8 +65,17 @@ proc initRuntimeCommands() {
...
-        source "mayaUsd_createStagesFromExistingLayer.mel";
     }
+
+    if (!`runTimeCommand -exists mayaUsdCreateStageFromExistingLayerOptions`) {
+        runTimeCommand -default true
+            -annotation "Create a USD Stage from an existing USD layer options"
+            -category   "Menu items.Maya USD"
+            -command    "mayaUsd_createStagesFromExistingLayerOptions"
+            mayaUsdCreateStageFromExistingLayerOptions;
+    }
```
fowlertADSK: Creating a runtime command + menu to open the Option Box .

plugin/adsk/scripts/mayaUsd_createStagesFromExistingLayer.mel
```diff
@@ -13,17 +13,28 @@
 // limitations under the License.
 //
 
-global string $stageFromExistingLayer_primPath = "";
-global string $stageFromExistingLayer_excludePrimPath = "";
-global int $stageFromExistingLayer_loadPayloads = 1;
+proc setOptionVars (int $forceFactorySettings)
```
fowlertADSK: Changing all of the global variables to optionVar's so they persist.

fowlertADSK: This proc can be called either to init the variables or reset them depending on the forceFactorySettings param.

plugin/adsk/scripts/mayaUsd_createStagesFromExistingLayer.mel
```diff
@@ -13,17 +13,28 @@
...
+	}        
+}
 
 // stageFromExistingLayer_UISetup
 // creates the options of the stageFromExistingLayer dialog
-global proc stageFromExistingLayer_UISetup(string $parent) {
+global proc string stageFromExistingLayer_UISetup(string $parent) {
     setParent $parent;
-    $parent = `scrollLayout -childResizable true`;
+    string $layout = `scrollLayout -childResizable true`;
```
fowlertADSK: Just changing the name of the variable instead of overwriting the one that is passed in because it looks a little cleaner.

plugin/adsk/scripts/mayaUsd_createStagesFromExistingLayer.mel
```diff
@@ -13,17 +13,28 @@
...
 
-    frameLayout -label "USD Layer Options";
+    frameLayout -label "USD Layer Options" -collapsable false;
     textFieldGrp -l "Prim Path:"
```
fowlertADSK: Options are minimal right now and having it be collapsible looks weird for now.

plugin/adsk/scripts/mayaUsd_createStagesFromExistingLayer.mel
```diff
@@ -36,31 +47,35 @@ global proc stageFromExistingLayer_UISetup(string $parent) {
         -ann "When on, loads all prims marked as payloads. When off, all prims marked as payloads and their children are not loaded." 
         -sbm "Loads prims marked as payloads" 
         loadPayloadsCheckBox;
+
+    return $layout;
```
fowlertADSK: Change the proc to return the layout it's creating so that the option box can attach it correctly.  When this same proc is called from the File Dialog UI it's handled a bit differently and doesn't need it.

plugin/adsk/scripts/mayaUsd_createStagesFromExistingLayer.mel
```diff
@@ -71,19 +86,99 @@ proc doCreateStage(string $fileName) {
...
-    setAttr ($shapeNode+".loadPayloads") $stageFromExistingLayer_loadPayloads;
+    setAttr -type "string" ($shapeNode+".primPath") $ppath;
+    setAttr -type "string" ($shapeNode+".excludePrimPaths") $exppath;
+    setAttr ($shapeNode+".loadPayloads") $loadp;
     connectAttr time1.outTime ($shapeNode+".time");
     select -r $shapeNode;
 }
 
+global proc createUSdStagesFromExistingCallback(string $parent, int $doIt)
+{
```
fowlertADSK: This is standard option box code...  Maya can call this proc to just store the current settings (Save and Close), or store it and execute the operation.

plugin/adsk/scripts/mayaUsd_createStagesFromExistingLayer.mel
```diff
@@ -92,14 +187,10 @@ global proc mayaUsd_createStagesFromExistingLayer() {
         -okCaption "Create"
         -optionsUICreate "stageFromExistingLayer_UISetup"
         -optionsUIInit "stageFromExistingLayer_UIInit"
-        -optionsUICommit2 "stageFromExistingLayer_UICommit"`;
+        -optionsUICommit "stageFromExistingLayer_UICommit"`;
```
fowlertADSK: We didn't need the "2" version of this flag since that's the one you use when you need the selected file.  We don't actually use this yet, and some of the code was going to be calling it with bogus empty strings, so I changed back to the flag that matches the behaviour for now since it's a bit more clear.

plugin/adsk/scripts/mayaUsd_createStagesFromExistingLayer.mel
```diff
@@ -71,19 +86,99 @@ proc doCreateStage(string $fileName) {
...
+}
+
+global proc mayaUsd_createStagesFromExistingLayerOptions()
+{
+    global int $gOptionBoxTemplateFrameSpacing;
+    
+	setOptionVars(false);
+	
+	string $commandName = "createUSdStagesFromExisting";
+	string $callback = ($commandName + "Callback");
```
mattyjams: Looks like some inconsistent whitespacing through here.

plugin/adsk/scripts/mayaUsd_createStagesFromExistingLayer.mel
```diff
@@ -71,19 +86,99 @@ proc doCreateStage(string $fileName) {
...
-    setAttr -type "string" ($shapeNode+".excludePrimPaths") $stageFromExistingLayer_excludePrimPath;
-    setAttr ($shapeNode+".loadPayloads") $stageFromExistingLayer_loadPayloads;
+    setAttr -type "string" ($shapeNode+".primPath") $ppath;
+    setAttr -type "string" ($shapeNode+".excludePrimPaths") $exppath;
+    setAttr ($shapeNode+".loadPayloads") $loadp;
     connectAttr time1.outTime ($shapeNode+".time");
     select -r $shapeNode;
 }
 
+global proc createUSdStagesFromExistingCallback(string $parent, int $doIt)
```
mattyjams: The `USd` in the function name looks a little odd here and in the function below, and seems a little inconsistent with other functions in this file that just say `createStage...` or whatever. Maybe just drop the "USd"?

plugin/adsk/scripts/mayaUsd_createStagesFromExistingLayer.mel
fowlertADSK: Yeah, looks like I bounced between editors that had different settings for tabs/spaces.  I'll clean it up.

plugin/adsk/scripts/mayaUsd_createStagesFromExistingLayer.mel
fowlertADSK: This is what happens when you type something by hand and then copy/paste it around.  Also like removing the "USD" entirely to match other functions.

fowlertADSK:
I cleaned up the white-space and got rid of the "USd" in those optionVar's.

PR #318, opened 2020-02-28
Qt support for older versions of Maya

seando-adsk:
Added Qt support for older versions of Maya by protecting some code with Qt version checks. Updated build doc.


CMakeLists.txt
```diff
@@ -103,8 +103,15 @@ if(CMAKE_WANT_UFE_BUILD)
 endif()
 
 if(DEFINED QT_LOCATION)
+    if(NOT DEFINED QT_VERSION)
```
seando-adsk: Normally we'll automatically set the Qt version based on the Maya you are building for. But allow using a specific version of Qt by setting QT_VERSION.

doc/build.md
```diff
@@ -114,7 +112,6 @@ BUILD_AL_PLUGIN             | builds the Animal Logic USD plugin and libraries.
 BUILD_HDMAYA                | builds the Maya-To-Hydra plugin and scene delegate.        | ON
 BUILD_TESTS                 | builds all unit tests.                                     | ON
 CMAKE_WANT_UFE_BUILD        | enables building with UFE (if found).                      | ON
-CMAKE_WANT_QT_BUILD         | enables building with Qt (if found).                       | ON
```
seando-adsk: This cmake option was previously removed.

PR #321, opened 2020-03-02
[cmake] remove duplicate def of mayaUsd_promoteHeaderListWithSubdir

elrond79:


kxl-adsk:
I believe we have this in PR which will show up soon (i.e. cleanup of CMakeLists). @HamedSabri-adsk can you confirm?

hamedsabri:
Yes, this is already cleaned up in my PR which should hopefully show up sometimes tomorrow.

HamedSabri-adsk:
@elrond79 This change is covered in #323.

PR #326, opened 2020-03-04
MAYA-103736 Clear frameSamples set

fowlertADSK:
Clear frameSamples set so that we don't add in the default value when the flag is used.


plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -93,6 +93,7 @@ UsdMayaExportTranslator::writer(const MFileObject &file,
                 theOption[1].split(',', filteredTypes);
             }
             else if (argName == "frameSample") {
```
kxl-adsk: Does this mean we can have the same argument name twice?
Additionally, do we have additional copy of this code somewhere? (potentially Pixar plugin?)

PR #328, opened 2020-03-04
MAYA-103731 - OAST: Jinja 2.10 WS report of CVE-2019-10906

seando-adsk:
#### MAYA-103731 - OAST: Jinja 2.10 WS report of CVE-2019-10906
* Update to Jinja 2.11.1

### DIFF
--- a/cmake/jinja.cmake
+++ b/cmake/jinja.cmake
@@ -41,7 +41,7 @@ function(init_jinja)
             message(FATAL_ERROR "JINJA_LOCATION not set")
         endif()
 
-        set(JINJA_ROOT "${JINJA_LOCATION}")
+        set(JINJA_ROOT "${JINJA_LOCATION}/src")
 
         # Add Jinja2 to the python path so that usdGenSchemas can run properly.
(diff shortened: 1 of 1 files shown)


cmake/jinja.cmake
```diff
@@ -41,7 +41,7 @@ function(init_jinja)
             message(FATAL_ERROR "JINJA_LOCATION not set")
         endif()
 
-        set(JINJA_ROOT "${JINJA_LOCATION}")
+        set(JINJA_ROOT "${JINJA_LOCATION}/src")
```
seando-adsk: In our internal Maya builds we use the .tar.gz files for both MarkupSafe & Jinja2. We extract these and add them to the PYTHONPATH. The Jinja 2.10 file was incorrectly packaged as the egg was not placed in the 'src' folder. That was fixed for the 2.11.1 download. So adjust the path here (notice it now matches the markupsafe path on line 27.

HamedSabri-adsk: Great fix. Thanks @seando-adsk

PR #329, opened 2020-03-04
Replace PXRUSDMAYA_DEFINE_READER macro with class declaration/definition to make the code easier to read and debug.

HamedSabri-adsk:
### Description of problem:
When using PXRUSDMAYA_DEFINE_READER Macro it is a bit hard to follow the code logic and debug.

### DIFF
--- a/lib/usd/translators/meshReader.cpp
+++ b/lib/usd/translators/meshReader.cpp
@@ -13,6 +13,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+// Modifications copyright (C) 2020 Autodesk
+//
+
 #include "pxr/pxr.h"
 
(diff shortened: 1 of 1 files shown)


mattyjams:
Looks good to me, thanks @HamedSabri-adsk!

For some historical context, `PXRUSDMAYA_DEFINE_READER` was the initial way of declaring a reader before we had a registry for import more like the one we have for export. Ideally, everything still using `PXRUSDMAYA_DEFINE_READER` would get converted to being class-based and we'd get rid of the `PXRUSDMAYA_DEFINE_READER` macro, but I'm ok doing that conversion in pieces.

HamedSabri-adsk:
@mattyjams Great and thanks for the historical context.

PR #345, opened 2020-03-09
adapt to UsdImagingDelegate::PopulateSelection param change

mattyjams:
With `USD_IMAGING_API_VERSION` 11, `UsdImagingDelegate::PopulateSelection()` expects to be given a USD path instead of a Hydra "index path". This helps to avoid some confusion about
the types of paths being passed around.

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -33,7 +33,7 @@ See Pixar's official github page for instructions on how to build USD: https://g
 
 |               |      ![](images/pxr.png)          |        
 |:------------: |:---------------:                  |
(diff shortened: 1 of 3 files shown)


kxl-adsk:
@mattyjams We missed our internal update of USD by a few days. I don't believe there is any rush in this PR, so we can keep it open for a week or two?

mattyjams:
@kxl-adsk This change shouldn't require the latest version of USD, as the "new" code is gated on `USD_IMAGING_API_VERSION`, but if you need to put it on hold for a while, then that's ok. We need this internally, so we're merging it into what we build and deploy.

I just rebased it following the merge of #346.

kxl-adsk:
How often USD_IMAGING_API_VERSION is updated? At every breaking change?

mattyjams:
I'm not quite sure exactly how rigorous we're being about that, but at least for this particular change, the API version was bumped for this specific API change. The core USD commit was https://github.com/PixarAnimationStudios/USD/commit/760904c80d8c38d83a4fca079d2414b8288e22cf.

kxl-adsk:
Roger that. Then we should be able to get it in sooner. Thank you for clarification @mattyjams

PR #347, opened 2020-03-09
Fix testPxrUsdPreviewSurfaceExport failure on Windows

HamedSabri-adsk:
Hi @mattyjams,

Can you please confirm if `${TEST_INSTALL_PREFIX}/maya/plugin/pxrUsdPreviewSurface/resources` entry is needed ?

The separator for env variables on Windows is ";"  and It doesn't seem that testwrapper.py is currently supporting that?

I removed  `${TEST_INSTALL_PREFIX}/maya/plugin/pxrUsdPreviewSurface/resources` all together and the test passes on Linux.

### DIFF
--- a/plugin/pxr/maya/plugin/pxrUsdPreviewSurface/CMakeLists.txt
+++ b/plugin/pxr/maya/plugin/pxrUsdPreviewSurface/CMakeLists.txt
@@ -89,7 +89,7 @@ pxr_register_test(testPxrUsdPreviewSurfaceExport
     TESTENV testPxrUsdPreviewSurfaceExport
     ENV
         MAYA_PLUG_IN_PATH=${TEST_INSTALL_PREFIX}/maya/plugin
-        MAYA_SCRIPT_PATH=${TEST_INSTALL_PREFIX}/maya/lib/usd/usdMaya/resources:${TEST_INSTALL_PREFIX}/maya/plugin/pxrUsdPreviewSurface/resources
+        MAYA_SCRIPT_PATH=${TEST_INSTALL_PREFIX}/maya/lib/usd/usdMaya/resources
         MAYA_DISABLE_CIP=1
         MAYA_NO_STANDALONE_ATEXIT=1
         MAYA_APP_DIR=<PXR_TEST_DIR>/maya_profile
--- a/plugin/pxr/maya/plugin/pxrUsdPreviewSurface/CMakeLists.txt
+++ b/plugin/pxr/maya/plugin/pxrUsdPreviewSurface/CMakeLists.txt
@@ -83,13 +83,22 @@ pxr_install_test_dir(
 )
 set(TEST_INSTALL_PREFIX "${PXR_INSTALL_PREFIX}")
 
+set(testMayaScriptPath
+    ${TEST_INSTALL_PREFIX}/maya/lib/usd/usdMaya/resources
+    ${TEST_INSTALL_PREFIX}/maya/plugin/pxrUsdPreviewSurface/resources
+)
+
+if (NOT IS_WINDOWS)
+    string(REPLACE ";" ":" testMayaScriptPath "${testMayaScriptPath}")
+endif()
+
 pxr_register_test(testPxrUsdPreviewSurfaceExport
     CUSTOM_PYTHON ${MAYA_PY_EXECUTABLE}
     COMMAND "${TEST_INSTALL_PREFIX}/tests/testPxrUsdPreviewSurfaceExport"
     TESTENV testPxrUsdPreviewSurfaceExport
     ENV
         MAYA_PLUG_IN_PATH=${TEST_INSTALL_PREFIX}/maya/plugin
-        MAYA_SCRIPT_PATH=${TEST_INSTALL_PREFIX}/maya/lib/usd/usdMaya/resources
+        MAYA_SCRIPT_PATH=${testMayaScriptPath}
         MAYA_DISABLE_CIP=1
         MAYA_NO_STANDALONE_ATEXIT=1
         MAYA_APP_DIR=<PXR_TEST_DIR>/maya_profile
--- a/plugin/pxr/maya/plugin/pxrUsdPreviewSurface/CMakeLists.txt
+++ b/plugin/pxr/maya/plugin/pxrUsdPreviewSurface/CMakeLists.txt
@@ -83,14 +83,12 @@ pxr_install_test_dir(
 )
 set(TEST_INSTALL_PREFIX "${PXR_INSTALL_PREFIX}")
 
+# TODO: disabling the second entry since currently testWrapper.py on Windows
+# can't handle the cases where there are multiple value entries. 
 set(testMayaScriptPath
     ${TEST_INSTALL_PREFIX}/maya/lib/usd/usdMaya/resources
-    ${TEST_INSTALL_PREFIX}/maya/plugin/pxrUsdPreviewSurface/resources
+    #${TEST_INSTALL_PREFIX}/maya/plugin/pxrUsdPreviewSurface/resources
 )
-
-if (NOT IS_WINDOWS)
-    string(REPLACE ";" ":" testMayaScriptPath "${testMayaScriptPath}")
-endif()
 
 pxr_register_test(testPxrUsdPreviewSurfaceExport
     CUSTOM_PYTHON ${MAYA_PY_EXECUTABLE}


mattyjams:
@HamedSabri-adsk I think as it stands now, the `.../pxrUsdPreviewSurface/resources` entry isn't technically required for this particular test. I mainly included it for completeness, since that directory contains plugin files generated by the build. New tests are often added by copy/pasta, so I wanted to make the original as complete as possible.

The only resource file currently in there that could even possibly be relevant as a `MAYA_SCRIPT_PATH` file is the attribute editor `AEpxrUsdPreviewSurfaceTemplate.mel` file, but it doesn't matter for this test since we're only exercising export. For the rest of the resources, the plugInfo.json is discovered via the `plug` library, and that's what makes the fragment XML files available.

So though we could technically remove that entry, I might argue that doing that isn't *strictly* correct. Maybe it would make sense for individually loadable plugins like this one to create a Maya module file? That way the plugin has one canonical declaration of all the env vars needed to point Maya at all of it's pieces. Curious to hear what you think!

mattyjams:
Ah, and sorry I missed your question about `testWrapper.py`. It looks like it's intending to support `;` in path variables on Windows?:
https://github.com/Autodesk/maya-usd/blob/1bf461bd88e17b76aa97df0ceb61d65c862d18a4/plugin/pxr/cmake/macros/testWrapper.py#L249

HamedSabri-adsk:
>  I mainly included it for completeness, since that directory contains plugin files generated by the build. New tests are often added by copy/pasta, so I wanted to make the original as complete as possible.

Great, thanks for confirming. 

> So though we could technically remove that entry, I might argue that doing that isn't strictly correct.
 
I agree. I removed that entry just for the sake of opening the PR and starting the discussion. 

> Maybe it would make sense for individually loadable plugins like this one to create a Maya module file? That way the plugin has one canonical declaration of all the env vars needed to point Maya at all of it's pieces. 

I wonder how much work it would be involved if we want to introduce this Maya module file and even if we have to since we are planning to bring the Pixar unit tests to lib and eventually abandon the testWrapper.py?

Let me explain the issue in bit more detail in this particular case and hopefully it makes things more clear and my apologies if it is long:

The trouble comes when we start parsing the `envvars` entries:
```
    # Add any envvars specified with --env-var options into the environment
    env = os.environ.copy()
    for varStr in args.envVars:
        try:
            k, v = varStr.split('=', 1)
            if k == 'PATH': 
                sys.stderr.write("Error: use --pre-path or --post-path to edit PATH.")
                sys.exit(1)
            v = v.replace('<PXR_TEST_DIR>', testDir)
            env[k] = v
        except IndexError:
            sys.stderr.write("Error: envvar '{0}' not of the form "
                             "key=value".format(varStr))
            sys.exit(1)
```
The `ENV` entries are parsed as a key,value pair with both as `str `type.
```
print "K Type= ", type(k)   // str
print "V Type= ", type(v)   // str
```
When you added the second entry to `MAYA_SCRIPT_PATH ` which by the way you are the first one to do so, testWrapper doesn't really know how to handle the separator because the value is always a string.
```
Key = MAYA_SCRIPT_PATH
Value= C:/workspace/install/RelWithDebInfo/plugin/pxr/maya/lib/usd/usdMaya/resources:C:/workspace/install/RelWithDebInfo/plugin/pxr/maya/plugin/pxrUsdP
```
and eventually the `--env-var` is set like below and Windows doesn't know how to deal with ` : `

`--env-var=MAYA_SCRIPT_PATH=C:/workspace/install/RelWithDebInfo/plugin/pxr/maya/lib/usd/usdMaya/resources:C:/workspace/install/RelWithDebInfo/plugin/pxr/maya/plugin/pxrUsdPreviewSurface/resources`

Now, 
`PATH`, `PXR_PLUGINPATH_NAME`, `PYTHONPATH ` are treated differently. :) If you look at any of the pxr_register_test's ENV entries, you won't find any traces of them.

testWrapper in fact prohibits setting 'PATH' as an entry 
```
if k == 'PATH': 
	sys.stderr.write("Error: use --pre-path or --post-path to edit PATH.")
	sys.exit(1)
```
and the following logics applies when it comes to consolidating the prepath, path, and postPaths and has no real affect on ENV entries:
```
    # Modify the PATH env var.  The delimiter depends on the platform.
    pathDelim = ';' if platform.system() == 'Windows' else ':'
    paths = env.get('PATH', '').split(pathDelim)
    paths = args.prePaths + paths + args.postPaths
    env['PATH'] = pathDelim.join(paths)
```
Now to even make things even more complicated, the separator logic for `prepath`, `PYTHONPATH`, `PXR_PLUGINPATH_NAME` are all handled in the `Public.cmake` 
```
       if(WIN32)
            string(REGEX REPLACE "\\\\" "/" _testPythonPath "${_testPythonPath}")
            string(REGEX REPLACE "\\\\" "/" _testPluginPath "${_testPluginPath}")
            string(REGEX REPLACE "\\\\" "/" _testPrePath "${_testPrePath}")
        else()
            string(REPLACE ";" ":" _testPythonPath "${_testPythonPath}")
            string(REPLACE ";" ":" _testPluginPath "${_testPluginPath}")
            string(REPLACE ";" ":" _testPrePath "${_testPrePath}")
        endif()
```
I hope this makes the problem a bit more clear. Please do let me know if any of the stuff I wrote is not clear.

mattyjams:
@HamedSabri-adsk Ahh, I think I'm following you now. And yeah, I can see how that would break on Windows.

It still seems like for correctness we should include both/all relevant paths though. Maybe in the short term, something like this in the pxrUsdPreviewSurface CMakeLists.txt?:

```
set(TEST_INSTALL_PREFIX "${PXR_INSTALL_PREFIX}")

set(testMayaScriptPath
    ${TEST_INSTALL_PREFIX}/maya/lib/usd/usdMaya/resources
    ${TEST_INSTALL_PREFIX}/maya/plugin/pxrUsdPreviewSurface/resources)

if (NOT IS_WINDOWS)
    string(REPLACE ";" ":" testMayaScriptPath "${testMayaScriptPath}")
endif()

...

pxr_register_test(testPxrUsdPreviewSurfaceExport
...
    ENV
        MAYA_SCRIPT_PATH=${testMayaScriptPath}
...
```

Longer term though, it still seems like it would be good to have some pattern for how individual plugins specify all the bits of Maya configuration that need to be set to bring in all of their parts, and a way to make sure they all layer cleanly when using multiple plugins.

HamedSabri-adsk:
Hey @mattyjams, I like your short term plan. Thanks. Please see e176f3d77c1

I agree with your longer term plan and it makes sense. I would still like to mention that Paul's PR https://github.com/Autodesk/maya-usd/pull/242 should help us with having one set of functionalities to handle unit tests for across all projects (AL, Pixar, Autodesk ).

HamedSabri-adsk:
Oops! sorry that's my bad. I should have ran the test before checking in the code. Ouch!

I had a quick look at this and it turned out that the trouble happens when testWrapper tries to parse the `varStr` argument

CMakeLists.txt
```
set(testMayaScriptPath
    ${TEST_INSTALL_PREFIX}/maya/lib/usd/usdMaya/resources
    ${TEST_INSTALL_PREFIX}/maya/plugin/pxrUsdPreviewSurface/resources
)
MAYA_SCRIPT_PATH=${testMayaScriptPath}
```
Below is how testWrapper.py sees this list entry when reading `varStr` argument. The second entry in testMayaScriptPath list is a new string without a key `MAYA_SCRIPT_PATH` and `=`
```
for varStr in args.envVars:
    try:
        print varStr
        k, v = varStr.split('=', 1)
```
```
MAYA_SCRIPT_PATH=C:/workspace/install/RelWithDebInfo/plugin/pxr/maya/lib/usd/usdMaya/resources
C:/workspace/install/RelWithDebInfo/plugin/pxr/maya/plugin/pxrUsdPreviewSurface/resources
```
which explains this error:
```
 k, v = varStr.split('=', 1)
 ValueError: need more than 1 value to unpack
```

kxl-adsk:
I remember seeing this code. What are the next steps @mattyjams , @HamedSabri-adsk ?

HamedSabri-adsk:
Hey @mattyjams, I was thinking that for the sake of not making any further changes in `testWrapper.py` , we could just temporarily comment out the second entry for now to pass this blocker. Once switching to #242 , we can put back the second entry. Would this work for you?

mattyjams:
Ahh... Is the issue that for the Windows case the variable remains a list because we don't coerce it to a string? Maybe it needs to be something like this (untested)?:

```
if (IS_WINDOWS)
    set(testMayaScriptPath "${testMayaScriptPath}")
else()
    string(REPLACE ";" ":" testMayaScriptPath "${testMayaScriptPath}")
endif()
```

If that doesn't work, then I'm ok just removing the second entry for now.

HamedSabri-adsk:
> if (IS_WINDOWS)
>     set(testMayaScriptPath "${testMayaScriptPath}")
> else()
>     string(REPLACE ";" ":" testMayaScriptPath "${testMayaScriptPath}")
> endif()

Hey @mattyjams, unfortunately that doesn't work. The second entry is still parsed as a new string and since this can't be a `Key,Value pair` entry, then we have a problem:
```
MAYA_SCRIPT_PATH=C:/workspace/install/RelWithDebInfo/plugin/pxr/maya/lib/usd/usdMaya/resources
C:/workspace/install/RelWithDebInfo/plugin/pxr/maya/plugin/pxrUsdPreviewSurface/resources
```
```
k, v = varStr.split('=', 1)
ValueError: need more than 1 value to unpack
```
I am going to remove the second entry and leave a comment in there for now.

PR #351, opened 2020-03-11
PXR_STRICT_BUILD_MODE turns on -Wunused-const-variable

elrond79:
this makes behavior between gcc and clang more similar, as clang enables this warning when -Wall is used, but gcc does not

First discussed here: https://github.com/Autodesk/maya-usd/pull/311#issuecomment-597189785

### DIFF
--- a/cmake/defaults/gccclangshareddefaults.cmake
+++ b/cmake/defaults/gccclangshareddefaults.cmake
@@ -14,16 +14,20 @@
 # limitations under the License.
 #
 
-# This file contains a set of flags/settings shared between our 
+# This file contains a set of flags/settings shared between our
 # GCC and Clang configs. This allows clangdefaults and gccdefaults
 # to remain minimal, marking the points where divergence is required.
 include(Options)
 
-# Turn on C++11; pxr won't build without it. 
+# Turn on C++11; pxr won't build without it.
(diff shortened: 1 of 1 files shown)


elrond79:
Hmm - so just built this, and got a number of errors - ie, from gtest and AL (from EventHandler.h).

I had thought from the reading the docs of clang that this was enabled (ie, [-Wall](https://clang.llvm.org/docs/DiagnosticsReference.html#wall) enables [-Wmost](https://clang.llvm.org/docs/DiagnosticsReference.html#wmost), which enables [-Wunused](https://clang.llvm.org/docs/DiagnosticsReference.html#wunused), which enables [-Wunused-variable](https://clang.llvm.org/docs/DiagnosticsReference.html#wunused-variable), which enables [-Wunused-const-variable](https://clang.llvm.org/docs/DiagnosticsReference.html#wunused-const-variable)), but perhaps this is not the case?  I'm going to close this PR for now, since I don't want to bother resolving me... but @kxl-adsk , I'm curious how you got the [error you mentioned], but didn't get the others I was seeing?

kxl-adsk:
@elrond79 no better answer than that there have to be more differences in warning configuration for the two compilers.

PR #356, opened 2020-03-13
Clone from existing shader instance to avoid extra cost.

HdC-adsk:
If the fallback shader map is not empty, clone any existing shader instance in the map instead of acquiring via `MShaderManager::getFragmentShader()`, which creates new shader fragment graph for each shader instance and causes expensive shader compilation and rebinding.

This change will speed up the first frame loading time considerably.

### DIFF
--- a/lib/render/vp2RenderDelegate/render_delegate.cpp
+++ b/lib/render/vp2RenderDelegate/render_delegate.cpp
@@ -260,9 +260,9 @@ namespace
             // map instead of acquiring via MShaderManager::getFragmentShader(),
             // which creates new shader fragment graph for each shader instance
             // and causes expensive shader compilation and rebinding.
-            auto iter = shaderMap._map.cbegin();
-            if (iter != shaderMap._map.cend()) {
-                shader = iter->second->clone();
+            it = shaderMap._map.begin();
+            if (it != shaderMap._map.end()) {
+                shader = it->second->clone();
             }
             else {
                 MHWRender::MRenderer* renderer = MHWRender::MRenderer::theRenderer();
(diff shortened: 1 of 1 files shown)


williamkrick:
I guess it doesn't really change the difficultly of varying the shaders more in the future.  If they have different parameters to set we'll have to handle that similar to color, and if the shaders really are different we'll have to update the search code to find a reasonable match to clone from.  But we can do all that when we need it.

We could add some debug code where in a debug build we clone from the map AND create one, and make sure they match.  That why if we do start varying the shaders more we'll immediately find this bit of code that needs to be updated.

HdC-adsk:
Yes. Currently fallback shader is only assigned to USD objects without material assignment and the only variant for shading is displayColor/displayOpacity only. So we can leave it as is for now.

For those objects with material assignment we are also calling the `MShaderManager::getFragmentShader()` API but that part needs more thinking on how to share "shader" among shader instances if the material network is identical. So I leave that part unchanged for now.

kxl-adsk:
@pilarmolinalopez is this something you would like to test before we merge?

HdC-adsk:
@pilarmolinalopez Thanks for validation.

Although I do expect some speedup for the playback framerate, the major speedup from the change is for the first frame loading time, esp. for large assets. In some of our large assets, we see 145s to 35s speedup on slower machine and 90s to 20s speedup on faster machine. Did you observe similar speedup for the first frame? Thanks again.

pilarmolinalopez:
I'll test that!

pilarmolinalopez:
hi @HdC-adsk - as you predicted, I'm seeing big speedup on first frame loading time:

- dev branch commit 00d651a (03/16/2020): First frame loading time = 12.63s
- this pr: First frame loading time = 5.96s

HdC-adsk:
Awesome. Thanks again for validation. Appreciated!

PR #358, opened 2020-03-16
Clean up build/usage requirements for hdMaya and mayaToHydra targets.

HamedSabri-adsk:
This is the first of many small PRs to clean up build/usage requirements logics for targets inside maya-usd/lib.


lib/usd/hdMaya/utils.cpp
```diff
@@ -14,7 +14,7 @@
 // limitations under the License.
 //
 
-#include <hdMaya/utils.h>
```
kxl-adsk: Utils is not a private header file. @HamedSabri-adsk please revert this change.

lib/usd/hdMaya/CMakeLists.txt
```diff
@@ -56,53 +53,17 @@ target_compile_definitions(${TARGET_NAME}
...
-        ${TBB_INCLUDE_DIRS}
-        ${CMAKE_BINARY_DIR}
 )
 
 # -----------------------------------------------------------------------------
 # link libraries
 # -----------------------------------------------------------------------------
 target_link_libraries(${TARGET_NAME} 
     PUBLIC 
-        arch
```
kxl-adsk: Nice!

lib/usd/hdMaya/utils.cpp
HamedSabri-adsk: @kxl-adsk Can you clarify what private header mean here?

We have a utils.h and utils.cpp sitting besides each other. Why do I need to do `hdMaya/utils.h` instead of simply `utils.h`

Besides, I have `${CMAKE_CURRENT_SOURCE_DIR}` added to `target_include_directories`  as `PRIVATE` requirement in `lib/usd/hdMaya/CMakeLists.txt`. 

```
target_include_directories(${TARGET_NAME} 
    PRIVATE
        ${CMAKE_BINARY_DIR}/include
        ${CMAKE_CURRENT_SOURCE_DIR}
)
```

elrond79: Hmm... so I was curious if this was / could be made a private header or not - and I think we may need to it to be public, because any hdMaya adapters that others might wish to implement will need/want it.

Currently, it's effectively public, since it's installed.  However, it's referenced in many places using "../hdMaya/utils.h" syntax, which is slightly confusing, and will need to be fixed to get everything in line with our style conventions.

HamedSabri-adsk: > 
> 
> Hmm... so I was curious if this was / could be made a private header or not - and I think we may need to it to be public, because any hdMaya adapters that others might wish to implement will need/want it.
> 
> Currently, it's effectively public, since it's installed. However, it's referenced in many places using "../hdMaya/utils.h" syntax, which is slightly confusing, and will need to be fixed to get everything in line with our style conventions.

Please see https://github.com/Autodesk/maya-usd/pull/369 and https://github.com/Autodesk/maya-usd/pull/370

PR #364, opened 2020-03-18
Fix for installing the gtest shared library on Linux/Mac.

seando-adsk:


seando-adsk:
I wonder if we should also change all the ${GTEST_LIBRARIES} -> GTest::GTest and remove the ${GTEST_INCLUDE_DIRS}?

Like what was done here:
https://github.com/Autodesk/maya-usd/pull/323/commits/01122bba2fc8d7c0ed9839e445c1358f3bbfcc8a

PR #365, opened 2020-03-18
Merge v0.1.0 to master

kxl-adsk:


kxl-adsk:
Indeed, but this impacts only our configuration.

HamedSabri-adsk:
lgtm. project builds, plugins load, and tests pass at least on Linux and MacOS (Release).

@kxl-adsk Regarding the Debug build, if you think this Ok I am going to remove the change request.

kxl-adsk:
Thank you for testing and please remove the change request. We have some thinking to do before solving the whole debug python issue.

PR #368, opened 2020-03-19
Tremblp/maya 103295/ufe path string usd

ppt-adsk:
This branch adds in support for UFE path strings in the mayaUsd plugin.  Most of the heavy lifting for path strings is done in the UFE library itself; UFE clients simply have to register their path component separator to the UFE library.  From the UFE Doxygen documentation:

In UFE, \ref Ufe::Path "paths" are full-fledged objects that are accessed and used through their object interfaces, in Python and C++.  To support scripting environments where such Python or C++ access is inconvenient or simply not possible, UFE provides an optional \ref Ufe::PathString "path string" service, which converts a parsable UFE path string to a UFE path.  A UFE path string has its segments separated by a path separator string, which by default is the single-character comma string ','.  In this way, UFE paths can be transported and manipulated as strings.  Converting a UFE path to a parsable UFE path string is also provided.  To use the path string service, a run-time registers the path component separator(s) it understands at initialization time, using \ref Ufe::PathString::registerPathComponentSeparator() .

Here is an example UFE path string, taken from the test in this pull request:

"|transform1|proxyShape1,/Room_set/Props/Ball_35"

Essentially, it is just a string representation of the UFE path, with UFE path segments separated by the path segment separator ','.

### DIFF
--- a/lib/ufe/Global.cpp
+++ b/lib/ufe/Global.cpp
@@ -33,6 +33,9 @@
 #if UFE_PREVIEW_VERSION_NUM >= 2009
 #include "UsdContextOpsHandler.h"
 #endif
+#if UFE_PREVIEW_VERSION_NUM >= 2011
+#include <ufe/pathString.h>
+#endif
 #else
 #include "UfeVersionCompat.h"
 #endif
@@ -119,6 +122,11 @@ MStatus initialize()
 		return MS::kFailure;
 
 	g_StagesSubject = StagesSubject::create();
+
+    // Register for UFE string to path service using path component separator '/'
+#if UFE_PREVIEW_VERSION_NUM >= 2011
+    UFE_V2(Ufe::PathString::registerPathComponentSeparator(g_USDRtid, '/');)
+#endif
 
 	return MS::kSuccess;
 }
--- a/test/lib/ufe/testDeleteCmd.py
+++ b/test/lib/ufe/testDeleteCmd.py
@@ -23,6 +23,11 @@ import ufe
 
 import unittest
 
+import os
+
+def childrenNames(children):
+    return [str(child.path().back()) for child in children]
+
 class DeleteCmdTestCase(unittest.TestCase):
     '''Verify the Maya delete command, for multiple runtimes.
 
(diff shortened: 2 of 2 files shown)


lib/ufe/Global.cpp
```diff
@@ -120,6 +123,11 @@ MStatus initialize()
 
 	g_StagesSubject = StagesSubject::create();
 
+    // Register for UFE string to path service using path component separator '/'
+#if UFE_PREVIEW_VERSION_NUM >= 2011
+    UFE_V2(Ufe::PathString::registerPathComponentSeparator(g_USDRtid, '/');)
```
ppt-adsk: Register with optional UFE path string functionality.  In USD, the path component separator is forward slash '/'.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -138,3 +140,100 @@ def childrenNames(children):
 
         self.assertNotIn(sphereShapeName, sphereChildrenNames)
         self.assertNotIn(ball35Name, propsChildrenNames)
+
+        # undo to restore state to original.
+        cmds.undo()
+
+    @unittest.skipIf(os.getenv('UFE_PREVIEW_VERSION_NUM', '0000') < '2011', 'testVisibility only available in UFE preview 2011 and greater')
```
ppt-adsk: Ack!  Copy-paste damage.  'testVisibility' should read 'testDeleteArgs'.  Will fix along with any other code review comments.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -138,3 +140,100 @@ def childrenNames(children):
...
+    def testDeleteArgs(self):
+        '''Delete Maya and USD objects passed as command arguments.'''
```
ppt-adsk: Pretty much the same test as delete using selection, only in this case the UFE arguments are passed as UFE path strings.

ppt-adsk:
Hi Pilar, if you have any questions about this pull requests, motivations behind the work, where it will be used, don't hesitate to ask.

ppt-adsk:
Thanks Pilar for the review.

robthebloke:
The only slight issue I have is that with the changing paths (i.e. the use of ','), that it would require me to go and change all the paths in unit tests for PR 371. It also will affect a fair number of tools within our pipeline that have already adopted the new ufe paths. Is this what you consider to be the final version, or is this a WIP? I don't mind taking the hit once, but don't want to be taking this hit frequently....

ppt-adsk:
@robthebloke UFE path strings are a new service, and are optional.  If you're directly filling in UFE paths segment by segment using your own strings, that is unaffected, and will remain valid, so nothing to do on your part.

ppt-adsk:
It's really to support Maya command command-line arguments, which must be strings, both in Python and MEL, so that is future work you cannot possibly be using, because it's unsupported.

PR #369, opened 2020-03-20
clean up part1

HamedSabri-adsk:
I was planning to bring over this PR a bit later but since there has a been a lot of confusion around this I decided to do it now to clarify things.

### What is the point of this PR?

targets linking against `mayaUsd ` library don't need to do anything extra other than declaring their dependence on mayaUsd. mayaUsd target now declares ${CMAKE_BINARY_DIR}/include publicly.

This change is based on earlier concern from @mattyjams and it makes sense to me. I hope it makes things more clear.

e.g
before:
```
target_include_directories(${potato}   
  PRIVATE
      ${CMAKE_BINARY_DIR}/include

target_link_libraries(${potato}   
  PRIVATE
      mayaUsd
```

after:
```
target_link_libraries(${potato}   
  PRIVATE
      mayaUsd
```


HamedSabri-adsk:
Thanks both @mattyjams @elrond79 . Glad it works.

PR #371, opened 2020-03-20
[al] Fix unit tests when UFE is enabled

robthebloke:
Apologies for the size of this PR, there was quite a lot that needed fixing up! This PR extends from the MayaTransformAPI by fixing up the unit test failures when VP2_RENDER_DELEGATE_PROXY and MAYA_WANT_UFE_SELECTION are turned on. 

The full list of changes:

* Added MayaUsdProxyShapeBase::notifyPreSelectionChanged / MayaUsdProxyShapeBase::notifyPostSelectionChanged, used to notify the AL plugin that the UFE selection has just changed. 
* Added MayaUsd::ufe::refreshStages(). It seems there is a lag between the UsdStage being loaded, and the notification sent to Ufe. This caused problems running the selection unit tests, because the process of createProxy -> loadStage -> tryToSelect would always fail (the g_stagesMap would be initialised after the stage). 
* Added a UsdTimeCode parameter to the constructor and create methods of the Ufe Rotate/Scale/Translate commands. The time value is pulled from the proxy shape base via: MayaUsdProxyShapeBase::getTime()
* Added support for Scale Pivot editing in UFE.
* Had to wire up access to the proxy shape (to get the time value) through the UsdTransform3d class.
* I had an issue with lib/ufe/Utils.cpp where the selection of instances would fail. To ensure something was selected, I stripped the ".proto_blah" from the end of the SdfPaths, which at least allowed instances to be selected. There is no selection highlighting at this time (I'm not familiar enough with the VP2 Render Delegate code). 
* Added a -ls flag to AL_usdmaya_ProxyShapeSelect. This is to allow us to try and isolate the changes between the previous transform paths (e.g. "|proxy|root|somePrim") and the new UFE paths (e.g. "|world|proxy|proxyShape/root/somePrim"). 
* Modified the AL::usdmaya::nodes::SelectionHelper class to be able to work with UFE (used primarily by the AL_usdmaya_ProxyShapeSelect command)
* Added WANT_UFE_BUILD to the Cmake files for the AL plugins (previously it was defined in the mayaUsd and AL_USDMaya libs, but not defined by the plugin builds, which caused UFE to never be initialised.
* Modified many of the selection unit tests to get them working with UFE based selection.


kxl-adsk:
@robthebloke should we close #299? This PR seems to be more complete. 
As well, PR #357 fixes small part of what you have it here (using utilities we already have)

lib/usd/utils/CMakeLists.txt
```diff
@@ -75,3 +83,74 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
...
+    ${UTILS_INCLUDE_LOCATION} 
+    ${MAYAUSDUTILS_INCLUDE_LOCATION} 
+    ${USD_INCLUDE_DIR}
+    )
+
+target_link_libraries(${MAYAUSDUTILS_PYTHON_LIBRARY_NAME}
+  ${TARGET_NAME}
+  ${python_LIBRARIES}
+  usdGeom
+)
```
HamedSabri-adsk: No keyword? This is a recipe for disaster.  `build `and `usage `requirement should be clear.

Please explicitly specify PRIVATE, PUBLIC, INTERFACE here.

lib/usd/utils/CMakeLists.txt
```diff
@@ -75,3 +83,74 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
...
+    MFB_ALT_PACKAGE_NAME=${TARGET_NAME}
+    MFB_PACKAGE_MODULE=${TARGET_NAME}
+)
+
+target_include_directories(${MAYAUSDUTILS_PYTHON_LIBRARY_NAME} 
+    PUBLIC
+    ${UTILS_INCLUDE_LOCATION} 
+    ${MAYAUSDUTILS_INCLUDE_LOCATION} 
+    ${USD_INCLUDE_DIR}
+    )
```
HamedSabri-adsk: closing parentheses seems off here.

lib/usd/utils/CMakeLists.txt
```diff
@@ -75,3 +83,74 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
...
+foreach(INPUT_FILE ${PY_INIT_FILES})
+  string(REPLACE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR}/${TARGET_NAME} OUTPUT_FILE ${INPUT_FILE})
+  execute_process(COMMAND ${CMAKE_COMMAND} -E copy ${INPUT_FILE} ${OUTPUT_FILE})
+endforeach()
+
+
+list(APPEND mayausdutils_python_source
+  module.cpp
+  wrapMayaTransformAPI.cpp
+)
```
HamedSabri-adsk: Why list? 

use set:
```
set(mayausdutils_python_source
  module.cpp
  wrapMayaTransformAPI.cpp
)
```

lib/usd/utils/CMakeLists.txt
```diff
@@ -75,3 +83,74 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
             DESTINATION ${CMAKE_INSTALL_PREFIX}/lib OPTIONAL
     )
 endif()
+
+# -----------------------------------------------------------------------------
+# python
+# -----------------------------------------------------------------------------
+
```
HamedSabri-adsk: empty space

lib/usd/utils/CMakeLists.txt
```diff
@@ -75,3 +83,74 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
...
+add_subdirectory(tests)
+
+#install python init files
+foreach(INPUT_FILE ${PY_INIT_FILES})
+  string(REPLACE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_INSTALL_PREFIX}/lib/python/${TARGET_NAME} OUTPUT_FILE ${INPUT_FILE})
+  install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E copy ${INPUT_FILE} ${OUTPUT_FILE})")
+  install(CODE "execute_process(COMMAND python -m compileall ${OUTPUT_FILE} ${CMAKE_INSTALL_PREFIX})")
+endforeach()
+
+
```
HamedSabri-adsk: two spaces.

lib/usd/utils/tests/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+  COMMAND
+    python -m unittest discover -s ${CMAKE_CURRENT_SOURCE_DIR}
+)
+
+set_tests_properties(${USDUTILS_PYTHON_TEST_NAME}
+  PROPERTIES
+    ENVIRONMENT PYTHONPATH=${CMAKE_BINARY_DIR}/${INSTALL_DIR_SUFFIX}/lib/usd/utils:${CMAKE_BINARY_DIR}/src:$ENV{PYTHONPATH} 
+)
+
+
```
HamedSabri-adsk: Please adopt the new way of adding unit tests using `mayaUsd_add_test`. There are ample examples around on how to use it for both python and gtest unit tests.

See:  https://github.com/Autodesk/maya-usd/blob/dev/cmake/test.cmake

lib/usd/utils/tests/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+    sdf
+    tf
+    gf
+    usd
+    usdGeom
+    vt
+    mayaUsdUtils
+    -lpthread -lm
+)
+
```
HamedSabri-adsk: I don't have access to my Linux machine for a while to test this but do we need to link against all of these libraries?

Instead of ${GTEST_LIBRARIES}, use GTest::GTest.  By doing this, you don't need to have ${GTEST_INCLUDE_DIRS} in the 
target_include_directories

lib/usd/utils/tests/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+list(APPEND usdutils_test_source
+    main.cpp
+    test_MayaTransformAPI.cpp
+)
+
+add_executable(${USDUTILS_TEST_EXECUTABLE_NAME}
+  ${usdutils_test_source}
+)
+
```
HamedSabri-adsk: Please use `target_sources` and explicitly declare the keyword entries.

lib/usd/utils/tests/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
+find_package(GTest REQUIRED)
+
+set(USDUTILS_TEST_EXECUTABLE_NAME AL_USDUtilsTests)
+set(USDUTILS_TEST_NAME GTest:AL_USDUtilsTests)
+set(USDUTILS_PYTHON_TEST_NAME Python:AL_USDUtilsTests)
+
+add_definitions(
+  -DAL_EXTRAS_TEST_DATA="${CMAKE_CURRENT_SOURCE_DIR}/data"
+)
+
```
HamedSabri-adsk: Use of add_definitions is prohibited.   use `target_compile_definitions ` instead and explicitly declare the keyword entries.

lib/usd/utils/tests/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
+find_package(GTest REQUIRED)
```
HamedSabri-adsk: There is no need to call this here. Project's top level cmake handles this.

lib/usd/utils/tests/CMakeLists.txt
HamedSabri-adsk: We don't bury unit tests inside subdirectories inside `maya-usd/lib`. All relevant unit tests should go into `maya-usd/test`.

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/commands/test_ProxyShapeSelect.cpp
```diff
@@ -57,28 +104,52 @@ TEST(ProxyShapeSelect, selectNode1)
...
+    else
+    {
+      auto selection = Ufe::GlobalSelection::get();
+      for(auto it : *selection)
+      {
+        bool found = false;
+        for(uint32_t j = 0; j < paths.size(); ++j)
+        {
+          auto str = it->path().string();
+          auto index = str.find_first_of('/');
```
ppt-adsk: Looks like you're trying to get the second segment in the UFE path, i.e.
it->path().getSegments()[1]

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/commands/test_ProxyShapeSelect.cpp
```diff
@@ -286,28 +373,52 @@ TEST(ProxyShapeSelect, selectNode2)
...
```
ppt-adsk: Again, looks like UFE second segment access through string manipulation.

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/commands/test_ProxyShapeSelect.cpp
```diff
@@ -964,34 +1153,61 @@ TEST(ProxyShapeSelect, selectSamePathTwice)
...
-  EXPECT_TRUE(proxy->isRequiredPath(SdfPath("/root")));
-  EXPECT_TRUE(proxy->isRequiredPath(SdfPath("/root/hip1")));
-  EXPECT_TRUE(proxy->isRequiredPath(SdfPath("/root/hip1/knee1")));
-  EXPECT_TRUE(proxy->isRequiredPath(SdfPath("/root/hip1/knee1/ankle1")));
-  EXPECT_TRUE(proxy->isRequiredPath(SdfPath("/root/hip1/knee1/ankle1/ltoe1")));
+  if(!MayaUsdProxyShapePlugin::useVP2_NativeUSD_Rendering())
+  {
+    MSelectionList sl;
+    MGlobal::getActiveSelectionList(sl);
+    EXPECT_EQ(1u, sl.length());
```
ppt-adsk: By virtue of the UFE selection list being a superset of the Maya selection, can't you just use the UFE selection throughout?  How can it be that the UFE selection would be inconsistent with the Maya selection?  If there's a way that this can happen, we need to plug that hole really, really quickly.

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/commands/test_ProxyShapeSelect.cpp
```diff
@@ -1359,188 +1643,218 @@ TEST(ProxyShapeSelect, pseudoRootSelect)
...
-  MGlobal::executeCommand("undo", false, true);
-
-  // State 1: proxy selected
-  EXPECT_EQ(0u, proxy->selectedPaths().size());
-  { SCOPED_TRACE(""); compareNodes({SdfPath("/")}); };
-
-  // Undo the initial replace
-  MGlobal::executeCommand("undo", false, true);
+  //
+  // RB: Disabling this test for now as this is currently broken in UFE pending some feedback.
```
ppt-adsk: If all else fails, you can always select Maya nodes using the Maya API, or the select command, which will automatically update the UFE selection.

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/nodes/test_TransformMatrix.cpp
```diff
@@ -1388,75 +1389,82 @@ TEST(Transform, emptyOpsNotMade)
...
     }
-
+    else
     {
-      SCOPED_TRACE("Querying after dirtying node");
-      MString cmd = MString("dgdirty \"") + xformMfn.partialPathName() + "\";";
-      status = MGlobal::executeCommand(cmd);
-      EXPECT_EQ(MS::kSuccess, status);
-      assertEmptyAfterQuerying();
+      /// \todo fix up test for UFE based selection. (Probably not needed?)
```
ppt-adsk: How does the test need to be fixed up?

lib/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -142,7 +142,7 @@ class ProxyRenderDelegate : public MHWRender::MPxSubSceneOverride
     void _FilterSelection();
     void _UpdateSelectionStates();
 
-    const MayaUsdProxyShapeBase*  _proxyShape{ nullptr }; //!< DG proxy shape node
```
ppt-adsk: Loss of constness seems like a shame.

ppt-adsk:
Great work, looking forward to iterating on this!  You seem to have found some pain points, hopefully we can explore these and resolve them.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.h
```diff
@@ -601,12 +606,27 @@ class ProxyShape
...
   MString getMayaPathFromUsdPrim(const UsdPrim& usdPrim) const;
 
+  /// \brief  A method that attempts to abstract away the setting of the pushToPrim attribute on a transform node. 
+  ///         Once support for parenting maya nodes to UFE objects has been added, refactor this method to provide
+  ///         similar read only support. 
+  /// \param  usdPath path to the prim you want to enable/disable pushToPrim
+  /// \param  pushToPrimState enable/disable pushToPrim
+  /// \return true if the node is valid
+  AL_USDMAYA_PUBLIC
+  bool enablePushToPrim(const SdfPath& usdPath, bool pushToPrimState) const;
```
kxl-adsk: This doesn't look to be related to the change you are proposing.

lib/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -545,14 +547,25 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
...
+        si = handler->createItem(_proxyShape->ufePath() + pathSegment);
+        if(!si)
+        {
+            TF_WARN("UFE runtime is not updated for the USD stage. Please save scene and reopen.");
+            return false;
+        }
     }
-
+    if(_proxyShape)
+        _proxyShape->notifyPreSelectionChanged();
```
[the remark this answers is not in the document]
kxl-adsk: Agreed, this doesn't seem like a change we want to make.

lib/ufe/UsdTransform3d.cpp
```diff
@@ -75,6 +80,32 @@ void UsdTransform3d::setItem(const UsdSceneItem::Ptr& item)
...
+	MString dagPathStr(pathStr.c_str() + 6, pos - 6);
+	MSelectionList sl;
+	sl.add(dagPathStr);
+	MDagPath dagPath;
+	sl.getDagPath(0, dagPath);
+	MFnDagNode fn(dagPath);
+	return (MayaUsdProxyShapeBase*)fn.userNode();
+}
+
+UsdTimeCode UsdTransform3d::timeCode() const
```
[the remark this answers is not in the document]
kxl-adsk: This was already done as part of #357.

lib/render/vp2RenderDelegate/proxyRenderDelegate.h
kxl-adsk: Indeed...but we will need a better solution to these new callbacks, so we will get back constness here.

lib/ufe/UsdTransform3d.cpp
```diff
@@ -91,78 +122,65 @@ Ufe::SceneItem::Ptr UsdTransform3d::sceneItem() const
...
-	return Ufe::Vector3d(x, y, z);
+	MayaUsdUtils::MayaTransformAPI api(fPrim);
+	auto translate = api.translate(timeCode());
+	return Ufe::Vector3d(translate[0], translate[1], translate[2]);
 }
 
 Ufe::RotateUndoableCommand::Ptr UsdTransform3d::rotateCmd()
 {
-	auto rotateCmd = UsdRotateUndoableCommand::create(fPrim, fItem->path(), fItem);
+	auto rotateCmd = UsdRotateUndoableCommand::create(fPrim, fItem->path(), fItem, timeCode());
```
kxl-adsk: I see at least three separate changes/PRs in here. 
1) the addition of new TransformAPI (this is already a separate PR)
2) enabling usage of new TransformAPI (can include new arguments to pass time, but shouldn't modify the behavior yet)
3) switching to set data at a `current time`

Points 2 and 3 require separate discussions.

PR #379, opened 2020-03-24
hdMaya: Adopt the include directive/order according to coding guideline.

HamedSabri-adsk:
This PR changes the `include directive/order ` in hdMaya according to guideline.

### DIFF
--- a/lib/usd/hdMaya/adapters/adapter.cpp
+++ b/lib/usd/hdMaya/adapters/adapter.cpp
@@ -13,15 +13,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-#include "adapter.h"
-#include "adapterDebugCodes.h"
+#include <hdMaya/adapters/adapter.h>
 
 #include <pxr/base/tf/type.h>
 
 #include <maya/MNodeMessage.h>
 
-#include "materialNetworkConverter.h"
-#include "mayaAttrs.h"
+#include <hdMaya/adapters/adapterDebugCodes.h>
+#include <hdMaya/adapters/materialNetworkConverter.h>
+#include <hdMaya/adapters/mayaAttrs.h>
 
 PXR_NAMESPACE_OPEN_SCOPE
 
--- a/lib/usd/hdMaya/adapters/adapter.h
+++ b/lib/usd/hdMaya/adapters/adapter.h
@@ -17,15 +17,14 @@
 #define HDMAYA_ADAPTER_H
 
 #include <pxr/pxr.h>
-
 #include <pxr/usd/sdf/path.h>
 
 #include <maya/MMessage.h>
 
 #include <vector>
 
-#include "../api.h"
-#include "../delegates/delegateCtx.h"
+#include <hdMaya/api.h>
+#include <hdMaya/delegates/delegateCtx.h>
 
 PXR_NAMESPACE_OPEN_SCOPE
 
--- a/lib/usd/hdMaya/adapters/lightAdapter.h
+++ b/lib/usd/hdMaya/adapters/lightAdapter.h
@@ -21,13 +21,13 @@
 #include <pxr/base/gf/frustum.h>
 
 #include <pxr/imaging/glf/simpleLight.h>
+#include <pxr/imaging/hd/light.h>
 #include <pxr/imaging/hdx/simpleLightTask.h>
 
 #include <maya/MFnLight.h>
 #include <maya/MFnNonExtendedLight.h>
 
 #include <hdMaya/adapters/dagAdapter.h>
-#include <pxr/imaging/hd/light.h>
 
 PXR_NAMESPACE_OPEN_SCOPE
 
--- a/lib/usd/hdMaya/adapters/lightAdapter.h
+++ b/lib/usd/hdMaya/adapters/lightAdapter.h
@@ -26,7 +26,7 @@
 #include <maya/MFnLight.h>
 #include <maya/MFnNonExtendedLight.h>
 
-#include "dagAdapter.h"
+#include <hdMaya/adapters/dagAdapter.h>
 #include <pxr/imaging/hd/light.h>
 
 PXR_NAMESPACE_OPEN_SCOPE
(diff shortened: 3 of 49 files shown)


lib/usd/hdMaya/adapters/lightAdapter.h
```diff
@@ -26,7 +26,7 @@
 #include <maya/MFnLight.h>
 #include <maya/MFnNonExtendedLight.h>
 
-#include "dagAdapter.h"
+#include <hdMaya/adapters/dagAdapter.h>
```
mattyjams: I think this one would go in its own group *below* `<pxr/imaging/hd/light.h>`?

lib/usd/hdMaya/adapters/materialNetworkConverter.cpp
```diff
@@ -31,7 +24,14 @@
 #include <maya/MPlugArray.h>
 #include <maya/MStatus.h>
 
-#include "../utils.h"
+#include <hdMaya/adapters/adapterDebugCodes.h>
+#include <hdMaya/adapters/materialAdapter.h>
+#include <hdMaya/adapters/mayaAttrs.h>
+#include <hdMaya/adapters/tokens.h>
+
```
mattyjams: Remove this blank line to put all the `hdMaya` includes in a group together?

lib/usd/hdMaya/adapters/proxyAdapter.h
```diff
@@ -1,15 +1,15 @@
 #ifndef HDMAYA_AL_PROXY_ADAPTER_H
 #define HDMAYA_AL_PROXY_ADAPTER_H
 
-#include "../delegates/proxyUsdImagingDelegate.h"
-#include "shapeAdapter.h"
+#include <hdMaya/delegates/proxyUsdImagingDelegate.h>
+#include <hdMaya/adapters/shapeAdapter.h>
```
mattyjams: Swap these two so "adapters" is before "delegates"?

lib/usd/hdMaya/delegates/sceneDelegate.cpp
```diff
@@ -28,24 +28,24 @@
...
 
+#include <pxr/imaging/hd/light.h>
 #include <pxr/imaging/hdx/renderSetupTask.h>
 #include <pxr/imaging/hdx/renderTask.h>
 #include <pxr/imaging/hdx/tokens.h>
 
-#include <maya/MDGMessage.h>
 #include <maya/MDagPath.h>
 #include <maya/MDagPathArray.h>
+#include <maya/MDGMessage.h>
```
mattyjams: I always waffle on this, whether to order capital letters before lower case, or to sort case-insensitively. :/

lib/usd/hdMaya/adapters/lightAdapter.h
HamedSabri-adsk: Thanks you @mattyjams . Please see 7cb7810a

lib/usd/hdMaya/adapters/materialNetworkConverter.cpp
HamedSabri-adsk: see 7cb7810a

lib/usd/hdMaya/adapters/proxyAdapter.h
HamedSabri-adsk: see 7cb7810a

lib/usd/hdMaya/delegates/sceneDelegate.cpp
HamedSabri-adsk: I hear you. I honestly don't have a strong preference. I can change it if you guys prefer it the other way any-times.

mattyjams: Yeah, I don't have a strong preference either. Just maybe a justification as to why it may have been that way. :)

lib/usd/hdMaya/adapters/adapter.cpp
```diff
@@ -13,15 +13,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-#include "adapter.h"
-#include "adapterDebugCodes.h"
+#include <hdMaya/adapters/adapter.h>
 
 #include <pxr/base/tf/type.h>
 
 #include <maya/MNodeMessage.h>
```
elrond79: According to the style guidelines, Autodesk + Maya comes before pxr + usd headers.  So these two would be flipped.

Of course, I think we're eventually planning on automating this via clang-format, so if you want to leave this be until we formalize our clang-format file, that's fine - we just need to make sure absolute/relative + <>/"" is handled (since clang-format won't change that.)

There's a few other places where maya vs pxr order is flipped, so I won't bother commenting on all of them.

lib/usd/hdMaya/adapters/adapter.h
```diff
@@ -17,15 +17,14 @@
 #define HDMAYA_ADAPTER_H
 
 #include <pxr/pxr.h>
-
 #include <pxr/usd/sdf/path.h>
 
 #include <maya/MMessage.h>
 
 #include <vector>
```
elrond79: vector, as a c++ standard lib, should come first.

lib/usd/hdMaya/delegates/sceneDelegate.cpp
elrond79: I want to go with whatever clang-format does by default for this, since I'd eventually like to start using that to handle header ordering automatically.  I wonder if maybe we should just create an initial `.clang-format` file, which disables everything EXCEPT the header ordering, and then just run that on the whole repo?  That way we can at least start having that be standardized, before we come to agreement on all the other nitty-gritty style details.

HamedSabri-adsk:
@mattyjams  Would it be Ok if I close this PR in favor of (#392, #388, #389, #390, #391)? Paul confirmed that he is OK.

mattyjams:
@HamedSabri-adsk: Sure, works for me. I'll try to make some time to look at those other PRs today. Thanks!

PR #381, opened 2020-03-24
MAYA-103421 - UFE selecting certain USD objects will throw errors

seando-adsk:
Modified all the attribute Get calls to use current time.
If attribute has no time samples, the Get will automatically fallback to default time.

MAYA-103421 - UFE selecting certain USD objects will throw errors on the outliner and script editor
MAYA-103266 - USD Materials Erroring with USD attribute does not match created attribute class type

### DIFF
--- a/lib/ufe/UsdAttribute.cpp
+++ b/lib/ufe/UsdAttribute.cpp
@@ -391,7 +391,7 @@ T TypedUsdAttribute<T>::get() const
 	if (!hasValue()) return T();
 
 	PXR_NS::VtValue vt;
-	if (fUsdAttr.Get(&vt, getCurrentTime(sceneItem())) && vt.IsHolding<T>())
+	if (fUsdAttr.Get(&vt, getCurrentTime(Ufe::Attribute::sceneItem())) && vt.IsHolding<T>())
 	{
 		return vt.UncheckedGet<T>();
 	}
@@ -404,7 +404,7 @@ template<typename T>
 void TypedUsdAttribute<T>::set(const T& value)
 {
 	T dummy;
-	UFE_ASSERT_MSG(fUsdAttr.Get<T>(&dummy, getCurrentTime(sceneItem())), kErrorMsgInvalidType);
+	UFE_ASSERT_MSG(fUsdAttr.Get<T>(&dummy, getCurrentTime(Ufe::Attribute::sceneItem())), kErrorMsgInvalidType);
 	bool b = setUsdAttr<T>(fUsdAttr, value);
 	UFE_ASSERT_MSG(b, kErrorMsgFailedSet);
 }
--- a/lib/ufe/UsdAttribute.cpp
+++ b/lib/ufe/UsdAttribute.cpp
@@ -16,6 +16,7 @@
 
 #include "UsdAttribute.h"
 #include "StagesSubject.h"
+#include "Utils.h"
 
 #include <pxr/base/tf/token.h>
 #include <pxr/base/vt/value.h>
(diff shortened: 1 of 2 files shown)


lib/ufe/UsdAttribute.cpp
```diff
@@ -79,12 +80,20 @@ bool setUsdAttr(const PXR_NS::UsdAttribute& attr, const T& value)
...
 }
 
-std::string getUsdAttributeValueAsString(const PXR_NS::UsdAttribute& attr)
+PXR_NS::UsdTimeCode getCurrentTime(const Ufe::SceneItem::Ptr& item)
+{
+	// Attributes with time samples will fail when calling Get with default time code.
+	// So we'll always use the current time when calling Get. If there are no time
+	// samples, it will fall-back to the default time code.
+	return MayaUsd::ufe::getTime(item->path());
+}
```
seando-adsk: New helper function that I'll call from every Get call below.

seando-adsk:
@robthebloke Rob, would you have time to review this?

lib/ufe/UsdAttribute.cpp
kxl-adsk: Wouldn't this helper function be useful to have in Utils.h?

lib/ufe/UsdAttribute.h
```diff
@@ -43,9 +43,9 @@ class UsdAttribute
 	bool hasValue() const;
 	std::string name() const;
 	std::string documentation() const;
-	std::string string() const;
+	std::string string(const Ufe::SceneItem::Ptr& item) const;
 
-public:
+protected:
```
kxl-adsk: Why not private?

lib/ufe/UsdAttribute.cpp
seando-adsk: There is already a getTime helper in Utils.h that takes a Ufe::Path. I created this one here mostly so I can have a single place that gets the time with the comment about why.

lib/ufe/UsdAttribute.h
seando-adsk: These members are needed by the derived classes.

kxl-adsk: I looked at derived classes and now I see why it's done this way. Thx.

lib/ufe/UsdAttribute.cpp
kxl-adsk: I was coming from the angle that maybe someone else will need it as well. Anyhow, not a blocker, we can deal with it next time we will see a similar method.

lib/ufe/UsdAttribute.cpp
```diff
@@ -79,12 +80,20 @@ bool setUsdAttr(const PXR_NS::UsdAttribute& attr, const T& value)
 	return attr.Set<T>(value);
 }
 
-std::string getUsdAttributeValueAsString(const PXR_NS::UsdAttribute& attr)
+PXR_NS::UsdTimeCode getCurrentTime(const Ufe::SceneItem::Ptr& item)
+{
+	// Attributes with time samples will fail when calling Get with default time code.
+	// So we'll always use the current time when calling Get. If there are no time
+	// samples, it will fall-back to the default time code.
+	return MayaUsd::ufe::getTime(item->path());
```
robthebloke: The way I approached this was to query the time via a virtual call to MayaUsdProxyShapeBase. The reason is that in the AL proxy shape, we compute the time value via the following equation:
```
outTime = timeOffset + timeScalar * time;
```
This was primarily to allow artists a bit more control over cache playback.

robthebloke: FYI: these are the attributes that define our time value:
https://github.com/Autodesk/maya-usd/blob/dev/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.h#L283

and this is where we compute the time: 
https://github.com/Autodesk/maya-usd/blob/dev/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp#L1476

kxl-adsk: @robthebloke please check the method we are having in utilities https://github.com/Autodesk/maya-usd/blob/dev/lib/ufe/Utils.cpp#L264. This method will call your version of getTime and time will be computed based on the equation you have mentioned.

PR #387, opened 2020-03-26
Cleanup "include directive/order" + "folder structures" 

HamedSabri-adsk:
Apologies in advanced for the large PR. This PR is intended to cleanup "include directive/order" under "lib" + "plugin/adsk" as well as "folder structures" under the lib subdirectory.

**include directive/order:**

All the *.h, *.cpp files under lib and plugin/adsk follow the coding guideline. 

**CMake changes:**

- All dir/subdir now get their own CMakeLists.txt
- ${CMAKE_CURRENT_SOURCE_DIR} is no longer added to the include path.
- All the projects now use the promotion header mechanism.
- mayaUsdPlugin promotion headers are now created under mayaUsdPlugin and not mayaUsd

**lib directory structure:**

There are now two subdirectories under lib:  
- **mayaUsd** : contain the core shared modules
- **usd** : contain shared libraries that depend only on USD and plug-ins to USD


elrond79:
At minimum, please separate out the header changes into their own commit.

HamedSabri-adsk:
Alright, I will try a different strategy... :)

PR #388, opened 2020-03-27
PART I: lib directory restructure

HamedSabri-adsk:
Ok, I have broken up the large PR into 5 parts.  Union of these 5 PRs is what makes the PR complete and working.  I also ran all the unit tests and they all pass at least on Windows.

lib directory is restructured and now contains two major sub-directories:
1. mayaUsd : contain the core shared modules
2. usd : contain shared libraries that depend only on USD and plug-ins to USD


HamedSabri-adsk:
> /// \file usdMaya/

Ahh, thanks for pointing out.  That's just me missing a few of them ( to be exact 10 ). :)  I already removed ~90 some of them. I will have a separate PR after all this to remove them.

HamedSabri-adsk:
Hi All,

I write here since this page doesn't have as many comments as the other 4 pages.

I've created 3 commits for PART III, IV, V that would address reviewer's feedback as well as adopting things to the latest code guideline. 

Project compiles and builds. All test successfully pass, all plugins load.  At least for me on MacOS and Windows. I don't have access to my Linux machine because of issues with VPN. 
I will also resolve the conflicts once the review is finished.

**PART III**
UPDATE: include directives and orders under **lib/usd** according to code guideline.
9d9404a6c11c3eb37ce41d0f60cbbc982b118078

**PART IV**
UPDATE: include directives and orders under **lib/mayaUsd** according to code guideline.
886119f576405e96acf84eb27b6419fc9fd993d4

**PART V**
UPDATE: include directives and orders under **plugin/adsk** according to code guideline.
3324c613c5d6217676dac3ee1b98ca82e5578c91

elrond79:
I started a review of your latest changes in Part III, and realized we had probably better get some clarification: https://github.com/Autodesk/maya-usd/pull/390#pullrequestreview-386875079

I had thought that we had decided that ALL public files were always using <> - and that the only exception was for the associated header of a .cpp file - ie, "foobar.h" for "foobar.cpp".

Did I have this wrong?

HamedSabri-adsk:
Ugh correct. Sorry about that, I totally missed it.

One other thing that I have noticed is that some of the projects ( e.g hdMaya , mayaUsdUI , etc ... )  promote and install header files that are totally necessary.

For example, why does hdMaya  need to promote and install api.h?  isn't api.h private.

Again, not something that I would change this PR series but we should make a separate PR to address them.

-H


________________________________

elrond79:
Well, I think that if any public headers use hdMaya/api.h, then transitively api.h would also have to be public.  Though I certainly wouldn't be surprised if there are other headers in there that don't really need to be public.  (For instance, I think mayaAttrs.h is right now, and it's questionable whether it should be.)

HamedSabri-adsk:
@elrond79  fixes for #390 (review) should be found here:

**PART III**
0acfcb2c426f8bb7baef7bd3a981dcd7b5b1cf82

**PART IV**
db250b344f9cea4764a7571cb2f782166362e2a3

PR #389, opened 2020-03-27
PART II: CMake clean ups under lib/mayaUsd and lib/usd

HamedSabri-adsk:
1. All sub directories under lib/mayaUsd and lib/usd get their own CMakeLists.txt
2. fix build/usage requirements for mayaUsdUtils and mayaUsd_Translators, testMayaUsdUI
3. don't rely on ${PROJECT_NAME} in lib/usd because of the folder restructure
4. use promote header mechanism across the project


lib/usd/translators/CMakeLists.txt
```diff
@@ -46,6 +46,41 @@ target_sources(${TARGET_NAME}
...
+    skelReader.h
+    xformReader.h
+)
+
+# -----------------------------------------------------------------------------
+# promoted headers
+# -----------------------------------------------------------------------------
+mayaUsd_promoteHeaderList(HEADERS ${headers}
+                          BASEDIR ${TARGET_NAME}
+)
```
mattyjams: Different style from the calls in other files?:
```
mayaUsd_promoteHeaderList( 
    HEADERS
        ${headers}
    BASEDIR
        ${TARGET_NAME}
)
```

lib/usd/ui/CMakeLists.txt
```diff
@@ -128,8 +102,8 @@ if(IS_WINDOWS)
     )
 endif()
 
-install(FILES ${mayaUsdUI_headers}
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/include/mayaUsd/ui
+install(FILES ${headers}
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/include/${PROJECT_NAME}/ui
```
mattyjams: `mayaUsd` vs. `${PROJECT_NAME}`?

lib/usd/ui/CMakeLists.txt
```diff
@@ -110,6 +71,19 @@ if(IS_MACOSX OR IS_LINUX)
...
+# -----------------------------------------------------------------------------
+set(headers
+    api.h
+)
+
+mayaUsd_promoteHeaderList( 
+    HEADERS
+        ${headers}
+    BASEDIR
+        ${PROJECT_NAME}/ui
```
mattyjams: Based on the commit comments and PR description, should this be `mayaUsd` instead of `${PROJECT_NAME}` since it's under the `lib/usd` directory?

lib/usd/ui/cmds/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+# -----------------------------------------------------------------------------
+# sources
+# -----------------------------------------------------------------------------
+target_sources(${PROJECT_NAME}
```
mattyjams: `mayaUsd` vs. `${PROJECT_NAME}` in this file?

lib/usd/ui/demo/CMakeLists.txt
```diff
@@ -28,19 +28,12 @@ target_sources(${PROJECT_NAME}
...
 # -----------------------------------------------------------------------------
-target_include_directories(${PROJECT_NAME}
-    PRIVATE
-        ${CMAKE_BINARY_DIR}/include
-)
 
 # -----------------------------------------------------------------------------
 # link libraries
 # -----------------------------------------------------------------------------
 target_link_libraries(${PROJECT_NAME}
```
mattyjams: `mayaUsd` vs. `${PROJECT_NAME}`?

lib/usd/ui/factories/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+# -----------------------------------------------------------------------------
+# sources
+# -----------------------------------------------------------------------------
+target_sources(${PROJECT_NAME}
```
mattyjams: `mayaUsd` vs. `${PROJECT_NAME}` in this file?

lib/usd/ui/views/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+# -----------------------------------------------------------------------------
+# sources
+# -----------------------------------------------------------------------------
+target_sources(${PROJECT_NAME}
```
mattyjams: `mayaUsd` vs. `${PROJECT_NAME}` in this file?

lib/usd/ui/CMakeLists.txt
HamedSabri-adsk: Hey @mattyjams  sorry for the misleading PR description around ${PROJECT_NAME}. I will fix the description but let me explain the problem:

The top-level cmake in super project `maya-usd/CMakeLists.txt` defines the project name
as `maya-usd`.  This is the root project. We then have bunch of mini projects underneath maya-usd project:

e,g
project(mayaUsd)
project(mayaUsdUI)
project(testMayaUsdUI)

`project` basically sets the name of the project, and stores it in the variable **PROJECT_NAME**.

Now `lib/usd/ui`  top-level cmake declares `project(mayaUsdUI)` and therefore ${PROJECT_NAME}  under this sub-directory returns `mayaUsdUI` which is a correct expected behavior.

Now, `lib/usd/schemas` is not really part of any projects and since we have restructured the sub-directories under lib, calling ${PROJECT_NAME} returns `maya-usd` from the targets that are not part of a mini projects ( e.g schemas, hdMaya, utils) . 
I have brought this up in our internal meetings and it is something to consider fixing ( something for sure for another day ). 

schema for whatever reasons needs to install it's stuff under `lib\python\mayaUsd`  and that's why ${PROJECT_NAME} would install things in  `lib\python\maya-usd` which is wrong.

I hope this makes sense.

lib/usd/ui/cmds/CMakeLists.txt
HamedSabri-adsk: https://github.com/Autodesk/maya-usd/pull/389#discussion_r399602894

lib/usd/ui/demo/CMakeLists.txt
HamedSabri-adsk: https://github.com/Autodesk/maya-usd/pull/389#discussion_r399602894

lib/usd/ui/factories/CMakeLists.txt
HamedSabri-adsk: https://github.com/Autodesk/maya-usd/pull/389#discussion_r399602894

lib/usd/ui/views/CMakeLists.txt
HamedSabri-adsk: https://github.com/Autodesk/maya-usd/pull/389#discussion_r399602894

lib/mayaUsd/render/vp2RenderDelegate/CMakeLists.txt
```diff
@@ -18,7 +18,22 @@ target_sources(${PROJECT_NAME}
 )
 
 set(headers
+    basisCurves.h
+    bboxGeom.h
+    CMakeLists.txt
```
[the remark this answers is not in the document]
HamedSabri-adsk: Ouch. that's advanced stuff. Lol  see 8cd1727fe28

lib/mayaUsd/fileio/utils/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+        readUtil.cpp
+        roundTripUtil.cpp
+        shadingUtil.cpp
+        userTaggedAttribute.cpp
+        writeUtil.cpp
+        xformStack.cpp
+)
+
+set(headers
+    shadingUtil.h
```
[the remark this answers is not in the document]
HamedSabri-adsk: see 8cd1727fe28

lib/mayaUsd/fileio/translators/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
+# -----------------------------------------------------------------------------
+# sources
+# -----------------------------------------------------------------------------
+target_sources(${PROJECT_NAME} 
+    PRIVATE
+        translatorRfMLight.cpp
```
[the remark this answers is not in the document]
HamedSabri-adsk: see 8cd1727fe28

lib/mayaUsd/fileio/shading/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+# -----------------------------------------------------------------------------
+# sources
+# -----------------------------------------------------------------------------
+target_sources(${PROJECT_NAME} 
+    PRIVATE
+        shadingModePxrRis.cpp
```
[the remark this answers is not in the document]
HamedSabri-adsk: see 8cd1727fe28

lib/mayaUsd/ufe/private/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+# promoted headers
+# -----------------------------------------------------------------------------
+mayaUsd_promoteHeaderList(HEADERS ${headers} SUBDIR ufe/private)
+
+# -----------------------------------------------------------------------------
+# install
+# -----------------------------------------------------------------------------
+install(FILES ${headers}
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/include/${PROJECT_NAME}/ufe/private
+)
```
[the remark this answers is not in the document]
HamedSabri-adsk: Ah, that's my mistake.  ufe/private is not supposed to be installed.  However, we still want to have these headers promoted so we can use it them to get the full path working.
e.g
#include <mayaUsd/ufe/private/InPathChange.h>

See 28d48e0

HamedSabri-adsk:
@mattyjams CMake can generate graphviz files that could show the dependencies between the targets in a project, libraries which are linked against, etc..  

I use it from time to time for sake of visualization. Here is the graphviz Only for `mayaUsd`.   You need to zoom in :) 

mayaUsd_graph_visualization

lib/mayaUsd/render/px_vp20/CMakeLists.txt
```diff
@@ -23,5 +23,5 @@ mayaUsd_promoteHeaderList(HEADERS ${headers} SUBDIR render/px_vp20)
 # install
 # -----------------------------------------------------------------------------
 install(FILES ${headers}
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/include/mayaUsd/render/px_vp20
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/include/${PROJECT_NAME}/render/px_vp20
```
seando-adsk: We have a lot of these, which are kind of hard-coded paths. I wonder if there is a cmake variable or way to use the relative path from project_name? Such that we could replace "${PROJECT_NAME}/render/px_vp20" with a variable.

lib/usd/ui/CMakeLists.txt
```diff
@@ -27,35 +27,9 @@ add_library(${PROJECT_NAME} SHARED)
...
-        cmds/USDImportDialogCmd.cpp
-        factories/TreeModelFactory.cpp
         images/ui.qrc
-        views/ItemDelegate.cpp
-        views/TreeItem.cpp
-        views/TreeModel.cpp
-        views/USDImportDialog.cpp
-        views/USDImportDialog.ui
-        views/USDQtUtil.cpp
 )
```
seando-adsk: I understand the principle here, but I think we should weigh the cost vs benefit of having a single cmake file vs many in sub-dirs for such a small project like this UI one. My personal opinion is that it was much simpler and easier to understand and edit as a single file. For example both the cmds/factories sub-dirs each only have a single. That seems such an overhead to have a separate cmakefile for each.

HamedSabri-adsk: Hey @seando-adsk ,

I can see your point when a project is small (e.g mayaUsdUI) with not many source files. I agree that it seems like an overhead, however the advantage of having CMakeLists.txt for each subdirectory starts to show as the project grows. We also follow the similar discipline inside Maya code base as well.

seando-adsk: Sure I can see that, but in this particular case this project will never grow that big. I don't see advantage to having cmake file that has a single source file, such as: lib/usd/ui/factories/CMakeLists.txt

lib/mayaUsd/render/vp2RenderDelegate/CMakeLists.txt
```diff
@@ -18,7 +18,21 @@ target_sources(${PROJECT_NAME}
...
+    material.h
+    mesh.h
     proxyRenderDelegate.h
+    render_delegate.h
+    render_param.h
+    render_pass.h
+    resource_registry.h
+    sampler.h
+    task_commit.h
+    tokens.h
```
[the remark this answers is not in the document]
HamedSabri-adsk: Thanks @mattyjams . Great catch. I will fix it.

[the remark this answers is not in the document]
HamedSabri-adsk: @mattyjams please see de407da1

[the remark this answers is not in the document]
elrond79: It seems odd to have headers that are promoted, but not installed.  That would make them in a bit of a gray area between public and private.  Are we sure they need to promoted?

I took a random look at one header - `draw_item.h` - and it seems like it should be fine being not promoted - it's always brought in via `#include "draw_item.h"`, and only from .cpp files in the same directory.

Have we tried not promoting any of these, other than `proxyRenderDelegate.h`?

lib/mayaUsd/render/vp2RenderDelegate/CMakeLists.txt
```diff
@@ -18,7 +18,21 @@ target_sources(${PROJECT_NAME}
...
 
 set(headers
+    basisCurves.h
+    bboxGeom.h
+    debugCodes.h
+    draw_item.h
+    instancer.h
+    material.h
+    mesh.h
     proxyRenderDelegate.h
```
elrond79, line 28: So, following the conversation with @mattyjams [here](https://github.com/Autodesk/maya-usd/pull/389/files#r400421862), it really seems like the only entry in this should be proxyRenderDelegate.h, and the rest should be removed.  I suppose I can't say this definitely, since I only checked the first 4 items, but all of those are only used locally, and with relative private-header syntax, and so don't need to be promoted.  I strongly suspect all the others in this list are the same

HamedSabri-adsk: @elrond79 sorry but you are looking at an outdated file here. 

As I mentioned in here https://github.com/Autodesk/maya-usd/pull/388#issuecomment-607543513 I created 3 commits to address the issues and feedbacks.

I also last night sent you a link to the changes you requested regarding <> vs "" for public headers. See https://github.com/Autodesk/maya-usd/pull/388#issuecomment-608220588

kxl-adsk: I'm starting to get concerned that we are trying to fix much more than these 5 PRs originally intended. Let's respect the work Hamed already put into this and help him bring it to the end.

HamedSabri-adsk: > 
> 
> @HamedSabri-adsk - I checked the latest from this PR, [de407da](https://github.com/Autodesk/maya-usd/commit/de407da17da5087b9b50068598cb8e0240e1a354). And it's still promoting a bunch of headers. I also checked the latest from [part 5](https://github.com/Autodesk/maya-usd/pull/392) - [3324c61](https://github.com/Autodesk/maya-usd/commit/3324c613c5d6217676dac3ee1b98ca82e5578c91) - and the changes are still there.
> 

@elrond79 see changes in 886119f5

elrond79: To the larger topic - to me, I see the basic problem is that larger PRs will have longer code reviews.  I think Hamed did a commendable job in splitting up this work into multiple pieces, and I personally think they could be merged in one-at-a-time, as they're ready, since they have reasonably well-defined boundaries.  (Again, props to @HamedSabri-adsk for that!)

Yes, that does mean that other PRs may have to resolve merge conflicts multiple times.  However, I personally like merging in work in multiple smaller chunks, anyway.

Also - in the future, when dealing specically with some of these style / formatting issues, I think a better approach might be to make changes to only a few files - 1-5, say - and get feedback on those first.  That way, any ambiguities / disagreements / misunderstandings regarding the style guidelines and how they should be applied can get hammered out, without having to make changes to the entire codebase along the way.  Then, once we're all seeing on the same page on that limited subset, we can try expanding the changes to either a larger subset or the whole repo.  I think that would make things less cumbersome.

Regardless - we all want to make it to the end, and I think we're nearly there.  I believe there are only a few more outstanding notes to address, and then we can move on!

lib/usd/ui/CMakeLists.txt
seando-adsk: Internally we discussed this "ui" folder and I thought we decided that since it was such a small project we would remove all those sub-folders views/cmds/factories and just have all the sources in one folder (except for demo which is a separate project). Therefore we don't really need these new ui/xxx/cmake files as they would just be removed anyways.

HamedSabri-adsk: @seando-adsk Yes this is addressed as part of my post here https://github.com/Autodesk/maya-usd/pull/388#issuecomment-607543513

See this commit for chanes 9d9404a6

lib/mayaUsd/render/vp2RenderDelegate/CMakeLists.txt
kxl-adsk: Sounds good...from my side, I'm trying to make sure we stay in the limits of what originally was intended to be done since it's easy to grow big changes. While saying that, I must admit I missed this commit https://github.com/Autodesk/maya-usd/pull/389/commits/5d11e6f1cfa3612855fa864e32cc7d1c494d50e9 and pulled the trigger too quickly.

elrond79: @HamedSabri-adsk - ah, thanks for the link to 886119f - and I see the changes are indeed properly in Part I!

PR #396, opened 2020-03-31
MAYA-104134 Add .usdz extension to the identifyFile function

fowlertADSK:
Add .usdz extension to the identifyFile function, which is used to determine if it is a supported file when doing a drag'n drop onto Maya.

### DIFF
--- a/plugin/adsk/plugin/importTranslator.cpp
+++ b/plugin/adsk/plugin/importTranslator.cpp
@@ -157,7 +157,8 @@ UsdMayaImportTranslator::identifyFile(
 
     if (fileExtension == UsdMayaTranslatorTokens->UsdFileExtensionDefault.GetText() ||
         fileExtension == UsdMayaTranslatorTokens->UsdFileExtensionASCII.GetText() ||
-        fileExtension == UsdMayaTranslatorTokens->UsdFileExtensionCrate.GetText()) {
+        fileExtension == UsdMayaTranslatorTokens->UsdFileExtensionCrate.GetText() ||
+        fileExtension == UsdMayaTranslatorTokens->UsdFileExtensionPackage.GetText()) {
         retValue = kIsMyFileType;
     }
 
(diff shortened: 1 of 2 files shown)


fowlertADSK:
I enabled .usdz import here: https://github.com/Autodesk/maya-usd/pull/313
That worked fine when using the file dialog because .usdz was made part of the file filter.

When a user drops a .usdz file on top of Maya to import it, Maya will call all the translators to see if they identify the file as something they understand.  This function was still parsing the file string to get the extension and then comparing against .usd/.usda/.usdc and not .usdz.

The pxr plugin has this exact code so it will also not import a .usdz when you try to drop it on Maya either.  I can add the exact same line there, or we can leave it as a mayaUSD only thing if we want.

ppt-adsk:
> The pxr plugin has this exact code so it will also not import a .usdz when you try to drop it on Maya either. I can add the exact same line there, or we can leave it as a mayaUSD only thing if we want.

Is this something that can be factored out easily into the mayaUsd/lib core and used across multiple plugins, or is it a bit more involved?

fowlertADSK:
> > The pxr plugin has this exact code so it will also not import a .usdz when you try to drop it on Maya either. I can add the exact same line there, or we can leave it as a mayaUSD only thing if we want.
> 
> Is this something that can be factored out easily into the mayaUsd/lib core and used across multiple plugins, or is it a bit more involved?

It could be.  Right now the common code that is used by multiple plugins are the "job" classes down in the lib but this code is up in the translators.  The translators, defined in each plugin, set things like the default options or what extensions they can support, but they all derive directly from MPxFileTranslator.  

For the commands I'm actually creating a new base command class that the others derive from so they can customize just the things that are different.  We could do that here as well but I have not planned for that refactoring yet.

fowlertADSK:
> Good catch @fowlertADSK! Looks good to me!
> 
> If it's not too much trouble, it would be great to add this for the Pixar plugin also:
> https://github.com/Autodesk/maya-usd/blob/dev/plugin/pxr/maya/lib/usdMaya/importTranslator.cpp#L122

Done.

PR #398, opened 2020-03-31
MAYA-103650 moving Pixar import/export commands into lib folder

fowlertADSK:
This change moves the implementation of the Pixar Import and Export commands down into the shared lib so that other plugins can reuse them.

Main changes:

1. Created new import/export commands in lib/fileio/commands/, baseImportCommand and baseExportCommand.

2. Copied the entire implementation from the Pixar commands, minus the type of Job that is used for Import into the base commands.

3. Added const identifiers for all of the short names of command flags instead of using hard-coded strings in the implementation.  This makes it a bit cleaner in the derived classes where they can now use the identifiers when calling Maya API functions like “isFlagSet”.  The names used for the long form of the flags are still queried and set from the Job Tokens.

4. Added a new derived command class in the mayaUSD plugin to register the two commands with Maya.  No customization at all at the moment, they only exist in order to register the command with Maya (using a different name than the Pixar plugin).

5. Renamed the Pixar c++ command classes (kept the file names the same) to make it a bit more clear that they are the Pixar classes.

6. The new Pixar classes now derive from the new classes down in the shared lib.  Implementation in the new base classes are identical to what was previously in the Pixar plugin save for two changes listed next.

7. I found an error when registering a couple command flags where the Pixar plugin was using the same short name for two different flags.  Both -frameStride and -filterTypes were using -ft as their short form.  You can see this by running “help usdExport” in the script editor before my change and you’ll see that filterTypes does not show up, I believe because the overlap of the short name.  The long form of the name still worked and I believe people must have been using that in any of their scripts.  I have left -filterTypes with a short form of -ft and changed -frameStride to use -fst.

8. The Pixar Import command needs to use a different Read Job than the new base command class uses, and MayaUSD wants, so it overrides a virtual initialization function and provides the base class with a differt Job.  As of right now all of the differences between the Pixar and MayaUSD Import commands (other than registering with a different name) come from the different Job classes used.

9. The Pixar Export command uses the same Write Job class as the MayaUSD plugin so no customization is required right now for Export.

Still TODO:

1. Was waiting on some refactoring related to automated tests to be completed before adding tests for these commands (this is the main reason why this is just a draft code review).

2. Want to take a look at the command flag documentation that Pixar moved into the repo from the main USD html documentation page and decide where that should live in our doc’s.


lib/fileio/commands/baseExportCommand.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+void* MayaUSDExportCommand::creator()
+{
+    return new MayaUSDExportCommand();
+}
+
+/* virtual */
+std::unique_ptr<UsdMaya_WriteJob> MayaUSDExportCommand::initializeWriteJob(const pxr::UsdMayaJobExportArgs & args)
+{
+    return std::unique_ptr<UsdMaya_WriteJob>(new UsdMaya_WriteJob(args));
```
fowlertADSK: Both the Pixar plugin and the new MayaUSD plugin use UsdMaya_WriteJob.  This virtual function is here to allow derived plugins to change that behavior, but right now they are still identical.

lib/fileio/commands/baseExportCommand.cpp
```diff
@@ -0,0 +1,328 @@
...
+{
+}
+
+MSyntax MayaUSDExportCommand::createSyntax()
+{
+    MSyntax syntax;
+
+    // These flags correspond to entries in
+    // UsdMayaJobExportArgs::GetDefaultDictionary.
+    syntax.addFlag(kMergeTransformAndShapeFlag,
```
fowlertADSK: A minor change, but all short names are now identifiers like this instead of hard-coded strings like were in the original Pixar source.

lib/fileio/commands/baseImportCommand.cpp
```diff
@@ -0,0 +1,262 @@
...
+MayaUSDImportCommand::creator()
+{
+    return new MayaUSDImportCommand();
+}
+
+/* virtual */
+std::unique_ptr<UsdMaya_ReadJob> MayaUSDImportCommand::initializeReadJob(const MayaUsd::ImportData & data, 
+        const UsdMayaJobImportArgs & args)
+{
+    return std::unique_ptr<UsdMaya_ReadJob>(new UsdMaya_ReadJob(data, args));
```
fowlertADSK: This is the one place that the Pixar plugin will do something different.  It will override this and return a pointer to a different type of Job, the MayaUSD command will just use the one the base class creates.

lib/fileio/commands/baseExportCommand.h
```diff
@@ -0,0 +1,109 @@
...
+    static constexpr auto kAppendFlagLong = "append";
+    static constexpr auto kFilterTypesFlag = "ft";
+    static constexpr auto kFilterTypesFlagLong = "filterTypes";
+    static constexpr auto kFileFlag = "f";
+    static constexpr auto kFileFlagLong = "file";
+    static constexpr auto kSelectionFlag = "sl";
+    static constexpr auto kSelectionFlagLong = "selection";
+    static constexpr auto kFrameSampleFlag = "fs";
+    static constexpr auto kFrameSampleFlagLong = "frameSample";
+    static constexpr auto kFrameStrideFlag = "fst";
```
fowlertADSK: Like I mention in my commit message, this short form previously overlapped with the short for of -filterTypes.  Changing this to -fst fixes that.

plugin/pxr/maya/lib/usdMaya/exportCommand.cpp
```diff
@@ -13,308 +13,21 @@
...
-    syntax.addFlag("-ppc",
-                   UsdMayaJobExportArgsTokens->pythonPostCallback.GetText(),
-                   MSyntax::kString);
-    syntax.addFlag("-v",
-                   UsdMayaJobExportArgsTokens->verbose.GetText(),
-                   MSyntax::kNoArg);
-
-    // These are additional flags under our control.
-    syntax.addFlag("-fr", "-frameRange", MSyntax::kDouble, MSyntax::kDouble);
-    syntax.addFlag("-ft", "-frameStride", MSyntax::kDouble);
```
fowlertADSK: All of this has now been moved to the base class, but here you'll see the short flag conflict (lines 156 and 164)

plugin/pxr/maya/lib/usdMaya/exportCommand.h
```diff
@@ -24,23 +24,19 @@
...
 #include <maya/MPxCommand.h>
 
+#include <mayaUsd/fileio/commands/baseExportCommand.h>
+
 PXR_NAMESPACE_OPEN_SCOPE
 
 
-class UsdMayaExportCommand : public MPxCommand
+class PxrMayaUSDExportCommand : public MayaUsd::MayaUSDExportCommand
 {
```
fowlertADSK: I have left the file name as it was before, but I did change the name of the c++ class to include "Pxr" as a way to distinguish it from the others.  Base class uses "base", new Maya USD class uses "adsk", at the start.

plugin/pxr/maya/lib/usdMaya/importCommand.cpp
```diff
@@ -17,253 +17,30 @@
...
-UsdMayaImportCommand::redoIt()
-{
-    if (!mUsdReadJob) {
-        return MS::kFailure;
-    }
-
-    bool success = mUsdReadJob->Redo();
-
-    return (success) ? MS::kSuccess : MS::kFailure;
+    return std::unique_ptr<UsdMaya_ReadJob>(new UsdMaya_ReadJobWithSceneAssembly(data, args));
```
fowlertADSK: Overriding the function that initializes the read job and returning the UsdMaya_ReadJobWithSceneAssembly that the Pixar plugin uses to customize the import behavior.

lib/fileio/CMakeLists.txt
```diff
@@ -25,6 +25,8 @@ target_sources(${PROJECT_NAME}
         writeJobContext.cpp
         chaser/chaser.cpp
         chaser/chaserRegistry.cpp
+        commands/baseExportCommand.cpp
```
ppt-adsk, line 28: I guess this will conflict with Hamed's branch and will need to be reworked.

lib/fileio/commands/baseExportCommand.h
```diff
@@ -0,0 +1,109 @@
...
+    // and some that are defined by this command itself.
+    // All short forms of the Maya flag names are defined here.
+    // All long forms of flags defined by the command are also here.
+    // All long forms of flags defined by the Arg Tokens are queried
+    // for and set when creating the MSyntax object.
+    // Derived classes can use the short forms of the flags when
+    // calling Maya functions like argData.isFlagSet()
+    //
+    // The list of short forms of flags defined as Arg Tokens:
+    static constexpr auto kDefaultMeshSchemeFlag = "dms";
```
ppt-adsk, line 44: This is nice of course, but I don't see any corresponding definition for these statics in the .cpp implementation.  Are you sure this compiles on all architectures?  I believe that Sean and I ran into exactly this problem during the development of the UFE attribute base class, and we had to add definitions in the .cpp.  I can show you the (Autodesk internal) code if you'd like.

lib/fileio/commands/baseImportCommand.h
```diff
@@ -0,0 +1,88 @@
...
+    // and some that are defined by this command itself.
+    // All short forms of the Maya flag names are defined here.
+    // All long forms of flags defined by the command are also here.
+    // All long forms of flags defined by the Arg Tokens are queried
+    // for and set when creating the MSyntax object.
+    // Derived classes can use the short forms of the flags when
+    // calling Maya functions like argData.isFlagSet()
+    //
+    // The list of short forms of flags defined as Arg Tokens:
+    static constexpr auto kShadingModeFlag = "shd";
```
ppt-adsk, line 44: Same question as for export re: lack of corresponding static implementation in the .cpp.

plugin/adsk/plugin/adskExportCommand.h
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef ADSK_MAYA_EXPORT_COMMAND_H
+#define ADSK_MAYA_EXPORT_COMMAND_H
+
+#include "base/api.h"
+
+#include <maya/MPxCommand.h>
```
ppt-adsk, line 21: Still needed?

plugin/adsk/plugin/adskImportCommand.h
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef ADSK_MAYA_IMPORT_COMMAND_H
+#define ADSK_MAYA_IMPORT_COMMAND_H
+
+#include "base/api.h"
+
+#include <maya/MPxCommand.h>
```
ppt-adsk, line 21: Still needed?

lib/fileio/commands/baseImportCommand.cpp
```diff
@@ -0,0 +1,262 @@
...
+    if (argData.isFlagSet(kPrimPathFlag))
+    {
+        // Get the value
+        MString tmpVal;
+        argData.getFlagArgument(kPrimPathFlag, 0, tmpVal);
+        mPrimPath = tmpVal.asChar();
+    }
+
+    // Add variant (variantSet, variant).  Multi-use
+    std::map<std::string,std::string> mVariants;
```
mattyjams, line 158: This likely also came from Pixar, but `Sdf` provides a type for this: `SdfVariantSelectionMap`. And in fact, that's the type that `ImportData` uses:
https://github.com/Autodesk/maya-usd/blob/ec85ed3bd75e7b502530f9564cebc5ea539d33a5/lib/fileio/importData.h#L121

lib/fileio/commands/baseImportCommand.cpp
```diff
@@ -0,0 +1,262 @@
...
+    {
+        MArgList tmpArgList;
+        status = argData.getFlagArgumentList(kVariantFlag, i, tmpArgList);
+        // Get the value
+        MString tmpKey = tmpArgList.asString(0, &status);
+        MString tmpVal = tmpArgList.asString(1, &status);
+        mVariants.insert( std::pair<std::string, std::string>(tmpKey.asChar(), tmpVal.asChar()) );
+    }
+
+    bool readAnimData = true;
```
mattyjams, line 169: This option *should* be off by default, so it probably makes slightly more sense to set this to `false` here.

lib/fileio/commands/baseImportCommand.h
```diff
@@ -0,0 +1,88 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef MAYA_IMPORT_COMMAND_H
+#define MAYA_IMPORT_COMMAND_H
+
+#include "../../base/api.h"
```
mattyjams, line 20: include as `<mayaUsd/base/api.h>`

plugin/adsk/plugin/adskExportCommand.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+const MString ADSKMayaUSDExportCommand::commandName("mayaUSDExport");
+
+ADSKMayaUSDExportCommand::ADSKMayaUSDExportCommand()
+{
+}
+
+ADSKMayaUSDExportCommand::~ADSKMayaUSDExportCommand()
+{
+}
```
mattyjams, line 28: Could we `= default;` these in the header instead?

plugin/adsk/plugin/adskExportCommand.h
```diff
@@ -0,0 +1,40 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef ADSK_MAYA_EXPORT_COMMAND_H
+#define ADSK_MAYA_EXPORT_COMMAND_H
+
+#include "base/api.h"
```
mattyjams, line 19: include as `<mayaUsd/base/api.h>`

plugin/adsk/plugin/adskImportCommand.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+const MString ADSKMayaUSDImportCommand::commandName("mayaUSDImport");
+
+ADSKMayaUSDImportCommand::ADSKMayaUSDImportCommand()
+{
+}
+
+ADSKMayaUSDImportCommand::~ADSKMayaUSDImportCommand()
+{
+}
```
mattyjams, line 28: Could we `= default;` these in the header instead?

plugin/adsk/plugin/adskImportCommand.h
```diff
@@ -0,0 +1,40 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef ADSK_MAYA_IMPORT_COMMAND_H
+#define ADSK_MAYA_IMPORT_COMMAND_H
+
+#include "base/api.h"
```
mattyjams, line 19: include as `<mayaUsd/base/api.h>`

plugin/pxr/maya/lib/usdMaya/exportCommand.h
```diff
@@ -24,23 +24,19 @@
...
-    ~UsdMayaExportCommand() override;
-
-    PXRUSDMAYA_API
-    MStatus doIt(const MArgList& args) override;
-    bool  isUndoable () const override { return false; };
-
+    PxrMayaUSDExportCommand();
     PXRUSDMAYA_API
-    static MSyntax  createSyntax();
+    ~PxrMayaUSDExportCommand() override;
```
mattyjams, line 38: ` = default;` and remove from cpp?

plugin/pxr/maya/lib/usdMaya/importCommand.h
```diff
@@ -22,40 +22,27 @@
...
-
-    PXRUSDMAYA_API
-    MStatus doIt(const MArgList& args) override;
-    PXRUSDMAYA_API
-    MStatus redoIt() override;
+    PxrMayaUSDImportCommand();
     PXRUSDMAYA_API
-    MStatus undoIt() override;
-    bool isUndoable() const override { return true; };
+    ~PxrMayaUSDImportCommand() override;
```
mattyjams, line 38: ` = default;` and remove from cpp?

plugin/pxr/maya/lib/usdMaya/importCommand.h
```diff
@@ -22,40 +22,27 @@
...
-    PXRUSDMAYA_API
-    static MSyntax createSyntax();
     PXRUSDMAYA_API
     static void* creator();
 
-  private:
-    UsdMaya_ReadJobWithSceneAssembly* mUsdReadJob;
+  protected:
+    virtual std::unique_ptr<UsdMaya_ReadJob> initializeReadJob(const MayaUsd::ImportData &, 
+        const UsdMayaJobImportArgs &) override;
```
mattyjams, line 45: Just the `override` keyword at the end here, and no `virtual` at the beginning.

plugin/adsk/plugin/adskImportCommand.h
kxl-adsk: I believe it's a different api.h header, see https://github.com/Autodesk/maya-usd/blob/dev/plugin/adsk/plugin/base/api.h 
So it should be `<mayaUsdPlugin/base/api.h>`

HamedSabri-adsk: We no longer promote private headers and therefore  `#include "base/api.h" `  is correct code. Please see latest changes in 3324c613c

kxl-adsk: Ha, good catch @HamedSabri-adsk, I wasn't aware it's a private header in the plugin. Then all is good.

mattyjams: Ahh, my bad. Sorry about that! Thanks @kxl-adsk and @HamedSabri-adsk!

lib/fileio/CMakeLists.txt
fowlertADSK: Yes, I put this up as a Draft PR because it was ready for comments and figured "why wait" but always expected that I'd wait for at least Hamed's changes to go in first.

lib/fileio/commands/baseExportCommand.h
fowlertADSK: I'll ping you about it

plugin/adsk/plugin/adskExportCommand.h
fowlertADSK: Doesn't look like it.

lib/fileio/commands/baseExportCommand.cpp
```diff
@@ -0,0 +1,328 @@
...
+#include <maya/MSelectionList.h>
+
+#include <mayaUsd/fileio/jobs/jobArgs.h>
+#include <mayaUsd/fileio/utils/writeUtil.h>
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+namespace {
+
+} // namespace
```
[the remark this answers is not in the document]
fowlertADSK: oops!  originally had been putting the flag strings in there but then moved them to the .h when I realized that they would eventually be used by the derived classes to.

lib/fileio/commands/baseExportCommand.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+MAYAUSD_NS_DEF {
+
+MayaUSDExportCommand::MayaUSDExportCommand()
+{
+}
+
+MayaUSDExportCommand::~MayaUSDExportCommand()
+{
+}
```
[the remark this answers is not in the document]
fowlertADSK: I think I'll go with the rule of zero and just remove them, i.e. http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero

lib/fileio/commands/baseExportCommand.h
```diff
@@ -0,0 +1,109 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef MAYA_EXPORT_COMMAND_H
+#define MAYA_EXPORT_COMMAND_H
+
+#include "../../base/api.h"
```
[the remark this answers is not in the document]
fowlertADSK: Yeah, I'll make this change.  In general I avoided making too many changes related to #includes here because of the on-going conversation about them and that we were probably going to batch edit everything in one big clean up step.  This one was copied from somewhere else I think, but it's the one style change that we all agree should be changed so I might as well do it now.

lib/fileio/commands/baseExportCommand.h
fowlertADSK: @ppt-adsk   I chatted with Sean briefly and I think the difference between this and UFE is that your class there is used in a template where I'm not here.  I can also find plenty of other uses like this in the repo so I think I'm ok here.  

I thought I could put it in there just in case, but then (if I'm understanding it correctly) I'd have to drop the usage of "auto" because I don't think I can just declare it there without assign it anything.

Long story short, I think we're good here, but it's an interesting c++ issue to understand.

lib/fileio/commands/baseImportCommand.cpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    GfInterval timeInterval;
+    if (readAnimData) {
+        if (argData.isFlagSet(kFrameRangeFlag)) {
+            double startTime = 1.0;
+            double endTime = 1.0;
+            argData.getFlagArgument(kFrameRangeFlag, 0, startTime);
+            argData.getFlagArgument(kFrameRangeFlag, 1, endTime);
+            timeInterval = GfInterval(startTime, endTime);
```
robthebloke, line 171: I'm assuming the user could never accidentally specify these the wrong way around?

lib/fileio/commands/baseImportCommand.cpp
```diff
@@ -0,0 +1,252 @@
...
+    // Add variant (variantSet, variant).  Multi-use
+    SdfVariantSelectionMap mVariants;
+    for (unsigned int i=0; i < argData.numberOfFlagUses(kVariantFlag); ++i)
+    {
+        MArgList tmpArgList;
+        status = argData.getFlagArgumentList(kVariantFlag, i, tmpArgList);
+        // Get the value
+        MString tmpKey = tmpArgList.asString(0, &status);
+        MString tmpVal = tmpArgList.asString(1, &status);
+        mVariants.insert( std::pair<std::string, std::string>(tmpKey.asChar(), tmpVal.asChar()) );
```
robthebloke, line 156:
``` c++
mVariants.emplace(tmpKey.asChar(), tmpVal.asChar());
```

lib/fileio/commands/baseImportCommand.cpp
```diff
@@ -0,0 +1,252 @@
...
+    {
+        // Get the value
+        MString tmpVal;
+        argData.getFlagArgument(kPrimPathFlag, 0, tmpVal);
+        mPrimPath = tmpVal.asChar();
+    }
+
+    // Add variant (variantSet, variant).  Multi-use
+    SdfVariantSelectionMap mVariants;
+    for (unsigned int i=0; i < argData.numberOfFlagUses(kVariantFlag); ++i)
```
robthebloke, line 149: The value of ```argData.numberOfFlagUses(kVariantFlag)``` isn't likely to change here. Take a copy.

lib/fileio/commands/baseImportCommand.cpp
```diff
@@ -0,0 +1,252 @@
...
+        return MS::kFailure;
+    }
+
+    std::string mPrimPath;
+    if (argData.isFlagSet(kPrimPathFlag))
+    {
+        // Get the value
+        MString tmpVal;
+        argData.getFlagArgument(kPrimPathFlag, 0, tmpVal);
+        mPrimPath = tmpVal.asChar();
```
robthebloke, line 144:
```c++
mPrimPath = UsdMayaUtil::convert(tmpVal);
```

lib/fileio/commands/baseImportCommand.cpp
```diff
@@ -0,0 +1,252 @@
...
+        UsdMayaUtil::GetDictionaryFromArgDatabase(
+            argData,
+            UsdMayaJobImportArgs::GetDefaultDictionary());
+
+    std::string mFileName;
+    if (argData.isFlagSet(kFileFlag)) {
+        // Get the value
+        MString tmpVal;
+        argData.getFlagArgument(kFileFlag, 0, tmpVal);
+        mFileName = tmpVal.asChar();
```
robthebloke, line 119:
```
mFileName = UsdMayaUtil::convert(tmpVal);
```

lib/fileio/commands/baseExportCommand.cpp
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    // Convert selection list to jobArgs dagPaths
+    UsdMayaUtil::MDagPathSet dagPaths;
+    for (unsigned int i=0; i < objSelList.length(); i++) {
+        MDagPath dagPath;
+        status = objSelList.getDagPath(i, dagPath);
+        if (status == MS::kSuccess)
+        {
+            dagPaths.insert(dagPath);
```
robthebloke, line 286: emplace?

lib/fileio/commands/baseExportCommand.cpp
```diff
@@ -0,0 +1,316 @@
...
+        argData.getFlagArgument(kFrameStrideFlag, 0, frameStride);
+    }
+
+    std::set<double> frameSamples;
+    unsigned int numFrameSamples = argData.numberOfFlagUses(kFrameSampleFlag);
+    for (unsigned int i = 0; i < numFrameSamples; ++i) {
+        MArgList tmpArgList;
+        argData.getFlagArgumentList(kFrameSampleFlag, i, tmpArgList);
+        frameSamples.insert(tmpArgList.asDouble(0));
+    }
```
robthebloke, line 263: This is a pretty horrible use of std::set. Use a vector, allocate once up front, and later on sort & filter. Then later, both iterating the keys, or using std::lower_bound, would be an improvement over the set.  Something along these lines
```c++
    unsigned int numFrameSamples = argData.numberOfFlagUses(kFrameSampleFlag);
    std::vector<double> frameSamples(numFrameSamples);

    for (unsigned int i = 0; i < numFrameSamples; ++i) {
        MArgList tmpArgList;
        argData.getFlagArgumentList(kFrameSampleFlag, i, tmpArgList);
        frameSamples[i] = tmpArgList.asDouble(0);
    }
    std::sort( frameSamples.begin(), frameSamples.end() ) ;
    frameSamples.erase( std::unique( frameSamples.begin(), frameSamples.end() ), frameSamples.end() ) ;
```

fowlertADSK: Hey @robthebloke , 

First apologies for leaving this review for so long.  All of this is copied code from the Pixar plugin down into the core and this draft review was mostly put here to give people the heads-up that it was getting moved.  We had this in-progress, but we knew we were going to pause work on a bit and wouldn’t be merged until at least some automated tests were added.  I didn’t even realized that you’d commented here until just recently but have made most of the changes you pointed out (and will push in a bit).

This one though…  The interface to UsdMayaWriteUtil::GetTimeSamples requires a std::set<double> so we do need that.  And if I understand how the “frame samples” flag is typically used, I think we would only normally have a couple values added here, so I don’t think pre-allocating a vector + sorting + filter is really required.

PR #392, opened 2020-03-27
PART V: clean up mayaUsdPlugin

HamedSabri-adsk:
- mayaUsdPlugin promotion headers are now created under mayaUsdPlugin and not mayaUsd
- clean up include order and directive

PR #400, opened 2020-04-01
codingGuidelines.md: clarify some details about headers

elrond79:
- make it explicit that associated headers don't use public formatting
- make it clear that all headers from this repo (maya-usd) go at end,
  not just ones from this library within maya-usd
- add a new include-order category for private headers
- remove some fancy unicode quotes from code blocks

Arose out of some discussion in https://github.com/Autodesk/maya-usd/pull/392 (and other associated PRs)


doc/codingGuidelines.md
```diff
@@ -166,14 +172,15 @@ Headers should be included in the following order, with each section separated b
 1. Related header
 2. C system headers
 3. C++ standard library headers
-4. Other libraries’ headers
+4. Other external libraries’ headers
 5. Autodesk + Maya headers
 6. Pixar + USD headers
-7. Your project’s headers
-8. Conditional includes
+7. All public headers from this project (maya-usd)
```
kxl-adsk, line 178: maya-usd is an example in here? if you are in hdMaya, then this is where you would put your public header files.

doc/codingGuidelines.md
```diff
@@ -156,8 +162,8 @@ All included public header files from outside and inside the project should be `
 
 Private project’s header files should be `#include`'d using the file name when in the same folder. Private headers may live in sub-directories, but they should never be included using "._" or ".._" as part of a relative paths. For example:
```
kxl-adsk, line 163: One comment I heard about this line was to make it more explicit about the fact we are using relative paths, i.e. not a full path as above.

kxl-adsk:
BTW. we will merge this only after PRs #388 - #392 are merged.

doc/codingGuidelines.md
elrond79: I meant maya-usd, the name of the whole repo.  So if you were in hdMaya, then you would put hdMaya and mayaUsd (and mayaUsdUI, etc) headers all in here.  I pushed an update which clarifies my intent a bit (replaced "project" with "repository").

This can be a matter for debate - I can think of three options here:
[I'm using your example of being in the hdMaya library, and deciding where to put hdMaya headers (same sub-unit) vs mayaUsd and mayaUsdUIheaders (same repo, different sub-unit).]

- **A.** Only headers from same sub-unit (ie, hdMaya) go in **7.**  Headers from maya-usd repo but other sub-units (ie, mayaUsd + mayaUsdUI) go in **4.**.
- **B.** All headers from the maya-usd repo (ie, hdMaya, mayaUsd, and mayaUsdUI) all go in **7.**.  Only headers from outside of the maya-usd repo go in **4.**
- **C.** We add a new category to separate same sub-unit headers from other sub-units, ie:
  
  - **7.** Headers from this repo, but a different sub-unit than current
  - **8.** Public headers form this sub-unit
  
  Obviously, exact wording would change once we determine exactly what we mean by a "sub-unit".
  
Personally, I prefer **A.** because:

- Conceptually, it's simplest + clearest - no need to figure out what is mean by a "sub-unit"
- As opposed to **B.*, it aligns with the most-general to most-specific order. In **B.**, we would have hdMaya headers come before maya headers, which violates the most-general to most-specific principle.
- It's easiest to enforce with clang-format - we simply need a single project-level `.clang-format` file, with a regexp to catch all headers from the maya-usd project (ie, `mayaUsd|mayaUsdUI|hdMaya|...`.  In theory, you could still enforce **B.** or **C.** with clang-format, but it would become a lot more complicated - you'd need to have a separate clang-format file at each sub-unit, just to override the header ordering rules.  And, since I don't think either clang-format or yaml allow for hierarchical overrides or "include" functionality<sup>1</sup>, we'd have to have complete copies at each level, with just the header ordering changed.

If we don't vote for **A.**, my next preference would be **C.** We'd be introducing yet another group, but assuming we can get things managed by .clang-format, this doesn't scare me overmuch.

Regardless, if we go with either **A.** or **C.**, we would still need to decide what was meant by a "sub-unit" - personally, I would vote for a broad definition - ie, have only 3 - pxr stuff (ie, in `plugin/pxr`), al stuff (in `plugin/al`), and everything else.  (For symmetry, if we also wanted to make `plugin/adsk` it's own sub-unit, and have 4 total, I'd also be fine with that.)  That means that, ie, `mayaUsd`, `mayaUsdUI`, and `hdMaya` would always get grouped together, but at least the stuff from the al and pixar plugins would be separated out, and it would keep the number of `.clang-format` files to a manageable level.

If we try to get more granular, we'll not only have an explosion in the number of `.clang-format` files, it gets a bit fuzzier about what is a sub-unit.  (Do we go by cmake-project? Because there's currently some libraries that don't have a separate cmake-project.  Do we go by output files, ie, libraries? If so, then what about utility headers with no associated .cpp?)

Finally - if we go with **B.**, then it's worth noting that even though things like hdMaya and mayaUsd will get lumped together, simply due to alphabetical sorting they will at least be mostly lumped next "similar things", ie:

```
#include <hdMaya/other.h>
#include <hdMaya/thingie.h>
#include <mayaUsd/bar.h>
#include <mayaUsd/foo.h>
#include <mayaUsd/ui/blah.h>
```

Notes:
1. I'd love to be wrong about this, but a quick search I did for yaml and and clang-format didn't turn up any way to do hierarchical overrides or includes...

doc/codingGuidelines.md
elrond79: Will make that change...

elrond79: See: https://github.com/Autodesk/maya-usd/pull/400/commits/dd99332698132e11a7416327a4089502ed7dd787

elrond79:
Clarified the relative-paths-for-private-includes issue you pointed out.

As for this going in after those other PRs, that's fine.

doc/codingGuidelines.md
kxl-adsk: My understanding was that we are heading towards option A (based on the discussion we had during TSC call). This is why the mention of `maya-usd` in the guideline confused me. Let's see what others think.

mattyjams: Yeah, I'd opt for **A** as well. I think that's most inline with what we've been discussing.

doc/codingGuidelines.md
```diff
@@ -148,16 +148,22 @@ In general, macros should be avoided (see [Modern C++](https://docs.google.com/d
 * Comments for users of classes and functions must be written in headers files. Comments in definition files are meant for contributors and maintainers.
 
 ### Include directive
+For source files (.cpp) with an associated header file (.h) that resides in the same directory, it should be `#include`'d with double quotes and no path.  This formatting should be followed regardless with whether the associated header is public or private. For example:
```
mattyjams, line 151: Grammar police: "regardless of"?

doc/codingGuidelines.md
```diff
@@ -148,16 +148,22 @@ In general, macros should be avoided (see [Modern C++](https://docs.google.com/d
...
+```
+
 All included public header files from outside and inside the project should be `#include`’d using angle brackets. For example:
 ```cpp
 #include <pxr/base/tf/stringUtils.h>
 #include <mayaUsd/nodes/stageData.h>
 ```
 
-Private project’s header files should be `#include`'d using the file name when in the same folder. Private headers may live in sub-directories, but they should never be included using "._" or ".._" as part of a relative paths. For example:
+Private project’s header files should be `#include`'d using the quotes, and a relative path. Private headers may live in the same directory or sub-directories, but they should never be included using "._" or ".._" as part of a relative paths. For example:
```
mattyjams, line 163: Towards the beginning: "using double quotes" instead of "using the quotes"
Towards the end: "as part of a relative path." instead of "as part of a relative path**s**."(plural)

doc/codingGuidelines.md
```diff
@@ -187,6 +194,8 @@ Headers should be included in the following order, with each section separated b
 #include <mayaUsd/fileio/shading/shadingModeRegistry.h>
```
mattyjams, line 194: Maybe add a group with a pxr header or two just before the mayaUsd ones for completeness?

doc/codingGuidelines.md
elrond79: Hmm - I'm curious to hear your folk's thoughts for why you prefer **A**.  It seems to me the worst of both worlds:

- **A.**
  :-1: Non-ideal header ordering (doesn't preserve most-generic to most-specific)
  :-1: Difficult to automate + enforce via clang-format
- **B.**
  :-1: Non-ideal header ordering (no distinction between this lib and other libs within maya-usd)
  :+1: Easy to automate + enforce via clang-format
- **C.**
  :+1: Best header ordering (preserves most-generic to most-specific, differentiates within maya-usd)
  :-1: Difficult to automate + enforce via clang-format

 To clarify what I mean about **A** not having good ordering - @mattyjams, you asked me to [add a pxr header, so we could show that it came before the mayaUsd headers](https://github.com/Autodesk/maya-usd/pull/400#discussion_r401970266).  The problem with **A** is that usually *won't* be the case - mayaUsd headers will generally come before pxr headers!  Here's an example of header ordering with the 3 options:

*Note - I'm using 10/20/30/40 so these arbitrary numbers aren't confused with the include-grouping numbers defined in `codingGuidelines.md`*

**A**
```
\\\file hdMaya/source.cpp

#include <vector>         // 10. most generic

#include <mayaUsd/foo.h>  // 30. more specific

#include <pxr/usd/bar.h>  // 20. more generic

#include <hdMaya/baz.h>   // 40. most specific
```

**B**
```
\\\file hdMaya/source.cpp

#include <vector>         // 10. most generic

#include <pxr/usd/bar.h>  // 20. more generic

#include <hdMaya/baz.h>   // 40. most specific
#include <mayaUsd/foo.h>  // 30. more specific
```

**C**
```
\\\file hdMaya/source.cpp

#include <vector>         // 10. most generic

#include <pxr/usd/bar.h>  // 20. more generic

#include <mayaUsd/foo.h>  // 30. more specific

#include <hdMaya/baz.h>   // 40. most specific
```

Note that though **B** gets the ordering in this case "wrong", the two that are switched are at least within the same *group*, so at least the ordering of the groups is always correct - as opposed to **A**, where the group ordering is generally wrong.

kxl-adsk: @elrond79 I see you are bringing automation often...and mentioning clang-format. Can you elaborate on how you want to achieve this? TMK clang-format allows you to sort within a group, but it doesn't re-order groups. 
I believe there is a way to get a check with clang-tidy, but this needs investigation.

elrond79: Oh, damnit... I just realized that in my initial comment, I was saying I preferred **A**, when I meant to say I preferred **B**.  
:man_facepalming:

So sorry for the confusion.  I'm going to edit my original comment, with strikethrough, so it matches my intent.

elrond79: Man, so sorry about that.  I messed up my own system... 🤦‍♂  I think my brain at some point just started associated A with "the one I liked"...  I introduced the letters in an attempt to make things more concrete, and instead I just made things way more confusing. Ugh...

elrond79: @kxl-adsk - Yes, clang-format has an option where it can enforce group ordering. It's not on by default - but if you set `IncludeBlocks` property to `Regroup` it will do it for you.  We actually used this on maya-to-hydra (when it was a separate repo), and I really liked it - was one less thing you never had to worry about.  Granted, it didn't always give us the "ideal" header ordering, but it was generally "mostly" right, and not having to think about it ever was definitely worth it in my book.

kxl-adsk: No worries...and let me think about it. FWIW. I wasn't aware of `Regroup` option...and now when I look at it, I don't see how it can be configured to capture some of our guidelines (example...not to group associated header files with private header files; or place std includes are the right place)

mattyjams: Ahh, so I went with **A** based on the broad definition of sub-unit, so I expected to see a result like your example for **B**. :P

I agree that **B** is the ordering I would like to see. I don't see a lot of value in breaking up the maya-usd repo into "projects" or "sub-units" such that includes of maya-usd headers end up in different groups. We could maybe make an exception for the plugin headers, but I'm not sure whether that's necessary either. It doesn't particularly bother me that a `mayaUsd` header would follow a `hdMaya` header, since I'd prefer to think of those as being in the same category.

As far as `clang-format` vs. `clang-tidy` I'm hoping we can get away with defining include category priorities as described by the "IncludeCategories" section here:
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#configuring-style-in-code

It sounds like you can also write more sophisticated checks for `clang-tidy`, but that looks to be a lot more involved, as you'd need to build it against LLVM:
https://clang.llvm.org/extra/clang-tidy/Contributing.html

I *think* `clang-format` should be sufficient.

kxl-adsk: From my perspective, the most important factor is automation. I don't think anyone really wants to spend time reviewing and commenting on include order. So thank you for bringing up `IncludeCategories` - this was the detail I was missing. And since you have experience with it, how we will separate `Related header` from `All private headers` ?

FYI. I will bring this up with our team to see where people are landing.

elrond79: If we go with **B**, I think clang-format should be able to handle all our categories.  As @mattyjams pointed out, clang-format actually has built-in support for the `Related header`... and, in fact, we probably won't even need to mess with the `IncludeIsMainRegex` - by default it picks up `include "foo.h"` as being the related header for `foo.cpp`, so I think we're good there.

I don't actually have extensive knowledge of .clang-format, either - Pal first introduced me to it when we was setting up maya-to-hydra, but I really saw it's benefits there.

If we go with **A** or **C** - I was thinking about it a bit more, and it WOULD be doable, with some more work - we'd have to have our .clang-format defined as a template, where cmake would fill in the appropriate "sub-unit" header regex everywhere we want that changed.  I would still prefer **B** though.

doc/codingGuidelines.md
elrond79: Agreed that we should probably add a pxr header to help clarify... will hold off on doing that until we make a decision [here](https://github.com/Autodesk/maya-usd/pull/400#discussion_r401844482), however.

doc/codingGuidelines.md
elrond79: Fixed in 2a53f95a0fae1e22002e158ba7985dacf9a9b3c7

elrond79:
Thanks for catching those mistakes @mattyjams - should be fixed now.

doc/codingGuidelines.md
kxl-adsk: @elrond79 Can you make a test to configure clang-format with rules described by option **B** and run it by a few files?

doc/codingGuidelines.md
```diff
@@ -148,16 +148,22 @@ In general, macros should be avoided (see [Modern C++](https://docs.google.com/d
 * Comments for users of classes and functions must be written in headers files. Comments in definition files are meant for contributors and maintainers.
 
 ### Include directive
+For source files (.cpp) with an associated header file (.h) that resides in the same directory, it should be `#include`'d with double quotes and no path.  This formatting should be followed regardless of with whether the associated header is public or private. For example:
```
mattyjams, line 151: Ah, sorry, this now reads "regardless of with whether" as opposed to "regardless of whether". :)

doc/codingGuidelines.md
mattyjams: Sounds good. I'm on board for option **B**.

doc/codingGuidelines.md
elrond79: I can, but I'm a bit busy at the moment.  I will make time to do this sometime next week.

elrond79:
Ok, reversed the header order, as discussed on the forum, to:

1. Related header
2. All private headers
3. All public headers from this repository (maya-usd)
4. Pixar + USD headers
5. Autodesk + Maya headers
6. Other libraries’ headers
7. C++ standard library headers
8. C system headers
9. Conditional includes

...and codified in .clang-format.

.clang-format
```diff
@@ -21,6 +21,57 @@ BreakConstructorInitializers: BeforeComma
 BreakInheritanceList: BeforeComma
 ColumnLimit: '100'
 FixNamespaceComments: 'true'
+IncludeBlocks: Regroup
+IncludeCategories:
+
+  # Desired final ordering:
```
kxl-adsk, line 27: @elrond79 You copy-pasted from wrong clang-format file, i.e. this doesn't match what we have below in coding guidelines document.

PR #390, opened 2020-03-27
PART III: Clean up include directives and order under lib/usd

HamedSabri-adsk:
Clean up include directives and orders under lib/usd.

PR #407, opened 2020-04-06
Address remaning feedbacks for lib/mayaUsd

HamedSabri-adsk:
Hi @elrond79 and @mattyjams ,

First of all thank you for your patience, detailed and thoughtful feedback, and very generous comments. It is really appreciated. 

As you have noticed, I've merged PR **5,4,3,2** to **1**  to make it easier to see where things are with all these new changes.

I am now creating 2 small PRs (one for lib/mayaUsd and one for lib/usd) to address the renaming issues that were brought up in previous PRs review. 

The changes are not that many and hopefully this should get us closer to the finishing line:

Changes in this PR :

- fix conditional orders based on coding guideline
- fix unnecessary promotion of private headers
- fix "" vs <> for a few public header
- fix grouping/sorting

NOTE: @mattyjams I didn't address your concern around shaderFragments.h in this PR since it requires a little bit of change. I totally agree with your comment and will have it fixed once this big PR goes in.

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
@@ -39,7 +39,6 @@
 #include <hdMaya/delegates/delegateRegistry.h>
 #include <hdMaya/delegates/sceneDelegate.h>
 #include <hdMaya/utils.h>
-
 #include <mayaUsd/render/px_vp20/utils.h>
 
 #include "pluginDebugCodes.h"
--- a/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
@@ -35,6 +35,18 @@
 #include <pxr/base/vt/value.h>
 #include <pxr/imaging/glf/contextCaps.h>
 #include <pxr/imaging/hd/rprim.h>
+
+#include <hdMaya/delegates/delegateRegistry.h>
+#include <hdMaya/delegates/sceneDelegate.h>
+#include <hdMaya/utils.h>
+
+#include <mayaUsd/render/px_vp20/utils.h>
+
+#include "pluginDebugCodes.h"
+#include "renderOverrideUtils.h"
+#include "tokens.h"
+#include "utils.h"
+
 #if USD_VERSION_NUM >= 1911
 #include <pxr/imaging/hd/rendererPluginRegistry.h>
 #else
@@ -56,17 +68,6 @@ PXR_NAMESPACE_CLOSE_SCOPE
 #include <ufe/observableSelection.h>
 #include <ufe/selectionNotification.h>
 #endif // WANT_UFE_BUILD
-
-#include <hdMaya/delegates/delegateRegistry.h>
-#include <hdMaya/delegates/sceneDelegate.h>
-#include <hdMaya/utils.h>
-
-#include <mayaUsd/render/px_vp20/utils.h>
-
-#include "pluginDebugCodes.h"
-#include "renderOverrideUtils.h"
-#include "tokens.h"
-#include "utils.h"
 
 PXR_NAMESPACE_OPEN_SCOPE
 
--- a/lib/mayaUsd/CMakeLists.txt
+++ b/lib/mayaUsd/CMakeLists.txt
@@ -58,8 +58,6 @@ target_include_directories(${PROJECT_NAME}
         ${PXR_INCLUDE_DIRS}
         $<$<BOOL:${UFE_FOUND}>:${UFE_INCLUDE_DIR}>
         ${CMAKE_BINARY_DIR}/include
-    PRIVATE
-        ${CMAKE_CURRENT_SOURCE_DIR}
 )
 
 # -----------------------------------------------------------------------------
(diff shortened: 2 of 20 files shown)


lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
```diff
@@ -35,6 +35,18 @@
 #include <pxr/base/vt/value.h>
 #include <pxr/imaging/glf/contextCaps.h>
 #include <pxr/imaging/hd/rprim.h>
+
+#include <hdMaya/delegates/delegateRegistry.h>
+#include <hdMaya/delegates/sceneDelegate.h>
+#include <hdMaya/utils.h>
+
+#include <mayaUsd/render/px_vp20/utils.h>
```
mattyjams, line 43: I think we're saying these should be one group?

HamedSabri-adsk: Please see b6cb901eb

HamedSabri-adsk:
> 
> 
> The changes here look good to me!
> 
> Just noted one thing, and then I think there are a handful of things for `lib/mayaUsd` still left from my last review of #391. I went through that list and resolved everything that was addressed.
> 
> Thanks @HamedSabri-adsk!

@mattyjams  see 1439d2e4c . shaderFragments.h is indeed a private header and doesn't have to be promoted.

mattyjams:
> @mattyjams see [1439d2e](https://github.com/Autodesk/maya-usd/commit/1439d2e4ca29a448728879fe67dec3d88b629c0d) . shaderFragments.h is indeed a private header and doesn't have to be promoted.

Sorry @HamedSabri-adsk! You actually had this right already. Since we have a cpp file in `mayaUsd/nodes` including a header from `mayaUsd/render`, that header **needs** to be public, so it should be getting both promoted and installed. I think we'd want to not only revert 1439d2e4ca29a448728879fe67dec3d88b629c0d, but also make sure that we install `${headers}` in the `vp2ShaderFragments/CMakeLists.txt`.
https://github.com/Autodesk/maya-usd/pull/391#discussion_r404544938

HamedSabri-adsk:
> Sorry @HamedSabri-adsk! You actually had this right already. Since we have a cpp file in `mayaUsd/nodes` including a header from `mayaUsd/render`, that header **needs** to be public, so it should be getting both promoted and installed. I think we'd want to not only revert [1439d2e](https://github.com/Autodesk/maya-usd/commit/1439d2e4ca29a448728879fe67dec3d88b629c0d), but also make sure that we install `${headers}` in the `vp2ShaderFragments/CMakeLists.txt`.
> [#391 (comment)](https://github.com/Autodesk/maya-usd/pull/391#discussion_r404544938)

Hey @mattyjams , no worries but I still agree with your original comment that shaderFragments.h is a private header. Here is why:

vp2ShaderFragments is not really a target and rather just a sub-directory in mayaUsd echo system.

shaderFragments.h is only used inside mayaUsd target ( e.g mayaUsd\nodes\proxyShapePlugin.cpp ) and doesn't really have to be promoted in order to be used. mayaUsd/nodes is not also a target and rather another sub-directory in mayaUsd echo system.

mattyjams:
> > Sorry @HamedSabri-adsk! You actually had this right already. Since we have a cpp file in `mayaUsd/nodes` including a header from `mayaUsd/render`, that header **needs** to be public, so it should be getting both promoted and installed. I think we'd want to not only revert [1439d2e](https://github.com/Autodesk/maya-usd/commit/1439d2e4ca29a448728879fe67dec3d88b629c0d), but also make sure that we install `${headers}` in the `vp2ShaderFragments/CMakeLists.txt`.
> > [#391 (comment)](https://github.com/Autodesk/maya-usd/pull/391#discussion_r404544938)
> 
> Hey @mattyjams , no worries but I still agree with your original comment that shaderFragments.h is a private header. Here is why:
> 
> vp2ShaderFragments is not really a target and rather just a sub-directory in mayaUsd echo system.
> 
> shaderFragments.h is only used inside mayaUsd target ( e.g mayaUsd\nodes\proxyShapePlugin.cpp ) and doesn't really have to be promoted in order to be used. mayaUsd/nodes is not also a target and rather another sub-directory in mayaUsd echo system.

Heh, in that case, I retract my original comment (https://github.com/Autodesk/maya-usd/pull/390#discussion_r404307161). :)

In #390, I had thought that we were mistakenly promoting `shaderFragments.h` because all I saw there were the changes to the CMakeLists.txt. Then when I looked at #391, I realized that it was indeed a public header and then understood why the CMakeLists.txt changes in #390 were there (https://github.com/Autodesk/maya-usd/pull/391#discussion_r404337448).

Having a cpp file in `mayaUsd/nodes` reach across directories to include a header from `mayaUsd/render` is not kosher to me unless that header is public (promoted and installed, and included with angle braces and the full path). Whether the build system treats them as distinct targets or not, from a code perspective they are distinct components.

As it stands right now, we're violating our coding guidelines in how we include the header:
```#include "render/vp2ShaderFragments/shaderFragments.h"```

And we also have to add `${CMAKE_CURRENT_SOURCE_DIR}` as an include directory to make that work, which we've decided we don't want to do.

HamedSabri-adsk:
@mattyjams I am happy to revert the change for the sake of unblocking us and we can discuss this in future PRs.  

I hope this helps.

HamedSabri-adsk:
@mattyjams see 5d3e00747. 

I believe this concludes all I wanted to cover in this PR and hopefully didn't muddy the water.

PR #409, opened 2020-04-07
Update mayaUsd_add_test macro calls in AL plugin to add support for additional schema paths

murphyeoin:
Small change to allow AL to continue to have our own separate schemas package whose PXR_PLUGINPATH_NAME can be passed to the test macro at runtime... this would be added by our top-level CMake file.
These are the only changes needed  to get our rez-based tests running - so thanks a lot to @HamedSabri-adsk for all his work getting the python test infrastructure working so smoothly compared to what was there before.


lib/usd/translators/CMakeLists.txt
```diff
@@ -82,6 +82,7 @@ target_link_libraries(${TARGET_NAME}
         usdUtils
         vt
         ${MAYA_LIBRARIES}
+        OpenMayaFX
```
murphyeoin, line 85: I know @robthebloke has added this before and it's been rejected (I can't find that PR at the moment) but it's definitely needed as a dependency at least in maya-2019, when running our tests using ALs test/build infrastructure

mattyjams: For what it's worth, we definitely needed this when usdMaya was in the USD repo (v19.11):
https://github.com/PixarAnimationStudios/USD/blob/4b1162957b2ad219e1635c81de8343be490e41fc/third_party/maya/plugin/pxrUsdTranslators/CMakeLists.txt#L24

This was because the translator for Maya `stroke` nodes used an OpenMayaFX header:
https://github.com/PixarAnimationStudios/USD/blob/4b1162957b2ad219e1635c81de8343be490e41fc/third_party/maya/plugin/pxrUsdTranslators/strokeWriter.cpp#L44

murphyeoin: Thanks @mattyjams ... The link error we're seeing is related to [MFnParticleSystem](https://help.autodesk.com/view/MAYAUL/2019/ENU/?guid=Maya_SDK_MERGED_cpp_ref_class_m_fn_particle_system_html) in [particleWriter.cpp](https://github.com/Autodesk/maya-usd/blob/1c48ce453961448b2916248322449b3737437f87/lib/usd/translators/particleWriter.cpp#L36)

murphyeoin:
Hi @HamedSabri-adsk I didn't know that adding a lib that way breaks windows - good to know. I will follow up on the ${MAYA_LIBRARIES} and let you know asap

murphyeoin:
@HamedSabri-adsk - that helped me find the problem (at our end) - thanks for your help... I've removed that line

PR #410, opened 2020-04-08
fix some of the ufe headers promoted incorrectly.

HamedSabri-adsk:
Hi @ppt-adsk, I noticed accidentally some of the ufe header files are not promoted correctly according to the UFE version. 

Looking back in history I noticed this originated from one of my PRs. Je suis de soleil.

### DIFF
--- a/lib/mayaUsd/ufe/CMakeLists.txt
+++ b/lib/mayaUsd/ufe/CMakeLists.txt
@@ -70,18 +70,18 @@ set(headers
     UsdUndoRenameCommand.h
     Utils.h
     UfeVersionCompat.h
+    UsdAttribute.h
+    UsdAttributes.h
+    UsdContextOps.h
+    UsdObject3d.h
+    UsdUndoCreateGroupCommand.h
 )
 
 if(CMAKE_UFE_V2_FEATURES_AVAILABLE)
     list(APPEND headers
-        UsdAttribute.h
-        UsdAttributes.h
         UsdAttributesHandler.h
-        UsdContextOps.h
         UsdContextOpsHandler.h
-        UsdObject3d.h
         UsdObject3dHandler.h
-        UsdUndoCreateGroupCommand.h
     )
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/CMakeLists.txt
```diff
@@ -70,18 +70,18 @@ set(headers
     UsdUndoRenameCommand.h
     Utils.h
     UfeVersionCompat.h
+    UsdAttribute.h
+    UsdAttributes.h
+    UsdContextOps.h
+    UsdObject3d.h
+    UsdUndoCreateGroupCommand.h
```
kxl-adsk, line 77: Only UsdUndoCreateGroupCommand.h seems to be V1. Everything else is V2 and should remain in the list below.

seando-adsk: Hamed, all of these are UFE V2 files and should remain where they are.

HamedSabri-adsk:
Let me revisit this carefully.  I put this PR last night prematurely. sigh...

HamedSabri-adsk:
> 
> 
> The real fix for the compilation of #388 with ufe v1 is to make conditional include of UsdUndoCreateGroupCommand in UsdHierarchy.cpp
> 
> ```
> #ifdef UFE_V2_FEATURES_AVAILABLE
> #include <mayaUsd/ufe/UsdUndoCreateGroupCommand.h>
> #endif
> ```

Indeed, Thanks @kxl-adsk .  I see now.......... What was I thinking? :) 

See b1c3c561ce0e7721ca06728f420c5482f9b8e32d

lib/mayaUsd/ufe/CMakeLists.txt
HamedSabri-adsk: Thanks @seando-adsk . Indeed. I had a momentary brain damage when I made the change :)

PR #411, opened 2020-04-08
Fix the build when project is not created with Qt.

HamedSabri-adsk:


plugin/adsk/plugin/CMakeLists.txt
```diff
@@ -51,10 +51,21 @@ target_include_directories(${TARGET_NAME}
...
 target_link_libraries(${TARGET_NAME}
     PRIVATE
-        $<$<BOOL:${Qt5_FOUND}>:mayaUsdUI>
         $<$<BOOL:${UFE_FOUND}>:${UFE_LIBRARY}>
 )
 
+if (${Qt5_FOUND})
+    target_link_libraries(${TARGET_NAME}
+        PRIVATE
+            mayaUsdUI
```
kxl-adsk, line 60: So we rely here on the fact that mayaUsdUI is dependent on MayaUsd?

HamedSabri-adsk: correct

seando-adsk: I think you can do this all on single line with:
```$<IF:$<BOOL:${Qt5_FOUND}>,mayaUsdUI,mayaUsd>```

PR #395, opened 2020-03-31
upgrade maya-usd project to the C++14 standard

mattyjams:
Conforming to [VFX Reference Platform](http://vfxplatform.com/) CY2018+ and following suit from core USD, this change sets the maya-usd project to use the C++14 standard.

Core USD moved to C++14 with release 20.05 in this commit:
https://github.com/PixarAnimationStudios/USD/commit/afc0d5e3d45eeaf6579b9699250da37cd9149c6f

PR #412, opened 2020-04-08
The aim of this PR is to allow compiler flags to be easily and clearly added within Maya USD build echo system.

HamedSabri-adsk:
### PR Goal
The aim of this PR is to allow compiler flags to be **easily** and **clearly** added within Maya USD build echo system using modern cmake.

### Background
When we originally started the Maya USD project, I brought in Pixar's compiler configuration( a.k.a defaults ) under the top level cmake to consume it's configurations. 

This was done for two reasons: 
1- It already worked out of the box when we migrated third-party/maya and it was convenient to use them for the sake of saving time.
2- other projects ( e.g Animal Logic ) was also consuming the exact same compiler configuration.

### Problems
1- Pixar's compiler flags/options have a global affect and setting/appending to CMAKE_CXX_FLAGS is no longer a good practice ( see CMake documentation around this ). Everytime a compiler flag/option is added/removed/modified, it affects the entire project and literally everyone (Animal Logic , mayaUsd, mayaUsdPlugin, pxrusd). 

2- Many of the compiler flags/options that came with defaults are either redundant or not needed at all which are carried over from core usd and old third-party/maya.

3- Hard to fine tune compiler options/flags around specific targets.

### Solution
Introducing `compiler_config.cmake` where all the compiler flags/options are applied to a specific cmake echo system (  e.g mayaUsd and mayaUsdPlugin ). We use modern cmake functions ( `target_compile_features`, `target_compile_options`, `target_compile_definitions`) to control the options and flags. All of the compiler flags for Gnu, Clang, VS are picked as needed and are 
**privately** added via `mayaUsd_compile_config` function.

### Where is cmake/defaults?
`cmake/defaults` is moved to it's original place under `plugin/pxr/cmake/defaults`. I also made a small surgery in top-level AL's cmakelist to continue consuming pxr/cmake/defaults as before.


cmake/compiler_config.cmake
```diff
@@ -0,0 +1,79 @@
...
+    -msse3
+    # disable warnings
+    -Wno-deprecated
+    -Wno-deprecated-declarations
+    -Wno-unused-local-typedefs
+)
+
+set(msvc_flags
+    # we want to be as strict as possible
+    /W3
```
HamedSabri-adsk, line 18: We currently have to use level 3 (/W3) instead of level 4(/W4) for Visual studio compiler since Pixar's USD core needs to adopt their API library to respect this level. Switching to /W4 is out of commission at the moment.

More on /W4 here:
https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=vs-2019

cmake/compiler_config.cmake
```diff
@@ -0,0 +1,79 @@
...
+# Don not use GNU extension 
+# Use -std=c++11 instead of -std=gnu++11
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+function(mayaUsd_compile_config TARGET)
+    # required compiler feature
+    target_compile_features(${TARGET} 
+        PRIVATE
+            $<$<VERSION_LESS:MAYA_APP_VERSION,2019>:cxx_std_11>
+            $<$<VERSION_GREATER_EQUAL:MAYA_APP_VERSION,2019>:cxx_std_14>
```
HamedSabri-adsk, line 62: Maya2018 ---> c++11
Maya2019 and above ---> c++14

plugin/pxr/cmake/macros/Private.cmake
```diff
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-include(Version)
```
HamedSabri-adsk, line 16: This is redundant. It is already added at the top level cmake.

cmake/compiler_config.cmake
mattyjams, line 62: I'm a little conflicted on burying the C++ language standard inside `mayaUsd_compile_config()`. It feels like this should be more of a repo-level setting. Do we envision having different parts of the repo on different standards? That would seem a bit weird to me.

And somewhat related to that, which parts of the tree do we expect would *not* call `mayaUsd_compile_config()` for their targets? For now, maybe that's the AL and pxr plugins, but do we expect that to be the case long term? I'm hoping to make some time in the not too distant future to remove all the pxr build stuff from the remaining pxr plugin code.

lib/mayaUsd/CMakeLists.txt
```diff
@@ -29,9 +29,17 @@ endif()
 find_package(Boost COMPONENTS filesystem REQUIRED)
 
 # -----------------------------------------------------------------------------
-# compile definitions
+# compiler configuration
 # -----------------------------------------------------------------------------
-target_compile_definitions(${PROJECT_NAME} 
+target_compile_definitions(${PROJECT_NAME}
+    PUBLIC
+        USD_VERSION_NUM=${USD_VERSION_NUM}
```
mattyjams, line 36: This seems ok for now, but since it comes from a dependency, maybe it should be on the upstream libraries to ensure this is set on the imported target with `INTERFACE_COMPILE_DEFINITIONS`?

plugin/al/CMakeLists.txt
```diff
@@ -51,6 +51,27 @@ set(AL_USDMAYA_LOCATION_NAME
...
 )
 
+#------------------------------------------------------------------------------
+# compiler configuration
+#------------------------------------------------------------------------------
+# CXXDefaults will set a variety of variables for the project.
+# Consume them here. This is an effort to keep the most common
+# build files readable.
+list(APPEND CMAKE_MODULE_PATH
+    ${PROJECT_SOURCE_DIR}/../pxr/cmake/defaults
```
mattyjams, line 61: Related to the comment above about setting the C++ language standard in `mayaUsd_compile_config()`, wouldn't we *also* need the change in #395 so that `CXXDefaults` specifies the language standard here? This does *not* use `mayaUsd_compile_config()`, so it seems like no standard is being specified.

plugin/pxr/CMakeLists.txt
```diff
@@ -9,15 +9,34 @@ set(PXR_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX}/${INSTALL_DIR_SUFFIX})
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
     ${PROJECT_SOURCE_DIR}/cmake/modules
     ${CMAKE_CURRENT_SOURCE_DIR}/cmake/macros
+    ${CMAKE_CURRENT_SOURCE_DIR}/cmake/defaults
```
mattyjams, line 12: Same thing here. No language standard is being specified as far as I can see.

cmake/compiler_config.cmake
HamedSabri-adsk: > I'm a little conflicted on burying the C++ language standard inside mayaUsd_compile_config(). 

There are two ways of adding C++ standard in modern CMake.

1- Setting the C++ standard directly before any targets being defined which is the approach you have in your PR:

```
set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

This setting applies to all targets `globally` which in this case is Ok because we obviously want all targets to use the exact same standard. 

2- Setting the C++ standard based on features which is a more finer grained approach

target_compile_features(foo PRIVATE|PUBLIC|INTERFACE cxx_std_11)

Now, there reason I went with this approach is because  `mayaUsd_compile_config` already use `target_compile_options` and `target_compile_definitions` and `target_compile_features` seems like a good fit there.

To be clear, every targets **must** add `mayaUsd_compile_config`.  Notice that all the entries for target_compile_options, target_compile_definitions, target_compile_features are declared as **PRIVATE** which is very important.  Any change in `mayaUsd_compile_config` affects **all** targets that are inheriting this function which simulates a global behavior. 

>  It feels like this should be more of a repo-level setting. Do we envision having different parts of the repo on different standards? That would seem a bit weird to me.

Not at all, all targets should have the same standard. 

> And somewhat related to that, which parts of the tree do we expect would not call mayaUsd_compile_config() for their targets?

Every targets **must** add `mayaUsd_compile_config`

plugin/pxr/CMakeLists.txt
HamedSabri-adsk: Not sure if I understand this but this is just a path for the module path.

The language standard are being applies by including `CXXDefaults`

See line 25 in plugin/pxr/CMakeLists.txt 

```
# compiler configuration
# CXXDefaults will set a variety of variables for the project.
# Consume them here. This is an effort to keep the most common
# build files readable.
include(CXXDefaults)

```

plugin/al/CMakeLists.txt
HamedSabri-adsk: I originally didn't want to bring mayaUsd_compile_config() into either AL or Pixar since that is outside the scope of this PR.

Both plugin/pixar and plugins/al targets are not affected by mayaUsd_compile_config intentionally and they are using the same compiler configuration as before.

You are correct, I didn't bring #395 but I believe I should have. :) I will make the change with proper switch between C++11 and C++14.

cmake/compiler_config.cmake
```diff
@@ -0,0 +1,79 @@
...
+
+    # Boost
+    BOOST_ALL_DYN_LINK
+    BOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE
+)
+
+#------------------------------------------------------------------------------
+# compiler configuration
+#------------------------------------------------------------------------------
+# Don not use GNU extension
```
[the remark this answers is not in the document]
HamedSabri-adsk: Ops! Good catch.

lib/mayaUsd/render/mayaToHydra/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ target_sources(${TARGET_NAME}
 )
 
 # -----------------------------------------------------------------------------
-# compile definitions
+# compiler configuration
 # -----------------------------------------------------------------------------
 target_compile_definitions(${TARGET_NAME}
     PRIVATE
```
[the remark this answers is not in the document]
HamedSabri-adsk: I don't believe I have `$<$<BOOL:${IS_LINUX}>:LINUX>` any where in `mayaUsd_compile_config`?

 I also wouldn't add this to the mayaUsd_compile_config since not all targets really need this definition.

cmake/compiler_config.cmake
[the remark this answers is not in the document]
HamedSabri-adsk: please see 20ea6b76

plugin/al/CMakeLists.txt
HamedSabri-adsk: Please see dc898ce4b2

lib/mayaUsd/render/mayaToHydra/CMakeLists.txt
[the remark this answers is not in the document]
mattyjams: Ahh, sorry, my mistake. Apparently I didn't scroll up far enough. :)

I saw this one:
https://github.com/Autodesk/maya-usd/blob/20ea6b767e710455dc30eeb65b3d89301378766d/lib/mayaUsd/CMakeLists.txt#L51

But failed to notice that was in mayaUsd's CMakeLists.txt, not `mayaUsd_compile_config()`. Ignore me!

cmake/compiler_config.cmake
mattyjams: Right. I guess I was reacting to two things:
- The fact that we're already breaking the "every target must add `mayaUsd_compile_config()`" rule, since the AL and pxr plugin code do not use it. But that's probably a whole other can of beans, and they probably should use it eventually.
- But more generally, we're saying that **every** target **must** add `mayaUsd_compile_config()` so that it picks up (effectively) global settings. It seems a little counter-intuitive to me to require authors to do something in their CMakeLists.txt so that they pick up the settings all targets need?

Anyway, it doesn't bother me that much, and I'm likely missing other context here. Just wanted to raise the question. Thanks for humoring me!

plugin/pxr/CMakeLists.txt
mattyjams: Sorry, I worded that poorly. I meant to say this was the same thing as in `plugin/al/CMakeLists.txt` where this file wouldn't pick up the C++14 language standard and would get the old C++11 one instead. But this is also addressed by https://github.com/Autodesk/maya-usd/commit/dc898ce4b24f4b0b250d8657117937758668adde, thanks!

plugin/pxr/cmake/defaults/CXXDefaults.cmake
```diff
@@ -17,6 +17,15 @@ include(CXXHelpers)
 include(Version)
 include(Options)
 
+# Require C++11/C++14
+if(${MAYA_APP_VERSION} STRLESS_EQUAL "2019")
+    set(CMAKE_CXX_STANDARD 11)
+elseif(${MAYA_APP_VERSION} STRGREATER_EQUAL "2019")
+    set(CMAKE_CXX_STANDARD 14)
+endif()
```
mattyjams, line 25: Unfortunately, I think the version of USD factors into this decision as well. USD 20.05 and later will **require** C++14, so for example I think attempting to build maya-usd for Maya 2018 against USD 20.05 would fail with this?

lib/usd/ui/CMakeLists.txt
```diff
@@ -44,7 +44,10 @@ target_sources(${PROJECT_NAME}
 # -----------------------------------------------------------------------------
 target_compile_definitions(${PROJECT_NAME}
     PRIVATE
+        # These flags are needed for Maya
         $<$<BOOL:${IS_MACOSX}>:OSMac_>
+        $<$<BOOL:${IS_LINUX}>:LINUX>
+
```
HamedSabri-adsk, line 50: I have noticed a lot of targets are adding these two flags privately.  Since these flags are needed for Maya, it makes more sense I make them as part of PUBLIC section in mayaUsd shared library.  I will have a separate PR to address this but for now it should be ok.

lib/mayaUsd/CMakeLists.txt
```diff
@@ -52,6 +51,13 @@ target_compile_definitions(${PROJECT_NAME}
         $<$<BOOL:${IS_LINUX}>:GLX_GLXEXT_PROTOTYPES>
 )
 
+if(DEFINED UFE_PREVIEW_VERSION_NUM)
+    target_compile_definitions(${PROJECT_NAME}
+        PUBLIC
+            UFE_PREVIEW_VERSION_NUM=${UFE_PREVIEW_VERSION_NUM}
+    )
+endif()
```
HamedSabri-adsk, line 59: Fix 2019 build issue on Windows ( Fatal Error C1017 invalid integer constant expression ). 

`UFE_PREVIEW_VERSION_NUM` is a new variable that was added recently and only defined when building against UFE `0.x.x`.

HamedSabri-adsk:
Hi @mattyjams, 

I have updated this PR to respect the recent changes in `dev` branch. Following this, I decided to finish what we started in here and get `plugin/pxr` and `plugin/al` to adopt compiler configuration ( mayaUsd_compile_config) to avoid duplication and future confusions. The work is done in these 2 PRs( PR #443, PR #445 ). 

Union of PR 412, 443, 445 makes this PR complete.  Also `defaults` folders, is no longer in use and I didn't delete it until your confirmation.

doc/build.md
```diff
@@ -110,6 +110,8 @@ BUILD_PXR_PLUGIN            | builds the Pixar USD plugin and libraries.
 BUILD_AL_PLUGIN             | builds the Animal Logic USD plugin and libraries.          | ON
 BUILD_HDMAYA                | builds the Maya-To-Hydra plugin and scene delegate.        | ON
 BUILD_TESTS                 | builds all unit tests.                                     | ON
+BUILD_STRICT_MODE           | enforces all warnings as errors.                           | ON
+BUILD_SHARED_LIBS			| build libraries as shared or static.						 | ON
```
mattyjams, line 114: Funky whitespacing. Tab characters, maybe?

HamedSabri-adsk: Hey @mattyjams , indeed funky whitespacing but one thing I have noticed is that is how github shows it when it wraps charcaters.  In my text editor, I usually have word wrap off when I work with *.md file.

I have also noticed that sometimes github doesn't properly visualize spaces vs tabs.

PR #414, opened 2020-04-09
add pre-emptive includes of sdf/types.h before M3dView.h for Maya 2019

mattyjams:
Following the merge of PRs #388 through #392, the header reorganization created some new sites where `M3dView.h` was being included earlier than `sdf/types.h`.

This change applies an existing pattern of including `sdf/types.h` early to work around this issue and fix compiling for Maya 2019 (and earlier).

### DIFF
--- a/lib/mayaUsd/render/pxrUsdMayaGL/batchRenderer.h
+++ b/lib/mayaUsd/render/pxrUsdMayaGL/batchRenderer.h
@@ -22,6 +22,19 @@
 #include <unordered_set>
 #include <utility>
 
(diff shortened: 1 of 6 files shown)


kxl-adsk:
@mattyjams Thank you and apologies for not catching this before the merge. We validated Maya 2019, but we are not yet ready to validate all supported platforms x all supported versions of Maya x all possible build configurations.

mattyjams:
No problem. My current work machine is Maya 2020, and clearly I didn't test the 2019 configuration either. Hopefully we'll be able to catch this kind of thing with continuous integration in the not-too-distant future.

Thanks for the quick merge @kxl-adsk!

PR #420, opened 2020-04-14
Pixar Mesh write translator/utilities refactoring

HamedSabri-adsk:
### PR Description 
This PR is the continuation of work on refactoring mesh translator/utilities in PR #343.

This PR involves refactoring Pixar's mesh writer class ( PxrUsdTranslators_MeshWriter ) by creating utilities and providing a simpler interface:

- All utilities for writing out mesh data are now housed under `UsdMayaMeshWriteUtils` namespace.
- `UsdUtilsSparseValueWriter` is now directly passed as an optional argument to all utility functions responsible for writing mesh data.
- Decoupled the logics for dealing with writing out joint and skin data. These utilities are now under `UsdMayaJointUtil` namespace.


lib/mayaUsd/utils/util.h
```diff
@@ -556,10 +556,14 @@ MString convert(const TfToken& token);
 MAYAUSD_CORE_PUBLIC
 std::string convert(const MString&);
 
-} // namespace UsdMayaUtil
+MAYAUSD_CORE_PUBLIC
+MDagPath getDagPath(const MFnDependencyNode& depNodeFn, const bool reportError = true);
```
mattyjams, line 560: Thanks for making this a public util function. I'm not sure why we hadn't done that to begin with. :/

It looks like there's a third place where the guts of this function are being used as well:
https://github.com/Autodesk/maya-usd/blob/802de0c946012427e9582054d4b7330f58fe5086/lib/mayaUsd/fileio/writeJobContext.cpp#L479

Can I hassle you to clean up the duplicate code there and call this instead?

HamedSabri-adsk: @mattyjams  Awesome. No hassle at all. Will do it soon.

HamedSabri-adsk: @mattyjams  please see 9c0ade65768d9a9266486b0574da98657a48146e

lib/mayaUsd/fileio/writeJobContext.cpp
```diff
@@ -476,16 +476,10 @@ UsdMayaWriteJobContext::CreatePrimWriter(
...
 
-        MStatus status;
-        const MDagPath dagPath = dagNodeFn.dagPath(&status);
-        if (status != MS::kSuccess || !dagPath.isValid()) {
-            TF_CODING_ERROR(
-                "Invalid MDagPath for MFnDagNode '%s'. Verify that it was "
-                "constructed using an MDagPath.",
-                dagNodeFn.fullPathName().asChar());
+        if (!dagPath.isValid()) {
             return nullptr;
```
mattyjams, line 483: The `try`/`catch` that wraps this whole thing should go away now, since that's handled by `getDagPath()`. The coding error that's in the catch below should go here.

mattyjams: The coding error is still missing.

HamedSabri-adsk: I was wondering where it would go? So the whole catch block should be replace with this? Not sure if I get the logic there.
```
if (writePath.IsEmpty()) {
    TF_CODING_ERROR(
        "No usdPath supplied for DG node '%s'.",
        UsdMayaUtil::GetMayaNodeName(depNodeFn.object()).c_str());
}
```

mattyjams: This function should be able to work with either DAG nodes or DG nodes, and the try/catch that was here previously is what was used to determine that. If you passed in an MFnDependencyNode that represents a DG node, you **must** also pass in an SdfPath. With the current changes, this will now **only** work for DAG nodes, which is incorrect.

`UsdMayaUtil::getDagPath()` is now what tells you whether the node is a DAG node or a DG node based on whether the MDagPath you get back is valid or not. So with that, here's what I think the first part of this function should look like:

```C++
    SdfPath writePath = usdPath;

    const MDagPath dagPath =
        UsdMayaUtil::getDagPath(depNodeFn, /* reportError = */ false);
    if (!dagPath.isValid()) {
        // This must be a DG node. usdPath must be supplied for DG nodes.
        if (writePath.IsEmpty()) {
            TF_CODING_ERROR(
                "No usdPath supplied for DG node '%s'.",
                UsdMayaUtil::GetMayaNodeName(depNodeFn.object()).c_str());
            return nullptr;
        }
    } else {
        if (dagPath.length() == 0u) {
            // This is the world root node. It can't have a prim writer.
            return nullptr;
        }

        if (writePath.IsEmpty()) {
            writePath = ConvertDagToUsdPath(dagPath);
        }

        const MFnDagNode dagNodeFn(dagPath);
        const bool instanced = dagNodeFn.isInstanced(/* indirect = */ false);
        if (mArgs.exportInstances && instanced && !forceUninstance) {
            // Deal with instances -- we use a special internal writer for them.
            return std::make_shared<UsdMaya_InstancedNodeWriter>(
                dagNodeFn,
                writePath,
                *this);
        }
    }

    ...
```

HamedSabri-adsk: @mattyjams  Ahhh..... it definitely makes sense. see b7a202b747ac4f45f7b513e7b1428ba8969f0949

lib/mayaUsd/python/wrapMeshWriteUtils.cpp
```diff
@@ -70,13 +70,16 @@ class DummyScopeClass{};
...
+    // TODO: HS, June 11, 2020 This logics is not accurate. Revisit this later.
+    // There are no test cases around is not being consumed in any tests.
+    /*
     scope s = class_<DummyScopeClass>("MeshWriteUtils", no_init)
 
         .def("GetMeshNormals", &_GetMeshNormals)
             .staticmethod("GetMeshNormals")
 
         ;
+    */
```
mattyjams: @HamedSabri-adsk: Can you please revert this change? True that the Python bindings for this are not used anywhere in maya-usd, but we have internal code that does this (using the old name):

```Python
from pxr.UsdMaya import MeshUtil
(normalsArray, interpolation) = MeshUtil.GetMeshNormals(...)
```

HamedSabri-adsk: Hey @mattyjams  Sure will do. But before that I want to understand a few things:

1-
I can see in the original code that GetMeshNormals is not a static function and yet it in the python binding is declared as `staticmethod`.  I am not an expert in Boost python binding but curious how this works?

https://github.com/PixarAnimationStudios/USD/commit/be80accecb33109fcc40471b4089c3748f50abc1

2- MeshUtil has now renamed MeshWriteUtils. I had to fix it in this commit:

https://github.com/Autodesk/maya-usd/pull/420/commits/a520d00fd70cac0421f5799e23bdbee5230935eb

Does this mean that we should revert these name changes as well?

mattyjams: For #1, we're binding `MeshWriteUtils.GetMeshNormals()` in Python to the `_GetMeshNormals()` function declared just above in `wrapMeshWriteUtils.cpp`, which is indeed declared static, and returns a boost tuple type:
https://github.com/Autodesk/maya-usd/blob/5803a4695f23006e8f8367c8422e32c14601d6d1/lib/mayaUsd/python/wrapMeshWriteUtils.cpp#L42

For #2, we do want the name changes so that files/classes/etc. are all in alignment, so that part should **not** be reverted. We can absorb the name change no problem, we just need a function with the new name to exist so that we can switch to it. :P

HamedSabri-adsk: @mattyjams Gotcha...  see be3c5450b31d787eb4616a084b739b8eabfa41f8

mattyjams: Great, thanks!

PR #421, opened 2020-04-14
PR II: Pixar Mesh write translator/utilities refactoring

HamedSabri-adsk:
### PART II

- Rename meshUtil(`*.h,*.cpp`) to meshReadUtils(`*.h,*.cpp`).
- Follow coding guideline for function name ( lowerCamelCase ).

### DIFF
--- a/lib/mayaUsd/fileio/utils/meshReadUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshReadUtils.cpp
@@ -15,7 +15,7 @@
 //
 // Modifications copyright (C) 2020 Autodesk
 //
-#include "meshUtil.h"
+#include "meshReadUtils.h"
 
 #include <maya/MFloatVector.h>
 #include <maya/MFloatVectorArray.h>
@@ -566,7 +566,7 @@ namespace
 
 // This can be customized for specific pipelines.
 bool
-UsdMayaMeshUtil::GetEmitNormalsTag(const MFnMesh& mesh, bool* value)
+UsdMayaMeshUtil::getEmitNormalsTag(const MFnMesh& mesh, bool* value)
 {
     MPlug plug = mesh.findPlug(MString(_meshTokens->USD_EmitNormals.GetText()));
     if (!plug.isNull()) {
@@ -578,7 +578,7 @@ UsdMayaMeshUtil::GetEmitNormalsTag(const MFnMesh& mesh, bool* value)
 }
 
 void
-UsdMayaMeshUtil::SetEmitNormalsTag(
+UsdMayaMeshUtil::setEmitNormalsTag(
         MFnMesh& meshFn,
         const bool emitNormals)
 {
@@ -596,7 +596,7 @@ UsdMayaMeshUtil::SetEmitNormalsTag(
 }
 
 bool
-UsdMayaMeshUtil::GetMeshNormals(
+UsdMayaMeshUtil::getMeshNormals(
         const MFnMesh& mesh,
         VtArray<GfVec3f>* normalsArray,
         TfToken* interpolation)
@@ -649,7 +649,7 @@ UsdMayaMeshUtil::GetMeshNormals(
 // the RenderMan for Maya int attribute.
 // XXX Maybe we should come up with a OSD centric nomenclature ??
 TfToken
-UsdMayaMeshUtil::GetSubdivScheme(const MFnMesh& mesh)
+UsdMayaMeshUtil::getSubdivScheme(const MFnMesh& mesh)
 {
     // Try grabbing the value via the adaptor first.
     TfToken schemeToken;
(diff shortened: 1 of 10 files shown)


lib/mayaUsd/fileio/utils/meshReadUtils.cpp
```diff
@@ -566,7 +566,7 @@ namespace
 
 // This can be customized for specific pipelines.
 bool
-UsdMayaMeshUtil::GetEmitNormalsTag(const MFnMesh& mesh, bool* value)
+UsdMayaMeshUtil::getEmitNormalsTag(const MFnMesh& mesh, bool* value)
```
mattyjams, line 569: We should probably also change the namespace from `UsdMayaMeshUtil` to `UsdMayaMeshReadUtils`?

lib/mayaUsd/python/wrapMeshUtil.cpp
```diff
@@ -59,7 +59,7 @@ _GetMeshNormals(const std::string& meshDagPath)
         return make_tuple(normalsArray, interpolation);
     }
 
-    UsdMayaMeshUtil::GetMeshNormals(meshObj, &normalsArray, &interpolation);
+    UsdMayaMeshUtil::getMeshNormals(meshObj, &normalsArray, &interpolation);
```
mattyjams, line 62: There's only one function with Python bindings here and it's a write-related utility method, so this file should probably be called `wrapMeshWriteUtils.cpp`

HamedSabri-adsk: see 1e492f25f45430e9b5d94c846baf5fc444e37519

lib/mayaUsd/fileio/utils/meshReadUtils.cpp
HamedSabri-adsk: see 1e492f25f45430e9b5d94c846baf5fc444e37519

HamedSabri-adsk:
> In the future, it would probably be better to collapse all of these separate PRs into a single PR with multiple commits so that it's easy to see the incremental progression commit by commit, but the final collapsed result is still easy to see. This also ensures that all of the discussion is kept in one place. This becomes particularly important in cases like this where the same code and files are touched in multiple commits.

The main reason for making multiple PRs was because sometimes there are so many comments and conversation in one PR that makes it super easy for one to get lost in there. My main goal was to make the PRs **short** and **compilable** and tried to communicate it in PR description that certain changes will be address in other PRs. 

I am not a big fan of Github review system and in my workflow always review PRs locally in my preferred Git GUI when the changes are more than 3-4 files.

Sadly I don't think this would change anytime soon and doesn't seem we can use better code review systems ( e.g code collaborator ) in this public project.

lib/mayaUsd/python/wrapMeshWriteUtils.cpp
```diff
@@ -59,12 +59,12 @@ _GetMeshNormals(const std::string& meshDagPath)
...
-    UsdMayaMeshUtil::GetMeshNormals(meshObj, &normalsArray, &interpolation);
+    UsdMayaMeshReadUtils::getMeshNormals(meshObj, &normalsArray, &interpolation);
 
     return make_tuple(normalsArray, interpolation);
 }
 
-// Dummy class for putting UsdMayaMeshUtil namespace functions in a Python
+// Dummy class for putting UsdMayaMeshReadUtils namespace functions in a Python
 // MeshUtil namespace.
 class DummyScopeClass{};
```
mattyjams, line 70: Ideally, the `wrapMeshUtil()` function below would be renamed to `wrapMeshWriteUtils()` as well to match the file name, which would also require a change in the module.cpp.

mattyjams:
> The main reason for making multiple PRs was because sometimes there are so many comments and conversation in one PR that makes it super easy for one to get lost in there. My main goal was to make the PRs **short** and **compilable** and tried to communicate it in PR description that certain changes will be address in other PRs.
> 
> I am not a big fan of Github review system and in my workflow always review PRs locally in my preferred Git GUI when the changes are more than 3-4 files.
> 
> Sadly I don't think this would change anytime soon and doesn't seem we can use better code review systems ( e.g code collaborator ) in this public project.

If there are going to be many comments and conversations, I would greatly prefer that they be collected in a single PR rather than spread across many PRs. And when the volume of change grows this large, there's likely to be conversation. The best way to control that is to keep PRs as small and focused as possible, and to organize the commits in each PR so that the steps needed to achieve the end result are clear.

When the same file is changed across a series of PRs, reviewers end up wasting time reviewing code in an earlier PR that may be completely different or totally irrelevant in a later PR.

GitHub's review system was clearly not intended to handle hundreds of files, so again, the best course is to keep PRs small. Reviewing your own changes before posting them is good practice, but effectively requiring your reviewers to do that also because a PR is so large puts additional burden on them. It also slows the turnaround time down for you the more work reviewers need to do.

lib/mayaUsd/python/wrapMeshWriteUtils.cpp
HamedSabri-adsk: @mattyjams  see 45f0f518a5ca86932203fc092ac53084eaaf00d7

PR #422, opened 2020-04-14
PART III: Pixar Mesh write translator/utilities refactoring 

HamedSabri-adsk:
### PART III

- Initial commit to introduce `TranslatorMeshWrite` class ( this class doesn't do much work at the moment )

- Created `meshWriteUtils` that houses utility functions operating on writing mesh data.

- Moved below utility functions from `meshReadUtils` to `meshWriteUtils` since they operate on write.

```
  GetMeshNormals()
  GetSubdivScheme()
  GetSubdivInterpBoundary()
  GetSubdivFVLinearInterpolation()
```

- Created two utility functions based on `isMeshValid()` and `_exportReferenceMesh()`


lib/mayaUsd/fileio/utils/meshWriteUtils.h
```diff
@@ -0,0 +1,76 @@
...
+#include <pxr/base/vt/array.h>
+#include <pxr/usd/usd/attribute.h>
+#include <pxr/usd/usdGeom/mesh.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+class UsdGeomMesh;
+
+// Utilities for dealing with writing USD from Maya mesh/subdiv tags.
+namespace UsdMayaMeshUtil
```
mattyjams, line 41: `UsdMayaMeshWriteUtils` namespace?

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+    const float* mayaRawPoints = referenceMesh.getRawPoints(&status);
+    const int numVertices = referenceMesh.numVertices();
+    VtArray<GfVec3f> points(numVertices);
+    for (int i = 0; i < numVertices; ++i) {
+        const int floatIndex = i * 3;
+        points[i].Set(mayaRawPoints[floatIndex],
+                        mayaRawPoints[floatIndex + 1],
+                        mayaRawPoints[floatIndex + 2]);
+    }
```
robthebloke, line 366: Just use memcpy. 
```
VtArray<GfVec3f> points(numVertices);
memcpy(points.data(), mayaRawPoints, numVertices * sizeof(float) * 3);
```

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -0,0 +1,380 @@
...
+        if (normalId < 0 ||
+                static_cast<size_t>(normalId) >= mayaNormals.length()) {
+            return false;
+        }
+
+        MFloatVector normal = mayaNormals[normalId];
+        (*normalsArray)[fvi][0] = normal[0];
+        (*normalsArray)[fvi][1] = normal[1];
+        (*normalsArray)[fvi][2] = normal[2];
+    }
```
robthebloke, line 164: You might want to use 
```
        MIntArray normalCounts, normalIndices;
        mesh.getNormalIds(normalCounts, normalIndices);
```
Rather than using the MItMeshFaceVertex  iterator :(

I'll direct you towards this:

https://github.com/Autodesk/maya-usd/blob/1737b11e03a37c599681b4a8802b417f0e34fb1f/plugin/al/usdmayautils/AL/usdmaya/utils/MeshUtils.cpp#L2180

That's the approach we've taken in the AL plugin (which also generates the perVert/perFace interp modes, and also has the option to write the normals out as a primvar,  which we need in a couple of places in our pipeline)

HamedSabri-adsk:
@mattyjams @robthebloke Thank you for great feedbacks. I have been busy with some work and will try to address them in coming days.

lib/mayaUsd/fileio/utils/meshWriteUtils.h
HamedSabri-adsk: @mattyjams see 80f2240f775911165812e65b20ca964e6af92a80

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
HamedSabri-adsk: @robthebloke see 2b1e3043d7984d5a63880395d1f35777cce85a9a

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
HamedSabri-adsk: @robthebloke see 2b1e304
Yes, I have already seen some old codes in Pixar code base that definitely has room for improvement.

PR #423, opened 2020-04-14
PART IV: Pixar Mesh write translator/utilities refactoring 

HamedSabri-adsk:
### PART IV

- Remove `meshWriter_Subdiv` and start using `UsdMayaMeshUtil::assignSubDivTagsToUSDPrim`

- Created 2 new template functions for writing attribute values. These functions now take UsdUtilsSparseValueWriter by reference and eventually would replace `_SetAttribute` template functions in UsdMayaPrimWriter.

Note: Pixar uses UsdUtilsSparseValueWriter utility class that manages sparse authoring of a set of UsdAttributes. UsdUtilsSparseAttrValueWriter is A utility class for authoring time-varying attribute values with simple run-length encoding, by skipping any redundant time-samples.


lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -377,4 +405,76 @@ UsdMayaMeshUtil::exportReferenceMesh(UsdGeomMesh& primSchema, MObject obj)
...
+    if (!TF_VERIFY(mayaCreaseVertIds.length() == mayaCreaseVertValues.length())) {
+        return;
+    }
+    if (mayaCreaseVertIds.length() > 0u) {
+        VtIntArray subdCornerIndices(mayaCreaseVertIds.length());
+        VtFloatArray subdCornerSharpnesses(mayaCreaseVertIds.length());
+        for (unsigned int i = 0u; i < mayaCreaseVertIds.length(); ++i) {
+            subdCornerIndices[i] = mayaCreaseVertIds[i];
+            subdCornerSharpnesses[i] = mayaCreaseVertValues[i];
+        }
```
robthebloke, line 426: Add a little inline utility method.... 
```c++
inline void doubleToFloat(float* const f, const double* const d, size_t c)
{
  for(size_t i = 0; i < c; ++i)
    f[i] = d[i];
}
```
and refactor to:
```c++
        VtIntArray subdCornerIndices(mayaCreaseVertIds.length());
        VtFloatArray subdCornerSharpnesses(mayaCreaseVertIds.length());
        memcpy(subdCornerIndices.data(), &mayaCreaseVertIds[0], mayaCreaseVertIds.length() * sizeof(int));
        doubleToFloat(subdCornerSharpnesses.data(), &mayaCreaseVertValues[0], mayaCreaseVertValues.length());
```
As for the reason why, take a look here:
https://godbolt.org/z/etT0v_

The above loop is effectively doing the same as copyA. If you look at the ASM, you'll noticce that all of the double->float conversions are being done via vcvtsd2ss (i.e. 1 value at a time). If however you use a dumb inline function (i.e. copyB), you'll notice that the generated ASM uses vcvtpd2ps (convert 4 at a time), no longer needs to make any function calls within the loop, and should make this method a wee bit faster allround...

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -377,4 +405,76 @@ UsdMayaMeshUtil::exportReferenceMesh(UsdGeomMesh& primSchema, MObject obj)
...
+    }
+    if (mayaCreaseEdgeIds.length() > 0u) {
+        std::vector<int> subdCreaseIndices(mayaCreaseEdgeIds.length() * 2);
+        std::vector<float> subdCreaseSharpnesses(mayaCreaseEdgeIds.length());
+        for (unsigned int i = 0u; i < mayaCreaseEdgeIds.length(); ++i) {
+            meshFn.getEdgeVertices(mayaCreaseEdgeIds[i], edgeVerts);
+            subdCreaseIndices[i * 2] = edgeVerts[0];
+            subdCreaseIndices[i * 2 + 1] = edgeVerts[1];
+            subdCreaseSharpnesses[i] = mayaCreaseEdgeValues[i];
+        }
```
robthebloke, line 451: It might be slightly nicer to do:
```c++
        std::vector<int> subdCreaseIndices(mayaCreaseEdgeIds.length() * 2);
        // just construct directly from the array data
        // by moving this out of the loop, you'll leverage simd ops here.
        std::vector<float> subdCreaseSharpnesses(&mayaCreaseEdgeIds[0], &mayaCreaseEdgeIds[0] + mayaCreaseEdgeIds.length());
        // avoid dso call by taking a copy of length. 
        for (unsigned int i = 0u, n = mayaCreaseEdgeIds.length(); i < n; ++i) {
            meshFn.getEdgeVertices(mayaCreaseEdgeIds[i], edgeVerts);
            subdCreaseIndices[i * 2] = edgeVerts[0];
            subdCreaseIndices[i * 2 + 1] = edgeVerts[1];
        }
```

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -377,4 +405,76 @@ UsdMayaMeshUtil::exportReferenceMesh(UsdGeomMesh& primSchema, MObject obj)
...
+
+        std::vector<int> numCreases;
+        std::vector<int> creases;
+        std::vector<float> creaseSharpnesses;
+        compressCreases(subdCreaseIndices, subdCreaseSharpnesses,
+                &numCreases, &creases, &creaseSharpnesses);
```
robthebloke, line 457: I assume compressCreases is inline? (or an anon method without any DSO export linkeage). 

If this results in an actual function call, none of the arrays you're passing in will ever be able to vectorise. If it's inline, you'll be good.

HamedSabri-adsk: @robthebloke I am not sure if I understand the connection between inline and vectorization?

From what I know, inline has nothing to do with vectorization and I quote:

https://en.cppreference.com/w/cpp/language/inline
> The original intent of the inline keyword was to serve as an indicator to the optimizer that inline substitution of a function is preferred over function call, that is, instead of executing the function call CPU instruction to transfer control to the function body, a copy of the function body is executed without generating the call. This avoids overhead created by the function call (passing the arguments and retrieving the result) but it may result in a larger executable as the code for the function has to be repeated multiple times.
> Since this meaning of the keyword inline is non-binding, compilers are free to use inline substitution for any function that's not marked inline, and are free to generate function calls to any function marked inline. Those optimization choices do not change the rules regarding multiple definitions and shared statics listed above.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
HamedSabri-adsk: @robthebloke Thanks. See 8e3a0accd75df424493e2911dcb5f0a10d629092

Just so I understand, bypassing the call per iteration is mainly because the compiler has no visibility into Maya's array structure so it cannot optimize it as well as your loop? Is that what you mean by `avoid dso call`?

lib/mayaUsd/fileio/utils/writeUtil.h
```diff
@@ -301,6 +302,34 @@ struct UsdMayaWriteUtil
...
+    /// leaving it in default-constructed state (value-initialized).
+    template <typename T>
+    static bool SetAttribute( const UsdAttribute& attr,
+                              T* value,
+                              UsdUtilsSparseValueWriter& valueWriter,
+                              const UsdTimeCode time = UsdTimeCode::Default()) 
+    {
+        return valueWriter.SetAttribute(attr, VtValue::Take(*value), time);
+    }
+
```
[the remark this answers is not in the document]
HamedSabri-adsk: @mattyjams you have probably seen by now but _SetAttribute eventually is removed:

https://github.com/Autodesk/maya-usd/pull/428#pullrequestreview-403932691

I didn't remove it in this PR to avoid making lots of additional changes.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
HamedSabri-adsk: @robthebloke  Good point and thanks for the godbolt example. 

Can I ask you if we can make this change in another PR? the reason I am asking is because the suggested `doubleToFloat`  utility functions is already part of `plugin\al\usdmayautils\AL\usdmaya\utils\MeshUtils.cpp`  and I already have a plan to bring some of these utilities to the core library so others can use them.  This is certainly a work for another PR.

HamedSabri-adsk:
@mattyjams @robthebloke  I believe I have addressed all the feedbacks in PR IV and V and don't see any major concerns here to stop things from moving forward. I am going to merge these 2 PRs to their target branch. I can always make any necessary changes before doing the final merge to dev.

PR #424, opened 2020-04-14
PART V: Pixar Mesh write translator/utilities refactoring 

HamedSabri-adsk:
### PART V
Created 3 utility functions for writing Maya mesh data to USD:

- vertex positions
- face/vertex indices
- invisible faces

Removed the "TODO: Use memcpy()" and used memcpy to copy the data directly.

### DIFF
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
@@ -483,22 +483,21 @@ UsdMayaMeshWriteUtils::writePointsData(const MFnMesh& meshFn,
     const uint32_t numVertices = meshFn.numVertices();
     VtVec3fArray points(numVertices);
     const float* pointsData = meshFn.getRawPoints(&status);
-    if(status)
-    {
-        // use memcpy() to copy the data. HS April 09, 2020
-        memcpy((GfVec3f*)points.data(), pointsData, sizeof(float) * 3 * numVertices);
 
-        VtVec3fArray extent(2);
-        // Compute the extent using the raw points
-        UsdGeomPointBased::ComputeExtent(points, &extent);
-
-        UsdMayaWriteUtil::SetAttribute(primSchema.GetPointsAttr(), &points, valueWriter, usdTime);
-        UsdMayaWriteUtil::SetAttribute(primSchema.CreateExtentAttr(), &extent, valueWriter,usdTime);
-    }
-    else
-    {
+    if(!status) {
         MGlobal::displayError(MString("Unable to access mesh vertices on mesh: ") + meshFn.fullPathName());
+        return;
     }
+    
+    // use memcpy() to copy the data. HS April 09, 2020
+    memcpy((GfVec3f*)points.data(), pointsData, sizeof(float) * 3 * numVertices);
+
+    VtVec3fArray extent(2);
+    // Compute the extent using the raw points
+    UsdGeomPointBased::ComputeExtent(points, &extent);
+
+    UsdMayaWriteUtil::SetAttribute(primSchema.GetPointsAttr(), &points, valueWriter, usdTime);
+    UsdMayaWriteUtil::SetAttribute(primSchema.CreateExtentAttr(), &extent, valueWriter,usdTime);
 }
 
 void 
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
@@ -36,6 +36,7 @@
 #include <pxr/base/tf/token.h>
 #include <pxr/base/vt/array.h>
 #include <pxr/usd/usdGeom/mesh.h>
+#include <pxr/usd/usdGeom/pointBased.h>
 #include <pxr/usd/usdGeom/primvar.h>
 #include <pxr/usd/usdGeom/tokens.h>
 #include <pxr/usd/usdUtils/pipeline.h>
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/fileio/utils/meshWriteUtils.h
```diff
@@ -73,6 +73,15 @@ namespace UsdMayaMeshUtil
     MAYAUSD_CORE_PUBLIC
     void assignSubDivTagsToUSDPrim(MFnMesh&, UsdGeomMesh&, UsdUtilsSparseValueWriter&);
 
+    MAYAUSD_CORE_PUBLIC
+    void writeVertexData(const MFnMesh&, UsdGeomMesh&, const UsdTimeCode&, UsdUtilsSparseValueWriter&);
```
mattyjams, line 77: Can we all this `writePointsData()` for consistency with the USD naming that we're actually authoring?

lib/usd/translators/meshWriter.cpp
```diff
@@ -160,44 +160,12 @@ PxrUsdTranslators_MeshWriter::writeMeshAttrs(
...
-            curFaceVertexIndex++;
-        }
-    }
-    _SetAttribute(primSchema.GetFaceVertexCountsAttr(), &faceVertexCounts, usdTime);
-    _SetAttribute(primSchema.GetFaceVertexIndicesAttr(), &faceVertexIndices, usdTime);
+    // Write faceVertexIndices
+    UsdMayaMeshUtil::writeFaceVertexIndicesData(geomMesh, primSchema, usdTime, *_GetSparseValueWriter());
 
     // Read subdiv scheme tagging. If not set, we default to defaultMeshScheme
     // flag (this is specified by the job args but defaults to catmullClark).
```
mattyjams, line 171: `writeInvisibleFacesData()` came from the block down below that handles holes. Shouldn't we be replacing that block with the new function?

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -477,4 +478,76 @@ UsdMayaMeshUtil::assignSubDivTagsToUSDPrim(MFnMesh& meshFn,
...
+void
+UsdMayaMeshUtil::writeVertexData(const MFnMesh& meshFn,
+                                 UsdGeomMesh& primSchema,
+                                 const UsdTimeCode& usdTime,
+                                 UsdUtilsSparseValueWriter& valueWriter)
+{
+    MStatus status{MS::kSuccess};
+
+    const uint32_t numVertices = meshFn.numVertices();
+    VtArray<GfVec3f> points(numVertices);
```
mattyjams, line 490: I know the original code used the templated VtArray, but it's slightly preferred to use the named types for these:
`VtVec3fArray`
`VtIntArray`
etc.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -477,4 +478,76 @@ UsdMayaMeshUtil::assignSubDivTagsToUSDPrim(MFnMesh& meshFn,
...
+    const float* pointsData = meshFn.getRawPoints(&status);
+    if(status)
+    {
```
mattyjams, line 493: It'd be a little bit cleaner to invert this condition and display the error and return if there's a problem. That way we can save an indent level for the code currently inside this block.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -477,4 +478,76 @@ UsdMayaMeshUtil::assignSubDivTagsToUSDPrim(MFnMesh& meshFn,
...
 
+void
+UsdMayaMeshUtil::writeVertexData(const MFnMesh& meshFn,
+                                 UsdGeomMesh& primSchema,
+                                 const UsdTimeCode& usdTime,
+                                 UsdUtilsSparseValueWriter& valueWriter)
+{
+    MStatus status{MS::kSuccess};
+
+    const uint32_t numVertices = meshFn.numVertices();
```
mattyjams, line 489: Looks like the return type here should is just `int`?

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -477,4 +478,76 @@ UsdMayaMeshUtil::assignSubDivTagsToUSDPrim(MFnMesh& meshFn,
...
+    UsdMayaWriteUtil::SetAttribute(primSchema.GetFaceVertexIndicesAttr(), &faceVertexIndices, valueWriter, usdTime);
+}
+
+void 
+UsdMayaMeshUtil::writeInvisibleFacesData(const MFnMesh& meshFn, 
+                                         UsdGeomMesh& primSchema, 
+                                         UsdUtilsSparseValueWriter& valueWriter)
+{
+    MUintArray mayaHoles = meshFn.getInvisibleFaces();
+    const uint32_t count = mayaHoles.length();
```
mattyjams, line 541: Just `unsigned int` here?

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -477,4 +478,76 @@ UsdMayaMeshUtil::assignSubDivTagsToUSDPrim(MFnMesh& meshFn,
...
+    if (count)
+    {
+        VtArray<int32_t> subdHoles(count);
```
mattyjams, line 544: `VtIntArray` here, since that's the type in USD.

lib/mayaUsd/fileio/utils/meshWriteUtils.h
HamedSabri-adsk: @mattyjams see a02159ccad0f0b0165f0214f1710f7680e67561e

lib/usd/translators/meshWriter.cpp
HamedSabri-adsk: This is my bad here. As you probably have seen, this change happens in the next PR. I am not really sure why I did it this way originally and I admit it is confusing. :(

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
HamedSabri-adsk: see 4ca16c08570949b8c0d73025d389139b03edbf30

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
HamedSabri-adsk: see ab0a321451a2c0d8be2d4c5237713e9a86e9eb61

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
HamedSabri-adsk: Correct, the return type of numVertices() is `int`. Both `int `and `uint32_t `are `4 bytes` in size so we should be Ok here.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
HamedSabri-adsk: I think uint32_t should be Ok here. Any particular concern why this needs to change here?

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
HamedSabri-adsk: see 4ca16c08570949b8c0d73025d389139b03edbf30

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
mattyjams: Right, the size is the same, but the semantics between the two types are not. I'd argue that `numVertices()` **should** return an unsigned type, but it doesn't. I usually prefer to store the returned value using the type that matches exactly, then do some kind of explicit casting if necessary, rather than introduce an implicit conversion.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
mattyjams: Same argument as above.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
HamedSabri-adsk: I am on the same boat as you on this.

PR #426, opened 2020-04-14
PART VII: Pixar Mesh write translator/utilities refactoring 

HamedSabri-adsk:
### PART VII

Removed `meshWriter_Skin` and created a new utility namespace `UsdMayaJointUtil` that houses functions for writing out joint and skin data. 

This work had to be done since writing skinning data happens during writeMeshAttrs call. There are more work needs to be done to make these utilities reusable in other places but this is outside the scope of this PR.


lib/usd/translators/CMakeLists.txt
```diff
@@ -33,18 +34,28 @@ target_sources(${TARGET_NAME}
...
+    jointWriteUtils.h
+)
+
+# -----------------------------------------------------------------------------
+# promoted headers
+# -----------------------------------------------------------------------------
+mayaUsd_promoteHeaderList(HEADERS ${headers}
+                          BASEDIR ${TARGET_NAME}
+)
+
```
mattyjams, line 58: If the intention is to move all the joint and skin related stuff into a reusable library, shouldn't this all go in `lib/mayaUsd/fileio/utils` alongside all the other read/write utils? It doesn't seem appropriate for this translators library to be the thing that provides them, especially so since it doesn't currently provide **any** headers at all.

lib/usd/translators/jointWriteUtils.h
```diff
@@ -0,0 +1,172 @@
...
+#include <pxr/usd/usdUtils/sparseValueWriter.h>
+#include <pxr/usd/usdSkel/bindingAPI.h>
+#include <pxr/usd/usdSkel/root.h>
+#include <pxr/usd/usdSkel/utils.h>
+#include <pxr/usd/usdSkel/skeleton.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+// Utilities for dealing with writing out joint and skin data.
+namespace UsdMayaJointUtil
```
mattyjams, line 45: The namespace should ideally align with the file name, so `UsdMayaJointWriteUtils` in this case.

lib/usd/translators/meshWriter.cpp
```diff
@@ -41,6 +41,7 @@
 #include <pxr/usd/usdGeom/pointBased.h>
 #include <pxr/usd/usdGeom/primvar.h>
 #include <pxr/usd/usdUtils/pipeline.h>
+#include <pxr/usd/usdSkel/root.h>
```
mattyjams, line 44: Move before `usdUtils/pipeline.h`.

lib/usd/translators/meshWriter.cpp
```diff
@@ -120,7 +129,36 @@ PxrUsdTranslators_MeshWriter::writeMeshAttrs(
...
+                                                                *_GetSparseValueWriter());
+
+            // Add all skel primvars to the exclude set.
+            // We don't want later processing to stomp on any of our data.
+            _excludeColorSets.insert({_tokens->skelJointIndices,
+                                      _tokens->skelJointWeights,
+                                      _tokens->skelGeomBindTransform});
+
+            // Mark the bindings for post processing.
+            _writeJobCtx.MarkSkelBindings(primSchema.GetPrim().GetPath(), skelPath, exportSkin);
```
mattyjams, line 160: Looking at the original code, I think this should be skipped if `_skelInputMesh` is a null object coming out of `writeSkinningData()`?

lib/usd/translators/CMakeLists.txt
HamedSabri-adsk: @mattyjams Yes, good point. I can't agree more. not really sure how I missed this.:) 

Please see 2d150798c6f9abc919325d0badf6dd4a185e3381

lib/usd/translators/jointWriteUtils.h
HamedSabri-adsk: Yes, I agree with your point but currently jointWriteUtils has a mixed of read and write functions and for the sake of time I didn't want add an extra work to separate them into to read/write namespaces. This can be broken down to jointWriteUtils [*.h,*.cpp] and jointReadUtils[*.h,*.cpp]
As I mentioned in the PR description, there is room for improvements here but this is something for another PR.

lib/usd/translators/meshWriter.cpp
HamedSabri-adsk: See 2d150798c6f9abc919325d0badf6dd4a185e3381

lib/usd/translators/meshWriter.cpp
HamedSabri-adsk: Correct, good catch. 

In the original version of writeSkinningData, we were doing early returns if objects were not valid and `_excludeColorSets.insert` and `_writeJobCtx.MarkSkelBindings` would get call right at the end. 

See 2d150798c6f9abc919325d0badf6dd4a185e3381

PR #427, opened 2020-04-14
PART VIII: Pixar Mesh write translator/utilities refactoring

HamedSabri-adsk:
### PART VIII

- Removed `meshWriter_Primvars` and created new utilities for writing color sets data.
- Added following utility functions that I missed in `PR #4`.
```
     writeSubdivInterpBound
     writeSubdivFVLinearInterpolation
     writeNormalsData
```


lib/usd/translators/meshWriter.cpp
```diff
@@ -129,21 +115,22 @@ PxrUsdTranslators_MeshWriter::writeMeshAttrs(
...
-            _skelInputMesh = MObject();
-        
+
+            m_skelInputMesh = MObject();
+
         }else {
 
             SdfPath skelPath;
-            _skelInputMesh = UsdMayaJointUtil::writeSkinningData(primSchema,
+            m_skelInputMesh = UsdMayaJointUtil::writeSkinningData(primSchema,
```
mattyjams, line 133: More irrelevant format and naming changes here and below. Those should go in a dedicated commit to avoid clouding the important changes.

lib/usd/translators/meshWriter.cpp
```diff
@@ -326,19 +282,18 @@ PxrUsdTranslators_MeshWriter::writeMeshAttrs(
...
+                                                  isDisplayColor,
+                                                  shadersRGBData,
+                                                  shadersAlphaData,
+                                                  shadersAssignmentIndices,
+                                                  &RGBData,
+                                                  &AlphaData,
+                                                  &interpolation,
+                                                  &assignmentIndices,
+                                                  &colorSetRep,
+                                                  &clamped)) {
```
mattyjams, line 296: clang-format will catch this, but for the purposes of this PR, restrict the change to just the function name, or include the formatting changes in a separate commit. This just makes an already large review even larger unnecessarily.

lib/usd/translators/meshWriter.cpp
```diff
@@ -413,34 +371,116 @@ PxrUsdTranslators_MeshWriter::writeMeshAttrs(
...
     // whether the finalMesh is animated.
-    return _skelInputMesh.isNull() ? _HasAnimCurves() : false;
+    return m_skelInputMesh.isNull() ? _HasAnimCurves() : false;
 }
 
+void
+PxrUsdTranslators_MeshWriter::cleanupPrimvars()
+{
+    if (!isMeshAnimated()) {
+        // Based on how SetPrimvar() works, the cleanup phase doesn't apply to
```
mattyjams, line 407: I think this now actually refers to the `setPrimvar()` function that's in an anonymous namespace in `meshWriteUtils.cpp`? That's an unfortunate hidden coupling.

lib/usd/translators/meshWriter.h
```diff
@@ -44,125 +44,37 @@ PXR_NAMESPACE_OPEN_SCOPE
...
-    void _SetPrimvar(
-            const UsdGeomPrimvar& primvar,
-            const VtIntArray& indices,
-            const VtValue& values,
-            const VtValue& defaultValue,
-            const UsdTimeCode& usdTime);
-
-    /// Cleans up any extra data authored by _SetPrimvar().
-    void _CleanupPrimvars();
+    /// Cleans up any extra data authored by SetPrimvar().
```
mattyjams, line 63: Again, this is actually `setPrimvar()` from `meshWriteUtils.cpp` that's not directly visible by this class.

lib/usd/translators/meshWriter.h
```diff
@@ -44,125 +44,37 @@ PXR_NAMESPACE_OPEN_SCOPE
...
 public:
-    PxrUsdTranslators_MeshWriter(
-            const MFnDependencyNode& depNodeFn,
-            const SdfPath& usdPath,
-            UsdMayaWriteJobContext& jobCtx);
+    PxrUsdTranslators_MeshWriter(const MFnDependencyNode& depNodeFn,
+                                 const SdfPath& usdPath,
+                                 UsdMayaWriteJobContext& jobCtx);
+
+    ~PxrUsdTranslators_MeshWriter() override {};
```
mattyjams, line 54: Not sure why this was added?

lib/mayaUsd/fileio/utils/meshWriteUtils.h
```diff
@@ -82,12 +88,47 @@ namespace UsdMayaMeshUtil
...
+    /// the unauthored/unpainted values in the color set will be filled in using
+    /// the shader values in \p shadersRGBData and \p shadersAlphaData if available.
+    /// Values are gathered per face vertex, but then the data is compressed to
+    /// vertex, uniform, or constant interpolation if possible.
+    /// Unauthored/unpainted values will be given the index -1.
+    MAYAUSD_CORE_PUBLIC
+    bool getMeshColorSetData( MFnMesh&, const MString&, bool, const VtArray<GfVec3f>&, const VtArray<float>&,
+                              const VtArray<int>&, VtArray<GfVec3f>*, VtArray<float>*, TfToken*, VtArray<int>*,
+                              MFnMesh::MColorRepresentation*, bool*);
+
```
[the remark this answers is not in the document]
HamedSabri-adsk: These changes dates back a while when I was just focusing on creating the utility functions without paying attention to argument names in the header. At the time, I was also making changes to reduce the number of passed parameters but eventually abandoned the work because it was a lot more work for the sake of PR size. That's why I probably have missed to put back the parameters name. My bad!

Speaking of parameters, a lot of these utility functions takes more than 6 or 7 arguments. There is even one that takes 12 arguments :). One way would be to group some of these parameters in a structure and pass it around.  Something to think about...

See 35c0c0fd6964f2970a512397546036a73af201df

lib/usd/translators/meshWriter.cpp
HamedSabri-adsk: I agree, probably too late now but I reverted m_skelInputMesh  to _skelInputMesh . Again these changes dates back a while before we decided on guidelines, etc...

See 35c0c0fd6964f2970a512397546036a73af201df

lib/usd/translators/meshWriter.cpp
HamedSabri-adsk: See f7f193bcf49fd73724dc65643429fa556f7e825b

lib/usd/translators/meshWriter.h
HamedSabri-adsk: Sorry this change was accidentally pulled into this PR.

See a7ccad1d0bc086f4074a6832fe5f97f36b65bdb2

PR #431, opened 2020-04-14
Failure to create shader instance when running command-line render

HdC-adsk:
The failure to create shader instance when running command-line render was a regression caused by #130 and a block for us to deploy automated render tests.

When running command-line render, plugin loading happens before VP2 is initialized. Because fragment registration relies on the access to the fragment manager which is created during VP2 initialization, we will have to delay the fragment registration until when it is needed instead of doing it at plugin loading.

PS: The last time I tested #130 and the command-line render worked because there was some other plugin forcing VP2 initialization to be done earlier. We don't want to force initialization of VP2 during plugin load, for compatibility with headless config used by other partners.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/render_delegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/render_delegate.cpp
@@ -31,6 +31,8 @@
 #include <pxr/imaging/hd/rprim.h>
 #include <pxr/imaging/hd/tokens.h>
 
+#include <mayaUsd/render/vp2ShaderFragments/shaderFragments.h>
+
 #include "basisCurves.h"
 #include "bboxGeom.h"
 #include "instancer.h"
@@ -125,9 +127,6 @@ namespace
         */
         void Initialize()
         {
-            if (_isInitialized)
-                return;
-
             MHWRender::MRenderer* renderer = MHWRender::MRenderer::theRenderer();
             const MHWRender::MShaderManager* shaderMgr =
                 renderer ? renderer->getShaderManager() : nullptr;
@@ -154,8 +153,6 @@ namespace
                 _3dFatPointShader->setParameter(_solidColorParameterName, white);
                 _3dFatPointShader->setParameter(_pointSizeParameterName, size);
             }
-
-            _isInitialized = true;
         }
 
         /*! \brief  Returns a fallback CPV shader instance when no material is bound.
@@ -284,8 +281,6 @@ namespace
         }
 
     private:
-        bool                    _isInitialized { false };  //!< Whether the shader cache is initialized
-
         //! Shader registry used by fallback shaders
         MShaderMap              _fallbackShaders[static_cast<size_t>(FallbackShaderType::kCount)];
         MShaderMap              _3dSolidShaders;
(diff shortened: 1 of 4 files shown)


HdC-adsk:
@mattyjams Could you test the branch on your headless config you mentioned in #130?

lib/mayaUsd/render/vp2RenderDelegate/render_delegate.cpp
```diff
@@ -361,7 +354,25 @@ namespace
     MSamplerStateCache sSamplerStates;  //!< Sampler state cache
     tbb::spin_rw_mutex sSamplerRWMutex; //!< Synchronization used to protect concurrent read from serial writes
 
+    MShaderCache sShaderCache;                      //!< Global shader cache to minimize the number of unique shaders.
     const HdVP2BBoxGeom* sSharedBBoxGeom = nullptr; //!< Shared geometry for all Rprims to display bounding box
+    std::once_flag sInitializeVP2ResourcesCalled;   //!< Whether or not InitializeVP2Resources() has been called.
+
+    void InitializeVP2Resources()
+    {
+        // Shader fragments can only be registered until VP2 has been initialized, thus the function
```
mattyjams, line 363: I think this should maybe be: "Shader fragments can only be registered **after** VP2 has been initialized"? As is, it sounds like you have to register fragments before, which is the opposite.

lib/mayaUsd/render/vp2RenderDelegate/render_delegate.cpp
```diff
@@ -361,7 +354,25 @@ namespace
...
+        // cannot be called when loading plugin (which can happen before VP2 initialization for
+        // command-line render). The fragments will be deregistered when plugin is unloaded.
+        HdVP2ShaderFragments::registerFragments();
+
+        // The shader cache should be initialized after registering shader fragments.
+        sShaderCache.Initialize();
+
+        // HdVP2BBoxGeom can only be instantiated from main thread until VP2 has been initialized.
```
mattyjams, line 371: Similar "until" -> "after" here?

lib/mayaUsd/render/vp2ShaderFragments/shaderFragments.cpp
```diff
@@ -130,22 +130,24 @@ namespace
     }
 }
 
-// Fragment registration
+bool HdVP2ShaderFragments::_registered = false;
+
+// Fragment registration should be done until VP2 has been initialized, to avoid any errors from
```
mattyjams, line 135: Similar "until" -> "after" here?

lib/mayaUsd/render/vp2ShaderFragments/shaderFragments.cpp
```diff
@@ -237,20 +239,18 @@ MStatus HdVP2ShaderFragments::registerFragments()
...
     }
 
     MHWRender::MFragmentManager* fragmentManager =
         theRenderer->getFragmentManager();
     if (!fragmentManager) {
-        return MS::kSuccess;
+        return MS::kFailure;
     }
 
     // De-register all fragment graphs.
```
mattyjams, line 256: Does `_registered` need to be set back to `false` somewhere at the bottom of this function (or in any of the error conditions?) in case the plugin is loaded, unloaded, and then reloaded again?

HdC-adsk:
@mattyjams Those are all good points. Thanks for rewording suggestions as well.

Did you get a chance to run your headless tests?

HdC-adsk:
Great to hear that. Some other maya customers can still have headless config for testing, so it is still useful to keep compatibility at that end.

Thanks for the extra effort to test it.

HdC-adsk:
@kxl-adsk Please merge when you see fit.

PR #432, opened 2020-04-15
One converter with examples

kxl-adsk:
Converted this now to actual PR. 

I will log separately an improvement to enhance this class to support "double-conversions", like translateMayaDoubleToUsdSinglePrecision. We will have to explore how it should be done. 

### Background
Translating data types between Maya and USD is implemented in each of our plugins, yet neither has a reusable interface.

When looking into providing a solution for mixed data parenting, animation, and constraints we realized that we will need a new and much more performant interface. Requirements:

- avoid unnecessary value translations via VtValue (e.g. avoid type-erased access to `Get` data on UsdAttribute)
- allow fetching the value to the same container
- allow caching of data converters to avoid expensive searches at runtime
- type checking should be performed when retrieving data converter, no extra penalty at runtime should be necessary
- one reusable set of interfaces to access converters

The list of currently implemented data converters you can find https://github.com/Autodesk/maya-usd/blob/kxl_adsk/one_converter_with_examples/lib/mayaUsd/utils/converter.cpp#L34. The goal was to cover all the possible cases, ranging from a simple data type, string, vectors, colors with gamma support, matrix, array types, and array attributes. 

In d26743f2c38ad9 you can find examples of how we can consolidate value conversion code, but it's removed from final PR.

### DIFF
--- a/lib/mayaUsd/fileio/shading/shadingModePxrRis.cpp
+++ b/lib/mayaUsd/fileio/shading/shadingModePxrRis.cpp
@@ -54,8 +54,8 @@
 #include <mayaUsd/fileio/translators/translatorUtil.h>
 #include <mayaUsd/fileio/utils/roundTripUtil.h>
 #include <mayaUsd/fileio/utils/writeUtil.h>
-#include <mayaUsd/utils/util.h>
 #include <mayaUsd/utils/converter.h>
+#include <mayaUsd/utils/util.h>
 
 using namespace MAYAUSD_NS;
 
--- a/lib/mayaUsd/fileio/shading/shadingModePxrRis.cpp
+++ b/lib/mayaUsd/fileio/shading/shadingModePxrRis.cpp
@@ -55,6 +55,9 @@
 #include <mayaUsd/fileio/utils/roundTripUtil.h>
 #include <mayaUsd/fileio/utils/writeUtil.h>
 #include <mayaUsd/utils/util.h>
+#include <mayaUsd/utils/converter.h>
+
+using namespace MAYAUSD_NS;
 
 PXR_NAMESPACE_OPEN_SCOPE
 
@@ -228,7 +231,7 @@ private:
             }
 
             const SdfValueTypeName attrTypeName =
-                UsdMayaWriteUtil::GetUsdTypeName(attrPlug);
+                Converter::getUsdTypeName(attrPlug);
             if (!attrTypeName) {
                 continue;
             }
--- a/lib/mayaUsd/fileio/utils/readUtil.cpp
+++ b/lib/mayaUsd/fileio/utils/readUtil.cpp
@@ -45,7 +45,10 @@
 
 #include <mayaUsd/fileio/utils/adaptor.h>
 #include <mayaUsd/utils/colorSpace.h>
+#include <mayaUsd/utils/converter.h>
 #include <mayaUsd/utils/util.h>
+
+using namespace MAYAUSD_NS;
 
 PXR_NAMESPACE_OPEN_SCOPE
 
@@ -414,6 +417,14 @@ bool UsdMayaReadUtil::SetMayaAttr(
         MPlug& attrPlug,
         const VtValue& newValue)
 {
+    SdfValueTypeName plugValueType = Converter::getUsdTypeName(attrPlug,false);
+    const auto* converter = Converter::find(plugValueType, attrPlug.isArray());
+    if(converter && newValue.GetType() == plugValueType.GetType()) {
+        ConverterArgs args;
+        converter->convert(newValue, attrPlug, args);
+        return true;
+    }
+    
     MDGModifier modifier;
     return SetMayaAttr(attrPlug, newValue, modifier);
 }
--- a/lib/mayaUsd/fileio/utils/readUtil.cpp
+++ b/lib/mayaUsd/fileio/utils/readUtil.cpp
@@ -46,6 +46,9 @@
 #include <mayaUsd/fileio/utils/adaptor.h>
 #include <mayaUsd/utils/colorSpace.h>
 #include <mayaUsd/utils/util.h>
+#include <mayaUsd/utils/converter.h>
+
+using namespace MAYAUSD_NS;
 
 PXR_NAMESPACE_OPEN_SCOPE
 
@@ -72,39 +75,6 @@ UsdMayaReadUtil::FindOrCreateMayaAttr(
     MDGModifier modifier;
     return FindOrCreateMayaAttr(typeName, variability, depNode, attrName,
             attrNiceName, modifier);
-}
-
-// Whether a plug's attribute is a typed attribute with given type.
-static bool
-_HasAttrType(const MPlug& plug, MFnData::Type type)
-{
-    MObject object = plug.attribute();
-    if (!object.hasFn(MFn::kTypedAttribute)) {
-        return false;
-    }
-
-    MFnTypedAttribute attr(object);
-    return attr.attrType() == type;
-}
-
-// Whether a plug's attribute is a numeric attribute with given type.
-static bool
-_HasNumericType(const MPlug& plug, MFnNumericData::Type type)
-{
-    MObject object = plug.attribute();
-    if (!object.hasFn(MFn::kNumericAttribute)) {
-        return false;
-    }
-
-    MFnNumericAttribute attr(object);
-    return attr.unitType() == type;
-}
-
-static bool
-_HasEnumType(const MPlug& plug)
-{
-    MObject object = plug.attribute();
-    return object.hasFn(MFn::kEnumAttribute);
 }
 
 static MObject
@@ -143,7 +113,7 @@ _FindOrCreateMayaTypedAttr(
     }
     else {
         // Found -- verify.
-        if (_HasAttrType(plug, type)) {
+        if (Converter::hasAttrType(plug, type)) {
             return plug.attribute();
         }
         else {
@@ -186,8 +156,8 @@ _FindOrCreateMayaNumericAttr(
     }
     else {
         // Found -- verify.
-        if (_HasNumericType(plug, type)
-            || (type == MFnNumericData::kInt && _HasEnumType(plug)))
+        if (Converter::hasNumericType(plug, type)
+            || (type == MFnNumericData::kInt && Converter::hasEnumType(plug)))
         {
             return plug.attribute();
         }
@@ -426,20 +396,20 @@ bool UsdMayaReadUtil::SetMayaAttr(
     bool ok = false;
     if (newValue.IsHolding<TfToken>()) {
         TfToken token = newValue.Get<TfToken>();
-        if (_HasAttrType(attrPlug, MFnData::kString)) {
+        if (Converter::hasAttrType(attrPlug, MFnData::kString)) {
             modifier.newPlugValueString(attrPlug, token.GetText());
             ok = true;
         }
     }
     else if (newValue.IsHolding<std::string>()) {
-        if (_HasAttrType(attrPlug, MFnData::kString)) {
+        if (Converter::hasAttrType(attrPlug, MFnData::kString)) {
             std::string str = newValue.Get<std::string>();
             modifier.newPlugValueString(attrPlug, str.c_str());
             ok = true;
         }
     }
     else if (newValue.IsHolding<SdfAssetPath>()) {
-        if (_HasAttrType(attrPlug, MFnData::kString)) {
+        if (Converter::hasAttrType(attrPlug, MFnData::kString)) {
             SdfAssetPath assetPath = newValue.Get<SdfAssetPath>();
             modifier.newPlugValueString(attrPlug,
                     assetPath.GetAssetPath().c_str());
@@ -447,7 +417,7 @@ bool UsdMayaReadUtil::SetMayaAttr(
         }
     }
     else if (newValue.IsHolding<GfMatrix4d>()) {
-        if (_HasAttrType(attrPlug, MFnData::kMatrix)) {
+        if (Converter::hasAttrType(attrPlug, MFnData::kMatrix)) {
             GfMatrix4d mat = newValue.Get<GfMatrix4d>();
             MMatrix mayaMat;
             for (size_t i = 0; i < 4; ++i) {
@@ -463,7 +433,7 @@ bool UsdMayaReadUtil::SetMayaAttr(
         }
     }
     else if (newValue.IsHolding<SdfTokenListOp>()) {
-        if (_HasAttrType(attrPlug, MFnData::kStringArray)) {
+        if (Converter::hasAttrType(attrPlug, MFnData::kStringArray)) {
             TfTokenVector result;
             newValue.Get<SdfTokenListOp>().ApplyOperations(&result);
             MStringArray mayaArr;
@@ -478,7 +448,7 @@ bool UsdMayaReadUtil::SetMayaAttr(
         }
     }
     else if (newValue.IsHolding<SdfStringListOp>()) {
-        if (_HasAttrType(attrPlug, MFnData::kStringArray)) {
+        if (Converter::hasAttrType(attrPlug, MFnData::kStringArray)) {
             std::vector<std::string> result;
             newValue.Get<SdfStringListOp>().ApplyOperations(&result);
             MStringArray mayaArr;
@@ -493,7 +463,7 @@ bool UsdMayaReadUtil::SetMayaAttr(
         }
     }
     else if (newValue.IsHolding<VtTokenArray>()) {
-        if (_HasAttrType(attrPlug, MFnData::kStringArray)) {
+        if (Converter::hasAttrType(attrPlug, MFnData::kStringArray)) {
             VtTokenArray arr = newValue.Get<VtTokenArray>();
             MStringArray mayaArr;
             for (const TfToken& tok : arr) {
@@ -507,7 +477,7 @@ bool UsdMayaReadUtil::SetMayaAttr(
         }
     }
     else if (newValue.IsHolding<VtStringArray>()) {
-        if (_HasAttrType(attrPlug, MFnData::kStringArray)) {
+        if (Converter::hasAttrType(attrPlug, MFnData::kStringArray)) {
             VtStringArray arr = newValue.Get<VtStringArray>();
             MStringArray mayaArr;
             for (const std::string& str : arr) {
@@ -521,7 +491,7 @@ bool UsdMayaReadUtil::SetMayaAttr(
         }
     }
     else if (newValue.IsHolding<VtDoubleArray>()) {
-        if (_HasAttrType(attrPlug, MFnData::kDoubleArray)) {
+        if (Converter::hasAttrType(attrPlug, MFnData::kDoubleArray)) {
             VtDoubleArray arr = newValue.Get<VtDoubleArray>();
             MDoubleArray mayaArr(arr.data(), arr.size());
             MFnDoubleArrayData data;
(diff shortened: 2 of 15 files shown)


kxl-adsk:
Time to move this forward. Any thoughts on questions I asked in the description?

plugin/al/usdmayautils/AL/usdmaya/utils/DgNodeHelper.cpp
```diff
@@ -5645,17 +5649,24 @@ MStatus DgNodeHelper::copyDynamicAttributes(MObject node, UsdPrim& prim, Animati
...
+          UsdAttribute usdAttr = prim.CreateAttribute(attributeName, !isArray ? SdfValueTypeNames->Matrix4d : SdfValueTypeNames->Matrix4dArray);
+          const auto* converter = Converter::find(plug, usdAttr);
+          if(converter) {
+            ConverterArgs args;
+            converter->convert(plug, usdAttr, args);
+            usdAttr.SetCustom(true);
+            break;
+          }
+            
           if(!isArray)
```
elrond79, line 5661: Hi Krystian - just to give my questions a bit more concrete footing, this is probably the piece that most confused me.  In this situation, we're ONLY dealing with Matrix4d, and Matrix4dArray types... both of which (according to [your doc](https://github.com/Autodesk/maya-usd/blob/kxl_adsk/one_converter_with_examples/lib/mayaUsd/utils/converter.cpp#L34)), should already be supported.  So I wasn't clear on in which situations `if(converter)` would be expected to return False, or why we would need to keep the old code.

Is it just that you THINK it should handle all the matrix conversions in this particular branch/scope... but you weren't SURE, and that's why you leave the check (and leave the old code in place)? And you wanted AL to confirm that this should work before removing the "old" stuff? Or are there in fact situations where you know/expect `if(converter)` to currently return False?

kxl-adsk: Commit 2e9beee is here to start the conversation about the adoption and make sure we all see the places where we currently do type conversion. This will help us validate the interface I'm proposing.
And commit 2e9beee isn't aiming at proposing the final injection points or strategy for adoption...and this is why I mentioned it won't be returned.

In this particular case, you are commenting on, my thinking was that the final solution shouldn't require this huge `switch`. Your question is valid...and I would as well expect we don't have to fall back to old code.

lib/mayaUsd/utils/converter.h
```diff
@@ -0,0 +1,331 @@
...
+        _handleToVtValue { handleToVtValue }, _vtValueoHandle { vtValueoHandle }
+    {}
+    
+    Converter() = delete;
+    ~Converter() = default;
+    
+    static const Converter* find(const SdfValueTypeName& typeName, bool isArrayPlug);
+    static const Converter* find(const MPlug& plug, const UsdAttribute& attr);
+    
+    const SdfValueTypeName& type() const
```
mattyjams, line 96: Not sure if calling this simply `type()` would confuse anyone, or if a more qualified `usdType()` or something is warranted.

lib/mayaUsd/utils/converter.h
```diff
@@ -0,0 +1,331 @@
...
+    { return _typeName; }
+    
+    bool validate(const MPlug& plug, const UsdAttribute& usdAttr, const bool translateMayaDoubleToUsdSinglePrecision) const
```
mattyjams, line 99: Following the thread from above, maybe this could take a `ConverterArgs` as its last parameter, and use a default-constructed one to make it an optional param?

lib/mayaUsd/utils/converter.h
```diff
@@ -0,0 +1,331 @@
...
+struct TypedConverter<MAYA_Type,USD_Type, typename std::enable_if<std::is_same<MAYA_Type, USD_Type>::value>::type>
+{
+    static void convert(const USD_Type& src, MAYA_Type& dst)
+    {
+        dst = src;
+    }
+};
+
+template <>
+struct TypedConverter<float3,GfVec3f>
```
mattyjams, line 184: Curious to hear your thoughts about how types that can be safely converted implicitly might look here.

For example:
```
src = float3, dst = GfVec3d    # OK
src = double3, dst = GfVec3f   # INVALID, potential data loss
```

test/lib/testMayaUsdConverter.py
```diff
@@ -0,0 +1,221 @@
...
+
+import unittest
+
+class MayaUsdConverterTestCase(unittest.TestCase):
+    """
+    Verify mayaUsd converter.
+    """
+
+    def createStage(self, layerName):
+        layer = Sdf.Layer.CreateAnonymous("Bool")
```
mattyjams, line 36: Weird name for a layer. ;)

lib/mayaUsd/fileio/utils/readUtil.cpp
```diff
@@ -414,6 +417,14 @@ bool UsdMayaReadUtil::SetMayaAttr(
         MPlug& attrPlug,
         const VtValue& newValue)
 {
+    SdfValueTypeName plugValueType = Converter::getUsdTypeName(attrPlug,false);
+    const auto* converter = Converter::find(plugValueType, attrPlug.isArray());
+    if(converter && newValue.GetType() == plugValueType.GetType()) {
```
mattyjams, line 422: How common a pattern do you think this typeName checking is going to be in client code?

It seems a bit cumbersome, so if it's going to be everywhere then I'm wondering whether it might make sense to push it down into the converters themselves, and maybe have `Converter::convert()` return a `bool` value instead of `void`, returning `true` for a successful conversion and `false` otherwise.

lib/mayaUsd/fileio/utils/readUtil.cpp
```diff
@@ -414,6 +417,14 @@ bool UsdMayaReadUtil::SetMayaAttr(
         MPlug& attrPlug,
         const VtValue& newValue)
 {
+    SdfValueTypeName plugValueType = Converter::getUsdTypeName(attrPlug,false);
```
mattyjams, line 420: Similar comment as elsewhere about potentially needing to plumb `translateMayaDoubleToUsdSinglePrecision` through here, or otherwise put it in a `ConverterArgs`.

lib/mayaUsd/utils/converter.h
```diff
@@ -0,0 +1,331 @@
...
+        _attrToModifier(attrToModifier), _plugToVtValue { plugToVtValue },
+        _vtValueToPlug { vtValueToPlug }, _vtValueToModifier(vtValueToModifier),
+        _handleToAttr(handleToAttr), _attrToHandle(attrToHandle),
+        _handleToVtValue { handleToVtValue }, _vtValueoHandle { vtValueoHandle }
+    {}
+    
+    Converter() = delete;
+    ~Converter() = default;
+    
+    static const Converter* find(const SdfValueTypeName& typeName, bool isArrayPlug);
```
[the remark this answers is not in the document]
kxl-adsk: The challenge here is around array attributes in Maya. See https://help.autodesk.com/view/MAYAUL/2020/ENU/?guid=Maya_SDK_MERGED_Dependency_graph_plug_ins_Complex_and_Dynamic_Attributes_html

This parameter is basically telling us if we should be looking for a converter for an array attribute or not.

lib/mayaUsd/utils/converter.h
```diff
@@ -0,0 +1,331 @@
...
+    static const Converter* find(const MPlug& plug, const UsdAttribute& attr);
```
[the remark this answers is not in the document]
kxl-adsk: It's to save on copy. We will get the actual value and store it directly on an attribute.
If you don't write to USD attribute, then we are proposing to use interface with VtValue.

lib/mayaUsd/utils/converter.cpp
```diff
@@ -0,0 +1,1105 @@
...
+    auto findIt = container.find(typeName);
+    if(findIt != container.end())
+        return &findIt->second;
+    else
+        return nullptr;
+}
+
+const Converter* Converter::find(const MPlug& plug, const UsdAttribute& attr)
+{
+    auto valueTypeName = getUsdTypeName(plug, false);
```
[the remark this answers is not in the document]
kxl-adsk: I was actually wondering how to handle this case. Should it be two-step conversion, for example first you convert double3 to GfVec3d...and then GfVec3d to GfVec3f? It's less performant, but will reduce the amount of code we have to write?

lib/mayaUsd/utils/converter.h
```diff
@@ -0,0 +1,331 @@
...
+#include <mayaUsd/base/api.h>
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+MAYAUSD_NS_DEF {
+    
+struct MAYAUSD_CORE_PUBLIC ConverterArgs final
+{
+    UsdTimeCode _timeCode{UsdTimeCode::Default()};
+    bool        _doGammaCorrection{true};
```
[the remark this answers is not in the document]
kxl-adsk: We are getting a bit outside of the initial feedback I was looking for this draft.

Re underscores - this is the convention we voted on and put in codingGuidelines.

Re `UsdTimeCode::EarliestTime` - very valid point, something we should look at globally. For now, let's keep it consistent.

Re `translateMayaDoubleToUsdSinglePrecision` and `ConverterArgs` - you read it very well, it is indeed a way to configure converters.

lib/mayaUsd/utils/converter.h
kxl-adsk: I will take a note of it...thx.

lib/mayaUsd/utils/converter.h
kxl-adsk: Good point!

lib/mayaUsd/utils/converter.h
kxl-adsk: I hope this comment shares my current thinking https://github.com/Autodesk/maya-usd/pull/432#discussion_r425930844

test/lib/testMayaUsdConverter.py
kxl-adsk: So true...I refactored the code to avoid duplication...probably this is how it happened.

lib/mayaUsd/fileio/utils/readUtil.cpp
kxl-adsk: There shouldn't be. IIRC, this is where we double to float conversion...and I wanted to filter it out for now. Clearly we need to decide how we will be dealing with double to float conversion.

kxl-adsk:
My next steps, I will add documentation and convert this Draft to PR.

lib/mayaUsd/utils/converter.h
kxl-adsk: Looking at it again now. We will need a design for this "double" conversion (i.e. one between maya and usd types, second to single precision). It won't happen as part of this PR, so I'm going to leave this method as is.

lib/mayaUsd/utils/converter.h
```diff
@@ -0,0 +1,461 @@
...
+        {
+            return (usdAttr.GetTypeName() == _usdTypeName)
+                && (getUsdTypeName(plug, false) == _usdTypeName);
+        }
+
+        //! \brief  Read current value from given plug and set it on destination attribute. Use
+        //! arguments to control converter behavior, like the time for setting value on attribute.
+        void convert(const MPlug& src, UsdAttribute& dst, const ConverterArgs& args) const
+        {
+            _plugToAttr(src, dst, args);
```
ppt-adsk, line 176: Is it possible for a converter not to exist, and that _plugToAttr would still be a nullptr?  Of course in that case you'll get a segmentation fault, but maybe that can never happen, or it's a good enough behavior to indicate a bug?

lib/mayaUsd/utils/converter.cpp
```diff
@@ -0,0 +1,1188 @@
...
+        } else if (attrObj.hasFn(MFn::kUnitAttribute)) {
+            MFnUnitAttribute unitAttrFn(attrObj);
+            unitDataType = unitAttrFn.unitType();
+        }
+
+        return true;
+    }
+
+    //---------------------------------------------------------------------------------
+    //! \brief  Type tranformation class to optain at compile time array type for a given Maya's
```
ppt-adsk, line 109: "optain" --> "obtain".

ppt-adsk: And a few other places.  Also, "a given Maya's" should be "a given Maya".

elrond79:
Think I'm going to give my official notice that I won't have time to get to this one anytime soon. Looks like we've already got some good feedback, though!

kxl-adsk:
@elrond79 no worries and thank you for letting me know.

lib/mayaUsd/utils/converter.h
kxl-adsk: We never generate converters that don't support all these interfaces. My current thinking is that we shouldn't go there and each converter should support them. Time will put this to the test.

lib/mayaUsd/fileio/shading/shadingModePxrRis.cpp
```diff
@@ -55,6 +55,9 @@
 #include <mayaUsd/fileio/utils/roundTripUtil.h>
 #include <mayaUsd/fileio/utils/writeUtil.h>
 #include <mayaUsd/utils/util.h>
+#include <mayaUsd/utils/converter.h>
```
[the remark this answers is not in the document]
kxl-adsk: see d59e8ac6f313c

lib/mayaUsd/fileio/utils/writeUtil.cpp
```diff
@@ -342,7 +163,7 @@ UsdMayaWriteUtil::GetOrCreateUsdAttr(
     }
 
     const SdfValueTypeName& typeName =
-        UsdMayaWriteUtil::GetUsdTypeName(attrPlug,
+        Converter::getUsdTypeName(attrPlug,
                                             translateMayaDoubleToUsdSinglePrecision);
```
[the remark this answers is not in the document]
kxl-adsk: see d59e8ac6f313c

test/lib/testMayaUsdConverter.py
```diff
@@ -0,0 +1,227 @@
...
+    Verify mayaUsd converter.
+    """
+
+    def createStage(self, layerName):
+        """
+        Create in-memory stage
+        """
+        layer = Sdf.Layer.CreateAnonymous(layerName)
+        stage = Usd.Stage.Open(layer.identifier)
+        self.assertNotEqual(stage, None)
```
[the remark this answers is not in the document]
kxl-adsk: every time I have a break in writing tests...I keep making this silly mistake. THX! see d59e8ac6f313c

test/lib/testMayaUsdConverter.py
```diff
@@ -0,0 +1,227 @@
...
+        converter.convertVt(value1,plug,args)
+        converter.convert(plug, attr, args)
+        
+        resultVt1 = converter.convertVt(plug,args)
+        self.assertEqual(resultVt1, Vt._test_Ident(value1))
+        
+        self.assertEqual(converter.validate(plug,attr), True)
+        self.assertEqual(converter.validate(errPlug,errAttr), False)
+        self.assertEqual(converter.validate(plug,errAttr), False)
+        self.assertEqual(converter.validate(errPlug,attr), False)
```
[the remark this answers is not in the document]
kxl-adsk: see d59e8ac6f313c

lib/mayaUsd/utils/converter.h
```diff
@@ -0,0 +1,461 @@
...
+    /*! \brief  Converter arguments are passed together using this convenient struct.
+
+        \note   Not all arguments will be applicable to all conversions.
+     */
+    struct MAYAUSD_CORE_PUBLIC ConverterArgs final {
+        //! When conversion needs to read / write to time, following argument will be used.
+        UsdTimeCode _timeCode { UsdTimeCode::Default() };
+        //! For types supporting gamma correction, following argument controls if it should be
+        //! applied
+        bool _doGammaCorrection { true };
```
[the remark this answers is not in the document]
kxl-adsk: Having different naming convention for private vs public got no traction when we discussed it internally.

lib/mayaUsd/utils/converter.h
```diff
@@ -0,0 +1,461 @@
...
+    };
+
+    /*! \brief Performant converter, providing several interfaces to translated between Maya and Usd
```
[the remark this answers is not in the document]
kxl-adsk: see d59e8ac6f313c

lib/mayaUsd/utils/converter.h
```diff
@@ -0,0 +1,461 @@
...
+        using UsdAttrToMDataHandleFn
+            = std::add_pointer<void(const UsdAttribute&, MDataHandle&, const ConverterArgs&)>::type;
+        using MDataHandleToVtValueFn
+            = std::add_pointer<void(const MDataHandle&, VtValue&, const ConverterArgs&)>::type;
+        using VtValueToMDataHandleFn
+            = std::add_pointer<void(const VtValue&, MDataHandle&, const ConverterArgs&)>::type;
+
+        //! \brief  Construct new converter. This should never be called directly other than during
+        //! generation of all supported converters.
+        Converter(
```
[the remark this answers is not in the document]
kxl-adsk: I couldn't convince myself that there is a need to do this...but since you thought about it too. See d59e8ac6f313c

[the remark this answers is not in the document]
kxl-adsk: FWIW. You can't friend a method from an anonymous namespace.

lib/mayaUsd/utils/converter.cpp
```diff
@@ -0,0 +1,1188 @@
...
+            createConverter<float3, GfVec3f>(converters, SdfValueTypeNames->Float3);
+            createConverter<double3, GfVec3d>(converters, SdfValueTypeNames->Double3);
+            createConverter<MMatrix, GfMatrix4d>(converters, SdfValueTypeNames->Matrix4d);
+
+            createConverter<float3, GfVec3f, NeedsGammaCorrection::kYes>(
+                converters, SdfValueTypeNames->Color3f);
+            createConverter<double3, GfVec3d, NeedsGammaCorrection::kYes>(
+                converters, SdfValueTypeNames->Color3d);
+
+            createConverter<MIntArray, VtArray<int>>(converters, SdfValueTypeNames->IntArray);
```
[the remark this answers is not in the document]
kxl-adsk: Indeed, it can be added later.

lib/mayaUsd/python/wrapConverter.cpp
```diff
@@ -0,0 +1,155 @@
...
+        .def("find", find1, return_value_policy<reference_existing_object>())
+        .def("find", find2, return_value_policy<reference_existing_object>())
+        .staticmethod("find")
+        .def("validate", validate)
+        .def("convert", convertMPlugToUsdAttr)
+        .def("convert", convertUsdAttrToMPlug)
+        .def("convertVt", convertMPlugToVtValue)
+        .def("convertVt", convertVtValueToMPlug)
+        .def("test_convertAndSetWithModifier", test_convertUsdAttrToMDGModifier)
+        .def("test_convertVtAndSetWithModifier", test_convertVtValueToMDGModifier);
```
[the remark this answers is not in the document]
kxl-adsk: I agree with you, this will be a nice improvement. Funny enough I was inspired by code from VtValue tests. I will consider making this change as a separate PR.

lib/mayaUsd/fileio/utils/writeUtil.cpp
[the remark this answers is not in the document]
mattyjams: There were actually a few instances of this. Looks like all the `Converter::getUsdTypeName()` calls that were added.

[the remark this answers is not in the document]
kxl-adsk: True, missed it - 0b3c4af693d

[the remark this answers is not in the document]
mattyjams: Cool, thanks!

PR #436, opened 2020-04-15
Enable consolidation for instanced render items

HdC-adsk:
The change will enable consolidation for instanced render items. It will take advantage of the MDI consolidation which is only available on OpenGL Core Profile mode in Maya.


cfmoore007:
Our in-house test :
Before - ~36ms per frame time,  27 fps  playback
After   -  ~14ms per frame time,  70 fps playback

PR #437, opened 2020-04-16
MAYA-101454 Clear stage cach on file new/open.

fowlertADSK:
This is a small change that reuses existing code that will reset the stage cache on New or Open, fixing some logged issues.  It is not a complete fix for how we deal with the cache though, e.g. deleting the proxy shape and re-importing the same USD file will still show you the cached version which is probably unexpected.  The listener here is only called on File -> New and Open.

We do plan on looking at how the Animal Logic plugin deals with this kind of thing and will most likely refactor some of that into the core lib to be shared by all plugins, but for right now this is an easy change that should help in a couple obvious workflows.

### DIFF
--- a/lib/mayaUsd/listeners/notice.cpp
+++ b/lib/mayaUsd/listeners/notice.cpp
@@ -24,7 +24,7 @@ PXR_NAMESPACE_OPEN_SCOPE
 
 namespace {
 
-// Siple ref count for how many plugins have initialized the callback.
+// Simple ref count for how many plugins have initialized the callback.
 // Used to keep the listener around until the last plugin asks for it
 // to be removed.
 static int _newOrOpenRegistrationCount = 0;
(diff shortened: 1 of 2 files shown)


plugin/adsk/plugin/plugin.cpp
```diff
@@ -134,6 +135,8 @@ MStatus initializePlugin(MObject obj)
         }
     }
 
+    UsdMayaSceneResetNotice::InstallListener();
```
fowlertADSK, line 138: This was already being installed by the Pixar plugin, but MayaUSD should also be using it.

lib/mayaUsd/listeners/notice.cpp
```diff
@@ -24,6 +24,11 @@ PXR_NAMESPACE_OPEN_SCOPE
 
 namespace {
 
+// Siple ref count for how many plugins have initialized the callback.
```
BigRoy, line 27: Siple or simple? :)

fowlertADSK: either :-)

PR #438, opened 2020-04-16
upgrade maya-usd project to the C++14 standard

mattyjams:
Re-opening this from the original #395 pull request...

Conforming to VFX Reference Platform CY2018+ and following suit from core USD, this change sets the maya-usd project to use the C++14 standard if we're either building for Maya 2019 or later, or if we're building against USD 20.05 or later.

Core USD moved to C++14 with release 20.05 in this commit:
https://github.com/PixarAnimationStudios/USD/commit/afc0d5e3d45eeaf6579b9699250da37cd9149c6f

### DIFF
--- a/cmake/defaults/CXXDefaults.cmake
+++ b/cmake/defaults/CXXDefaults.cmake
@@ -20,8 +20,7 @@ include(Options)
 # Require C++14 if we're either building for Maya 2019 or later, or if we're
 # building against USD 20.05 or later. Otherwise require C++11.
 if ((MAYA_APP_VERSION VERSION_GREATER_EQUAL 2019) OR
-        ((DEFINED USD_VERSION_NUM) AND
-         (USD_VERSION_NUM VERSION_GREATER_EQUAL 2005)))
+        (USD_VERSION_NUM VERSION_GREATER_EQUAL 2005))
     set(CMAKE_CXX_STANDARD 14)
 else()
     set(CMAKE_CXX_STANDARD 11)
@@ -59,9 +58,7 @@ if (PXR_OVERRIDE_PLUGINPATH_NAME)
     _add_define("PXR_PLUGINPATH_NAME=${PXR_OVERRIDE_PLUGINPATH_NAME}")
 endif()
 
-if (DEFINED USD_VERSION_NUM)
-    _add_define("USD_VERSION_NUM=${USD_VERSION_NUM}")
-endif()
+_add_define("USD_VERSION_NUM=${USD_VERSION_NUM}")
 
 if (DEFINED UFE_PREVIEW_VERSION_NUM)
     _add_define("UFE_PREVIEW_VERSION_NUM=${UFE_PREVIEW_VERSION_NUM}")
--- a/cmake/defaults/CXXDefaults.cmake
+++ b/cmake/defaults/CXXDefaults.cmake
@@ -17,6 +17,19 @@ include(CXXHelpers)
 include(Version)
 include(Options)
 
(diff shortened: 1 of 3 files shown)


cmake/defaults/CXXDefaults.cmake
```diff
@@ -17,6 +17,19 @@ include(CXXHelpers)
 include(Version)
 include(Options)
 
+# Require C++14 if we're either building for Maya 2019 or later, or if we're
+# building against USD 20.05 or later. Otherwise require C++11.
+if ((MAYA_APP_VERSION VERSION_GREATER_EQUAL 2019) OR
+        ((DEFINED USD_VERSION_NUM) AND
+         (USD_VERSION_NUM VERSION_GREATER_EQUAL 2005)))
```
HamedSabri-adsk, line 24: Just 1 minor things, do we really need (DEFINED USD_VERSION_NUM)?  Why do we need to check for it?

mattyjams: Yeah, I was a little wary of that myself, but we already had something similar lower in that file:
https://github.com/Autodesk/maya-usd/blob/11b87a9097ca40eb9cf94f87d8504d32a962b9da/cmake/defaults/CXXDefaults.cmake#L49

It looks like it's technically possible to make it out of FindUSD.cmake without having gotten a value for `USD_VERSION_NUM`:
https://github.com/Autodesk/maya-usd/blob/11b87a9097ca40eb9cf94f87d8504d32a962b9da/cmake/modules/FindUSD.cmake#L102

Maybe that's the case for especially old releases of USD? I can't remember when that was added.

HamedSabri-adsk: Not a biggie at all but I believe it is safe to not check for this and the reason is that this variable always exist as part of logic that you showed me in FindUSD.cmake. This variable was introduced by Paul ( 848de8fe6c32 ) back in October of last year.

The reason I brought was because for certain variables ( e.g UFE_PREVIEW_VERSION_NUM ), we do want to check because they simply don't exist when building against certain version of things.

See my comments here:

https://github.com/Autodesk/maya-usd/pull/412/commits/42bec8b333d3d7662f762074d4a16979acf2cac5

Cheers.

mattyjams: Right, but since `USD_VERSION_NUM` is only set inside an `if()` block, it's *technically* possible for it not to be set:
https://github.com/Autodesk/maya-usd/blob/11b87a9097ca40eb9cf94f87d8504d32a962b9da/cmake/modules/FindUSD.cmake#L93

It would admittedly have to be for the pretty bizarre reason that `pxr/pxr.h` was missing even though `USD_INCLUDE_DIR` is set, since the `find_package_handle_standard_args()` call validates `USD_INCLUDE_DIR`. The build is likely to fail in some spectacular way if that's the case.

I can add `USD_VERSION` and `USD_VERSION_NUM` as `REQUIRED_VAR`s  to `find_package_handle_standard_args()` and remove the `defined()` checks? It might be better anyway if we fail faster during CMake config, as that would hopefully be easier to diagnose than some incomprehensible compile failure later on.

HamedSabri-adsk: Cool, that make sense to me. Good call!

mattyjams: Cool. Done in bc4d81f88750982c665507712594642b1b548dfd and b7a7f206df50a75115aef1bde4db4f3d21015187.

HamedSabri-adsk: Awesome! Thank you again for quick respond.

PR #439, opened 2020-04-16
add a performance test for measuring the time to duplicate a pxrUsdProxyShape

mattyjams:
We're working to address a performance regression in the Pixar batch renderer having to do with its use of many Hydra render tasks. We're mainly looking for some short term relief for our production artists in order to buy ourselves some time to work on adopting the VP2.0 render delegate.

This test is modeled after our other pxrUsdMayaGL performance tests, for example:
https://github.com/Autodesk/maya-usd/blob/11b87a9097ca40eb9cf94f87d8504d32a962b9da/plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeDrawPerformance.py#L180

Like the other pxrUsdMayaGL tests in the repo here, these are unfortunately not yet being run through ctest, so this PR is probably mostly irrelevant outside of Pixar.


plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeDuplicatePerformance.py
```diff
@@ -0,0 +1,164 @@
...
+        # it comes out too dark.
+        cmds.setAttr("defaultColorMgtGlobals.outputTransformEnabled", 1)
+
+        # Do the render.
+        cmds.ogsRender(camera=self._cameraName, currentFrame=True, width=960,
+            height=540)
+
+    def testPerfGridOfCubeModelDuplicate(self):
+        """
+        Tests speed of duplicating a model reference assembly/proxy when many
```
kxl-adsk, line 130: How many are `many`?

mattyjams: Yeah, sorry, we've been intentionally leaving the comments vague in terms of numbers to try to avoid them going stale in case we need to scale up the tests to keep pace with development.

This test is scaled back a bit from usual production workloads to make for a reasonable unit test, but at the moment it's a 5x5x5 cubic grid for a total of 125 of `pxrUsdProxyShape`. Each one of those proxy shapes references a USD stage containing 10x10x10 Mesh prims. Because of our use (read: abuse) of the Hydra API and our high render task counts, we're tripping over some O(N^2) behavior during Hydra sync where we visit every rprim in the index for every render task:
https://github.com/PixarAnimationStudios/USD/blob/6af4d8afd02682ee4365d1b7d60683e95acf12ac/pxr/imaging/hd/dirtyList.cpp#L107

Internally, this particular test takes about 4 seconds to duplicate one of those proxy shapes. With production assets, it's not uncommon to have 1000+ proxy shapes in a medium to large set, and artists are sometimes seeing duplicates take as much as 15-20 seconds.

We're working to address that now, so I'll likely have some follow-up PRs around this for `pxrUsdMayaGL`.

kxl-adsk: Thank you for providing extra info. Have you tried comparing performance with VP2RenderDelegate? FYI. We are actively working on improving instancing and consolidation performance with subscene override (an API used by VP2RenderDelegate). It would be good to know if there is something else we should be measuring when doing this work.

mattyjams: We haven't yet tested this particular issue with the VP2.0 render delegate, but we'll hopefully be looking at that soon. Once I'm able to devote some time to  that, one of the first things will probably be to make sure that all of these pxrUsdMayaGL tests can be run externally so that we're measuring the same things for both viewport integrations.

PR #443, opened 2020-04-19
plugin/pxr is now adopting mayaUsd_compile_config routine

HamedSabri-adsk:
Following PR #412, plugin/pxr is now getting the compiler configuration by using mayaUsd_compile_config.

I didn't delete `defaults` sub-directory and left it there by itself for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,9 @@ option(BUILD_TESTS "Build tests." ON)
 option(BUILD_STRICT_MODE "Enforce all warnings as errors." ON)
 option(BUILD_SHARED_LIBS "Build libraries as shared or static." ON)
 option(CMAKE_WANT_UFE_BUILD "Enable building with UFE (if found)." ON)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -29,6 +29,9 @@ option(BUILD_TESTS "Build tests." ON)
 option(BUILD_STRICT_MODE "Enforce all warnings as errors." ON)
 option(BUILD_SHARED_LIBS "Build libraries as shared or static." ON)
 option(CMAKE_WANT_UFE_BUILD "Enable building with UFE (if found)." ON)
+option(PXR_ENABLE_PYTHON_SUPPORT "Enable Python based components for USD" ON)
+option(PXR_BUILD_TESTS "Build tests" ON)
+option(PXR_BUILD_MONOLITHIC "Build a monolithic library." OFF)
```
mattyjams, line 34: These options and their corresponding code should probably all get removed, but we can punt on that for now.

PR #445, opened 2020-04-19
plugin/al is now adopting mayaUsd_compile_config routine

HamedSabri-adsk:
Following PR #412, plugin/al is now getting the compiler configuration by using mayaUsd_compile_config.


murphyeoin:
Thanks @mattyjams - I've skimmed this PR and #412 fairly quickly, and the idea of consolidating all of this configuration sounds great. I can't see anything obvious that would break our build so  I'll merge these in and run them with our build system and report back. Thanks @HamedSabri-adsk

HamedSabri-adsk:
@murphyeoin @mattyjams  Awesome. Thank you both.

murphyeoin:
Hi @HamedSabri-adsk @mattyjams ran it through our build system and it's looking good. The only error I got was a cmake error with https://github.com/Autodesk/maya-usd/commit/4eb3d237b700e01857eff50697949b4683f0cf9a, not part of this PR, but pulled in as part of the update to dev. Works on our open source build
For reference:
```
CMake Error at /scratch/dev/maya-usd/plugin/al/lib/AL_USDMaya/CMakeLists.txt:153 (add_library):
  Target "AL_USDMaya" links to target "$<IF:0,Boost::thread,optimized" but
  the target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
```
Definitely looks related to cmake import targets.. I don't know enough about that new cmake syntax to wonder why it's evaluating when we're not in Windows .... either way this PR is fine by me

HamedSabri-adsk:
Thank you @murphyeoin. Yeah, this looks strange and my hunch tells me that this could be related to the buggy Boost module ( FindBoost.cmake ).  FWIW, we don't get this error on our side with the Linux VM machine we have setup.

The reason we have the below condition is because we found an issue on MacOS with linking against boost's thread library via import target and therefore we had to switch to old way of calling Boost_THREAD_LIBRARY variable directly. We kept the same thing for Linux to get called.

`
$<IF:$<BOOL:${IS_WINDOWS}>,Boost::thread,${Boost_THREAD_LIBRARY}>`

Can you please do me a favor and as a quick hack replace ${Boost_THREAD_LIBRARY} with `Boost::thread` and see if things work?

murphyeoin:
Hi @HamedSabri-adsk that does indeed work... The FindBoost.cmake we're using is the one that ships with cmake-3.16, the documentation for which states:
```
Boost_<C>_LIBRARY      - Libraries to link for component <C> (may include
                           target_link_libraries debug/optimized keywords)
  ```
Thanks!
Eoin

HamedSabri-adsk:
Hi @murphyeoin , glad to hear that. I will make a PR to address this. 

Cheers

PR #449, opened 2020-04-20
updates from Hdx de-boostification and Hd reorganization

mattyjams:
Two recent post-20.05 commits in core USD made some Hdx pointer types use std::shared_ptr instead of boost::shared_ptr, and a third commit moved some Hd enum values from hd/enums.h to hd/types.h.

The changes here update the maya-usd code to account for these core USD commits.

### DIFF
--- a/lib/usd/hdMaya/adapters/lightAdapter.cpp
+++ b/lib/usd/hdMaya/adapters/lightAdapter.cpp
@@ -33,6 +33,12 @@
 #include <hdMaya/adapters/constantShadowMatrix.h>
 #include <hdMaya/adapters/mayaAttrs.h>
 
+#if HDX_API_VERSION < 7
+#include <boost/shared_ptr.hpp>
+#include <boost/make_shared.hpp>
+#endif
+
+
 PXR_NAMESPACE_OPEN_SCOPE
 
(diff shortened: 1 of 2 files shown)


kxl-adsk:
@mattyjams We just adopted 20.05 and will stay for a few weeks on it while we deal with python3. I see HDX_API_VERSION guards - we shouldn't have any issue in merging this change even without adopting the latest dev branch? 

FWIW. we will have to validate that everything works on our end with the latest dev branch before merging.

mattyjams:
@kxl-adsk: Yeah, that's right. The `boost::shared_ptr` vs. `std::shared_ptr` changes are gated by `HDX_API_VERSION`, so they should be safe with or without everything else from 20.05. And the addition of the extra `hd` header should be safe too since that's been there in core USD since at least 19.07.

AwanPydrall:
#101

Pada tanggal Sel, 21 Apr 2020 23.25, Matt Johnson <notifications@github.com>
menulis:

AwanPydrall:
A1

Pada tanggal Rab, 22 Apr 2020 03.30, Muhammad Qoyum Akhiriawan <
makhiriawan@gmail.com> menulis:

PR #451, opened 2020-04-21
remove virtual dispatch for PxrMayaHdShapeAdapter::IsViewport2()

mattyjams:
Following on from #450, we want to start being more explicit about the division between the legacy viewport and Viewport 2.0 so that it's easier to tear out the legacy stuff a little bit at a time.

There might also be a slight performance improvement here with `PxrMayaHdShapeAdapter::IsViewport2()` no longer dispatching virtually.

### DIFF
--- a/lib/mayaUsd/render/pxrUsdMayaGL/instancerImager.h
+++ b/lib/mayaUsd/render/pxrUsdMayaGL/instancerImager.h
@@ -72,7 +72,9 @@ public:
         ContinueTrackingOnDisconnectDelegate delegate);
 
     /// Factory function for creating instancer shape adapters.
-    typedef std::function<UsdMayaGL_InstancerShapeAdapter*()> InstancerShapeAdapterFactory;
+    using InstancerShapeAdapterFactory =
+        std::function<UsdMayaGL_InstancerShapeAdapter*(
+            const bool /* isViewport2 */)>;
 
     /// Set the factory function for creating instancer shape adapters.
     MAYAUSD_CORE_PUBLIC
@@ -177,7 +179,8 @@ private:
     /// no factory has been set, returns a UsdMayaGL_InstancerShapeAdapter base
     /// class object.  The caller must manage the lifescope of the returned
     /// object.
-    static UsdMayaGL_InstancerShapeAdapter* CreateInstancerShapeAdapter();
+    static UsdMayaGL_InstancerShapeAdapter* CreateInstancerShapeAdapter(
+            const bool isViewport2);
 
     UsdMayaGL_InstancerImager();
     ~UsdMayaGL_InstancerImager();
--- a/lib/mayaUsd/render/pxrUsdMayaGL/instancerImager.h
+++ b/lib/mayaUsd/render/pxrUsdMayaGL/instancerImager.h
@@ -73,8 +73,7 @@ public:
 
     /// Factory function for creating instancer shape adapters.
     using InstancerShapeAdapterFactory =
-        std::function<UsdMayaGL_InstancerShapeAdapter*(
-            const bool /* isViewport2 */)>;
+        std::function<UsdMayaGL_InstancerShapeAdapter*(bool /* isViewport2 */)>;
 
     /// Set the factory function for creating instancer shape adapters.
     MAYAUSD_CORE_PUBLIC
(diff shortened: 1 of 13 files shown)


lib/mayaUsd/render/pxrUsdMayaGL/instancerImager.h
```diff
@@ -177,7 +179,8 @@ class UsdMayaGL_InstancerImager : public TfWeakBase {
     /// no factory has been set, returns a UsdMayaGL_InstancerShapeAdapter base
     /// class object.  The caller must manage the lifescope of the returned
     /// object.
-    static UsdMayaGL_InstancerShapeAdapter* CreateInstancerShapeAdapter();
+    static UsdMayaGL_InstancerShapeAdapter* CreateInstancerShapeAdapter(
+            const bool isViewport2);
```
HdC-adsk, line 183: constness on the pass-by-value argument is not necessary. Can you remove it?

Regarding to coding style, we are likely to have 100 chars per line so there is no need to break a new line.

HdC-adsk: @kxl-adsk Have we published the coding style guildline somewhere for partners?

lib/mayaUsd/render/pxrUsdMayaGL/proxyDrawOverride.cpp
```diff
@@ -64,10 +64,11 @@ UsdMayaProxyDrawOverride::UsdMayaProxyDrawOverride(const MObject& obj) :
         MHWRender::MPxDrawOverride(obj,
                                    UsdMayaProxyDrawOverride::draw
 #if MAYA_API_VERSION >= 201651
```
HdC-adsk, line 66: If we don't care about pre-2018 maya versions, we should also remove this  conditional compilation.

lib/mayaUsd/render/pxrUsdMayaGL/instancerImager.h
HdC-adsk: Another question: can this argument be true by default?

lib/mayaUsd/render/pxrUsdMayaGL/shapeAdapter.h
```diff
@@ -201,14 +201,15 @@ class PxrMayaHdShapeAdapter
...
-        /// version of Sync() that is used to populate it.
+        /// The shape adapter's viewport renderer affiliation must be specified
+        /// when it is constructed.
         ///
         /// Returns true if the shape adapter should be used for batched
         /// drawing/selection in Viewport 2.0, or false if it should be used
         /// in the legacy viewport.
-        MAYAUSD_CORE_PUBLIC
-        virtual bool IsViewport2() const;
+        bool IsViewport2() const {
```
HdC-adsk, line 210: Just to be sure, we don't need to export this method?

lib/mayaUsd/render/pxrUsdMayaGL/instancerImager.h
kxl-adsk: clang-format is published on TSC forum...pending review.

lib/mayaUsd/render/pxrUsdMayaGL/proxyDrawOverride.cpp
mattyjams: Makes sense. I can put up a separate PR for that later since it's not directly related to the changes here, and I'm sure there are a bunch of other places where we're testing for pre-2018 versions.

lib/mayaUsd/render/pxrUsdMayaGL/shapeAdapter.h
mattyjams: Right, we don't need to export this anymore since it's now defined inline right here.

lib/mayaUsd/render/pxrUsdMayaGL/instancerImager.h
mattyjams: On the argument const-ness, yeah, I can remove that.

About defaulting it to true, I had thought of doing that initially, but if we made it default to true here, it seemed like it should default to true all the way down to the shape adapter constructors. It seemed safer to me to require callers to specify a value rather than accidentally end up creating a shape adapter for the wrong viewport. If you feel strongly though, I can revisit that.

mattyjams: Rebased and removed the const qualifiers in f0d222549112146c7ef372041d5f9011edb5d91d.

lib/mayaUsd/render/pxrUsdMayaGL/proxyDrawOverride.cpp
HdC-adsk: LGTM.

AwanPydrall:
AwanPydral

Jengky

Pada tanggal Sel, 21 Apr 2020 18.57, Krystian Ligenza <
notifications@github.com> menulis:

PR #452, opened 2020-04-21
MAYA-102232 Check the index vs the size of the prim data array

fowlertADSK:


lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -136,10 +136,20 @@ namespace {
             if (requiresUnsharedVertices) {
                 const VtIntArray& faceVertexIndices = topology.GetFaceVertexIndices();
                 if (numVertices == faceVertexIndices.size()) {
+                    unsigned int dataSize = primvarData.size();
                     for (size_t v = 0; v < numVertices; v++) {
                         SRC_TYPE* pointer = reinterpret_cast<SRC_TYPE*>(
                             reinterpret_cast<float*>(&vertexBuffer[v]) + channelOffset);
-                        *pointer = primvarData[faceVertexIndices[v]];
```
fowlertADSK, line 142: We have checks comparing the size of primvarData and faceVertexIndices arrays, but we weren't checking the actual value that we are getting out of faceVertexIndices when using it as an index to the primvarData array.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -136,10 +136,20 @@ namespace {
...
+                        unsigned int index = faceVertexIndices[v];
+                        if (index < dataSize) {
+                            *pointer = primvarData[faceVertexIndices[v]];
+                        } else {
+                            TF_DEBUG(HDVP2_DEBUG_MESH).Msg("Invalid face vertex index '%s': "
+                                                    "primvar %s has %zu elements, while its topology "
+                                                    "references only upto element index %zu.\n",
+                                                    rprimId.asChar(), primvarName.GetText(),
+                                                    primvarData.size(), numVertices);                            
+                        }
```
fowlertADSK, line 152: Stealing this error message from the block of code just a bit down from here in the file where we were looking at almost the exact same thing.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -136,10 +136,20 @@ namespace {
...
-                        *pointer = primvarData[faceVertexIndices[v]];
+                        unsigned int index = faceVertexIndices[v];
+                        if (index < dataSize) {
+                            *pointer = primvarData[faceVertexIndices[v]];
+                        } else {
+                            TF_DEBUG(HDVP2_DEBUG_MESH).Msg("Invalid face vertex index '%s': "
+                                                    "primvar %s has %zu elements, while its topology "
+                                                    "references only upto element index %zu.\n",
+                                                    rprimId.asChar(), primvarName.GetText(),
+                                                    primvarData.size(), numVertices);
```
kxl-adsk, line 151: primvarData.size() == dataSize...as for debugging...wouldn't knowing `index` be more useful than `numVertices` ?

fowlertADSK: You're talking about the print statement printing out index?  Maybe I suppose, although with this kind of thing the actual # might not help you at all.  The fact that something was out of bounds + the name of the object is what you probably need and not the actual number.

kxl-adsk: Agreed. Then how about we simply remove # from the message?

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
HdC-adsk: It sounds like a corrupted scene or a bug in upstream software producing USD data causing our inefficiency (due to more checking).

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
HdC-adsk: Sorry what do you mean?

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -136,10 +136,20 @@ namespace {
...
                 const VtIntArray& faceVertexIndices = topology.GetFaceVertexIndices();
                 if (numVertices == faceVertexIndices.size()) {
+                    unsigned int dataSize = primvarData.size();
                     for (size_t v = 0; v < numVertices; v++) {
                         SRC_TYPE* pointer = reinterpret_cast<SRC_TYPE*>(
                             reinterpret_cast<float*>(&vertexBuffer[v]) + channelOffset);
-                        *pointer = primvarData[faceVertexIndices[v]];
+                        unsigned int index = faceVertexIndices[v];
+                        if (index < dataSize) {
+                            *pointer = primvarData[faceVertexIndices[v]];
```
HdC-adsk, line 145: Use index to avoid another lookup.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -136,10 +136,20 @@ namespace {
...
+                        } else {
+                            TF_DEBUG(HDVP2_DEBUG_MESH).Msg("Invalid face vertex index '%s': "
+                                                    "primvar %s has %zu elements, while its topology "
+                                                    "references only upto element index %zu.\n",
```
HdC-adsk, line 149: In this case, it references out of bound elements.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
fowlertADSK: Oops, yes absolutely.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -136,10 +136,20 @@ namespace {
...
-                        *pointer = primvarData[faceVertexIndices[v]];
+                        unsigned int index = faceVertexIndices[v];
+                        if (index < dataSize) {
+                            *pointer = primvarData[index];
+                        } else {
+                            TF_DEBUG(HDVP2_DEBUG_MESH).Msg("Invalid Hydra prim '%s': "
+                                                    "primvar %s has %zu elements, while its topology "
+                                                    "references face vertex index %zu.\n",
+                                                    rprimId.asChar(), primvarName.GetText(),
+                                                    primvarData.size(), index);
```
kxl-adsk, line 151: You already have acquired `primvarData.size()` and stored it as `dataSize`

PR #455, opened 2020-04-22
Tremblp/maya 103815/ufe parenting usd 00

ppt-adsk:


ppt-adsk:
This is the implementation of the UFE parenting interface in maya-usd.  It relies on UFE parenting integration into Maya that will only be available in Maya Preview Release 115.  The current implementation is incomplete: the UFE implementation allows for arbitrary insert child position, but the current implementation provides append only.  Furthermore, parenting at the top level of the USD hierarchy (i.e. below the pseudo-root, which in Maya is hosted by the Maya proxy shape node itself) is not supported.  Only USD to USD parenting is allowed.

The USD parenting code itself is almost identical to the existing USD namespace edit code we have, which is rename.  Therefore, most of this pull request is in implementing new UFE interface requirements, refactoring code for simplicity, writing a test for parenting, etc.

lib/mayaUsd/ufe/UsdRotateUndoableCommand.cpp
```diff
@@ -20,67 +20,61 @@
...
+    performImp(0, 0, 0);	// Add an empty rotate
 }
 
-void UsdRotateUndoableCommand::perform()
+void UsdRotateUndoableCommand::performImp(double x, double y, double z)
 {
-	// No-op, use rotate to move the object
-	// The Maya rotate command directly invokes our rotate() method in its
-	// redoIt(), which is invoked both for the inital rotate and the redo.
+	rotateOp(prim(), path(), x, y, z);
```
ppt-adsk, line 77: The fact that perform() was a no-op was a very incomplete implementation based on the behavior of the Maya move, rotate and scale commands, which perform undo themselves by simply setting the previous value.  For those uses, undo is not invoked.

When used in absolute mode, the parent command changes the local transform of an object.  Its undo should not be concerned about setting the previous values for the parented object's translate, rotate, and scale: the undo commands should do this themselves (separation of concerns).

lib/mayaUsd/ufe/UsdRotateUndoableCommand.h
```diff
@@ -31,40 +32,45 @@ namespace ufe {
 //! \brief Absolute rotation command of the given prim.
 /*!
 	Ability to perform undo to restore the original rotation value.
-	As of 06/07/2018, redo is a no op as Maya re-does the operation for redo
  */
-class MAYAUSD_CORE_PUBLIC UsdRotateUndoableCommand : public Ufe::RotateUndoableCommand
+class MAYAUSD_CORE_PUBLIC UsdRotateUndoableCommand : public Ufe::RotateUndoableCommand, public UsdTRSUndoableCommandBase<GfVec3f>
```
ppt-adsk, line 36: Derive from the helper class that contains all the factored-out, common code for translation, rotate, scale.

lib/mayaUsd/ufe/UsdRotateUndoableCommand.h
```diff
@@ -31,40 +32,45 @@ namespace ufe {
...
 	UsdRotateUndoableCommand(const UsdRotateUndoableCommand&) = delete;
 	UsdRotateUndoableCommand& operator=(const UsdRotateUndoableCommand&) = delete;
 	UsdRotateUndoableCommand(UsdRotateUndoableCommand&&) = delete;
 	UsdRotateUndoableCommand& operator=(UsdRotateUndoableCommand&&) = delete;
 
-	//! Create a UsdRotateUndoableCommand from a USD prim, UFE path and UFE scene item.
-	static UsdRotateUndoableCommand::Ptr create(const UsdPrim& prim, const Ufe::Path& ufePath, const Ufe::SceneItem::Ptr& item);
+	//! Create a UsdRotateUndoableCommand from a UFE scene item.  The
+	//! command is not executed.
+	static UsdRotateUndoableCommand::Ptr create(
```
ppt-adsk, line 48: You can now set the initial x, y, and z values right at the point of creation.

lib/mayaUsd/ufe/UsdRotateUndoableCommand.h
```diff
@@ -31,40 +32,45 @@ namespace ufe {
...
 
 private:
-	UsdPrim fPrim;
-	Ufe::Path fPath;
-	UsdAttribute fRotateAttrib;
-	GfVec3f fPrevRotateValue;
+
+    static TfToken rotXYZ;
+
+    TfToken attributeName() const override { return rotXYZ; }
```
ppt-adsk, line 67: Implement the TRS helper base class hooks.

lib/mayaUsd/ufe/UsdScaleUndoableCommand.cpp
```diff
@@ -17,65 +17,60 @@
```
ppt-adsk, line 17: Mostly same as rotate: most of the code moved to the helper base class, and undo is now implemented.

lib/mayaUsd/ufe/UsdScaleUndoableCommand.h
```diff
@@ -20,6 +20,7 @@
 #include <pxr/usd/usd/attribute.h>
```
ppt-adsk, line 20: Same as rotate.

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.cpp
```diff
@@ -0,0 +1,112 @@
+//
```
ppt-adsk, line 1: This is the new base class for shared code between translate, rotate, and scale.

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+    // See
+    // https://stackoverflow.com/questions/17853212/using-shared-from-this-in-templated-classes
+    // for explanation of this->shared_from_this() in templated class.
+    attribute().Get(&fPrevValue);
+    Ufe::Scene::instance().addObjectPathChangeObserver(this->shared_from_this());
+}
+
+template<class V>
+void UsdTRSUndoableCommandBase<V>::operator()(
```
ppt-adsk, line 54: If the object we translate, rotate, scale sees its path changed (e.g. when reparenting or renaming), make sure we keep up with its new path.

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.cpp
```diff
@@ -0,0 +1,112 @@
...
+    attribute().Set(fPrevValue);
+    // Todo : We would want to remove the xformOp
+    // (SD-06/07/2018) Haven't found a clean way to do it - would need to investigate
+}
+
+template<class V>
+void UsdTRSUndoableCommandBase<V>::redoImp()
+{
+    if (fDoneOnce) {
+        attribute().Set(fNewValue);
```
ppt-adsk, line 78: If the command has been done once, we know there is an attribute on the prim for the translate, rotate, or scale.  Otherwise, we must invoke the code that creates the attribute if required.

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.cpp
```diff
@@ -0,0 +1,112 @@
...
+{
+    if (notification->previousPath() == path()) {
+        fItem = std::dynamic_pointer_cast<UsdSceneItem>(notification->item());
+    }
+}
+
+template<class V>
+void UsdTRSUndoableCommandBase<V>::perform(double x, double y, double z)
+{
+    fNewValue = V(x, y, z);
```
ppt-adsk, line 97: When called, update our redo value, and set the value onto the attribute.

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.h
```diff
@@ -0,0 +1,101 @@
...
+//! \brief Base class for translate, rotate, scale undoable commands.
+//
+// UsdRotateUndoableCommand causes USD assertion when used with GfVec3d:
+//
+// Coding Error: in _SetValueImpl at line 5619 of [...]/usd/pxr/usd/usd/stage.cpp -- Type mismatch for </pCylinder1/pCube1.xformOp:rotateXYZ>: expected 'GfVec3f', got 'GfVec3d'
+//
+// Therefore, this class is templated on GfVec3f or GfVec3d, so rotate can use
+// GfVec3f.  To be investigated.  PPT, 9-Apr-2020.
+//
+template<class V>
```
ppt-adsk, line 40: Templated on the GfVec type.  Translate is GfVec3d, rotate and scale are GfVec3f, otherwise we crash.  No time to investigate why at this point.

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.h
```diff
@@ -0,0 +1,101 @@
...
+
+    // Initialize the command.
+    void initialize();
+
+    // Undo and redo implementations.
+    void undoImp();
+    void redoImp();
+
+    // Set the new value of the command (for redo), and execute the command.
+    void perform(double x, double y, double z);
```
ppt-adsk, line 58: Calls performImp(), which implements the translate, rotate, or scale.

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.h
```diff
@@ -0,0 +1,101 @@
...
+    UsdSceneItem::Ptr fItem;
+    V                 fPrevValue;
+    V                 fNewValue;
+    bool              fOpAdded{false};
+    bool              fDoneOnce{false};
+}; // UsdTRSUndoableCommandBase
+
+// shared_ptr requires public ctor, dtor, so derive a class for it.
+template<class T>
+struct MakeSharedEnabler : public T {
```
ppt-adsk, line 94: Utility class to allow make_shared to be called.  Since we've set the derived translate, rotate, and scale constructors to be protected, make_shared could not be called on them.

lib/mayaUsd/ufe/UsdTransform3d.cpp
```diff
@@ -60,17 +60,26 @@ namespace {
 	}
 }
 
-UsdTransform3d::UsdTransform3d()
-	: Transform3d()
+UsdTransform3d::UsdTransform3d() : Transform3d()
 {
 }
 
+UsdTransform3d::UsdTransform3d(const UsdSceneItem::Ptr& item)
```
ppt-adsk, line 67: Had to go through setItem() because the default constructor did not take an argument, inconvenient, fixed.

lib/mayaUsd/ufe/UsdTransform3d.cpp
```diff
@@ -91,11 +100,12 @@ Ufe::SceneItem::Ptr UsdTransform3d::sceneItem() const
 	return fItem;
 }
 
-Ufe::TranslateUndoableCommand::Ptr UsdTransform3d::translateCmd()
+#if UFE_PREVIEW_VERSION_NUM >= 2013
+Ufe::TranslateUndoableCommand::Ptr UsdTransform3d::translateCmd(double x, double y, double z)
 {
-	auto translateCmd = UsdTranslateUndoableCommand::create(fPrim, fItem->path(), fItem);
-	return translateCmd;
+	return UsdTranslateUndoableCommand::create(fItem, x, y, z);
```
ppt-adsk, line 106: Return the undoable command that performs the translation.

lib/mayaUsd/ufe/UsdTransform3d.cpp
```diff
@@ -118,22 +128,73 @@ Ufe::Vector3d UsdTransform3d::translation() const
 	return Ufe::Vector3d(x, y, z);
 }
 
-Ufe::RotateUndoableCommand::Ptr UsdTransform3d::rotateCmd()
+#if UFE_PREVIEW_VERSION_NUM >= 2013
+Ufe::Vector3d UsdTransform3d::rotation() const
```
ppt-adsk, line 132: With UFE < 0.2.13, you could query translation, but not rotate or scale.  Oversight fixed in UFE, implemented here in maya-usd.

lib/mayaUsd/ufe/UsdTransform3d.cpp
```diff
@@ -118,22 +128,73 @@ Ufe::Vector3d UsdTransform3d::translation() const
...
 }
 
+#if UFE_PREVIEW_VERSION_NUM >= 2013
+Ufe::ScaleUndoableCommand::Ptr UsdTransform3d::scaleCmd(double x, double y, double z)
+{
+	return UsdScaleUndoableCommand::create(fItem, x, y, z);
+}
+#endif
+
+#if !defined(UFE_V2_FEATURES_AVAILABLE) && defined(UFE_V1_FEATURES_AVAILABLE)
```
ppt-adsk, line 182: Old UFE 1.0 interfaces.  Should no longer be used in UFE 2.0.

lib/mayaUsd/ufe/UsdTransform3d.h
```diff
@@ -45,18 +46,30 @@ class MAYAUSD_CORE_PUBLIC UsdTransform3d : public Ufe::Transform3d
...
 	static UsdTransform3d::Ptr create();
+	static UsdTransform3d::Ptr create(const UsdSceneItem::Ptr& item);
 
 	void setItem(const UsdSceneItem::Ptr& item);
 
 	// Ufe::Transform3d overrides
 	const Ufe::Path& path() const override;
 	Ufe::SceneItem::Ptr sceneItem() const override;
+#ifdef UFE_V2_FEATURES_AVAILABLE
+#if UFE_PREVIEW_VERSION_NUM >= 2013
```
ppt-adsk, line 57: New UFE interfaces.

lib/mayaUsd/ufe/UsdTransform3dHandler.h
```diff
@@ -44,9 +44,6 @@ class MAYAUSD_CORE_PUBLIC UsdTransform3dHandler : public Ufe::Transform3dHandler
 	// Ufe::Transform3dHandler overrides
 	Ufe::Transform3d::Ptr transform3d(const Ufe::SceneItem::Ptr& item) const override;
 
-private:
```
ppt-adsk, line 47: Transform handler had a UsdTransform3d as a data member, which was therefore essentially a singleton, because there is only one handler for a given interface type in the UFE run-time.  This was a misguided optimization, as it meant that as soon as you tried to create more than one Transform3d interface, one per object for different USD objects (which is frequent), you would in fact get the same interface object attached to all USD objects, and you'd be scratching your head trying to figure out why the underlying object's transformation had changed without your making calls on that particular object.  Just create a transform 3d interface object per USD object, as needed.

lib/mayaUsd/ufe/UsdTranslateUndoableCommand.cpp
```diff
@@ -1,5 +1,5 @@
 //
```
ppt-adsk, line 1: Same as rotate and scale.

lib/mayaUsd/ufe/UsdTranslateUndoableCommand.h
```diff
@@ -1,5 +1,5 @@
 //
```
ppt-adsk, line 1: Same as rotate and scale.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -0,0 +1,161 @@
+//
```
ppt-adsk, line 1: This is one of the two core files in the review: the implementation of the UFE insert child command interface.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -0,0 +1,161 @@
...
+    const UsdSceneItem::Ptr& pos
+)
+{
+    return std::make_shared<UsdUndoInsertChildCommand>(parent, child, pos);
+}
+
+bool UsdUndoInsertChildCommand::insertChildRedo()
+{
+    // See comments in UsdUndoRenameCommand.cpp.
+    bool status = SdfCopySpec(fLayer, fUsdSrcPath, fLayer, fUsdDstPath);
```
ppt-adsk, line 80: If this looks very much like the rename implementation, it's because it is: the same strategy is used.  We copy the USD source hierarchy into the destination (to reparent), then inactivate the source.

We are currently investigating alternative implementations for rename and reparent.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -0,0 +1,161 @@
...
+#include <pxr/usd/sdf/copyUtils.h>
+#include <pxr/base/tf/token.h>
+
+MAYAUSD_NS_DEF {
+namespace ufe {
+
+UsdUndoInsertChildCommand::UsdUndoInsertChildCommand(
+    const UsdSceneItem::Ptr& parent,
+    const UsdSceneItem::Ptr& child,
+    const UsdSceneItem::Ptr& pos
```
ppt-adsk, line 42: Note that currently the pos argument is ignored, and reparent is append only.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -152,12 +152,8 @@ std::string uniqueName(const TfToken::HashSet& existingNames, std::string srcNam
 	return dstName;
 }
 
-std::string uniqueChildName(const Ufe::SceneItem::Ptr& parent, const Ufe::Path& childPath)
+std::string uniqueChildName(const UsdSceneItem::Ptr& usdParent, const Ufe::Path& childPath)
```
ppt-adsk, line 155: We always have a UsdSceneItem in maya-usd, no sense in passing around a base class object and downcasting.

test/lib/ufe/test-samples/parentCmd/simpleSceneMayaPlusUSD_TRS.ma
```diff
@@ -0,0 +1,242 @@
+//Maya ASCII 2020 scene
```
ppt-adsk, line 1: A test scene that has a simple Maya scene, and the USD version of it under a proxy shape.  This makes comparison between Maya parenting and USD parenting behavior trivial: just perform the operation on the Maya objects, and the USD objects should behave the same and have the same transform values.

test/lib/ufe/testParentCmd.py
```diff
@@ -0,0 +1,222 @@
...
+        # Load a file that has the same scene in both the Maya Dag
+        # hierarchy and the USD hierarchy.
+        filePath = os.path.join(os.path.dirname(os.path.realpath(__file__)), "test-samples", "parentCmd", "simpleSceneMayaPlusUSD_TRS.ma" )
+        cmds.file(filePath, force=True, open=True)
+
+        # Clear selection to start off
+        cmds.select(clear=True)
+
+    @unittest.skipUnless(mayaUtils.previewReleaseVersion() >= 115, 'Parent functionality only available in Maya Preview Release 115 or later.') 
+    def testParentRelative(self):
```
ppt-adsk, line 65: Test parent -relative.  This is the simpler case, as the object's local transformation is unaffected.

test/lib/ufe/testParentCmd.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # Undo to bring scene to its original state.
+        for i in range(4):
+            cmds.undo()
+
+        cylChildren = cylHier.children()
+        self.assertEqual(len(cylChildren), 0)
+
+    @unittest.skipUnless(mayaUtils.previewReleaseVersion() >= 115, 'Parent functionality only available in Maya Preview Release 115 or later.') 
+    def testParentAbsolute(self):
```
ppt-adsk, line 149: Test parent -absolute.  This changes the object's local transformation, but keeps its world transformation unchanged.  This means that undo must restore the local transformation.

test/lib/ufe/ufeTestUtils/mayaUtils.py
```diff
@@ -127,3 +130,17 @@ def openTopLayerScene():
     filePath = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "test-samples", "ballset", "StandaloneScene", "top_layer.ma" )
     cmds.file(filePath, force=True, open=True)
     
+def previewReleaseVersion():
```
ppt-adsk, line 133: Utility function to return the Maya preview release version.  The parenting test must not be run for a Maya version before preview release 115 (unless you use the special override environment variable).

ppt-adsk:
With latest commit, parenting to (top-level) proxy shape is now supported.

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.cpp
ppt-adsk: The attribute is added, but perform() does conversion to the common transform API.  Without it, assertions occur on rotate and scale, such as this one:
Runtime Error: in _SetValueImpl at line 5604 of /mnt/dataTwoTwenty/tremblp/projects/ecg-usd-build/ecg-usd-build/usd/pxr/usd/usd/stage.cpp -- Empty typeName for </pCube1.xformOp:rotateXYZ>
I will add this comment to the implementation.

ppt-adsk:
> 
> 
> Looks good to me! Looking forward to testing it. Thanks!

Thanks @pilarmolinalopez for the review!

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.cpp
kxl-adsk: Let's keep this as a post review action, but I would prefer to unify how we are dealing with the problem (selection commands vs TRS command base) and rely on UFE path. Dealing with notifications can be tricky and surprising.

PR #465, opened 2020-04-25
Enable exception handling for VS compiler

HamedSabri-adsk:
Full compiler support for the Standard C++exception handling model that safely unwinds stack objects requires /EHsc (recommended)

### DIFF
--- a/cmake/compiler_config.cmake
+++ b/cmake/compiler_config.cmake
@@ -24,6 +24,8 @@ set(msvc_flags
     /Zc:rvalueCast
     # enable multiprocessor builds.
     /MP
+    # enable exception handling.
+    /EHsc
     # disable warnings
     /wd4244
     /wd4267
--- a/lib/mayaUsd/CMakeLists.txt
+++ b/lib/mayaUsd/CMakeLists.txt
@@ -49,6 +49,9 @@ target_compile_definitions(${PROJECT_NAME}
         $<$<BOOL:${IS_LINUX}>:LINUX>
         $<$<BOOL:${IS_LINUX}>:GL_GLEXT_PROTOTYPES>
         $<$<BOOL:${IS_LINUX}>:GLX_GLXEXT_PROTOTYPES>
+
+        # this flag is needed when building maya-usd in Maya
+        $<$<BOOL:${IS_WINDOWS}>:WIN32>
 )
 
 if(DEFINED UFE_PREVIEW_VERSION_NUM)


cmake/compiler_config.cmake
```diff
@@ -24,6 +24,8 @@ set(msvc_flags
     /Zc:rvalueCast
     # enable multiprocessor builds.
     /MP
+    # enable exception handling.
+    /EHsc
```
HamedSabri-adsk, line 28: Full compiler support for the Standard C++ exception handling model that safely unwinds stack objects requires /EHsc (recommended). We previously had support this flag but I missed to add it to the compiler_config

lib/mayaUsd/CMakeLists.txt
```diff
@@ -49,6 +49,9 @@ target_compile_definitions(${PROJECT_NAME}
         $<$<BOOL:${IS_LINUX}>:LINUX>
         $<$<BOOL:${IS_LINUX}>:GL_GLEXT_PROTOTYPES>
         $<$<BOOL:${IS_LINUX}>:GLX_GLXEXT_PROTOTYPES>
+
+        # this flag is needed when building maya-usd in Maya
+        $<$<BOOL:${IS_WINDOWS}>:WIN32>
```
HamedSabri-adsk, line 54: Without 'WIN32' we get the following compiler error:
```
maya\src\components\usd\maya-usd\lib\mayausd\ufe\ProxyShapeHierarchyHandler.h(39): warning C4275: non dll-interface class 'Ufe_v0::HierarchyHandler' used as base for dll-interface class 'MayaUsd_v0::ufe::ProxyShapeHierarchyHandler'
 ufe-0.2.12\common\include\ufe/hierarchyHandler.h(25): note: see declaration of 'Ufe_v0::HierarchyHandler'
```
This flags was previously set in MayaUSD.cmake via CMAKE_CXX_FLAGS. CMAKE_CXX_FLAGS from other cmake echo systems no longer influence maya-usd.  

There were some compiler flags/defs (e.g _WINDOWS, /GR) which I don't see why is needed.

seando-adsk: The reason I added those three flags in our internal MayaUSD.cmake was to fix some linker errors when compiling the repo from inside our internal Maya build. Those three flags came from comparing the standalone compilation vs the one inside of Maya. At the time I remember searching the maya-usd repo to find where these three flags were added, but I could not find them.

PR #468, opened 2020-04-27
Opt in to diffuse color multidraw consolidation

williamkrick:
This change enables multidraw consolidation for render items using the fallback shader in MayaUSD. 

The Maya API this change relies on is a prototype API being introduced in PR115. It is appropriate for testing only.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -1287,9 +1287,6 @@ HdVP2BasisCurves::_UpdateDrawItem(
         // If available, something changed
         if (stateToCommit._shader != nullptr) {
             renderItem->setShader(stateToCommit._shader);
-#if MAYA_API_VERSION >= 20210000
-            renderItem->setWantDiffuseColorMultiDrawConsolidation(stateToCommit._isFallbackShader);
-#endif
             renderItem->setTreatAsTransparent(stateToCommit._isTransparent);
         }
 
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -108,6 +108,9 @@ namespace {
         //! If true, associate geometric buffers to the render item and trigger consolidation/instancing update
         bool _geometryDirty{ false };
 
+        //! Is _shader a fallback shader
+        bool _isFallbackShader { false };
+
         //! Construct valid commit state
         CommitState(HdVP2DrawItem& item) : _drawItemData(item.GetRenderItemData())
         {}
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -1280,6 +1287,7 @@ HdVP2BasisCurves::_UpdateDrawItem(
         // If available, something changed
         if (stateToCommit._shader != nullptr) {
             renderItem->setShader(stateToCommit._shader);
+            renderItem->setWantDiffuseColorMultiDrawConsolidation(stateToCommit._isFallbackShader);
```
HdC-adsk, line 1290: I guess it is probably ok for now to not enable color consolidation for curves with material assignment, but at some point we will need to do it.

HdC-adsk, line 1290: But... You still need to protect it with MAYA_API_VERSION >= 20210000.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1612,6 +1612,7 @@ MHWRender::MRenderItem* HdVP2Mesh::_CreateSmoothHullRenderItem(const MString& na
     renderItem->castsShadows(true);
     renderItem->receivesShadows(true);
     renderItem->setShader(_delegate->GetFallbackShader(kOpaqueGray));
+    renderItem->setWantDiffuseColorMultiDrawConsolidation(true);
```
HdC-adsk, line 1615: You still need to protect it with MAYA_API_VERSION >= 20210000.

HdC-adsk:
Can we also enable the flag for material-produced shaders?

williamkrick:
> 
> 
> Can we also enable the flag for material-produced shaders?

I am not sure. Part of the problem is some materials were getting the flag that didn't support it. We want to prevent that from happening. Can we test those materials in MayaUSD to see if they have a diffuseColor parameter?

HdC-adsk:
> 
> 
> > Can we also enable the flag for material-produced shaders?
> 
> I am not sure. Part of the problem is some materials were getting the flag that didn't support it. We want to prevent that from happening. Can we test those materials in MayaUSD to see if they have a diffuseColor parameter?

Yes, but I think the API call itself can check the existence of diffuseColor. If not, we could have an explicit lookup with MShaderInstance::parameterList().

williamkrick:
> 
> 
> > > Can we also enable the flag for material-produced shaders?
> > 
> > 
> > I am not sure. Part of the problem is some materials were getting the flag that didn't support it. We want to prevent that from happening. Can we test those materials in MayaUSD to see if they have a diffuseColor parameter?
> 
> Yes, but I think the API call itself can check the existence of diffuseColor. If not, we could have an explicit lookup with MShaderInstance::parameterList().

Yes you are right. Actually the version I'm already working on does a much better job checking that setting the flag is valid and rejecting invalid calls so we should be able to attempt to mark everything as wantMultiDrawConsolidation(). Do you have a suggestion where I can flag those materials off the top of your head? If not I'll dig though and find a place.

HdC-adsk:
My first try would be adding the API call here:

https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/render/vp2RenderDelegate/material.cpp#L701

This is where the shader instance has just been produced from the full material network.

HdC-adsk:
Based on the offline chat, I don't think we have to add color consolidation for material in the same PR. Appoving the PR.

lib/mayaUsd/render/vp2RenderDelegate/render_delegate.cpp
```diff
@@ -273,7 +273,8 @@ namespace
                     shader = shaderMgr->getFragmentShader(
                         _fallbackShaderNames[index], _structOutputName, true);
 #if MAYA_API_VERSION >= 20210000
-                    shader->setWantDiffuseColorMultiDrawConsolidation(true);
+                    MString diffuseColorParamName("diffuseColor");
```
HdC-adsk, line 276: Please use _diffuseColorParameterName which has been defined in the file.

PR #469, opened 2020-04-27
Add support for outline selection hilighting and disabling color-quantization

marsupial:
Outline hi-lighting was added in 22.05, and color-quantization can make things extremely slow.

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/renderGlobals.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/renderGlobals.cpp
@@ -45,6 +45,8 @@ TF_DEFINE_PRIVATE_TOKENS(
     (mtohColorSelectionHighlightColorA)
     (mtohWireframeSelectionHighlight)
     (mtohSelectionOverlay)
+    (mtohColorQuantization)
+    (mtohSelectionOutline)
     (mtohEnableMotionSamples)
     );
 // clang-format on
@@ -113,6 +115,22 @@ void _CreateNumericAttribute(
     node.addAttribute(creator());
 }
 
+template <typename T>
+void _CreateNumericAttribute(
+    MFnDependencyNode& node, const TfToken& attrName, MFnNumericData::Type type,
+    typename std::enable_if<std::is_pod<T>::value, T>::type defValue) {
+    _CreateNumericAttribute(
+        node, attrName, type,
+        [&]() -> MObject {
+            MFnNumericAttribute nAttr;
+            const auto o = nAttr.create(
+                        attrName.GetText(), attrName.GetText(),
+                        type);
+            nAttr.setDefault(defValue);
+            return o;
+        });
+}
+
 void _CreateColorAttribute(
     MFnDependencyNode& node, const TfToken& attrName, const TfToken& attrAName,
     const GfVec4f& defValue) {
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/render/mayaToHydra/renderGlobals.cpp
```diff
@@ -113,6 +115,22 @@ void _CreateNumericAttribute(
     node.addAttribute(creator());
 }
 
+template <typename T>
+void _CreateNumericAttribute(
+    MFnDependencyNode& node, const TfToken& attrName, MFnNumericData::Type type,
+    typename std::enable_if<std::is_pod<T>::value, T>::type defValue) {
```
elrond79, line 121: Definitely approve of this addition (and the usage of enable_if!)

lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
```diff
@@ -450,6 +450,17 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext) {
...
 
+#if USD_VERSION_NUM >= 2005
+    if (_globals.outlineSelectionWidth != 0.f) {
+        _taskController->SetSelectionOutlineRadius(_globals.outlineSelectionWidth);
+        _taskController->SetSelectionEnableOutline(true);
+    } else
+        _taskController->SetSelectionEnableOutline(false);
+#endif
+#if USD_VERSION_NUM >= 1911
+    _taskController->SetColorizeQuantizationEnabled(_globals.enableColorQuantization);
```
elrond79, line 461: So, I actually didn't even know this option existed before... however, looking at the docs for `SetColorizeQuantizationEnabled` in `taskController.h`, I saw:

```
    /// Turns the colorize task color quantization on or off.
    /// XXX: This is a temporary function that will be soon deprecated. Please
    //       avoid calling it.
```

...which makes me a little nervous.  However, if this thing is hurting performance, and there's no other way to disable it, I'm not sure what other option we have...?

lib/mayaUsd/render/mayaToHydra/utils.cpp
```diff
@@ -56,6 +56,18 @@ global proc {{override}}OptionBox() {
...
     attrControlGrp -label "Highlight Color for Selected Objects" -attribute "defaultRenderGlobals.mtohColorSelectionHighlightColor" -changeCommand $cc;
+)mel"
+#if USD_VERSION_NUM >= 2005
+R"mel(
+    attrControlGrp -label "Highlight outline (in pixels, 0 to disable)" -attribute "defaultRenderGlobals.mtohSelectionOutline" -changeCommand $cc;
+)mel"
+#endif
+#if USD_VERSION_NUM >= 1911
+R"mel(
+    attrControlGrp -label "Enable color quantization" -attribute "defaultRenderGlobals.mtohColorQuantization" -changeCommand $cc;
```
elrond79, line 67: Maybe change this to "Enable color quantization for selection-coloring", since that's a bit more descriptive.

Alternatively... I'd be fine if we didn't even make this an option, and just left it always off.

marsupial: The color quantization is on the frame buffer as a whole, it just happens to run in the 'colorizeTask'.
It's in 19.11, 20.02, and 20.05, so not sure how soon it's going to be deprecated.

lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
marsupial: It's in 19.11, 20.02, and 20.05, so not sure how soon it's going to be deprecated, if when it does, can easily ifdef for only those versions it's available?

I believe in 20.05 it became threaded, so it's less painful, but it's horrific before then.

elrond79: Makes sense to me...

lib/mayaUsd/render/mayaToHydra/utils.cpp
elrond79: Oh, really? Yuck... maybe I'm not understanding what this is doing, or why it's there, then - my knowledge of color science / color management is limited.  But I'm not sure in what situations you'd want this on...?

marsupial: It does two things, a quantization/dithering (so that banding is less apparent) and conversion from linear->sRGB.

I imagine tying into Maya's color-management will eventually be the way to handle such things, but until then I'd like an easy way to make sure this is off.

marsupial:
Hmm. Must have back ported that into our 19.11

lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
mattyjams: Just following up on this, as I'm currently reconciling our internal maya-usd branch. This was actually removed as of this core USD commit in the dev branch:
https://github.com/PixarAnimationStudios/USD/commit/584c27d495047aa9a7231df64eec9e22396e5651

I just opened #481 to restrict calling this function to USD versions between 19.11 and 20.05.

PR #470, opened 2020-04-27
Fix leak of UsdImagingDelegate

marsupial:
There's an issue with the following, when a viewport triggers a display before the second command is run:
```
string $proxy = `createNode mayaUsdProxyShape`;
setAttr ($proxy + ".filePath") <path>
```

### DIFF
--- a/lib/usd/hdMaya/adapters/proxyAdapter.cpp
+++ b/lib/usd/hdMaya/adapters/proxyAdapter.cpp
@@ -143,13 +143,13 @@ void HdMayaProxyAdapter::PopulateSelectedPaths(
 }
 
 void HdMayaProxyAdapter::CreateUsdImagingDelegate() {
-    // Why do this release when we do a reset right below? Because we want
+    // Why do this reset when we do another right below? Because we want
     // to make sure we delete the old delegate before creating a new one
     // (the reset statement below will first create a new one, THEN delete
     // the old one). Why do we care? In case they have the same _renderIndex
     // - if so, the delete may clear out items from the renderIndex that the
     // constructor potentially adds
-    _usdDelegate.release();
+    _usdDelegate.reset();
     _usdDelegate.reset(new HdMayaProxyUsdImagingDelegate(
         &GetDelegate()->GetRenderIndex(),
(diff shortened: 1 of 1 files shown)


lib/usd/hdMaya/adapters/proxyAdapter.cpp
```diff
@@ -143,13 +149,13 @@ void HdMayaProxyAdapter::PopulateSelectedPaths(
...
 void HdMayaProxyAdapter::CreateUsdImagingDelegate() {
-    // Why do this release when we do a reset right below? Because we want
+    // Why do this reset when we do another right below? Because we want
     // to make sure we delete the old delegate before creating a new one
     // (the reset statement below will first create a new one, THEN delete
     // the old one). Why do we care? In case they have the same _renderIndex
     // - if so, the delete may clear out items from the renderIndex that the
     // constructor potentially adds
-    _usdDelegate.release();
+    _usdDelegate.reset();
```
elrond79, line 158: Oops, good catch!

kxl-adsk:
@marsupial We are making changes around this code as we speak. One change will be to proxy to always create a stage...additional changes you can review here #474. Let's revisit this PR once the above changes land in dev.

seando-adsk:
I'll wait to review this PR until after 474 goes in.

marsupial:
Is it worth splitting off the leak-fix (`_usdDelegate.release();` => `_usdDelegate.reset();`) into t a separate PR or does that just make things more complicated ?

kxl-adsk:
@marsupial I don't think it's worth doing at this point. Having said that, it's a good practice to not bundle separate issues in one PR.

BTW. I would like #478 to go in before you revisit your fixes. You will notice that we will no longer have a proxy shape without a stage.

kxl-adsk:
@marsupial all changes are now in dev, can you please rebase this work and adapt to the new behavior of proxy shapes with the empty file path.

marsupial:
Just moving this to fix the leak. Looks like the original issue is gone.

kxl-adsk:
Perfect, thank you.

PR #471, opened 2020-04-27
Teach the import cmd ui how to import usd-assets and apply to proxy-shapes

marsupial:
Current code assumes an on-disk 'usd' file, which blocks possible ArAssetResolver from running.
Adds another `mayaUsd_createStageFromAsset` mel procedure to open the UI and apply settings to a proxy.

### DIFF
--- a/lib/usd/ui/USDImportDialogCmd.cpp
+++ b/lib/usd/ui/USDImportDialogCmd.cpp
@@ -86,13 +86,10 @@ MStatus USDImportDialogCmd::applyToProxy(const MString& proxyPath)
 		return status;
 
 	MObject proxyShapeObj = proxyShapeDagPath.node(&status);
-	CHECK_MSTATUS_AND_RETURN(status, status);
-	if (status.error())
-		return status;
+	CHECK_MSTATUS_AND_RETURN_IT(status);
 
     MFnDependencyNode fn(proxyShapeObj, &status);
-	if (status.error())
-		return status;
+	CHECK_MSTATUS_AND_RETURN_IT(status);
 
 	if (fn.typeName() != MString("mayaUsdProxyShape"))
 		return MS::kInvalidParameter;
@@ -102,20 +99,17 @@ MStatus USDImportDialogCmd::applyToProxy(const MString& proxyPath)
 		return MS::kInvalidParameter;
 
 	MPlug primPath = fn.findPlug("primPath", &status);
-	if (status.error())
-		return status;
+	CHECK_MSTATUS_AND_RETURN_IT(status);
+
 	MPlug filePath = fn.findPlug("filePath", &status);
-	if (status.error())
-		return status;
+	CHECK_MSTATUS_AND_RETURN_IT(status);
 
 	ImportData& importData = ImportData::instance();
 	primPath.setValue(MString(importData.rootPrimPath().c_str()));
-	if (status.error())
-		return status;
+	CHECK_MSTATUS_AND_RETURN_IT(status);
 
 	filePath.setValue(MString(importData.filename().c_str()));
-	if (status.error())
-		return status;
+	CHECK_MSTATUS_AND_RETURN_IT(status);
 
 	auto rootPrim = proxyShape->usdPrim();
 	if (!rootPrim)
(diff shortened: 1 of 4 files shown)


kxl-adsk:
@marsupial I think you merged two changes into one PR:
1) fix for ArAssetResolver
2) new workflow leveraging import dialog

Assuming I got it right, how about we split this change into two PRs? I know @wtelford may have some feedback about the workflow and we may need to iterate a bit more on it...while it shouldn't stop fixing 1).

wtelford:
@marsupial Can you provide more details about the mayaUsd_createStageFromAsset change and its behavior? Screenshots would be great.
Is this a change to the import dialog (translation of USD to Maya data), or to the proxyShape creation workflow (Create Stage From File)?

marsupial:
The fix for ArAssetResolver isn't really worth much without the other, because it would still be requiring on-disk assets for load due to the invocation via (Create Stage From File).

**mayaUsd_createStageFromAsset** is a simple procedure that when given an asset-path (from an external source/browser), will invoke the same UI for picking variants that is shown from (Create Stage From File), then create the proxy node, name it, and apply the user's variant selections.

Essentially it's similar to **Create Stage From File**, but for external content management where the/a file browser isn't wanted.

kxl-adsk:
@marsupial we discussed this change internally. We are going to accept this change (of course we will still do the code review), but keep in mind we have plans to leverage 'tree view' dialog in proxy creation workflows. At that point, we may change some of the interfaces and/or behaviors you are introducing in this PR.

marsupial:
Totally fine/expect things will move forward (or even change significantly), we just need aspects of this functionality now to even evaluate, and aligning these goals so that such enhancements aren't living solely over here seemed good.

lib/usd/ui/USDImportDialogCmd.cpp
```diff
@@ -66,6 +76,69 @@ void* USDImportDialogCmd::creator()
...
+	MSelectionList selection;
+	selection.add(proxyPath);
+	MStatus status = selection.getDagPath(0, proxyShapeDagPath);
+	if (status.error())
+		return status;
+
+	MObject proxyShapeObj = proxyShapeDagPath.node(&status);
+	CHECK_MSTATUS_AND_RETURN(status, status);
+	if (status.error())
+		return status;
```
seando-adsk, line 91: Here you can use CHECK_MSTATUS_AND_RETURN_IT(status) and you don't need the extra check for .error() since the macro returns if status != kSuccess.

lib/usd/ui/USDImportDialogCmd.cpp
```diff
@@ -93,30 +166,55 @@ MStatus USDImportDialogCmd::doIt(const MArgList& args)
...
-				// Don't set the stage pop mask until we solve how to use it together with
-				// the root prim path.
-				//importData.setStagePopulationMask(usdImportDialog->stagePopulationMask());
-				importData.setPrimVariantSelections(usdImportDialog->primVariantSelections());
-			}
+			std::unique_ptr<IUSDImportView> usdImportDialog(
+				new USDImportDialog(assetPath.asChar(), &importData,
+					                usdQtUtil, MQtUtil::mainWindow()));
+			if (!usdImportDialog->execute())
+				return MS::kFailure;
```
seando-adsk, line 206: According to the doc for the execute() interface method: "True if the user applied the changes, false if they canceled." We do not want to return an error from the command when they cancel. Thus the previous if check (to set the importData) and then always return MS::kSuccess.

lib/usd/ui/USDImportDialogCmd.cpp
```diff
@@ -93,30 +166,55 @@ MStatus USDImportDialogCmd::doIt(const MArgList& args)
 		return MS::kSuccess;
 	}
 
+	if(argData.isFlagSet(kApplyToProxyFlag))
```
seando-adsk, line 169: Can you add a comment here that says when using this flag, no command object is expected.

lib/usd/ui/USDImportDialogCmd.h
```diff
@@ -25,6 +25,8 @@
 MAYAUSD_NS_DEF {
 
 class MAYAUSD_UI_PUBLIC USDImportDialogCmd : public MPxCommand {
+    MStatus applyToProxy(const MString& proxyPath);
```
seando-adsk, line 28: I would prefer if you created a new "private" method section after line 42 with this method. This matches how the rest of this project was coded. Thanks.

PR #474, opened 2020-04-29
MAYA-104544 Handle changing the USD stage

williamkrick:
The cached usdStage in the proxyRenderDelegate can change. If that happens delete everything cached in the ProxyRenderDelegate and clear the subscene container so VP2 draws correctly.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -758,8 +758,7 @@ MayaUsdProxyShapeBase::preEvaluation(const MDGContext& context, const MEvaluatio
         evaluationNode.dirtyPlugExists(filePathAttr) ||
         evaluationNode.dirtyPlugExists(primPathAttr) ||
         evaluationNode.dirtyPlugExists(loadPayloadsAttr) ||
-        evaluationNode.dirtyPlugExists(inStageDataAttr) ||
-        evaluationNode.dirtyPlugExists(inStageDataCachedAttr)) {
+        evaluationNode.dirtyPlugExists(inStageDataAttr)) {
         _IncreaseUsdStageVersion();
     }
 
@@ -769,11 +768,11 @@ MayaUsdProxyShapeBase::preEvaluation(const MDGContext& context, const MEvaluatio
 MStatus
 MayaUsdProxyShapeBase::postEvaluation(const  MDGContext& context, const MEvaluationNode& evaluationNode, PostEvaluationType evalType)
 {
-    // If/when the MPxDrawOverride for the proxy shape specifies
-    // isAlwaysDirty=false to improve performance, we must be sure to notify
-    // the Maya renderer that the geometry is dirty and needs to be redrawn
-    // when any plug on the proxy shape is dirtied.
-    MHWRender::MRenderer::setGeometryDrawDirty(thisMObject());
+    // When a node is evaluated by evaluation manager setDependentsDirty is not called. The functionality in setDependentsDirty needs
+    // to be duplicated in preEvaluation or postEvaluation. I don't think we need the call to setGeometryDrawDirty() in
+    // setDependentsDirty, but there may be a workflow I'm not seeing that does require it. I'm leaving this here commented out as a
+    // reminder that we should either have both calls to setGeometryDrawDirty, or no calls to setGeometryDrawDirty.
+    // MHWRender::MRenderer::setGeometryDrawDirty(thisMObject());
 
     return MStatus::kSuccess;
 }
@@ -796,19 +795,11 @@ MayaUsdProxyShapeBase::setDependentsDirty(const MPlug& plug, MPlugArray& plugArr
         plug == filePathAttr ||
         plug == primPathAttr ||
         plug == loadPayloadsAttr ||
-        plug == inStageDataAttr ||
-        plug == inStageDataCachedAttr) {
+        plug == inStageDataAttr) {
         _IncreaseUsdStageVersion();
     }
 
     return MPxSurfaceShape::setDependentsDirty(plug, plugArray);
-}
-
-/* virtual */
-bool
-MayaUsdProxyShapeBase::setInternalValue(const MPlug& plug, const MDataHandle& dataHandle)
-{
-    return MPxSurfaceShape::setInternalValue(plug, dataHandle);
 }
 
 UsdPrim
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -48,6 +48,7 @@
 #include <maya/MString.h>
 #include <maya/MTime.h>
 #include <maya/MViewport2Renderer.h>
+#include <maya/MEvaluationNode.h>
 
 #include <pxr/base/gf/bbox3d.h>
 #include <pxr/base/gf/range3d.h>
@@ -746,6 +747,37 @@ MayaUsdProxyShapeBase::isStageValid() const
     return true;
 }
 
+MStatus
+MayaUsdProxyShapeBase::preEvaluation(const MDGContext& context, const MEvaluationNode& evaluationNode)
+{
+    if (evaluationNode.dirtyPlugExists(excludePrimPathsAttr)) {
+        _IncreaseExcludePrimPathsVersion();
+    }
+    else if (evaluationNode.dirtyPlugExists(outStageDataAttr) ||
+        // All the plugs that affect outStageDataAttr
+        evaluationNode.dirtyPlugExists(filePathAttr) ||
+        evaluationNode.dirtyPlugExists(primPathAttr) ||
+        evaluationNode.dirtyPlugExists(loadPayloadsAttr) ||
+        evaluationNode.dirtyPlugExists(inStageDataAttr) ||
+        evaluationNode.dirtyPlugExists(inStageDataCachedAttr)) {
+        _IncreaseUsdStageVersion();
+    }
+
+    return MStatus::kSuccess;
+}
+
+MStatus
+MayaUsdProxyShapeBase::postEvaluation(const  MDGContext& context, const MEvaluationNode& evaluationNode, PostEvaluationType evalType)
+{
+    // If/when the MPxDrawOverride for the proxy shape specifies
+    // isAlwaysDirty=false to improve performance, we must be sure to notify
+    // the Maya renderer that the geometry is dirty and needs to be redrawn
+    // when any plug on the proxy shape is dirtied.
+    MHWRender::MRenderer::setGeometryDrawDirty(thisMObject());
+
+    return MStatus::kSuccess;
+}
+
 /* virtual */
 MStatus
 MayaUsdProxyShapeBase::setDependentsDirty(const MPlug& plug, MPlugArray& plugArray)
@@ -755,6 +787,20 @@ MayaUsdProxyShapeBase::setDependentsDirty(const MPlug& plug, MPlugArray& plugArr
     // the Maya renderer that the geometry is dirty and needs to be redrawn
     // when any plug on the proxy shape is dirtied.
     MHWRender::MRenderer::setGeometryDrawDirty(thisMObject());
+
+    if (plug == excludePrimPathsAttr) {
+        _IncreaseExcludePrimPathsVersion();
+    }
+    else if (plug == outStageDataAttr ||
+        // All the plugs that affect outStageDataAttr
+        plug == filePathAttr ||
+        plug == primPathAttr ||
+        plug == loadPayloadsAttr ||
+        plug == inStageDataAttr ||
+        plug == inStageDataCachedAttr) {
+        _IncreaseUsdStageVersion();
+    }
+
     return MPxSurfaceShape::setDependentsDirty(plug, plugArray);
 }
 
@@ -762,9 +808,6 @@ MayaUsdProxyShapeBase::setDependentsDirty(const MPlug& plug, MPlugArray& plugArr
 bool
 MayaUsdProxyShapeBase::setInternalValue(const MPlug& plug, const MDataHandle& dataHandle)
 {
-    if (plug == excludePrimPathsAttr) {
-        _IncreaseExcludePrimPathsVersion();
-    }
     return MPxSurfaceShape::setInternalValue(plug, dataHandle);
 }
 
@@ -843,6 +886,11 @@ MayaUsdProxyShapeBase::getUsdStage() const
     else
         return {};
 
+}
+
+size_t
+MayaUsdProxyShapeBase::getUsdStageVersion() const {
+    return _UsdStageVersion;
 }
 
 SdfPathVector
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -247,8 +247,6 @@ void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
     UsdStageRefPtr newUsdStage = _proxyShape->getUsdStage();
     if (_usdStage != newUsdStage)
     {
-        _ClearRenderDelegate();
-
         // delete everything so we stop drawing the old stage and draw the new one
         _ClearRenderDelegate();
         _dummyTasks.clear();
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -192,14 +192,17 @@ MHWRender::MPxSubSceneOverride* ProxyRenderDelegate::Creator(const MObject& obj)
 ProxyRenderDelegate::ProxyRenderDelegate(const MObject& obj)
 : MHWRender::MPxSubSceneOverride(obj)
 {
-    MDagPath::getAPathTo(obj, _proxyDagPath);
+    MDagPath proxyDagPath;
+    MDagPath::getAPathTo(obj, proxyDagPath);
 
     const MFnDependencyNode fnDepNode(obj);
-    _proxyShape = static_cast<MayaUsdProxyShapeBase*>(fnDepNode.userNode());
+    _proxyShapeData.reset(new ProxyShapeData(static_cast<MayaUsdProxyShapeBase*>(fnDepNode.userNode()), proxyDagPath));
 }
 
 //! \brief  Destructor
 ProxyRenderDelegate::~ProxyRenderDelegate() {
+    _ClearRenderDelegate();
+
 #if !defined(WANT_UFE_BUILD)
     if (_mayaSelectionCallbackId != 0) {
         MMessage::removeCallback(_mayaSelectionCallbackId);
@@ -227,8 +230,6 @@ bool ProxyRenderDelegate::requiresUpdate(const MSubSceneContainer& container, co
 void ProxyRenderDelegate::_ClearRenderDelegate()
 {
     // The order of deletion matters. Some orders cause crashes.
-    // This order matches the deletion order when a ProxyRenderDelegate is destroyed.
-    // Class member variables are destroyed in the reverse order in which they appear in the class declaration.
 
     _sceneDelegate.reset();
     _taskController.reset();
@@ -239,17 +240,17 @@ void ProxyRenderDelegate::_ClearRenderDelegate()
 //! \brief  One time initialization of this drawing routine
 void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
 
-    if (_proxyShape == nullptr)
+    if (_proxyShapeData->ProxyShape() == nullptr)
         return;
 
-    if (!_proxyShapeData.isUsdStageUpToDate(*_proxyShape))
+    if (!_proxyShapeData->IsUsdStageUpToDate())
     {
         // delete everything so we stop drawing the old stage and draw the new one
         _ClearRenderDelegate();
         _dummyTasks.clear();
         container.clear();
 
-        _usdStage = _proxyShape->getUsdStage();
+        _proxyShapeData->UpdateUsdStage();
     }
     
     // No need to run all the checks if we got till the end
@@ -286,8 +287,8 @@ void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
             TfMakeValidIdentifier(
                 TfStringPrintf(
                     "Proxy_%s_%p",
-                    _proxyShape->name().asChar(),
-                    _proxyShape));
+                    _proxyShapeData->ProxyShape()->name().asChar(),
+                    _proxyShapeData->ProxyShape()));
         const SdfPath delegateID =
             SdfPath::AbsoluteRootPath().AppendChild(TfToken(delegateName));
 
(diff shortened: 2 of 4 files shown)


williamkrick:
The changes I've done aren't perfect. In some cases when I load a new usd stage some of the old stage keeps on drawing in the viewport. Looking at the documentation for MPxSubSceneOverride it seems like clearing the MSubSceneContainer should be enough to prevent anything from the old stage from drawing, so I'm guessing there is still a bug in Maya which is causing that problem.

I'll wait until @HdC-adsk makes some progress on the drawing issues he's looking at before tackling this additional issue, in case they are duplicates.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -229,18 +226,42 @@ bool ProxyRenderDelegate::requiresUpdate(const MSubSceneContainer& container, co
...
 
-    if (!_usdStage) {
-        _usdStage = _proxyShape->getUsdStage();
+    UsdStageRefPtr newUsdStage = _proxyShape->getUsdStage();
+    if (_usdStage != newUsdStage)
+    {
+        _ClearRenderDelegate();
+
+        // delete everything so we stop drawing the old stage and draw the new one
+        _ClearRenderDelegate();
```
kxl-adsk, line 253: you don't have to do it twice...see line 250.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -229,18 +226,42 @@ bool ProxyRenderDelegate::requiresUpdate(const MSubSceneContainer& container, co
...
+        _dummyTasks.clear();
+
+        _isPopulated = false;
+        container.clear();
```
kxl-adsk, line 257: You are saying this is not enough to clear the subscene state?

williamkrick: According to the API tech docs tt is supposed to be enough to clear the subscene but it doesn't work correctly. Sometimes things get leftover and I have to ```ogs -reset``` to get rid of them all.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
williamkrick: Oops, yes once should be enough. I made the helper method because the deletion order matters, if we use the wrong order we can crash deleting the renderIndex. I ended up copying the ordering from the destructor and that seems to work fine.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
HdC-adsk: Maybe that reveals some bugs in VP2?

HdC-adsk: Just saw the other comment of yours above. I believe the correctness issue I am looking at are not a duplicate of this one. See my comments in JIRA for details.

williamkrick: Yes I think there is something to fix in VP2. Maybe related to what you are already looking at @HdC-adsk. I'll put notes on which scenes I used to reproduce the issue in the JIRA story so we can validate it after you have a fix.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -229,18 +226,40 @@ bool ProxyRenderDelegate::requiresUpdate(const MSubSceneContainer& container, co
     return true;
 }
 
+void ProxyRenderDelegate::_ClearRenderDelegate()
+{
+    delete _sceneDelegate;
```
HdC-adsk, line 231: Instead of using the plain pointers as data member, a better way is to use unique_ptr.

Then this manual cleanup can be avoided and data members in the proxyRenderDelegate class can be declared consistently.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
HdC-adsk: I think we commented at the same time? The issue I am looking at is caused by transparent instances, so I don't think it is the same issue as leftover ghost objects.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -229,32 +224,54 @@ bool ProxyRenderDelegate::requiresUpdate(const MSubSceneContainer& container, co
...
 
     if (!_renderIndex) {
         MProfilingScope subProfilingScope(HdVP2RenderDelegate::sProfilerCategory,
             MProfiler::kColorD_L1, "Allocate RenderIndex");
 #if USD_VERSION_NUM > 2002
-        _renderIndex = HdRenderIndex::New(_renderDelegate, HdDriverVector());
+        _renderIndex.reset(HdRenderIndex::New(_renderDelegate.get(), HdDriverVector()));
 #else
-        _renderIndex = HdRenderIndex::New(_renderDelegate);
+        _renderIndex.reset(HdRenderIndex::New(_renderDelegate));
```
HdC-adsk, line 274: This is a bit different from Line 272. Is it intentional?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -229,32 +224,54 @@ bool ProxyRenderDelegate::requiresUpdate(const MSubSceneContainer& container, co
     return true;
 }
 
+void ProxyRenderDelegate::_ClearRenderDelegate()
+{
+    // The order of deletion matters. Some orders cause crashes.
+    // This order matches the deletion order when a ProxyRenderDelegate is destroyed.
+    // Class member variables are destroyed in the reverse order in which they appear in the class declaration.
+
+    _sceneDelegate.reset(nullptr);
```
kxl-adsk, line 233: I would expect that you don't have to be explicit about `nullptr` here and can just call `reset()`

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -229,32 +224,54 @@ bool ProxyRenderDelegate::requiresUpdate(const MSubSceneContainer& container, co
...
-        return;
+void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
 
     if (_proxyShape == nullptr)
         return;
 
-    if (!_usdStage) {
-        _usdStage = _proxyShape->getUsdStage();
+    UsdStageRefPtr newUsdStage = _proxyShape->getUsdStage();
+    if (_usdStage != newUsdStage)
```
kxl-adsk, line 246: I was thinking more about this. We currently have two code which needs to re-populate the render index:
- file path change
- exclude prim paths change

If you look at the code in _Populate method, you will notice `_proxyShape->getExcludePrimPathsVersion() != _excludePrimPathsVersion`. I believe this code no longer works as expected and we will have to do a very similar thing when excluded prims change. 

How about we merge the two together and instead of testing the stage at line 246, you will test `stageVersion`? Then in proxyShapeBase, we would bump the stageVersion on excluded prims change (basically rename _IncreaseExcludePrimPathsVersion to _IncreaseStageVersion) and from https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/nodes/proxyShapeBase.cpp#L531

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
williamkrick: No, I didn't test against building against an USD with version num <= 2002.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
williamkrick: Yes I think it does make sense to combine those two ideas. Is the MayaUsdProxyShapeBase a class that AL and Pixar derive from but I don't see them making any calls to getExcludePrimPathsVersion(), so we should be good to change this functionality a little bit without much impact on them.

In the future if there is a scenario where we'd want more fine grained information about what changed (the exclude prim paths vs. the stage) we could split the version number into two, one for exclude paths and one for the stage. Right now we'd be doing the same work either way so combining them should work.

williamkrick: I went in a started making the changes and I learned a few new things.

1. excludePrimPathsAttr doesn't affect outStageDataAttr, so combing changes to excludePrimPathsAttr with changes to outStageDataAttr into a single version counter implies some knowledge of the consumers of that data. In our case we know that either one changing will require us to call populate again, but MayaUsdProxyShapeBase doesn't know that.

1. There _are_ some other attributes that do affect what ends up in outStageDataAttr: filePathAttr as we already identified, but loadPayloadsAttr also impacts it. And kind of indirectly primPathAttr.

1. primPathAttr is a little different because the data in outStageDataAttr is a MayaUsdStageData which has two values, the UsdStage & the SdfPath prim path.

I think in MayaUsdProxyShapeBase it is wrong to combine dirty information because that implies knowledge of the downstream consumers of that information, and we don't control all the downstream consumers. That means that for example the dirty counter _ExcludePrimPathVersion is fine, but a combined counter for _IncreaseStageVersion that increments when excludePrimPathsAttr changes is not fine.

We need separate version counters for all the output data on MayaUsdProxyShapeBase. Those counters can be on the proxy shape itself, or they can be in the render delegates (we'd have to create a mixin class with the functionality so customers could re-use it). I think it's probably simpler to just put it on the ProxyShapeBase.

We could also put the version counters for the stage and primPath inside the MayaUsdStageData.

What I will do is for each output attribute create the version counter on MayaUsdProxyShapeBase and set them up to increment properly. That way everyone can use them. Then in ProxyRenderDelegate we'll have to track all the version ids we care about to do the updates when necessary. Once we have that basic version in the PR we can decide what tweaks we want to the system.

williamkrick: We don't need version counters for simple numeric types which are easy to store & compare vs. their old value.

williamkrick: The new version is going the direction I think we should take.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
williamkrick: I agree I think some items are getting lost in consolidated world. I'll log an internal issue to investigate. It could be caused by our recent changes.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -147,20 +147,33 @@ class ProxyRenderDelegate : public MHWRender::MPxSubSceneOverride
...
-    size_t              _excludePrimPathsVersion{ 0 }; //!< Last version of exluded prims used during render index populate
-
-    bool                _isPopulated{ false };      //!< If false, scene delegate wasn't populated yet within render index
-    bool                _selectionChanged{ false }; //!< Whether there is any selection change or not
-    bool                _isProxySelected{ false };  //!< Whether the proxy shape is selected
-    MColor              _wireframeColor;            //!< Wireframe color assigned to the proxy shape
+    // We need to track when output data on the proxy shape changes. For simple numeric types we cache the last value
+    // read from _proxyShape. For complicated types we keep a version number of the last value we read to make
+    // comparisons fast.
+    class ProxyShapeData
```
HdC-adsk, line 161: The class name is probably a little bit confusing. How about ProxyShapeVersion?

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -746,6 +747,37 @@ MayaUsdProxyShapeBase::isStageValid() const
...
 
+MStatus
+MayaUsdProxyShapeBase::preEvaluation(const MDGContext& context, const MEvaluationNode& evaluationNode)
+{
+    if (evaluationNode.dirtyPlugExists(excludePrimPathsAttr)) {
+        _IncreaseExcludePrimPathsVersion();
+    }
+    else if (evaluationNode.dirtyPlugExists(outStageDataAttr) ||
+        // All the plugs that affect outStageDataAttr
+        evaluationNode.dirtyPlugExists(filePathAttr) ||
```
kxl-adsk, line 758: I understand that all of these attributes can be connected/animated?

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -746,6 +747,37 @@ MayaUsdProxyShapeBase::isStageValid() const
...
+}
+
+MStatus
+MayaUsdProxyShapeBase::postEvaluation(const  MDGContext& context, const MEvaluationNode& evaluationNode, PostEvaluationType evalType)
+{
+    // If/when the MPxDrawOverride for the proxy shape specifies
+    // isAlwaysDirty=false to improve performance, we must be sure to notify
+    // the Maya renderer that the geometry is dirty and needs to be redrawn
+    // when any plug on the proxy shape is dirtied.
+    MHWRender::MRenderer::setGeometryDrawDirty(thisMObject());
```
kxl-adsk, line 776: I'm not sure why we need this? Some attributes (like input time) have `setAffectsAppearance` which I believe is what was causing the refresh.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -755,16 +787,27 @@ MayaUsdProxyShapeBase::setDependentsDirty(const MPlug& plug, MPlugArray& plugArr
     // the Maya renderer that the geometry is dirty and needs to be redrawn
     // when any plug on the proxy shape is dirtied.
     MHWRender::MRenderer::setGeometryDrawDirty(thisMObject());
+
+    if (plug == excludePrimPathsAttr) {
+        _IncreaseExcludePrimPathsVersion();
+    }
+    else if (plug == outStageDataAttr ||
+        // All the plugs that affect outStageDataAttr
+        plug == filePathAttr ||
```
kxl-adsk, line 796: FYI @seando-adsk

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
williamkrick: I'm thinking about moving the proxyShape pointer, the _usdStage and anything else we get directly from the proxyShape into this class. What do you think about that idea?

HdC-adsk: I agree.

lib/mayaUsd/nodes/proxyShapeBase.cpp
williamkrick: I don't know why we need it either which is why I don't want to remove it. Not sure which workflows need testing. setDependentsDirty has it and if the node goes into EM setDependentsDirty won't get called, so this is more correct. I think probably we need to figure out when exactly we need to do the refreshes but the change is large enough already I didn't want to add that investigation to it.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -144,20 +145,35 @@ class ProxyRenderDelegate : public MHWRender::MPxSubSceneOverride
...
-
-    bool                _isPopulated{ false };      //!< If false, scene delegate wasn't populated yet within render index
-    bool                _selectionChanged{ false }; //!< Whether there is any selection change or not
-    bool                _isProxySelected{ false };  //!< Whether the proxy shape is selected
-    MColor              _wireframeColor;            //!< Wireframe color assigned to the proxy shape
+    HdEngine                _engine;                    //!< Hydra engine responsible for running synchronization between scene delegate and VP2RenderDelegate
+    HdTaskSharedPtrVector   _dummyTasks;                //!< Dummy task to bootstrap data preparation inside Hydra engine
+    UsdStageRefPtr          _usdStage;              //!< USD stage pointer
+
+    // Do not re-order these 4 member variables relative to one another. The order they are declared here is the reverse order of deletion, and the deletion order matters!
```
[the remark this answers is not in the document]
williamkrick: Yes I can do that. Then the unique_ptr is really just a reminder that proxyRenderDelegate owns those pointers vs. actually doing anything, but I think improving the clarity of the code is still valuable.

lib/mayaUsd/nodes/proxyShapeBase.cpp
williamkrick: inStageDataCachedAttr being not writeable is enough that it can't be animated and so could be removed from this list.

lib/mayaUsd/nodes/proxyShapeBase.cpp
kxl-adsk: The flip side is - nobody will know why it was added. It's a rather simple change to make, so let's wait for repro case.

williamkrick: I still prefer it there. If you want I will comment out the call and leave a comment explaining why I added it.

kxl-adsk: Documenting why we think it should be there (i.e. symmetry with setDependentsDirty) and waiting for confirmation before we enable the code sounds reasonable to me.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -141,35 +141,38 @@ class ProxyRenderDelegate : public MHWRender::MPxSubSceneOverride
     void _FilterSelection();
     void _UpdateSelectionStates();
 
-    const MayaUsdProxyShapeBase*  _proxyShape{ nullptr }; //!< DG proxy shape node
-    MDagPath                      _proxyDagPath;          //!< DAG path of the proxy shape (assuming no DAG instancing)
+    // We need to track when output data on the proxy shape changes. For simple numeric types we cache the last value
```
HdC-adsk, line 144: Now the class doc looks not matching what the whole class is.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -141,35 +141,38 @@ class ProxyRenderDelegate : public MHWRender::MPxSubSceneOverride
...
+    // read from _proxyShape. For complicated types we keep a version number of the last value we read to make
+    // comparisons fast.
+    class ProxyShapeData
+    {
+        const MayaUsdProxyShapeBase*    _proxyShape{ nullptr };     //!< DG proxy shape node
```
HdC-adsk, line 149: Can it be `const MayaUsdProxyShapeBase* const`?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -141,35 +141,38 @@ class ProxyRenderDelegate : public MHWRender::MPxSubSceneOverride
...
+        MDagPath                        _proxyDagPath;              //!< DAG path of the proxy shape (assuming no DAG instancing)
+        UsdStageRefPtr                  _usdStage;                  //!< USD stage pointer
+        size_t                          _excludePrimsVersion{ 0 };  //!< Last version of exluded prims used during render index populate
+        size_t                          _usdStageVersion{ 0 };      //!< Last version of stage used during render index populate
+    public:
+        ProxyShapeData(const MayaUsdProxyShapeBase* proxyShape, const MDagPath& proxyDatPath);
```
HdC-adsk, line 155: Typo: `proxyDatPath` -> `proxyDagPath`

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -141,35 +141,38 @@ class ProxyRenderDelegate : public MHWRender::MPxSubSceneOverride
...
 
-    const MayaUsdProxyShapeBase*  _proxyShape{ nullptr }; //!< DG proxy shape node
-    MDagPath                      _proxyDagPath;          //!< DAG path of the proxy shape (assuming no DAG instancing)
+    // We need to track when output data on the proxy shape changes. For simple numeric types we cache the last value
+    // read from _proxyShape. For complicated types we keep a version number of the last value we read to make
+    // comparisons fast.
+    class ProxyShapeData
+    {
+        const MayaUsdProxyShapeBase*    _proxyShape{ nullptr };     //!< DG proxy shape node
+        MDagPath                        _proxyDagPath;              //!< DAG path of the proxy shape (assuming no DAG instancing)
```
HdC-adsk, line 150: Can it be `const MDagPath`?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -729,17 +730,27 @@ const MColor& ProxyRenderDelegate::GetWireframeColor() const
...
 // ProxyShapeData
-bool ProxyRenderDelegate::ProxyShapeData::isUsdStageUpToDate(const MayaUsdProxyShapeBase& proxyShape) const {
-    return proxyShape.getUsdStageVersion() == _usdStageVersion;
+ProxyRenderDelegate::ProxyShapeData::ProxyShapeData(const MayaUsdProxyShapeBase* proxyShape, const MDagPath& proxyDagPath)
+    : _proxyShape(proxyShape)
+    , _proxyDagPath(proxyDagPath)
+{
+    assert(_proxyShape);
+}
+const MayaUsdProxyShapeBase* ProxyRenderDelegate::ProxyShapeData::ProxyShape() const { return _proxyShape; }
```
HdC-adsk, line 739: Can these functions be inline?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -736,20 +736,20 @@ ProxyRenderDelegate::ProxyShapeData::ProxyShapeData(const MayaUsdProxyShapeBase*
 {
     assert(_proxyShape);
 }
-const MayaUsdProxyShapeBase* ProxyRenderDelegate::ProxyShapeData::ProxyShape() const { return _proxyShape; }
-const MDagPath& ProxyRenderDelegate::ProxyShapeData::ProxyDagPath() const { return _proxyDagPath; }
-UsdStageRefPtr ProxyRenderDelegate::ProxyShapeData::UsdStage() const { return _usdStage; }
-void ProxyRenderDelegate::ProxyShapeData::UpdateUsdStage() { _usdStage = _proxyShape->getUsdStage(); }
-bool ProxyRenderDelegate::ProxyShapeData::IsUsdStageUpToDate() const {
+inline const MayaUsdProxyShapeBase* ProxyRenderDelegate::ProxyShapeData::ProxyShape() const { return _proxyShape; }
```
HdC-adsk, line 739: I assume these inline functions won't be called from other cpp files?

Info from https://isocpp.org/wiki/faq/inline-functions#overview-inline-fns

> The reason you (almost always) put the definition (the {...} part) of an inline function in a header file is to avoid “unresolved external” errors from the linker. That error will occur if you put the inline function’s definition in a .cpp file and if that function is called from some other .cpp file.

williamkrick: Yes no other cpp files need them. If we wanted to use ProxyShapeData in some other class we'd have to move them to the header but we're not doing that right now, and this avoids head bloat of include proxyShapeBash.h in proxyRenderDelegate.h.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -140,24 +137,46 @@ class ProxyRenderDelegate : public MHWRender::MPxSubSceneOverride
...
+        const MayaUsdProxyShapeBase* ProxyShape() const;
+        const MDagPath& ProxyDagPath() const;
+        UsdStageRefPtr UsdStage() const;
+        void UpdateUsdStage();
+        bool IsUsdStageUpToDate() const;
+        void UsdStageUpdated();
+        bool IsExcludePrimsUpToDate() const;
+        void ExcludePrimsUpdated();
+    };
+    std::unique_ptr<ProxyShapeData>          _proxyShapeData;
```
kxl-adsk, line 165: You seem to assume everywhere that proxyShapeData exists...so why to make it dynamically allocated?

williamkrick: I want to have a constructor to set the proxyShape pointer and MDagPath pointer. Those things need to get pulled out of the MObject passed into the ProxyRenderDelegate constructor. Basically if I made it a member I'd need to have the body of the ProxyRenderDelegate constructor appear in it's member initializers or give up the const-ness of some members of ProxyShapeData.

PR #475, opened 2020-04-29
MAYA-104215: improve renaming restriction on prim objects

HamedSabri-adsk:
### PR Changes:
This PR brings new improvements to rename command and adds new restrictions:

- Reworked rename command implementation.
- Only allow edits on the target layer. error out if the object is not defined on the target layer.
- Error out if object to rename has any definitions or overs on any stronger or weaker layers. A Reference to a weaker layer should be nameable if it meets all other criteria here.
- Error out if the object to be renamed has been referenced by another layer. Internal references (references without a file path specified) from the same file, should be nameable. 
- Created utilities for common USD operations under `lib\usd\utils\util.h `

### Unit tests:
- Previous `testRename` is renamed to `testRenameUndo`. 
- Added 3 additional tests for newly added restriction logic.

```
testRename
testRenameUndo
testRenameRestrictionSameLayerDef
testRenameRestrictionOtherLayerOpinions
testRenameRestrictionExternalReference
```


lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -42,18 +42,29 @@ UsdUndoRenameCommand::UsdUndoRenameCommand(const UsdSceneItem::Ptr& srcItem, con
...
+
+    // check if the target layer has any opinions that affects selected prim
+    if (!isTargetLayerHaveOpinion(prim)) {
+        auto possibleTargetLayer = targetLayerWithOpion(prim);
+        std::string err = TfStringPrintf("Unable to rename [%s] on current target layer, " 
+                                         "Please set [%s] as the target layer to proceed", 
+                                         prim.GetName().GetString(), 
+                                         possibleTargetLayer->GetDisplayName());
+        throw std::runtime_error(err.c_str());
+    }
```
HamedSabri-adsk, line 67: We check here if the target layer has any opinions that affects selected prim. If it fails, we then find the possible target layer and append it's display name to the error message. Finaly we throw a runtime_error message.

lib/mayaUsd/ufe/UsdUndoRenameCommand.h
```diff
@@ -59,12 +59,12 @@ class MAYAUSD_CORE_PUBLIC UsdUndoRenameCommand : public Ufe::UndoableCommand
...
-    UsdSceneItem::Ptr fUfeSrcItem;
-	SdfPath fUsdSrcPath;
-	UsdSceneItem::Ptr fUfeDstItem;
-	SdfPath fUsdDstPath;
+	UsdStageWeakPtr m_fStage;
+	SdfLayerHandle m_fLayer;
+    UsdSceneItem::Ptr m_fUfeSrcItem;
+	SdfPath m_fUsdSrcPath;
+	UsdSceneItem::Ptr m_fUfeDstItem;
+	SdfPath m_fUsdDstPath;
```
HamedSabri-adsk, line 67: Following the C++ guideline, member variables need to be prefixed with `m_ `

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -109,8 +109,6 @@ SdfLayerHandle defPrimSpecLayer(const UsdPrim& prim)
 
 	SdfLayerHandle defLayer;
 	auto layerStack = prim.GetStage()->GetLayerStack();
-    auto stage = prim.GetStage();
-    auto primFromPath = stage->GetPrimAtPath(prim.GetPath());
```
HamedSabri-adsk, line 113: remove unused variables.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -124,6 +122,38 @@ SdfLayerHandle defPrimSpecLayer(const UsdPrim& prim)
...
+    auto primSpec = layer->GetPrimAtPath(prim.GetPath());
+
+    // to know whether the target layer contains any opinions that 
+    // affect a particular prim, there must be a primSpec for that prim
+    if (!primSpec) {
+        return false;
+    }
+
+    return true;
+}
```
HamedSabri-adsk, line 138: To find out if a specific layer contains any opinions that affect a particular prim, we can use the Sdf's GetPrimAtPath() for that prim.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -124,6 +122,38 @@ SdfLayerHandle defPrimSpecLayer(const UsdPrim& prim)
...
+        // affect a particular prim, there must be a primSpec for that prim
+        auto primSpec = layer->GetPrimAtPath(prim.GetPath());
+        if (primSpec) {
+            targetLayer = layer;
+            break;
+        }
+    }
+    return targetLayer;
+}
+
```
HamedSabri-adsk, line 156: This function is used to give the user a hint to what target layer to set in order to rename a particular prim.

lib/mayaUsd/ufe/Utils.h
```diff
@@ -56,6 +56,14 @@ bool isRootChild(const Ufe::Path& path);
...
 
+//! Check if the target layer has any opinions that affects a particular prim
+MAYAUSD_CORE_PUBLIC
+bool isTargetLayerHaveOpinion(const UsdPrim& prim);
+
+//! Return the target layer that has any opinions on a particular prim
+MAYAUSD_CORE_PUBLIC
+SdfLayerHandle targetLayerWithOpion(const UsdPrim& prim);
+
 MAYAUSD_CORE_PUBLIC
```
HamedSabri-adsk, line 67: defPrimSpecLayer, isTargetLayerHaveOpinion, targetLayerWithOpion functions don't really operate on UFE and they all only rely on USD functionalities.

I am thinking if `maya-usd\lib\usd\utils` is a better place to house such functionalities.

kxl-adsk: Agreed, it has no dependency on UFE, nor Maya.

lib/mayaUsd/ufe/UsdUndoRenameCommand.h
kxl-adsk: Actually, we use `_` prefix - see https://github.com/Autodesk/maya-usd/blob/dev/doc/codingGuidelines.md#naming-file-type-variable-constant-function-namespace-macro-template-parameters-schema-names

New modules should start using these guidelines, but we still have to plan how to adapt the existing code.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -124,6 +122,38 @@ SdfLayerHandle defPrimSpecLayer(const UsdPrim& prim)
...
+    // to know whether the target layer contains any opinions that 
+    // affect a particular prim, there must be a primSpec for that prim
+    if (!primSpec) {
+        return false;
+    }
+
+    return true;
+}
+
+SdfLayerHandle targetLayerWithOpion(const UsdPrim& prim)
```
ppt-adsk, line 140: Typo: "Opion" --> "Opinion".
Also, isn't it the case that this will return the highest-priority layer that has an opinion, either def or over?  It would be good if the function would say that, e.g.
SdfLayerHandle strongestLayerWithOpinion(const UsdPrim& prim)
or
SdfLayerHandle strongestLayerWithPrimSpec(const UsdPrim& prim)


A minor point is that the existing function in the Utils file is called defPrimSpecLayer, i.e. uses "prim spec" rather than "opinion".  Perhaps we should standardize on one terminology or the other.

lib/mayaUsd/ufe/Utils.h
ppt-adsk: That's very true, but then to be consistent you'd need to make this change to other functions in the file.  There are other utility functions in this file that are Maya-only and don't depend on UFE, so it makes you wonder where those should go --- strictly speaking not in the ufe directory.

lib/mayaUsd/ufe/Utils.h
```diff
@@ -56,6 +56,14 @@ bool isRootChild(const Ufe::Path& path);
...
 SdfLayerHandle defPrimSpecLayer(const UsdPrim& prim);
 
+//! Check if the target layer has any opinions that affects a particular prim
+MAYAUSD_CORE_PUBLIC
+bool isTargetLayerHaveOpinion(const UsdPrim& prim);
```
ppt-adsk, line 61: I feel the naming of these two functions is confusing, because they both talk about "target layer", but they're not related.
"isTargetLayerHaveOpinion" is really answering the question: does the edit target layer on the prim's stage have an opinion?
"targetLayerWithOpion" [sic] is answering the question: what is the strongest layer that has an opinion for this prim?  From the point of view of this function, there is no target involved, and therefore I would remove the word "target" here.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -42,18 +42,29 @@ UsdUndoRenameCommand::UsdUndoRenameCommand(const UsdSceneItem::Ptr& srcItem, con
...
-	if (!fLayer) {
+	m_fUsdDstPath = prim.GetParent().GetPath().AppendChild(TfToken(newName.string()));
+	m_fLayer = defPrimSpecLayer(prim);
+	if (!m_fLayer) {
 		std::string err = TfStringPrintf("No prim found at %s", prim.GetPath().GetString().c_str());
 		throw std::runtime_error(err.c_str());
 	}
+
+    // check if the target layer has any opinions that affects selected prim
+    if (!isTargetLayerHaveOpinion(prim)) {
```
ppt-adsk, line 60: So here if the edit target doesn't have an opinion, we tell the user to set the edit target to the strongest layer than does have an opinion.  O.K., maybe that makes sense, even if the strongest layer only has "over" opinion.

But what if the edit target is on a weak layer and the weak layer DOES have an opinion, that is overridden by an opinion in a stronger layer?  We're saying that's O.K.?  The user will rename in the weak layer, but the stronger layer will still have its opinion and it will be composed into the final result, no?  That doesn't seem to meet Will's acceptance criteria that "Trying to rename an object with opinions on stronger layers error out."  Perhaps what you're showing here is really work in progress and you're going to fix this?  Or is there something I haven't understood?

lib/mayaUsd/ufe/UsdUndoRenameCommand.h
HamedSabri-adsk: My bad, I was sill thinking about "m_" :)   Fixed in 54166d2b6bc

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -124,6 +122,38 @@ SdfLayerHandle defPrimSpecLayer(const UsdPrim& prim)
...
 
+bool isTargetLayerHaveOpinion(const UsdPrim& prim)
+{
+    auto editTarget = prim.GetStage()->GetEditTarget();
+    auto layer = editTarget.GetLayer();
+    auto primSpec = layer->GetPrimAtPath(prim.GetPath());
+
+    // to know whether the target layer contains any opinions that 
+    // affect a particular prim, there must be a primSpec for that prim
+    if (!primSpec) {
```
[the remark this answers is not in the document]
HamedSabri-adsk: Good point @ppt-adsk  but I don't find any "explicit operator bool" both in SdfSpec and SdfPrimSpec

lib/mayaUsd/ufe/Utils.cpp
HamedSabri-adsk: Please see 54166d2b6

I agree with standardization on one terminology, I picked "prim spec" instead of "opinion" for now.
Regarding the function names, I made following changes and dropped the target. Let me know if this reads better:

**Before:**
```
isTargetLayerHaveOpinion 
targetLayerWithOpion      
```
**After:**
```
doesLayerHavePrimSpec  
strongestLayerWithPrimSpec
```

lib/mayaUsd/ufe/Utils.h
HamedSabri-adsk: Please see 54166d2b6

I agree. I made following changes and dropped the target. Let me know if this reads better:

**Before:**
```
isTargetLayerHaveOpinion 
```
**After:**
```
doesLayerHavePrimSpec  
```

"doesLayerHavePrimSpec" answers this question:  Does the layer that you are currently on have any primSpec (a.k.a opinion) on the selected prim?

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
HamedSabri-adsk: Very valid question and I should have elaborated more on the change here:

if (!MayaUsdUtils::doesLayerHavePrimSpec(prim)) solves this request:

`Trying to rename an object that was not defined on the target layer errors out`

but as you also mentioned it doesn't really solve this request:

`Trying to rename an object with opinions on stronger layers error out.`

I am working on addressing this as we speak.

lib/mayaUsd/ufe/Utils.h
HamedSabri-adsk: See 54166d2b6
@ppt-adsk  correct, for the sake of making this PR short I only moved out the USD-only utility functions from lib/maya-usd to lib/usd.

lib/usd/utils/util.cpp
```diff
@@ -0,0 +1,81 @@
...
+    // affect a particular prim, there must be a primSpec for that prim
+    if (!primSpec) {
+        return false;
+    }
+
+    return true;
+}
+
+SdfLayerHandle
+MayaUsdUtils::strongestLayerWithPrimSpec(const UsdPrim& prim)
```
HamedSabri-adsk, line 64: extra MayaUsdUtils:: in line 64 and 48  was added by accident. I will fix this in the next PR.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
HamedSabri-adsk: @ppt-adsk please see 8f8295f3b8.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -42,18 +44,44 @@ UsdUndoRenameCommand::UsdUndoRenameCommand(const UsdSceneItem::Ptr& srcItem, con
...
+    else
+    {
+        auto layers = MayaUsdUtils::layersWithOpinion(prim);
+
+        if (layers.size() > 1) {
+            std::string layerNames;
+            for (auto layer : layers) {
+                layerNames.append("[" + layer->GetDisplayName() + "]" + ",");
+            }
+            layerNames.pop_back();
```
ppt-adsk, line 79: Why omit the last layer?

lib/usd/utils/util.cpp
```diff
@@ -0,0 +1,98 @@
...
+        if (primSpec && (primSpec->GetSpecifier() == SdfSpecifierDef)) {
+            defLayer = layer;
+            break;
+        }
+    }
+    return defLayer;
+}
+
+std::vector<SdfLayerHandle>
+layersWithOpinion(const UsdPrim& prim)
```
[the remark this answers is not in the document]
HamedSabri-adsk: please see 4c67b698066f1e18b8c40caf6e3a1158066c52d2

lib/usd/utils/util.cpp
```diff
@@ -0,0 +1,98 @@
...
+{
+    // get the list of PrimSpecs that provide opinions for this prim
+    // ordered from strongest to weakest opinion.
+    const auto& primStack = prim.GetPrimStack();
+
+    std::vector<SdfLayerHandle> layersWithOpion;
```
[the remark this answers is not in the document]
HamedSabri-adsk: see 4c67b698066f1e18b8c40caf6e3a1158066c52d2

Some how I can't type `opinion` properly :)

lib/usd/utils/util.cpp
```diff
@@ -0,0 +1,98 @@
...
+    for (auto primSpec : primStack) {
+        layersWithOpion.emplace_back(primSpec->GetLayer());
+    }
+
+    return layersWithOpion;
+}
+
+bool
+doesLayerHavePrimSpec(const UsdPrim& prim)
```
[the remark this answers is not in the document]
HamedSabri-adsk: Good point, I like your suggested name.  

see 4c67b698066f1e18b8c40caf6e3a1158066c52d2

lib/usd/utils/util.h
```diff
@@ -0,0 +1,46 @@
...
+
+    //! Return a list of layers in strength order that have opinions on a prim
+    MAYA_USD_UTILS_PUBLIC
+    std::vector<SdfLayerHandle> layersWithOpinion(const UsdPrim&);
+
+    //! Check if a layer has any opinions that affects a particular prim
+    MAYA_USD_UTILS_PUBLIC
+    bool doesLayerHavePrimSpec(const UsdPrim&);
+
+    //! Return the layer that has any opinions on a particular prim
```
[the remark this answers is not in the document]
HamedSabri-adsk: see 4c67b698066f1e18b8c40caf6e3a1158066c52d2

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -42,18 +44,44 @@ UsdUndoRenameCommand::UsdUndoRenameCommand(const UsdSceneItem::Ptr& srcItem, con
...
 	const UsdPrim& prim = srcItem->prim();
-	fStage = prim.GetStage();
-	fUfeSrcItem = srcItem;
-	fUsdSrcPath = prim.GetPath();
-	// Every call to rename() (through execute(), undo() or redo()) removes
+	_stage = prim.GetStage();
+	_ufeSrcItem = srcItem;
+	_usdSrcPath = prim.GetPath();
+
+    // Every call to rename() (through execute(), undo() or redo()) removes
```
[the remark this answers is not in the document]
HamedSabri-adsk: see 4c67b698066f1e18b8c40caf6e3a1158066c52d2

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -42,18 +44,44 @@ UsdUndoRenameCommand::UsdUndoRenameCommand(const UsdSceneItem::Ptr& srcItem, con
...
-	if (!fLayer) {
+	_usdDstPath = prim.GetParent().GetPath().AppendChild(TfToken(newName.string()));
+	_layer = MayaUsdUtils::defPrimSpecLayer(prim);
+	if (!_layer) {
 		std::string err = TfStringPrintf("No prim found at %s", prim.GetPath().GetString().c_str());
 		throw std::runtime_error(err.c_str());
 	}
+
+    // if the current layer doesn't have any opinions that affects selected prim
+    if (!MayaUsdUtils::doesLayerHavePrimSpec(prim)) {
```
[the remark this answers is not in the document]
HamedSabri-adsk: Good point. Correct, we could also use `_layer != prim.GetStage()->GetEditTarget().GetLayer()` instead. I keep using doesEditTargetLayerHavePrimSpec which is a bit more generic. Hope this is Ok.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
HamedSabri-adsk: see 4c67b698066f1e18b8c40caf6e3a1158066c52d2

I rename `layerNames` to `layerDisplayNames `so it reads better.

I am not omitting any layers and rather the extra "," at the end of `layerDisplayNames ` string.

This is the message the design wants to show up as part of the error message:

`Opinions exist in [layer name 1], [layer name 2], [layer name 3], [layer name 4].”`

So when I iterate through the layers, I get their GetDisplayName and concatenate them in the
layerDisplayNames string separated by ","

```
for (auto layer : layers) {
    layerDisplayNames.append("[" + layer->GetDisplayName() + "]" + ",");
}
layerDisplayNames.pop_back();
```
There are better ways of doing this, but this is one way I did.

test/lib/ufe/testRename.py
```diff
@@ -149,7 +166,52 @@ def childrenNames(children):
         propsChildren = propsHierarchy.children()
         propsChildrenNames = childrenNames(propsChildren)
 
-        self.assertIn(ball35RenName, propsChildrenNames)
-        self.assertNotIn('Ball_35', propsChildrenNames)
+        self.assertIn(pCylinder1RenName, propsChildrenNames)
+        self.assertNotIn('pCylinder1', propsChildrenNames)
         self.assertEqual(len(propsChildren), len(propsChildrenPre))
```
HamedSabri-adsk, line 171: @ppt-adsk  edited code logic in `testRename` to use `usdCylinder.ma` scene since with new rename restrictions `top_layer.ma` doesn't meet the requirements for testing undo/redo

test/lib/ufe/testRename.py
```diff
@@ -149,7 +166,52 @@ def childrenNames(children):
...
+        stage = mayaUsd.ufe.getStage(str(mayaPathSegment))
+
+        # set the edit target to Assembly_room_set.usda
+        stage.SetEditTarget(stage.GetLayerStack()[2])
+        self.assertEqual(stage.GetEditTarget().GetLayer().GetDisplayName(), "Assembly_room_set.usda")
+
+        # expect the exception happens
+        with self.assertRaises(RuntimeError):
+           newName = 'Ball_35_Renamed'
+           cmds.rename(newName)
```
HamedSabri-adsk, line 217: @ppt-adsk  I added two test cases for rename restriction operations. I created two separate tests:

`testRenameRestriction` is to test if the prim that we are about to rename is in the correct layer

`testRenameRestriction2` is to test if the prim that we are about to rename doesn't have any definitions or opinions on other layers.

If you think `testRenameRestriction2` is ugly, I am happy to change it with something more meaningful.

test/lib/ufe/ufeTestUtils/mayaUtils.py
```diff
@@ -126,4 +126,15 @@ def openTopLayerScene():
...
+    filePath = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "test-samples", "cylinder", "usdCylinder.ma" )
+    cmds.file(filePath, force=True, open=True)
+
+def openTwoSpheresScene():
+    filePath = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "test-samples", "twoSpheres", "twoSpheres.ma" )
+    cmds.file(filePath, force=True, open=True)
+
+def openSphereAnimatedRadiusScene():
+    filePath = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "test-samples", "sphereAnimatedRadius", "sphereAnimatedRadiusProxyShape.ma" )
+    cmds.file(filePath, force=True, open=True)
```
HamedSabri-adsk, line 140: refactored code around opening maya test scenes and introduced three utility functions (openCylinderScene, openTwoSpheresScene, openSphereAnimatedRadiusScene)

test/lib/ufe/testRename.py
```diff
@@ -149,7 +166,52 @@ def childrenNames(children):
         propsChildren = propsHierarchy.children()
         propsChildrenNames = childrenNames(propsChildren)
 
-        self.assertIn(ball35RenName, propsChildrenNames)
-        self.assertNotIn('Ball_35', propsChildrenNames)
+        self.assertIn(pCylinder1RenName, propsChildrenNames)
+        self.assertNotIn('pCylinder1', propsChildrenNames)
         self.assertEqual(len(propsChildren), len(propsChildrenPre))
 
+    def testRenameRestriction(self):
```
ppt-adsk, line 173: How about "testRenameRestrictionSameLayerDef", or similar?

test/lib/ufe/testRename.py
```diff
@@ -149,7 +166,52 @@ def childrenNames(children):
...
+        # check GetLayerStack behavior
+        self.assertEqual(stage.GetLayerStack()[0], stage.GetSessionLayer())
+        self.assertEqual(stage.GetEditTarget().GetLayer(), stage.GetSessionLayer())
+
+        # expect the exception happens
+        with self.assertRaises(RuntimeError):
+            newName = 'Ball_35_Renamed'
+            cmds.rename(newName)
+
+    def testRenameRestriction2(self):
```
ppt-adsk, line 196: How about "testRenameRestrictionOtherLayerOpinions", or similar?

test/lib/ufe/testRename.py
HamedSabri-adsk: @ppt-adsk  Merci Pierre. please see 581932b82e3053af6cebad02c829fb36263c15a6

ppt-adsk:
Great job Hamed!

lib/mayaUsd/ufe/Utils.h
```diff
@@ -82,5 +82,9 @@ MDagPath nameToDagPath(const std::string& name);
 MAYAUSD_CORE_PUBLIC
 UsdTimeCode getTime(const Ufe::Path& path);
 
+//! Object renamed scene notification
+MAYAUSD_CORE_PUBLIC
+void sendRenameNotification(const Ufe::SceneItem::Ptr& item, const Ufe::Path& previousPath);
+
```
HamedSabri-adsk, line 88: Created a utility function for sending  UFE scene rename notification.

lib/mayaUsd/ufe/UsdUndoRenameCommand.h
```diff
@@ -19,9 +19,6 @@
 #include <ufe/pathComponent.h>
 #include <ufe/undoableCommand.h>
 
-#include <pxr/usd/sdf/path.h>
-#include <pxr/usd/usd/prim.h>
-
```
HamedSabri-adsk, line 24: These headers are not needed here.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -88,84 +89,85 @@ UsdUndoRenameCommand::~UsdUndoRenameCommand()
...
+
             _ufeDstItem = nullptr;
         }
-	}
+    }
     else {
-        UFE_LOG(std::string("Warning: RemovePrim(") +
+        UFE_LOG(std::string("Warning: SdfCopySpec(") +
                 _usdDstPath.GetString() + std::string(") failed."));
     }
```
HamedSabri-adsk, line 168: On Undo, we do an opposite of Redo:

1- Call SdfCopySpec to copy ALL the child specs from _usdDstPath to _usdSrcPath. This is a crucial call.
2- Remove the prim at the _usdDstPath.
3- Create a new prim from _usdSrcPath and check if have a valid object to continue.
4- set _ufeSrcItem by calling createSiblingSceneItem ( read comments in code )
5- send rename notification
6- set _ufeDstItem to nullptr

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -88,84 +89,85 @@ UsdUndoRenameCommand::~UsdUndoRenameCommand()
...
         if (status) {
             // The renamed scene item is a "sibling" of its original name.
-            auto ufeSrcPath = _ufeSrcItem->path();
-            _ufeDstItem = createSiblingSceneItem(
-                ufeSrcPath, _usdDstPath.GetElementString());
+            _ufeDstItem = createSiblingSceneItem(_ufeSrcItem->path(), _usdDstPath.GetElementString());
 
-            Ufe::ObjectRename notification(_ufeDstItem, ufeSrcPath);
-            Ufe::Scene::notifyObjectPathChange(notification);
+            sendRenameNotification(_ufeDstItem, _ufeSrcItem->path());
```
HamedSabri-adsk, line 122: minor refactoring here.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -88,84 +89,85 @@ UsdUndoRenameCommand::~UsdUndoRenameCommand()
...
+    // would be the edit target layer be the destination:
+    // _layer = _stage->GetEditTarget().GetLayer()
+    bool status = SdfCopySpec(_layer, _usdSrcPath, _layer, _usdDstPath);
+    if (status) {
+        // remove all scene description for the given path and 
+        // its subtree in the current UsdEditTarget 
+        {
+            UsdEditContext ctx(_stage, _layer);
+            status = _stage->RemovePrim(_ufeSrcItem->prim().GetPath());
+        }
```
HamedSabri-adsk, line 116: With the new restriction in place it is now safe to remove the prim from the scene graph rather that deactivating it.When RemovePrim is called, I simply pass the `_ufeSrcItem->prim().GetPath()` rather than creating a new srcPrim which would be really unnecessary.

Redo operation:
1- Call SdfCopySpec to copy ALL the child specs from _usdSrcPath to _usdDstPath. This is a crucial call.
2- Remove the prim at _ufeSrcItem USD path.
3- set _ufeDstItem by calling createSiblingSceneItem
4- send rename notification

lib/mayaUsd/ufe/UsdUndoRenameCommand.h
```diff
@@ -34,39 +31,39 @@ namespace ufe {
...
     UsdSceneItem::Ptr _ufeSrcItem;
-	SdfPath _usdSrcPath;
+    SdfPath _usdSrcPath;
 
-	UsdSceneItem::Ptr _ufeDstItem;
-	SdfPath _usdDstPath;
+    UsdSceneItem::Ptr _ufeDstItem;
+    SdfPath _usdDstPath;
 
 }; // UsdUndoRenameCommand
```
HamedSabri-adsk, line 69: I have converted tabs to dots. If you find them distracting, Github allow hiding whitespace changes during review:

Please see 
https://medium.com/@kevinpmcc/hide-whitespace-for-easier-github-code-reviews-d3f476d03863

HamedSabri-adsk:
@ppt-adsk  I have the first stab at removing the cruft in USD files as objects are named and renamed. please see 20be9ac681b1fde7a44abc4e9312db15bd916d1b

HamedSabri-adsk:
@ppt-adsk  Ah....... I am starting to think why we need to rely on `SdfCopySpec` any more. 

I honestly think we can simply rename a prim in place. I just realized that `SdfPrimSpec` has a routine called `SetName` to set the prim name:

https://graphics.pixar.com/usd/docs/api/class_sdf_prim_spec.html#aa1f241d28f26488867a276eca6d54405

This would really simplify a lot of things. Stay tuned.....

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -122,7 +125,14 @@ bool UsdUndoRenameCommand::renameRedo()
     // the renamed scene item is a "sibling" of its original name.
     _ufeDstItem = createSiblingSceneItem(_ufeSrcItem->path(), _newName);
     sendRenameNotification(_ufeDstItem, _ufeSrcItem->path());
- 
+
+    // SdfLayer is a "simple" container, and all it knows about defaultPrim is that it is a piece of token-valued layer metadata.  
+    // It is only the higher-level Usd and Pcp modules that know that it is identifying a prim on the stage.  
+    // One must use the SdfLayer API for setting the defaultPrim when you rename the prim it identifies.
+    if(primNameStr == stageDefPrimNameStr){
```
HamedSabri-adsk, line 132: @ppt-adsk We don't need to do a comparison here.  if(stageDefPrimNameStr) does the work.  Will fix it in the next commit,

HamedSabri-adsk: Hmm, I don't think that would work now that I think about it. please ignore my last comment.

HamedSabri-adsk: @ppt-adsk please see d84d82fcaa2f377e77114aaf9adf25f5be8c601b

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -88,110 +103,109 @@ UsdUndoRenameCommand::~UsdUndoRenameCommand()
...
+
+    return true;
+}
 
 void UsdUndoRenameCommand::undo()
 {
-	// MAYA-92264: Pixar bug prevents undo from working.  Try again with USD
-	// version 0.8.5 or later.  PPT, 7-Jul-2018.
-	try {
+    // MAYA-92264: Pixar bug prevents undo from working.  Try again with USD
```
ppt-adsk, line 191: This comment is no longer valid, right?

PR #478, opened 2020-05-04
MAYA-104213 - add a new USD stage to an empty scene

seando-adsk:
#### MAYA-104213 - As a user, I'd like to add a new USD stage to an empty scene

* New USD notice type for stage invalidate. Send this notice when certain attributes (like filepath) are changed. From the StagesSubject receiver of this notice we also send the new Ufe::ObjectInvalidate() notification.
* When proxy shape compute is called if no filepath was attached, we create an in memory stage.
* Added dirty state to UsdStageMap. When dirty it will rebuild when a new request for stage info is made.
* Cleanup: renamed "UsdMayaProxyStageSetNotice" -> "MayaUsdProxyStageSetNotice".
* Added new menu item "Stage with New Layer".

### DIFF
--- a/lib/mayaUsd/listeners/proxyShapeNotice.h
+++ b/lib/mayaUsd/listeners/proxyShapeNotice.h
@@ -13,8 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-#ifndef USDMAYA_PROXYSTAGE_NOTICE_H
-#define USDMAYA_PROXYSTAGE_NOTICE_H
+#ifndef MAYAUSD_PROXYSTAGE_NOTICE_H
+#define MAYAUSD_PROXYSTAGE_NOTICE_H
 
 #include <maya/MObject.h>
 
@@ -27,11 +27,11 @@ PXR_NAMESPACE_OPEN_SCOPE
 class MayaUsdProxyShapeBase;
 
 /// Notice sent when the ProxyShape loads a new stage
-class UsdMayaProxyStageSetNotice : public TfNotice
+class MayaUsdProxyStageBaseNotice : public TfNotice
 {
 public:
     MAYAUSD_CORE_PUBLIC
-    UsdMayaProxyStageSetNotice(const MayaUsdProxyShapeBase& proxy);
+    MayaUsdProxyStageBaseNotice(const MayaUsdProxyShapeBase& proxy);
 
     /// Get proxy shape which had stage set
     MAYAUSD_CORE_PUBLIC
@@ -39,6 +39,18 @@ public:
 
 private:
     const MayaUsdProxyShapeBase& _proxy;
+};
+
+class MayaUsdProxyStageSetNotice : public MayaUsdProxyStageBaseNotice
+{
+public:
+    using MayaUsdProxyStageBaseNotice::MayaUsdProxyStageBaseNotice;
+};
+
+class MayaUsdProxyStageInvalidateNotice : public MayaUsdProxyStageBaseNotice
+{
+public:
+    using MayaUsdProxyStageBaseNotice::MayaUsdProxyStageBaseNotice;
 };
 
 PXR_NAMESPACE_CLOSE_SCOPE
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -496,6 +496,11 @@ MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
 
             usdStage->SetEditTarget(usdStage->GetSessionLayer());
         }
+        else if (!usdStage) {
+            // Create a new stage in memory with an anonymous root layer.
+            UsdStageCacheContext ctx(UsdMayaStageCache::Get());
+            usdStage = UsdStage::CreateInMemory("", loadSet);
+        }
 
         if (usdStage) {
             primPath = usdStage->GetPseudoRoot().GetPath();
@@ -621,7 +626,7 @@ MayaUsdProxyShapeBase::computeOutStageData(MDataBlock& dataBlock)
                   this,
                   std::placeholders::_1));
 
-    UsdMayaProxyStageSetNotice(*this).Send();
+    MayaUsdProxyStageSetNotice(*this).Send();
 
     return MS::kSuccess;
 }
@@ -797,6 +802,7 @@ MayaUsdProxyShapeBase::setDependentsDirty(const MPlug& plug, MPlugArray& plugArr
         plug == loadPayloadsAttr ||
         plug == inStageDataAttr) {
         _IncreaseUsdStageVersion();
+        MayaUsdProxyStageInvalidateNotice(*this).Send();
     }
 
     return MPxSurfaceShape::setDependentsDirty(plug, plugArray);
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -26,6 +26,7 @@
 #include <mayaUsd/ufe/ProxyShapeHandler.h>
 #include <mayaUsd/ufe/UsdStageMap.h>
 #include <mayaUsd/ufe/Utils.h>
+#include <mayaUsd/nodes/proxyShapeBase.h>
 
 #include "private/InPathChange.h"
 
@@ -100,6 +101,7 @@ StagesSubject::StagesSubject()
 
 	TfWeakPtr<StagesSubject> me(this);
 	TfNotice::Register(me, &StagesSubject::onStageSet);
+	TfNotice::Register(me, &StagesSubject::onStageInvalidate);
 }
 
 StagesSubject::~StagesSubject()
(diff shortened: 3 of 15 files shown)


lib/mayaUsd/listeners/proxyShapeNotice.h
```diff
@@ -27,11 +27,11 @@ PXR_NAMESPACE_OPEN_SCOPE
 class MayaUsdProxyShapeBase;
 
 /// Notice sent when the ProxyShape loads a new stage
-class UsdMayaProxyStageSetNotice : public TfNotice
+class MayaUsdProxyStageBaseNotice : public TfNotice
```
seando-adsk, line 30: Renamed this one "UsdMaya" -> "MayaUsd" and made it a base notice class. Below created this existing one (deriving from this base) and new one for stage invalidate.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -496,6 +496,11 @@ MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
 
             usdStage->SetEditTarget(usdStage->GetSessionLayer());
         }
+        else if (!usdStage) {
+            // Create a new stage in memory with an anonymous root layer.
+            UsdStageCacheContext ctx(UsdMayaStageCache::Get());
+            usdStage = UsdStage::CreateInMemory("", loadSet);
```
seando-adsk, line 502: The empty string is for the identifier, which is used when naming the root layer. Here we just want it named as the default "anon", rather than "tmp.usda"

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -797,6 +802,7 @@ MayaUsdProxyShapeBase::setDependentsDirty(const MPlug& plug, MPlugArray& plugArr
         plug == loadPayloadsAttr ||
         plug == inStageDataAttr) {
         _IncreaseUsdStageVersion();
+        MayaUsdProxyStageInvalidateNotice(*this).Send();
```
seando-adsk, line 805: Send the new stage invalidate notice whenever one of the plugs that affects outStageDataAttr changes.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -170,29 +172,15 @@ void StagesSubject::afterOpen()
 	// frequent, we won't implement this for now.  PPT, 22-Dec-2017.
 	std::for_each(std::begin(fStageListeners), std::end(fStageListeners),
 		[](StageListenerMap::value_type element) { TfNotice::Revoke(element.second); } );
-
-	StagesSubject::Ptr me(this);
-	for (auto stage : ProxyShapeHandler::getAllStages())
-	{
-		fStageListeners[stage] = TfNotice::Register(
-			me, &StagesSubject::stageChanged, stage);
-	}
```
seando-adsk, line 179: We should not get the stages here as that will pull, but the new stage may not be ready yet.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -170,29 +172,15 @@ void StagesSubject::afterOpen()
...
 	// - get their stage.
-	g_StageMap.clear();
-	auto proxyShapeNames = ProxyShapeHandler::getAllNames();
-	for (const auto& psn : proxyShapeNames)
-	{
-		MDagPath dag = nameToDagPath(psn);
-		Ufe::Path ufePath = dagPathToUfe(dag);
-		auto stage = ProxyShapeHandler::dagPathToStage(psn);
-		g_StageMap.addItem(ufePath, stage);
-	}
```
seando-adsk, line 195: All this code was moved into the stage map. We just set it dirty now. It will be rebuilt when queried.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -277,9 +265,29 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
...
+	// We should have no listerners and stage map is dirty.
+	TF_VERIFY(g_StageMap.isDirty());
+	TF_VERIFY(fStageListeners.empty());
+
+	StagesSubject::Ptr me(this);
+	for (auto stage : ProxyShapeHandler::getAllStages())
+	{
+		fStageListeners[stage] = TfNotice::Register(
+			me, &StagesSubject::stageChanged, stage);
+	}
```
seando-adsk, line 279: When we get the stage set notice is when we can register. The stage(s) is(are) valid at this point.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -277,9 +265,29 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
...
+
+void StagesSubject::onStageInvalidate(const MayaUsdProxyStageInvalidateNotice& notice)
 {
 	afterOpen();
+
+#if UFE_PREVIEW_VERSION_NUM >= 2014
+	Ufe::SceneItem::Ptr sceneItem = Ufe::Hierarchy::createItem(notice.GetProxyShape().ufePath());
+	auto notification = Ufe::SubtreeInvalidate(sceneItem);
+	Ufe::Scene::notifySubtreeInvalidate(notification);
+#endif
```
seando-adsk, line 290: Send a new UFE subtree invalidate notice when the stage is invalidated. Clients like the Outliner can use this to re-build.

lib/mayaUsd/ufe/UsdStageMap.cpp
```diff
@@ -92,8 +93,10 @@ void UsdStageMap::addItem(const Ufe::Path& path, UsdStageWeakPtr stage)
 	fStageToObject[stage] = proxyShape;
 }
 
-UsdStageWeakPtr UsdStageMap::stage(const Ufe::Path& path) const
+UsdStageWeakPtr UsdStageMap::stage(const Ufe::Path& path)
 {
+	rebuildIfDirty();
```
seando-adsk, line 98: All the query methods on the stage map, will now rebuild first if the map is flagged as dirty.

lib/mayaUsd/ufe/UsdStageMap.cpp
```diff
@@ -106,19 +109,37 @@ UsdStageWeakPtr UsdStageMap::stage(const Ufe::Path& path) const
...
+	if (!fDirty) return;
+
+	auto proxyShapeNames = ProxyShapeHandler::getAllNames();
+	for (const auto& psn : proxyShapeNames)
+	{
+		MDagPath dag = nameToDagPath(psn);
+		Ufe::Path ufePath = dagPathToUfe(dag);
+		auto stage = ProxyShapeHandler::dagPathToStage(psn);
+		addItem(ufePath, stage);
+	}
```
seando-adsk, line 141: This code was moved from the StagesSubject.

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -103,6 +112,7 @@ global proc mayaUsdMenu_createMenuCallback() {
                 -label "Universal Scene Description (USD)" 
                 -annotation "Create a USD stage" 
                 -version $mayaVersion`;
+            menuItem -runTimeCommand mayaUsdCreateStageWithNewLayer;
```
seando-adsk, line 115: New menu item to create a new empty USD stage.

plugin/adsk/scripts/mayaUsd_createStageWithNewLayer.py
```diff
@@ -0,0 +1,32 @@
...
+    - Proxyshape
+    - Stage
+    - Session Layer
+    - Anonymous Root Layer (this is set as the target layer)
+    """
+
+    # Simply create a proxy shape. Since it does not have a USD file associated
+    # (in the .filePath attribute), the proxy shape base will create an empty
+    # stage in memory. This will create the session and root layer as well.
+    shapeNode = cmds.createNode('mayaUsdProxyShape', name='stageShape')
```
seando-adsk, line 32: Because of the changes in the proxyshapebase (to create stage in memory when no file attr) we simply need to create a proxy shape node here. That will create the stage, which will create the root layer and session layer.

seando-adsk:
@pilarmolinalopez Would you be available to review these changes?

seando-adsk:
@kxl-adsk I fixed the three failing tests with [c5bc21d](c5bc21d9138987121f47e113f2ac126117ae1c3a).

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
```diff
@@ -398,6 +398,18 @@ MStatus ProxyShape::setDependentsDirty(const MPlug& plugBeingDirtied, MPlugArray
   {
     MHWRender::MRenderer::setGeometryDrawDirty(thisMObject(), true);
   }
+
+  if (plugBeingDirtied == outStageData() ||
```
kxl-adsk, line 402: ProxyShapeBase is a base class...shouldn't we just forward it to the base class instead of making a copy?

PR #481, opened 2020-05-05
remove use of HdxTaskController::SetColorizeQuantizationEnabled() after USD 20.05

mattyjams:
Following up on #469, the questionable `HdxTaskController::SetColorizeQuantizationEnabled()` was actually removed fairly soon after the release of core USD 20.05 in this commit:
https://github.com/PixarAnimationStudios/USD/commit/584c27d495047aa9a7231df64eec9e22396e5651

The change here ensures we stop using it for USD versions after 20.05.

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
@@ -457,7 +457,7 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext) {
     } else
         _taskController->SetSelectionEnableOutline(false);
 #endif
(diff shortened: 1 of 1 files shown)


mattyjams:
@marsupial: If you have a chance, could you give this a try against the current core USD dev branch and make sure you're still seeing the performance you expect? Just want to make sure there isn't something else needed here to prevent a regression for you. Thanks!

marsupial:
I won't have a chance to do a trial on dev until the end of the week at the earliest.  If it has been removed, then hopefully means they've either fixed it or really removed it entirely.

Either way, if there are performance regressions, that should probably be handled on the USD side at this point.

...so LGTM!...

PR #483, opened 2020-05-06
Improve rename logics

HamedSabri-adsk:
This change simplifies the rename operation by simply renaming a prim in place via `SdfPrimSpec's SetName` routine and removes the old work-flow involving SdfCopySpec.


lib/usd/utils/util.h
```diff
@@ -41,6 +41,10 @@ namespace MayaUsdUtils {
     MAYA_USD_UTILS_PUBLIC
     SdfLayerHandle strongestLayerWithPrimSpec(const UsdPrim&);
 
+    //! Return a PrimSpec for the argument prim in the layer containing the stage's current edit target.
+    MAYA_USD_UTILS_PUBLIC
+    SdfPrimSpecHandle getPrimSpecAtEditTarget(UsdStageWeakPtr, const UsdPrim&);
+
 } // namespace MayaUsdUtils
```
HamedSabri-adsk, line 48: added a utility function for getting a PrimSpec from a prim path in the layer containing the stage's current edit target.

lib/mayaUsd/ufe/UsdUndoRenameCommand.h
```diff
@@ -48,22 +48,18 @@ class MAYAUSD_CORE_PUBLIC UsdUndoRenameCommand : public Ufe::UndoableCommand
...
 
     UsdSceneItem::Ptr _ufeSrcItem;
-    SdfPath _usdSrcPath;
-
     UsdSceneItem::Ptr _ufeDstItem;
-    SdfPath _usdDstPath;
+
+    const UsdPrim& _prim;
+    UsdStageWeakPtr _stage;
+    std::string _newName;
```
HamedSabri-adsk, line 63: Simplified member variables. We no longer need _usdSrcPath, _usdDstPath, and _layer.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -42,35 +42,24 @@ namespace ufe {
 
 UsdUndoRenameCommand::UsdUndoRenameCommand(const UsdSceneItem::Ptr& srcItem, const Ufe::PathComponent& newName)
     : Ufe::UndoableCommand()
+    , _ufeSrcItem(srcItem)
+    , _ufeDstItem(nullptr)
+    , _prim(srcItem->prim())
+    , _stage(_prim.GetStage())
+    , _newName(newName.string())
```
HamedSabri-adsk, line 49: Use initializer list.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -101,79 +90,49 @@ UsdSceneItem::Ptr UsdUndoRenameCommand::renamedItem() const
...
+        sendRenameNotification(_ufeDstItem, _ufeSrcItem->path());
     }
     else {
-        UFE_LOG(std::string("Warning: SdfCopySpec(") +
-                _usdSrcPath.GetString() + std::string(") failed."));
+        return false;
     }
 
-    return status;
+    return true;
```
HamedSabri-adsk, line 109: On Redo:
1- Get the SdfPrimSpec to the _prim and check it's validity.
2- Set it's name to the new name: primSpec->SetName(_newName);
3- Set _ufeDstItem and send UFE scene notification

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -101,79 +90,49 @@ UsdSceneItem::Ptr UsdUndoRenameCommand::renamedItem() const
...
+        _ufeDstItem = nullptr;
     }
     else {
-        UFE_LOG(std::string("Warning: SdfCopySpec(") +
-                _usdDstPath.GetString() + std::string(") failed."));
+        return false;
     }
 
-    return status;
+    return true;
```
HamedSabri-adsk, line 133: On Undo:
1- Get the SdfPrimSpec to the _ufeDstItem->prim() and check it's validity.
2- Set it's name to the new name: primSpec->SetName(_prim.GetPath().GetElementString());
3- Set _ufeSrcItem and send UFE scene notification
4- Set _ufeDstItem = nullptr

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -101,79 +90,49 @@ UsdSceneItem::Ptr UsdUndoRenameCommand::renamedItem() const
...
-            #endif
+        // set prim's name
+        primSpec->SetName(_prim.GetPath().GetElementString());
 
-            // I shouldn't have to again create a sibling sceneItem here since we already have a valid _ufeSrcItem
-            // however, I get random crashes if I don't which needs furthur investigation.  HS, 6-May-2020.
-            _ufeSrcItem = createSiblingSceneItem(_ufeDstItem->path(), _usdSrcPath.GetElementString());
+        // shouldn't have to again create a sibling sceneItem here since we already have a valid _ufeSrcItem
+        // however, I get random crashes if I don't which needs furthur investigation.  HS, 6-May-2020.
+        _ufeSrcItem = createSiblingSceneItem(_ufeDstItem->path(), _prim.GetPath().GetElementString());
```
ppt-adsk, line 123: Isn't the prim in the original _ufeSrcItem now invalid?

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -91,45 +92,45 @@ UsdSceneItem::Ptr UsdUndoRenameCommand::renamedItem() const
...
     }
-    else {
+
+    // set prim's name
+    // XXX: SetName successfuly returns true but when you examine the _prim.GetName()
+    // after the rename, the prim name shows the original name HS, 6-May-2020.
+    bool status = primSpec->SetName(_newName);
+    if (!status) {
         return false;
     }
```
HamedSabri-adsk, line 105: I do an early check for primSpec and SetName result. 

@ppt-adsk  there is something troubling happening here which I am trying to understand:

On Redo, I get the primspec for _prim object and rename it's name to the _newName. This operation return successful BUT when I query the _prim's name right after, I still get the _prim's original name. Huh???

I then set _ufeDstItem and when I print it's USD name and it gives me the _newName which is expected.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -47,6 +47,7 @@ UsdUndoRenameCommand::UsdUndoRenameCommand(const UsdSceneItem::Ptr& srcItem, con
     , _prim(srcItem->prim())
     , _stage(_prim.GetStage())
     , _newName(newName.string())
+    , _oldName(_prim.GetName().GetString())
```
HamedSabri-adsk, line 50: Added _oldName that holds the _prim name before it is changed.

HamedSabri-adsk:
> 
> 
> I think to best demonstrate that this change is robust I would like to see an addition to the rename test where you access a child of the renamed prim, and try to do something with it e.g. use the hierarchy interface on the child to get its parent, which should now be the renamed prim. This way we would know that the subtree of prims under the renamed prim is in a good state.

@ppt-adsk Very valid point. I will update the rename test.

test/lib/ufe/testRename.py
```diff
@@ -96,6 +96,49 @@ def assertStageAndPrimAccess(
...
+
+        # set primspec name
+        primspec.name = "TreeBase_potato"
+
+        assert stage.GetPrimAtPath('/TreeBase_potato')
+        assert stage.GetPrimAtPath('/TreeBase_potato/leavesXform')
+
+        # HS, 6-May-2020. defualtPrim in null?? This can't be null....
+        # defualtPrim = stage.GetDefaultPrim()
+        # self.assertEqual(defualtPrim.GetName(), 'TreeBase_potato')
```
HamedSabri-adsk, line 139: @ppt-adsk  Unless I am missing something really obvious, defualtPrim can not be null in line 138??

Both assert passed in line 134,135 which to me indicates the subtree under the renamed prim is in a good state.

test/lib/ufe/testRename.py
```diff
@@ -96,6 +96,49 @@ def assertStageAndPrimAccess(
...
+
+    def testRenameUndo(self):
```
HamedSabri-adsk, line 141: renamed original testRename to testRenameUndo.

test/lib/ufe/testRename.py
HamedSabri-adsk: OK, I did a little more digging into this and the good news is that the tree/subtree  graph is in a good state after the rename.  It is just that we can't no longer rely on the stage which looks like a bug on USD side to me or perhaps there is a rational reason for this behavior that I don't know.
 
```
# set primspec name
primspec.name = "TreeBase_potato"

potatoPrim = stage.GetPrimAtPath('/TreeBase_potato')
self.assertEqual(potatoPrim.GetName(), 'TreeBase_potato')   // Pass
```

Traverse the stage

```
/TreeBase_potato
/TreeBase_potato/leavesXform
/TreeBase_potato/leavesXform/leaves
/TreeBase_potato/trunk
```

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -60,8 +60,23 @@ UsdUndoRenameCommand::UsdUndoRenameCommand(const UsdSceneItem::Ptr& srcItem, con
...
+                // this will be empty in the case of an internal reference.
+                if (ref.GetAssetPath().empty()) { 
+                    return;
+                }else{
+                   std::string err = TfStringPrintf("Unable to rename referenced object [%s]", 
+                                                    _prim.GetName().GetString().c_str());
+                   throw std::runtime_error(err.c_str());
+                }
+            }
+        }
```
HamedSabri-adsk, line 78: Using lower-level SdfReference to detect internal vs external reference to meet following condition:

`Internal references (references without a file path specified) from the same file, should be renamable and the references string updated. //Error: Unable to rename referenced object [object name]`

test/lib/ufe/testRename.py
```diff
@@ -267,3 +267,39 @@ def testRenameRestrictionOtherLayerOpinions(self):
...
+        leafRefPrim = usdUtils.getPrimFromSceneItem(treebaseItem)
+
+        # leafRefPrim should have an Authored References
+        self.assertTrue(leafRefPrim.HasAuthoredReferences())
+
+        # expect the exception to happen
+        with self.assertRaises(RuntimeError):
+           newName = 'leaf_ref_2_renaming'
+           cmds.rename(newName)
+
```
HamedSabri-adsk, line 304: Added a simple unit test around rename restriction for an external reference.

NOTE: I originally didn't want to add `treeRef.usda` and `treeRef.ma` and my plan was to use already added `tree.ma` and programmatically add internal reference, however AddInternalReference didn't seem to give me what I wanted.

What I wanted was to add a `typeless` prim, append an internal reference, author bunch of attributes and be done.
e.g 
```
    def "leaf_ref_1" (
        append references = </TreeBase/leavesXform/greenLeaf>
    )
    {
        double3 xformOp:translate = (2.0, 2.5, 0)
        uniform token[] xformOpOrder = ["xformOp:translate"]
        double radius = 0.5
    }
```

Here is how I originally planned to add in code:
```
refPrim = stage.DefinePrim('/TreeBase/leaf_ref_1')
refPrim.GetReferences().AddInternalReference('/TreeBase/leavesXform/greenLeaf')

# author refPrim's xform and radius
xform = UsdGeom.Xformable(refPrim)
xform.AddTranslateOp().Set(Gf.Vec3d(2.0, 2.5, 0))

radiusAttr = refPrim.GetAttribute('radius')
radiusAttr.Set(0.5)
```
Instead I get something like this:
```
def Sphere "leaf_ref_1"
{
    float3[] extent = [(-2, -2, -2), (2, 2, 2)]
    color3f[] primvars:displayColor = [(0, 1, 0)]
    double radius = 0.5
    double3 xformOp:translate = (2, 2.5, 0)
    uniform token[] xformOpOrder = ["xformOp:translate"]
}
```

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
ppt-adsk: Your first sentence (before //Error) would make a great code comment right after line 63.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -42,44 +42,49 @@ namespace ufe {
...
                                          possibleTargetLayer->GetDisplayName().c_str());
         throw std::runtime_error(err.c_str());
     }
     else
     {
-        auto layers = MayaUsdUtils::layersWithPrimSpec(prim);
+        // account for internal vs external references
+        if (_prim.HasAuthoredReferences()) {
+            auto primSpec = MayaUsdUtils::getPrimSpecAtEditTarget(_stage, _prim);
+            for (const SdfReference& ref : primSpec->GetReferenceList().GetAddedOrExplicitItems()) {
```
ppt-adsk, line 66: Feels strange that this is going through a loop...  Because in the loop body, the if part of the conditional returns, and the else part of the conditional throws, so there is no way you'll get anything beyond the first item in the loop!

Also, is it possible for _prim.HasAuthoredReferences() to be true, but primSpec->GetReferenceList().GetAddedOrExplicitItmes() to return an empty list?

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -42,44 +42,49 @@ namespace ufe {
...
+                                         _prim.GetName().GetString().c_str(),
                                          possibleTargetLayer->GetDisplayName().c_str());
         throw std::runtime_error(err.c_str());
     }
     else
     {
-        auto layers = MayaUsdUtils::layersWithPrimSpec(prim);
+        // account for internal vs external references
+        if (_prim.HasAuthoredReferences()) {
+            auto primSpec = MayaUsdUtils::getPrimSpecAtEditTarget(_stage, _prim);
```
ppt-adsk, line 65: Given that you've submitted a pull request to Pixar to better handle this functionality, wouldn't it be good to encapsulate this behind a utility function so that this code won't have to change when internal reference testing improves?  The encapsulation would be specially useful since you'll have to support different implementations, depending on the version of USD that's being used.  E.g. something like
bool isInternalReference(primSpec)

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
HamedSabri-adsk: @ppt-adsk  yes that was my first impression about the loop but that's what we have got. 

`Because in the loop body, the if part of the conditional returns, and the else part of the conditional throws, so there is no way you'll get anything beyond the first item in the loop!`

Yes I agree. This is a first pass and I am sure I am going to modify this code as I understand more edge cases.

`Also, is it possible for _prim.HasAuthoredReferences() to be true, but primSpec->GetReferenceList().GetAddedOrExplicitItmes() to return an empty list?`

I don't think that would be the case at least by looking at the low-level code. I could be wrong.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
HamedSabri-adsk: Good point, will make one in the next commit.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
HamedSabri-adsk: @ppt-adsk please see a305af95bbe17a8e7a68e8e807335ccc80fe043f

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
HamedSabri-adsk: @ppt-adsk please see a305af95bbe17a8e7a68e8e807335ccc80fe043f

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -61,18 +61,17 @@ UsdUndoRenameCommand::UsdUndoRenameCommand(const UsdSceneItem::Ptr& srcItem, con
...
-                // this will be empty in the case of an internal reference.
-                if (ref.GetAssetPath().empty()) { 
-                    return;
-                }else{
-                   std::string err = TfStringPrintf("Unable to rename referenced object [%s]", 
-                                                    prim.GetName().GetString().c_str());
-                   throw std::runtime_error(err.c_str());
-                }
+
+            if(MayaUsdUtils::isInternalReference(primSpec)) {
```
ppt-adsk, line 69: Nit-pick: could just be
if (!MayaUsdUtils::isInternalReference(primSpec)) {
    throw blablabla;
}

lib/usd/utils/util.cpp
```diff
@@ -101,4 +101,20 @@ getPrimSpecAtEditTarget(UsdStageWeakPtr stage, const UsdPrim& prim)
...
+bool
+isInternalReference(const SdfPrimSpecHandle& primSpec)
+{
+    bool isInternalRef{false};
+
+    for (const SdfReference& ref : primSpec->GetReferenceList().GetAddedOrExplicitItems()) {
+        // GetAssetPath returns the asset path to the root layer of the referenced layer
+        // this will be empty in the case of an internal reference.
+        if (ref.GetAssetPath().empty()) {
+            isInternalRef = true;
```
ppt-adsk, line 113: Don't you want to
break;
at this point?

HamedSabri-adsk: @ppt-adsk see 610f2ce099cceab4f2727e7df0774aa39d7ea815

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
HamedSabri-adsk: @ppt-adsk see 80aa35e4160b7c35147042f45c49d727ac3289f5

HamedSabri-adsk:
@ppt-adsk please see 3cf2acce328bf62b5f0952aa952b8ba586905ee4

PR #485, opened 2020-05-06
When the stage data changes clear the cache so that the contents of

williamkrick:
out stage data reflect the current state of the proxy shape.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -760,6 +760,7 @@ MayaUsdProxyShapeBase::preEvaluation(const MDGContext& context, const MEvaluatio
         evaluationNode.dirtyPlugExists(loadPayloadsAttr) ||
         evaluationNode.dirtyPlugExists(inStageDataAttr)) {
         _IncreaseUsdStageVersion();
+        UsdMayaStageCache::Clear();
     }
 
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -760,6 +760,7 @@ MayaUsdProxyShapeBase::preEvaluation(const MDGContext& context, const MEvaluatio
         evaluationNode.dirtyPlugExists(loadPayloadsAttr) ||
         evaluationNode.dirtyPlugExists(inStageDataAttr)) {
         _IncreaseUsdStageVersion();
+        UsdMayaStageCache::Clear();
```
williamkrick, line 763: I think the behavior we want is for the contents of outStageDataAttr to be fully dependent on the current values of the attributes that affect it, and this change achieves that behavior. When the outStageDataAttr changes, clear the cache so that exactly that is in the cache matches what the user has asked for.

kxl-adsk: That's too big of a hammer. We can't clear all the caches here. As well, you can't do this if the proxy hasn't loaded the stage (so not all attributes will affect it). 
So we either need to find good rules for when stage can be removed from the case...or maybe there is a way to load the same stage twice, but with different parameters?

PR #488, opened 2020-05-08
Tremblp/maya 104617/add selection test

ppt-adsk:


ppt-adsk:
While debugging a selection issue, I noticed that maya-usd did not have a dedicated selection test that exercised undo / redo.  The selection issue ended up being separate, but the test is still valuable.

lib/mayaUsd/render/mayaToHydra/renderGlobals.cpp
```diff
@@ -177,10 +177,12 @@ void _CreateBoolAttribute(
     _CreateNumericAttribute<bool>(node, attrName, MFnNumericData::kBoolean, defValue);
 }
 
+#if USD_VERSION_NUM >= 2005
```
ppt-adsk, line 180: Fixed unused function warning as error.

PR #489, opened 2020-05-08
MAYA-104214 - As a user, I'd like to add a USD prim to an existing stage

seando-adsk:
#### MAYA-104214 - As a user, I'd like to add a USD prim to an existing stage
* Added new AE templates for basic prim types.
* New ProxyShapeContextOpsHandler to decorate the one from Maya.
* Send new Ufe::ObjectDestroyed notif when USD prim is deleted.
* New context item sub-menu: "Add New Prim" with various prim types. Full undo/redo support thur new command. Added new context ops test "testAddNewPrim".
* Fixed a bug where "visbility" context menu item was added to prims that don't support visibility attribute.


lib/mayaUsd/ae/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ if (CMAKE_UFE_V2_FEATURES_AVAILABLE)
     install(FILES __init__.py DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/python/ufe_ae/usd/nodes)
 
     # Maya Attribute Editor python template files
-    foreach(_SUBDIR base camera material mesh scope shader xform)
+    foreach(_SUBDIR base camera capsule cone cube cylinder material mesh scope shader sphere xform)
```
seando-adsk, line 11: Created AE templates for the missing prim types that are part of the menu. They are all just copies of the existing templates that have everything in the "AEBaseTemplate"

lib/mayaUsd/ufe/Global.cpp
```diff
@@ -96,8 +105,14 @@ MStatus initialize()
...
 	g_MayaHierarchyHandler = Ufe::RunTimeMgr::instance().hierarchyHandler(g_MayaRtid);
-	auto proxyShapeHandler = ProxyShapeHierarchyHandler::create(g_MayaHierarchyHandler);
-	Ufe::RunTimeMgr::instance().setHierarchyHandler(g_MayaRtid, proxyShapeHandler);
+	auto proxyShapeHierHandler = ProxyShapeHierarchyHandler::create(g_MayaHierarchyHandler);
+	Ufe::RunTimeMgr::instance().setHierarchyHandler(g_MayaRtid, proxyShapeHierHandler);
+
+#if UFE_PREVIEW_VERSION_NUM >= 2015
+	g_MayaContextOpsHandler = Ufe::RunTimeMgr::instance().contextOpsHandler(g_MayaRtid);
+	auto proxyShapeContextOpsHandler = ProxyShapeContextOpsHandler::create(g_MayaContextOpsHandler);
+	Ufe::RunTimeMgr::instance().setContextOpsHandler(g_MayaRtid, proxyShapeContextOpsHandler);
```
seando-adsk, line 114: Decorate the Maya context ops handler, just like we do with the Maya hier handler. Note: currently the Maya context ops handler is not used so it's null.

lib/mayaUsd/ufe/ProxyShapeContextOpsHandler.cpp
```diff
@@ -0,0 +1,80 @@
...
+	{
+		// UsdContextOps expects a UsdSceneItem which wraps a prim, so
+		// create one using the pseudo-root and our own path.
+		UsdStageWeakPtr stage = getStage(item->path());
+		if (stage)
+		{
+			auto usdItem = UsdSceneItem::create(item->path(), stage->GetPseudoRoot());
+			auto usdContextOpsHandler = UsdContextOpsHandler::create();
+			auto cOps = usdContextOpsHandler->contextOps(usdItem);
+			UsdContextOps::Ptr usdCOps = std::dynamic_pointer_cast<UsdContextOps>(cOps);
```
seando-adsk, line 58: From Maya the input Ufe::SceneItem will be of an internal type to Maya. We need to create a UsdSceneItem (that has a prim).

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -236,6 +236,13 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
 				Ufe::Scene::notifyObjectDelete(notification);
 			}
 		}
+#if UFE_PREVIEW_VERSION_NUM >= 2015
+		else if (!prim.IsValid())
+		{
+			auto notification = Ufe::ObjectDestroyed(ufePath);
+			Ufe::Scene::notifyObjectDelete(notification);
+		}
+#endif
```
seando-adsk, line 245: New scene notification added in Ufe. When this message is sent the object is no longer valid. Thus we don't use the scene item, but instead just the path.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -54,6 +56,58 @@ class SetVariantSelectionUndoableCommand : public Ufe::UndoableCommand
...
+    {
+        if (_stage) {
+            _stage->RemovePrim(_primPath);
+        }
+    }
+
+    void redo() override
+    {
+        if (_stage) {
+            auto newPrim = _stage->DefinePrim(_primPath, _primToken);
```
seando-adsk, line 99: Create a prim with the specified path and of the specified type.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -100,23 +154,28 @@ Ufe::ContextOps::Items UsdContextOps::getItems(
...
-            auto visibility =
-                std::dynamic_pointer_cast<Ufe::AttributeEnumString>(
-                    attributes->attribute(UsdGeomTokens->visibility));
-            if (visibility) {
-                auto current = visibility->get();
-                const std::string l = (current == UsdGeomTokens->invisible) ?
-                    std::string("Make Visible") : std::string("Make Invisible");
-                items.emplace_back("Toggle Visibility", l);
+        // Top-level items.  Variant sets and visibility. Do not add for gateway type node.
+        if (!fIsAGatewayType) {
```
seando-adsk, line 158: Lines 159-174 were indented inside this new if test. The gateway node will be the mayausd proxy shape.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -100,23 +154,28 @@ Ufe::ContextOps::Items UsdContextOps::getItems(
...
+            if (fPrim.HasVariantSets()) {
+                items.emplace_back(
+                    "Variant Sets", "Variant Sets", Ufe::ContextItem::kHasChildren);
+            }
+            auto attributes = Ufe::Attributes::attributes(sceneItem());
+            if (attributes && attributes->hasAttribute(UsdGeomTokens->visibility)) {
```
seando-adsk, line 164: Added the hasAttribute() test here to fix a bug with prims that don't have a visibility attribute to toggle.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -54,6 +56,58 @@ class SetVariantSelectionUndoableCommand : public Ufe::UndoableCommand
...
+
+            // Build (and store) the path for the new prim with the unique name.
+            PXR_NS::SdfPath usdItemPath = usdSceneItem->prim().GetPath();
+            _primPath = usdItemPath.AppendChild(PXR_NS::TfToken(newPrimName));
+
+            // The type of prim we were asked to create.
+            if (itemPath[1] == "Def")
+                _primToken = TfToken();     // create typeless prim
+            else
+                _primToken = TfToken(itemPath[1]);
```
seando-adsk, line 85: Type of prim to create. Def is typeless.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -54,6 +56,58 @@ class SetVariantSelectionUndoableCommand : public Ufe::UndoableCommand
...
+    : Ufe::UndoableCommand()
+    {
+        // First get the stage from the proxy shape.
+        auto ufePath = usdSceneItem->path();
+        auto segments = ufePath.getSegments();
+        auto dagSegment = segments[0];
+        _stage = MayaUsd::ufe::getStage(Ufe::Path(dagSegment));
+        if (_stage) {
+            // Rename the new prim for uniqueness, if needed.
+            Ufe::Path newUfePath = ufePath + itemPath[1];
```
ppt-adsk, line 74: You've tried this, so clearly it works, but I'm just not sure how :)  In the parenting code, when I parent just below the proxy shape, I needed to take care of the fact that the newly-added component is in a different run-time:
https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp#L53
Why don't you have the same problem?  If you just concatenate the prim name to the existing proxy shape UFE path, you'll get a path that is in the Maya run-time, which is incorrect.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -54,6 +56,58 @@ class SetVariantSelectionUndoableCommand : public Ufe::UndoableCommand
     const std::string fNewSelection;
 };
 
+//! \brief Undoable command for add new prim
+class AddNewPrimUndoableCommand : public Ufe::UndoableCommand
+{
+public:
+    AddNewPrimUndoableCommand(const MayaUsd::ufe::UsdSceneItem::Ptr& usdSceneItem,
+                              const Ufe::ContextOps::ItemPath& itemPath)
```
ppt-adsk, line 64: Looks like you're only using itemPath[1].  If that's the case, couldn't we just pass that in, and give it a more readable name, i.e. const std::string& primType, or something similar?

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -54,6 +56,58 @@ class SetVariantSelectionUndoableCommand : public Ufe::UndoableCommand
...
+            Ufe::Path newUfePath = ufePath + itemPath[1];
+            auto newPrimName = uniqueChildName(usdSceneItem, newUfePath);
+
+            // Build (and store) the path for the new prim with the unique name.
+            PXR_NS::SdfPath usdItemPath = usdSceneItem->prim().GetPath();
+            _primPath = usdItemPath.AppendChild(PXR_NS::TfToken(newPrimName));
+
+            // The type of prim we were asked to create.
+            if (itemPath[1] == "Def")
+                _primToken = TfToken();     // create typeless prim
```
ppt-adsk, line 83: Nit-pick: could be in the constructor's initializer list as

_primToken(itemPath[1] == "Def" ? TfToken() : TfToken(itemPath[1]))

lib/mayaUsd/ufe/UsdContextOps.cpp
ppt-adsk: As I discussed with you, I'm still trying to figure out how we could have done this with a base class, and avoided the type-based conditional...

If we had a base class for the context ops that did not have variant sets and visibility, and a derived class that would have variant sets and visibility first, then would call the base class, would that work?  Ordering would be preserved.

Maybe I'm too attached to that idea, and maybe it's just a minor improvement, but my feeling is that it would help with separation of concerns.  The base class would take care of the proxy shape-only context ops, and the derived class would add in visibility and variant sets.  Let me know what you think.

lib/mayaUsd/ufe/UsdContextOps.cpp
ppt-adsk: Thanks.  For my own education, which ones don't?  The typeless defs?

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -184,6 +255,17 @@ Ufe::UndoableCommand::Ptr UsdContextOps::doOpCmd(const ItemPath& itemPath)
...
+    else if (!itemPath.empty() && (itemPath[0] == "Add New Prim")) {
+        // Operation is to create a new prim of the type specified.
+        if (itemPath.size() != 2u) {
+            TF_CODING_ERROR("Wrong number of arguments");
+            return nullptr;
+        }
+
+        // At this point we know we have 2 arguments to execute the operation.
+        // itemPath[1] contains the new prim type to create.
+        return std::make_shared<AddNewPrimUndoableCommand>(fItem, itemPath);
```
ppt-adsk, line 267: As per above I'd just pass in itemPath[1] 'cause that's all you need.

test/lib/ufe/testContextOps.py
```diff
@@ -140,3 +154,66 @@ def shadingVariant():
         self.assertEqual(shadingVariant(), 'Cue')
 
         cmds.undo()
+
+
+    def testAddNewPrim(self):
```
ppt-adsk, line 159: Thanks for the test!

lib/mayaUsd/ufe/UsdContextOps.cpp
seando-adsk: Typeless = yes, but also shaders, materials

lib/mayaUsd/ufe/UsdContextOps.cpp
ppt-adsk: I agree with you, the base class approach isn't flexible enough to accomodate the ordering we'll need.  Case closed.

lib/mayaUsd/ufe/UsdContextOps.cpp
ppt-adsk: Ah, of course.  Can you add that in as a code comment, to avoid further questions?

lib/mayaUsd/ufe/UsdContextOps.cpp
ppt-adsk: I think it's O.K. like this, but I'd add in a comment somehow to say something like "Hey I know this isn't 100% correct, but all uniqueName does is take the back() of the path, so it's fine."

lib/mayaUsd/ufe/UsdContextOps.cpp
ppt-adsk: O.K.

lib/mayaUsd/ufe/UsdContextOps.cpp
ppt-adsk: O.K.

lib/mayaUsd/ufe/UsdContextOps.cpp
seando-adsk: Done. I also added another code comment for the visibility.

seando-adsk:
I discovered a problem while doing some more local testing. We want to ignore resycn paths that prim paths. See commit [17c9a05](https://github.com/Autodesk/maya-usd/pull/489/commits/17c9a0556105dc5857b6345bc65fe46f55840f50)

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -204,15 +204,21 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
 	auto stage = notice.GetStage();
 	for (const auto& changedPath : notice.GetResyncedPaths())
 	{
+		// When visibility is toggled for the first time or you add a xformop we enter
+		// here with a resync path. However the changedPath is not a prim path, so we
+		// don't care about it. In those cases, the changePath will contain something like:
+		//   "/<prim>.visibility"
+		//   "/<prim>.xformOp:translate"
+		if (!changedPath.IsPrimPath())
+			continue;
```
seando-adsk, line 213: At line 221 we specifically ask for the prim at path. So if this changed path is not a prim path we will get an invalid prim back.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -184,6 +260,17 @@ Ufe::UndoableCommand::Ptr UsdContextOps::doOpCmd(const ItemPath& itemPath)
         return visibility->setCmd(
             current == UsdGeomTokens->invisible ? UsdGeomTokens->inherited : UsdGeomTokens->invisible);
     } // Visibility
+    else if (!itemPath.empty() && (itemPath[0] == "Add New Prim")) {
```
pilarmolinalopez, line 263: Super minor comment: it may be a good idea to keep these operation strings (like "Add New Prim") in a unique place. That way it'd be easier to change if we ever need to. I've seen the "Add New Prim" string already three times during this review.

seando-adsk: Yes good idea. See commit [14d8d4b](https://github.com/Autodesk/maya-usd/pull/489/commits/14d8d4ba1fb39d2bb690ea1d3306b28964551749)

seando-adsk:
Since this all deals with context menu changes, I've also added one more small change in [this commit](https://github.com/Autodesk/maya-usd/pull/489/commits/2045683ac8ecb381d16f9224a24a5e39b11bc7d6) for opening the USD Layer Editor.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -32,6 +32,35 @@
 
 namespace {
 
+// Ufe::ContextItem strings
+// - the "Item" describe the operation to be performed.
+// - the "Label" is used in the context menu (can be localized).
+static constexpr char kUSDLayerEditorItem[] = "USD Layer Editor";
```
ppt-adsk, line 38: Nit-pick: is this some MSVC-specific requirement?  Because otherwise, you're in an unnamed namespace, you don't need the static linkage specifier.

PR #495, opened 2020-05-13
MAYA-104677 Update MayaUSD to use the new version of the prototype API.

williamkrick:
This change replaces the changes in https://github.com/Autodesk/maya-usd/pull/468 to use an updated prototype API in Maya.


williamkrick:
I forgot to remove that _isFallbackShader stuff from the original, I'll update this with another commit that does that.

williamkrick:
Actually I think I should push these changes to https://github.com/Autodesk/maya-usd/pull/468 rather than having a separate pull request.

PR #496, opened 2020-05-13
Support more incoming xformOps

marsupial:
A start at trying to support things other than the USDXformCommonAPI.
In particular, incoming `xformOp:transform` (matrix4d) operations.

Things that would be nice to solve whether using this as a starting point or not:
--
- Support for editing and overlaying on top of existing `xformOp:transform`
- Layering of ops that can completely preserve the incoming state. (With the ability to tag the new ops based on context/department)
 - Undo that fully restores the state prior to the edit. This PR adds ops individually and will remove them from the stack on undo.
 - More customizable base implementation. (The current implementation is scattered across a couple of files and base classes that make it hard to grok the control flow and add features).

Usage
--
Runtime behavior can be adjusted with two env-vars
`UFE_XFORM_OP_SUFFIX=<string>`, will look for and add ops with a suffix in the form of `xformOp:<type>:<string>`

`UFE_XFORM_OP_XFORM=[0|1]`, will add a single `xformOp:transform:$UFE_XFORM_OP_SUFFIX ` to the stack rather than the 5 individual translate, pivot, rotate, inv-pivot, scale.

Currently editing rotation on top of an incoming stack can get wacky, I'm hoping this can be addressed more with UFE-additions in 115, but as it is now, for editing an existing `xformOp:transform`, `UFE_XFORM_OP_SUFFIX=''` and `UFE_XFORM_OP_XFORM='1'` is the most stable as it involves less matrix->euler decomposition and fighting.

Additionally rotatePivot support isn't really included yet.  Adding it shouldn't be too difficult, but not a priority here.

Example file to compare editing off master vs with this PR.
--
```

#usda 1.0
(
    doc = """Generated from Composed Stage of root layer 
"""
)

def Xform "xform1"
{
    matrix4d xformOp:transform = ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, -100, 0, 1) )
    uniform token[] xformOpOrder = ["xformOp:transform"]

    def Sphere "sphere1"
    {
        double radius = 1
        matrix4d xformOp:transform = ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (100, 0, 0, 1) )
        uniform token[] xformOpOrder = ["xformOp:transform"]
    }

    def Cube "cube1"
    {
        double size = 2
        matrix4d xformOp:transform = ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 100, 1) )
        uniform token[] xformOpOrder = ["xformOp:transform"]
    }
}

def Sphere "sphere2"
{
    double radius = 1
    matrix4d xformOp:transform = ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (100, 100, 0, 1) )
    uniform token[] xformOpOrder = ["xformOp:transform"]
}

def Cube "cube2"
{
    double size = 2
    matrix4d xformOp:transform = ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 100, 100, 1) )
    uniform token[] xformOpOrder = ["xformOp:transform"]
}

def Cube "cube4"
{
    double size = 4
}
```


lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.h
```diff
@@ -46,22 +48,22 @@ namespace ufe {
...
 protected:
 
-    UsdTRSUndoableCommandBase(
-        const UsdSceneItem::Ptr& item, double x, double y, double z);
-    ~UsdTRSUndoableCommandBase() = default;
+    UsdTRSUndoableCommandBase(const UsdSceneItem::Ptr& item,
+        GfVec3f vec, UsdGeomXformOp::Type opType);
+
+    UsdTRSUndoableCommandBase(const UsdSceneItem::Ptr& item,
+        GfVec3d vec, UsdGeomXformOp::Type opType);
```
robthebloke, line 60: Specialising funcs based on _what can be converted to GfVec3f/GfVec3d_, isn't the same as specialising based on the type imho. What is the motivation for avoiding const references here? (If there is a reason for it, you probably want to consider __vectorcall for the WIn32 build).

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.h
```diff
@@ -72,38 +74,26 @@ class MAYAUSD_CORE_PUBLIC UsdTRSUndoableCommandBase : public Ufe::Observer,
...
 
 // shared_ptr requires public ctor, dtor, so derive a class for it.
 template<class T>
 struct MakeSharedEnabler : public T {
-    MakeSharedEnabler(
-        const UsdSceneItem::Ptr& item, double x, double y, double z)
-        : T(item, x, y, z) {}
+    template <typename V>
+    MakeSharedEnabler(const UsdSceneItem::Ptr& item, V vec)
+        : T(item, std::move(vec)) {}
```
robthebloke, line 96: std::move cannot move an immovable object.

lib/mayaUsd/ufe/UsdTransform3d.cpp
```diff
@@ -103,126 +103,89 @@ Ufe::SceneItem::Ptr UsdTransform3d::sceneItem() const
...
+	float x{0}, y{0}, z{0};
 	const TfToken xpivot("xformOp:translate:pivot");
 	if (fPrim.HasAttribute(xpivot))
 	{
 		// Initially, attribute can be created, but have no value.
 		GfVec3f v;
 		if (fPrim.GetAttribute(xpivot).Get<GfVec3f>(&v,getTime(path())))
 		{
-			x = v[0]; y = v[1]; z = v[2];
+			std::tie(x, y, z) = { v[0], v[1], v[2] };
```
robthebloke, line 188: How is this an 'improvement'? 
Keep it simple, Keep it stupid, and assume some poor inturn will have to pick up and maintain this code in the future. Make life easy for them, and write clear, dumb, and maintainable code.

lib/mayaUsd/ufe/UsdTranslateUndoableCommand.cpp
```diff
@@ -20,46 +20,34 @@
 MAYAUSD_NS_DEF {
 namespace ufe {
 
-TfToken UsdTranslateUndoableCommand::xlate("xformOp:translate");
-
 UsdTranslateUndoableCommand::UsdTranslateUndoableCommand(
-    const UsdSceneItem::Ptr& item, double x, double y, double z
+    const UsdSceneItem::Ptr& item, GfVec3d translate
```
robthebloke, line 24: What is your reasoning for not using const references?

lib/mayaUsd/ufe/UsdRotateUndoableCommand.h
```diff
@@ -43,34 +43,20 @@ class MAYAUSD_CORE_PUBLIC UsdRotateUndoableCommand : public Ufe::RotateUndoableC
 	UsdRotateUndoableCommand(UsdRotateUndoableCommand&&) = delete;
 	UsdRotateUndoableCommand& operator=(UsdRotateUndoableCommand&&) = delete;
 
+    //! Construct a UsdRotateUndoableCommand.  The command is not executed.
+    UsdRotateUndoableCommand(const UsdSceneItem::Ptr& item, GfVec3f rotate = {0,0,0});
```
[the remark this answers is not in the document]
marsupial: Could you clarify this? Does Maya->UFE->USD support editing of a non XYZ rotation now ?

marsupial:
Most of what's been said isn't an issue for me, but it's a large amount of one-liners, and I'd like to make sure were on the same page on some of this fundamentally before stuff before I'm spending time adapting to how you'd like the code to look.

On that topic, can we move to a more useful prose, comments like  and "Pointless use", "How is this an 'improvement?", "Rotation order ???????" can be a bit _off-putting_ to say the least.

As far as the use of std::move I tend to do that for transient objects so that I don't have to re-do a ton of work if an API changes below me (and yes it has happened). Fine moving to const reference.

lib/mayaUsd/ufe/UsdRotateUndoableCommand.h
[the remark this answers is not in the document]
robthebloke: We've had support for rotation orders for months now (and scale pivots, and rotate axes). It's been blocked in a PR that was never approved by you guys which I've finally given up on, and so now you're repeating all the work we've already done. 

Realistically the only way forward on this is to separate mayaUsd/ufe out into it's own repo so we can ignore these changes I guess?

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.cpp
```diff
@@ -17,41 +17,447 @@
...
-{}
+namespace {
+
+// Placeholder for delivering an op-suffix via env-var or Maya setting
+std::string getOpSuffix()
+{
+    // os.environ['UFE_XFORMOP_SUFFIX']='??' to use a new suffix
+    const char* UFE_XFORMOP_SUFFIX = ::getenv("UFE_XFORM_OP_SUFFIX");
+
+    return UFE_XFORMOP_SUFFIX ? UFE_XFORMOP_SUFFIX : "";
```
robthebloke, line 38: This won't work. We need to support all of the naming conventions, not just one.

robthebloke:
> On that topic, can we move to a more useful prose, comments like and "Pointless use", "How is this an 'improvement?", "Rotation order ???????" can be a bit _off-putting_ to say the least.

Sorry about that, possibly a result of me losing my shit when it comes to this transform stuff. 

How does any of this relate to the discussions we've had about transforms (which mainly just seems to be me shouting into a void?). Which of *THOSE* issues are you trying to solve here? Can we at least first get some discussion going about how this stuff is supposed to work, before you go off and commit a load more changes to the codebase that won't work for us?

What changes are you committing? Who agreed to those changes? What plan are they following? I was not aware anyone had agreed to a plan yet? So why is this code being developed if there is no plan? 

How does this improve upon the code that we've submitted, and you've already rejected? It has clear regressions from our internal build, so of course I'm going to get a bit shitty _(especially given the BS hoops you guys get me to jump through)_.  You don't support rotation orders, you don't support scale pivots, you don't support rotate axes, and most importantly, you have yet to complete, or even start, any design work for this UFE/USD integration. 

This PR will not get approval from me until you show me the design documents that led to this work. The usage of std::move is probably the least of my concerns at this point in time.

lib/mayaUsd/ufe/UsdTransform3d.cpp
robthebloke: Which debugger? (and define 'choke').

kxl-adsk:
@marsupial Thank you for sharing the work with us.

As flexible handling of different transform stacks is of interest to many people, and also important across several interop scenarios, we’d like to form a working group to identify needs, and figure out how we can all contribute to get this done. We’re hoping you can participate in that process.

pilarmolinalopez:
hi @marsupial - thanks for all your work on this! We finally had some time to play with it in pr115 and it's looking very good to us :)

I first tested with the usd test file you attached. Viewport manipulation adds the correct overs to the session layer. I also tested undoing transformations and it worked well! My only comment here is that sometimes the session layer keeps harmless overs. For instance, with the example file, if we manipulate `cube2` and then undo, this is the resulting session layer:
```
#usda 1.0
 
over "cube2"
{
    matrix4d xformOp:transform.timeSamples = {
        1: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 10, 10, 1) ),
    }
}
```
This over is not changing anything since its value matches the original `xformOp:transform` in `cube2`. If I'm not mistaken we get the over because `XformOpUndo::undoIt` is restoring the initial value to a specific time, instead of the default time. However the original xformOp in the usd file has no time samples. In fact, if we manipulate a prim that has no previous xformOp, like `cube4`, when we undo we get an empty over:
```
#usda 1.0
 
over "cube4"
{
    uniform token[] xformOpOrder = []
}
```
I know this is part of a bigger conversation: default time vs current time. Just wanted to leave this info here.

I also tested manipulating existing usd scenes with a custom suffix. This is very important for us since we always use it internally. It worked great. Thanks!

marsupial:
I think everything you're saying is pretty accurate.

The support for time was very preliminary when we realized this needs to be a much larger topic and kind of stopped there (as it involves even more changes to the Maya/UFE layer and the global `getTime` function).

As you alluded to, there's all sorts of caveats in editing time and preserving the default or prior values.  The last time I checked copying all time samples into the session layer would also have to occur, otherwise you're essentially authoring a new stack with one sample at that specific time.

To me it seems that copy -should- occur by default, as a user explicitly clearing all time-samples seems a trivial thing to make more convenient.

kxl-adsk:
With ufe-manipulation branch now merged, there shouldn’t be anything missing in terms of supporting matrix stacks and undo/redo support. With that merge we also show how to use chain of responsibilities to avoid modifying default behaviors and let studios configure manipulation support (more in https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/ufe/UsdTransform3d.md)

I’m closing this PR.

PR #501, opened 2020-05-19
remove support for Maya releases older than 2018

mattyjams:
CMake will enforce that maya-usd be built with at least Maya 2018 (any API version of that release).

@murphyeoin @robthebloke: Just want to make sure this is cool with you, and that you're not still trying to support any Maya releases before 2018, since I included `plugin/al` stuff in this.


CMakeLists.txt
```diff
@@ -91,7 +91,7 @@ if (BUILD_USDMAYA_SCHEMAS)
     include(cmake/jinja.cmake)
 endif()
 
-find_package(Maya REQUIRED)
+find_package(Maya 2018 REQUIRED)
```
HamedSabri-adsk, line 94: @mattyjams  Great work and thanks for tackling this one. All the changes look good to me. just one minor thing:

You are now explicitly adding version number `2018 ` after the <PackageName> follow by REQUIRED keyword.  What happens then here with 2019 or above?

mattyjams: The requested version of "2018" will get passed into the `FindMaya.cmake` module. At the bottom of that module, we specify that `MAYA_APP_VERSION` is the version variable to use, so `find_package_handle_standard_args()` will compare the requested version of "2018" to whatever it found ("2019", "2020", etc.). Since we don't specify `EXACT` to `find_package()`, the Maya module will deem the found version compatible.

PR #503, opened 2020-05-19
MAYA-104401 Add/Clear Reference menu op's

fowlertADSK:
Added "Add Reference" and "Clear All References" menu items to the Outliner popup menu for UFE objects.  This is just a quick initial implementation, a separate Issue will be logged by Design to cover a more complete/advanced workflow.

### DIFF
--- a/lib/mayaUsd/ufe/UsdContextOps.cpp
+++ b/lib/mayaUsd/ufe/UsdContextOps.cpp
@@ -180,30 +180,30 @@ public:
     AddReferenceUndoableCommand(
         const UsdPrim& prim,
         const std::string& filePath
-    ) : fPrim(prim),
-        fSdfRef(),
-        fFilePath(filePath)
+    ) : _prim(prim),
+        _sdfRef(),
+        _filePath(filePath)
     {}
 
     void undo() override { 
-        if (fPrim.IsValid()) {
-            UsdReferences primRefs = fPrim.GetReferences();
-            primRefs.RemoveReference(fSdfRef);
+        if (_prim.IsValid()) {
+            UsdReferences primRefs = _prim.GetReferences();
+            primRefs.RemoveReference(_sdfRef);
         }
     }
 
     void redo() override { 
-        if (fPrim.IsValid()) {
-            fSdfRef = SdfReference(fFilePath);
-            UsdReferences primRefs = fPrim.GetReferences();
-            primRefs.AddReference(fSdfRef);
+        if (_prim.IsValid()) {
+            _sdfRef = SdfReference(_filePath);
+            UsdReferences primRefs = _prim.GetReferences();
+            primRefs.AddReference(_sdfRef);
         }
     }
 
 private:
-    UsdPrim fPrim;
-    SdfReference fSdfRef;
-    const std::string fFilePath;
+    UsdPrim _prim;
+    SdfReference _sdfRef;
+    const std::string _filePath;
 };
 const std::string AddReferenceUndoableCommand::commandName("Add Reference...");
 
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -140,6 +144,97 @@ class AddNewPrimUndoableCommand : public Ufe::UndoableCommand
     PXR_NS::TfToken _primToken;
 };
 
+const char* selectUSDFileScript = R"(
+global proc string SelectUSDFileForAddReference()
+{
+    string $result[] = `fileDialog2 
+        -fileMode 1
+        -caption "Add Reference to USD Prim"
+        -fileFilter "USD Files (*.usd *.usda *.usdc);;*.usd;;*.usda;;*.usdc"`;
```
fowlertADSK, line 153: It's my understanding that .usdz should not be in the list of files that can be referenced, so just adding the other three here..

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -140,6 +144,97 @@ class AddNewPrimUndoableCommand : public Ufe::UndoableCommand
...
+    static const std::string commandName;
+    static const MString cancelRemoval;
+
+    ClearAllReferencesUndoableCommand(
+        const UsdPrim& prim
+    ) : fPrim(prim)
+    {}
+
+    void undo() override { 
+
```
fowlertADSK, line 222: Not currently providing an Undo for clearing all references.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -140,6 +144,97 @@ class AddNewPrimUndoableCommand : public Ufe::UndoableCommand
...
+    if (0 == size($result))
+        return "";
+    else
+        return $result[0];
+}
+SelectUSDFileForAddReference();
+)";
+
+const char* clearAllReferencesConfirmScript = R"(
```
fowlertADSK, line 163: For the file dialog to choose a reference and a confirm box when clearing I'm just adding a quick and dirty MEL script as a string.  Design has been thinking about what are final design should look like, but this should be good enough to start iterating with them on it.

lib/mayaUsd/ufe/UsdContextOps.cpp
ppt-adsk: Is it because it's technically difficult, or because you're waiting for design to think about it?  Perhaps we should add a separate JIRA entry for this one, so that it doesn't fall through the cracks.

ppt-adsk:
Once we get ClearAllReferences undo implemented, it would be good to add an automated test for this code.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -140,6 +144,97 @@ class AddNewPrimUndoableCommand : public Ufe::UndoableCommand
...
+            fSdfRef = SdfReference(fFilePath);
+            UsdReferences primRefs = fPrim.GetReferences();
+            primRefs.AddReference(fSdfRef);
+        }
+    }
+
+private:
+    UsdPrim fPrim;
+    SdfReference fSdfRef;
+    const std::string fFilePath;
```
seando-adsk, line 206: All new code should use coding standard, member variables should have _

lib/mayaUsd/ufe/UsdContextOps.cpp
fowlertADSK: Bit of both actually.  The request here was for something pretty basic, really just enough to get started with testing how the final design should look/work.  If I'd seen a really obvious way of dealing with it then I would have added it, but it doesn't look like there is a straight forward get'er for the complete list of references to be stored (and then put back as part of Undo).

PR #507, opened 2020-05-19
remove unused method warning: _CreateFloatAttribute 

robthebloke:
The usage of _CreateFloatAttribute is conditionally compiled. When using older versions of usd (e.g. 20.2), a compiler warning is generated for the unused method. Conditionally compile the method as well to remove the warning.


kxl-adsk:
Hi @robthebloke, I believe this is already fixed in dev - see https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/render/mayaToHydra/renderGlobals.cpp#L180

PR #508, opened 2020-05-20
Add override flag for setting WORKING_DIRECTORY

HamedSabri-adsk:
Added a new flag to allow overriding `WORKING_DIRECTORY`. If this flag is not set, ${CMAKE_CURRENT_SOURCE_DIR} is used hence the default behavior. 

UFE tests are supposed to run from build directory. We already copy all the python tests and resource to build directory and therefore WORKING_DIRECTORY needs to be set to ${CMAKE_CURRENT_BINARY_DIR}

### DIFF
--- a/cmake/test.cmake
+++ b/cmake/test.cmake
@@ -15,6 +15,7 @@ endfunction()
 #   PYTHON_COMMAND     - Python code to execute; should call sys.exit
 #                        with an appropriate exitcode to indicate success
 #                        or failure.
+#   WORKING_DIRECTORY  - Directory from which the test executable will be called.
 #   COMMAND            - Command line to execute as a test
 #   NO_STANDALONE_INIT - Only allowable with PYTHON_MODULE or
 #                        PYTHON_COMMAND. With those modes, this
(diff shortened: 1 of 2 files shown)


cmake/test.cmake
```diff
@@ -74,6 +75,13 @@ function(mayaUsd_add_test test_name)
             "used with PYTHON_MODULE or PYTHON_COMMAND")
     endif()
 
+    # set the working_dir
+    if(PREFIX_WORKING_DIRECTORY)
+        set(working_dir ${PREFIX_WORKING_DIRECTORY})
+    else()
+        set(working_dir ${CMAKE_CURRENT_SOURCE_DIR})
+    endif()
+
```
HamedSabri-adsk, line 84: Now that I look at this again, I realize this can simplified. I will make a commit to address this.
if(NOT PREFIX_WORKING_DIRECTORY)
      set(working_dir ${CMAKE_CURRENT_SOURCE_DIR})
endif()

elrond79: Hmm... I don't quite understand your fix - I think the `set(working_dir ${PREFIX_WORKING_DIRECTORY})` is needed, isn't it?   Or are there other changes, too? I guess I'll wait for the actual commit.

HamedSabri-adsk: @elrond79  Sorry, Yes ignore my comment here. I am not sure what I was thinking when I wrote this comment. It has been a long day...... 

Everything as is should be fine. Thanks.

PR #510, opened 2020-05-20
[al] Handle anonymous layer in ProxyShape

robthebloke:
Summary
This change allows ProxyShape to correctly load an anonymous root layer if the file path is set to an anonymous layer id. The anonymous root layer is serialised by the layer manager.

Note on file path
Note that the filePath on the ProxyShape is subject to change when an anonymous layer is used. This is necessary because an anonymous layer cannot be re-created with the same exact anon identitifier. This prevents the ProxyShape's file path from deviating from the layer manager serialisation records, and by doing so guarantees fetching the right layer when the Maya scene is reloaded.

Tests
A couple of Python tests are provided to verify this feature.

### DIFF
--- a/plugin/al/docs/layers.md
+++ b/plugin/al/docs/layers.md
@@ -16,7 +16,25 @@ Normally, the default Edit Target in USD will be the Root Layer of the scene, bu
 ###### Modifications for exisiting layers
 After doing in-memory edits to our USD scene changes(typically via Maya) we then translate our USD scene, which is a filepath to the root layer and the serialised content of all the modified in-memory layers that are tracked by the LayerManager, into our renderers scene description for rendering. 
 
-#### Commands 
+###### Serialised anonymous layers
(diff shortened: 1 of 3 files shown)


kxl-adsk:
@robthebloke this change was conflicting with another PR you opened yesterday. Please resolve conflicts.

plugin/al/docs/layers.md
```diff
@@ -16,7 +16,25 @@ Normally, the default Edit Target in USD will be the Root Layer of the scene, bu
 ###### Modifications for exisiting layers
 After doing in-memory edits to our USD scene changes(typically via Maya) we then translate our USD scene, which is a filepath to the root layer and the serialised content of all the modified in-memory layers that are tracked by the LayerManager, into our renderers scene description for rendering. 
 
-#### Commands 
+###### Serialised anonymous layers
+An anonymous root layer can be imported into the ProxyShape by passing the `stageId` into the AL_usdmaya_ProxyShapeImport command. This way the layer manager serialises changes that are made in the root layer without any backing `.usda` or `.usdc` file. As an example this follow script sets up a scene with a single sphere in an anonymous layer, and import that into a ProxyShape.
```
kxl-adsk, line 20: How are these anonymous root layers created? Can they have any sublayers?

PR #512, opened 2020-05-21
[al] Ensure excluded prim changed are correctly handled

robthebloke:
1. Ensure the tagged excluded prims are updated correctly when a meta data value is changed.
2. Rebuild the UsdImagingDelegate used within the VP2 render delegate when the excluded prim paths change. Previous implementation assumed re-populating the delegate would be enough (it actually needs to be rebuilt)

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -281,18 +281,7 @@ void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
         MProfilingScope subProfilingScope(HdVP2RenderDelegate::sProfilerCategory,
             MProfiler::kColorD_L1, "Allocate SceneDelegate");
 
-        // Make sure the delegate name is a valid identifier, since it may
-        // include colons if the proxy node is in a Maya namespace.
-        const std::string delegateName =
-            TfMakeValidIdentifier(
-                TfStringPrintf(
-                    "Proxy_%s_%p",
-                    _proxyShapeData->ProxyShape()->name().asChar(),
-                    _proxyShapeData->ProxyShape()));
-        const SdfPath delegateID =
-            SdfPath::AbsoluteRootPath().AppendChild(TfToken(delegateName));
-
-        _sceneDelegate.reset(new UsdImagingDelegate(_renderIndex.get(), delegateID));
+        const SdfPath delegateID =_ResetSceneDelegate();
 
         _taskController.reset(new HdxTaskController(_renderIndex.get(),
             delegateID.AppendChild(TfToken(TfStringPrintf("_UsdImaging_VP2_%p", this))) ));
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -342,6 +331,12 @@ bool ProxyRenderDelegate::_Populate() {
         MProfilingScope subProfilingScope(HdVP2RenderDelegate::sProfilerCategory,
             MProfiler::kColorD_L1, "Populate");
 
+        // if previously populated, destroy the old delegate to be able to populate the new excluded paths
```
kxl-adsk, line 334: Given we already use ProxyRenderDelegate::_InitRenderDelegate as the only place to detect changes and re-initialize VP2RenderDelegate. How about we leave the creation of scene delegate in there (so no need for _ResetSceneDelegate method) but similar to how IsUsdStageUpToDate check is done in _InitRenderDelegate, we add a check for IsExcludePrimsUpToDate to reset the scene delegate? 

@williamkrick Does this match your thinking?

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/proxy/ProxyShapeMetaData.cpp
```diff
@@ -92,16 +94,15 @@ void ProxyShape::processChangedMetaData(const SdfPathVector& resyncedPaths, cons
         m_lockManager.setInherited(path);
       }
     }
-    m_lockManager.sort();
```
kxl-adsk, line 95: Why this sort is not needed anymore?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
williamkrick: I agree we should keep all of those checks & re-initialization together and they are already in _InitRenderDelegate(). If we can simplify the expression in _Populate to just check _isPopulated and then do the work then I would say move the destructive part out to _initRenderDelegate along with the "extra" part of the expression testing IsUsdStageUpToDate() and IsExcludePrimsUpToDate(). If we can't clean up the expression we might have to re-think the method we have and how we are organizing the code.

The documentation for _InitRenderDelegate will need to be updated, right now it says one time initialization but it is already not a one-time thing. Maybe we should also rename that function to reflect that fact that it is not just initialization but also responding to changes that require re-initialization?

I suppose the real question we are answering is how do we want to organize the code. Should we have a function per object or a function per task? I think what is there is more like a function per task approach, so moving the re-initialize stuff out makes sense.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -15,6 +15,7 @@
 //
 #include "proxyRenderDelegate.h"
 
+#include <maya/M3dView.h>
```
kxl-adsk, line 18: What needs this include?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -319,6 +320,12 @@ void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
             }
         }
     }
+    else
```
kxl-adsk, line 323: GitHub doesn't allow me to place a comment at line 255, but this is where I would expect this check to happen.

Otherwise, I would expect _isInitialized() at line 258 return true and we never get till here to check exclude prims version.

kxl-adsk:
We have opened #821 to address the problem. Closing this PR since there was no action for a long time.

Issue #430, opened 2020-04-14
[MAYA-102232] HumanFemale.walk.usd crashes Maya on creating the stage


Xaia:
**Describe the bug**
Creating the stage with HumanFemale.walk.usd kills Maya instantly.  

**Steps to reproduce**
1. Create stage with HumanFemale.walk.usd - provided by Pixar
2. Observe 

**Attachments**
![kf8YtxD7rf](https://user-images.githubusercontent.com/60515654/79241707-b3235f00-7e73-11ea-949c-d6acd038a52e.gif)

**Specs (if applicable):**
 - Windows 10
 -  Maya PR114

santosg87:
thanks for logging this @Xaia !

we are currently investigating this behavior. we have it logged internally as MAYA-102232.

HdC-adsk:
The fix for the crash is now being review. The general support for USDSkel will be implemented at some other time.

kxl-adsk:
@Xaia Pull Request #514 fixed the issue. Please validate the fix and close this issue. FWIW. If you rely on Maya Preview Releases, then you will have to wait for the next beta release.

kxl-adsk:
We validated this change internally. @Xaia Not sure how much time you will need to validate it, so I'm closing it for now. Feel free to reopen or log a new issue once you get a chance to try it.

PR #514, opened 2020-05-21
Fixing the crash when loading USDSkel

HdC-adsk:
**Issue**
Described in #430 

**Cause**
The crash occurred because it tried to compute smooth normals
from an empty point array due to missing support of USDSkel.

**Solution**
The change will fix the crash itself, and the real support
for USDSkel will be implemented at some other time.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
@@ -888,10 +888,11 @@ void HdVP2Mesh::_UpdateDrawItem(
                 adjacency->GetSharedAdjacencyBuilderComputation(&topology);
             adjacencyComputation->Resolve(); // IS the adjacency updated now?
 
-            // The topology doesn't have to reference all of the points, thus
-            // we compute the number of normals as required by the topology.
+            // Only the points referenced by the topology are used to compute
+            // smooth normals.
             normals = Hd_SmoothNormals::ComputeSmoothNormals(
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -888,10 +888,11 @@ void HdVP2Mesh::_UpdateDrawItem(
...
             adjacencyComputation->Resolve(); // IS the adjacency updated now?
 
-            // The topology doesn't have to reference all of the points, thus
-            // we compute the number of normals as required by the topology.
+            // Only the points referenced by the topology are used to compute
+            // smooth normals.
             normals = Hd_SmoothNormals::ComputeSmoothNormals(
-                adjacency.get(), topology.GetNumPoints(),
+                adjacency.get(),
+                _meshSharedData._points.size(),
```
HdC-adsk, line 895: The points array is empty because VP2RenderDelegate currently doesn't support USDSkel yet. If we passed in the number of points referenced by topology, it would end up with referencing a null array.

The original code tried to minimize the compute, but the same optimization has already been considered by the HdSmoothNormals::ComputeSmoothNormals() method at https://github.com/PixarAnimationStudios/USD/blob/master/pxr/imaging/hd/smoothNormals.cpp#L90, where 
`numAdjPoints` is the number of points referenced by topology and `numPoints` is the size of the points array. The minimum of the two will determine the size of the output normals.

PR #515, opened 2020-05-21
Improved mayaUsd_copyDirectory routine to allow more control over copying dir/files.

HamedSabri-adsk:
Rewrote mayaUsd_copyDirectory routine to allow more control over copying dir/files.

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -100,9 +100,9 @@ endfunction()
 # and add a relative path from the origin to the target.
 function(mayaUsd_add_rpath rpathRef target)
     if(IS_ABSOLUTE "${target}")
-	    # init_rpath calls get_filename_component([...] REALPATH), which does
-		# symlink resolution, so we must do the same, otherwise relative path
-		# determination below will fail.
+        # init_rpath calls get_filename_component([...] REALPATH), which does
+        # symlink resolution, so we must do the same, otherwise relative path
+        # determination below will fail.
         get_filename_component(target "${target}" REALPATH)
         # Make target relative to $ORIGIN (which is the first element in
         # rpath when initialized with mayaUsd_init_rpath()).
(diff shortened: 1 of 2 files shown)


test/lib/ufe/CMakeLists.txt
```diff
@@ -40,18 +40,12 @@ endif()
...
-mayaUsd_copyDirectory(${TARGET_NAME} DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
-                             DIRECTORY ufeTestUtils)
-mayaUsd_copyDirectory(${TARGET_NAME} DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
-                             DIRECTORY ufeTestPlugins)
-mayaUsd_copyFiles(${TARGET_NAME} DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
-                         FILES ${test_script_files})
-mayaUsd_copyFiles(${TARGET_NAME} DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
-                         FILES ${test_support_files})
+# copy all the resources and python scripts to build directory
+mayaUsd_copyDirectory(${TARGET_NAME}
```
ppt-adsk, line 44: Nice!

cmake/utils.cmake
```diff
@@ -256,52 +256,72 @@ endfunction()
...
-     if(PREFIX_DIRECTORY)
-        set(directory ${PREFIX_DIRECTORY})
-        get_filename_component(directory "${directory}" ABSOLUTE)
-        get_filename_component(dir_name "${directory}" NAME)
-    endif()
-    if(NOT PREFIX_DIRECTORY)
-        message(FATAL_ERROR "DIRECTORY keyword is not specified.")
+    # exclude
+    if(NOT PREFIX_EXCLUDE)
+        unset(PREFIX_EXCLUDE)
```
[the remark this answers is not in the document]
HamedSabri-adsk: Hey @ppt-adsk  exactly, I mainly added it to avoid UB when PREFIX_EXCLUDE is not specified, same thing for exclude_files and glob_files.  
This is based on what I was reading the other day:
https://gitlab.kitware.com/cmake/cmake/commit/546a32897839f015440ea0cc6c8f5b486359d26f

test/lib/ufe/CMakeLists.txt
```diff
@@ -40,18 +40,12 @@ endif()
...
-mayaUsd_copyDirectory(${TARGET_NAME} DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
-                             DIRECTORY ufeScripts)
-mayaUsd_copyDirectory(${TARGET_NAME} DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
-                             DIRECTORY ufeTestUtils)
-mayaUsd_copyDirectory(${TARGET_NAME} DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
-                             DIRECTORY ufeTestPlugins)
-mayaUsd_copyFiles(${TARGET_NAME} DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
-                         FILES ${test_script_files})
-mayaUsd_copyFiles(${TARGET_NAME} DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
-                         FILES ${test_support_files})
```
mattyjams, line 54: It looks like using the new form of `mayaUsd_copyDirectory()` is actually going to result in *slightly* different behavior here. I was initially going to suggest that I don't think we need to define `test_support_files` anymore, but I then realized that we're now going to **always** install everything in this directory, even if the Maya release you're building for does not support V2 or a compatible preview version of UFE. Any concerns about that?

HamedSabri-adsk: test_support_files can be removed if not used. This can be cleaned up in another PR.  

We unconditionally always want to install all of python tests and resources from source dir to build directory regardless of Maya or UFE version being used.  We could exclude things from being copied during mayaUsd_copyDirectory but that's redundant and doesn't gain us anything at the end.

@mattyjams does this answer your question?

mattyjams: Yup, thanks. No real concerns from me. I just wanted to raise the point that this isn't exactly equivalent to what was here before. But as you said, what files get installed doesn't really matter as long as we're still only registering the right set of tests, which is definitely still the case. Thanks!

PR #516, opened 2020-05-21
[al] Make TranslatePrim update behaviour consistent with Proxy Shape update behaviour

robthebloke:
When an onObjectsChanged callback causes prims with schema translators to be called, if a translator plugin supports update, the following occurs:

update() is called on the translator
postImport() is called on the translator
generateUniqueKey() is called on the translator by context->updateUniqueKey()
any extraData Plugins have their update and postImport methods called
we import any prims which weren't updateable because they didn't already exist in the scene
We now call the same code from TranslatePrim, so the behaviours are now consistent.

Have also removed a double update/import/teardown call which was happening in TranslatePrim


plugin/al/plugin/AL_USDMayaTestPlugin/py/testTranslators.py
```diff
@@ -353,7 +395,6 @@ def test_variantSwitch_listener_from_different_stage(self):
 
         # Make a dummy stage that mimics prim path found in test data
         otherHandle = tempfile.NamedTemporaryFile(delete=True, suffix=".usda")
-        otherHandle.close()
```
kxl-adsk, line 356: Not sure what was the reason to revert this change, but now this test if failing on Windows with the following error:
```84: ======================================================================
84: ERROR: test_variantSwitch_listener_from_different_stage (testTranslators.TestPythonTranslators)
84: Test listener only responds to changes made to layers found in proxy shape owned stages.
84: ----------------------------------------------------------------------
84: Traceback (most recent call last):
84:   File "testTranslators.py", line 403, in test_variantSwitch_listener_from_different_stage
84:     stage.Export(otherHandle.name)
84: ErrorException:
84:     Error in 'pxrInternal_v0_19__pxrReserved__::SdfTextFileFormat::WriteToFile' at line 327 in file e:\dev\usd\usd\pxr\usd\lib\sdf\textfileformat.cpp : 'Failed to rename temporary file 'c:\users\ligenzk\appdata\local\temp\1/tmpyiz96t.4g4edy' to 'c:\users\ligenzk\appdata\local\temp\1\tmpyiz96t.usda': Access is denied.
84: '```

FWIW. We discussed this different python interpreted behavior in few other tickets, here is one https://github.com/Autodesk/maya-usd/pull/273#issuecomment-589585095

Issue #285, opened 2020-02-14
[MAYA-104542] proxyShape does not draw nodes with assigned purpose


ablev-dwa:
**Describe the bug**
Any USD prim that has a purpose (proxy, render, guide) does not draw in VP2 as a proxyShape using the dev branch build of maya-usd

**Steps to reproduce**
1. usdMaya_CreateUsdProxyShape;
2. Select triangle_proxy.usd
3. It doesn't draw in VP2
4. usdMaya_CreateUsdProxyShape;
5. Select triangle_nopurpose.usd
6. It renders fine in VP2

[triangles.zip](https://github.com/Autodesk/maya-usd/files/4202146/triangles.zip)

**Expected behavior**
Whatever purposes are selected on the proxyShape node should be displayed in VP2

**Specs (if applicable):**
 - Linux CentOS7
 - Maya 2020 (released version)
 - maya-usd commit 567b470a34e690b71adb84ae5628c81ff944697c

mattyjams:
Hey @ablev-dwa! :)

I'm assuming this is when using the VP 2.0 render delegate (`HD_VP2_RENDER_DELEGATE=1`)? If you set `HD_VP2_RENDER_DELEGATE=0`, which should switch you to using the Pixar batch renderer, do you see purpose being respected?

I suspect this may be either related to or the same issue as #261 that we ran into when starting to test the new render delegate.

ablev-dwa:
Ah yes! Setting ```HD_VP2_RENDER_DELEGATE=0``` gets everything to appear. I assumed that was important since it was set in usdMaya.mod but I don't really know what I'm doing here so I didn't think to change it. All of our geometry has a purpose so I was pretty confused when nothing was showing up! Thanks @mattyjams

kxl-adsk:
Indeed this looks very similar to #261. We currently don't support render tags in VP2RenderDelegate, but it should be something not too hard to add given we use HdEngine for the population of VP2 render items.

santosd:
Took a look at this in PR113 and am able to reproduce the issue which doesn't show the proxy with purpose. However, setting the `HD_VP2_RENDER_DELEGATE=0` env doesn't work for me. Switching to Hydra as the renderer in the viewport doesn't work either. @kxl-adsk I am wondering do I need to have the pxr usd plugin installed for that to work?

mattyjams:
[...]

PR #517, opened 2020-05-21
Set the render tags on our dummy render task correctly.

williamkrick:
Trigger enough updates so that we can show/hide items whose visibility changes.

There is a big comment in ProxyRenderDelegate::_UpdateRenderTags() which explains the challenges and the approach taken.

https://github.com/Autodesk/maya-usd/issues/285

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -627,11 +627,11 @@ void HdVP2BasisCurves::Sync(
         _sharedData.bounds.SetMatrix(delegate->GetTransform(id));
     }
 
-    auto* const          param = static_cast<HdVP2RenderParam*>(_delegate->GetRenderParam());
-    ProxyRenderDelegate& drawScene = param->GetDrawScene();
-
     if (HdChangeTracker::IsVisibilityDirty(*dirtyBits, id) || 
         *dirtyBits & HdChangeTracker::DirtyRenderTag) {
+
+        auto* const          param = static_cast<HdVP2RenderParam*>(_delegate->GetRenderParam());
+        ProxyRenderDelegate& drawScene = param->GetDrawScene();
         _sharedData.visible = delegate->GetVisible(id)
             && drawScene.DrawRenderTag(delegate->GetRenderIndex().GetRenderTag(GetId()));
     }
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -627,8 +627,13 @@ void HdVP2BasisCurves::Sync(
         _sharedData.bounds.SetMatrix(delegate->GetTransform(id));
     }
 
-    if (HdChangeTracker::IsVisibilityDirty(*dirtyBits, id)) {
-        _sharedData.visible = delegate->GetVisible(id);
+    auto* const          param = static_cast<HdVP2RenderParam*>(_delegate->GetRenderParam());
+    ProxyRenderDelegate& drawScene = param->GetDrawScene();
+
+    if (HdChangeTracker::IsVisibilityDirty(*dirtyBits, id) || 
+        *dirtyBits & HdChangeTracker::DirtyRenderTag) {
+        _sharedData.visible = delegate->GetVisible(id)
+            && drawScene.DrawRenderTag(delegate->GetRenderIndex().GetRenderTag(GetId()));
     }
 
     *dirtyBits = HdChangeTracker::Clean;
--- a/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
@@ -1305,8 +1305,10 @@ void HdVP2Mesh::_UpdateDrawItem(
         }
     }
 
-    if (itemDirtyBits & HdChangeTracker::DirtyVisibility) {
-        drawItemData._enabled = drawItem->GetVisible();
+    if (itemDirtyBits & HdChangeTracker::DirtyVisibility
+        || itemDirtyBits & HdChangeTracker::DirtyRenderTag) {
+        drawItemData._enabled
+            = drawItem->GetVisible() && drawScene.DrawRenderTag(renderIndex.GetRenderTag(GetId()));
         stateToCommit._enabled = &drawItemData._enabled;
     }
 
--- a/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
@@ -1305,8 +1305,7 @@ void HdVP2Mesh::_UpdateDrawItem(
         }
     }
 
-    if (itemDirtyBits & HdChangeTracker::DirtyVisibility
-        || itemDirtyBits & HdChangeTracker::DirtyRenderTag) {
+    if (itemDirtyBits & (HdChangeTracker::DirtyVisibility | HdChangeTracker::DirtyRenderTag)) {
         drawItemData._enabled
             = drawItem->GetVisible() && drawScene.DrawRenderTag(renderIndex.GetRenderTag(GetId()));
         stateToCommit._enabled = &drawItemData._enabled;
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -177,6 +177,62 @@ namespace
     }
 #endif
 
+// Copied from renderIndex.cpp, the code that does HdRenderIndex::GetDrawItems. But I just want the rprimIds, I don't want to go all the way to draw items.
+struct _FilterParam {
+    const HdRprimCollection& collection;
+    const TfTokenVector&     renderTags;
+    const HdRenderIndex*     renderIndex;
+};
+
+bool _DrawItemFilterPredicate(const SdfPath& rprimID, const void* predicateParam)
+{
+    const _FilterParam* filterParam = static_cast<const _FilterParam*>(predicateParam);
+
+    const HdRprimCollection& collection = filterParam->collection;
+    const TfTokenVector&     renderTags = filterParam->renderTags;
+    const HdRenderIndex*     renderIndex = filterParam->renderIndex;
+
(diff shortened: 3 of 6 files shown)


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -699,6 +702,159 @@ void ProxyRenderDelegate::_UpdateSelectionStates()
...
+        }
+        if (_proxyShapeData->DrawGuidePurpose()) {
+            renderTags.push_back(HdRenderTagTokens->guide);
+        }
+        _taskController->SetRenderTags(renderTags);
+    }
+}
+
+
+// Copied from renderIndex.cpp, the code that does HdRenderIndex::GetDrawItems. But I just want the rprimIds, I don't want to go all the way to draw items.
```
williamkrick, line 786: These are all copied out of USD proper. I don't see a way to do it without executing some code like this. USD also has a few near-duplicate versions of this function. _DrawItemFilterPredicate and _DirtyListFilterPredicate are quite similar in how they filter for renderTags, _DirtyListFilterPredicate just adds the dirty filter as well.

williamkrick: The better way of doing this would be to modify USD. HdRenderIndex::GetDrawItems() kind of does what I want, but it goes a step further in converting the rprimIds into HdDrawItems that I don't want. The functionality I want could get split out into a separate helper, essentially moving _GetFilteredRprims to HdRenderIndex and avoiding the code duplication.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -699,6 +702,159 @@ void ProxyRenderDelegate::_UpdateSelectionStates()
...
+struct _FilterParam {
+    const HdRprimCollection& collection;
+    const TfTokenVector&     renderTags;
+    const HdRenderIndex*     renderIndex;
+};
+
+static bool _DrawItemFilterPredicate(const SdfPath& rprimID, const void* predicateParam)
```
kxl-adsk, line 793: How about we put this method in anonymous namespace we already have at the top of this file? In such case, the static keyword can be ignored.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -699,6 +702,159 @@ void ProxyRenderDelegate::_UpdateSelectionStates()
...
+    //  3: Update the render tags with the new final render tags so that the next
+    //     real call to execute will update all the now visible items.
+    bool renderPurposeChanged, proxyPurposeChanged, guidePurposeChanged;
+    _proxyShapeData->UpdatePurpose(
+        &renderPurposeChanged, &proxyPurposeChanged, &guidePurposeChanged);
+    if (renderPurposeChanged || proxyPurposeChanged || guidePurposeChanged)
+    {
+        // Build the list of render tags which were added or removed (changed)
+        // and the list of render tags which were removed.
+        TfTokenVector changedRenderTags;
```
kxl-adsk, line 738: Let's add profiler instrumentation to this method. This will allow us to keep an eye on performance and scalability.

williamkrick:
ecg-mayausd-branch-preflight-master-public run 106

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -627,8 +627,13 @@ void HdVP2BasisCurves::Sync(
         _sharedData.bounds.SetMatrix(delegate->GetTransform(id));
     }
 
-    if (HdChangeTracker::IsVisibilityDirty(*dirtyBits, id)) {
-        _sharedData.visible = delegate->GetVisible(id);
+    auto* const          param = static_cast<HdVP2RenderParam*>(_delegate->GetRenderParam());
+    ProxyRenderDelegate& drawScene = param->GetDrawScene();
```
HdC-adsk, line 631: These two lines can moved inside the if clause.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1305,8 +1305,10 @@ void HdVP2Mesh::_UpdateDrawItem(
         }
     }
 
-    if (itemDirtyBits & HdChangeTracker::DirtyVisibility) {
-        drawItemData._enabled = drawItem->GetVisible();
+    if (itemDirtyBits & HdChangeTracker::DirtyVisibility
+        || itemDirtyBits & HdChangeTracker::DirtyRenderTag) {
```
HdC-adsk, line 1309: Instead of using logical OR, the expression can be changed to use bitwise OR:

`if (itemDirtyBits  & (HdChangeTracker::DirtyVisibility | HdChangeTracker::DirtyRenderTag))`

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -752,5 +928,37 @@ inline bool ProxyRenderDelegate::ProxyShapeData::IsExcludePrimsUpToDate() const
...
+{
+    return _drawProxyPurpose;
+}
+inline bool ProxyRenderDelegate::ProxyShapeData::DrawGuidePurpose() const
+{
+    return _drawGuidePurpose;
+}
 
-PXR_NAMESPACE_CLOSE_SCOPE
+    PXR_NAMESPACE_CLOSE_SCOPE
```
HdC-adsk, line 964: No indention here.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -161,6 +169,10 @@ class ProxyRenderDelegate : public MHWRender::MPxSubSceneOverride
         void UsdStageUpdated();
         bool IsExcludePrimsUpToDate() const;
         void ExcludePrimsUpdated();
+        void UpdatePurpose(bool* drawRenderPurposeChanged, bool* drawProxyPurposeChanged, bool* _drawGuidePurposeChanged);
```
HdC-adsk, line 172: `_drawGuidePurposeChanged` -> `drawGuidePurposeChanged`

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -752,5 +928,37 @@ inline bool ProxyRenderDelegate::ProxyShapeData::IsExcludePrimsUpToDate() const
 inline void ProxyRenderDelegate::ProxyShapeData::ExcludePrimsUpdated() {
     _excludePrimsVersion = _proxyShape->getExcludePrimPathsVersion();
 }
+inline void ProxyRenderDelegate::ProxyShapeData::UpdatePurpose(
```
HdC-adsk, line 931: Do we want this code to be inline?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -699,6 +758,105 @@ void ProxyRenderDelegate::_UpdateSelectionStates()
...
+            renderTags.push_back(HdRenderTagTokens->proxy);
+        }
+        if (_proxyShapeData->DrawGuidePurpose()) {
+            renderTags.push_back(HdRenderTagTokens->guide);
+        }
+        _taskController->SetRenderTags(renderTags);
+    }
+}
+
+SdfPathVector ProxyRenderDelegate::_GetFilteredRprims(
```
HdC-adsk, line 844: Please add doc for each new method.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -699,6 +758,105 @@ void ProxyRenderDelegate::_UpdateSelectionStates()
...
+    // This function needs to do three things:
+    //  1: Make sure every rprim with a render tag whose visibility changed gets
+    //     marked dirty. This will ensure the upcoming execute call will update
+    //     the visibility of the MRenderItems in MPxSubSceneOverride.
+    //  2: Make a special call to Execute() with only the render tags which have
+    //     been removed. We need this extra call to hide the previously shown
+    //     items.
+    //  3: Update the render tags with the new final render tags so that the next
+    //     real call to execute will update all the now visible items.
+    bool renderPurposeChanged, proxyPurposeChanged, guidePurposeChanged;
```
HdC-adsk, line 787: It seems safer to initialize them to false.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -699,6 +758,105 @@ void ProxyRenderDelegate::_UpdateSelectionStates()
...
+                removedRenderTags.push_back(HdRenderTagTokens->proxy);
+        }
+        if (guidePurposeChanged) {
+            changedRenderTags.push_back(HdRenderTagTokens->guide);
+            if (!_proxyShapeData->DrawGuidePurpose())
+                removedRenderTags.push_back(HdRenderTagTokens->guide);
+        }
+
+        // Mark all the rprims which have a render tag which changed dirty
+        SdfPathVector    rprimsToDirty = _GetFilteredRprims(*(_defaultCollection.get()), changedRenderTags);
```
HdC-adsk, line 816: `*(_defaultCollection.get())` can be replaced with `*_defaultCollection`

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -699,6 +758,105 @@ void ProxyRenderDelegate::_UpdateSelectionStates()
...
+        HdChangeTracker& changeTracker = _renderIndex->GetChangeTracker();
+        for (auto id : rprimsToDirty) {
```
HdC-adsk, line 818: Is auto& better here?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
williamkrick: We only call it once so it's not bloating the code right now.

williamkrick:
ecg-mayausd-branch-preflight-master-public run 109

PR #519, opened 2020-05-21
Use separate UsdStageCache's for stages that loadAll or loadNone. 

williamkrick:
Rename forcePopulate to loadAll to match the enum name. Remove the default parameter value for loadAll so future callers have to make a choice, they will be more likely to choose correctly.

The stage cache does know about different session layers and so it won't hit the cache if the session layers don't match. The interesting code is at https://github.com/PixarAnimationStudios/USD/blob/master/pxr/usd/usd/stage.cpp#L975.

There you can see the call to the cache to FindOneMatching takes all the arguments except for the InitialLoadSet.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -481,7 +481,7 @@ MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
         }
 
         if (SdfLayerRefPtr rootLayer = SdfLayer::FindOrOpen(fileString)) {
-            UsdStageCacheContext ctx(UsdMayaStageCache::Get());
+            UsdStageCacheContext ctx(UsdMayaStageCache::Get(loadSet == UsdStage::InitialLoadSet::LoadAll));
             SdfLayerRefPtr sessionLayer = computeSessionLayer(dataBlock);
             if (sessionLayer) {
                 usdStage = UsdStage::Open(rootLayer,
@@ -498,7 +498,7 @@ MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
         }
         else if (!usdStage) {
             // Create a new stage in memory with an anonymous root layer.
-            UsdStageCacheContext ctx(UsdMayaStageCache::Get());
+            UsdStageCacheContext ctx(UsdMayaStageCache::Get(loadSet == UsdStage::InitialLoadSet::LoadAll));
             usdStage = UsdStage::CreateInMemory("", loadSet);
         }
 
@@ -755,6 +755,8 @@ MayaUsdProxyShapeBase::isStageValid() const
 MStatus
 MayaUsdProxyShapeBase::preEvaluation(const MDGContext& context, const MEvaluationNode& evaluationNode)
 {
+    // Any logic here should have an equivalent implementation in MayaUsdProxyShapeBase::setDependentsDirty().
+
     if (evaluationNode.dirtyPlugExists(excludePrimPathsAttr)) {
         _IncreaseExcludePrimPathsVersion();
     }
@@ -765,6 +767,7 @@ MayaUsdProxyShapeBase::preEvaluation(const MDGContext& context, const MEvaluatio
         evaluationNode.dirtyPlugExists(loadPayloadsAttr) ||
         evaluationNode.dirtyPlugExists(inStageDataAttr)) {
         _IncreaseUsdStageVersion();
+        MayaUsdProxyStageInvalidateNotice(*this).Send();
     }
 
     return MStatus::kSuccess;
@@ -786,6 +789,8 @@ MayaUsdProxyShapeBase::postEvaluation(const  MDGContext& context, const MEvaluat
 MStatus
 MayaUsdProxyShapeBase::setDependentsDirty(const MPlug& plug, MPlugArray& plugArray)
 {
+    // Any logic here should have an equivalent implementation in MayaUsdProxyShapeBase::preEvaluation() or postEvaluation().
+
     // If/when the MPxDrawOverride for the proxy shape specifies
     // isAlwaysDirty=false to improve performance, we must be sure to notify
     // the Maya renderer that the geometry is dirty and needs to be redrawn
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -496,9 +496,8 @@ MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
 
             usdStage->SetEditTarget(usdStage->GetSessionLayer());
         }
-        else if (!usdStage) {
+        else {
             // Create a new stage in memory with an anonymous root layer.
-            UsdStageCacheContext ctx(UsdMayaStageCache::Get(loadSet == UsdStage::InitialLoadSet::LoadAll));
             usdStage = UsdStage::CreateInMemory("", loadSet);
         }
 
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -480,25 +480,32 @@ MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
             loadSet = UsdStage::InitialLoadSet::LoadNone;
         }
 
-        if (SdfLayerRefPtr rootLayer = SdfLayer::FindOrOpen(fileString)) {
+        {
+            // When opening or creating stages we must have an active UsdStageCache.
+            // The stage cache is the only one who holds a strong reference to the
+            // UsdStage. See https://github.com/Autodesk/maya-usd/issues/528 for
+            // more information.
             UsdStageCacheContext ctx(UsdMayaStageCache::Get(loadSet == UsdStage::InitialLoadSet::LoadAll));
-            SdfLayerRefPtr sessionLayer = computeSessionLayer(dataBlock);
-            if (sessionLayer) {
-                usdStage = UsdStage::Open(rootLayer,
-                        sessionLayer,
-                        ArGetResolver().GetCurrentContext(),
-                        loadSet);
-            } else {
-                usdStage = UsdStage::Open(rootLayer,
-                        ArGetResolver().GetCurrentContext(),
-                        loadSet);
-            }
+            
+            if (SdfLayerRefPtr rootLayer = SdfLayer::FindOrOpen(fileString)) {
+                SdfLayerRefPtr sessionLayer = computeSessionLayer(dataBlock);
+                if (sessionLayer) {
+                    usdStage = UsdStage::Open(rootLayer,
+                            sessionLayer,
+                            ArGetResolver().GetCurrentContext(),
+                            loadSet);
+                } else {
+                    usdStage = UsdStage::Open(rootLayer,
+                            ArGetResolver().GetCurrentContext(),
+                            loadSet);
+                }
 
-            usdStage->SetEditTarget(usdStage->GetSessionLayer());
-        }
-        else {
-            // Create a new stage in memory with an anonymous root layer.
-            usdStage = UsdStage::CreateInMemory("", loadSet);
+                usdStage->SetEditTarget(usdStage->GetSessionLayer());
+            }
+            else {
+                // Create a new stage in memory with an anonymous root layer.
+                usdStage = UsdStage::CreateInMemory("", loadSet);
+            }
         }
 
         if (usdStage) {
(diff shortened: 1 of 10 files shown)


lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -765,6 +767,7 @@ MayaUsdProxyShapeBase::preEvaluation(const MDGContext& context, const MEvaluatio
         evaluationNode.dirtyPlugExists(loadPayloadsAttr) ||
         evaluationNode.dirtyPlugExists(inStageDataAttr)) {
         _IncreaseUsdStageVersion();
+        MayaUsdProxyStageInvalidateNotice(*this).Send();
```
williamkrick, line 770: Sean I think you just forgot to add this here?

seando-adsk: Yes I guess. I wasn't aware that what we did in setDependantsDirty needed to have matching here. Under what circumstance would this method be called and not setDependantsDirty?

lib/mayaUsd/python/wrapStageCache.cpp
```diff
@@ -33,7 +33,7 @@ void wrapStageCache()
     class_<UsdMayaStageCache>("StageCache")
 
         .def("Get", &UsdMayaStageCache::Get,
-             (args("forcePopulate") = true),
+             args("loadAll"),
```
seando-adsk, line 36: By removing the default = true, doesn't that mean that anyone who uses the python Get() method now has to provide an argument?

williamkrick: Yes exactly. No callers in C++ ever set the parameter and some were using it wrong, so my assumption is python callers will be having the same type of error. Where should I be searching for existing python scripts to update? I don't want to break existing user scripts if I can avoid it.

lib/mayaUsd/nodes/proxyShapeBase.cpp
williamkrick: If the EM claims the proxyShape node for parallel evaluation then the proxyShape won't get dirty propagation and setDependentsDirty won't be called. preEvaluate & postEvaluate get called when the node is claimed by the EM, so whatever logic was happening during dirty propagation needs to occur there instead.

lib/mayaUsd/python/wrapStageCache.cpp
seando-adsk: I would search this maya-usd repo.

test/lib/testMayaUsdPythonImport.py
```diff
@@ -32,5 +32,5 @@ def testImportModule(self):
         # mayaUsd library also loaded its dependencies (i.e. the core USD
         # libraries). We test the type name as a string to ensure that we're
         # not causing USD libraries to be loaded any other way.
-        stageCache = mayaUsdLib.StageCache.Get()
+        stageCache = mayaUsdLib.StageCache.Get(True)
```
williamkrick, line 35: I believe this is the only python script that uses the StageCache python class I modified. There is also an Animal Logic version of the StageCache in python, most of the uses in the repo are their version that doesn't have a parameter.

lib/mayaUsd/utils/stageCache.h
```diff
@@ -34,7 +34,7 @@ class UsdMayaStageCache
     /// 2 stage caches are maintained; 1 for stages that have been
     /// force-populated, and 1 for stages that have not been force-populated.
```
mattyjams, line 35: Would be good to update the docstring here. Maybe just move the comments added in the cpp into the docstring here instead so that they get included in the generated documentation?

lib/mayaUsd/python/wrapStageCache.cpp
mattyjams: Good catch, @seando-adsk. This is indeed technically an API breaking change for Python clients. Thanks @williamkrick for updating the usage in this repo. We have internal `StageCache` usages that we'll need to update as well though. I'm not sure whether we have a good mechanism for things like this to inform people when updates to client code might be required.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -498,7 +498,7 @@ MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
         }
         else if (!usdStage) {
             // Create a new stage in memory with an anonymous root layer.
-            UsdStageCacheContext ctx(UsdMayaStageCache::Get());
+            UsdStageCacheContext ctx(UsdMayaStageCache::Get(loadSet == UsdStage::InitialLoadSet::LoadAll));
             usdStage = UsdStage::CreateInMemory("", loadSet);
         }
```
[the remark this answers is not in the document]
williamkrick: I think you are right on both ideas. We don't need to check the value of usdStage in the else if, it should just be an else. I traced back through the code in CreateInMemory and I think the intention is we won't get another anonymous layer with  the same name so there is no point it checking the cache for it, or for putting it into the cache. I'll make those changes.

williamkrick:
> 
> 
> Just some minor cosmetic comments, but otherwise looks good to me!
> 
> Worth noting though that any maya-usd client code with parameter-less usages of `mayaUsd.lib.StageCache.Get()` in Python will need to be updated following this change.

I've updated the only test which is calling mayaUsd.lib.StageCache.Get(), I don't see any more uses in maya-usd.

Removing that default value is a breaking change from any users point of view, @kxl-adsk do we have a process to notify customers about changes that may break their existing scripts that are outside of maya-usd? How can I make this as painless as possible for users?

At the end of the day I don't really need to remove the default parameter, so if we want to avoid breaking changes we can do it. We will need to break the "python API" at some point, if we don't have a process in place should we start thinking about it now?

mattyjams:
> Removing that default value is a breaking change from any users point of view, @kxl-adsk do we have a process to notify customers about changes that may break their existing scripts that are outside of maya-usd? How can I make this as painless as possible for users?
> 
> At the end of the day I don't really need to remove the default parameter, so if we want to avoid breaking changes we can do it. We will need to break the "python API" at some point, if we don't have a process in place should we start thinking about it now?
> 
> Flagged this one do-not-merge-yet until we come to a decision on this issue.

Thanks @williamkrick! Yeah, to be sure, I have no objection at all to the API change, and it's obviously easy for calling code to absorb. Just wanted raise the question as to the process, since I'm sure this won't be the first such case of this. It doesn't need to gate this PR, but definitely worth thinking about as we move forward.

williamkrick:
ecg-mayausd-branch-preflight-master-public run 93

lib/mayaUsd/nodes/proxyShapeBase.cpp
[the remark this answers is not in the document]
seando-adsk: @kxl-adsk Do you remember why we added we were using the stage cache here?

kxl-adsk:
We don't currently have any good place to announce such changes. All existing channels won't be able to reach everybody. Something for TSC to discuss - note taken.

For this particular one, I doubt many people will be affected other than Pixar. So if @mattyjams you are cool with it, I wouldn't change plans.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -496,9 +496,8 @@ MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
 
             usdStage->SetEditTarget(usdStage->GetSessionLayer());
         }
-        else if (!usdStage) {
+        else {
             // Create a new stage in memory with an anonymous root layer.
-            UsdStageCacheContext ctx(UsdMayaStageCache::Get());
```
kxl-adsk, line 501: I saw a question from @seando-adsk in my notifications but now I can't find it on GitHub.

It is true that currently we may not have any workflow which would require this stage to be inserted into the cache. At the same time, I don't see what could go wrong if we do insert it and keep the code symmetric with the case where we open the stage from file.

williamkrick: I think you couldn't find it because the conversation got resolved. Here is the link:

https://github.com/Autodesk/maya-usd/pull/519#discussion_r429422220

I unresolved that conversation so lets continue talking there.

lib/mayaUsd/nodes/proxyShapeBase.cpp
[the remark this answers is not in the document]
williamkrick: By @kxl-adsk in another comment thread:

> It is true that currently we may not have any workflow which would require this stage to be inserted into the cache. At the same time, I don't see what could go wrong if we do insert it and keep the code symmetric with the case where we open the stage from file.

I'm not sure having the anonymous stage in the cache is a good idea. Is it possible for two different proxyShapes to open the same anonymous stage at the same time to benefit from the cache? Normally the point of the cache is to prevent loading the stage twice, but if the stage can't be shared then it doesn't help us, but maybe it can hurt us.

@seando-adsk What is the process for an anonymous stage getting a name? To avoid "leaking" stages in the cache we'd need additional code when we rename anonymous stages to remove the anonymous stage from the cache and re-add it with the new name. The cache is an associative container with the stage name as one of the possible keys, so to prevent it from becoming invalid we need the remove & re-add. Actually any time we attempt to rename a stage we'd need to remove & add, depending on how the code works.

williamkrick:
> 
> 
> We don't currently have any good place to announce such changes. All existing channels won't be able to reach everybody. Something for TSC to discuss - note taken.
> 
> For this particular one, I doubt many people will be affected other than Pixar. So if @mattyjams you are cool with it, I wouldn't change plans.

If y'all are happy that this change is minor enough that we don't need to wait on the notification channel being ready I'm fine with that. We just need to finalize the discussion in https://github.com/Autodesk/maya-usd/pull/519#discussion_r429422220 before I'm ready to merge.

lib/mayaUsd/nodes/proxyShapeBase.cpp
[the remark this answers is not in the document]
seando-adsk: I had purposely not given this anonymous stage a name because our design team was still trying to sort how how to name them (taking the new Layer Editor into account). They have now given me the naming as "anonymousLayer1". So I will be making this change soon.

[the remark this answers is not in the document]
williamkrick: If anonymousLayer1 already exists in a different proxyShape then the stage would be named anonymousLayer2 instead? If it re-uses the same name & uses the cache then whatever is already in anonymousLayer1 would show up here as well, which I don't think is intended.

@seando-adsk is there code to rename stages somewhere?

I keep on having more and more questions. What if two proxyShapes share a stage and then one of them renames it? Do they both now share the stage with the updated name? Do they have different stages? And what should be in the stage cache. This issue feels too big for my loadPayloads change here.

I think what we should do is have a separate discussion in the future when we start working on multi-proxy workflows and view this as just an incremental improvement rather than fixing all the issues with stage caching.

williamkrick:
> If y'all are happy that this change is minor enough that we don't need to wait on the notification channel being ready I'm fine with that. We just need to finalize the discussion in [#519 (comment)](https://github.com/Autodesk/maya-usd/pull/519#discussion_r429422220) before I'm ready to merge.

I'm not totally happy with the state of stage caching right now, but I think there is a bigger discussion on stage caching that we need to have before we can continue, and most of the issues are not a high priority for us right now. This change makes things incrementally better, but not perfect, so I feel it can be merged now.

lib/mayaUsd/nodes/proxyShapeBase.cpp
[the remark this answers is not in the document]
ppt-adsk: I agree with your suggestion about discussing multi-proxy workflows.  Sharing stages between multiple proxy shapes feels like a form of instancing, and I know that currently there is UFE code that depends on a strict one to one mapping between proxy shapes and stages, i.e. no sharing of stages between proxy shapes.  We would need to confirm whether rendering code makes this assumption as well.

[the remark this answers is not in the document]
kxl-adsk: The way we are currently using stages with in-memory root layers, I don't think we would run into a problem. BUT I see your point @williamkrick ! Let's proceed with your current change where you removed the stage cache context from the in-memory stage.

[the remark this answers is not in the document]
williamkrick: One automated test failure lead me to the reason we _have_ to have the stage cache: the proxyShape only holds a weak reference to the `UsdStage` and so without the cache holding a strong reference the stage is deleted when the local strong pointer goes out of scope at the end of this function.

I logged https://github.com/Autodesk/maya-usd/issues/528 for the issue of the proxyShape not actually keeping its UsdStage alive, and brought back the stage cache for now.

williamkrick:
ecg-mayausd-branch-preflight-master-public number 105

PR #521, opened 2020-05-25
MAYA-104946 new add ref's menu should only be on prims, not proxy shape

fowlertADSK:


lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -312,14 +312,17 @@ Ufe::ContextOps::Items UsdContextOps::getItems(
...
+
         // Top level item - Add New Prim (for all context op types).
         items.emplace_back(
             kUSDAddNewPrimItem, kUSDAddNewPrimLabel, Ufe::ContextItem::kHasChildren);
 
-        items.emplace_back(AddReferenceUndoableCommand::commandName,
-                            AddReferenceUndoableCommand::commandName);
-        items.emplace_back(ClearAllReferencesUndoableCommand::commandName,
-                            ClearAllReferencesUndoableCommand::commandName);
+        if (!fIsAGatewayType) {
```
fowlertADSK, line 320: Missed the check for proxy shape in my original commit.  Add/Clear References menu items should only appear on the prims.

And I did mean to start another fIsAGatewayType section here instead of adding this one to the first group.  We want the ref's menu items to show up last, not before "Add New Prim" which is the more commonly used item.

seando-adsk: Yes this is fine. We can (and should) order this menu however we want. Not based on gateway node or now.

Issue #525, opened 2020-05-25
MTOH draws above camera guides and gate


marsupial:
**Describe the bug**
When trying to show the camera resolution/gate in the viewport, any delegate will draw on top of whatever Maya did.

**Steps to reproduce**
1. Enable film or resolution gate for the viewport camera.
2. Load **mtoh**
3. Select a delegate

**Expected behavior**
Not to draw on top of the guides

santosd:
@marsupial I tested this out in PR115 and I was not able to replicate the issue. However, I am occasionally hitting a crash when enabling mtoh after enabling the display of the resolution gate.

marsupial:
If you're building from dev, this is no longer an issue and can be closed (#526 & #567).
Just started testing PR115 last week, and am not seeing a crash related to this.

santosd:
Thank you @marsupial I will go ahead and close this out. I will see if others are hitting a crash on my end and if so I will log an issue for that.

PR #526, opened 2020-05-25
[mtoh] Split pre and post scene passes so camera guides go above Hydra.

marsupial:
#525

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/renderOverrideUtils.h
+++ b/lib/mayaUsd/render/mayaToHydra/renderOverrideUtils.h
@@ -38,7 +38,7 @@ public:
 
     MSceneFilterOption renderFilterOverride() override {
         return _drawSelectionOverlay
-                   ? kRenderUIItems
+                   ? kRenderPreSceneUIItems
                    : MHWRender::MSceneRender::renderFilterOverride();
     }
 
@@ -66,12 +66,17 @@ public:
         : MHWRender::MSceneRender(name) {}
 
     MUint64 getObjectTypeExclusions() override {
-        return ~MHWRender::MFrameContext::kExcludeManipulators;
+        // kExcludeHoldOuts is used so that camera-guides are rendered here.
+        return ~(kExcludeManipulators | kExcludeHoldOuts);
     }
 
     MHWRender::MClearOperation& clearOperation() override {
         mClearOperation.setMask(MHWRender::MClearOperation::kClearNone);
         return mClearOperation;
+    }
+
+    MSceneFilterOption renderFilterOverride() override {
+        return kRenderPostSceneUIItems;
     }
 };
 


lib/mayaUsd/render/mayaToHydra/renderOverrideUtils.h
```diff
@@ -66,13 +66,17 @@ class HdMayaManipulatorRender : public MHWRender::MSceneRender {
...
+        return MHWRender::MSceneRender::getObjectTypeExclusions() | ~(kExcludeManipulators | kExcludeHoldOuts);
     }
 
     MHWRender::MClearOperation& clearOperation() override {
         mClearOperation.setMask(MHWRender::MClearOperation::kClearNone);
         return mClearOperation;
     }
+
+    MSceneFilterOption renderFilterOverride() override {
+        return MSceneFilterOption(MHWRender::MSceneRender::renderFilterOverride() | kRenderPostSceneUIItems);
```
HdC-adsk, line 78: Can you return kRenderPostSceneUIItems directly? From API doc, MSceneRender::renderFilterOverride() returns kNoSceneFilterOverride which equals 0.

lib/mayaUsd/render/mayaToHydra/renderOverrideUtils.h
```diff
@@ -66,13 +66,17 @@ class HdMayaManipulatorRender : public MHWRender::MSceneRender {
         : MHWRender::MSceneRender(name) {}
 
     MUint64 getObjectTypeExclusions() override {
-        return ~MHWRender::MFrameContext::kExcludeManipulators;
+        return MHWRender::MSceneRender::getObjectTypeExclusions() | ~(kExcludeManipulators | kExcludeHoldOuts);
```
HdC-adsk, line 69: Is this change relevant to the issue? Why do you exclude holdouts?

`MHWRender::MSceneRender::getObjectTypeExclusions()` returns 0, so there is no need to use it in the bitmask op.

HdC-adsk:
I don't know well enough about mtoh implementation to review the impact to any existing workflow, adding @elrond79 to review too.

marsupial:
I just called through to the base classes as I didn't know they returned 0.  So I can test to see if removing them still works.

The other change is relevant, as kExcludeHoldOuts is what winds up drawing the camera guides from what I've seen.

elrond79:
Haven't tested the code, but the code changes make sense to me, assuming that kExcludeHoldOuts is where camera guides are grouped.  (Which seems odd, but I'll take your word for it!)

HdC-adsk:
Yes, a bit odd. Maybe adding a comment to explain the purpose of excluding holdout would be better.

HdC-adsk:
> 
> 
> I just called through to the base classes as I didn't know they returned 0. So I can test to see if removing them still works.
> 
> The other change is relevant, as kExcludeHoldOuts is what winds up drawing the camera guides from what I've seen.

@marsupial How is the test doing?

marsupial:
Sadly no time! Monday for sure.

kxl-adsk:
@marsupial I see you (force) pushed changes two days ago to the branch. We are not receiving notification for this action from GitHub. For future PRs, please leave a comment in the PR when reviewers should review new changes...or you can re-request the review from a specific person which will send a notification.

PR #527, opened 2020-05-26
MAYA-104964 Registering loaded/unloaded callbacks with the plugin.

fowlertADSK:
Moving existing menu create/delete calls there and adding auto-load of the ufeSupport plugin.

### DIFF
--- a/plugin/adsk/plugin/plugin.cpp
+++ b/plugin/adsk/plugin/plugin.cpp
@@ -109,7 +109,7 @@ MStatus initializePlugin(MObject obj)
 #endif
 
     plugin.registerUI("mayaUsd_pluginUICreation", "mayaUsd_pluginUIDeletion", 
-        "mayaUsd_pluginBatchLoad", "mayaUsd_pluginBatchUnoad");
+        "mayaUsd_pluginBatchLoad", "mayaUsd_pluginBatchUnload");
 
     // As of 2-Aug-2019, these PlugPlugin translators are not loaded
     // automatically.  To be investigated.  A duplicate of this code is in the
--- a/plugin/adsk/plugin/plugin.cpp
+++ b/plugin/adsk/plugin/plugin.cpp
@@ -108,8 +108,8 @@ MStatus initializePlugin(MObject obj)
     }
 #endif
 
-    MGlobal::executeCommand("source \"mayaUsdMenu.mel\"");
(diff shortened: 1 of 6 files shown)


plugin/adsk/plugin/plugin.cpp
```diff
@@ -108,8 +108,8 @@ MStatus initializePlugin(MObject obj)
     }
 #endif
 
-    MGlobal::executeCommand("source \"mayaUsdMenu.mel\"");
-    MGlobal::executeCommand("mayaUsdMenu_loadui");
+    plugin.registerUI("mayaUsd_pluginUICreation", "mayaUsd_pluginUIDeletion", 
+        "mayaUsd_pluginBatchLoad", "mayaUsd_pluginBatchUnoad");
```
fowlertADSK, line 113: Moving anything we do on load and unload into the MEL proc's that are now registered with the plugin.  For now that's the menu creation + autoload of the ufeSupport plugin.

I am also leaving in empty callbacks for when you load/unload in batch mode.  Better to have the scripts in place if/when we need them instead of someone not realizing that MFnPlugin does have separate callbacks in those situations and then trying to code something specific in another place.

plugin/adsk/scripts/mayaUsd_pluginUICreation.mel
```diff
@@ -0,0 +1,15 @@
...
+// Callback registered with the MFnPlugin for when the plugin is loaded.
+// Used to initialize any UI we create.
+{
+    source "mayaUsdMenu.mel";
+    mayaUsdMenu_loadui;
+
+    int $mjv = `about -majorVersion`;
+
+    if (((2021 <= $mjv)) && (!`pluginInfo -q -loaded "ufeSupport"`)) {
+        if (catchQuiet(`loadPlugin "ufeSupport"`)) {
```
fowlertADSK, line 11: I don't think we have an easy way to determine in a MEL script if we've been built with UFE support or not, but I don't think this is a big deal for the ufeSupport plugin.  It's pretty much a standalone thing, and it's currently autoloaded by Maya, so it should not be a bad thing if we try to load it from a MayaUSD plugin that someone built without UFE disabled.
The change I made here is to at least check for a version of Maya starting at 2021, and to spit out a warning if there was a problem loading it.

PR #531, opened 2020-05-27
replace deprecated GetPlatformDefaultHgi() with CreatePlatformDefaultHgi() for USD > 20.05

mattyjams:
This change corresponds to core USD commit https://github.com/PixarAnimationStudios/USD/commit/9d75ff82b700b00ef79552a6ceadbd82666e4b5b

PR #530, opened 2020-05-27
clarify that HdRepr owns the draw items by using std::unique_ptr

mattyjams:
This change corresponds to core USD commit https://github.com/PixarAnimationStudios/USD/commit/8a8ab0b274dccfe335737dff611593dadbfb6a40

PR #534, opened 2020-05-28
Updates to support core USD imaging-related changes in latest dev branch

mattyjams:
This PR includes and replaces PRs #530 and #531.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -1413,12 +1413,11 @@ HdDirtyBits HdVP2BasisCurves::_PropagateDirtyBits(HdDirtyBits bits) const
     // Propagate dirty bits to all draw items.
     for (const std::pair<TfToken, HdReprSharedPtr>& pair : _reprs) {
         const HdReprSharedPtr& repr = pair.second;
+        const auto& items = repr->GetDrawItems();
 #if HD_API_VERSION < 35
-        const HdRepr::DrawItems& items = repr->GetDrawItems();
         for (HdDrawItem* item : items) {
             if (HdVP2DrawItem* drawItem = static_cast<HdVP2DrawItem*>(item)) {
 #else
-        const HdRepr::DrawItemUniquePtrVector &items = repr->GetDrawItems();
         for (const HdRepr::DrawItemUniquePtr &item : items) {
             if (HdVP2DrawItem * const drawItem =
                         static_cast<HdVP2DrawItem*>(item.get())) {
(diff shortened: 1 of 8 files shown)


kxl-adsk:
We are still pretty much locked to 20.05 due to on-going work on Python3. @mattyjams can you confirm that this change should have no effect on us since all the changes should be guarded with USD version check?

mattyjams:
@kxl-adsk: That's right. These changes will continue with past USD releases (including 20.05, the current master branch), as well as the current dev branch, whose commit I identified in the build.md.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -1412,9 +1413,16 @@ HdDirtyBits HdVP2BasisCurves::_PropagateDirtyBits(HdDirtyBits bits) const
     // Propagate dirty bits to all draw items.
     for (const std::pair<TfToken, HdReprSharedPtr>& pair : _reprs) {
         const HdReprSharedPtr& repr = pair.second;
+#if HD_API_VERSION < 35
         const HdRepr::DrawItems& items = repr->GetDrawItems();
```
kxl-adsk, line 1417: Would simple const auto& work in this case?

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -1472,9 +1480,16 @@ void HdVP2BasisCurves::_InitRepr(TfToken const &reprToken, HdDirtyBits *dirtyBit
         _reprs.begin(), _reprs.end(), _ReprComparator(reprToken));
     if (it != _reprs.end()) {
         const HdReprSharedPtr& repr = it->second;
+#if HD_API_VERSION < 35
         const HdRepr::DrawItems& items = repr->GetDrawItems();
```
kxl-adsk, line 1484: Same question about `auto` here.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -573,9 +574,16 @@ HdDirtyBits HdVP2Mesh::_PropagateDirtyBits(HdDirtyBits bits) const {
     // Propagate dirty bits to all draw items.
     for (const std::pair<TfToken, HdReprSharedPtr>& pair : _reprs) {
         const HdReprSharedPtr& repr = pair.second;
+#if HD_API_VERSION < 35
         const HdRepr::DrawItems& items = repr->GetDrawItems();
```
kxl-adsk, line 578: We have some refactoring to do here to avoid duplication of code. Either way, the same comments from basisCurves apply in this file.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
mattyjams: Yeah, I think it probably would.

I'm not the original author of these changes so I'm not certain about their intent, but my guess is that they chose not to use `auto` in order to make the change in API more obvious to the reader. I can pull the `GetDrawItems()` call out of the conditional and use `const auto&` instead if you prefer that though.

mattyjams: Addressed by commit https://github.com/Autodesk/maya-usd/pull/534/commits/fd9aef010ea3702625668fa869e67dec2d2ab962.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
mattyjams: Addressed by commit https://github.com/Autodesk/maya-usd/pull/534/commits/fd9aef010ea3702625668fa869e67dec2d2ab962.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
mattyjams: Addressed by commit https://github.com/Autodesk/maya-usd/pull/534/commits/fd9aef010ea3702625668fa869e67dec2d2ab962.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
kxl-adsk: I was actually hoping it will make a bigger change...but didn't notice pointer vs reference difference. I will give up on it & thx for trying.

PR #537, opened 2020-05-29
MAYA-103697 selection update on new objects

HdC-adsk:
When a prim switches modeling variant, the children of the prim can probably be changed to a different set of prims, with different hierarchy and paths. Thus the UFE selection observer will need to also listen to `Ufe::ObjectAdd` notification for Hydra selection update.

The other part of the commit will update the Hydra selection with respect to the UFE global selection  when creating a new stage or a new prim.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -46,7 +46,9 @@
 #include <ufe/globalSelection.h>
 #include <ufe/observableSelection.h>
 #include <ufe/runTimeMgr.h>
+#include <ufe/scene.h>
 #include <ufe/sceneItem.h>
+#include <ufe/sceneNotification.h>
 #include <ufe/selectionNotification.h>
 #endif
 
@@ -157,9 +159,8 @@ namespace
                 return;
             }
 
-            auto selectionChanged =
-                dynamic_cast<const Ufe::SelectionChanged*>(&notification);
-            if (selectionChanged != nullptr) {
+            if (dynamic_cast<const Ufe::SelectionChanged*>(&notification) ||
+                dynamic_cast<const Ufe::ObjectAdd*>(&notification)) {
                 _proxyRenderDelegate.SelectionChanged();
             }
         }
@@ -361,11 +362,14 @@ void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
 
 #if defined(WANT_UFE_BUILD)
         if (!_ufeSelectionObserver) {
+            _ufeSelectionObserver = std::make_shared<UfeSelectionObserver>(*this);
+
             auto globalSelection = Ufe::GlobalSelection::get();
             if (globalSelection) {
-                _ufeSelectionObserver = std::make_shared<UfeSelectionObserver>(*this);
                 globalSelection->addObserver(_ufeSelectionObserver);
             }
+
+            Ufe::Scene::instance().addObjectAddObserver(_ufeSelectionObserver);
         }
 #else
         // Without UFE, support basic selection highlight at proxy shape level.
@@ -879,11 +883,11 @@ ProxyRenderDelegate::GetPrimSelectionStatus(const SdfPath& path) const
     }
 
     const HdSelection::PrimSelectionState* state = GetPrimSelectionState(path);
-    if (state && state->fullySelected) {
-        return kFullySelected;
+    if (state) {
+        return state->fullySelected ? kFullySelected : kPartiallySelected;
     }
 
-    return state ? kPartiallySelected : kUnselected;
+    return kUnselected;
 }
 
 //! \brief  Query the wireframe color assigned to the proxy shape.
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -142,10 +142,10 @@ namespace
     }
 
 #if defined(WANT_UFE_BUILD)
-    class UfeSelectionObserver : public Ufe::Observer
+    class UfeObserver : public Ufe::Observer
     {
     public:
-        UfeSelectionObserver(ProxyRenderDelegate& proxyRenderDelegate)
+        UfeObserver(ProxyRenderDelegate& proxyRenderDelegate)
             : Ufe::Observer()
             , _proxyRenderDelegate(proxyRenderDelegate)
         {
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -879,11 +883,11 @@ ProxyRenderDelegate::GetPrimSelectionStatus(const SdfPath& path) const
...
 
     const HdSelection::PrimSelectionState* state = GetPrimSelectionState(path);
-    if (state && state->fullySelected) {
-        return kFullySelected;
+    if (state) {
+        return state->fullySelected ? kFullySelected : kPartiallySelected;
     }
 
-    return state ? kPartiallySelected : kUnselected;
+    return kUnselected;
```
HdC-adsk, line 890: Tiny refactoring change here for better looking and logical order.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -186,7 +186,7 @@ class ProxyRenderDelegate : public MHWRender::MPxSubSceneOverride
     std::unique_ptr<UsdImagingDelegate> _sceneDelegate; //!< USD scene delegate
 
     bool                    _isPopulated{ false };      //!< If false, scene delegate wasn't populated yet within render index
-    bool                    _selectionChanged{ false }; //!< Whether there is any selection change or not
+    bool                    _selectionChanged{ true };  //!< Whether there is any selection change or not
```
HdC-adsk, line 189: At the frame creating a proxy shape, we need to update the Hydra selection.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -157,9 +159,8 @@ namespace
                 return;
             }
 
-            auto selectionChanged =
-                dynamic_cast<const Ufe::SelectionChanged*>(&notification);
-            if (selectionChanged != nullptr) {
+            if (dynamic_cast<const Ufe::SelectionChanged*>(&notification) ||
+                dynamic_cast<const Ufe::ObjectAdd*>(&notification)) {
```
HdC-adsk, line 163: So we listen to `ObjectAdd` for Hydra selection update. This will then make sure a new prim or a variant switched prim to be highlighted if it or any of its ancestor is selected.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
HdC-adsk: Because the script to create the stage will select the proxy shape, this change means the new stage will be highlighted, which is what we want (MAYA-102103).

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -361,11 +362,14 @@ void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
 
 #if defined(WANT_UFE_BUILD)
         if (!_ufeSelectionObserver) {
+            _ufeSelectionObserver = std::make_shared<UfeSelectionObserver>(*this);
```
ppt-adsk, line 365: Feels funny to still have it called selection observer, when it's observing both the selection and the scene...  Maybe just plain ufeObserver?

HdC-adsk: Good point. I was thinking about the same but somehow gave up probably I thought its purpose is for selection update. But yes, I can rename it.

HdC-adsk: Unluckily if I rename it to `_ufeObserver`, I hit a VS compiler bug that would produce an error: 

`d:\git\maya\worktrees\main\maya\src\components\usd\maya-usd\lib\mayausd\rend er\vp2renderdelegate\proxyRenderDelegate.h(201): error C2208: 'std::shared_ptr fe_v0::Observer>': no members defined using this type`

And I am not the only one hitting the issue:

https://stackoverflow.com/questions/23914815/what-is-the-error-in-my-shared-ptr-declaration

I will have to rename it to `_ufeChangeObserver`. Or if you have better suggestions?

HdC-adsk: I am renaming it to just `_observer`.

ppt-adsk: Works for me.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -360,12 +361,15 @@ void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
...
+            _observer = std::make_shared<UfeObserver>(*this);
+
             auto globalSelection = Ufe::GlobalSelection::get();
             if (globalSelection) {
-                _ufeSelectionObserver = std::make_shared<UfeSelectionObserver>(*this);
-                globalSelection->addObserver(_ufeSelectionObserver);
+                globalSelection->addObserver(_observer);
             }
+
+            Ufe::Scene::instance().addObjectAddObserver(_observer);
```
kxl-adsk, line 372: Does it make sense to add this object add observer even if there is no global selection list?

HdC-adsk: I am not sure. In what scenario there won't be global selection list?

The check of the global selection list was intended to avoid crash. But maybe it is always non-null.

ppt-adsk: Certainly in Maya there will always be a global selection list.  If the global selection list were null, many, many selection-based workflows would either not work or crash.  It is created during Maya initialization, so we could assert that it is non-null.

HdC-adsk: Even for command line render?

HdC-adsk: Chatted with @ppt-adsk and the answer is yes.

kxl-adsk: Are you then going to switch it to TF_CODING_ERROR?

HdC-adsk: Done with the new commit.

Issue #89, opened 2019-10-30
Coding Error in _LoadTexture


suekieza:
Hey there,

refactoring_sandbox branch, Pulled down today with USD 1d08054

This barrel works just fine in Solaris, and usdview. But imports black(no textures) in maya 2019.1
FYI: I've check and the textures are in the right place. Not sure why maya is looking for them here: /home/suesauer/Documents/maya-usd/lib/render/vp2RenderDelegate/material.cpp 

The error it gives in maya. 

Error CODE:
`// Breakdown for file: /home/suesauer/Documents/LOPS_DEMO_FILES/Library/Assets/Barrels/BarrelsAsset.usd
[-2.14748e+07%](0ms) LoadStage
  [-2.14748e+07%](0ms) PostLoadProcess
    [-2.14748e+07%](0ms) CreateTransformChains
    [-2.14748e+07%](0ms) HuntForNativePrims
    [-2.14748e+07%](0ms) PostImportLogic
    [-2.14748e+07%](0ms) CreatePrims
  [-2.14748e+07%](0ms) OpeningUsdStage
    [-2.14748e+07%](0ms) UsdStageOpen
    [-2.14748e+07%](0ms) OpenRootLayer
    [-2.14748e+07%](0ms) OpeningSessionLayer
 // 
setFilterScript "initialShadingGroup";
// Result: 0 // 
setFilterScript "initialParticleSE";
// Result: 0 // 
setFilterScript "defaultLightSet";
// Result: 1 // 
setFilterScript "defaultObjectSet";
// Result: 1 // 
ls -type mayaUsdProxyShapeBase -long;
// Result: |AL_usdmaya_Proxy|AL_usdmaya_ProxyShape // 
ls -type mayaUsdProxyShapeBase -long;
// Result: |AL_usdmaya_Proxy|AL_usdmaya_ProxyShape // 
// Error: Failed verification: ' image ' -- Coding Error in _LoadTexture at line 275 of /home/suesauer/Documents/maya-usd/lib/render/vp2RenderDelegate/material.cpp // 
// Error: Failed verification: ' image ' -- Coding Error in _LoadTexture at line 275 of /home/suesauer/Documents/maya-usd/lib/render/vp2RenderDelegate/material.cpp // 
// Error: Failed verification: ' image ' -- Coding Error in _LoadTexture at line 275 of /home/suesauer/Documents/maya-usd/lib/render/vp2RenderDelegate/material.cpp // 
[...]

PR #538, opened 2020-05-29
MAYA-101853 Changes necessary for UDIM texture support.

williamkrick:
This change depends on a change in USD, commit [4e426565](https://github.com/PixarAnimationStudios/USD/commit/4e42656543f4e3a313ce31a81c27477d4dcb64b9), which isn't merged into the standard version of USD that we build against, so it won't actually work yet. But the source itself can compile fine without that change and the behavior isn't any worse, so I will commit the changes anyway.

This commit + 4e426565 fixes issue #89

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -430,12 +430,6 @@ MHWRender::MTexture* _LoadTexture(
         return _LoadUdimTexture(path, isColorSpaceSRGB, uvScaleOffset);
 #endif
 
-    // set a default uv scale of (1.0, 1.0) and an offset of (0.0f, 0.0f);
-    uvScaleOffset.append(1.0f);
-    uvScaleOffset.append(1.0f);
-    uvScaleOffset.append(0.0f);
-    uvScaleOffset.append(0.0f);
-
     MHWRender::MRenderer* const renderer = MHWRender::MRenderer::theRenderer();
     MHWRender::MTextureManager* const textureMgr =
         renderer ? renderer->getTextureManager() : nullptr;
@@ -975,8 +969,16 @@ HdVP2Material::_AcquireTexture(const std::string& path)
     HdVP2TextureInfo& info = _textureMap[path];
     info._texture.reset(texture);
     info._isColorSpaceSRGB = isSRGB;
-    info._stScale.Set(uvScaleOffset[0], uvScaleOffset[1]); // The first 2 elements are the scale, the 2nd two elements are the offset.
-    info._stOffset.Set(uvScaleOffset[2], uvScaleOffset[3]);
+    if (uvScaleOffset.length() > 0) {
+        TF_VERIFY(uvScaleOffset.length() == 4);
+        info._stScale.Set(uvScaleOffset[0], uvScaleOffset[1]); // The first 2 elements are the scale
+        info._stOffset.Set(uvScaleOffset[2], uvScaleOffset[3]);// The next two elements are the offset
+    }
+    else {
+        // set a default uv scale of (1.0, 1.0) and an offset of (0.0f, 0.0f);
+        info._stScale.Set(1.0f, 1.0f);
+        info._stOffset.Set(0.0f, 0.0f);
+    }
     return info;
 }
 
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -412,7 +412,7 @@ MHWRender::MTexture* _LoadUdimTexture(
 
     return texture;
 #else
-    #return nullptr;
+    return nullptr;
 #endif
 }
 
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -37,11 +37,13 @@
 #include <pxr/base/gf/vec4f.h>
 #include <pxr/base/tf/diagnostic.h>
 #include <pxr/imaging/glf/image.h>
+#include <pxr/imaging/glf/udimTexture.h>
 #include <pxr/imaging/hd/sceneDelegate.h>
 #include <pxr/usd/ar/packageUtils.h>
 #include <pxr/usd/sdf/assetPath.h>
 #include <pxr/usd/usdHydra/tokens.h>
 #include <pxr/usdImaging/usdImaging/tokens.h>
+#include <pxr/usdImaging/usdImaging/textureUtils.h>
 
 #include "debugCodes.h"
 #include "render_delegate.h"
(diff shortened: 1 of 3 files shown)


HdC-adsk:
Maybe ask QA to upload a RTS test so we know when it is fixed.

williamkrick:
> 
> 
> Maybe ask QA to upload a RTS test so we know when it is fixed.

Good idea, I will follow up in JIRA with George.

lib/mayaUsd/render/vp2RenderDelegate/material.h
```diff
@@ -61,6 +61,8 @@ using HdVP2TextureUniquePtr = std::unique_ptr<
 struct HdVP2TextureInfo
 {
     HdVP2TextureUniquePtr  _texture;          //!< Unique pointer of the texture
+    float                  _stScale[2];       //!< UV scale for tiled textures
+    float                  _stOffset[2];      //!< UV offset for tiled textures
```
HdC-adsk, line 65: Maybe you would like to try GfVec2f.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -823,11 +933,16 @@ HdVP2Material::_AcquireTexture(const std::string& path)
...
+    MFloatArray          uvScaleOffset;
+    MHWRender::MTexture* texture = _LoadTexture(path, isSRGB, uvScaleOffset);
 
     HdVP2TextureInfo& info = _textureMap[path];
     info._texture.reset(texture);
     info._isColorSpaceSRGB = isSRGB;
+    info._stScale[0] = uvScaleOffset[0]; // The first 2 elements are the scale, the 2nd two elements are the offset.
+    info._stScale[1] = uvScaleOffset[1];
+    info._stOffset[0] = uvScaleOffset[2];
+    info._stOffset[1] = uvScaleOffset[3];
```
HdC-adsk, line 945: If you use GfVec2f and also change the signature of _LoadTexture to accept two GfVec2f objects, the code here can be simplified a bit.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -783,6 +883,16 @@ void HdVP2Material::_UpdateShaderInstance(const HdMaterialNetwork& mat)
...
+                    if (status) {
+                        paramName = nodeName + "stScale";
+                        const float* val = &(info._stScale[0]);
+                        status = _surfaceShader->setParameter(paramName, val);
+                    }
+                    if (status) {
+                        paramName = nodeName + "stOffset";
+                        const float* val = &(info._stOffset[0]);
+                        status = _surfaceShader->setParameter(paramName, val);
+                    }
```
HdC-adsk, line 895: If you use GfVec2f, the code can be simplified to:

`status = _surfaceShader->setParameter(paramName, info._stScale.data());`

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
williamkrick: I think I need them in the HdVP2TextureInfo in case we hit the texture cache. Using the GfVec2f type makes sense for sure.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -291,13 +296,108 @@ MHWRender::MSamplerStateDesc _GetSamplerStateDesc(const HdMaterialNode& node)
...
+        return nullptr;
+    }
+
+    MString textureName(path.c_str()); // used for caching, using the string with <UDIM> in it is fine
+    MStringArray tilePaths;
+    MFloatArray tilePositions;
+    // HdSt sets the tile limit to the max number of textures in an array of 2d textures. We don't
+    // want to set it too high because UsdImaging_GetUdimTiles will search for every one on disk.
+    int tileLimit = 100;
+    std::vector<std::tuple<int, TfToken>> tiles = UsdImaging_GetUdimTiles(path, tileLimit);
```
HdC-adsk, line 338: Maybe we need to add a TF_WARN if there are more tiles than 100?

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -291,13 +296,108 @@ MHWRender::MSamplerStateDesc _GetSamplerStateDesc(const HdMaterialNode& node)
...
+#else
+    #return nullptr;
+#endif
+}
+
 //! Load texture from the specified path
 MHWRender::MTexture* _LoadTexture(
     const std::string& path,
-    bool& isColorSpaceSRGB)
+    bool& isColorSpaceSRGB, MFloatArray& uvScaleOffset)
```
HdC-adsk, line 385: Break a new line for new params :)

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
williamkrick: UsdImaging_GetUdimTiles looks for 100 tiles on disk, so I don't directly know if there is a 101st tile to warn about. What I will do is set the tile limit to 100, but ask for 101 tiles. If the 101st tile is present I will emit the warning.

HdC-adsk: I didn't know it is limit from USD side. If so, maybe rename the tileLimit to usdUdimTileLimit to be more specific?

williamkrick:
ecg-mayausd-branch-preflight-master-public run 119

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -291,13 +296,146 @@ MHWRender::MSamplerStateDesc _GetSamplerStateDesc(const HdMaterialNode& node)
...
+    );
+
+    for(unsigned int i=0; i<failedTilePaths.length(); i++)
+    {
+        TF_WARN("Failed to load <UDIM> texture tile %s", failedTilePaths[i].asChar());
+    }
+
+    return texture;
+#else
+    #return nullptr;
```
kxl-adsk, line 415: Typo which causes compilation error with older USD versions.

williamkrick:
MayaUsd - Build External GitHub Branch 121

williamkrick:
MayaUsd - Build External GitHub Branch run 126

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -823,11 +963,22 @@ HdVP2Material::_AcquireTexture(const std::string& path)
...
     info._isColorSpaceSRGB = isSRGB;
+    if (uvScaleOffset.length() > 0) {
+        TF_VERIFY(uvScaleOffset.length() == 4);
+        info._stScale.Set(uvScaleOffset[0], uvScaleOffset[1]); // The first 2 elements are the scale
+        info._stOffset.Set(uvScaleOffset[2], uvScaleOffset[3]);// The next two elements are the offset
+    }
+    else {
+        // set a default uv scale of (1.0, 1.0) and an offset of (0.0f, 0.0f);
+        info._stScale.Set(1.0f, 1.0f);
+        info._stOffset.Set(0.0f, 0.0f);
```
HdC-adsk, line 980: Maybe we'd better initialize the members for HdVP2TextureInfo?

williamkrick:
MayaUsd - Build External GitHub Branch run 128

PR #540, opened 2020-05-29
remove hard-coded "pxr" namespace in favor of PXR_NS in baseExportCommand

mattyjams:
This is a small follow-up fix to #398 and accounts for the case when PXR_USE_NAMESPACES is disabled (which is how Pixar currently builds internally) or when USD is built with anything other than "pxr" as the namespace.

### DIFF
--- a/lib/mayaUsd/commands/baseExportCommand.cpp
+++ b/lib/mayaUsd/commands/baseExportCommand.cpp
@@ -172,7 +172,7 @@ void* MayaUSDExportCommand::creator()
 }
 
 /* virtual */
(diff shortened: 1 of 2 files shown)


mattyjams:
> Looks good, thx for the fix.
> 
> Is there something we should change in our build setup in order to catch this kind of problem? Don't think I've ever heard of anyone not using "pxr" or disabling namespaces

Yeah, we got pretty far down the road building stuff on top of USD before we added C++ namespacing to it, so our internal builds still don't use namespaces.

I'm not sure how we might be able to catch this. My best guess would be some kind of custom git pre-commit hook that does a simple search for hard-coded namespacing like `pxr::`.

PR #541, opened 2020-05-29
expanded the name collision search to choose nonconflciting names for…

TonyKaap:
… duplicated items

When building a list of item names that would cause collisions in new items, include inactive and unloaded items.  (Otherwise new items will be subject to the opinions about the inactive and unloaded items, which is incorrect.)

getChildren() is equivalent to getFilteredChildren(UsdPrimDefaultPredicate);
UsdPrimDefaultPredicate is equivalent to:
UsdPrimIsDefined && UsdPrimIsLoaded && !UsdPrimIsAbstract && UsdPrimIsActive;

So the change here is to stop filtering the unloaded and inactive items from this list of reserved names that is getting built.

This matches usdview's use of unloaded and inactive items in its listing of items inside appController.py => _computeDisplayPredicate(...)

### DIFF
--- a/test/lib/ufe/testDuplicateCmd.py
+++ b/test/lib/ufe/testDuplicateCmd.py
@@ -48,10 +48,6 @@ class DuplicateCmdTestCase(unittest.TestCase):
         if not cls.pluginsLoaded:
             cls.pluginsLoaded = mayaUtils.isMayaUsdPluginLoaded()
 
-    @classmethod    
-    def childrenNames(children):
-        return [str(child.path().back()) for child in children]
-
     def setUp(self):
         ''' Called initially to set up the Maya test environment '''
         # Load plugins
@@ -146,7 +142,8 @@ class DuplicateCmdTestCase(unittest.TestCase):
         self.assertNotIn(ball35DupName, propsChildrenNames)
 
         # MAYA-92264: because of USD bug, redo doesn't work.
-        """return
+        """
+        return
         cmds.redo()
 
         snIter = iter(ufe.GlobalSelection.get())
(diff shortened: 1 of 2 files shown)


test/lib/ufe/testDuplicateCmd.py
```diff
@@ -161,3 +162,36 @@ def childrenNames(children):
...
+            ufe.GlobalSelection.get().append(propChild)
+        cmds.delete()
+
+        worldHierarchy = ufe.Hierarchy.hierarchy(worldItem)
+        worldChildren = worldHierarchy.children()
+        propsHierarchy = ufe.Hierarchy.hierarchy(propsItem)
+        propsChildren = propsHierarchy.children()
+        propsChildrenPostDel = propsHierarchy.children()
+
+        # Duplicate Ball_1, which should create Ball_36, not Ball_2
```
ppt-adsk, line 184: Kinda feels like you should check this, right after line 191, no?  In the spirit of moving assertions out of comments and into code.

TonyKaap: Are you suggesting an explicit string check to verify that it's making Ball_36?  That's probably reasonable.  My primary concern was just to avoid using Ball_2 (or any of the even-numbered prop names, since they've been deactivated).

ppt-adsk: Exactly, yes, and you can then remove the comment.

TonyKaap: I added the test.  For clarify, I left the portion of the comment explaining that I'm duplicating Ball_1, but eliminated the part describing what I hoped it would do -- enforcing that by the assertion.

test/lib/ufe/testDuplicateCmd.py
```diff
@@ -47,7 +47,11 @@ class DuplicateCmdTestCase(unittest.TestCase):
     def setUpClass(cls):
         if not cls.pluginsLoaded:
             cls.pluginsLoaded = mayaUtils.isMayaUsdPluginLoaded()
-    
+
+    @classmethod
```
ppt-adsk, line 51: Unused?

TonyKaap: That's a workflow error on my part.  Trying to juggle too many tasks this morning.  Let me clean that up before you review anymore.

ppt-adsk: :)  No trouble, cheers.

PR #543, opened 2020-06-01
Remove the call to the old multidraw API, and use the new xml based 

williamkrick:
api to enable multidraw consolidation.

The "API" (the XML attribute name) is still being finalized but I wanted this up to show an example of how it is used.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/render_delegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/render_delegate.cpp
@@ -272,9 +272,6 @@ namespace
                 if (TF_VERIFY(shaderMgr)) {
                     shader = shaderMgr->getFragmentShader(
                         _fallbackShaderNames[index], _structOutputName, true);
-#if MAYA_API_VERSION >= 20210000
-                    shader->setWantMultiDrawConsolidation(_diffuseColorParameterName, true);
-#endif
                 }
             }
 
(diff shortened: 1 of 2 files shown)


HdC-adsk:
@williamkrick Sorry I totally missed this review. Do you still need this PR?

HdC-adsk:
It doesn't look like we still need this PR, given we already added the new API on MShaderInstance.

williamkrick:
> 
> 
> @williamkrick Sorry I totally missed this review. Do you still need this PR?



> 
> 
> It doesn't look like we still need this PR, given we already added the new API on MShaderInstance.

We do need this change. The interface on MShaderInstance is going to be removed in favor of marking up the XML describing the shader. This change uses the XML method to make the fallbackShader diffuseColor a multidraw parameter.

This will be merged once PR116 goes out because the changes don't work until the corresponding Maya changes are available. Once PR116 goes out I'll remove the MShaderInstance API from Maya.

HdC-adsk:
I don't understand why we would remove MShaderInstance API. In some cases, plug-ins don't own a shader fragment xml file, they will have to rely on the MShaderInstance API to mark a parameter multi-draw required after they get a stock shader from MShaderManager.

williamkrick:
> 
> 
> I don't understand why we would remove MShaderInstance API. In some cases, plug-ins don't own a shader fragment xml file, they will have to rely on the MShaderInstance API to mark a parameter multi-draw required after they get a stock shader from MShaderManager.

Having the multidraw API on MShaderInstance is confusing. All other MShaderInstance functions affect only that MShaderInstance. If you clone an MShaderInstance then they share the underlying effect, but even if you do something like call MShaderInstance::addInputFragment that will create a new effect and not modify any clones.

The problem with multidraw is the value isn't like a shader parameter that gets stored with the effect instance, it is a property of the shader that gets stored with the effect. So if you did something like clone an MShaderInstance a bunch of times and then set the multidraw on or off, either we'd have to create a new effect each time (and prevent multidraw from working), or have setting the multidraw value on one MShaderInstance affect all the clones.

Really the correct place for the API to be would be on a hypothetical MShader API class that would represent the shader created from XML or another source.

I think the answer for things like stock shaders is they already have the multidraw value set, the user shouldn't have to ask for the good performance manually.

HdC-adsk:
Makes sense. Missing the ARenderEffect concept in the API has been a pain for a few things, including this one. You've convinced me.

PR #552, opened 2020-06-02
Make Maya module compatible with mayapy2 which was introduced in 2021.

HamedSabri-adsk:


HamedSabri-adsk:
Closing this, I am going to make this change as part of a next PR.

PR #555, opened 2020-06-02
Support Python3 for Pixar's UsdMaya and unit tests

HamedSabri-adsk:
This PR adds Python3 compatibility support for Pixar's MayaUSD and unit tests. 

Note: Out of all tests, testUsdExportPackage is not passing right now in python 3. 
See https://github.com/Autodesk/maya-usd/pull/555#discussion_r434170321

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,13 @@ message(STATUS "   PYTHON_INCLUDE_DIRS = ${PYTHON_INCLUDE_DIRS}")
 message(STATUS "   PYTHON_LIBRARIES    = ${PYTHON_LIBRARIES}")
 message(STATUS "   Python_EXECUTABLE   = ${Python_EXECUTABLE}")
 
+# The building of the schemas requires some extra python packages which may or
+# may not be installed. If they aren't we provide an option to add them from
+# a python egg.
+if (BUILD_USDMAYA_SCHEMAS)
+    include(cmake/jinja.cmake)
+endif()
+
 find_package(Maya 2018 REQUIRED)
 find_package(USD REQUIRED)
 include(cmake/usd.cmake)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,13 +90,6 @@ message(STATUS "   PYTHON_INCLUDE_DIRS = ${PYTHON_INCLUDE_DIRS}")
 message(STATUS "   PYTHON_LIBRARIES    = ${PYTHON_LIBRARIES}")
 message(STATUS "   Python_EXECUTABLE   = ${Python_EXECUTABLE}")
 
-# The building of the schemas requires some extra python packages which may or
-# may not be installed. If they aren't we provide an option to add them from
-# a python egg.
-if (BUILD_USDMAYA_SCHEMAS)
-    include(cmake/jinja.cmake)
-endif()
-
 find_package(Maya 2018 REQUIRED)
 find_package(USD REQUIRED)
 include(cmake/usd.cmake)
--- a/build.py
+++ b/build.py
@@ -47,6 +47,8 @@ def PrintError(error):
         traceback.print_exc()
     print("ERROR:", error)
 
+def Python3():
+    return sys.version_info.major == 3
 ############################################################
 def Windows():
     return platform.system() == "Windows"
@@ -558,6 +560,7 @@ if __name__ == "__main__":
       Build directory           {buildDir}
       Install directory         {instDir}
       Variant                   {buildVariant}
+      Python 3:                 {enablePython3}
       Python Debug              {debugPython}
       CMake generator           {cmakeGenerator}"""
 
@@ -588,6 +591,7 @@ if __name__ == "__main__":
         ctestArgs=context.ctestArgs,
         buildVariant=BuildVariant(context),
         debugPython=("On" if context.debugPython else "Off"),
+        enablePython3=("On" if Python3() else "Off"),
         cmakeGenerator=("Default" if not context.cmakeGenerator
                         else context.cmakeGenerator)
     )
--- a/cmake/modules/FindMaya.cmake
+++ b/cmake/modules/FindMaya.cmake
@@ -256,7 +256,20 @@ endif()
 # swtich between mayapy and mayapy2
 set(mayapy_exe mayapy)
 if(${MAYA_APP_VERSION} STRGREATER_EQUAL "2021")
-    if(NOT BUILD_WITH_PYTHON_3)
+    # check to see if we have a mayapy2 executable
+    find_program(MAYA_PY_EXECUTABLE2
+            mayapy2
+        HINTS
+            "${MAYA_LOCATION}"
+            "$ENV{MAYA_LOCATION}"
+            "${MAYA_BASE_DIR}"
+        PATH_SUFFIXES
+            Maya.app/Contents/bin/
+            bin/
+        DOC
+            "Maya's Python executable path"
+    )
+    if(NOT BUILD_WITH_PYTHON_3 AND MAYA_PY_EXECUTABLE2)
         set(mayapy_exe mayapy2)
     endif()
 endif()
--- a/cmake/modules/FindMaya.cmake
+++ b/cmake/modules/FindMaya.cmake
@@ -239,21 +239,7 @@ find_program(MAYA_EXECUTABLE
         "Maya's executable path"
 )
 
-find_program(MAYA_PY_EXECUTABLE
-        mayapy
-    HINTS
-        "${MAYA_LOCATION}"
-        "$ENV{MAYA_LOCATION}"
-        "${MAYA_BASE_DIR}"
-    PATH_SUFFIXES
-        Maya.app/Contents/bin/
-        bin/
-    DOC
-        "Maya's Python executable path"
-)
-
 if(MAYA_INCLUDE_DIRS AND EXISTS "${MAYA_INCLUDE_DIR}/maya/MTypes.h")
-
     # Tease the MAYA_API_VERSION numbers from the lib headers
     file(STRINGS ${MAYA_INCLUDE_DIR}/maya/MTypes.h TMP REGEX "#define MAYA_API_VERSION.*$")
     string(REGEX MATCHALL "[0-9]+" MAYA_API_VERSION ${TMP})
@@ -265,8 +251,28 @@ if(MAYA_INCLUDE_DIRS AND EXISTS "${MAYA_INCLUDE_DIR}/maya/MTypes.h")
     else()
         string(SUBSTRING ${MAYA_API_VERSION} "0" "4" MAYA_APP_VERSION)
     endif()
-
 endif()
+
+# swtich between mayapy and mayapy2
+set(mayapy_exe mayapy)
+if(${MAYA_APP_VERSION} STRGREATER_EQUAL "2021")
+    if(NOT BUILD_WITH_PYTHON_3)
+        set(mayapy_exe mayapy2)
+    endif()
+endif()
+
+find_program(MAYA_PY_EXECUTABLE
+        ${mayapy_exe}
+    HINTS
+        "${MAYA_LOCATION}"
+        "$ENV{MAYA_LOCATION}"
+        "${MAYA_BASE_DIR}"
+    PATH_SUFFIXES
+        Maya.app/Contents/bin/
+        bin/
+    DOC
+        "Maya's Python executable path"
+)
 
 # handle the QUIETLY and REQUIRED arguments and set MAYA_FOUND to TRUE if
 # all listed variables are TRUE
--- a/plugin/pxr/cmake/macros/Public.cmake
+++ b/plugin/pxr/cmake/macros/Public.cmake
@@ -190,7 +190,6 @@ function(pxr_setup_python)
     install(
         FILES
           ${CMAKE_CURRENT_BINARY_DIR}/__init__.py
-          ${CMAKE_CURRENT_BINARY_DIR}/__init__.pyc
         DESTINATION
           ${INSTALL_DIR_SUFFIX}/${installPrefix}
     )
--- a/plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeDrawPerformance.py
+++ b/plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeDrawPerformance.py
@@ -15,6 +15,8 @@
 # limitations under the License.
 #
 
+from future.utils import iteritems
+
 from pxr import UsdMaya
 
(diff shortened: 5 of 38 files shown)


build.py
```diff
@@ -588,6 +591,7 @@ def __init__(self, args):
         ctestArgs=context.ctestArgs,
         buildVariant=BuildVariant(context),
         debugPython=("On" if context.debugPython else "Off"),
+        enablePython3=("On" if Python3() else "Off"),
```
HamedSabri-adsk, line 594: Simple change to output Python3 state in the output console.

CMakeLists.txt
```diff
@@ -90,13 +90,6 @@ message(STATUS "   PYTHON_INCLUDE_DIRS = ${PYTHON_INCLUDE_DIRS}")
 message(STATUS "   PYTHON_LIBRARIES    = ${PYTHON_LIBRARIES}")
 message(STATUS "   Python_EXECUTABLE   = ${Python_EXECUTABLE}")
 
-# The building of the schemas requires some extra python packages which may or
-# may not be installed. If they aren't we provide an option to add them from
-# a python egg.
-if (BUILD_USDMAYA_SCHEMAS)
-    include(cmake/jinja.cmake)
-endif()
-
```
HamedSabri-adsk, line 99: @seando-adsk  This was a bit of a surprise since this code never is called when the clean build happens. 

BUILD_USDMAYA_SCHEMAS is an option that is set in AL.  This value is not cached until AL's CmakeLists is called which happens later.

cmake/modules/FindMaya.cmake
```diff
@@ -265,9 +251,29 @@ if(MAYA_INCLUDE_DIRS AND EXISTS "${MAYA_INCLUDE_DIR}/maya/MTypes.h")
...
+        "${MAYA_LOCATION}"
+        "$ENV{MAYA_LOCATION}"
+        "${MAYA_BASE_DIR}"
+    PATH_SUFFIXES
+        Maya.app/Contents/bin/
+        bin/
+    DOC
+        "Maya's Python executable path"
+)
+
```
HamedSabri-adsk, line 276: Added a switch to handle mayapy executable name in dual Maya python.

plugin/pxr/cmake/macros/Public.cmake
```diff
@@ -190,7 +190,6 @@ function(pxr_setup_python)
     install(
         FILES
           ${CMAKE_CURRENT_BINARY_DIR}/__init__.py
-          ${CMAKE_CURRENT_BINARY_DIR}/__init__.pyc
```
HamedSabri-adsk, line 193: There is no need to install compiled bytecode for __init__  either for python2 or 3.

In python3 compiled byte-code  are stored in `__pycache__`, that means we have to have copy the compiled bytecode from `__pycache__` which is not really necessary, hence removing this all together.

plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeDrawPerformance.py
```diff
@@ -47,7 +49,7 @@ def setUpClass(cls):
     @classmethod
     def tearDownClass(cls):
         statsOutputLines = []
-        for profileScopeName, elapsedTime in cls._profileScopeMetrics.iteritems():
+        for profileScopeName, elapsedTime in iteritems(cls._profileScopeMetrics):
```
HamedSabri-adsk, line 52: In python3, dict.iteritems is no longer available. Using python.future function iteritems()

Note: mayapy shipped with Maya already comes with `future` package installed

plugin/pxr/maya/lib/usdMaya/testenv/testUsdExportPackage.py
```diff
@@ -24,7 +23,6 @@
 
 from pxr import Usd, UsdGeom
 
-
```
HamedSabri-adsk, line 27: @mattyjams  This test is still failing in Python3 on Windows:

The reason it fails is because GetFileNames which is simply a list of strings is different from the expected list.

```
zipFile = Usd.ZipFile.Open(usdFile)
fileNames = zipFile.GetFileNames()
self.assertEqual(fileNames, [
    "MyAwesomePackage.usdc",
    "ReferenceModel.usda",
    "BaseModel.usda",
    "card.png"
])
```
After some debugging , I noticed the list entries in the python 3 version have the path to temp folder which is really bizarre:
```
['Users/sabrih/AppData/Local/Temp/tmpp7c6l4yj/MyAwesomePackage.usdc', 
'Users/sabrih/AppData/Local/Temp/tmpp7c6l4yj/ReferenceModel.usda', 
'Users/sabrih/AppData/Local/Temp/tmpp7c6l4yj/BaseModel.usda', 
'Users/sabrih/AppData/Local/Temp/tmpp7c6l4yj/card.png']
```

VS the python 2 version doesn't have this temporary paths:

`['MyAwesomePackage.usdc', 'ReferenceModel.usda', 'BaseModel.usda', 'card.png']`

To me it looks like something to do with `UsdZipFile`.

plugin/pxr/maya/lib/usdMaya/testenv/testUsdExportPointInstancer.py
```diff
@@ -340,6 +341,10 @@ def testMashInstancePaths(self):
         self._TestInstancePaths("MASH1_Instancer")
 
     def testMashVisibility(self):
+
+        # assertCountEqual in python 3 is equivalent to assertItemsEqual
+        if sys.version_info[0] >= 3:
+            self.assertItemsEqual = self.assertCountEqual
```
HamedSabri-adsk, line 347: Python 3 no longer has assertItemsEqual and instead this function has been renamed to assertCountEqual.

added `self.assertItemsEqual = self.assertCountEqual` for compatibility between Python 2 and Python 3

plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeDuplicatePerformance.py
```diff
@@ -15,6 +15,8 @@
 # limitations under the License.
 #
 
+from future.utils import iteritems
```
mattyjams, line 18: Is the `iteritems(cls._profileScopeMetrics)` change from testProxyShapeDrawPerformance.py above missing in `tearDownClass()` below?

These tests don't yet run with CMake/CTest, so I don't expect that they'd show up in your test results.

plugin/pxr/maya/lib/usdMaya/testenv/testUsdExportPackage.py
mattyjams: Hmm, if that's the case, then it seems like that might be an issue with core USD?

Can you reproduce this in a non-Maya context, and if so, could you open an issue against the core USD GitHub project, if so?

plugin/pxr/maya/lib/usdMaya/testenv/testUsdImportXforms.py
```diff
@@ -122,8 +125,7 @@ def testImportMayaXformVariations(self):
             attrVals = {}
             allNodes.append(node)
             allExpected[node] = attrVals
-            for enabled, (attr, (valMin, valMax)) in itertools.izip(enabledArray,
-                                                                    ATTRS.iteritems()):
```
mattyjams, line 126: Does anything else still use `itertools`, or can we also remove that import from the top of this function?

plugin/pxr/maya/lib/usdMaya/testenv/testUsdMayaDiagnosticDelegate.py
```diff
@@ -144,10 +147,13 @@ def testBatching_DelegateRemoved(self):
...
             cmds.unloadPlugin('pxrUsd', force=True)
 
-            for i in xrange(5):
+            for i in range(5):
                 Tf.Status("no delegate, this will be lost %d" % i)
         log = self._StopRecording()
 
+        # assertCountEqual in python 3 is equivalent to assertItemsEqual
+        if sys.version_info[0] >= 3:
+            self.assertItemsEqual = self.assertCountEqual
```
mattyjams, line 156: Maybe do this in `setUpClass()` to avoid repeating it?

plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeDuplicatePerformance.py
HamedSabri-adsk: Hmm, I missed one???  See e594505ae0d7ba2258f47edfb48bad0b558c2424

Correct, it looks like we are ignoring the tests in pxrUsdMayaGL and that is why I didn't see this error.

Anyways, all tests in usdMaya pass successful except one that I mentioned earlier.

image

plugin/pxr/maya/lib/usdMaya/testenv/testUsdImportXforms.py
HamedSabri-adsk: Yes, itertools is still used. See line 121:

` for i, enabledArray in enumerate(itertools.product((False, True), repeat=len(ATTRS))):`

plugin/pxr/maya/lib/usdMaya/testenv/testUsdMayaDiagnosticDelegate.py
HamedSabri-adsk: See e594505ae0d7ba2258f47edfb48bad0b558c2424

plugin/pxr/maya/lib/usdMaya/testenv/testUsdImportXforms.py
mattyjams: Ah, duh. Sorry!

plugin/pxr/maya/lib/usdMaya/testenv/testUsdExportPackage.py
HamedSabri-adsk: @mattyjams  see 28f6222c0e0532115fc03c1da4fa2812313aa79b

I ran part of the test in question in a non-Maya context both in Python 2 and 3.  In my test I am **not** calling `cmds.usdExport `and as it shown below in both cases fileNames do contain temporary paths in the names. So I don't think this has anything to do with UsdZipFile. 

I suspect it either has something to do with `os.path.abspath` or the call to `usdExport`. For now I have disabled this assert for Python 3 until further investigation.

Python 3
```
Python 3.7.7 (tags/v3.7.7:d7c567b08f, Mar 10 2020, 10:41:24) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> from pxr import Usd, UsdGeom
>>> usdFile = os.path.abspath('C:\\Users\\sabrih\\AppData\\Local\\Temp\\tmpjwhzno7b\\MyAwesomePackage.usdz')
>>> zipFile = Usd.ZipFile.Open(usdFile)
>>> fileNames = zipFile.GetFileNames()
>>> print(fileNames)
['Users/sabrih/AppData/Local/Temp/tmpjwhzno7b/MyAwesomePackage.usdc', 'Users/sabrih/AppData/Local/Temp/tmpjwhzno7b/ReferenceModel.usda', 'Users/sabrih/AppData/Local/Temp/tmpjwhzno7b/BaseModel.usda', 'Users/sabrih/AppData/Local/Temp/tmpjwhzno7b/card.png']
```

Python 2
```
Python 2.7.15 (v2.7.15:ca079a3ea3, Apr 30 2018, 16:30:26) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> from pxr import Usd, UsdGeom
>>> usdFile = os.path.abspath('C:\\Users\\sabrih\\AppData\\Local\\Temp\\tmpjwhzno7b\\MyAwesomePackage.usdz')
>>> zipFile = Usd.ZipFile.Open(usdFile)
>>> fileNames = zipFile.GetFileNames()
>>> print(fileNames)
['Users/sabrih/AppData/Local/Temp/tmpjwhzno7b/MyAwesomePackage.usdc', 'Users/sabrih/AppData/Local/Temp/tmpjwhzno7b/ReferenceModel.usda', 'Users/sabrih/AppData/Local/Temp/tmpjwhzno7b/BaseModel.usda', 'Users/sabrih/AppData/Local/Temp/tmpjwhzno7b/card.png']
```

build.py
HamedSabri-adsk: @seando-adsk  Correct, just a simple indication in the output console:

image

CMakeLists.txt
HamedSabri-adsk: Agree, I had to take it out eventually since I was getting bunch of weird issues with Python 3 when doing incremental builds. Let's revisit this together.

build.py
mattyjams: I think @seando-adsk's point was that that's indicating that `build.py` itself is running using Python 3, and **not** necessarily indicating that you're building for Python 3. You could run `build.py` using Python 3 but still be building maya-usd for Python 2, in which case that indicator might be misleading.

HamedSabri-adsk: Yes I already realized that :) I agree with you that this could be misleading and in fact in my mind was thinking to change it. something for another day.

CMakeLists.txt
seando-adsk: I realized that this needs to remain here. From the Maya build and our internal ecg-maya-usd repo we are setting this flag "BUILD_USDMAYA_SCHEMAS=ON" and providing the jinja/markupsafe locations. The jinja.cmake will check to see if jinja2 and markupsafe are part of the python executable. If not, then it will add them to the python path from the locations provided.

HamedSabri-adsk: Hey @seando-adsk see 389bc58645baebaf3fc54211d76430428ad4f713

cmake/modules/FindMaya.cmake
```diff
@@ -265,9 +251,42 @@ if(MAYA_INCLUDE_DIRS AND EXISTS "${MAYA_INCLUDE_DIR}/maya/MTypes.h")
     else()
         string(SUBSTRING ${MAYA_API_VERSION} "0" "4" MAYA_APP_VERSION)
     endif()
+endif()
 
+# swtich between mayapy and mayapy2
```
seando-adsk, line 256: Minor typo "switch"

HamedSabri-adsk:
> I was able to validate that failing tests are passing once Python's` future` package is installed. What's left to do before we can merge this PR is to document this new requirement for running tests in https://github.com/Autodesk/maya-usd/blob/dev/doc/build.md

Thank you @kxl-adsk for the steps. Please see b7aad657324ad3533e6ff686950ac783113f8209

PR #558, opened 2020-06-03
MAYA-104542 Handle rprim purpose changing to a hidden purpose correctly

williamkrick:


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -390,6 +390,17 @@ void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
                 break;
             }
         }
+
+        // Set the render tags to be all the tags. Vp2RenderDelegate implements
+        // render tags as a per-render item setting, so we always need to visit
+        // all the rprims to handle cases when an rprim changes from a displayed
+        // tag to a hidden tag.
```
williamkrick, line 397: The performance downside of this change is for animated rprims which have a purpose which is not displayed. In that case we'll do all the updates for those rprims and then not draw the item, which is unfortunate.

We might be able to work around it by having a special case for visibility, where we update the visibility of an item first. Then, for hidden items don't bother updating anything else, just leave it all dirty.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -826,26 +838,6 @@ void ProxyRenderDelegate::_UpdateRenderTags()
         for (auto& id : rprimsToDirty) {
             changeTracker.MarkRprimDirty(id, HdChangeTracker::DirtyRenderTag);
         }
-
-        // Make a special pass over the removed render tags so that the objects
-        // can hide themselves.
```
williamkrick, line 831: No need for the special pass because we visit all the dirty rprims already.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -826,26 +838,6 @@ void ProxyRenderDelegate::_UpdateRenderTags()
...
-        if (_proxyShapeData->DrawRenderPurpose()) {
-            renderTags.push_back(HdRenderTagTokens->render);
-        }
-        if (_proxyShapeData->DrawProxyPurpose()) {
-            renderTags.push_back(HdRenderTagTokens->proxy);
-        }
-        if (_proxyShapeData->DrawGuidePurpose()) {
-            renderTags.push_back(HdRenderTagTokens->guide);
-        }
-        _taskController->SetRenderTags(renderTags);
```
williamkrick, line 848: Don't set the render tags to the current set of drawn tags. We leave them as all render tags, because we have to catch those cases where an rprim goes from a visible tag to a hidden tag.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1332,16 +1332,19 @@ void HdVP2Mesh::_UpdateDrawItem(
...
-            = drawItem->GetVisible() && drawScene.DrawRenderTag(renderIndex.GetRenderTag(GetId()));
+        drawItemData._enabled = drawItemVisible();
         stateToCommit._enabled = &drawItemData._enabled;
     }
 
     if (isDedicatedSelectionHighlightItem) {
         if (itemDirtyBits & DirtySelectionHighlight) {
             const bool enable =
-                (_selectionState != kUnselected) && drawItem->GetVisible();
+                (_selectionState != kUnselected) && drawItemVisible();
```
HdC-adsk, line 1347: Interestingly my change for the UsdSkel geometry corruption issue is also fixing this part.

HdC-adsk: So my fix will also include a check on the points array. If it is empty, the render item will be disabled. I will push my branch soon so that we can coordinate on the conflict.

williamkrick:
MayaUsd - Build External GitHub Branch run 142

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
kxl-adsk: With this change, it sounds like purpose will only allow us to save on drawing, but not on the cost to sync changes? What's the implication for memory usage of big static scenes, time to the first frame, etc?

cfmoore007:
Here are my results ( a scene with some instanced trees ) :

PR558_Results

There is a clear improvement in shaded guide display before/after.

Christopher

williamkrick:
> There is a clear improvement in shaded guide display before/after.
> 
> Christopher

@cfmoore007 Your performance results surprised me, I wasn't expecting anything to go faster.

I'm haven't been able to reproduce your slow result with 38c3ea in `Guide Purpose` mode. My numbers before and after the change are similar to your 38c3ea+PR558 numbers. Am I correct in assuming that when you did this test only `Guide Puropose` and not `Proxy Purpose`? Anything else to reproduce this beyond create the proxy shape and set the file name?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
williamkrick: I think you are right we will create all the guide objects and then not draw them. Basically we have to poll every rprim every frame to see if it goes its visibility dirtied because it's purpose changed.

I checked again how HdSt handles the render tag on an rprim changing to a hidden tag:

- HdChangeTracker::MarkRprimDirty() is called with DirtyRenderTag as one of the dirty bits, which increments rprimIndexVersion.

- In HdSt_RenderPass::_PrepareDrawItems() they detect the change in rprimIndexVersion and rebuild the list of draw items to draw (filtered by render tags).

This is what we are missing. We don't have an equivalent search over all the draw items to only update their "visibility". Items get hidden by being left out of the _drawItems list, and therefore excluded from their draw batches in _PrepareCommandBuffers.
I'm trying use our existing pass over all the rprims in execute to turn items on and off as necessary, to do that I explicitly need to consider each MRenderItem and turn on or off its visibility.

I can improve it with a combination of two changes:

- Have an early exit in Sync for rprims that are hidden by render tags, leaving the dirty bits intact so they update correctly when made visible. That prevents creating Maya stuff for guide items which are not drawn.

- Only force the pass over all render tags when that HdChangeTracker::rprimIndexVersion has changed. That prevents visiting everything and checking its visibility every refresh (I'm assuming we'd have lots of dirty + hidden stuff because we skip the update for hidden things). We'd only visit the stuff that passes the render tags test.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -365,6 +365,20 @@ void HdVP2Mesh::Sync(
...
+    // existing render items because they should not be drawn.
+    auto* const          param = static_cast<HdVP2RenderParam*>(_delegate->GetRenderParam());
+    ProxyRenderDelegate& drawScene = param->GetDrawScene();
+    if (!drawScene.DrawRenderTag(delegate->GetRenderIndex().GetRenderTag(GetId()))) {
+        _HideAllDrawItems(reprToken);
+        // clearing visibility here is wrong. It is a work around to handle purpose changing
+        // marking visibility dirty instead of render tag. See UsdImagingGprimAdapter::ProcessPropertyChange
+        // for where this happens.
+        *dirtyBits &= ~( HdChangeTracker::DirtyRenderTag | HdChangeTracker::DirtyVisibility );
+        return;
```
williamkrick, line 379: This is the risky bit of the change. We don't have much experience or tests leaving items dirty & hidden. I tested a scene with animation where I change the purpose of the items so they are hidden, change time, and change the purpose so they are drawn, and that simple workflow is correct.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1520,6 +1534,35 @@ void HdVP2Mesh::_UpdateDrawItem(
...
+
+        auto* drawItem = static_cast<HdVP2DrawItem*>(curRepr->GetDrawItem(drawItemIndex++));
+        if (!drawItem)
+            continue;
+        MHWRender::MRenderItem* renderItem = drawItem->GetRenderItem();
+        if (!renderItem)
+            continue;
+
+        _delegate->GetVP2ResourceRegistry().EnqueueCommit([renderItem]() {
+            renderItem->enable(false);
```
williamkrick, line 1561: Setting enable can change which world the renderItem in, we need those changes to be serialized so use the commit system to do it from the main thread.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
williamkrick: This pattern will have to be repeated in the other HdVP2 shapes because they all have the same problem. Ideally we can find a way for the different shapes to share code.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -803,37 +805,28 @@ void ProxyRenderDelegate::_UpdateRenderTags()
...
-        // Make a special pass over the removed render tags so that the objects
-        // can hide themselves.
-        _taskController->SetRenderTags(removedRenderTags);
-        _engine.Execute(_renderIndex.get(), &_dummyTasks);
-
-        // Set the new render tags correctly.The regular execute will cause
-        // any newly visible rprims will update and mark themselves visible.
+    // When the render tag on an rprim changes we do a pass over all rprims to update
+    // their visibility. The frame after we do the pass over all the tags, set the tags back to
+    // the minimum set of tags.
```
williamkrick, line 828: In the new version we call Execute once with all the tags (to hide those visible->hidden rprims) and the next update we go back to calling Execute with the minimum set of render tags.

cfmoore007:
> > There is a clear improvement in shaded guide display before/after.
> > Christopher
> 
> @cfmoore007 Your performance results surprised me, I wasn't expecting anything to go faster.
> 
> I'm haven't been able to reproduce your slow result with 38c3ea in `Guide Purpose` mode. My numbers before and after the change are similar to your 38c3ea+PR558 numbers. Am I correct in assuming that when you did this test only `Guide Puropose` and not `Proxy Purpose`? Anything else to reproduce this beyond create the proxy shape and set the file name?

- I've updated my previous post.  Not sure what happened, but I can confirm the roughly match
PR558_Results2

I did my tests in both guide and proxy.  And, yes all I did was create the proxyShape and hit 'F' to frame the viewport.
Viewport size was the same in all tests.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
kxl-adsk: Soon we will need ThreadRun to free main-thread and get processing of commits started as soon as possible.

lib/mayaUsd/render/vp2RenderDelegate/mesh.h
```diff
@@ -102,6 +102,8 @@ class HdVP2Mesh final : public HdMesh {
         const HdMeshReprDesc& desc,
         bool requireSmoothNormals, bool requireFlatNormals);
 
+    void _HideAllDrawItems(const TfToken& reprToken);
```
kxl-adsk, line 105: We haven't adopted `_` as our naming convention for methods. No need to change it now, but we should stop using it.

PR #559, opened 2020-06-03
MAYA-104981 Disable all degenerated render items.

HdC-adsk:
Currently due to missing support for UsdSkel, the mesh will have no points data, but it still has the topology data which will reference the points. This could cause corrupted geometry in the viewport.

The commit will disable all degenerated render items.

Along with the fix I clean up the code that determines the enable/disable state of the render item.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -628,13 +628,12 @@ void HdVP2BasisCurves::Sync(
         _sharedData.bounds.SetMatrix(delegate->GetTransform(id));
     }
 
-    if (HdChangeTracker::IsVisibilityDirty(*dirtyBits, id) || 
-        *dirtyBits & HdChangeTracker::DirtyRenderTag) {
+    if (HdChangeTracker::IsVisibilityDirty(*dirtyBits, id)) {
+        _sharedData.visible = delegate->GetVisible(id);
+    }
 
-        auto* const          param = static_cast<HdVP2RenderParam*>(_delegate->GetRenderParam());
-        ProxyRenderDelegate& drawScene = param->GetDrawScene();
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -1193,28 +1192,26 @@ HdVP2BasisCurves::_UpdateDrawItem(
...
-            }
+        else if (drawMode == MHWRender::MGeometry::kBoundingBox) {
+            enable = enable && !range.IsEmpty();
+        }
+
+        enable = enable && drawScene.DrawRenderTag(_curvesSharedData._renderTag);
+
+        if (drawItemData._enabled != enable) {
+            drawItemData._enabled = enable;
+            stateToCommit._enabled = &drawItemData._enabled;
```
HdC-adsk, line 1214: Streamline the code about the enable/disable state.

HdC-adsk: @williamkrick The change here will conflict with #558. I think this change will be clearer and streamline all the relevant code about enable/disable state. It also avoids the query of the render tag when there is points change (e.g. deformation). What do you think?

HdC-adsk: Maybe this change also fixes a potential issue that if you switch the purpose of the prim and then switch the viewport mode the render item state won't be updated?

williamkrick: Sure lets do it your way. I'll take that part of the changes out of #558. We still need _updateRenderTags because we still need to mark rprims that are becoming hidden or visible dirty. Without marking them dirty we don't call Sync and so can't update.

I'll update #558 to be what I think is correct and we can discuss more.

PR #560, opened 2020-06-03
add Trace profiling to pxr interactive performance tests

mattyjams:
These tests are still not being run through CMake/CTest, but we've been running them internally in our SCons-based build. This change makes them drop performance metrics generated using the Trace library to better determine where the time is being spent.

### DIFF
--- a/plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeDrawPerformance.py
+++ b/plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeDrawPerformance.py
@@ -18,6 +18,7 @@
 from pxr import UsdMaya
 
 from pxr import Tf
+from pxr import Trace
 
 from maya import cmds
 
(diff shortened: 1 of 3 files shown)


plugin/pxr/maya/lib/pxrUsdMayaGL/testenv/testProxyShapeDrawPerformance.py
```diff
@@ -75,15 +76,26 @@ def _ProfileScope(self, profileScopeName):
         exit and stores the elapsed time in the class' metrics dictionary.
         """
         stopwatch = Tf.Stopwatch()
+        collector = Trace.Collector()
```
kxl-adsk, line 79: What's the difference between stopwatch and trace collector?

mattyjams: `stopwatch` is pretty much just like using Python's `time()` to measure wall clock time by recording a start and end time and computing the elapsed time from them.

Enabling the `Trace` stuff is actually causing all of the `TRACE_FUNCTION`, `TRACE_SCOPE`, etc. declarations in USD itself to measure how much time is spent inside those things, more like `MProfilerScope`.

Internally, we're just verifying that the wall clock time from the Stopwatch is under an expected threshold, but looking at the traces can help identify issues if there's a failure.

kxl-adsk: I'm guessing there is a cost in using `Trace` - should it be implicitly enabled in the test or developer should opt-in when he needs to investigate it?

mattyjams: There is a small bit of overhead to adding traces, which indeed gets bigger when tracing is enabled. It's discussed a bit at the bottom of the API docs page here:
https://graphics.pixar.com/usd/docs/api/trace_page_front.html

For these tests, we opted to just enable tracing all the time. They may sometimes run on heterogeneous hardware, so it's useful to be able to just grab the trace files out of the test run artifacts. We haven't seen the overhead be significant enough to meaningfully impact the overall wall clock time.

PR #564, opened 2020-06-04
A couple of minor fixes for python3 support.

seando-adsk:


cmake/test.cmake
```diff
@@ -191,6 +191,12 @@ finally:
     list(APPEND mayaUsd_varname_PXR_PLUGINPATH_NAME
          "${CMAKE_INSTALL_PREFIX}/plugin/al/plugin")
 
+    if(IS_WINDOWS AND DEFINED ENV{PYTHONHOME})
+        # If the environment contains a PYTHONHOME, also set the path to
+        # that folder so that we can find the python DLLs.
+        list(APPEND mayaUsd_varname_PATH $ENV{PYTHONHOME})
+    endif()
```
seando-adsk, line 198: @HamedSabri-adsk As we discussed, when the PYTHONHOME env var is set (which we'll do from our internal builds) we also add that to the path. This is so the tests can find the python37.dll

cmake/utils.cmake
```diff
@@ -68,7 +68,7 @@ function(mayaUsd_find_python_module module)
             set(${module}_FIND_REQUIRED TRUE)
         endif()
         execute_process(COMMAND "${Python_EXECUTABLE}" "-c"
-            "import re, ${module}; print re.compile('/__init__.py.*').sub('',${module}.__file__)"
+            "from __future__ import print_function; import re, ${module}; print(re.compile('/__init__.py.*').sub('',${module}.__file__))"
```
seando-adsk, line 71: The python code is used to determine if a given module (such as jinja2) exists. It needed to be made python3 compatible.

PR #565, opened 2020-06-04
MAYA-105122 trigger only on assignment to ecp-maya-devops-adsk

smithw-adsk:
The diff looks bad. The real change is only 2 lines:

` runs-on: ubuntu-latest`
`if: contains(toJson(github.event.pull_request.assignee), 'ecp-maya-devops-adsk')`

I took out the runs-on matrix for simplicity.
The if: condition makes the GitHub Actions build show as "skipped" instead of "failed", which is nice. You can see an example here: https://github.com/smithw-adsk/maya-usd/pull/16


kxl-adsk:
Awesome. Thank you @smithw-adsk !

Issue #522, opened 2020-05-25
Visibility not respected for instanced prims 


eddiehoyle:
**Describe the bug**
Changing instanced prims visibility state to `"invisible"` isn't respected, draws prim at origin instead.

**Steps to reproduce**
1. Import to Maya a layer that contains a prim with `instanceable = True` and which references some imageable data.
2. Via the `UsdGeom.Imageable` API, set the visibility attribute on the prim to `"inherited"` or `"invisible"`
3. Observe the prim being redrawn at the origin when `instanced = True` and visibility is set from `"inherited"` to `"invisible"`
4. If the prim is not instanced, the visibility works as expected

Here's a snippet:
```python
from maya import cmds
from pxr import Sdf, UsdGeom

layer_ref = Sdf.Layer.CreateAnonymous()
layer_ref.ImportFromString("""#sdf 1.4.2
(
    defaultPrim = "root"
)

def Xform "root"
{
    def Mesh "plane_M_geo"
    {
        float3[] extent = [(-5, 0, -5), (5, 0, 5)]
        int[] faceVertexCounts = [4]
        int[] faceVertexIndices = [0, 1, 3, 2]
        normal3f[] normals = [(0, 1, 0)]
        point3f[] points = [(-5, 0, 5), (5, 0, 5), (-5, 0, -5), (5, 0, -5)]
    }
}""")
layer_ref.Export("/tmp/ref.usda")


layer_src = Sdf.Layer.CreateAnonymous()
layer_src.ImportFromString("""#sdf 1.4.2
(
    defaultPrim = "root"
)

def Xform "root" (
    instanceable = true
    kind = "component"
    prepend references = @/tmp/ref.usda@

)
{
    double3 xformOp:translate = (-10, 0, 0)
    uniform token[] xformOpOrder = ["xformOp:translate"]
}""")
layer_src.Export("/tmp/src.usda")

# Pseudocode for importing "/tmp/src.usda" via Proxy Shape 
# cmds.loadPlugin("...") 
# cmds.usdMaya_CreateUsdProxyShape(...)
# proxy = usdMaya.getProxy(...)
stage = proxy.getUsdStage()
prim = stage.GetPrimAtPath("/root")
vis = UsdGeom.Imageable(prim).GetVisibilityAttr()
```

Run these next blocks separately and repeatedly:

```python
# Works
prim.SetInstanceable(False)
vis.Set("inherited" if vis.Get() == "invisible" else "invisible")

# Weird
[...]

Issue #404, opened 2020-04-03
[MAYA-105120] PointInstancer prototypes load as visible


fdan:
**Describe the bug**
When I load a set with a point instancer, the instances show correctly in the viewport, but the prototypes are also visible.  They should not be. 

**Steps to reproduce**
1. Download and unzip my attache set
2. Bring into Maya via Create -> Universal Scene Description (USD) -> Stage From Existing Layer and navigate to `assets/set/world/model/model/assembly/world_model_model_assembly.v046.usd`
3. In the viewport, you will see a single giant tree which is one of the instancer prototypes that should be hidden

**Expected behavior**
Point instancer instances should be visible, but prototypes should not be.  

**Attachments**
In this image, you can see the prototypes are visible
![image](https://user-images.githubusercontent.com/1701261/78403560-45dc2680-7648-11ea-8f90-78166d7464e9.png)

Whereas in usdview they are not - I take this to be correct?
![image](https://user-images.githubusercontent.com/1701261/78403611-60160480-7648-11ea-8dc0-d53cd1a21550.png)

[world_set.zip](https://github.com/Autodesk/maya-usd/files/4429854/world_set.zip)

**Specs (if applicable):**
 - OSX 10.14
 - Apple LLVM version 10.0.1 (clang-1001.0.46.4)- (I used the prebuild binary in release 0.1.0)
 - Maya 2020
 - Maya USD commit SHA 00d651a8af3af752160531a0841579d1bc0f41ea

**Additional context**

santosd:
Hello @fdan, thank you for posting this issue here. I took a look at it in PR114 and was able to reproduce the issue in the latest build of the Maya USD. I will set this issue as reproduced and will get one of our devs to take a look at it.

Thank you.

williamkrick:
@fdan #570 which fixes this issue has been merged. Can you test it to verify?

santosd:
@fdan I tested it out on my end with the new changes and this seems to be working well after the merged changes. Please give it a try and let me know or feel free to close this issue out.

fdan:
Cheers, I haven’t had a chance to confirm unfortunately but trust you folks that it’s solved.

PR #570, opened 2020-06-08
MAYA-105120 Hide MRenderItems created for instancers with zero instances.

williamkrick:
As a part of setting up an instanced MRenderItem in Vp2RenderDelegate we'd set the prototype geometry on the item. At this point the MRenderItem looks to Maya like any other drawable item (with an identity transform matrix). Then we set the instance transforms and the MRenderItem changes to an instanced item and we handle the drawing differently.

The problem is when there is a USD instances with no instances we never do that second step, so the item stays a regular Maya item and we draw it normally, which is not what we want to happen.

The solution is detect that there is an instancer with zero instance transforms and hide the MRenderItem in those cases.

#404 #522

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -1075,6 +1075,9 @@ HdVP2BasisCurves::_UpdateDrawItem(
     // The current instancer invalidation tracking makes it hard for
     // us to tell whether transforms will be dirty, so this code
     // pulls them every time something changes.
+    // If the mesh is instanced but has 0 instance transforms remember that
+    // so the render item can be hidden.
+    bool instancerWithNoInstances = false;
     if (!GetInstancerId().IsEmpty()) {
 
         // Retrieve instance transforms from the instancer.
@@ -1083,10 +1086,14 @@ HdVP2BasisCurves::_UpdateDrawItem(
             ComputeInstanceTransforms(id);
 
         MMatrix instanceMatrix;
+        const unsigned int instanceCount = transforms.size();
 
-        if (isDedicatedSelectionHighlightItem) {
+        if (0 == instanceCount) {
+            instancerWithNoInstances = true;
+        }
+        else if (isDedicatedSelectionHighlightItem) {
             if (_selectionState == kFullySelected) {
-                stateToCommit._instanceTransforms.setLength(transforms.size());
+                stateToCommit._instanceTransforms.setLength(instanceCount);
                 for (size_t i = 0; i < transforms.size(); ++i) {
                     transforms[i].Get(instanceMatrix.matrix);
                     instanceMatrix = worldMatrix * instanceMatrix;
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1356,8 +1362,9 @@ void HdVP2Mesh::_UpdateDrawItem(
                          HdChangeTracker::DirtyRenderTag |
                          HdChangeTracker::DirtyPoints |
                          HdChangeTracker::DirtyExtent |
-                         DirtySelectionHighlight)) {
-        bool enable = drawItem->GetVisible() && !_meshSharedData._points.empty();
+                         DirtySelectionHighlight) ||
+        instancerWithNoInstances) {
```
williamkrick, line 1366: In the instancer section it ignores dirty flags, so unfortunately we have to re-hide the instancer with no instances every frame, but changing from hidden to hidden should have no cost.

HdC-adsk: But if there is nothing else to trigger the sync, we won't run into this code at all. So I would not worry about the cost too much.

HdC-adsk: Maybe protect the bitwise AND with parentheses.

HdC-adsk:
Please add the same fix to basisCurves.

williamkrick:
> 
> 
> basisCurves will need the same fix.

We need a way to share code between basisCurves & mesh. There is a lot of duplicate stuff between the two of them.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -1193,12 +1199,13 @@ HdVP2BasisCurves::_UpdateDrawItem(
...
     // Determine if the render item should be enabled or not.
-    if (itemDirtyBits & (HdChangeTracker::DirtyVisibility |
+    if ((itemDirtyBits & (HdChangeTracker::DirtyVisibility |
                          HdChangeTracker::DirtyRenderTag |
                          HdChangeTracker::DirtyPoints |
                          HdChangeTracker::DirtyExtent |
-                         DirtySelectionHighlight)) {
-        bool enable = drawItem->GetVisible() && !_curvesSharedData._points.empty();
+                         DirtySelectionHighlight)) ||
+                         !GetInstancerId().IsEmpty()) {
```
williamkrick, line 1207: The old expression checking instancerWithNoInstances wouldn't re-show the render item when going from 0 instances to 1 or more instances.

HdC-adsk:
> 
> 
> > basisCurves will need the same fix.
> 
> We need a way to share code between basisCurves & mesh. There is a lot of duplicate stuff between the two of them.

Extractly. I think this can be done when we start to implement the third Rprim class.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
HdC-adsk: Do you find a way to interactively add instances?

williamkrick: I think the python API may be able to do it, but I didn't actually get it to work. I just realized it was a problem when updating the code.

PR #573, opened 2020-06-09
Support Python 3 Compatibility for AL plugin

HamedSabri-adsk:


plugin/al/usdtransaction/AL/usd/transaction/tests/testTransaction.py
```diff
@@ -13,6 +14,9 @@ def setUp(self):
         self._closed = 0
         self._changed = []
         self._resynced = []
+        # assertCountEqual in python 3 is equivalent to assertItemsEqual
+        if sys.version_info[0] >= 3:
```
seando-adsk, line 18: FYI, can also use sys.version_info.major

PR #575, opened 2020-06-09
MAYA-104936 instanceIndices may contain duplicate indices.

williamkrick:
Only add each instance transform at most once.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
@@ -16,7 +16,6 @@
 #include "mesh.h"
 
 #include <numeric>
-#include <unordered_set>
 
 #include <maya/MMatrix.h>
 #include <maya/MProfiler.h>
--- a/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
@@ -16,6 +16,7 @@
 #include "mesh.h"
 
 #include <numeric>
+#include <unordered_set>
 
 #include <maya/MMatrix.h>
 #include <maya/MProfiler.h>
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1286,13 +1287,18 @@ void HdVP2Mesh::_UpdateDrawItem(
                 }
             }
             else if (auto state = drawScene.GetPrimSelectionState(id)) {
-                for (const auto& indexArray : state->instanceIndices) {
+                std::unordered_set<int> selectedIndices;
```
williamkrick, line 1290: Each index can only be in the unordered_set once, so this approach ensures we won't add the same instance transform to the instance batch multiple times.

I also tried to get an answer to the question, "Should GetPrimSelectionState be returning duplicate instance indices?" but I didn't find a good example of it in use, and tracing the code that builds up the selection was taking a long time, so I decided to post up publicly and see if we can figure out who I can ask about what GetPrimSelectionState is supposed to be doing.

williamkrick: There also might be a way to have the selection state we get back from USD to tell us that the object is fully selected. We store that kFullySelected status in _selectionState, and test it at line 1281. I'm not sure why we don't get fully selected here, by selected the parent USDXform it seems like the instancer should be fully selected?

kxl-adsk: It's a shame we have to do this and I'm looking forward to reading the response to your question. 

If we have to stay with this pattern for longer, I wonder if std::vector<bool> (it's a specialization of std::vector) wouldn't be more efficient (both memory-wise and access time).

williamkrick:
@cfmoore007 This is some defensive code that "fixes" the issue you reported with drawing instanced selection. I'm not sure this is the code we'll check in, but this change does give what I believe will be the final performance, if you are interested in testing.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
williamkrick: For std::vector I guess it depends on how sparse the selection is. If things are mostly fully selected the vector is good, if things are mostly not selected the map is better for memory.

@HdC-adsk gave me a lead to follow so I'll see if we're creating the selection from UFE wrong, or if there is a USD bug I can log.

williamkrick: Initial testing shows we call UsdImagingDelegate::PopulateSelection() once in a reasonable looking way, with UsdImagingDelegate::ALL_INSTNACES set. And that returns an HdSelection with the N^2 instanceIndices in it, so I'm thinking I need to log an issue with USD.

williamkrick: I guess things are mostly all selected, so I could use std::vector<bool> and pre-allocate it because I know the max size. That should be fast & small.

PR #583, opened 2020-06-10
[mtoh] Push WANT_UFE_BUILD definition into target_compile_definitions.

marsupial:
-DWANT_UFE_BUILD isn't making it into the _mtoh_ build process.


marsupial:
Could you please be specific about what you would like changed, I moved this **WANT_UFE_BUILD** to **PUBLIC** because you asked to.

Again, *WANT_UFE_BUILD** is used in headers so anything that builds/links against mayaUsd/mtoh/hdMaya afterward needs to know what this is set to.

Issue #589 is because _Autodesk_ is not dealing with this properly ... on your end... if/when you fix that all this back and forth and time becomes irrelevant.

PR #585, opened 2020-06-10
fix up header includes using "./" and "../" and other misc. clean-up

mattyjams:
This fixes up the remaining instances where headers were being included with a path using `./` or `../`. While I was in there, I also addressed a few other coding guidelines issues for the header includes in those files.

### DIFF
--- a/lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.h
+++ b/lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.h
@@ -15,8 +15,8 @@
 //
 #pragma once
 
-#include "../base/api.h"
-#include "UsdSceneItem.h"
+#include "mayaUsd/base/api.h"
+#include "mayaUsd/ufe/UsdSceneItem.h"
 
 #include <ufe/transform3dUndoableCommands.h>
 #include <ufe/observer.h>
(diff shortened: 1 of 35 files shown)


mattyjams:
> @mattyjams lgtm! I still see in bunch of other places the traces of ../ ./. Can you please address them in this PR?
> 
> e.g
> 
> ```
> https://github.com/mattyjams/maya-usd/blob/pr/fix_relative_include_paths_in_AL_plugins/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/AnimationTranslator.h
> 
> https://github.com/mattyjams/maya-usd/blob/pr/fix_relative_include_paths_in_AL_plugins/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/ImportTranslator.h
> 
> https://github.com/mattyjams/maya-usd/blob/pr/fix_relative_include_paths_in_AL_plugins/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/DgNodeTranslator.h
> ```

Ahh, good catch @HamedSabri-adsk. Sorry, my grep'ing missed those.

That added a bunch more files in `plugin/al` that are addressed in https://github.com/Autodesk/maya-usd/pull/585/commits/99fcff5e2f873c15542949eafb37e3edd22d9112, and two files in `mayaUsd/ufe` addressed in https://github.com/Autodesk/maya-usd/pull/585/commits/bdd436f46fa1da61aa3b4de9bd6fd1bc1b5f88d4.

PR #587, opened 2020-06-11
MAYA-105284 Track changes to render tags more carefully to avoid setting the tags

williamkrick:
every refresh.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -556,8 +556,7 @@ void HdVP2BasisCurves::Sync(
     if (!drawScene.DrawRenderTag(delegate->GetRenderIndex().GetRenderTag(GetId()))) {
         _HideAllDrawItems(reprToken);
         *dirtyBits &= ~( HdChangeTracker::DirtyRenderTag
-#if USD_VERSION_NUM <= 2005
-        // Purpose changing marks visibility dirty instead of render tag.
+#ifdef ENABLE_RENDERTAG_VISIBILITY_WORKAROUND
          | HdChangeTracker::DirtyVisibility
 #endif
           );
@@ -649,8 +648,7 @@ void HdVP2BasisCurves::Sync(
     }
 
     if (*dirtyBits & (HdChangeTracker::DirtyRenderTag
-#if USD_VERSION_NUM <= 2005
-    //purpose changing sets DirtyVisibility instead of DirtyRenderTag
+#ifdef ENABLE_RENDERTAG_VISIBILITY_WORKAROUND
         |HdChangeTracker::DirtyVisibility
 #endif
     )) {
@@ -1695,8 +1693,7 @@ void HdVP2BasisCurves::_HideAllDrawItems(const TfToken& reprToken) {
         if (!renderItem)
             continue;
 
-        HdVP2DrawItem::RenderItemData& drawItemData(drawItem->GetRenderItemData());
-        drawItemData._enabled = false;
+        drawItem->GetRenderItemData()._enabled = false;
 
         _delegate->GetVP2ResourceRegistry().EnqueueCommit([renderItem]() {
             renderItem->enable(false);
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -548,6 +548,22 @@ void HdVP2BasisCurves::Sync(
         return;
     }
 
+    // We don't update the repr if it is hidden by the render tags (purpose)
+    // of the ProxyRenderDelegate. In additional, we need to hide any already
+    // existing render items because they should not be drawn.
+    auto* const          param = static_cast<HdVP2RenderParam*>(_delegate->GetRenderParam());
+    ProxyRenderDelegate& drawScene = param->GetDrawScene();
+    if (!drawScene.DrawRenderTag(delegate->GetRenderIndex().GetRenderTag(GetId()))) {
+        _HideAllDrawItems(reprToken);
+        *dirtyBits &= ~( HdChangeTracker::DirtyRenderTag
+#if USD_VERSION_NUM <= 2005
+        // Purpose changing marks visibility dirty instead of render tag.
+         | HdChangeTracker::DirtyVisibility
+#endif
+          );
+        return;
+    }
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -823,24 +823,11 @@ void ProxyRenderDelegate::_UpdateRenderTags()
...
-            renderTags.push_back(HdRenderTagTokens->guide);
-        }
-        _taskController->SetRenderTags(renderTags);
-        _taskRenderTagsValid = true;
-    }
+    // The renderTagsVersion increments when the render tags on an rprim changes, or when the
+    // global render tags are set. Check to see if the render tags version has changed since
+    // the last time we set the render tags so we know if there is a change to an individual
+    // rprim or not.
+    bool rprimRenderTagChanged = (_renderTagVersion == changeTracker.GetRenderTagVersion());
```
HdC-adsk, line 830: I think this is checking the global render tag instead of the rprim render tag. The var name seems to be confusing.

williamkrick: RenderTagVersion from the change tracker increments when the render tags on the tasks are set (the global render tags) or when HdChangeTracker::DirtyRenderTag gets set on an rprim. It is one value tracking two things.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -548,6 +548,22 @@ void HdVP2BasisCurves::Sync(
         return;
     }
 
+    // We don't update the repr if it is hidden by the render tags (purpose)
```
williamkrick, line 551: The same optimization that is in mesh.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -548,6 +548,22 @@ void HdVP2BasisCurves::Sync(
...
+    // of the ProxyRenderDelegate. In additional, we need to hide any already
+    // existing render items because they should not be drawn.
+    auto* const          param = static_cast<HdVP2RenderParam*>(_delegate->GetRenderParam());
+    ProxyRenderDelegate& drawScene = param->GetDrawScene();
+    if (!drawScene.DrawRenderTag(delegate->GetRenderIndex().GetRenderTag(GetId()))) {
+        _HideAllDrawItems(reprToken);
+        *dirtyBits &= ~( HdChangeTracker::DirtyRenderTag
+#if USD_VERSION_NUM <= 2005
```
williamkrick, line 559: Pixar is in the process of fixing this one, hopefully the fix is in the next version. If node we may have to adjust this check.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1587,6 +1594,9 @@ void HdVP2Mesh::_HideAllDrawItems(const TfToken& reprToken) {
         if (!renderItem)
             continue;
 
+        HdVP2DrawItem::RenderItemData& drawItemData(drawItem->GetRenderItemData());
+        drawItemData._enabled = false;
```
williamkrick, line 1598: This was a real problem. Later when we are calling MRenderItem::enable() we check to see if the new value is the same as the old value. If we don't update drawItemData._enabled then that check thinks the MRenderItem is enabled when it is not, and we don't show the rprim correctly.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -823,24 +823,16 @@ void ProxyRenderDelegate::_UpdateRenderTags()
...
-    }
+    // The renderTagsVersion increments when the render tags on an rprim changes, or when the
+    // global render tags are set. Check to see if the render tags version has changed since
+    // the last time we set the render tags so we know if there is a change to an individual
+    // rprim or not.
+    bool rprimRenderTagChanged = (_renderTagVersion != changeTracker.GetRenderTagVersion());
+#if USD_VERSION_NUM <= 2005
+    // in v20.05 and earlier when the purpose of an rprim changes the visibility gets dirtied,
+    // and that doesn't update the render tag version.
+    rprimRenderTagChanged = rprimRenderTagChanged || ( _visibilityVersion != changeTracker.GetVisibilityChangeCount());
```
williamkrick, line 834: Extra check necessary until purpose changing correctly marks DirtyRenderTag instead of DirtyVisibility.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
kxl-adsk: People may have already adopted a newer version of USD and won't get proper behavior because of above assumption. Let's not assume things will get fixed...but instead document places which will require conditional compilation when it will be available in USD.

One way of doing it is to create a new macro which will say ENABLE_RENDERTAG_VISIBILITY_WORKAROUND and add a comment in there pointing to GitHub issue in USD (if we have one)...and/or usd-interest discussion.

williamkrick: Yea that is a nice way of doing it. I can put the documentation in the code once where I define the macro, and then the rest of the places test the macro and don't have to say much, because the macro explains why that code is there already.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -849,16 +839,36 @@ void ProxyRenderDelegate::_UpdateRenderTags()
...
+            renderTags.push_back(HdRenderTagTokens->proxy);
+        }
+        if (_proxyShapeData->DrawGuidePurpose()) {
+            renderTags.push_back(HdRenderTagTokens->guide);
+        }
+        _taskController->SetRenderTags(renderTags);
+        _taskRenderTagsValid = true;
+    }
+    _renderTagVersion = changeTracker.GetRenderTagVersion();
+#if USD_VERSION_NUM <= 2005
```
kxl-adsk, line 869: This one is not a workaround?

williamkrick: It is, I'm just bad at coding.

PR #590, opened 2020-06-15
MAYA-105123 wait for build to start, then wait for completion and pos…

smithw-adsk:
…t results

### DIFF
--- a/.github/workflows/maya-usd-preflight-launcher.yml
+++ b/.github/workflows/maya-usd-preflight-launcher.yml
@@ -9,9 +9,9 @@ on:
 
 jobs:
 
-  # Wait for remote build to start, report back
-  build_start:
-      timeout-minutes: 180
+  # Wait for remote build to start and finish, report results
+  build_preflight:
+      timeout-minutes: 400
       runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/maya-usd-preflight-launcher.yml
```diff
@@ -9,28 +9,51 @@ on:
...
             path: transfer
 
+        # Echo the file name - it's hard to find the run_id in the UI
+        - name: Echo the expected start file name
+          run: "echo ${{ github.run_id }}_${{ github.run_number }}_start.txt"
+
+        # Wait for remote build to start
+        - name: Wait until remote build starts
+          shell: bash
+          # 1080 = 180 minutes in 10 second increments. There will be overhead for the git pull command, so wait time will be more than 180 minutes
```
kxl-adsk, line 34: From the end-user perspective, I don't really care if this sleep will be 10 seconds or 1 minute. How about we reduce the frequency of these calls?

BTW. Typically we count the time that is elapsed, not the number of iterations. Is this something hard to achieve here?

smithw-adsk: That's a good idea. 1-minute sleeps is vastly more intuitive. I'll add that now.

We could be more precise with the timing, but I think the rough count is sufficient for the timeout purpose.

PR #591, opened 2020-06-17
MAYA-103187 Add option to save .usd files as binary or ascii.

fowlertADSK:


fowlertADSK:
This adds an option to our Export UI to set the internal format for .usd files to be either ascii or binary, default is binary.

1. Added a new defaultUSDFormat token in jobArgs with only usda and usdc as the options.
2. Menu to choose between the two formats in the export UI.
3. Added a flag for it in the base export command.  It will show up as an option in both ADSK and Pixar export commands but have only added it to the UI for ADSK.

lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -250,6 +250,13 @@ UsdMayaJobExportArgs::UsdMayaJobExportArgs(
                     UsdGeomTokens->bilinear,
                     UsdGeomTokens->none
                 })),
+        defaultUSDFormat(
+            _Token(userArgs,
+                UsdMayaJobExportArgsTokens->defaultUSDFormat,
+                UsdMayaTranslatorTokens->UsdFileExtensionCrate,
+                {
+                    UsdMayaTranslatorTokens->UsdFileExtensionASCII
+                })),
```
mattyjams, line 259: Since we're talking about file formats here, I think it might be slightly more correct to use the format ID tokens provided by USD:

```C++
#include <pxr/usd/usd/usdaFileFormat.h>
#include <pxr/usd/usd/usdcFileFormat.h>
...
                UsdUsdcFileFormatTokens->Id,
                {
                    UsdUsdaFileFormatTokens->Id
                })),
...
```

lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -444,6 +452,8 @@ UsdMayaJobExportArgs::GetDefaultDictionary()
         d[UsdMayaJobExportArgsTokens->defaultCameras] = false;
         d[UsdMayaJobExportArgsTokens->defaultMeshScheme] = 
                 UsdGeomTokens->catmullClark.GetString();
+        d[UsdMayaJobExportArgsTokens->defaultUSDFormat] = 
+                UsdMayaTranslatorTokens->UsdFileExtensionCrate.GetString();
```
mattyjams, line 456: Similar here: ```UsdUsdcFileFormatTokens->Id```

lib/mayaUsd/fileio/jobs/jobArgs.h
```diff
@@ -53,6 +53,7 @@ TF_DECLARE_PUBLIC_TOKENS(
     (chaserArgs) \
     (compatibility) \
     (defaultCameras) \
+    (defaultUSDFormat) \
```
mattyjams, line 56: Move this down one line after `defaultMeshScheme`?

lib/mayaUsd/fileio/writeJobContext.cpp
```diff
@@ -388,7 +388,9 @@ UsdMayaWriteJobContext::_OpenFile(const std::string& filename, bool append)
             layer = existingLayer;
         }
         else {
-            layer = SdfLayer::CreateNew(filename);
+            SdfLayer::FileFormatArguments args;
+            args["format"] = mArgs.defaultUSDFormat.GetString();
```
mattyjams, line 392: Could use another USD provided token here:

```C++
#include <pxr/usd/usd/usdFileFormat.h>
...
args[UsdUsdFileFormatTokens->FormatArg] = mArgs.defaultUSDFormat.GetString();
...
```

fowlertADSK:
Thx for the pointers @mattyjams   FYI I am going to write some kind of automated test for this, but will sync up with Pierre on https://github.com/Autodesk/maya-usd/pull/595 first, so that PR will probably go in before this one.  Not in a big rush for this one, just wanted to put it up last week in case we wanted it in the next Maya beta.

kxl-adsk:
Closing...test was added and new PR opened. See #630

PR #595, opened 2020-06-18
Make import / export tests run from core.

ppt-adsk:
This pull request moves the majority of import / export tests down to the maya-usd core, and has them run using the mayaUsdPlugin.  Only those tests that have a dependence on the Pixar plugin (for Maya scene assembly support) were kept in the Pixar plugin subtree.


plugin/pxr/maya/lib/usdMaya/testenv/UsdExportPointInstancerTest/InstancerTest.ma
```diff
@@ -0,0 +1,262 @@
+//Maya ASCII 2016 scene
```
ppt-adsk, line 1: This file was actually removed by
https://github.com/Autodesk/maya-usd/commit/7cd3c48e173434ec9cdb074f7a2adfade4fd0258
Will fix.

plugin/pxr/maya/lib/usdMaya/testenv/testUsdExportPointInstancer.py
```diff
@@ -52,9 +52,17 @@ def tearDownClass(cls):
     def setUpClass(cls):
         standalone.initialize('usd')
 
-        cmds.loadPlugin("MASH")
-
-        scene = "InstancerTestMash.ma"
+        # Choose the test file based on whether the MASH plugin is available.
```
ppt-adsk, line 55: Again, incorrect revert of
https://github.com/Autodesk/maya-usd/commit/7cd3c48e173434ec9cdb074f7a2adfade4fd0258
will fix.

test/lib/usd/importExport/testUsdExportStripNamespaces.py
```diff
@@ -49,8 +48,22 @@ def testExportWithClashStripping(self):
 
         usdFilePath = os.path.abspath('UsdExportStripNamespaces_EXPORTED.usda')
 
-        errorRegexp = "Multiple dag nodes map to the same prim path" \
-            ".+|cube1 - |foo:cube1.*"
+        # https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/fileio/jobs/writeJob.cpp#L814
```
ppt-adsk, line 51: Help with this would be appreciated.

ppt-adsk:
Many, many thanks to @HamedSabri-adsk for his help on this branch!

test/lib/usd/importExport/fixturesUtils.py
```diff
@@ -0,0 +1,74 @@
...
+    '''Test class import setup for tests that do not write to the file system.
+
+    This function:
+    - Initializes Maya standalone use.
+    - Loads the plugin
+    - Returns the original directory from the argument.
+    '''
+    (testDir, testFile) = _setUpClass(modulePathName)
+
+    return testDir
```
HamedSabri-adsk, line 74: @ppt-adsk This is a great utility. I am wondering if it makes sense to have fixturesUtils.py move out of here to a higher directory ( e.g test\lib\usd ).  In case we bring in more pixar tests here, they can also leverage this.

test/lib/usd/importExport/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+    EXCLUDE "*.txt"
+)
+
+foreach(script ${test_script_files})
+    mayaUsd_get_unittest_target(target ${script})
+    mayaUsd_add_test(${target}
+        PYTHON_MODULE ${target}
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+        ENV
+            "MAYA_DISABLE_CIP=1"
```
HamedSabri-adsk, line 68: I believe we can set below two Maya environment variables in `test.cmake`, so don't have to worry about setting them every time.

```
MAYA_DISABLE_CIP=1  Avoid fatal crash on start-up.
MAYA_DISABLE_CER=1  Customer Error Reporting.
```
These can be added to the list of ENV variables here:
```
    # without "MAYA_NO_STANDALONE_ATEXIT=1", standalone.uninitialize() will
    # set exitcode to 0
    set_property(TEST "${test_name}" APPEND PROPERTY ENVIRONMENT
        "MAYA_NO_STANDALONE_ATEXIT=1"
        "MAYA_DISABLE_CIP=1"
        "MAYA_DISABLE_CER=1")
```

test/lib/usd/importExport/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+    testUsdExportDisplayColor.py
+    testUsdExportEulerFilter.py
+    testUsdExportFilterTypes.py
+    testUsdExportFrameOffset.py
+    testUsdExportInstances.py
+    testUsdExportLocator.py
+    testUsdExportMesh.py
+    testUsdExportNurbsCurve.py
+    testUsdExportOpenLayer.py
+    testUsdExportOverImport.py
```
HamedSabri-adsk, line 17: Hey @ppt-adsk , 

I did a clean build of maya-usd with your changes on Windows.  Ran all the tests and ctest reported that all tests pass:

```
100% tests passed, 0 tests failed out of 104

Label Time Summary:
importExport         = 667.42 sec*proc (39 tests)
ufe                  = 390.75 sec*proc (19 tests)
usdPreviewSurface    =  33.33 sec*proc (2 tests)

Total Test time (real) =  85.62 sec
```

I then ran the tests again but ctest reports that `testUsdExportOverImport ` failed. Do you see this behavior on your side as well?

Here is what -VV shows:
```
40: ======================================================================
40: FAIL: testImportModifyAndExportCubeModel (testUsdExportOverImport.testUsdExportOverImport)
40: ----------------------------------------------------------------------
40: Traceback (most recent call last):
40:   File "testUsdExportOverImport.py", line 73, in testImportModifyAndExportCubeModel
40:   File "testUsdExportOverImport.py", line 50, in _ValidateUsdBeforeExport
40: AssertionError: invalid null prim is not true
40:
40: ----------------------------------------------------------------------
40: Ran 1 test in 0.323s
40:
40: FAILED (failures=1)
1/1 Test #40: testUsdExportOverImport ..........***Failed    6.35 sec
```

HamedSabri-adsk:
@ppt-adsk Great job. You also did a lot of the work here as well. 

Aside from the testUsdExportOverImport failure and couple of minor comments, everything else looks good to me. 

Merci

mattyjams:
Thanks for tackling this @ppt-adsk!

I unfortunately probably won't get a chance to dig into this in detail until early next week, but I'm excited to check it out!

ppt-adsk:
@mattyjams no worries, thanks for having a look.

ppt-adsk:
importExport directory renamed translators at @kxl-adsk 's request, to match library hierarchy and help discoverability.

test/lib/usd/translators/fixturesUtils.py
```diff
@@ -0,0 +1,74 @@
...
+    (testDir, testFile) = _setUpClass(modulePathName)
+    outputName = os.path.splitext(testFile)[0]+suffix+'Output'
+
+    outputPath = os.path.join(testDir, outputName)
+    if os.path.exists(outputPath):
+        # Remove previous test run output.
+        shutil.rmtree(outputPath)
+
+    os.mkdir(outputPath)
+    os.chdir(outputPath)
```
mattyjams, line 60: Curious about this, since by the time we hit this we've obviously already launched Maya from some other directory. Would Maya itself possibly still write `workspace.mel` or swatch files or other stuff into the run directory, and do we care if so?

test/lib/usd/translators/fixturesUtils.py
```diff
@@ -0,0 +1,74 @@
...
+import os
+import shutil
+
+_exportTranslatorName = "USD Export"
+
+def _setUpClass(modulePathName):
+    '''Common code for setUpClass() and readOnlySetUpClass()'''
+    standalone.initialize('usd')
+
+    cmds.loadPlugin('mayaUsdPlugin', quiet=True)
```
mattyjams, line 29: Is it worth making the plugin load optional maybe? I'm thinking for export tests in particular, is it worth verifying that the `requires` statement in the Maya file we're loading is what causes the necessary plugin(s) to get loaded to ensure that we're not masking some kind of plugin load/dependency issue by loading manually?

We were admittedly somewhat sloppy about that in the original tests, but we eventually got a bit better about only explicitly loading when necessary.

test/lib/usd/translators/testUsdExportStripNamespaces.py
```diff
@@ -49,8 +48,22 @@ def testExportWithClashStripping(self):
...
+        # does indeed generate the proper TF_RUNTIME_ERROR originally tested
+        # below, and
+        # http://graphics.pixar.com/usd/docs/api/group__group__tf___diagnostic.html#ga4abf7754e5dbf161d2a5a4160fd3b891
+        # describes that "Generally, an error handling delegate will take action
+        # to turn this error into a python exception [...]".
+        #
+        # Unfortunately, at time of writing, this is not happening correctly,
+        # as the Python exception is generated, but the detailed error string
+        # is lost.  We are left with the the generic 'Maya command error'
+        # string, which is far weaker.  PPT, 16-Jun-20.
```
mattyjams, line 61: Hmm, interesting. This test passes currently with the original regex, so was the problem caused by the file moves somehow?

Is there an issue open for this, or can we create one if not?

test/lib/usd/translators/testUsdImportRfMLight.py
```diff
@@ -41,10 +41,8 @@ def setUpClass(cls):
         cmds.loadPlugin('RenderMan_for_Maya', quiet=True)
 
         # Import from USD.
-        usdFilePath = os.path.abspath('RfMLightsTest.usda')
-        cmds.loadPlugin('pxrUsd')
-        cmds.usdImport(file=usdFilePath, shadingMode='pxrRis',
-            readAnimData=True)
+        usdFilePath = os.path.join(os.path.dirname(os.path.realpath(__file__)), "UsdImportRfMLightTest", "RfMLightsTest.usda")
```
mattyjams, line 44: Not using `fixtureUtils` for this?

test/lib/usd/translators/fixturesUtils.py
ppt-adsk: @JGamache-autodesk has suggested the following improvements, which I believe touch upon what you're mentioning:

MAYA_APP_DIR -> should point to somewhere inside that folder. Makes sure tests are always run with fresh user prefs, and allows investigating prefs added or modified by the test.
MAYA_PROJECTS_DIR -> should point to %{ MAYA_APP_DIR}/projects, especially if the test renders to disk using the default output settings or exports a USD scene to the default output path by only specifying a file name.

This is for later improvement, as needed.

test/lib/usd/translators/testUsdExportStripNamespaces.py
ppt-adsk: Excellent suggestion:
https://github.com/Autodesk/maya-usd/issues/601

ppt-adsk:
> Looks good to me, thanks @ppt-adsk!

Thanks @mattyjams for the review!

> Just one request: Can we make the `fixturesUtils` changes to the `testUsdExportRfMLight.py` test? I don't see any modifications to that file. No need to make sure it actually runs and passes. I can verify that it's working when I merge this internally.

Done, see
https://github.com/Autodesk/maya-usd/pull/595/commits/5d1f55824fefc3319119ba94122f64e5b83844b6

> And on that note, fair warning that I may have some small follow-ups to this once I'm able to wrestle our internal build and test system to adapt to these changes. One thing I think might be problematic for us is that we currently install the test scripts in a different location than the test run directory, so I may need to make `fixturesUtils.setUpClass()` slightly more flexible to be able to handle that somehow.

That sounds like a great improvement.

test/lib/usd/translators/fixturesUtils.py
ppt-adsk: Done, see
https://github.com/Autodesk/maya-usd/pull/595/commits/5d1f55824fefc3319119ba94122f64e5b83844b6

mattyjams: Cool, thanks!

mattyjams:
> > Just one request: Can we make the `fixturesUtils` changes to the `testUsdExportRfMLight.py` test? I don't see any modifications to that file. No need to make sure it actually runs and passes. I can verify that it's working when I merge this internally.
> 
> Done, see
> [5d1f558](https://github.com/Autodesk/maya-usd/commit/5d1f55824fefc3319119ba94122f64e5b83844b6)

Thanks @ppt-adsk! I think 5d1f55824fefc3319119ba94122f64e5b83844b6 covers `testUsdImportRfMLight.py`, but `testUsdExportRfMLight.py` needs the same treatment? It looks like it's currently unchanged.

ppt-adsk:
> Thanks @ppt-adsk! I think [5d1f558](https://github.com/Autodesk/maya-usd/commit/5d1f55824fefc3319119ba94122f64e5b83844b6) covers `testUsdImportRfMLight.py`, but `testUsdExportRfMLight.py` needs the same treatment? It looks like it's currently unchanged.

Ah, sorry, missed that one.  Done in 
https://github.com/Autodesk/maya-usd/pull/595/commits/51bee0447e751d902bd550cd3bba4a8a03b920af

PR #600, opened 2020-06-24
MAYA-103462 Filter out wheel events on the variant set combo boxes

fowlertADSK:


lib/usd/ui/ItemDelegate.cpp
```diff
@@ -257,6 +258,9 @@ QLayout* VariantsEditorWidget::createVariantSet(const ItemDelegate* itemDelegate
 
 	cb->addItems(varNames);
 	cb->setCurrentText(varSel);
+	cb->setFocusPolicy( Qt::StrongFocus );
+	cb->installEventFilter(this);
```
fowlertADSK, line 262: Setting to StrongFocus will remove "wheel focus" from the policy.  This is not required to block the wheel event from scrolling the variants, but it will stop it from highlighting it with the wheel scroll event.  Without this it looks kind of odd when you scroll down the view and one of the items highlights even though it doesn't actually edit anything.

PR #602, opened 2020-06-25
MAYA-105232 - Fix for finding boost

seando-adsk:


lib/mayaUsd/CMakeLists.txt
```diff
@@ -20,8 +20,8 @@ add_library(${PROJECT_NAME} SHARED)
 #------------------------------------------------------------------------------
 # modules and definitions
 #------------------------------------------------------------------------------
-if(NOT DEFINED BOOST_ROOT)
-    set(BOOST_ROOT ${PXR_USD_LOCATION})
+if(NOT DEFINED ENV{BOOST_ROOT})
+    set(ENV{BOOST_ROOT} ${PXR_USD_LOCATION})
```
seando-adsk, line 24: We are setting this variable to tell the find_package(Boost) to find the boost cmake files inside the USD folder. However if you look at FindBoost.cmake from any cmake modules folder you will find the code below. You can see that it is actually looking for an environment variable, not a cmake variable.

````
#-------------------------------------------------------------------------------
# Before we go searching, check whether a boost cmake package is available, unless
# the user specifically asked NOT to search for one.
#
# If Boost_DIR is set, this behaves as any find_package call would. If not,
# it looks at BOOST_ROOT and BOOSTROOT to find Boost.
#
if (NOT Boost_NO_BOOST_CMAKE)
  # If Boost_DIR is not set, look for BOOSTROOT and BOOST_ROOT as alternatives,
  # since these are more conventional for Boost.
  if ("$ENV{Boost_DIR}" STREQUAL "")
    if (NOT "$ENV{BOOST_ROOT}" STREQUAL "")
      set(ENV{Boost_DIR} $ENV{BOOST_ROOT})
    elseif (NOT "$ENV{BOOSTROOT}" STREQUAL "")
      set(ENV{Boost_DIR} $ENV{BOOSTROOT})
    endif()
  endif()
````

mattyjams: Hmm... I'm not sure that we should need to do this. If I'm reading the docs and the source correctly, it looks like it should accept **either** CMake variables or environment variables, and that it prefers the CMake variables.

Since we're using CMake versions newer than 3.12,[ CMake policy 00074 ](https://cmake.org/cmake/help/v3.17/policy/CMP0074.html)should use the new behavior by default and it would seem that `<package name>_ROOT` CMake variables should be supported.

That snippet of the `FindBoost.cmake` does indeed look at the environment variables, but it stuffs the result into another environment variable `Boost_DIR`.

Later on, it looks like it checks for the CMake variable version of `BOOST_ROOT` before it looks for the environment variable version:
https://gitlab.kitware.com/cmake/cmake/-/blob/v3.17.3/Modules/FindBoost.cmake#L1551

I know that at least when I'm building core USD, I am specifying the path to Boost using the `BOOST_ROOT` CMake variable, and it is not set as an environment variable. What's the issue you're seeing with the maya-usd build?

seando-adsk: The issue I had was when using a USD compiled with python3 the boost python filename is boost_python37-xxx.lib. When using python2 to execute build.py it was not finding the boost python. What we want by setting the ENV BOOST_ROOT is for the find_package [here](https://gitlab.kitware.com/cmake/cmake/-/blob/v3.17.3/Modules/FindBoost.cmake#L444) to find the boost in the USD build (because we set it to PXR_USD_LOCATION). Then it will enter the if on line 451 and return on line 476 skipping the rest of the file. The file in the USD build **.../lib/cmake/Boost-1.70.0/BoostConfig.cmake** is correctly setup based on the python you built for (2 or 3).

mattyjams: Ok, fair enough. I've been bypassing USD's `build_usd.py` and pointing my maya-usd builds at a specific installation of Boost using `Boost_NO_BOOST_CMAKE`, so I hadn't come across this. But Boost is such a nightmare, I have no doubt that other incantations might expose weird issues like this.

PR #609, opened 2020-06-29
add maya.standalone initialize and uninitialize to testPxrUsdPreviewSurfaceExport test

mattyjams:


mattyjams:
> That initialize would only be necessary if the test is run from another test suite with a different harness. Is this the case?

That is indeed the case. Internally, we run our builds and tests using a SCons-based system, so we bypass everything CMake-related. This test is failing for us internally without this change.

For the CMake infrastructure, I'm not a big fan of the `test.cmake` manipulating/re-writing the test code before execution. I'd argue that the test scripts should be complete in and of themselves, but we can look to discuss and possibly address that later.

mattyjams:
> In that case, code is completely fine since calling initialize twice is not an error, so this test should run fine under both harness.

Thanks @JGamache-autodesk! I've also verified that this test continues to pass when run via the CMake build.

PR #612, opened 2020-06-30
MAYA-105538 minor fix for column width of some text.

fowlertADSK:
Minor change to adjust the column width of some text in the Import UI.  The chosen width is important for the animation details below the Hierarchy View button, but the text that is displayed when the button is invisible doesn't need the extra width.  The current width can effect other parts of the UI, minor problem but can be annoying.


seando-adsk:
What does this change? Do you have a snapshot before and after? Natalia gave me strict guidelines about how to line up items.

fowlertADSK:
> What does this change? Do you have a snapshot before and after? Natalia gave me strict guidelines about how to line up items.

Here's a picture that shows the hierarchy view button + the text when nothing is selected.  You can see that the UI in the sections above the USD options are affected by changing to the text.  It's the extra padding in the column width for that text that forces everything over, and the UI above the options adjust with it.  Flip between having a .usd file selected or not and you'll see that nothing in the USD options change, but you get a horizontal scroll bar + UI around it changes.

This PR should change nothing at all in the USD options, it only avoids having the UI around our options be effected.

image

PR #615, opened 2020-06-30
[mtoh] Add playblasting support.

marsupial:


lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
```diff
@@ -354,7 +363,7 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext) {
     // }
     TF_DEBUG(HDMAYA_RENDEROVERRIDE_RENDER)
         .Msg("MtohRenderOverride::Render()\n");
-    auto renderFrame = [&]() {
+    auto renderFrame = [&](bool markTime = false) {
         const auto originX = 0;
```
pmolodo, line 367: In what situations is renderFrame(false) called?

lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
```diff
@@ -377,8 +386,44 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext) {
...
+            // XXX: Is this better as user-configurable ?
+            constexpr auto msWait = std::chrono::duration<float, std::milli>(100);
+
+            std::shared_ptr<HdxProgressiveTask> progressiveTask =
+                std::dynamic_pointer_cast<HdxProgressiveTask>(tasks.front());
+            if (progressiveTask) {
+                HdTaskSharedPtrVector renderOnly = { progressiveTask };
+                _engine.Execute(_renderIndex, &renderOnly);
+
+                while (!progressiveTask->IsConverged()) {
```
pmolodo, line 410: Is there an easy way that we can add a check to see if the user has requested a cancel to this loop?  Even in normal situations, convergence could take a while, and in worst case, the underlying renderer has some sort of a bug which makes it never properly converge.

marsupial: Uhhh yeah, I seem to have dropped the **_playBlasting** check in a refactor..

lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
marsupial: I have trouble following the logic, but there's a case below the first renderFrame where Storm is called again (looks like to draw wireframe).

marsupial:
Updated with cancelaltion.

lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
pmolodo: Ah, missed that... makes sense then!

lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
pmolodo: Awesome, thanks for the fix!

kxl-adsk:
@marsupial Do you have any update on this PR? Would be great to fix compilation error with older USD versions and get it in.

marsupial:
Moving to 20.08, so I am updating it, but not sure how much it's in the cards for support < 20.05 here though...you might be on your own with that.

kxl-adsk:
We won't be able to accept the PR if it doesn't build with supported USD versions.

marsupial:
Understand, but there isn't much need (or resources) to support these old versions on our side.
Hopefully the latest alleviates the issues going back to 20.02

PR #617, opened 2020-06-30
Updates to support core USD release 20.08

mattyjams:
The `_engine` property was made private starting with `USDIMAGINGGL_API_VERSION` 6 in core USD commit https://github.com/PixarAnimationStudios/USD/commit/c03508ba9850bccabcd4b84e217533bd909352a7. It can still be accessed using the protected `_GetHdEngine()` accessor.

These changes should continue to work before and after that version.

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -31,7 +31,7 @@ See Pixar's official github page for instructions on how to build USD: https://g
 
 |               |      ![](images/pxr.png)          |        
 |:------------: |:---------------:                  |
-|  CommitID/Tags | master: [v19.07](https://github.com/PixarAnimationStudios/USD/releases/tag/v19.07) or [v19.11](https://github.com/PixarAnimationStudios/USD/releases/tag/v19.11) or [v20.02](https://github.com/PixarAnimationStudios/USD/releases/tag/v20.02) or [v20.05](https://github.com/PixarAnimationStudios/USD/releases/tag/v20.05) <br> dev: [e7ff8e8](https://github.com/PixarAnimationStudios/USD/commit/e7ff8e85b422b04a6d685908a32e4b394ef04504) |
+|  CommitID/Tags | release: [v19.07](https://github.com/PixarAnimationStudios/USD/releases/tag/v19.07) or [v19.11](https://github.com/PixarAnimationStudios/USD/releases/tag/v19.11) or [v20.02](https://github.com/PixarAnimationStudios/USD/releases/tag/v20.02) or [v20.05](https://github.com/PixarAnimationStudios/USD/releases/tag/v20.05) or [v20.08](https://github.com/PixarAnimationStudios/USD/releases/tag/v20.08) <br> dev: [238d0f4](https://github.com/PixarAnimationStudios/USD/commit/238d0f4b09d595955d3b5819db427e270756cc24) |
 
 For additional information on building Pixar USD, see the ***Additional Build Instruction*** section below.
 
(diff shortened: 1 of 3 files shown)


kxl-adsk:
@mattyjams I was testing the latest USD changes and noticed that the following test is failing with it
```
14 - testUsdExportAsClip (Failed)
```
Verbose results
```
14: ======================================================================
14: FAIL: testExportAsClip (testUsdExportAsClip.testUsdExportAsClip)
14: ----------------------------------------------------------------------
14: Traceback (most recent call last):
14:   File "testUsdExportAsClip.py", line 123, in testExportAsClip
14:   File "testUsdExportAsClip.py", line 65, in _ValidateSamples
14: AssertionError: different values found on frame: 15
14: non clip: inherited
14: clips:    None
14:
14: ----------------------------------------------------------------------
14: Ran 1 test in 1.083s
```

Can you reproduce this on your end?

mattyjams:
@kxl-adsk Ahh, sorry, I must have missed that earlier, but I actually am seeing that test failure on my end as well.

I did some digging, and it looks like this particular test was actually disabled for us internally a while back since we were seeing it fail intermittently in a similar way. Apparently that change to turn the test off was never translated from our SCons build to the CMake build, so it has just continued to run in the maya-usd repo. Now it seems that there may be a regression between 20.05 and later versions of USD that we weren't able to detect with the test disabled.

I'm able to reproduce the issue just using the USD API in Python and cutting Maya out of the loop, so I don't think there's any issue on the maya-usd side. I've provided some additional notes to the core USD team to see if that helps identify the issue with stitched clips and value resolution.

If it sounds ok to you, I can simply disable that test in CMake for now and then re-enable it once we're able to address the core USD issue and turn the test back on?

kxl-adsk:
> this particular test was actually disabled for us internally a while back since we were seeing it fail intermittently in a similar way

This is very interesting. So this particular test is very stable (with previous versions of USD) on all platforms when running all tests serially. It's as well stable on windows BUT fails randomly on MacOS.

It is fine to disable the test for the unreleased USD version, i.e. 20.08. We should address the stability during parallel test execution before re-enabling as well. Do you mind logging the GitHub issue to re-enable the test and assign it to yourself?

mattyjams:
@kxl-adsk: I worked with the core USD team on this and it turns out that this just needed a test-only fix. The behavior of value clips and stitching was overhauled between 20.05 and 20.08, and we now need to pass an additional option when stitching to preserve the earlier results. See 9f1a0c4e9e6fd49aac860c54c8694a2bc0d84eb9 for that change. I'm not sure whether this will address the intermittent failure issues that you were seeing, but we at least shouldn't need to disable the test.

The documentation on value clips has been updated to describe the new behavior:
https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usd/usd/doxygen/valueClips.dox

I had to wait until the necessary changes to core USD were pushed publicly, so I also updated the build.md to reflect the new core USD commit SHA, which is also the one identified by the `v20.08-rc1` tag.

Re-rebranding this PR as "update to USD `v20.08-rc1`"...

kxl-adsk:
We have adopted an earlier version commit of USD 20.08 internally and if we merge this change now, we are getting
```
06:22:55     ======================================================================
06:22:55     ERROR: testExportAsClip (testUsdExportAsClip.testUsdExportAsClip)
06:22:55     ----------------------------------------------------------------------
06:22:55     Traceback (most recent call last):
06:22:55       File "testUsdExportAsClip.py", line 120, in testExportAsClip
06:22:55     ArgumentError: Python argument types in
06:22:55         pxr.UsdUtils._usdUtils.StitchClips(Layer, list, str)
06:22:55     did not match C++ signature:
06:22:55         StitchClips(pxrInternal_v0_20__pxrReserved__::TfWeakPtr<pxrInternal_v0_20__pxrReserved__::SdfLayer> resultLayer, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > clipLayerFiles, pxrInternal_v0_20__pxrReserved__::SdfPath clipPath, boost::python::api::object startFrame=None, boost::python::api::object endFrame=None, boost::python::api::object clipSet=None)
06:22:55     
06:22:55     ----------------------------------------------------------------------
06:22:55     Ran 1 test in 2.510s
```

We will have to adopt v20.08 final release first before merging. @mattyjams please let me know if this causes you trouble.

mattyjams:
@kxl-adsk: Yeah, unfortunately the dependency on the newly added `interpolateMissingClipValues` parameter means that this test won't work for core USD commits between 20.05 and https://github.com/PixarAnimationStudios/USD/commit/92ca9958eac6efe26a851b1f6b8a233f1607b991.

No problem for us waiting until maya-usd adopts the official 20.08 USD release.

mattyjams:
Just rebased this and updated build.md to tag it against core USD rc3 so people running into issues like #651 and #654 can build from this branch directly if needed.

mattyjams:
Still just keeping this up to date. Rebased and tagged against core USD rc4.

kxl-adsk:
Thank you, Matt. We are waiting for the final 20.08 release before adopting it internally.

mattyjams:
USD 20.08 is now officially released, so one last update to `doc/build.md`. Note that the release branch for core USD is now called `release` instead of `master`, so that is also reflected in the docs now.

mattyjams:
Rebased to pickup the fix from #676 and help keep this branch directly buildable.

PR #618, opened 2020-07-01
export UV sets as USD Texture Coordinate value types by default

mattyjams:
This change flips the default value of the `PIXMAYA_WRITE_UV_AS_FLOAT2` env setting variable so that UV sets export to USD as `texCoord2f`-typed instead of `float2`-typed. The resulting values are handled the same by consumers but the additional context about the role of the value's type can provide some useful information for DCCs and pipelines.

The old behavior of exporting UV sets as `float2`-typed can be restored by setting the environment variable `PIXMAYA_WRITE_UV_AS_FLOAT2` to `true`.


mattyjams:
See the usd-interest thread here for a bit more discussion and context:
https://groups.google.com/forum/#!topic/usd-interest/I6loL1KCY5k

kxl-adsk:
@mattyjams Thank you!

mattyjams:
> There are bunch of other places where we are still writing out UVs as float2, could you also apply the same changes for them for sake of consistency?

Done. Rebased the change and replaced `float2` with `texCoord2f` for all UV sets in https://github.com/Autodesk/maya-usd/pull/618/commits/0c0ebaeba183d76e4bc5956302ec781a58d72d40.

Issue #153, opened 2019-12-04
[MAYA-102530] Unable to select an instance if it is the only instance of a master in VP2


radon199:
Hey,

We tag all references as instanceable in our layouts. This ends up with many masters that may have only a single instance.

I am able to select references with the Adsk Plugin that have at least two instances, but if there is only one ( ie one master and one instance of that master ) I get an warning and the whole proxy is selected.

`Warning: in getInstancedSelectionPath at line 517 of /sw/dev/rousselp/src/MTCR/maya_usd/build/platform-linux/arch-x86_64/os-CentOS-7/boost-1.61.0/python-2.7/maya-2019/maya_usd_stub/usd-19.11.m2/maya_usd/src/maya_usd/lib/render/vp2RenderDelegate/proxyRenderDelegate.cpp -- UFE runtime is not updated for the USD stage. Please save scene and reopen.
`

Reopening the scene does not fix the issue.

**Steps to reproduce :**
1. Open a stage that has a reference that will produce a single master and instance ( ie a Usd Reference that only exists once in the stage but is still tagged as instanceable )
2. Attempt to select this instance with VP2.

cfmoore007:
Actually, this seems to be a duplicate of : https://github.com/Autodesk/maya-usd/issues/153

cfmoore007:
@radon199 - I may have made a duplicate issue ... Can you confirm if your prims show up in the UFE Outliner in this scenario ?

radon199:
Only up to the instanceable prim, everything from the master is hidden.

I wasn’t sure that was relevant because we treat masters that way within our pipeline.

Your issue has more information, except for the fact that viewport selection does work for me with 2 or more instances of the same prim. It is selecting the instanceable prim in this case. 

It should be noted that selecting the instance is the default behavior for usdview/hydra. Usdview however will draw prims under a master to its tree.

cfmoore007:
Ok, I'll leave the other issue open then, as it's slightly different.  Thanks for clarifying.

HdC-adsk:
It is probably a regression due to the draw performance optimization for single-instanced Rprims.

HdC-adsk:
Logged with MAYA-102530.

HdC-adsk:
@santosg87 I just tried but failed to reproduce this issue. Can you reproduce it?

HdC-adsk:
@santosg87 @santosd Guys, have you been able to reproduce the issue?

santosd:
[...]

PR #620, opened 2020-07-03
Fix the regression for selecting single instance objects.

HdC-adsk:
The regression #153 was caused by not using instanced draw for single-instance objects and reproducible on Maya 2020 and before.

Recent Maya preview releases already support consolidation for instanced render items so the performance workaround has been removed, therefore the regression won't be reproduced for those preview releases.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -612,28 +612,44 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
     // is used instead to extract Rprim id.
     const std::string renderItemName = renderItem.name().asChar();
     const auto pos = renderItemName.find_last_of(USD_UFE_SEPARATOR);
-    SdfPath rprimId(renderItemName.substr(0, pos));
+    const SdfPath rprimId(renderItemName.substr(0, pos));
 
-    // If the selection hit comes from an instanced render item, its instance
-    // transform matrices should have been sorted according to USD instance ID,
-    // therefore drawInstID is usdInstID plus 1 considering VP2 defines the
-    // instance ID of the first instance as 1.
+    // If drawInstID is positive, it means the selection hit comes from one instanced render item,
+    // in this case its instance transform matrices have been sorted w.r.t. USD instance ID, thus
+    // usdInstID is drawInstID minus 1 because VP2 instance IDs start from 1.  Otherwise the
+    // selection hit is from one regular render item, but the Rprim can be either plain or single
+    // instance, because we don't use instanced draw for single instance render items in order to
+    // improve draw performance in Maya 2020 and before.
     const int drawInstID = intersection.instanceID();
-    const int usdInstID = drawInstID - 1;
+
 #if defined(USD_IMAGING_API_VERSION) && USD_IMAGING_API_VERSION >= 13
-    rprimId = _sceneDelegate->GetScenePrimPath(rprimId, usdInstID);
+    const int usdInstID = drawInstID > 0 ? drawInstID - 1 : 0;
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -612,28 +612,44 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
...
+        indexPath = _sceneDelegate->GetPathForInstanceIndex(rprimId, 0, nullptr);
+        usdPath = _sceneDelegate->ConvertIndexPathToCachePath(indexPath);
+    }
+#endif
 
     const Ufe::PathSegment pathSegment(usdPath.GetText(), USD_UFE_RUNTIME_ID, USD_UFE_SEPARATOR);
     const Ufe::SceneItem::Ptr& si = handler->createItem(_proxyShapeData->ProxyShape()->ufePath() + pathSegment);
     if (!si) {
-        TF_WARN("UFE runtime is not updated for the USD stage. Please save scene and reopen.");
+        TF_WARN("Failed to create UFE scene item for Rprim '%s'", rprimId.GetText());
```
HdC-adsk, line 652: The previous warning message is no longer valid.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -612,28 +612,44 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
...
+        indexPath = _sceneDelegate->GetPathForInstanceIndex(rprimId, usdInstID, nullptr);
     }
-#endif
+    else {
+        indexPath = rprimId;
+    }
+
+    SdfPath usdPath = _sceneDelegate->ConvertIndexPathToCachePath(indexPath);
 
-    const SdfPath usdPath(_sceneDelegate->ConvertIndexPathToCachePath(rprimId));
```
HdC-adsk, line 631: ConvertIndexPathToCachePath() is only required when calling with GetPathForInstanceIndex().

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -612,28 +612,44 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
...
+    // If drawInstID is positive, it means the selection hit comes from one instanced render item,
+    // in this case its instance transform matrices have been sorted w.r.t. USD instance ID, thus
+    // usdInstID is drawInstID minus 1 because VP2 instance IDs start from 1.  Otherwise the
+    // selection hit is from one regular render item, but the Rprim can be either plain or single
+    // instance, because we don't use instanced draw for single instance render items in order to
+    // improve draw performance in Maya 2020 and before.
     const int drawInstID = intersection.instanceID();
-    const int usdInstID = drawInstID - 1;
+
 #if defined(USD_IMAGING_API_VERSION) && USD_IMAGING_API_VERSION >= 13
```
HdC-adsk, line 625: I fixed the single instance selection issue for both conditional compilation branches.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -612,28 +612,44 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
...
-    rprimId = _sceneDelegate->GetScenePrimPath(rprimId, usdInstID);
+    const int usdInstID = drawInstID > 0 ? drawInstID - 1 : 0;
+    SdfPath usdPath = _sceneDelegate->GetScenePrimPath(rprimId, usdInstID);
```
HdC-adsk, line 627: @mattyjams I think this new API also handles nested instancing, selection for nested instancing case can work as expected (before this API we could only select the highest instancing group but not an exact instance from the viewport). Very nice!

PR #621, opened 2020-07-03
MAYA-102319 Basic useRegistry import support

JGamache-autodesk:
Add the useRegistry import mode which uses an info:id-based registry to process individual UsdShade nodes.
Implemented basic import of UsdPreviewSurface and UsdUVTexture nodes to plugin pxrUsdPreviewSurface materials.
Split monolithic materialWriter file into proper C++ hierarchy.

Note 1: does not convert to Maya native materials (yet).
Note 2: does not solve the issue of having multiple importers for a single info:id (for commonly used names like "standard_surface")


lib/mayaUsd/nodes/proxyShapePlugin.cpp
```diff
@@ -129,7 +129,7 @@ MStatus
 MayaUsdProxyShapePlugin::finalize(MFnPlugin& plugin)
 {
     // If more than one plugin still has us registered, do nothing.
-    if (_registrationCount-- > 1) {
+    if (!_registrationCount || _registrationCount-- > 1) {
```
JGamache-autodesk, line 132: Request from previous review: fix negative registration counts.

lib/usd/translators/CMakeLists.txt
```diff
@@ -17,14 +17,12 @@ target_sources(${TARGET_NAME}
...
         cameraReader.cpp
         cameraWriter.cpp
-        fileTextureWriter.cpp
         instancerWriter.cpp
         jointWriter.cpp
         lightReader.cpp
         lightWriter.cpp
         locatorWriter.cpp
         materialReader.cpp
-        materialWriter.cpp
```
JGamache-autodesk, line 27: As requested in a previous review: split the monolithic materialWriter.cpp file per dependency node.

lib/usd/translators/shading/usdMaterialWriter.h
```diff
@@ -0,0 +1,68 @@
...
+PXR_NAMESPACE_OPEN_SCOPE
+
+class SdfPath;
+class TfToken;
+class UsdMayaWriteJobContext;
+class UsdShadeShader;
+class UsdTimeCode;
+
+/// Shader writer for exporting Maya's material shading nodes to USD.
+class PxrUsdTranslators_MaterialWriter : public UsdMayaShaderWriter
```
JGamache-autodesk, line 36: Utility base class for often-used functions.

lib/usd/translators/shading/usdUVTextureReader.cpp
```diff
@@ -0,0 +1,277 @@
...
+    // Connect it:
+    MString connectCmd;
+    connectCmd.format("fileTexturePlacementConnect \"^1s\" \"^2s\"", depFn.name(), uvDepFn.name());
+    MGlobal::displayWarning(connectCmd);
+    status == MGlobal::executeCommand(connectCmd);
+    if (status != MS::kSuccess) {
+        return false;
+    }
+
+    // File
```
JGamache-autodesk, line 153: Does the exact reverse of usdFileTextureWriter.

lib/usd/pxrUsdPreviewSurface/usdPreviewSurfaceReader.cpp
```diff
@@ -0,0 +1,147 @@
...
+    UsdShadeShader shaderSchema = UsdShadeShader(prim);
+    if (!shaderSchema) {
+        return false;
+    }
+
+    MStatus           status;
+    MFnDependencyNode depFn;
+    if (!(UsdMayaTranslatorUtil::CreateShaderNode(
+              MString(prim.GetName().GetText()),
+              PxrMayaUsdPreviewSurfaceTokens->MayaTypeName.GetText(),
```
JGamache-autodesk, line 84: Importing back to pxrUsdPreviewSurface plugin material at this point int time.
Importing to Maya native materials to happen in the next PR.

lib/usd/pxrUsdPreviewSurface/usdPreviewSurfaceReader.cpp
```diff
@@ -0,0 +1,160 @@
...
+        auto mayaAttrName = GetMayaNameForUsdAttrName(input.GetFullName());
+        if (mayaAttrName == TfToken()) {
+            continue;
+        }
+        MPlug mayaAttr = depFn.findPlug(mayaAttrName.GetText(), true, &status);
+        if (status == MS::kSuccess) {
+            if (mayaAttrName == PxrMayaUsdPreviewSurfaceTokens->DiffuseColorAttrName
+                || mayaAttrName == PxrMayaUsdPreviewSurfaceTokens->EmissiveColorAttrName
+                || mayaAttrName == PxrMayaUsdPreviewSurfaceTokens->SpecularColorAttrName) {
+                // These values are exported unscaled, so they also reimport unscaled:
```
JGamache-autodesk, line 101: Will have to check to see if we want them scaled on export.

lib/usd/translators/shading/usdUVTextureReader.cpp
```diff
@@ -0,0 +1,309 @@
...
+          && uvDepFn.setObject(uvObj))) {
+        // we need to make sure assumes those types are loaded..
+        TF_RUNTIME_ERROR(
+            "Could not create node of type '%s' for shader '%s'.\n",
+            _tokens->place2dTexture.GetText(),
+            prim.GetName().GetText());
+        return false;
+    }
+
+    // Connect manually (fileTexturePlacementConnect is not available in batch):
```
JGamache-autodesk, line 177: Code from MEL script fileTexturePlacementConnect copied here since the script is not sourced in batch.

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -390,8 +399,209 @@ TF_REGISTRY_FUNCTION_WITH_TAG(UsdMayaShadingModeExportContext, useRegistry)
     );
 }
 
+namespace {
+
+class UseRegistryShadingModeImporter {
```
kxl-adsk, line 404: Please document this class and methods

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -390,8 +399,209 @@ TF_REGISTRY_FUNCTION_WITH_TAG(UsdMayaShadingModeExportContext, useRegistry)
...
+        // No shader reader exists for this node yet, so create one.
+        UsdMayaShaderReaderSharedPtr shaderReader;
+
+        TfToken shaderId;
+        shaderSchema.GetIdAttr().Get(&shaderId);
+
+        if (UsdMayaShaderReaderRegistry::ReaderFactoryFn factoryFn
+            = UsdMayaShaderReaderRegistry::Find(shaderId.GetText())) {
+
+            UsdMayaJobImportArgs defaultJobArgs = UsdMayaJobImportArgs::CreateFromDictionary(
```
kxl-adsk, line 495: Why we don't try to get the actual arguments from read job?

lib/usd/pxrUsdPreviewSurface/usdPreviewSurfaceReader.cpp
```diff
@@ -0,0 +1,160 @@
...
+                    useModifier = true;
+                }
+            } else {
+                // This call always scales colors:
+                UsdMayaReadUtil::SetMayaAttr(mayaAttr, input);
+            }
+        }
+    }
+    if (useModifier) {
+        modifier.doIt();
```
kxl-adsk, line 119: Why do we have to set it with DG modifier?

test/lib/usd/pxrUsdPreviewSurface/testPxrUsdPreviewSurfaceExportImportRoundtrip.py
```diff
@@ -0,0 +1,138 @@
...
+        test_dir = os.path.join(os.path.abspath('.'),
+                                "PxrUsdPreviewSurfaceExportTest")
+
+        cmds.workspace(test_dir, o=True)
+
+    @classmethod
+    def tearDownClass(cls):
+        standalone.uninitialize()
+
+    @unittest.skip("Needs investigation on MacOS")
```
kxl-adsk, line 46: This test is crashing on MacOS...same if I try to import assets with materials.
I disabled the test since we gave it to QA to play with. Please re-enable.

lib/usd/translators/shading/usdUVTextureReader.cpp
```diff
@@ -0,0 +1,309 @@
...
+              _tokens->file.GetText(),
+              UsdMayaShadingNodeType::Texture,
+              &status,
+              &_mayaObject)
+          && depFn.setObject(_mayaObject))) {
+        // we need to make sure assumes those types are loaded..
+        TF_RUNTIME_ERROR(
+            "Could not create node of type '%s' for shader '%s'.\n",
+            _tokens->file.GetText(),
+            prim.GetName().GetText());
```
mattyjams, line 155: Might want the full shader path here rather than just the name, and similarly elsewhere.

lib/usd/translators/shading/usdUVTextureReader.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+    return true;
+}
+
+/* virtual */
+TfToken PxrMayaUsdUVTexture_Reader::GetMayaNameForUsdAttrName(const TfToken& usdAttrName)
+{
+    if (_mayaObject.isNull()) {
+        return TfToken();
+    }
```
mattyjams, line 287: Remove this check.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -191,7 +191,21 @@ global proc int mayaUsdTranslatorExport (string $parent,
...
+            string $exporters[] = `mayaUSDListShadingModes -export`;
+            optionMenuGrp -l "Materials:" shadingModePopup;
+            for ($modeName in $exporters) {
+                string $niceName = $modeName;
+                if ($modeName == "displayColor") {
+                    $niceName = "Material Colors";
+                } else if ($modeName == "pxrRis") {
+                    $niceName = "RfM Shaders";
+                } else if ($modeName == "useRegistry") {
+                    $niceName = "Preview Surface Shaders";
```
mattyjams, line 204: Again, this is ok as a placeholder for now, but we might want some symmetry between import and export. Looks like import uses "USD Shade"?

test/lib/usd/pxrUsdPreviewSurface/testPxrUsdPreviewSurfaceExportImportRoundtrip.py
```diff
@@ -0,0 +1,138 @@
...
+
+from maya import cmds
+from maya import standalone
+
+
+class testPxrUsdPreviewSurfaceExport(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        standalone.initialize('usd')
```
mattyjams, line 35: Should this be using the new(-ish) `fixtureUtils`?

test/lib/usd/pxrUsdPreviewSurface/testPxrUsdPreviewSurfaceExportImportRoundtrip.py
```diff
@@ -0,0 +1,138 @@
...
+                         ".mirrorU", ".mirrorV", ".stagger", ".wrapU",
+                         ".wrapV", ".repeatUV", ".offset", ".rotateUV",
+                         ".noiseUV", ".vertexUvOne", ".vertexUvTwo",
+                         ".vertexUvThree", ".vertexCameraOne"):
+            cmds.connectAttr(uv_node + att_name, file_node + att_name, f=True)
+
+        cmds.connectAttr(file_node + ".outColor",
+                         material_node + ".diffuseColor", f=True)
+
+        txfile = "PxrUsdPreviewSurfaceExportTest/Brazilian_rosewood_pxr128.png"
```
mattyjams, line 85: This test should just duplicate the texture map file into its own test directory rather than reach across into the test data of another test. We shouldn't really assume that all of the test data lives in directories right next to each other.

lib/mayaUsd/fileio/shaderReaderRegistry.h
```diff
@@ -0,0 +1,137 @@
...
+/// PXRUSDMAYA_REGISTER_SHADER_READER(usdInfoId, readerClass) to register a
+/// reader class with the registry.
+///
+/// In order for the core system to discover the plugin, you need a
+/// \c plugInfo.json that contains the usdInfoId and the Maya plugin to load:
+/// \code
+/// {
+///     "UsdMaya": {
+///         "ShaderReader": {
+///             "mayaPlugin": "myMayaPlugin",
```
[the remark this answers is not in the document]
JGamache-autodesk: The preview surface was moved to mayaUsd plugin, so it is in memory at this point in time.

This also means that we can remove https://github.com/Autodesk/maya-usd/blob/dev/lib/usd/translators/plugInfo.json without any ill effect since mayaUsd_Translator is always in memory as well.

lib/mayaUsd/fileio/shaderReaderRegistry.h
```diff
@@ -0,0 +1,137 @@
...
+///             UsdMayaReadeJobContext& jobCtx) {
+///         // ...
+///     }
+/// };
+/// PXRUSDMAYA_REGISTER_WRITER(myCustomMayaNode, MyReader);
+/// \endcode
+#define PXRUSDMAYA_REGISTER_SHADER_READER(usdInfoId, readerClass)                         \
+    TF_REGISTRY_FUNCTION_WITH_TAG(UsdMayaShaderReaderRegistry, usdInfoId##_##readerClass) \
+    {                                                                                     \
+        UsdMayaShaderReaderRegistry::Register(                                            \
```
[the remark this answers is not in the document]
JGamache-autodesk: This will add a more helpful error that will help understand the unhelpful error messages about not being able to convert the lambda to a ReaderFactoryFn (because the lambda could not return a UsdMayaPrimReaderSharedPtr),

lib/usd/pxrUsdPreviewSurface/usdPreviewSurfaceReader.cpp
JGamache-autodesk: Yes, I am actually curious why UsdMayaReadUtil::SetMayaAttr does that. It would seem easier to simply set the value directly into the MPlug.

lib/mayaUsd/fileio/shaderReaderRegistry.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <pxr/base/tf/token.h>
+#include <pxr/pxr.h>
+
+#include <map>
+#include <string>
+#include <utility>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+TF_DEFINE_PRIVATE_TOKENS(_tokens, (UsdMaya)(ShaderReader));
```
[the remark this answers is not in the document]
JGamache-autodesk: I added WhitespaceSensitiveMacros: ['TF_DEFINE_PRIVATE_TOKENS']
To the clang-format rules. This should preserve the layout for this macro.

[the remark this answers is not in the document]
JGamache-autodesk: Sadly, this is available only in the latest LLVM 11. Might be a little too advanced yet.

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -390,8 +399,209 @@ TF_REGISTRY_FUNCTION_WITH_TAG(UsdMayaShadingModeExportContext, useRegistry)
...
+    }
+
+    MObject Read()
+    {
+        const UsdShadeMaterial& shadeMaterial = _context->GetShadeMaterial();
+        if (!shadeMaterial) {
+            return MObject();
+        }
+
+        UsdShadeShader surfaceShader = shadeMaterial.ComputeSurfaceSource();
```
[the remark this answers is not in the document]
JGamache-autodesk: Exactly. I first need to do a working export to a shading dialect that supports "displacement" and "volume" so I can import it back and make sure the roundtrip works.

lib/usd/translators/shading/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+    PRIVATE
+        usdBlinnWriter.cpp
+        usdFileTextureWriter.cpp
+        usdLambertWriter.cpp
+        usdMaterialWriter.cpp
+        usdPhongWriter.cpp
+        usdPhongEWriter.cpp
+        usdReflectWriter.cpp
+        usdStandardSurfaceWriter.cpp
+        usdUVTextureReader.cpp
```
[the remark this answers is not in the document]
JGamache-autodesk: At some point in the future we will have the option of exporting to multiple shading dialects.

Which naming scheme do you prefer?
1- Each dialect in its own folder:
   translators/shading/usdPreviewSurface/blinnWriter.cpp
    translators/shading/materialX/blinnWriter.cpp
    translators/shading/maya/blinnWriter.cpp
2- Dialect as a prefix:
   translators/shading/usdBlinnWriter.cpp
    translators/shading/materialxBlinnWriter.cpp
    translators/shading/mayaBlinnWriter.cpp
With option 2, the file name will match the class name.

lib/usd/pxrUsdPreviewSurface/usdPreviewSurfaceReader.cpp
```diff
@@ -0,0 +1,160 @@
...
+    const auto&    prim = _GetArgs().GetUsdPrim();
+    UsdShadeShader shaderSchema = UsdShadeShader(prim);
+    if (!shaderSchema) {
+        return false;
+    }
+
+    MStatus           status;
+    MFnDependencyNode depFn;
+    if (!(UsdMayaTranslatorUtil::CreateShaderNode(
+              MString(prim.GetName().GetText()),
```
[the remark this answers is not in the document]
JGamache-autodesk: Maya will automatically append a numeric increments to make sure the name is unique, but that might not be the best solution. I suspect we will have proponents for both options, which might require exposing that as an import option instead of going one way or the other.

On the same subject, it would be nice to produce less materials when possible: if we have the same cup/plate/crayon asset file coming from 2 different paths, it would be nice to be able to produce a single Maya material that link both geometries. The Kitchen_set demo scene has a nice two-toned tiled floor with should, in a perfect world, result in only 2 lambert materials when importing using the "Display color" shading mode.

Do you think this would be of interest?

test/lib/usd/translators/testUsdExportImportRoundtripPreviewSurface.py
```diff
@@ -82,7 +82,8 @@ def testPxrUsdPreviewSurfaceRoundtrip(self):
         cmds.connectAttr(file_node + ".outColor",
                          material_node + ".diffuseColor", f=True)
 
-        txfile = "PxrUsdPreviewSurfaceExportTest/Brazilian_rosewood_pxr128.png"
+        txfile = "/".join(["UsdExportImportRoundtripPreviewSurface",
```
mattyjams, line 85: `os.path.sep` here instead of `"/"`, or just use `os.path.join()` instead?

lib/usd/pxrUsdPreviewSurface/usdPreviewSurfaceReader.cpp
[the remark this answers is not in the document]
mattyjams: I think it's less about trying to ensure that the node names are unique, and more important that the translation is as consistent as possible across multiple roundtrips. When there are potentially existing overs on the prims being exported, it would be easy for those to fall off if a name collision on import caused the node name to change, which then affects the prim path when re-exported.

And yeah, I think material deduplication would be an interesting feature to pursue if we can detect that the "source" materials in USD are all the same material (e.g. bindings from multiple gprims across references that all resolve to what is ultimately a single material).

lib/usd/translators/shading/CMakeLists.txt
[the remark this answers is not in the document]
mattyjams: Hmm. I'm not quite sure I'm following yet how the selection of an export "dialect" impacts what ultimately gets authored in USD, and how/if there's an overlap there with UsdShade's concept of renderContext. There ongoing discussion here, so I guess I wouldn't worry too much about the naming for now.

PR #622, opened 2020-07-03
MAYA-105232 - misc fixes for python3

seando-adsk:


build.py
```diff
@@ -99,7 +99,7 @@ def GetVisualStudioCompilerAndVersion():
         # VisualStudioVersion environment variable should be set by the
         # Visual Studio Command Prompt.
         match = re.search(
-            "(\d+).(\d+)",
+            r"(\d+)\.(\d+)",
```
seando-adsk, line 102: When using python 3 this gives the following warning:
````
build.py:102: DeprecationWarning: invalid escape sequence \d
      "(\d+).(\d+)",
````
I believe you got this function from the USD build script. I checked and they made this fix a while ago. The env var has something like "15.0" in it. The **r** tells python this is a raw string which should not be interpreted. And you need **\.** to find the . otherwise that is the regex matching any char.

File: plugin/al/usdtransaction/CMakeLists.txt [hunk 2]
```diff
@@ -34,7 +34,6 @@ foreach(INPUT_FILE ${PY_INIT_FILES})
   get_filename_component(OUTPUT_PATH ${OUTPUT_FILE} DIRECTORY)
   install(FILES
         ${INPUT_FILE}  # .py files
-        ${INPUT_FILE}c # .pyc files
```
seando-adsk, line 37: We should not install .pyc files. They are not needed and they aren't there in python3 anyways.

build.py
HamedSabri-adsk: Ugh I see. Great find. Didn't realize that.

plugin/al/usdtransaction/CMakeLists.txt
HamedSabri-adsk: Makes sense.

PR #623, opened 2020-07-05
Wait after configuring cache before making changes. Re-enable tests.

kxl-adsk:


test/lib/testMayaUsdProxyAccessor.py
```diff
@@ -105,6 +105,8 @@ def __enter__(self):
         self.auto_key_chars = cmds.autoKeyframe(q=True, characterOption=True)
         cmds.autoKeyframe(e=True, state=False)
 
+        self.waitForCache(None)
```
kxl-adsk, line 108: This is the actual fix...we make sure configuration changes were applied before starting changes in the data model.

ppt-adsk: As discussed, if you pass in the unit test argument to waitForCache() in the context constructor, you'll be able to preserve the assertion on previous line 162.  Perhaps not a crucial thing, but even if it's just a minor improvement you can consider it.

PR #629, opened 2020-07-07
MAYA-105538 adjust button width

fowlertADSK:


plugin/adsk/scripts/mayaUsdTranslatorImport.mel
```diff
@@ -45,6 +45,7 @@ global proc int mayaUsdTranslatorImport (string $parent,
     int $bResult = 0;
     int $cw1 = 180;
     int $cw2 = 300;
+    int $bw = 125;
```
fowlertADSK, line 48: Instead of just hiding the button, which continues to take up space, I'll now adjust the width to 1 just like we do with the text.  I also tried setting "manage" to false which is documented to not take up any space, but the rowLayout still has the column set to have a width and it doesn't adjust the way I'd like.

plugin/adsk/scripts/mayaUsdTranslatorImport.mel
```diff
@@ -59,11 +60,11 @@ global proc int mayaUsdTranslatorImport (string $parent,
...
                 ;
-                textField -ed false -nbg true -text "Please select a file to enable this option" -w $cw1 mayaUsdTranslator_SelectFileField;
-                button -label "Hierarchy View" -c "string $usdFile = currentFileDialogSelection(); usdImportDialog $usdFile;" mayaUsdTranslator_ViewHierBtn;
+                textField -ed false -nbg true -text "Please select a file to enable this option" -w $cw2 mayaUsdTranslator_SelectFileField;
```
fowlertADSK, line 63: Putting it back to cw2 so it doesn't chop off the first word.

plugin/adsk/scripts/mayaUsdTranslatorImport.mel
```diff
@@ -59,11 +60,11 @@ global proc int mayaUsdTranslatorImport (string $parent,
...
+                button -label "Hierarchy View" -c "{string $usdFile = currentFileDialogSelection(); usdImportDialog $usdFile;}" mayaUsdTranslator_ViewHierBtn;
```
fowlertADSK, line 64: This is a fix to not put $usdFile as a global var

plugin/adsk/scripts/mayaUsdTranslatorImport.mel
fowlertADSK: And since we'll be explicitly flipping the width of the button we need to pick a regular width instead of going with the default, so adding a variable like some of the others.

plugin/adsk/scripts/mayaUsdTranslatorImport.mel
```diff
@@ -167,18 +168,16 @@ global proc int mayaUsdTranslatorImport (string $parent,
             {
                 // Hide the field (and shrink it) and show the button (which will align with label).
                 textField -e -vis false -w 1 mayaUsdTranslator_SelectFileField;
+                button -e -visible true -w $bw mayaUsdTranslator_ViewHierBtn;
```
seando-adsk, line 171: The previous code below has a button "exists" test. And we have a exists test for the textfield. I guess we should keep that test here and line 177?

fowlertADSK: I started a comment that explained this but maybe I didn't actually leave it here.  I didn't bother with the extra "exists" check because the logic that decides if this UI should be created or not creates both of them so if one exists then the other is guaranteed to be there as well.

seando-adsk: Okay looks good then.

PR #630, opened 2020-07-07
MAYA-103187 Add option to save .usd files as binary or ascii

fowlertADSK:
This branch is everything from https://github.com/Autodesk/maya-usd/pull/591 including some requested changes to use the file format id's, and a test added, but redone on the latest dev branch.  I had set this aside while there was some ongoing work on the tests and when coming back to it after everything else went in I ended up just cherry picking in the original changes I had into a new branch.

### DIFF
--- a/test/lib/usd/translators/testUsdExportFileFormat.py
+++ b/test/lib/usd/translators/testUsdExportFileFormat.py
@@ -1,0 +1,143 @@
+#!/pxrpythonsubst
+#
+# Copyright 2020 Autodesk
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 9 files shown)


test/lib/usd/translators/testUsdExportFileFormat.py
```diff
@@ -0,0 +1,143 @@
+#!/pxrpythonsubst
```
fowlertADSK, line 1: This test is the one new change from the original PR.  All this does is export a poly sphere from Maya using the 'file' command as well as the mayaUSDExport command, checking the default option (which should be usdc) as well as explicitly setting the .usd format to usda and usdc.

test/lib/usd/translators/testUsdExportFileFormat.py
```diff
@@ -0,0 +1,143 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        standalone.uninitialize()
+
+    def GetUnderlyingUsdFileFormat(self, layer):
+        if Sdf.FileFormat.FindById('usda').CanRead(layer.identifier):
+            return 'usda'
+        elif Sdf.FileFormat.FindById('usdc').CanRead(layer.identifier):
+            return 'usdc'
+        return ''
```
fowlertADSK, line 45: Took me some time to figure out how to check what the .usd file format is from Python, the c++ class that also does this doesn't look like it's exposed in Python, but I did find this in a similar core USD test.

HamedSabri-adsk: @fowlertADSK  Nice find!

PR #631, opened 2020-07-07
MAYA-105621 - MayaUSD: install USD from artifact beside plugin

seando-adsk:
#### MAYA-105621 - MayaUSD: install USD from artifact beside plugin
* When given a USD relative path, set mod file and path/pythonpath accordingly.
* Specific mod file for Windows (with paths) and Linux/Mac. Added .mod specific tags.


CMakeLists.txt
```diff
@@ -177,7 +177,16 @@ endif()
 #------------------------------------------------------------------------------
 # install
 #------------------------------------------------------------------------------
-configure_file(mayaUSD.mod.template ${PROJECT_BINARY_DIR}/mayaUSD.mod)
+if (DEFINED MAYAUSD_TO_USD_RELATIVE_PATH)
+    set(USD_INSTALL_LOCATION "${CMAKE_INSTALL_PREFIX}/${MAYAUSD_TO_USD_RELATIVE_PATH}")
+else()
+    set(USD_INSTALL_LOCATION ${PXR_USD_LOCATION})
+endif()
```
seando-adsk, line 184: When given a USD relative path (from the ADSK internal build system) we will use that for the USD path in the .mod file.

CMakeLists.txt
```diff
@@ -177,7 +177,16 @@ endif()
...
+if (IS_WINDOWS)
+    configure_file(mayaUSD_Win.mod.template ${PROJECT_BINARY_DIR}/mayaUSD.mod)
+else()
+    configure_file(mayaUSD.mod.template ${PROJECT_BINARY_DIR}/mayaUSD.mod)
+endif()
```
seando-adsk, line 189: Created Windows specific .mod file (with PATH) and one for Linux/Mac (no PATH)

cmake/usd.cmake
```diff
@@ -15,9 +15,15 @@
 #
 function(init_usd)
     # Adjust PYTHONPATH, PATH
-    mayaUsd_append_path_to_env_var("PYTHONPATH" "${PXR_USD_LOCATION}/lib/python")
+    if (DEFINED MAYAUSD_TO_USD_RELATIVE_PATH)
+        set(USD_INSTALL_LOCATION "${CMAKE_INSTALL_PREFIX}/${MAYAUSD_TO_USD_RELATIVE_PATH}")
+    else()
+        set(USD_INSTALL_LOCATION ${PXR_USD_LOCATION})
+    endif()
```
seando-adsk, line 22: Again when given a USD relative path, use that for the PATH and PYTHONPATH.

mayaUSD.mod.template
```diff
@@ -1,32 +1,37 @@
-+ USD ${USD_VERSION} ${PXR_USD_LOCATION}
++ USD ${USD_VERSION} ${USD_INSTALL_LOCATION}
```
seando-adsk, line 1: Changed to new variable which is set from top-level CMakeLists.txt above.

mayaUSD.mod.template
```diff
@@ -1,32 +1,37 @@
...
+icons: 
+plug-ins: 
+presets: 
+scripts: 
+resources:
```
seando-adsk, line 6: Use Maya module specific tags to not add anything specific for these paths since USD is not a Maya plugin and doesn't need Maya to know about its icons/presets/etc.

mayaUSD.mod.template
```diff
@@ -1,32 +1,37 @@
-+ USD ${USD_VERSION} ${PXR_USD_LOCATION}
++ USD ${USD_VERSION} ${USD_INSTALL_LOCATION}
+icons: 
+plug-ins: 
+presets: 
+scripts: 
+resources: 
 PYTHONPATH+:=lib/python
-PATH+:=bin
-PATH+:=lib
```
seando-adsk, line 4: These PATH vars are no needed on Linux/Mac. There is a Windows specific .mod template now.

mayaUSD.mod.template
```diff
@@ -1,32 +1,37 @@
...
 
 + PXR_USDMAYA ${USD_VERSION} ${CMAKE_INSTALL_PREFIX}/plugin/pxr
+icons: maya/lib/usd/usdMaya/resources
+plug-ins: maya/plugin
+scripts: maya/lib/usd/usdMaya/resources
```
seando-adsk, line 12: Use module tags rather than setting the vars directly.

mayaUSD_Win.mod.template
```diff
@@ -0,0 +1,42 @@
++ USD ${USD_VERSION} ${USD_INSTALL_LOCATION}
+icons: 
+plug-ins: 
+presets: 
+scripts: 
+resources: 
+PYTHONPATH+:=lib/python
+PATH+:=bin
+PATH+:=lib
```
seando-adsk, line 9: Windows specific mod template which just has these extra PATH lines. Otherwise it's identical.

cmake/usd.cmake
seando-adsk: I realized (from a clean build) that this change breaks the build. We need the PATH and PYTHONPATH to usd during the configure step for generating the schemas. So for that step we need to use USD from it's original location.

But then for running the tests we would want PATH & PYTHONPATH to point to the USD in the install folder.  I need to think some more how to fix this.

cmake/test.cmake
```diff
@@ -197,11 +197,22 @@ finally:
...
+        set(USD_INSTALL_LOCATION ${PXR_USD_LOCATION})
+    endif()
+    list(APPEND mayaUsd_varname_PYTHONPATH
+        "${USD_INSTALL_LOCATION}/lib/python")
+    if(IS_WINDOWS)
+        list(APPEND mayaUsd_varname_PATH
+            "${USD_INSTALL_LOCATION}/bin")
+        list(APPEND mayaUsd_varname_PATH
+            "${USD_INSTALL_LOCATION}/lib")
+    endif()
```
seando-adsk, line 215: I reverted my change in usd.cmake where I was using the USD from install folder because it was wrong. In our internal builds USD is only copied to install folder during install stage. During configure stage we'll need to use USD from the given input USD location.

However at test time we now check if we were given a USD relative path. If so we use'll that one and add USD to the PYTHONPATH from there. 

Here I've only add USD to PATH when on Windows, just like usd.cmake does.

I believe that even all the PATH items above should only need to be done on Windows, but I didn't change them so as to not pollute this PR.

CMakeLists.txt
seando-adsk: Note: "USD_INSTALL_LOCATION" is a variable inside the module template files.

cmake/test.cmake
```diff
@@ -197,9 +197,30 @@ finally:
...
+
+    # NOTE: this should come after any setting of PATH/PYTHONPATH so
+    #       that our entries will come first.
+    # Inherit any existing PATH/PYTHONPATH, but keep it at the end.
+    # This is needed (especially for PATH) because we will overwrite
+    # both with the values from our list and we need to keep any
+    # system entries.
     list(APPEND mayaUsd_varname_PATH $ENV{PATH})
     list(APPEND mayaUsd_varname_PYTHONPATH $ENV{PYTHONPATH})
```
seando-adsk, line 225: I realized this is important, especially for PATH. We will overwrite PATH with the value(s) set here. So we need to make sure we append the existing PATH otherwise we'll erase all the system entries and things won't work.

seando-adsk:
@mattyjams Can you have quick look at these changes to make sure nothing conflicts with the way you are building? Most of the changes are around "MAYAUSD_TO_USD_RELATIVE_PATH" which we use in our internal build system to generate relative rpaths. I split the Maya module template file into two (one Windows specific with PATH entries, and the other for Linux/Mac which use rpath). I also added some extra tags so Maya doesn't add extra paths for icons/scripts/etc for USD which aren't needed or wanted, since it is not a Maya plugin.

PR #634, opened 2020-07-08
Add menu in Hierarchy View to reset the model and variant selections

fowlertADSK:
New menu item in the Hierarchy View dialog that will reset all of the edited variant sets back to what is in the usd file, and sets the checked state back to the root selection

### DIFF
--- a/lib/usd/ui/TreeModel.cpp
+++ b/lib/usd/ui/TreeModel.cpp
@@ -20,6 +20,8 @@
 #include <QtWidgets/QTreeView>
 #include <QtCore/QSortFilterProxyModel>
 
+#include <pxr/usd/usd/variantSets.h>
+
 #include <mayaUsdUI/ui/TreeItem.h>
 #include <mayaUsdUI/ui/ItemDelegate.h>
 #include <mayaUsdUI/ui/IMayaMQtUtil.h>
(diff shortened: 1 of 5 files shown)


lib/usd/ui/USDImportDialog.ui
```diff
@@ -23,7 +23,75 @@
...
+        <enum>QAbstractItemView::ExtendedSelection</enum>
+       </property>
+       <property name="uniformRowHeights">
+        <bool>false</bool>
+       </property>
+      </widget>
+     </item>
+    </layout>
+   </item>
+   <item row="4" column="0">
```
fowlertADSK, line 94: .ui file will be hard to review, but the reason it looks like so much has changed in the diff is because the new menu item goes at the top, so it becomes row 0.  This pushes all of the other items down a row and ends up creating a diff like this.

I found it harder than expected to add the menu bar + item at the top of the dialog, there is no interactive way of doing this in Qt Designer as far as I can tell unless we change the base class type of the dialog to a window.  I did find a way to add the menu bar as a widget to a layout and ended up basically hand-editing this file to make it work.

lib/usd/ui/USDImportDialog.cpp
```diff
@@ -71,6 +71,7 @@ USDImportDialog::USDImportDialog(const std::string& filename, const ImportData*
 	fUI->treeView->setAlternatingRowColors(true);
 	fUI->treeView->setSelectionMode(QAbstractItemView::SingleSelection);
 	QObject::connect(fUI->treeView, SIGNAL(clicked(const QModelIndex&)), this, SLOT(onItemClicked(const QModelIndex&)));
+	QObject::connect(fUI->actionReset_File, SIGNAL(triggered(bool)), this, SLOT(onResetFileTriggered()));
```
fowlertADSK, line 74: Hooking up the menu item action to the reset function.  It's not a check-box menu so I'm ignoring the bool state variable.

lib/usd/ui/TreeModel.cpp
```diff
@@ -51,6 +53,49 @@ TreeItem* findTreeItem(TreeModel* treeModel, const QModelIndex& parent, std::fun
...
+	for (int r=0; r<treeModel->rowCount(parent); ++r)
+	{
+		QModelIndex variantIndex = treeModel->index(r, TreeModel::kTreeColumn_Variants, parent);
+		TreeItem* variantItem = static_cast<TreeItem*>(treeModel->itemFromIndex(variantIndex));
+		assert(variantItem);
+
+		if (variantItem->variantSelectionModified()) 
+		{
+			resetVariantToPrimSelection(variantItem);
+		}
```
seando-adsk, line 89: Here you assert, but then just use variantItem. So if it happened to be nullptr you would crash and thus never get to the call to resetVariantToPrimSelection where inside you have an if(variantItem) check. I would be more inclined to put the nullptr check on line 86 and move the assert into resetVariantToPrimSelection and remove the if check there. This makes resetVariantToPrimSelection a helper that is assuming valid input.

lib/usd/ui/USDImportDialog.cpp
```diff
@@ -172,4 +173,13 @@ void USDImportDialog::onItemClicked(const QModelIndex& index)
 	}
 }
 
+void USDImportDialog::onResetFileTriggered()
+{
+	if (fTreeModel)
```
seando-adsk, line 178: What is our coding standard stance on if checks? In Maya only a bool is a boolean, so if the type is not a boolean don't use it as one.

lib/usd/ui/USDImportDialog.ui
seando-adsk: Okay as long as it still works. Sometimes even when editing in Qt Designer and adding one thing the diff was all messed up.

lib/usd/ui/TreeModel.cpp
fowlertADSK: Yeah, good point.  Will change.

lib/usd/ui/USDImportDialog.cpp
fowlertADSK: Not sure if we've listed this in the coding standard or not, but I can add a "nullptr !=" here.

PR #637, opened 2020-07-08
Use proper boost target name

seando-adsk:


plugin/al/lib/AL_USDMaya/CMakeLists.txt
```diff
@@ -201,7 +201,7 @@ target_link_libraries(${LIBRARY_NAME}
     usdImagingGL
     vt
     Boost::python
-    $<IF:$<BOOL:${IS_WINDOWS}>,Boost::thread,${Boost_THREAD_LIBRARY}>
+    Boost::thread
```
seando-adsk, line 204: Because of my change in PR #602 , we should now use the proper boost target name here. This is needed for python3 builds where Boost_THREAD_LIBRARY is not defined.

seando-adsk:
@mattyjams Sorry I had to rework it again to handle both our python 2 and 3 builds.

plugin/al/lib/AL_USDMaya/CMakeLists.txt
```diff
@@ -182,6 +182,15 @@ target_include_directories(
...
 )
 
+# There were bugs in both cmake and findboost with the boost version string. So in
+# order to support all boost and cmake versions, we check how the boost version is
+# defined and create our version comparison string.
+if(${Boost_VERSION} MATCHES "^([0-9]+)\\.([0-9]+)\\.([0-9]+)$")
+    set(boost_1_70_0_ver_string "1.70.0")
+else()
+    set(boost_1_70_0_ver_string "107000")
+endif()
```
seando-adsk, line 192: USD uses a different version of boost on Windows (1.70.0) and Linux/Mac, python 2 (1.66.0) and python 3 (1.70.0). Because of this the Boost_VERSION string has a different format between the two versions. And Boost_VERSION_STRING is not always set. There is also a fix in cmake 3.15, but we support 3.13-3.17.

    -- Boost_VERSION = 106100
    -- Boost_VERSION_STRING = 1.61.0

    -- Boost_VERSION = 1.70.0
    -- Boost_VERSION_STRING =

So in order to work around all of this I detect the format of the boost version and set a string for 1.70.0 to compare against.

plugin/al/lib/AL_USDMaya/CMakeLists.txt
```diff
@@ -201,7 +210,7 @@ target_link_libraries(${LIBRARY_NAME}
     usdImagingGL
     vt
     Boost::python
-    $<IF:$<BOOL:${IS_WINDOWS}>,Boost::thread,${Boost_THREAD_LIBRARY}>
+    $<IF:$<VERSION_GREATER_EQUAL:${Boost_VERSION},${boost_1_70_0_ver_string}>,Boost::thread,${Boost_THREAD_LIBRARY}>
```
seando-adsk, line 213: If we are 1.70.0 or greater use the Boost::thread, otherwise use Boost_THREAD_LIBRARY. Note: in 1.70.0, Boost_THREAD_LIBRARY is not set because we are setting boost_root to the USD location so it uses the boost config cmake files there.

I would like to use Boost::thread all the time, but that causes a problem with v1.66.0 on the Mac in that it brings in Boost::chrono, but we don't have that.

seando-adsk:
The more boost we can remove the better! Thanks.

PR #638, opened 2020-07-08
Minor update to doc regarding pip install

seando-adsk:


seando-adsk:
I used the new instructions to download and install pip into Maya 2020 and add the future package.

PR #641, opened 2020-07-09
Selection by kind

HdC-adsk:
The pull request is to add a "UsdSelectionKind" option var to override selection
behavior (GUI support is coming separately).

The default behavior is to select the exact prim which is picked in the viewport.

If the option var is assigned with a non-empty string to indicate a Kind token,
the nearest ancestor that is assigned with that Kind or a subcategory of that
Kind will be selected when a prim is picked in the viewport. If the prim doesn't
have any qualified ancester, it falls back to the default behavior (but it can be
slower due to the cost of walking the scene hierarchy).

If the option var is undefined or empty, the default behavior will be used to
select the exact prim.

For details about Kind, please refer to [Kind : Extensible Categorization](https://graphics.pixar.com/usd/docs/api/kind_page_front.html).

Examples:

// Select the nearing ancester which is a model or its subcategory.
optionVar -sv "UsdSelectionKind" "model";

// Select the nearing ancester which is a group or its subcategory.
optionVar -sv "UsdSelectionKind" "group";

// Select the nearing ancester which is a user-defined kind or its subcategory.
optionVar -sv "UsdSelectionKind" "user-defined";

// Select the exact prim that is picked in the viewport.
optionVar -sv "UsdSelectionKind" "";

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -25,6 +25,9 @@
 #include <pxr/base/tf/diagnostic.h>
 #include <pxr/base/tf/stringUtils.h>
 #include <pxr/usdImaging/usdImaging/delegate.h>
+#include <pxr/usd/kind/registry.h>
+#include <pxr/usd/usd/prim.h>
+#include <pxr/usd/usd/modelAPI.h>
 #include <pxr/imaging/hdx/renderTask.h>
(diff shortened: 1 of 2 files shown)


cfmoore007:
From a usability perspective, it appears to work as advertised.

```
def Xform "apple" (
    kind = "component"
)
```

**1) select by component**
cmds.optionVar(sv=('UsdSelectionKind','component'))

optionVar_component

**2) optionVar empty ( select prim )**
cmds.optionVar(sv=('UsdSelectionKind',''))
optionVar_empty

Thanks, @HdC-adsk !

PR #646, opened 2020-07-10
[mtoh] Draw all the things that mtoh will not in the post-render pass.

marsupial:
There still seems to be a lot of things not drawn when in mtoh (Nurbs, SubdivSurf, GreasePencils).
This may or may not have been introduced in aeb87c46, but this simplifies the Pre & Post render passes to just not draw Meshes & PluginShapes (for UsdProxies) in the _PostRender_.

Additionally the HdSt menu for selection is removed, as I seem to be able to accomplish this with the other two checkboxes for wireframe & overlay.  And I haven't been able to get the menu to show what it's purpose is. (perhaps @pmolodo can shed some light on what it does or if the original intent is still valid).

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/renderGlobals.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/renderGlobals.cpp
@@ -299,8 +299,6 @@ MObject MtohCreateRenderGlobals() {
                 defGlobals.delegateParams.maximumShadowMapResolution);
             return o;
         });
-    static const TfTokenVector selectionOverlays{MtohTokens->UseHdSt,
-                                                 MtohTokens->UseVp2};
     _CreateBoolAttribute(
         node, _tokens->mtohWireframeSelectionHighlight,
         defGlobals.wireframeSelectionHighlight);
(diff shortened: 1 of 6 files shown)


lib/mayaUsd/render/mayaToHydra/renderGlobals.cpp
```diff
@@ -312,9 +301,6 @@ MObject MtohCreateRenderGlobals() {
         });
     static const TfTokenVector selectionOverlays{MtohTokens->UseHdSt,
                                                  MtohTokens->UseVp2};
```
pmolodo, line 303: This static var - and the MtohTokens it uses - should probably also be removed. I've pushed a commit that you can add to this PR to do that:

https://github.com/LumaPictures/maya-usd/commit/b99dea4ab927c8fa2b900e821eaec62a1d052508

...or I can make a new PR after this one, whatever you'd prefer.

marsupial:
Updated with  e4dc5c9 appended.

pmolodo:
Great, thanks! @kxl-adsk, let's get this merged!

PR #648, opened 2020-07-10
[mtoh] Move some Storm only task/params into the Storm block.

marsupial:
These parameters never make it to anyone but Storm, as a ShadowTask is only created when it is the renderer.

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
@@ -455,21 +455,6 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext) {
         it->PreFrame(drawContext);
     }
 
-    // TODO: Is there a way to improve this? Quite silly.
-    auto enableShadows = true;
-    auto* lightParam = drawContext.getLightParameterInformation(
-        0, MHWRender::MDrawContext::kFilteredIgnoreLightLimit);
-    if (lightParam != nullptr) {
-        MIntArray intVals;
-        if (lightParam->getParameter(
-                MHWRender::MLightParameterInformation::kGlobalShadowOn,
-                intVals) &&
-            intVals.length() > 0) {
-            enableShadows = intVals[0] != 0;
-        }
-    }
-    _taskController->SetEnableShadows(enableShadows);
-
     HdxRenderTaskParams params;
     params.enableLighting = true;
     params.enableSceneMaterials = true;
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
```diff
@@ -406,21 +406,6 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext) {
...
     }
 
-    // TODO: Is there a way to improve this? Quite silly.
-    auto enableShadows = true;
-    auto* lightParam = drawContext.getLightParameterInformation(
-        0, MHWRender::MDrawContext::kFilteredIgnoreLightLimit);
-    if (lightParam != nullptr) {
-        MIntArray intVals;
-        if (lightParam->getParameter(
-                MHWRender::MLightParameterInformation::kGlobalShadowOn,
```
kxl-adsk, line 416: What makes this HdSt specific?

marsupial: The light Lighting and Shadow tasks are only ever created for Storm, and without those tasks those parameters do nothing.

kxl-adsk: Looking at the [comment in Hydra](https://github.com/PixarAnimationStudios/USD/blob/ebac0a8b6703f4fa1c27115f1f013bb9819662f4/pxr/imaging/hdx/taskController.cpp#L269) and wondering how we can make sure the code in here will stay in sync with Hydra.

One option would be to add TF_CODING_ERROR in `else` case (line 478) which would validate that lighting and shadow tasks were not created?

Another option would be to keep setting these parameters outside of `_isUsingHdSt` but add a condition to check if a specific task (lighting or shadow) is created.

Either one would work for me...and would save us debugging time when Hydra changes.

marsupial: To me it seems that would just be duplicating the **if-else** outside of _taskController.cpp_ with not a clear picture of how/when it changes what would be needed (i.e. I'm not convinced those will ever be common to all delegates).

kxl-adsk: We are making assumptions over here and matching the hardcoded choice from the task controller. It is a good habit to provide validation for these assumptions so if they ever change, we won't have to spend hours investigating what's going on. Especially that we will forget about this code as soon as it's merged.

BUT, maybe the reason you are not convinced is that there is no easy way to add this validation? I looked quickly and can't find any way to check if _shadowTaskId or _simpleLightTaskId is empty. Iterating over all tasks and casting doesn't look ideal. If that's the case, let's document it better in the source code, point to actual tasks that these parameters are affecting and what in task controller is responsible for creating them.

kxl-adsk:
@marsupial when you force push commits and override the history it makes it hard to follow the changes since review comments are no longer pointing to valid commits and we can't diff before and after. Would it be possible in the future to simply push new commits when making new changes and when resolving conflicts with dev branch, just merge it instead of rebasing?

PR #649, opened 2020-07-10
[mtoh] Remove explicit call to HdChangeTracker::RprimInserted.

marsupial:
HdRenderIndex::InsertRprim will do this using HdRprim::GetInitialDirtyBitsMask().
Additionally, the previous use of HdChangeTracker::AllDirty could cause over-invalidation:
 If the delegate doesn't have knowledge of those bits,  it won't necessarily clear them leading to `HdRprim::Sync` called on every `engine.Execute` invocation.


HdC-adsk:
The change shouldn't cause any functional change, because what you remove is an unused argument of ` HdMayaDelegateCtx::InsertRprim()`.

PR #650, opened 2020-07-11
[mtoh] Fix stage-repopulation.

marsupial:
When rendering via **mtoh**, changes to **UsdProxy.filepath** aren't working.
(If not always, at least when **VP_RENDER_DELEGATE=0**)

### DIFF
--- a/lib/usd/hdMaya/adapters/proxyAdapter.cpp
+++ b/lib/usd/hdMaya/adapters/proxyAdapter.cpp
@@ -176,7 +176,13 @@ void HdMayaProxyAdapter::_OnStageSet(const MayaUsdProxyStageSetNotice& notice)
                 "(ProxyShape: "
                 "%s)\n",
                 GetDagPath().partialPathName().asChar());
+
         CreateUsdImagingDelegate();
+        auto stage = _proxy->getUsdStage();
+        if (_usdDelegate && stage) {
+            _usdDelegate->Populate(stage->GetPseudoRoot());
+            _isPopulated = true;
+        }
     }
 }
 


kxl-adsk:
Similar problem we were solving in VP2RenderDelegate - see condition https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp#L421 and implementation details for ProxyRenderDelegate::ProxyShapeData::IsUsdStageUpToDate() and ProxyRenderDelegate::ProxyShapeData::IsExcludePrimsUpToDate()

I will ask @williamkrick to review this change and point out opportunities for aligning the two render paths on the same mechanism to detect when stage on proxy shape changed.

williamkrick:
> 
> 
> Similar problem we were solving in VP2RenderDelegate - see condition https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp#L421 and implementation details for ProxyRenderDelegate::ProxyShapeData::IsUsdStageUpToDate() and ProxyRenderDelegate::ProxyShapeData::IsExcludePrimsUpToDate()
> 
> I will ask @williamkrick to review this change and point out opportunities for aligning the two render paths on the same mechanism to detect when stage on proxy shape changed.

My approach in Vp2RenderDelegate is to hold all the data we use from the ProxyShape on a class (ProxyShapeData) and do all the version tracking there. ProxyShapeData could be refactored into ProxyShapeBase or a similar common area and re-used by mtoh. ProxyShapeData can directly compare cached values to the current ProxyShapeBase data to detect changes in simple types, or compare versionIDs to detect changes in complex types.

Right now y'all have a callback registered `TfNotice::Register(me, &HdMayaProxyAdapter::_OnStageSet);` so so know exactly when the proxy is changing. I'm not sure where that system comes from, we don't use that in Vp2RenderDelegate.

lib/usd/hdMaya/adapters/proxyAdapter.cpp
```diff
@@ -155,7 +149,9 @@ void HdMayaProxyAdapter::CreateUsdImagingDelegate() {
         _id.AppendChild(TfToken(TfStringPrintf(
             "ProxyDelegate_%s_%p", _proxy->name().asChar(), _proxy))),
         _proxy, GetDagPath()));
-    _isPopulated = false;
+
+    _usdDelegate->Populate(stage->GetPseudoRoot());
+    _isPopulated = true;
```
williamkrick, line 154: I don't like the side effect of calling `_usdDelegate->Populate` inside of `CreateUsdImagingDelegate()`. Would it be possible to instead call `HdMayaProxyAdapter::Populate()` from `_OnStageSet` after the `CreateUsdImagingDelegate` call there? The code right now is confusing because both orderings of calls is safe (`Populate` can automatically call `CreateUsdImagingDelegate` & `CreateUsdImagingDelegate` can call `Populate`).

I guess the real question is why have two methods if you can call them in either order and get the right answer? And if the two separate methods are different enough to really be necessary, are they named correctly to describe what they actually do?

lib/usd/hdMaya/adapters/proxyAdapter.cpp
```diff
@@ -61,18 +61,11 @@ void HdMayaProxyAdapter::Populate() {
     TF_DEBUG(HDMAYA_AL_POPULATE)
         .Msg("HdMayaProxyDelegate::Populating %s\n", _proxy->name().asChar());
 
-    auto stage = _proxy->getUsdStage();
```
kxl-adsk, line 64: It's true that the base proxy shape currently always has a valid stage. This doesn't have to be true a) in future, b) for plugin proxy shapes (example AL proxy shape). Please bring back this validation.

PR #652, opened 2020-07-12
Convenience function for printing the list of queried composition arcs in order

HamedSabri-adsk:
Often time I need to debug the queried composition arcs in C++ and this function helps to pretty-print them.

Same thing can also be achieved in python. See testUsdPrimCompositionQuery here :

https://github.com/PixarAnimationStudios/USD/blob/ebac0a8b6703f4fa1c27115f1f013bb9819662f4/pxr/usd/usd/testenv/testUsdPrimCompositionQuery.py#L28

### DIFF
--- a/lib/usd/utils/util.cpp
+++ b/lib/usd/utils/util.cpp
@@ -16,14 +16,63 @@
 
 #include "util.h"
 
+#include <pxr/usd/pcp/layerStack.h>
 #include <pxr/usd/sdf/layer.h>
 #include <pxr/usd/usd/prim.h>
 #include <pxr/usd/usd/primCompositionQuery.h>
 #include <pxr/usd/usd/stage.h>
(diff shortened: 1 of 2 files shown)


lib/usd/utils/util.cpp
```diff
@@ -137,4 +186,23 @@ hasSpecs(const UsdPrim& prim)
     return found;
 }
 
+
+void
+printCompositionQuery(const UsdPrim& prim)
```
ppt-adsk, line 191: Might be useful to pass in an std::ostream&, instead of forcing use of std::cout.

HamedSabri-adsk: @ppt-adsk please see de4ed7c4b7dda2c6c78ba34b618e29d088bdd0fe

PR #653, opened 2020-07-12
Minor cmake clean up pxr plugin

HamedSabri-adsk:


plugin/pxr/maya/lib/usdMaya/CMakeLists.txt
```diff
@@ -26,7 +25,6 @@ pxr_shared_library(${PXR_PACKAGE}
         mayaUsd
 
     INCLUDE_DIRS
-        ${Boost_INCLUDE_DIRS}
```
HamedSabri-adsk, line 29: these variables are empty?? The mystery part is how usdMaya links against Boost?

HamedSabri-adsk:
> If an upstream library is included and used directly by a downstream library, the downstream library should explicitly declare that dependency.

Respectfully, I have a different view on this. Explicitly declaring upstream library again in the downstream library is totally redundant.  If upstream library is properly using PRIVATE/PUBLIC/INTERFACE keywords, one can look up the upstream dependencies to find out what they are. 

For example, **mayaUsdUtils** is declaring  gf, usd, sdf as PUBLIC.  The downstream library DiffCore doesn't really need to repeat them again. 

mayaUsdUtils
https://github.com/Autodesk/maya-usd/blob/461668a75deeb46d36cf55566b119010ff491a7d/lib/usd/utils/CMakeLists.txt#L42

DiffCore
https://github.com/Autodesk/maya-usd/blob/461668a75deeb46d36cf55566b119010ff491a7d/test/lib/usd/utils/CMakeLists.txt#L23

I am happy to revert the change if you still disagree.  The important change in this PR is getting rid of those empty boost entries. They are very confusing and Sean and I were wondering how usdMaya builds without them. That's the real mystery :)

mattyjams:
Fair enough. I definitely see your point, and if the only reason that we were declaring that `usdMaya` has a dependency on `vt`, for example, was strictly because some other library that `usdMaya` depends on (`mayaUsd` in this case) is dependent on `vt`, then I completely agree with you. In that case the re-declaration is entirely redundant and should be removed.

In the case here though, `usdMaya` itself **directly** uses symbols from many of the libraries being removed by this change. So to me, it becomes an extension of the "include what you use" principle, or I guess now "link what you use".

Consider the case where in some future change, for whatever reason we decide to remove all usage of `vt` from `mayaUsd` and we remove `vt` as a dependency for `mayaUsd`. `usdMaya` still has `#include` directives for and its own usage of `vt`, so building `usdMaya` would suddenly break because of an otherwise inconsequential change in an upstream library. We would actually need to add `vt` as a dependency to `usdMaya` to fix that. Making sure we declare that direct dependency, even though we would otherwise pick it up transitively from some other library, helps make the build more robust against such changes in linkages.

So I'd still contend that many of these declared dependencies should remain here (basically any library that is `#include`'d by a code file in `usdMaya` should appear in the list of `LIBRARIES` in its CMakeLists.txt. `vt` is indeed still being used, but `plug` no longer is, so that would be a good candidate for removal.

HamedSabri-adsk:
> Consider the case where in some future change, for whatever reason we decide to remove all usage of vt from mayaUsd and we remove vt as a dependency for mayaUsd. usdMaya still has #include directives for and its own usage of vt, so building usdMaya would suddenly break because of an otherwise inconsequential change in an upstream library. We would actually need to add vt as a dependency to usdMaya to fix that. Making sure we declare that direct dependency, even though we would otherwise pick it up transitively from some other library, helps make the build more robust against such changes in linkages.

@mattyjams Thanks for the explanation. It makes sense to me now.

plugin/pxr/maya/lib/usdMaya/CMakeLists.txt
```diff
@@ -6,7 +6,6 @@ pxr_shared_library(${PXR_PACKAGE}
         gf
         js
         kind
-        plug
         sdf
         tf
         usd
```
mattyjams, line 11: Of the libraries still listed here, it looks like you could now also remove these if you wanted to:

```
js
usdLux
usdRi
usdShade
usdSkel
```

HamedSabri-adsk: Done. Please see f70e798850e98e562be556e0fdfdaeb52df3b188

PR #655, opened 2020-07-13
Clean up cmake variable names according to guidelines

HamedSabri-adsk:


cmake/test.cmake
```diff
@@ -75,64 +75,64 @@ function(mayaUsd_add_test test_name)
             "used with PYTHON_MODULE or PYTHON_COMMAND")
     endif()
 
-    # set the working_dir
+    # set the WORKING_DIR
```
seando-adsk, line 78: Minor, but since this is a comment I think it should not be uppercase and be "working dir"

cmake/test.cmake
```diff
@@ -75,64 +75,64 @@ function(mayaUsd_add_test test_name)
...
-        set(command ${PREFIX_COMMAND})
+        set(COMMAND ${PREFIX_COMMAND})
     endif()
 
-    if(pytest_code)
+    if(PYTEST_CODE)
         if(NOT PREFIX_NO_STANDALONE_INIT)
             # first, indent pycode
-            mayaUsd_indent(indented_pytest_code "${pytest_code}")
+            mayaUsd_indent(indented_PYTEST_CODE "${PYTEST_CODE}")
```
seando-adsk, line 107: Why is "indented" still lowercase?

cmake/test.cmake
```diff
@@ -75,64 +75,64 @@ function(mayaUsd_add_test test_name)
...
-main(module=${module_name})
+import ${MODULE_NAME}
+main(module=${MODULE_NAME})
 ")
     elseif(PREFIX_PYTHON_COMMAND)
-        set(pytest_code "${PREFIX_PYTHON_COMMAND}")
+        set(PYTEST_CODE "${PREFIX_PYTHON_COMMAND}")
     else()
-        set(command ${PREFIX_COMMAND})
+        set(COMMAND ${PREFIX_COMMAND})
```
seando-adsk, line 101: I actually think that "COMMAND" is too generic for a variable name.

lib/mayaUsd/base/CMakeLists.txt
```diff
@@ -6,27 +6,27 @@ target_sources(${PROJECT_NAME}
...
 
-set(headers
+set(HEADERS
     api.h
     debugCodes.h
 )
 
 # -----------------------------------------------------------------------------
-# promoted headers
+# promoted HEADERS
```
seando-adsk, line 15: Again comment should not be uppercase

HamedSabri-adsk:
@seando-adsk Thanks, I will address them all today. There is also a bug introduced with this change that affects the RPATH logics which I am fixing as well.

cmake/test.cmake
HamedSabri-adsk: @seando-adsk  please see 07d962479409f7375694679beed213da71859b0c

cmake/test.cmake
HamedSabri-adsk: Addressed 07d962479409f7375694679beed213da71859b0c

lib/mayaUsd/base/CMakeLists.txt
HamedSabri-adsk: Addressed 07d962479409f7375694679beed213da71859b0c

cmake/test.cmake
HamedSabri-adsk: indented_PYTEST_CODE is a function argument. My goal for this PR was to tackle the set variables.

cmake/utils.cmake
```diff
@@ -92,8 +92,8 @@ function(mayaUsd_init_rpath rpathRef origin)
     endif()
     # mayaUsd_add_rpath uses REALPATH, so we must make sure we always
     # do so here too, to get the right relative path
-    get_filename_component(origin "${origin}" REALPATH)
-    set(${rpathRef} "${origin}" PARENT_SCOPE)
+    get_filename_component(ORIGIN "${ORIGIN}" REALPATH)
+    set(${rpathRef} "${ORIGIN}" PARENT_SCOPE)
 endfunction()
```
HamedSabri-adsk, line 97: This is a good example of why lower/upper case matters. :)

cmake/test.cmake
seando-adsk: Okay, but you changed the "pytest_code" to uppercase so now it looks weird being part lower and part upper.

PR #659, opened 2020-07-14
MAYA-104778: Handle two more illegal cases:

HamedSabri-adsk:
- Spaces are replaced with "_"
- Names are not allowed to start with digits.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
@@ -37,6 +37,8 @@
 #include <cassert>
 #endif
 
+#include <cctype>
+
 MAYAUSD_NS_DEF {
 namespace ufe {
 
@@ -131,8 +133,13 @@ bool UsdUndoRenameCommand::renameRedo()
         _newName = uniqueName(childrenNames, _newName);
     }
 
+    // names are not allowed to start to digit numbers
+    if(std::isdigit(_newName.at(0))){
+        _newName = prim.GetName();
+    }
+
     // all special characters are replaced with `_`
-    const std::string specialChars{"~!@#$%^&*()-=+,.?`':{}|<>[]/"};
+    const std::string specialChars{"~!@#$%^&*()-=+,.?`':{}|<>[]/' '"};
     std::replace_if(_newName.begin(), _newName.end(), [&](auto c){
         return std::string::npos != specialChars.find(c);
     }, '_');
--- a/lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
@@ -139,7 +139,7 @@ bool UsdUndoRenameCommand::renameRedo()
     }
 
     // all special characters are replaced with `_`
-    const std::string specialChars{"~!@#$%^&*()-=+,.?`':{}|<>[]/' '"};
+    const std::string specialChars{"~!@#$%^&*()-=+,.?`':{}|<>[]/ "};
     std::replace_if(_newName.begin(), _newName.end(), [&](auto c){
         return std::string::npos != specialChars.find(c);
     }, '_');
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -131,8 +133,13 @@ bool UsdUndoRenameCommand::renameRedo()
...
     }
 
+    // names are not allowed to start to digit numbers
+    if(std::isdigit(_newName.at(0))){
+        _newName = prim.GetName();
+    }
+
     // all special characters are replaced with `_`
-    const std::string specialChars{"~!@#$%^&*()-=+,.?`':{}|<>[]/"};
+    const std::string specialChars{"~!@#$%^&*()-=+,.?`':{}|<>[]/' '"};
```
HamedSabri-adsk, line 142: Space also needs to be replaced with under score.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -131,8 +133,13 @@ bool UsdUndoRenameCommand::renameRedo()
         _newName = uniqueName(childrenNames, _newName);
     }
 
+    // names are not allowed to start to digit numbers
+    if(std::isdigit(_newName.at(0))){
+        _newName = prim.GetName();
+    }
```
HamedSabri-adsk, line 139: Names are not allowed to start with digits and should to replaced with the old name. This is based on Natalia's request to match Maya's behavior.

ppt-adsk: Hmm, Maya's behavior is confusing / buggy.  Here's what I get when I rename a sphere with a leading numeric:
select -r pSphere1 ;
rename "pSphere1" "1pSphere1";
// Warning: line 1: New name contains invalid characters. Illegal characters were converted to "_". //
// Result: pSphere1 //

Yes, you do get the old name back, but you also get a warning saying "illegal character" will be replaced with an underscore --- and then you don't get the underscore...

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
ppt-adsk: Think you added the space AND the two single quotes, which you don't need because you're in a string.  In fact, the single quote is already in your string, so now you have it 3 times...

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
HamedSabri-adsk: > 
> 
> Hmm, Maya's behavior is confusing / buggy. Here's what I get when I rename a sphere with a leading numeric:
> select -r pSphere1 ;
> rename "pSphere1" "1pSphere1";
> // Warning: line 1: New name contains invalid characters. Illegal characters were converted to "_". //
> // Result: pSphere1 //
> 
> Yes, you do get the old name back, but you also get a warning saying "illegal character" will be replaced with an underscore --- and then you don't get the underscore...

@ppt-adsk Can't agree more. I have noticed this as well and it is very confusing. 

I've already brought this up with design team and not really sure if we want to mimic exactly how Maya's handling all this cases. 

My main goal for now is to prevent Maya crash when illegal characters are used in names.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
HamedSabri-adsk: @ppt-adsk Duh!! Thanks.  I could use std::isspace for sake of readability but that's something for later.  All these string manipulation and logics could also go to somewhere like stringUtils.h later on.

See bb59f68ab5fa7d55a44d264dce73a67a2384ce89

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
ppt-adsk: O.K., fair enough, so for now, no crash, but no warning, and no underscore.  Works for me.

PR #660, opened 2020-07-15
Fix anonymous layer default name

HamedSabri-adsk:


lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -584,7 +585,7 @@ MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
             }
             else {
                 // Create a new stage in memory with an anonymous root layer.
-                usdStage = UsdStage::CreateInMemory("", loadSet);
+                usdStage = UsdStage::CreateInMemory(kanonymousLayerName, loadSet);
```
HamedSabri-adsk, line 588: @seando-adsk  Setting identifier passed to **CreateInMemory** routine automatically sets the anonymous name.

See here:
https://github.com/PixarAnimationStudios/USD/blob/ebac0a8b6703f4fa1c27115f1f013bb9819662f4/pxr/usd/usd/stage.cpp#L777

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -98,6 +98,7 @@ TF_DEFINE_PUBLIC_TOKENS(MayaUsdProxyShapeBaseTokens,
 MayaUsdProxyShapeBase::ClosestPointDelegate
 MayaUsdProxyShapeBase::_sharedClosestPointDelegate = nullptr;
 
+const auto kanonymousLayerName{"anonymousLayer1"};
```
seando-adsk, line 101: createInMemory takes a std::string, so instead of auto why not make this a std::string? 

According to our coding standards constants "should be named with a leading “k” followed by UpperCamelCase". So this should be "kA..."

HamedSabri-adsk: auto already deduces the type to std::string.

HamedSabri-adsk: @seando-adsk please see 5dd82214fad78999da9c022c81cb6d2e588356d3

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -98,6 +98,7 @@ TF_DEFINE_PUBLIC_TOKENS(MayaUsdProxyShapeBaseTokens,
 MayaUsdProxyShapeBase::ClosestPointDelegate
 MayaUsdProxyShapeBase::_sharedClosestPointDelegate = nullptr;
 
+const std::string kAnonymousLayerName{"anonymousLayer1"};
```
kxl-adsk, line 101: Not a blocker for the change...but non exported definitions we prefer to put in the unnamed namespace - https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities. Consider adapting to this rule in a new PR.

PR #663, opened 2020-07-15
MAYA-104858: fix regression after UsdUndoAddNewPrimCommand was introduced.

HamedSabri-adsk:


lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -93,63 +93,6 @@ class SetVariantSelectionUndoableCommand : public Ufe::UndoableCommand
...
-                TF_RUNTIME_ERROR("Failed to create new prim type: %s", _primToken.GetText());
-        }
-    }
-
-private:
-    PXR_NS::UsdStageWeakPtr _stage;
-    PXR_NS::SdfPath _primPath;
-    PXR_NS::TfToken _primToken;
-};
-#endif
```
HamedSabri-adsk, line 151: Removed duplicated AddNewPrimUndoableCommand since we now are using UsdUndoAddNewPrimCommand.

lib/mayaUsd/ufe/UsdUndoAddNewPrimCommand.cpp
```diff
@@ -48,7 +48,7 @@ UsdUndoAddNewPrimCommand::UsdUndoAddNewPrimCommand(const UsdSceneItem::Ptr& usdS
     _stage = usdSceneItem->prim().GetStage();
     if (_stage) {
         // Append the parent path and the requested name into a full ufe path.
-        _newUfePath = appendToPath(ufePath, name);
+        _newUfePath = appendToPath(ufePath, name + std::to_string(1));
```
HamedSabri-adsk, line 51: Added my earlier fix for MAYA-104858 here.

lib/mayaUsd/ufe/UsdContextOps.cpp
ppt-adsk: Not really a duplicate, it's for UFE versions before 0.2.17.

lib/mayaUsd/ufe/UsdUndoAddNewPrimCommand.cpp
ppt-adsk, line 51: Could just be name + "1".

ppt-adsk: Did this cause a test to fail to indicate the regression?  If not, we're missing a test...

HamedSabri-adsk: @ppt-adsk please see 324bff09d53cab1f693fed35c9f0a293b435a34c

HamedSabri-adsk: It looks not. This is the new class that @fowlertADSK introduced.  Agreed, we need a test for this.

lib/mayaUsd/ufe/UsdContextOps.cpp
HamedSabri-adsk: @ppt-adsk Correct me if I am wrong but I don't see anything in the removed code that relied on older version of UFE. 

Almost all of code in AddNewPrimUndoableCommand got duplicated in the new UsdUndoAddNewPrimCommand class with a few logic changes which are not really relied on UFE interface changes.

Keeping AddNewPrimUndoableCommand  and UsdUndoAddNewPrimCommand  together is very confusing and hard to maintain.

fowlertADSK: Yes, Hamed and I talked about this one yesterday.  I think when I was first going through the code updating to 2.17 I was wrapping everything in the ifdef's, but this one could have been removed entirely.  There was my fix to not use the name as the type when creating a group, but that doesn't require us to keep the old version around.

ppt-adsk: O.K., you guys know what you're talking about.

PR #664, opened 2020-07-15
Remove testUsdMayaAppDir

kxl-adsk:
Recently a script exploit was identified which could corrupt Maya’s software’s settings, run malicious code, and be propagated to other Maya files (*.ma and .mb). To read more about it, [go here](https://www.autodesk.com/trust/security-advisories/adsk-sa-2020-0003?_ga=2.57833711.457183772.1594544936-1238440735.1590257173) and [here](https://knowledge.autodesk.com/support/maya/troubleshooting/caas/sfdcarticles/sfdcarticles/How-to-diagnose-and-clean-Maya-ScriptExploit-issues.html).

We just landed a new security option of not running userSetup .mel/.py scripts by default ... and this caused testUsdMayaAppDir to stop working (you might have noticed PF coming back red for last two days). It's not possible to change the default from the script (otherwise we would continue to have security issues), so we started to wonder what is the real value of keeping testUsdMayaAppDir. We couldn't find one thus we propose to remove this test.


pmolodo:
So... I don't really have any objection to removing this particular test, per-se... as long as we're still setting the MAYA_APP_DIR to a fresh / empty directory on every test run, or have some other way of ensuring that every test runs with default preferences.

Otherwise, we risk user preferences affecting the tests, making them less reproducible.  I originally created https://github.com/PixarAnimationStudios/USD/pull/209/files after running into a problem with a test failing in some circumstances and not others, and it ended up being due to a user preference (though I no longer remember the details).

pmolodo:
Also - note that you'll want to remove this dir as well if you're getting rid of the test:
https://github.com/Autodesk/maya-usd/tree/dev/plugin/pxr/maya/lib/usdMaya/testenv/UsdMayaAppDirTest

kxl-adsk:
@pmolodo Thank you for this info. It totally makes sense to start tests with default preferences. The removal of the regression test in only because it relies on userSetup.py which by default won't be executed due to security reasons I highlighted in the description.

Regarding the folder I forgot to remove - good catch - see 256d690f6c

PR #665, opened 2020-07-16
Added implementation for new UFE unparent interface (MAYA-105276)

ppt-adsk:
This pull request implements the Ufe::Hierarchy::defaultParent() method.  This is used when unparenting, a.k.a parent -world.  In this implementation, the default parent of USD nodes is their proxy shape, so that calling parent -w on them will place them as a child of the corresponding USD stage pseudo-root.

### DIFF
--- a/lib/mayaUsd/ufe/Global.cpp
+++ b/lib/mayaUsd/ufe/Global.cpp
@@ -174,5 +174,10 @@ Ufe::Rtid getUsdRunTimeId()
     return g_USDRtid;
 }
 
+Ufe::Rtid getMayaRunTimeId()
+{
+    return g_MayaRtid;
+}
+
 } // namespace ufe
 } // namespace MayaUsd
--- a/lib/mayaUsd/ufe/Global.h
+++ b/lib/mayaUsd/ufe/Global.h
@@ -38,5 +38,9 @@ MStatus finalize();
 MAYAUSD_CORE_PUBLIC
 Ufe::Rtid getUsdRunTimeId();
 
+//! Return the run-time ID allocated to Maya.
+MAYAUSD_CORE_PUBLIC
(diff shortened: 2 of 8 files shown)


ppt-adsk:
Hi @pilarmolinalopez , I hope this is of some interest to you and your team.

pilarmolinalopez:
Thanks @ppt-adsk ! I'll take a look.

ppt-adsk:
> Can we move the insertChild functions next to insertChildCmd in each of the .cpp files instead of having them by themselves at the bottom?

Sounds good, will do.  Will wait for suggestions from Pilar and fix eveything in a single commit.

> Only thing I’m starting to wonder about is the usefulness of the non-cmd functions that implement UFE interfaces but that’s a separate conversation. This review isn't the place to have that conversation, but for someone taking a look at this review it will look odd that "insertChild" can just return null. I believe that this is because everyone goes through the Cmd interface and the Cmd does not necessarily have to use the non-Cmd function as its implementation correct?

The non-cmd versions are there because if you're writing code that is non-interactive (and therefore does not need undo / redo), it's really awkward to call the cmd version of the interface, call execute on the resulting cmd object, then ask the cmd object for the result of the call, when all you want is to call e.g. insertedChild = parent.insertChild(child, position);

What has been suggested before (and that I like) is that the non-cmd versions be virtual, but have a default implementation right in UFE that does exactly what I mentioned, create the command object, call execute on it, and return the result.  That way, if the run-time implementer doesn't want to bother with details, they can just implement the cmd version and that is sufficient.

But as you say, this is a conversation for a later branch.

kxl-adsk:
@ppt-adsk I'm marking this PR as do-not-merge-yet since there are changes you still want to make. Please remove that label when PR is ready for merge.

PR #666, opened 2020-07-17
MAYA-105732 Activate displayColors on mesh only if one imported colorSets is displayable

JGamache-autodesk:
Only turn on the displayColors attribute on the mesh if the loaded color set looks like a color.

This means that loading the Kitchen_set scene will still result in having these colorSets: 
    float[] primvars:__faceindex = [...]
    float3[] primvars:PreMenvPosingRefPose = [...]
But the will not cause the objects to be shown as black in the viewport.

### DIFF
--- a/lib/mayaUsd/fileio/utils/meshReadUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshReadUtils.cpp
@@ -49,6 +49,9 @@
 #include <mayaUsd/utils/colorSpace.h>
 #include <mayaUsd/utils/util.h>
 
+#include <algorithm>
+#include <cctype>
+
 PXR_NAMESPACE_OPEN_SCOPE
 
 TF_DEFINE_PUBLIC_TOKENS(UsdMayaMeshColorSetTokens,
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/fileio/utils/meshReadUtils.cpp
```diff
@@ -657,6 +660,38 @@ UsdMayaMeshReadUtils::assignPrimvarsToMesh(const UsdGeomMesh& mesh,
...
+            if (typeName == SdfValueTypeNames->FloatArray
+                && name != UsdMayaMeshColorSetTokens->DisplayOpacityColorSetName) {
+                // Could be anything, forcing us to guess...
+                std::string lowerCaseName(name.GetString());
+                std::transform(
+                    lowerCaseName.begin(),
+                    lowerCaseName.end(),
+                    lowerCaseName.begin(),
+                    [](unsigned char c) { return std::tolower(c); });
+                if (lowerCaseName.find("color") == std::string::npos
```
JGamache-autodesk, line 672: That is an incredibly fragile test. Any suggestion for better filtering would be appreciated.

JGamache-autodesk:
OK. Changed the implementation to only focus on the incorrect display of those meshes with custom data that are not colors.

PR #672, opened 2020-07-21
MAYA-105743 Use the correct type when creating a new group

fowlertADSK:
Made a typo in my group command change.  Problem was that I used the variable for the name of the new object when defining the prim type.  Fixed that and added to the group test to make sure it's correct.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoAddNewPrimCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoAddNewPrimCommand.cpp
@@ -64,7 +64,7 @@ UsdUndoAddNewPrimCommand::UsdUndoAddNewPrimCommand(const UsdSceneItem::Ptr& usdS
 
         // The type of prim we were asked to create.
         // Note: "Def" means create typeless prim.
-        _primToken = (type.empty() || type == "Def") ? TfToken() : TfToken(name);
+        _primToken = (type.empty() || type == "Def") ? TfToken() : TfToken(type);
     }
 }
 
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/UsdUndoAddNewPrimCommand.cpp
```diff
@@ -64,7 +64,7 @@ UsdUndoAddNewPrimCommand::UsdUndoAddNewPrimCommand(const UsdSceneItem::Ptr& usdS
 
         // The type of prim we were asked to create.
         // Note: "Def" means create typeless prim.
-        _primToken = (type.empty() || type == "Def") ? TfToken() : TfToken(name);
+        _primToken = (type.empty() || type == "Def") ? TfToken() : TfToken(type);
```
fowlertADSK, line 67: No idea how this got left as "name" seeing as this problem was the first thing I discovered when implementing the new command.

test/lib/ufe/testGroupCmd.py
```diff
@@ -92,6 +92,12 @@ def testUsdGroup(self):
         # convention in Maya.
         newGroupPath = parentPath + ufe.PathComponent("newGroup1")
 
+        # Make sure the new group item has the correct Usd type
+        newGroupItem = ufe.Hierarchy.createItem(newGroupPath)
+        newGroupPrim = usdUtils.getPrimFromSceneItem(newGroupItem)
+        newGroupType = newGroupPrim.GetTypeName()
+        self.assertEqual(newGroupType, 'Xform')
```
fowlertADSK, line 99: Make sure the new group is an Xform.  Our new group command always uses the Xform type, the context op's code in the Outliner is the one that can pass a number of different types.

PR #661, opened 2020-07-15
MAYA-105368: set the edittarget to default into RootLayer instead of SessionLayer

HamedSabri-adsk:

PR #673, opened 2020-07-21
Apply Restriction Rules for Parenting

HamedSabri-adsk:
This PR applies the same restriction rules that we currently have for "Rename" to "Re-parent" operation.


lib/mayaUsd/ufe/UsdUndoInsertChildCommand.h
```diff
@@ -64,14 +61,18 @@ class MAYAUSD_CORE_PUBLIC UsdUndoInsertChildCommand : public Ufe::UndoableComman
...
+    UsdSceneItem::Ptr _ufeSrcItem;
+    UsdSceneItem::Ptr _ufeDstItem;
+
+    SdfPath _usdSrcPath;
+    SdfPath _usdDstPath;
+
+    SdfLayerHandle _childLayer;
+    SdfLayerHandle _parentLayer;
+
+    Ufe::Path _ufeDstPath;
```
HamedSabri-adsk, line 76: Cleaned up this class according to our guidelines.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -50,46 +51,46 @@ struct MakeSharedEnabler : public T {
...
+                                                     const UsdSceneItem::Ptr& child,
+                                                     const UsdSceneItem::Ptr& /* pos */) 
+    : Ufe::UndoableCommand()
+    , _ufeSrcItem(child)
+    , _stage(child->prim().GetStage())
+    , _usdSrcPath(child->prim().GetPath())
 {
+    const auto& childPrim = child->prim();
+    const auto& parentPrim = parent->prim();
+
```
HamedSabri-adsk, line 65: cleaned up initializing class member variables.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -62,6 +63,10 @@ UsdUndoInsertChildCommand::UsdUndoInsertChildCommand(const UsdSceneItem::Ptr& pa
     const auto& childPrim = child->prim();
     const auto& parentPrim = parent->prim();
 
+    // Apply restriction rules
+    ufe::applyCommandRestriction(childPrim, "reparent");
+    ufe::applyCommandRestriction(parentPrim, "reparent");
+
```
HamedSabri-adsk, line 69: @ppt-adsk here I am applying the restriction rules very early. These rules are applied both to the child and parent prims.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -78,22 +83,13 @@ UsdUndoInsertChildCommand::UsdUndoInsertChildCommand(const UsdSceneItem::Ptr& pa
...
-        MayaUsdUtils::defPrimSpecLayer(parentPrim);
-
-    if (!_parentLayer) {
-        std::string err = TfStringPrintf("No parent prim found at %s", parentPrim.GetPath().GetString().c_str());
-        throw std::runtime_error(err.c_str());
-    }
+    _parentLayer = parentPrim.IsPseudoRoot() 
+        ? _stage->GetEditTarget().GetLayer() 
+        : MayaUsdUtils::defPrimSpecLayer(parentPrim);
 }
```
HamedSabri-adsk: Both "_childLayer" and "_parentLayer" should be now in the correct target layers, therefore they can simply take "_stage->GetEditTarget().GetLayer()".

@ppt-adsk does "_parentLayer"  still need to check for "parentPrim.IsPseudoRoot()"? I would like to understand when this happens.

test/lib/ufe/testParentCmd.py
```diff
@@ -85,6 +85,17 @@ def testParentRelative(self):
...
+        # get the USD stage
+        stage = mayaUsd.ufe.getStage(str(shapeSegment))
+
+        # check GetLayerStack behavior
+        self.assertEqual(stage.GetLayerStack()[0], stage.GetSessionLayer())
+        self.assertEqual(stage.GetEditTarget().GetLayer(), stage.GetSessionLayer())
+
+        # set the edit target to the root layer
+        stage.SetEditTarget(stage.GetRootLayer())
+        self.assertEqual(stage.GetEditTarget().GetLayer(), stage.GetRootLayer())
```
HamedSabri-adsk, line 97: @ppt-adsk We now need to make sure we are in the correct target layer before making any edits.  

Also please note that this logic would soon change once #661 is in. I will refactor this chunk of code since the same logics is repeated for many of our test cases.

test/lib/ufe/testParentCmd.py
```diff
@@ -244,57 +266,36 @@ def testParentToProxyShape(self):
...
-                self, sphereWorldListPre, matrixToList(sphereWorld), places=6)
-    
-            # Undo.
-            cmds.undo()
-    
-            shapeChildren = shapeHier.children()
-            self.assertNotIn("pSphere1", childrenNames(shapeChildren))
+            # path strings.Expect the exception happens
+            with self.assertRaises(RuntimeError):
+                cmds.parent("|mayaUsdProxy1|mayaUsdProxyShape1,/pCylinder1/pCube1/pSphere1", "|mayaUsdProxy1|mayaUsdProxyShape1")
```
HamedSabri-adsk, line 298: @ppt-adsk This test would have passed prior to applying the rules but now it fails because of third rule we have in place:

` Cannot reparent with definitions or opinions on other layers. Opinions exist in [simpleHierarchy.usda],[simpleHierarchy-session.usda]`

I am going to create a new PR to move the deleted part under `with self.assertRaises(RuntimeError):`. I really shouldn't have removed the code now that I think more about it.

test/lib/ufe/testGroupCmd.py
```diff
@@ -95,26 +100,25 @@ def testUsdGroup(self):
...
-        self.assertEqual(len(parentChildrenRedo), 34)
+        self.assertEqual(len(parentChildrenRedo), 5)
 
         childPathsRedo = set([child.path() for child in parentChildrenRedo])
         self.assertTrue(newGroupPath in childPathsRedo)
-        self.assertTrue(ball25Path not in childPathsRedo)
-        self.assertTrue(ball35Path not in childPathsRedo)
-
+        self.assertTrue(ball5Path not in childPathsRedo)
+        self.assertTrue(ball3Path not in childPathsRedo)
```
HamedSabri-adsk, line 124: Hey @fowlertADSK , prior to applying the restriction rules, one could use top_layer.ma scene for writing this test case. However, with the new rules in place, we can't really use it: 

The first error you would hit is this:

`Cannot reparent [Ball_25] defined on another layer. Please set [Assembly_room_set.usda] as the target layer to proceed`

And once you go to the correct target layer, you would hit this.

`Cannot reparent [Ball_25] with definitions or opinions on other layers. Opinions exist in [Ball.usd],[edits.usda],[Assembly_room_set.usda],[Ball.maya.usd],[Ball.shadingVariants.usda]`

Hence, creating a simpler version of ball_set scene and applying your previous group logics to it.

test/lib/ufe/test-samples/groupBalls/ballset.ma
```diff
@@ -0,0 +1,189 @@
+//Maya ASCII 2020 scene
```
fowlertADSK, line 1: Can you resave this in something like Maya 2019 or 2018 if you have it?  If not, I can grab the file and send an updated version to you (not sure if for this file it's as easy as typing in "2018" by hand).  Matt caught that in one of my recent changes and I installed 2018 so I'd have it around.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
fowlertADSK: Is this really the only change/fix here?  Other than name changes I see a couple places where we've removed nullptr checks and then this is the only real change, do I have that right?

test/lib/ufe/test-samples/groupBalls/ballset.ma
HamedSabri-adsk: Sure thing.  Yes just changing the text should do it since we are dealing with the ASCII.  I have been doing it many times before.

The important part to change is this: `requires maya "2020";` 

We have other Maya test scenes that are still 2020. I think we should change them as well ( something for another PR ). I wonder how they could load in older Maya cut since @kxl-adsk usually runs tests against 2019.

fowlertADSK: There is an option to ignore the version, it might be on by default these days?  Changing the requires line usually works, especially in simple files that we create for automated tests that don't have much in them, but it can definitely fail depending on what exactly is in there.

Something else I usually do when creating automated test files is to now save the UI state in the file.  It's usually not needed for the test, it makes the file a lot bigger, and it's something that can mess up interactive tests occasionally.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
HamedSabri-adsk: Hey @fowlertADSK I have broken my PR into 3 parts to make it easier to review:

1- c35855050166920ad0b2f6bdc223fa56cd135fae
2- 0202eecc5b6e41b3578d092ef468feadc45e42fb
3- baf159546a2d50b71cd9738103cd9624500d96e7

The 2 line changes that you pointed out are part of section 2. There are also 2 important changes in section 2 which are related to cleaning up _childLayer and _parentLayer.

Section 3 (Fix failing unit tests) is what affects your earlier changes. If I recall correctly you told me last time that you were planning to make changes to testGroupCmd. I wanted to make sure we are aligned on this.

test/lib/ufe/test-samples/groupBalls/ballset.ma
HamedSabri-adsk: @fowlertADSK please see e123bad7942ab2dd0e392e481f18786c79e9e01a

PR #675, opened 2020-07-21
MAYA-102330 Add UI to display the number of selected prims and changed variants

fowlertADSK:
1. Help menu in the Hierarchy View dialog.

2. Might not be called from anywhere, but noticed a typo in ImportData::hasVariantSelections that I fixed.

3. I think we were leaking every ItemDelegate we created.  Qt docs say that QAbstractItemView::setItemDelegate does not take ownership of the delegate you pass to it but we were calling it like it would.

4. Made the checkbox column not resizable. 

5. Main part:
Adding UI to display a couple counts as information to the user.  One is the total number of prims that are checked, and the other is the number of variants that have been modified and are checked.  This info is displayed both in the Hierarchy View and also the file dialog import options after exiting the HV.

In the file dialog import options we also added a button to reset the import data.

The model is responsible for doing all of the counting.
The model sends signals to the dialog when it has updated the counts and the dialog will update the labels that display the info.
The delegate is where the editing of the variants happens so I connect it to the model with a signal/slot so that updating a variant will trigger the model to update the count.

The file dialog options UI just needs to display the same #’s that the Hierarchy View was displaying, so I store those two #’s in the ImportData and then added a couple query flags to the command so the options UI can get at it.

### DIFF
--- a/lib/mayaUsd/fileio/importData.cpp
+++ b/lib/mayaUsd/fileio/importData.cpp
@@ -28,6 +28,8 @@ constexpr const char* kRootPrimPath = "/";
 ImportData::ImportData()
 	: fLoadSet(UsdStage::InitialLoadSet::LoadAll)
 	, fRootPrimPath(kRootPrimPath)
+	, fPrimsInScopeCount(0)
+	, fSwitchedVariantCount(0)
 {
 }
 
@@ -35,6 +37,8 @@ ImportData::ImportData(const std::string& f)
 	: fLoadSet(UsdStage::InitialLoadSet::LoadAll)
 	, fRootPrimPath(kRootPrimPath)
 	, fFilename(f)
+	, fPrimsInScopeCount(0)
+	, fSwitchedVariantCount(0)	
 {
 }
 
@@ -60,6 +64,8 @@ void ImportData::clearData()
 	fPrimVariants.clear();
 	fFilename.clear();
 	fRootPrimPath = kRootPrimPath;
+	fPrimsInScopeCount = 0;
+	fSwitchedVariantCount =0;	
 }
 
 bool ImportData::empty() const
@@ -124,7 +130,7 @@ void ImportData::setStageInitialLoadSet(UsdStage::InitialLoadSet loadSet)
 
 bool ImportData::hasVariantSelections() const
 {
-	return !(fRootVariants.empty() || fPrimVariants.empty());
+	return !(fRootVariants.empty() && fPrimVariants.empty());
 }
 
 const SdfVariantSelectionMap& ImportData::rootVariantSelections() const
@@ -155,6 +161,26 @@ void ImportData::setPrimVariantSelections(const PrimVariantSelections& vars)
 void ImportData::setPrimVariantSelections(PrimVariantSelections&& vars)
 {
 	fPrimVariants = std::move(vars);
+}
+
+void ImportData::setPrimsInScopeCount(int count)
+{
+	fPrimsInScopeCount = count;
+}
+
+void ImportData::switchedVariantCount(int count)
+{
+	fSwitchedVariantCount = count;
+}
+
+int ImportData::primsInScopeCount() const
+{
+	return fPrimsInScopeCount;
+}
+
+int ImportData::switchedVariantCount() const
+{
+	return fSwitchedVariantCount;
 }
 
 } // namespace MayaUsd
--- a/lib/mayaUsd/fileio/importData.cpp
+++ b/lib/mayaUsd/fileio/importData.cpp
@@ -168,7 +168,7 @@ void ImportData::setPrimsInScopeCount(int count)
 	fPrimsInScopeCount = count;
 }
 
-void ImportData::switchedVariantCount(int count)
+void ImportData::setSwitchedVariantCount(int count)
 {
 	fSwitchedVariantCount = count;
 }
--- a/lib/mayaUsd/fileio/importData.h
+++ b/lib/mayaUsd/fileio/importData.h
@@ -115,6 +115,11 @@ public:
 	void setPrimVariantSelections(const PrimVariantSelections& vars);
 	void setPrimVariantSelections(PrimVariantSelections&& vars);
 
+	void setPrimsInScopeCount(int count);
+	void switchedVariantCount(int count);
+	int primsInScopeCount() const;
+	int switchedVariantCount() const;
+
 private:
 	UsdStagePopulationMask		fPopMask;
 	UsdStage::InitialLoadSet	fLoadSet;
@@ -122,6 +127,9 @@ private:
 	PrimVariantSelections		fPrimVariants;
 	std::string					fRootPrimPath;
 	std::string					fFilename;
+
+	int							fPrimsInScopeCount;
+	int							fSwitchedVariantCount;
 };
 
 } // namespace MayaUsd
--- a/lib/mayaUsd/fileio/importData.h
+++ b/lib/mayaUsd/fileio/importData.h
@@ -115,10 +115,16 @@ public:
 	void setPrimVariantSelections(const PrimVariantSelections& vars);
 	void setPrimVariantSelections(PrimVariantSelections&& vars);
 
+	//@{
+	//! Set and get the number of prims to be imported and the number of prims within that scope
+	//! that had a variant changed from what is currently set in the USD file.
+	//! These values are stored here as a way of communicating choices made between the various
+	//! Import options UI, and are used for display purposes only.
 	void setPrimsInScopeCount(int count);
-	void switchedVariantCount(int count);
+	void setSwitchedVariantCount(int count);
 	int primsInScopeCount() const;
 	int switchedVariantCount() const;
+	//@}
 
 private:
 	UsdStagePopulationMask		fPopMask;
--- a/lib/usd/ui/ItemDelegate.cpp
+++ b/lib/usd/ui/ItemDelegate.cpp
@@ -118,6 +118,7 @@ void ItemDelegate::setModelData(QWidget* editor, QAbstractItemModel* model, cons
 		if (nullptr != treeItem)
 		{
 			treeItem->setVariantSelectionModified();
+			Q_EMIT variantModified();
 		}
 	}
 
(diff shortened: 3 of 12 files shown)


lib/usd/ui/USDImportDialogCmd.cpp
```diff
@@ -202,7 +221,7 @@ MStatus USDImportDialogCmd::doIt(const MArgList& args)
 			// Creating the View can pause Maya, usually only briefly but it's noticable, so we'll toggle the wait cursor to show that it's working.
 			QApplication::setOverrideCursor(QCursor(Qt::WaitCursor));
 
-			std::unique_ptr<IUSDImportView> usdImportDialog(new USDImportDialog(assetPath.asChar(), &importData, usdQtUtil, MQtUtil::mainWindow()));
+			std::unique_ptr<USDImportDialog> usdImportDialog(new USDImportDialog(assetPath.asChar(), &importData, usdQtUtil, MQtUtil::mainWindow()));
```
fowlertADSK, line 224: Changing to use the derived class so I can call the functions that are not part of the base interface.  We can probably get rid of that interface class if we wanted to, but I have not bothered here.

lib/usd/ui/USDImportDialog.ui
```diff
@@ -23,17 +23,6 @@
    <bool>false</bool>
   </property>
```
fowlertADSK, line 24: Qt Designer always seems to want to change the order that things show up here, which makes it hard to diff.  Changes made here are to add the Help menu item and then the two labels to display the counts.

lib/usd/ui/USDImportDialog.h
```diff
@@ -79,6 +86,8 @@ private Q_SLOTS:
 	std::unique_ptr<TreeModel> fTreeModel;
 	// Reference to the Proxy Model used to sort and filter the USD file hierarchy:
 	std::unique_ptr<QSortFilterProxyModel> fProxyModel;
+	// Reference to the delegate we set on the tree view:
+	std::unique_ptr<ItemDelegate> fItemDelegate;
```
fowlertADSK, line 90: Moving the delegate to a member variable fixes the memory leak and makes it easier to hook one of its signals up to the model.

lib/usd/ui/USDImportDialog.cpp
```diff
@@ -17,8 +17,9 @@
 
 #include <stdexcept>
 
+#include <maya/MGlobal.h>
```
fowlertADSK, line 20: To execute a command to show the help.

lib/usd/ui/USDImportDialog.cpp
```diff
@@ -72,15 +79,20 @@ USDImportDialog::USDImportDialog(const std::string& filename, const ImportData*
...
 	header->setStretchLastSection(true);
 	header->setDefaultAlignment(Qt::AlignLeft | Qt::AlignVCenter);
 
+	fItemDelegate = std::unique_ptr<ItemDelegate>(new ItemDelegate(fUI->treeView));
+	QObject::connect(fItemDelegate.get(), SIGNAL(variantModified()), 
+		fTreeModel.get(), SLOT(updateModifiedVariantCount()));
+
 	// Set our item delegate on the treeview so we can customize the drawing of
 	// the variant sets.
-	fUI->treeView->setItemDelegate(new ItemDelegate(fUI->treeView));
```
fowlertADSK, line 83: I think this was a memory leak.

lib/usd/ui/USDImportDialog.cpp
```diff
@@ -182,4 +195,33 @@ void USDImportDialog::onResetFileTriggered()
 	}
 }
 
+void USDImportDialog::onHierarchyViewHelpTriggered()
+{
+	MGlobal::executeCommand("showHelp \"UsdHierarchyView\"");
```
fowlertADSK, line 200: I have a separate task to update the Maya HelpTable to make this link work.

lib/usd/ui/TreeModel.cpp
```diff
@@ -73,6 +73,7 @@ void resetVariantToPrimSelection(TreeItem* variantItem)
 	}
 
 	variantItem->setData(qtVarNames, ItemDelegate::kVariantSelectionRole);
+	variantItem->resetVariantSelectionModified();
```
fowlertADSK, line 76: Already had this function to reset the variant selection, but also need to tell the TreeItem that this has happened because we use its state when counting.

plugin/adsk/scripts/mayaUsdTranslatorImport.mel
```diff
@@ -15,6 +15,20 @@
...
+
+proc mayaUsdTranslatorImport_hideInfo()
+{
+    if (`rowLayout -exists mayaUsd_NBPrimsInScopeRL`) 
+    {
+        rowLayout -edit -visible false mayaUsd_NBPrimsInScopeRL;
+        rowLayout -edit -visible false mayaUsd_NBSwitchedVariantsRL;
+
+        text -edit -label "---" mayaUsd_NBPrimsInScopeLabel;
+        text -edit -label "---" mayaUsd_NBSwitchedVariantsLabel;
```
fowlertADSK, line 27: These text labels are hidden if we don’t have current info to display, so technically we don’t have to come up with a display that indicates that the value hasn’t been set, but I do it anyways.

plugin/adsk/scripts/mayaUsdTranslatorImport.mel
```diff
@@ -54,18 +68,38 @@ global proc int mayaUsdTranslatorImport (string $parent,
 
         if (`exists usdImportDialog`)
         {
-            rowLayout -numberOfColumns 3 -cw 1 $cw1 -cat 1 "right" 0;
+            rowLayout -numberOfColumns 4 -cw 1 $cw1 -cat 1 "right" 0;
```
fowlertADSK, line 71: 4th item in the rowLayout is the new reset button.

plugin/adsk/scripts/mayaUsdTranslatorImport.mel
```diff
@@ -54,18 +68,38 @@ global proc int mayaUsdTranslatorImport (string $parent,
...
                 // Hide the button by default.
                 button -e -visible false -w 1 mayaUsdTranslator_ViewHierBtn;
+                iconTextButton -e -visible false mayaUsdTranslator_ResetBtn;
+            setParent ..;
+
+            rowLayout -numberOfColumns 3 -cw 1 $cw1 -cat 1 "right" 0 -cal 1 "right" mayaUsd_NBPrimsInScopeRL;
+                text -label "";
+                text -label $nbPrimsInScopeLabel;
+                text -label "---" mayaUsd_NBPrimsInScopeLabel;
+            setParent ..;
```
fowlertADSK, line 94: I added an empty text label at the start of these rowLayouts because it was just easier to line up with the UI above it.

lib/usd/ui/ItemDelegate.cpp
```diff
@@ -118,6 +118,7 @@ void ItemDelegate::setModelData(QWidget* editor, QAbstractItemModel* model, cons
 		if (nullptr != treeItem)
 		{
 			treeItem->setVariantSelectionModified();
+			Q_EMIT variantModified();
```
fowlertADSK, line 121: I guess I could have cast the model pointer that is passed into this function to a TreeModel and told it directly to update its count (which is what this signal will end up triggering), but I was already hooking up signal/slots to make sure all the UI gets updated correctly so also doing this with a signal made sense.

lib/mayaUsd/fileio/importData.cpp
```diff
@@ -124,7 +130,7 @@ void ImportData::setStageInitialLoadSet(UsdStage::InitialLoadSet loadSet)
 
 bool ImportData::hasVariantSelections() const
 {
-	return !(fRootVariants.empty() || fPrimVariants.empty());
+	return !(fRootVariants.empty() && fPrimVariants.empty());
```
fowlertADSK, line 133: Logic error I noticed.  Don't think it was actually called from anywhere so no existing defect for QA to test.

lib/mayaUsd/fileio/importData.h
```diff
@@ -115,13 +115,21 @@ class MAYAUSD_CORE_PUBLIC ImportData
 	void setPrimVariantSelections(const PrimVariantSelections& vars);
 	void setPrimVariantSelections(PrimVariantSelections&& vars);
 
+	void setPrimsInScopeCount(int count);
+	void switchedVariantCount(int count);
```
seando-adsk, line 119: I think this should be "setSwitchedVariantCount"

seando-adsk: Also please add documentation tags like all the other members have.

lib/usd/ui/TreeItem.h
```diff
@@ -97,6 +97,8 @@ class MAYAUSD_UI_PUBLIC TreeItem : public QStandardItem
 	//! Only valid for kVariants type.
 	void setVariantSelectionModified();
 
+	void resetVariantSelectionModified() { fVariantSelectionModified = false; }
```
seando-adsk, line 100: member doc?

lib/usd/ui/USDImportDialog.cpp
seando-adsk: Yeah I think you are right. In the docs it says that "QAbstractView does not take ownership of delegate". Thanks for catching this.

plugin/adsk/scripts/mayaUsdTranslatorImport.mel
```diff
@@ -232,3 +268,29 @@ global proc string mayaUsdTranslatorImport_AppendFromDialog(string $currentOptio
...
+    if ("" != $result)
+    {
+        int $pc = `usdImportDialog -q -primCount`;
+        int $swc = `usdImportDialog -q -switchedVariantCount`;
+
+        text -edit -label $pc mayaUsd_NBPrimsInScopeLabel;
+        text -edit -label $swc mayaUsd_NBSwitchedVariantsLabel;
+
+        rowLayout -edit -visible true mayaUsd_NBPrimsInScopeRL;
+        rowLayout -edit -visible true mayaUsd_NBSwitchedVariantsRL;
```
seando-adsk, line 285: Since you have a hideInfo() helper, I think it would be a good idea to have a setInfo() helper.

PR #676, opened 2020-07-21
Fix compilation when UFE_PREVIEW_VERSION_NUM < 2017.

marsupial:


lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -357,10 +357,11 @@ Ufe::UndoableCommand::Ptr UsdContextOps::doOpCmd(const ItemPath& itemPath)
             TF_CODING_ERROR("Wrong number of arguments");
             return nullptr;
         }
-
+#if UFE_PREVIEW_VERSION_NUM >= 2017
         // At this point we know we have 2 arguments to execute the operation.
         // itemPath[1] contains the new prim type to create.
         return UsdUndoAddNewPrimCommand::create(fItem, itemPath[1], itemPath[1]);
+#endif
```
fowlertADSK, line 364: Ok, you are the second person today that has complained about compile issues with this file and we've finally noticed what is wrong here.  This change is not correct, what we need to do is remove the #ifdef up on line 37 of this file where we conditionally #include UsdUndoAddNewPrimCommand.h.  

What happened was I created the original file and had the #ifdef's to check for version 2017.  Then Hamed cleaned it up because that file does work with any version, it did not require the version check at all (my other changes in that PR did require the 2017 ifdef though).  But what we missed was that we still had the conditional #include up there, which now fails to build on versions < 2017.

I will go clean that up and push a fix, or if you beat me to it you could always make the change yourself and update this PR.  Either way, thanks for helping out.

PR #677, opened 2020-07-21
MAYA-104839 - Prim specs disappear from Outliner after rename

seando-adsk:
#### MAYA-104839 - Prim specs disappear from Outliner after rename
* Don't send UFE delete when in path change (rename).

### DIFF
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -231,7 +231,7 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
 			}
 		}
 #if UFE_PREVIEW_VERSION_NUM >= 2015
-		else if (!prim.IsValid())
+		else if (!prim.IsValid() && !InPathChange::inPathChange())
 		{
 			auto notification = Ufe::ObjectDestroyed(ufePath);
 			Ufe::Scene::notifyObjectDelete(notification);
--- a/test/lib/ufe/testRename.py
+++ b/test/lib/ufe/testRename.py
@@ -25,6 +25,24 @@ import mayaUsd.ufe
 import unittest
 import re
 
+class TestObserver(ufe.Observer):
+    def __init__(self):
+        super(TestObserver, self).__init__()
+        self.unexpectedNotif = 0
+        self.renameNotif = 0
+
+    def __call__(self, notification):
+        if isinstance(notification, (ufe.ObjectAdd, ufe.ObjectDelete)):
+            self.unexpectedNotif += 1
+        if isinstance(notification, ufe.ObjectRename):
+            self.renameNotif += 1
+
+    def notifications(self):
+        return self.renameNotif
+
+    def receivedUnexpectedNotif(self):
+        return self.unexpectedNotif > 0
+
 class RenameTestCase(unittest.TestCase):
     '''Test renaming a UFE scene item and its ancestors.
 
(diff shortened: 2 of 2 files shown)


lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -231,7 +231,7 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
 			}
 		}
 #if UFE_PREVIEW_VERSION_NUM >= 2015
-		else if (!prim.IsValid())
+		else if (!prim.IsValid() && !InPathChange::inPathChange())
```
seando-adsk, line 234: Don't send the delete when we are in a path change. The UFE rename notif is sent [here](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp#L130)

ppt-adsk: Shame that we didn't have a failing test to report this.  Feels like we're missing code coverage.

kxl-adsk: Indeed, code coverage is high on my wish list. What stops us from adding a test now?

seando-adsk: How would we even test this?

ppt-adsk: Some kind of rename failure, I would hope.

test/lib/ufe/testRename.py
```diff
@@ -25,6 +25,24 @@
...
 
+class TestObserver(ufe.Observer):
+    def __init__(self):
+        super(TestObserver, self).__init__()
+        self.unexpectedNotif = 0
+        self.renameNotif = 0
+
+    def __call__(self, notification):
+        if isinstance(notification, (ufe.ObjectAdd, ufe.ObjectDelete)):
+            self.unexpectedNotif += 1
```
seando-adsk, line 36: These are the notifs that we should **not** get during a rename.

test/lib/ufe/testRename.py
```diff
@@ -25,6 +25,24 @@
...
+        if isinstance(notification, ufe.ObjectRename):
+            self.renameNotif += 1
```
seando-adsk, line 38: We do expect a rename notif.

test/lib/ufe/testRename.py
```diff
@@ -120,14 +138,14 @@ def testRename(self):
         self.assertTrue(stage.GetRootLayer().GetPrimAtPath("/TreeBase"))
 
         # get default prim
-        defualtPrim = stage.GetDefaultPrim()
-        self.assertEqual(defualtPrim.GetName(), 'TreeBase')
+        defaultPrim = stage.GetDefaultPrim()
```
seando-adsk, line 141: Fix spelling

test/lib/ufe/testRename.py
```diff
@@ -380,3 +398,48 @@ def testRenameSpecialCharacter(self):
...
+        ufe.Scene.addObjectDeleteObserver(ufeObs)
+        ufe.Scene.addObjectPathChangeObserver(ufeObs)
+
+        # rename
+        newName = 'pCylinder1_Renamed'
+        cmds.rename(newName)
+
+        # After the rename we should have 1 rename notif and no unexepected notifs.
+        self.assertEqual(ufeObs.notifications(), 1)
+        self.assertFalse(ufeObs.receivedUnexpectedNotif())
```
seando-adsk, line 445: Here we rename a UFE item. We should get a single UFE rename notif and no unexpected notifs (object add or delete). The test passes now (with my fix) and I ran it without my fix (above) and it failed (as it should).

PR #680, opened 2020-07-23
Fix the wrong queried value for nodeLayerStack entry.

HamedSabri-adsk:


HamedSabri-adsk:
Thanks @ppt-adsk ,  Yes definitely. I have another PR coming up soon that would simplify the logic in MayaUsdUtils::layersWithContribution(const UsdPrim& prim) by leveraging the UsdPrimCompositionQueryArc utility and will be backed an automated test.

PR #681, opened 2020-07-23
MAYA-105608 Non-UFE selection code path couldn't work with instances

HdC-adsk:
When picking instances from viewport, the non-UFE selection code path
wasn't able to select the proxy DAG path as expected because the default
VP2 selection logic cannot handle selection for non-DAG instances.

Thus in the plugin side we need to determine the selected DAG path to
avoid relying on the default VP2 selection logic.

Note: This fix is for Maya 2018 only. Besides, in the maya side we are fixing
a selection issue for transparent instances for all relevant Maya versions.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -607,7 +607,7 @@ void ProxyRenderDelegate::updateSelectionGranularity(
     }
 }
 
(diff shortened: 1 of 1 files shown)


kxl-adsk:
It is possible to build maya-usd without UFE support, so any Maya version can be affected by this bug.

HdC-adsk:
To reproduce the issue, you would need to use Maya 2018 to create a stage from a usd file which contains instanced objects.

Then when you pick the instances in the viewport, nothing will be selected.

PR #685, opened 2020-07-24
MAYA-105735 Clean up empty map1 texcoord

JGamache-autodesk:
If the imported USD mesh does not have a "map1" or a "st" texcoord
stream, then we make sure to get rid of the default created map1
texcoord on the Maya side.


kxl-adsk:
@marsupial You raised some concerns about the UV set support. Please have a look at this PR.

marsupial:
If there could be a fallback to try for 'uv' if neither 'st' or 'map1' exists, that would be great.
'uv' is the dominant form here, also exported out of Houdini (not LOPs).

JGamache-autodesk:
@marsupial Maya does always fallback on the UV stream at index 0 on the mesh. Without this fix, the default empty stream "map1" was left in place at index zero and prevented textures from showing correctly.
With the fix, a Houdini USD file with a single UV stream called "uv" will import in a Maya mesh with a single UV stream called "uv". Since this stream is at index zero, the textures will automatically use them and display correctly.

Fixes most of the issues I have seen so far with USD files not showing textures. This does not fix meshes/materials using multiple UV streams. Is this is a common workflow?

marsupial:
Ok great, sorry for the noise: from reading it looked like only _st_ & _map1_ were honored.
Right now multiple sets _for us_ aren't super-important, but there are discussions on some workflows that would make it more important in the coming months.

lib/mayaUsd/fileio/utils/meshReadUtils.cpp
```diff
@@ -608,7 +624,25 @@ UsdMayaMeshReadUtils::assignPrimvarsToMesh(const UsdGeomMesh& mesh,
 
     // GETTING PRIMVARS
     const std::vector<UsdGeomPrimvar> primvars = mesh.GetPrimvars();
-    TF_FOR_ALL(iter, primvars) 
+
+    // Maya always has a map1 UV set. We need to find out if there is any stream in the file that
+    // will use that slot. If not, the first texcoord stream to load will replace the default map1
+    // stream.
+    bool hasDefaultUVSet = false;
+    TF_FOR_ALL(iter, primvars)
```
mattyjams, line 632: We've been slowly yanking out usages of `TF_FOR_ALL` as we come across them, so these might be good ones to replace with regular range-based for loops:

```C++
for (const UsdGeomPrimvar& primvar : primvars) {
```

lib/mayaUsd/fileio/utils/meshReadUtils.cpp
```diff
@@ -58,6 +58,9 @@ TF_DEFINE_PUBLIC_TOKENS(UsdMayaMeshColorSetTokens,
 TF_DEFINE_PRIVATE_TOKENS(
     _meshTokens,
 
+    // Default UV set name in Maya
+    (map1)
```
mattyjams, line 62: It might be nice to have this token available publicly in a place that's accessible to both this file and `meshWriteUtils.cpp` so we can avoid hard-coding the string there also:
https://github.com/Autodesk/maya-usd/blob/e504341c1cbf455653d9ff7c2c3033735cfca8d2/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp#L838

PR #687, opened 2020-07-28
MAYA-105977: Update restriction error reports based on latest design.

HamedSabri-adsk:
Updated the strings for re-parent/rename restriction rules. There are currently 3 rules in place:

**1.** Check if a particular node has any specs to contribute to the final composed prim. 

`Error Message: Cannot [reparent/rename] [prim_name]. It does not make any contributions in the current layer because its specs are in an external composition arc. Please open [filename.usd] to make direct edits.`

 **2.** Check if we are in the correct edit target to make an edit.

`Error Message:  Cannot [reparent/rename] [prim_name]. It is defined on another layer. Please set [layername] as the target layer to proceed.
`

 **3.** Check if we have more than 2 layers that contribute to the final composed prim. 

`Error Message: Cannot [reparent/rename] [prim_name]. It has definitions or opinions on other layers. Opinions exist in [layer name 1], [layer name 2], [layer name 3], [layer name 4].`

### DIFF
--- a/lib/mayaUsd/ufe/private/Utils.cpp
+++ b/lib/mayaUsd/ufe/private/Utils.cpp
@@ -36,95 +36,95 @@ namespace ufe {
 
 UsdGeomXformCommonAPI convertToCompatibleCommonAPI(const UsdPrim& prim)
 {
-	// As we are using USD's XformCommonAPI which supports only the following xformOps :
-	//    ["xformOp:translate", "xformOp:translate:pivot", "xformOp:rotateXYZ", "xformOp:scale", "!invert!xformOp:translate:pivot"]
-	// We are extending the supported xform Operations with :
-	//    ["xformOp:rotateX", "xformOp:rotateY", "xformOp:rotateZ"]
-	// Where we convert these into xformOp:rotateXYZ.
+    // As we are using USD's XformCommonAPI which supports only the following xformOps :
+    //    ["xformOp:translate", "xformOp:translate:pivot", "xformOp:rotateXYZ", "xformOp:scale", "!invert!xformOp:translate:pivot"]
+    // We are extending the supported xform Operations with :
+    //    ["xformOp:rotateX", "xformOp:rotateY", "xformOp:rotateZ"]
+    // Where we convert these into xformOp:rotateXYZ.
 
-	static TfToken rotX("xformOp:rotateX");
-	static TfToken rotY("xformOp:rotateY");
-	static TfToken rotZ("xformOp:rotateZ");
-	static TfToken rotXYZ("xformOp:rotateXYZ");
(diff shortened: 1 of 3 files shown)


lib/usd/utils/util.h
```diff
@@ -53,6 +53,10 @@ namespace MayaUsdUtils {
     MAYA_USD_UTILS_PUBLIC
     bool hasSpecs(const UsdPrim&);
 
+    //! Returns the layer in composition arc where HasSpecs is set to true
+    MAYA_USD_UTILS_PUBLIC
+    std::vector<SdfLayerHandle> layerInCompositionArcsWithSpec(const UsdPrim& prim);
```
HamedSabri-adsk, line 58: @ppt-adsk added this new utility function since the new design for the first restriction rule requires the layer's name to be displayed in the message.

Please see the new string for first rule:

`Error Message: Cannot [reparent/rename] [prim_name]. It does not make any contributions in the current layer because its specs are in an external composition arc. Please open [filename.usd] to make direct edits.`

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -174,82 +181,82 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
...
+            if (!primXform.SetPivot(GfVec3f(x, y, z)))
+                throw std::runtime_error("Unable to SetPivot after conversion to CommonAPI.");
+        }
+        catch (const std::exception& e) {
+            std::string err = TfStringPrintf("Failed to set pivot for prim %s - %s",
+                path.string().c_str(), e.what());
+            UFE_LOG(err.c_str());
+            throw;
+        }
+    }
```
HamedSabri-adsk, line 259: Please ignore the tab conversion to dot changes by enabling "Hide whitespace changes" on github.

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -158,8 +166,7 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
                 layerDisplayNames.append("[" + layer->GetDisplayName() + "]" + ",");
             }
             layerDisplayNames.pop_back();
-            std::string err = TfStringPrintf("Cannot %s [%s] with definitions or opinions on other layers. "
-                                             "Opinions exist in %s",
+            std::string err = TfStringPrintf("Cannot %s [%s]. It has definitions or opinions on other layers. Opinions exist in %s",
                                              commandName.c_str(),
                                              prim.GetName().GetString().c_str(), 
                                              layerDisplayNames.c_str());
```
HamedSabri-adsk, line 172: @ppt-adsk Updating all the strings inside applyCommandRestriction routine.

PR #690, opened 2020-07-29
Test - apply clang-format to repo (no intent on merging it yet)

kxl-adsk:
We added clang-format to the repo with a configuration that TSC agreed on. We have to start exploring strategies for adopting this formatting for the entire repo, but before we do so, let's review how the repo will look. We may discover things that have to be still tweaked in the configuration.

The goal of this `draft` PR is to validate the current clang-format configuration. No intent to merge it.

One discovery we already did - this PR doesn't compile because of include directives reordering 
```
In file included from /git/usd/maya-usd-pr/lib/mayaUsd/render/px_vp20/glslProgram.cpp:22:
In file included from /git/usd/_workspace/Pixar_USD/include/pxr/imaging/glf/glew.h:32:
/git/usd/_workspace/Pixar_USD/include/GL/glew.h:85:2: error: gl.h included before glew.h
```


seando-adsk:
One thing I noticed is that it changed a for to all on a single line. Was that intentional?

`        for (const auto& attr : primAttrs) { names.push_back(attr.GetName()); }
`

kxl-adsk:
@seando-adsk In our clang-format configuration file we have `AllowShortLoopsOnASingleLine` set to true, i.e. `while (true) continue;` can be put on a single line (based on clang-format documentation).  

Now your example is an interesting one because if I test it with [clang-format configurator](https://zed0.co.uk/clang-format-configurator/) (with maya-usd clang-format loaded) it formats the code this way:
```
for (const auto& attr : primAttrs) {
    names.push_back(attr.GetName());
}
```
I don't understand where the difference comes from. Maybe I should paste more of maya-usd code to reproduce the problem.

pmolodo:
So, regarding the differences between what you see on clang-format configurator, and when you ran it locally... one unfortunate detail about clang-format I recently discovered is that it has NO guarantees for stability across different version.

That is, if you have the exact same clang-format file, and it's executed on the same code by two different versions of clang-format, you may get different output.  (Even if all options are supported by both versions.)  And unlike, say, cmake, there's no way to specify, "I want the formatting to match version XX" in the clang-format file.

See, for instance, this thread:

https://stackoverflow.com/questions/59395507/how-to-maintain-a-clang-format-file-for-different-clang-format-versions

That's a pretty big burn - and may mean we will need to eventually have server-side running of clang-format, just to enforce that the "correct" version is used.

pmolodo:
Another thing I recently discovered - clang-format has no way to enforce conversion of this:

```cpp
if (condition)
    doSomething();
```

...into this:

```cpp
if (condition) {
    doSomething();
}
```

Apparently, you clang-format considers such changes "oustide of it's scope", and you need to use `clang-tidy` to make such fixes:

http://clang.llvm.org/extra/clang-tidy/checks/readability-braces-around-statements.html

While I'm fairly ambivalent on most style changes, the exceptions are those where I feel one option or the other will encourage / reduce potential bugs... and the above situation is a classic example of that.  So I'd really like to discourage the first option.

Since adding another tool, with another config file (.clang-tidy) feels like a bit much to ask right now, I'm thinking instead that we should change `AllowShortIfStatementsOnASingleLine` to `Always`.  That way, the first example will be converted to:

```cpp
if (condition) doSomething();
```

...which at least doesn't invite a bug as soon as someone wants to add something else to the block.

It would also match some of our other options - for instance, we already have `AllowShortLoopsOnASingleLine`, `AllowShortCaseLabelsOnASingleLine`, and `AllowShortLambdasOnASingleLine` enabled, so there would be a certain symmetry if we turned on `AllowShortIfStatementsOnASingleLine` as well.  (And we might also want to consider turning on `AllowShortLoopsOnASingleLine`?)

Thoughts?

pmolodo:
Also - the discovery of the if-statement thing came up when doing a code review.  Aside from what we enforce (or want to enforce) via automated tools - is it ok to at least say that we'd like to avoid this pattern?

```cpp
if (condition)
    doSomething()
```

?

ppt-adsk:
> 
> 
> Another thing I recently discovered - clang-format has no way to enforce conversion of this:
> 
> ```c++
> if (condition)
>     doSomething();
> ```
> 
> ...into this:
> 
> ```c++
> if (condition) {
>     doSomething();
> }
> ```
> 
> Apparently, you clang-format considers such changes "oustide of it's scope", and you need to use `clang-tidy` to make such fixes:
> 
> http://clang.llvm.org/extra/clang-tidy/checks/readability-braces-around-statements.html
> 
> While I'm fairly ambivalent on most style changes, the exceptions are those where I feel one option or the other will encourage / reduce potential bugs... and the above situation is a classic example of that. So I'd really like to discourage the first option.
> 
> Since adding another tool, with another config file (.clang-tidy) feels like a bit much to ask right now, I'm thinking instead that we should change `AllowShortIfStatementsOnASingleLine` to `Always`. That way, the first example will be converted to:
> 
> ```c++
> if (condition) doSomething();
> ```
> 
> ...which at least doesn't invite a bug as soon as someone wants to add something else to the block.
> 
> It would also match some of our other options - for instance, we already have `AllowShortLoopsOnASingleLine`, `AllowShortCaseLabelsOnASingleLine`, and `AllowShortLambdasOnASingleLine` enabled, so there would be a certain symmetry if we turned on `AllowShortIfStatementsOnASingleLine` as well. (And we might also want to consider turning on `AllowShortBlocksOnASingleLine`?)
> 
> Thoughts?

I agree with you:
- We should stick to a single tool.
- Single-statement ifs should go on a single line.

mattyjams:
> Also - the discovery of the if-statement thing came up when doing a code review. Aside from what we enforce (or want to enforce) via automated tools - is it ok to at least say that we'd like to avoid this pattern?
> 
> ```c++
> if (condition)
>     doSomething()
> ```
> 
> ?

Yeah, I definitely agree about avoiding this brace-less multi-line `if` pattern.

I'm actually not a big fan of `AllowShort<Whatevers>OnASingleLine` in general, but since it sounds like we can't get great "enforcement" of the brace usage without also introducing `clang-tidy`, I'd be ok turning on `AllowShortIfStatementsOnASingleLine` at least for now.

I haven't looked at much of the rest of this PR yet, but I'm hoping to make some time to read through it in the next few days.

kxl-adsk:
There is one argument that people always bring up when discussing a single line `if` statements - the ability to set a breakpoint when debugging. I know it doesn't look as important until you actually need it. Because of this, I would prefer we explore options for recommending 
```cpp
if (condition) {
    doSomething();
}
```

We haven't yet adopted clang-format, so let's not think about clang-tidy (even if it's a very powerful tool that would allow us to modernize the code base a lot!). At the very least we can include this in our coding guidelines. It will save us the need to spend time debating it in code reviews - do you think it's worth doing?

pmolodo:
Yeah, putting it in the guidelines sounds like a good step for now.

lib/mayaUsd/commands/baseExportCommand.h
```diff
@@ -18,92 +18,90 @@
...
-    static constexpr auto kFrameSampleFlag = "fs";
-    static constexpr auto kFrameSampleFlagLong = "frameSample";
-    static constexpr auto kFrameStrideFlag = "fst";
-    static constexpr auto kFrameStrideFlagLong = "frameStride";
-    static constexpr auto kFrameRangeFlag = "fr";
-    static constexpr auto kFrameRangeFlagLong = "frameRange";
 
-    MStatus doIt(const MArgList& args) override;
-    bool  isUndoable () const override { return false; };
+    class MAYAUSD_CORE_PUBLIC MayaUSDExportCommand : public MPxCommand {
```
pmolodo, line 32: Just noticed that we have class opening-braces on the same line - is this intentional? Don't care much personally, but it seemed a bit odd to have functions on a new line, but classes on the same, so wanted to confirm this was the intent...

lib/mayaUsd/commands/baseExportCommand.cpp
```diff
@@ -16,287 +16,295 @@
...
 #include <maya/MSelectionList.h>
+#include <maya/MSyntax.h>
 
-#include <mayaUsd/fileio/jobs/jobArgs.h>
-#include <mayaUsd/fileio/utils/writeUtil.h>
+#include <vector>
 
 PXR_NAMESPACE_USING_DIRECTIVE
 
-MAYAUSD_NS_DEF {
```
pmolodo, line 35: MAYAUSD_NS_DEF isn't detected as a namespace, and so clang-format applies an indent to everything inside.

I tested using clang's `NamespaceMacros` option; unfortunately, it didn't help.  (Apparently those have a very narrow use case - namespace macros where the name is given as an arg.)

The only way we can avoid this is if we instead start using something like `PXR_NAMESPACE_OPEN_SCOPE` and `PXR_NAMESPACE_CLOSE_SCOPE` pairs, which also include the braces - ie, `MAYAUSD_NAMESPACE_OPEN_SCOPE` and `MAYAUSD_NAMESPACE_CLOSE_SCOPE`.  (We could keep the `MAYAUSD_NS_DEF` for backwards compatibility with dependent code already using it, but not use it within this repo.)

Would we rather do that, or just be ok with the `MAYAUSD_NS_DEF` contents being indented?

pmolodo (commented):
Noticed some potential issues with the current .clang-format

lib/mayaUsd/commands/baseExportCommand.cpp
kxl-adsk: This was raised by @luceric-adsk internally. The idea is to change `MAYAUSD_NS_DEF` and take the namespace keyword out of it.

lib/mayaUsd/commands/baseExportCommand.h
kxl-adsk: I believe this was inherited from `WebKit` and we never changed the default. Looking over the repo, we are pretty consistent in opening braces to be on a new line so I don't think anyone will disagree here.

PR #694, opened 2020-08-03
[mayaUsd] load Plug libraries when finding and loading in the registry

mattyjams:
This change ensures that even if a `Plug` plugin does not identify a Maya plugin to be loaded in the "mayaPlugin" field, we still instruct the `Plug` library to load its plugin, since doing so might be required to cause readers/writers/adapters/etc. to register.

This became problematic after some restructuring of the maya-usd repo such that the translator code was moved under `lib/usd/translators`, which though it is still considered "core" mayaUsd, it is built as a separate component library. Simply importing `mayaUsd` in Python does **not** cause that library to be loaded, but loading the `pxrUsd` Maya plugin does. The prim writers and readers are still identified in the `plugInfo.json`, but there is no longer a `pxrUsdTranslators` plugin on the Maya side that goes with them.

This manifested as `UsdMayaAdaptors` not working in Python unless you specifically loaded the `pxrUsd` (or `mayaUsdPlugin`) plugin beforehand. With this change, that shouldn't be necessary.

To verify this, the `loadPlugin("pxrUsd")` command was removed from the `testUsdMayaAdaptor` test to make sure that the library is loaded correctly through the registry discovery mechanism. The one gotcha here is that the undo/redo test **does** depend on the `pxrUsd` plugin being loaded because the adaptors can be used with or without undo support, and it is only available when the plugin **is** loaded. To deal with that, the undo/redo test was just moved into its own test script rather than trying to carefully control when the plugin gets loaded and having to ensure that the suite of tests run in a
specific order.


plugin/pxr/maya/lib/usdMaya/testenv/testUsdMayaAdaptorUndoRedo.py
```diff
@@ -0,0 +1,80 @@
...
+
+import unittest
+
+
+class testUsdMayaAdaptorUndoRedo(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        standalone.initialize('usd')
+
+        # We load the pxrUsd plugin here to ensure that the usdUndoHelperCmd
```
kxl-adsk, line 33: We would like to move / add tests which can work with either plugin outside of `plugins/...` and into `test/...`. I believe both testUsdMayaAdaptor and testUsdMayaAdaptorUndoRedo fall into the above category?

mattyjams: @kxl-adsk: Sounds good. I just rebased the original change as https://github.com/Autodesk/maya-usd/pull/694/commits/a95b06260fd9b9e8376f6ff3f17c5b4cedfde638, and then migrated all of the `UsdMayaAdaptor` tests to `test/lib/usd/translators`. I moved them all since it seems like they should all stay together, and the ones that need plugins loaded will work with either `pxrUsd` or `mayaUsdPlugin`.

I had to fix one thing in `fixturesUtils.py` which I mentioned in https://github.com/Autodesk/maya-usd/pull/613#issue-442143534, which is to only create aliases for the export and import commands if we're actually loading the plugin. Otherwise the tests where we're specifically trying *not* to load plugins would fail there. That fix is in https://github.com/Autodesk/maya-usd/pull/694/commits/c56253f496488e3c0ce5647646e72ebc21460b74.

kxl-adsk:
@mattyjams Thank you for trying the PF. You can for now ignore `2020 Windows python2` test failure. We have been seeing this happening randomly. Still haven't discovered the reason.

mattyjams:
> @mattyjams Thank you for trying the PF. You can for now ignore `2020 Windows python2` test failure. We have been seeing this happening randomly. Still haven't discovered the reason.

@kxl-adsk: Cool, thanks! Yeah, thought I'd give it a shot. :)
The log made it sound like it was a permissions or storage issue with the build machine, but seems weird that it would only affect some tests (in the middle of the run, no less) and not others.

PR #698, opened 2020-08-05
MAYA-105944 - MayaUSD plugin version in plugin manager doesn't match …

seando-adsk:
#### MAYA-105944 - MayaUSD plugin version in plugin manager doesn't match the release version
* Set the plugin version from the cmake variable 'MAYAUSD_VERSION'

### DIFF
--- a/plugin/adsk/plugin/CMakeLists.txt
+++ b/plugin/adsk/plugin/CMakeLists.txt
@@ -41,6 +41,13 @@ target_compile_definitions(${TARGET_NAME}
         $<$<BOOL:${UFE_FOUND}>:WANT_UFE_BUILD>
 )
 
+if(DEFINED MAYAUSD_VERSION)
+    target_compile_definitions(${TARGET_NAME}
+        PRIVATE
+            MAYAUSD_VERSION=${MAYAUSD_VERSION}
+    )
+endif()
+
 mayaUsd_compile_config(${TARGET_NAME})
 
 # -----------------------------------------------------------------------------
--- a/plugin/adsk/plugin/plugin.cpp
+++ b/plugin/adsk/plugin/plugin.cpp
@@ -53,6 +53,13 @@
 #include <mayaUsd/ufe/Global.h>
 #endif
 
+#if defined(MAYAUSD_VERSION)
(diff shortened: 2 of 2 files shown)


plugin/adsk/plugin/CMakeLists.txt
```diff
@@ -41,6 +41,13 @@ target_compile_definitions(${TARGET_NAME}
         $<$<BOOL:${UFE_FOUND}>:WANT_UFE_BUILD>
 )
 
+if(DEFINED MAYAUSD_VERSION)
+    target_compile_definitions(${TARGET_NAME}
+        PRIVATE
+            MAYAUSD_VERSION=${MAYAUSD_VERSION}
+    )
+endif()
```
seando-adsk, line 49: This adds `-DMAYAUSD_VERSION=0.3.0` to the compile line.

plugin/adsk/plugin/plugin.cpp
```diff
@@ -53,6 +53,13 @@
 #include <mayaUsd/ufe/Global.h>
 #endif
 
+#if defined(MAYAUSD_VERSION)
+#define STRINGIFY(x) #x
+#define TOSTRING(x) STRINGIFY(x)
+#else
+#error "MAYAUSD_VERSION is not defined"
+#endif
```
seando-adsk, line 61: Macro trick to stringify a define. I cause compiler error if the MAYAUSD_VERSION is not defined.

plugin/adsk/plugin/plugin.cpp
```diff
@@ -78,7 +85,7 @@ MAYAUSD_PLUGIN_PUBLIC
 MStatus initializePlugin(MObject obj)
 {
     MStatus status;
-    MFnPlugin plugin(obj, "Autodesk", "1.0", "Any");
+    MFnPlugin plugin(obj, "Autodesk", TOSTRING(MAYAUSD_VERSION), "Any");
```
seando-adsk, line 88: This will be replaced with `"0.3.0"`

HamedSabri-adsk: @seando-adsk  Nice, makes sense. Thanks.

PR #699, opened 2020-08-05
fixed unused-but-set-variable on compiling with Maya 2018 earlier than 2018.7

kremersa:
fixed unused-but-set-variable on compiling with Maya 2018 earlier than 2018.7


kxl-adsk:
@kremersa I can't find signed CLA by you - did you send us one? Please refer to https://github.com/Autodesk/maya-usd/blob/dev/doc/CONTRIBUTING.md in case you have questions.

kremersa:
@kxl-adsk I just signed and emailed it. Thanks for checking

kxl-adsk:
We got it. Thank you.

PR #702, opened 2020-08-09
Fix windows build when compiling for Maya 2018 or without UFE

kxl-adsk:
The compiler generates following warning as an error when building on windows for Maya 2018 (or without UFE):
```
e:\maya-usd_pr\lib\mayausd\commands\editTargetCommand.h(53): error C2220: warning treated as error - no 'object' file generated
e:\maya-usd_pr\lib\mayausd\commands\editTargetCommand.h(53): warning C4251: 'MayaUsd::EditTargetCommand::_setEditTarget': class 'std::unique_ptr<MayaUsd::Impl::SetEditTarget,std::default_delete<_Ty>>' needs to have dll-interface to be used by clients of class 'MayaUsd::EditTargetCommand'
        with
        [
            _Ty=MayaUsd::Impl::SetEditTarget
        ]
e:\maya-usd_pr\lib\mayausd\commands\editTargetCommand.h(53): note: see declaration of 'std::unique_ptr<MayaUsd::Impl::SetEditTarget,std::default_delete<_Ty>>'
        with
        [
            _Ty=MayaUsd::Impl::SetEditTarget
        ]
```

There can't be any clients of _setEditTarget, so a simple solution to this problem is to export only what's needed.

### DIFF
--- a/lib/mayaUsd/commands/editTargetCommand.h
+++ b/lib/mayaUsd/commands/editTargetCommand.h
@@ -29,17 +29,29 @@ namespace Impl {
 class SetEditTarget;
 }
 
-class MAYAUSD_CORE_PUBLIC EditTargetCommand : public MPxCommand {
+class EditTargetCommand : public MPxCommand {
 public:
     // plugin registration requirements
+    MAYAUSD_CORE_PUBLIC
     static const char commandName[];
+    
+    MAYAUSD_CORE_PUBLIC
     static void*      creator();
+    
+    MAYAUSD_CORE_PUBLIC
     static MSyntax    createSyntax();
 
     // MPxCommand callbacks
+    MAYAUSD_CORE_PUBLIC
     MStatus doIt(const MArgList& argList) override;
+    
+    MAYAUSD_CORE_PUBLIC
     MStatus undoIt() override;
+    
+    MAYAUSD_CORE_PUBLIC
     MStatus redoIt() override;
+    
+    MAYAUSD_CORE_PUBLIC
     bool    isUndoable() const override;
 
 private:


lib/mayaUsd/commands/editTargetCommand.h
```diff
@@ -29,17 +29,29 @@ namespace Impl {
 class SetEditTarget;
 }
 
-class MAYAUSD_CORE_PUBLIC EditTargetCommand : public MPxCommand {
+class EditTargetCommand : public MPxCommand {
 public:
     // plugin registration requirements
+    MAYAUSD_CORE_PUBLIC
```
luceric-adsk, line 35: It seems heavy-handed and doesn't examplain why no issue why the LayerEditorCommand, which is written almost the same way.
I suggest you instead try to change the std::unique_ptr<Impl::SetEditTarget> to std::shared_ptr instead. It's probably an issue with the definition of unique_ptr in microsoft

ppt-adsk:
Changing the ownership model of the pointer doesn't sound great to me...  If it's a single owner, it should be a unique_ptr.  As per
https://stackoverflow.com/questions/16419318/one-way-of-eliminating-c4251-warning-when-using-stl-classes-in-the-dll-interface
either per-member export (which I agree is tedious) or turning off the warning with a pragma seem appropriate.

kxl-adsk:
@luceric-adsk Changing the pointer type to a shared ptr makes no difference in this case. Disabling the warning is even worst than per-member export, especially that we already have this pattern used in other places.

ppt-adsk:
> 
> 
> @luceric-adsk Changing the pointer type to a shared ptr makes no difference in this case. Disabling the warning is even worst than per-member export, especially that we already have this pattern used in other places.

To be clear, I would disable the warning in this file only.  Though per-member exports minimize the number of exports, they are also a maintenance problem when you're doing cross-platform development, as Linux and macOS developers sometimes forget to export symbols they add, and builds break (though of course with pre-flights you'll catch this).  Your call, but I'd go with disabling in the file, with per-member export being fine, of course.

luceric-adsk:
> Changing the ownership model of the pointer doesn't sound great to me... If it's a single owner, it should be a unique_ptr. As per
> https://stackoverflow.com/questions/16419318/one-way-of-eliminating-c4251-warning-when-using-stl-classes-in-the-dll-interface
> either per-member export (which I agree is tedious) or turning off the warning with a pragma seem appropriate.

Changing the type of the pointer indeed doesn't make sense, logically, but I've had to do that in the other command because the Microsoft compiler. It's probably related to the compiler generating the default move constructor and then that interacts with the unique_ptr, and would not be an issue with shared_tr.    Since I don't currently have a windows setup, I have not had the time to track down what is the proper workaround. I want to write as little code as possible and not pepper the code with explicit constructors and move operators without fully understand the source of the problem, because that brings other maintenance issues

PR #703, opened 2020-08-10
MAYA-105732 Do not auto turn on displayColors

JGamache-autodesk:
The vertex color streams are sometimes hijacked to store custom
attribute values that are not representable as colors. This is the case
in the Kitchen_Set scene with pose information that show as black when
displayColors is turned on. Let the user manually turn that feature on
if he knows for sure there are colors to display.

### DIFF
--- a/lib/mayaUsd/fileio/utils/meshReadUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshReadUtils.cpp
@@ -526,15 +526,14 @@ namespace
                     if (csRep == MFnMesh::kRGB || csRep == MFnMesh::kRGBA) 
                     {
                         meshFn.setCurrentColorSetName(colorSetName);
+                        MPlug plg = meshFn.findPlug("displayColors");
+                        if (!plg.isNull()) {
+                            plg.setBool(true);
+                        }
                     }
                     break;
                 }
             }
-
-            // It is tempting to enable "displayColors" here on the mesh to display the imported
-            // vertex colors. This would be fine if those streams were only vertex colors, but they
-            // are very frequently used to store custom per vertex values which would not display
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/fileio/utils/meshReadUtils.cpp
```diff
@@ -531,10 +531,10 @@ namespace
...
             }
 
-            MPlug plg = meshFn.findPlug("displayColors");
-            if (!plg.isNull()) {
-                plg.setBool(true);
-            }
+            // It is tempting to enable "displayColors" here on the mesh to display the imported
+            // vertex colors. This would be fine if those streams were only vertex colors, but they
+            // are very frequently used to store custom per vertex values which would not display
+            // properly. Let the user turn that on manually.
```
mattyjams, line 537: I agree that the original code here was probably incorrect. As written, if you had one RGB or RGBA color primvar on the mesh named `RandomGarbage`, we'd end up turning on the `displayColors` attribute and the resulting non-display color set would be displayed. Indeed, I don't think we want to do that.

I think the intent here was that only if the primvar was named `displayColors` (`UsdMayaMeshPrimvarTokens->DisplayColorColorSetName`) do we both make it the current color set **and** turn on the `displayColors` attribute. So given that, what would you think about moving this find-and-set-plug stuff inside the block above where we call `setCurrentColorSetName()`?

My concern is that legitimately authored assets with just a `displayColors` primvar will suddenly start coming in gray when imported as a result of this change. They'll look "incorrect" to people unless they think to dig into the mesh and turn `displayColors` on. Particularly when there's no material bound to the USD prim, the `displayColor` primvar could be really important to the display of the mesh.

PR #704, opened 2020-08-10
remove Reload() override from HdVP2Material for core USD 20.11+

mattyjams:
This virtual function was removed from `HdMaterial` in core USD commit https://github.com/PixarAnimationStudios/USD/commit/5814e68c62dbb82c366933f72566f34d35da4774

The removed override didn't actually do anything and was only invoked indirectly by HdEngine, but it was a pure virtual function, so we continue to define it for earlier versions of USD.


mattyjams:
Closing this out in favor of #717 which includes this change and handling for a few other recent core USD updates.

PR #706, opened 2020-08-11
add initial "interactive" unit test using testPxrUsdMayaGL from the Pixar plugin

mattyjams:
We've had a long-standing issue where the "interactive" (run using a full session of Maya with the UI) tests we have that exercise the Pixar batch renderer have never been run as part of the open source repo, even back when the code was in the core USD repo. With the more Maya-focused testing infrastructure now available in maya-usd, this is a first example of an interactive test, using the very simple `testPxrUsdMayaGL` test from the Pixar plugin.

A few ideas/questions for consideration:
- I'm able to run this test successfully on Linux, but I'm interested to see how it performs on other platforms and at other facilities.
- I'm interested to see if/how we can get such tests working using the existing CI setup. I'm not sure whether we're guaranteed that a windowing system will be available and/or whether launching the full Maya UI is possible on the current test infrastructure. If not, we may need to add a mechanism similar to core USD's [`PXR_HEADLESS_TEST_MODE`](https://github.com/PixarAnimationStudios/USD/blob/5746d47e36ee7b34f98089391dabb6a56568241e/cmake/defaults/Options.cmake#L26) that gates whether tests like these are included in the build.
- Internally, I'm having to use the `ADDITIONAL_LD_PRELOAD` that I included similar to how others are using `ADDITIONAL_LD_LIBRARY_PATH`, and others may need to do this as well. Maya appears to ship with a very old version of GLEW that has proven to be incompatible with Hydra which expects support for OpenGL 4.5. Animal Logic has some documentation that also describes the issue here: https://github.com/Autodesk/maya-usd/blob/1c48ce453961448b2916248322449b3737437f87/plugin/al/docs/build.md#L7


cmake/test.cmake
```diff
@@ -97,6 +113,13 @@ main(module=${MODULE_NAME})
 ")
     elseif(PREFIX_PYTHON_COMMAND)
         set(PYTEST_CODE "${PREFIX_PYTHON_COMMAND}")
+    elseif(PREFIX_PYTHON_SCRIPT)
+        if (PREFIX_INTERACTIVE)
+            set(MEL_EXECFILE_COMMAND "python(\"execfile(\\\"${PREFIX_PYTHON_SCRIPT}\\\")\")")
```
mattyjams, line 118: This is the primary reason I opted to only support `INTERACTIVE` with `PYTHON_SCRIPT`. We quickly descend into escape sequence hell trying to make sure Python code stored in a CMake variable survives intact when filtered through the guts of CMake, through the command execution, and into Maya's `-command` option.

kxl-adsk:
@mattyjams I didn't look at the code yet, but just wanted to set the right level of expectations when it comes to our CI - none of the agents have GPUs.

mattyjams:
> @mattyjams I didn't look at the code yet, but just wanted to set the right level of expectations when it comes to our CI - none of the agents have GPUs.

@kxl-adsk: Ahh, good to know, thanks!

Sounds like we **do** need a `MAYAUSD_HEADLESS_TEST_MODE` option then. In that case, the next question there is whether it should default to `OFF` like `PXR_HEADLESS_TEST_MODE` does and headless CI configurations should be configured to turn it `ON`, or whether it should default `ON` and developers and GUI/GPU-friendly CI systems need to remember to turn it `OFF`. I'd definitely prefer the former versus the latter, but curious to hear what you all think too.

I'm assuming at some point we'll have other GUI tests like these that exercise things like selection with the VP 2.0 render delegate, or some of the mayaUsd UI components?

mattyjams:
Thanks for taking a look @seando-adsk and @HamedSabri-adsk, and for the heads up about the problem with Python 3. I'll sort that out and also add the headless mode option and update this soon.

mattyjams:
I re-worked the Python we use when launching Maya interactively in a way that *should* be supported in both Python 2 and Python 3, though admittedly I'm only currently setup to test using Python 2. That was updated here:
https://github.com/Autodesk/maya-usd/pull/706/commits/52ef07028badd005c8afed2f72b339791270fd6c#diff-05227e8d327447ea9b12228004739d5eR118.

@HamedSabri-adsk, if you could give this branch another try with a Python 3 version of Maya, that would be much appreciated. :)

And I ended up taking a different approach for headless CI configurations and am just applying an "interactive" label to all tests that use the `INTERACTIVE` option when adding the test. This should allow bypassing all of these tests by simply using the `--label-exclude interactive` option to `ctest`. That addition is in commit https://github.com/Autodesk/maya-usd/commit/8cae100beec07b7eec16af26a43af433cc484a28. I think I like this better than the CMake variable approach of `PXR_HEADLESS_TEST_MODE` since it doesn't require a CMake reconfigure. You can just re-run `ctest` with or without that option using the same build. I'm open to other names for the "interactive" label if people have a strong preference there.

Related to that, I did a pass in https://github.com/Autodesk/maya-usd/commit/d85cc0d62439f6449b260ab06bc6ff3a9e38e9c3 to make sure that we're always appending to `LABELS` rather than setting them. If we do more of this type of thing where we set labels in higher-level CMake files, we'll want to make sure that we're not stomping those when lower-level CMake files want to add others.

Finally, I also rebased this branch on the current dev.

I'll stick with just the `testPxrUsdMayaGL` test here for now to focus more on the infrastructure stuff, and I'll put up a conversion of the other `pxrUsdMayaGL` tests in a follow-up PR.

mattyjams:
Sweet! Thanks for giving it a try @HamedSabri-adsk!

test/lib/render/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_subdirectory(pxrUsdMayaGL)
```
kxl-adsk, line 1: @mattyjams The folder structure under test should match the folder structure of the source tree. I know we have currently some cleanup to do with ufe, but can you adjust this folder structure? So instead of `test/lib/render/` be `test/lib/mayaUsd/render` ?

PR #707, opened 2020-08-12
MAYA-104987 Allows converting preview surface to Maya native shaders

### LINKED ISSUE #693: Standard Surface support for USD export
Is 'Standard Surface' material export currently intended to work, or is it on the road map for support? My understanding is that Autodesk intend to retire the legacy shaders (lambert, blinn, etc ) in favour of the built in (as of 2020) Standard Surface.

My testing suggests that there is currently no support (even for displayColor inheritance from standard surface baseColor). Is that correct or have I encountered a bug? (simple repo scene attached. Export with 'Material Colors' enabled - lambert colour is passed through as displayColor, standard surface colour is ignored and displayColor defaults to 1,1,1 )

[standardSurfaceDisplayColor.zip](https://github.com/Autodesk/maya-usd/files/5010280/standardSurfaceDisplayColor.zip)

If it's still "to do", it would be great to have support for baseColor->displayColor as a first step as we are aiming to transition to the standard surface for basic viewport shading. Parity with lambert would be a big help.

thanks

JGamache-autodesk:
Also allows converting displayColor to other Maya shader nodes.
Also fixes #693 by adding export of Maya standardSurface to displayColors.


lib/usd/translators/shading/usdBlinnReader.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+/* static */
+UsdMayaShaderReader::ContextSupport
+PxrUsdTranslators_BlinnReader::CanImport(const UsdMayaJobImportArgs& importArgs)
+{
+    return importArgs.shadingConversion == UsdMayaShadingConversionTokens->blinn
+        ? ContextSupport::Supported
+        : ContextSupport::Unsupported;
+}
+
```
JGamache-autodesk, line 93: The implementation is very generic and can be found in usdMaterialReader.h and .cpp. We only need auxiliary functions to help with value conversions and required side-effects.

lib/usd/translators/shading/usdBlinnReader.cpp
```diff
@@ -0,0 +1,134 @@
...
+    MFnDependencyNode& shaderFn) const
+{
+    if (mayaAttrName == _tokens->specularColor) {
+        MFnBlinnShader blinnFn;
+        blinnFn.setObject(shaderFn.object());
+        MColor color(blinnFn.specularColor());
+        float  scale(blinnFn.specularRollOff());
+        color /= scale;
+        blinnFn.setSpecularColor(color);
+        blinnFn.setSpecularRollOff(1.0f);
```
JGamache-autodesk, line 110: On export we scale the specular color by the rollOff. So on import we must set the rollOff to 1.

lib/usd/translators/shading/usdLambertWriter.cpp
```diff
@@ -111,11 +111,16 @@ void PxrUsdTranslators_LambertWriter::WriteSpecular(const UsdTimeCode& usdTime)
...
-    // Using specular workflow, but with default black specular color.
+    // Using specular workflow, but enforced black specular color.
     shaderSchema
         .CreateInput(
             PxrMayaUsdPreviewSurfaceTokens->UseSpecularWorkflowAttrName, SdfValueTypeNames->Int)
         .Set(1, usdTime);
+
+    shaderSchema
+        .CreateInput(PxrMayaUsdPreviewSurfaceTokens->SpecularColorAttrName, SdfValueTypeNames->Color3f)
+        .Set(GfVec3f(0.0f, 0.0f, 0.0f), usdTime);
```
JGamache-autodesk, line 122: We need the value to be "set" instead of "default" to zero if we want to read it back correctly on import.

lib/usd/translators/shading/usdPhongReader.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+const TfToken& PxrUsdTranslators_PhongReader::_GetMayaNodeTypeName() const
+{
+    return UsdMayaShadingConversionTokens->phong;
+}
+
+void PxrUsdTranslators_PhongReader::_ConvertToMaya(const TfToken& mayaAttrName, VtValue& usdValue)
+    const
+{
+    if (mayaAttrName == _tokens->cosinePower && usdValue.IsHolding<float>()) {
```
JGamache-autodesk, line 101: Need to convert back roughness to cosinePower using the reverse of the formula used in the exporter.

lib/usd/translators/shading/usdStandardSurfaceWriter.cpp
```diff
@@ -140,13 +140,12 @@ void PxrUsdTranslators_StandardSurfaceWriter::Write(const UsdTimeCode& usdTime)
                 PxrMayaUsdPreviewSurfaceTokens->UseSpecularWorkflowAttrName, SdfValueTypeNames->Int)
             .Set(1, usdTime);
 
-        AuthorShaderInputFromScaledShadingNodeAttr(
+        AuthorShaderInputFromShadingNodeAttr(
```
JGamache-autodesk, line 143: The specular color is not scaled by the specular coefficient in the PBR implementation, so we should not scale here.

lib/usd/translators/shading/usdUVTextureReader.cpp
```diff
@@ -183,7 +183,7 @@ bool PxrMayaUsdUVTexture_Reader::Read(UsdMayaPrimReaderContext* context)
             // nesting of referenced assets. Use absolute path instead if USD was
             // able to resolve. A better fix will require providing an asset
             // resolver to Maya that can resolve the file correctly using the
-            // MPxAssertResolver API.
+            // MPxFileResolver API.
```
JGamache-autodesk, line 186: Typo.

lib/mayaUsd/fileio/shaderReader.h
```diff
@@ -37,6 +37,19 @@ class UsdMayaShaderReader : public UsdMayaPrimReader {
     MAYAUSD_CORE_PUBLIC
     UsdMayaShaderReader(const UsdMayaPrimReaderArgs&);
 
+    /// The level of support a reader can offer for a given context
+    ///
+    /// A basic writer that gives correct results across most contexts should
```
JGamache-autodesk, line 42: Should be reader.

lib/mayaUsd/fileio/shaderReader.h
```diff
@@ -37,6 +37,19 @@ class UsdMayaShaderReader : public UsdMayaPrimReader {
...
+    /// report `Fallback`, while a specialized writer that really shines in a
+    /// given context should report `Supported` when the context is right and
+    /// `Unsupported` if the context is not as expected.
+    enum class ContextSupport { Supported, Fallback, Unsupported };
+
+    /// This static function is expected for all shader writers and allows
```
JGamache-autodesk, line 48: readers

lib/mayaUsd/fileio/shading/shadingModeDisplayColor.cpp
```diff
@@ -304,40 +325,86 @@ DEFINE_SHADING_MODE_IMPORTER(displayColor, context)
...
+        return MObject();
+    }
+
+    MPlug outputPlug;
+#if MAYA_API_VERSION >= 20200000
+    if (shadingConversion != UsdMayaShadingConversionTokens->standardSurface) {
+#endif
+        MFnLambertShader lambertFn;
+        lambertFn.setObject(shadingObj);
+        lambertFn.setName(shaderName.c_str());
```
mattyjams, line 367: Is the `setName()` here redundant now because we've already created the node with this name above?

lib/mayaUsd/fileio/shading/shadingModeDisplayColor.cpp
```diff
@@ -304,40 +325,86 @@ DEFINE_SHADING_MODE_IMPORTER(displayColor, context)
...
+            shaderParentPath.AppendChild(TfToken(lambertFn.name().asChar()));
+        context->AddCreatedObject(lambertPath, shadingObj);
+
+        outputPlug = lambertFn.findPlug("outColor", &status);
+        CHECK_MSTATUS_AND_RETURN(status, MObject());
+#if MAYA_API_VERSION >= 20200000
+    } else {
+        MFnStandardSurfaceShader surfaceFn;
+        surfaceFn.setObject(shadingObj);
+        surfaceFn.setName(shaderName.c_str());
```
mattyjams, line 389: Same here about redundant `setName()`?

lib/usd/translators/shading/usdBlinnReader.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include <maya/MFnDependencyNode.h>
+#include <maya/MObject.h>
+#include <maya/MPlug.h>
+#include <maya/MStatus.h>
+
+#include <basePxrUsdPreviewSurface/usdPreviewSurface.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+/// Shader writer for importing UsdPreviewSurface to Maya's lambert material nodes
```
mattyjams, line 48: Update doc string for this blinn reader.

lib/usd/translators/shading/usdBlinnReader.cpp
```diff
@@ -0,0 +1,134 @@
...
+class PxrUsdTranslators_BlinnReader : public PxrUsdTranslators_LambertReader {
+    typedef PxrUsdTranslators_LambertReader _BaseClass;
```
mattyjams, line 50: `using` rather than `typedef`.

lib/usd/translators/shading/usdMaterialReader.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    context->RegisterNewMayaNode(prim.GetPath().GetString(), mayaObject);
+
+    for (const UsdShadeInput& input : shaderSchema.GetInputs()) {
+        TfToken baseName = GetMayaNameForUsdAttrName(input.GetFullName());
+        if (baseName.IsEmpty()) {
+            continue;
+        }
+        _OnReadAttribute(baseName, depFn);
```
mattyjams, line 92: Is it worth documenting on the `_OnReadAttribute()` function declaration that it gets called **before** the value from USD is written into the Maya attribute? In the derived classes' implementations of the functions, it might not be immediately obvious that you don't yet have the values from USD in there.

lib/usd/translators/shading/usdBlinnReader.cpp
mattyjams: I mentioned this in `usdMaterialReader.cpp` where `_OnReadAttribute()` gets called, but this code actually looks kind of odd until you realize that this is a "prep" step **before** we apply the values from USD, since we wouldn't have read a "rolloff" attribute, for example.

lib/usd/translators/shading/usdBlinnReader.cpp
```diff
@@ -0,0 +1,134 @@
...
+    if (mayaAttrName == _tokens->specularColor) {
+        MFnBlinnShader blinnFn;
+        blinnFn.setObject(shaderFn.object());
+        MColor color(blinnFn.specularColor());
+        float  scale(blinnFn.specularRollOff());
+        color /= scale;
+        blinnFn.setSpecularColor(color);
+        blinnFn.setSpecularRollOff(1.0f);
+    } else {
+        return _BaseClass::_OnReadAttribute(mayaAttrName, shaderFn);
```
mattyjams, line 112: No need for the `return`, since this functions returns `void`?

lib/usd/translators/shading/usdLambertReader.cpp
```diff
@@ -0,0 +1,118 @@
...
+    if (mayaAttrName == _tokens->color) {
+        MFnLambertShader lambertFn;
+        lambertFn.setObject(shaderFn.object());
+        MColor color(lambertFn.color());
+        float  scale(lambertFn.diffuseCoeff());
+        color /= scale;
+        lambertFn.setColor(color);
+        lambertFn.setDiffuseCoeff(1.0f);
+    } else {
+        return _BaseClass::_OnReadAttribute(mayaAttrName, shaderFn);
```
mattyjams, line 94: Remove `return`?

lib/usd/translators/shading/usdLambertReader.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include "usdMaterialReader.h"
+
+#include <pxr/pxr.h>
+
+#include <maya/MFnDependencyNode.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+/// Shader writer for importing UsdPreviewSurface to Maya's lambert material nodes
```
mattyjams, line 29: "reader" rather than "writer"

lib/usd/translators/shading/usdLambertReader.h
```diff
@@ -0,0 +1,50 @@
...
+class PxrUsdTranslators_LambertReader : public PxrUsdTranslators_MaterialReader {
+    typedef PxrUsdTranslators_MaterialReader _BaseClass;
```
mattyjams, line 31: `using` rather than `typedef`

lib/usd/translators/shading/usdMaterialReader.cpp
```diff
@@ -0,0 +1,136 @@
...
+    if (!shaderSchema) {
+        return false;
+    }
+
+    const TfToken&    mayaNodeTypeName = _GetMayaNodeTypeName();
+    MStatus           status;
+    MObject           mayaObject;
+    MFnDependencyNode depFn;
+    if (!(UsdMayaTranslatorUtil::CreateShaderNode(
+              MString(prim.GetName().GetText()),
```
mattyjams, line 71: Should we be using more of the prim's path in the Maya node name here to avoid collisions when the shader prims in different namespaces in USD share the same name?

lib/usd/translators/shading/usdMaterialReader.cpp
```diff
@@ -0,0 +1,136 @@
...
+              mayaNodeTypeName.GetText(),
+              UsdMayaShadingNodeType::Shader,
+              &status,
+              &mayaObject)
+          && depFn.setObject(mayaObject))) {
+        // we need to make sure assumes those types are loaded..
```
mattyjams, line 77: Same question as in the other instance of this comment. Is this an `XXX`?

lib/usd/translators/shading/usdMaterialReader.h
```diff
@@ -0,0 +1,62 @@
...
+#include <maya/MFnDependencyNode.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+class UsdMayaPrimReaderArgs;
+class TfToken;
+struct UsdMayaJobImportArgs;
+class UsdShadeShader;
+
+/// Shader writer for importing UsdPreviewSurface to Maya's native material nodes
```
mattyjams, line 34: "reader" rather than "writer"

lib/usd/translators/shading/usdPhongEReader.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <maya/MFnDependencyNode.h>
+#include <maya/MObject.h>
+#include <maya/MPlug.h>
+#include <maya/MStatus.h>
+
+#include <basePxrUsdPreviewSurface/usdPreviewSurface.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+/// Shader writer for importing UsdPreviewSurface to Maya's lambert material nodes
```
mattyjams, line 47: Update doc string for this phongE reader.

lib/usd/translators/shading/usdPhongEReader.cpp
```diff
@@ -0,0 +1,115 @@
...
+class PxrUsdTranslators_PhongEReader : public PxrUsdTranslators_LambertReader {
+    typedef PxrUsdTranslators_LambertReader _BaseClass;
```
mattyjams, line 49: `using` rather than `typedef`

lib/usd/translators/shading/usdPhongReader.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include <maya/MFnDependencyNode.h>
+#include <maya/MObject.h>
+#include <maya/MPlug.h>
+#include <maya/MStatus.h>
+
+#include <basePxrUsdPreviewSurface/usdPreviewSurface.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+/// Shader writer for importing UsdPreviewSurface to Maya's lambert material nodes
```
mattyjams, line 47: Update doc string for this phong reader.

lib/usd/translators/shading/usdPhongReader.cpp
```diff
@@ -0,0 +1,128 @@
...
+class PxrUsdTranslators_PhongReader : public PxrUsdTranslators_LambertReader {
+    typedef PxrUsdTranslators_LambertReader _BaseClass;
```
mattyjams, line 49: `using` rather than `typedef`

lib/usd/translators/shading/usdStandardSurfaceReader.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include <maya/MFnStandardSurfaceShader.h>
+#include <maya/MObject.h>
+#include <maya/MPlug.h>
+#include <maya/MStatus.h>
+
+#include <basePxrUsdPreviewSurface/usdPreviewSurface.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+/// Shader writer for importing UsdPreviewSurface to Maya's lambert material nodes
```
mattyjams, line 48: Update doc string for this standardSurface reader.

lib/usd/translators/shading/usdStandardSurfaceReader.cpp
```diff
@@ -0,0 +1,177 @@
...
+class PxrUsdTranslators_StandardSurfaceReader : public PxrUsdTranslators_MaterialReader {
+    typedef PxrUsdTranslators_MaterialReader _BaseClass;
```
mattyjams, line 50: `using` rather than `typedef`

plugin/pxr/doc/README.md
```diff
@@ -38,7 +38,8 @@ The plugin creates two commands: `usdImport` and `usdExport`, and will also regi
...
-| `-shd` | `-shadingMode` | string | `displayColor` | Enable importing of materials according to a defined shading mode. Allowed values are: `none` (extract no shading data from the USD), `displayColor` (if there are bound materials in the USD, create corresponding Lambertian shaders and bind them to the appropriate Maya geometry nodes), `pxrRis` (attempt to reconstruct a Maya shading network from (presumed) Renderman RIS shading networks in the USD) |
+| `-shd` | `-shadingMode` | string | `displayColor` | Enable importing of materials according to a defined shading mode. Allowed values are: `none` (extract no shading data from the USD), `displayColor` (if there are bound materials in the USD, create corresponding Lambertian shaders and bind them to the appropriate Maya geometry nodes), `pxrRis` (attempt to reconstruct a Maya shading network from (presumed) Renderman RIS shading networks in the USD), `useRegistry` (attempt to reconstruct a Maya shading network from (presumed) UsdPreviewSurface shading networks in the USD) |
```
mattyjams, line 41: I think the default is actually `useRegistry` now instead of `displayColor`?

lib/mayaUsd/fileio/shaderReaderRegistry.h
```diff
@@ -58,15 +58,15 @@ PXR_NAMESPACE_OPEN_SCOPE
 /// reader plugin for some Maya built-in type, you can register your own
 /// plugin for that Maya built-in type.
 struct UsdMayaShaderReaderRegistry {
+    /// Predicate function, i.e. a function that can tell the level of support
+    /// the reader function will provide for a given context.
+    typedef std::function<UsdMayaShaderReader::ContextSupport(const UsdMayaJobImportArgs&)>
+        ContextPredicateFn;
```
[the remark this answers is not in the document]
kxl-adsk: Yes, we are. Thx.

lib/mayaUsd/fileio/shaderReaderRegistry.cpp
```diff
@@ -40,51 +40,98 @@ TF_DEFINE_PRIVATE_TOKENS(
...
-        return ret;
+    _Registry::const_iterator it = _Find(usdInfoId, importArgs);
+    
+    if (it != _reg.end()) {
+        return it->second._fn;
     }
 
+    // Try adding more writers via plugin load:
     static const TfTokenVector SCOPE = { _tokens->UsdMaya, _tokens->ShaderReader };
     UsdMaya_RegistryHelper::FindAndLoadMayaPlug(SCOPE, usdInfoId);
```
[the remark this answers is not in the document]
JGamache-autodesk: It is a pattern I saw used in all of the Prim{reader|updater|writer} registries, where a double TfMapLookup is done, with a nullptr dummy used to prevent loading plugins twice. It is replicated here in expanded form, but that is due to unordered_multimap usage.

lib/mayaUsd/fileio/shading/shadingModeDisplayColor.cpp
```diff
@@ -304,40 +325,86 @@ DEFINE_SHADING_MODE_IMPORTER(displayColor, context)
...
+    // Construct the selected shader.
+    MObject           shadingObj;
+    UsdMayaTranslatorUtil::CreateShaderNode(
+              MString(shaderName.c_str()),
+              shadingConversion.GetText(),
+              UsdMayaShadingNodeType::Shader,
+              &status,
+              &shadingObj);
+    if (status != MS::kSuccess) {
+        // we need to make sure assumes those types are loaded..
```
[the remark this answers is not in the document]
JGamache-autodesk: Invalid comment. The types involved are native and always loaded.

lib/usd/translators/shading/usdMaterialReader.cpp
JGamache-autodesk: The node name will be made unique by Maya by appending or incrementing a last digit. It would be less surprising (to me) to have blueRubber become blueRubber1 than to see very long path prefixed names.

plugin/pxr/doc/README.md
JGamache-autodesk: Still displayColor on import. I have to rework the import dialog in another task and this might even involve reworking the strategy and defaults. We currently have a dropdown for the "where from", and another one for the "how", and we would like something more user-friendly.

lib/usd/translators/shading/usdMaterialReader.cpp
mattyjams: I think the concern would be that the material/shader names could be unstable across round-trips in Maya then. Within the asset itself, this might be ok, but for anything else that references the exported asset and then imparts other opinions, you run the risk of those opinions falling off if the prim names change.

JGamache-autodesk: Makes sense indeed. We would need to detect when the path prefix was already added, otherwise the material name will accumulate a path prefix at each roundtrip.
In the case where an asset is referenced multiple times with the same set of opinions, should we try to collapse its materials when possible? What would be the names? I suspect we will have to add some code to reduce the number of materials generated on import, like having only two materials generated for the two-toned tile floor of the Kitchen_set.

mattyjams: I think we've talked about it before, but naively my thought was that we might be able to translate the USD namespace hierarchy into a Maya namespaced name. For example:

UsdShade prim: `/Chair/Materials/Wood/WoodGrain`
Maya shader node: `:Chair:Materials:Wood:WoodGrain`

And then as far as collapsing, and because the Kitchen_set doesn't actually have any materials in USD, I guess you're thinking about the displayColors shading mode? I guess collapsing could be a special case for that shading mode if we really wanted that, but I'd say that we might want to move away from and eventually deprecate the displayColors shading mode altogether. If there's no material in USD, it doesn't really feel like we should invent one based on the displayColors. We could just assign the Maya geometry the initialShadingGroup and turn on the geometry's `displayColors` attribute.

For all other shading modes and when there is a material prim to work with, USD somewhat enforces that the shaders connected to a Material live under it in namespace, so I would think it should be a pretty direct translation to shading nodes in Maya on import.

JGamache-autodesk: I will have to check if Maya namespaces can be used like that. That would be nice. I have opened a ticket in our internal tracker to track this improvement, which will probably be in a follow-up PR.

lib/mayaUsd/fileio/shading/shadingModeDisplayColor.cpp
```diff
@@ -114,9 +135,9 @@ class DisplayColorShadingModeExporter : public UsdMayaShadingModeExporter
         // (of the common graycale conversion) on re-import
         // The average is compute from the Maya color as is
         VtFloatArray displayOpacityAry;
-        const float transparencyAvg = (mayaTransparency[0] +
-                                       mayaTransparency[1] +
-                                       mayaTransparency[2]) / 3.0f;
+        const float transparencyAvg = (transparency[0] +
+                                       transparency[1] +
+                                       transparency[2]) / 3.0f;
```
mattyjams, line 140: Just wanted to flag here that this particular part of the change introduced a subtle "regression" from the old behavior. This is now using the linearized `transparency` from the lambert shading node rather than the previously un-linearized `mayaTransparency`.

I opened #751 which restores the previous behavior, but as I note there, this shading mode uses this value in a somewhat bizarre way. I don't expect it would have any practical impact on any what exporting materials using the `displayColors` shading mode.

PR #715, opened 2020-08-15
MAYA-105263 Improve draw performance for dense mesh

HdC-adsk:
The face vertex indices can jump drastically and cause cache hit rate
to be low when drawing. Thus we need to sort the vertices so it can be
more cache-friendly.

After sorting, a remapping table is populated and used to prepare vertex
buffers. This remapping table is updated only when topology is changed,
so it won't happen often in common workflows.

Previously we have had another remapping code path for vertex unsharing
when there is any uniform or face-varying primvar. The commit will unify
the two remapping processes with a single remapping table, making it
easier to maintain.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
@@ -392,7 +392,8 @@ void HdVP2Mesh::Sync(
             _meshSharedData._renderingToSceneFaceVtxIds = faceVertexIndices;
 
             // Fill with sequentially increasing values, starting from 0. The
-            // new face vertex indices will be used to populate index data.
+            // new face vertex indices will be used to populate index data for
+            // unshared vertex layout.
             std::iota(newFaceVertexIndices.begin(), newFaceVertexIndices.end(), 0);
         }
         else {
--- a/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
@@ -163,9 +163,6 @@ namespace {
             break;
         case HdInterpolationUniform:
         {
-            // Unshared vertex layout is required for uniform primvars, in this case
-            // renderingToSceneFaceVtxIds is a natural sequence starting from 0,
-            // thus we can avoid looking up in the table.
             const VtIntArray& faceVertexCounts = topology.GetFaceVertexCounts();
             const size_t numFaces = faceVertexCounts.size();
             if (numFaces <= primvarData.size()) {
@@ -205,7 +202,9 @@ namespace {
         case HdInterpolationFaceVarying:
             // Unshared vertex layout is required for face-varying primvars, in
             // this case renderingToSceneFaceVtxIds is a natural sequence starting
-            // from 0, thus we can avoid looking up in the table.
+            // from 0, thus we can save a lookup into the table. If the assumption
+            // about the natural sequence is changed, we will need the lookup and
+            // remap indices.
             if (numVertices <= primvarData.size()) {
                 // If the primvar has more data than needed, we issue a warning,
                 // but don't skip the primvar update. Truncate the buffer to the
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -406,43 +380,54 @@ void HdVP2Mesh::Sync(
...
-                    topology.GetOrientation(),
-                    topology.GetFaceVertexCounts(),
-                    newFaceVtxIds,
-                    topology.GetHoleIndices(),
-                    topology.GetRefineLevel()
-                )
-            );
+            // new face vertex indices will then be used to assemble all primvar
+            // vertex buffers.
+            std::iota(newFaceVertexIndices.begin(), newFaceVertexIndices.end(), 0);
```
williamkrick, line 397: Is there anywhere we actually use the iota buffer? I saw in the algorithm that you don't actually use the buffer when you detect there is no re-ordering. Can we skip filling the buffer at all? You should add a comment here if we don't use the buffer when it goes through this code path, so that if someone changes things they know that they need to update the places using the buffer as well.

lib/mayaUsd/render/vp2RenderDelegate/mesh.h
```diff
@@ -37,23 +46,26 @@ class HdVP2RenderDelegate;
     draw data from these shared data as needed.
 */
 struct HdVP2MeshSharedData {
```
williamkrick, line 48: If this is going to be the mesh topology should we re-name the struct as well?

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -133,55 +134,58 @@ namespace {
...
+                    "requires %zu vertices, while the number of elements in "
+                    "renderingToSceneFaceVtxIds is %zu. Skipping primvar update.",
+                    rprimId.asChar(), numVertices, renderingToSceneFaceVtxIds.size());
+
+                memset(vertexBuffer, 0, sizeof(DEST_TYPE) * numVertices);
+            }
+            break;
+        case HdInterpolationUniform:
+        {
+            // Unshared vertex layout is required for uniform primvars, in this case
```
williamkrick, line 166: Can we assert here that renderingToSceneVertexIds is holding what we think it is?

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
HdC-adsk: The iota buffer is used to populate index buffer at https://github.com/Autodesk/maya-usd/blob/HdC-adsk/MAYA-105263/improve-dense-mesh-drawing-performance/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp#L862. I will add comment here to make it clear.

lib/mayaUsd/render/vp2RenderDelegate/mesh.h
HdC-adsk: I don't quite follow what you mean here. If I guess it correctly, it is the doc change that leads to your question?

There is still primvar data in the struct - i.e. _primvarSourceMap.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
HdC-adsk: Hmm... That would be a linear loop to assert its element values? I don't see big values of this assertion, maybe if we hit issue we could add back the _requiresUnsharedVertices for the verification purpose.

lib/mayaUsd/render/vp2RenderDelegate/mesh.h
williamkrick: Yes the documentation change cause the question.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
williamkrick: I want something to catch the case where someone modifies how the buffer is generated for the unshared case, but then don't realize they need to update this algorithm because the data isn't used. I don't think we're going to hit bad data and that will cause a problem, I think we'll potentially modify the code and then not understand why things don't work.

It doesn't have to be a 100% verification, we could do something like in debug builds test every 1000th value and make sure it is what we expect. Probably that'll catch the case I'm worried about.

HdC-adsk: I think this makes sense. Maybe for the moment I could add a comment to point to the other code area when changing the array data.

HdC-adsk: But on the other hand, I don't see how we would change the new face vertex ids for unsharing case...

HdC-adsk:
testParentCmd failed but it is a known issue being fixed by @ppt-adsk .

PR #718, opened 2020-08-17
MAYA-106113: Adopt to new UFE interface changes around scene composite notification support

HamedSabri-adsk:


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -439,7 +439,7 @@ void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
                 globalSelection->addObserver(_observer);
             }
 
-            Ufe::Scene::instance().addObjectAddObserver(_observer);
+            Ufe::Scene::instance().addObserver(_observer);
```
ppt-adsk, line 442: Do we need to do something in the observer now that we'll receive notifications other than object add?  I.e. dynamic cast the notification and early out if not object add?

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.cpp
```diff
@@ -47,7 +47,7 @@ void UsdTRSUndoableCommandBase<V>::initialize()
     // https://stackoverflow.com/questions/17853212/using-shared-from-this-in-templated-classes
     // for explanation of this->shared_from_this() in templated class.
     attribute().Get(&fPrevValue);
-    Ufe::Scene::instance().addObjectPathChangeObserver(this->shared_from_this());
+    Ufe::Scene::instance().addObserver(this->shared_from_this());
```
ppt-adsk, line 50: In this case the filtering for rename or reparent notification is done at lines 58-63, so we're fine.

lib/mayaUsd/ufe/Utils.h
```diff
@@ -85,7 +85,7 @@ template <class T>
 void sendNotification(const Ufe::SceneItem::Ptr& item, const Ufe::Path& previousPath)
 {
     T notification(item, previousPath);
-    Ufe::Scene::notifyObjectPathChange(notification);
+    Ufe::Scene::instance().notify(notification);
```
ppt-adsk, line 88: This class can now be used for any notification that takes an item and a path.  Probably only path change notifications will ever have this requirement, but in theory it's now a more general utility :)

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
ppt-adsk: Answered my own question: no, because the observer is doing a dynamic_cast for the appropriate notification types already, at line 234.

test/lib/ufe/testRename.py
```diff
@@ -420,14 +420,10 @@ def testRenameNotifications(self):
...
-        self.assertFalse(ufe.Scene.hasObjectPathChangeObserver(ufeObs))
-
-        # Add the UFE observers we want to test
-        ufe.Scene.addObjectAddObserver(ufeObs)
-        ufe.Scene.addObjectDeleteObserver(ufeObs)
-        ufe.Scene.addObjectPathChangeObserver(ufeObs)
+        self.assertFalse(ufe.Scene.hasObserver(ufeObs))
+
+        # Add the UFE observer we want to test
+        ufe.Scene.addObserver(ufeObs)
```
HamedSabri-adsk, line 426: @ppt-adsk  

Adopting to new UFE interface changes for testRename. Simply calling Subject's "addObserver" and "hasObserver" once.

test/lib/ufe/testObservableScene.py
```diff
@@ -0,0 +1,103 @@
...
+        ufe.Scene.notify(ufe.ObjectAdd(itemA))
+
+        # we should now have an ObjectAdd notification
+        self.checkNotifications(snObs, [1,0,0,0,0,0])
+
+        # HS 2020: I was thinking that we should be able to pass ufe.scene to NotificationGuard
+        # since this is now driving from Subject but Pybind 11 slapped me :) 
+        
+        # Composite notifications with guard.
+        # with ufe.NotificationGuard(ufe.Scene):
```
HamedSabri-adsk, line 101: @ppt-adsk  Initial attempt on writing a unit test for testing Scene Composite Notification :) 

I added all the necessary pyhton bindings using the lambda trick for notify, nbObservers, addObserver, removeObserver, hasObserver. 

So far so good until line 95, however I hit a problem when I tried to create a NotificationGuard for the Scene. 

My thinking was that I should be able to pass `ufe.scene` to `NotificationGuard` since we are now driving from Subject but Python slapped me on the hand. :)  This is the error I get:

```
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. ufe.PyUfe.NotificationGuard(arg0: ufe.PyUfe.Subject)
```

test/lib/ufe/testRename.py
seando-adsk: I'm a bit confused here. When you call addObserver() now what types of notifs are you observing for? Previously we would only get add/delete/pathchange notifs. Now what will we get?

HamedSabri-adsk: @seando-adsk 

The only notification used in this test case is always ObjectRename. We know that because that's what the test is doing ( testRenameNotifications ). See Line 35

Other notifications in this case are irrelevant. I don't think we even need to any checks for other notification types ( e.g ObjectAdd, ObjectDelete) in TestObserver ( 

Previously we were adding 3 observers to the scene (ObjectAdd, ObjectDelet, ObjectPathChange) when we only just needed one.

seando-adsk: Actually no. If you look at the original TestObserver you see that it was making sure that during the rename we only received an UFE ObjectRename, and that we did **not** receive any ObjectAdd or ObjectDelete. 

In this test we did not ask for SubtreeInvalidate observation. So if this test somehow triggered a UFE subtree invalidate notif this observer would not get that notif since it never observed on it. Has that behavior changed now and when you observe on the scene you will get all of the scene notifs?

HamedSabri-adsk: In the original test, you are adding 3 observers but when are the other scene notifications(e.g ObjectAdd, ObjectDelete) are sent?

Scene still gets all of the "scene notifs" as I am demonstrating in **testObservableScene.py**

The Scene notifications are ObjectAdd, ObjectDelete, ObjectPathChange, SubtreeInvalidate and the newly added Composition 

In the new design, Scene adds the observer **only once** since it drives directly from subject. Scene can also now `publicly calls notify`.

Sorry, if this is still confusing, we can talk internally and I can walk you through the though process.

PR #721, opened 2020-08-18
MAYA-104985 Allow multiple export converters to co-exist in useRegistry export

JGamache-autodesk:
The material exporter now supports multiple exporter classes for the same Maya dependency node type.
A new parameter was added to the export command to specify a converter on export and the registry will choose the exporter that best matches the converter found in the export options.
A test exporter was written to make sure the mechanism works, but no new functionality was added to the current exporters.


lib/mayaUsd/commands/baseListShadingModesCommand.cpp
```diff
@@ -59,6 +166,9 @@ MayaUSDListShadingModesCommand::createSyntax() {
     MSyntax syntax;
     syntax.addFlag("-ex", "-export", MSyntax::kNoArg);
     syntax.addFlag("-im", "-import", MSyntax::kNoArg);
+    syntax.addFlag("-eo", "-exportOptions", MSyntax::kString);
+    syntax.addFlag("-ea", "-exportAnnotation", MSyntax::kString);
+    syntax.addFlag("-fen", "-findExportName", MSyntax::kString);
```
JGamache-autodesk, line 171: These new options help the UI build the material export dropdown menu and moves the knowledge about the nice name and annotation closer to the source.

lib/mayaUsd/fileio/shading/shadingModeDisplayColor.cpp
```diff
@@ -233,6 +233,8 @@ TF_REGISTRY_FUNCTION_WITH_TAG(UsdMayaShadingModeExportContext, displayColor)
 {
     UsdMayaShadingModeRegistry::GetInstance().RegisterExporter(
         "displayColor",
+        "Display Colors",
+        "Exports the diffuse color of the bound shader as a displayColor primvar on the USD mesh.",
```
JGamache-autodesk, line 237: The nice name and annotations are now where the exporter is implemented.

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -119,35 +120,45 @@ class UseRegistryShadingModeExporter : public UsdMayaShadingModeExporter
...
+
+            UsdMayaShaderWriterSharedPtr shaderWriter
+                = std::dynamic_pointer_cast<UsdMayaShaderWriter>(primWriter);
 
             // Store the shader writer pointer whether we succeeded or not so
             // that we don't repeatedly attempt and fail to create it for the
             // same node.
             shaderWriterMap[nodeHandle] = shaderWriter;
 
+            shaderWriter->Write(UsdTimeCode::Default());
```
JGamache-autodesk, line 144: Make sure the shaderWriter writes only once. When returning an existing writer from the cache above, it will not write a second time.

test/lib/usd/plugin/mayaGenericWriter.cpp
```diff
@@ -0,0 +1,262 @@
...
+    _tokens,
+    // Render context:
+    (maya)
+);
+
+REGISTER_SHADING_MODE_RENDER_CONTEXT(
+    _tokens->maya,
+    "Maya Shaders",
+    "Dumps the bound shader in a Maya UsdShade network that can only be used for import. Will not "
+    "render in the Maya viewport or usdView.");
```
JGamache-autodesk, line 95: The "Will not render" is why this class is in a test module and not in the main plugin. We have no plans to support this quite hackish mode.

lib/mayaUsd/fileio/shading/shadingModeRegistry.cpp
```diff
@@ -63,7 +80,23 @@ UsdMayaShadingModeRegistry::_GetExporter(const TfToken& name)
...
+    const auto it = _exportReg.find(name);
+    return it == _exportReg.end() ? _kEmptyString : it->second._niceName;
+}
+
+const std::string&
+UsdMayaShadingModeRegistry::_GetExporterDescription(const TfToken& name) {
+    UsdMaya_RegistryHelper::LoadShadingModePlugins();
+    TfRegistryManager::GetInstance().SubscribeTo<UsdMayaShadingModeExportContext>();
+    const auto it = _exportReg.find(name);
+    return it == _exportReg.end() ? _kEmptyString : it->second._description;
```
mattyjams, line 99: Maybe refactor these three with a `_GetExporter()` that returns the new struct so we don't have almost identical code repeated three times?

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
mattyjams: I don't think we should be doing **any** authoring here. That seems like an odd side effect for a function that sounds like it just gets you the thing you need to do the authoring.

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -340,41 +347,64 @@ class UseRegistryShadingModeExporter : public UsdMayaShadingModeExporter
...
                     context.GetSurfaceShaderPlug(),
                     context);
+            TfToken entryPointName = UsdShadeTokens->surface;
+
+            // For preview we will use the universal context, and for anything else we will prefix
+            if (renderContext != UsdMayaShadingModeTokens->preview) {
+                entryPointName = TfToken(
+                    TfStringPrintf("%s:%s", renderContext.GetText(), entryPointName.GetText())
+                        .c_str());
+            }
```
mattyjams, line 365: We should **not** be doing this ourselves. This and maybe `CreateShaderOutputAndConnectMaterial()` should be refactored to use `CreateSurfaceOutput()`, `CreateDisplacementOutput()`, and `CreateVolumeOutput()` from `UsdShadeMaterial`, and we should pass the appropriate renderContext into that:
https://github.com/PixarAnimationStudios/USD/blob/5d1771d23b935740bcd5bfb985a92a6652e070c0/pxr/usd/usdShade/material.h#L324

And again, we're talking about the "universal" renderContext, so we should use that term rather than "preview" to avoid confusion.

lib/usd/pxrUsdPreviewSurface/usdPreviewSurfaceWriter.cpp
```diff
@@ -45,11 +46,18 @@
...
+PXRUSDMAYA_REGISTER_SHADER_WRITER(
     pxrUsdPreviewSurface,
     PxrMayaUsdPreviewSurface_Writer);
 
+UsdMayaShaderWriter::ContextSupport
+PxrMayaUsdPreviewSurface_Writer::CanExport(const UsdMayaJobExportArgs& exportArgs)
+{
+    return exportArgs.renderContext == UsdMayaShadingModeTokens->preview
+        ? ContextSupport::Supported
+        : ContextSupport::Unsupported;
```
mattyjams, line 58: Shouldn't this at least be fallback? If your renderContext is something other than the universal context, and no other writer offers support, would you get no shader authored for a pxrUsdPreviewSurface node in Maya?

plugin/pxr/doc/README.md
```diff
@@ -105,7 +105,8 @@ Short flag | Long flag | Type | Default | Description
...
+`-shd` | `-shadingMode` | string | `displayColor` | Set the shading schema to use. Valid values are: `none`: export no shading data to the USD, `displayColor`: unless there is a colorset named `displayColor` on a Mesh, export the diffuse color of its bound shader as `displayColor` primvar on the USD Mesh, `pxrRis`: export the authored Maya shading networks, applying the same translations applied by RenderMan for Maya to the shader types, `useRegistry`: Use a registry based to export the Maya shading network to an equivalent UsdShade network.
+`-rdc` | `-renderContext` | string | `preview` | Selects a render context for export. The default value `preview` will export to a UsdPreviewSurface shading netword. A plugin mechanism allows more render contexts to be registered. Use the `mayaUSDListShadingModes` command to explore the possible options.
```
mattyjams, line 109: typo: network

And sorry to keep harping on it, but "preview" is not a render context. I'd again advocate for "universal". We're authoring UsdPreviewSurface since we expect that to be consumable by virtually all renderers.

test/lib/usd/translators/testUsdExportCustomRenderContext.py
```diff
@@ -0,0 +1,83 @@
...
+    def tearDownClass(cls):
+        standalone.uninitialize()
+
+    def testUsdExportCustomRenderContext(self):
+        """
+        Tests a custom exporter for a render context that exists in an unloaded
+        plugin.
+        """
+        plugin_path = os.path.join(self.input_path, "..", "plugin")
+        Plug.Registry().RegisterPlugins(plugin_path)
```
mattyjams, line 51: Can we do this by setting `PXR_PLUGINPATH_NAME` in the build system rather than having the test assume it knows how files are organized above it?

lib/mayaUsd/fileio/shading/shadingModeRegistry.cpp
```diff
@@ -119,6 +152,53 @@ UsdMayaShadingModeRegistry::_ListImporters() {
     return ret;
 }
 
+REGISTER_SHADING_MODE_RENDER_CONTEXT(
+    UsdMayaShadingModeTokens->preview,
+    "USD Preview Surface",
+    "Exports the bound shader as a USD preview surface UsdShade network.");
```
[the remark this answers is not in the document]
JGamache-autodesk: I tried to merge two concepts (convertingTo and renderContext). Bad idea. Let's be explicit then:
REGISTER_SHADING_MODE_CONVERSION(
    UsdMayaShadingModeTokens->UsdPreviewSurface,  // <- What type of shading node to write 
    UsdMayaShadingModeTokens->universal, // <- renderContext to use
    "USD Preview Surface", // <- Nice name
    "Exports the bound..." // <- Tooltip
);
Would this move me out of "dicey terminology" territory?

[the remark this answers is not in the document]
mattyjams: Ah, ok. Yeah, I think using a dedicated term like "conversion" makes a lot more sense than trying to use renderContext directly.

Do you still think it's necessary to associate a renderContext with the conversion mode when registering one? I would think that the selection of a particular shader ID to convert to would specify an exporter, and that that exporter would make the appropriate choice of renderContext based on which renderers would support that shader. So in the case of specifying conversion to UsdPreviewSurface, it would be the exporter for that shader ID that specifies the renderContext as universal rather than having to specify that as a property of the conversion mode. There may be some corner cases I'm not thinking of.

And smaller note, but for the tokens used here, I'd suggest using the existing ones provided by USD:
```C++
UsdImagingTokens->UsdPreviewSurface
UsdShadeTokens->universalRenderContext
```

lib/mayaUsd/fileio/shading/shadingModeRegistry.h
```diff
@@ -69,48 +70,114 @@ class UsdMayaShadingModeRegistry : public TfWeakBase
...
+    /// most well known is the default context where UsdPreviewSurface shaders are used. This
+    /// registry allows introducing other render contexts as necessary to support other renderers
+    /// and DCC. Look in "UsdShade Material Assignment", section "Refinement 2: Material Purpose"
+    /// for more details.
+    ///
+    /// To register a render context, you need to use the REGISTER_SHADING_MODE_RENDER_CONTEXT macro
+    /// for each render contexts supported by the library. Multiple registration is supported, so
+    /// each plugin should declare once the render contexts it supports.
+    ///
+    static TfTokenVector ListRenderContexts() { return GetInstance()._ListRenderContexts(); }
```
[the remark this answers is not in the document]
JGamache-autodesk: After some searching I was able to find where I got render context and material purpose confused. I now understand the difference and have reworked the code accordingly.

lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -337,6 +337,11 @@ UsdMayaJobExportArgs::UsdMayaJobExportArgs(
                 UsdMayaJobExportArgsTokens->shadingMode,
                 UsdMayaShadingModeTokens->none,
                 UsdMayaShadingModeRegistry::ListExporters())),
+        renderContext(
+            _Token(userArgs,
+                UsdMayaJobExportArgsTokens->renderContext,
+                UsdMayaShadingModeTokens->preview,
```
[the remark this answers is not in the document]
JGamache-autodesk: The "preview" token will be completely removed from this PR. The default export will be to UsdPreviewSurface shaders in the universal context.

test/lib/usd/CMakeLists.txt
```diff
@@ -2,3 +2,4 @@ add_subdirectory(pxrUsdPreviewSurface)
 add_subdirectory(schemas)
 add_subdirectory(utils)
 add_subdirectory(translators)
+add_subdirectory(plugin)
```
mattyjams, line 5: Move to the top to keep this list sorted?

test/lib/usd/translators/CMakeLists.txt
```diff
@@ -112,3 +112,23 @@ mayaUsd_add_test(testUsdImportUVSetsFloat
...
+)
+set_property(TEST testUsdExportCustomConverter APPEND PROPERTY LABELS translators)
+
+# Test uses standardSurface, which was introduced in Maya 2020.
+if (MAYA_APP_VERSION VERSION_GREATER_EQUAL 2020) 
+    mayaUsd_add_test(testUsdMayaListShadingModesCommand
+        PYTHON_MODULE testUsdMayaListShadingModesCommand
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+        ENV
+            "MAYA_PLUG_IN_PATH=${CMAKE_CURRENT_BINARY_DIR}/../plugin"
```
mattyjams, line 131: It looks like we intentionally omit `PXR_PLUGINPATH_NAME` here to exercise plugin loading in the test. Maybe add a comment mentioning that? My first thought was that it was missing here until I read the test.

test/lib/usd/translators/testUsdExportCustomConverter.py
```diff
@@ -0,0 +1,80 @@
...
+                  options=";".join(options),
+                  typ="USD Export", pr=True, ea=True)
+        cmds.file(defaultExtensions=default_ext_setting)
+
+        # Make sure we have a Maya standardSurface material in the USD file:
+        stage = Usd.Stage.Open(usd_path)
+        material = stage.GetPrimAtPath(
+            "/pCube1/Looks/standardSurface2SG/standardSurface2")
+        shader = UsdShade.Shader(material)
+        self.assertEqual(shader.GetIdAttr().Get(), "standardSurface")
```
mattyjams, line 76: Does this test work for Maya releases before 2020? It's not registered as needing Maya 2020+ in the CMakeLists.txt, and the Maya scene looks like it requires 2020. I'm not sure whether we'd expect to see a standardSurface node in the Maya scene, or exported to USD as a result for Maya 2019?

test/lib/usd/translators/testUsdMayaListShadingModesCommand.py
```diff
@@ -0,0 +1,81 @@
...
+        self.assertTrue(len(modes(exportAnnotation=preview)) > 5)
+        self.assertEqual(modes(exportOptions=preview), preview_option)
+
+        # And we should not have this one yet
+        self.assertFalse(maya_shaders in exporters)
+
+        # Load the test plugin directly to get the "Maya shading" export. This
+        # plugin was not discoverable by USD so it was not loaded when asking
+        # for the modes the first time. We will test discoverability in the
+        # testUsdExportCustomRenderContext.py test.
```
mattyjams, line 68: I think this means to refer to testUsdExportCustomConverter.py now?

test/lib/usd/translators/testUsdMayaListShadingModesCommand.py
```diff
@@ -0,0 +1,81 @@
...
+import os
+import unittest
+
+from maya import cmds
+from maya import standalone
+
+import fixturesUtils
+
+
+class testUsdExportImportRoundtripPreviewSurface(unittest.TestCase):
```
mattyjams, line 27: Update the test class name.

test/lib/usd/translators/CMakeLists.txt
```diff
@@ -112,3 +112,23 @@ mayaUsd_add_test(testUsdImportUVSetsFloat
...
+    PYTHON_MODULE testUsdExportCustomConverter
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    ENV
+        "MAYA_PLUG_IN_PATH=${CMAKE_CURRENT_BINARY_DIR}/../plugin"
+        "PXR_PLUGINPATH_NAME=${CMAKE_CURRENT_BINARY_DIR}/../plugin"
+)
+set_property(TEST testUsdExportCustomConverter APPEND PROPERTY LABELS translators)
+
+# Test uses standardSurface, which was introduced in Maya 2020.
+if (MAYA_APP_VERSION VERSION_GREATER_EQUAL 2020)
```
mattyjams, line 126: After reading through the tests, it's looking like maybe the 2020+ check was intended to guard the `testUsdExportCustomConverter` test instead?

PR #726, opened 2020-08-19
MAYA-103701 - UFE: Maya crashes after switching variant

seando-adsk:
#### MAYA-103701 - UFE: Maya crashes when selecting a new object after switching variant

* Removed special case for add/remove reference and instead created special case for add or delete operations. In these cases we send the UFE add/delete notifs. Otherwise we fallback and send the UFE subtree invalidate notif.
* In the tests that add new prims and test delete, added a UFE observer to make sure we are getting the correct notifs.

### DIFF
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -219,34 +219,48 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
 			if (!sceneItem)
 				continue;
 
-			if (prim.IsActive())
+			// Special case when we know the operation came from either
+			// the add or delete of our UFE/USD implementation.
+			if (InAddOrDeleteOperation::inAddOrDeleteOperation())
 			{
-				if (InAddOrRemoveReference::inAddOrRemoveReference())
-				{
-#if UFE_PREVIEW_VERSION_NUM >= 2014
-					// When we are in an add or remove reference we send the
-					// UFE subtree invalidate notif instead.
-					auto notification = Ufe::SubtreeInvalidate(sceneItem);
-					Ufe::Scene::notifySubtreeInvalidate(notification);
-#endif
-				}
-				else
+				if (prim.IsActive())
 				{
 					auto notification = Ufe::ObjectAdd(sceneItem);
 					Ufe::Scene::notifyObjectAdd(notification);
 				}
+				else
+				{
+					auto notification = Ufe::ObjectPostDelete(sceneItem);
+					Ufe::Scene::notifyObjectDelete(notification);
+				}
 			}
 			else
 			{
-				auto notification = Ufe::ObjectPostDelete(sceneItem);
-				Ufe::Scene::notifyObjectDelete(notification);
+#if UFE_PREVIEW_VERSION_NUM >= 2014
+				// According to USD docs for GetResyncedPaths():
+				// - Resyncs imply entire subtree invalidation of all descendant prims and properties.
+				// So we send the UFE subtree invalidate notif.
+				auto notification = Ufe::SubtreeInvalidate(sceneItem);
+				Ufe::Scene::notifySubtreeInvalidate(notification);
+#endif
 			}
 		}
 #if UFE_PREVIEW_VERSION_NUM >= 2015
 		else if (!prim.IsValid() && !InPathChange::inPathChange())
 		{
-			auto notification = Ufe::ObjectDestroyed(ufePath);
-			Ufe::Scene::notifyObjectDelete(notification);
+			if (InAddOrDeleteOperation::inAddOrDeleteOperation())
+			{
+				auto notification = Ufe::ObjectDestroyed(ufePath);
+				Ufe::Scene::notifyObjectDelete(notification);
+			}
+			else
+			{
+#if UFE_PREVIEW_VERSION_NUM >= 2014
+				auto sceneItem = Ufe::Hierarchy::createItem(ufePath);
+				auto notification = Ufe::SubtreeInvalidate(sceneItem);
+				Ufe::Scene::notifySubtreeInvalidate(notification);
+#endif
+			}
 		}
 #endif
 	}
--- a/lib/mayaUsd/ufe/UsdContextOps.cpp
+++ b/lib/mayaUsd/ufe/UsdContextOps.cpp
@@ -35,8 +35,6 @@
 #include <mayaUsd/ufe/UsdSceneItem.h>
 #include <mayaUsd/ufe/UsdUndoAddNewPrimCommand.h>
 
-#include "private/UfeNotifGuard.h"
-
 namespace {
 
 // Ufe::ContextItem strings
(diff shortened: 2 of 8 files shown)


lib/mayaUsd/ufe/private/UfeNotifGuard.h
```diff
@@ -40,20 +40,20 @@ class InPathChange
 	static bool inGuard;
 };
 
-//! \brief Helper class to scope when we are in an add or remove reference operation.
-class InAddOrRemoveReference
+//! \brief Helper class to scope when we are in an add or delete operation.
+class InAddOrDeleteOperation
```
seando-adsk, line 44: As we discussed, removed the special case for add/remove reference and instead made a special case for add/delete of prims.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -219,34 +219,48 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
 			if (!sceneItem)
 				continue;
 
-			if (prim.IsActive())
+			// Special case when we know the operation came from either
+			// the add or delete of our UFE/USD implementation.
+			if (InAddOrDeleteOperation::inAddOrDeleteOperation())
```
seando-adsk, line 224: As we discussed, if we are in an add/delete prim operation, we'll send the appropriate add/delete UFE notif.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -219,34 +219,48 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
...
 			{
-				auto notification = Ufe::ObjectPostDelete(sceneItem);
-				Ufe::Scene::notifyObjectDelete(notification);
+#if UFE_PREVIEW_VERSION_NUM >= 2014
+				// According to USD docs for GetResyncedPaths():
+				// - Resyncs imply entire subtree invalidation of all descendant prims and properties.
+				// So we send the UFE subtree invalidate notif.
+				auto notification = Ufe::SubtreeInvalidate(sceneItem);
+				Ufe::Scene::notifySubtreeInvalidate(notification);
+#endif
```
seando-adsk, line 245: Otherwise send the UFE subtree invalidate. Question: do we need anything for UFE < 0.2.14?

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -135,15 +133,13 @@ class AddReferenceUndoableCommand : public Ufe::UndoableCommand
 
     void undo() override { 
         if (_prim.IsValid()) {
-            MayaUsd::ufe::InAddOrRemoveReference ar;
```
seando-adsk, line 138: Removed special case for add/remove reference. This will still send the UFE subtree invalidate, since it will not fall into the add/delete prim case.

test/lib/ufe/testContextOps.py
```diff
@@ -41,14 +63,14 @@ def setUp(self):
         ''' Called initially to set up the maya test environment '''
         # Load plugins
         self.assertTrue(self.pluginsLoaded)
-        
+
+        # This test requires no additional setup.
+        if self._testMethodName == 'testAddNewPrim':
```
seando-adsk, line 68: The add new prim test only needs a proxy shape (not the top layer scene).

test/lib/ufe/testContextOps.py
```diff
@@ -158,6 +180,16 @@ def shadingVariant():
 
     @unittest.skipIf(os.getenv('UFE_PREVIEW_VERSION_NUM', '0000') < '2015', 'testAddNewPrim only available in UFE preview version 0.2.15 and greater')
     def testAddNewPrim(self):
+
+        # Create a proxy shape with empty stage to start with.
+        import mayaUsd_createStageWithNewLayer
+        mayaUsd_createStageWithNewLayer.createStageWithNewLayer()
```
seando-adsk, line 186: Move here from the setup()

test/lib/ufe/testContextOps.py
```diff
@@ -158,6 +180,16 @@ def shadingVariant():
...
+
+        # Create our UFE notification observer
+        ufeObs = TestAddPrimObserver()
+        ufe.Scene.addObjectDeleteObserver(ufeObs)
+        ufe.Scene.addObjectAddObserver(ufeObs)
```
seando-adsk, line 191: Use a UFE observer helper class to make sure when we add/remove prims we get the proper UFE add/delete notifs.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -68,6 +90,11 @@ def setUp(self):
     def testDelete(self):
         '''Delete Maya and USD objects.'''
 
+        # Create our UFE notification observer
+        ufeObs = TestObserver()
+        ufe.Scene.addObjectDeleteObserver(ufeObs)
+        ufe.Scene.addObjectAddObserver(ufeObs)
```
seando-adsk, line 96: Same thing for the delete operation (which does the prim deactivate/activate)

test/lib/ufe/testDeleteCmd.py
```diff
@@ -209,10 +263,14 @@ def testDeleteArgs(self):
         self.assertNotIn(sphereShapeName, sphereChildrenNames)
         self.assertNotIn(ball35Name, propsChildrenNames)
         self.assertNotIn(ball34Name, propsChildrenNames)
-        self.assertFalse(cmds.ls("|pSphere1|pSphereShape1"))
+        self.assertFalse(cmds.objExists("|pSphere1|pSphereShape1"))
```
seando-adsk, line 266: The ls command is actually undoable. So this test wasn't really testing the undo of the cmds.delete from line 250. My new test case caught this since the notifs weren't right.

seando-adsk: Funny thing is that objExists says its also undoable (in docs), but this case works.

lib/mayaUsd/ufe/StagesSubject.cpp
ppt-adsk: No.  We only support the current Maya preview release, and it has 0.2.20.  We are supposed to go back and clean out these conditionals, but we rarely have.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -219,34 +219,48 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
...
-			auto notification = Ufe::ObjectDestroyed(ufePath);
-			Ufe::Scene::notifyObjectDelete(notification);
+			if (InAddOrDeleteOperation::inAddOrDeleteOperation())
+			{
+				auto notification = Ufe::ObjectDestroyed(ufePath);
+				Ufe::Scene::notifyObjectDelete(notification);
+			}
+			else
+			{
+#if UFE_PREVIEW_VERSION_NUM >= 2014
```
ppt-adsk, line 258: You don't need this any more, we're at 2020.

test/lib/ufe/testDeleteCmd.py
ppt-adsk: Nice catch!  Uh, objExists is undoable?  What does that mean?  It's a pure query...  Maybe if you undo objExists and it returned true, the object will become an anti-matter version of itself?  I'm struggling with this, I admit.

PR #729, opened 2020-08-20
Proof of concept for USD matrix transform op editing.

ppt-adsk:
# Editing Matrix Transform Op Proof of Concept

This branch is a proof of concept for editing USD matrix transform ops
with maya-usd.  A more detailed architectural overview is given in
lib/mayaUsd/ufe/UsdTransform3d.md .

## Relevant Requirements 

As per lib/mayaUsd/ufe/UsdTransform3d.md , the following two
requirements are the focus of this branch:
- 2 *Manipulation code will use an interface that takes care of transformation
operations management. Pipelines may choose to change this interface at runtime
to improve interop between DCCs.*  This is met by the existing customizable
Transform3dHandler system, which can create custom Transform3d interfaces.
Pipelines can leverage this mechanism to introduce studio-specific overrides
without modifying the default implementation or having to maintain and resolve
conflicts when the default implementation changes. In other words, the default
implementation and studio overrides will be leveraging a stable common
interface.
- 6 *UFE API should provide UFE-USD plugin with an interface to handle the
in-place editing of any arbitrary transformation stacks, i.e. target any
transformation op.*  The foundations for meeting this requirement is the
addition of the `Transform3dHandler::editTransform3d()` method.  As presented
in the proof of concept MatrixOp handler implementation, this allows writing
custom Transform3d interfaces targeting specific transform ops.  We are open to
feedback to identify potential other UFE API gaps when it comes to
requirement 6.

## Files of Interest

An important demonstration in this branch is the capability to
entirely replace at run-time the factory that creates Transform3d
interface objects, the Transform3dHandler.  This allows a plugin to
develop its own Transform3dHandler and its own Transform3d, without
changing maya-usd core code, and to be able to have UFE use it.

For the Autodesk plugin, we replace the default Transform3dHandler
with a new one in file plugin/adsk/plugin/plugin.cpp .  The new
Transform3dHandler sets up a chain of responsibility, with the new
UsdTransform3dMatrixOpHandler being given the existing
Transform3dHandler as an argument, so it can later delegate to it.
Please note that *this chain of responsibility software structure is
entirely optional!* The main point is that you can replace the
Transform3dHandler for your plugin with your own Transform3dHandler,
which can be a single, monolithic handler, or a chain of more granular
handlers, or any implementation of the Transform3dHandler interface
you choose.

File lib/mayaUsd/ufe/UsdTransform3dMatrixOp.cpp is where most of the
new code is.  It has the UsdTransform3dMatrixOp implementation of the
Ufe::Transform3d interface, which targets an individual matrix
transform op.  It also has the UsdTransform3dMatrixOpHandler, which
shows how to set up the Transform3dHandler chain of responsibility.
The handler is given an item for which to create a Transform3d
interface.  If it can, it creates and returns a UsdTransform3dMatrixOp
interface.  If it cannot, it passes on to the next handler in the
chain, which in the case of the Autodesk plugin is the existing
UsdTransform3dHandler.

## Use in Maya

To use this branch in Maya to edit a given matrix transform op, the
transform op's name must be given to maya-usd.  This branch provides
no real workflow around the support, so a simple environment variable
is used.  The following Python script can be used to set the transform
op's name for the selected prim to be
"xformOp:transform:relative:_1:RIG",
"xformOp:transform:relative:_2:WORLD", or
"xformOp:transform:relative:_3:DIRECTION", respectively.  Notifying of
a transform change through UFE updates the Maya manipulator.  If no
matrix transform op name is given, the first matrix transform op will
be edited.  If a non-existent matrix transform op name is given, no
editing will be done.  If no matrix transform op exists, the current
maya-usd Transform3dHandler will be used.
```Python
from __future__ import print_function
import os
import ufe

def setMatrixOp(name):
    os.environ['MAYA_USD_MATRIX_XFORM_OP_NAME'] = name
    sn = ufe.GlobalSelection.get()
    if not sn:
        print("Empty selection, cannot notify Transform3d observers.")
        return
    ufe.Transform3d.notify(sn.front().path())

setMatrixOp("xformOp:transform:relative:_1:RIG")
setMatrixOp("xformOp:transform:relative:_2:WORLD")
setMatrixOp("xformOp:transform:relative:_3:DIRECTION")
```

### DIFF
--- a/lib/mayaUsd/ufe/UsdTransform3dMatrixOp.cpp
+++ b/lib/mayaUsd/ufe/UsdTransform3dMatrixOp.cpp
@@ -96,7 +96,7 @@ public:
     void redo() { fOp.GetAttr().Set(fNewOpValue,  fTime); }
 
     UsdGeomXformOp          fOp;
-    const UsdTimeCode       fTime;
+    const UsdTimeCode       fTime; // Authoring time
     const VtValue           fPrevOpValue;
     VtValue                 fNewOpValue;
 };
@@ -112,7 +112,8 @@ public:
         const UsdTimeCode&       time
     ) : Ufe::TranslateUndoableCommand(item), UsdTRSUndoableCmdBase(op, time)
     {
-        fOpTransform = op.GetOpTransform(time);
+        // We always read from proxy shape time.
+        fOpTransform = op.GetOpTransform(getTime(item->path()));
     }
 
     void undo() override { UsdTRSUndoableCmdBase::undo(); }
@@ -142,7 +143,8 @@ public:
         const UsdTimeCode&       time
     ) : Ufe::RotateUndoableCommand(item), UsdTRSUndoableCmdBase(op, time)
     {
-        GfMatrix4d opTransform = op.GetOpTransform(time);
+        // We always read from proxy shape time.
+        GfMatrix4d opTransform = op.GetOpTransform(getTime(item->path()));
 
         // Other matrix decomposition code from AL:
         // from https://github.com/AnimalLogic/maya-usd/blob/8852bdbb1fc904ac80543cd6103489097fa00154/lib/usd/utils/MayaTransformAPI.cpp#L979-L1055
@@ -190,7 +192,8 @@ public:
         const UsdTimeCode&       time
     ) : Ufe::ScaleUndoableCommand(item), UsdTRSUndoableCmdBase(op, time)
     {
-        GfMatrix4d opTransform = op.GetOpTransform(time);
+        // We always read from proxy shape time.
+        GfMatrix4d opTransform = op.GetOpTransform(getTime(item->path()));
 
         // Other matrix decomposition code from AL:
         // from https://github.com/AnimalLogic/maya-usd/blob/8852bdbb1fc904ac80543cd6103489097fa00154/lib/usd/utils/MayaTransformAPI.cpp#L979-L1055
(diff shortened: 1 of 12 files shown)


lib/mayaUsd/ufe/UsdTransform3dMatrixOp.cpp
```diff
@@ -0,0 +1,375 @@
...
+) const
+{
+    // This method can be used to edit the 3D transform of the argument, but at
+    // time of writing this is not implemented in UsdTransform3dMatrixOp, and
+    // our UsdTransform3dBaseHandler base class does not know how to edit the
+    // argument either.  Simply delegate to the next handler in the list.
+    return _nextHandler->transform3d(item);
+}
+
+Ufe::Transform3d::Ptr UsdTransform3dMatrixOpHandler::editTransform3d(
```
ppt-adsk, line 338: This is the new handler that determines whether it can create a UsdTransform3dMatrixOp interface.  If it cannot, it delegates to the next handler.

plugin/adsk/plugin/plugin.cpp
```diff
@@ -122,6 +130,18 @@ MStatus initializePlugin(MObject obj)
...
+    // responsibility: first try the single matrix op at arbitrary position in
+    // the transform stack, then pass it off to the core maya-usd
+    // Transform3dHandler (which at time of writing converts to the USD common
+    // transform API).
+	auto& runTimeMgr = Ufe::RunTimeMgr::instance();
+	auto usdRtid = MAYAUSD_NS::ufe::getUsdRunTimeId();
+    g_Transform3dHandler = runTimeMgr.transform3dHandler(usdRtid);
+	auto matrixHandler = MAYAUSD_NS::ufe::UsdTransform3dMatrixOpHandler::create(
+		g_Transform3dHandler);
+	runTimeMgr.setTransform3dHandler(usdRtid, matrixHandler);
```
ppt-adsk, line 144: This is where we replace the Transform3dHandler with our own, for our plugin.

PR #734, opened 2020-08-23
Removing code blocks for USD 19.07

kxl-adsk:
With recent support added for USD 20.08, we are dropping support for fairly old 19.07.

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -31,7 +31,7 @@ See Pixar's official github page for instructions on how to build USD: https://g
 
 |               |      ![](images/pxr.png)          |        
 |:------------: |:---------------:                  |
-|  CommitID/Tags | release: [v19.07](https://github.com/PixarAnimationStudios/USD/releases/tag/v19.07) or [v19.11](https://github.com/PixarAnimationStudios/USD/releases/tag/v19.11) or [v20.02](https://github.com/PixarAnimationStudios/USD/releases/tag/v20.02) or [v20.05](https://github.com/PixarAnimationStudios/USD/releases/tag/v20.05) or [v20.08](https://github.com/PixarAnimationStudios/USD/releases/tag/v20.08) <br> dev: [8a13d20](https://github.com/PixarAnimationStudios/USD/commit/8a13d20062d14d486b7e130d252c4652aab0263e) |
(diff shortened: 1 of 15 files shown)


kxl-adsk:
@HamedSabri-adsk what's the string that you were looking for? I examined a few files that you have listed, but then do not have code conditionally compiled for `19.07`. Is it possible that you made a mistake and looked for `19.11`?

HamedSabri-adsk:
@kxl-adsk right, I explicitly searched for both 19.07 and 19.11. 

The places that I mentioned above are where we have 19.11.  If I am not mistaking, there are places in those files that we still implicitly use code from earlier version of 19.11.

e.g lib/mayaUsd/render/mayaToHydra/renderOverride.cpp

```
#if USD_VERSION_NUM >= 1911
#include <pxr/imaging/hd/rendererPluginRegistry.h>
#else
#include <pxr/imaging/hdx/rendererPluginRegistry.h>
PXR_NAMESPACE_OPEN_SCOPE
using HdRendererPluginRegistry = HdxRendererPluginRegistry;
PXR_NAMESPACE_CLOSE_SCOPE
#endif
```
The else condition above happens to older version of 19.11.  I was under the impression that these needs to be also removed?

kxl-adsk:
Good catch...my weekend brain missed that!

kxl-adsk:
@HamedSabri-adsk see https://github.com/Autodesk/maya-usd/pull/734/commits/4a3876851c20502b3e8664c1f7917708d63f9f88

HamedSabri-adsk:
@kxl-adsk  Thank you for the changes. 

There are couple of places in lib/mayaUsd/render/pxrUsdMayaGL/batchRenderer.cpp that can be cleaned up :

**line 880 - 888**
**line 990 - 1006**
**line 1266 -1276**

kxl-adsk:
@HamedSabri-adsk what exactly you propose to remove at these lines? For example
```
#if USD_VERSION_NUM > 1911
                "        HIT:\n"
                "            delegateId      : %s\n"
                "            objectId        : %s\n"
                "            normalizedDepth : %f\n",
                hit.delegateId.GetText(),
                hit.objectId.GetText(),
                hit.normalizedDepth);
#else
                "        HIT:\n"
                "            delegateId: %s\n"
                "            objectId  : %s\n"
                "            ndcDepth  : %f\n",
                hit.delegateId.GetText(),
                hit.objectId.GetText(),
                hit.ndcDepth);
#endif
```

Else case is required for 19.11 version.

HamedSabri-adsk:
@kxl-adsk  the else condition applies to older version of 19.11

https://github.com/Autodesk/maya-usd/blob/4a3876851c20502b3e8664c1f7917708d63f9f88/lib/mayaUsd/render/pxrUsdMayaGL/batchRenderer.cpp#L888-L896

PR #736, opened 2020-08-24
Remove unused map

HamedSabri-adsk:
Removing unused map in UsdMayaImportTranslator::reader


plugin/adsk/plugin/importTranslator.cpp
```diff
@@ -58,7 +58,6 @@ UsdMayaImportTranslator::reader(
         MPxFileTranslator::FileAccessMode  /*mode*/)
 {
     std::string fileName(file.fullName().asChar(), file.fullName().length());
-    std::map<std::string, std::string> variants;
```
HamedSabri-adsk, line 61: Hmm.... I am wondering why this was not caught by the compiler?

kxl-adsk: indeed!

PR #737, opened 2020-08-24
Eliminate any std::bind in favor of lambda

HamedSabri-adsk:
Favour lambda over std::bind to avoid the unnecessary overheads that comes with std::bind

### DIFF
--- a/lib/mayaUsd/fileio/functorPrimReader.cpp
+++ b/lib/mayaUsd/fileio/functorPrimReader.cpp
@@ -45,10 +45,9 @@ UsdMayaPrimReaderRegistry::ReaderFactoryFn
 UsdMaya_FunctorPrimReader::CreateFactory(
     UsdMayaPrimReaderRegistry::ReaderFn readerFn)
 {
-    return std::bind(
-            Create,
-            std::placeholders::_1,
-            readerFn);
+    return [=](const UsdMayaPrimReaderArgs& args) { 
+        return Create(args, readerFn); 
+    };
 }
 
 PXR_NAMESPACE_CLOSE_SCOPE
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/fileio/functorPrimWriter.cpp
```diff
@@ -107,12 +107,11 @@ UsdMayaPrimWriterRegistry::WriterFactoryFn
...
-    return std::bind(
-            Create,
-            std::placeholders::_1,
-            std::placeholders::_2,
-            std::placeholders::_3,
-            fn);
+    return [=](const MFnDependencyNode& depNodeFn, 
+               const SdfPath& usdPath, 
+               UsdMayaWriteJobContext& jobCtx) { 
+        return Create(depNodeFn, usdPath, jobCtx, fn);
```
HamedSabri-adsk, line 113: `fn` is explicitly captured by copy ( notice = )

lib/mayaUsd/fileio/functorPrimReader.cpp
```diff
@@ -45,10 +45,9 @@ UsdMayaPrimReaderRegistry::ReaderFactoryFn
 UsdMaya_FunctorPrimReader::CreateFactory(
     UsdMayaPrimReaderRegistry::ReaderFn readerFn)
 {
-    return std::bind(
-            Create,
-            std::placeholders::_1,
-            readerFn);
+    return [=](const UsdMayaPrimReaderArgs& args) { 
+        return Create(args, readerFn); 
+    };
```
HamedSabri-adsk, line 50: `readerFn` is explicitly captured by copy ( notice = )

PR #738, opened 2020-08-24
adapt to refactoring in core USD of registry handling of prim specs versus prim definitions

mattyjams:
These changes were made by the core USD team to allow some refactoring in the schema registry. They make use of API added in core USD version 20.05 so older APIs can be removed for later versions of USD. Support for older versions of USD is maintained as well.

The changes here are required to build against core USD at commit https://github.com/PixarAnimationStudios/USD/commit/0e3171a02b5b854c36d58c4ca74c7a8fa3ac2b8a or later.


mattyjams:
Rebased to resolve merge conflict in `doc/build.md` following PR #734.

PR #741, opened 2020-08-25
migrate testProxyShapeDrawLighting and testProxyShapeDrawPurpose to test/... area

mattyjams:
Following on from #706, this moves the `testProxyShapeDrawLighting` and `testProxyShapeDrawPurpose` tests from `plugin/pxr/maya/lib/pxrUsdMayaGL` to `test/lib/mayaUsd/render/pxrUsdMayaGL`.

Internally, the images generated by these tests are compared against the baseline images to verify that the results match visually. Otherwise, there are no assertions in the test scripts themselves. If/when we add a system for performing image diffs to the CMake testing infrastructure, we could add that for these tests.


mattyjams:
> @mattyjams great work! Newly added tests passed successfully on my Windows + Maya beta .
> 
> Test #5: testProxyShapeDrawLighting ....... Passed 42.78 sec
> Test #6: testProxyShapeDrawPurpose ........ Passed 16.67 sec

Great, thanks for giving it a try @HamedSabri-adsk!

PR #751, opened 2020-08-29
use the un-linearized lambert transparency to author displayOpacity when using displayColors shading mode

mattyjams:
Prior to PR #707, the un-linearized transparency value was being used to compute a transparency "average", which was then used to author `displayOpacity` on the bound gprim if it did not already have an authored `displayOpacity`, as well as a `displayOpacity` input on the shader prim. It looks like an unintended side effect of #707 was that it was made to use the linearized transparency for this purpose instead. Note the use of the linearized `transparency` instead of the previously un-linearized `mayaTransparency` here:
https://github.com/Autodesk/maya-usd/pull/707/files#diff-5fe2e05011c04068f0f7249a36061532R138

Both of these behaviors seem somewhat dubious. It doesn't really seem like the shading mode exporter should be authoring `displayOpacity` of the bound gprim, since that is more a property of the geometry than the material. It also doesn't appear as if the `displayOpacity` input being authored on the shader prim is used for anything.

Ignoring those two oddities for now, the changes here should address the "regression" introduced by #707. We can revisit the behavior of the displayColors shading mode (or better yet work towards removing it!) in follow-up commits.

### DIFF
--- a/lib/mayaUsd/fileio/shading/shadingModeDisplayColor.cpp
+++ b/lib/mayaUsd/fileio/shading/shadingModeDisplayColor.cpp
@@ -99,6 +99,13 @@ private:
 
         GfVec3f color;
         GfVec3f transparency;
+
+        // We might use the Maya shading node's transparency to author a scalar
+        // displayOpacity value on the UsdGeomGprim, as well as an input on the
+        // shader prim. How we compute that value will depend on which Maya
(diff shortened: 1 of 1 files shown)


JGamache-autodesk (approved):
OK, I see the mistake I made in previous PR. Fix looks good to me.

PR #757, opened 2020-08-31
[mtoh] Add option to schedule updates through Qt instead of Maya.

marsupial:
This PR addresses two outstanding issues:
  - The refresh on idle callback being done for every delegate separately, and incurring a bit of a penalty to get the update.
  - There's an outstanding issue in Maya where a user holding the mouse down button will block all updates....We'd like to think our users are going to be actively using their cursor and would still like visual feedback in those cases.

From offline conversations it seems 2 is in your court to solve, but this provides a work-around until then.
I does introduce Qt as a dependency into **mtoh** (which I think no-one really wants) but we need to get around the issue at hand.

**-DMTOH_QT_UPDATES=OFF** can be used to make **mtoh** use a modified version of the old code-path (using an MTimer callback, but that version moves away from `MGlobal::executeOnIdle('refresh -f')` in favor of `M3dView::scheduleRefreshAllViews()`

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt
+++ b/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt
@@ -2,6 +2,17 @@ set(TARGET_NAME mtoh)
 
 add_library(${TARGET_NAME} SHARED)
 
+
+# -----------------------------------------------------------------------------
+# options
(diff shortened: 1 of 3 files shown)


kxl-adsk:
This has been logged internally as MAYA-110426 as an improvement for Maya. I will close this PR.

PR #761, opened 2020-09-03
migrate more pxrUsdMayaGL tests from plugin/pxr/... to test/...

mattyjams:
These commits move a whole bunch more interactive tests previously in the `plugin/pxr/...` area over to the project-level `test/...` area.

All of the tests were "sanitized" to remove usage of the `pxrUsd` plugin and the `pxrUsdProxyShape` and `pxrUsdReferenceAssembly` nodes. They are now all authored in terms of the `mayaUsdPlugin` and `mayaUsdProxyShape`.

I realized that I forgot to do this sanitization in #741, so I'll put up a separate PR that cleans up `testProxyShapeDrawLighting` and `testProxyShapeDrawPurpose`.

One thing to note about `testProxyShapeDrawColorAccuracy` and `testProxyShapeDrawVisibility`: These tests previously took advantage of the feature with `pxrUsdProxyShape` to set variants directly on the proxy shape. This doesn't exist on `mayaUsdProxyShape`, since the expectation is that you would make those selections directly in USD through UFE. For those two, I added a new "wrapper" USD file that makes the shadingVariant selection and pointed the proxies at those instead of the originals.

Sorry this ended up being fairly large, but it's *mostly* just file moves. Each test was migrated in its own commit. There were touch-ups to the Python test scripts, and then probably the bulk of the actual diffs are in the Maya scenes where I had to replace `pxrUsdProxyShape` with `mayaUsdProxyShape`.


test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeDrawAndTransform.py
```diff
@@ -47,15 +46,6 @@ def setUp(self):
...
 
-    def _LoadAssemblies(self):
-        loadStartTime = cmds.timerX()
-
-        UsdMaya.LoadReferenceAssemblies()
-
-        loadElapsedTime = cmds.timerX(startTime=loadStartTime)
-
-        Tf.Status("Load Elapsed Time: %f" % loadElapsedTime)
-
```
mattyjams, line 58: No assemblies anymore, so assembly-related code was removed.

test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeDrawColors.py
```diff
@@ -21,30 +21,36 @@
...
 class testProxyShapeDrawColors(unittest.TestCase):
+
     @classmethod
     def setUpClass(cls):
         # The test USD data is authored Z-up, so make sure Maya is configured
         # that way too.
         cmds.upAxis(axis='z')
 
-        cmds.loadPlugin('pxrUsd')
+        cmds.loadPlugin('mayaUsdPlugin')
```
mattyjams, line 33: We still explicitly load `mayaUsdPlugin` here because the Maya scene doesn't have a `requires` line for it, and because we use `mayaUSDExport()` and `mayaUSDImport()` below.

test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeDrawPerformance.py
```diff
@@ -15,10 +15,6 @@
 # limitations under the License.
 #
 
-from future.utils import iteritems
```
mattyjams, line 18: I dropped this from all of the tests to remove this dependency.

test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeDrawPerformance.py
```diff
@@ -50,7 +46,8 @@ def setUpClass(cls):
     @classmethod
     def tearDownClass(cls):
         statsOutputLines = []
-        for profileScopeName, elapsedTime in iteritems(cls._profileScopeMetrics):
+        for profileScopeName in cls._profileScopeMetrics.keys():
+            elapsedTime = cls._profileScopeMetrics[profileScopeName]
```
mattyjams, line 50: Pretty straightforward tweak to remove `iteritems`.

test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeDrawPerformance.py
```diff
@@ -59,8 +56,8 @@ def tearDownClass(cls):
             }
             statsOutputLines.append(json.dumps(statsDict))
 
-        statsOutput = '\n'.join(statsOutputLines)
-        perfStatsFilePath = '%s/perfStats.raw' % cls._testDir
+        statsOutput = os.linesep.join(statsOutputLines)
+        perfStatsFilePath = os.path.join(cls._testDir, 'perfStats.raw')
```
mattyjams, line 60: A few other tweaks like these to make sure all platforms are supported.

test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeDrawUsdChangeProcessing.py
```diff
@@ -84,39 +99,6 @@ def _RunTest(self, dagPathName):
...
-        self._testName = 'UsdChangeProcessingTest_Assembly'
-
-        mayaSceneFile = '%s.ma' % self._testName
-        mayaSceneFullPath = os.path.abspath(mayaSceneFile)
-        cmds.file(mayaSceneFullPath, open=True, force=True)
-
-        UsdMaya.LoadReferenceAssemblies()
-
-        assemblyDagPathName = '|UsdChangeProcessingTest|Primitive'
-        self._RunTest(assemblyDagPathName)
```
mattyjams, line 102: The assembly case was removed, so everything was collapsed from `_RunTest()` into `testUsdChangeProcessingProxy()`.

test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeLiveSurface.py
```diff
@@ -15,35 +15,17 @@
...
-    from PySide2.QtTest import QTest
-    from PySide2.QtWidgets import QWidget
-
-    from shiboken2 import wrapInstance
-else:
-    from PySide import QtCore
-    from PySide.QtTest import QTest
-    from PySide.QtGui import QWidget
-
-    from shiboken import wrapInstance
```
mattyjams, line 42: Removed all the support for PySide, dropping Maya 2016 and earlier.

mattyjams:
One small addition here with 17f789afbf5d655ba17b392c8e74a11fd1597a96. The Pixar batch renderer actually represents the USD in the `testProxyShapeLiveSurface` test in the wrong location since the Pixar workflows never really supported directly referencing a root prim with a transformation on it. That test assumes the transformation is being ignored, so I just removed it from the USD. This ensures that it matches the VP2.0 render delegate, which does support this correctly.

HamedSabri-adsk (approved):

test/lib/mayaUsd/render/pxrUsdMayaGL/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ set(TEST_SCRIPT_FILES
     testProxyShapeDuplicatePerformance.py
     testProxyShapeLiveSurface.py
     testProxyShapeRendererSceneMessages.py
+    testProxyShapeSelectionPerformance.py
```
HamedSabri-adsk, line 17: @mattyjams 

As a FYI, I noticed that both testProxyShapeLiveSurface and testProxyShapeSelectionPerformance fail on my Windows after syncing to the latest dev changes.

@kxl-adsk  I believe you mentioned the same failure on your side.  I don't think I notice these failures when I originally reviewed this PR but they show up now. 

Requires more investigation.

```
98% tests passed, 2 tests failed out of 127

Label Time Summary:
interactive          = 847.06 sec*proc (14 tests)
pxrUsdMayaGL         = 847.06 sec*proc (14 tests)
translators          = 643.22 sec*proc (52 tests)
ufe                  = 264.30 sec*proc (21 tests)
usdPreviewSurface    =  20.88 sec*proc (2 tests)

Total Test time (real) = 212.37 sec

The following tests FAILED:
         15 - testProxyShapeLiveSurface (Failed)
         17 - testProxyShapeSelectionPerformance (Failed)
```

mattyjams: Ahh, sorry about that @HamedSabri-adsk. They've been passing for me on Linux. I briefly had access to a Windows machine before we left the studio, but I've since lost that. :/

Do you have any logs handy from failed runs of either of those tests? And I assume all of the other `pxrUsdMayaGL` tests are passing?

Just a wild guess without any more detail for the moment, but the one thing common to these two tests and absent from all the others is the shenanigans they pull dealing with PySide2 and shiboken2 so that they can emulate clicking in the viewports. Maybe there's something there that's problematic on Windows?

Let me know what I can do to help get to the bottom of that!

kxl-adsk: Indeed I saw these when validating your change @mattyjams and before merging. I was confused because @HamedSabri-adsk run the tests and they were passing. The explanation I found is that we can be using builds of Maya with asserts enabled. When we do, these tests will cause Maya to assert when the application is exiting. This then causes the test to fail. You won't get it with the official build of Maya.

The assert seems to be complaining that there are still some shared GL widgets alive when we are releasing resources.

mattyjams: Thanks for the extra info @kxl-adsk. Makes sense then why I might not be seeing those failed asserts.

I took a wild stab at addressing this in #777, so we can continue the investigation there.

PR #764, opened 2020-09-04
remove test dependencies on Python future past, and builtin.zip

mattyjams:
This removes the usage of `future`, `past`, and `builtin.zip` in tests. This simplifies the Python environment setup when attempting to run tests on currently released versions of Maya with Python 2, but should still support Python 3 as well.

Note that this is only part of the `future` removal. The other part of it takes place in #761 for the `pxrUsdMayaGL` tests.

No future. No past. Only now. :)


HamedSabri-adsk (approved):
Nice work @mattyjams . I agree with your comment :)

PR #768, opened 2020-09-09
Fix build script with python 3 where each line starting with 'b and ends with \n' making it hard to read the log.

HamedSabri-adsk:
When running build.py with python3, each line starts with 'b and ends with \n' making it hard to read the log.

Using codec module fixes the problem: https://docs.python.org/3/library/codecs.html

It appears the same fix was also introduced in USD's build_usd.py few months ago:

```
Commit 4f0a1640
by sunyab, 25/03/2020 04:43 PM
committed by pixar-oss
parent 7f64522e
child a9c9786e
```

### DIFF
--- a/build.py
+++ b/build.py
@@ -5,6 +5,7 @@ from distutils.spawn import find_executable
 
 import argparse
 import contextlib
+import codecs
 import datetime
 import distutils
 import multiprocessing
@@ -124,7 +125,7 @@ def Run(context, cmd):
     """Run the specified command in a subprocess."""
     PrintInfo('Running "{cmd}"'.format(cmd=cmd))
 
-    with open(context.logFileLocation, "a") as logfile:
+    with codecs.open(context.logFileLocation, "a", "utf-8") as logfile:
         logfile.write("#####################################################################################" + "\n")
         logfile.write("log date: " + datetime.datetime.now().strftime("%Y-%m-%d %H:%M") + "\n")
         commitID,commitData = GetGitHeadInfo(context)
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -124,7 +125,7 @@ def Run(context, cmd):
     """Run the specified command in a subprocess."""
     PrintInfo('Running "{cmd}"'.format(cmd=cmd))
 
-    with open(context.logFileLocation, "a") as logfile:
+    with codecs.open(context.logFileLocation, "a", "utf-8") as logfile:
```
kxl-adsk, line 128: I had a fix locally for it as well...but I didn't have to change anything else than line 150, i.e. remove encode and write simply content of `l`. 

How can I see the issue which forced you to use codecs?

HamedSabri-adsk: `I had a fix locally for it as well...but I didn't have to change anything else than line 150, i.e. remove encode and write simply content of l.`

If I recall correctly, I had the code originally that way and it was working on some platforms but not others.

`How can I see the issue which forced you to use codecs?`
Use python 3 to run the build script. 

e.g
`C:\Python37\python.exe build.py .....`

kxl-adsk: That is exactly what I'm doing...and all I needed is a change to switch from `logfile.write(str(l.encode("utf8")))` to `logfile.write(l)`.

HamedSabri-adsk: Yeah, I see but we had random failures with that approach.  I think the codecs approach should be fine. I did a few tests and seems to be working.

Please give it a try and see if it solves things for you as well.

kxl-adsk: From offline discussion - this looks to be platform-specific and this is why I only required a smaller fix on MacOS.

kxl-adsk (approved):

PR #771, opened 2020-09-11
Clean up setting _ufeDstItem in renameUndo() 

HamedSabri-adsk:
Minor clean up for the purpose of code readability.  No need to set _ufeDstItem to nullptr at the end of renameUndo(). I already set _ufeDstItem to nullptr in the initializer list.


ppt-adsk (changes requested):
As per my comments on
https://github.com/Autodesk/maya-usd/pull/772
you should simply remove _ufeSrcItem and _ufeDstItem as data members completely.

Therefore, you can just close this pull request.

PR #772, opened 2020-09-11
MAYA-106376: fix undo/redo re-parenting crash due to accessing a stale prim.

HamedSabri-adsk:
This PR addresses the random undo/redo crash after performing multiple re-parenting operations.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
@@ -124,16 +124,20 @@ bool UsdUndoInsertChildCommand::insertChildRedo()
     bool status = SdfCopySpec(_childLayer, _usdSrcPath, _parentLayer, _usdDstPath);
     if (status)
     {
+        // we shouldn't rely on UsdSceneItem to access the UsdPrim since 
+        // it could be stale. Instead we should get the USDPrim from the Ufe::Path
+        const auto& desPrimUsd = ufePathToPrim(_ufeDstPath);
+
         // remove all scene description for the given path and 
         // its subtree in the current UsdEditTarget 
         {
-            auto stage = _ufeSrcItem->prim().GetStage();
+            auto stage = desPrimUsd.GetStage();
             UsdEditContext ctx(stage, _childLayer);
             status = stage->RemovePrim(_usdSrcPath);
         }
 
         if (status) {
-            _ufeDstItem = UsdSceneItem::create(_ufeDstPath, ufePathToPrim(_ufeDstPath));
+            _ufeDstItem = UsdSceneItem::create(_ufeDstPath, desPrimUsd);
             sendNotification<Ufe::ObjectReparent>(_ufeDstItem, _ufeSrcPath);
         }
     }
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -150,16 +154,20 @@ bool UsdUndoInsertChildCommand::insertChildUndo()
...
     {
+        // we shouldn't rely on UsdSceneItem to access the UsdPrim since 
+        // it could be stale. Instead we should get the USDPrim from the Ufe::Path
+        const auto& srcPrimUsd = ufePathToPrim(_ufeSrcPath);
+
         // remove all scene description for the given path and 
         // its subtree in the current UsdEditTarget
         {
-            auto stage = _ufeDstItem->prim().GetStage();
+            auto stage = srcPrimUsd.GetStage();
```
ppt-adsk, line 164: You sure about this?  The original code was getting the stage from the destination item to remove its prim, so I think you should be doing the same.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -124,16 +124,20 @@ bool UsdUndoInsertChildCommand::insertChildRedo()
     bool status = SdfCopySpec(_childLayer, _usdSrcPath, _parentLayer, _usdDstPath);
     if (status)
     {
+        // we shouldn't rely on UsdSceneItem to access the UsdPrim since 
+        // it could be stale. Instead we should get the USDPrim from the Ufe::Path
+        const auto& desPrimUsd = ufePathToPrim(_ufeDstPath);
```
ppt-adsk, line 129: Same a line 164: this will give the wrong stage at line 134.

ppt-adsk (changes requested):
Good catch!  In general, any command which stores UFE scene items (or USD prims, same issue) is suspicious.

I think you should simply remove the _ufeSrcItem and _ufeDstItem data members, because they are no longer useful, and doing this would completely remove the possibility of relying on stale prims.  Just use local variables where UFE scene items are needed.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
ppt-adsk: Naming nit-pick: should use "dst" or "des" prefix for "destination", but not both.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
HamedSabri-adsk: @ppt-adsk  Duh!!  addressed in 9948c423fa9ba43020d6c029f3d008dd53d01df6

HamedSabri-adsk:
> I think you should simply remove the _ufeSrcItem and _ufeDstItem data members, because they are no longer useful, and doing this would completely remove the possibility of relying on stale prims. Just use local variables where UFE scene items are needed.

Good point but we can't remove _ufeDstItem since it needs to be passed to insertedChild().

For now, I removed _ufeSrcItem since it serves no purpose to be stored.  Addressed in  c29d1e48169c7fd8d855b6659b7454c7ee2bcc67

ppt-adsk (approved):

PR #773, opened 2020-09-11
MAYA-106188 - As a user, I would like to activate and deactivate prims

seando-adsk:
#### MAYA-106188 - As a user, I would like to activate and deactivate prims
* Implemented new UFE interface UIInfoHandler.
* Added support for child filtering with new Ufe::Hierarchy method "filteredChildren" and Ufe::HierarchyHandler method "childFilter". In USD we support a single child filter for "Inactive Prims".
* Added context menu for "{Active/Deactivate} Prim".
* Added test cases for new interface, new methods and new menu.

### DIFF
--- a/lib/mayaUsd/ufe/Global.cpp
+++ b/lib/mayaUsd/ufe/Global.cpp
@@ -36,6 +36,9 @@
 #include <mayaUsd/ufe/UsdObject3dHandler.h>
 #include <mayaUsd/ufe/UsdContextOpsHandler.h>
 #include <mayaUsd/ufe/ProxyShapeContextOpsHandler.h>
+#if UFE_PREVIEW_VERSION_NUM >= 2022
+#include <mayaUsd/ufe/UsdUIInfoHandler.h>
+#endif
 #include <ufe/pathString.h>
 #else
 #include <mayaUsd/ufe/UfeVersionCompat.h>
@@ -115,11 +118,17 @@ MStatus initialize()
 	UFE_V2(auto usdAttributesHandler = UsdAttributesHandler::create();)
 	UFE_V2(auto usdObject3dHandler = UsdObject3dHandler::create();)
 	UFE_V2(auto usdContextOpsHandler = UsdContextOpsHandler::create();)
+#if UFE_PREVIEW_VERSION_NUM >= 2022
+	UFE_V2(auto usdUIInfoHandler = UsdUIInfoHandler::create();)
+#endif
 	g_USDRtid = Ufe::RunTimeMgr::instance().register_(
 		kUSDRunTimeName, usdHierHandler, usdTrans3dHandler, 
         usdSceneItemOpsHandler
         UFE_V2(, usdAttributesHandler, usdObject3dHandler)
         UFE_V2(, usdContextOpsHandler)
+#if UFE_PREVIEW_VERSION_NUM >= 2022
+        UFE_V2(, usdUIInfoHandler)
+#endif
     );
 
 #if !defined(NDEBUG)
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchyHandler.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchyHandler.cpp
@@ -57,5 +57,14 @@ Ufe::SceneItem::Ptr ProxyShapeHierarchyHandler::createItem(const Ufe::Path& path
 	return fMayaHierarchyHandler->createItem(path);
 }
 
+#ifdef UFE_V2_FEATURES_AVAILABLE
+#if UFE_PREVIEW_VERSION_NUM >= 2022
+Ufe::Hierarchy::ChildFilter ProxyShapeHierarchyHandler::childFilter() const
+{
+	return Ufe::Hierarchy::ChildFilter();
+}
+#endif
+#endif
+
 } // namespace ufe
 } // namespace MayaUsd
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchyHandler.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchyHandler.cpp
@@ -61,7 +61,9 @@ Ufe::SceneItem::Ptr ProxyShapeHierarchyHandler::createItem(const Ufe::Path& path
 #if UFE_PREVIEW_VERSION_NUM >= 2022
 Ufe::Hierarchy::ChildFilter ProxyShapeHierarchyHandler::childFilter() const
 {
-	return Ufe::Hierarchy::ChildFilter();
+	Ufe::Hierarchy::ChildFilter childFilters;
+	childFilters.emplace_back("InactivePrims", "Inactive Prims", true);
+	return childFilters;
 }
 #endif
 #endif
--- a/lib/mayaUsd/ufe/UsdContextOps.cpp
+++ b/lib/mayaUsd/ufe/UsdContextOps.cpp
@@ -15,6 +15,8 @@
 //
 #include "UsdContextOps.h"
 
+#include "private/UfeNotifGuard.h"
+
 #include <cassert>
 
 #include <maya/MGlobal.h>
@@ -47,6 +49,9 @@ static constexpr char kUSDVariantSetsLabel[] = "Variant Sets";
 static constexpr char kUSDToggleVisibilityItem[] = "Toggle Visibility";
 static constexpr char kUSDMakeVisibleLabel[] = "Make Visible";
 static constexpr char kUSDMakeInvisibleLabel[] = "Make Invisible";
+static constexpr char kUSDToggleActiveStateItem[] = "Toggle Active State";
+static constexpr char kUSDActivatePrimLabel[] = "Activate Prim";
+static constexpr char kUSDDeactivatePrimLabel[] = "Deactivate Prim";
 static constexpr char kUSDAddNewPrimItem[] = "Add New Prim";
 static constexpr char kUSDAddNewPrimLabel[] = "Add New Prim";
 static constexpr char kUSDDefPrimItem[] = "Def";
@@ -88,6 +93,33 @@ private:
     UsdVariantSet     fVarSet;
     const std::string fOldSelection;
     const std::string fNewSelection;
+};
+
+//! \brief Undoable command for prim active state change
+class ToggleActiveStateCommand : public Ufe::UndoableCommand
+{
+public:
+    ToggleActiveStateCommand(const UsdPrim& prim)
+        : _prim(prim)
+    {
+        _active = _prim.IsActive();
+    }
+
+    void undo() override
+    {
+        MayaUsd::ufe::InAddOrDeleteOperation ad;
+        _prim.SetActive(_active);
+    }
+
+    void redo() override
+    {
+        MayaUsd::ufe::InAddOrDeleteOperation ad;
+        _prim.SetActive(!_active);
+    }
+
+private:
+    UsdPrim _prim;
+    bool _active;
 };
 
 const char* selectUSDFileScript = R"(
@@ -235,12 +267,14 @@ Ufe::ContextOps::Items UsdContextOps::getItems(
             items.emplace_back(Ufe::ContextItem::kSeparator);
         }
 
-        // Top-level items.  Variant sets and visibility. Do not add for gateway type node.
+        // Top-level items (do not add for gateway type node):
         if (!fIsAGatewayType) {
+            // Variant sets:
             if (prim().HasVariantSets()) {
                 items.emplace_back(
                     kUSDVariantSetsItem, kUSDVariantSetsLabel, Ufe::ContextItem::kHasChildren);
             }
+            // Visibility:
             // If the item has a visibility attribute, add menu item to change visibility.
             // Note: certain prim types such as shaders & materials don't support visibility.
             auto attributes = Ufe::Attributes::attributes(sceneItem());
@@ -255,6 +289,8 @@ Ufe::ContextOps::Items UsdContextOps::getItems(
                     items.emplace_back(kUSDToggleVisibilityItem, l);
                 }
             }
+            // Prim active state:
+            items.emplace_back(kUSDToggleActiveStateItem, prim().IsActive() ? kUSDDeactivatePrimLabel : kUSDActivatePrimLabel);
         }
 
         // Top level item - Add New Prim (for all context op types).
@@ -344,6 +380,9 @@ Ufe::UndoableCommand::Ptr UsdContextOps::doOpCmd(const ItemPath& itemPath)
         return visibility->setCmd(
             current == UsdGeomTokens->invisible ? UsdGeomTokens->inherited : UsdGeomTokens->invisible);
     } // Visibility
+    else if (itemPath[0] == kUSDToggleActiveStateItem) {
+        return std::make_shared<ToggleActiveStateCommand>(prim());
+    } // ActiveState
     else if (!itemPath.empty() && (itemPath[0] == kUSDAddNewPrimItem)) {
         // Operation is to create a new prim of the type specified.
         if (itemPath.size() != 2u) {
--- a/lib/mayaUsd/ufe/UsdContextOps.cpp
+++ b/lib/mayaUsd/ufe/UsdContextOps.cpp
@@ -100,25 +100,37 @@ class ToggleActiveStateCommand : public Ufe::UndoableCommand
 {
 public:
     ToggleActiveStateCommand(const UsdPrim& prim)
-        : _prim(prim)
     {
-        _active = _prim.IsActive();
+        _stage = prim.GetStage();
+        _primPath = prim.GetPath();
+        _active = prim.IsActive();
     }
 
     void undo() override
     {
-        MayaUsd::ufe::InAddOrDeleteOperation ad;
-        _prim.SetActive(_active);
+        if (_stage) {
+            UsdPrim prim = _stage->GetPrimAtPath(_primPath);
+            if (prim.IsValid()) {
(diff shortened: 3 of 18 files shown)


lib/mayaUsd/ufe/Global.cpp
```diff
@@ -115,11 +118,17 @@ MStatus initialize()
 	UFE_V2(auto usdAttributesHandler = UsdAttributesHandler::create();)
 	UFE_V2(auto usdObject3dHandler = UsdObject3dHandler::create();)
 	UFE_V2(auto usdContextOpsHandler = UsdContextOpsHandler::create();)
+#if UFE_PREVIEW_VERSION_NUM >= 2022
+	UFE_V2(auto usdUIInfoHandler = UsdUIInfoHandler::create();)
+#endif
```
seando-adsk, line 123: A lot of my code will be double protected: UFE_V2 and UFE preview version 0.2.22

lib/mayaUsd/ufe/ProxyShapeHierarchyHandler.cpp
```diff
@@ -57,5 +57,14 @@ Ufe::SceneItem::Ptr ProxyShapeHierarchyHandler::createItem(const Ufe::Path& path
 	return fMayaHierarchyHandler->createItem(path);
 }
 
+#ifdef UFE_V2_FEATURES_AVAILABLE
+#if UFE_PREVIEW_VERSION_NUM >= 2022
+Ufe::Hierarchy::ChildFilter ProxyShapeHierarchyHandler::childFilter() const
+{
+	return Ufe::Hierarchy::ChildFilter();
```
seando-adsk, line 64: Return empty list - we don't (yet) support any child filters for the proxy shape.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -47,6 +49,9 @@ static constexpr char kUSDVariantSetsLabel[] = "Variant Sets";
 static constexpr char kUSDToggleVisibilityItem[] = "Toggle Visibility";
 static constexpr char kUSDMakeVisibleLabel[] = "Make Visible";
 static constexpr char kUSDMakeInvisibleLabel[] = "Make Invisible";
+static constexpr char kUSDToggleActiveStateItem[] = "Toggle Active State";
+static constexpr char kUSDActivatePrimLabel[] = "Activate Prim";
+static constexpr char kUSDDeactivatePrimLabel[] = "Deactivate Prim";
```
seando-adsk, line 54: New context menu item to activate/deactivate a prim

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -90,6 +95,33 @@ class SetVariantSelectionUndoableCommand : public Ufe::UndoableCommand
...
+public:
+    ToggleActiveStateCommand(const UsdPrim& prim)
+        : _prim(prim)
+    {
+        _active = _prim.IsActive();
+    }
+
+    void undo() override
+    {
+        MayaUsd::ufe::InAddOrDeleteOperation ad;
```
seando-adsk, line 110: Technically we are not an add or delete operation but by activate/deactivate a prim we want to send the UFE add or delete notifs.

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -41,19 +42,16 @@
 #endif
 
 namespace {
-	UsdPrimSiblingRange filteredChildren( const UsdPrim& prim )
+	UsdPrimSiblingRange getUSDFilteredChildren(const UsdPrim& prim, const Usd_PrimFlagsPredicate pred = UsdPrimDefaultPredicate)
```
seando-adsk, line 45: Added the predicate as input (with default) so that it can be used by new filtered children method.

seando-adsk, line 45: Renamed helper function as it conflicted with new filteredChildren method.

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -101,18 +99,43 @@ Ufe::SceneItem::Ptr UsdHierarchy::sceneItem() const
...
 }
 
 Ufe::SceneItemList UsdHierarchy::children() const
 {
-	// Return USD children only, i.e. children within this run-time.
-	Ufe::SceneItemList children;
-	for (auto child : filteredChildren(prim()))
-	{
-		children.emplace_back(UsdSceneItem::create(fItem->path() + child.GetName(), child));
-	}
```
seando-adsk, line 114: Moved this code into new private method createUFEChildList()

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -101,18 +99,43 @@ Ufe::SceneItem::Ptr UsdHierarchy::sceneItem() const
...
+#ifdef UFE_V2_FEATURES_AVAILABLE
+#if UFE_PREVIEW_VERSION_NUM >= 2022
+Ufe::SceneItemList UsdHierarchy::filteredChildren(const ChildFilter& childFilter) const
+{
+    // Note: for now the only child filter flag we support is "Inactive Prims".
+    //       See UsdHierarchyHandler::childFilter()
+    if ((childFilter.size() == 1) && (childFilter.front().name == "InactivePrims"))
+    {
+        // See uniqueChildName() for explanation of USD filter predicate.
+        Usd_PrimFlagsPredicate flags = childFilter.front().value ? UsdPrimIsDefined && !UsdPrimIsAbstract
```
seando-adsk, line 119: Leave off the UsdPrimIsActive flag so we'll also get inactive prims.

lib/mayaUsd/ufe/UsdHierarchyHandler.cpp
```diff
@@ -52,5 +52,17 @@ Ufe::SceneItem::Ptr UsdHierarchyHandler::createItem(const Ufe::Path& path) const
...
 }
 
+#ifdef UFE_V2_FEATURES_AVAILABLE
+#if UFE_PREVIEW_VERSION_NUM >= 2022
+Ufe::Hierarchy::ChildFilter UsdHierarchyHandler::childFilter() const
+{
+    Ufe::ChildFilterFlag inactivePrims("InactivePrims", "Inactive Prims", true);
+    Ufe::Hierarchy::ChildFilter childFilters;
+    childFilters.emplace_back(inactivePrims);
+    return childFilters;
```
seando-adsk, line 62: For now, we support a single child filter for inactive prims.

lib/mayaUsd/ufe/UsdSceneItemOps.cpp
```diff
@@ -60,14 +75,24 @@ Ufe::SceneItem::Ptr UsdSceneItemOps::sceneItem() const
...
-	deleteCmd->execute();
-	return deleteCmd;
+	if (prim().IsActive()) {
+		auto deleteCmd = UsdUndoDeleteCommand::create(prim());
+		deleteCmd->execute();
+		return deleteCmd;
+	}
+
+	displayWarning(prim(), kWarningCannotDeactivePrim);
+	return nullptr;
```
seando-adsk, line 85: As per spec - if user tries to delete an inactive prim, print a warning message and don't execute command.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -0,0 +1,81 @@
...
+	UsdSceneItem::Ptr usdItem = std::dynamic_pointer_cast<UsdSceneItem>(item);
+#if !defined(NDEBUG)
+	assert(usdItem);
+#endif
+	if (usdItem)
+	{
+		if (!usdItem->prim().IsActive())
+		{
+			changed = true;
+			info.fontStrikeout = true;
```
seando-adsk, line 55: We've changed something so return true. In Maya this is helpful to know if the input/output struct values need to be copied back.

Set the font to strikeout.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -0,0 +1,81 @@
...
+				(outlinerInvisibleColor.length() == 3))
+			{
+				double rgb[3];
+				outlinerInvisibleColor.get(rgb);
+				info.textFgColor.set(static_cast<float>(rgb[0]), static_cast<float>(rgb[1]), static_cast<float>(rgb[2]));
+			}
+			else
+			{
+				info.textFgColor.set(0.403922f, 0.403922f, 0.403922f);
+			}
```
seando-adsk, line 67: Get the current Maya Outliner invisible color and set that for inactive prim. If we cannot get color just use a hard-coded gray. This happens when we run the tests. The displayRGBColor command will fail.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -0,0 +1,81 @@
...
+		}
+	}
+
+	return changed;
+}
+
+std::string UsdUIInfoHandler::getLongRunTimeLabel() const
+{
+	return "Universal Scene Description";
```
seando-adsk, line 76: Used by the menu divider (for a label).

test/lib/ufe/testChildFilter.py
```diff
@@ -0,0 +1,83 @@
...
+        # Load plugins
+        self.assertTrue(self.pluginsLoaded)
+
+        # Open ballset.ma scene in test-samples
+        mayaUtils.openGroupBallsScene()
+
+        # Clear selection to start off
+        cmds.select(clear=True)
+
+    def testFilteredChildren(self):
```
seando-adsk, line 52: Basically this test will load the balls scene which has 6 balls. I verify that "Props" prim has 6 children. Then I deactivate one and verify that I now have 5 children and the deactivated ball isn't one of them. Then use the filteredChildren(InactivePrims) and verify that I get 6 and that the deactivated ball is one of them.

test/lib/ufe/testContextOps.py
```diff
@@ -149,6 +152,22 @@ def testDoOp(self):
...
+        self.assertTrue(self.ball35Prim.IsActive())
+
+        ufeCmd.execute(cmd)
+        self.assertFalse(self.ball35Prim.IsActive())
+
+        cmds.undo()
+        self.assertTrue(self.ball35Prim.IsActive())
+
+        cmds.redo()
+        self.assertFalse(self.ball35Prim.IsActive())
```
seando-adsk, line 169: Test the toggle active state (with undo/redo).

test/lib/ufe/testUIInfoHandler.py
```diff
@@ -0,0 +1,77 @@
...
+        # Load plugins
+        self.assertTrue(self.pluginsLoaded)
+
+        # Open ballset.ma scene in test-samples
+        mayaUtils.openGroupBallsScene()
+
+        # Clear selection to start off
+        cmds.select(clear=True)
+
+    def testUIInfo(self):
```
seando-adsk, line 51: Basically this tests the new UIInfoHandler interface. Make sure we have a long runtime label. And that the treeViewCellInfo will do something for an inactive prim.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -90,6 +95,33 @@ class SetVariantSelectionUndoableCommand : public Ufe::UndoableCommand
...
+    }
+
+    void redo() override
+    {
+        MayaUsd::ufe::InAddOrDeleteOperation ad;
+        _prim.SetActive(!_active);
+    }
+
+private:
+    UsdPrim _prim;
```
ppt-adsk, line 121: On working with Hamed on parenting bugs I've become increasingly weary about storing prims (or scene items, same thing) as command data members, because they can go stale.  Imagine the following sequence of commands:
- Inactivate prim (prim stored in this command)
- Reparent prim (prim removed using UsdStage::removePrim())
- Undo reparent
- Undo inactivate --- boom.
You should be able to demonstrate this fairly easily.  Solution is the same everywhere, and I know you've implemented it in the past: use a path.  In this case it can be the SdfPath and the UsdStage of the prim (that would get my vote), or you can use the Ufe::Path and call Hierarchy::createItem(), then downcast to the UsdSceneItem and get the prim.

lib/mayaUsd/ufe/ProxyShapeHierarchyHandler.cpp
ppt-adsk: Doesn't this mean that inactive children of the USD pseudo-root (which the proxy shape represents) can't be shown?

test/lib/ufe/testChildFilter.py
```diff
@@ -0,0 +1,83 @@
...
+        # Ensure we have one USD child filter
+        rid = ufe.RunTimeMgr.instance().getId('USD')
+        usdHierHndlr = ufe.RunTimeMgr.instance().hierarchyHandler(rid)
+        cf = usdHierHndlr.childFilter()
+        self.assertEqual(1, len(cf))
+
+        # Make sure the USD hierarchy handler has an inactive prims filter
+        self.assertEqual('InactivePrims', cf[0].name)
+
+        # Toggle "Inactive Prims" on and get the filtered chlidren with
```
ppt-adsk, line 78: da chlidren

test/lib/ufe/testChildFilter.py
ppt-adsk: Nice test, thanks.

ppt-adsk (changes requested):

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -101,18 +99,43 @@ Ufe::SceneItem::Ptr UsdHierarchy::sceneItem() const
...
+    return Ufe::SceneItemList();
+}
+#endif
+#endif
+
+Ufe::SceneItemList UsdHierarchy::createUFEChildList(const UsdPrimSiblingRange& range) const
+{
+    // Return UFE child list from input USD child list.
+    Ufe::SceneItemList children;
+    for (auto child : range)
```
HamedSabri-adsk, line 134: @seando-adsk  Unnecessary copy here. Use `const auto& child `

HamedSabri-adsk (changes requested):

lib/mayaUsd/ufe/UsdHierarchyHandler.cpp
HamedSabri-adsk, line 62: construct ChildFilterFlag object in place, creating inactivePrims is not necessary:

```
Ufe::Hierarchy::ChildFilter childFilters;
childFilters.emplace_back("InactivePrims", "Inactive Prims", true);
```

lib/mayaUsd/ufe/ProxyShapeHierarchyHandler.cpp
seando-adsk: Oops I forgot about that case. I'll fix it.

test/lib/ufe/testChildFilter.py
seando-adsk: its a new word I invented

lib/mayaUsd/ufe/UsdHierarchy.cpp
seando-adsk: Good catch - thanks.

lib/mayaUsd/ufe/UsdHierarchyHandler.cpp
seando-adsk: Another good catch - thanks.

lib/mayaUsd/ufe/UsdContextOps.cpp
seando-adsk: Ah yes good thinking.

lib/mayaUsd/ufe/ProxyShapeHierarchyHandler.cpp
seando-adsk: See commit e19fa54bb6955fb94e66aef00a39385e250dd73a

lib/mayaUsd/ufe/UsdContextOps.cpp
seando-adsk: See commit e19fa54bb6955fb94e66aef00a39385e250dd73a

lib/mayaUsd/ufe/UsdHierarchyHandler.cpp
seando-adsk: See commit e19fa54bb6955fb94e66aef00a39385e250dd73a

HamedSabri-adsk (approved):
LGTM! Thank you.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -100,25 +100,37 @@ class ToggleActiveStateCommand : public Ufe::UndoableCommand
 {
 public:
     ToggleActiveStateCommand(const UsdPrim& prim)
-        : _prim(prim)
     {
-        _active = _prim.IsActive();
+        _stage = prim.GetStage();
```
ppt-adsk, line 104: Nit-pick: could have put everything in the constructor initializer list, rather than having the default constructor called, then doing the assignment:
ToggleActiveStateCommand(const UsdPrim& prim) : _stage(prim.GetStage()), _primPath(prim.GetPath()), _active(prim.IsActive) {}

ppt-adsk (approved):

seando-adsk:
Ready to merge. My preflight failed on the 2020_python2_osx because of machine connection problem. All the other builds passed.

PR #777, opened 2020-09-16
delete testing UI and resource handles in live surface and selection performance tests

mattyjams:
I'm hoping this might help resolve the issue raised by @HamedSabri-adsk and @kxl-adsk in https://github.com/Autodesk/maya-usd/pull/761#discussion_r487392722.

This is unfortunately just a shot in the dark though since I'm not able to reproduce the failed assertions when Maya exits on my end.

### DIFF
--- a/test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeLiveSurface.py
+++ b/test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeLiveSurface.py
@@ -47,19 +47,14 @@ class testProxyShapeLiveSurface(unittest.TestCase):
         mayaSceneFullPath = os.path.join(self._inputDir, mayaSceneFile)
         cmds.file(mayaSceneFullPath, open=True, force=True)
 
-    def testObjectPosition(self):
-        """
-        Tests that an object created interactively is positioned correctly on
-        the live surface.
-        """
         # Create a new custom viewport.
-        window = cmds.window(widthHeight=(500, 400))
+        self._window = cmds.window(widthHeight=(500, 400))
         cmds.paneLayout()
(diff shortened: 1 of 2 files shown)


HamedSabri-adsk:
Hi @mattyjams . Apologies for leaving this review a bit long. 

I tried this PR again with Maya beta both ( binaries and local build ) and still get the failures on Windows.  I don't currently have time to look deep into this and don't have a good guess as what could be the main cause.

One thing, I have noticed with these interactive tests is that I don't get any extra information in the console when the tests fail. I tried `extra-verbose` or `output-on-failure` but nothing shows up in  the Window's console.

kxl-adsk:
@mattyjams Based on feedback from Hamed - would you like to still put this change in?

mattyjams:
Hey guys, sorry to hear this doesn't improve things on Windows. :(

I think I'd still advocate to put this change in since it does clean the tests up a little bit.

@HamedSabri-adsk: To your question about the test output, that's something we've seen as well. For my own testing, I often need to comment out the `cmds.quit()` call at the end of the tests to see what happened, since much of the important output only goes to the script editor and ends up disappearing when Maya shuts down. We work around this internally in a test harness script that installs a filter callback using `MCommandMessage.addCommandOutputFilterCallback()` and writes the output to std out. I can look into bringing that over into test.cmake if that would be of interest to you all, just let me know.

kxl-adsk:
@mattyjams please run PF and we will review your change.

mattyjams:
@kxl-adsk: Just kicked off a pre-flight run, but I don't think these tests are run by pre-flight since they're both interactive?

kxl-adsk:
True.

kxl-adsk (approved):

PR #778, opened 2020-09-16
MAYA-106222 - Fix registry errors for preview surface readers and wri…

JGamache-autodesk:
…ters

Fixes the registration of the UsdPreviewSurface reader and writer. The registration needs to be done in a Tf registry context in order for the automatic removal of the reader/writer to be correctly done when the plugins are unloaded.
The directly visible impact is that this will fix the following coding error message that appeared with PR 744:
`Coding Error: in AddUnloader at line 301 of C:\mayaUsd\fileio\registryHelper.cpp -- Couldn't add unload function (was this function called from outside a TF_REGISTRY_FUNCTION block?)`


kxl-adsk (approved):

kxl-adsk:
@JGamache-autodesk Do you mind updating the description to explain the problem and your fix.

mattyjams (approved):
Looks good to me!

PR #782, opened 2020-09-18
MAYA-106190 - Prim type icons

seando-adsk:
#### MAYA-106629 - Generic USD icon in the main menu

* Moved "ae" templates folder (from "lib/mayaUsd/") to new "resources" folder.
* Added new "resources/icons" folder with all the prim Outliner icons. The icons are part of the MayaUsd_LIB so added icon folder in module path.
* In UFE context menu added icon for "USD Layer Editor menu item.
* In UFE context "Add New Prim" submenu added icons for all the prim types.
* Implemented "treeViewIcon" override in UsdUIInfoHandler interface.
* Added a few icons to the adsk MayaUsd plugin.
* * mayaUsdProxyShape - outliner icon
* * USD_stage/USD_generic used in menus
* Removed "_usd" from proxy shape name, when creating a stage from file.


lib/mayaUsd/CMakeLists.txt
```diff
@@ -163,14 +163,14 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/lib/usd/plugInfo.json
 # -----------------------------------------------------------------------------
 # subdirectories
 # -----------------------------------------------------------------------------
-add_subdirectory(ae)
```
seando-adsk, line 166: After discussion with Krystian, we decided to move this "ae" folder underneath a new "resources" folder. So below you'll see all the ae template (py) files being moved.

lib/mayaUsd/resources/icons/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+        # The _100.png files need to be installed without the _100. This is the
+        # base icon name that is used. Maya will automatically choose the _150/_200
+        # image if neeeded.
+        install(FILES "out_USD_${ICON_BASE}_100.png"
+            DESTINATION "${CMAKE_INSTALL_PREFIX}/lib/icons"
+            RENAME "out_USD_${ICON_BASE}.png"
+        )
+        install(FILES "out_USD_${ICON_BASE}_150.png" "out_USD_${ICON_BASE}_200.png"
+            DESTINATION "${CMAKE_INSTALL_PREFIX}/lib/icons"
+        )
```
seando-adsk, line 43: Install all the Outliner icon files (for the various prim types).

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -275,7 +287,11 @@ Ufe::ContextOps::Items UsdContextOps::getItems(
         int hasLayerEditorCmd{0};
         MGlobal::executeCommand("runTimeCommand -exists UsdLayerEditor", hasLayerEditorCmd);
         if (hasLayerEditorCmd) {
+#if UFE_PREVIEW_VERSION_NUM >= 2023
+            items.emplace_back(kUSDLayerEditorItem, kUSDLayerEditorLabel, kUSDLayerEditorImage);
```
seando-adsk, line 291: Added icon for the "USD Layer Editor" context menu item. And for the "Add New Prim" types below.

lib/mayaUsd/ufe/UsdSceneItem.cpp
```diff
@@ -62,8 +62,7 @@ std::vector<std::string> UsdSceneItem::ancestorNodeTypes() const
 	const TfType& schemaType = fPrim.GetPrimTypeInfo().GetSchemaType();
 #endif
 	if (!schemaType) {
-		TF_CODING_ERROR("Could not find prim type '%s' for prim %s",
-			fPrim.GetTypeName().GetText(), UsdDescribe(fPrim).c_str());
```
seando-adsk, line 66: This error was too extreme. There are nodes which don't have a schema type, so just silently return empty ancestor types.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -71,6 +73,48 @@ bool UsdUIInfoHandler::treeViewCellInfo(const Ufe::SceneItem::Ptr& item, Ufe::Ce
 	return changed;
 }
 
+std::string UsdUIInfoHandler::treeViewIcon(const Ufe::SceneItem::Ptr& item) const
```
seando-adsk, line 76: Override this UFE method to provide an image for specific node types. Avoids the ancestor types call.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -71,6 +73,48 @@ bool UsdUIInfoHandler::treeViewCellInfo(const Ufe::SceneItem::Ptr& item, Ufe::Ce
...
+{
+	// Special case for nullptr input.
+	if (!item) {
+		return "out_USD_UsdTyped.png";	// Default USD icon
+	}
```
seando-adsk, line 81: Can call with nullptr to get a default USD icon.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -71,6 +73,48 @@ bool UsdUIInfoHandler::treeViewCellInfo(const Ufe::SceneItem::Ptr& item, Ufe::Ce
...
+		{"Volume",				"out_USD_Volume.png"}
+	};
+
+	const auto search = supportedTypes.find(item->nodeType());
+	if (search != supportedTypes.cend()) {
+		return search->second;
+	}
+
+	// No specific node type icon was found.
+	return "";
```
seando-adsk, line 115: No specific node type icon provided. Will just ancestor types.

modules/mayaUSD.mod.template
```diff
@@ -8,7 +8,7 @@ PYTHONPATH+:=lib/python
 USD_LOCATION:=
 
 + MayaUSD_LIB ${MAYAUSD_VERSION} ${CMAKE_INSTALL_PREFIX}
-icons: 
+icons: lib/icons
```
seando-adsk, line 11: All the outliner icons above were added to this folder (as part of the base lib). So that any of the 3 plugins will get them.

plugin/adsk/icons/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+# ---------------------------------------------------------------------------------------------
+# install
+# ---------------------------------------------------------------------------------------------
+
+set(PLUGIN_ICONS
+    out_mayaUsdProxyShape
+    USD_generic
+    USD_stage
+)
```
seando-adsk, line 9: These icons are specific to the adsk mayausd plugin.

plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
```diff
@@ -81,9 +81,7 @@ global proc stageFromFile_UICommit(string $parent) {
 proc doCreateStage(string $fileName) {
     // actually load the file
     string $baseName = capitalizeString(`basenameEx $fileName`);
-    if( isValidObjectName($baseName) )
-        $baseName += "_usd";
```
seando-adsk, line 85: As per spec from Nat, remove this _usd from proxy shape name.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
ppt-adsk: It is the case that if you add a new icon, you'll have to make changes in two places, here and in the CMakeLists.txt file, right?  That's a bit unfortunate, but trying to devise a system to make a single change in one place seems difficult.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -71,6 +73,48 @@ bool UsdUIInfoHandler::treeViewCellInfo(const Ufe::SceneItem::Ptr& item, Ufe::Ce
...
 
+std::string UsdUIInfoHandler::treeViewIcon(const Ufe::SceneItem::Ptr& item) const
+{
+	// Special case for nullptr input.
+	if (!item) {
+		return "out_USD_UsdTyped.png";	// Default USD icon
+	}
+
+	// We support these node types directly.
+	static const std::map<std::string, std::string> supportedTypes{
```
ppt-adsk, line 84: Just to avoid repetition, could we have a really small map with the exceptions (empty string and "AL_MayaReference"), and a set for the rest?  I.e.

static const std::map<std::string, std::string> exceptionsTypes {
{"",					"out_USD_Def.png"},
{"AL_MayaReference",	"out_USD_mayaReference.png"}};
if (inTheExceptionsTypes) {
   return that;
}
static const std::set<std::string> supportedTypes {"BlendShape", "Camera", ...};
if (inSupportedTypes) {
    return "out_USD_" + theType + ".png";
}

seando-adsk: It is not really repetition as the map key is node type, which are never repeated. I see your point, but I like the current table. It is very clear what icon filename is associated with each node type. What benefit do you see from your suggestion? It actually requires a 2nd search.

ppt-adsk: Don't think the second search is a big deal.  The repetition is in having the type twice, once in the key type, once in the mapped type, when the mapped type is simply the key type with a fixed prefix and a fixed suffix.  Therefore, the set feels more appropriate, because there really is just one value in the associative data structure, the key type.

If you like the way it is now, I don't have that strong an opinion on it, so I'll leave it up to you.

ppt-adsk (approved):

PR #783, opened 2020-09-18
Fix Python 3 warnings

seando-adsk:
Fix Python 3 warnings (also works in Python 2).
See https://docs.python.org/3.3/library/unittest.html#deprecated-aliases

### DIFF
--- a/plugin/pxr/maya/lib/usdMaya/testenv/testUsdMayaDiagnosticDelegate.py
+++ b/plugin/pxr/maya/lib/usdMaya/testenv/testUsdMayaDiagnosticDelegate.py
@@ -31,6 +31,12 @@ class testUsdMayaDiagnosticDelegate(unittest.TestCase):
     def setUpClass(cls):
         standalone.initialize('usd')
 
+        # Deprecated since version 3.2: assertRegexpMatches and assertRaisesRegexp
+        # have been renamed to assertRegex() and assertRaisesRegex()
+        if sys.version_info.major < 3 or sys.version_info.minor < 2:
+            cls.assertRegex = cls.assertRegexpMatches
+            cls.assertRaisesRegex = cls.assertRaisesRegexp
+
     @classmethod
     def tearDownClass(cls):
         standalone.uninitialize()
--- a/plugin/pxr/maya/lib/usdMaya/testenv/testUsdMayaDiagnosticDelegate.py
+++ b/plugin/pxr/maya/lib/usdMaya/testenv/testUsdMayaDiagnosticDelegate.py
@@ -96,7 +96,7 @@ class testUsdMayaDiagnosticDelegate(unittest.TestCase):
         log = self._StopRecording()
         self.assertEqual(len(log), 1)
         logText, logCode = log[0]
-        self.assertRegexpMatches(logText,
+        self.assertRegex(logText,
                 "^Python coding error: blah -- Coding Error in "
                 "__main__\.testError at line [0-9]+ of ")
         self.assertEqual(logCode, OM.MCommandMessage.kError)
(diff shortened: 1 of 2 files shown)


mattyjams (changes requested):
I haven't tried pulling this change into our environment, but I'm not sure that the new function names are available in our current version of Python (2.7.16):

```Python
> python
Python 2.7.16 (default, Mar  9 2020, 17:47:06) 
[GCC 6.3.1 20170216 (Red Hat 6.3.1-3)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> 
>>> from unittest import TestCase
>>> 
>>> TestCase.assertRegex
  File "<stdin>", line 1, in <module>
AttributeError: type object 'TestCase' has no attribute 'assertRegex'
>>> TestCase.assertRegexpMatches
<unbound method TestCase.assertRegexpMatches>
>>> 
>>> TestCase.assertRaisesRegex
  File "<stdin>", line 1, in <module>
AttributeError: type object 'TestCase' has no attribute 'assertRaisesRegex'
>>> TestCase.assertRaisesRegexp
<unbound method TestCase.assertRaisesRegexp>
```

I do see the new names and the deprecated old names in our version of Python 3 though (3.6.8):

```python
python3
Python 3.6.8 (default, Sep 26 2019, 11:57:09) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-39)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
>>> from unittest import TestCase
>>> 
>>> TestCase.assertRegex
<function TestCase.assertRegex at 0x7fca4c5f0510>
>>> TestCase.assertRegexpMatches
<function TestCase._deprecate.<locals>.deprecated_func at 0x7fca4c5f0ae8>
>>> 
>>> TestCase.assertRaisesRegex
<function TestCase.assertRaisesRegex at 0x7fca4c5f0400>
>>> TestCase.assertRaisesRegexp
<function TestCase._deprecate.<locals>.deprecated_func at 0x7fca4c5f0a60>
```

Is it possible your Python 2 test run wasn't actually running against Python 2?

Maybe we could do a `sys.version` check in `setUpClass()` and monkey patch in the old functions with the new names if we're running against Python 2?

seando-adsk:
@mattyjams You're right sorry about that. I see the same thing with my Python2 version. I don't know what I did - maybe I only ran the tests in python3. 

These are changes I was asked to make by another developer on the Maya team. I was told they were python2 compatible, but I'm questioning that now. I'll get back to you.

seando-adsk:
@mattyjams It turns out these functions have only been deprecated since 3.2 (https://docs.python.org/3.3/library/unittest.html#deprecated-aliases):
`Deprecated since version 3.2: assertRegexpMatches and assertRaisesRegexp have been renamed to assertRegex() and assertRaisesRegex()`

Do you still want me to fix this using your monkey patch suggestion or do you have another suggestion?

mattyjams:
> @mattyjams It turns out these functions have only been deprecated since 3.2 (https://docs.python.org/3.3/library/unittest.html#deprecated-aliases):
> `Deprecated since version 3.2: assertRegexpMatches and assertRaisesRegexp have been renamed to assertRegex() and assertRaisesRegex()`
> 
> Do you still want me to fix this using your monkey patch suggestion or do you have another suggestion?

@seando-adsk: That was just the first idea that came to mind, but I'm open to another solution if you have one.

To make it a little more concrete, this is more or less what I was thinking:

```Python
    @classmethod
    def setUpClass(cls):
        ...
        if sys.version_info.major < 3 or sys.version_info.minor < 2:
            cls.assertRegex = cls.assertRegexpMatches
            cls.assertRaisesRegex = cls.assertRaisesRegexp
        ...
```

mattyjams (approved):
Cool, thanks @seando-adsk! Glad that worked out. Looks good to me!

PR #785, opened 2020-09-21
Adding Cache Id to the Base Proxy Node

alsoussi:
Adding support for having load from Cache Id in the Base proxy node. This feature was available in the Animal Logic Proxy Node so we are promoting it (sort of) and refactoring the code to inherit this property instead of define it. 

Couple of notes to mention, we are sticking with 32-bit int for now as it was the case in the animal logic plugin. The Pixar code specifies long int for the cache id but as that is platform dependent and could cause some compatibility issues we opted to do the same as the animal logic implementation. 

Another note to mention is that the animal logic code was not specifying the behavior of disconnect, we do to make sure the cache id is reset properly. 

Another note to mention is that we decided to split the cache id from being an in/out port to having two sperate in cache id port and out cache id port. But in order not to break compatibility with AL we kept the in port and an in/out port and now we have an extra out cache id port. This is because if you are loading from a file but want a cache id, querying that port will add the stage to the cache, even if it didn't come form the cache id. This seems cleaner and more efficient moving forward. We are happy to refactor the code on the AL side to adopt this change but wanted to get some feedback from the team on this matter as they may have other tests or pipelines depending on that port with that name. Please let us know!


test/lib/testMayaUsdCacheId.py
```diff
@@ -0,0 +1,219 @@
...
+        self.SC = pxr.UsdUtils.StageCache.Get()
+
+    def setUp(self):
+
+        cmds.file(force=True, new=True)
+        cmds.loadPlugin("mayaUsdPlugin", quiet=True)
+        self.assertTrue(
+            cmds.pluginInfo("mayaUsdPlugin", query=True, loaded=True))
+
+        tmpfile = tempfile.NamedTemporaryFile(delete=True, suffix=".usda")
```
kxl-adsk, line 47: @ppt-adsk can you please review the regression test?

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -517,80 +570,89 @@ MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
...
-        // Get input attr values
-        const MString file = dataBlock.inputValue(filePathAttr, &retValue).asString();
+        // Check if we have a Stage from the Cache Id
+        const auto cacheIdNum = dataBlock.inputValue(stageCacheIdAttr, &retValue).asInt();
         CHECK_MSTATUS_AND_RETURN_IT(retValue);
+        UsdStageRefPtr usdStage;
+        const auto     cacheId = UsdStageCache::Id::FromLongInt(cacheIdNum);
+        const auto stageCached = cacheId.IsValid() && UsdUtilsStageCache::Get().Contains(cacheId);
+        if (stageCached) {
+            usdStage = UsdUtilsStageCache::Get().Find(cacheId);
```
williamkrick, line 580: There may be a problem here. There are multiple UsdStageCache objects in use which could lead to incorrect behavior.

1. UsdUtilsStageCache holds a static UsdStageCache in UsdUtilsStageCache::Get().
2. UsdMayaStageCache holds two static UsdStageCache objects in its own UsdMayaStageCache::Get() implementation.

Is it possible for an input cacheId to be from a stage which is in a cache which is not the UsdUtilsStageCache? If so then I guess we fall back on the old code and try to read the stage from disk. But that might not be the correct behavior.

If we have a scene with two proxy shapes and the first proxy shape loads a file from disk (and so the stage ends up in the UsdMayaStageCache and then we connect the output cache id of that shape to the input of the other proxy shape, the 2nd proxy shape will search in the UsdUtilsStageCache and fail to find it.

I guess it works because MayaUsdProxyShapeBase::computeOutStageCacheId() inserts the stage into the UsdUtilsStageCache no matter what.

So the behavior seems fine. And the UsdStageCache doesn't duplicate the stage data, it just adds another reference to the existing stage smart pointer.

This turned into a bit of a run on train of thought with no actual action required by the code author, but I'll leave it here as a useful comment understanding how the different caches are working together safely.

test/lib/testMayaUsdCacheId.py
ppt-adsk, line 47: I'm confused.  The documentation to NamedTemporaryFile() says that the file is deleted as soon as it's closed, i.e. at line 48.  Are you making this call just to get a temporary file name?

If so, you might want to have a look at the test fixtures code we've been using
https://github.com/Autodesk/maya-usd/blob/dev/test/testUtils/fixturesUtils.py
which places test output in a known directory, where you can inspect it after the fact, and cleans out this directory in subsequent runs.

luceric-adsk: I've used that NamedTemporaryFile trick in the layer command tests, and it is used in a few places in the AL plugin tests, testLayerManager.py, testProxyShape.py and testTranslators.py

lib/mayaUsd/nodes/proxyShapeBase.cpp
alsoussi: So as you deduced it won't fail. When you connect it to the output cache id it WILL put that in UsdUtilsStageCache regardless where it loads it from.

I think the MayaUSDCache is an implementation detail that might not exist moving forward for example if we use strong references for the stage.

test/lib/testMayaUsdCacheId.py
alsoussi: Yup, seems to work well but I can switch it if you feel strongly about it.

williamkrick (approved):

test/lib/testMayaUsdCacheId.py
kxl-adsk: AL had issues with NamedTemporaryFile in the past (https://github.com/Autodesk/maya-usd/pull/273#issuecomment-589585095). I don't see why we would need to follow this pattern. @alsoussi please switch to using https://github.com/Autodesk/maya-usd/blob/dev/test/testUtils/fixturesUtils.py

alsoussi: Done!

alsoussi: @kxl-adsk

luceric-adsk (approved):

test/lib/testMayaUsdCacheId.py
luceric-adsk: the tests are failing

      File "/home/S/jenkins/workspace/ecg-mayausd-branch-preflight-python3-linux/build/RelWithDebInfo/test/lib/testMayaUsdCacheId.py", line 47, in setUp
        cmds.file(filePath, force=True, open=True)
    RuntimeError: File not found: "/home/S/jenkins/workspace/ecg-mayausd-branch-preflight-python3-linux/build/RelWithDebInfo/test/lib/MayaUsdPythonCacheIdTestCase/MayaUsdPythonCacheIdTestCase.ma"

(check pre-flight build below, look for "Artifact" link at the top right to download the log)

alsoussi: Yup you are right @luceric-adsk I've fixed them @kxl-adsk

PR #786, opened 2020-09-21
MAYA-105175: Rewrote the rename restriction algorithm from scratch to cover more edge cases.

HamedSabri-adsk:
This PR makes a major improvement in the way "Rename" restriction behaves in Maya.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
@@ -67,6 +67,10 @@ UsdUndoInsertChildCommand::UsdUndoInsertChildCommand(const UsdSceneItem::Ptr& pa
     const auto& childPrim = child->prim();
     const auto& parentPrim = parent->prim();
 
+    // Apply restriction rules
+    ufe::applyCommandRestriction(childPrim, "reparent");
+    ufe::applyCommandRestriction(parentPrim, "reparent");
+
     // First, check if we need to rename the child.
     const auto& childName = uniqueChildName(parent, child->path());
 
@@ -81,21 +85,11 @@ UsdUndoInsertChildCommand::UsdUndoInsertChildCommand(const UsdSceneItem::Ptr& pa
         _ufeDstPath = parent->path() + Ufe::PathSegment(
             Ufe::PathComponent(childName), cRtId, cSep);
     }
-    _usdDstPath = parent->prim().GetPath().AppendChild(TfToken(childName));
+    _usdDstPath = parentPrim.GetPath().AppendChild(TfToken(childName));
 
-    _childLayer = child->prim().GetStage()->GetEditTarget().GetLayer();
+    _childLayer = childPrim.GetStage()->GetEditTarget().GetLayer();
 
-    // If parent prim is the pseudo-root, no def primSpec will be found, so
-    // just use the edit target layer.
-    _parentLayer = parentPrim.IsPseudoRoot() 
-        ? parent->prim().GetStage()->GetEditTarget().GetLayer() 
-        : MayaUsdUtils::defPrimSpecLayer(parentPrim);
-
-    // Apply restriction rules
-    if(!parentPrim.IsPseudoRoot()){
-        ufe::applyCommandRestriction(childPrim, "reparent");
-        ufe::applyCommandRestriction(parentPrim, "reparent");
-    }
+    _parentLayer = parentPrim.GetStage()->GetEditTarget().GetLayer(); 
 }
 
 UsdUndoInsertChildCommand::~UsdUndoInsertChildCommand()
--- a/lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
@@ -67,10 +67,6 @@ UsdUndoInsertChildCommand::UsdUndoInsertChildCommand(const UsdSceneItem::Ptr& pa
     const auto& childPrim = child->prim();
     const auto& parentPrim = parent->prim();
 
-    // Apply restriction rules
-    ufe::applyCommandRestriction(childPrim, "reparent");
-    ufe::applyCommandRestriction(parentPrim, "reparent");
-
     // First, check if we need to rename the child.
     const auto& childName = uniqueChildName(parent, child->path());
 
@@ -94,6 +90,12 @@ UsdUndoInsertChildCommand::UsdUndoInsertChildCommand(const UsdSceneItem::Ptr& pa
     _parentLayer = parentPrim.IsPseudoRoot() 
         ? parent->prim().GetStage()->GetEditTarget().GetLayer() 
         : MayaUsdUtils::defPrimSpecLayer(parentPrim);
+
+    // Apply restriction rules
+    if(!parentPrim.IsPseudoRoot()){
+        ufe::applyCommandRestriction(childPrim, "reparent");
+        ufe::applyCommandRestriction(parentPrim, "reparent");
+    }
 }
 
 UsdUndoInsertChildCommand::~UsdUndoInsertChildCommand()
--- a/lib/mayaUsd/ufe/private/Utils.cpp
+++ b/lib/mayaUsd/ufe/private/Utils.cpp
@@ -133,7 +133,7 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
     auto primStack = prim.GetPrimStack();
     std::string layerDisplayName;
 
-    // check to see if there is a spec at the edit target layer. 
+    // iterate over the prim stack, starting at the highest-priority layer.
     for (const auto& spec : primStack) 
     {
         if (primSpec != spec) {
@@ -141,12 +141,15 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
         }
         else
         {
-            //  break out if the spec already exist
-            if (primSpec == spec && spec->GetSpecifier() == SdfSpecifierDef) {
-                break;
+            // if exists a def primSpec
+            if (spec->GetSpecifier() == SdfSpecifierDef) {
+                if(spec->HasReferences()) {
+                    break;
+                }
+                continue;
             }
 
-            // break out if this is an over that has a reference.
+            // if exists an over that has a reference.
             if(spec->GetSpecifier() == SdfSpecifierOver && spec->HasReferences()) {
                 break;
             }
--- a/lib/mayaUsd/ufe/private/Utils.cpp
+++ b/lib/mayaUsd/ufe/private/Utils.cpp
@@ -134,9 +134,12 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
     std::string layerDisplayName;
 
     // check to see if there is a spec at the edit target layer. 
-    if (primSpec)
+    for (const auto& spec : primStack) 
     {
-        for (const auto& spec : primStack) 
+        if (primSpec != spec) {
+            layerDisplayName.append("[" + spec->GetLayer()->GetDisplayName() + "]" + ",");
+        }
+        else
         {
             // skip if the spec already exist
             if (primSpec == spec && spec->GetSpecifier() == SdfSpecifierDef && primSpec->GetLayer() == spec->GetLayer()) {
@@ -159,22 +162,10 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
                 break;
             }
         }
-
-        if(!layerDisplayName.empty())
-        {
-            layerDisplayName.pop_back();
-            std::string err = TfStringPrintf("Cannot %s [%s]. It is defined on another layer. Please set %s as the target layer to proceed.",
-                                     commandName.c_str(),
-                                     prim.GetName().GetString().c_str(), 
-                                     layerDisplayName.c_str());
-            throw std::runtime_error(err.c_str());
-        }
     }
-    else
+
+    if(!layerDisplayName.empty())
     {
-        for (const auto& spec : primStack) {
-            layerDisplayName.append("[" + spec->GetLayer()->GetDisplayName() + "]" + ",");
-        }
         layerDisplayName.pop_back();
         std::string err = TfStringPrintf("Cannot %s [%s]. It is defined on another layer. Please set %s as the target layer to proceed.",
                                  commandName.c_str(),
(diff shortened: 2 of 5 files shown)


test/lib/ufe/testRename.py
```diff
@@ -270,29 +270,6 @@ def testRenameRestrictionSameLayerDef(self):
...
-        stage = mayaUsd.ufe.getStage(str(mayaPathSegment))
-
-        # set the edit target to Assembly_room_set.usda
-        stage.SetEditTarget(stage.GetLayerStack()[2])
-        self.assertEqual(stage.GetEditTarget().GetLayer().GetDisplayName(), "Assembly_room_set.usda")
-
-        # expect the exception happens
-        with self.assertRaises(RuntimeError):
-           newName = 'Ball_35_Renamed'
-           cmds.rename(newName)
```
HamedSabri-adsk, line 294: testRenameRestrictionOtherLayerOpinions no longer makes any sense.  Removing it all together.

lib/usd/utils/util.h
```diff
@@ -49,14 +37,6 @@ namespace MayaUsdUtils {
...
     bool isInternalReference(const SdfPrimSpecHandle&);
 
-    //! Returns true if the target node has any specs to contribute to the composed prim.
-    MAYA_USD_UTILS_PUBLIC
-    bool hasSpecs(const UsdPrim&);
-
-    //! Returns the layer in composition arc where HasSpecs is set to true
-    MAYA_USD_UTILS_PUBLIC
-    std::vector<SdfLayerHandle> layerInCompositionArcsWithSpec(const UsdPrim& prim);
-
```
HamedSabri-adsk, line 59: Cleaned up the usd utility functions. We no longer use them.

UsdPrimCompositionQuery is a nice utility but doesn't deal with SubLayer comp arcs. For the sake of this task, I stopped using it altogether.

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -129,53 +129,61 @@ UsdGeomXformCommonAPI convertToCompatibleCommonAPI(const UsdPrim& prim)
...
+            layerDisplayName.append("[" + spec->GetLayer()->GetDisplayName() + "]" + ",");
         }
+        layerDisplayName.pop_back();
+        std::string err = TfStringPrintf("Cannot %s [%s]. It is defined on another layer. Please set %s as the target layer to proceed.",
+                                 commandName.c_str(),
+                                 prim.GetName().GetString().c_str(), 
+                                 layerDisplayName.c_str());
+        throw std::runtime_error(err.c_str());
     }
 }
```
HamedSabri-adsk, line 188: This algorithm heavily relies on dealing with "Prim's PrimStack" rather than "Stage's LayerStack" which makes it much easier to interrogate about the SdfPrimSpec(s) in different composition arcs.

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -129,53 +129,61 @@ UsdGeomXformCommonAPI convertToCompatibleCommonAPI(const UsdPrim& prim)
...
+                if (sessionLayer == spec->GetLayer()){
+                    layerDisplayName.append("[" + spec->GetLayer()->GetDisplayName() + "]" + ",");
+                }
+            }
+
+            // if the def primspec is in another layer other than current stage's local layer.
+            if (spec->GetSpecifier() == SdfSpecifierDef && primSpec->GetLayer() != spec->GetLayer()) {
+                layerDisplayName.append("[" + spec->GetLayer()->GetDisplayName() + "]" + ",");
+                break;
+            }
```
HamedSabri-adsk, line 163: Handle different cases around "over" specifier.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -94,6 +90,12 @@ UsdUndoInsertChildCommand::UsdUndoInsertChildCommand(const UsdSceneItem::Ptr& pa
     _parentLayer = parentPrim.IsPseudoRoot() 
         ? parent->prim().GetStage()->GetEditTarget().GetLayer() 
         : MayaUsdUtils::defPrimSpecLayer(parentPrim);
+
+    // Apply restriction rules
+    if(!parentPrim.IsPseudoRoot()){
+        ufe::applyCommandRestriction(childPrim, "reparent");
+        ufe::applyCommandRestriction(parentPrim, "reparent");
+    }
```
HamedSabri-adsk, line 98: I was unconditionally applying the restriction rules here but realized that If parent prim is the pseudo-root, no def primSpec will be found and instead USD returns the "over".

This case happens when one tries to parent a prim under proxy shape ('/').

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -159,22 +162,10 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
...
-            std::string err = TfStringPrintf("Cannot %s [%s]. It is defined on another layer. Please set %s as the target layer to proceed.",
-                                     commandName.c_str(),
-                                     prim.GetName().GetString().c_str(), 
-                                     layerDisplayName.c_str());
-            throw std::runtime_error(err.c_str());
-        }
     }
-    else
+
+    if(!layerDisplayName.empty())
```
HamedSabri-adsk, line 167: Simplified the logic by removing the redundant primSpec check and loop through primStack.

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -150,10 +150,7 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
 
             // if the over exist in the session layer
             if (spec->GetSpecifier() == SdfSpecifierOver) {
-                auto sessionLayer = prim.GetStage()->GetSessionLayer();
-                if (sessionLayer == spec->GetLayer()){
-                    layerDisplayName.append("[" + spec->GetLayer()->GetDisplayName() + "]" + ",");
-                }
+                layerDisplayName.append("[" + spec->GetLayer()->GetDisplayName() + "]" + ",");
```
HamedSabri-adsk, line 156: Remove redundant check to distinguish if the "over" specifier is in the session layer.

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -129,54 +129,49 @@ UsdGeomXformCommonAPI convertToCompatibleCommonAPI(const UsdPrim& prim)
...
+
+            // if exists an over sepc
+            if (spec->GetSpecifier() == SdfSpecifierOver) {
+                layerDisplayName.append("[" + layerName + "]" + ",");
+            }
         }
     }
+
+    if(!layerDisplayName.empty()) {
+        layerDisplayName.pop_back();
```
ppt-adsk, line 168: Why the pop_back()?

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -129,54 +129,49 @@ UsdGeomXformCommonAPI convertToCompatibleCommonAPI(const UsdPrim& prim)
...
-            // we are interested in all the arcs after root
-            std::for_each(std::next(layers.begin()),layers.end(), [&](const auto& it) {
-                 layerDisplayNames.append("[" + it->GetDisplayName() + "]" + ",");
-            });
+        if (primSpec != spec) {
+            layerDisplayName.append("[" + layerName + "]" + ",");
+        }
+        else {
+            // if exist a primSpec with reference
+            if(spec->HasReferences()) {
```
ppt-adsk, line 146: Should comment on why we abort the loop here.

test/lib/ufe/testRename.py
ppt-adsk: Why does it no longer make sense?

HamedSabri-adsk: @ppt-adsk 
What this test was doing was to exercise the following:

"A prim composed of multiple primSpecs can not be renamed". 

So in the case of '/Room_set/Props/Ball_35' in Assembly_room_set.usda

```
def "Ball_35" (
    add references = @./Ball.usd@
    variants = {
        string shadingVariant = "Ball_8"
    }
)
{
    double3 xformOp:translate = (6.43, 20.0025, 0)
    uniform token[] xformOpOrder = ["xformOp:translate"]
}
```
Ball_35 is a result of composition from different primSpecs:

Assembly_room_set.usda	  ------>  root
Ball.usd                  ------>  reference
Ball.shadingVariants.usda ------>  reference
Ball.shadingVariants.usda ------>  variant
Ball.maya.usd             ------>  reference

This was the third rule that we previously had and was very restrictive. 

one should be able to rename Ball_35 because its a concrete type ( def ) in a layer(Assembly_room_set) that has defined it.

ppt-adsk: Ah, O.K., I guess I was misled by the test docstring, which still sounded relevant.

HamedSabri-adsk: I can't agree more. Indeed misleading.

lib/mayaUsd/ufe/private/Utils.cpp
HamedSabri-adsk: The main reason for the skipping the reference is to not clash with the **over** that may be created in **the stage's local sessioLayer or other layers**.

Consider the case, where we have a stage that has opened a USD file called **RefExample.usda** and in that file we have 
a prim with **over** which has an external reference to another Usd file called **HelloWorld.usda**.

**RefExample.usda**
```
over "refSphere2" (
    prepend references = @./HelloWorld.usda@
)
```

**HelloWorld.usda**
```
#usda 1.0
(
    defaultPrim = "hello"
)

def Xform "hello"
{
	...........
}
```

**All PrimStack knows at this point is that "refSphere2" has two specs in 2 different places in order from strong to weak:**

```
RefExample.usda  --->  over
HelloWorld.usda  --->  def
```

Now in line 161, I unconditionally check if there is an override and then record it's layer name which then results in error message to be displayed which is not desirable:

```
// if exists an over sepc
if (spec->GetSpecifier() == SdfSpecifierOver) {
    layerDisplayName.append("[" + layerName + "]" + ",");
}
```

one should be able to rename "refSphere2" since it is created in "RefExample.usda" and itself doesn't have any other "over" in the stage's local layers.

This is why I previously had a condition to check if the "over" was created in sessioLayer or not.

Hope this makes sense.

lib/mayaUsd/ufe/private/Utils.cpp
HamedSabri-adsk: the pop_back is just to remove the last "," when there are multiple layers

`[bob],[foo], `  

I pop back the last "," so it looks

`[bob],[foo]`

ppt-adsk: Right, you've explained this to me once before...  A shortened version of your explanation would make a great code comment.

lib/mayaUsd/ufe/private/Utils.cpp
ppt-adsk: It does.  Can you distill this into a code comment somehow?  I know it's a bit challenging, you don't want to have a 2-page code comment, but it would still be nice.

HamedSabri-adsk: Ugh yes of course! I was planning to do it after the planning meeting.

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -132,43 +134,46 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
...
-        layerDisplayName.pop_back();
-        std::string err = TfStringPrintf("Cannot %s [%s]. It is defined on another layer. Please set %s as the target layer to proceed.",
+        std::string err = TfStringPrintf("Cannot %s [%s]. %s. Please set %s as the target layer to proceed.",
                                  commandName.c_str(),
-                                 prim.GetName().GetString().c_str(), 
+                                 prim.GetName().GetString().c_str(),
+                                 message.c_str(),
                                  layerDisplayName.c_str());
         throw std::runtime_error(err.c_str());
     }
```
HamedSabri-adsk, line 179: @ppt-adsk  

- polished the algorithm slightly to make it easier to reason about the code and added appropriate comments.
- added proper messaging. we now only need to deal with two messages ( no more fancy words ) :
1. Cannot rename something because **It is defined on another layer**
2. Cannot rename something because **It has a stronger opinion on another layer**

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -81,21 +85,11 @@ UsdUndoInsertChildCommand::UsdUndoInsertChildCommand(const UsdSceneItem::Ptr& pa
...
-
-    _childLayer = child->prim().GetStage()->GetEditTarget().GetLayer();
+    _usdDstPath = parentPrim.GetPath().AppendChild(TfToken(childName));
 
-    // If parent prim is the pseudo-root, no def primSpec will be found, so
-    // just use the edit target layer.
-    _parentLayer = parentPrim.IsPseudoRoot() 
-        ? parent->prim().GetStage()->GetEditTarget().GetLayer() 
-        : MayaUsdUtils::defPrimSpecLayer(parentPrim);
+    _childLayer = childPrim.GetStage()->GetEditTarget().GetLayer();
```
HamedSabri-adsk, line 90: @ppt-adsk  clean up the legacy old code for getting the appropriate layer for _parentLayer .  

ufe::applyCommandRestriction in line 71 and 72 will always put us in a position to use appropriate layer to use for performing the parent operation.

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -131,13 +131,20 @@ UsdGeomXformCommonAPI convertToCompatibleCommonAPI(const UsdPrim& prim)
 
 void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName)
 {
+    // return early if prim is the pseudo-root.
+    // this is a special case and could happen when one tries to drag a prim under the 
+    // proxy shape in outliner. Also note if prim is the pseudo-root, no def primSpec will be found.
+    if (prim.IsPseudoRoot()){
+        return;
+    }
+
```
HamedSabri-adsk, line 140: @ppt-adsk per out offline discussion, PseudoRoot() is a special case and should be treated differently.

After some thinking, it makes more sense to simply return early if we are dealing with the root layer.

ppt-adsk (approved):

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -25,6 +25,8 @@
 
 #include <mayaUsdUtils/util.h>
 
+#include <iostream>
```
HamedSabri-adsk, line 28: just noticed I left out to clean up this iostream. Will have it remove in another PR.

PR #787, opened 2020-09-21
extract RenderMan for Maya shading export from "pxrRis" shadingMode for use with "useRegistry"

mattyjams:
There was actually hardly anything RfM-specific happening in the shader export side of the `pxrRis` shadingMode, so this change pulls all of that out into a generic shader writer. These then get registered per node-type-name-to-shader-ID mapping with the material conversion `rendermanForMaya`, and the exported shaders are bound using the `ri` `renderContext`.

The eventual goal is that this plus a forthcoming rewrite of the import side will allow us to remove the `pxrRis` shadingMode.

Note that none of this actually requires RenderMan for Maya to be loaded or even installed. The `pxrRis` shadingMode and now these new translators support mapping some Maya-native node types to `Pxr`-prefixed shader IDs.

I'm working on the import side of this as well, but there are a few extra complications there that will likely require some more infrastructure-y changes to the `useRegistry` shadingMode itself, as well as some more convenient API in `UsdShade`.

Looking at the commit history might make it a bit easier to follow the progression here than just looking at the flat diffs, but let me know if there are any questions or concerns!


lib/usd/translators/shading/rfmShaderTranslation.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <pxr/base/tf/token.h>
+
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+
+TF_DEFINE_PRIVATE_TOKENS(
+    _tokens,
+
+    ((conversionName, "rendermanForMaya"))
```
kxl-adsk, line 34: @mattyjams Would it be problematic if we make this translator a separate plugin?

mattyjams: @kxl-adsk: I had thought of that, but I was a little reluctant to do so right now. My short term goal here is to get rid of the `pxrRis` shadingMode and replace it with this. Anyone who had been using that would need to start using the `useRegistry` shadingMode instead, so I was hesitant to force them to make that change *and* load something else. Especially so since with the use of the genericShaderWriter, there would be almost nothing to the plugin. If you all have strong feelings that this be separated out though, I could do that. I'd just need some guidance as to what structure/naming would make sense for that.

I think ideally, this could one day be owned by the RenderMan team and just ship as part of their plugin once this shading-related API stabilizes and is available in public releases.

JGamache-autodesk (approved):

JGamache-autodesk:
Looks good. Should not be too difficult to move to the Renderman plugin once the API stabilizes. At that moment, the Pxr part of the RfmNodesToShaderIds table should probably be auto-discovered from the plugin instead of copied in the table.

The genericShaderWriter is quite nice. It obsoletes 90% of the code in maya-usd/dev/test/lib/usd/plugin/mayaGenericWriter.cpp
That class requires attribute names to match 100% between Maya and UsdShade, so maybe renaming it to symmetricShaderWriter would help highlight the capabilities?

lib/usd/translators/shading/rfmShaderTranslation.cpp
kxl-adsk: I see your argument for not having much left in the plugin. Moving it over to RenderMan team would match our own discussions around Arnold and translators for Arnold.

How about we then make this translator conditionally compiled? The point to make here is that not everyone has access to RenderMan and we are considering taking it out of the binaries we release.

mattyjams:
> Looks good. Should not be too difficult to move to the Renderman plugin once the API stabilizes. At that moment, the Pxr part of the RfmNodesToShaderIds table should probably be auto-discovered from the plugin instead of copied in the table.
> 
> The genericShaderWriter is quite nice. It obsoletes 90% of the code in maya-usd/dev/test/lib/usd/plugin/mayaGenericWriter.cpp
> That class requires attribute names to match 100% between Maya and UsdShade, so maybe renaming it to symmetricShaderWriter would help highlight the capabilities?

Sounds good to me. I wasn't quite sure what to call this thing, but "symmetric" does seem to fit a bit better than "generic". I rebased these changes and updated the naming in https://github.com/Autodesk/maya-usd/pull/787/commits/35d7ba70e53f2b5b1c7c53214b3a2b883eca945e. The only other difference from before is that I noticed you were using `MFnDependencyNode::reorderedAttribute()` here:
https://github.com/Autodesk/maya-usd/blob/4365cc6675f666ee8ec23bc21c12ef68b3a5d5c7/test/lib/usd/plugin/mayaGenericWriter.cpp#L184
That seemed like a good practice, so I applied that in the symmetric writer here too.

Once this goes in, I can put up a follow-up change that revisits that `mayaGenericWriter` test plugin and makes it take advantage of this new writer. And yeah, down the line, it would definitely be better to pull that list of RfM nodes/mappings programmatically somehow rather than enumerating them all in that header file.

Thanks @JGamache-autodesk!

lib/usd/translators/shading/rfmShaderTranslation.cpp
mattyjams: Yeah, that makes sense. I can see not necessarily wanting to ship RenderMan-related stuff with the base mayaUsd installer, though there isn't currently a way to cull the `pxrRis` shadingMode. But hopefully we're killing that soon anyway!

In the rebased commits I just pushed, you'll see a `BUILD_RFM_TRANSLATORS` build option was added in https://github.com/Autodesk/maya-usd/pull/787/commits/fc8efb0b686310a2e87c46fe514d5d83aeff979f. This needed to go at the top-level since the option is used in both the `lib/...` and `test/...` trees, and you'll see the option checked again to conditionally add the test in  https://github.com/Autodesk/maya-usd/pull/787/commits/2d951e4b008db3146b353ee1403630cc0512b9b0.

JGamache-autodesk (approved):

lib/usd/translators/shading/rfmShaderTranslation.cpp
kxl-adsk: Thank you for adding `BUILD_RFM_TRANSLATORS`, @mattyjams.

I'm definitely for killing pxrRis shadingMode - is this something you are planning on doing in the near future or it's more on us to execute on it?

mattyjams: Yeah, I'm planning on doing that once `useRegistry` provides a complete replacement for the functionality `pxrRis` provided. We don't actually use the `pxrRis` shadingMode internally, so most of the work is probably just making sure that anyone external who does use it is aware that it's being removed.

I'm planning to kill the `displayColor` shadingMode too, though that might be slightly more contentious since it seems more likely that people actively use that one. Hopefully not though. With Jerry's work to support Maya-native materials, I don't see much of a need to keep `displayColor` anymore.

mattyjams:
Hey @kxl-adsk, I ran one last pre-flight check on this PR but it looks like the master/Windows/Python 3 build hit some sort of failure. It looks like all of the tests failed, but the process failed during some packaging step towards the end, maybe where it archives build artifacts?:
https://github.com/Autodesk/maya-usd/runs/1161153739?check_suite_focus=true

All the other platforms and build configurations look fine.

Let me know if I should kick off another one for this PR. Thanks!

lib/usd/translators/shading/rfmShaderTranslation.cpp
kxl-adsk: > I'm planning on doing that once useRegistry provides a complete replacement for the functionality pxrRis provided

What would you consider as still missing?

kxl-adsk:
@mattyjams no need to run it again. It's something our DevOps team is tracking and working hard to eliminate.

lib/usd/translators/shading/rfmShaderTranslation.cpp
mattyjams: This PR only adds export support, so I have a follow-up one I'm working on that'll take care of import.

A few nuances off the top of my head that we'll still need to deal with:
- exporting needs to be able to consider non-built-in shading plugs that some renderers use on the `shadingEngine` node:
https://github.com/Autodesk/maya-usd/blob/91d0bfe820f74e65d4d4f91049d8171fbf1b5dcf/lib/mayaUsd/fileio/shading/shadingModePxrRis.cpp#L83
RenderMan for Maya obviously does this, but I think Arnold might as well? I think the intent is to help differentiate between the shader used for previewing in the Maya viewport and the renderer-specific shader used by the renderer.
- Similarly, import likely needs to support creating these non-built-in plugs and making shading connections to them, for example when a Material prim has both an `outputs:surface` and an `outputs:ri:surface` terminal, or potentially if we want to start considering [materialPurpose (e.g. preview vs. full)](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usd/usdShade/materialBindingAPI.h#L77).
- `useRegistry` currently only supports importing the universal renderContext terminals of Material prims. The export of RfM shaders here writes them into the `ri` renderContext, so `useRegistry` will not currently see them on import. This will tie into the feature above so that the import knows which plugs to connect these shaders to on the `shadingEngine`.

I'm working on (or will be working on) all of these things and will put up PRs for them, and happy to discuss the intricacies of each as we dig into them.

JGamache-autodesk: Seems we need to get in touch as I am also looking into expanding the material import code to support an import path similar to the materialConversion that was added recently on export.
Some information:
- Arnold does use the same shading group attachment points as regular surface materials, so it does not allow a material purpose concept in Maya.
- We would like to be able to pass multiple shadingMode to the import command to allow specifying a list of locations where we want to search for materials in USD. I thought the search would end on the first surface material found, but having custom entry points for RfM might mean reconsidering how the search terminates. The import options would be simplified to try everywhere, and a user wanting full control over where to search could do so at the command level.
- The usdListShadingModes command would be expanded to also list materialConversions that have an import path. This means expanding the materialConversion registry with import information.

mattyjams: @JGamache-autodesk: Yup, definitely. Sorry, I didn't mean to suggest that any of the things I mentioned were solved problems with well understood solutions that just needed implementing. I'm sure we'll be talking about all of these things and more as we keep going. PRs (or maybe draft PRs?) might be a good way to anchor some of the discussion around something concrete as we break down each piece, but I'm expecting we'll need to iterate a bit to find the best way forward.

Good to know about Arnold, as well as the other ideas you have in mind. All sounds great to me. Catch up with you about it soon!

PR #790, opened 2020-09-22
add missing "import sys" to testPxrUsdPreviewSurfaceDraw test

mattyjams:


kxl-adsk:
@mattyjams How did we miss this before?

mattyjams:
> @mattyjams How did we miss this before?

@kxl-adsk: I think the trick is that the CMake build adds a little preamble when adding the test that includes a `import sys`:
https://github.com/Autodesk/maya-usd/blob/77ff438fe876858faff32b8fb96cfd7c25085906/cmake/test.cmake#L109

I tripped over this when moving some stuff around in our internal SCons-based build.

kxl-adsk (approved):

Issue #589, opened 2020-06-11
WANT_UFE_BUILD scattered throughout build process


marsupial:
**Describe the issue**
**WANT_UFE_BUILD** can wind up controlling vtable layout and availability of functions.
Currently it's usage is scattered throughout CMake and headers in mayaUsd, hdMaya, mayaToHydra, AL-usdmaya.  Usually as **PRIVATE** sometimes as **PUBLIC** (#583).

Seems better to be held in a spot where all sub-projects consume the value, and anyone building against a distribution can ingest the MACRO properly (i.e. in an installed CMake target).

kxl-adsk:
@marsupial We discussed this internally and agreed that `WANT_UFE_BUILD` should be `PUBLIC` in https://github.com/Autodesk/maya-usd/blob/f2ccf7ae4012ba5a23a5b9cbe7e0947dd63160d9/lib/mayaUsd/CMakeLists.txt#L47

kxl-adsk:
Fixed in dev branch.

PR #793, opened 2020-09-23
Make WANT_UFE_BUILD def public.

HamedSabri-adsk:
Make WANT_UFE_BUILD def public per discussion in issue https://github.com/Autodesk/maya-usd/issues/589


seando-adsk:
@kxl-adsk Can you review the change I just pushed to fix the failing test.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -316,15 +318,21 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
...
 void StagesSubject::onStageSet(const MayaUsdProxyStageSetNotice& notice)
 {
-	// We should have no listerners and stage map is dirty.
-	TF_VERIFY(g_StageMap.isDirty());
-	TF_VERIFY(fStageListeners.empty());
-
-	StagesSubject::Ptr me(this);
-	for (auto stage : ProxyShapeHandler::getAllStages())
+	// Handle re-entrant onStageSet
+	if (!stageSetGuardCount.load())
```
kxl-adsk, line 322: You will have atomic load, but the store on line 324 is not guaranteed to happen without any other thread reading the current value of stageSetGuardCount. Here is how you can make it (and then remove the line 324)
```suggestion
	bool expectedState = false;
	if (stageSetGuardCount.compare_exchange_strong(expectedState,true))
```

seando-adsk:
Thanks @kxl-adsk. That was so cool I committed the change right from the PR (after testing it locally, of course).

kxl-adsk:
Almost perfect...I just forgot to remove the assignment from line 325.

seando-adsk:
@kxl-adsk Done.

kxl-adsk (approved):

kxl-adsk (approved):

PR #799, opened 2020-09-25
extract RenderMan for Maya shading import from "pxrRis" shadingMode for use with "useRegistry"

mattyjams:
Continuing on from #787, these changes pull the import support from the `pxrRis` shadingMode and turns it into a symmetric shader reader for use with the `useRegistry` shadingMode.

There's definitely still more to do here and more issues to figure out, but this rounds out the core of the symmetric import/export. We'll need to catch up on and figure out how to deal with some of the finer details like custom plugs on Maya `shadingEngine` nodes and non-universal renderContexts on USD `Shader` prim terminal outputs.

On that last point, one important gotcha here is that the added `testUsdImportRfMShaders` test USD has its material using the universal renderContext instead of the `ri` renderContext that the `rendermanForMaya` material conversion generates at export. This means that strictly speaking, shading authored using this scheme won't yet roundtrip. We'll need some additional work in the `useRegistry` shadingMode itself to add support for importing from non-universal renderContexts, so that will come in future PRs.


test/lib/usd/translators/UsdImportRfMShadersTest/MarbleCube.usda
```diff
@@ -0,0 +1,73 @@
...
+    {
+        def Material "MarbleCubeSG"
+        {
+            # XXX: Note that the rendermanForMaya material conversion at export
+            # will create this surface terminal output using the "ri"
+            # renderContext as opposed to using the universal renderContext as
+            # we do here. Further work is needed in the "useRegistry"
+            # shadingMode to be able to support finding materials in
+            # renderContexts other than universal.
+            token outputs:surface.connect = </MarbleCube/Materials/MarbleCubeSG/MarbleLambert.outputs:surface>
```
mattyjams, line 47: Here's where we're cheating for now. If you were to run the export side of this test, it would author the terminal using the `ri` renderContext:

```
token outputs:ri:surface.connect = ...
```

But right now, the `useRegistry` shadingMode doesn't know to look anywhere other than the universal renderContext when looking for shaders to import. I'll revisit this to use the correct renderContext once we add that support.

JGamache-autodesk (approved):
Looks good to me.

PR #803, opened 2020-09-28
Pr/update custom converter test plugin to use symmetric writer

mattyjams:
One last bit of follow-up from the discussion in https://github.com/Autodesk/maya-usd/pull/787#issuecomment-698165564.

This updates the `usdTestPlugin` used in some shading export testing to take advantage of the symmetric shader writer that was added in that PR.

I ended up moving the symmetric shader reader and writer classes to `lib/mayaUsd/fileio/shading` as I was doing this. I started off without that, but things quickly started to look weird. There are no public headers installed from `lib/usd/translators`, and if we added any, we would start seeing include directives like this:

```C++
#include <mayaUsd_Translators/symmetricShaderWriter.h>
```

That seemed pretty undesirable to me. The symmetric shader reader and writer are also not translators themselves, but rather things you would use to implement translation, as we do here, so it makes more sense to me for them to live alongside the other core shading translation framework code anyway.


kxl-adsk:
@mattyjams The move to the core library makes perfect sense!

JGamache-autodesk (approved):
Good. I already have plans for the symmetric reader. Thanks!

lib/mayaUsd/fileio/shading/symmetricShaderReader.h
```diff
@@ -48,14 +48,14 @@ PXR_NAMESPACE_OPEN_SCOPE
...
-///     PxrUsdTranslators_SymmetricShaderReader::RegisterReader(
+///     UsdMayaSymmetricShaderReader::RegisterReader(
 ///         TfToken("MyUsdShaderId"),
 ///         TfToken("myMayaNodeTypeName"),
 ///         /* shadingConversionName = */ TfToken("myShadingConversionName"));
 /// };
 /// \endcode
 ///
-class PxrUsdTranslators_SymmetricShaderReader : public UsdMayaShaderReader
+class UsdMayaSymmetricShaderReader : public UsdMayaShaderReader
```
JGamache-autodesk, line 58: That file is missing a lot of "MAYAUSD_CORE_PUBLIC" for the DLL exports on windows.

lib/mayaUsd/fileio/shading/symmetricShaderWriter.h
```diff
@@ -53,14 +53,14 @@ PXR_NAMESPACE_OPEN_SCOPE
...
-///     PxrUsdTranslators_SymmetricShaderWriter::RegisterWriter(
+///     UsdMayaSymmetricShaderWriter::RegisterWriter(
 ///         TfToken("myMayaNodeTypeName"),
 ///         TfToken("MyUsdShaderId"),
 ///         /* materialConversionName = */ TfToken("myMaterialConversionName"));
 /// };
 /// \endcode
 ///
-class PxrUsdTranslators_SymmetricShaderWriter : public UsdMayaShaderWriter
+class UsdMayaSymmetricShaderWriter : public UsdMayaShaderWriter
```
JGamache-autodesk, line 63: Same here. Missing MAYAUSD_CORE_PUBLIC tags for DLL export declaration.

lib/mayaUsd/fileio/shading/symmetricShaderReader.h
mattyjams: Yup, just saw that in the failed preflight run. Fixing that up now. Thanks!

mattyjams: Should be addressed in https://github.com/Autodesk/maya-usd/pull/803/commits/6cd7599a2e4ef525f50766ad1ed161f84d62de2b. I kicked off another preflight to confirm.

lib/mayaUsd/fileio/shading/symmetricShaderWriter.h
mattyjams: Should be addressed in https://github.com/Autodesk/maya-usd/pull/803/commits/6cd7599a2e4ef525f50766ad1ed161f84d62de2b.

mattyjams:
Ok, looks like preflight is happy again after having added the export macros.

Thanks again, @JGamache-autodesk!

PR #805, opened 2020-09-28
[CMake/CTest] Use -pythonver argument for testsuite.

marsupial:
We're running into issues where running tests on the python-2 builds are invoking the python-3 version of Maya.

### DIFF
--- a/cmake/test.cmake
+++ b/cmake/test.cmake
@@ -122,7 +122,7 @@ compileMode = \\\"exec\\\"\\\; \
 globals = {\\\"__file__\\\": file, \\\"__name__\\\": \\\"__main__\\\"}\\\; \
 exec(compile(open(file, openMode).read(), file, compileMode), globals)\
 \")")
-            set(COMMAND_CALL ${MAYA_EXECUTABLE} -c ${MEL_PY_EXEC_COMMAND})
+            set(COMMAND_CALL ${MAYA_EXECUTABLE} -pythonver ${Python_VERSION_MAJOR} -c ${MEL_PY_EXEC_COMMAND})
         else()
             set(COMMAND_CALL ${MAYA_PY_EXECUTABLE} ${PREFIX_PYTHON_SCRIPT})
         endif()


HamedSabri-adsk:
@marsupial  can you please be more specific and describe the issues?  What tests are these?

marsupial:
Your test suite. Wen we build for python-2.7 you invoke **maya**, but that defaults to the python-3 version ?

mattyjams (dismissed):
Ah, I believe the change here would be relevant for any tests that use the `INTERACTIVE` option to `mayaUsd_add_test()`, which is currently only the tests in `test/lib/mayaUsd/render/pxrUsdMayaGL`. Sorry, that was probably my oversight, as I haven't yet tried running these tests against a preview release.

The modified line is the only one that will end up in a direct launch of the `maya` binary. All of the other paths through `mayaUsd_add_test()` look like they bottom out calling `MAYA_PY_EXECUTABLE`, which should be appropriately set to either `mayapy` or `mayapy2` from `FindMaya.cmake`:
https://github.com/Autodesk/maya-usd/blob/81ae65c08341d58df542d74051430d0506a2998b/cmake/modules/FindMaya.cmake#L256

The change looks good to me. Thanks @marsupial!

cmake/test.cmake
```diff
@@ -122,7 +122,7 @@ compileMode = \\\"exec\\\"\\\; \
 globals = {\\\"__file__\\\": file, \\\"__name__\\\": \\\"__main__\\\"}\\\; \
 exec(compile(open(file, openMode).read(), file, compileMode), globals)\
 \")")
-            set(COMMAND_CALL ${MAYA_EXECUTABLE} -c ${MEL_PY_EXEC_COMMAND})
+            set(COMMAND_CALL ${MAYA_EXECUTABLE} -pythonver ${Python_VERSION_MAJOR} -c ${MEL_PY_EXEC_COMMAND})
```
seando-adsk, line 125: I don't doubt that "Python_VERSION_MAJOR" is being set, but where would that be? In our python.cmake we are setting "PYTHON_VERSION_MAJOR" (all caps) here: https://github.com/Autodesk/maya-usd/blob/dev/cmake/python.cmake#L144

The -pythonver flag does not exist in Maya 2020 and earlier. Won't we need a check here for the Maya version?

I wonder if we could instead append the "-pythonver" flag in FindMaya.cmake for the variable MAYA_EXECUTABLE (for Maya 2021 and up). That way anyone using MAYA_EXECUTABLE would get the correct python version?

mattyjams: Ahh, good catch @seando-adsk! That would indeed break for Maya 2020 and earlier. Clearly I didn't try it. :/

Bundling it into `MAYA_EXECUTABLE` in FindMaya.cmake would be one option, but I wonder if instead we should just be explicitly setting the `MAYA_PYTHON_VERSION` environment variable for tests somewhere. That way `MAYA_EXECUTABLE` continues to be just a path to the executable, and earlier releases that don't understand `MAYA_PYTHON_VERSION` will just ignore it.

As far as `Python_VERSION_MAJOR`, that should be getting set by CMake's own FindPython module(s):
https://cmake.org/cmake/help/latest/module/FindPython.html#result-variables

Since the version of Python we build with can be controlled with `BUILD_WITH_PYTHON_3`, maybe we should be setting a variable (or `MAYA_PYTHON_VERSION` itself?) in FindMaya.cmake so we know which version to use for tests.

seando-adsk: Ah yes, I agree that setting `MAYA_PYTHON_VERSION` for the tests is a better idea. For the tests I would suggest in this block: https://github.com/Autodesk/maya-usd/blob/dev/cmake/test.cmake#L309

mattyjams: Cools, thanks for the pointer @seando-adsk!

@marsupial: Since I'm probably to blame for the headache here, let me know if you want to just close this PR and I can put together a new one that fixes this.

mattyjams: Hey again @marsupial, I just put up #806 which implements the other solution we were talking about here. If you have a chance, could you try running with those changes and let me know if that fixes things for you?

Thanks!

marsupial: ENV_VAR should be fine for us, just chose the cmd-line flag as it seemed 'more explicit'

marsupial:
Solution in https://github.com/Autodesk/maya-usd/pull/806 supports versions other than 2021

PR #806, opened 2020-09-30
fix Python version handling for interactive tests

mattyjams:
This should address the issue raised in #805. Since the interactive tests launch the `maya` binary rather than `mayapy` or `mayapy2`, they were always running using Python 3 in Maya 2021/preview releases.

In addition to setting `MAYA_PY_EXECUTABLE` to either `mayapy` or `mayapy2` in `FindMaya.cmake`, we now also set `MAYA_PY_VERSION` to either `2` or `3` as appropriate. We then use that CMake variable to set `MAYA_PYTHON_VERSION` when setting up the environment for running the tests. Maya 2020 and earlier will simply ignore that environment variable.

I've verified that the interactive tests pass running Maya PR 119 and Python 2 on Linux. I'm not actually setup to build against Python 3 yet, so if someone is able to verify that configuration for me, that would be great. These tests get skipped by the preflight builds, so any issues likely would not show up there.

### DIFF
--- a/cmake/modules/FindMaya.cmake
+++ b/cmake/modules/FindMaya.cmake
@@ -253,9 +253,12 @@ if(MAYA_INCLUDE_DIRS AND EXISTS "${MAYA_INCLUDE_DIR}/maya/MTypes.h")
     endif()
 endif()
 
-# swtich between mayapy and mayapy2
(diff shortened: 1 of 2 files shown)


seando-adsk (approved):
Looks good to me.

PR #810, opened 2020-10-01
always register and de-register the PxrMayaHdImagingShape and its draw override

mattyjams:
While the USD proxy shape classes included in the maya-usd repo all use either the (legacy) Pixar batch renderer *or* the Viewport 2.0 render delegate, there may be other custom node types that may or may not leverage USD but still want to take advantage of aggregated drawing using Hydra.

As an example, internally at Pixar we have a custom Maya shape node that is a severely stripped down approximation of a `UsdGeomPointInstancer`. This shape maintains its own data structures as a mix of USD and non-USD data, but it still uses the `PxrMayaHdImagingShape` and the Pixar batch renderer for drawing. I'm not sure whether other studios have custom shape nodes that leverage the Pixar batch renderer and its shape adapter system in this way, but it was definitely written with the intention of supporting that.

In our case, the eventual goal is that once we've fully transitioned to the Viewport 2.0 render delegate and UFE, we'll be authoring `UsdGeomPointInstancer` prims directly under a `mayaUsdProxyShape` and will no longer need this custom node type. As a result, we don't see value putting effort into making the custom shape work with the Viewport 2.0 render delegate. Instead, we're hoping that it can continue to use the Pixar batch renderer as we transition, and then we'll retire the shape once we've switched.

I don't think there should be any ill effects from always registering and de-registering `PxrMayaHdImagingShape` and its draw override. If no nodes are created in the scene that call `PxrMayaHdImagingShape::GetOrCreateInstance()` in their `postConstructor()` override, we shouldn't see any imaging nodes created and the Pixar batch renderer shouldn't get instantiated either.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapePlugin.cpp
+++ b/lib/mayaUsd/nodes/proxyShapePlugin.cpp
@@ -102,24 +102,28 @@ MayaUsdProxyShapePlugin::initialize(MFnPlugin& plugin)
             MayaUsdProxyShapeBase::displayFilterLabel,
             UsdMayaProxyDrawOverride::drawDbClassification);
         CHECK_MSTATUS(status);
-
-        // Hybrid Hydra / VP2 rendering uses the PxrMayaHdImagingShape for draw
-        // aggregation of all proxy shapes.
-        status = plugin.registerShape(
-            PxrMayaHdImagingShape::typeName,
-            PxrMayaHdImagingShape::typeId,
-            PxrMayaHdImagingShape::creator,
-            PxrMayaHdImagingShape::initialize,
-            PxrMayaHdImagingShapeUI::creator,
-            &PxrMayaHdImagingShapeDrawOverride::drawDbClassification);
-        CHECK_MSTATUS(status);
-
-        status = MHWRender::MDrawRegistry::registerDrawOverrideCreator(
-            PxrMayaHdImagingShapeDrawOverride::drawDbClassification,
-            _RegistrantId,
-            PxrMayaHdImagingShapeDrawOverride::creator);
-        CHECK_MSTATUS(status);
     }
+
+    // We register the PxrMayaHdImagingShape regardless of whether the Viewport
+    // 2.0 render delegate is enabled for the USD proxy shape node types. There
+    // may be other non-proxy shape node types in use that still want to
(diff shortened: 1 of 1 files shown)


mattyjams:
To put it another way, right now enabling the Viewport 2.0 render delegate for USD proxy shape node types effectively disables the Pixar batch renderer for **all** shape node types, not just USD proxies. We'd like to still be able to use the batch renderer for non-proxy shape types to smooth the transition.

ppt-adsk (approved):

HdC-adsk:
I am not familiar with the `PxrMayaHdImagingShape` class. If we enable the congregated draw for this node type, will it introduce duplicated drawing for the same objects from both VP2RenderDelegate and the draw override you enable in this commit?

mattyjams:
> I am not familiar with the `PxrMayaHdImagingShape` class. If we enable the congregated draw for this node type, will it introduce duplicated drawing for the same objects from both VP2RenderDelegate and the draw override you enable in this commit?

@HdC-adsk: Good question. No, we should be ok there. We effectively have two layers of protection against that.

The proxy shape classes guard against even creating an instance of the `PxrMayaHdImagingShape` when we're using the Viewport 2.0 render delegate:
https://github.com/Autodesk/maya-usd/blob/29b37a61c4f8369e07026bb04c7db62b489d35e6/plugin/adsk/plugin/ProxyShape.cpp#L64
https://github.com/Autodesk/maya-usd/blob/9090482a0293677f9192c5a6fdaa308a86208b69/plugin/pxr/maya/lib/usdMaya/proxyShape.cpp#L318

So if none of these custom node types are in play, we really shouldn't end up with a `PxrMayaHdImagingShape` node in the scene when the Viewport 2.0 render delegate is enabled.

Even if we do have a `PxrMayaHdImagingShape` node in the scene, we're still not registering the legacy Pixar batch renderer `UsdMayaProxyDrawOverride` for proxy shapes when the Viewport 2.0 render delegate is enabled, so no proxy shape rprims would get populated in the batch renderer, and there would be nothing for it to draw.

HdC-adsk (approved):

HdC-adsk:
No worry about the preflight build errors. They are relevant to a new API class we recently added and will fix it soon.

Issue #366, opened 2020-03-19
[MAYA-102675] userExportedAttributesUI | cannot import name QStringListModel


mhamid3d:
The ```userExportedAttributesUI``` python file is trying to import ```QStringListModel``` from ```PySide2.QtGui```. On PySide2 I found that this class in fact lives in ```QtCore``` instead.

1. Open any version of Maya using PySide2, open Script Editor
2. Run the following code in a python tab:
```
from pxr.UsdMaya import userExportedAttributesUI
```

This code yields the following error:

```# Error: ImportError: file /.../plugin/pxr/lib/python/pxr/UsdMaya/userExportedAttributesUI.py line 52: cannot import name QStringListModel # ```

SPECS -
 - Linux CentOs 7 x64
 - Maya 2020

santosd:
I was able to reproduce this in PR114 with the Pixar plug. I think this one is definitely specific to the pxr plug-in. Maya USD doesn't contain userExportedAttributesUI from what I can see. Will speak with the team to see what we can do about this issue.

santosg87:
it's worth giving a try on this with an older build that hasn't updated Qt to confirm.

mattyjams:
Hi @mhamid3d, thanks for reporting this, and sorry it has languished so long. It somehow managed to slip by me earlier this year, and we've not yet tried to use this UI with Maya 2020+ which is where it starts to appear, so we hadn't tripped over it yet.

I just put up #812 which should address this.

mattyjams:
This should be resolved in the dev branch with the merge of #812.

Thanks!

PR #812, opened 2020-10-02
handle namespace migration of QStringListModel between Maya 2019 and Maya 2020 in userExportedAttributesUI

mattyjams:
Maya 2020 and later use Qt/PySide2 5.12.5, where `QStringListModel` is exported to Python with QtCore, which is also where it lives in C++. Maya 2019 and earlier use Qt 5.6.1 and PySide2 2.0.0-alpha, where `QStringListModel` was incorrectly exported with QtGui.

See this bug for more detail:
https://bugreports.qt.io/browse/PYSIDE-614

This change addresses #366. I also took this opportunity to remove support for Maya 2016 and earlier.


seando-adsk (approved):
Looks good to me.

Issue #769, opened 2020-09-10
[MAYA-106867] Discarding Edits Can Crash Maya


wjwilson:
**Describe the bug**
Discarding edits on a USD layer can crash Maya if it leads to a change in the hierarchy. In particular, if it leads to a reference being removed.

**Steps to reproduce**
1. Create a USD stage with at least one prim in it. I'm loading a simple USD file that looks like this:
```
#usda 1.0

def Xform "Xform1"
{
}
````
2. Reference a second USD file into the stage. I'm using a file that looks like this:
```
#usda 1.0
(
    defaultPrim = "top"
)

def Xform "top"
{
    def Cube "cube"
    {
    }
}

```
3. The viewport doesn't seem to redraw automatically so do something to jiggle it (click in the outliner or viewport, so long as the geometry becomes visible)
4. In the USD Layer Editor right-click and Discard Edits on whichever layer is active.
5. The geometry will disappear from the viewport, but the reference will still show in the hierarchy in the outliner.
6. Do something to jiggle the update again (click in the outliner, viewport, change selection, etc)
7. Maya will crash

**Expected behavior**
On clicking Discard Edits, USD stage will revert to the state on disk. The outliner and viewport will reflect this. Maya won't crash.

**Specs (if applicable):**
 - Windows 10
 - Maya PR118
 - Maya USD 0.3.0 bundled with PR
 - Pixar USD 20.08 bundled with PR

**Additional context**
I've crashed Maya about 300 times in the last 24 hours and sent a CER each time, if the information there is useful then CER_226117302 is one of them.

santosg87:
thanks for logging this! we have it logged internally as well. this will add more info to help us narrow down the issue. :)

santosd:
Hi @wjwilson, This has been resolved and the fix should be in the next PR.

PR #817, opened 2020-10-05
MAYA-106867 [Github #769] Crash after discarding edits on an anon layer

luceric-adsk:


lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -258,7 +258,13 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
 #ifdef UFE_V2_FEATURES_AVAILABLE
 		else if (!prim.IsValid() && !InPathChange::inPathChange())
 		{
-			if (InAddOrDeleteOperation::inAddOrDeleteOperation())
```
ppt-adsk, line 261: Hmm, this is quite strange, and is reversing the logic.  The idea was that we generate a SubtreeInvalidate notification unless we know better, i.e. unless we know we're in an add or delete operation (here delete, I would hope).  Therefore, SubtreeInvalidate is the default, and delete will only be sent if the code implementing the delete has put in a delete guard.

luceric-adsk: Thanks for your review @ppt-adsk . I hope I didn't mess up the logic. What's happening is that we are getting a delete notification, but we did not initiate this delete and therefore inAddOrDeleteOperation is false.  The deletion was caused by a reload() on a layer.  sceneItem will be remain null if InaddOrDeleteOperation is true, so it should be passing to the ObjectDestroyed notification as before.  createItem will return nullptr in the case of this deletion notification, and therefore it will also go to the ObjectDestroyed notification.

seando-adsk: I would just move the createItem above the original line 261 and then change 261 to

`if (!sceneItem || InAddOrDeleteOperation::inAddOrDeleteOperation())`

seando-adsk: @ppt-adsk The case we are trying to catch here is when the prim was deleted by outside sources and not going thru our delete operations. In this case since the prim is deleted, then createItem() returns nullptr which is then packaged up in the SubtreeInvalidate message. The Outliner gets it, expects a valid item to invalidate and crashes because its null.

seando-adsk (approved):

PR #818, opened 2020-10-06
clean up pxr CMake and remove irrelevant build options

mattyjams:
There was a bunch of junk still hanging around in the pxr CMake setup that was leftover from the merge in from the core USD repo. The pxr `CXXDefaults.cmake` wasn't being included by anything anymore, so that allowed the removal of everything in `plugin/pxr/cmake/defaults`.

There's more that could be cleaned up here, but this is a start at removing the cruft.


HamedSabri-adsk (approved):
Lgtm! Thank for doing this.

PR #819, opened 2020-10-06
MAYA-102621 don't snap to active objects when using point snapping

HdC-adsk:


lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -1222,7 +1224,9 @@ HdVP2BasisCurves::_UpdateDrawItem(
                          DirtySelectionHighlight))) {
         bool enable = drawItem->GetVisible() && !_curvesSharedData._points.empty() && !instancerWithNoInstances;
```
williamkrick, line 1225: What does enable get used for? It looks like we set it and then never use it?

williamkrick: Ah sorry, I didn't have the code fully expanded.

williamkrick (approved):

PR #821, opened 2020-10-07
MAYA-107249 When excluded prims changes we don't know if some prims were removed 

williamkrick:
from it and so we don't know to bring them back. The easy solution is to reset everything when excluded prims changes, that gives us the correct result slowly. We don't have any workflows built on top of this right now so leaving the performance issue in the backlog.

In addition to fixing the defect I also cleaned up how we understand changes to the proxy shape in the render delegate. I've tried to concentrate all the checks for invalid data into a new function _ClearInvalidData. Then the rest of the code after that (_InitRenderDelegate, _Populate, etc) doesn't directly check the _proxyShapeData for changes. Instead it looks at the data it is supposed to generate and does work based on that (similar to how it works on the first frame when we create the proxyRenderDelegate). I think that will help keep the logic controlling what actually happens when proxy shape attributes change simpler.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -366,32 +366,46 @@ void ProxyRenderDelegate::_ClearRenderDelegate()
     _renderIndex.reset();
     _renderDelegate.reset();
 
+    _dummyTasks.clear();
+
     // reset any version ids or dirty information that doesn't make sense if we clear
     // the render index.
     _renderTagVersion = 0;
     _visibilityVersion = 0;
     _taskRenderTagsValid = false;
+    _isPopulated = false;
 }
 
-//! \brief  One time initialization of this drawing routine
-void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -475,28 +489,29 @@ void ProxyRenderDelegate::_InitRenderDelegate(MSubSceneContainer& container) {
...
     if (!_isInitialized())
         return false;
 
-    if (_proxyShapeData->UsdStage() && (!_isPopulated || !_proxyShapeData->IsUsdStageUpToDate() || !_proxyShapeData->IsExcludePrimsUpToDate()) ) {
+    if (_proxyShapeData->UsdStage() && !_isPopulated ) {
         MProfilingScope subProfilingScope(HdVP2RenderDelegate::sProfilerCategory,
             MProfiler::kColorD_L1, "Populate");
 
-        // It might have been already populated, clear it if so.
+        // Remove any excluded prims before populating
```
williamkrick, line 503: I considered moving this code out of _Populate as well, because it is modifying the render index rather than populating, but I didn't have a strong enough reason to do it so I left it for now.

kxl-adsk (approved):

PR #822, opened 2020-10-07
MAYA-102319 - Material import options rework

### LINKED ISSUE #520: primvars:displayColor always included in export regardless of shadingMode
**Describe the bug**
My understanding from https://github.com/Autodesk/maya-usd/blob/dev/plugin/pxr/doc/README.md is that specifying shadingMode=none for USD export should skip any shading and displayColor export.
From my testing it appears that primvars:displayColor is always exported, deriving values from assigned shaders.
Specifying shadingMode as an export option does appear to result in any material definitions being excluded from the export, but it's not clear what value to pass to that option in order to not export displayColor.

**Steps to reproduce**
1. Create sphere
2. Assign lambert shader with obvious 'color' value
3. Select sphere and run export command
cmds.file("out.usd",
  force=True,
  typ="USD Export",
  exportSelected=True,
  options="shadingMode=none",
)
4. Inspect the resulting usd

**Expected behavior**
Specifying "shadingMode=none" as an export option should not export displayColor primvars.

JGamache-autodesk:
The major item of this work is to remove the material import dropdown in favor of a single checkbox for importing materials or not. The strategy is to provide a list of shadingModes to try in the import options and let the code execute each one in sequence until a material is found.

Major work:
The shadingMode import option is now a list of shading modes to try.
  - Added code to extract array of tokens from options string.
  - Added code to generate options string containing arrays
  - The mayaUSDListShadingModes command will list the modes in a predefined order that tries more complex shading modes first.
  - The pxrRis shadingMode import has been made more strict to let other shading modes succeed.
  - Shading modes that declare a renderContext will no longer explore the universal context. This allow universal context shading modes importers to successfully run.

Added materialConversion support for import modes. This mirrors the export mode exactly.
  - Added registry functions to register an importer with materialConversion.
  - Shares and uses the renderContext concept used for export.
  - Renamed shadingConversion to preferredMaterial to prevent confusion. This represents an optional "convertTo" that can be added on top of the "convertFrom" that materialConversion represents.

Other work:
The displayColor shading mode is removed in favor of a better handling of the exportDisplayColor geometry option. This fixes #520.
  - The exportDisplayColor now default to false
  - The exportOptions MEL script has been reworked to still present a Display Color material export that toggles the exportDisplayColor import option.

Minor work:
- SetPlugValue can now handle 4x4 matrices
- UsdPreviewSurface added to the list of preferred materials on import. Support for displayColor import/export added.
- PhongE removed from list of preferred materials on import.
- Symmetric shader writer now skips hidden attributes.
- vp2RenderDelegate will convert linear displayColor to Maya to provide better color match between the original lambert and the displayColor.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
@@ -55,7 +55,7 @@ namespace
 	template<typename T>
 	void addMetadataCount(const T& op, std::string& tooltip, bool& needComma, const std::string& singular, const std::string& plural)
 	{
-		T::ItemVector refs;
+		typename T::ItemVector refs;
 		op.ApplyOperations(&refs);
 		if (!refs.empty()) {
 			addMetadataStrings(refs.size(), tooltip, needComma, singular, plural);


lib/mayaUsd/commands/baseImportCommand.cpp
```diff
@@ -44,9 +44,10 @@ MayaUSDImportCommand::createSyntax()
     // UsdMayaJobImportArgs::GetDefaultDictionary.
     syntax.addFlag(kShadingModeFlag,
                    UsdMayaJobImportArgsTokens->shadingMode.GetText(),
-                   MSyntax::kString);
-    syntax.addFlag(kShadingConversionFlag,
-                   UsdMayaJobImportArgsTokens->shadingConversion.GetText(),
+                   MSyntax::kString, MSyntax::kString);
+    syntax.makeFlagMultiUse(kShadingModeFlag);
```
JGamache-autodesk, line 48: This option is now multi (array) and takes pairs of strings (shadingMode, materialConversion)

lib/mayaUsd/commands/baseListShadingModesCommand.cpp
```diff
@@ -75,41 +82,59 @@ MayaUSDListShadingModesCommand::doIt(const MArgList& args) {
...
     if (argData.isFlagSet("export")) {
-        // We have these default exporters which are always 1-2-3 in the options:
-        appendToResult(UsdMayaShadingModeRegistry::GetExportConversionInfo(
+        // We have these default exporters which are always 1-2 in the options:
+        appendToResult(UsdMayaShadingModeRegistry::GetMaterialConversionInfo(
                            UsdImagingTokens->UsdPreviewSurface)
                            .niceName.GetText());
-        appendToResult(
-            UsdMayaShadingModeRegistry::GetExporterNiceName(UsdMayaShadingModeTokens->displayColor)
-                .c_str());
```
JGamache-autodesk, line 84: displayColor is no longer a shadingMode export option. We still allow it via the exportDisplayColor geom export option.

lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -466,7 +511,7 @@ UsdMayaJobExportArgs::GetDefaultDictionary()
         d[UsdMayaJobExportArgsTokens->eulerFilter] = false;
         d[UsdMayaJobExportArgsTokens->exportCollectionBasedBindings] = false;
         d[UsdMayaJobExportArgsTokens->exportColorSets] = true;
-        d[UsdMayaJobExportArgsTokens->exportDisplayColor] = true;
+        d[UsdMayaJobExportArgsTokens->exportDisplayColor] = false;
```
JGamache-autodesk, line 514: Defaults to false, but can become true when the material export option is "Display Colors" (see mayaUsdTranslatorExport.mel for details).

lib/mayaUsd/fileio/shading/shadingModePxrRis.cpp
```diff
@@ -415,6 +432,16 @@ _GetOrCreateShaderObject(
...
 
+    if (shadingNodeType == UsdMayaShadingNodeType::Shader) {
+        // We specifically reload only Pxr type shaders and let other importers
+        // work on unsupported nodes:
+        TfToken shaderId;
+        shaderSchema.GetIdAttr().Get(&shaderId);
+        if (_GetMayaTypeNameForShaderId(shaderId, false).IsEmpty()) {
+            return shaderObj;
+        }
+    }
```
JGamache-autodesk, line 443: This still allows all utility nodes to be Maya nodes. Only surface shaders are checked here.

lib/mayaUsd/fileio/shading/symmetricShaderReader.cpp
```diff
@@ -84,10 +84,10 @@ UsdMayaSymmetricShaderReader::RegisterReader(
 UsdMayaShaderReader::ContextSupport
 UsdMayaSymmetricShaderReader::CanImport(
         const UsdMayaJobImportArgs& importArgs,
-        const TfToken& shadingConversionName)
+        const TfToken& materialConversion)
 {
-    if (shadingConversionName.IsEmpty() ||
-            importArgs.shadingConversion == shadingConversionName) {
+    if (materialConversion.IsEmpty()
+        || importArgs.shadingModes.front().second == materialConversion) {
```
JGamache-autodesk, line 90: Compares against the shading mode import option currently being considered.

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
```diff
@@ -71,15 +70,22 @@ UsdMayaTranslatorMaterial::Read(
...
 
-    if (UsdMayaShadingModeImporter importer =
-            UsdMayaShadingModeRegistry::GetImporter(jobArguments.shadingMode)) {
-        shadingEngine = importer(&c, jobArguments);
-    }
+    UsdMayaJobImportArgs localArguments = jobArguments;
+    for (const auto& shadingMode : jobArguments.shadingModes) {
+        if (shadingMode.first == UsdMayaShadingModeTokens->none) {
+            break;
+        }
```
JGamache-autodesk, line 77: Stop searching when "None" option is encountered.

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
```diff
@@ -71,15 +70,22 @@ UsdMayaTranslatorMaterial::Read(
...
+            = UsdMayaShadingModeRegistry::GetImporter(shadingMode.first)) {
+            localArguments.shadingModes = UsdMayaJobImportArgs::ShadingModes(1, shadingMode);
+            shadingEngine = importer(&c, localArguments);
+        }
 
-    if (!shadingEngine.isNull()) {
-        c.AddCreatedObject(shadeMaterial.GetPrim(), shadingEngine);
+        if (!shadingEngine.isNull()) {
+            c.AddCreatedObject(shadeMaterial.GetPrim(), shadingEngine);
+            return shadingEngine;
```
JGamache-autodesk, line 86: Stop searching when an importer succeeds in crating a shading engine.
Does not work with importers that could share existing shading engines by adding specialized surface ports (RenderMan  is a known example). To be added later.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1013,7 +1014,7 @@ void HdVP2Mesh::_UpdateDrawItem(
                 // Use fallback shader if there is no material binding or we
                 // failed to create a shader instance from the material.
                 if (!stateToCommit._shader) {
-                    const GfVec3f& color = colorArray[0];
+                    const GfVec3f& color = UsdMayaColorSpace::ConvertLinearToMaya(colorArray[0]);
```
JGamache-autodesk, line 1017: @HamedSabri-adsk Do you think this is a good idea? Should I extend it to the other supported geometries?

lib/mayaUsd/utils/util.cpp
```diff
@@ -869,16 +870,14 @@ _GetColorAndTransparencyFromDepNode(
...
-    }
-    MPlug transparencyPlug = d.findPlug("transparency", true, &status);
-    if (!status) {
-        return false;
-    }
-
     if (rgb) {
+        MPlug colorPlug = d.findPlug("color", true, &status);
+        if (!status) {
+            colorPlug = d.findPlug("diffuseColor", true, &status);
```
JGamache-autodesk, line 876: For UsdPreviewSurface support.

lib/mayaUsd/utils/util.cpp
```diff
@@ -1590,7 +1599,23 @@ UsdMayaUtil::setPlugValue(
             }
         }
     }
-    else if (val.IsHolding<VtDoubleArray>()) {
+    else if (val.IsHolding<GfMatrix4d>()) {
```
JGamache-autodesk, line 1602: Found by noticing a message from the TF_CODING_ERROR("Unsupported type..." ) from line 1863 in the log of the RenderMan import test. This error message was not sufficient to cause the test to fail though. Should we investigate if that is possible?

lib/mayaUsd/utils/util.cpp
```diff
@@ -1953,35 +1978,143 @@ UsdMayaUtil::GetDictionaryFromArgDatabase(
...
+        // To prevent quoting issues in MEL scripts, we expect a string that is
+        // completely unquoted, containing an array of booleans, tokens, or
+        // other arrays:
+        //       [0,1,true,false]
+        //       [none,default]
+        //       [[useRegistry,UsdPreviewSurface],[displayColors,default]]
+        // So, to be able to parse this as valid JSON, we need to add quotes at
+        // each "[[," to alphanumeric transition. 
+        // NOTE: The array contains *tokens*, not freeform strings. So we can
+        //       skip spaces and consider all commas to be element separators.
```
JGamache-autodesk, line 2033: This might be too fragile. One option would be to make the options string true JSON, but since it is parsed by MEL on the Maya side, this will mean escaping a lot of double quotes.

lib/mayaUsd/utils/util.cpp
```diff
@@ -1953,35 +1978,143 @@ UsdMayaUtil::GetDictionaryFromArgDatabase(
...
+    } else {
         TF_CODING_ERROR("Unknown flag '%s'", key.c_str());
     }
 
     return VtValue();
 }
 
+std::pair<bool, std::string>
+UsdMayaUtil::ValueToArgument(const VtValue& value) {
+    if (value.IsHolding<bool>()) {
```
JGamache-autodesk, line 2088: Provide a function to do the reverse of the parsing. Helps with symmetry.

test/lib/usd/translators/UsdImportCustomConverterTest/UsdImportCustomConverterTest.usda
```diff
@@ -0,0 +1,182 @@
+#usda 1.0
+(
+    defaultPrim = "pCube1"
+    metersPerUnit = 0.01
+    upAxis = "Y"
+)
+
```
JGamache-autodesk, line 7: Provides a mix of materials to test import scenarios (display color, preview surface, maya dump).

test/lib/usd/translators/testUsdExportShadingInstanced.py
```diff
@@ -39,7 +39,7 @@ def setUpClass(cls):
 
         usdFilePath = os.path.abspath('InstancedShading.usda')
         cmds.usdExport(mergeTransformAndShape=True, file=usdFilePath,
-                shadingMode='displayColor', exportInstances=True,
+                shadingMode='useRegistry', exportInstances=True,
```
mattyjams, line 42: Should `shadingMode` get a list value here?

test/lib/usd/translators/testUsdExportShadingInstanced.py
```diff
@@ -53,7 +53,7 @@ def setUpClass(cls):
 
         usdFilePath = os.path.abspath('NestedInstancedShading.usda')
         cmds.usdExport(mergeTransformAndShape=True, file=usdFilePath,
-                shadingMode='displayColor', exportInstances=True,
+                shadingMode='useRegistry', exportInstances=True,
```
mattyjams, line 56: Should `shadingMode` get a list value here?

test/lib/usd/translators/testUsdImportCustomConverter.py
```diff
@@ -0,0 +1,182 @@
...
+                self.assertTrue(shaders[0] in self.default_materials)
+
+    def testImportCommand(self):
+        """
+        Tests a custom exporter for a conversion that exists in an unloaded
+        plugin using the mayaUSDImport command.
+        """
+        modes = [["useRegistry", "maya"],
+                 ["useRegistry", "UsdPreviewSurface"],
+                 ["displayColor", "default"]]
```
mattyjams, line 71: Is `default` significant here, or is it just ignored?

test/lib/usd/translators/testUsdImportCustomConverter.py
```diff
@@ -0,0 +1,182 @@
...
+        cmds.mayaUSDImport(file=self.usd_path, shadingMode=modes,
+                           preferredMaterial="none", primPath="/")
```
mattyjams, line 73: Shouldn't need `primPath="/"`?

test/lib/usd/translators/testUsdImportCustomConverter.py
```diff
@@ -0,0 +1,182 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from pxr import Usd
+from pxr import UsdShade
+
+import json
```
mattyjams, line 21: Don't think `json` is used in this test script?

test/lib/usd/translators/testUsdImportCustomConverter.py
```diff
@@ -0,0 +1,182 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from pxr import Usd
+from pxr import UsdShade
```
mattyjams, line 19: Looks like all the validation is done on the Maya side, so I think these imports can be removed.

test/lib/usd/translators/UsdImportRfMShadersTest/MarbleCube.usda
```diff
@@ -40,13 +40,7 @@ def Xform "MarbleCube" (
...
         def Material "MarbleCubeSG"
         {
-            # XXX: Note that the rendermanForMaya material conversion at export
-            # will create this surface terminal output using the "ri"
-            # renderContext as opposed to using the universal renderContext as
-            # we do here. Further work is needed in the "useRegistry"
-            # shadingMode to be able to support finding materials in
-            # renderContexts other than universal.
-            token outputs:surface.connect = </MarbleCube/Materials/MarbleCubeSG/MarbleLambert.outputs:surface>
+            token outputs:ri:surface.connect = </MarbleCube/Materials/MarbleCubeSG/MarbleLambert.outputs:surface>
```
mattyjams, line 43: Thanks for patching this up!

mattyjams (approved):
Looks good to me! I had a few minor notes, but nothing too critical.

I have to admit that just reading through the code, it's a little confusing to understand how the conversion aspect of it works, but I'm hoping that becomes clearer once we start playing with it.

lib/mayaUsd/fileio/shading/symmetricShaderWriter.cpp
```diff
@@ -115,8 +115,9 @@ UsdMayaSymmetricShaderWriter::UsdMayaSymmetricShaderWriter(
     for (unsigned int i = 0u; i < depNodeFn.attributeCount(); ++i) {
         const MObject attrObj = depNodeFn.reorderedAttribute(i);
         MPlug attrPlug = depNodeFn.findPlug(attrObj, true);
+        MFnAttribute attrFn(attrObj);
```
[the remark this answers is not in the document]
JGamache-autodesk: evaluateNumElements is not const, so the MPlug needs to stay as is,

test/lib/usd/translators/testUsdImportCustomConverter.py
JGamache-autodesk: It is ignored for "displayColor" shading mode importer.
I do not like passing an unused argument, but I feel this is only transitional: in a future where we keep only the "useRegistry" importer, we can remove the first argument and use only a list of materialConversion instead of a list of {mode, conversion}.

PR #827, opened 2020-10-13
Prepare for upcoming UFE interface changes to support Maya transform …

ppt-adsk:
…stack.

The changes are mostly minor renames, to conform to upcoming UFE 0.2.25 interfaces.

### DIFF
--- a/lib/mayaUsd/ufe/UsdRotatePivotTranslateUndoableCommand.h
+++ b/lib/mayaUsd/ufe/UsdRotatePivotTranslateUndoableCommand.h
@@ -61,6 +61,7 @@ public:
     void undo() override;
     void redo() override;
 #if UFE_PREVIEW_VERSION_NUM >= 2025
+//#ifdef UFE_V2_FEATURES_AVAILABLE
     bool set(double x, double y, double z) override;
 #else
     bool translate(double x, double y, double z) override;
--- a/lib/mayaUsd/ufe/UsdRotatePivotTranslateUndoableCommand.h
+++ b/lib/mayaUsd/ufe/UsdRotatePivotTranslateUndoableCommand.h
@@ -1,5 +1,5 @@
 //
-// Copyright 2019 Autodesk
+// Copyright 2020 Autodesk
 //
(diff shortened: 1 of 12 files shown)


kxl-adsk (approved):
LGTM!

lib/mayaUsd/ufe/UsdRotatePivotTranslateUndoableCommand.h
```diff
@@ -60,7 +60,11 @@ class MAYAUSD_CORE_PUBLIC UsdRotatePivotTranslateUndoableCommand : public Ufe::T
     // Ufe::TranslateUndoableCommand overrides
     void undo() override;
     void redo() override;
+#if UFE_PREVIEW_VERSION_NUM >= 2025
+    bool set(double x, double y, double z) override;
+#else
     bool translate(double x, double y, double z) override;
+#endif
```
seando-adsk, line 67: Looking at UFE v1 I see the translate method exists. So I think you need more than just a UFE_PREVIEW_VERSION_NUM check here. Because later when we cleanup and remove these (because Maya PR has this version) we will still need an ifdef to support the UFE v1 version with translate (not set) method.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -55,7 +55,7 @@ namespace
 	template<typename T>
 	void addMetadataCount(const T& op, std::string& tooltip, bool& needComma, const std::string& singular, const std::string& plural)
 	{
-		T::ItemVector refs;
+		typename T::ItemVector refs;
```
seando-adsk, line 58: Thanks for fixing this.

lib/mayaUsd/ufe/UsdRotatePivotTranslateUndoableCommand.h
seando-adsk: Great work thanks. I think we should use this same pattern moving forwards. It will be a great help when we come back and cleanup this UFE preview checks.

seando-adsk (approved):

PR #836, opened 2020-10-16
add translation support between lambert transparency and UsdPreviewSurface opacity

mattyjams:
We have some internal test cases that were previously using the `displayColor` shadingMode, and in switching them over to use the new round robin-style shadingMode handling from #822 instead, I noticed that `lambert` shaders were not having their `transparency` attribute properly round-tripped out to USD as `opacity` and then back.

We could definitely use some better unit testing in this area, but I wanted to get the fix for this up sooner than later.

### DIFF
--- a/lib/usd/translators/shading/usdLambertReader.cpp
+++ b/lib/usd/translators/shading/usdLambertReader.cpp
@@ -22,6 +22,7 @@
 #include <mayaUsd/fileio/utils/readUtil.h>
 #include <mayaUsd/utils/util.h>
 
+#include <pxr/base/gf/vec3f.h>
 #include <pxr/base/tf/diagnostic.h>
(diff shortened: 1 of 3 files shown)


mattyjams:
Sorry, I rebased https://github.com/Autodesk/maya-usd/pull/836/commits/2069976381a13e1ba812b959f192b12adf6ef425 as https://github.com/Autodesk/maya-usd/pull/836/commits/bc1fe10f6eeb4e970e8199577a1f961e4766a182 just to fix a commit message typo. No code changes.

JGamache-autodesk (approved):
Thanks! I was not sure if we wanted to export that parameter since it would not work with connected image nodes, but for scalar values this will work fine.

PR #841, opened 2020-10-16
Devtoolset-9 fixes

JGamache-autodesk:
GCC 9 will warn if using memcopy to overwrite the memory area of a non-POD class. The warning is correct, but in these cases, we know that these classes are POD so we can proceed and indicate the intent with a void* cast.

### DIFF
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
@@ -579,7 +579,7 @@ UsdMayaMeshWriteUtils::exportReferenceMesh(UsdGeomMesh& primSchema, MObject obj)
     const int numVertices = referenceMesh.numVertices();
     VtVec3fArray points(numVertices);
 
-    memcpy(points.data(), mayaRawPoints, numVertices * sizeof(float) * 3);
+    memcpy((void*)points.data(), mayaRawPoints, numVertices * sizeof(float) * 3);
 
     UsdGeomPrimvar primVar = primSchema.CreatePrimvar(
         UsdUtilsGetPrefName(),
@@ -687,7 +687,7 @@ UsdMayaMeshWriteUtils::writePointsData(const MFnMesh& meshFn,
     }
 
     // use memcpy() to copy the data. HS April 09, 2020
-    memcpy((GfVec3f*)points.data(), pointsData, sizeof(float) * 3 * numVertices);
+    memcpy((void*)points.data(), pointsData, sizeof(float) * 3 * numVertices);
 
     VtVec3fArray extent(2);
     // Compute the extent using the raw points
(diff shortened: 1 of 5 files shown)


plugin/al/plugin/AL_USDMayaTestPlugin/test_translators_NurbsCurveTranslator.cpp
```diff
@@ -66,15 +66,15 @@ MObject createNurbStage(bool useSingleWidth=false)
   VtVec3fArray points = VtVec3fArray(5);
   for(uint32_t i = 0 ; i< pointsa.size(); ++i)
   {
-    memcpy(&points[i], &pointsa[i], 3*sizeof(float));
+    memcpy((void*)&points[i], &pointsa[i], 3*sizeof(float));
```
mattyjams, line 69: Same suggestion as above, but here I think I'd also remove the loops and include the array size in the range being copied?

HamedSabri-adsk (approved):
Thanks @JGamache-autodesk . Lgmt!

mattyjams (changes requested):
I'm not a big fan of the `void*` cast, so I had an alternate suggestion that might work. I can't actually try it myself though since I'm not setup to build with `gcc` 9 at the moment.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -579,7 +579,7 @@ UsdMayaMeshWriteUtils::exportReferenceMesh(UsdGeomMesh& primSchema, MObject obj)
     const int numVertices = referenceMesh.numVertices();
     VtVec3fArray points(numVertices);
 
-    memcpy(points.data(), mayaRawPoints, numVertices * sizeof(float) * 3);
+    memcpy((void*)points.data(), mayaRawPoints, numVertices * sizeof(float) * 3);
```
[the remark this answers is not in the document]
JGamache-autodesk: Using cdata() on the std::vector turns it into a GfVec3f C pointer, then using ->data would turn it into a full C float pointer. I do not see how the C++ compiler could optimize std::copy given raw C pointers.
Can we meet in the middle? By reshaping the source raw float* into a GfVec3f pointer, then we can use an optimized C++ constructor and skip the zero-fill we had previously.

plugin/al/plugin/AL_USDMayaTestPlugin/test_translators_NurbsCurveTranslator.cpp
JGamache-autodesk: Yes, I think we can skip the intermediates and the copy as well.

mattyjams (approved):
Looks good to me!

HamedSabri-adsk (approved):
I like this better. Thanks.

PR #842, opened 2020-10-16
Clang format fixes

pmolodo:
A few clang-format related tweaks, in preparation of running on whole repo:

- made line break before open-brace for classes (https://github.com/Autodesk/maya-usd/pull/690#discussion_r505968056)
- replaced `MAYAUSD_NS_DEF` with `namesapce MAYAUSD_NS` (https://github.com/Autodesk/maya-usd/pull/690#discussion_r505949492)
- protected some nicely-indented #if formatting in lib/mayaUsd/base/api.h


ppt-adsk:
Hey @pmolodo, thanks for doing this.  My only concern is with replacing MAYAUSD_NS_DEF with MAYAUSD_NS.  You've expanded out the namespace token, which is the right thing to do for clang-format.  However, MAYAUSD_NS is the namespace that is unversioned, while MAYAUSD_VERSIONED_NS is the versioned namespace.  Perhaps we can question whether symbols in the maya-usd namespace need to be versioned or not; because maya-usd defines a core library that can be used by more than one plugin I would tend to err on the safe side and version them.

Also, MAYAUSD_NS_DEF abstracted away whether we were doing a Doxygen build or not --- and of course we're currently not doing anything of the sort.  But again, since maya-usd is defining a core library, it doesn't seem outlandish to think we some day may want to do such a thing.

My suggestion would be to keep MAYAUSD_NS_DEF and simply take the namespace token out of it.  Hope this makes sense.

pmolodo:
Ah, makes sense.  Well... should we consider just making a new macro for the non-namespace-token version?  Mostly to preserve backward-compatibility for dependent code already using `MAYAUSD_NS_DEF` (and which may not use clang-format)?

kxl-adsk (commented):
We favor the non-namespace-token version because of clang-format and refactoring the code after the change is trivial. Keeping both, IMHO may be confusing for developers.

With the above, I can't find reasons for preserving backward-compatibility.

kxl-adsk (approved):
Thank you, @pmolodo. LGTM!

lib/mayaUsd/base/mayaUsd.h.src
```diff
@@ -36,11 +36,10 @@ namespace MAYAUSD_VERSIONED_NS {}
 namespace MAYAUSD_NS {
     using namespace MAYAUSD_VERSIONED_NS;
 }
- 
-// Macros to place the MayaUsd symbols in the versioned namespace, which is how
-// they will appear in the shared library, e.g. MayaUsd_v1::Class.
+
+// Deprecated macro (because it causes problems with clang-format)
```
ppt-adsk, line 40: This comment is no longer accurate, right?  We should restore the previous comment, perhaps fixing it to change the plural "Macros" to "Macro".

ppt-adsk (commented):
Thanks for doing this @pmolodo, very appreciated.  Just this last nit-pick and I think we're done.

kxl-adsk:
@pmolodo PF failed on all platforms - looks like changes went in that are were using the macro. Please merge the latest changes from dev and fix compilation errors.

pmolodo:
Fixed the comment and rebased against latest dev.  Tested that it compiles.

PR #846, opened 2020-10-18
Enable USD version check for components distributed separately

kxl-adsk:
USD plugins that are distributed in separate packages should perform runtime USD version check to avoid potential crashes caused by version mismatch. This PR adds:
- `MAYA_USD_VERSION`  and `MAYAUSD_VERSION` env variables via MayaUSD mod file 
- `MAYA_PXR_PLUGINPATH_NAME` env variable and a service that will append `<USD_VERSION>` to each path and call RegisterPlugins for them. This service runs only once per Maya session and is triggered by either MayaUSD plugin or mtoh (whichever is loaded first). For example, MayaUSD distribution with USD 20.11 will attempt registration of each path in MAYA_PXR_PLUGINPATH_NAME with the `2011` subdirectory.


kxl-adsk:
Still need to add a test, but wanted to open it to collect feedback.

lib/mayaUsd/nodes/proxyShapePlugin.cpp
```diff
@@ -62,6 +63,9 @@ MayaUsdProxyShapePlugin::initialize(MFnPlugin& plugin)
         return MS::kSuccess;
     }
 
+    // Call one time registration of plugins compiled for same USD version as MayaUSD plugin.
+    MAYAUSD_NS::registerVersionedPlugins();
```
seando-adsk, line 67: Here in the core, you should just use MayaUsd::

lib/mayaUsd/render/mayaToHydra/plugin.cpp
```diff
@@ -34,6 +36,9 @@ PXR_NAMESPACE_USING_DIRECTIVE
 PLUGIN_EXPORT MStatus initializePlugin(MObject obj) {
     MStatus ret = MS::kSuccess;
 
+    // Call one time registration of plugins compiled for same USD version as MayaUSD plugin.
+    MAYAUSD_NS::registerVersionedPlugins();
```
seando-adsk, line 40: Same

lib/mayaUsd/utils/plugRegistryHelper.h
```diff
@@ -0,0 +1,67 @@
...
+//
+#ifndef MAYAUSD_PLUGREGISTRYHELPER_H
+#define MAYAUSD_PLUGREGISTRYHELPER_H
+
+#include <mayaUsd/base/api.h>
+
+#include <pxr/base/plug/registry.h>
+#include <pxr/base/tf/getenv.h>
+#include <pxr/base/tf/pathUtils.h>
+#include "pxr/base/tf/stringUtils.h"
```
seando-adsk, line 24: Why does this one have " and not <

kxl-adsk: Copy-paste error, thx!

lib/mayaUsd/nodes/proxyShapePlugin.cpp
kxl-adsk: As per offline discussion, I removed all usages of MAYAUSD_NS.

seando-adsk (approved):

seando-adsk (changes requested):
Will you fix the " vs < in plugRegistryHelper.h?

seando-adsk (approved):

PR #848, opened 2020-10-19
export lambert and descendants more sparsely

mattyjams:
This addresses some test failures we started seeing when adapting them to use the new cascading `shadingMode` setup. The tests do straight USD file diffs of an initial export from Maya versus a re-import and re-export of that initial export. Because the initial export uses the lambert shader writer, `emissiveColor` and `specularColor` were both getting exported as black, which is unnecessary since that's the default value for those inputs anyway:
https://github.com/PixarAnimationStudios/USD/blob/d8a405a1344480f859f025c4f97085143efacb53/pxr/usdImaging/plugin/usdShaders/shaders/shaderDefs.usda#L34

When re-imported, those shaders became usdPreviewSurface nodes in Maya instead of lambert nodes. The reader sets those attributes using the default values which has no effect. When the usdPreviewSurface writer then runs during the second export, it does **not** author `emissiveColor` or `specularColor` since they are not authored in Maya.

The diff of the two USD files would then show `emissiveColor` and `specularColor` present in the first export but absent in the second export. The default-authored values aren't needed in the first export, so these changes just avoid authoring them.

### DIFF
--- a/lib/usd/translators/shading/usdLambertWriter.cpp
+++ b/lib/usd/translators/shading/usdLambertWriter.cpp
@@ -28,6 +28,7 @@
 #include <pxr/usd/usdShade/tokens.h>
 
 #include <maya/MFnDependencyNode.h>
+#include <maya/MPlug.h>
 #include <maya/MStatus.h>
 
 #include <basePxrUsdPreviewSurface/usdPreviewSurface.h>
@@ -111,12 +112,15 @@ void PxrUsdTranslators_LambertWriter::Write(const UsdTimeCode& usdTime)
         }
     }
 
+    // Since incandescence in Maya and emissiveColor in UsdPreviewSurface are
+    // both black by default, only author it in USD if it is authored in Maya.
     AuthorShaderInputFromShadingNodeAttr(
         depNodeFn,
         _tokens->incandescence,
         shaderSchema,
         PxrMayaUsdPreviewSurfaceTokens->EmissiveColorAttrName,
-        usdTime);
+        usdTime,
+        /* ignoreIfUnauthored = */ true);
 
     // Exported, but unsupported in hdStorm.
     AuthorShaderInputFromShadingNodeAttr(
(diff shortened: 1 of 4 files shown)


lib/usd/translators/shading/usdLambertWriter.cpp
```diff
@@ -110,12 +111,21 @@ void PxrUsdTranslators_LambertWriter::Write(const UsdTimeCode& usdTime)
...
-        PxrMayaUsdPreviewSurfaceTokens->EmissiveColorAttrName,
-        usdTime);
+    // Since incandescence in Maya and emissiveColor in UsdPreviewSurface are
+    // both black by default, only author it in USD if it is authored in Maya.
+    const MPlug incandescencePlug =
+        depNodeFn.findPlug(
+            depNodeFn.attribute(_tokens->incandescence.GetText()),
+            /* wantNetworkedPlug = */ true,
+            &status);
+    if (status == MS::kSuccess && UsdMayaUtil::IsAuthored(incandescencePlug)) {
```
JGamache-autodesk, line 121: Maybe that check should be in PxrUsdTranslators_MaterialWriter::AuthorShaderInputFromScaledShadingNodeAttr?

JGamache-autodesk (approved):

lib/usd/translators/shading/usdLambertWriter.cpp
mattyjams: Yeah, I had had that thought too, so I'll make that change. Thanks!

mattyjams: I turned this into an optional parameter to `AuthorShaderInputFromShadingNodeAttr()` in https://github.com/Autodesk/maya-usd/pull/848/commits/7db9ee89e9149f50c28ee44bf92caca9fc8cbd1d and make use of it for the `lambert` writer in https://github.com/Autodesk/maya-usd/pull/848/commits/e81ca1f3bec715894e310e89061b96607f26e9e7.

JGamache-autodesk (approved):

kxl-adsk:
@mattyjams Please resolve merge conflicts.

mattyjams:
> @mattyjams Please resolve merge conflicts.

@kxl-adsk Done! Rebased and resolved conflicts.

PR #849, opened 2020-10-19
disable shading import in UsdMayaTranslatorCamera::ReadToCamera()

mattyjams:
Some additional fallout following #822. This is a workaround specific to `ReadToCamera()`, but we may want to reconsider how/when the `UsdPreviewSurface` shading mode conversion is registered to fix this for all potential cases.

Code that links against mayaUsd but otherwise does not load any Maya plugins currently encounters an error that looks like this when it calls `ReadToCamera()`:

```
Error in '_shadingModesImportArgs' at line 244 in file ext/mayapkg/maya_open_source/maya-usd/lib/mayaUsd/fileio/jobs/jobArgs.cpp : 'Unknown material conversion 'UsdPreviewSurface''```
```

The issue is that the `UsdPreviewSurface` shading conversion is specified as the default in the default args dictionary in the mayaUsd core, but that conversion is only registered when one of the `mayaUsdPlugin`, `usdPreviewSurface`, or `pxrUsdPreviewSurface` plugins are loaded.

Since in this case we're only interested in importing cameras, we can work around that issue for now by overriding `shadingMode` in the default args dictionary and disabling shading import.

### DIFF
--- a/lib/mayaUsd/fileio/translators/translatorCamera.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorCamera.cpp
@@ -27,6 +27,8 @@
 #include <pxr/base/gf/vec2f.h>
 #include <pxr/base/tf/token.h>
 #include <pxr/base/tf/type.h>
+#include <pxr/base/vt/dictionary.h>
+#include <pxr/base/vt/value.h>
 #include <pxr/usd/sdf/path.h>
 #include <pxr/usd/usd/prim.h>
 #include <pxr/usd/usd/timeCode.h>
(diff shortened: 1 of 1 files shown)


kxl-adsk:
@mattyjams I'm not familiar with ReadToCamera, where is it used?

mattyjams:
> @mattyjams I'm not familiar with ReadToCamera, where is it used?

@kxl-adsk It looks like it's not actually used by any other code in maya-usd, but internally we have a simple tool that uses it to transfer a static camera in USD to an existing Maya camera.

`ReadToCamera()` is pretty much the same thing as `Read()` except that it bypasses all of the node creation and imports onto a camera node already in the scene. I imagine this is probably functionality that we'd like all of the translators to have at some point, though admittedly probably not implemented like this. We just do so much with cameras that we added this helper function as a special case.

kxl-adsk:
I see how this is useful and it would be great to have a test for it.

kxl-adsk (approved):

PR #851, opened 2020-10-20
MAYA-105389 - Displacement import

JGamache-autodesk:


lib/mayaUsd/fileio/shaderReader.h
```diff
@@ -50,6 +50,18 @@ class UsdMayaShaderReader : public UsdMayaPrimReader {
...
+    /// USD attribute named \p usdAttrName on \p mayaObject.
+    ///
+    /// The default implementation always returns an empty plug, which
+    /// effectively prevents any connections from being authored to or from
+    /// the imported shader nodes. Derived classes should override this and
+    /// return the corresponding attribute names for the USD attributes
+    /// that should be considered for connections.
+    MAYAUSD_CORE_PUBLIC
+    virtual MPlug
+    GetMayaPlugForUsdAttrName(const TfToken& usdAttrName, const MObject& mayaObject) const;
```
JGamache-autodesk, line 63: Allows connection to a secondary object instead of the main one stored in the ReadContext.

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -433,48 +433,81 @@ class UseRegistryShadingModeImporter {
             return MObject();
         }
 
+        MPlug surfaceOutputPlug, volumeOutputPlug, displacementOutputPlug;
```
JGamache-autodesk, line 436: We now read all known inputs of the material, which is symmetric with what the exporter writes.

lib/usd/translators/shading/usdMaterialReader.cpp
```diff
@@ -84,13 +86,27 @@ bool PxrUsdTranslators_MaterialReader::Read(UsdMayaPrimReaderContext* context)
...
+            std::string shaderName = prim.GetName().GetString();
+            shaderName += "_displacement";
+            UsdMayaTranslatorUtil::CreateShaderNode(
+                shaderName.c_str(),
+                "displacementShader",
+                UsdMayaShadingNodeType::Shader,
+                &status,
+                &_displacementShader);
+            // this displacement shader will get connected as we explore
+            // the displacement input of the UsdShadeMaterial.
```
JGamache-autodesk, line 100: The material reader, which is responsible for backconverting from UsdPreviewSurface will now hold on to a displacementShader when there is an authored displacement input on the material.

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -433,48 +433,81 @@ class UseRegistryShadingModeImporter {
...
+                }
+            }
         }
 
-        MPlug shaderOutputPlug = _GetSourcePlug(surfaceShader, UsdShadeTokens->surface);
-        if (shaderOutputPlug.isNull()) {
-            return MObject();
-        }
+        if (!surfaceOutputPlug.isNull() || !volumeOutputPlug.isNull()
+            || !displacementOutputPlug.isNull()) {
```
mattyjams, line 475: Maybe invert this and return so we can save an indent level for everything below?:

```C++
        if(surfaceOutputPlug.isNull() && volumeOutputPlug.isNull() &&
                displacementOutputPlug.isNull()) {
            return MObject();
        }

        // Create the shading engine.
        ....
```

lib/usd/translators/shading/usdMaterialReader.h
```diff
@@ -36,6 +36,11 @@ class PxrUsdTranslators_MaterialReader : public UsdMayaShaderReader {
 public:
     PxrUsdTranslators_MaterialReader(const UsdMayaPrimReaderArgs&);
 
+    /// Get the name of the Maya shading plug that corresponds to the
+    /// USD attribute named \p usdAttrName on \p mayaObject.
```
mattyjams, line 40: Similar comment as above about the docstring:

```C++
    /// Get the Maya shading plug on \p mayaObject that corresponds to the USD
    /// attribute named \p usdAttrName.
```

mattyjams (approved):
A few minor comments and suggestions, but otherwise this looks good to me.

I'm still a bit concerned about how much assumption of `UsdPreviewSurface` is getting baked into some of this translation, though I recognize that that will likely be the vast majority of use cases at least initially. I would bet though at some point we'll need a dedicated shader reader specifically for `displacementShader` Maya nodes for networks in USD where there's no `UsdPreviewSurface` involved?

test/lib/usd/translators/UsdImportDisplacementTest/SimpleDisplacement.usda
```diff
@@ -0,0 +1,86 @@
+#usda 1.0
+(
+    defaultPrim = "pPlane16"
+    metersPerUnit = 0.01
+    upAxis = "Y"
+)
+
+def Mesh "pPlane16" (
```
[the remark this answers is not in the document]
JGamache-autodesk: We have a much bigger issue here. This is exactly how Maya currently exports materials if you export a single mesh with a material. I did not hand-edit the file at all.

JGamache-autodesk:
I tried to limit UsdPreviewSurface knowledge to usdMaterialReader.cpp, where it belongs. The support at the shaderReader level is more for any importer that has to deal with a 1-to-N situation, where a single UsdShade node expands to more than one Maya dependency node, and where connections to auxiliary nodes are expected. UsdPreviewSurface was just conveniently available as a demo node for this feature.

test/lib/usd/translators/UsdImportDisplacementTest/SimpleDisplacement.usda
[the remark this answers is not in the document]
mattyjams: Ah, I see. For our assets, we always organize everything under a single transform node when working in Maya so that when we export it to USD there's a definitive defaultPrim. This makes things a lot easier when you then author references to that USD.

I think this is where we figure out where to author the `Materials` scope:
https://github.com/Autodesk/maya-usd/blob/dc910a9da47299ceb916c8b0b0a12505b1ab958d/lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp#L336

Sounds like maybe after we've found a root prim path but **before** we append the materialsScopeName to it, we should check whether it's an `Xform`, and if it's not, just author the `Materials` scope under the `AbsoluteRootPath`?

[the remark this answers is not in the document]
JGamache-autodesk: Yes. That would potentially solve an issue we have where "Looks" transforms are multiplying in Maya. You get one for each export-import cycle. I will copy this discussion to the internal bug we opened to track the breeding Looks and see if we can schedule it a bit earlier.

PR #860, opened 2020-10-21
MAYA-106690 - Merge subsets connected to same material

JGamache-autodesk:
If you connect one set of faces to a material, then later connect a second set of faces to the same material, there will be two separate sets in Maya. When exporting to USD we need to merge the faces into a single GeomSubset.

### DIFF
--- a/lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
+++ b/lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
@@ -511,11 +511,11 @@ UsdMayaShadingModeExportContext::MakeStandardMaterialPrim(
                     // Update and continue:
                     VtIntArray mergedIndices;
                     UsdAttribute indicesAttribute = faceSubset.GetIndicesAttr();
-                    indicesAttribute.Get(&mergedIndices, UsdTimeCode::Default());
+                    indicesAttribute.Get(&mergedIndices);
                     std::set<int> uniqueIndices(mergedIndices.cbegin(), mergedIndices.cend());
                     uniqueIndices.insert(faceIndices.cbegin(), faceIndices.cend());
                     mergedIndices.assign(uniqueIndices.cbegin(), uniqueIndices.cend());
-                    indicesAttribute.Set(mergedIndices, UsdTimeCode::Default());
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -492,8 +492,35 @@ UsdMayaShadingModeExportContext::MakeStandardMaterialPrim(
...
+
+                if (faceSubset) {
+                    // Update and continue:
+                    VtIntArray mergedIndices;
+                    UsdAttribute indicesAttribute = faceSubset.GetIndicesAttr();
+                    indicesAttribute.Get(&mergedIndices, UsdTimeCode::Default());
+                    std::set<int> uniqueIndices(mergedIndices.cbegin(), mergedIndices.cend());
+                    uniqueIndices.insert(faceIndices.cbegin(), faceIndices.cend());
+                    mergedIndices.assign(uniqueIndices.cbegin(), uniqueIndices.cend());
+                    indicesAttribute.Set(mergedIndices, UsdTimeCode::Default());
```
mattyjams, line 518: I might leave off the `time` parameter to the `UsdAttribute` `Get()` and `Set()` calls. It's optional and what we're passing here matches the default, but their presence suggests that we're intending to do something specific related to time, which I don't think we are. I think it would be fairly weird if the binding subsets were time varying, if that's even supported at all in Maya.

test/lib/usd/translators/UsdExportGeomSubsetTest/UsdExportGeomSubsetTest.ma
```diff
@@ -0,0 +1,348 @@
+//Maya ASCII 2021ff01 scene
+//Name: ExportGeomSubsetTest.ma
+//Last modified: Wed, Oct 21, 2020 10:54:20 AM
+//Codeset: 1252
+requires maya "2021ff01";
```
mattyjams, line 5: Similar comment as in another PR, is this ok when running the test against Maya 2019 or Maya 2020?

mattyjams (approved):
Just one tiny suggestion, but otherwise looks good to me!

PR #861, opened 2020-10-21
MAYA-107461  expose stage reset notices to python

luceric-adsk:
expose these notice objects so I can catch them in python


seando-adsk (approved):
Looks good.

PR #862, opened 2020-10-21
MAYA-106892 - Export and import color space information on file nodes

JGamache-autodesk:
On export, if the color space of a file node was set (differs from the global rule) then we export it as a colorSpace metadata on the file UsdShadeInput.
On import, if there is a colorSpace on the USD file input, we set it in the Maya file node.

### DIFF
--- a/lib/usd/translators/shading/usdFileTextureWriter.cpp
+++ b/lib/usd/translators/shading/usdFileTextureWriter.cpp
@@ -223,11 +223,6 @@ PxrUsdTranslators_FileTextureWriter::Write(const UsdTimeCode& usdTime)
         return;
     }
 
-    MString colorRuleCmd;
-    colorRuleCmd.format(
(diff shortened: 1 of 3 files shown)


lib/usd/translators/shading/usdFileTextureWriter.cpp
```diff
@@ -221,6 +223,11 @@ PxrUsdTranslators_FileTextureWriter::Write(const UsdTimeCode& usdTime)
         return;
     }
 
+    MString colorRuleCmd;
+    colorRuleCmd.format(
+        "colorManagementFileRules -evaluate \"^1s\";", fileTextureNamePlug.asString());
+    const MString colorSpaceByRule = MGlobal::executeCommandStringResult(colorRuleCmd);
+
```
mattyjams, line 230: Maybe move this closer to where we get the `colorSpace` plug, or even inside that `if (status == MS::kSuccess) {` block?

mattyjams (approved):
Looks good to me!

PR #864, opened 2020-10-22
MAYA-107591: Subtree invalidation support on StageProxy ( a.k.a pseudo-root ).

HamedSabri-adsk:
This PR fixes the cases where subtree invalidation doesn't occur in the outliner where all the children are the descendant of stage proxy shape ( a.k.a pseudo-root).

This issue became obvious during the current ongoing work on reordering support where outliner stopped to refresh.

### DIFF
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -203,7 +203,7 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
 		// each Maya Dag path instancing the proxy shape / stage.
 		Ufe::Path ufePath;
 		UsdPrim prim;
-		if (changedPath.GetString() == "/")
+		if (changedPath.IsAbsoluteRootPath())
 		{
 			ufePath = stagePath(sender);
 			prim = stage->GetPseudoRoot();
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -194,16 +194,27 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
 		// don't care about it. In those cases, the changePath will contain something like:
 		//   "/<prim>.visibility"
 		//   "/<prim>.xformOp:translate"
-		if (!changedPath.IsPrimPath())
+		if (changedPath.IsPropertyPath())
 			continue;
 
-		const std::string& usdPrimPathStr = changedPath.GetPrimPath().GetString();
 		// Assume proxy shapes (and thus stages) cannot be instanced.  We can
 		// therefore map the stage to a single UFE path.  Lifting this
 		// restriction would mean sending one add or delete notification for
 		// each Maya Dag path instancing the proxy shape / stage.
-		Ufe::Path ufePath = stagePath(sender) + Ufe::PathSegment(usdPrimPathStr, g_USDRtid, '/');
-		auto prim = stage->GetPrimAtPath(changedPath);
+		Ufe::Path ufePath;
+		UsdPrim prim;
+		if (changedPath.GetString() == "/")
+		{
+			ufePath = stagePath(sender);
+			prim = stage->GetPseudoRoot();
+		}
+		else 
+		{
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -194,16 +194,27 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
 		// don't care about it. In those cases, the changePath will contain something like:
 		//   "/<prim>.visibility"
 		//   "/<prim>.xformOp:translate"
-		if (!changedPath.IsPrimPath())
+		if (changedPath.IsPropertyPath())
 			continue;
```
HamedSabri-adsk, line 198: `IsPrimPath()` checks  if a path identifies a prim but what we wanted here is to check if path identifies a property. 

calling isPrimPath() on "/" would return false and therefore subtree invalidation never occurs.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -194,16 +194,27 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
...
+			ufePath = stagePath(sender);
+			prim = stage->GetPseudoRoot();
+		}
+		else 
+		{
+			const std::string& usdPrimPathStr = changedPath.GetPrimPath().GetString();
+			ufePath = stagePath(sender) + Ufe::PathSegment(usdPrimPathStr, g_USDRtid, '/');
+			prim = stage->GetPrimAtPath(changedPath);
+		}
+
```
HamedSabri-adsk, line 217: Add conditional check to set `ufePath `and `prim` in the case where the changedPath is a pseduo-root.

seando-adsk (approved):
Looks good, thanks.

ppt-adsk (approved):

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -194,16 +194,27 @@ void StagesSubject::stageChanged(UsdNotice::ObjectsChanged const& notice, UsdSta
...
-		const std::string& usdPrimPathStr = changedPath.GetPrimPath().GetString();
 		// Assume proxy shapes (and thus stages) cannot be instanced.  We can
 		// therefore map the stage to a single UFE path.  Lifting this
 		// restriction would mean sending one add or delete notification for
 		// each Maya Dag path instancing the proxy shape / stage.
-		Ufe::Path ufePath = stagePath(sender) + Ufe::PathSegment(usdPrimPathStr, g_USDRtid, '/');
-		auto prim = stage->GetPrimAtPath(changedPath);
+		Ufe::Path ufePath;
+		UsdPrim prim;
+		if (changedPath.GetString() == "/")
```
kxl-adsk, line 206: How about we replace this with
```
changedPath == SdfPath::AbsoluteRootPath()
```

mattyjams: There's also this:

```C++
if (changedPath.IsAbsoluteRootPath()) {
...
```

https://github.com/PixarAnimationStudios/USD/blob/d8a405a1344480f859f025c4f97085143efacb53/pxr/usd/sdf/path.h#L338

HamedSabri-adsk: Ahhh... that's what I was looking for. :)   Addressed in d109480e4277aac04786963e4e374cf3aa09a52c

kxl-adsk:
One bad news - IsAbsoluteRootPath is not available with 19.11. We are getting the following compilation error
```E:\Dev\usd\maya-usd_pr\lib\mayaUsd\ufe\StagesSubject.cpp(206): error C2039: 'IsAbsoluteRootPath': is not a member of 'pxrInternal_v0_19__pxrReserved__::SdfPath'
E:\Dev\usd\_workspaces_latest\Pixar_USD_19.11\include\pxr/usd/pcp/primIndex.h(63): note: see declaration of 'pxrInternal_v0_19__pxrReserved__::SdfPath'
```

kxl-adsk:
But looks like SdfPath::AbsoluteRootPath() is available - https://github.com/PixarAnimationStudios/USD/blob/4b1162957b2ad219e1635c81de8343be490e41fc/pxr/usd/lib/sdf/path.h#L295

seando-adsk (approved):

PR #865, opened 2020-10-23
MAYA-107167 - Keep USDZ texture paths relative

### LINKED ISSUE #795: [MAYA-107167] USDZ export writes paths that aren't inclusive to the usdz
**Describe the bug**
There was a regression to this repo that caused usdz files to write references that aren't inclusive to the zip file.
Previously, the paths would be written relative to the internals of the usdz and would include the referenced dependencies into the usdz file, however this no longer works.

This was due to a change in [cb77d92c63](https://github.com/Autodesk/maya-usd/commit/cb77d92c63) `lib/usd/translators/shading/usdFileTextureWriter.cpp`
**Steps to reproduce**
1. Export asset as a usdz
2. Check the file path reference on the Shader file1 prim.

**Expected behavior**
Prior to this change, the filepath would resolve as `0/checkerboard.png` since the path and texture would be inlined into the usdz.
After this change, the path is relative to the workspace and no longer inlines the texture into the zip.

We disabled this block of code, and it's back to the expected behaviour, so I think revisiting this code will resolve the issue.

**Attachments**
I've attached the Maya workspace this is generated from, and the usdz is inside it as well

[bugreports.zip](https://github.com/Autodesk/maya-usd/files/5272256/bugreports.zip)

**Specs (if applicable):**
 - macOS Catalina and Big Sur
 - Compiler version: Apple clang version 12.0.0 (clang-1200.0.32.5) (reproducible on older clang versions too)
 - Maya 2018.5 and 2020.0
 - Maya USD commit SHA 4365cc6675f666ee8ec23bc21c12ef68b3a5d5c7
 - Pixar USD commit SHA 71b4baace2044ea4400ba802e91667f9ebe342f0

**Additional context**
We are still using `pxrUSD.bundle` to do our exporting.

JGamache-autodesk:
If the path requires a relationship to resolves, we keep it relative.
Fixes #795


lib/usd/translators/shading/usdUVTextureReader.cpp
```diff
@@ -179,12 +179,15 @@ bool PxrMayaUsdUVTexture_Reader::Read(UsdMayaPrimReaderContext* context)
     UsdShadeInput usdInput = shaderSchema.GetInput(_tokens->file);
     if (usdInput && usdInput.Get(&val) && val.IsHolding<SdfAssetPath>()) {
         std::string filePath = val.UncheckedGet<SdfAssetPath>().GetResolvedPath();
-        if (!filePath.empty()) {
+        if (!filePath.empty()
+            && filePath[filePath.length() - 1]
+                != SdfPathTokens->relationshipTargetEnd.GetString()[0]) {
```
mattyjams, line 184: We should use `Ar` here to test this instead:

```C++
#include <pxr/usd/ar/packageUtils.h>
...
        if (!filePath.empty() && !ArIsPackageRelativePath(filePath)) {
...
```

mattyjams (changes requested):
There's a utility function available in the `Ar` library that we should use instead of the test that's there now. Otherwise, these changes make sense to me.

But I'm not seeing how this change addresses #795. That issue has to do with export, and this only deals with import. I would guess that it's the stuff here that @dgovil is talking about needing to bypass in order to see the desired asset paths in the exported package?:
https://github.com/Autodesk/maya-usd/blob/96d40b7e6c019e47902364d979b3758545631299/lib/usd/translators/shading/usdFileTextureWriter.cpp#L228

JGamache-autodesk:
I missed the "We are still using pxrUSD.bundle to do our exporting.". I will re-check and fix that code path.

dgovil:
Thanks for fixing up the pxrUSD.bundle too! We're about to transition to the mayaUSD.bundle now so hopefully I won't be an imposition on y'all for much longer in that regard, but this will be great for all our code that still calls the pixar export command instead (so hard to track down code across multiple teams and get them all ported)

mattyjams (approved):

PR #868, opened 2020-10-26
[CMake] Add mayaUtils.openTestScene and mayaUtils.getTestScene

marsupial:
Fixes relative paths to test-scenes copied to **${CMAKE_BINARY_DIR}** when compiled as a sub-project.
  * testParentCmd.py, testSelection.py

Refactor duplicate **os.path.join** and **cmds.file** across other tests scenes into one place
  * other tests


ppt-adsk (approved):
Thanks!

PR #869, opened 2020-10-26
wrap UFE runtime ID getters for Maya and USD to Python

mattyjams:
We're starting to work more UFE awareness into some of our internal pipeline tools and having the UFE runtime IDs more readily accessible to client code would be handy.


seando-adsk (approved):
All good, thanks for adding this and fixing the existing hard-coded values.

PR #876, opened 2020-10-30
MAYA-106486 Handle UV linkage on material export

JGamache-autodesk:
We now explore the UV linkage and make sure the UV stream name in the material matches the UV set name on the shape.
We will create as many material specializations as needed to match the shape streams in case of UV stream names inconsistency in a group of shapes sharing a single material.
We no longer automatically rename "map1" to "st" since the UV streams in the material will now be correctly set. You can use the following pair of environment variables to bring back the renaming behaviour: MAYAUSD_EXPORT_MAP1_AS_PRIMARY_UV_SET and MAYAUSD_IMPORT_PRIMARY_UV_SET_AS_MAP1.


lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -419,17 +422,25 @@ _UninstancePrim(
...
-        SdfPathSet * const boundPrimPaths) const
+UsdPrim UsdMayaShadingModeExportContext::MakeAndBindStandardMaterialPrim(
+    const AssignmentVector& assignmentsToBind,
+    const std::string&      name,
+    SdfPathSet* const       boundPrimPaths) const
+{
+    UsdPrim materialPrim = MakeStandardMaterialPrim(assignmentsToBind, name);
+    BindStandardMaterialPrim(materialPrim, assignmentsToBind, boundPrimPaths);
+    return materialPrim;
+}
```
JGamache-autodesk, line 433: This function is split in two parts since the useRegistry exporter now wants to have the material exported before the binding is done.

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -440,111 +451,246 @@ UsdMayaShadingModeExportContext::MakeStandardMaterialPrim(
...
-        UsdShadeMaterial material = UsdShadeMaterial::Define(
-                GetUsdStage(), materialPath);
+    if (UsdPrim materialParent
+        = _GetMaterialParent(stage, GetExportArgs().materialsScopeName, assignmentsToBind)) {
+        SdfPath          materialPath = materialParent.GetPath().AppendChild(TfToken(materialName));
+        UsdShadeMaterial material = UsdShadeMaterial::Define(GetUsdStage(), materialPath);
 
         UsdPrim materialPrim = material.GetPrim();
 
-        // could use this to determine where we want to export.
```
JGamache-autodesk, line 454: The binding code exported in a separate function.

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -440,111 +451,246 @@ UsdMayaShadingModeExportContext::MakeStandardMaterialPrim(
...
+    _UVMappingManager uvMappingManager(material, assignmentsToBind);
+
+    UsdStageRefPtr stage = GetUsdStage();
+    TfToken        materialNameToken(materialPrim.GetName());
+    for (const auto& iter : assignmentsToBind) {
+        const SdfPath&    boundPrimPath = iter.boundPrimPath;
+        const VtIntArray& faceIndices = iter.faceIndices;
+        const TfToken&    shapeName = iter.shapeName;
+
+        const UsdShadeMaterial& materialToBind = uvMappingManager.getMaterial(shapeName);
```
JGamache-autodesk, line 612: Main difference in the extracted code for binding materials is here. We ask the mapping manager for the material to bind. In case of inconsistent texcoord naming in the shape set, this function can return specialized materials that fit the needs of the current shape.

lib/mayaUsd/fileio/utils/meshReadUtils.cpp
```diff
@@ -245,12 +245,7 @@ namespace
         MString uvSetName(primvarName.GetText());
         bool createUVSet = true;
 
-        if (primvarName == UsdUtilsGetPrimaryUVSetName()) {
-            // We assume that the primary USD UV set maps to Maya's default 'map1'
-            // set which always exists, so we shouldn't try to create it.
-            uvSetName = UsdMayaMeshPrimvarTokens->DefaultMayaTexcoordName.GetText();
-            createUVSet = false;
-        } else if (!hasDefaultUVSet) {
+        if (!hasDefaultUVSet) {
```
JGamache-autodesk, line 248: The removes the automatic "st"-> "map1" renaming on import. Should I create an env var for shops that have internal tools that depend on this behavior?

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -828,17 +828,9 @@ UsdMayaMeshWriteUtils::writeUVSetsAsVec2fPrimvars(const MFnMesh& meshFn,
             continue;
         }
 
-        // XXX: We should be able to configure the UV map name that triggers this
-        // behavior, and the name to which it exports.
-        // The UV Set "map1" is renamed st. This is a Pixar/USD convention.
-        TfToken setName(uvSetNames[i].asChar());
-        if (setName == UsdMayaMeshPrimvarTokens->DefaultMayaTexcoordName.GetText()) {
-            setName = UsdUtilsGetPrimaryUVSetName();
```
JGamache-autodesk, line 836: The removes the automatic "map1"-> "st" renaming on export. Should I create an env var for shops that have internal tools that depend on this behavior?

test/lib/usd/translators/testUsdExportUVSetMappings.py
```diff
@@ -0,0 +1,85 @@
...
+        '''
+        self.assertTrue(self._stage)
+
+    def testExportUVSetMappings(self):
+        '''
+        Tests that exporting multiple Maya planes with varying UV mappings
+        setups results in USD data with material specializations:
+        '''
+        expected = [
+            ("/pPlane1", "/blinn1SG_map1_map1_map1", "map1", "map1", "map1"),
```
JGamache-autodesk, line 64: No longer renamed to st

lib/mayaUsd/fileio/shading/shadingModeExporterContext.h
```diff
@@ -105,18 +110,39 @@ class UsdMayaShadingModeExportContext
     AssignmentVector GetAssignments() const;
 
     /// Use this function to create a UsdShadeMaterial prim at the "standard"
-    /// location.  The "standard" location may change depending on arguments
-    /// that are passed to the export script.
+    /// location and bind it to all assignments provided. The "standard" location
+    /// may change depending on arguments that are passed to the export script.
+    ///
+    /// Note: Call this function if correct UV linking is not a requirement. Use
+    ///       the two-stage version otherwise.
```
mattyjams, line 117: Are there use cases where you'd prefer **not** to incorporate UV linking? I'm wondering whether there's a strong reason why we should keep `MakeAndBindStandardMaterialPrim()`, or if we should just remove it and always use `MakeStandardMaterialPrim()` followed by `BindStandardMaterialPrim()`. `shadingModePxrRis.cpp` and `mayaToHydra/shadingModeExporter.cpp` seem to be the only usages of it that I can see.

lib/mayaUsd/fileio/utils/meshReadUtils.cpp
mattyjams: Yes, please! Export is more of a concern for us than import (more detail on that below), but it would be good to have a `TfEnvSetting` we could set for each direction.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
mattyjams: Yes, definitely. Having the Maya USD export portion of our asset builds suddenly writing `map1` primvars instead of `st` primvars would be a non-trivial pipeline change for us, so at least for now, we'll need to be able to opt out of that.

Including a setting for import, maybe a pair of `TfEnvSettings` like this?:
```C++
TF_DEFINE_ENV_SETTING(
    MAYAUSD_IMPORT_PRIMARY_UV_SET_AS_MAP1,
    false,
    "Translates the primary UV set in USD to the map1 UV set in Maya. "
    "When disabled, UV set names are translated directly (st in USD becomes st in Maya).");

TF_DEFINE_ENV_SETTING(
    MAYAUSD_EXPORT_MAP1_AS_PRIMARY_UV_SET,
    false,
    "Translates the map1 UV set in Maya to the primary UV set in USD. "
    "When disabled, UV set names are translated directly (map1 in Maya becomes map1 in USD).");
```

Thanks! :)

mattyjams (approved):
We will indeed need `TfEnvSettings` to be able to fallback to the `map1`->`st` translation, particularly for export. But otherwise, looks good to me!

Thanks @JGamache-autodesk!

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
HamedSabri-adsk: Yes agreed, having the env variable to make a switch would be useful. Thanks

HamedSabri-adsk (approved):
Looking good! Thanks.

mattyjams:
Just wanted to flag here that after bringing in this change, we're seeing an issue where exported materials specialize themselves. There isn't any UV set re-mapping happening in these cases so I'm not sure what's going on yet, but there looks to be some fallout to this that needs to be addressed.

JGamache-autodesk:
Yes, missing an important early return in:

    const UsdShadeMaterial& getMaterial(const TfToken& shapeName)
    {
        if (_uvNamesToMaterial.empty()) {
            return _material;
        }

PR #880, opened 2020-11-02
MAYA-107246 - USD: outliner does not update when duplicating items

seando-adsk:
#### MAYA-107246 - USD: outliner does not update dynamically when duplicating items

* Outliner was being sent incorrect UFE notif (SubtreeInvalidate) when duplicating an item. It should get ObjectAdd.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
@@ -27,6 +27,8 @@
 
 #include <mayaUsdUtils/util.h>
 
+#include "private/UfeNotifGuard.h"
+
 namespace MAYAUSD_NS_DEF {
 namespace ufe {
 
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
```diff
@@ -88,6 +90,7 @@ bool UsdUndoDuplicateCommand::duplicate(const SdfLayerHandle& layer, const SdfPa
 	// We use the source layer as the destination.  An alternate workflow
 	// would be the edit target layer be the destination:
 	// layer = self._stage.GetEditTarget().GetLayer()
+	MayaUsd::ufe::InAddOrDeleteOperation ad;
```
seando-adsk, line 93: Without this in StagesSubject we'll end up sending a SubtreeInvalidate notif which is not correct. When duplicating we need an ObjectAdd. You'll see below on line 114, we explicitly send and ObjectDelete.

seando-adsk: I debated between putting this notif here or in the redo() below - which calls this static method. I think its better here because this is also called from the sceneitemops.

ppt-adsk (approved):

PR #881, opened 2020-11-02
fix call to non-existent function in pxrUsd plugin's usdTranslatorImport

mattyjams:
The existing call to `usdTranslatorImport_SetOptionMenuByAnnotation()` was replaced with a call to `usdTranslatorImport_SetShadingOptionMenuByAnnotation()` in #865, but the latter does not exist anywhere. This reverts that change.


mattyjams:
@JGamache-autodesk: This change got lumped into #865, which was an otherwise non-shading-related PR. Was there some other intention here?:
https://github.com/Autodesk/maya-usd/pull/865/files#diff-dda3f8c38b95d3d9a4339f961596e4f7c3e596ab3c342aa1ae831ff1510e3443R163

JGamache-autodesk (approved):
Sorry about that. It was experimental code that was removed and was not cleaned-up correctly.

mattyjams:
> Sorry about that. It was experimental code that was removed and was not cleaned-up correctly.

Ah, cool. No problem. Thanks for confirming!

PR #883, opened 2020-11-03
MAYA-106486 UDIM Import/Export

JGamache-autodesk:
Correctly detect and handle UDIM on import and export.


lib/usd/translators/shading/usdUVTextureReader.cpp
```diff
@@ -189,8 +195,40 @@ bool PxrMayaUsdUVTexture_Reader::Read(UsdMayaPrimReaderContext* context)
...
+                        std::string resolvedName
+                            = SdfComputeAssetPathRelativeToLayer(res.GetLayer(), udimPath);
+
+                        if (!resolvedName.empty() && !ArIsPackageRelativePath(resolvedName)
+                            && resolvedName != udimPath) {
+                            udimPath = resolvedName;
+                            break;
+                        }
+                    }
+                    val = SdfAssetPath(udimPath);
```
mattyjams, line 229: As of USD 20.08, there's a `UsdImaging_ResolveMaterialParamValue()` utility that might be useful here:
https://github.com/PixarAnimationStudios/USD/blob/71b4baace2044ea4400ba802e91667f9ebe342f0/pxr/usdImaging/usdImaging/materialParamUtils.h#L45

mattyjams (approved):
Looks good to me!

lib/usd/translators/shading/usdUVTextureReader.cpp
JGamache-autodesk: Would love to use that function, but it is currently private:
https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usdImaging/usdImaging/CMakeLists.txt#L68
Maybe in the next version?

PR #843, opened 2020-10-16
Test of applying clang-format to whole repo

pmolodo:
So that we can preview what code will look like with latest changes to .clang-format, and for use with testing of merging into existing PRs / branches.

Instructions for applying clang-format to an outstanding branch / PR
====================================================================
These instructions assume that:
- `$MAYA_USD_REPO` points to the root directory of a checkout of the maya-usd source repository
- `$COMMIT_BEFORE_FORMAT` points at the commit in the dev branch immediately before clang-format is applied (322f10207cc2d84318b19fec4741e6e42d8f2d8c in my example clang_format_test branch)
- `$FORMAT_COMMIT` points at the commit in the dev branch where clang-format was applied (eb6c346570ce56c03739a8cc4baaa4be8e4fe695 in my example clang_format_test branch)
- `$FORMAT_BRANCH` points at the head of the branch, containing any commits after `$FORMAT_COMMIT` (once we do this for real, this would just be `dev`, but is `clang_format_test` in my example branch)
- `$YOUR_BRANCH` is the name of the branch which does not have clang-format applied, but want to merge dev (with clang-format) into

Download and install clang-tools 10.0 (generally included as part of llvm):
- MacOS: https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/clang+llvm-10.0.0-x86_64-apple-darwin.tar.xz
- Windows: https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/LLVM-10.0.0-win64.exe
- Ubuntu 18.04: https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/clang+llvm-10.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz
- CentOS:
- On CentOS, I had to build myself - rough instructions follow:
  Make sure a modern version of cmake, and python-3, are both on your path, then:
  ```
  sudo yum install libedit-devel libxml2-devel ncurses-devel swig ninja-build devtoolset-9
  source /opt/rh/devtoolset-9/enable
  git clone -b release/10.x --single-branch --depth 1 https://github.com/llvm/llvm-project.git
  cd llvm-project
  mkdir build
  cd build
[...]

PR #886, opened 2020-11-04
Apply clang-format

kxl-adsk:
Time to start leveraging clang-format in maya-usd. Below are updated instructions from #843 

Instructions for applying clang-format to an outstanding branch / PR
====================================================================
These instructions assume that:
- `$MAYA_USD_REPO` points to the root directory of a checkout of the maya-usd source repository
- `$COMMIT_BEFORE_FORMAT` points at the commit in the dev branch immediately before clang-format is applied (this is 5cfeb1b045189ef0dcc1a67ab503cda915b98954)
- `$FORMAT_COMMIT` points at the commit in the dev branch where clang-format was applied (this is 6bd514da818b7769797c5195024f7f87631f4d22)
- `$FORMAT_BRANCH` points at the head of the branch, containing any commits after `$FORMAT_COMMIT` (dev branch)
- `$YOUR_BRANCH` is the name of the branch which does not have clang-format applied, but want to merge dev (with clang-format) into

Download and install clang-tools 10.0 (generally included as part of llvm):
- MacOS: use brew or https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/clang+llvm-10.0.0-x86_64-apple-darwin.tar.xz
- Windows: VS2019 comes with v10.0 or https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/LLVM-10.0.0-win64.exe
- Ubuntu 18.04: https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/clang+llvm-10.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz
- CentOS: build yourself - rough instructions follow:
  Make sure a modern version of cmake, and python-3, are both on your path, then:
  ```
  sudo yum install libedit-devel libxml2-devel ncurses-devel swig ninja-build devtoolset-9
  source /opt/rh/devtoolset-9/enable
  git clone -b release/10.x --single-branch --depth 1 https://github.com/llvm/llvm-project.git
  cd llvm-project
  mkdir build
  cd build
  cmake -D CMAKE_INSTALL_PREFIX=/usr/local/clang-10.0 -D CMAKE_BUILD_TYPE=RELEASE -D CMAKE_VERBOSE_MAKEFILE=1 -D LLVM_ENABLE_PROJECTS="clang;clang-tools-extra;compiler-rt;debuginfo-tests;libc;libclc;libcxx;libcxxabi;libunwind;lld;lldb;mlir;openmp;parallel-libs;polly;pstl" -D LLVM_ENABLE_BINDINGS=0 -D LLVM_TARGETS_TO_BUILD=all -GNinja ../llvm
  ninja-build install
  ```
Merge + apply clang-format to your branch (these instructions assume Linux, you may need to adapt to other OSes):
```
cd $MAYA_USD_REPO
git checkout $YOUR_BRANCH
git merge $COMMIT_BEFORE_FORMAT
# resolve any merge errors that may have resulted from previous command
# this will technically create a merge in git, but won't actually change the
# state of any files
git merge -s ours $FORMAT_COMMIT
# now, apply the formatting to your branch...for Linux:
find . -regextype posix-egrep -regex "$(echo '.*('$(echo $(cat .clang-format-include) | tr ' ' '\|')')')" -exec clang-format -i {} \+
# for MacOS
# find -E . -regex "$(echo '.*('$(echo $(cat .clang-format-include) | tr ' ' '\|')')')" -exec clang-format -i {} \+
# ...then ammend the commit, to make this the new "merged" result
git commit -a --amend --no-edit
# now merge in any of the latest stuff, that came AFTER the clang-format changes
git merge $FORMAT_BRANCH
# OPTIONAL: do a sanity check, to make sure that the diff of your branch to
#           the format branch now only contains "substantive" changes (ie, not a
#           bunch of formatting changes!)
git diff $FORMAT_BRANCH
# ...or, if you have a gui difftool set up in your .gitconfig:
git difftool $FORMAT_BRANCH
```


lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -380,69 +380,53 @@ _computeLoadAndUnloadItems(const UsdPrim& prim)
...
 UsdContextOps::Ptr UsdContextOps::create(const UsdSceneItem::Ptr& item)
 {
-	return std::make_shared<UsdContextOps>(item);
+    return std::make_shared<UsdContextOps>(item);
 }
 
-void UsdContextOps::setItem(const UsdSceneItem::Ptr& item)
-{
-	fItem = item;
-}
```
seando-adsk, line 422: Why this? I think its okay when this in a header file, but not in a cpp.

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -244,18 +228,18 @@ Ufe::SceneItem::Ptr UsdHierarchy::defaultParent() const
...
+    // create a reorder command and pass in the parent and its reordered children list
+    return UsdUndoReorderCommand::create(downcast(sceneItem())->prim(), orderedTokens);
 }
 #endif
 
 #endif // UFE_V2_FEATURES_AVAILABLE
 
 } // namespace ufe
-} // namespace MayaUsd
+} // namespace MAYAUSD_NS_DEF
```
seando-adsk, line 245: This is a shame. The macro name is really not useful here.

lib/usd/ui/TreeModel.cpp
```diff
@@ -15,386 +15,360 @@
...
-
-#include <QtWidgets/QTreeView>
-#include <QtCore/QSortFilterProxyModel>
-
 #include <pxr/usd/usd/variantSets.h>
 
-#include <mayaUsdUI/ui/TreeItem.h>
-#include <mayaUsdUI/ui/ItemDelegate.h>
+#include <QtCore/QSortFilterProxyModel>
+#include <QtWidgets/QTreeView>
```
seando-adsk, line 21: It seems like these should be part of group 6 and thus be moved to line 25?

lib/mayaUsd/ufe/UsdContextOps.cpp
kxl-adsk: I don't know of a way to have separate rules for header files vs cpp. Given it shouldn't affect our ability to debug the method, let's not re-open the discussion and keep the existing configuration.

lib/mayaUsd/ufe/UsdHierarchy.cpp
kxl-adsk: Clang-format auto-generates this comment and currently, we are giving it a macro. Something that may be newer versions of clang-format will handle better.

lib/usd/ui/TreeModel.cpp
kxl-adsk: Good catch. This is because we don't recognize mayaUsdUI in https://github.com/Autodesk/maya-usd/blob/dev/.clang-format#L56. I will fix this and create a new PR.

kxl-adsk: PR opened for it https://github.com/Autodesk/maya-usd/pull/888

PR #889, opened 2020-11-04
updates for building with latest post-20.11 dev branch of core USD

mattyjams:
This PR includes a bunch of changes that go along with commits recently pushed to the dev branch of core USD. All supported releases of USD are still supported with these changes. The build.md was updated to reflect support up to the current dev branch commit.


mattyjams:
Rebased to absorb clang-format changes from #890, and re-ran clang-format v10.0.0 on the modified files.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -1559,7 +1570,7 @@ void HdVP2BasisCurves::_UpdateRepr(HdSceneDelegate* sceneDelegate, TfToken const
 HdDirtyBits HdVP2BasisCurves::GetInitialDirtyBitsMask() const
 {
     constexpr HdDirtyBits bits = HdChangeTracker::InitRepr | HdChangeTracker::DirtyExtent
-        | HdChangeTracker::DirtyInstanceIndex | HdChangeTracker::DirtyNormals
+        | HdChangeTracker::DirtyInstancer | HdChangeTracker::DirtyNormals
```
HdC-adsk, line 1573: Does this change need to be protected by `HD_API_VERSION >= 36` condition?

lib/mayaUsd/render/vp2RenderDelegate/instancer.cpp
```diff
@@ -77,12 +82,18 @@ void HdVP2Instancer::_SyncPrimvars()
 
     // Use the double-checked locking pattern to check if this instancer's
     // primvars are dirty.
-    int dirtyBits = changeTracker.GetInstancerDirtyBits(id);
-    if (HdChangeTracker::IsAnyPrimvarDirty(dirtyBits, id)) {
+    HdDirtyBits dirtyBits = changeTracker.GetInstancerDirtyBits(id);
+    if (HdChangeTracker::IsAnyPrimvarDirty(dirtyBits, id)
+        || HdChangeTracker::IsInstancerDirty(dirtyBits, id)) {
```
HdC-adsk, line 87: Does the change need conditional compilation on `HD_API_VERSION`?

lib/mayaUsd/render/vp2RenderDelegate/instancer.cpp
```diff
@@ -77,12 +82,18 @@ void HdVP2Instancer::_SyncPrimvars()
...
         std::lock_guard<std::mutex> lock(_instanceLock);
 
         // If not dirty, then another thread did the job
         dirtyBits = changeTracker.GetInstancerDirtyBits(id);
+
+#if defined(HD_API_VERSION) && HD_API_VERSION >= 36
+        _UpdateInstancer(GetDelegate(), &dirtyBits);
```
HdC-adsk, line 94: It looks like this should only be done when the instancer is dirty?

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -522,7 +527,7 @@ HdDirtyBits HdVP2Mesh::GetInitialDirtyBitsMask() const
     constexpr HdDirtyBits bits = HdChangeTracker::InitRepr | HdChangeTracker::DirtyPoints
         | HdChangeTracker::DirtyTopology | HdChangeTracker::DirtyTransform
         | HdChangeTracker::DirtyMaterialId | HdChangeTracker::DirtyPrimvar
-        | HdChangeTracker::DirtyVisibility | HdChangeTracker::DirtyInstanceIndex
+        | HdChangeTracker::DirtyVisibility | HdChangeTracker::DirtyInstancer
```
HdC-adsk, line 530: Does it need to be protected with the macro conditional compilation?

lib/usd/hdMaya/adapters/aiSkydomeLightAdapter.cpp
```diff
@@ -65,7 +65,11 @@ class HdMayaAiSkyDomeLightAdapter : public HdMayaLightAdapter
             return VtValue(light.findPlug("aiExposure", true).asFloat());
         } else if (paramName == HdLightTokens->normalize) {
             return VtValue(light.findPlug("aiNormalize", true).asBool());
+#if USD_VERSION_NUM >= 2102
```
HdC-adsk, line 68: For my own education, when should we use `USD_VERSION_NUM` and when should we use `HD_API_VERSION`?

HdC-adsk (approved):
LGTM.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
mattyjams: Good question. I asked that myself when I reviewed internally. :)

The GPU team said that `DirtyInstanceIndex` doesn't appear to be used by this code, so removing it doesn't matter. `DirtyInstancer` is used if `HD_API_VERSION > 36` but even if we're linking against older versions of USD/Hydra, setting it in the initial bits won't matter.

lib/mayaUsd/render/vp2RenderDelegate/instancer.cpp
mattyjams: This should be fine without API guards too. `IsInstancerDirty()` and the `DirtyInstancer` dirty bit it uses are both present back to at least USD 19.11.

lib/mayaUsd/render/vp2RenderDelegate/instancer.cpp
mattyjams: Yup, and the base class implementation in `HdInstancer` checks that:
https://github.com/PixarAnimationStudios/USD/blob/f7b5f25e0d7bd951439485841304aa430683ab23/pxr/imaging/hd/instancer.cpp#L119

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
mattyjams: Nope, ok here too, similar to https://github.com/Autodesk/maya-usd/pull/889#discussion_r518217643.

lib/usd/hdMaya/adapters/aiSkydomeLightAdapter.cpp
mattyjams: Yeah, sorry, this is admittedly not very obvious.

Internally, our development cadence runs faster than the public USD release cycle, and the Hydra API in particular is changing fairly rapidly. We've been needing more fine-grained API guards to deal with that as we attempt to support several releases and mixed-component builds (internal-only stuff built on top of USD) across multiple shows. There isn't really a rigid structure around when API-level guards (`HD_API_VERSION`, `USD_IMAGING_API_VERSION`, etc.) get added or incremented. It's just done as needed to address a particular API change.

So externally to Pixar, the go-to should generally be `USD_VERSION_NUM`, since we don't really intend to support individual components of the public USD distribution being built separately from each other.

PR #891, opened 2020-11-04
MAYA-106486 Fix self-specialization bug with textureless materials

JGamache-autodesk:
Thanks to @mattyjams for the heads-up and the fix. Added unit test.


mattyjams (approved):
Looks good to me! Thanks @JGamache-autodesk for extending the test to check for this!

PR #892, opened 2020-11-04
MAYA-106486 Fix map1 import

JGamache-autodesk:
When a primvar named map1 is encountered in a UsdMesh, re-use the existing map1 uvSet. Regression from PR 876 which now allows exporting map1 as itself.


mattyjams (approved):
Looks good to me! I *thought* there might have been something like this lurking in #876, but wasn't sure, so I didn't end up commenting on it there.

Thanks @JGamache-autodesk!

PR #890, opened 2020-11-04
Apply clang format

kxl-adsk:
Time to start leveraging clang-format in maya-usd. Below are updated instructions from #843 

Instructions for applying clang-format to an outstanding branch / PR
====================================================================
These instructions assume that:
- `$MAYA_USD_REPO` points to the root directory of a checkout of the maya-usd source repository
- `$COMMIT_BEFORE_FORMAT` points at the commit in the dev branch immediately before clang-format is applied (this is 0fe90bcce893dfbd4422f70bda84435f5721dd43)
- `$FORMAT_COMMIT` points at the commit in the dev branch where clang-format was applied (this is 7c64b5df1f1cdbf879aaef12ed64b7fdd77a3af1)
- `$FORMAT_BRANCH` points at the head of the branch, containing any commits after `$FORMAT_COMMIT` (dev branch)
- `$YOUR_BRANCH` is the name of the branch which does not have clang-format applied, but want to merge dev (with clang-format) into

Download and install clang-tools 10.0 (generally included as part of llvm):
- MacOS: use brew or https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/clang+llvm-10.0.0-x86_64-apple-darwin.tar.xz
- Windows: VS2019 comes with v10.0 or https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/LLVM-10.0.0-win64.exe
- Ubuntu 18.04: https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/clang+llvm-10.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz
- CentOS: build yourself - rough instructions follow:
  Make sure a modern version of cmake, and python-3, are both on your path, then:
  ```
  sudo yum install libedit-devel libxml2-devel ncurses-devel swig ninja-build devtoolset-9
  source /opt/rh/devtoolset-9/enable
  git clone -b release/10.x --single-branch --depth 1 https://github.com/llvm/llvm-project.git
  cd llvm-project
  mkdir build
  cd build
[...]

PR #895, opened 2020-11-05
Add blame ignore file with clang-format commit sha

kxl-adsk:
Suggestion from https://github.com/Autodesk/maya-usd/pull/890#issuecomment-722347918


aloysbaillet (approved):
Thanks Krystian!

PR #902, opened 2020-11-06
MAYA-106486 Import UV set mappings

JGamache-autodesk:
Applies uvChooser nodes whenever the selected UV is not the default UV.
Will merge specialized materials that were created on export solely to handle UV set names mismatches.


lib/usd/translators/meshReader.cpp
```diff
@@ -127,12 +127,13 @@ bool MayaUsdPrimReaderMesh::Read(UsdMayaPrimReaderContext* context)
         }
     }
 
-    // assign material
-    assignMaterial(mesh, _GetArgs(), meshRead.meshObject(), context);
-
     // assign primvars to mesh
     UsdMayaMeshReadUtils::assignPrimvarsToMesh(
         mesh, meshRead.meshObject(), _GetArgs().GetExcludePrimvarNames());
+
```
JGamache-autodesk, line 133: UV needs to be read and present in the mesh before the material is bound to insure proper UV linkage.

lib/usd/translators/shading/usdUVTextureReader.cpp
```diff
@@ -137,6 +140,16 @@ bool PxrMayaUsdUVTexture_Reader::Read(UsdMayaPrimReaderContext* context)
...
+    {
+        MPlug uvPlug = uvDepFn.findPlug(_tokens->outUV.GetText(), true, &status);
+        MPlug filePlug = depFn.findPlug(_tokens->uvCoord.GetText(), true, &status);
+        UsdMayaUtil::Connect(uvPlug, filePlug, false);
+    }
+    {
+        MPlug uvPlug = uvDepFn.findPlug(_tokens->outUvFilterSize.GetText(), true, &status);
+        MPlug filePlug = depFn.findPlug(_tokens->uvFilterSize.GetText(), true, &status);
+        UsdMayaUtil::Connect(uvPlug, filePlug, false);
+    }
```
JGamache-autodesk, line 152: These missing connections were revealed by UV linking.

test/lib/usd/translators/UsdImportUVSetMappingsTest/UsdImportUVSetMappings.usda
```diff
@@ -0,0 +1,424 @@
...
+        }
+    }
+}
+
+def Material "blinn3SG_uvSet1" (
+    prepend specializes = </blinn3SG>
+)
+{
+    token inputs:file5:varname = "uvSet1"
+    token not:mergeable:anymore = "becauseModified"
```
JGamache-autodesk, line 373: Testing _IsMergeableMaterial

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
```diff
@@ -51,6 +58,48 @@
...
+// mergeable material.
+bool _IsMergeableMaterial(const UsdShadeMaterial& shadeMaterial)
+{
+    if (!shadeMaterial || !shadeMaterial.HasBaseMaterial()) {
+        return false;
+    }
+    const PcpPrimIndex&   primIndex = shadeMaterial.GetPrim().GetPrimIndex();
+    PcpNodeRef            primRoot = primIndex.GetRootNode();
+    SdfPath               primPath = primIndex.GetPath();
+    const SdfLayerHandle& layer = primRoot.GetLayerStack()->GetLayerTree()->GetLayer();
```
mattyjams, line 77: I'm not sure I understand why we're trying to find the root layer here. Don't we want to be considering the fully composed stage? Materials could be added or specialized, or opinions could be expressed, across multiple layers.

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
```diff
@@ -51,6 +58,48 @@
...
+        }
+        std::vector<std::string> splitName = SdfPath::TokenizeIdentifier(path.GetName());
+        // We allow only ["inputs", "XXX", "varname"]
+        if (splitName.size() != 3 || splitName[0] != _tokens->inputs.GetString()
+            || splitName[2] != _tokens->varname.GetString()) {
+            retVal = false;
+        }
+    };
+
+    layer->Traverse(primPath, testFunction);
```
mattyjams, line 97: Along the lines of the above comment, it seems like we'd want a stage-based traversal here rather than a prim-based one.

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
```diff
@@ -89,10 +143,87 @@ MObject UsdMayaTranslatorMaterial::Read(
     return shadingEngine;
 }
 
+namespace {
+using _UVBindings = std::map<TfToken, TfToken>;
```
mattyjams, line 147: Should the `_GetUVBindingsFromMaterial()` and `_BindUVs()` functions below also go in this anonymous namespace?

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
```diff
@@ -89,10 +143,87 @@ MObject UsdMayaTranslatorMaterial::Read(
...
+}
+
+static void _BindUVs(const MDagPath& shapeDagPath, const _UVBindings& uvBindings)
+{
+    if (uvBindings.empty()) {
+        return;
+    }
+
+    MStatus status;
+    MFnMesh meshFn(shapeDagPath.node(), &status);
```
mattyjams, line 195: I think this can be constructed directly with an `MDagPath`, right?

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
```diff
@@ -89,10 +143,87 @@ MObject UsdMayaTranslatorMaterial::Read(
...
+            continue;
+        }
+        MString uvLinkCommand("uvLink -make -uvs \"");
+        uvLinkCommand += shapeDagPath.fullPathName();
+        uvLinkCommand += ".uvSet[";
+        uvLinkCommand += uvSetIndex;
+        uvLinkCommand += "].uvSetName\" -t \"";
+        uvLinkCommand += iter->second.GetText(); // texture name
+        uvLinkCommand += "\";";
+        status = MGlobal::executeCommand(uvLinkCommand);
```
mattyjams, line 218: Yikes. I take it there's no C++ API for this?

lib/usd/translators/shading/usdUVTextureReader.cpp
```diff
@@ -67,6 +67,9 @@ TF_DEFINE_PRIVATE_TOKENS(
     // UsdUVTexture Input Names
     (bias)(fallback)(scale)(wrapS)(wrapT)
 
+    // uv connections:
+    (outUvFilterSize)(uvFilterSize)(outUV)(uvCoord)
```
mattyjams, line 71: Similar to the `clang-format` comment above, these would ideally be one per line.

mattyjams (changes requested):
This overall looks good to me, but I'm not sure about the layer spelunking we're doing in `_IsMergeableMaterial()`. Am I missing something there? Can you help me understand why we're only considering the root layer? I would have thought we'd want to consider the fully composed stage instead.

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
JGamache-autodesk: There is no direct C++ API to do this. The command has some interesting logic that would have to be reproduced. This includes navigation from the texture node to the placement node, and the decision to add or reuse a uvChooser depending on the inputs of the placement node.

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
```diff
@@ -51,6 +55,63 @@
...
+        }
+        std::vector<std::string> splitName = SdfPath::TokenizeIdentifier(path.GetName());
+        // We allow only ["inputs", "<texture_name>", "varname"]
+        if (splitName.size() != 3 || splitName[0] != _tokens->inputs.GetString()
+            || splitName[2] != _tokens->varname.GetString()) {
+            retVal = false;
+        }
+    };
+
+    specializationArc.GetIntroducingLayer()->Traverse(primPath, isVarnameOpinion);
```
mattyjams, line 109: The layer-based traversal at the bottom here has the potential to be way more expensive than it needs to be. At the point where we start the traversal, all we know is that we're looking at a specialized material. The root of the traversal is the primSpec in the layer that introduced the specialize arc. That layer may also have an arbitrary number of properties and/or namespace children authored for that Material, and we're going to end up visiting **all** of those paths when really we know whether or not we should stop after inspecting only the Material primSpec and its properties.

With the caveat that this is untested, this is kind of what I'm thinking for this, beginning right after we verify that there are only two arcs:

```C++
    // This is a little more robust than grabbing a specific arc index.
    UsdPrimCompositionQuery::Filter filter;
    filter.arcTypeFilter = UsdPrimCompositionQuery::ArcTypeFilter::Specialize;
    query.SetFilter(filter);
    arcs = query.GetCompositionArcs();
    const UsdPrimCompositionQueryArc specializationArc = arcs.front();

    const SdfLayerHandle layer = specializationArc.GetIntroducingLayer();
    const SdfPrimSpecHandle primSpec = layer->GetPrimAtPath(shadeMaterial.GetPath());

    // If the primSpec that specializes the base material introduces other
    // namespace children, it can't be merged.
    if (!primSpec->GetNameChildren().empty()) {
        return false;
    }

    // Check that the only properties authored are varname inputs.
    for (const SdfPropertySpecHandle& propSpec : primSpec->GetProperties()) {
        const SdfPath propPath = propSpec->GetPath();

        const std::vector<std::string> splitName =
            SdfPath::TokenizeIdentifier(propPath.GetName());
        // We allow only ["inputs", "<texture_name>", "varname"]
        if (splitName.size() != 3u ||
                splitName[0u] != _tokens->inputs.GetString() ||
                splitName[2u] != _tokens->varname.GetString()) {
            return false
        }
    }

    return true;
```

mattyjams (changes requested):
I think the `SdfLayer::Traverse()` currently at the bottom of `_IsMergeableMaterial()` could be replaced with something less expensive. We obviously can't assume that the USD we're importing came from Maya, so it seems very possible (and probably likely) that we might be importing a specialized material with potentially lots of scene description authored on or underneath it.

mattyjams (approved):
Looks good to me! Thanks @JGamache-autodesk!

Issue #607, opened 2020-06-29
pxrUsdPreviewSurface opacity support broken following PR #574


mattyjams:
**Describe the bug**
Following the merge of #574, the `pxrUsdPreviewSurface` shading node now uses the fragment graph XML from the VP 2.0 render delegate instead of its own version. One important difference there is that the render delegate version uses an `opacity` property instead of a `transparency` property. There were no changes to `pxrUsdPreviewSurface` or its shading node override to account for this, so as a result the `opacity` control on `pxrUsdPreviewSurface` is no longer working.

This was uncovered by running the `testPxrUsdPreviewSurfaceDraw` test internally at Pixar after merging the PR. This test is unfortunately not being run in the open source repo.

I was able to fix the issue by reverting the `UsdPreviewSurface` fragment graph in vp2ShaderFragments to use `transparency` instead as it did before it was copied there from `pxrUsdPreviewSurface`. That change can be seen in https://github.com/mattyjams/maya-usd/commit/1a6628a0dd9e743035bfebfb346fce9a4fbb55f0.

I'm not exactly sure how or if that fragment graph is used by the VP2.0 render delegate, but I would guess that we don't actually want to fix this that way. It would seem then that we need to modify the shading node override for `pxrUsdPreviewSurface`, and/or modify the shading node itself. I remember running into problems though when writing the node because of the inversion from opacity to transparency, and the fact that the attribute declared in `transparencyParameter()` of the override seems to be expected to have "transparency" semantics (zero is fully opaque and one is fully transparent).

**Steps to reproduce**
Open the test scene here:
https://github.com/Autodesk/maya-usd/blob/dev/test/lib/usd/pxrUsdPreviewSurface/PxrUsdPreviewSurfaceDrawTest/PxrUsdPreviewSurfaceDrawTest.ma

You should notice that in the third "column" from the right, the spheres are all opaque and ramp from solid white at the front to middle gray in the back.

This is the image being produced when I run the test:
![PxrUsdPreviewSurfaceDrawTest_value_ramps_broken_opacity](https://user-images.githubusercontent.com/2974964/86035342-0af05180-b9f1-11ea-9520-47900669471e.png)

**Expected behavior**
The spheres in the third "column" from the right should ramp from fully transparent to fully opaque, matching the baseline image here:
[...]

PR #903, opened 2020-11-06
MAYA-106797 Fix transparency issues for pxrUsdPreviewSurface shader node

HdC-adsk:
**Issues**

1. When the `pxrUsdPreviewSurface` shader node is connected with a texture through its opacity plug, the shader fragment connection fails because there is no "opacity" parameter on the surface shader fragment to be connected with a texture fragment. This is a regression of 1c0de920f8903b9a0d9fd804094f96d15fdf7af1, which replaces "opacity" with "transparency".

2. VP2 transparency test returns false negative value and results in incorrect render.

**Changes**

* a27a8e42676cde6e84092b67608acc5e96e3576d: Reverted 1c0de920f8903b9a0d9fd804094f96d15fdf7af1. Now VP2RenderDelegate and pxrUsdPreviewSurface shader node are using the same surface shader fragment for UsdPreviewSurface, which exposes the "opacity" parameter instead of "transparency". With this change, we are able to generate the correct shader code when an opacity texture is connected to pxrUsdPreviewSurface.

* ff8f8439b39e32960b6a1072cdf64ab35f7e1e72: Reworked on #607 and fixed the transparency on/off issue for two cases: 1) When the opacity plug is connected to textures (not including anim curves), transparency is enabled. 2) When the opacity plug is not connected (or connected to anim curves), we evaluate its value and enable transparency if that value is not 1.0.


mattyjams (approved):
Looks good to me, thanks @HdC-adsk!

I addressed some of the same `clang-format` stuff in #904. I can rebase those changes if this gets merged first.

kxl-adsk:
I already merged #904 since this change is still pending review from @JGamache-autodesk. @HdC-adsk you will have to fix merge conflicts.

JGamache-autodesk (approved):

HdC-adsk:
@kxl-adsk The merge conflicts have been resolved. Please merge.

PR #904, opened 2020-11-06
wrap token definitions in clang-format off/clang-format on

mattyjams:
It looks like `clang-format` is pretty seriously mangling some of the code that defines public and private tokens, so this change wraps many of these in `// clang-format off` and `// clang-format on`.

### DIFF
--- a/lib/mayaUsd/commands/baseListShadingModesCommand.cpp
+++ b/lib/mayaUsd/commands/baseListShadingModesCommand.cpp
@@ -30,14 +30,18 @@ PXR_NAMESPACE_USING_DIRECTIVE
 
 namespace MAYAUSD_NS_DEF {
 
+// clang-format off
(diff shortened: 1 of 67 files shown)


kxl-adsk (approved):

kxl-adsk:
Thank you @mattyjams. I heard from @JGamache-autodesk that the newer version of clang-format allows excluding macros, but I wasn't able to find the actual configuration flag in the documentation.

JGamache-autodesk:
The flag name is `WhitespaceSensitiveMacros`. It makes its appearance in Clang-format 11: https://releases.llvm.org/11.0.0/tools/clang/docs/ClangFormatStyleOptions.html

kxl-adsk:
Awesome, thank you @JGamache-autodesk!

PR #905, opened 2020-11-09
MAYA-103547 Use the common utility function to check error conditions

fowlertADSK:
There is a utility function in Utils that can be shared between some of the commands (e.g. duplicate and rename).  Using it to get the correct logic and proper error message.

There is still a question about what the exact logic/message should be for the commands, but if we want to change it then that will happen in a separate PR.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
@@ -16,6 +16,7 @@
 #include "UsdUndoDuplicateCommand.h"
 
 #include "private/UfeNotifGuard.h"
+#include "private/Utils.h"
 
 #include <mayaUsd/ufe/Utils.h>
 #include <mayaUsdUtils/util.h>
@@ -59,6 +60,8 @@ void UsdUndoDuplicateCommand::primInfo(
     SdfPath&        usdDstPath,
     SdfLayerHandle& srcLayer)
 {
+    ufe::applyCommandRestriction(srcPrim, "duplicate");
+
     auto             parent = srcPrim.GetParent();
     TfToken::HashSet childrenNames;
     for (auto child : parent.GetFilteredChildren(UsdPrimIsDefined && !UsdPrimIsAbstract)) {
@@ -69,19 +72,6 @@ void UsdUndoDuplicateCommand::primInfo(
     // has a numerical suffix, increment it, otherwise append "1" to it.
     auto dstName = uniqueName(childrenNames, srcPrim.GetName());
     usdDstPath = parent.GetPath().AppendChild(TfToken(dstName));
-
-    // Iterate over the layer stack, starting at the highest-priority layer.
-    // The source layer is the one in which there exists a def primSpec, not
-    // an over.  An alternative would have beeen to call Sdf.CopySpec for
-    // each layer in which there is an over or a def, until we reach the
-    // layer with a def primSpec.  This would preserve the visual appearance
-    // of the duplicate.  PPT, 12-Jun-2018.
-    srcLayer = MayaUsdUtils::defPrimSpecLayer(srcPrim);
-    if (!srcLayer) {
-        std::string err
-            = TfStringPrintf("No prim found at %s", srcPrim.GetPath().GetString().c_str());
-        throw std::runtime_error(err.c_str());
-    }
 }
 
 /*static*/
--- a/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
@@ -36,11 +36,10 @@ UsdUndoDuplicateCommand::UsdUndoDuplicateCommand(
     const UsdPrim&   srcPrim,
     const Ufe::Path& ufeSrcPath)
     : Ufe::UndoableCommand()
-    , fSrcPrim(srcPrim)
-    , fUfeSrcPath(ufeSrcPath)
+    , _srcPrim(srcPrim)
+    , _ufeSrcPath(ufeSrcPath)
 {
-    fStage = fSrcPrim.GetStage();
-    primInfo(srcPrim, fUsdDstPath, fLayer);
+    primInfo(srcPrim, _usdDstPath, _layer);
 }
 
 UsdUndoDuplicateCommand::~UsdUndoDuplicateCommand() { }
@@ -52,7 +51,7 @@ UsdUndoDuplicateCommand::create(const UsdPrim& srcPrim, const Ufe::Path& ufeSrcP
     return std::make_shared<UsdUndoDuplicateCommand>(srcPrim, ufeSrcPath);
 }
 
-const SdfPath& UsdUndoDuplicateCommand::usdDstPath() const { return fUsdDstPath; }
+const SdfPath& UsdUndoDuplicateCommand::usdDstPath() const { return _usdDstPath; }
 
 /*static*/
 void UsdUndoDuplicateCommand::primInfo(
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
```diff
@@ -59,6 +60,8 @@ void UsdUndoDuplicateCommand::primInfo(
     SdfPath&        usdDstPath,
     SdfLayerHandle& srcLayer)
 {
+    ufe::applyCommandRestriction(srcPrim, "duplicate");
```
HamedSabri-adsk, line 63: 1- both static flavors of primInfo() can go away.  All the logics in primInfo can be moved directly to the constructor. 

https://github.com/Autodesk/maya-usd/blob/5ece36c5c45449b348d76ae73523e25179dc984f/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.h#L61

https://github.com/Autodesk/maya-usd/blob/5ece36c5c45449b348d76ae73523e25179dc984f/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.h#L53

2- fStage can go away. We don't need to save it . It can be obtained from fUfeSrcPath. 

3- I believe there is a bug around the layer. You should get layer from the srcPrim and store in 

`fLayer = GetStage()->GetEditTarget().GetLayer()`

lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
```diff
@@ -69,19 +72,6 @@ void UsdUndoDuplicateCommand::primInfo(
     // has a numerical suffix, increment it, otherwise append "1" to it.
     auto dstName = uniqueName(childrenNames, srcPrim.GetName());
     usdDstPath = parent.GetPath().AppendChild(TfToken(dstName));
```
HamedSabri-adsk, line 74: line 66-73 can be factored out. See uniqueChildName utility function:

https://github.com/Autodesk/maya-usd/blob/91994c05e287ca22d522556eee3d82eb4c93ecc4/lib/mayaUsd/ufe/Utils.cpp#L123

HamedSabri-adsk (changes requested):
It would be nice to clean up this class and also back this change up with a test case.

lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
HamedSabri-adsk: Nit-pick: these includes can go away.

#include <mayaUsd/ufe/UsdSceneItem.h>
#include <pxr/usd/usd/stage.h>

all the fxxx member variables should follow the coding standard.

fowlertADSK: 1. primInfo() doesn't even have an implementation as far as I can tell, so I'll just get rid of it.  The other primInfo static is used by Scene Item Ops so can't be moved to the constructor.

2. Yes, we can get the stage when needed.  I think I'd have to use _srcPrim.GetStage() though, fUfeSrcPath doesn't know anything about Usd.

3. Hmm.  fLayer is currently set by passing it to primInfo where it's set to the whatever layer the def primspec is on.  What you're suggesting is to change this to always grab the current edit target layer?  With the new restriction we want to enforce, I think those two should always be the same, but there is a comment in the next function down that says we chose to not just grab the edit target layer.

4. I removed one include that didn't look like it was needed.  I only see stage.h in the .cpp though, which is needed.

5. Yes, I can switch to the new naming convention for variables while I'm here.

lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
fowlertADSK: Done

fowlertADSK:
> It would be nice to clean up this class and also back this new change with a test case.
The existing test cases in testDuplicateCmd.py should cover the cases we're checking right now.  The only real difference at the moment is that the error message itself should give more details.

The requested changes in MAYA-107582 would be where we will probably want to add some new coverage.  At least that's what I think, unless you have a specific case that isn't covered right now?

lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
```diff
@@ -72,6 +72,13 @@ void UsdUndoDuplicateCommand::primInfo(
...
+
+    srcLayer = MayaUsdUtils::defPrimSpecLayer(srcPrim);
+    if (!srcLayer) {
+        std::string err
+            = TfStringPrintf("No prim found at %s", srcPrim.GetPath().GetString().c_str());
+        throw std::runtime_error(err.c_str());
+    }
 }
```
HamedSabri-adsk: Hey @fowlertADSK   There is no need to call `MayaUsdUtils::defPrimSpecLayer `. This routine can be removed all together later. It is currently called in some Ufe1.0 but I believe this can be cleaned up as well.
https://github.com/Autodesk/maya-usd/blob/91994c05e287ca22d522556eee3d82eb4c93ecc4/lib/mayaUsd/ufe/UsdHierarchy.cpp#L149

> What you're suggesting is to change this to always grab the current edit target layer? With the new restriction we want to enforce, I think those two should always be the same, but there is a comment in the next function down that says we chose to not just grab the edit target layer.

By calling applyCommandRestriction(), you are guaranteed to be the correct layer. That's the whole idea, that this restrict function puts you in a correct layer to perform your change.
This layer can be retrieved from the edit target

_layer = srcPrim.GetStage()->GetEditTarget().GetLayer();

> The other primInfo static is used by Scene Item Ops so can't be moved to the constructor.

The name primInfo is confusing and can be simplified. This is how I would imagine the constructor to be 

```
{
    ufe::applyCommandRestriction(srcPrim, "duplicate");

    _layer = _srcPrim.GetStage()->GetEditTarget().GetLayer();

    auto parentPrim = srcPrim.GetParent();
    auto dstName = uniqueChildName(parentPrim, srcPrim.GetName());
    _usdDstPath = parentPrim.GetPath().AppendChild(TfToken(dstName));
}
```
Similar to UsdUndoRenameCommand, UsdUndoDuplicateCommand can return the duplicate Scene item which can be returned by UsdSceneItemOps::duplicateItem()

```
UsdSceneItem::Ptr duplicatedItem() const;  { return _ufeDuplicateItem; }

Ufe::SceneItem::Ptr UsdSceneItemOps::duplicateItem()
{
    auto duplicateCmd = UsdUndoDuplicateCommand::create(prim(), fItem->path());
    duplicateCmd->execute();
    return duplicateCmd->duplicatedItem();
}

Ufe::Duplicate UsdSceneItemOps::duplicateItemCmd()
{
    auto duplicateCmd = UsdUndoDuplicateCommand::create(prim(), fItem->path());
    duplicateCmd->execute();
    return Ufe::Duplicate(duplicateCmd->duplicatedItem(), duplicateCmd);
}
```

I hope it makes sense.

lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
```diff
@@ -109,23 +98,23 @@ void UsdUndoDuplicateCommand::undo()
...
 #endif
 
-    fStage->RemovePrim(fUsdDstPath);
+    _srcPrim.GetStage()->RemovePrim(_usdDstPath);
 }
 
 void UsdUndoDuplicateCommand::redo()
 {
     // MAYA-92264: Pixar bug prevents redo from working.  Try again with USD
     // version 0.8.5 or later.  PPT, 28-May-2018.
```
ppt-adsk, line 115: Well clearly redo() is working, so the comment is obsolete!

fowlertADSK: @HamedSabri-adsk   This is not the comment I was talking about, I meant the comment on lines 82-84:
    // We use the source layer as the destination.  An alternate workflow
    // would be the edit target layer be the destination:
    // layer = self._stage.GetEditTarget().GetLayer()
The static duplicate function assumes that they should be on the same layer, but the constructor of the command assumes that it's the edit target.

fowlertADSK: And for the redo comment...  I originally missed this comment, but I did see a similar comment in the test.  I removed that comment and reenabled the test for redo.

HamedSabri-adsk: Yeah, that comment can also be clean up. 

I believe both srcLayer and dstLayer that are being passed to SdfCopySpec should be the same. This is the _layer that is populated in the constructor. 

My thinking is that the duplicate always happen in one single layer. Kind of similar to how rename works.

HamedSabri-adsk (approved):
LGTM! Thank you for the extra clean up. It certainly makes it easier to follow the logics.

lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
```diff
@@ -16,116 +16,107 @@
...
 #include <pxr/usd/sdf/copyUtils.h>
+#include <pxr/usd/usd/prim.h>
 #include <pxr/usd/usd/stage.h>
 
 #include <ufe/log.h>
+#include <ufe/path.h>
 #include <ufe/scene.h>
 #include <ufe/sceneNotification.h>
 
+#include <iostream>
```
HamedSabri-adsk, line 34: Nit-pick: this include can be removed.

PR #908, opened 2020-11-09
Organize Gitignore, and add more editor,platform and build specific patterns

dgovil:
This adds some general setup exclusions to the Git repo. This should apply to all Jetbrains users and a few CLI editor folks.

Just putting this in to trial our process of getting PR's in. We're in the process of cleaning up some larger internal changes for submission as well.

For details:
- `.idea/cmake-build-*` are primarily for CLion users
- The remainder are primarily for vim/emacs folks with their completions and cmake integration setups.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,19 +1,86 @@
-*build/
+##### Editor Specific Ignores #####
+
+## Common Editor Resources ##
+GTAGS
+GRTAGS
+GPATH
+*.clang_complete
+compile_flags.txt
+compile_commands.json
+# Git Merge conflict files
+*.orig
+
+## Visual Studio / Visual Studio Code ##
+*.vs
+*.vscode
+*.rsuser
+*.suo
+*.user
+*.userosscache
+*.sln.docstates
+
+## Jetbrains (CLion/PyCharm etc) ##
+.idea/
+
+## Eclipse ##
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -5,6 +5,8 @@
 *.settings/
 *.vs
 *.vscode
+.idea/
```
kxl-adsk, line 8: @dgovil would you mind adding comments to this file and making the file a little more organized into categories? See this repository for some inspiration - https://github.com/github/gitignore/tree/master/Global

dgovil: Sure thing. I'll take a pass tomorrow. I assume you're fine with me moving about any of the stuff y'all already had in that case.

kxl-adsk: Yes, please. FWIW. Apologies for some extra cleanup you will have to do here, but it will make the file easier to extend.

dgovil: Okay, organized it into 3 groups: Editor, Build and Platform. Groups denoted by `#####`. Subgroups within that denoted by `##` and then comments for the one or two ones that I felt remained ambiguous after the grouping.

kxl-adsk (approved):
Perfect. Thank you! We can merge this once we get to the bottom of the CLA discussion.

PR #912, opened 2020-11-11
Add support for Boost_NO_BOOST_CMAKE=ON for Boost::python

dgovil:
This PR is a reflection of our internal changes to support Boost_NO_BOOST_CMAKE=ON for using boost::python along custom paths. 

For a set of Boost Versions, the CMake setup made it difficult to use your own Boost library paths.
Prior to Boost 1.70, certain versions of CMake failed to find Boosts Python library due to the addition of version suffixes.
From Boost 1.70 onwards, it is possible for Boost to optionally support CMake files, but can also be problematic if you want to sub in your own paths. Occasionally without Boost_NO_BOOST_CMAKE, CMake will resolve to other versions of boost even if you specify an explicit boost root path. Hence the need for Boost_NO_BOOST_CMAKE.
See https://stackoverflow.com/a/63094072 for more details.

This doesn't appear to affect any other of the Boost requirements for Maya USD, and so we only manage that one specific library here.

### DIFF
--- a/plugin/al/CMakeLists.txt
+++ b/plugin/al/CMakeLists.txt
@@ -25,7 +25,6 @@ option(SKIP_USDMAYA_TESTS "Build tests" OFF)
 # Adding specific components forces calls to _Boost_find_library, which
 # is the rationale for listing them here.
 set(Boost_FIND_COMPONENTS
-     python
      thread
 )
 if(NEED_BOOST_FILESYSTEM)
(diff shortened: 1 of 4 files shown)


HamedSabri-adsk (approved):
Lgtm! Thank you. 

We have had many headaches with Boost python in the past and it's a bit disappointing that we have to go through all these troubles.

PR #913, opened 2020-11-11
Add FPS metadata to exported usd files if non static

dgovil:
The current exporter behavior writes out frames as if the USD is a the current scenes uiUnits, but does not actually set the fps attributes in the exported layer. This causes downstream readers to treat it as 24fps as is default.

This change adds both TimeCodesPerSecond and FramesPerSecond, if the exporter is exporting animation. This will allow downstream readers to playback at the correct timing.

As a side note...it is surprising that MTimeUnit doesn't have an API method to convert to float values. Would be great in the future if that was exposed, rather than resorting to a hardcoded mapping here.

### DIFF
--- a/lib/mayaUsd/utils/util.cpp
+++ b/lib/mayaUsd/utils/util.cpp
@@ -2078,3 +2078,172 @@ void UsdMayaUtil::GetFilteredSelectionToExport(
             dagPaths.emplace(pIter);
     }
 }
+
+float UsdMayaUtil::ConvertMTimeUnitToFloat(const MTime::Unit& unit)
+{
+    float ret = 0.f;
+    switch (unit) {
+    case MTime::k2FPS: {
+        ret = 2.f;
+    } break;
+    case MTime::k3FPS: {
+        ret = 3.f;
+    } break;
+    case MTime::k4FPS: {
+        ret = 4.f;
+    } break;
+    case MTime::k5FPS: {
+        ret = 5.f;
+    } break;
+    case MTime::k6FPS: {
+        ret = 6.f;
+    } break;
+    case MTime::k8FPS: {
+        ret = 8.f;
+    } break;
+    case MTime::k10FPS: {
+        ret = 10.f;
+    } break;
+    case MTime::k12FPS: {
+        ret = 12.f;
+    } break;
+    case MTime::k15FPS: {
+        ret = 15.f;
+    } break;
+    case MTime::k16FPS: {
+        ret = 16.f;
+    } break;
+    case MTime::k20FPS: {
+        ret = 20.f;
+    } break;
+    case MTime::k23_976FPS: {
+        ret = 23.976f;
+    } break;
+    case MTime::k24FPS: {
+        ret = 24.f;
+    } break;
+    case MTime::k25FPS: {
+        ret = 25.f;
+    } break;
+    case MTime::k29_97FPS: {
+        ret = 29.97f;
+    } break;
+    case MTime::k29_97DF: {
+        ret = 29.97f;
+    } break;
+    case MTime::k30FPS: {
+        ret = 30.f;
+    } break;
+    case MTime::k40FPS: {
+        ret = 40.f;
+    } break;
+    case MTime::k47_952FPS: {
+        ret = 47.952f;
+    } break;
+    case MTime::k48FPS: {
+        ret = 48.f;
+    } break;
+    case MTime::k50FPS: {
+        ret = 50.f;
+    } break;
+    case MTime::k59_94FPS: {
+        ret = 59.94f;
+    } break;
+    case MTime::k60FPS: {
+        ret = 60.f;
+    } break;
+    case MTime::k75FPS: {
+        ret = 75.f;
+    } break;
+    case MTime::k80FPS: {
+        ret = 80.f;
+    } break;
+#if MAYA_API_VERSION >= 20200000
+    case MTime::k90FPS: {
+        ret = 90.f;
+    } break;
+#endif
+    case MTime::k100FPS: {
+        ret = 100.f;
+    } break;
+    case MTime::k120FPS: {
+        ret = 120.f;
+    } break;
+    case MTime::k125FPS: {
+        ret = 125.f;
+    } break;
+    case MTime::k150FPS: {
+        ret = 150.f;
+    } break;
+    case MTime::k200FPS: {
+        ret = 200.f;
+    } break;
+    case MTime::k240FPS: {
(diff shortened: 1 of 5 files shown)


kxl-adsk:
I glance over the changes and notice a change in formatting - we have configured clang-format for maya-usd repository and pull-request authors are required to apply clang-format (using version 10.0.0) to the modified files. Can you please make sure to do it for all your open PRs?

dgovil:
@kxl-adsk Are you seeing any formatting changes in particular? 

I'm not sure I see the formatting issues after running clang-format manually again (did this for all the PRs). I ran `clang-format -i` on the modified `.cpp` and `.h` files, and the files were identical.  Additionally I have clang format running as part of my CLion setup and it's not detecting any style infractions.

Also tried running the command specified in [your PR](https://github.com/Autodesk/maya-usd/pull/890) `find -E . -regex "$(echo '.*('$(echo $(cat .clang-format-include) | tr ' ' '\|')')')" -exec clang-format -i {} \+` , however that modified a bunch of files that are not part of any of my PRs:

```
        modified:   lib/mayaUsd/fileio/transformWriter.cpp
        modified:   lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
        modified:   lib/mayaUsd/render/vp2RenderDelegate/render_delegate.h
        modified:   lib/mayaUsd/render/vp2RenderDelegate/render_param.h
        modified:   lib/mayaUsd/ufe/UsdAttributes.cpp
        modified:   lib/mayaUsd/ufe/UsdRotatePivotTranslateUndoableCommand.h
        modified:   lib/mayaUsd/ufe/UsdRotateUndoableCommand.h
        modified:   lib/mayaUsd/ufe/UsdScaleUndoableCommand.h
        modified:   lib/mayaUsd/ufe/UsdTranslateUndoableCommand.h
        modified:   lib/usd/translators/nurbsSurfaceWriter.cpp
        modified:   lib/usd/utils/SIMD.h
        modified:   plugin/al/lib/AL_USDMaya/AL/usdmaya/Global.h
        modified:   plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.h
        modified:   plugin/al/mayautils/AL/maya/event/MayaEventManager.cpp
        modified:   plugin/al/mayautils/AL/maya/event/MayaEventManager.h
        modified:   plugin/al/mayautils/AL/maya/utils/FileTranslatorOptions.cpp
        modified:   plugin/al/mayautils/AL/maya/utils/FileTranslatorOptions.h
        modified:   plugin/al/translators/Mesh.cpp
        modified:   plugin/pxr/maya/lib/usdMaya/referenceAssembly.cpp
```

For reference, I'm using your clang format file from bb0704ab6cfde7f158419837a98a8a5aa477814d
My clang-format version is 11.0.0 on a Mac. Is it possible that our clang formats are configured differently perhaps?

kxl-adsk:
It's the change in utils.h where I was expecting indentation - https://github.com/Autodesk/maya-usd/pull/913/files#diff-68214d17766b3d5a5678b676f7939a4d4899e99d1f5152a8a8c0e24e933d7078R575

But the more important is the version of clang-format. You mentioned version **11.0.0** but we ask developers to all use version **10.0.0.** This is because different versions produce different results for the same clang-format configuration file.

dgovil:
I tried formatting the files again with clang-format 10 (downloaded from the link in https://github.com/Autodesk/maya-usd/pull/890 since homebrew no longer has a recipe for 10 available) and it's still not marking these files as needing changes. Running it across the entirety of the repo did show a different set of modified files than clang-format 11 so that does show the versions vary quite a bit.

Just to confirm though, if you run clang-format on this branch, does it change `utils.h`? I just want to verify if it's an issue in the config versus something on our end here. 

Here are the results of running clang-format 10 across the whole repo on my end:

[formatting.patch.txt](https://github.com/Autodesk/maya-usd/files/5531933/formatting.patch.txt)

```
	modified:   lib/mayaUsd/ufe/UsdAttributes.cpp
	modified:   plugin/al/mayautils/AL/maya/event/MayaEventManager.cpp
	modified:   plugin/al/mayautils/AL/maya/event/MayaEventManager.h
	modified:   plugin/al/mayautils/AL/maya/utils/FileTranslatorOptions.cpp
```

kxl-adsk:
Re `utils.h` - you are right, I must have looked at it too quickly.

Re clang-format version and what's currently still showing up after running it. Indeed, we still have 4 files that are not showing a difference after you run clang-format v10.0.0 on the entire repo. We have discovered just recently that in some cases, clang-format may have to be applied more than once. I will open a PR for the remaining files.

dgovil:
Ok great to know it's not some deeper issue. I'll move around the unittests per the other PR and then update this one.

dgovil:
Okay, updated the test location. Hopefully should be good to go, but let me know if you see anything else of concern. I'll update the other PRs shortly too.

test/lib/usd/translators/testUsdExportLayerAttributes.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python
```
mattyjams, line 1: It's mostly a vestige from the core USD repo at this point, but the rest of the tests here use this shebang line:

```Python
#!/pxrpythonsubst
```

Just for consistency, I'd suggest using that. That way if/when we do any kind of shebang replacement, we'll catch this test too.

lib/mayaUsd/utils/util.cpp
```diff
@@ -2078,3 +2078,172 @@ void UsdMayaUtil::GetFilteredSelectionToExport(
...
+        ret = 44100.f;
+    } break;
+    case MTime::k48000FPS: {
+        ret = 48000.f;
+    } break;
+    case MTime::kHours: {
+        ret = (1.0 / 3600);
+    } break;
+    case MTime::kMinutes: {
+        ret = (1.0 / 60);
```
mattyjams, line 2230: Might throw the `f` on the operands here just to keep  everything in `float`-land and avoid the implicit downcast from `double`.

mattyjams (approved):
Just some small cosmetic suggestions, but otherwise looks good to me! Thanks @dgovil!

test/lib/usd/translators/testUsdExportLayerAttributes.py
dgovil: Done!

lib/mayaUsd/utils/util.cpp
dgovil: Done

PR #914, opened 2020-11-11
Fixes for out of bound errors causing segfaults

dgovil:
We discovered some out of bounds errors that we hit when exporting certain skeletons. This would result in segfaults.
These are our internal fixes to prevent the crashing.


kxl-adsk:
@dgovil Can you add regression tests to cover the cases you fixed?

dgovil:
Sure, good call. I'll try and pare down an asset.

lib/usd/translators/jointWriter.cpp
```diff
@@ -184,9 +184,10 @@ static bool _FindDagPoseMembers(
 
     MPlugArray inputs;
 
-    indices->clear();
-    indices->resize(
-        std::min(membersPlug.numElements(), static_cast<unsigned int>(dagPaths.size())), -1);
+    indices.clear();
+    // TODO: (yliangsiew) Need to know why we even have mismatched array sizes here.
+    indices.resize(
+        std::max(membersPlug.numElements(), static_cast<unsigned int>(dagPaths.size())), -1);
```
mattyjams, line 190: The way this reads to me, I actually don't think either of these are correct. Based on what the code is attempting to do, I think the intent was for `indices` to have the same size as `dagPaths`:

```C++
indices.resize(dagPaths.size(), -1);
```

We look through the elements of `membersPlug` below and mark the indices of all of the nodes represented there. If there are elements remaining in `dagPaths` that were **not** represented, that ends up being an error and we return false.

Note that `indices` also needs to be a `std::vector` of `int`, not `unsigned int`, for the default value of -1 to be meaningful.

lib/usd/translators/jointWriter.cpp
```diff
@@ -197,19 +198,20 @@ static bool _FindDagPoseMembers(
...
-                (*indices)[it->second] = i;
+                indices[it->second] = i;
             }
         }
     }
 
     // Validate that all of the input dagPaths are members.
-    for (size_t i = 0; i < indices->size(); ++i) {
-        int index = (*indices)[i];
+    for (size_t i = 0; i < dagPaths.size(); ++i) {
```
mattyjams, line 207: If I'm on the right track with the comments above, then I think it's still appropriate to iterate over `indices` here rather than `dagPaths`, since they will both be the same size, and `indices` is the thing we've been filling with data.

mattyjams (changes requested):
I'm actually not all that familiar with this code, but after a read through I think there were a few issues here from before your changes. I think they can be addressed relatively easily and would also address the crashes you're seeing, but apologies in advance if the feedback here ventures off a bit from what was presented in this PR. :)

In short, I think if `indices`/`memberIndices` is a vector of `int` instead of `unsigned int`, and if its size is always exactly `dagPaths.size()`, the code makes a lot more sense.

dgovil:
Thanks for the feedback! I'll check in with our engineer who was fixing this for us and we'll take another pass at it, and update the PR accordingly.

HamedSabri-adsk:
@dgovil   can you please describe the steps that is causing the segfaults? I am still not clear what and when this happens.

I also agree with Matt observations based on what I read so far.

dgovil:
@HamedSabri-adsk  the segfaults happen when we export specific skinned asset out of Maya. This only happens for a few assets that trigger an out of bounds error, whereas most are fine.

We're trying to create a shareable version of the asset to demo the issue. Proving to be a little tricky to make it non revalatory while preserving the crashing, but I'm hoping we'll be able to provide a test case next week.

dgovil:
@kxl-adsk I've added a synthetic asset that causes the issue on our end as a test.

Our engineer (who I will get on our CLA soon so he can contribute directly) has narrowed it down to being a heap corruption that occurs at a higher rate if you export this asset multiple times. It's not 100% consistent so I am only exporting it once here, since it seems really temperamental to various conditions. I'm not sure this will necessarily help catch regressions, but it should hopefully be at least a good point of reference.

Here is his lldb backtrace that shows when we hit the issue. 
[crashlog.txt](https://github.com/Autodesk/maya-usd/files/5557171/crashlog.txt)


```
Status: in _BeginWriting at line 236 of /Users/yliangsiew/Git/work/dev/maya-usd/lib/mayaUsd/fileio/jobs/writeJob.cpp -- Opening layer '/Users/yliangsiew/Downloads/tmp-2D50DBFF-EC46-204F-A7E9-CFAFD368CAB7.usdc' for writing
Maya(53410,0x10018fe00) malloc: Incorrect checksum for freed object 0x224021d08: probably modified after being freed.
Corrupt value: 0x3900000038
Maya(53410,0x10018fe00) malloc: *** set a breakpoint in malloc_error_break to debug
Process 53410 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007fff2056a462 libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff2056a462 <+10>: jae    0x7fff2056a46c            ; <+20>
    0x7fff2056a464 <+12>: mov    rdi, rax
    0x7fff2056a467 <+15>: jmp    0x7fff205646a1            ; cerror_nocancel
    0x7fff2056a46c <+20>: ret
Target 0: (Maya) stopped.
```

lib/mayaUsd/utils/diagnosticDelegate.cpp
```diff
@@ -127,7 +127,7 @@ void UsdMayaDiagnosticDelegate::IssueWarning(const TfWarning& warning)
     }
 
     if (ArchIsMainThread()) {
-        MGlobal::displayWarning(warning.GetCommentary().c_str());
+        MGlobal::displayWarning(MString(warning.GetCommentary().c_str()));
```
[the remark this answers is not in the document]
dgovil: Just replying on behalf of our engineer debugging this...

Ok, after some investigation, the heap corruption was affecting this part of the
memory (i.e. printing the string). The implicit conversion itself is not the
issue and that explicit cast is now removed.

lib/usd/translators/jointWriter.cpp
```diff
@@ -164,10 +164,10 @@ static MObject _FindBindPose(const MDagPath& dagPath)
...
     const std::vector<MDagPath>& dagPaths,
-    std::vector<unsigned int>*   indices)
+    std::vector<unsigned int>&   indices)
 {
     MStatus status;
-    MPlug   membersPlug = dagPoseDep.findPlug("members", &status);
+    MPlug   membersPlug = dagPoseDep.findPlug("members", false, &status);
     CHECK_MSTATUS_AND_RETURN(status, false);
 
     // Build a map of dagPath->index.
```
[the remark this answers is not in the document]
dgovil: Done. Thanks for pointing out that existing type.

lib/usd/translators/jointWriter.cpp
dgovil: I think this has to remain `unsigned int`, because it's used to store
potentially the max. no. of elements in an array plug, which is an `unsigned
int`:
```
for (unsigned int i = 0; i < membersPlug.numElements(); ++i) {
    MPlug memberPlug = membersPlug[i];
    memberPlug.connectedTo(inputs, /*asDst*/ true, /*asSrc*/ false);
    for (unsigned int j = 0; j < inputs.length(); ++j) {
        MObjectHandle connNode(inputs[j].node());
        auto          it = pathIndexMap.find(connNode);
        if (it != pathIndexMap.end()) {
            (*indices)[it->second] = i;
        }
    }
}
```
However, if we can assume that `dagPaths` and `indices` must be the same size,
then the default value of `-1` (which I agree makes zero sense) can now be
dropped. I just did `std::max` as a hacky hotfix previously.

lib/usd/translators/jointWriter.cpp
dgovil: This is also now reverted since we are betting on the assumption that `dagPaths`
and `indices` must be the same size and that the `pathIndexMap`'s 2nd member
will not return a value larger than said size.

lib/usd/translators/jointWriter.cpp
mattyjams: Fair enough. I buy that we want to support the maximum possible number of elements in an array plug. I'd argue though that as it stands now, we would hit an error for any elements with indices that exceed the capacity of `int` because of this:
https://github.com/Autodesk/maya-usd/blob/566b7a006f867d07c395ae7911e543a60bda6b00/lib/usd/translators/jointWriter.cpp#L200
We're grabbing an `unsigned int` there and stuffing it in a signed `int`, so any values larger than `std::numeric_limits<int>::max()` are going to wrap around and become negative. The `if (index < 0) {` that follows will then be `true` and we'll bail out in error.
Admittedly, to reach values that trigger that, `membersPlug.numElements()` would need to be up above two billion, so that particular issue is mostly academic. :)

The main issue though is that we need a way to detect whether all of the passed in `dagPaths` were indeed present in the `members` plug. Previously, I think the `indices` vector was trying to do double duty with the `-1` default value and mark which `dagPaths` had been seen so that the final loop could verify that they were all accounted for. We don't have any default value now, so with uninitialized values there's no way to tell if one was missing. But there may also not be an available `unsigned int` value we could use as a sentinel in `indices`. Maybe we just need a separate structure to account for that? Something like this after we do the `indices.resize()`?:
```C++
std::vector<bool> seenPaths(dagPaths.size(), false);
```

Then in the `membersPlug.numElements()` loop, we just mark each path as having been seen when we store an index for it:
```C++
seenPaths[it->second] = true;
```

The validation loop at the end would then just check for any `false` values.

kxl-adsk:
@dgovil Please check the PF results - your new test if failing - https://github.com/Autodesk/maya-usd/runs/1418407982?check_suite_focus=true. You can see logs after downloading the artifact produced by the check.

dgovil:
@kxl-adsk It looked to be caused by requires for Turtle. I removed all the requires from the `.ma` file. I believe that should let it work, but I don't have a good way on my end to verify since the plugins are part of stock Maya.

Is there a docker image or sanitization script available to do more thorough tests with?

kxl-adsk:
You can run tests with the provided `build.py` script. This is what we are running internally as part of the preflight.

dgovil:
Yeah we use that too, but it still seems to load plugins like Turtle,  which the CI environment seems to have stripped out? It's not a big deal. I don't anticipate checking in too many maya files, but it does mean that we'd hit situations like above where the Maya file had unavailable requires, even though the base maya install has them available. 

Anyway, I digress. I'll see if I can restrict our test setup here to only allow a subset of plugins to prevent requires issues in the future.

lib/usd/translators/jointWriter.cpp
dgovil: Thanks for catching that. You're right; there's still potential for overflow here. I've gone ahead and made the requested changes (i.e. adding a visited array) that should hopefully avoid this issue.

lib/usd/translators/jointWriter.cpp
```diff
@@ -167,27 +167,22 @@ static bool _FindDagPoseMembers(
...
 
     MPlugArray inputs;
 
+    size_t numDagPaths = dagPaths.size();
     indices->clear();
-    indices->resize(
-        std::min(membersPlug.numElements(), static_cast<unsigned int>(dagPaths.size())), -1);
+    indices->resize(numDagPaths);
 
+    std::vector<uint8_t> visitedIndices(numDagPaths, 0);
```
mattyjams, line 185: I would think `std::vector<bool>` here would be more compact?

lib/usd/translators/jointWriter.cpp
```diff
@@ -198,14 +193,15 @@ static bool _FindDagPoseMembers(
...
         }
     }
 
     // Validate that all of the input dagPaths are members.
-    for (size_t i = 0; i < indices->size(); ++i) {
-        int index = (*indices)[i];
-        if (index < 0) {
+    for (size_t i = 0; i < visitedIndices.size(); ++i) {
+        int index = visitedIndices[i];
+        if (index != 1) {
```
mattyjams, line 204: We still need to pull a value from `indices` here. The `visitedIndices` value is going to be either `0` or `1`:

```C++
    for (size_t i = 0u; i < visitedIndices.size(); ++i) {
        const bool visited = visitedIndices[i];
        if (!visited) {
            ... // Error
        }
        const unsigned int index = (*indices)[i];
        ...
```

Or if you prefer to use `uint8_t`:
```C++
    for (size_t i = 0u; i < visitedIndices.size(); ++i) {
        const uint8_t visited = visitedIndices[i];
        if (visited != 1u) {
            ... // Error
        }
        const unsigned int index = (*indices)[i];
        ...
```

mattyjams (changes requested):

lib/usd/translators/jointWriter.cpp
ysiewappl: (Hi! I'm the engineer that's working on this PR and previously implemented our hotfix internally, and making @dgovil suffer being the middleman :P)

I think I generally avoid `std::vector<bool>` as the guarantees about compactness and contiguous memory storage vary across implementation, along with a host of other issues (i.e. http://www.gotw.ca/publications/N1211.pdf and https://isocpp.org/blog/2012/11/on-vectorbool). It's not a complete dealbreaker, but I'd argue that an array of ints that is guaranteed to be stored contiguously without worrying about future issues is easier than saving on bytes here since we're generally going to be iterating over the entire array in the common case. But if there's a strong opinion that we should squeeze out every last bit we can, again, not a dealbreaker; just let me know and I can swap it out.

lib/usd/translators/jointWriter.cpp
ysiewappl: You're right; this is a silly mistake on my part, I totally missed that the warning string would have the wrong index otherwise. I have fixed this internally and @dgovil will submit the update soon. Thanks for catching this!

lib/usd/translators/jointWriter.cpp
mattyjams: Hey @ysiewappl! Thanks for the info! I suspected that might be the case, but just wasn't sure without any additional context. No strong opinions in favor of `std::vector<bool>`, so this is fine by me!

lib/usd/translators/jointWriter.cpp
dgovil: Pushed the changes.

mattyjams: Cool, no problem! With the truncated diff GitHub was showing, I failed to notice that we actually need the `index` value **in** the error message, so apologies for the slightly misleading suggestions. :) Sounds like you've got it covered though. Thanks!

kxl-adsk:
@dgovil Last PF came back again all red and I don't think it has anything to do with warnings about missing nodes (mostly because plugins were not loaded when we run tests in mayapy). 

I see one runtime error and I wonder if it happens for you as well when this test passes:
```    
Runtime Error: in _VerifyOrMakeSkelRoot at line 105 of S:\jenkins\workspace\ecg-mayausd-branch-preflight-python3-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\fileio\translators\skelBindingsProcessor.cpp -- 
The prim </joint1> is a root prim, so it has no ancestors that can be converted to a SkelRoot. 
(USD requires that skinned meshes and skeletons be encapsulated under a SkelRoot.) 
Try grouping this prim under a parent group.
```

dgovil:
@kxl-adsk apologies, my bad, the Maya file had gotten messed up between some merges and removing the Turtle plugins etc... 
I restored it and took a second pass at cleaning it up. Also adjusted the test slightly.

Everything should be running fine now, fingers crossed. Verified on our end

Edit: Actually one sec...it seems like the PR is not picking up my latest commit. d263ad8fddfcefac394758dad3a3d162bfef29bf should be the PR commit. I'll look into what's up...

dgovil:
Okay it must have been a weird Github server side snafu. It was stuck on an old commit, but has since propagated to this PR.
Should be good now.

mattyjams (approved):
Looks good to me now! Sorry we seem to have set a bit of a trap for you with this, but thanks @dgovil and @ysiewappl for fixing it up!

HamedSabri-adsk (approved):

PR #917, opened 2020-11-12
Add opacity threshold to USD Preview Surface Shader

dgovil:
Opacity Threshold is an attribute on the USD preview surface shader that wasn't previously exposed to Maya, however is useful for artists to define the cutoff level for their opacity maps.

This exposes it in the attribute editor, makes sure it serializes properly to USD and that it clips the opacity accordingly. 

There's a very minimal test included here that we used to check after merging upstream commits, but I hope we can add more material serialization tests in the future so I'm including it here.

### DIFF
--- a/lib/usd/pxrUsdPreviewSurface/usdPreviewSurface.cpp
+++ b/lib/usd/pxrUsdPreviewSurface/usdPreviewSurface.cpp
@@ -42,6 +42,7 @@ TF_DEFINE_PUBLIC_TOKENS(
     PxrMayaUsdPreviewSurfaceTokens,
     PXRUSDPREVIEWSURFACE_USD_PREVIEW_SURFACE_TOKENS);
 
+
 /* static */
 void* PxrMayaUsdPreviewSurface::creator() { return new PxrMayaUsdPreviewSurface(); }
 
@@ -60,6 +61,7 @@ MStatus PxrMayaUsdPreviewSurface::initialize()
     MObject normalAttr;
     MObject occlusionAttr;
     MObject opacityAttr;
+    MObject opacityThresholdAttr;
     MObject roughnessAttr;
     MObject specularColorAttr;
     MObject useSpecularWorkflowAttr;
@@ -228,6 +230,24 @@ MStatus PxrMayaUsdPreviewSurface::initialize()
     status = addAttribute(opacityAttr);
     CHECK_MSTATUS_AND_RETURN_IT(status);
 
+    opacityThresholdAttr = numericAttrFn.create(
+        PxrMayaUsdPreviewSurfaceTokens->OpacityThresholdAttrName.GetText(),
+        "opt",
+        MFnNumericData::kFloat,
+        0.0,
+        &status);
+    CHECK_MSTATUS_AND_RETURN_IT(status);
+    status = numericAttrFn.setSoftMin(0.0);
+    CHECK_MSTATUS_AND_RETURN_IT(status);
+    status = numericAttrFn.setSoftMax(1.0);
+    CHECK_MSTATUS_AND_RETURN_IT(status);
+    status = numericAttrFn.setKeyable(true);
+    CHECK_MSTATUS_AND_RETURN_IT(status);
+    status = numericAttrFn.setAffectsAppearance(true);
+    CHECK_MSTATUS_AND_RETURN_IT(status);
+    status = addAttribute(opacityThresholdAttr);
+    CHECK_MSTATUS_AND_RETURN_IT(status);
+
     roughnessAttr = numericAttrFn.create(
         PxrMayaUsdPreviewSurfaceTokens->RoughnessAttrName.GetText(),
         "rgh",
@@ -345,6 +365,8 @@ MStatus PxrMayaUsdPreviewSurface::initialize()
     CHECK_MSTATUS_AND_RETURN_IT(status);
     status = attributeAffects(opacityAttr, outColorAttr);
     CHECK_MSTATUS_AND_RETURN_IT(status);
+    status = attributeAffects(opacityThresholdAttr, outColorAttr);
+    CHECK_MSTATUS_AND_RETURN_IT(status);
     status = attributeAffects(roughnessAttr, outColorAttr);
     CHECK_MSTATUS_AND_RETURN_IT(status);
     status = attributeAffects(specularColorAttr, outColorAttr);
@@ -353,6 +375,8 @@ MStatus PxrMayaUsdPreviewSurface::initialize()
     CHECK_MSTATUS_AND_RETURN_IT(status);
 
     status = attributeAffects(opacityAttr, outTransparencyAttr);
+    CHECK_MSTATUS_AND_RETURN_IT(status);
+    status = attributeAffects(opacityThresholdAttr, outTransparencyAttr);
     CHECK_MSTATUS_AND_RETURN_IT(status);
 
     status = attributeAffects(opacityAttr, outTransparencyOnAttr);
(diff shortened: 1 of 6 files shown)


test/lib/testMayaUsdExportPreviewShader.py
```diff
@@ -0,0 +1,63 @@
...
+    cmds.select(node)
+    cmds.hyperShade(assign=shader)
+    return shader
+
+class MayaUsdExportPreviewSurface(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cmds.loadPlugin('mayaUsdPlugin')
+        cls.temp_dir = tempfile.mkdtemp()
```
JGamache-autodesk, line 42: You might want to consider moving this test to test/lib/usd/translators 
In that folder there are utilities that will help setup a full test environments in the build area where the output files will be preserved, allowing easy investigation in case of problems. Look for `inputPath = fixturesUtils.setUpClass(__file__)` in files like https://github.com/Autodesk/maya-usd/blob/dev/test/lib/usd/translators/testUsdExportRfMShaders.py

dgovil: @JGamache-autodesk Thanks! I'll switch over our tests to live there. Quick question regarding naming...
I've been calling our tests `testMayaUsdxxx` but the ones in the translator section are `testUsdxxx` or `textUsdMayaxxx`. Which naming should I be going with?

kxl-adsk: We haven't really set any naming convention for tests other than it should start with `test`. For translators, please follow the naming convention you see in` https://github.com/Autodesk/maya-usd/blob/dev/test/lib/usd/translators/`, i.e. `testUsdxxx`.

dgovil: Okay moved this over.

test/lib/usd/translators/testUsdExportPreviewShader.py
```diff
@@ -0,0 +1,69 @@
...
+from maya import cmds
+from maya import standalone
+from pxr import Usd
+
+
+def apply_preview_surface(node, shader=None):
+    """Apply a USD Preview Surface shader to the given node"""
+    shader = shader or cmds.shadingNode('usdPreviewSurface', asShader=True)
+    cmds.select(node)
+    cmds.hyperShade(assign=shader)
```
JGamache-autodesk, line 32: Sadly, the `hypershade` command is not available in batch mode. This means that the shader must be setup the painful way.
I think the quickest way out of this would be to add opacity threshold checks in the existing `test\lib\usd\translators\testUsdExportImportRoundtripPreviewSurface.py` test.

dgovil: Ah good to know. We've been running tests in UI mode so it didn't come up. I'll move this over to the other test you mentioned.

kxl-adsk: I'm not sure if you have privileges to change the Assignee, but this is the way to trigger a preflight for pull-requests. Results for your pull-request can be found in the checks tab - see https://github.com/Autodesk/maya-usd/pull/917/checks

dgovil: Okay switched over to being in the `testUsdExportImportRoundtripPreviewSurface`.
Also set it up so it imports the value as well.

dgovil: @kxl-adsk I do not have that permission it looks like.

JGamache-autodesk (approved):

lib/usd/pxrUsdPreviewSurface/usdPreviewSurface.cpp
```diff
@@ -399,7 +424,16 @@ MStatus PxrMayaUsdPreviewSurface::compute(const MPlug& plug, MDataBlock& dataBlo
             = depNodeFn.attribute(PxrMayaUsdPreviewSurfaceTokens->OpacityAttrName.GetText());
         const MDataHandle opacityData = dataBlock.inputValue(opacityAttr, &status);
         CHECK_MSTATUS(status);
-        const float opacity = opacityData.asFloat();
+        float opacity = opacityData.asFloat();
+
+        const MDataHandle opacityThresholdData
```
mattyjams, line 429: To remove needing to make it a static member, I think we'd re-fetch `opacityThresholdAttr` by name here similar to what we're doing with `opacityAttr` just above.

lib/usd/pxrUsdPreviewSurface/usdPreviewSurface.h
```diff
@@ -76,6 +77,8 @@ class PxrMayaUsdPreviewSurface : public MPxNode
     PXRUSDPREVIEWSURFACE_API
     MStatus compute(const MPlug& plug, MDataBlock& dataBlock) override;
 
+    static MObject opacityThresholdAttr;
+
```
mattyjams, line 81: As noted elsewhere, I don't think we want to expose this as a `public static` member.

test/lib/usd/translators/testUsdExportImportRoundtripPreviewSurface.py
```diff
@@ -68,6 +68,7 @@ def testUsdPreviewSurfaceRoundtrip(self):
         cmds.setAttr(material_node + ".specularColor", 0.125, 0.25, 0.75,
                      type="double3")
         cmds.setAttr(material_node + ".useSpecularWorkflow", True)
+        cmds.setAttr("{}.opacityThreshold".format(material_node), 0.5)
```
mattyjams, line 71: Sorry, this is a really obnoxious nit-pick, but I'd stick with the convention already in use here and use `+` rather than `format()`. I actually prefer what you used personally, but just so someone looking at this code later doesn't wonder whether there's a reason it's treated differently from the other attributes:
```Python
cmds.setAttr(materialNode + ".opacityThreshold", 0.5)
```

mattyjams (changes requested):
Thanks for this @dgovil! Just two things I think we should address before it goes in:
- I don't think we need `opacityThresholdAttr` to be `public`/`static`, especially since none of the other attributes are.
- I think we want to use `<` instead of `<=` when testing the threshold.

Other than that, what's here looks good to me!

Note though that this is only part of the story. What's in `compute()` really only comes into play for software renders, or in preparation for feeding values into the shading fragment graph, so there's another chunk of work to this where we'd want to add support for `opacityThreshold` in the viewport.

We used to have a mechanism for keeping USD's `previewSurface.glslfx` and the shading fragments used by `pxrUsdPreviewSurface` in sync (or at least flagging when they were out of sync) back when they were both in the core USD repo, but since the split we've lost that. There have been a bunch of other changes to USD's implementation recently and I've been meaning to take a pass at sync'ing the two back up again.

lib/usd/pxrUsdPreviewSurface/usdPreviewSurface.cpp
```diff
@@ -42,6 +42,9 @@ TF_DEFINE_PUBLIC_TOKENS(
     PxrMayaUsdPreviewSurfaceTokens,
     PXRUSDPREVIEWSURFACE_USD_PREVIEW_SURFACE_TOKENS);
 
+// Attributes
+MObject PxrMayaUsdPreviewSurface::opacityThresholdAttr;
```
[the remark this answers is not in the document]
JGamache-autodesk: Yes, the issue is that there are 2 usdPreviewSurface shaders, one is in mayaUsdPlugin and the other in pxrUsdPreviewSurface. They both share the implementation seen here, so you would need two static attributes and a way to map each one by plugin. This is why refetching by name is done.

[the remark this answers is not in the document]
dgovil: You're right. I moved it to be initialized like the rest. We had some other (deprecated) internal code that depended on that IIRC, but this cleans that up.

lib/usd/pxrUsdPreviewSurface/usdPreviewSurface.cpp
```diff
@@ -399,7 +424,16 @@ MStatus PxrMayaUsdPreviewSurface::compute(const MPlug& plug, MDataBlock& dataBlo
...
         CHECK_MSTATUS(status);
-        const float opacity = opacityData.asFloat();
+        float opacity = opacityData.asFloat();
+
+        const MDataHandle opacityThresholdData
+            = dataBlock.inputValue(opacityThresholdAttr, &status);
+        CHECK_MSTATUS(status);
+        const float opacityThreshold = opacityThresholdData.asFloat();
+
+        if (opacity <= opacityThreshold) {
```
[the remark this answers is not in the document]
dgovil: Good catch. Switched to `<`

dgovil:
@mattyjams I fixed up your notes. Agreed, that this doesn't resolve the actual shader work. We hadn't really bothered on our end since shader transparency was broken till recently. If/when you do sync up the shaders, that would be great.

mattyjams (approved):
Cool, thanks @dgovil! Looks great now!

No problem about the viewport support. I'll see if I can carve out any time to take a crack at that myself in the next few weeks.

kxl-adsk:
@mattyjams 

> We used to have a mechanism for keeping USD's previewSurface.glslfx and the shading fragments used by pxrUsdPreviewSurface in sync (or at least flagging when they were out of sync) back when they were both in the core USD repo, but since the split we've lost that. There have been a bunch of other changes to USD's implementation recently and I've been meaning to take a pass at sync'ing the two back up again.

Yes, we are aware of this too. It's a task on our end as well, but we haven't got a chance to prioritize it yet.

PR #918, opened 2020-11-12
Apply clang-format two more times to make sure we have everything formatted

kxl-adsk:
After PR #890 we still had few files that weren't fully formatted by clang-format
```
	modified:   lib/mayaUsd/ufe/UsdAttributes.cpp
	modified:   plugin/al/mayautils/AL/maya/event/MayaEventManager.cpp
	modified:   plugin/al/mayautils/AL/maya/event/MayaEventManager.h
	modified:   plugin/al/mayautils/AL/maya/utils/FileTranslatorOptions.cpp
```

For some reason, clang-format required to be applied twice to the above files (and to be precise, three times for lib/mayaUsd/ufe/UsdAttributes.cpp).

It's not something I was aware of, but luckily it's only these 4 files we missed.

### DIFF
--- a/lib/mayaUsd/ufe/UsdAttributes.cpp
+++ b/lib/mayaUsd/ufe/UsdAttributes.cpp
@@ -127,85 +127,17 @@ UsdAttributes::getUfeTypeForAttribute(const PXR_NS::UsdAttribute& usdAttr) const
 {
     // Map the USD type into UFE type.
     static const std::unordered_map<size_t, Ufe::Attribute::Type> sUsdTypeToUfe {
-        { SdfValueTypeNames->Bool.GetHash(), Ufe::Attribute::kBool }, // bool
-        //		{SdfValueTypeNames->UChar.GetHash(), Ufe::Attribute::kUnknown},
-        //// uint8_t
-        { SdfValueTypeNames->Int.GetHash(), Ufe::Attribute::kInt }, // int32_t
-        //		{SdfValueTypeNames->UInt.GetHash(), Ufe::Attribute::kUnknown},
-        //// uint32_t
-        //		{SdfValueTypeNames->Int64.GetHash(), Ufe::Attribute::kInt},
-        //// int64_t 		{SdfValueTypeNames->UInt64.GetHash(),
-        /// Ufe::Attribute::kUnknown}, / uint64_t
-        //		{SdfValueTypeNames->Half.GetHash(), Ufe::Attribute::kUnknown},
-        //// GfHalf
-        { SdfValueTypeNames->Float.GetHash(), Ufe::Attribute::kFloat },      // float
-        { SdfValueTypeNames->Double.GetHash(), Ufe::Attribute::kDouble },    // double
-        { SdfValueTypeNames->String.GetHash(), Ufe::Attribute::kString },    // std::string
-        { SdfValueTypeNames->Token.GetHash(), Ufe::Attribute::kEnumString }, // TfToken
-        //		{SdfValueTypeNames->Asset.GetHash(), Ufe::Attribute::kUnknown},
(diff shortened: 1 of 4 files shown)


kxl-adsk:
FYI @dgovil

plugin/al/mayautils/AL/maya/event/MayaEventManager.cpp
```diff
@@ -921,27 +926,31 @@ void MayaEventHandler::initDagMessage(MayaCallbackInfo& cbi)
...
+                                  &cbi); */
         break;
     case DagMessage::kInstanceRemoved:
         cbi.mayaCallback = MDagMessage::addInstanceRemovedCallback(bindParentChildFunction, &cbi);
         break;
-    case DagMessage::kInstanceRemovedDagPath: /* cbi.mayaCallback =
-                                                 MDagMessage::addInstanceRemovedDagPathCallback(bindNodeStringBoolFunc,
-                                                 &cbi); */
+    case DagMessage::
+        kInstanceRemovedDagPath: /* cbi.mayaCallback =
```
kxl-adsk, line 946: This is pretty ugly...wasn't expecting clang-format to break it this way. Must have something to do with commented code after.

Something to remember and revisit eventually.

lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -126,89 +126,87 @@ Ufe::Attribute::Type
...
-                // SdfValueTypeName Normal3hArray, Normal3fArray, Normal3dArray;
-                // SdfValueTypeName Color3hArray,  Color3fArray,  Color3dArray;
-                // SdfValueTypeName Color4hArray,  Color4fArray,  Color4dArray;
-                // SdfValueTypeName QuathArray,    QuatfArray,    QuatdArray;
-                // SdfValueTypeName Matrix2dArray, Matrix3dArray, Matrix4dArray;
-                // SdfValueTypeName Frame4dArray;
-                // SdfValueTypeName TexCoord2hArray, TexCoord2fArray, TexCoord2dArray;
-                // SdfValueTypeName TexCoord3hArray, TexCoord3fArray, TexCoord3dArray;
-            };
+    static const std::unordered_map<size_t, Ufe::Attribute::Type> sUsdTypeToUfe {
```
seando-adsk, line 129: I really hate this formatting and this make it impossible to read. Can you please revert this map back to its original state **before the first** clang-format was applied. Then you can manually change the tabs to spaces. And add in the clang-format off/on comment so this block will remain unchanged by clang-format.

plugin/al/mayautils/AL/maya/event/MayaEventManager.cpp
seando-adsk: Why not put it back to what it was before and use the clang-format off/on comment?

seando-adsk (changes requested):

lib/mayaUsd/ufe/UsdAttributes.cpp
kxl-adsk: Removed dead code and fixed based on our discussion.

plugin/al/mayautils/AL/maya/event/MayaEventManager.cpp
kxl-adsk: I don't want developers to have to fallback to disabling clang-format in too many cases. We will update clang-format version at one point. This will be a good opportunity to revising some of the clang-format settings and prevent this type of formatting.

For that reason, I'm not going to use the clang-format off/on comment.

lib/mayaUsd/ufe/UsdAttributes.cpp
seando-adsk: Much better, thanks.

seando-adsk (approved):

PR #922, opened 2020-11-16
MAYA-107239 - Proxy Shape AE template cleanup

seando-adsk:
#### MAYA-107239 - Don't show attributes that aren't supported in the Stage AE template

* First pass at cleaning up the Proxy Shape AE template.

### DIFF
--- a/plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
+++ b/plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
@@ -1,11 +1,204 @@
 global proc AEmayaUsdProxyShapeTemplate( string $nodeName )
 {
-	editorTemplate -beginScrollLayout;
+    editorTemplate -beginScrollLayout;
 
-		// include/call base class/node attributes
-		AEsurfaceShapeTemplate $nodeName;
+        // Object Display (from AEdagNodeCommon.mel):
+        editorTemplate -beginLayout (uiRes("m_AEshapeTemplate.kObjectDisplay"));
+            editorTemplate -beginNoOptimize;
+                editorTemplate -suppress "template";
+                editorTemplate -suppress "hideOnPlayback";
+                editorTemplate -suppress "displayHWEnvironment";
+                editorTemplate -suppress "ignoreHwShader";
+                editorTemplate -suppress "intermediateObject";
+                editorTemplate -suppress "hardwareFogMultiplier";
+                editorTemplate -addControl "visibility";
+                editorTemplate -label (uiRes("m_AEdagNodeCommon.kLODVisibility")) -addControl "lodVisibility";
+                editorTemplate -callCustom "AEhiddenInOutlinerNew" "AEhiddenInOutlinerReplace" "hiddenInOutliner";
+                editorTemplate -callCustom "AEuseOutlinerColorNew" "AEuseOutlinerColorReplace" "useOutlinerColor";
+                editorTemplate -callCustom "AEoutlinerColorNew" "AEoutlinerColorReplace" "outlinerColor";
+            editorTemplate -endNoOptimize;
+        editorTemplate -endLayout;
 
-	editorTemplate -addExtraControls;
+        // Bounding Box Information (from AEdagNodeCommon.mel):
+        editorTemplate -beginLayout (uiRes("m_AEdagNodeCommon.kBoundingBoxInformation"));
+            editorTemplate -suppress "boundingBox";
+            editorTemplate -callCustom AEnewBbox AEreplaceBbox "boundingBoxMin" "boundingBoxMax";
+        editorTemplate -endLayout;
 
-	editorTemplate -endScrollLayout;
+        // UUID (from AEdependNodeTemplate.mel):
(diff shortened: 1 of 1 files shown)


plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
```diff
@@ -1,11 +1,204 @@
...
+        editorTemplate -beginLayout (uiRes("m_AEshapeTemplate.kObjectDisplay"));
+            editorTemplate -beginNoOptimize;
+                editorTemplate -suppress "template";
+                editorTemplate -suppress "hideOnPlayback";
+                editorTemplate -suppress "displayHWEnvironment";
+                editorTemplate -suppress "ignoreHwShader";
+                editorTemplate -suppress "intermediateObject";
+                editorTemplate -suppress "hardwareFogMultiplier";
+                editorTemplate -addControl "visibility";
+                editorTemplate -label (uiRes("m_AEdagNodeCommon.kLODVisibility")) -addControl "lodVisibility";
```
seando-adsk, line 15: Current Maya has a label for "Visibility" as well, but that tag doesn't exist in Maya 2020.

plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
```diff
@@ -1,11 +1,204 @@
...
+        editorTemplate -suppress "ghostColorPreA";
+        editorTemplate -suppress "ghostColorPost";
+        editorTemplate -suppress "ghostColorPostR";
+        editorTemplate -suppress "ghostColorPostG";
+        editorTemplate -suppress "ghostColorPostB";
+        editorTemplate -suppress "ghostColorPostA";
+        editorTemplate -suppress "ghostDriver";
+        editorTemplate -suppress "ghostUseDriver";
+        editorTemplate -suppress "ghostRangeStart";
+        editorTemplate -suppress "ghostRangeEnd";
```
seando-adsk, line 112: Some of these (including these two) are from Maya 2020 (either removed or renamed). If an attribute doesn't exist the suppress/addControl will silently return.

seando-adsk:
@fowlertADSK First pass at cleanup of the MayaUsd proxy shape AE template. Basically removing a bunch of sections (by suppressing all the attributes). I no longer derive from the shape template, but instead fully define one here. It works in both Maya 2020 and current PR.

plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
```diff
@@ -1,11 +1,204 @@
 global proc AEmayaUsdProxyShapeTemplate( string $nodeName )
 {
-	editorTemplate -beginScrollLayout;
+    editorTemplate -beginScrollLayout;
 
-		// include/call base class/node attributes
-		AEsurfaceShapeTemplate $nodeName;
+        // Object Display (from AEdagNodeCommon.mel):
+        editorTemplate -beginLayout (uiRes("m_AEshapeTemplate.kObjectDisplay"));
```
fowlertADSK, line 6: Is the uiRes call going to fail if AEdagNodeCommon.mel hasn't been sourced yet?  You were calling AEsurfaceShapeTemplate before (which calls a bunch of other base Templates).  Almost anything would end calling it, but if the AE is not being shown on startup and the first time it's shown we have a USD node selected...will it spit out an error?

plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
fowlertADSK: Any reason why you bother reusing a string defined in Maya at all?  Everything else is kind of hard-coded and in English only, but a couple of them will come from Maya and possibly be in another language.  It should work, and the odds of Maya changing (or moving the location of) one of those strings is pretty small, but just wondering why you're even bothering.

fowlertADSK (approved):
uiRes calls are ok, but I did leave a question about why you're bothering with them.

Also, double check that those strings are available even if this template is the very first thing the AE displays.  

Other than that, looks good.

plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
seando-adsk: I tried your steps just to be sure. But even at startup the AE is empty and if I just create an empty stage that is the first object to be displayed in the AE.

uiRes is just a MEL proc that actually sources the correct xxx.res.mel file. It doesn't need AEdagNodeCommon.mel.

plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
seando-adsk: Because the proxy shape is a Maya node, so I figured we should try and support the localization as much as possible.

seando-adsk:
@kxl-adsk This change is just a single MEL file that is only copied to the install folder. There is nothing to build, so a preflight is not needed. Please merge when you're ready.

PR #924, opened 2020-11-17
Export uiUnits as metersPerUnit Metadata in USD file

dgovil:
This PR changes the metersPerUnit to respect the users uiUnit value instead of using Maya's internal unit value.
Previously, the exporter would always export as Maya's internal unit (cm). This seems because the exporter didn't want to handle conversion, however since it's writing out values directly anyway, I don't believe any conversion is required on write out.

At least for the standard case of meshes, transforms and skeletons, it seems more correct to write out the uiUnit instead of defaulting to the internal units.

In the event that someone would prefer the previous behaviour, I have gated it behind an env variable  : `MAYAUSD_EXPORT_INTERNAL_DISTANCE_UNITS`

### DIFF
--- a/lib/mayaUsd/fileio/jobs/writeJob.cpp
+++ b/lib/mayaUsd/fileio/jobs/writeJob.cpp
@@ -70,6 +70,12 @@
 
 PXR_NAMESPACE_OPEN_SCOPE
 
+TF_DEFINE_ENV_SETTING(
+    MAYAUSD_EXPORT_INTERNAL_DISTANCE_UNITS,
+    false,
+    "Sets metersPerUnit based on the internal Maya unit (cm)"
+    "When disabled, uses the UI Unit instead.");
+
 UsdMaya_WriteJob::UsdMaya_WriteJob(const UsdMayaJobExportArgs& iArgs)
     : mJobCtx(iArgs)
     , _modelKindProcessor(new UsdMaya_ModelKindProcessor(iArgs))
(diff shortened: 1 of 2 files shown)


koen-v:
Cool, just glanced over the code. Hopefully I missed the important part, but it looks like it just sets the flag for units per meter in the usd, does it also change the actual vertex coordinates? If you change the meta data in the file, but not the mesh, things get even more confusing.

Thanks,
Koen

dgovil:
@koen-v I do not change the vertex coordinates since ,afaik from testing, it was already writing out values based on the uiUnits (where 1 unit is 1 foot if your scene is set to feet for example) but it was always forcing the resulting usd file to have centimeters as the metadata.

koen-v:
Ok, thanks for testing and good to know people are looking at this. Let me know if you could use a hand, we have some resources to help if needed.

Cheers,
Koen

dgovil:
@koen-v Thanks. i'm not sure when we'll get time to resolve it on our end. Hopefully the autodesk folks or your folks are able to solve it, but agree, it would be great to try and solve it.

mattyjams:
Thanks @koen-v, that's my concern with this change as well, and why I haven't reviewed it yet. I wanted to do some testing to remind myself why we always export using Maya's internal units.

I did the following test using both 'cm' and 'foot' as the uiUnit:
- File -> New
- Set the uiUnit (Window -> Settings/Preferences -> Preferences, Settings in the left-hand column, then Working Units -> Linear, Save)
- Create -> Polygon Primitives -> Cube
- Export to USD

When the working unit was 'cm', the exported points were:
```
point3f[] points = [(-0.5, -0.5, 0.5), (0.5, -0.5, 0.5), (-0.5, 0.5, 0.5), (0.5, 0.5, 0.5), (-0.5, 0.5, -0.5), (0.5, 0.5, -0.5), (-0.5, -0.5, -0.5), (0.5, -0.5, -0.5)]
```

So Maya create a cube that was 1 unit on each side.

When the working unit was 'foot', the exported points were:
```
point3f[] points = [(-15.24, -15.24, 15.24), (15.24, -15.24, 15.24), (-15.24, 15.24, 15.24), (15.24, 15.24, 15.24), (-15.24, 15.24, -15.24), (15.24, 15.24, -15.24), (-15.24, -15.24, -15.24), (15.24, -15.24, -15.24)]
```

So again Maya create a cube that was 1 unit on each side (1 foot = 30.48cm).

All of the distance units we're collecting for export are in 'cm' as the internal unit, so without a major project to inject unit conversion all over the place, enforcing that the `metersPerUnit=0.01` in USD seems the only correct thing to do?

dgovil:
@mattyjams Yep, my bad. You're right that keeping it as `metersPerUnit=0.01` is correct, though it does put restrictions on the rest of the pipeline outside Maya to do conversion on read/load.

I think the only right thing to do would be to modify all queries to convert on write. Though that could negatively impact the write performance significantly.

mattyjams:
@dgovil: Cool, yeah, I definitely agree it would be great to have support for handling unit conversions in both export and import. We (Pixar, at least) just haven't had a strong enough need for that functionality for the work to bubble up in priority.

PR #925, opened 2020-11-18
Fowlert/maya 106722/adding tests

fowlertADSK:
Adding a test for attempting to group Usd objects from two separate proxies (was a crash).
Adding a test for duplicate -smartTransform.
Both tests need to be guarded until the next Maya PR.

Design also requested an update to a button label so I added that too.


plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
```diff
@@ -144,7 +144,7 @@ global proc mayaUsd_createStageFromFileOptions()
         $parent;
 
     string $applyCloseBtn = getOptionBoxApplyAndCloseBtn();
-    button -edit -label "Create Stage"
+    button -edit -label "Create Stage from File"
```
fowlertADSK, line 147: Small update Design requested.

test/lib/ufe/testDuplicateCmd.py
```diff
@@ -18,11 +18,14 @@
 
 import maya.cmds as cmds
 
-import usdUtils, mayaUtils
+import usdUtils
+import mayaUtils
```
fowlertADSK, line 22: A bunch of these changes are because of the clang format now running.  I think it was the plan to just update files as we go?  I'm running the clang file that is checked into the repo so I'm assuming that we do want these edits.

test/lib/ufe/testDuplicateCmd.py
```diff
@@ -183,5 +186,24 @@ def childrenNames(children):
...
+        cmds.move(10, 0, 0, r=True)
+        smartDup = cmds.duplicate(smartTransform=True)
+
+        usdTorusItem = ufeUtils.createUfeSceneItem(torusDagPath, '/pTorus3')
+        torusT3d = ufe.Transform3d.transform3d(usdTorusItem)
+        transVector = torusT3d.inclusiveMatrix().matrix[-1]
+
+        correctResult = [20, 0, 0, 1]
+
+        self.assertEqual(correctResult, transVector)
```
fowlertADSK, line 209: Test is to just duplicate, make a move, and duplicate with the smartTransform flag without changing selection at all.  In this case it should apply the same move to the second new object.

I added a couple simple files for the other test I'm adding and realized later that I'm using those files in this test as well, which wasn't intentional but should be fine, it's just a torus and a sphere.

test/lib/ufe/testGroupCmd.py
```diff
@@ -127,3 +134,18 @@ def testUsdGroup(self):
...
+        usdSphere = sphereDagPath + ",/pSphere1"
+
+        torusFile = mayaUtils.getTestScene("groupCmd", "torus.usda")
+        torusDagPath, torusStage = mayaUtils.createProxyFromFile(torusFile)
+        usdTorus = torusDagPath + ",/pTorus1"
+
+        try:
+            cmds.group(usdSphere, usdTorus)
+        except Exception as e:
+            self.assertTrue('cannot group across usd proxies')
```
fowlertADSK, line 151: Creating two different proxies from files (sphere in one, torus in the other), then selecting an object from each and trying to group them.  This was previously a crash, but will be fixed in the next Maya PR.

ppt-adsk (approved):
Thanks for the tests!

PR #926, opened 2020-11-18
Allow recursive compute with proxy accessor

kxl-adsk:
The proxy accessor is allowing DG dependencies between Maya's data model and USD. This means we can connect input from a DG node to proxy shape to drive USD attributes AND we can connect the proxy shape outputs to make USD attributes drive other DG nodes.

This also allows injecting DG dependencies between two USD attributes, like for example creating parent constraint between two prims. In order to compute such a setup correctly, we need to allow nested / recursive compute with the proxy accessor.

This PR is doing exactly that and an example of such DG setup is provided as a regression test.


lib/mayaUsd/base/id.h
```diff
@@ -0,0 +1,145 @@
...
+
+#ifndef MAYAUSD_ID_H
+#define MAYAUSD_ID_H
+
+#include <climits>
+#include <string>
+
+namespace MAYAUSD_NS_DEF {
+
+/*! \class Id
```
kxl-adsk, line 25: We have been starting to use "version" numbers in base proxy shape and VP2RenderDelegate. Tracking what different int's and long's mean can be tricky, so we are bringing this little class to solve it.

lib/usd/pxrUsdPreviewSurface/usdPreviewSurface.cpp
```diff
@@ -425,8 +424,8 @@ MStatus PxrMayaUsdPreviewSurface::compute(const MPlug& plug, MDataBlock& dataBlo
         CHECK_MSTATUS(status);
         float opacity = opacityData.asFloat();
 
-        MObject opacityThresholdAttr
-            = depNodeFn.attribute(PxrMayaUsdPreviewSurfaceTokens->OpacityThresholdAttrName.GetText());
+        MObject opacityThresholdAttr = depNodeFn.attribute(
```
kxl-adsk, line 427: Only formatting change from clang-format

test/lib/CMakeLists.txt
```diff
@@ -5,12 +5,12 @@ set(TEST_SCRIPT_FILES
     testMayaUsdConverter.py
     testMayaUsdPythonImport.py
     testMayaUsdLayerEditorCommands.py
+    testMayaUsdCacheId.py
 )
 
-if (MAYA_APP_VERSION VERSION_GREATER 2020)
+if (UFE_FOUND AND MAYA_APP_VERSION VERSION_GREATER 2020)
```
kxl-adsk, line 11: Proxy accessor test requires UFE.

test/lib/testMayaUsdProxyAccessor.py
```diff
@@ -53,6 +53,7 @@ def setUpClass(cls):
         """
         if not cls.pluginsLoaded:
             cls.pluginsLoaded = mayaUtils.isMayaUsdPluginLoaded()
+            cmds.loadPlugin('matrixNodes')
```
kxl-adsk, line 56: For inverseMatrix node

lib/mayaUsd/nodes/proxyAccessor.cpp
```diff
@@ -50,6 +51,94 @@
 
 namespace MAYAUSD_NS_DEF {
 
+/*! /brief  Scoped object setting up compute context for accessor
+
+    Proxy accessor supports nested compute that allows injecting DG dependencies to USD. More
```
williamkrick, line 56: Is there an extra space before "Proxy"?

williamkrick: Maybe the other lines have only 3 spaces.

lib/mayaUsd/nodes/proxyAccessor.cpp
```diff
@@ -50,6 +51,94 @@
...
+        _proxyInclusiveMatrix = proxyDagPath.inclusiveMatrix();
+
+        // Only increment evaluation ID for top most evaluation scope.
+        // Nested scope is allowed, but shouldn't in really be needed.
+        if (!_restoreState) {
+            accessor._evaluationId.next();
+        }
+    }
+
+    //! \brief  Construct compute context for both inputs
```
williamkrick, line 92: This comment cut off?

williamkrick (approved):

lib/mayaUsd/nodes/proxyAccessor.cpp
kxl-adsk: Good catch!

lib/mayaUsd/nodes/proxyAccessor.cpp
kxl-adsk: Fixed.

williamkrick (approved):

PR #927, opened 2020-11-18
USD method GetSchemaTypeName() was added in v20.05

seando-adsk:
USD method UsdSchemaRegistry::GetSchemaTypeName() was only added in v20.05.


kxl-adsk (approved):

seando-adsk:
@kxl-adsk Can you test this locally with your USD v20.02 and then merge if okay.

PR #928, opened 2020-11-18
MAYA-107484 - Time is not connected when creating a new empty USD stage

seando-adsk:
#### MAYA-107484 - Time is not connected when creating a new empty USD stage
* Connected time to proxy shape node

### DIFF
--- a/plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
+++ b/plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
@@ -192,5 +192,9 @@ global proc mayaUsd_createStageFromFile() {
     if (size($result) > 0) {
         doCreateStage($result[0]);
     }
-        
+}
+
+global proc mayaUsd_createStageFromFilePath(string $usdFile)
+{
+    doCreateStage($usdFile);
 }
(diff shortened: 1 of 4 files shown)


kxl-adsk (approved):

HamedSabri-adsk (approved):
Looking good!

plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
```diff
@@ -192,5 +192,9 @@ global proc mayaUsd_createStageFromFile() {
     if (size($result) > 0) {
         doCreateStage($result[0]);
     }
-        
+}
+
+global proc mayaUsd_createStageFromFilePath(string $usdFile)
+{
+    doCreateStage($usdFile);
```
seando-adsk, line 199: Simple global proc that allows me to do the same as proc above, but sending in a file path, skipping the file dialog.

seando-adsk (commented):
I added a new test for the create stages commands. I test that the proxy node was created, selected and that the time is connected.

kxl-adsk (approved):
Thank you, @seando-adsk, for adding the test.

HamedSabri-adsk (approved):

PR #930, opened 2020-11-19
Fixed current compiler warnings

dgovil:
There were several small warnings that were tripping up our compilation when trying to compile with -Wall

They're all errors like this:

```
/Users/dhruvgovil/Projects/maya-usd/lib/mayaUsd/commands/baseListShadingModesCommand.cpp:117:25: warning: loop variable 'c' of type 'const pxrInternal_v0_20__pxrReserved__::TfToken' creates a copy from type 'const pxrInternal_v0_20__pxrReserved__::TfToken' [-Wrange-loop-analysis]
        for (const auto c : UsdMayaShadingModeRegistry::ListMaterialConversions()) {
```

### DIFF
--- a/lib/mayaUsd/fileio/instancedNodeWriter.cpp
+++ b/lib/mayaUsd/fileio/instancedNodeWriter.cpp
@@ -168,7 +168,7 @@ UsdMaya_InstancedNodeWriter::UsdMaya_InstancedNodeWriter(
             // Replace prefixes to obtain DAG-USD path mapping.
             const UsdMayaUtil::MDagPathMap<SdfPath>& writerMapping
                 = writer->GetDagToUsdPathMapping();
-            for (const std::pair<MDagPath, SdfPath>& pair : writerMapping) {
+            for (const auto& pair : writerMapping) {
                 const MDagPath& dagPathInMaster = pair.first;
                 const SdfPath&  usdPathInMaster = pair.second;
 
--- a/lib/mayaUsd/fileio/registryHelper.cpp
+++ b/lib/mayaUsd/fileio/registryHelper.cpp
@@ -238,7 +238,7 @@ VtDictionary UsdMaya_RegistryHelper::GetComposedInfoDictionary(const std::vector
         if (_ReadNestedDict(plugin->GetMetadata(), scope, &curJsDict)) {
             const VtValue curValue = JsConvertToContainerType<VtValue, VtDictionary>(curJsDict);
             if (curValue.IsHolding<VtDictionary>()) {
-                for (const std::pair<std::string, VtValue>& pair :
+                for (const std::pair<std::string, VtValue> pair :
                      curValue.UncheckedGet<VtDictionary>()) {
                     result[pair.first] = pair.second;
                     keyDefinitionSites[pair.first].push_back(plugin->GetName());
@@ -254,7 +254,7 @@ VtDictionary UsdMaya_RegistryHelper::GetComposedInfoDictionary(const std::vector
     }
 
     // Validate that keys are only defined once globally.
-    for (const std::pair<std::string, std::vector<std::string>>& pair : keyDefinitionSites) {
+    for (const auto& pair : keyDefinitionSites) {
         if (pair.second.size() != 1) {
             TF_RUNTIME_ERROR(
                 "Key '%s' is defined in multiple plugins (%s). "
--- a/lib/mayaUsd/utils/util.cpp
+++ b/lib/mayaUsd/utils/util.cpp
@@ -1682,7 +1682,7 @@ VtDictionary UsdMayaUtil::GetDictionaryFromArgDatabase(
     //     Python command API. If single arg per flag, make it a vector of
     //     strings. Multi arg per flag, vector of vector of strings.
     VtDictionary args;
-    for (const std::pair<std::string, VtValue>& entry : guideDict) {
+    for (const auto& entry : guideDict) {
         const std::string& key = entry.first;
         const VtValue&     guideValue = entry.second;
         if (!argData.isFlagSet(key.c_str())) {
--- a/cmake/compiler_config.cmake
+++ b/cmake/compiler_config.cmake
@@ -13,6 +13,12 @@ set(GNU_CLANG_FLAGS
     -Wno-unused-local-typedefs
 )
 
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    list(APPEND GNU_CLANG_FLAGS
+            -Wrange-loop-analysis
+        )
+endif()
+
 set(MSVC_FLAGS
     # we want to be as strict as possible
     /W3
--- a/lib/mayaUsd/commands/baseListShadingModesCommand.cpp
+++ b/lib/mayaUsd/commands/baseListShadingModesCommand.cpp
@@ -114,7 +114,7 @@ MStatus MayaUSDListShadingModesCommand::doIt(const MArgList& args)
         // This is to be used when importing via the dialog. Finer grained import is available at
         // the command level.
         //
-        for (const auto c : UsdMayaShadingModeRegistry::ListMaterialConversions()) {
+        for (const auto &c : UsdMayaShadingModeRegistry::ListMaterialConversions()) {
             if (c != UsdImagingTokens->UsdPreviewSurface) {
                 auto const& info = UsdMayaShadingModeRegistry::GetMaterialConversionInfo(c);
                 if (info.hasImporter) {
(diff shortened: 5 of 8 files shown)


lib/mayaUsd/fileio/registryHelper.cpp
```diff
@@ -254,7 +254,7 @@ VtDictionary UsdMaya_RegistryHelper::GetComposedInfoDictionary(const std::vector
     }
 
     // Validate that keys are only defined once globally.
-    for (const std::pair<std::string, std::vector<std::string>>& pair : keyDefinitionSites) {
+    for (const std::pair<std::string, std::vector<std::string>> pair : keyDefinitionSites) {
```
HamedSabri-adsk, line 257: please use auto. 
`for (const auto pair : keyDefinitionSites)`

Once we switch to C++17, We can leverage the  Structured bindings feature.

lib/mayaUsd/utils/util.cpp
```diff
@@ -1682,7 +1682,7 @@ VtDictionary UsdMayaUtil::GetDictionaryFromArgDatabase(
     //     Python command API. If single arg per flag, make it a vector of
     //     strings. Multi arg per flag, vector of vector of strings.
     VtDictionary args;
-    for (const std::pair<std::string, VtValue>& entry : guideDict) {
+    for (const std::pair<std::string, VtValue> entry : guideDict) {
```
HamedSabri-adsk, line 1685: same as above comment

lib/mayaUsd/fileio/instancedNodeWriter.cpp
```diff
@@ -168,7 +168,7 @@ UsdMaya_InstancedNodeWriter::UsdMaya_InstancedNodeWriter(
             // Replace prefixes to obtain DAG-USD path mapping.
             const UsdMayaUtil::MDagPathMap<SdfPath>& writerMapping
                 = writer->GetDagToUsdPathMapping();
-            for (const std::pair<MDagPath, SdfPath>& pair : writerMapping) {
+            for (const std::pair<MDagPath, SdfPath> pair : writerMapping) {
```
HamedSabri-adsk, line 171: same comment as others.

HamedSabri-adsk (changes requested):
Looking good! Thanks. 

Nit-pick: see my comment about auto.

lib/mayaUsd/fileio/instancedNodeWriter.cpp
```diff
@@ -168,7 +168,7 @@ UsdMaya_InstancedNodeWriter::UsdMaya_InstancedNodeWriter(
             // Replace prefixes to obtain DAG-USD path mapping.
             const UsdMayaUtil::MDagPathMap<SdfPath>& writerMapping
                 = writer->GetDagToUsdPathMapping();
-            for (const std::pair<MDagPath, SdfPath>& pair : writerMapping) {
```
kxl-adsk, line 171: I understand the above cases, where we want to avoid copies (and it's such a common mistake)...but I'm curious what exactly is the compiler warning in this case?

dgovil:
@HamedSabri-adsk Done! I am all for more compiler auto inference.

kxl-adsk (commented):
I left one question in the review to get a better sense of the warning you are fixing by requesting an actual copy.

There is as well something bigger here that we may be missing. Unless we make sure PF can detect warnings you are fixing on all platforms, people will still be able to break your builds.

lib/mayaUsd/fileio/instancedNodeWriter.cpp
dgovil: This was the compiler warning for that specific area.

```
/Users/dhruvgovil/Projects/maya-usd/lib/mayaUsd/fileio/instancedNodeWriter.cpp:171:54: warning: loop variable 'pair' has type 'const std::pair<MDagPath, SdfPath> &' but is initialized with type 'const std::__1::__map_const_iterator<std::__1::__tree_const_iterator<std::__1::__value_type<Autodesk::Maya::OpenMaya20200000::MDagPath, pxrInternal_v0_20__pxrReserved__::SdfPath>, std::__1::__tree_node<std::__1::__value_type<Autodesk::Maya::OpenMaya20200000::MDagPath, pxrInternal_v0_20__pxrReserved__::SdfPath>, void *> *, long> >::value_type' (aka 'const pair<const Autodesk::Maya::OpenMaya20200000::MDagPath, pxrInternal_v0_20__pxrReserved__::SdfPath>') resulting in a copy [-Wrange-loop-analysis]
            for (const std::pair<MDagPath, SdfPath>& pair : writerMapping) {
                                                     ^
/Users/dhruvgovil/Projects/maya-usd/lib/mayaUsd/fileio/instancedNodeWriter.cpp:171:18: note: use non-reference type 'std::pair<MDagPath, SdfPath>' to keep the copy or type 'const std::__1::__map_const_iterator<std::__1::__tree_const_iterator<std::__1::__value_type<Autodesk::Maya::OpenMaya20200000::MDagPath, pxrInternal_v0_20__pxrReserved__::SdfPath>, std::__1::__tree_node<std::__1::__value_type<Autodesk::Maya::OpenMaya20200000::MDagPath, pxrInternal_v0_20__pxrReserved__::SdfPath>, void *> *, long> >::value_type &' (aka 'const pair<const Autodesk::Maya::OpenMaya20200000::MDagPath, pxrInternal_v0_20__pxrReserved__::SdfPath> &') to prevent copying
            for (const std::pair<MDagPath, SdfPath>& pair : writerMapping) {
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

dgovil:
For reference, this is the entirety of the compiler logs. We're ignoring warnings from the devkit itself, and just eliminating the ones from this repo.

[Uploading compiler_warnings.txt…]()

HamedSabri-adsk (approved):
Thank you @dgovil

kxl-adsk (changes requested):
@dgovil Thank you for the log. This makes sense. 

Going back to my previous point, I suggest we enable `-Wrange-loop-analysis` in [compiler configuration](https://github.com/Autodesk/maya-usd/blob/dev/cmake/compiler_config.cmake). This will guarantee you won't have to fix these type of errors again. Would be great as well to check if there is a matching MSVC flag, so Windows developers won't get surprises in PF.

lib/mayaUsd/fileio/instancedNodeWriter.cpp
dgovil: Enabling `-Wrange-loop-analysis` suggested adding back the `&` so I put it back. I suspect it was a false positive of some kind without the explicit flag. Not exactly sure why it suggested removing it in the first place, because reading the actual code...it makes sense to me to leave the `&` in.

dgovil:
Okay, added `-Wrange-loop-analysis` explicitly, which suggested a couple changes. 
Specifically restoring the `&` to `lib/mayaUsd/fileio/registryHelper.cpp:257:21` and `lib/mayaUsd/fileio/instancedNodeWriter.cpp:171:29`. This makes more sense to me reading the code, so I suspect it was a false positive without the explicit flag..

Regarding msvc, I do not have a windows machine available to me at the moment, and can't find a comparable flag on their flag documentation. Hopefully someone more familiar with msvc can add those flags in later?

kxl-adsk:
Preflight came back and we will have to pay the price for having different compilers per platform. There is no `-Wrange-loop-analysis` for GCC, so you will have to put it in its own category. Apologies for not catching that earlier.

As for Windows - no worries and thx for checking.

HamedSabri-adsk:
`Okay, added -Wrange-loop-analysis explicitly, which suggested a couple changes. Also updated my build setup to build against the 2020.3 devkit, which might have contributed.
`
There is no VS equivalent. This is a clang only flag and is off by default.

https://clang.llvm.org/docs/DiagnosticsReference.html#id640

By enabling this flag we are also enabling -Wrange-loop-bind-reference and -Wrange-loop-construct.

These flags can result in false positives sometimes. 

https://reviews.llvm.org/D73007

https://quuxplusone.github.io/blog/2020/08/26/wrange-loop-analysis/

dgovil:
I'll modify it so that it's only enabled on Mac platforms. I figured GCC would have just ignored it if it didn't exist since it has to deal with other version specific warning flags, but c'est la vie.

dgovil:
Okay, I modified it so there's a specific section for CLANG args. It matches against all Clang variants for now (Clang/AppleClang).

```
if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
    list(APPEND GNU_CLANG_FLAGS
            -Wrange-loop-analysis
        )
endif()
```

kxl-adsk (approved):

PR #934, opened 2020-11-20
MAYA-106722 Maya check callback to see if there are any unsaved edits

fowlertADSK:
This adds a file new and open check callback so we can warn the user that they have USD edits that are not saved.  The "check" callbacks from Maya pass a bool variable that you can set to false in order to block the file IO operation.  Meaning, before Maya actually opens another file or creates a new scene we can cancel it and stay in the current scene.

We are still iterating on the entire workflow of warning the user and then helping them to quick save the edits, so at the moment this is only in the ADSK plugin.  Once we finalize the workflow we can look at what might be useful down in the shared lib.

### DIFF
--- a/plugin/adsk/plugin/sceneResetCheck.cpp
+++ b/plugin/adsk/plugin/sceneResetCheck.cpp
@@ -64,11 +64,11 @@ static void _OnMayaNewOrOpenSceneCheckCallback(bool* retCode, void*)
                 continue;
             }
 
-            std::string          temp;
             SdfLayerHandleVector allLayers = stage->GetLayerStack(true);
             for (auto layer : allLayers) {
                 if (layer->IsDirty()) {
                     atLeastOneDirty = true;
+                    break;
                 }
             }
         }
--- a/plugin/adsk/plugin/sceneResetCheck.cpp
+++ b/plugin/adsk/plugin/sceneResetCheck.cpp
@@ -1,0 +1,109 @@
+//
+// Copyright 2020 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "sceneResetCheck.h"
+
+#include "ProxyShape.h"
+
+#include <maya/MFnDependencyNode.h>
+#include <maya/MGlobal.h>
+#include <maya/MItDependencyNodes.h>
+#include <maya/MMessage.h>
+#include <maya/MSceneMessage.h>
+
+namespace {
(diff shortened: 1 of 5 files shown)


kxl-adsk (changes requested):
Even if we enable this for the Autodesk plugin only, I don't see why we wouldn't put sceneResetCheck in the mayaUsd library. Please move it.

plugin/adsk/plugin/sceneResetCheck.cpp
```diff
@@ -0,0 +1,115 @@
...
+        MObject mobj = iter.item();
+        fn.setObject(mobj);
+        if (MayaUsd::ProxyShape::typeId == fn.typeId()) {
+            MayaUsdProxyShapeBase* pShape = static_cast<MayaUsdProxyShapeBase*>(fn.userNode());
+            UsdStageRefPtr         stage = pShape ? pShape->getUsdStage() : nullptr;
+            if (!stage) {
+                continue;
+            }
+
+            std::string          temp;
```
seando-adsk, line 67: This variable is unused.

plugin/adsk/plugin/sceneResetCheck.cpp
```diff
@@ -0,0 +1,115 @@
...
+            SdfLayerHandleVector allLayers = stage->GetLayerStack(true);
+            for (auto layer : allLayers) {
+                if (layer->IsDirty()) {
+                    atLeastOneDirty = true;
```
seando-adsk, line 71: Why not break here?

seando-adsk (changes requested):

plugin/adsk/plugin/sceneResetCheck.cpp
fowlertADSK: The for loop will exit immediately, it checks atLeastOneDirty.  I think I had originally been printing out some information while debugging something and just left it like this (but completely missed the temp variable you pointed out).  I could remove the check in the for loop and add an extra line to "break" here if you prefer that style.

seando-adsk: Actually it won't because you are in a for-loop (allLayers) inside of that other for-loop (that checks atLeastOneDirty).

fowlertADSK:
> Even if we enable this for the Autodesk plugin only, I don't see why we wouldn't put sceneResetCheck in the mayaUsd library. Please move it.

Were you thinking about adding some kind of registration where plugins would provide their proxy shape node id or something to tell the lib to look at it for dirty layers?

One of the reasons I didn’t bother adding it to the common lib yet, other than just that we’re still in-progress on this, was that I believe the Animal Logic ability to save edits to the Maya scene leave the Usd layers dirty.  So even though they have saved everything and can recover it, Maya will still provide this warning dialog.  If this is moved down to the lib then I guess the easiest way to avoid this is the registration where we only look at proxy shapes that are requesting this check.  I’m assuming that you are voting to do that and enable it for the ADSK and Pixar nodes, but not the Animal Logic node?

Or were you voting to move it to the shared lib but only have it look at the ADSK nodes?

kxl-adsk:
Maybe even simpler than that. Let's move `sceneResetCheck.cpp` and `sceneResetCheck.h` to the core library. Make sure any studio plugin (not limited to AL or PXR) that derives from ProxyShapeBase can register the check (so deal with the node type). Finally, as part of this change, we only make Autodesk proxy to register this check.

kxl-adsk:
@fowlertADSK I believe this PR can be now closed?

PR #935, opened 2020-11-20
Remove redundant toMString function and pragma include guard.

marsupial:


mattyjams (approved):
Looks good to me! Thanks for the follow-up here, @marsupial!

PR #936, opened 2020-11-20
RFC - sample implementation of VP2.0 render delegate PointInstancer selection support

mattyjams:
Posting this just to help generate some discussion about how to support selecting instances from `PointInstancer` prims when using the Viewport 2.0 render delegate. I'm a little unclear around how/whether individual instances should be identified by unique `Ufe::Path`, so I wanted to show what's necessary on the imaging/selection side at least to help inform what we might want to do with the data model.

Using the `PointInstancerCubes.ma` test scene as an example, here's what the current selection behavior looks like:
before

Notice that **all** instances of the prototype are always selected, regardless of whether the top or bottom cube is picked.

And here's what the behavior looks like with this change:
after

This brings the behavior roughly in line with how `usdview` behaves when the "Pick mode" is set to "Instances".


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -136,13 +136,14 @@ void PopulateSelection(
...
     SdfPath usdPath = usdItem->prim().GetPath();
+    const int instanceIndex = usdItem->instanceIndex();
 
 #if !defined(USD_IMAGING_API_VERSION) || USD_IMAGING_API_VERSION < 11
     usdPath = sceneDelegate.ConvertCachePathToIndexPath(usdPath);
 #endif
 
     sceneDelegate.PopulateSelection(
-        HdSelection::HighlightModeSelect, usdPath, UsdImagingDelegate::ALL_INSTANCES, result);
+        HdSelection::HighlightModeSelect, usdPath, instanceIndex, result);
```
mattyjams, line 146: This is the reason the current code selects all instances of the prototype. If a particular instance was selected, we want to be sure to populate the Hydra selection with the index of that instance.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -726,10 +727,18 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
     // instance, because we don't use instanced draw for single instance render items in order to
     // improve draw performance in Maya 2020 and before.
     const int drawInstID = intersection.instanceID();
+    int instanceIndex = UsdImagingDelegate::ALL_INSTANCES;
 
 #if defined(USD_IMAGING_API_VERSION) && USD_IMAGING_API_VERSION >= 13
     const int usdInstID = drawInstID > 0 ? drawInstID - 1 : 0;
-    SdfPath   usdPath = _sceneDelegate->GetScenePrimPath(rprimId, usdInstID);
+    HdInstancerContext instancerContext;
+    SdfPath usdPath = _sceneDelegate->GetScenePrimPath(rprimId, usdInstID, &instancerContext);
```
mattyjams, line 735: For point instances, `GetScenePrimPath()` is going to give us back the path to the prototype prim that is being instanced. We want to make sure we capture the instancer context so we can find the `PointInstancer` that is doing the instancing:
https://github.com/PixarAnimationStudios/USD/blob/d8a405a1344480f859f025c4f97085143efacb53/pxr/usdImaging/usdImaging/delegate.h#L389

When multiple levels of instancing are involved, we only ever allow selecting (and eventually manipulating) the top-level instance, at least for now, since making changes at lower levels may have wider-reaching effects.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -776,13 +785,29 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
...
     const Ufe::PathSegment pathSegment(usdPath.GetText(), USD_UFE_RUNTIME_ID, USD_UFE_SEPARATOR);
-    const Ufe::SceneItem::Ptr& si
-        = handler->createItem(_proxyShapeData->ProxyShape()->ufePath() + pathSegment);
+    Ufe::Path sceneItemPath(_proxyShapeData->ProxyShape()->ufePath() + pathSegment);
+
+    if (instanceIndex != UsdImagingDelegate::ALL_INSTANCES) {
+        const std::string& primName = sceneItemPath.back().string();
+        const Ufe::PathComponent primInstanceComponent(
+            TfStringPrintf("%s.instance_%d", primName.c_str(), instanceIndex));
+        sceneItemPath = sceneItemPath.sibling(primInstanceComponent);
```
mattyjams, line 794: Here's where things start getting a little sketchy. It's seeming like we need a `Ufe::Path` that is unique to each particular instance. If we don't do this, when we add the scene items to the UFE selection below, it appears to be collapsing multiple items into one when they all share the same path, even though we're setting a different `instanceIndex` on each one. In `PopulateSelection()`, we only get a single `SceneItem` per path out of the global selection.

I'm intentionally only inventing this `PathComponent` on the UFE side to avoid polluting any tables on the USD side by creating `TfTokens` or `SdfPaths` there that we'll never use again. I would imagine though that it's a similar issue on the UFE side, so I'm a little concerned about the performance implications of having a `Ufe::Path` per instance. `PointInstancer` prims can easily reach instance counts into the hundreds of thousands or more.

This disconnect though between the `Ufe::Path` not exactly matching the `SdfPath` of the `UsdPrim` it represents could be problematic. For example, the outliner currently seems to get confused when an instance is selected. It highlights all ancestors of the `PointInstancer` using the "secondary" highlighting, but doesn't highlight the `PointInstancer` itself.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -776,13 +785,29 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
...
         return false;
     }
 
+    auto usdItem = std::dynamic_pointer_cast<MayaUsd::ufe::UsdSceneItem>(si);
+    if (!usdItem) {
+        TF_WARN("Failed to cast UFE scene item to UsdSceneItem for Rprim '%s'", rprimId.GetText());
+        return false;
+    }
+
+    usdItem->setInstanceIndex(instanceIndex);
```
mattyjams, line 809: We'll need this value both in `PopulateSelection()` here in the render delegate to make sure the right instance is put in the Hydra selection, but also later when creating the transform handler for the `PointInstancer`, since this will be the index into the various array attributes for the `positions`/`orientations`/`scales`/etc. that we'll need to modify.

For performance, I think we'll really want to keep this available somewhere as an `int` and not have to do any string manipulation to try to yank it out of the `Ufe::Path`. Selecting a large swatch of instances with a marquee selection and then manipulating them all as a group could be a pretty common use case.

lib/mayaUsd/ufe/UsdSceneItem.h
```diff
@@ -42,10 +46,21 @@ class MAYAUSD_CORE_PUBLIC UsdSceneItem : public Ufe::SceneItem
...
 
     const UsdPrim& prim() const { return fPrim; }
 
+    void setInstanceIndex(int instanceIndex) {
+        _instanceIndex = instanceIndex;
+    }
+
+    int instanceIndex() const {
+        return _instanceIndex;
+    }
```
mattyjams, line 62: I'm not convinced at all that this is the best way to handle storing and retrieving the `instanceIndex`, but this at least enables this prototype of selection highlighting.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -81,7 +81,8 @@ UsdPrim ufePathToPrim(const Ufe::Path& path)
 
     UsdPrim prim;
     if (auto stage = getStage(Ufe::Path(segments[0]))) {
-        prim = stage->GetPrimAtPath(SdfPath(segments[1].string()));
+        const SdfPath usdPath = SdfPath(segments[1].string());
+        prim = stage->GetPrimAtPath(usdPath.GetPrimPath());
```
mattyjams, line 85: So maybe I lied... `SdfPath` here is getting a string that will include the `instance_XXX` that we added to the `Ufe::Path` in the render delegate, so we may actually be polluting things on the USD side as well.

In any case, this ensures that if we did tack on the instance component of the path there, we strip it back off when getting the `UsdPrim` for that instance.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
ppt-adsk: Yes, Ufe::Selection will ignore duplicate items, i.e. items that compare equal.  Currently, item comparison is hard-coded to compare only paths.  We can explore the implications of changing this.

Yes, UFE has a problem similar to USD, in that Ufe::PathComponent is like a TfToken, a fixed, unmutable string stored in a table.  I can make a few observations here:
- Each PathComponent is a small object, and the strings for the integer indices are small, and stored once.
- If we have Ufe::PathComponents for an instancer, those strings will get re-used, won't they?
- In the UFE implementation, the table that maps strings to path components is typedef std::unordered_map<std::string, std::weak_ptr<Ufe::PathComponent::Data> > StringTable.  We may be able to garbage collect the table if a stored weak_ptr is expired.
- The Ufe::PathComponent::Data is opaque from the point of view of the PathComponent interface.  We may be able to play games there and have a union that stores an integer index.  This is starting to sound like a bit of a stretch, admittedly.
- The performance impact of having a Ufe::Path per instance would have to be measured.  Since each path component is the size of a shared_ptr, and assuming we have say 10 path component in a Ufe::Path, then we're talking the size of 10 x shared_ptr x 100000 instances, which doesn't sound so bad.  What other implications are you concerned about?
- We would have to figure out what the trouble with the Outliner is.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
ppt-adsk: What is an instance index in the USD data model?  Is it a property of a prim?

lib/mayaUsd/ufe/UsdSceneItem.h
ppt-adsk: Is an instance index something that can be associated with any prim?  Only certain kinds of prims?  Perhaps this should be a derived class?  Or maybe we can have some kind of TfToken, VtValue pair map associated with a UsdSceneItem and an instance index would be one such pair?

ppt-adsk (commented):
Since we're exploring things here, did you explore the idea of tacking on the integer index itself as a path component, and not munging strings together?  Is that a more or less palatable option?  What would be the advantages and disadvantages?  What UsdPrim would be associated with such a scheme?  I can think of one advantage: you wouldn't need to store the integer index in the UsdSceneItem, since you'd have it as the last component of the Ufe::Path --- albeit as a string.  Maybe that's O.K.?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
HdC-adsk: I can definitely see why the change is required, but I don't understand yet how selection highlight could work on exact instance before. What is the difference between point instancer and regular instancing (the one I see in treePatch.usda) when it comes to PopulateSelection()?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
HdC-adsk: I think you made a change before to allow picking the precise instance from nested instancing. Looks like this is not a desired behavior any more. Could you elaborate on the side effects?

kxl-adsk:
With #1027 now open, is there value in keeping this Draft PR open?

mattyjams:
> With #1027 now open, is there value in keeping this Draft PR open?

I left this open mainly because #1027 stops short of the changes in the VP2 render delegate and only represents the "foundational" UFE stuff. I'm planning to put the viewport changes up as a separate PR if/when we're happy with #1027. I also wasn't convinced that I had adequately responded to all the feedback here (thanks again for that!).

We can migrate this discussion to #1027 (and subsequent PRs) and I can just close this one if that sounds preferable to you?

kxl-adsk:
Thank you @mattyjams. I think you can reference feedback in subsequent PRs if you have addressed it or if more discussion is required. 
And apologies for being slow with #1027, we have an internal deadline to meet that is taking all of our attention.

mattyjams:
Sounds good! Closing this one...

PR #937, opened 2020-11-20
Fix missing ARCH_CONSTRUCTOR code with VS2019

JGamache-autodesk:
Copied verbatim from https://github.com/PixarAnimationStudios/USD/commit/acf7c132affd6489b2f96c8b94d689e7508a90e6

### DIFF
--- a/cmake/compiler_config.cmake
+++ b/cmake/compiler_config.cmake
@@ -20,7 +20,6 @@ set(MSVC_FLAGS
     # enable pdb generation.
     /Zi
     # standards compliant.
-    /Zc:inline
     /Zc:rvalueCast
     # enable multiprocessor builds.
     /MP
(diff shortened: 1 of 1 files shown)


HamedSabri-adsk (approved):
Excellent find! Thank you.

cmake/compiler_config.cmake
```diff
@@ -38,6 +37,19 @@ set(MSVC_FLAGS
...
+# versions.
+#
+# For more details, see:
+# https://developercommunity.visualstudio.com/content/problem/914943/zcinline-removes-extern-symbols-inside-anonymous-n.html
+if (MSVC_VERSION GREATER_EQUAL 1920)
+    set(MSVC_FLAGS "${MSVC_FLAGS} /Zc:inline-")
+else()
+    set(MSVC_FLAGS "${MSVC_FLAGS} /Zc:inline")
+endif()
+
```
HamedSabri-adsk, line 52: Nit-pick: You could also use regular expression to have you if/else logic in one line and keep it inside the MSVC_FLAGS list.

`$<IF:$<VERSION_GREATER_EQUAL:${MSVC_VERSION},1920>,/Zc:inline-,/Zc:inline>`

JGamache-autodesk: That is a better syntax indeed. Thanks!

HamedSabri-adsk (approved):

PR #939, opened 2020-11-23
Make the call to register_ using the new handlers struct.

williamkrick:
UFE has a minor API breaking change, this PR calls register_ in the new way.


lib/mayaUsd/ufe/Global.cpp
```diff
@@ -112,6 +112,18 @@ MStatus initialize()
     Ufe::RunTimeMgr::instance().setContextOpsHandler(g_MayaRtid, proxyShapeContextOpsHandler);
 #endif
 
+#if UFE_PREVIEW_VERSION_NUM >= 2027
```
ppt-adsk, line 115: Gonna have to be 2028...

ppt-adsk (changes requested):

ppt-adsk (approved):

PR #944, opened 2020-11-26
Proxy Shape AE Template - Part 2

seando-adsk:
#### JIRA items:
MAYA-107386 As a user, on the stage AE template I'd like to connect to time
MAYA-107350 As a user, on the stage AE template I'd like to see what my default prim is
MAYA-107338 As a user, on the stage AE template I'd like to see what my Root Layer is
MAYA-107273 As a user, on the stage AE template I'd like to toggle Purpose=Default

* Create new AETemplatesHelpers.py file which is used by the MEL templates to query USD (thru python API).
* Organized proxy shape template in sections: "Stage", "Stage Source" and "Stage Display".
* Suppressed 'Stage Cache ID' and 'Out Stage Cache ID'.
* Added new non-editable text fields (for display purposes only) for "Root Layer" and "Default Prim".
* Added tooltips to some controls.

### DIFF
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -316,8 +316,13 @@ void StagesSubject::onStageInvalidate(const MayaUsdProxyStageInvalidateNotice& n
     afterOpen();
 
 #ifdef UFE_V2_FEATURES_AVAILABLE
-    Ufe::SceneItem::Ptr sceneItem = Ufe::Hierarchy::createItem(notice.GetProxyShape().ufePath());
-    Ufe::Scene::instance().notify(Ufe::SubtreeInvalidate(sceneItem));
+    auto p = notice.GetProxyShape().ufePath();
+    if (!p.empty()) {
+        Ufe::SceneItem::Ptr sceneItem = Ufe::Hierarchy::createItem(p);
+        if (sceneItem) {
+            Ufe::Scene::instance().notify(Ufe::SubtreeInvalidate(sceneItem));
+        }
+    }
 #endif
 }
 
--- a/plugin/adsk/scripts/AETemplateHelpers.py
+++ b/plugin/adsk/scripts/AETemplateHelpers.py
@@ -2,8 +2,6 @@ import ufe
 import mayaUsd.ufe
 
 def GetDefaultPrimName(proxyShape):
-    if not proxyShape.startswith('|world'):
-        proxyShape = '|world' + proxyShape
     try:
         proxyStage = mayaUsd.ufe.getStage(proxyShape)
         if proxyStage:
@@ -15,8 +13,6 @@ def GetDefaultPrimName(proxyShape):
     return ''
 
 def GetRootLayerName(proxyShape):
-    if not proxyShape.startswith('|world'):
-        proxyShape = '|world' + proxyShape
     try:
         proxyStage = mayaUsd.ufe.getStage(proxyShape)
         if proxyStage:
--- a/plugin/adsk/scripts/AETemplateHelpers.py
+++ b/plugin/adsk/scripts/AETemplateHelpers.py
@@ -1,0 +1,28 @@
+import ufe
+import mayaUsd.ufe
+
+def GetDefaultPrimName(proxyShape):
+    if not proxyShape.startswith('|world'):
+        proxyShape = '|world' + proxyShape
+    try:
+        proxyStage = mayaUsd.ufe.getStage(proxyShape)
+        if proxyStage:
+            defPrim = proxyStage.GetDefaultPrim()
+            if defPrim:
+                return defPrim.GetName()
+    except:
+        pass
+    return ''
+
+def GetRootLayerName(proxyShape):
+    if not proxyShape.startswith('|world'):
+        proxyShape = '|world' + proxyShape
+    try:
+        proxyStage = mayaUsd.ufe.getStage(proxyShape)
+        if proxyStage:
+            rootLayer = proxyStage.GetRootLayer()
+            if rootLayer:
+                return rootLayer.GetDisplayName() if rootLayer.anonymous else rootLayer.realPath
+    except:
+        pass
+    return ''
--- a/plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
+++ b/plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
@@ -1,6 +1,65 @@
+global proc AEmayaUsdProxyShapeInfoNew(string $input)
+{
+    textFieldGrp -label "Root Layer" -editable 0
+                 -ann "Identifies the root layer of a stage. If a file path is shown in this field, the root layer is a file on disk. If a layerName is shown in this field, the root layer is an anonymous layer."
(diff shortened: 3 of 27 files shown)


plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -0,0 +1,28 @@
...
+        proxyShape = '|world' + proxyShape
+    try:
+        proxyStage = mayaUsd.ufe.getStage(proxyShape)
+        if proxyStage:
+            rootLayer = proxyStage.GetRootLayer()
+            if rootLayer:
+                return rootLayer.GetDisplayName() if rootLayer.anonymous else rootLayer.realPath
+    except:
+        pass
+    return ''
```
seando-adsk, line 28: I created this python file to contains AE template helpers that I call from the actual MEL template.

plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
```diff
@@ -1,7 +1,66 @@
...
+        editorTemplate -suppress "complexity";
+        editorTemplate -suppress "inStageData";
+        editorTemplate -suppress "stageCacheId";
+        editorTemplate -suppress "outStageCacheId";
+
+        editorTemplate -beginLayout "Stage" -collapse 0;
+            // We aren't displaying the "inStageData" attribute in the ProxyShapeInfo. We simply
+            // pass it in order to get the current nodename. Cannot use the input one here as it
+            // changes based on which proxy shape the AE is viewing.
+            editorTemplate -callCustom "AEmayaUsdProxyShapeInfoNew" "AEmayaUsdProxyShapeInfoReplace" inStageData;
```
seando-adsk, line 45: This adds the "Root Layer" and "Default Prim" (non-editable) fields. These aren't actual attributes on the proxy shape. The fields are for display purposes only.

plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
```diff
@@ -1,7 +1,66 @@
...
+    string $nodeName = plugNode($input);
+    string $res[] = `ls -l $nodeName`;
+    string $fullNodeName = $res[0];
+
+    // Call python helpers to get the info we want to display.
+    string $rootLayer   = `python("import AETemplateHelpers; AETemplateHelpers.GetRootLayerName('" + $fullNodeName + "')")`;
+    string $defaultPrim = `python("import AETemplateHelpers; AETemplateHelpers.GetDefaultPrimName('" + $fullNodeName + "')")`;
+
+    textFieldGrp -edit -text $rootLayer mayaUsdProxyShapeRootLayer;
+    textFieldGrp -edit -text $defaultPrim mayaUsdProxyShapeDefaultPrim;
```
seando-adsk, line 26: This will update the two fields with the values from the current proxy shape being shown.

plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
```diff
@@ -1,7 +1,66 @@
+global proc AEmayaUsdProxyShapeInfoNew(string $input)
+{
+    textFieldGrp -label "Root Layer" -editable 0
+                 -ann "Identifies the root layer of a stage. If a file path is shown in this field, the root layer is a file on disk. If a layerName is shown in this field, the root layer is an anonymous layer."
+                 mayaUsdProxyShapeRootLayer;
+
+    textFieldGrp -label "Default Prim" -editable 0
+                 -ann "As part of its metadata, each stage can identify a default prim. This is the primitive that is referenced in if you reference in a file."
+                 mayaUsdProxyShapeDefaultPrim;
```
seando-adsk, line 9: Create two text fields for displaying info.

seando-adsk (commented):
Proxy shape AE template organization.

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -0,0 +1,28 @@
+import ufe
+import mayaUsd.ufe
+
+def GetDefaultPrimName(proxyShape):
+    if not proxyShape.startswith('|world'):
+        proxyShape = '|world' + proxyShape
```
ppt-adsk, line 6: We should not do this.  We should fix the implementation of getStage() in wrapUtils.cpp so that it converts the argument path string to a path using Ufe::PathString::path().

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -0,0 +1,28 @@
...
+            defPrim = proxyStage.GetDefaultPrim()
+            if defPrim:
+                return defPrim.GetName()
+    except:
+        pass
+    return ''
+
+def GetRootLayerName(proxyShape):
+    if not proxyShape.startswith('|world'):
+        proxyShape = '|world' + proxyShape
```
ppt-adsk, line 19: Same as above: let's fix the wrapper instead of working around it.

ppt-adsk (changes requested):

plugin/adsk/scripts/AETemplateHelpers.py
seando-adsk: As we discussed I cleaned up all the usage of "|world"

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
```diff
@@ -1905,23 +1905,6 @@ MSelectionMask ProxyShape::getShapeSelectionMask() const
...
-    MDagPath thisPath;
-    MDagPath::getAPathTo(thisMObject(), thisPath);
-
-    // MDagPath does not include |world to its full path naem
-    MString fullpath = "|world" + thisPath.fullPathName();
-
-    return Ufe::PathSegment(fullpath.asChar(), MAYA_UFE_RUNTIME_ID, MAYA_UFE_SEPARATOR);
-}
-
-Ufe::Path ProxyShape::ufePath() const { return Ufe::Path(ProxyShape::ufePathSegment()); }
```
seando-adsk, line 1922: This exact method already exists in the base class (which I fixed for the |world). The ufePathSegment() method above is not needed as you can get that from this ufePath() method.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -316,8 +316,13 @@ void StagesSubject::onStageInvalidate(const MayaUsdProxyStageInvalidateNotice& n
     afterOpen();
 
 #ifdef UFE_V2_FEATURES_AVAILABLE
-    Ufe::SceneItem::Ptr sceneItem = Ufe::Hierarchy::createItem(notice.GetProxyShape().ufePath());
-    Ufe::Scene::instance().notify(Ufe::SubtreeInvalidate(sceneItem));
+    auto p = notice.GetProxyShape().ufePath();
```
seando-adsk, line 319: It is possible for this path to return empty when called before the proxy shape is ready. The MDagPath in the proxy shape is empty. The previous code (before my world cleanup) would thus return a Ufe::path to "|world" which was incorrect.

ppt-adsk (approved):
Thanks for the cleanup!

fowlertADSK (approved):
LGTM

PR #948, opened 2020-11-30
Integrate Ufe::Transform3d::editTransform3d() hint support.

ppt-adsk:
Integration of Ufe::Transform3d::editTransform3d() hint.  Quick manual testing seems to work.  No automated test yet, but will work on this in parallel.


lib/mayaUsd/ufe/Global.cpp
```diff
@@ -112,6 +112,17 @@ MStatus initialize()
     Ufe::RunTimeMgr::instance().setContextOpsHandler(g_MayaRtid, proxyShapeContextOpsHandler);
 #endif
 
+#if UFE_PREVIEW_VERSION_NUM >= 2028
```
ppt-adsk, line 115: Update to @williamkrick changes in UFE.

lib/mayaUsd/ufe/UsdTransform3dMatrixOp.cpp
```diff
@@ -433,8 +436,19 @@ UsdTransform3dMatrixOpHandler::editTransform3d(const Ufe::SceneItem::Ptr& item)
     });
     bool foundMatrix = (i != xformOps.end());
 
-    return foundMatrix ? UsdTransform3dMatrixOp::create(usdItem, *i)
-                       : _nextHandler->editTransform3d(item);
+    // If we've found a matrix op, but there is a more local non-matrix op in
```
ppt-adsk, line 439: This is a crucial change: if there is a more local non-matrix op, let the next handler in the chain handle the edit.  This is because when a Maya fallback stack is appended to the transform ops, it must be the stack on which operations are performed.  For example, if a fallback pivot transform op is added, any further rotation on the prim must be a fallback rotation, on which the fallback pivot will apply.

lib/mayaUsd/ufe/UsdTransform3dMatrixOp.cpp
```diff
@@ -433,8 +436,19 @@ UsdTransform3dMatrixOpHandler::editTransform3d(const Ufe::SceneItem::Ptr& item)
...
+    // the stack, the more local op should be used to handle the edit.
+    bool moreLocalNonMatrix = foundMatrix ?
+        (std::find_if(i, xformOps.end(), [](const UsdGeomXformOp& op)
+        { return op.GetOpType() != UsdGeomXformOp::TypeTransform; }) 
+         != xformOps.end()) : false;
+
+    // We can't handle pivot edits, so in that case pass on to the next handler.
+    return
+        (foundMatrix && !moreLocalNonMatrix && 
+         (hint.type() != Ufe::EditTransform3dHint::EditPivot)) ?
```
ppt-adsk, line 449: Use the EditTransform3dHint to pass on the edit to the next Transform3dHandler, if a pivot edit.

lib/mayaUsd/ufe/Global.cpp
kxl-adsk: This is already in the dev branch. We should have brought it with a merge to avoid conflicts.

kxl-adsk (approved):

PR #950, opened 2020-11-30
Reuse shader effect for duplicate material networks.

HdC-adsk:
Implemented a shader instance cache to allow reuse of shader effect for duplicate material networks.
* This avoids repetitive compilation of the same shader effect, and thus reduces the load time.
* This removes a blocker for material consolidation, and thus will reduce the playback time. Maya has some other blockers for material consolidation that should be fixed separately.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -503,7 +503,7 @@ _LoadTexture(const std::string& path, bool& isColorSpaceSRGB, MFloatArray& uvSca
 #if USD_VERSION_NUM >= 2102
     HioImageSharedPtr image = HioImage::OpenForReading(path);
 #else
-    GlfImageSharedPtr image = GlfImage::OpenForReading(path);
+    GlfImageSharedPtr     image = GlfImage::OpenForReading(path);
 #endif
     if (!TF_VERIFY(image)) {
         return nullptr;
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -113,35 +113,82 @@ inline bool _IsUsdUVTexture(const HdMaterialNode& node)
     return (node.identifier == UsdImagingTokens->UsdUVTexture);
 }
 
-//! Helper utility function to print nodes, connections and primvars in the
+//! Helper function to generate a XML string about nodes, relationships and primvars in the
 //! specified material network.
-void _PrintMaterialNetwork(
-    const std::string&       label,
-    const SdfPath&           id,
-    const HdMaterialNetwork& mat)
+std::string _GenerateXMLString(const HdMaterialNetwork& mat, bool includeParams=true)
 {
-    std::cout << label << " material network for " << id << "\n";
+    std::string result;
+    result.reserve(1024);
 
-    std::cout << "  --Node--\n";
-    for (HdMaterialNode const& node : mat.nodes) {
-        std::cout << "    " << node.path << "\n";
-        std::cout << "    " << node.identifier << "\n";
+    if (!mat.nodes.empty()) {
+        result += "<nodes>\n";
 
-        for (auto const& entry : node.parameters) {
-            std::cout << "      param " << entry.first << ": " << TfStringify(entry.second) << "\n";
+        if (includeParams) {
+            for (HdMaterialNode const& node : mat.nodes) {
+                result += "  <node name=\"";
+                result += node.path.GetName();
+                result += "\" identifier=\"";
+                result += node.identifier;
+                result += "\">\n";
+
+                result += "    <params>\n";
+
+                for (auto const& parameter : node.parameters) {
+                    result += "      <param name=\"";
+                    result += parameter.first;
+                    result += "\" value=\"";
+                    result += TfStringify(parameter.second);
+                    result += "\"/>\n";
+                }
+
+                result += "    </params>\n";
+
+                result += "  </node>\n";
+            }
+        } else {
+            for (HdMaterialNode const& node : mat.nodes) {
+                result += "  <node name=\"";
+                result += node.path.GetName();
+                result += "\" identifier=\"";
+                result += node.identifier;
+                result += "\"/>\n";
+            }
         }
+
+        result += "</nodes>\n";
     }
 
-    std::cout << "  --Connections--\n";
-    for (const HdMaterialRelationship& rel : mat.relationships) {
-        std::cout << "    " << rel.inputId << "." << rel.inputName << "->" << rel.outputId << "."
-                  << rel.outputName << "\n";
+    if (!mat.relationships.empty()) {
+        result += "<relationships>\n";
+
+        for (const HdMaterialRelationship& rel : mat.relationships) {
+            result += "  <relationship from=\"";
+            result += rel.inputId.GetName();
+            result += ".";
+            result += rel.inputName;
+            result += "\" to=\"";
+            result += rel.outputId.GetName();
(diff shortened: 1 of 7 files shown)


lib/mayaUsd/render/vp2RenderDelegate/render_delegate.h
```diff
@@ -155,10 +159,12 @@ class HdVP2RenderDelegate final : public HdRenderDelegate
         _resourceRegistry; //!< Shared and unused by VP2 resource registry
 
     std::unique_ptr<HdVP2RenderParam>
-            _renderParam; //!< Render param used to provided access to VP2 during prim synchronization
-    SdfPath _id;          //!< Render delegate IDs
+        _renderParam; //!< Render param used to provided access to VP2 during prim synchronization
+    SdfPath _id;      //!< Render delegate ID
```
HdC-adsk, line 163: It is a change by clang-format.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -113,35 +113,82 @@ inline bool _IsUsdUVTexture(const HdMaterialNode& node)
...
 }
 
-//! Helper utility function to print nodes, connections and primvars in the
+//! Helper function to generate a XML string about nodes, relationships and primvars in the
 //! specified material network.
-void _PrintMaterialNetwork(
-    const std::string&       label,
-    const SdfPath&           id,
-    const HdMaterialNetwork& mat)
+std::string _GenerateXMLString(const HdMaterialNetwork& mat, bool includeParams=true)
```
kxl-adsk, line 118: Looking at HdMaterialNode and HdMaterialRelationship, it has members that can return a hash. Did you consider generating hash directly out of objects without having to create a big string and computing a hash for it?

lib/mayaUsd/render/vp2RenderDelegate/render_delegate.h
kxl-adsk: It doesn't seem to like comments after a member declaration. Wondering if there is a better configuration for it OR we should start simply putting documentation above the member.

kxl-adsk (approved):

lib/mayaUsd/render/vp2RenderDelegate/render_delegate.h
HdC-adsk: Indeed. I was thinking about having the doc above the member, but that would be another refactoring change. I am not familiar with clang-format, so I am not sure if there is a special configuration for this.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
HdC-adsk: Yes, computing the hash from the members is an option but we would have to use the `HdMaterialNetwork` as the key to index the shader cache?

If so, it wouldn't work because the equality comparison between `HdMaterialNetwork` includes the path comparison, making it too strict to support reuse.

JGamache-autodesk: The material network is a large stream of TfTokens, each one having already a hash computed. So you could probably generate code to combine the hashes in a way that is very similar to that XML representation and use the final hash as key for the shader instance. USD uses boost::hash_combine a lot for these scenarios. This could save quite a few CPU cycles.

HdC-adsk: My concern is that we cannot use hash as the key to index the shader cache because the hash code can collide, meaning two different networks (with different nodes or different relationships) can produce the same hash.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -113,35 +113,82 @@ inline bool _IsUsdUVTexture(const HdMaterialNode& node)
...
+    std::string result;
+    result.reserve(1024);
+
+    if (!mat.nodes.empty()) {
+        result += "<nodes>\n";
+
+        if (includeParams) {
+            for (HdMaterialNode const& node : mat.nodes) {
+                result += "  <node name=\"";
+                result += node.path.GetName();
```
JGamache-autodesk, line 129: Not sure about this one. You definitely do not want the full path since it includes the material name and this would prevent useful merges. The problem I see is with ambiguous paths inside the material. Have you noticed how often the name "TexCoordReader" comes back in a Maya export? You probably need the relative path rooted at the material: something that truncates `/pSphere1/Looks/standardSurface2SG/file1/TexCoordReader.outputs` into `./file1/TexCoordReader`.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
HdC-adsk: I was thinking of using a stripped material network (which removes material paths and params) as the key so that we can use the proposed hash_combine approach, but using it as index also has higher cost than a single-token key. So there is pros & cons for each approach. I would prefer to keep this in mind and when we find the code become a hotspot we explore the above proposal.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
HdC-adsk: Good point, Jerry. I never noticed the hierarchy in the material network. I will try to reproduce this scenario first. Thanks.

JGamache-autodesk (changes requested):
This will indeed save a lot of shader recompilations. Thanks!

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
JGamache-autodesk: Maybe use something like std::vector<TfToken> as key then? That would be more compact than generating a large string and the hash_combine strategy would still work.

HdC-adsk: Using std::vector as key is similar to using a stripped material network as key, both would need more equality comparison than single token, while saving compute cost of hash code. I can try this after PR121 based on profiling.

HdC-adsk:
I have to rebase the branch so that it can compiled with latest UFE code. Aside from that, the new commit 59ac9e6 is to address the change proposed by @JGamache-autodesk .

JGamache-autodesk (approved):

JGamache-autodesk (approved):

HdC-adsk:
@kxl-adsk Please merge.

HdC-adsk:
Another commit for this PR: 528f2c1. Click the commit link to see the reasons please.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -482,7 +503,7 @@ _LoadTexture(const std::string& path, bool& isColorSpaceSRGB, MFloatArray& uvSca
 #if USD_VERSION_NUM >= 2102
     HioImageSharedPtr image = HioImage::OpenForReading(path);
 #else
-    GlfImageSharedPtr     image = GlfImage::OpenForReading(path);
+    GlfImageSharedPtr image = GlfImage::OpenForReading(path);
```
HdC-adsk, line 506: @kxl-adsk Any idea why github clang-format-linter doesn't like this change?

JGamache-autodesk (approved):

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
HdC-adsk: It is one more mysterious mystery about clang format. I will revert this change anyway so we don't have to wait on it.

HdC-adsk: I've logged the problem in the wiki.

PR #952, opened 2020-11-30
Added Blendshape export functionality

ysiewappl:
Hi all:

This PR adds support for exporting blendshapes out of Maya in line with the `UsdSkelBlendShape` schema for a variety of scenarios, including single base - single target, single base - multiple targets (incl. inbetween), multiple base - multiple targets, writing out combined extents, and writing out animation. Some limitations include:

- Currently no support for "chained" blendshape deformations or other intermediate deformers in the deform stack.
- "Layered" blendshape deformers will end up having individual shapes baked out individually.
- Combined extents only take blendshapes into account and no other deformers.

I ran into an issue while implementing this, which can be summarized by the following:

Maya does not evaluate certain plugs on a blendshape node until evaluation of the node is forced. This is a problem because certain information like the indices of components on a blendshape cannot be determined _unless_ the blendshape node itself is evaluated.

Steps to  Reproduce:

1. Create a cube and a blendshape for it with one vertex moved.
2. Run the following code:

```
import maya.OpenMaya as om
sl = om.MSelectionList()
p = om.MPlug()
sl.add('blendShape1.inputTarget[0].inputTargetGroup[0].inputTargetItem[6000].inputComponentsTarget')
sl.getPlug(0, p)

dh = p.asMDataHandle()
d= dh.data()
fn = om.MFnComponentListData(d)
print fn.length()
```

3. You should have gotten an error when attempting to attach the function set. Now twiddle the weight on the blendshape deformer and twiddle it back and then run the same code again.
4. The code suddenly works without any other modifications to the scene or its data.

We're currently working around this by basically twiddling the blendshape node's weights through OpenMaya prior to export, which is not ideal since that could potentially be expensive for denser meshes. Hopefully someone from ADSK could help advise here on any possible remedies.

### DIFF
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
@@ -53,162 +53,6 @@
 
 #define MAYA_ATTR_NAME_INMESH "inMesh"
 
-MString mayaGetUniqueNameOfDAGNode(const MObject& node)
-{
-    assert(!node.isNull());
-    assert(node.hasFn(MFn::kDagNode));
-    MStatus    stat;
-    MFnDagNode fnNode(node, &stat);
-    CHECK_MSTATUS_AND_RETURN(stat, MString());
-    MString nodeName = fnNode.partialPathName(&stat);
-    return nodeName;
-}
-
-MPlug mayaFindChildPlugWithName(const MPlug& parent, const MString& name)
-{
-    MPlug sentinel;
-    if (!parent.isCompound() || parent.isNull()) {
-        return sentinel;
-    }
-    MStatus      stat;
-    unsigned int numChildren = parent.numChildren(&stat);
-    CHECK_MSTATUS_AND_RETURN(stat, sentinel);
-    if (numChildren == 0) {
-        return sentinel;
-    }
-
-    MFnAttribute fnAttr;
-
-    // TODO: (yliangsiew) for a certain threshold of child plugs, might want to
-    //       binary search instead.
-    for (unsigned int i = 0; i < numChildren; ++i) {
-        MPlug plgChild = parent.child(i, &stat);
-        CHECK_MSTATUS_AND_RETURN(stat, sentinel);
-        MObject attrChild = plgChild.attribute(&stat);
-        CHECK_MSTATUS_AND_RETURN(stat, sentinel);
-        stat = fnAttr.setObject(attrChild);
-        CHECK_MSTATUS_AND_RETURN(stat, sentinel);
-        const MString attrName = fnAttr.name();
-        if (attrName == name) {
-            return plgChild;
-        }
-    }
-
-    return sentinel;
-}
-
-MStatus mayaGetSkinClusterConnectedToMesh(const MObject& mesh, MObject& skinCluster)
-{
-    // TODO: (yliangsiew) Do we care about multiple skinCluster layers? How do we even want
-    //        to deal with that, if at all?
-    MStatus stat;
-    if (!mesh.hasFn(MFn::kMesh)) {
-        return MStatus::kInvalidParameter;
-    }
-
-    MFnDependencyNode fnNode(mesh, &stat);
-    CHECK_MSTATUS_AND_RETURN_IT(stat);
-
-    MPlug inMeshPlug = fnNode.findPlug(MAYA_ATTR_NAME_INMESH, false, &stat);
-    CHECK_MSTATUS_AND_RETURN_IT(stat);
-
-    bool isDest = inMeshPlug.isDestination(&stat);
-    CHECK_MSTATUS_AND_RETURN_IT(stat);
-    if (!isDest) {
-        return MStatus::kFailure;
-    }
-    MPlug srcPlug = inMeshPlug.source(&stat);
-    CHECK_MSTATUS_AND_RETURN_IT(stat);
-    if (srcPlug.isNull()) {
-        return MStatus::kFailure;
-    }
-
-    skinCluster = srcPlug.node(&stat);
-
-    CHECK_MSTATUS_AND_RETURN_IT(stat);
-    if (!skinCluster.hasFn(MFn::kSkinClusterFilter)) {
-        return MStatus::kFailure;
-    }
-
-    return stat;
-}
-
-MStatus mayaGetSkinClustersUpstreamOfMesh(const MObject& mesh, MObjectArray& skinClusters)
-{
-    MStatus stat;
-    if (!mesh.hasFn(MFn::kMesh)) {
-        return MStatus::kInvalidParameter;
-    }
-
-    skinClusters.clear();
-    MObject            searchObj = MObject(mesh);
-    MItDependencyGraph itDg(
-        searchObj,
-        MFn::kInvalid,
-        MItDependencyGraph::kUpstream,
-        MItDependencyGraph::kDepthFirst,
-        MItDependencyGraph::kNodeLevel,
-        &stat);
-    while (!itDg.isDone()) {
-        MObject curNode = itDg.currentItem();
-        if (curNode.hasFn(MFn::kSkinClusterFilter)) {
-            skinClusters.append(curNode);
-        }
-        itDg.next();
-    }
-
-    return stat;
-}
-
-bool mayaSearchMIntArray(const int a, const MIntArray& array, unsigned int* idx)
-{
-    for (unsigned int i = 0; i < array.length(); ++i) {
-        if (array[i] == a) {
-            if (idx != NULL) {
-                *idx = i;
-            }
-            return true;
-        }
-    }
-    if (idx != NULL) {
-        *idx = -1;
-    }
-    return false;
-}
-
-MBoundingBox mayaCalcBBoxOfMeshes(const MObjectArray& meshes)
-{
-    unsigned int numMeshes = meshes.length();
-    MFnMesh      fnMesh;
-    MStatus      stat;
-    MVector      a;
-    MVector      b;
-    for (unsigned int i = 0; i < numMeshes; ++i) {
-        MObject curMesh = meshes[i];
-        assert(curMesh.hasFn(MFn::kMesh));
-        fnMesh.setObject(curMesh);
-        unsigned int numVertices = fnMesh.numVertices();
-        const float* meshPts = fnMesh.getRawPoints(&stat);
-        for (unsigned int j = 0; j < numVertices; ++j) {
-            float x = meshPts[j * 3];
-            float y = meshPts[(j * 3) + 1];
-            float z = meshPts[(j * 3) + 2];
-
-            a.x = x < a.x ? x : a.x;
-            b.x = x > b.x ? x : b.x;
-
-            a.y = y < a.y ? y : a.y;
-            b.y = y > b.y ? y : b.y;
-
-            a.z = z < a.z ? z : a.z;
-            b.z = z > b.z ? z : b.z;
-        }
-    }
-
-    MBoundingBox result = MBoundingBox(MPoint(a), MPoint(b));
-    return result;
-}
-
 PXR_NAMESPACE_OPEN_SCOPE
 
 // These tokens are supported Maya attributes used for Mesh surfaces
@@ -499,6 +343,105 @@ GfVec3f LinearColorFromColorSet(const MColor& mayaColor, bool shouldConvertToLin
 }
 
 } // anonymous namespace
+
+MStatus
+UsdMayaMeshWriteUtils::getSkinClusterConnectedToMesh(const MObject& mesh, MObject& skinCluster)
+{
+    // TODO: (yliangsiew) Do we care about multiple skinCluster layers? How do we even want
+    //        to deal with that, if at all?
+    MStatus stat;
+    if (!mesh.hasFn(MFn::kMesh)) {
+        return MStatus::kInvalidParameter;
+    }
+
+    MFnDependencyNode fnNode(mesh, &stat);
+    CHECK_MSTATUS_AND_RETURN_IT(stat);
+
+    MPlug inMeshPlug = fnNode.findPlug(MAYA_ATTR_NAME_INMESH, false, &stat);
+    CHECK_MSTATUS_AND_RETURN_IT(stat);
+
+    bool isDest = inMeshPlug.isDestination(&stat);
+    CHECK_MSTATUS_AND_RETURN_IT(stat);
+    if (!isDest) {
+        return MStatus::kFailure;
+    }
+    MPlug srcPlug = inMeshPlug.source(&stat);
+    CHECK_MSTATUS_AND_RETURN_IT(stat);
+    if (srcPlug.isNull()) {
+        return MStatus::kFailure;
+    }
+
+    skinCluster = srcPlug.node(&stat);
+
+    CHECK_MSTATUS_AND_RETURN_IT(stat);
+    if (!skinCluster.hasFn(MFn::kSkinClusterFilter)) {
+        return MStatus::kFailure;
+    }
+
+    return stat;
+}
+
+MStatus UsdMayaMeshWriteUtils::getSkinClustersUpstreamOfMesh(
+    const MObject& mesh,
+    MObjectArray&  skinClusters)
+{
+    MStatus stat;
+    if (!mesh.hasFn(MFn::kMesh)) {
+        return MStatus::kInvalidParameter;
+    }
+
+    skinClusters.clear();
+    MObject            searchObj = MObject(mesh);
+    MItDependencyGraph itDg(
+        searchObj,
+        MFn::kInvalid,
+        MItDependencyGraph::kUpstream,
+        MItDependencyGraph::kDepthFirst,
+        MItDependencyGraph::kNodeLevel,
+        &stat);
+    while (!itDg.isDone()) {
+        MObject curNode = itDg.currentItem();
+        if (curNode.hasFn(MFn::kSkinClusterFilter)) {
+            skinClusters.append(curNode);
+        }
+        itDg.next();
+    }
+
+    return stat;
+}
+
+MBoundingBox UsdMayaMeshWriteUtils::calcBBoxOfMeshes(const MObjectArray& meshes)
+{
+    unsigned int numMeshes = meshes.length();
+    MFnMesh      fnMesh;
+    MStatus      stat;
+    MVector      a;
+    MVector      b;
+    for (unsigned int i = 0; i < numMeshes; ++i) {
+        MObject curMesh = meshes[i];
+        TF_VERIFY(curMesh.hasFn(MFn::kMesh));
+        fnMesh.setObject(curMesh);
+        unsigned int numVertices = fnMesh.numVertices();
+        const float* meshPts = fnMesh.getRawPoints(&stat);
+        for (unsigned int j = 0; j < numVertices; ++j) {
+            float x = meshPts[j * 3];
+            float y = meshPts[(j * 3) + 1];
+            float z = meshPts[(j * 3) + 2];
+
+            a.x = x < a.x ? x : a.x;
+            b.x = x > b.x ? x : b.x;
+
+            a.y = y < a.y ? y : a.y;
+            b.y = y > b.y ? y : b.y;
+
+            a.z = z < a.z ? z : a.z;
+            b.z = z > b.z ? z : b.z;
+        }
+    }
+
+    MBoundingBox result = MBoundingBox(MPoint(a), MPoint(b));
+    return result;
+}
 
 bool UsdMayaMeshWriteUtils::getMeshNormals(
     const MFnMesh& mesh,
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
@@ -36,14 +36,178 @@
 #include <pxr/usd/usdGeom/tokens.h>
 #include <pxr/usd/usdUtils/pipeline.h>
 
+#include <maya/MBoundingBox.h>
+#include <maya/MFnAttribute.h>
+#include <maya/MFnMesh.h>
 #include <maya/MFnSet.h>
 #include <maya/MGlobal.h>
 #include <maya/MIntArray.h>
+#include <maya/MItDependencyGraph.h>
 #include <maya/MItMeshFaceVertex.h>
 #include <maya/MPlug.h>
 #include <maya/MPlugArray.h>
+#include <maya/MPoint.h>
 #include <maya/MStatus.h>
 #include <maya/MUintArray.h>
+#include <maya/MVector.h>
+
+#define MAYA_ATTR_NAME_INMESH "inMesh"
+
+MString mayaGetUniqueNameOfDAGNode(const MObject& node)
+{
+    assert(!node.isNull());
+    assert(node.hasFn(MFn::kDagNode));
+    MStatus    stat;
+    MFnDagNode fnNode(node, &stat);
+    CHECK_MSTATUS_AND_RETURN(stat, MString());
+    MString nodeName = fnNode.partialPathName(&stat);
+    return nodeName;
+}
+
+MPlug mayaFindChildPlugWithName(const MPlug& parent, const MString& name)
+{
+    MPlug sentinel;
+    if (!parent.isCompound() || parent.isNull()) {
+        return sentinel;
+    }
+    MStatus      stat;
+    unsigned int numChildren = parent.numChildren(&stat);
+    CHECK_MSTATUS_AND_RETURN(stat, sentinel);
+    if (numChildren == 0) {
+        return sentinel;
+    }
+
+    MFnAttribute fnAttr;
+
+    // TODO: (yliangsiew) for a certain threshold of child plugs, might want to
+    //       binary search instead.
+    for (unsigned int i = 0; i < numChildren; ++i) {
+        MPlug plgChild = parent.child(i, &stat);
+        CHECK_MSTATUS_AND_RETURN(stat, sentinel);
+        MObject attrChild = plgChild.attribute(&stat);
+        CHECK_MSTATUS_AND_RETURN(stat, sentinel);
+        stat = fnAttr.setObject(attrChild);
+        CHECK_MSTATUS_AND_RETURN(stat, sentinel);
+        const MString attrName = fnAttr.name();
+        if (attrName == name) {
+            return plgChild;
+        }
+    }
+
+    return sentinel;
+}
+
+MStatus mayaGetSkinClusterConnectedToMesh(const MObject& mesh, MObject& skinCluster)
+{
+    // TODO: (yliangsiew) Do we care about multiple skinCluster layers? How do we even want
+    //        to deal with that, if at all?
+    MStatus stat;
+    if (!mesh.hasFn(MFn::kMesh)) {
+        return MStatus::kInvalidParameter;
+    }
+
+    MFnDependencyNode fnNode(mesh, &stat);
+    CHECK_MSTATUS_AND_RETURN_IT(stat);
+
+    MPlug inMeshPlug = fnNode.findPlug(MAYA_ATTR_NAME_INMESH, false, &stat);
+    CHECK_MSTATUS_AND_RETURN_IT(stat);
+
+    bool isDest = inMeshPlug.isDestination(&stat);
+    CHECK_MSTATUS_AND_RETURN_IT(stat);
+    if (!isDest) {
+        return MStatus::kFailure;
+    }
+    MPlug srcPlug = inMeshPlug.source(&stat);
+    CHECK_MSTATUS_AND_RETURN_IT(stat);
+    if (srcPlug.isNull()) {
+        return MStatus::kFailure;
+    }
+
+    skinCluster = srcPlug.node(&stat);
+
+    CHECK_MSTATUS_AND_RETURN_IT(stat);
+    if (!skinCluster.hasFn(MFn::kSkinClusterFilter)) {
+        return MStatus::kFailure;
+    }
+
+    return stat;
+}
+
+MStatus mayaGetSkinClustersUpstreamOfMesh(const MObject& mesh, MObjectArray& skinClusters)
+{
+    MStatus stat;
+    if (!mesh.hasFn(MFn::kMesh)) {
+        return MStatus::kInvalidParameter;
+    }
+
+    skinClusters.clear();
+    MObject            searchObj = MObject(mesh);
+    MItDependencyGraph itDg(
+        searchObj,
+        MFn::kInvalid,
+        MItDependencyGraph::kUpstream,
+        MItDependencyGraph::kDepthFirst,
+        MItDependencyGraph::kNodeLevel,
+        &stat);
+    while (!itDg.isDone()) {
+        MObject curNode = itDg.currentItem();
+        if (curNode.hasFn(MFn::kSkinClusterFilter)) {
+            skinClusters.append(curNode);
+        }
+        itDg.next();
+    }
+
+    return stat;
+}
+
+bool mayaSearchMIntArray(const int a, const MIntArray& array, unsigned int* idx)
+{
+    for (unsigned int i = 0; i < array.length(); ++i) {
+        if (array[i] == a) {
+            if (idx != NULL) {
+                *idx = i;
+            }
+            return true;
+        }
+    }
+    if (idx != NULL) {
+        *idx = -1;
+    }
+    return false;
+}
+
+MBoundingBox mayaCalcBBoxOfMeshes(const MObjectArray& meshes)
+{
+    unsigned int numMeshes = meshes.length();
+    MFnMesh      fnMesh;
+    MStatus      stat;
+    MVector      a;
+    MVector      b;
+    for (unsigned int i = 0; i < numMeshes; ++i) {
+        MObject curMesh = meshes[i];
+        assert(curMesh.hasFn(MFn::kMesh));
+        fnMesh.setObject(curMesh);
+        unsigned int numVertices = fnMesh.numVertices();
+        const float* meshPts = fnMesh.getRawPoints(&stat);
+        for (unsigned int j = 0; j < numVertices; ++j) {
+            float x = meshPts[j * 3];
+            float y = meshPts[(j * 3) + 1];
+            float z = meshPts[(j * 3) + 2];
+
+            a.x = x < a.x ? x : a.x;
+            b.x = x > b.x ? x : b.x;
+
+            a.y = y < a.y ? y : a.y;
+            b.y = y > b.y ? y : b.y;
+
+            a.z = z < a.z ? z : a.z;
+            b.z = z > b.z ? z : b.z;
+        }
+    }
+
+    MBoundingBox result = MBoundingBox(MPoint(a), MPoint(b));
+    return result;
+}
 
 PXR_NAMESPACE_OPEN_SCOPE
 
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
@@ -386,7 +386,7 @@ MStatus UsdMayaMeshWriteUtils::getSkinClustersUpstreamOfMesh(
     MObjectArray&  skinClusters)
 {
     MStatus stat;
-    if (!mesh.hasFn(MFn::kMesh)) {
+    if (mesh.isNull() || !mesh.hasFn(MFn::kMesh)) {
         return MStatus::kInvalidParameter;
     }
 
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.h
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.h
@@ -35,25 +35,12 @@
 #include <maya/MObject.h>
 #include <maya/MString.h>
 
-/**
- * Gets the minimum unique name of a DAG node.
- *
- * @param node  The node to find the name of.
- *
- * @return      The name as a Maya string.
- */
-MString mayaGetUniqueNameOfDAGNode(const MObject& node);
+PXR_NAMESPACE_OPEN_SCOPE
 
-/**
- * Finds a child plug with the given `name`.
- *
- * @param parent    The parent plug to start the search from.
- * @param name      The name of the child plug to find. This should be the short name.
- *
- * @return          The plug if it can be found, or a null `MPlug` otherwise.
- */
-MPlug mayaFindChildPlugWithName(const MPlug& parent, const MString& name);
+class UsdGeomMesh;
 
+// Utilities for dealing with writing USD from Maya mesh/subdiv tags.
+namespace UsdMayaMeshWriteUtils {
 /**
  * Finds a skinCluster directly connected upstream in the DG to the given mesh.
  *
@@ -63,10 +50,11 @@ MPlug mayaFindChildPlugWithName(const MPlug& parent, const MString& name);
  *
  * @return                  `MStatus::kSuccess` if the operation completed successfully.
  */
-MStatus mayaGetSkinClusterConnectedToMesh(const MObject& mesh, MObject& skinCluster);
+MAYAUSD_CORE_PUBLIC
+MStatus getSkinClusterConnectedToMesh(const MObject& mesh, MObject& skinCluster);
 
 /**
- * Similar to `mayaGetSkinClusterConnectedToMesh`, except that instead of finding a
+ * Similar to `getSkinClusterConnectedToMesh`, except that instead of finding a
  * directly-connected skinCluster, it searches the DG upstream of the mesh for any
  * other skinClusters as well.
  *
@@ -75,19 +63,8 @@ MStatus mayaGetSkinClusterConnectedToMesh(const MObject& mesh, MObject& skinClus
  *
  * @return                  `MStatus::kSuccess` if the operation completed successfully.
  */
-MStatus mayaGetSkinClustersUpstreamOfMesh(const MObject& mesh, MObjectArray& skinClusters);
-
-/**
- * Searches the given array for an element.
- *
- * @param a         The element to search for.
- * @param array     The array to search within.
- * @param idx       Storage for the index of the element within the array if it exists.
- *                  If it does not exist, this will be undefined.
- *
- * @return          Returns ``true`` if the element exists in the array, ``false`` otherwise.
- */
-bool mayaSearchMIntArray(const int a, const MIntArray& array, unsigned int* idx);
+MAYAUSD_CORE_PUBLIC
+MStatus getSkinClustersUpstreamOfMesh(const MObject& mesh, MObjectArray& skinClusters);
 
 /**
  * Calculates the union bounding box of a given array of meshes.
@@ -96,14 +73,9 @@ bool mayaSearchMIntArray(const int a, const MIntArray& array, unsigned int* idx)
  *
  * @return          The union bounding box.
  */
-MBoundingBox mayaCalcBBoxOfMeshes(const MObjectArray& meshes);
+MAYAUSD_CORE_PUBLIC
+MBoundingBox calcBBoxOfMeshes(const MObjectArray& meshes);
 
-PXR_NAMESPACE_OPEN_SCOPE
-
-class UsdGeomMesh;
-
-// Utilities for dealing with writing USD from Maya mesh/subdiv tags.
-namespace UsdMayaMeshWriteUtils {
 /// Helper method for getting Maya mesh normals as a VtVec3fArray.
 MAYAUSD_CORE_PUBLIC
 bool getMeshNormals(const MFnMesh& mesh, VtVec3fArray* normalsArray, TfToken* interpolation);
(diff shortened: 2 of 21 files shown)


williamkrick:
@ysiewappl You are correct several of the inputTarget attributes are only computed as a side effect of computing the aOutputGeom plug that requires the data. If part of your step 1 does trigger that evaluation, say by the object drawing in the the viewport, then I would expect the python script to work. I did a quick test locally in 2020.3 and performed my step 1 interactively in the viewport using the shape editor, and I didn't get an error. Can you share the details of your step 1 so I can more closely reproduce what you are seeing? I'm guessing you are doing the export as a part of a batch job and that's where you see the issue?

lib/usd/translators/meshWriter.cpp
```diff
@@ -54,6 +59,57 @@
...
+    stat = mayaGetSkinClusterConnectedToMesh(mesh, skinCluster);
+    if (stat) {
+        searchObject = MObject(skinCluster);
+    } else {
+        searchObject = MObject(mesh);
+    }
+
+    // TODO: (yliangsiew) Problem: if there are _intermediate deformers between blendshapes,
+    // then what do we do? Like blendshape1 -> wrap -> blendshape2. This won't find
+    // that correctly...
```
williamkrick, line 80: Maya's GPU Deformer has a similar problem where it has to parse deformation graphs and figure out what can be supported by GPU evaluation and what is not. We do something similar to your itDg except we don't set the node type filter and do that part manually in the loop, testing to see if each upstream node is something supported as we go.

You could do something similar here. The code knows the set of supported deformer types that can be exported to UsdSkel so you can search upwards and find either the static base mesh, or the set of boundary nodes which are animated but not supported.

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -0,0 +1,931 @@
...
+            stat = fnBS.targetItemIndexList(weightIndex, baseObj, targetItemIndexList);
+            CHECK_MSTATUS_AND_RETURN_IT(stat);
+
+            for (unsigned int k = 0; k < targetItemIndexList.length(); ++k) {
+                float curWeight = fnBS.weight(weightIndex, &stat);
+                CHECK_MSTATUS_AND_RETURN_IT(stat);
+                origWeightVals.append(curWeight);
+                // NOTE: (yliangsiew) For in-between shapes, need to trigger at _all_
+                // full weight values of each target so as to populate the components
+                // list for each of the targets. Yea, this is dumb.
```
williamkrick, line 388: I see it is even more annoying that I realized. The blend shape is doing lazy evaluation and only computing input target information for targets it needs, so in the in-between case there will be lots of targets that don't get evaluated.

lib/usd/translators/meshWriter.cpp
williamkrick: Thinking about it more carefully I think there is an issue with the MItDependencyGraph usage you have. The intent of the code is to follow the deformation chain upstream to find the original mesh. That means you should only be following connections which are carrying mesh data, not other data.

Imagine a scene where there are two separate meshes being driven by blend shapes. If the weight attribute of blendShape1 is driven by the weight attribute of blendShape2 then when searching for the upstream original mesh of blendShape1 the iterator will find both blendShape1 and blendShape2. If blendShape2 is found first then an incorrect inputGeo (originalMesh2) can be returned, a mesh which can have no relationship to deformedMesh1.

Our GPU Override code solves this by limiting the search plugs to only geometry plugs which affect the output geometry & which must match the output geometry's topology. So for a blend shape the upstream search plugs are inputTargetGeom and inputGeom. We have lots of cases where inputTargetGeom has incoming deformation.

williamkrick (changes requested):
What happens when a scene with animated blend shape targets is exported?

ysiewappl:
> @ysiewappl You are correct several of the inputTarget attributes are only computed as a side effect of computing the aOutputGeom plug that requires the data. If part of your step 1 does trigger that evaluation, say by the object drawing in the the viewport, then I would expect the python script to work. I did a quick test locally in 2020.3 and performed my step 1 interactively in the viewport using the shape editor, and I didn't get an error. Can you share the details of your step 1 so I can more closely reproduce what you are seeing? I'm guessing you are doing the export as a part of a batch job and that's where you see the issue?

Hi:

Run the following code before the aforementioned Python code and you should see the issue

```polyCube;
duplicate -rr;
select -r pCube2.vtx[2] ;
move -r -0.781387 0.620818 1.190132 ;
select -r pCube2 ;
select -tgl pCube1 ;
blendShape;

# Error: RuntimeError: file /local/S/Maya_2020_DI/build/RelWithDebInfo/runTime/Maya.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python2.7/site-packages/maya/OpenMaya.py line 4881: (kInvalidParameter): Argument is a NULL pointer # ```
Once you twiddle the weights, you can run the Python code again and you should get the proper result, which is:

```import maya.OpenMaya as om
sl = om.MSelectionList()
p = om.MPlug()
sl.add('blendShape1.inputTarget[0].inputTargetGroup[0].inputTargetItem[6000].inputComponentsTarget')
sl.getPlug(0, p)

dh = p.asMDataHandle()
d= dh.data()
fn = om.MFnComponentListData(d)
print fn.length()
1```

File: lib/usd/translators/meshWriter_BlendShapes.cpp [hunk 3]
```diff
@@ -0,0 +1,931 @@
...
+    for (unsigned int i = 0; i < numOffsets; ++i) {
+        MPoint curPt = mptTargetOffsets[i];
+        targetOffsets[i] = GfVec3f(curPt.x, curPt.y, curPt.z);
+    }
+
+    MPlug plgTargetGeom = mayaFindChildPlugWithName(plgInputTargetItem, "inputGeomTarget");
+    if (plgTargetGeom.isNull()) {
+        return MStatus::kFailure;
+    }
+    MObject targetGeom = plgTargetGeom.asMObject();
```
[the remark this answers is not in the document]
ysiewappl: You're right, animated targets are another limitation I didn't think about here. After some internal discussion, it appears that UsdSkelBlendShape offsets and normalOffsets are currently uniforms, which means that they can't be time-sampled apparently and thus it would be tricky to support animated blendshape targets the way Maya does. I think the suggestion you gave about blocking an export of animated shape targets (or warning) makes the most sense for now and I'll see about adding this. What do you think? (Warning VS hard block of export)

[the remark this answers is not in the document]
williamkrick: Warning vs. a blocking error should be an export configuration, where the default value is to fail but the user can override the value and get the best-effort export that can be produced.

The same approach can help with some of the other ambiguous bits of code around what happens when there are multiple upstream blend shapes, or if there is a wrap deformer between blend shapes. In those cases you can error out saying the exporter doesn't support that pattern, but you can try forcing it and see what you get.

lib/usd/translators/meshWriter.cpp
ysiewappl: Could you elaborate a little on this setup that you're referring to? I've tried to re-create as described above. In the case shown in the diagram, whether starting from pCube1baseShape or pCube2baseShape, I think either should find the correct corresponding origShape node (since I exit iteration once the mesh is found, thus it shouldn't find the other blendshape node, regardless of the fact that there is a connection between the two blendshape nodes by virtue of the weight attribute of one shape driving the other). I might be misunderstanding here.

image

lib/usd/translators/meshWriter_BlendShapes.cpp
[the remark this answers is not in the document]
ysiewappl: When you say "export configuration", do you mean having an additional flag for the mayaUSDExport command i.e. ignoreErrors or similar? I can certainly add a catch-all option for users to override what happens; or are you referring to something else?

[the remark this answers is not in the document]
williamkrick: Yes an additional flag on the export command.

lib/usd/translators/meshWriter.cpp
williamkrick: You are correct the code you have can't return the input of the wrong blend shape. I misunderstood the power of MFnGeometryFilter::inputShapeAtIndex().

The problem with this function is if there are more deformers upstream of the blend shape's input geometry then inputShapeAtIndex skips over them and goes directly to a shape. There could be other deformers between the discovered mesh and the input to the blend shape node. For example those tweak nodes do not have to be empty, or there could be other animated deformers present. In that case the original mesh data you read won't match the mesh data Maya has at the input of the blend shape, and the deformation exported to USD won't match Maya.

In addition to that the discovered mesh node doesn't have to be static. The mesh can have an input mesh connection that can be animated with deformation as well.

The problem is Maya can support any combination of these nodes together. The code here must do more pattern matching to ensure that the deformation described by the DG is Maya matches a case that the exporter understands. 

My suggestion to you would be to be extremely specific about the configuration of supported nodes to limit what is allowed to be exported and what is not. Trying to handle exporting arbitrary graphs of deformers is a huge task that I don't want this change to get bogged down in. Focus this change on doing a really good job exporting all the features of a blend shape node (which you already have) but in a simple graph context. The only thing missing is verifying that the graph of deformer nodes matches something the exporter can handle.

lib/usd/translators/meshWriter_BlendShapes.cpp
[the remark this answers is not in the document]
dgovil: Do you think it would make sense to have a single flag that all aspects of the exporter could follow? Something like a `--strict`. 

We could make blendshapes be the first aspect of the exporter that follows that flag, but it would potentially be nicer UX to have a single flag for everything to follow eventually, rather than adding per feature flags?

Of course things that need more granularity could define more flags in the future so it wouldn't need to be mutually exclusive to have a common flag and per feature flags.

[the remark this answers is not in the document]
williamkrick: Yes I think starting with a single flag is the right approach. As you say it can be extended in the future.

We do need to emit a warning for everything that is unsupported but exported anyway so that the user knows what is happening.

lib/usd/translators/meshWriter.cpp
ysiewappl: I have added a new `ignoreWarnings` flag to the `mayaUSDExport` command, which
will control this behaviour (defaulting to `false`). If any unrecognized node is
found, the export will be blocked. Currently the only nodes whitelisted are any
nodes supporting the Fn sets:

- kGroupParts
- kMesh (i.e. mesh piping output to another mesh)

For tweak nodes, they are checked to see if they have any active effect; if not,
they are whitelisted as well. For deformers, they are checked to see if their
envelopes are set to `0.0f` and if so, whitelisted as well.

All other types of nodes will trigger an error message on export. Hopefully this
addresses the issue that you're bringing up?

lib/usd/translators/meshWriter_BlendShapes.cpp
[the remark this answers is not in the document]
ysiewappl: Added; please see: https://github.com/Autodesk/maya-usd/pull/952#discussion_r535728053

lib/usd/translators/meshWriter.cpp
williamkrick: Yes my concerns have been addressed. You've done a good job validating that the deformation expressed in the DG matches the capabilities of the export code.

There is one more thing to test for mesh nodes. Meshes can have tweaks embedded in them in the pnts attribute. I think normally it'll be empty but you can put something in there through MFnMesh::setPoint, in which case the mesh won't be a pass through and safe for your whitelist.

ysiewappl: > There is one more thing to test for mesh nodes. Meshes can have tweaks embedded in them in the pnts attribute. I think normally it'll be empty but you can put something in there through MFnMesh::setPoint, in which case the mesh won't be a pass through and safe for your whitelist.

This should now be addressed as part of `7c733f5`.

williamkrick (approved):

ysiewappl:
Hi @williamkrick and @kxl-adsk:
We're running into another issue here that I'd like to get some info regarding Maya's blendshape node implementation, if possible. Currently we've found that blendshape targets that have been "baked" into the blendshape node (i.e. no more explicit connections in the DG between the target meshes and the inputGeomTarget plugs) will not work with this current code since it relies heavily on the assumption that `MFnBlendShapeDeformer::getTargets` will return the target `MObject` meshes for a given weight index, and then I can walk back from each mesh back to the blendshape node to figure out the rest of the information I need.

If these connections do not exist in the DG, getTargets gives an empty `MObjectArray`. Is there any other way I can figure out, for a given weight index, which inputTargetItems in their corresponding inputTargetGroups are going to be associated with it? I don't quite see a way I can infer this from the available public attributes on the blendshape node itself nor a way for me to replicate what `MFnBlendShapeDeformer::getTargets` in OpenMaya is doing internally. I'm sure the information must be available on the node somewhere, but I don't quite see how I can get to it. If I'm missing something, please let me know.

An example of what I mean by baking is shown below, where the setup before/after "baking" is shown:
image
image

williamkrick:
After baking the target data can be stored in the inputPointsTarget attribute. Each value in inputPointsTarget is the difference vector between the input geometry & the target geometry at the time when the bake occurred. This value is targetPosition - inputPosition. This is one of the first steps to computing the blend shape deformer. After baking we store that delta rather than re-computing it every frame.

I think for your case it should be possible to reconstruct what the value of inputGeomTarget would be by reversing the calculation so that targetPosition = inputPosition + delta. In that way you can get a target to use as an input to the usdSkel blend shape.

For your second question about in-between weights you are correct. If there are no in-between targets then there is one element in inputTargetItem at index 6000. If there are in-betweens then there will still be an entry at 6000, but there will also be entries between 5000 and 5000. The mapping is index = wt * 1000 + 5000. As the weight increases the actual used target changes between the various in-betweens.

ysiewappl:
Hi @williamkrick:

Thanks for the response. Unfortunately, I think I might not have made my question clear enough. To clarify: I'm more asking how to replicate the functionality of `MFnBlendShapeDeformer::getTargets` while not having actual targets connected to the blendshape node (i.e. not about re-creating "baked" blendshape targets and/or how to determine "activation weight" for a given target).  This is important, because it gets me-

- The appropriate logical index for a parent `inputTarget` plug
- Also the appropriate logical index for a parent `inputTargetGroup`

-that the target mesh is connected to and that a specific `weightIndex` is driving for a given `baseObject`. Maybe a better way to frame my question was: based on the diagram provided, are my assumptions above about how a given weight attribute decides which `inputTargetGroup` items to use for driving the final `outputGeometry` accurate?

(To put it another way: ideally, `MFnBlendShapeDeformer::targetItemIndexList` would return me not just `targetItemIndices`, but also the `inputTargetIndex` and `inputTargetGroupIndices` as well of _where_ I should be looking under the appropriate `inputTargetGroup` plug, so that I would be able to infer this information myself.)

lib/mayaUsd/fileio/utils/meshWriteUtils.h
```diff
@@ -29,11 +29,75 @@
...
 #include <maya/MString.h>
 
+/**
+ * Gets the minimum unique name of a DAG node.
+ *
+ * @param node  The node to find the name of.
+ *
+ * @return      The name as a Maya string.
+ */
+MString mayaGetUniqueNameOfDAGNode(const MObject& node);
```
mattyjams, line 45: Noted this in the .cpp, but is the minimal uniqueness important, or can we remove this and just use `UsdMayaUtil::GetMayaNodeName()`?

williamkrick:
I think I understand what you are asking now. Your assumptions under the diagram are correct.

To compute the output deformed geometry in outputGeometry you need to read all the inputTargetItem's that affect that outputGeometry.

To read an inputTargetItem you need an index for 3 levels of multi attribute: inputTarget, inputTargetGroup and inputTargetItem.

- For inputTarget, the index matches the index of inputGeometry and outputGeometry.
- For inputTargetGroup you'll have to iterate over all of them that exist. The index of the inputTargetGroup will match the index into the weight attribute.
- For inputTargetItem, index varies depending on the current weight value, according to the formula index = wt * 1000 + 5000

In practice we don't see in-betweens used that often, so inputTargetItem almost always gets an index of 6000.

Unfortunately I don't see any methods in MFnBlendShapeDeformer that understand inputPointsTarget, which means you'll have to do it manually. I do that using MPlug and MArrayDataHandle:

This is mostly copied from a bit of internal code that uses the API, so some bits might not make perfect sense but it should be possible to figure out what it is doing.

```
MObject attributeInputTargetGroup; // you have to set these up with MNodeClass::Attribute
MPlug inputTargetGroupPlug(node, attributeInputTargetGroup);
MArrayDataHandle inputTargetGroupArray(inputTargetGroupDataHandle, &status);

const unsigned int inputTargetGroupCount = inputTargetGroupArray.elementCount();
for(unsigned int i=0; i< inputTargetGroupCount; i++, inputTargetGroupArray.next())
{
    int inputTargetGroupLogicalIndex = inputTargetGroupArray.elementIndex();
    MDataHandle inputTargetGroupStructure = inputTargetGroupArray.inputValue(&status);
    MDataHandle inputTargetItemDataHandle = inputTargetGroupStructure.child(attributeInputTargetItem);
    MArrayDataHandle inputTargetItemArray = MArrayDataHandle(inputTargetItemDataHAndle, &status);
    const unsigned int inputTargetItemCount = inputTargetItemArray.elementCount();
    //inputTargetItemCount is usually 1
    //can use the same array iteration pattern to find them all

    //we can extract weights like
    MArrayDataHandle weightArray = block.inputArrayValue(attributeWeight, &status);
    weightArray.jumpToElement(inputTargetGroupLogicalIndex);
    MDataHandle weightData = weightArray.inputValue(&status);
    float weight = weightData.asFloat();

    //for each inputTargetGroupStructure you have to figure out is there something in inputPointsTarget or inputGeomTarget.
}
```

It is not safe to do the iteration based on the weight attribute. It is possible that a weight value exists in the weight array but no inputTargetGroup exists in the targets.

lib/mayaUsd/fileio/utils/meshWriteUtils.h
```diff
@@ -29,11 +29,75 @@
...
+
+/**
+ * Finds a child plug with the given `name`.
+ *
+ * @param parent    The parent plug to start the search from.
+ * @param name      The name of the child plug to find. This should be the short name.
+ *
+ * @return          The plug if it can be found, or a null `MPlug` otherwise.
+ */
+MPlug mayaFindChildPlugWithName(const MPlug& parent, const MString& name);
```
mattyjams, line 55: This seems generally useful and not specific to meshes at all, so I'd suggest that this go in `mayaUsd/utils/util.[h,cpp]`, maybe next to the existing `GetPlugByName()` as `GetChildPlugByName()`:
https://github.com/Autodesk/maya-usd/blob/1439297b4ba79bdcffcfa6551b143ea7ac46f99a/lib/mayaUsd/utils/util.h#L175

lib/mayaUsd/fileio/utils/meshWriteUtils.h
```diff
@@ -29,11 +29,75 @@
...
+ * Searches the given array for an element.
+ *
+ * @param a         The element to search for.
+ * @param array     The array to search within.
+ * @param idx       Storage for the index of the element within the array if it exists.
+ *                  If it does not exist, this will be undefined.
+ *
+ * @return          Returns ``true`` if the element exists in the array, ``false`` otherwise.
+ */
+bool mayaSearchMIntArray(const int a, const MIntArray& array, unsigned int* idx);
```
mattyjams, line 90: This is also not mesh specific and seems better kept in `mayaUsd/utils/util.[h,cpp]`. No good suggestion as to where in there would be best though. :/

lib/mayaUsd/fileio/utils/meshWriteUtils.h
```diff
@@ -29,11 +29,75 @@
 #include <pxr/usd/usdGeom/mesh.h>
 #include <pxr/usd/usdUtils/sparseValueWriter.h>
 
+#include <maya/MBoundingBox.h>
 #include <maya/MDagPath.h>
 #include <maya/MFnMesh.h>
 #include <maya/MObject.h>
 #include <maya/MString.h>
```
mattyjams, line 37: The naming and namespacing is getting a bit muddied in here. The new functions are all in the global namespace with a `maya` prefix, while everything else is inside `PXR_NS`, then further scoped to `UsdMayaMeshWriteUtils`. Is there a reason the mesh-specific functions can't go in the `UsdMayaMeshWriteUtils` namespace? And it would be great to move the non-mesh specific functions out of this file.

lib/usd/translators/meshWriter.cpp
```diff
@@ -41,19 +43,107 @@
...
 
+#include <cassert>
+#include <cfloat>
 #include <set>
 #include <string>
 #include <vector>
 
+MObject mayaFindOrigMeshFromBlendShapeTarget(const MObject& mesh, MObjectArray* intermediates)
+{
+    assert(mesh.hasFn(MFn::kMesh));
```
mattyjams, line 67: Remove assert.

lib/usd/translators/meshWriter.cpp
```diff
@@ -41,19 +43,107 @@
...
+    } else {
+        searchObject = MObject(mesh);
+    }
+
+    // NOTE: (yliangsiew) Problem: if there are _intermediate deformers between blendshapes,
+    // then oh-no: what do we do? Like blendshape1 -> wrap -> blendshape2. This won't find
+    // that correctly...so we just tell the client what we found and let them decide how to
+    // handle it.
+    MFnGeometryFilter fnGeoFilter;
+    assert(MObjectHandle(searchObject).isValid());
```
mattyjams, line 87: Remove assert.

lib/usd/translators/meshWriter.cpp
```diff
@@ -41,19 +43,107 @@
...
+        MItDependencyGraph::kUpstream,
+        MItDependencyGraph::kDepthFirst,
+        MItDependencyGraph::kPlugLevel,
+        &stat);
+    MFnDependencyNode fnNode;
+    for (; !itDg.isDone(); itDg.next()) {
+        MObject curBlendShape = itDg.currentItem();
+        assert(curBlendShape.hasFn(MFn::kBlendShape));
+        MPlug outputGeomPlug = itDg.thisPlug();
+        assert(outputGeomPlug.isElement() == true);
```
mattyjams, line 101: Remove asserts.

lib/usd/translators/meshWriter.cpp
```diff
@@ -87,6 +177,35 @@ PxrUsdTranslators_MeshWriter::PxrUsdTranslators_MeshWriter(
...
 void PxrUsdTranslators_MeshWriter::PostExport() { cleanupPrimvars(); }
 
+bool PxrUsdTranslators_MeshWriter::writeAnimatedMeshExtents(
+    const MObject&     deformedMesh,
+    const UsdTimeCode& usdTime)
+{
+    // NOTE: (yliangsiew) We also cache the animated extents out here; this
+    // will be written at the SkelRoot level later on.
+    assert(!deformedMesh.isNull());
+    assert(deformedMesh.hasFn(MFn::kMesh));
```
mattyjams, line 187: Remove asserts.

lib/usd/translators/meshWriter.cpp
```diff
@@ -163,13 +284,171 @@ bool PxrUsdTranslators_MeshWriter::writeMeshAttrs(
...
+                    CHECK_MSTATUS_AND_RETURN(status, false);
+                    float envelope = fnGeoFilt.envelope();
+                    if (fabs(envelope - 0.0f) < FLT_EPSILON) {
+                        continue; // deformer has no effect
+                    }
+
+                    if (curIntermediate.hasFn(MFn::kTweak)) {
+                        // NOTE: (yliangsiew) Make sure tweak really has no effect even if it's on
+                        MPlug plgVLists = fnGeoFilt.findPlug("vlist");
+                        assert(plgVLists.isArray());
```
mattyjams, line 319: Remove asserts all through here.

lib/usd/translators/meshWriter.cpp
```diff
@@ -163,13 +284,171 @@ bool PxrUsdTranslators_MeshWriter::writeMeshAttrs(
...
+                                                "Could not determine the original blendshape "
+                                                "source mesh due to the tweak node: "
+                                                + fnNode.name()
+                                                + ". Please either bake it down or remove the "
+                                                  "edits and attempt the export process again, or "
+                                                  "specify -ignoreWarnings.");
+                                            TF_RUNTIME_ERROR(
+                                                "Could not determine the original blendshape "
+                                                "source mesh due to a non-empty tweak node, "
+                                                "aborting export.");
```
mattyjams, line 351: Are both of these necessary? I would have thought that the `TF_RUNTIME_ERROR` would be sufficient, and that the diagnostic delegate we install would take care of surfacing the message using `MGlobal::displayError()`:
https://github.com/Autodesk/maya-usd/blob/1439297b4ba79bdcffcfa6551b143ea7ac46f99a/lib/mayaUsd/utils/diagnosticDelegate.cpp#L104

And then a few more cases of this below.

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -0,0 +1,960 @@
...
+
+#include <algorithm>
+#include <complex>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#define MAYA_BLENDSHAPE_EVAL_HOTFIX 1
+
+#define MAYA_ATTR_NAME_WEIGHT "weight"
```
mattyjams, line 57: Prefer typed, named constants over `#define`s.

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -0,0 +1,960 @@
...
+
+struct MayaBlendShapeWeightDatum
+{
+    MObjectArray targetMeshes; // The target shape(s) to hit. (i.e. multiple shapes would be using
+                               // Maya's "in-betweens" feature.)
+    MIntArray
+              inputTargetGroupIndices; // The input group indices at which each target is connected under.
```
mattyjams, line 64: Looks like `clang-format` mangled this? It seems to handle comments better when they're on the line before rather than trailing like this.

ysiewappl:
Hi @williamkrick:

Thanks very much for clarifying and providing your sample code (especially for your note regarding a possible edge case when iterating over the weight attributes, I wasn't aware of that)! I'm currently working on changes to the code. However, do you think it is possible to merge this initially and for me to address this (i.e. supporting baked blendshape targets as well) in a separate PR?

williamkrick:
Yes I would approve merging it now with the understanding that more changes are coming to address the baked targets.

Those future changes could be as small as detecting baked targets and throwing an error, but I would suggest handling the baked targets (which I think you are working on) because we see that relatively frequently in Maya.

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -0,0 +1,960 @@
...
+            info.weightDatas.push_back(weightInfo);
+        }
+        outInfos.push_back(info);
+    }
+    return stat;
+}
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+VtIntArray getUnionOfVtIntArrays(const VtIntArray* const arrays, const size_t count)
```
mattyjams, line 227: Am I missing where this is used, or is it not used anywhere? I was wondering whether it could take a `std::vector` or `VtArray` of `VtIntArray` and avoid the separate count, but even better if we can just get rid of it.

mattyjams (changes requested):
I didn't look too terribly closely at the blend shape-specific stuff since you all are well into discussions on that, and I'd be pretty out of my depth anyway.

From what I saw, I think the two main things that could use cleaning up are removing all of the `assert`s and making sure the newly added utility functions find good homes in the appropriate files and namespaces.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -36,14 +36,178 @@
...
+#include <maya/MItDependencyGraph.h>
 #include <maya/MItMeshFaceVertex.h>
 #include <maya/MPlug.h>
 #include <maya/MPlugArray.h>
+#include <maya/MPoint.h>
 #include <maya/MStatus.h>
 #include <maya/MUintArray.h>
+#include <maya/MVector.h>
+
+#define MAYA_ATTR_NAME_INMESH "inMesh"
```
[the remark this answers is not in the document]
ysiewappl: I mostly rely on macros for Maya attribute names since some areas require `const char *` or `MString`. I'm more inclined to something like

```static constexpr char kMayaAttrNameInMesh[] = "inMesh";```
though, rather than an anonymous namespace. Is there a reason that it helps? I'm actually not sure.

File: lib/mayaUsd/fileio/utils/meshWriteUtils.cpp [hunk 18]
```diff
@@ -36,14 +36,178 @@
...
+
+MString mayaGetUniqueNameOfDAGNode(const MObject& node)
+{
+    assert(!node.isNull());
+    assert(node.hasFn(MFn::kDagNode));
```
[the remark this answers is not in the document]
ysiewappl: This brings up a question for me: Are Maya-USD builds that are released built with asserts enabled? If so, I agree that these can be removed, but I'd wonder if there was a better way I can have runtime asserts for development purposes where I actually _do_ want to halt execution immediately when I'm not handling an error.

lib/mayaUsd/fileio/writeJobContext.cpp
```diff
@@ -590,4 +590,13 @@ void UsdMayaWriteJobContext::MarkSkelBindings(
     _skelBindingsProcessor->MarkBindings(path, skelPath, config);
 }
 
+bool UsdMayaWriteJobContext::UpdateSkelBindingsWithExtent(
+    const UsdStagePtr&  stage,
+    const VtVec3fArray& bbox,
+    const UsdTimeCode&  timeSample)
+{
+    bool bStat = _skelBindingsProcessor->UpdateSkelRootsWithExtent(stage, bbox, timeSample);
+    return bStat;
```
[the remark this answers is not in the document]
ysiewappl: I can one-line this; usually I prefer having these split out just because it's easier to debug when stepping through code.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -36,14 +36,178 @@
...
+    MStatus    stat;
+    MFnDagNode fnNode(node, &stat);
+    CHECK_MSTATUS_AND_RETURN(stat, MString());
+    MString nodeName = fnNode.partialPathName(&stat);
```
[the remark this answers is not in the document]
ysiewappl: It's used to format the name of the final UsdSkelBlendShape name. The reason I prefer to get a unique name is that I've run into namespace collision issues several times before in varying scenarios and usually the solve there is just this; making sure that there's no chance for a collision.

```                    MString curTargetMeshNameMStr = mayaGetUniqueNameOfDAGNode(targetMesh);
                    assert(curTargetMeshNameMStr.length() != 0);
                    std::string curTargetMeshName
                        = TfMakeValidIdentifier(std::string(curTargetMeshNameMStr.asChar()));
                    SdfPath usdBlendShapePath
                        = primSchemaPath.AppendChild(TfToken(curTargetMeshName));
                    UsdSkelBlendShape usdBlendShape
                        = UsdSkelBlendShape::Define(this->GetUsdStage(), usdBlendShapePath);
                    if (!usdBlendShape) {
                        TF_RUNTIME_ERROR(
                            "Could not create blendshape primitive: <%s>",
                            usdBlendShapePath.GetText());
                        return MObject::kNullObj;
                    }```

File: lib/mayaUsd/fileio/utils/meshWriteUtils.cpp [hunk 21]
```diff
@@ -36,14 +36,178 @@
...
+
+            a.y = y < a.y ? y : a.y;
+            b.y = y > b.y ? y : b.y;
+
+            a.z = z < a.z ? z : a.z;
+            b.z = z > b.z ? z : b.z;
+        }
+    }
+
+    MBoundingBox result = MBoundingBox(MPoint(a), MPoint(b));
```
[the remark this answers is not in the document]
ysiewappl: Yes. However, that relies on calling expand() on each and every point (which implementation isn't known to me publicly). I was thinking that it was probably more efficient and simpler to compute it as implemented.

lib/mayaUsd/fileio/utils/meshWriteUtils.h
ysiewappl: As above: I think it potentially could be if you had multiple target meshes with the same names under different transform groups. In that case, we could have them saved with the same name in the resulting usd file, but it might make more sense to have them saved with the minimum unique name so that it's clear which blendshape target is which.

lib/usd/translators/meshWriter.cpp
ysiewappl: I actually did not see any status messages show up in Maya when using TF_RUNTIME_ERROR (they only show up in the terminal stdout), which is why I was making explicit calls to display the messages as well. If there's something I'm missing here like a config or a compile flag, I'd love to know so I don't have to essentially duplicate these calls.

lib/mayaUsd/fileio/utils/meshWriteUtils.h
ysiewappl: I was having trouble trying to decide where exactly to place my utlity functions. I'd be ok with moving the mesh-specific ones to that namespace, though generally I wasn't expecting to expose them to other translation units until they were actually needed.

lib/usd/translators/meshWriter_BlendShapes.cpp
ysiewappl: > Prefer typed, named constants over #defines.
I'd be happy, as above, to change this to something like

```static constexpr kMayaAttrNameWeight[] = "weight";```
if that fits better.

File: lib/usd/translators/meshWriter_BlendShapes.cpp (hunk 15 above)
Comment (ysiewappl) [reply lib/usd/translators/meshWriter_BlendShapes.cpp:64]: Yea, clang-format "adjusted" this. I can re-arrange the comment to see if that helps.

File: lib/mayaUsd/fileio/utils/meshWriteUtils.h [hunk 22]
```diff
@@ -29,11 +29,75 @@
...
+ * Similar to `mayaGetSkinClusterConnectedToMesh`, except that instead of finding a
+ * directly-connected skinCluster, it searches the DG upstream of the mesh for any
+ * other skinClusters as well.
+ *
+ * @param mesh              The mesh to start the search from.
+ * @param skinClusters      Storage for the result.
+ *
+ * @return                  `MStatus::kSuccess` if the operation completed successfully.
+ */
+MStatus mayaGetSkinClustersUpstreamOfMesh(const MObject& mesh, MObjectArray& skinClusters);
```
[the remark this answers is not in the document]
ysiewappl: That's very possible, since MSVC symbol visibility is private by default and the other way round on macOS/Linux. Unfortunately I don't have a Windows machine at work to cross-compile against. Though I wonder...shouldn't the CI system on here have caught this when building against MSVC?

lib/usd/translators/meshWriter_BlendShapes.cpp
ysiewappl: You're right, this is leftover from a previous implementation attempt I did. Will remove.

lib/mayaUsd/fileio/utils/meshWriteUtils.h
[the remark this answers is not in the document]
kxl-adsk: Agreed, I'm surprised by this one. Once a change is ready to merge, we always run a few more checks to cover configurations that are not yet handled by PF. Maybe we would see it there.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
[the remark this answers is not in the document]
kxl-adsk: The preference here is to leverage USD diagnostics over asserts - https://graphics.pixar.com/usd/docs/api/page_tf__diagnostic.html. You have much more control over them and depending on your choice, debugging can be opt-in only if necessary.

lib/mayaUsd/fileio/utils/meshWriteUtils.h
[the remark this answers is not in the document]
ysiewappl: Addressed in 6460336.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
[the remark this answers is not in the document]
ysiewappl: Addressed in 6460336. I'm using `TF_VERIFY` after looking through the available options, which seems to make the most sense?

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -36,14 +36,178 @@
...
+    MFnDagNode fnNode(node, &stat);
+    CHECK_MSTATUS_AND_RETURN(stat, MString());
+    MString nodeName = fnNode.partialPathName(&stat);
+    return nodeName;
+}
+
+MPlug mayaFindChildPlugWithName(const MPlug& parent, const MString& name)
+{
+    MPlug sentinel;
+    if (!parent.isCompound() || parent.isNull()) {
```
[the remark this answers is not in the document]
ysiewappl: Addressed in 6460336.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -36,14 +36,178 @@
...
+    }
+
+    return stat;
+}
+
+bool mayaSearchMIntArray(const int a, const MIntArray& array, unsigned int* idx)
+{
+    for (unsigned int i = 0; i < array.length(); ++i) {
+        if (array[i] == a) {
+            if (idx != NULL) {
```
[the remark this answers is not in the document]
ysiewappl: Addressed in 6460336.

lib/mayaUsd/fileio/utils/meshWriteUtils.h
ysiewappl: Addressed in 6460336.

lib/mayaUsd/fileio/utils/meshWriteUtils.h
ysiewappl: Addressed in 6460336.

lib/usd/translators/meshWriter.cpp
ysiewappl: This, and other asserts, have been addressed in 6460336.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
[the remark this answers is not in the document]
mattyjams: Cool, yeah, a `const char[]` sounds good to me.

For the anonymous namespace, I don't think it really makes a difference in this case. I tend to lean towards that over using `static` just in case at some point we end up wanting to add an internal class or something, in which case `static` doesn't work and we'd already have a namespace for it to go into.

The Cpp core guidelines have an item on using anonymous namespaces, but the justification is admittedly pretty thin:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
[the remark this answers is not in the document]
mattyjams: Sounds good, thanks for changing that!

One note though that `TF_VERIFY` doesn't stop execution, so you may want to use a pattern like:

```C++
if (!TF_VERIFY(...)) {
   return ...;
}
```

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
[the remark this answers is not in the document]
mattyjams: Yup, I definitely get the desire to have a guaranteed unique name. `UsdMayaUtil::GetMayaNodeName()` will actually get you that too. It just uses `MFnDagNode::fullPathName()` as opposed to `MFnDagNode::partialPathName()`, so you likely end up with something a bit longer than `partialPathName()` would give you.

lib/usd/translators/meshWriter.cpp
```diff
@@ -41,21 +43,108 @@
...
+        TF_VERIFY(curBlendShape.hasFn(MFn::kBlendShape));
+        MPlug outputGeomPlug = itDg.thisPlug();
+        TF_VERIFY(outputGeomPlug.isElement() == true);
+        unsigned int outputGeomPlugIdx = outputGeomPlug.logicalIndex();
+
+        // NOTE: (yliangsiew) Because we can have multiple output
+        // deformed meshes from a single blendshape deformer, we have
+        // to walk back up the graph using the connected index to find
+        // out what the _actual_ base mesh was.
+        if (intermediates == NULL) {
```
mattyjams, line 109: `nullptr`

lib/usd/translators/meshWriter.cpp
```diff
@@ -163,13 +283,171 @@ bool PxrUsdTranslators_MeshWriter::writeMeshAttrs(
...
+
+    // NOTE: (yliangsiew) Because we need to write out the _actual_ base mesh,
+    // not the deformed mesh as as result of blendshapes, if there is a blendshape
+    // in the deform stack here, we walk past it to the original shape instead.
+    // Also check if the mesh is a valid DG node (mesh geo subsets are
+    // kMeshData in cases where a single mesh has multiple face
+    // assignments to materials.)
+    if (exportArgs.exportBlendShapes && geomMeshObj.hasFn(MFn::kDependencyNode)) {
+        if (exportArgs.ignoreWarnings) {
+            geomMeshObj = mayaFindOrigMeshFromBlendShapeTarget(geomMeshObj, NULL);
```
mattyjams, line 295: `nullptr`

lib/mayaUsd/fileio/utils/meshWriteUtils.h
mattyjams: Yeah, namespacing in general is a little confusing and all over the place in general, the result of a few iterations of big code moves and merges. It'll take a focused clean-up pass to address a bunch of that later, but good to at least have these related functions grouped together for now, thanks!

lib/usd/translators/meshWriter.cpp
mattyjams: Ahh, sorry. Apparently we currently only install the diagnostic delegate I was talking about for the `pxrUsd` plugin:
https://github.com/Autodesk/maya-usd/blob/dev/plugin/pxr/maya/plugin/pxrUsd/plugin.cpp#L164

The `mayaUsdPlugin` does **not** install that delegate, which is probably why the `TF_RUNTIME_ERROR` doesn't get translated into a Maya error message.

lib/usd/translators/meshWriter_BlendShapes.cpp
mattyjams: Sure, `const char[]` sounds good to me!

lib/mayaUsd/utils/util.cpp
```diff
@@ -185,6 +185,17 @@ std::string UsdMayaUtil::GetMayaNodeName(const MObject& mayaNode)
     return nodeName.asChar();
 }
 
+MString UsdMayaUtil::GetUniqueNameOfDAGNode(const MObject& node)
+{
+    TF_VERIFY(!node.isNull());
+    TF_VERIFY(node.hasFn(MFn::kDagNode));
```
mattyjams, line 191: I'll just mention on this one instance here that `TF_VERIFY` won't prevent execution from continuing through the rest of this function, so you might want to do something like this:

```C++
if (!TF_VERIFY(!node.isNull() && node.hasFn(MFn::kDagNode)) {
    return MString();
}
...
```

mattyjams (approved):
Looks good to me now!

I noticed two stray `NULL`s leftover, and wondered whether in some of the `TF_VERIFY` cases you might want to early return if they fail.

I'll follow-up on getting the diagnostic delegate installed for the `mayaUsdPlugin` so we don't need the duplicate `TF_...` and `MGlobal::display...` calls.

Thanks @ysiewappl!

ysiewappl:
Just as a note, I am currently preparing a sizeable update to this PR that will introduce support for baked blendshape targets as well. After some internal discussion, we've decided that updating this current PR would be preferable to filing a new one. Please don't merge this quite yet (we are currently reviewing the update internally before submission)

kxl-adsk:
@ysiewappl there was quite a lot of discussion already in this PR. How about you close this one before opening a new one? This way we don't lose old conversations, but as well we don't have to re-read what was already resolved.

ysiewappl:
> @ysiewappl there was quite a lot of discussion already in this PR. How about you close this one before opening a new one? This way we don't lose old conversations, but as well we don't have to re-read what was already resolved.
Sounds good to me. Once we've approved everything internally, I will close this PR and submit the same branch again under a different PR.

lib/usd/translators/meshWriter.cpp
mattyjams: Just posted #1003 which I think should address this and let us remove the extra `MGlobal::...` calls.

ysiewappl:
Closing this PR and opening a new one.

PR #954, opened 2020-12-01
Add support for UV transform value exports

dgovil:
This commit adds support for UV transform exports.
We have not yet implemented the import functionality.


lib/usd/translators/shading/usdFileTextureWriter.cpp
```diff
@@ -452,6 +472,178 @@ void PxrUsdTranslators_FileTextureWriter::Write(const UsdTimeCode& usdTime)
...
+    if (UsdMayaUtil::IsAuthored(offsetPlug)) {
+        for (size_t i = 0u; i < GfVec2f::dimension; ++i) {
+            translationValue[i] = offsetPlug.child(i).asFloat(&status);
+            if (status != MS::kSuccess) {
+                return;
+            }
+        }
+
+        if (translationValue != translationFallbackValue) {
+            transformationsAreAuthored |= true;
```
JGamache-autodesk, line 521: Will work, but this is a bitwise-or. Why not use `=` here?

JGamache-autodesk (approved):

lib/usd/translators/shading/usdFileTextureWriter.cpp
dgovil: Good catch. Think it was just vestigial but benign. Changed it to a simple `=`

dgovil:
Apologies, the test is failing due to default maya plugins again. I'll fix up the Maya file and  and try and institute better validation checks on our side to make sure we aren't committing files with other plugins. 

```
# Traceback (most recent call last):
    #   File "<maya console>", line 1, in <module>
    # RuntimeError: Plug-in, "mtoa", was not found on MAYA_PLUG_IN_PATH.
```

dgovil:
Okay I've removed mtoa and any `.ai_*` setattrs. That should satisfy the CI setup I hope.
I'm not sure I can kick off a new CI by myself, but hopefully should work for whenever you do it.

kxl-adsk:
Started one. You will as well notice that we have clang-format check as well.

dgovil:
> Started one. You will as well notice that we have clang-format check as well.

Thanks! and yes, that format check will be handy (and I'm glad to see this PR passed it)

dgovil:
Looks like the CI failed getting results back

`fatal: could not read Username for 'https://github.com': No such device or address`

kxl-adsk:
That’s new. I will start it again.

kxl-adsk:
@dgovil There was an incident with GitHub yesterday - could be related - https://www.githubstatus.com/incidents/sxthxvdlljk4

New PF passed.

dgovil:
Ah makes sense. Thanks for looking into it too!

kxl-adsk:
Credit goes to @smithw-adsk

lib/usd/translators/shading/usdFileTextureWriter.cpp
```diff
@@ -452,6 +472,178 @@ void PxrUsdTranslators_FileTextureWriter::Write(const UsdTimeCode& usdTime)
...
+        if (status != MS::kSuccess) {
+            return;
+        }
+
+        if (rotationValue != rotationFallbackValue) {
+            transformationsAreAuthored |= true;
+        }
+
+        // Convert rotation from radians to degrees
+        rotationValue = ((rotationValue / M_PI) * 180.0f);
```
mattyjams, line 545: `pxr/base/gf/math.h` Provides function for this:

```C++
rotationValue = GfRadiansToDegrees(rotationValue)
```

lib/usd/translators/shading/usdFileTextureWriter.cpp
```diff
@@ -452,6 +472,178 @@ void PxrUsdTranslators_FileTextureWriter::Write(const UsdTimeCode& usdTime)
...
+    GfMatrix4f translateXform
+        = GfMatrix4f().SetTranslate(GfVec3f(translationValue[0], translationValue[1], 0));
+    GfRotation rotation = GfRotation(GfVec3f::ZAxis(), rotationValue);
+    GfMatrix4f rotationXform = GfMatrix4f().SetRotate(rotation);
+    GfVec3f    scale;
+    if (fabs(scaleValue[0]) <= std::numeric_limits<float>::epsilon()
+        || fabs(scaleValue[1]) <= std::numeric_limits<float>::epsilon()) {
+        TF_WARN(
+            "At least one of the components of RepeatUV are set to zero.  To avoid divide by zero "
+            "exceptions, these values are changed to the smallest finite float greater than zero.");
```
mattyjams, line 613: Might be nice to include the full name of the problematic node in this message to help track it down. You can get that with this:

```C++
UsdMayaUtil::GetMayaNodeName(GetMayaObject())
```

test/lib/usd/translators/testUsdExportUVTransforms.py
```diff
@@ -0,0 +1,97 @@
...
+        transform = findTransformPrim(prim)
+        if transform:
+            return transform
+
+
+class testUsdExportUVTransforms(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.temp_dir = fixturesUtils.setUpClass(__file__, initializeStandalone=False)
```
mattyjams, line 47: I think we actually **do** want the `standalone.initialize()` call that this would otherwise do, so I think remove the `initializeStandlone=False` here?

test/lib/usd/translators/testUsdExportUVTransforms.py
```diff
@@ -0,0 +1,97 @@
...
+
+        usd_file_path = os.path.join(cls.temp_dir, "UsdExportUVTransforms.usda")
+        cmds.mayaUSDExport(mergeTransformAndShape=True, file=usd_file_path,
+                           shadingMode='useRegistry')
+
+        cls.stage = Usd.Stage.Open(usd_file_path)
+
+    @classmethod
+    def tearDownClass(cls):
+        return
```
mattyjams, line 59: And remove this `return`?

mattyjams (commented):
Mostly looks good to me! Just two small suggestions for the new code.

In the test though, I'm a little unclear about the use of `maya.standalone`. I would think that we **do** want to initialize with the `fixturesUtils.setUpClass()` call, and then `uninitialize()` in `tearDownClass()`?

test/lib/usd/translators/testUsdExportUVTransforms.py
dgovil: Ah my bad. I write the tests interactively, so had disabled that and forgotten to remove the statements. Same for the return below. I'll remove those.

lib/usd/translators/shading/usdFileTextureWriter.cpp
dgovil: Thanks, swapped that out.

lib/usd/translators/shading/usdFileTextureWriter.cpp
dgovil: Good call, I've swapped that out.

dgovil:
@mattyjams Thanks for the suggestions. Addressed that and pushed a new commit with all the changes.

mattyjams (approved):
Looks great now! Thanks @dgovil!

kxl-adsk:
@dgovil You PR is almost ready to go in, just please address the clang-format-lint check error.

dgovil:
@kxl-adsk Done.

PR #955, opened 2020-12-01
MAYA-108270 Improve draw performance for Rprims without extent

HdC-adsk:
If a Rprim has no extent, we will assign a null bounding box to its render item and
Maya will compute the bounding box from the position data. With the bounding box
the render item will be eligible for consolidation.


williamkrick (approved):

kxl-adsk (approved):

HdC-adsk:
@kxl-adsk This PR can be merged.

PR #957, opened 2020-12-01
Fix broken test when run from parent-project.

marsupial:
Is there a chance you can add a sub-project build to your internal test-runners so this doesn't keep breaking ?
```
cmake_minimum_required(VERSION 3.13)
project(SubProjectBuild)
add_subdirectory(autodesk-maya-usd)
```

### DIFF
--- a/test/lib/testMayaUsdCreateStageCommands.py
+++ b/test/lib/testMayaUsdCreateStageCommands.py
@@ -21,6 +21,7 @@ import platform
 import unittest
 from maya import cmds
 import maya.mel as mel
+import mayaUtils
 
 class MayaUsdCreateStageCommandsTestCase(unittest.TestCase):
     """Test the MEL commands that are used to create a USD stage."""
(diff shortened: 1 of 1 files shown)


test/lib/testMayaUsdCreateStageCommands.py
```diff
@@ -59,7 +57,7 @@ def testCreateStageFromFile(self):
         # we can call what it does once the file is choose.
         # Note: on ballFilePath we replace \ with / to stop the \ as
         #       being interpreted.
-        ballFilePath = os.path.normpath(self.getTestScene('ballset', 'StandaloneScene', 'top_layer.usda')).replace('\\', '/')
+        ballFilePath = os.path.normpath(mayaUtils.getTestScene('ballset', 'StandaloneScene', 'top_layer.usda')).replace('\\', '/')
```
seando-adsk, line 60: Sorry, but what does this fix? I purposely copied the getTestScene() into this test to make it completely self contained. It doesn't rely or need the mayaUtils (which also imports ufe and ufeUtils). This test needs to run against previous versions of Maya.

marsupial: It fixes building/testing from a parent-project, where _maya-usd_ is included as a sub-project.
Lines 127-175 of _mayaUtils.py_ have no dependencies on **ufe**, so could easily be moved to _testUtils.py_

seando-adsk: Okay I understand now. I agree with moving those lines to a new _testUtils.py_ and then using that here instead of _mayaUtils_.

marsupial:
Superseded by #1034

PR #959, opened 2020-12-02
More code clean up in UsdTransform3dMayaXformStack

HamedSabri-adsk:
- Adjust UndoBlock to capture the opFunc functor that sets/adds to the transform op order.
- Clean up old logics around recreateOp, _prevOpValue, _attrName.


ppt-adsk (approved):
Thanks Hamed!

PR #961, opened 2020-12-02
remove dependencies on old Hydra texture system from hdMaya for USD releases after 20.11

mattyjams:
Our Hydra team is working on removing parts of the old Hydra texture system for the next core USD release.

With Luma's recent work, it looks like all of the necessary support for the newer material network-based system is already in place, so the changes here mostly just pre-process the old code away for USD releases after 20.11.

### DIFF
--- a/lib/usd/hdMaya/adapters/materialAdapter.cpp
+++ b/lib/usd/hdMaya/adapters/materialAdapter.cpp
@@ -21,26 +21,29 @@
 #include <hdMaya/adapters/tokens.h>
 #include <hdMaya/utils.h>
 
-#include <pxr/base/tf/fileUtils.h>
-#include <pxr/imaging/glf/contextCaps.h>
-#include <pxr/imaging/glf/textureRegistry.h>
-#include <pxr/imaging/glf/udimTexture.h>
-#include <pxr/imaging/hd/instanceRegistry.h>
 #include <pxr/imaging/hd/material.h>
-#include <pxr/imaging/hd/resourceRegistry.h>
-#include <pxr/imaging/hdSt/resourceRegistry.h>
-#include <pxr/imaging/hdSt/textureResource.h>
-#include <pxr/imaging/hdSt/textureResourceHandle.h>
 #include <pxr/imaging/hio/glslfx.h>
 #include <pxr/usd/sdf/types.h>
 #include <pxr/usd/sdr/registry.h>
-#include <pxr/usdImaging/usdImaging/textureUtils.h>
 #include <pxr/usdImaging/usdImaging/tokens.h>
 #include <pxr/usdImaging/usdImagingGL/package.h>
 
 #include <maya/MNodeMessage.h>
 #include <maya/MPlug.h>
 #include <maya/MPlugArray.h>
+
+#if USD_VERSION_NUM < 2011
+#include <pxr/base/tf/fileUtils.h>
+#include <pxr/imaging/glf/contextCaps.h>
+#include <pxr/imaging/glf/textureRegistry.h>
+#include <pxr/imaging/glf/udimTexture.h>
+#include <pxr/imaging/hd/instanceRegistry.h>
+#include <pxr/imaging/hd/resourceRegistry.h>
+#include <pxr/imaging/hdSt/resourceRegistry.h>
+#include <pxr/imaging/hdSt/textureResource.h>
+#include <pxr/imaging/hdSt/textureResourceHandle.h>
+#include <pxr/usdImaging/usdImaging/textureUtils.h>
+#endif // USD_VERSION_NUM < 2011
 
 PXR_NAMESPACE_OPEN_SCOPE
 
@@ -105,7 +108,19 @@ void HdMayaMaterialAdapter::Populate()
     _isPopulated = true;
 }
 
-#if USD_VERSION_NUM <= 1911
+#if USD_VERSION_NUM > 1911 && USD_VERSION_NUM < 2011
+
+HdTextureResource::ID HdMayaMaterialAdapter::GetTextureResourceID(const TfToken& paramName)
+{
+    return {};
+}
+
+HdTextureResourceSharedPtr HdMayaMaterialAdapter::GetTextureResource(const SdfPath& textureShaderId)
+{
+    return {};
+}
+
+#elif USD_VERSION_NUM <= 1911
 
 // We can't store the shader here explicitly, since it causes a deadlock
 // due to library dependencies.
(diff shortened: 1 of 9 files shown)


mattyjams:
@pmolodo: Could I bug you to have a look at this one when you get a chance? Just want to make sure we haven't missed anything here, since we don't make heavy use of mtoh ourselves.

Thanks!

lib/usd/hdMaya/adapters/materialAdapter.cpp
```diff
@@ -105,7 +108,19 @@ void HdMayaMaterialAdapter::Populate()
...
+{
+    return {};
+}
+
+HdTextureResourceSharedPtr HdMayaMaterialAdapter::GetTextureResource(const SdfPath& textureShaderId)
+{
+    return {};
+}
+
+#elif USD_VERSION_NUM <= 1911
```
pmolodo, line 123: this elif branch will never be entered - anything <= 1911 is by definition also <= 2011, so the first branch would be entered.

Either the order needs to be reversed (first check <= 1911, then elif <= 2011), or the first if check needs to be changed to:

```
#if USD_VERSION_NUM <= 2011 && USD_VERISON_NUM > 1911
```

lib/usd/hdMaya/adapters/materialAdapter.cpp
```diff
@@ -317,7 +325,29 @@ class HdMayaShadingEngineAdapter : public HdMayaMaterialAdapter
...
+            }
+            return GetFileTextureResource(
+                *mObjPtr,
+                GetFileTexturePath(MFnDependencyNode(*mObjPtr)),
+                GetDelegate()->GetParams().textureMemoryPerTexture);
+        }
+        return {};
+    }
+
+#elif USD_VERSION_NUM <= 1911
```
pmolodo, line 350: this elif branch will never be entered - see above note

lib/usd/hdMaya/adapters/materialAdapter.h
```diff
@@ -43,7 +43,14 @@ class HdMayaMaterialAdapter : public HdMayaAdapter
...
 
-#if USD_VERSION_NUM <= 1911
+#if USD_VERSION_NUM <= 2011
+
+    HDMAYA_API
+    virtual HdTextureResourceSharedPtr GetTextureResource(const SdfPath& textureShaderId);
+    HDMAYA_API
+    virtual HdTextureResource::ID GetTextureResourceID(const TfToken& paramName);
+
+#elif USD_VERSION_NUM <= 1911
```
pmolodo, line 53: this elif branch will never be entered - see above note

pmolodo (changes requested):
So, I haven't looked at the texture changes coming in 21.02 yet - is it the case that texture resource handling is handled more "automatically", and none of these overrides are required at all in order to get texture support?  Or are these changes effectively just disabling texture support for now (until we can add it back in a way that's compatible with 21.02, later?)

lib/usd/hdMaya/adapters/materialAdapter.cpp
mattyjams: Ugh, exactly right, great catch. I'd protest that these weren't my changes, but I did review them and I completely glossed over that. :/

I'll fix these up!

mattyjams:
> So, I haven't looked at the texture changes coming in 21.02 yet - is it the case that texture resource handling is handled more "automatically", and none of these overrides are required at all in order to get texture support? Or are these changes effectively just disabling texture support for now (until we can add it back in a way that's compatible with 21.02, later?)

I think it should be the former. I don't actually expect that there will be major changes needed to **add** support for 21.02, but just that vestiges of the old texture system are being removed, so we want to drop code in `hdMaya` that used those. The "new" system I was alluding to is actually in 20.11 now, where the scene delegate constructs an `HdMaterialNetworkMap` and returns that from `GetMaterialResource()`:

https://github.com/PixarAnimationStudios/USD/blob/release/pxr/imaging/hd/sceneDelegate.h#L598

It looks like all of that is already in place in `hdMaya` now. The actual handling of the texture resources should indeed now be handled auto-magically within Hydra itself based on the information in the material network.

So my hope is that texture support should happily continue working, but I was hoping you could help verify that. :)

lib/usd/hdMaya/adapters/materialAdapter.cpp
mattyjams: Addressed by d9ea7b2abc0f27f79638e3602c129a9812e73fd1.

lib/usd/hdMaya/adapters/materialAdapter.cpp
mattyjams: Addressed by d9ea7b2abc0f27f79638e3602c129a9812e73fd1.

lib/usd/hdMaya/adapters/materialAdapter.h
mattyjams: Addressed by d9ea7b2abc0f27f79638e3602c129a9812e73fd1.

pmolodo:
Oh - if the changes are in 20.11, then why are most of the checks for <= 20.11? Shouldn't they be < 20.11?

mattyjams:
> Oh - if the changes are in 20.11, then why are most of the checks for <= 20.11? Shouldn't they be < 20.11?

Good question. I don't think there's any particular reason why the deprecated API was left in for `== 20.11` other than that the Hydra team is midway through removing the old API and it made for a somewhat cleaner break. Would you prefer that I change it to `< 20.11` instead?

In any case, I rebased these changes and fixed up the `#if`/`#elif` stuff you noted, so thanks again for that! One additional change I added drops out the now-unused debug codes that go along with the removed `GetTextureResource()` and `GetTextureResourceID()`.

pmolodo:
If there's no benefit to supporting both APIs, then I think we should drop the old one in 20.11.  That will make the code read better, mean the old stuff will get dropped sooner, and also make it easier for me to test (since I already have a pre-compiled 20.11 at home, but haven't compiled the latest tip of dev yet!)

mattyjams:
> If there's no benefit to supporting both APIs, then I think we should drop the old one in 20.11. That will make the code read better, mean the old stuff will get dropped sooner, and also make it easier for me to test (since I already have a pre-compiled 20.11, but haven't compiled the latest tip of dev yet!)

Cool, works for me! I just rebased and pushed that change.

Let me know how your testing goes! Thanks!

pmolodo:
Ok, I built this and ran some basic tests, and it seems to be working fine!

pmolodo (approved):
Looks good!

mattyjams:
@pmolodo: Great, glad to hear it! Thank you, sir!

PR #964, opened 2020-12-03
MAYA-108214 - Attributes categorized by schema

seando-adsk:
#### MAYA-108214 - Attributes categorized by schema
* New "usdschemabase" template which is at top level of the schema inheritance scheme so all prim types will use it. All the node type templates are deprecated and will be removed in the future.

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
@@ -1,0 +1,124 @@
+import fnmatch
+import ufe
+import mayaUsd.ufe as mu
+import maya.internal.common.ufe_ae.template as ufeAeTemplate
+
+# We manually import all the classes which have a 'GetSchemaAttributeNames'
+# method so we have access to it and the 'pythonClass' method.
+from pxr import Usd, UsdGeom, UsdLux, UsdRender, UsdRi, UsdShade, UsdSkel, UsdUI, UsdVol
+
+# SchemaBase template class for categorization of the attributes.
+class AETemplate(ufeAeTemplate.Template):
+    '''
+    This system of schema inherits groups attributes per schema helping the user
+    learn how attributes are stored.
+    '''
+    def __init__(self, ufeSceneItem):
+        # Get the UFE Attributes interface for this scene item.
+        self.attrS = ufe.Attributes.attributes(ufeSceneItem)
+
+        # And finally call our base class init which will in turn
+        # build the UI.
+        super(AETemplate, self).__init__(ufeSceneItem)
+
+    def createSection(self, layoutName, attrList, collapse=False):
+        # We create the section named "layoutName" if at least one
+        # of the attributes from the input list exists.
+        for attr in attrList:
+            if self.attrS.hasAttribute(attr):
+                with ufeAeTemplate.Layout(self, layoutName, collapse):
+                    # Note: addControls will silently skip any attributes
+                    #       which do not exist.
+                    self.addControls(attrList)
+                return
+
+    def sectionNameFromSchema(self, schemaTypeName):
+        '''Return a section name from the input schema type name. This section
+        name is a pretty name used in the UI.'''
+
+        # List of special rules for adjusting the base schema names.
+        prefixToAdjust = {
+            'UsdAbc' : '',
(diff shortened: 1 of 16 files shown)


lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -0,0 +1,124 @@
+import fnmatch
+import ufe
+import mayaUsd.ufe as mu
```
ppt-adsk, line 3: mu?

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -0,0 +1,124 @@
...
+            schemaTypeName = schemaType.typeName
+            sectionName = self.sectionNameFromSchema(schemaTypeName)
+            if schemaType.pythonClass:
+                attrsToAdd = schemaType.pythonClass.GetSchemaAttributeNames(False)
+
+                # We have a special case when building the Xformable section.
+                if schemaTypeName == 'UsdGeomXformable':
+                    self.createTransformAttributesSection(ufeSceneItem, sectionName, attrsToAdd)
+                else:
+                    self.createSection(sectionName, attrsToAdd)
```
seando-adsk, line 124: This method completely changed from ae/base/ae_template.py. We no longer have hard-coded sections and attributes. Instead using the ancestor node types I get all the schemas and for each one get the attributes.

The only exception is the Xformable, where we still display the transform attributes by xformoporder.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -0,0 +1,124 @@
...
+                    catName = catName + str(' ') + str(schemaTypeName[i])
+                else:
+                    catName = catName + str(schemaTypeName[i])
+            elif schemaTypeName[i] == '_':
+                continue
+            else:
+                catName = catName + str(schemaTypeName[i])
+
+        # Make each word start with an uppercase.
+        catName = catName.title()
```
seando-adsk, line 81: New helper to create the section title from the schema name. There are a bunch of rules (see JIRA item) for how to parse and format the string.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -0,0 +1,124 @@
...
+            'UsdUI' : '',
+            'UsdUtils' : '',
+            'UsdVol' : ''
+        }
+        for p, r in prefixToAdjust.items():
+            if schemaTypeName.startswith(p):
+                schemaTypeName = schemaTypeName.replace(p, r, 1)
+                break
+
+        # Put a space in the name when proceeded with a capital letter.
```
ppt-adsk, line 64: Not quite sure here, but think you mean either "preceded by" or "followed by".

ppt-adsk (approved):
Just this minor comment detail, but otherwise LGTM.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
seando-adsk: Does that matter? Isn't this just a name to use instead of "mayaUsd.ufe.xxx"

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
seando-adsk: Ah yes, typo

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
ppt-adsk: You are absolutely correct in saying that there is no problem on executing the above code and getting correct results.  Just wondering about the very terse name...  m can be Maya or maya-usd, u can be USD or UFE...  Kinda hard to find a more descriptive name that's shorter than mayaUsd.ufe, I guess...  You can leave it as is if you like.

seando-adsk: Your right. I only used it once, so I've just removed that

ppt-adsk (approved):

seando-adsk:
@kxl-adsk If the PF passes, please merge whenever.

PR #965, opened 2020-12-03
MAYA-108549: Fix undo crash when switching between target layers in a stage

HamedSabri-adsk:
This PR fixes the crash when switching between target layers in a stage.

### DIFF
--- a/lib/mayaUsd/ufe/StagesSubject.h
+++ b/lib/mayaUsd/ufe/StagesSubject.h
@@ -73,7 +73,7 @@ private:
     //! Call the stageChanged() methods on stage observers.
     void stageChanged(UsdNotice::ObjectsChanged const& notice, UsdStageWeakPtr const& sender);
 
-#if UFE_PREVIEW_VERSION_NUM >= 2029
+#if UFE_PREVIEW_VERSION_NUM >= 2025
     //! Call the stageEditTargetChanged() methods on stage observers.
     void stageEditTargetChanged(
         UsdNotice::StageEditTargetChanged const& notice,
@@ -88,7 +88,7 @@ private:
     void onStageInvalidate(const MayaUsdProxyStageInvalidateNotice& notice);
 
     // Array of Notice::Key for registered listener
-#if UFE_PREVIEW_VERSION_NUM >= 2029
+#if UFE_PREVIEW_VERSION_NUM >= 2025
     using NoticeKeys = std::array<TfNotice::Key, 2>;
 #else
     using NoticeKeys = std::array<TfNotice::Key, 1>;
--- a/lib/mayaUsd/undo/UsdUndoManager.cpp
+++ b/lib/mayaUsd/undo/UsdUndoManager.cpp
@@ -29,16 +29,9 @@ UsdUndoManager& UsdUndoManager::instance()
     return undoManager;
 }
 
-void UsdUndoManager::trackStatesOnNewStage(const SdfLayerHandle& layer)
+void UsdUndoManager::trackLayerStates(const SdfLayerHandle& layer)
 {
-    _layerStateDelegate = TfCreateRefPtr(new UsdUndoStateDelegate);
-
-    layer->SetStateDelegate(_layerStateDelegate);
-}
-
-void UsdUndoManager::trackStatesOnEditTargetChange(const SdfLayerHandle& layer) 
-{
-    layer->SetStateDelegate(_layerStateDelegate);
+    layer->SetStateDelegate(UsdUndoStateDelegate::New());
 }
 
 void UsdUndoManager::addInverse(InvertFunc func)
--- a/lib/mayaUsd/undo/UsdUndoManager.cpp
+++ b/lib/mayaUsd/undo/UsdUndoManager.cpp
@@ -29,9 +29,16 @@ UsdUndoManager& UsdUndoManager::instance()
     return undoManager;
 }
 
-void UsdUndoManager::trackLayerStates(const SdfLayerHandle& layer)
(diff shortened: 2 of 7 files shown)


lib/mayaUsd/undo/UsdUndoManager.cpp
```diff
@@ -29,9 +29,16 @@ UsdUndoManager& UsdUndoManager::instance()
...
-void UsdUndoManager::trackLayerStates(const SdfLayerHandle& layer)
+void UsdUndoManager::trackStatesOnNewStage(const SdfLayerHandle& layer)
 {
-    layer->SetStateDelegate(UsdUndoStateDelegate::New());
+    _layerStateDelegate = TfCreateRefPtr(new UsdUndoStateDelegate);
+
+    layer->SetStateDelegate(_layerStateDelegate);
+}
+
+void UsdUndoManager::trackStatesOnEditTargetChange(const SdfLayerHandle& layer)
```
ppt-adsk, line 39: I'm a bit perplexed here.  If we set a new state delegate on layer creation, then why do we have to set it on edit target change?  Is it because of initialization order and the fact that certain layers may have been created before the UsdUndoManager starts tracking changes?

lib/mayaUsd/undo/UsdUndoManager.cpp
```diff
@@ -29,9 +29,16 @@ UsdUndoManager& UsdUndoManager::instance()
     return undoManager;
 }
 
-void UsdUndoManager::trackLayerStates(const SdfLayerHandle& layer)
+void UsdUndoManager::trackStatesOnNewStage(const SdfLayerHandle& layer)
 {
-    layer->SetStateDelegate(UsdUndoStateDelegate::New());
+    _layerStateDelegate = TfCreateRefPtr(new UsdUndoStateDelegate);
```
ppt-adsk, line 34: Why do we save this as a data member?  Since the UsdUndoManager is a singleton, doesn't that mean that more than one layer can share the same UsdUndoStateDelegate?

Thinking out loud here, could it be that the UsdUndoManager should track a set of layers, and when it sees one that isn't in its set, whether than be through a layer creation notification or an edit target change notification, it will give the unknown layer a newly-created UsdUndoStateDelegate, then will add the layer to its set of known layers?

lib/mayaUsd/undo/UsdUndoManager.cpp
HamedSabri-adsk: Every time we change the target edit layer, UsdUndoStateDelegate::_OnSetLayer is invoked where it
needs to set the _layer for tracking changes. 

We do need to call layer->SetStateDelegate() when we switch between layers otherwise how would the layer know what state delegate needs to use?

Notice that in trackStatesOnEditTargetChange, I am only passing the _layerStateDelegate object that was created on stage set. 

More than one layer needs to share the same StateDelegate .
```
void UsdUndoManager::trackStatesOnEditTargetChange(const SdfLayerHandle& layer) 
{
    layer->SetStateDelegate(_layerStateDelegate);
}
```

lib/mayaUsd/undo/UsdUndoManager.cpp
HamedSabri-adsk: > Why do we save this as a data member? 

The main reason I need to save _layerStateDelegate as a data member is for undo/redo purposes between layers itself.


> Since the UsdUndoManager is a singleton, doesn't that mean that more than one layer can share the same UsdUndoStateDelegate?

Exactly, more than one layer can share the same UsdUndoStateDelegate in a same stage.

> it will give the unknown layer a newly-created UsdUndoStateDelegate, then will add the layer to its set of known layers?

Let's have a chat around this. I would like to understand this approach better.

lib/mayaUsd/ufe/StagesSubject.h
```diff
@@ -88,7 +88,7 @@ class MAYAUSD_CORE_PUBLIC StagesSubject : public TfWeakBase
     void onStageInvalidate(const MayaUsdProxyStageInvalidateNotice& notice);
 
     // Array of Notice::Key for registered listener
-#if UFE_PREVIEW_VERSION_NUM >= 2029
+#if UFE_PREVIEW_VERSION_NUM >= 2025
```
HamedSabri-adsk, line 91: @ppt-adsk  Based on our internal discussion,  lowering the version to cover existing shipped PR.

lib/mayaUsd/undo/UsdUndoManager.cpp
```diff
@@ -31,7 +31,13 @@ UsdUndoManager& UsdUndoManager::instance()
...
 void UsdUndoManager::trackLayerStates(const SdfLayerHandle& layer)
 {
-    layer->SetStateDelegate(UsdUndoStateDelegate::New());
+    // Check if the layer has already been given a UsdUndoStateDelegate
+    // if the cast fails that means we need to set a new one.
+    auto usdUndoStateDelegatePtr
+        = TfDynamic_cast<UsdUndoStateDelegatePtr>(layer->GetStateDelegate());
+    if (!usdUndoStateDelegatePtr) {
+        layer->SetStateDelegate(UsdUndoStateDelegate::New());
+    }
```
HamedSabri-adsk, line 40: @ppt-adsk Thanks for great feedback.  Using a dynamic_cast to determine if a layer is already registered with our UsdUndoStateDelegate. If the cast fails, that means it needs to be set with a new one. The UsdUndoStateDelegate is stateless and doesn't need to be shared among layers.

ppt-adsk (approved):

kxl-adsk (approved):
LGTM

PR #966, opened 2020-12-03
MAYA-106937 Prototype support for GPU Normals and OSD using glsl. 

williamkrick:
The code is disabled by default.

The GPU evaluation prototype is two separate parts, the normal calculation code and the OSD
code.

The normal calculation code is enabled by setting HDVP2_USE_GPU_NORMAL_COMPUTATION=1 at runtime.
The normal calculation code is close to being stable enough for general use, but hasn't had
enough polish to enable by default.

The OSD code requires the normal calculation code to be enabled to use. OSD is enabled
by compiling with HDVP2_ENABLE_GPU_OSD. The OSD code is much less stable then the normals
calculation code and comes with a number of huge.

OSD Limitations:
    * No OSD adaptive support
    * scenes with animation behave poorly
    * selection in the viewport is very slow
    * toggling VP2 consolidation world off and on will cause crashes
    * some objects draw with incorrect indexing

### DIFF
--- a/cmake/modules/FindMaya.cmake
+++ b/cmake/modules/FindMaya.cmake
@@ -207,7 +207,8 @@ foreach(MAYA_LIB
     IMFbase
     tbb
     cg
-    cgGL)
+    cgGL
+    clew)
 
     find_library(MAYA_${MAYA_LIB}_LIBRARY
             ${MAYA_LIB}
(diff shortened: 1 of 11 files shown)


HdC-adsk (approved):

kxl-adsk (changes requested):
This change doesn't compile with USD 20.02. I'm getting the following error:
```
maya-usd_pr\lib\mayaUsd\render\vp2RenderDelegate\meshViewportCompute.cpp(250): error C2065: 'PxOsdOpenSubdivTokens': undeclared identifier
```

kxl-adsk (approved):

PR #969, opened 2020-12-04
MAYA-108545: Prevent the crash when authoring edits to instance proxies.

HamedSabri-adsk:
According to USD docs, editing scene description via instance proxies and their properties is not allowed.

### DIFF
--- a/lib/mayaUsd/ufe/UsdTransform3dHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdTransform3dHandler.cpp
@@ -17,6 +17,8 @@
 
 #include <mayaUsd/ufe/UsdSceneItem.h>
 
+#include <maya/MGlobal.h>
+
 namespace MAYAUSD_NS_DEF {
 namespace ufe {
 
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/UsdTransform3dHandler.cpp
```diff
@@ -44,6 +46,14 @@ Ufe::Transform3d::Ptr UsdTransform3dHandler::transform3d(const Ufe::SceneItem::P
...
 #endif
 
+    // According to USD docs, editing scene description via instance proxies and their properties is
+    // not allowed.
+    // https://graphics.pixar.com/usd/docs/api/_usd__page__scenegraph_instancing.html#Usd_ScenegraphInstancing_InstanceProxies
+    if (usdItem->prim().IsInstanceProxy()) {
+        MGlobal::displayError("Authoring to an instance proxy is not allowed.");
+        return nullptr;
+    }
+
```
HamedSabri-adsk, line 56: This code is duplicated in 2 places. Would be nice to refactor it. Perhaps as part of "lib/mayaUsd/ufe/Utils"?

ppt-adsk: Agreed.  Perhaps when we collect more illegal cases, we can do the refactoring.

ppt-adsk (approved):

PR #970, opened 2020-12-04
Convert USD timeSamples to Maya FPS

dgovil:
Hi,
This PR adds support for conversion on import from the USD's timeCodesPerUnits to the Maya scenes uiUnit values.

Since USD technically supports arbitrary fps, I do not try and map from their FPS values to one of MTime units, and instead calculate a TimeSampleMultiplier value.

I believe I've caught all the uses of MTimeArray and MTime, and my tests cover most object types that I believe are supported, with the exception of NurbsCurves since it doesn't seem like the importer actually imports them? At least I couldn't coerce an import.


kxl-adsk:
@dgovil Apologies for taking long to get to your pull requests. We have some internal deadlines that affected our ability to get to this PR.

dgovil:
@kxl-adsk best of luck with that, that's understandable.

kxl-adsk:
@dgovil I was looking for someone who could help us review this change. From one side, it looks pretty straightforward...but remembering all the trickery with supporting different time codes, like 29.976, I didn't want to overlook something that would cause us to affect precision or roundtripping of data.

I asked @langlor-autodesk to take a look and at the very least ask questions that will confirm that we are not ignoring anything important here.

dgovil:
Thanks! yes, i would appreciate the double checking on precision losses. Should be easy to modify the tests accordingly too

lib/mayaUsd/fileio/translators/translatorCamera.cpp
```diff
@@ -213,7 +216,12 @@ static bool _TranslateAnimatedUsdAttributeToPlug(
     MTimeArray   timeArray;
     MDoubleArray valueArray;
     if (!_GetTimeAndValueArrayForUsdAttribute(
-            usdAttr, args.GetTimeInterval(), &timeArray, &valueArray, convertToUnit)) {
+            usdAttr,
+            args.GetTimeInterval(),
+            &timeArray,
+            &valueArray,
+            convertToUnit,
+            context->GetTimeSampleMultiplier())) {
```
mattyjams, line 224: I think we want to allow `context` to be `nullptr` in these translator code paths, so maybe something like this:
```C++
(context != nullptr) ? context->GetTimeSampleMultiplier() : 1.0)) {
```

More on this below.

lib/mayaUsd/fileio/translators/translatorCamera.cpp
```diff
@@ -239,7 +247,12 @@ static bool _TranslateAnimatedUsdAttributeToPlugs(
     MDoubleArray valueArray1;
     MDoubleArray valueArray2;
     if (!_GetTimeAndValueArraysForUsdAttribute(
-            usdAttr, args.GetTimeInterval(), &timeArray, &valueArray1, &valueArray2)) {
+            usdAttr,
+            args.GetTimeInterval(),
+            &timeArray,
+            &valueArray1,
+            &valueArray2,
+            context->GetTimeSampleMultiplier())) {
```
mattyjams, line 255: Similar here:
```C++
(context != nullptr) ? context->GetTimeSampleMultiplier() : 1.0)) {
```

lib/mayaUsd/fileio/translators/translatorCurves.cpp
```diff
@@ -242,10 +242,11 @@ bool UsdMayaTranslatorCurves::Create(
         MFnAnimCurve animFn;
 
         // Construct the time array to be used for all the keys
-        MTimeArray timeArray;
-        timeArray.setLength(numTimeSamples);
+        MTime::Unit timeUnit = MTime::uiUnit();
+        double      timeSampleMultiplier = context->GetTimeSampleMultiplier();
```
mattyjams, line 246: Allow context to be `nullptr`:
```C++
double      timeSampleMultiplier = (context != nullptr) ? context->GetTimeSampleMultiplier() : 1.0;
```

lib/mayaUsd/fileio/translators/translatorMesh.cpp
```diff
@@ -298,11 +299,20 @@ TranslatorMeshRead::TranslatorMeshRead(
...
     MFnAnimCurve animFn;
 
+    // Get the values needed to convert time to the current maya scenes framerate
+    MTime::Unit timeUnit = MTime::uiUnit();
+    if (!context) {
+        TF_RUNTIME_ERROR("Received null context when writing <%s>", prim.GetPath().GetText());
+        *status = MS::kFailure;
+        return;
+    }
+    double timeSampleMultiplier = context->GetTimeSampleMultiplier();
```
mattyjams, line 309: The original intent with these translator classes was that they could be used outside the context of an import operation to do "in-memory" translations, so you weren't necessarily required to have a context. I'm not sure whether that's still workable now, but I think we'd eventually want to provide that support, so I'd suggest allowing `context` to be `nullptr`:

Allow context to be `nullptr`:
```C++
MTime::Unit timeUnit = MTime::uiUnit();
double      timeSampleMultiplier = (context != nullptr) ? context->GetTimeSampleMultiplier() : 1.0;
```

lib/mayaUsd/fileio/translators/translatorNurbsPatch.cpp
```diff
@@ -264,10 +264,12 @@ bool UsdMayaTranslatorNurbsPatch::Read(
         MFnAnimCurve animFn;
 
         // Construct the time array to be used for all the keys
-        MTimeArray timeArray;
+        MTime::Unit timeUnit = MTime::uiUnit();
+        double      timeSampleMultiplier = context->GetTimeSampleMultiplier();
```
mattyjams, line 268: Allow context to be `nullptr`:
```C++
double      timeSampleMultiplier = (context != nullptr) ? context->GetTimeSampleMultiplier() : 1.0;
```

lib/mayaUsd/fileio/translators/translatorPrim.cpp
```diff
@@ -33,6 +33,8 @@ void UsdMayaTranslatorPrim::Read(
     UsdMayaPrimReaderContext*    context)
 {
     UsdGeomImageable primSchema(prim);
+    MTime::Unit      timeUnit = MTime::uiUnit();
+    double           timeSampleMultiplier = context->GetTimeSampleMultiplier();
```
mattyjams, line 37: Allow context to be `nullptr`:
```C++
double      timeSampleMultiplier = (context != nullptr) ? context->GetTimeSampleMultiplier() : 1.0;
```

lib/mayaUsd/fileio/translators/translatorSkel.cpp
```diff
@@ -476,8 +476,10 @@ bool _CopyAnimFromSkel(
     _GetJointAnimTimeSamples(skelQuery, args, &usdTimes);
     MTimeArray mayaTimes;
     mayaTimes.setLength(usdTimes.size());
+    MTime::Unit timeUnit = MTime::uiUnit();
+    double      timeSampleMultiplier = context->GetTimeSampleMultiplier();
```
mattyjams, line 480: Allow context to be `nullptr`:
```C++
double      timeSampleMultiplier = (context != nullptr) ? context->GetTimeSampleMultiplier() : 1.0;
```

lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -210,6 +210,9 @@ static bool _pushUSDXformOpToMayaXform(
     const UsdMayaPrimReaderArgs&    args,
     const UsdMayaPrimReaderContext* context)
 {
+    MTime::Unit timeUnit = MTime::uiUnit();
+    double      timeSampleMultiplier = context->GetTimeSampleMultiplier();
```
mattyjams, line 214: Allow context to be `nullptr`:
```C++
double      timeSampleMultiplier = (context != nullptr) ? context->GetTimeSampleMultiplier() : 1.0;
```

lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -380,6 +383,9 @@ static bool _pushUSDXformToMayaXform(
 
     std::vector<UsdTimeCode> timeCodes;
     MTimeArray               timeArray;
+    MTime::Unit              timeUnit = MTime::uiUnit();
+
+    double timeSampleMultiplier = context->GetTimeSampleMultiplier();
```
mattyjams, line 388: Allow context to be `nullptr`:
```C++
double      timeSampleMultiplier = (context != nullptr) ? context->GetTimeSampleMultiplier() : 1.0;
```

mattyjams (commented):
I'll leave it to @fowlertADSK and @langlor-autodesk for the final approval, but it generally makes sense to me.

I left a bunch of comments, but they're all around the same request: It would be good **not** to require a prim reader context in the various translator classes. At some point, we want to have better support for ad-hoc translation outside the context of an import. That was the aspiration initially, though it's probably not a clean enough API for that purpose at the moment.

dgovil:
Thanks @mattyjams , and thanks for highlighting all the uses. Did a quick `rg` to see if there were any others, but you got them all.

Just rebuilt and updated the commit.

mattyjams:
> Thanks @mattyjams , and thanks for highlighting all the uses. Did a quick `rg` to see if there were any others, but you got them all.
> 
> Just rebuilt and updated the commit.

Sweet, thanks @dgovil! :)

langlor-autodesk:
I don't know for USD but Maya internally handles times using a fixed point representation which ensure a constant precision across the entire time domain. However MTime only exposes these times using floating points. These floating points are more precise than Maya internal time close to time 0 but they become less precise than Maya for large time values. This difference of precision could affect round trip workflows if the Maya USD plugin supports such workflows. When testing, please ensure tests cover non integer framerates and fractional keyframe values.

langlor-autodesk (approved):
The comments I left are not justifying to hold the change. For me it can be merged. Please invite me to future reviews involving time handling

lib/mayaUsd/fileio/jobs/readJob.cpp
```diff
@@ -500,4 +508,13 @@ void UsdMaya_ReadJob::SetMayaRootDagPath(const MDagPath& mayaRootDagPath)
 
 const MDagPath& UsdMaya_ReadJob::GetMayaRootDagPath() const { return mMayaRootDagPath; }
 
+double UsdMaya_ReadJob::timeSampleMultiplier() const { return mTimeSampleMultiplier; }
+
+double UsdMaya_ReadJob::_setTimeSampleMultiplierFrom(const double layerFPS)
+{
+    float sceneFPS = UsdMayaUtil::GetSceneMTimeUnitAsFloat();
```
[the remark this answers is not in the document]
dgovil: This function is only used in one other place for USD metadata so I can change it to double instead of float.

Is there a list of constant values and/or mathematical formulae to use for correctness? Alternatively I can just update this PR with the changes to double tomorrow and if you can let me know which values should be modified, that would be awesome.

Semi related, it would be awesome if a future API version of MTime provided these out of the box to prevent discrepancies in different implementations that each provide their own mapping.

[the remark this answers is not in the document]
langlor-autodesk: The only non integer framerates that Maya currently supports (24df, 30df, 48df and 60fc) are all based on the same multiplier 1000/1001. So for 24df = 24*1000/1001 ~= 23.976, 30df ~= 29.97 and 60df=60*1000/1001 ~= 59.94
Unless USD does not support the notion of partial frame, rounding can usually and should usually be avoided.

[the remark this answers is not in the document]
dgovil: @langlor-autodesk Thanks, I updated the PR with a change to `util.cpp/h` to change it all to doubles. 
I also changed all the drop frame conversions to `wholeFrame * 1000.0 / 1001`. 

Hopefully everything looks good now? I was unsure as to the difference between `k29_97FPS` and `k29_97DF` however, so used the same math for both. I hope that's correct.

langlor-autodesk (approved):
Thanks, I now see that framerates are now as precise as the double precision,

kxl-adsk:
@dgovil Can you please address clang format errors.

dgovil:
@kxl-adsk Done, apologies, didn't think it would want to unindent those lines with the latest changes.

lib/mayaUsd/utils/util.cpp
```diff
@@ -2079,171 +2079,171 @@ void UsdMayaUtil::GetFilteredSelectionToExport(
...
-        ret = 48.f;
+        ret = 48.0;
     } break;
     case MTime::k50FPS: {
-        ret = 50.f;
+        ret = 50.0;
     } break;
     case MTime::k59_94FPS: {
-        ret = 59.94f;
+        ret = (60.0 * 1000.0) / 100;
```
mattyjams, line 2150: Hey @dgovil! Sorry, only just noticing this as I'm integrating these changes internally.

I think `1001.0` was probably what was intended here rather than `100`?

dgovil: oh yes! I must have hit backspace. let me put up a new PR

dgovil: https://github.com/Autodesk/maya-usd/pull/1010

PR #980, opened 2020-12-07
Use RUNPATH instead of RPATH for linux plugins

fabal:
This bakes runtime paths using RUNPATH instead of RPATH for linux plugins (i.e. modules) as well.


HamedSabri-adsk (approved):
Looking good. Thank you @fabal 

Can you please provide a bit more information about the case where you needed this change?  Was this the case with mayaUsdPlugin?

fabal:
That's correct @HamedSabri-adsk , this is the case with mayaUsdPlugin.

PR #985, opened 2020-12-08
MAYA-106079 Skip Looks scopes

JGamache-autodesk:
Skip importing non-empty UsdScope that only contains UsdShade items.

### DIFF
--- a/lib/usd/translators/scopeReader.cpp
+++ b/lib/usd/translators/scopeReader.cpp
@@ -19,8 +19,7 @@
 #include <pxr/pxr.h>
 #include <pxr/usd/usd/prim.h>
 #include <pxr/usd/usdGeom/scope.h>
-#include <pxr/usd/usdShade/nodeGraph.h>
-#include <pxr/usd/usdShade/shader.h>
+#include <pxr/usd/usdShade/connectableAPI.h>
 
 #include <maya/MObject.h>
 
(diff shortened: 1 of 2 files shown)


lib/usd/translators/scopeReader.cpp
```diff
@@ -27,7 +29,23 @@ PXR_NAMESPACE_OPEN_SCOPE
...
+        if (child.IsA<UsdShadeNodeGraph>() || child.IsA<UsdShadeShader>()) {
+            hasLookData = true;
+        } else {
+            hasNonLookData = true;
+            break;
+        }
+    }
+    if (hasLookData && !hasNonLookData) {
+        return false;
+    }
```
mattyjams, line 46: Makes sense to me, but a few somewhat cosmetic notes:
- Though `Looks` is unfortunately still the default for the material scope name, I think we'd like to transition away from using that term, and really we're just interested in finding *any* `UsdGeomScope` prims with authored scene description below them that is only shading information.
- It might be a bit more robust to check whether the child prim is compatible with `UsdShadeConnectableAPI` rather than looking specifically at its node graph-ness or shader-ness.

So with those two small tweaks, something like this?:
```C++
    // If this scope contains only UsdShade nodes, just skip.
    bool hasShadingData = false;
    bool hasNonShadingData = false;
    for (const auto& child : usdPrim.GetChildren()) {
        if (UsdShadeConnectableAPI(child)) {
            hasShadingData = true;
        } else {
            hasNonShadingData = true;
            break;
        }
    }
    if (hasShadingData && !hasNonShadingData) {
        return false;
    }
```

mattyjams (approved):
Just some minor suggestions, but otherwise looks good to me!

PR #988, opened 2020-12-09
MAYA-107516 USD Layer Editor

luceric-adsk:
The Maya USD Layer Editor provides a hierarchical view used to create, edit and save USD layers


lib/usd/ui/layerEditor/qtUtils.h
```diff
@@ -0,0 +1,68 @@
...
+
+#ifdef Q_OS_DARWIN
+const bool IS_MAC_OS = true;
+#else
+const bool IS_MAC_OS = false;
+#endif
+
+extern QtUtils* utils;
+
+template <class T> inline T DPIScale(T pixel) { return static_cast<T>(pixel * utils->dpiScale()); }
```
seando-adsk, line 64: By doing this you've lost what we do internally in Maya when scaling values. At 100/200% it is not a problem, but at 150% we ran into problems when scaling and trying to keep drawing items lined up. So we look at even/odd of number before calc and then after we force the scaled value to match even/odd or original value.

lib/usd/ui/layerEditor/resources.qrc
```diff
@@ -0,0 +1,34 @@
+<RCC>
+    <qresource prefix="/UsdLayerEditor">
+
+        <file alias="LE_save_all_100.png">resources/save_all_100.png</file>
```
seando-adsk, line 4: Is there a reason to use the "alias" tag. In Maya we only put the alias on the _100 version to name it without the _100

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -162,3 +221,111 @@ global proc mayaUsdMenu_unloadui() {
...
+	    menuItem -label $label -enable $enabled -c $cmd;
+	}
+
+
+    if (!$isAnonymousLayer) {
+        $label = getMayaUsdString("kMenuClear");
+	    $cmd = makeCommand($panelName, "clearLayer");
+	    $enabled = 1;
+	    menuItem -label $label -enable $enabled -c $cmd;
+	}
```
seando-adsk, line 329: The spacing is not all the same one these lines, and the closing } doesn't match indentation of {

seando-adsk (approved):
I quickly went thru as there were so many files. I didn't see anything major - just a few minor issues.

HamedSabri-adsk (changes requested):
See MAYA-108697

plugin/adsk/scripts/mayaUsdMenu.mel
luceric-adsk: ok, this actually all lined up, but let me convert those tabs to spaces

lib/usd/ui/layerEditor/resources.qrc
luceric-adsk: it's because the alias removes the subdirectory name

lib/mayaUsd/commands/layerEditorWindowCommand.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+    // editor name
+    syntax.addArg(MSyntax::kString);
+
+    return syntax;
+}
+
+LayerEditorWindowCommand::LayerEditorWindowCommand()
+{
+    //
```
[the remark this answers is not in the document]
luceric-adsk: sorry, this is about preventing clang-format from putting that method on a single line. it's hard to find it afterwards.

lib/usd/ui/layerEditor/qtUtils.h
luceric-adsk: I think you're right, but we tried the layer at 150% and it looked fine.  Right now I do not want to mess around with it, but it could be rerouted through MQtUtils in the future, as you have down in your treeview

lib/usd/ui/layerEditor/mayaLayerEditorWindow.cpp
```diff
@@ -160,6 +154,12 @@ void MayaLayerEditorWindow::onCreateUI()
     setCentralWidget(_layerEditor);
     _layerEditor->show();
     _sessionState.registerNotifications();
+
+    connect(
```
luceric-adsk, line 158: fix for MAYA-108697 @HamedSabri-adsk

lib/usd/ui/layerEditor/abstractCommandHook.h
```diff
@@ -0,0 +1,111 @@
...
+
+#include "pxr/usd/sdf/layer.h"
+
+#include <pxr/usd/usd/stage.h>
+
+#include <QtCore/QString>
+
+#include <string>
+
+namespace UsdLayerEditor {
```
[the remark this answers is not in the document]
luceric-adsk: Listen, at this moment I've decided the layer editor is not exported from the UI lib and is therefore not in the mayaUsd API and therefore does not need to versionned.  This can be changed in the future, but right now, it doesn't need to be in the public MayaUsd namespace.  The mel commands, however, are.

HamedSabri-adsk (approved):

PR #990, opened 2020-12-09
Fowlert/maya 108091/save layers dialog

fowlertADSK:
Main change is to add a new "Save Layers" dialog that can batch save all anonymous layers.

Couple minor fixes to:
Fixing the MEL code that determines what folder to start the file dialog in when there is no current Maya scene.
Change a couple dialogs to not have the "?" buttons in the title bar.


lib/usd/ui/layerEditor/loadLayersDialog.cpp
```diff
@@ -167,7 +167,7 @@ void LayerPathRow::setAsRowInserter(bool setIt)
 }
 
 LoadLayersDialog::LoadLayersDialog(LayerTreeItem* in_treeItem, QWidget* in_parent)
-    : QDialog(in_parent)
+    : QDialog(in_parent, Qt::WindowTitleHint | Qt::WindowSystemMenuHint | Qt::WindowCloseButtonHint)
```
fowlertADSK, line 170: On Windows at least, this will remove the ? button from the dialog title bar.

lib/usd/ui/layerEditor/mayaSessionState.cpp
```diff
@@ -305,12 +305,27 @@ void MayaSessionState::setupCreateMenu(QMenu* in_menu)
...
 
+const char* getCurrentSaveAsFolderScript = R"(
+global proc string MayaSessionState_GetCurrentSaveAsFolder()
+{
+    string $sceneFolder = dirname(`file -q -sceneName`);
+    if ("" == $sceneFolder)
+    {
+        string $workspaceLocation = `workspace -q -fn`;
+        string $scenesFolder = `workspace -q -fileRuleEntry "scene"`;
+        $sceneFolder = $workspaceLocation + "/" + $scenesFolder;
```
fowlertADSK, line 316: First check if there is a current scene, but if not then query the workspace for a good starting folder.  Without this check the dialog will typically open in the current working directory, which could be wherever the Maya executable is.

lib/usd/ui/layerEditor/saveLayersDialog.cpp
```diff
@@ -0,0 +1,414 @@
...
+            QFileInfo fInfo(path);
+            if (fInfo.exists()) {
+                existingFiles.append(path);
+            }
+        }
+    }
+
+    if (!existingFiles.isEmpty()) {
+        QString message
+            = QString("%1 file(s) already exist and will be overwritten.  Do you want to continue?")
```
luceric-adsk, line 406: string not translated like the other. Also, that "i" and the "count" in the for loop should be declared in the "for"

luceric-adsk (approved):

lib/usd/ui/layerEditor/saveLayersDialog.cpp
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    // Ok/Cancel button area
+    auto buttonsLayout = new QHBoxLayout();
+    QtUtils::initLayoutMargins(buttonsLayout, DPIScale(20));
+    buttonsLayout->addStretch();
+    auto okButton = new QPushButton("Save Stage", this);
+    connect(okButton, &QPushButton::clicked, this, &SaveLayersDialog::onSaveAll);
+    okButton->setDefault(true);
+    auto cancelButton = new QPushButton("Cancel", this);
```
seando-adsk, line 280: These strings should be moved to the new string resource file.

PR #992, opened 2020-12-11
MAYA-108519 Choose an appropriate location to open the file dialog.

fowlertADSK:
Adding a proc to figure out a good starting location for a file dialog that is going to read in a USD file.  Right now it just defaults to whatever the current working directory is, or whatever folder it was in last.

I do plan on eventually refactoring a few other places that open dialogs, but right now just keeping this one as a local proc.  We have some calls in c++, both in the plugin and in the Outliner in Maya, that I want to clean up but it's not a huge priority right now.


seando-adsk (approved):
Looks good.

PR #993, opened 2020-12-11
Clean up Visibility Logics

HamedSabri-adsk:
Use UsdGeomImageable's GetVisibilityAttr, MakeVisible, MakeInvisible API directly.

### DIFF
--- a/lib/mayaUsd/ufe/UsdObject3d.cpp
+++ b/lib/mayaUsd/ufe/UsdObject3d.cpp
@@ -81,14 +81,13 @@ Ufe::BBox3d UsdObject3d::boundingBox() const
 bool UsdObject3d::visibility() const
 {
     TfToken visibilityToken;
-    auto visAttr = UsdGeomImageable(fPrim).GetVisibilityAttr();
+    auto    visAttr = UsdGeomImageable(fPrim).GetVisibilityAttr();
     visAttr.Get(&visibilityToken);
 
     bool result;
-    if (visibilityToken == UsdGeomTokens->invisible) { 
+    if (visibilityToken == UsdGeomTokens->invisible) {
         result = false;
-    }
-    else if (visibilityToken == UsdGeomTokens->inherited) {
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/UsdObject3d.cpp
```diff
@@ -99,19 +80,23 @@ Ufe::BBox3d UsdObject3d::boundingBox() const
...
 bool UsdObject3d::visibility() const
 {
-    auto visAttr = getVisibilityAttribute(sceneItem());
-    if (visAttr) {
-        return (visAttr->get() != UsdGeomTokens->invisible);
+    TfToken visibilityToken;
+    auto    visAttr = UsdGeomImageable(fPrim).GetVisibilityAttr();
+    visAttr.Get(&visibilityToken);
+
+    bool result;
```
ppt-adsk, line 87: Your code below is structured such that there might be token values other than invisible or inherited; if that's the case (and I don't believe it is possible), result would be uninitialized.  For safety and conciseness, I would go with something like the existing code, i.e.
return visibilityToken != UsdGeomTokens->invisible;

ppt-adsk (changes requested):

lib/mayaUsd/ufe/UsdObject3d.cpp
HamedSabri-adsk: Yeah, I thought about this but couldn't convince myself that this ever change. I mean what else could be added? :)
https://graphics.pixar.com/usd/docs/api/class_usd_geom_imageable.html#a64bae16002f3448a6cdeee55e8ebf1fc
I am happy to change it, would be less branching as well.

HamedSabri-adsk: Please see 696db13a64c0795142b7035d0d78d95fc0607782

ppt-adsk (approved):

PR #994, opened 2020-12-11
MAYA-107449 Layer disappears when moved under last layer.

spinell-adsk:
When a layer in the Layer Editor is moved (drag & drop) under the last layer in a hierarchy it will disappear.

drag & drop a layer is a two steps operation. 
 1- Removed the layer from it's original index position
 2- Adding the layer to is new index location. 

The problem is we can't add a layer at the end of the layer list because of a wrong comparaison.
So, the layer is removed but never added at the new location.


fowlertADSK (approved):
Looks good to me!

PR #998, opened 2020-12-11
use the PXR_NS macro instead of pxr in the layer editor

mattyjams:
Just a small follow-up from the merge of the layer editor work from #988:

The `PXR_NS` macro should be used rather than hard-coding `pxr` as the namespace when referring to symbols in USD in case the core USD build is configured to use a different namespace (or in our case, no namespace with `PXR_USE_NAMESPACES=0`).

### DIFF
--- a/lib/usd/ui/layerEditor/stageSelectorWidget.h
+++ b/lib/usd/ui/layerEditor/stageSelectorWidget.h
@@ -39,12 +39,13 @@ public:
     StageSelectorWidget(SessionState* in_sessionState, QWidget* in_parent);
 
 protected:
-    void                setSessionState(SessionState* in_sessionState);
-    pxr::UsdStageRefPtr selectedStage();
+    void                   setSessionState(SessionState* in_sessionState);
+    PXR_NS::UsdStageRefPtr selectedStage();
 
     // slot:
-    void updateFromSessionState(pxr::UsdStageRefPtr const& stageToSelect = pxr::UsdStageRefPtr());
-    void stageRenamed(std::string const& name, pxr::UsdStageRefPtr const& stage);
+    void
+         updateFromSessionState(PXR_NS::UsdStageRefPtr const& stageToSelect = PXR_NS::UsdStageRefPtr());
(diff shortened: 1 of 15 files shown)


lib/usd/ui/layerEditor/stageSelectorWidget.h
```diff
@@ -39,12 +39,13 @@ class StageSelectorWidget : public QWidget
...
-    void                setSessionState(SessionState* in_sessionState);
-    pxr::UsdStageRefPtr selectedStage();
+    void                   setSessionState(SessionState* in_sessionState);
+    PXR_NS::UsdStageRefPtr selectedStage();
 
     // slot:
-    void updateFromSessionState(pxr::UsdStageRefPtr const& stageToSelect = pxr::UsdStageRefPtr());
-    void stageRenamed(std::string const& name, pxr::UsdStageRefPtr const& stage);
+    void
+         updateFromSessionState(PXR_NS::UsdStageRefPtr const& stageToSelect = PXR_NS::UsdStageRefPtr());
```
mattyjams, line 47: An unfortunate choice made by `clang-format` here. :(

kxl-adsk: We keep track of issues like this. If you have to know the configuration that would prevent this, please send it my way and I will put it in our internal wiki. Once we will be considering adopting clang-format v11, we will revisit configuration settings to consider tweaks for annoying parts people are finding.

kxl-adsk (approved):

lib/usd/ui/layerEditor/stageSelectorWidget.h
mattyjams: Thanks @kxl-adsk! Looking at the [current set of options](https://clang.llvm.org/docs/ClangFormatStyleOptions.html), I think it might be `IndentWrappedFunctionNames: 'false'`? I haven't had a chance to build a newer version of clang-format to try it out though. Looks like it was known as `IndentFunctionDeclarationAfterType` for a while too: https://github.com/llvm/llvm-project/commit/c75e1effed73e1a3f7e9b509634648d71c1dddcf

PR #1006, opened 2020-12-15
duplicate baseline images for testMtohVisibility to isolate it from other tests

mattyjams:
Though this does create extra copies of the image files, it seems preferable to have each test be fully self-contained and not have a hidden dependency on some other test(s).

Internally, we need this since we do not install the test scripts and the baseline images in the same directories relative to each other, so the test is unable to find the baselines when we attempt to find them from some other test.


test/lib/mayaUsd/render/mayaToHydra/testMtohVisibility.py
```diff
@@ -1,7 +1,3 @@
-import os
```
mattyjams, line 1: `os` became unused after my changes below, but `sys` and `unittest` were never used, so I just removed all three.

pmolodo:
Looks good - only comment is that, if we are going to be doing a copy, let's give "instances_0.png" a name that makes more sense in the new context - ie, when this was a symlink, it was formerly named "nothing.png"

pmolodo (changes requested):
Looks good - only comment is that, if we are going to be doing a copy, let's give "instances_0.png" a name that makes more sense in the new context - ie, when this was a symlink, it was formerly named "nothing.png"

mattyjams:
> Looks good - only comment is that, if we are going to be doing a copy, let's give "instances_0.png" a name that makes more sense in the new context - ie, when this was a symlink, it was formerly named "nothing.png"

Ahh, good call, sorry. I somehow managed to neither read the variable name in the test nor actually look at the image, so I failed to notice that it was actually empty! Fixed in 019d055241a3df938988ab262a9e72cae7083187.

pmolodo (approved):
Looks good! And I wouldn't have expected you to notice that the name used to be something else at some point in the past - "instances_0.png" was definitely the most obvious choice!

PR #1007, opened 2020-12-15
ensure that GL is initialized when mtoh creates the Storm render delegate

mattyjams:
I'm not sure how others aren't running into this, but we're seeing all of the newly added `mtoh` tests crashing because GL/GLEW isn't fully initialized at the point where it discovers the Storm renderer plugin and tries to call `GlfContextCaps::InitInstance()`.

Prior to core USD 21.02, this was handled by calling `GlfGlewInit()`, but from 21.02 on, the GLEW dependency has been removed, and `GarchGLApiLoad()` is  used instead.

The same pattern of calling `GlfGlewInit()`/`GarchGLApiLoad()` before calling `GlfContextCaps::InitInstance()` can be seen in some of the unit tests in core USD, for example:
https://github.com/PixarAnimationStudios/USD/blob/e9172af0ccea9896990f5ac093c3d6063ac7cab5/pxr/imaging/hdSt/unitTestGLDrawing.cpp#L94

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/utils.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/utils.cpp
@@ -13,6 +13,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+
+// GLEW must be included early (for core USD < 21.02), but we need pxr.h first
+// so that USD_VERSION_NUM has the correct value.
+// We also disable clang-format for this block, since otherwise v10.0.0 fails
+// to recognize that "utils.h" is the related header.
+// clang-format off
+#include <pxr/pxr.h>
+#if USD_VERSION_NUM < 2102
+#include <pxr/imaging/glf/glew.h>
+#endif
+// clang-format on
+
 #include "utils.h"
 
 #include "renderGlobals.h"
@@ -23,6 +35,10 @@
 #include <pxr/imaging/hd/rendererPluginRegistry.h>
 
 #include <maya/MGlobal.h>
+
+#if USD_VERSION_NUM >= 2102
+#include <pxr/imaging/garch/glApi.h>
+#endif
 
 PXR_NAMESPACE_OPEN_SCOPE
 
@@ -51,8 +67,14 @@ MtohInitializeRenderPlugins()
             }
 
             // XXX: As of 22.02, this needs to be called for Storm
-            if (pluginDesc.id == MtohTokens->HdStormRendererPlugin)
+            if (pluginDesc.id == MtohTokens->HdStormRendererPlugin) {
+#if USD_VERSION_NUM >= 2102
+                GarchGLApiLoad();
+#else
+                GlfGlewInit();
+#endif
                 GlfContextCaps::InitInstance();
+            }
 
             HdRenderDelegate* delegate
                 = plugin->IsSupported() ? plugin->CreateRenderDelegate() : nullptr;


lib/mayaUsd/render/mayaToHydra/utils.cpp
```diff
@@ -13,6 +13,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+
+// GLEW must be included early (for core USD < 21.02), but we need pxr.h first
+// so that USD_VERSION_NUM has the correct value.
+// We also disable clang-format for this block, since otherwise v10.0.0 fails
+// to recognize that "utils.h" is the related header.
+// clang-format off
```
mattyjams, line 21: Another `clang-format` oddity here, but when I added this block of includes and the conditional, `clang-format` v10.0.0 starts failing to recognize that `"utils.h"` is the related header for this cpp file, and it tries to put it at the bottom of the "private" header group at line 32. I opted to disable `clang-format` here rather than absorb that change.

lib/mayaUsd/render/mayaToHydra/utils.cpp
```diff
@@ -24,6 +36,10 @@
 
 #include <maya/MGlobal.h>
 
+#if USD_VERSION_NUM >= 2102
+#include <pxr/imaging/garch/glApi.h>
+#endif
```
mattyjams, line 41: Unlike GLEW, this `garch` include does **not** need to come first.

pmolodo:
Hmm... well, note sure why @mattyjams was encountering this, and I didn't. (Nor Hamed or the pre-flight tests...) However, since it should be fine to call  `GlfGlewInit` more than once, this looks good / makes sense to me.

pmolodo (approved):
Hmm... well, note sure why @mattyjams was encountering this, and I didn't. (Nor Hamed or the pre-flight tests...) However, since it should be fine to call GlfGlewInit more than once, this looks good / makes sense to me.

lib/mayaUsd/render/mayaToHydra/utils.cpp
kxl-adsk: Do you happen to know if v11.0.0 will has it fixed?

kxl-adsk:
Apologies @mattyjams, this somehow pass undetected in our pre-merge validation. I tested again, and you are correct that dev was broken with dev post 20.11.

lib/mayaUsd/render/mayaToHydra/utils.cpp
mattyjams: > Do you happen to know if v11.0.0 will has it fixed?

I just built clang v11.0.0 and gave it a try (using the current configuration) and I get the same results. It seems to get tripped up by the conditional that wraps the `glew.h` `#include`. If I remove the `#if`/`#else`, it correctly figures out that `utils.h` should stay where it is, just after `glew.h` but before the other "local" headers.

lib/mayaUsd/render/mayaToHydra/utils.cpp
mattyjams: I had a follow-up chat with the Hydra team about this. They were saying that the newly added GL loader doesn't break the way `glew.h` does if it's not `#include`'d first, it's probably still good practice to include it early. I've got a new PR coming shortly with a bunch more instances of this, so I'll revisit this one as part of those changes. I'll also amend the clang-format config so that it includes both `glew.h` and `glApi.h` in the first group.

The Hydra team also just made a change to `GlfContextCaps::InitInstance()` so that it calls `GarchGLApiLoad()` itself, so we'd then be able to remove the call I had to add here for core USD 21.02+.

PR #1008, opened 2020-12-16
MAYA-108814 Support viewport display filtering for meshes and basisCurves

HdC-adsk:
* If polygons are excluded from display, USD meshes will be filtered out.
* If NURBS curves are excluded from display, USD curves will be filtered out.
* If plugin shapes are excluded from display, all USD objects will be filtered out.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -32,6 +32,7 @@
 #include <pxr/imaging/hd/tokens.h>
 #include <pxr/imaging/hd/version.h>
 
+#include <maya/MFrameContext.h>
 #include <maya/MMatrix.h>
 #include <maya/MProfiler.h>
 #include <maya/MSelectionMask.h>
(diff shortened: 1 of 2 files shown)


HdC-adsk:
@kxl-adsk I talked with Will again whether we need the "fine-grain" override control and we feel that it is safer not to do it the way you suggested. 

The reason is that we don't have any doc about the viewport object-type filters as its behavior is quite intuitive, I assume users would expect all plug-in render items to be excluded from viewport when user chooses to exclude plug-in shapes. If we changed it to a "fine-grain" override control then we would introduce a slightly more complex logic that we had to document and made it clear.

Note the behavior is an internal implementation detail so we can change it if required by users without modifying the API, therfore we can wait for a strong requirement before changing the logic.

williamkrick (approved):

kxl-adsk (approved):

HdC-adsk:
I need to run PF after Maya version including new API is blessed.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -1666,6 +1667,10 @@ MHWRender::MRenderItem* HdVP2BasisCurves::_CreateWireRenderItem(const MString& n
     renderItem->setShader(_delegate->Get3dSolidShader(kOpaqueGray));
     renderItem->setSelectionMask(MSelectionMask::kSelectCurves);
 
+#if MAYA_API_VERSION >= 20210000
```
mattyjams, line 1670: Hey all, is it possible that we might need to use a higher API version number here? We're continuing to build against PR 121 at the moment while we get our licenses sorted out for the newly released PR 122, but compilation against PR 121 seems to be failing with `setObjectTypeExclusionFlag()` not being found.

kxl-adsk: For beta releases, we only support the last release, i.e. PR122.

PR #1009, opened 2020-12-16
[tests] interactive tests ensure output streams are flushed before quitting

pmolodo:
When developing #915, we uncovered an issue where [output streams weren't being flushing before interactive maya tests quit](https://github.com/Autodesk/maya-usd/pull/915#issuecomment-738472429). 

This applies the fix developed in that PR to all interactive maya tests, so we can be sure to see all their output on stdout.

### DIFF
--- a/plugin/al/plugin/AL_USDMayaTestPlugin/py/testLayerManager.py
+++ b/plugin/al/plugin/AL_USDMayaTestPlugin/py/testLayerManager.py
@@ -23,6 +23,8 @@ import unittest
 from maya import cmds
 from AL.usdmaya import ProxyShape
 
+import fixturesUtils
+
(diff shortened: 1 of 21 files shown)


pmolodo:
One note - it seems I had my editor's option to automatically strip trailing whitespace on when I was making these changes.  There's not too many of them, though, so hopefully it isn't too confusing.  If you feel it is, I can separate out the whitespace changes.

kxl-adsk (approved):

kxl-adsk:
LGTM but since most of the latest interactive tests were added by @mattyjams, I would like to make sure it works for him too.

mattyjams (approved):
Looks good to me too! Thanks @pmolodo!

PR #1010, opened 2020-12-16
Fix 59.95FPS double conversion

dgovil:
Thanks to @mattyjams for catching this typo with this specific frame rate. All the other drop frames look fine.
Apologies for that typo.


mattyjams (approved):
Looks good to me! Thanks for the quick fix!

kxl-adsk (approved):
Good catch indeed!

PR #1013, opened 2020-12-17
Fix a build issue on Linux with maya<2020

pmolodo:
On maya<=2019, maya ends up including Xlib.h, which does

```
#define Bool int
```

...which messes with SdfValueTypeNames->Bool. So on Linux with Maya<2020,
this just does an `#undef Bool`.

This may only be an issue on specific Linux distributions, since it's tied
to the system supplied GL/glx.h and Xlib.h headers.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeUI.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeUI.cpp
@@ -15,6 +15,19 @@
 //
 #include "AL/usdmaya/nodes/ProxyShapeUI.h"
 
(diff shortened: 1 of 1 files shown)


mattyjams:
Ugh, yeah, we tripped over that as well and ended up littering a bunch of code with early `#include`s of `sdf/types.h`.

The awful details are described here:
https://github.com/Autodesk/maya-usd/blob/ae0cb7cf326e6df3ace5cbe212242a9a22b89129/lib/mayaUsd/render/pxrUsdMayaGL/batchRenderer.h#L25

But I notice the include of `sdf/types.h` is **not** there. Maybe the `#include` became disassociated from that comment block as a result of `clang-format`? We don't build for Maya 2019 anymore, so we haven't noticed it.

kxl-adsk (approved):

PR #1015, opened 2020-12-17
MAYA-106310 - Crash when querying target layer and nothing is assigne…

spinell-adsk:
…d the target

- Update the mayaUsdLayerEditor cmd to set the root layer as the edit target when the command is used the remove a layer and this layer is the edit target

### DIFF
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -110,6 +110,7 @@ class InsertRemoveSubPathBase : public BaseCmd
 public:
     int         _index = -1;
     std::string _subPath;
+    std::string _proxyShapePath;
 
     InsertRemoveSubPathBase(CmdId id)
         : BaseCmd(id)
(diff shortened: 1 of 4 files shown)


spinell-adsk:
Removed a layer that's the current edit target with the command mayaUsdLayerEditor will cause a crash when querying the current edit target. UsdStage will return a dangling pointer when the edit layer is removed.

To fix this issue, I updated the command  mayaUsdLayerEditor to set the root layer as the edit layer when removing a layer and this layer is the edit target.

To be able to access to the UsdStage inside the command, I updated the signature for removing a layer.
 - Before:  mayaUsdLayerEditor -e removeSubPath $index $layer
 - Now:     mayaUsdLayerEditor -e removeSubPath $index $shape $layer

lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -136,6 +137,18 @@ class InsertRemoveSubPathBase : public BaseCmd
...
+
+            // if the layer to remove is the current edit target,
+            // set the root layer as the current edit target
+            auto subLayerHandle = SdfLayer::FindRelativeToLayer(layer, _subPath);
+            auto stage = getStage();
+            auto currentTarget = stage->GetEditTarget().GetLayer();
+            if (currentTarget
+                && currentTarget->GetIdentifier() == subLayerHandle->GetIdentifier()) {
+                _isEditTarget = true;
+                stage->SetEditTarget(stage->GetRootLayer());
```
kxl-adsk, line 149: Would be worth considering matching the logic with base proxy shape, ie fallback to session layer of root layer can’t be edited or we forced session layer with option var.

kxl-adsk (approved):

PR #1016, opened 2020-12-17
Added Blendshape export functionality

ysiewappl:
Hi all:

**This is a continuation of the PR originally submitted at #952. Please refer to that one for more comments and previous discussion. This update fixes both "baked" blendshape targets and also an issue that sees blendshape offsets being doubled-up if the first frame being exported has any of the blendshape targets activated as well.**

This PR adds support for exporting blendshapes out of Maya in line with the UsdSkelBlendShape schema for a variety of scenarios, including single base - single target, single base - multiple targets (incl. inbetween), multiple base - multiple targets, "baked" blendshape targets (i.e. no explicit target meshes in the scene), writing out combined extents, and writing out animation. Some limitations include:

Currently no support for "chained" blendshape deformations or other intermediate deformers in the deform stack.
"Layered" blendshape deformers will end up having individual shapes baked out individually.
Combined extents only take blendshapes into account and no other deformers.
I ran into an issue while implementing this, which can be summarized by the following:

Maya does not evaluate certain plugs on a blendshape node until evaluation of the node is forced. This is a problem because certain information like the indices of components on a blendshape cannot be determined unless the blendshape node itself is evaluated.

Steps to Reproduce:

Create a cube and a blendshape for it with one vertex moved.

Run the following code:
```
import maya.OpenMaya as om
sl = om.MSelectionList()
p = om.MPlug()
sl.add('blendShape1.inputTarget[0].inputTargetGroup[0].inputTargetItem[6000].inputComponentsTarget')
sl.getPlug(0, p)

dh = p.asMDataHandle()
d= dh.data()
fn = om.MFnComponentListData(d)
print fn.length()
```

You should have gotten an error when attempting to attach the function set. Now twiddle the weight on the blendshape deformer and twiddle it back and then run the same code again.
The code suddenly works without any other modifications to the scene or its data.
We're currently working around this by basically twiddling the blendshape node's weights through OpenMaya prior to export, which is not ideal since that could potentially be expensive for denser meshes. Hopefully someone from ADSK could help advise here on any possible remedies.


lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -36,14 +36,22 @@
...
+#include <maya/MItDependencyGraph.h>
 #include <maya/MItMeshFaceVertex.h>
 #include <maya/MPlug.h>
 #include <maya/MPlugArray.h>
+#include <maya/MPoint.h>
 #include <maya/MStatus.h>
 #include <maya/MUintArray.h>
+#include <maya/MVector.h>
+
+#define MAYA_ATTR_NAME_INMESH "inMesh"
```
mattyjams, line 54: `static constexpr char kMayaAttrNameInMesh[] = "inMesh";` here?

lib/mayaUsd/utils/util.h
```diff
@@ -580,6 +601,22 @@ float ConvertMTimeUnitToFloat(const MTime::Unit& unit);
 MAYAUSD_CORE_PUBLIC
 float GetSceneMTimeUnitAsFloat();
 
+/**
+ * Searches the given array for an element.
+ *
+ * @param a         The element to search for.
+ * @param array     The array to search within.
+ * @param idx       Storage for the index of the element within the array if it exists.
+ *                  If it does not exist, this will be undefined.
```
mattyjams, line 610: Maybe mention that `idx` can be `nullptr` if you're only interested in whether the array contains the element and don't care what the index is, and/or make it optional with `unsigned int* idx = nullptr`?

lib/usd/translators/meshWriter.cpp
```diff
@@ -15,6 +15,10 @@
 //
 #include "meshWriter.h"
 
+#include "maya/MApiNamespace.h"
+#include "maya/MStatus.h"
+#include "pxr/base/tf/diagnostic.h"
```
mattyjams, line 20: Angle braces for these includes, and it looks like that maybe confused clang-format, as I'd expect it to have grouped these with their respective `maya` and `pxr` groups below.

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -0,0 +1,981 @@
...
+#include <maya/MPointArray.h>
+#include <maya/MStatus.h>
+
+#include <algorithm>
+#include <complex>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#define MAYA_BLENDSHAPE_EVAL_HOTFIX 1
```
mattyjams, line 55: `constexpr` `int` or `bool` here?

I see this gets used to compile out the call to `mayaBlendShapeTriggerAllTargets()` below, but if we set this to `0`/`false`, I would guess we'd see complaints from the compiler about the function being unused. Any reason not to make that just a regular `if()`?

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -0,0 +1,981 @@
...
+                            &stat); // NOTE: (yliangsiew) The target name is set as an alias, so
+                                    // we'll use that instead of calling our target "weight_".
+                        CHECK_MSTATUS_AND_RETURN(stat, MObject::kNullObj);
+                        // NOTE: (yliangsiew) Because a single weight can drive multiple targets, we
+                        // have to put a numeric suffix in the target name.
+                        int   lenSuffix = snprintf(nullptr, 0, "%zu", k);
+                        char* suffix = (char*)malloc((sizeof(char) * lenSuffix) + 1);
+                        snprintf(suffix, lenSuffix + 1, "%zu", k);
+                        curTargetNameMStr = plgBlendShapeName + MString(suffix);
+                        free(suffix);
```
mattyjams, line 641: Maybe just `TfStringPrintf()` here?:

```C++
curTargetNameMStr = TfStringPrintf("%s%zu", plgBlendShapeName.asChar(), k).c_str();
```

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -0,0 +1,981 @@
...
+                            &stat); // NOTE: (yliangsiew) The target name is set as an alias, so
+                                    // we'll use that instead of calling our target "weight_".
+                        CHECK_MSTATUS_AND_RETURN(stat, MObject::kNullObj);
+                        // NOTE: (yliangsiew) Because a single weight can drive multiple targets, we
+                        // have to put a numeric suffix in the target name.
+                        int   lenSuffix = snprintf(nullptr, 0, "%zu", k);
+                        char* suffix = (char*)malloc((sizeof(char) * lenSuffix) + 1);
+                        snprintf(suffix, lenSuffix, "%zu", k);
+                        curTargetNameMStr = plgBlendShapeName + MString(suffix);
+                        free(suffix);
```
mattyjams, line 812: Maybe just `TfStringPrintf()` here?:

```C++
curTargetNameMStr = TfStringPrintf("%s%zu", plgBlendShapeName.asChar(), k).c_str();
```

mattyjams (commented):
Just a few cosmetic notes from me, but otherwise looks good.

I noticed that there are still a fair number of `TF_VERIFY`s that execution will plow right through where maybe we'd want to check its return value and `return`/`continue`/etc if the check fails.

I'll defer again to @williamkrick to approve though, since he's far more knowledgeable on the details of this than me.

Thanks @ysiewappl!

lib/usd/translators/meshWriter_BlendShapes.cpp
ysiewappl: I just realized that this isn't my intention; it's meant to actually make sure the code isn't included at all in the final plugin if the `#define` is on. I've gone ahead and enclosed the function under that `#define` as well.

lib/usd/translators/meshWriter_BlendShapes.cpp
ysiewappl: Resolved in 2e3a363.

lib/usd/translators/meshWriter_BlendShapes.cpp
ysiewappl: Resolved in 2e3a363.

lib/usd/translators/meshWriter.cpp
ysiewappl: Resolved in 2e3a363.

lib/usd/translators/meshWriter.cpp
```diff
@@ -180,7 +511,61 @@ bool PxrUsdTranslators_MeshWriter::writeMeshAttrs(
...
-    UsdMayaMeshWriteUtils::writePointsData(geomMesh, primSchema, usdTime, _GetSparseValueWriter());
+    /*
+     NOTE: (yliangsiew) Because we cannot assume that the first frame of export
+     will have no blendshape targets activated, and we want to write out the
+     points _without_ the influence of any blendshapes, (or any other deformers,
+     for that matter; just that we haven't implemented support for other
+     deformers yet, so until we do that, we can leave the effect of other
+     deformers "baked" into the base/"pref" pose) we need to deactivate all the
+     blendshape targets here _before_ writing out the data.
+    */
```
williamkrick, line 522: I think you can achieve this in a simpler way by reading MFnGeometryFilter::getInputGeometry(), or directly reading the inputGeometry attribute of the blend shape. That will give you the input mesh without any of the blend shape's modifications. And it will leave any any deformations which occur after. We could have something like blendShape->skinCluster->Wrap->mesh. In the code that is here you'd end up with just the wrap applied, and then using that as the input to the blendShape isn't correct.

The inputGeometry could also be animated. I see some checks above for dealing with time but I'm not sure they capture this fact.

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -0,0 +1,977 @@
...
+            weightInfo.inputTargetIndex = outputGeomPlugIdx;
+
+            stat = fnBlendShape.targetItemIndexList(
+                weightInfo.weightIndex, outputGeo, weightInfo.targetItemIndices);
+            CHECK_MSTATUS_AND_RETURN_IT(stat);
+
+            unsigned int numInTgtGrps = plgInTgtGrps.numElements();
+            if (numInTgtGrps == 0) {
+                continue;
+            }
```
williamkrick, line 370: I think what you have written here will work, but please test a scene were you have a weight value for a non-existent target group. Maya can handle that case (by ignoring the entry which I think you are successfully doing here) so there could be scenes out there with more weights then targets.

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -0,0 +1,977 @@
...
+        TF_WARN(
+            "Cannot find any blendshape deformers for the mesh: %s",
+            deformedMeshDagPath.fullPathName().asChar());
+        return MObject::kNullObj;
+    case 1: break;
+    default:
+        // TODO: (yliangsiew) For multiple blend shape deformers, what do we do?
+        // Do we collapse the shapes from multiple blend targets together, or just
+        // write out only the "closest" blendshape deformer's targets? Or just write all
+        // of them and print this warning to end-users?
```
williamkrick, line 576: I think eventually we'll want a 1 to 1 mapping between deformer nodes in Maya and UsdSkel compute items written to USD. Doing a special case to merge two blend shapes in a row won't help when there's a wrap deformer between them. We'll eventually have to support writing a graph of UsdSkel compute items to the USD file.

I think this change, even though it has a bunch of limitations, is a good step towards that final goal.

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -0,0 +1,977 @@
...
+                size_t numOfTargets = weightInfo.targets.size();
+                for (size_t k = 0; k < numOfTargets; ++k) {
+                    MayaBlendShapeTargetDatum targetDatum = weightInfo.targets[k];
+                    MObject                   targetMesh = targetDatum.targetMesh;
+                    MString                   curTargetNameMStr;
+                    if (!targetMesh.isNull()) {
+                        // NOTE: (yliangsiew) Because UsdSkelBlendShape does not
+                        // support animated targets (the `normalOffsets` and
+                        // `offsets` attributes are defined as uniforms), we cannot
+                        // fully support it in the exporter either.
```
williamkrick, line 611: Ah that is an interesting limitation. We'd need UsdSkelBlendShape targets to accept the same data type that the input & output geometry are to have a more general solution where we can connect a graph of computes together.

williamkrick (approved):

williamkrick:
@ysiewappl there is some discussion we need to have before this change is merged.

It was my understanding from in December that after the original commit for blend shape support you would be continuing on with more work. Is all the work you have planned now included in this pull request or is more work coming?

If you are planning to do more work then I am fine with the minor style comments I've made here being fixed in the next change. If this is the end of the work then I'd like to see those loose ends tied off before this is merged.

After the changes get merged our MayaUSD QAs will test this functionality and be logging bugs for things that don't work. I need you to provide a list of what is expected to work and what is not expected to work for our QAs to validate. Clearly everything in Maya is not supported, so we need to clearly state the limitations of the blend shape export so we know what to expect, and so that other users of MayaUSD know what to expect. Something like what are the supported deformation setups (blendShape->skinCluster->mesh ?) would be perfect.

dgovil:
@williamkrick just as a heads up, Yi Liang is out for a little bit longer so I'll wait for him to be back to respond accurately.

I believe all of our intended changes are represented in this PR.

lib/usd/translators/meshWriter.cpp
```diff
@@ -41,21 +42,178 @@
...
+        // out what the _actual_ base mesh was.
+        if (intermediates == nullptr) {
+            MFnGeometryFilter fnGeoFilter;
+            fnGeoFilter.setObject(curBlendShape);
+            MObject inputGeo = fnGeoFilter.inputShapeAtIndex(outputGeomPlugIdx, &stat);
+            if (inputGeo.hasFn(MFn::kMesh)) {
+                return inputGeo;
+            }
+        } else {
+            intermediates->clear();
```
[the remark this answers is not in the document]
ysiewappl: This is updated in `17e1b0227`.

lib/usd/translators/meshWriter.cpp
```diff
@@ -163,13 +352,155 @@ bool PxrUsdTranslators_MeshWriter::writeMeshAttrs(
...
+    // in the deform stack here, we walk past it to the original shape instead.
+    // Also check if the mesh is a valid DG node (mesh geo subsets are
+    // kMeshData in cases where a single mesh has multiple face
+    // assignments to materials.)
+    if (exportArgs.exportBlendShapes && geomMeshObj.hasFn(MFn::kDependencyNode)) {
+        if (exportArgs.ignoreWarnings) {
+            geomMeshObj = mayaFindOrigMeshFromBlendShapeTarget(geomMeshObj, nullptr);
+        } else {
+            MObjectArray intermediates;
+            geomMeshObj = mayaFindOrigMeshFromBlendShapeTarget(geomMeshObj, &intermediates);
```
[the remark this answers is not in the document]
ysiewappl: This is implemented in `17e1b0227`.

lib/usd/translators/meshWriter.cpp
```diff
@@ -41,21 +42,178 @@
...
 
+MStatus mayaDisableAllBlendShapesForMesh(
+    const MObject& mesh,
+    MObjectArray&  blendShapeNodes,
+    MFloatArray&   envelopes)
+{
+    MStatus status;
+    // NOTE: (yliangsiew) If there's a skinCluster, find that first since that
+    // will be the intermediate to the blendShape node. If not, just search for any
+    // blendshape deformers upstream of the mesh.
```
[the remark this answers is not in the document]
ysiewappl: Hi, thanks for the review!

I think the code currently checks if there is a skinCluster; if not, it will
start the search from the mesh instead, so it should be finding all blendshape
deformers upstream regardless. This is the same code being used in
`mayaFindOrigMeshFromBlendShapeTarget`.

lib/usd/translators/meshWriter.cpp
```diff
@@ -41,21 +42,178 @@
...
+            envelopes.append(curEnvelope);
+            fnGeoFilter.setEnvelope(0.0f);
+        }
+    }
+
+    return MStatus::kSuccess;
+}
+
+MStatus
+mayaEnableGeometryFilterNodesForMesh(const MObjectArray& nodes, const MFloatArray& envelopes)
```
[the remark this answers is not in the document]
ysiewappl: This is basically because we disable the blendshape nodes in the first call, and
since we already know which nodes we disabled, we don't have to find them again.
The intention here is to write out the points of the mesh in the "pref" pose.

lib/usd/translators/meshWriter.cpp
ysiewappl: This might be a better approach to finding the "pref" pose and I can give this a
shot...though I'm a little concerned about just how much I need to support here,
especially when other deformers get thrown into the mix in the future.

lib/usd/translators/meshWriter.cpp
```diff
@@ -163,13 +352,155 @@ bool PxrUsdTranslators_MeshWriter::writeMeshAttrs(
...
+                                            MFnDependencyNode fnNode(curIntermediate, &status);
+                                            CHECK_MSTATUS_AND_RETURN(status, false);
+                                            TF_RUNTIME_ERROR(
+                                                "Could not determine the original blendshape "
+                                                "source mesh due to a non-empty tweak node: %s. "
+                                                "Please either bake it down or remove the "
+                                                "edits and attempt the export process again, or "
+                                                "specify -ignoreWarnings.",
+                                                fnNode.name().asChar());
+                                            return false;
```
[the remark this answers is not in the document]
ysiewappl: With the latest updates; nothing. The prim would simply not be written out to
the file. The expectation I had was that the user would notice the error message
being written on export and address the issue. The `ignoreWarnings` flag would 
then be for users who "don't care" and just want to export with the
aforementioned issues anyway.

ysiewappl:
Hi @williamkrick:

> It was my understanding from in December that after the original commit for blend shape support you would be continuing on with more work. Is all the work you have planned now included in this pull request or is more work coming?

- All work related to this feature is done (unless our team(s) find another
last-minute bug).

> If you are planning to do more work then I am fine with the minor style comments I've made here being fixed in the next change. If this is the end of the work then I'd like to see those loose ends tied off before this is merged.

- I've addressed the housekeeping comments and am currently looking at the issue
  of more accurately determining the pref pose of the mesh, as mentioned above.
  I should have an answer for you soon.

> After the changes get merged our MayaUSD QAs will test this functionality and be logging bugs for things that don't work. I need you to provide a list of what is expected to work and what is not expected to work for our QAs to validate. Clearly everything in Maya is not supported, so we need to clearly state the limitations of the blend shape export so we know what to expect, and so that other users of MayaUSD know what to expect. Something like what are the supported deformation setups (blendShape->skinCluster->mesh ?) would be perfect.

- Certainly. I have a sample Maya scene which I am using to test my export
  configuration against; I could provide it here for you rather than attempting
  to type out every single possible combination of exports if that might be more
  useful?

williamkrick:
@ysiewappl yes a test scene with examples would be fantastic. Can provide a brief text overview of what is expected to work vs. not work as well? You don't have to include everything but just give a high level idea of what to expect.

After that I approve of this PR being merged.

ysiewappl:
> @ysiewappl yes a test scene with examples would be fantastic. Can provide a brief text overview of what is expected to work vs. not work as well? You don't have to include everything but just give a high level idea of what to expect.
Hi:

I've updated the unit test for the blendshapes a tiny bit, and uploaded one of
the test scenes I use for verifying output. The scene contains various basic
test cases, which include:

- Single base to single target, non-animated/animated
- Single base to multiple in-betweens, non-animated/animated
- Multiple bases to multiple targets, animated
- A setup where there is an extra squash deformer in the stack
- A setup with "baked" blendshape targets
- A setup with a skinned cube with "baked" blendshape targets

kxl-adsk (changes requested):
@ysiewappl I tested your build against older versions of USD (the minimum version we support is 19.11) and I'm getting build failure. We don't have it yet covered by PF, so this is why you didn't see it in reports. Here is what I did locally to pass the build error (and it's what will have to be done before we can merge the change)
```
index 0e6b2081d..9da727ff7 100644
--- a/lib/usd/translators/meshWriter_BlendShapes.cpp
+++ b/lib/usd/translators/meshWriter_BlendShapes.cpp
@@ -397,7 +397,7 @@ MStatus mayaGetBlendShapeInfosForMesh(
                     plgInComponentsTgt, indices);
                 for (unsigned int m = 0; m < indices.length(); ++m) {
-                    meshTargetDatum.indices.emplace_back(indices[m]);
+                    meshTargetDatum.indices.push_back(indices[m]);
                 }

                 unsigned int numComponentIndices = meshTargetDatum.indices.size();
@@ -446,7 +446,7 @@ MStatus mayaGetBlendShapeInfosForMesh(
                     MPointArray ptDeltas = fnPtArrayData.array();
                     for (unsigned int m = 0; m < numComponentIndices; ++m) {
                         MPoint pt = ptDeltas[m];
-                        meshTargetDatum.ptOffsets.emplace_back(GfVec3f(pt.x, pt.y, pt.z));
+                        meshTargetDatum.ptOffsets.push_back(GfVec3f(pt.x, pt.y, pt.z));
                     }
                 }
                 weightInfo.targets.emplace_back(meshTargetDatum);
```

In addition, I'm seeing the new test (`testUsdExportBlendshapes`) failing. I will re-trigger the PF once you fix the build error reported above to confirm.

ysiewappl:
There's several points where I call `emplace_back` instead of `push_back`. Are those really the only two points of failure? I would have expected to see some other calls fail as well if c++17 support wasn't assumed. Nevertheless I've changed them all to use `push_back` instead just in case.

ysiewappl:
@kxl-adsk I'm not sure why the Windows builds failed, but I made a commit based on a guess (regarding the filename passed to `MFileIO::open`. If you could help confirm the actual issue (since I don't have a Windows machine here for testing the build against), I'd very much appreciate it so I can get everything resolved and finally get this PR merged. Thanks!

kxl-adsk (approved):
Thank you, builds and tests are passing now.

kxl-adsk:
FWIW. You didn't have to replace all emplac_back. Just the two I mentioned in my comment was enough.

kxl-adsk:
I'm going to merge this change and pass it to our QA. I will share the files and short description you provided. If you can think of anything else to mention when testing, please don't hesitate to let me know.

Finally, thank you for your hard work on this one.

dgovil:
Awesome! Thank you everyone for your feedback on this, and @ysiewappl for putting this all together.

PR #1024, opened 2020-12-21
Fix incorrect `bindTransforms` being determined during export

ysiewappl:
Hi:

This PR fixes an issue that was noticed while I was working on #1016, where: under the situation where a joint does not have its `bindPose` attribute populated with a correct bind transform matrix, the resulting exported USD is incorrect (object has an offset that appears different to what is represented in Maya). This is because the exporter assumes a default identity matrix if it cannot find data on the plug to use.

An example of this problem is shown below, where you can see that the offending bind transform was assumed to be an identity matrix by the exporter (both spheres should have the same Y-value):
image


kxl-adsk:
@ysiewappl Thank you for another great contribution. 

Just wanted to let you know that on Autodesk side we have an extended holiday break, returning in January.

Have a wonderful holiday season!

test/lib/usd/translators/testUsdExportBindTransform.py
```diff
@@ -0,0 +1,60 @@
+#!/pxrpythonsubst
```
kxl-adsk, line 1: Since it’s a new file, please apply change following #1026

test/lib/usd/translators/testUsdExportBindTransform.py
```diff
@@ -0,0 +1,60 @@
...
+    def testBindTransformExport(self):
+        temp_export_file = os.path.join(self.temp_dir, 'test_output.usda')
+        cmds.select('a', r=True)
+        cmds.mayaUSDExport(v=True, sl=True, f=temp_export_file, skl='auto', skn='auto',
+                           defaultMeshScheme='catmullClark')
+
+        stage = Usd.Stage.Open(temp_export_file)
+        prim = stage.GetPrimAtPath('/a/b/c')
+        bindTransforms = prim.GetAttribute('bindTransforms').Get()
+        self.assertNotEqual(bindTransforms, [[1, 0, 0, 0],
```
kxl-adsk, line 53: Even better would be to test for the matrix we expect to read from dag pose.

kxl-adsk (approved):

test/lib/usd/translators/testUsdExportBindTransform.py
ysiewappl: Resolved in 71d7bb0.

test/lib/usd/translators/testUsdExportBindTransform.py
ysiewappl: Resolved in 71d7bb0.

Issue #645, opened 2020-07-10
Remove USD_VERSION_NUMBER and replace with PXR_VERSION


marsupial:
**Is your feature request related to a problem? Please describe.**
The build system introduces a **USD_VERSION_NUMBER** when **PXR_VERSION** already exists since at least 19.11.
It just makes it more complicated and harder to read.

**Describe the solution you'd like**
Replace usage of **USD_VERSION_NUMBER** with **PXR_VERSION** and remove the CMake generation of this constant.

kxl-adsk:
@marsupial When we first introduced `USD_VERSION_NUMBER`, `PXR_VERSION` was not available on all supported USD versions, i.e. we had to support 19.01 and up. With 19.11 being now a minimal supported version, we can rely on the pxr header file and `PXR_VERSION ` for code conditional compilation.
We still need something custom for CMake - this is where we will need to keep `USD_VERSION_NUMBER`

mattyjams:
This should be addressed by #1181.

PR #1025, opened 2020-12-22
updates for building with latest post-20.11 dev branch of core USD

mattyjams:
A few small changes here adapting to recent updates in `garch`, `glf`, and `hdSt` in core USD.

With core USD 21.02, the pxr imaging code no longer has a dependency on GLEW and instead uses its own GL loader in the `garch` library. Where code for USD prior to 21.02 would have called `GlfGlewInit()`, for 21.02 and later it should now call `GarchGLApiLoad()` instead. `GlfContextCaps::InitInstance()` calls `GarchGLApiLoad()` itself, so any calls of `GlfContextCaps::InitInstance()` do not need an additional load call.

Though it's not a strict requirement for `garch/glApi.h` to be included before any other GL header like it was for `glew.h`, our Hydra team tells me it's probably still good practice, so it was added to the highest priority group in the `clang-format` config.

Some UDIM texture utilities also moved from `glf` to `hdSt`, so the usage of that utility in `vp2RenderDelegate/material.cpp` was updated accordingly.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyDrawOverride.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyDrawOverride.cpp
@@ -13,7 +13,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+
+// GL loading library needs to be included before any other OpenGL headers.
+// We also disable clang-format for this block, since otherwise v10.0.0 fails
+// to recognize that "ProxyDrawOverride.h" is the related header.
+// clang-format off
+#include <pxr/pxr.h>
+#if USD_VERSION_NUM < 2102
 #include <pxr/imaging/glf/glew.h>
+#else
+#include <pxr/imaging/garch/glApi.h>
+#endif
+// clang-format on
 
 #include "AL/usdmaya/nodes/ProxyDrawOverride.h"
 
--- a/.clang-format
+++ b/.clang-format
@@ -43,9 +43,9 @@ IncludeCategories:
   # 8. C system headers
   # 9. Conditional includes
 
-  # 0. Glew must be included before any other GL header
+  # 0. GL loaders must be included before any other GL header
   # Negative priority puts it above the default IncludeIsMainRegex
-  - Regex:           '<pxr/imaging/glf/glew.h>'
+  - Regex:           '<pxr/imaging/(garch/glApi.h|glf/glew.h)>'
     Priority:        -1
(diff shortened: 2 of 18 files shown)


plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyDrawOverride.cpp
```diff
@@ -13,7 +13,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+
+// GL loading library needs to be included before any other OpenGL headers.
+// We also disable clang-format for this block, since otherwise v10.0.0 fails
+// to recognize that "ProxyDrawOverride.h" is the related header.
+// clang-format off
+#include <pxr/pxr.h>
```
kxl-adsk, line 21: Is this include order intentional?

mattyjams: It is, yeah. Internally, we define `USD_VERSION_NUM=PXR_VERSION`, so we need to include `pxr.h` first so that `USD_VERSION_NUM` resolves to a valid value. I had left a comment to that effect in an earlier PR, but didn't apply the same comment here:
https://github.com/Autodesk/maya-usd/blob/51d22f04f685253fe54e7429de1d0a4be49e3b80/lib/mayaUsd/render/mayaToHydra/utils.cpp#L17

The eventual goal would be that we could drop the maya-usd specific `USD_VERSION_NUM` and just use `PXR_VERSION` everywhere, since that's provided by USD. Now that I'm looking at it again, I think we'd actually be ready to do that now. @pmolodo fixed `PXR_VERSION` to make it a useable value for version checks in https://github.com/PixarAnimationStudios/USD/pull/886, which made it into core USD 19.11, so with that as the current minimum core USD version, it should be safe to switch over.

kxl-adsk: This reminds me https://github.com/Autodesk/maya-usd/issues/645

kxl-adsk (approved):

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyDrawOverride.cpp
mattyjams: > This reminds me #645

Ah, yup, well referenced. :)

PR #1026, opened 2020-12-22
Remove unused shebang.py script, and /pxrythonsubst

pmolodo:
`shebang.py` was leftover cruft from the Pixar repository, that was no longer used.

It did a find/replace on the `#!/pxrpythonsubst` lines in the test scripts, but since `shebang.py` is not run anymore, I just replaced these with `#!/usr/bin/env mayapy`.

The project never makes use of the shebang line anyway, since we support windows, so we always explcitly call mayapy with the script as an arg... so we could eliminate it altogether.  But by leaving it as mayapy, it may at least be occasionally useful for testers running things manually... and at least documents that it should be run with mayapy.


kxl-adsk (approved):

mattyjams (approved):
Thanks for doing this @pmolodo! I had had this on my to-do list for a while, but never got around to it.

The change looks good to me!

Question though, maybe for @kxl-adsk: Does `mayapy` support Python version switching the same way `maya` does? Looking at the `mayapy` and `mayapy2` script executables on Linux, and then giving it a try, it doesn't look to me like it does:

```
> env MAYA_PYTHON_VERSION=2 <Maya PR 121 installation>/bin/mayapy --version
Python 3.7.7
```

So it is handy to be able to still run the test scripts directly with `mayapy` in the shebang, but you'll be forced into using Python 3. I don't know if that's something we want/need to address or just a gotcha to be aware of.

kxl-adsk:
I had exactly the same question regarding python version before approving this PR. I don’t believe people will run these test with shebang and python 2 is going away. 

Your question will have to wait till next year.

pmolodo:
Yeah, given that the shebang is essentially "unsupported", as far as this project is concerned, AND mayapy2 is a temporary construct, I think it's probably not worth worrying about.

PR #1027, opened 2020-12-22
Add UFE path handling and utilities to support paths that identify point instances of a PointInstancer

mattyjams:
This PR represents all of the plumbing necessary to work with UFE paths that identify individual instances of a `PointInstancer` prim. It was distilled from #936 as well as some offline discussion. This PR stops short of making the Viewport 2.0 render delegate point instance selection-aware. That and manipulation will be added in subsequent PRs.

The path to a point instance is constructed by simply appending the instance's `instanceIndex` as a final component of the USD `Ufe::PathSegment`, for example: `/PointInstancer/12345`. This is a valid path in UFE, but it is **not** a valid `SdfPath`, since all elements of an `SdfPath` need to be C identifiers and cannot start with a digit. We take advantage of this fact to quickly identify `Ufe::Paths` to point instances and avoid having to do too much string manipulation. Utilities were added and wrapped to Python to handle adding or stripping the `instanceIndex` component when necessary. All of the path manipulation specifics about how point instance paths are constructed or parsed should be consolidated in `lib/mayaUsd/ufe/Utils.cpp`, so if we choose to change the path representation or apply a similar pattern to other prim component types (e.g. individual points/edges/faces of a `Mesh` prim), we should hopefully only have one place we'll need to change it.

Test USD files and Maya scenes with proxies that use them were added at three different scales: 14 instances, 7,000 instances, and 70,000 instances. The test added in this PR only makes use of the 14 instance sample, but the imaging and manipulation tests to be added in subsequent PRs will also use the large scale samples.


test/lib/ufe/testPointInstances.py
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python
```
mattyjams, line 1: Just noticed I need to update this to conform to #1026.

mattyjams: Addressed in rebased commit 5d6e227ac8e6fa8df7a37e40c58d0cee6dc0e137.

mattyjams:
Just rebased to resolve conflicts from the merge of #1039.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -76,25 +85,95 @@ UsdStageWeakPtr getStage(const Ufe::Path& path) { return g_StageMap.stage(path);
...
+
+    const UsdPrim usdPrim = ufePathToPrim(path);
+    if (!usdPrim || !usdPrim.IsA<UsdGeomPointInstancer>()) {
+        return instanceIndex;
+    }
+
+    // Once more as above in usdPathToUfePathSegment() and
+    // stripInstanceIndexFromUfePath(), a path component at the tail of the
+    // path that begins with a digit is assumed to represent an instance index.
+    const std::string& tailComponentString = path.back().string();
```
ppt-adsk, line 168: Might want to factor out the test as an (inline) predicate in the unnamed namespace in this file, and use it in both places.

lib/mayaUsd/ufe/wrapUtils.cpp
```diff
@@ -81,36 +88,103 @@ std::string stagePath(UsdStageWeakPtr stage)
...
+{
+#ifdef UFE_V2_FEATURES_AVAILABLE
+    return ufe::ufePathToPrim(Ufe::PathString::path(ufePathString));
+#else
+    Ufe::Path       path = _UfeV1StringToUsdPath(ufePathString);
+
+    // If there are fewer than two segments, there cannot be a USD segment, so
+    // return an invalid UsdPrim.
+    const size_t numSegments =
+#if UFE_PREVIEW_VERSION_NUM >= 2006
```
ppt-adsk, line 152: Because this code path is UFE v1 only, you should be able to get rid of this UFE_PREVIEW_VERSION_NUM pre-processor conditional.

lib/mayaUsd/ufe/wrapUtils.cpp
```diff
@@ -81,36 +88,103 @@ std::string stagePath(UsdStageWeakPtr stage)
...
+{
+#ifdef UFE_V2_FEATURES_AVAILABLE
+    return ufe::ufePathToInstanceIndex(Ufe::PathString::path(ufePathString));
+#else
+    Ufe::Path path = _UfeV1StringToUsdPath(ufePathString);
+
+    // If there are fewer than two segments, there cannot be a USD segment, so
+    // return ALL_INSTANCES.
+    const size_t numSegments =
+#if UFE_PREVIEW_VERSION_NUM >= 2006
```
ppt-adsk, line 175: Same as line 152.

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -117,14 +120,15 @@ Ufe::SceneItemList UsdHierarchy::createUFEChildList(const UsdPrimSiblingRange& r
     // Return UFE child list from input USD child list.
     Ufe::SceneItemList children;
     for (const auto& child : range) {
-        children.emplace_back(UsdSceneItem::create(fItem->path() + child.GetName(), child));
```
ppt-adsk, line 120: Not sure I understand why we need the stripInstanceIndexFromUfePath() call.  Is it the case that a point instancer instance can have children?  Because here we're in the parent, collecting children from USD.

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -117,14 +120,15 @@ Ufe::SceneItemList UsdHierarchy::createUFEChildList(const UsdPrimSiblingRange& r
...
+        children.emplace_back(
+            UsdSceneItem::create(stripInstanceIndexFromUfePath(path()) + child.GetName(), child));
     }
     return children;
 }
 
 Ufe::SceneItem::Ptr UsdHierarchy::parent() const
 {
-    return UsdSceneItem::create(fItem->path().pop(), prim().GetParent());
+    return UsdSceneItem::create(stripInstanceIndexFromUfePath(path()).pop(), prim().GetParent());
```
ppt-adsk, line 131: Not sure I understand this one either.  If you have a point instancer index UFE scene item, by pop'ing its tail, you will still get the parent, no?

And therefore if you do both a stripInstanceIndexFromUfePath() AND a pop, you will have removed one too many path components?

Also, I'm unsure about the call to prim().GetParent().  If I'm a point instancer instance, is that the correct call to make?

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -144,7 +148,7 @@ Ufe::AppendedChild UsdHierarchy::appendChild(const Ufe::SceneItem::Ptr& child)
     auto           stage = childPrim.GetStage();
     auto           ufeSrcPath = usdChild->path();
     auto           usdSrcPath = childPrim.GetPath();
-    auto           ufeDstPath = fItem->path() + childName;
+    auto           ufeDstPath = stripInstanceIndexFromUfePath(path()) + childName;
```
ppt-adsk, line 151: Again, not sure from a data model perspective why we need the strip call.

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -216,11 +220,11 @@ Ufe::SceneItem::Ptr UsdHierarchy::defaultParent() const
 {
     // Default parent for USD nodes is the pseudo-root of their stage, which is
     // represented by the proxy shape.
-    auto path = fItem->path();
+    Ufe::Path ufePath = stripInstanceIndexFromUfePath(path());
```
ppt-adsk, line 223: If we're going to pop off the whole tail segment, no point in calling stripInstanceIndexFromUfePath().

test/lib/ufe/testPointInstances.py
```diff
@@ -0,0 +1,113 @@
...
+        # Create a UFE path to a PointInstancer prim with an instanceIndex on
+        # the end. This path uniquely identifies a specific point instance.
+        ufePath = ufe.Path([
+            mayaUtils.createUfePathSegment('|UsdProxy|UsdProxyShape'),
+            usdUtils.createUfePathSegment('/PointInstancerGrid/PointInstancer/7')])
+        ufeItem = ufe.Hierarchy.createItem(ufePath)
+
+        # Verify that we turned the path into a scene item correctly and that
+        # we can access the PointInstancer prim.
+        self.assertEqual(ufeItem.nodeType(), 'PointInstancer')
```
ppt-adsk, line 58: So this means that the nodeType() of PointInstancer/7 and PointInstancer are the same?  Is that awkward, or is that exactly what we want?

test/lib/ufe/testPointInstances.py
```diff
@@ -0,0 +1,113 @@
...
+        ufeAttrs = ufe.Attributes.attributes(ufeItem)
+        self.assertTrue(ufeAttrs.hasAttribute(UsdGeom.Tokens.positions))
+
+        # XXX: Note that UsdSceneItem currently has no wrapping to Python, so
+        # we are not able to verify that we can query the scene item directly
+        # for its instance index. We can exercise the Utils functions though.
+
+        # XXX: Another note, would be nice here to be able to do
+        # ufePath.string() or str(ufePath) to produce this path string instead.
+        ufePathString = ','.join([str(segment) for segment in ufePath.segments])
```
ppt-adsk, line 69: Just call ufe.PathString.string(ufePath)
but I agree with you nonetheless.

test/lib/ufe/testPointInstances.py
```diff
@@ -0,0 +1,113 @@
...
+        self.assertEqual(instanceIndex, 7)
+
+        # Tests for the usdPathToUfePathSegment() utility function.
+        ufePathSegmentString = mayaUsdUfe.usdPathToUfePathSegment(prim.GetPrimPath())
+        self.assertEqual(ufePathSegmentString, '/PointInstancerGrid/PointInstancer')
+
+        ufePathSegmentString = mayaUsdUfe.usdPathToUfePathSegment(prim.GetPrimPath(), 7)
+        self.assertEqual(ufePathSegmentString, '/PointInstancerGrid/PointInstancer/7')
+
+        # XXX: Similar to the issue above, Ufe::Path::string() yields a string
```
ppt-adsk, line 85: Is the comment still relevant?  You're not creating scene items in lines 88-91.

test/lib/ufe/testPointInstances.py
```diff
@@ -0,0 +1,113 @@
...
+        self.assertEqual(
+            ufePrimPathString,
+            '|world|UsdProxy|UsdProxyShape/PointInstancerGrid/PointInstancer')
+
+        # Create a path to the PointInstancer itself and verify that the
+        # instanceIndex is ALL_INSTANCES.
+        ufePath = ufe.Path([
+            mayaUtils.createUfePathSegment('|UsdProxy|UsdProxyShape'),
+            usdUtils.createUfePathSegment('/PointInstancerGrid/PointInstancer')])
+        ufePathString = ','.join([str(segment) for segment in ufePath.segments])
```
ppt-adsk, line 98: ufePathString = ufe.PathString.string(ufePath)

test/lib/ufe/testPointInstances.py
```diff
@@ -0,0 +1,113 @@
...
+
+        instanceIndex = mayaUsdUfe.ufePathToInstanceIndex(ufePathString)
+        self.assertEqual(instanceIndex, -1)
```
ppt-adsk, line 101: Would there be extra kosher-ness if the comparison was with ALL_INSTANCES?

test/lib/ufe/testPointInstances.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Create a path to a non-PointInstancer prim but append an instance
+        # index to it. This isn't really a valid path, but the utility function
+        # should recognize this case and return an instanceIndex of
+        # ALL_INSTANCES.
+        ufePath = ufe.Path([
+            mayaUtils.createUfePathSegment('|UsdProxy|UsdProxyShape'),
+            usdUtils.createUfePathSegment('/PointInstancerGrid/7')])
+        ufePathString = ','.join([str(segment) for segment in ufePath.segments])
```
ppt-adsk, line 110: ufe.PathString.string()

ppt-adsk (commented):
Looks simple enough!  There are most likely some things where I don't completely understand the UFE parenting model for point instancer instances (pointed out in code comments), but otherwise seems like a good data model place to start!

lib/mayaUsd/ufe/Utils.cpp
mattyjams: Good idea. Done.

lib/mayaUsd/ufe/UsdHierarchy.cpp
mattyjams: Yup, good catch. Fixed!

lib/mayaUsd/ufe/UsdHierarchy.cpp
mattyjams: Hmm, I may be misunderstanding when some of this `UsdHierarchy` API gets invoked then. I was assuming that we might get into situations where this instance of `UsdHierarchy` has an `fItem` that represents a point instance, but maybe that's not actually possible?

When a `UsdSceneItem` represents a point instance, its `prim()` member will return the `PointInstancer` prim that created the instance. There is no representation of the instance itself anywhere in USD. As a result, I'm not sure that it makes sense to think of a point instance as having a parent. Instead, there's a `PointInstancer` prim that created it.

So my goal then in adding these `stripInstanceIndexFromUfePath()` calls was to ensure that if we somehow are in a case where this `UsdHierarchy`'s `fItem` is a point instance, we **always** work in terms of the `PointInstancer` prim instead when computing parents or children, since the point instance really has neither and is more a product of imaging.

Maybe these overrides should be detecting when `fItem` is a point instance and return `nullptr` or  empty lists or the equivalent instead?

lib/mayaUsd/ufe/Utils.cpp
mattyjams: Addressed in dde6e1e46df178891d348b4d645cb4395b7546ff and 5a2f62f2ba8779ac48a339e524f14315eec23fec.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -76,25 +85,95 @@ UsdStageWeakPtr getStage(const Ufe::Path& path) { return g_StageMap.stage(path);
 
 Ufe::Path stagePath(UsdStageWeakPtr stage) { return g_StageMap.path(stage); }
 
+Ufe::PathSegment usdPathToUfePathSegment(const SdfPath& usdPath, int instanceIndex)
+{
+    const Ufe::Rtid usdRuntimeId = getUsdRunTimeId();
+    const char      separator = SdfPathTokens->childDelimiter.GetText()[0u];
```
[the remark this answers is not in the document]
mattyjams: Addressed in 66d869aea2c482dce0e24ee1299eb05689daedf6.

lib/mayaUsd/ufe/wrapUtils.cpp
mattyjams: Addressed in 34329b23f94983b0fb48c1104f197b99a36caf24.

lib/mayaUsd/ufe/wrapUtils.cpp
mattyjams: Addressed in 34329b23f94983b0fb48c1104f197b99a36caf24.

lib/mayaUsd/ufe/UsdHierarchy.cpp
mattyjams: Reverted these changes in rebased commit 1ced6783c1524a5a9fa82446bd33193642641e87.

test/lib/ufe/testPointInstances.py
mattyjams: Addressed in 32299f7a276e9e0040b5fd06890515509012f204.

test/lib/ufe/testPointInstances.py
mattyjams: Addressed in 1d344b1db3abc06e9b58396c0e181b49ece3e979 and 32299f7a276e9e0040b5fd06890515509012f204.

test/lib/ufe/testPointInstances.py
mattyjams: Addressed in 32299f7a276e9e0040b5fd06890515509012f204.

test/lib/ufe/testPointInstances.py
mattyjams: Addressed in 32299f7a276e9e0040b5fd06890515509012f204.

lib/mayaUsd/ufe/UsdHierarchy.cpp
ppt-adsk: O.K., let me see if I have my head wrapped around the UFE data model for point instancers and point instances --- and I still have a question :)
- A point instance UFE scene item has no parent, and has no children.  Not having a parent is a departure from existing UFE practice, which may uncover latent bugs, but retrieving a parent should go through the Ufe::Hierarchy interface, which you'll be defining.  Returning a nullptr parent and an empty child list is perfect.
- The prim of a point instance UFE scene item is its point instancer prim.  This makes sense.
- What are the children of a point instancer?
We should make sure all of this is as obvious as possible in the implementation and comments.
-

test/lib/ufe/testPointInstances.py
ppt-adsk: Given your other explanations, this makes sense.

test/lib/ufe/testPointInstances.py
ppt-adsk: Agreed.  Maybe this needs a quick code comment?

lib/mayaUsd/ufe/UsdHierarchy.cpp
mattyjams: Yeah, all good insights and questions. Hopefully my understanding is starting to congeal here too. :)

On the USD side, when we're talking about parents or children, we are virtually always talking about **namespace** parents or **namespace** children. These would be defined prims with unique locations in USD namespace. Point instances have no such location in USD namespace. They are instead generated based on data members of a `PointInstancer` prim and a "prototype" that is being instanced, where the prototype refers to some location in USD namespace which may or may not be a hierarchy of prims. In terms of USD, a point instance itself doesn't really have either a parent or children then. A `PointInstancer` is an actual prim with namespace parents, so it clearly has a parent, but the point instances that it generates would **not** be considered its children. It's common for `PointInstancer` prims to have a `prototypes` scope as a child where all of its prototypes are defined, so it's really only proper USD prims that would be considered children of a `PointInstancer`.

On the UFE side, I think carrying over the child-less property of point instances from USD makes sense, so there is no way to traverse down from a point instance. It **is** useful to be able to get from a point instance to the `PointInstancer` that generated it though, in particular for pick-walking, so I think it does make sense to have UFE report that the "parent" of a point instance is its `PointInstancer`. For the `PointInstancer` itself, their hierarchy should be treated the same way as any other non-`PointInstancer` prim, where only USD namespace parents and children are considered UFE parents and children.

I re-worked this a bit and tried to make this a bit clearer with some additional comments.

lib/mayaUsd/ufe/UsdHierarchy.cpp
mattyjams: Good note here. I think you're right that the double-pop here is **not** what we want for point instances, which hopefully jibes with my explanation above.

lib/mayaUsd/ufe/UsdHierarchy.cpp
mattyjams: I think you're right here too, though I'm admittedly not sure about the proper way to handle this when point instances and `PointInstancer` prims are either the parent or the child. This function is also UFE v1 only, so I'm not sure if it would still come into play?

test/lib/ufe/testPointInstances.py
mattyjams: Yup, added a comment here to explain this.

lib/mayaUsd/ufe/UsdHierarchy.cpp
mattyjams: I took another crack at this in 95a4d4174531e488facd72c1789ebef3434ff469. Let me know what you think!

I also added a pick-walking test in 9c5099586fc182b547eb306d728329634e2de3bc.

lib/mayaUsd/ufe/UsdHierarchy.cpp
mattyjams: Reverted these changes as part of 95a4d4174531e488facd72c1789ebef3434ff469.

lib/mayaUsd/ufe/UsdHierarchy.cpp
mattyjams: Reverted these changes as part of 95a4d4174531e488facd72c1789ebef3434ff469, though I'm not certain whether we need some kind of handling for append/insert when `PointInstancer` and/or point instances are involved as either the parent or the child?

test/lib/ufe/testPointInstances.py
mattyjams: Added a comment in 9c5099586fc182b547eb306d728329634e2de3bc.

test/lib/ufe/testPointInstances.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # Pick-walk up from the point instance.
+        cmds.pickWalk(direction='up')
+
+        # The selection should have been replaced with the PointInstancer
+        ufeItem = globalSelection.front()
+        ufePathString = ufe.PathString.string(ufeItem.path())
+        self.assertEqual(
+            ufePathString,
+            '|UsdProxy|UsdProxyShape,/PointInstancerGrid/PointInstancer')
```
ppt-adsk, line 140: Nice.

ppt-adsk (commented):
Just one change where you can add a point instancer and point instance UFE data model section to UsdSceneItem, and I think we're done!

lib/mayaUsd/ufe/UsdHierarchy.cpp
mattyjams: Sounds good. I just added some documentation to UsdSceneItem.h to capture everything we've been discussing here. You should see that in a71c3dd30a13c12c96ebe076bca90779f299fe37. Give it a read through and let me know what you think.

I added that in a rebase (apologies, I know that's been somewhat complicating reviewing this), so although the commits were rewritten, nothing else was changed.

ppt-adsk: Love it, very thorough, thanks.  Now the only challenge is to get people to RTFM, but we have something to point them to.

ppt-adsk (approved):

Issue #972, opened 2020-12-04
[MAYA-107682] usdz export writes incorrect texture paths


dgovil:
**Describe the bug**

This is an issue that builds on https://github.com/Autodesk/maya-usd/issues/795.

The resulting usdz file now does contain the texture, but it now contains the entire full path hierarchy to the file instead.
See this screenshot showing the usdz files hierarchy. Note, this could potentially be revelatory for certain people if the hierarchy included project names etc...

The usdz file is attached : 
[asset.usdz.zip](https://github.com/Autodesk/maya-usd/files/5645328/asset.usdz.zip)

 
![image](https://user-images.githubusercontent.com/4443090/101211216-a84c5300-362b-11eb-9ede-9f07e2377535.png)

The resulting usd file inside the usdz also does not point to the correct texture path, instead pointing to the path as it was in the Maya file. 

```python
def Shader "file1"
{
    uniform token info:id = "UsdUVTexture"
    asset inputs:file = @../sourceimages/checkerboard.png@
}
```

The workaround of disabling the block of code mentioned in https://github.com/Autodesk/maya-usd/issues/795 works to let us workaround this behaviour.

While the change in the https://github.com/Autodesk/maya-usd/pull/865 partially fixed the original issue so the texture is included, it actually creates a different issue as described here. Apologies for not catching this sooner, as the change in the PR did satisfy a specific part of our pipeline that obscured the second part of this issue. Testing in a fully vanilla environment exposes this next bit which is still problematic.

**Steps to reproduce**
1. Open Maya project from https://github.com/Autodesk/maya-usd/issues/795
2. Export usdz file
3. Expand the usdz file and look at the resulting file and USD texture reference path as described above.

**Expected behavior**
I would expect the texture to be written such that it's using the same relative path as the maya material does, just inlined into the Maya file.
If the texture lives outside the Maya workspace, then it would be best to keep the old behaviour where it was stored in `0/textureName.jpg` etc....

**Specs (if applicable):**
 - macOS Catalina and Big Sur
 - Compiler version: Apple clang version 12.0.0 (clang-1200.0.32.5) (reproducible on older clang versions too)
 - Maya 2018.5 and 2020.0
 - Maya USD commit SHA ddcd319b7785a8cc586b3106ecd16cba25f9cb3c
 - Pixar USD commit SHA d8a405a1344480f859f025c4f97085143efacb53

JGamache-autodesk:
[...]

PR #1028, opened 2020-12-23
Fix texture writer path resolver to work relative to final export path

dgovil:
This PR fixes an issue where the texture writer in the exporter expects the stage being written to be the final export path.
`.usdz` files however switch the exporter to write to a temporary location. Therefore prim writers etc do not have access to the final path for their logic. This change makes it so that the fileName is part of the export args, to be available to the rest of the exporter code base.

This addresses the issue in https://github.com/Autodesk/maya-usd/issues/972 and should make usdz exports once again compatible with AR QuickLook on iOS, and preview on Mac.

Also I realize everyone is on, or will be on, holiday break, so no rush on responding to this PR. I just wanted to post it before our break starts too. Hope everyone has a great holiday season.


dgovil:
Oops, missed some arnold attributes when cleansing the maya file. Updated PR with those removed.

lib/mayaUsd/commands/baseExportCommand.cpp
```diff
@@ -205,7 +205,7 @@ MStatus MayaUSDExportCommand::doIt(const MArgList& args)
         }
 
         // Read all of the dictionary args first.
-        const VtDictionary userArgs = UsdMayaUtil::GetDictionaryFromArgDatabase(
+        VtDictionary userArgs = UsdMayaUtil::GetDictionaryFromArgDatabase(
```
mattyjams, line 208: I'm a little wary of dropping the `const` here, as I think we want to make sure we maintain a store that's guaranteed to contain the actual arguments the user passed in. More below...

lib/mayaUsd/commands/baseExportCommand.cpp
```diff
@@ -231,6 +231,8 @@ MStatus MayaUSDExportCommand::doIt(const MArgList& args)
             if (fileName.empty()) {
                 fileName = tmpVal.asChar();
             }
+
+            userArgs["fileName"] = fileName;
```
mattyjams, line 235: It feels a bit weird to be stuffing a (likely) processed string value back into `userArgs`, since it may very well be different from what the user actually specified. The other wrinkle here is that we know we have some work to do to properly support USD's `Ar` path resolvers, and we're baking in a resolved path that uses Maya's resolver.

What do you think of adding a `GetResolvedFileName()` method to `UsdMayaJobExportArgs` that does this resolution and returns the result? Something like:

```C++
std::string UsdMayaJobExportArgs::GetResolvedFileName() // declared const in the header
{
    MFileObject fileObj;
    fileObj.setRawFullName(fileName);

    // Make sure it's an absolute path
    fileObj.setRawFullName(fileObj.resolvedFullName());
    const std::string resolvedFileName = absoluteFile.resolvedFullName().asChar();

    if (!resolvedFileName.empty()) {
        return resolvedFileName;
    }

    return fileName;
}
```

We could then replace the `if` block with just this:

```C++
if (argData.isFlagSet(kFileFlag)) {
    fileName = userArgs.GetResolvedFileName();
}
```

And we would also call `GetResolvedFileName()` in the file texture writer. That way we have the original, unmodified `fileName` argument available if we need it for something, and we also have the path resolution stuff separated out so we can revisit it when we get to the `Ar` work.

lib/usd/translators/shading/usdFileTextureWriter.cpp
```diff
@@ -279,7 +279,10 @@ void PxrUsdTranslators_FileTextureWriter::Write(const UsdTimeCode& usdTime)
     //          minimal interop. It will be replaced by proper use of Maya and
     //          USD asset resolvers. For package files, the exporter needs full
     //          paths.
-    const std::string& fileName = GetUsdStage()->GetRootLayer()->GetRealPath();
+
+    // We use the ExportArgs fileName here instead of the USD root layer path
+    // to make sure that we are basing logic of the final export location
+    const std::string& fileName = _GetExportArgs().fileName;
```
mattyjams, line 285: We'll want to drop the `&` making this a reference if we call `GetResolvedFileName()` here.

mattyjams (changes requested):
The changes make sense to me, but I think we want to try to keep the `userArgs` dictionary in a state where it contains the exact arguments the user passed in. I had a suggestion of adding a new public method to `UsdMayaJobExportArgs` that would hopefully make that work. Let me know what you think!

lib/mayaUsd/commands/baseExportCommand.cpp
dgovil: Yeah that's a fair call. I figured it would be representative of the users fileName argument to the exporter, which is why I was putting it into the userArgs, but you're right that it is mutated so isn't exactly 1:1.

I'll make the necessary changes this week and update the PR

dgovil: Updated the PR with the changes.
The UsdMayaJobExportArgs constructors now take fileName as an argument to instantiate it, and added a GetResolvedFileName method to return the resolved file name to the caller. 

I believe that should address the concerns you raised, but let me know if there's anything else I missed.

JGamache-autodesk (approved):

lib/mayaUsd/commands/baseExportCommand.cpp
mattyjams: Cool, that's looking better already.

I think I was imagining though that we'd keep your original changes to jobArgs.h and jobArgs.cpp that add `fileName` as a public token and member of the struct. That way, the default dictionary would have a `fileName` key in it so that when we call `GetDictionaryFromArgDatabase()`, the returned `userArgs` dictionary has that key and its value from the Maya args in it:
https://github.com/Autodesk/maya-usd/blob/dcd4094f5353d6be96fb7b61b17a41686b58ad5a/lib/mayaUsd/commands/baseExportCommand.cpp#L208

We then don't need to plumb the `fileName` down through the `UsdMayaJobExportArgs` functions separately:
https://github.com/Autodesk/maya-usd/blob/dcd4094f5353d6be96fb7b61b17a41686b58ad5a/lib/mayaUsd/commands/baseExportCommand.cpp#L288

We'd basically be treating the `fileName` arg the same as we do every other arg that can be specified to the command, and just have an additional function on `UsdMayaJobExportArgs` that processes it into a resolved value. With the most recent changes, nothing has access to the unresolved `fileName` value as far as I can tell?

dgovil: Ah I see what you're saying. I'll have to figure out how to get the file attribute working in GetDictionaryFromArgDatabase. Since I was setting it externally to the userArgs earlier, it seems like it never actually was getting set up here.

dgovil: Okay pushed a new update. I've renamed the `fileName` token to `file` to stay close to the input arguments, and have that publicly available now.

mattyjams: Ahh, sorry for the confusion. I missed the `fileName`/`file` disconnect there. But yeah, it's looking great to me now. Thanks for doing that!

mattyjams (approved):
Looks good to me now! Thanks @dgovil!

PR #1035, opened 2021-01-04
MAYA-106780  Layer Editor "Any Edits" query flag

fowlertADSK:
Adding ability to register node id's with the Layer Editor so that Maya can determine that there are unsaved changes
Added a new flag on the Layer Editor command that Maya can query and then warn a user before resetting the scene.

### DIFF
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -16,16 +16,20 @@
 
 #include "layerEditorCommand.h"
 
+#include <mayaUsd/nodes/proxyShapeBase.h>
 #include <mayaUsd/utils/query.h>
 
 #include <pxr/base/tf/diagnostic.h>
 
 #include <maya/MArgList.h>
 #include <maya/MArgParser.h>
+#include <maya/MFnDependencyNode.h>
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -46,6 +50,11 @@ const char kAddAnonSublayerFlagL[] = "addAnonymous";
 const char kMuteLayerFlag[] = "mt";
 const char kMuteLayerFlagL[] = "muteLayer";
 
+const char kAnyDirtyLayerFlag[] = "ad";
+const char kAnyDirtyLayerFlagL[] = "anyDirty";
+
+std::set<unsigned int> _supportedTypesForDirtyQuery;
```
seando-adsk, line 56: Since the method that add/remove from this set are static shouldn't this also be a static data member?

fowlertADSK: I don't think it's required because this is declared in the unnamed namespace?

seando-adsk (approved):

kxl-adsk (changes requested):
Please rework the check so it works for any proxy shape derived from the base class.

Issue #1005, opened 2020-12-15
custom compile of v0.6.0 : Layer Editor Window does not open or the Create Stage From File Menu Item


cfmoore007:
**Describe the bug**
Compiling v0.6.0 or dev@c411bab9 is having menu items issue for me.

In particular, the Create->Universal Scene Description->Create Stage From File option results in an error : // Error: file: F:/APPLICATIONS/Autodesk/Maya2021/scripts/others/getPluginResource.mel line 66: Plug-in resource lookup failed, resource ("mayaUsdPlugin","kCreate") is not registered

Secondly, the Layer Editor does not open, but does not complain about anything ( i.e : no errors or warnings ).  I tested this by using the 'Create Stage From New Layer' option.  I verified that the layerEditor is under the lib/usd/ui/ directory.

Finally, I tried the download of v0.6.0 using the installer.  This version works fine.  The issue seems limited to my custom build.  The last working build I tried was dev@37ea8e0.

**Steps to reproduce**
( Create Stage From File )
1. Create->Universal Scene Description->Create Stage From File
2. see Plug-in resource lookup error

( Missing Layer Editor )
1. Create->Universal Scene Description->Create Stage From New Layer
2. Right-click on stage and select the Layer Editor
3. Witness no Layer Editor nor warning/error displayed

**Expected behavior**
The menu items work.  They did prior to this build

**Attachments**
If applicable, add screenshots, sample files, etc to help explain your problem.

**Specs (if applicable):**
 - OS & version [Windows 10]
 - Compiler & version [VS2017]
 - Maya version [e.g. Maya PR 121]
 - Maya USD commit SHA [dev at c411bab9 ]
 - Pixar USD commit SHA [v20.11]

**Additional context**
N/A

fowlertADSK:
@cfmoore007   I just synced back to the dev branch and was kicking off a build against the latest Maya PR, so I'll try to see if I can reproduce the problem.  We did add some string resource handling in a recent change.  The plug-in isn't localized at the moment (English only) but it's now set up to be able to at some point in the future.

fowlertADSK:
@cfmoore007   I downloaded the latest Maya beta + devkit and built the exact same SHA as you and can't reproduce a problem.  I'm guessing you haven't changed anything about how you build the plug-in have you?  

And what's your .mod file look like, any difference between what you're using for your build vs. the installed one?

cfmoore007:
Nothing seems out of the ordinary here : 

---
[...]

PR #1037, opened 2021-01-05
Cleanup UI folder

seando-adsk:
#### Cleanup UI folder
* Suppress TBB deprecated messages.
* Find proper debug TBB lib.
* When Layer Editor is not built (i.e. Qt not found) do not add the menu or register command.
* Moved importDialog into subfolder.


cmake/compiler_config.cmake
```diff
@@ -107,4 +107,10 @@ function(mayaUsd_compile_config TARGET)
                 ${MSVC_DEFINITIONS}
         )
     endif()
+
+    # Remove annoying TBB warnings.
+    target_compile_definitions(${TARGET}
+        PRIVATE
+            TBB_SUPPRESS_DEPRECATED_MESSAGES
+    )
```
seando-adsk, line 115: I got tired of seeing hundreds of TBB warnings polluting our build log. This started with the update of TBB in Maya.

cmake/modules/FindMaya.cmake
```diff
@@ -205,11 +205,17 @@ foreach(MAYA_LIB
...
     Foundation
     IMFbase
-    tbb
     cg
     cgGL
-    clew)
+    clew
+)
+if (CMAKE_BUILD_TYPE MATCHES Debug)
+    list(APPEND MAYA_LIBS_TO_FIND tbb_debug)
```
seando-adsk, line 213: I compile in debug all the time and just happened to notice one day that this one wasn't being found correctly in debug. We aren't using it in MayaUsd, which is why it never caused a problem.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -50,8 +50,10 @@ namespace {
 // - the "Image" is used for icon in the context menu. Directly used std::string
 //   for these so the emplace_back() will choose the right constructor. With char[]
 //   it would convert that param to a bool and choose the wrong constructor.
+#ifdef WANT_QT_BUILD
 static constexpr char    kUSDLayerEditorItem[] = "USD Layer Editor";
 static constexpr char    kUSDLayerEditorLabel[] = "USD Layer Editor...";
+#endif
```
seando-adsk, line 56: When we don't have Qt, we don't build the Layer Editor so it should not be available in any menu.

lib/usd/ui/CMakeLists.txt
```diff
@@ -23,23 +23,14 @@ set(CMAKE_AUTOUIC ON)
 add_library(${PROJECT_NAME} SHARED)
 
 add_subdirectory(layerEditor)
+add_subdirectory(importDialog)
```
seando-adsk, line 26: With the addition of the Layer Editor, the "ui" folder now contains more than just the import dialog. So I moved the import dialog code to a sub-folder.

lib/usd/ui/CMakeLists.txt
```diff
@@ -143,5 +125,5 @@ install(FILES ${HEADERS}
 # -----------------------------------------------------------------------------
 if (BUILD_TESTS)
     # A simple executable used to test the import dialog.
-    add_subdirectory(demo)
+    add_subdirectory(importDialogDemo)
```
seando-adsk, line 128: I just renamed this folder to make it more obvious what it is.

plugin/adsk/plugin/plugin.cpp
```diff
@@ -208,7 +208,9 @@ MStatus initializePlugin(MObject obj)
     registerCommandCheck<MayaUsd::ADSKMayaUSDImportCommand>(plugin);
     registerCommandCheck<MayaUsd::EditTargetCommand>(plugin);
     registerCommandCheck<MayaUsd::LayerEditorCommand>(plugin);
+#if defined(WANT_QT_BUILD)
     registerCommandCheck<MayaUsd::LayerEditorWindowCommand>(plugin);
+#endif
```
seando-adsk, line 213: When no Qt, don't register the command that is used to open the Layer Editor (since we didn't build the layer editor). Note: I left the two commands above in since they are built even when no Qt (they are in a different folder from the "ui/layerEditor"

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -101,15 +101,17 @@ proc initRuntimeCommands() {
...
+        if (!`runTimeCommand -exists mayaUsdOpenUsdLayerEditor`) {
+            runTimeCommand -default true
+                -label 		"USD Layer Editor"
+                -annotation "Organize and edit USD data in layers"
+                -category   "Menu items.Common.Windows.General Editors"
+                -command 	"mayaUsdLayerEditorWindow mayaUsdLayerEditor"
+                -image		"USD_generic.png"
+                mayaUsdOpenUsdLayerEditor;
+        }
+    }
```
seando-adsk, line 114: Don't create the runtime command to open the LE if there is no command for the LE window.

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -165,7 +167,7 @@ global proc mayaUsdMenu_windowMenuCallback() {
 // mayaUsdMenu_windowMenuCallback
 // setup the items in Maya's "Window->General Editors" menu
 global proc mayaUsdMenu_generalEditorsMenuCallback() {
-    if (!(`menuItem -query -exists wmUsdLayerEditorMenuitem`))
+    if (`exists mayaUsdLayerEditorWindow` && !(`menuItem -query -exists wmUsdLayerEditorMenuitem`))
```
seando-adsk, line 170: This makes sure to not add the "USD Layer Editor" under the Maya "Windows" menu.

seando-adsk:
Note: this will fix the issue discovered in #1005 where they had built without Qt (so no Layer Editor), but the Layer Editor menu item was there and they got an error when clicking on it.

HamedSabri-adsk (approved):
Looking good. Thank you for doing this.

plugin/adsk/scripts/mayaUsdMenu.mel
fowlertADSK: All of these "exists" checks are for the existence of the command and not the UI right?

seando-adsk: Yes. One of my other changes is only register the "mayaUsdLayerEditorWindow" command when the Layer Editor was built (i.e. Qt is found). So if we don't have this command we won't create the runtime command to open the LE either.

fowlertADSK (approved):
LGTM

PR #1039, opened 2021-01-05
MAYA-107303 - UFE: crash (debug only) with multiple UFE items in single string

seando-adsk:
#### MAYA-107303 - UFE: crash (debug only) in MEL parser with multiple UFE items in single string

* When asking for UsdPrim from Ufe Path we cannot assume the path is valid and therefore should not assert.

### DIFF
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -50,6 +50,10 @@ template <> struct iterator_traits<MStringArray::Iterator>
 } // namespace std
 #endif
 
+namespace {
+    static constexpr auto kIllegalUSDPath = "Illegal USD run-time path %s.";
+}
+
 namespace MAYAUSD_NS_DEF {
 namespace ufe {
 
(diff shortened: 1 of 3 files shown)


test/lib/usd/translators/CMakeLists.txt
```diff
@@ -72,6 +71,15 @@ set(TEST_SCRIPT_FILES
...
 )
 
+if(BUILD_PXR_PLUGIN)
+    # This test uses the file "UsdExportUVTransforms.ma" which
+    # requires the plugin "pxrUsdPreviewSurface" that is built by the
+    # Pixar plugin.
+    list(APPEND TEST_SCRIPT_FILES
+        testUsdExportUVTransforms.py
+    )
+endif()
```
seando-adsk, line 81: I had built locally only the maya-usd plugin and got this test failing. Found it is because of the ma file it loads (there is another example of this).

seando-adsk (commented):
@ppt-adsk Changes as we discussed.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -50,6 +50,10 @@ template <> struct iterator_traits<MStringArray::Iterator>
 } // namespace std
 #endif
 
+namespace {
+    static constexpr auto kIllegalUSDPath = "Illegal USD run-time path %s.";
```
ppt-adsk, line 54: Nit-pick: since this is in the unnamed namespace, static is redundant.

ppt-adsk (approved):

test/lib/usd/translators/CMakeLists.txt
kxl-adsk: A better approach would be to port this test to use usdPreviewSurface node that is independent of Pixar plugin - see https://github.com/Autodesk/maya-usd/blob/dev/lib/usd/pxrUsdPreviewSurface/usdPreviewSurfacePlugin.cpp

seando-adsk: Let's do that in a separate PR. I copied this way of skipping the test from another place. So there are multiple tests that need to be ported.

ppt-adsk (approved):

PR #1042, opened 2021-01-06
Migrate more tests from plugin/pxr to core mayaUsd

mattyjams:
All of these tests exercise functionality that is in mayaUsd itself and not specific to the `pxrUsd` plugin or other code in `plugin/pxr`.

This is in the continuing effort to thin out the `plugin/pxr` code leaving behind only the assembly-related stuff that we intend to ultimately get rid of.


test/lib/mayaUsd/fileio/utils/testXformStack.py
```diff
@@ -15,33 +15,32 @@
...
     @classmethod
     def setUpClass(cls):
-        standalone.initialize('usd')
-        cmds.loadPlugin('pxrUsd')
+        fixturesUtils.setUpClass(__file__)
 
     @classmethod
     def tearDownClass(cls):
         standalone.uninitialize()
-
```
mattyjams, line 44: You may want to exclude commit 8e39493fd99d192f130e9d360f30c4b29255c548 or otherwise ignore whitespace when looking through this PR, since it causes GitHub to identify `testXformStack.py` as a "large diff".

mattyjams:
Rebased to resolve conflicts from the merge of #1039.

kxl-adsk (approved):
Thank you for continuing this migration, Matt. Not many comments from my side but I stopped for a second on `testProxyShape`. We want this test to only validate features of the base proxy shape class and studio-specific features should be tested in separate test files. To communicate this intent, how about we rename this test to `testProxyShapeBase` and operate on proxyShapeBase node (i.e. `mayaUsdProxyShapeBase` instead of `mayaUsdProxyShape`)

mattyjams:
Thanks @kxl-adsk! Good catch. I just rebased the changes and addressed your proxy shape base note in https://github.com/Autodesk/maya-usd/pull/1042/commits/6420c884e5353e0d1d9f237dcbcc27852d8a4a51.

kxl-adsk (approved):
LGTM, thank you!

mattyjams:
Hopefully one last amendment to this: I separated out all of the `pxrHdImagingShape`-related stuff into its own dedicated test in https://github.com/Autodesk/maya-usd/pull/1042/commits/a8cb3708e52c05633fc700d250147167c3f82107. While `mayaUsdProxyShape` will automatically create a `pxrHdImagingShape` when the Viewport 2.0 render delegate is disabled, `mayaUsdProxyShapeBase` will **not** (and should not, since it is also the base type for the Animal Logic proxy shape).

kxl-adsk (approved):
LGTM!

PR #1043, opened 2021-01-07
MAYA-108707 Share shader effect among similar-enough material networks

HdC-adsk:
Replace the authored node paths with simplified paths in the form of "node#".
By doing so we will be able to reuse shader effects among material networks
which have the same node identifiers and relationships but different node
paths, reduce shader compilation overhead and enable material consolidation
for faster rendering.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -100,6 +100,10 @@ TF_DEFINE_PRIVATE_TOKENS(
 );
 // clang-format on
 
+//! Prefix of VP2-specific simplified node path to enable reuse of shader effects among material
+//! networks which have the same node identifiers and relationships but different node paths.
+const std::string kNodePathPrefix = "node";
+
 //! Helper utility function to test whether a node is a UsdShade primvar reader.
 bool _IsUsdPrimvarReader(const HdMaterialNode& node)
 {
@@ -125,30 +129,15 @@ std::string _GenerateXMLString(const HdMaterialNetwork& materialNetwork, bool in
     std::string result;
 
     if (ARCH_LIKELY(!materialNetwork.nodes.empty())) {
-        const size_t numNodes = materialNetwork.nodes.size();
-
-        SdfPathVector paths;
-        paths.reserve(numNodes);
-
-        for (size_t i = 0; i < numNodes; i++) {
-            paths.push_back(materialNetwork.nodes[i].path);
-        }
-
-        // We use concise relative paths to generate the same string for duplicate material
-        // networks.
-        const SdfPathVector conciseRelativePaths = SdfPath::GetConciseRelativePaths(paths);
-
         // Reserve enough memory to avoid memory reallocation.
         result.reserve(1024);
 
         result += "<nodes>\n";
 
         if (includeParams) {
-            for (size_t i = 0; i < numNodes; i++) {
-                const HdMaterialNode& node = materialNetwork.nodes[i];
-
+            for (const HdMaterialNode& node : materialNetwork.nodes) {
                 result += "  <node path=\"";
-                result += conciseRelativePaths[i].GetString();
+                result += node.path.GetString();
                 result += "\" id=\"";
                 result += node.identifier;
                 result += "\">\n";
@@ -168,11 +157,9 @@ std::string _GenerateXMLString(const HdMaterialNetwork& materialNetwork, bool in
                 result += "  </node>\n";
             }
         } else {
-            for (size_t i = 0; i < numNodes; i++) {
-                const HdMaterialNode& node = materialNetwork.nodes[i];
-
+            for (const HdMaterialNode& node : materialNetwork.nodes) {
                 result += "  <node path=\"";
-                result += conciseRelativePaths[i].GetString();
+                result += node.path.GetString();
                 result += "\" id=\"";
                 result += node.identifier;
                 result += "\"/>\n";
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -220,100 +202,6 @@ std::string _GenerateXMLString(const HdMaterialNetwork& materialNetwork, bool in
     return result;
 }
 
-//! Helper utility function to apply VP2-specific fixes to the material network.
-//! - Add passthrough nodes to read vector component(s).
-//! - Fix UsdImagingMaterialAdapter issue for not producing primvar requirements.
-//! - Temporary workaround of missing support for normal map.
-void _ApplyVP2Fixes(HdMaterialNetwork& outNet, const HdMaterialNetwork& inNet)
```
HdC-adsk, line 227: The method is moved down into the class section because it is now a member of HdVP2Material. I need to do so for accessing new member of the class.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -125,30 +129,15 @@ std::string _GenerateXMLString(const HdMaterialNetwork& materialNetwork, bool in
...
-        SdfPathVector paths;
-        paths.reserve(numNodes);
-
-        for (size_t i = 0; i < numNodes; i++) {
-            paths.push_back(materialNetwork.nodes[i].path);
-        }
-
-        // We use concise relative paths to generate the same string for duplicate material
-        // networks.
-        const SdfPathVector conciseRelativePaths = SdfPath::GetConciseRelativePaths(paths);
```
HdC-adsk, line 139: Now that we've created the simplified paths which have only one unique token, we don't need to worry about name collision inside a material network so we can remove the part of the code here.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -497,6 +385,9 @@ _LoadUdimTexture(const std::string& path, bool& isColorSpaceSRGB, MFloatArray& u
 MHWRender::MTexture*
 _LoadTexture(const std::string& path, bool& isColorSpaceSRGB, MFloatArray& uvScaleOffset)
 {
+    MProfilingScope profilingScope(
```
HdC-adsk, line 388: Adding some profiler event to help identifying bottlenecks.

In the NV attic scene, the biggest bottleneck is texture loading actually which takes 54 seconds. The optimization of this commit reduces material translation and compilation, improves consolidation, but won't help with texture loading overhead. The first frame drops from the original 70s to 54s, while HdSt takes only 30+ seconds. I think we need to understand why texture loading is slower than HdSt, maybe it is a task for next release.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -796,6 +699,128 @@ change tracker for use in the first sync of this prim.
...
+    }
+
+    // Update the relationships to use the new node paths.
+    for (const HdMaterialRelationship& rel : inNet.relationships) {
+        tmpNet.relationships.push_back(rel);
+
+        HdMaterialRelationship& outRel = tmpNet.relationships.back();
+        outRel.inputId = _nodePathMap[outRel.inputId];
+        outRel.outputId = _nodePathMap[outRel.outputId];
+    }
```
HdC-adsk, line 740: These lines are new code for the optimization.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
HdC-adsk: FWIW, NV Attic scene contains 188 4k textures and a dozen of smaller textures.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
HdC-adsk: I am still running more tests. Please challenge the code to catch bugs early.

williamkrick (approved):

lib/mayaUsd/render/vp2RenderDelegate/material.h
```diff
@@ -84,13 +84,17 @@ class HdVP2Material final : public HdMaterial
...
 private:
+    void _ApplyVP2Fixes(HdMaterialNetwork& outNet, const HdMaterialNetwork& inNet);
     MHWRender::MShaderInstance* _CreateShaderInstance(const HdMaterialNetwork& mat);
     void                        _UpdateShaderInstance(const HdMaterialNetwork& mat);
     const HdVP2TextureInfo&     _AcquireTexture(const std::string& path);
 
     HdVP2RenderDelegate* const
         _renderDelegate; //!< VP2 render delegate for which this material was created
 
+    std::unordered_map<SdfPath, SdfPath, SdfPath::Hash>
```
kxl-adsk, line 95: How long is this container needed during the lifetime of the material? In other words, does it have to be a member?

kxl-adsk (approved):

lib/mayaUsd/render/vp2RenderDelegate/material.h
HdC-adsk: It is needed during the whole lifetime of the material. Each material needs the mapping info to find the converted node names which are used to get fragment parameters.

HdC-adsk:
@kxl-adsk Jerry is probably busy for his work and won't be able to review. I've done more testing and marked this ready for merge.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -796,6 +699,128 @@ change tracker for use in the first sync of this prim.
...
+
+    // Replace the authored node paths with simplified paths in the form of "node#". By doing so
+    // we will be able to reuse shader effects among material networks which have the same node
+    // identifiers and relationships but different node paths, reduce shader compilation overhead
+    // and enable material consolidation for faster rendering.
+    for (const HdMaterialNode& node : inNet.nodes) {
+        tmpNet.nodes.push_back(node);
+
+        HdMaterialNode& outNode = tmpNet.nodes.back();
+        outNode.path = SdfPath(kNodePathPrefix + std::to_string(++nodeCounter));
```
JGamache-autodesk, line 728: That is a very good idea as it will collapse networks with similar topologies and is really safe to use in the case of UsdPreviewSurface since all nodes have dissimilar parameter names and will probably never result in ambiguity.

However, for MaterialX, where we can have many math nodes where all the inputs are called "input1" and " input2" and all outputs called "out" you will merge together networks that differ only by the use of an "add" instead of a "multiply" node. The fix is actually simple. Instead of using a single "node" string + number, you should be using info:id + number.

HdC-adsk: Good point. I was actually doing the approach with identifier + number and thought it was not necessary for UsdShade material. The change is very small, so we can do it when we implement MaterialX.

HdC-adsk: Hmm... Second thought. I can do it now because it is actually also safer than the "node" prefix.

kxl-adsk: Yes, we should not defer this work.

HdC-adsk: The follow-up of this conversation is at https://github.com/Autodesk/maya-usd/pull/1048 .

PR #1047, opened 2021-01-08
MAYA-108067: ordering is no longer an issue after the new improvements on the Maya side.

HamedSabri-adsk:
Call order should be fixed now after my PR changes in Maya. Please see MAYA-106958.

### DIFF
--- a/test/lib/ufe/testComboCmd.py
+++ b/test/lib/ufe/testComboCmd.py
@@ -613,7 +613,7 @@ class ComboCmdTestCase(testTRSBase.TRSTestCaseBase):
                 "xformOp:rotateZ", "!invert!xformOp:translate:pivot",
                 "xformOp:rotateXYZ:maya_fallback")))
 
-    @unittest.skipIf(mayaUtils.previewReleaseVersion() < 122 , 'Fallback transform op handling only available in Maya Preview Release 121 or later.')
+    @unittest.skipIf(mayaUtils.previewReleaseVersion() < 121 , 'Fallback transform op handling only available in Maya Preview Release 121 or later.')
     def testFallback(self):
         '''Transformable not handled by standard Transform3d handlers must be
     handled by fallback handler.'''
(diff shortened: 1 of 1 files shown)


test/lib/ufe/testComboCmd.py
```diff
@@ -613,12 +613,8 @@ def testFallbackCases(self):
                 "xformOp:rotateZ", "!invert!xformOp:translate:pivot",
                 "xformOp:rotateXYZ:maya_fallback")))
 
-    # Name test such that it runs last.  Otherwise, it runs before 
-    # testRotateScalePivotCompensation(), and causes it to fail.  To be 
-    # investigated --- MAYA-108067.  Fails on Linux, both Python 2 and 3
-    # --- MAYA-108612.
     @unittest.skipIf(mayaUtils.previewReleaseVersion() < 121 or platform.system() == 'Linux', 'Fallback transform op handling only available in Maya Preview Release 121 or later.')
```
ppt-adsk, line 616: Need to adjust the skipIf to remove the Linux condition.

Also, I wonder if you should just change the previewReleaseVersion() check to be <= 121, since your Maya-side fix will only be in PR122.

ppt-adsk (changes requested):

test/lib/ufe/testComboCmd.py
```diff
@@ -613,7 +613,7 @@ def testFallbackCases(self):
                 "xformOp:rotateZ", "!invert!xformOp:translate:pivot",
                 "xformOp:rotateXYZ:maya_fallback")))
 
-    @unittest.skipIf(mayaUtils.previewReleaseVersion() < 121 or platform.system() == 'Linux', 'Fallback transform op handling only available in Maya Preview Release 121 or later.')
+    @unittest.skipIf(mayaUtils.previewReleaseVersion() < 122 , 'Fallback transform op handling only available in Maya Preview Release 121 or later.')
```
ppt-adsk, line 616: You'll need to adjust the message...  Still says 121.  Maybe something like:
"This test requires Maya functionality only available in Maya Preview Release 122 or later."

ppt-adsk (changes requested):

ppt-adsk (approved):

test/lib/ufe/testComboCmd.py
HamedSabri-adsk: @ppt-adsk Ah Good point... I finally got the message correct. lol

See eb064a5b1dfcccfaa8f232bb081e131df073593b

HamedSabri-adsk:
Closing this for now since I noticed the issue is still present with my internal changes. Needs further investigation.

PR #1050, opened 2021-01-08
Fix for UFE v1 code path with missing "|world".

seando-adsk:


lib/mayaUsd/ufe/wrapUtils.cpp
```diff
@@ -70,7 +70,16 @@ UsdStageWeakPtr getStage(const std::string& ufePathString)
     // This function works on a single-segment path, i.e. the Maya Dag path
     // segment to the proxy shape.  We know the Maya run-time ID is 1,
     // separator is '|'.
-    return ufe::getStage(Ufe::Path(Ufe::PathSegment(ufePathString, 1, '|')));
+    // The helper function proxyShapeHandle() assumes Maya path starts
+    // with "|world" and will pop it off. So make sure our string has it.
+    // Note: std::string starts_with only added for C++20. So use diff trick.
+    std::string proxyPath;
+    if (ufePathString.rfind("|world", 0) == std::string::npos) {
```
ppt-adsk, line 77: Nice trick.

ppt-adsk (approved):

kxl-adsk (changes requested):
Please add regression test and cover all ways we can access the stage (with and without ufe)

ppt-adsk (approved):

seando-adsk:
@mattyjams I added to you review the change I made to _test/lib/ufe/testUfePythonImport.py_

lib/mayaUsd/ufe/wrapUtils.cpp
```diff
@@ -113,13 +122,11 @@ UsdPrim ufePathToPrim(const std::string& ufePathString)
...
 {
-    def("getPrimFromRawItem", getPrimFromRawItem);
-
 #ifdef UFE_V2_FEATURES_AVAILABLE
+    def("getPrimFromRawItem", getPrimFromRawItem);
     def("getNodeNameFromRawItem", getNodeNameFromRawItem);
-#endif
-
     def("getNodeTypeFromRawItem", getNodeTypeFromRawItem);
+#endif
```
seando-adsk, line 129: The raw item (for ufe scene item) was only added in UFE v2. So these three wrappers that use raw item should only be avail in v2.

test/lib/ufe/testPythonWrappers.py
```diff
@@ -0,0 +1,101 @@
...
+    @classmethod
+    def setUpClass(cls):
+        if not cls.pluginsLoaded:
+            cls.pluginsLoaded = mayaUtils.isMayaUsdPluginLoaded()
+
+    def setUp(self):
+        ''' Called initially to set up the Maya test environment '''
+        self.assertTrue(self.pluginsLoaded)
+
+    def testWrappers(self):
```
seando-adsk, line 43: New test where I test all the maya-usd ufe wrapper methods.

test/lib/ufe/testUfePythonImport.py
```diff
@@ -26,13 +26,19 @@ class UfePythonImportTestCase(unittest.TestCase):
...
         # ufe library also loaded its dependencies (i.e. the core USD
         # libraries). We test the type name as a string to ensure that we're
         # not causing USD libraries to be loaded any other way.
-        invalidPrim = mayaUsdUfe.getPrimFromRawItem(0)
+        if ufe.VersionInfo.getMajorVersion() >= 2:
+            # raw item only introduced in UFE v2
+            invalidPrim = mayaUsdUfe.getPrimFromRawItem(0)
+        else:
+            # In UFE v2 if the Nb segments == 1, returns invalid prim
+            invalidPrim = mayaUsdUfe.ufePathToPrim('a')
```
seando-adsk, line 41: This test is avail in UFE v1, but it was using a ufe wrapper method that I've made only avail in UFE v2. So for the v1 case use a different wrapper method. We cannot just use ufePathToPrim for both UFE versions, since in UFE v2 ufePathToPrim uses different code that will assert with an invalid input path.

kxl-adsk (approved):
Excellent test, thank you!

mattyjams (approved):
Looks good to me!

test/lib/ufe/testUfePythonImport.py
```diff
@@ -26,13 +26,19 @@ class UfePythonImportTestCase(unittest.TestCase):
...
         from mayaUsd import ufe as mayaUsdUfe
+        import ufe
 
         # Test calling a function that depends on USD. This exercises the
         # script module loader registry function and ensures that loading the
         # ufe library also loaded its dependencies (i.e. the core USD
         # libraries). We test the type name as a string to ensure that we're
         # not causing USD libraries to be loaded any other way.
-        invalidPrim = mayaUsdUfe.getPrimFromRawItem(0)
+        if ufe.VersionInfo.getMajorVersion() == 1:
```
seando-adsk, line 36: Sorry, I realized my previous code wasn't going to work once we switched to UFE v3 (with 0.3.0). UFE v1 always has major version 1.

ppt-adsk (approved):

PR #1057, opened 2021-01-11
perrauo/maya 105659/fix crash when loading unloading the mtoh plugin directx mode

perrauo-adsk:
* Fixed crash which occurs since callbacks for the render override were never removed.
* This is because the render override lifetime is not taken in charge by MViewport2Renderer::registerOverride and not destroyed to remove the callbacks.


perrauo-adsk:
How do I start a preflight? adsk pf will not work.

seando-adsk:
@perrauo-adsk Something looks wrong with this PR - it has commits other than yours. Once the review is done you can start a preflight by **assigning** the PR to "ecp-maya-devops-adsk".

PR #1059, opened 2021-01-11
perrauo/MAYA-105659/fix-crash-when-loading-unloading-the-mtoh-plugin-in-directx-mode

perrauo-adsk:
* Fixed crash which occurs since callbacks for the render override were never removed.
* This is because the render override lifetime is not taken in charge by MViewport2Renderer::registerOverride and not destroyed to remove the callbacks.


pmolodo (approved):
Looks good to me, thanks!

PR #1060, opened 2021-01-11
Work around gcc 6.3.1 bug, and re-enable test.

ppt-adsk:
Direct manipulation failures on Linux were caused by a gcc 6.3.1 compiler problem.  This pull request implements an easy work-around for the issue.


kxl-adsk (approved):
Great catch!

PR #1061, opened 2021-01-11
MAYA-106780 - Editing USD should mark scene as dirty

seando-adsk:
#### MAYA-106780 - Editing USD should mark scene as dirty
* Added catch-all condition in response to USD stage changes. If no UFE notif is sent, but we were given paths, send a UFE subtreeInvalidate notif.

### DIFF
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -192,9 +192,6 @@ void StagesSubject::stageChanged(
     if (stagePath(sender).empty())
         return;
 
-#ifdef UFE_V2_FEATURES_AVAILABLE
-    bool anyUfeV2NotifSent = false;
-#endif
     auto stage = notice.GetStage();
     for (const auto& changedPath : notice.GetResyncedPaths()) {
         // When visibility is toggled for the first time or you add a xformop we enter
@@ -237,7 +234,6 @@ void StagesSubject::stageChanged(
                 if (prim.IsActive()) {
 #ifdef UFE_V2_FEATURES_AVAILABLE
                     Ufe::Scene::instance().notify(Ufe::ObjectAdd(sceneItem));
-                    anyUfeV2NotifSent = true;
 #else
                     auto notification = Ufe::ObjectAdd(sceneItem);
                     Ufe::Scene::notifyObjectAdd(notification);
@@ -245,7 +241,6 @@ void StagesSubject::stageChanged(
                 } else {
 #ifdef UFE_V2_FEATURES_AVAILABLE
                     Ufe::Scene::instance().notify(Ufe::ObjectPostDelete(sceneItem));
-                    anyUfeV2NotifSent = true;
 #else
                     auto notification = Ufe::ObjectPostDelete(sceneItem);
                     Ufe::Scene::notifyObjectDelete(notification);
@@ -258,7 +253,6 @@ void StagesSubject::stageChanged(
                 // - Resyncs imply entire subtree invalidation of all descendant prims and
                 // properties. So we send the UFE subtree invalidate notif.
                 Ufe::Scene::instance().notify(Ufe::SubtreeInvalidate(sceneItem));
-                anyUfeV2NotifSent = true;
             }
 #endif
         }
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -296,9 +305,37 @@ void StagesSubject::stageChanged(
...
+            Ufe::Path ufePath;
+            if (changedPath == SdfPath::AbsoluteRootPath()) {
+                ufePath = stagePath(sender);
+            } else {
+                const std::string& usdPrimPathStr = changedPath.GetPrimPath().GetString();
+                ufePath = stagePath(sender) + Ufe::PathSegment(usdPrimPathStr, g_USDRtid, '/');
             }
+
+            auto sceneItem = Ufe::Hierarchy::createItem(ufePath);
+            Ufe::Scene::instance().notify(Ufe::SubtreeInvalidate(sceneItem));
```
ppt-adsk, line 335: This feels very awkward.  SubtreeInvalidate is meant to imply a structural change to the hierarchy, but you're adding info-only changes to it.  I don't think we should be doing that.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -296,9 +305,37 @@ void StagesSubject::stageChanged(
...
+#endif
+            }
+        }
+    }
+
+#ifdef UFE_V2_FEATURES_AVAILABLE
+    // If we get here and didn't send any UFE notif, but yet have paths
+    // then we'll send a catch-all subtree invalidate notification.
+    if (!anyUfeV2NotifSent) {
+        SdfPathVector allPaths = (SdfPathVector)notice.GetResyncedPaths();
```
ppt-adsk, line 319: We're already looping over resync'ed paths at line 199.  Why can't the catch-all SubtreeInvalidate be in that loop?

seando-adsk: Because at lines 319-323 I merge all the paths together into a single SdfPathVector and then a tline 324 it removes any that are descendant paths. This way if you have two paths: /tree/one and /tree, we just end up with /tree.

lib/mayaUsd/ufe/StagesSubject.cpp
seando-adsk: We discussed this internally and since we have no other UFE notif to send this was the only thing we could do. One idea was the creation of a new UFE notif (such as SceneDirty) but we decided against that. @kxl-adsk Can you provide more info?

ppt-adsk: I agree that the catch-all UFE notification for structural change (i.e. ResyncedPaths) should be SubtreeInvalidate --- but isn't that the case already?

For info-only changes, we should be able to extract an attribute out of the notice, and shouldn't need SubtreeInvalidate --- no?

seando-adsk: Okay I reworked it as we discussed to send Ufe attribute changed notif for the ChangedInfoOnlyPaths. I also incorporated the upcoming UFE changes (.36) to change the notif() method.

I also added the special case when both resync and changed only paths are empty.

ppt-adsk (approved):

seando-adsk:
@ppt-adsk Oops sorry I forgot to commit the change to the test.

ppt-adsk (approved):

test/lib/ufe/testAttribute.py
```diff
@@ -32,8 +34,12 @@ def __init__(self):
...
 
     def __call__(self, notification):
-        if isinstance(notification, ufe.AttributeChanged):
-            self._notifications += 1
+        if(os.getenv('UFE_PREVIEW_VERSION_NUM', '0000') >= '2036'):
+            if isinstance(notification, ufe.AttributeValueChanged):
+                self._notifications += 1
+        else:
+            if isinstance(notification, ufe.AttributeChanged):
+                self._notifications += 1
```
seando-adsk, line 42: Oops sorry. Of course the test passed locally for me because I had UFE 0.2.36. I forgot to account for older UFE versions.

ppt-adsk (approved):

seando-adsk:
@kxl-adsk I had a random test timeout on **master Windows python3** only.
`    The following tests FAILED:`
`    	116 - testUsdImportXforms (Timeout)`

Please merge. In the meantime, I'll re-run the preflight just to verify that it was indeed random.

PR #1065, opened 2021-01-12
Make VP2RenderDelegate enabled by default

kxl-adsk:
We have been enabling VP2RenderDelegate via mod file for a very long time now. It's time to switch the behavior and have this rendering solution be the default one. 

The new `MAYAUSD_DISABLE_VP2_RENDER_DELEGATE ` env variable can be used to switch back to the legacy drawing.


HdC-adsk (approved):

lib/mayaUsd/nodes/proxyShapePlugin.cpp
```diff
@@ -48,9 +48,9 @@ MString _registrantPluginName;
 bool _useVP2RenderDelegate = false;
 
 TF_DEFINE_ENV_SETTING(
-    VP2_RENDER_DELEGATE_PROXY,
+    MAYAUSD_DISABLE_VP2_RENDER_DELEGATE,
     false,
-    "Switch proxy shape rendering to VP2 render delegate.");
+    "Disable VP2RenderDelegate from handling rendering of the proxy shape. Switch to legacy mode");
```
mattyjams, line 53: I'm all for "namespacing" the env setting by prepending the `MAYAUSD_` prefix, so no issue with the name of this variable being changed.

I do have a mild aversion though to options that are of the "enable this setting to disable this feature/behavior" variety. I feel like they often involve some additional mental logic gymnastics to deduce the meaning when they're used as below in the form of "if not disabled then use the new thing". Is that preferable to an inverse setting like this?:

```C++
TF_DEFINE_ENV_SETTING(
    MAYAUSD_ENABLE_VP2_RENDER_DELEGATE,
    true,
    "Use the VP2 render delegate for proxy shape rendering. Disable to switch to legacy mode.");
```

Contentious though it may be, this follows the example of `MAYA_ENABLE_LEGACY_VIEWPORT`, which is admittedly disabled by default, but it's a little more straightforward to reason about. Turn the option off to disable the feature, and turn it on (or leave it alone) to enable it. To my eye, `MAYAUSD_ENABLE_VP2_RENDER_DELEGATE=0` a bit more clearly suggests that you want the VP2 render delegate **disabled**.

mattyjams (approved):
This works for me, but I have a slight preference for the env setting to be called `MAYAUSD_ENABLE_VP2_RENDER_DELEGATE` and be enabled by default instead. I can let that go if you feel strongly otherwise though.

kxl-adsk:
I’m not a big fan of long lasting env variables. They are good for a short period of time but eventually get forgotten. 

The main thing we want to communicate with VP2RenderDelegate env variable is that you are going to be off the default rendering path if you opt-in. Additionally, since this path puts you on legacy drawing, you should not expect we will keep it forever. I don’t think the proposed MAYAUSD_ENABLE_VP2_RENDER_DELEGATE does the job to communicate it. 

Maya’s MAYA_ENABLE_LEGACY_VIEWPORT fits perfectly above the description. It puts you off the default and it can disappear whenever legacy support will be stopped. We could use a similar name here, but I fear it’s not descriptive enough. This is why we chose to include VP2RenderDelegate in the name. 

I do see your point, but I hope my answer is satisfactory.

PR #1073, opened 2021-01-13
Implement setMatrixCmd for Maya transform stack.

ppt-adsk:


lib/mayaUsd/ufe/UsdTransform3dMatrixOp.cpp
```diff
@@ -355,6 +355,18 @@ Ufe::ScaleUndoableCommand::Ptr UsdTransform3dMatrixOp::scaleCmd(double x, double
         UsdTimeCode::Default());
 }
 
+Ufe::SetMatrix4dUndoableCommand::Ptr UsdTransform3dMatrixOp::setMatrixCmd(const Ufe::Matrix4d& m)
+{
+    return nullptr;
```
ppt-adsk, line 360: parent -absolute won't work for matrix op prims yet, but implementing this should be trivial (see setMatrix()) below.

lib/mayaUsd/ufe/UsdTransform3dMatrixOp.cpp
```diff
@@ -355,6 +355,18 @@ Ufe::ScaleUndoableCommand::Ptr UsdTransform3dMatrixOp::scaleCmd(double x, double
...
+}
+
+void UsdTransform3dMatrixOp::setMatrix(const Ufe::Matrix4d& m) { _op.Set(toUsd(m)); }
```
ppt-adsk, line 363: This is how little work parent -absolute has to do for matrix op prims: just set their (local) matrix.

lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
```diff
@@ -169,6 +173,72 @@ createTransform3d(const Ufe::SceneItem::Ptr& item, NextTransform3dFn nextTransfo
     return stackOps.empty() ? nextTransform3dFn() : UsdTransform3dMayaXformStack::create(usdItem);
 }
 
+// Class for setMatrixCmd() implementation.  Should be rolled into a future
+// command class with full undo / redo support
```
ppt-adsk, line 177: This temporary command just does undo / redo with values.  Will be rolled into the up-coming Uber value undo / redo class.

lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
```diff
@@ -169,6 +173,72 @@ createTransform3d(const Ufe::SceneItem::Ptr& item, NextTransform3dFn nextTransfo
...
+        const Ufe::Vector3d& oldT,
+        const Ufe::Vector3d& oldR,
+        const Ufe::Vector3d& oldS,
+        const Ufe::Matrix4d& newM)
+        : Ufe::SetMatrix4dUndoableCommand(path)
+        , _oldT(oldT)
+        , _oldR(oldR)
+        , _oldS(oldS)
+    {
+        // Decompose new matrix to extract TRS.  Neither GfMatrix4d::Factor
```
ppt-adsk, line 192: Since now the parent -absolute code no longer decomposes the local matrix into TRS, we do it here in the plugin, using essentially the same code as what Maya used to use, namely MTransformationMatrix.

lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
```diff
@@ -169,6 +173,72 @@ createTransform3d(const Ufe::SceneItem::Ptr& item, NextTransform3dFn nextTransfo
...
+        // nor GfTransform decomposition provide results that match Maya,
+        // so use MTransformationMatrix.  Struggle a bit to get the
+        // Ufe::Matrix4d into MTransformationMatrix --- even though their
+        // underlying matrix layout is exactly the same.
+        double mArray[4][4];
+        std::memcpy(mArray, &newM.matrix[0][0], sizeof(double) * 16);
+        MMatrix                              m(mArray);
```
kxl-adsk, line 199: This looks less painful
```suggestion
        MMatrix                              m;
        std::memcpy(m[0], &newM.matrix[0][0], sizeof(double) * 16);
```

kxl-adsk (approved):
Do we already have a test for parenting with absolute flag and maya xform stack? (if we would have test coverage data, I wouldn't have to ask :) )

ppt-adsk:
> 
> 
> Do we already have a test for parenting with absolute flag and maya xform stack? (if we would have test coverage data, I wouldn't have to ask :) )

Yes, it's testParentCmd, which was failing when I did the matrix decomposition with the Pixar code base.  Now that I'm using the Maya MTransformationMatrix, the test passes.

kxl-adsk:
I don't see this PR re-enabling any test?

ppt-adsk:
No, it uses the existing testParentCmd.  During development of this change, testParentCmd was failing, and with the solution I have, testParentCmd passes.

When we bless the new Maya, testParentCmd will exercise the new setMatrixCmd() code path, as it does for me locally, and it will still pass.

kxl-adsk (approved):

PR #1075, opened 2021-01-14
MAYA-107402: A better error message when renaming a prim inside a variantset.

HamedSabri-adsk:
A better error message when renaming a prim inside a variantset.


ppt-adsk (commented):
Do we have a test for this?  I searched for variant in test/lib/ufe, but only found context ops testing...

HamedSabri-adsk:
> 
> 
> Do we have a test for this? I searched for variant in test/lib/ufe, but only found context ops testing...

No I don't have a test case for it.  I will add one to the testRename.py if it is needed.

ppt-adsk:
> 
> 
> > Do we have a test for this? I searched for variant in test/lib/ufe, but only found context ops testing...
> 
> No I don't have a test case for it. I will add one to the testRename.py if it is needed.

Yeah, would be nice.

ppt-adsk (approved):
Thanks for the test!

PR #1078, opened 2021-01-14
UFE v2 transition

seando-adsk:
Temporary setting of UFE_PREVIEW_VERSION_NUM while we transition to UFE v2.0.0.

### DIFF
--- a/cmake/modules/FindUFE.cmake
+++ b/cmake/modules/FindUFE.cmake
@@ -46,6 +46,10 @@ if(UFE_INCLUDE_DIR AND EXISTS "${UFE_INCLUDE_DIR}/ufe/ufe.h")
 
     if("${UFE_MAJOR_VERSION}" STREQUAL "0")
         math(EXPR UFE_PREVIEW_VERSION_NUM "${UFE_MINOR_VERSION} * 1000 + ${UFE_PATCH_LEVEL}")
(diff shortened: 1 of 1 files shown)


cmake/modules/FindUFE.cmake
```diff
@@ -46,6 +46,10 @@ if(UFE_INCLUDE_DIR AND EXISTS "${UFE_INCLUDE_DIR}/ufe/ufe.h")
 
     if("${UFE_MAJOR_VERSION}" STREQUAL "0")
         math(EXPR UFE_PREVIEW_VERSION_NUM "${UFE_MINOR_VERSION} * 1000 + ${UFE_PATCH_LEVEL}")
+    elseif("${UFE_VERSION}" STREQUAL "2.0.0")
+        # Temporary. Once next Maya PR is released with UFE v2.0.0 this should
+        # be removed (along with all the UFE_PREVIEW_VERSION_NUM checks).
+        set(UFE_PREVIEW_VERSION_NUM 2036)
```
seando-adsk, line 52: Internally switching to UFE v2.0.0 means that line 47 is not true anymore so UFE_PREVIEW_VERSION_NUM will no longer be set. This is a temporary measure to have it set to the last ufe version (0.2.36).

ppt-adsk (approved):

PR #1081, opened 2021-01-15
Custom layer data export

dgovil:
Draft PR to aid this discussion https://github.com/Autodesk/maya-usd/discussions/896#discussioncomment-284792
Adds the ability for users to specify an array of triplets to be exported as customLayerData on the root layer.
Supports basic types like string, bool, int, float and double for now.

### DIFF
--- a/lib/mayaUsd/commands/baseExportCommand.cpp
+++ b/lib/mayaUsd/commands/baseExportCommand.cpp
@@ -161,6 +161,14 @@ MSyntax MayaUSDExportCommand::createSyntax()
         UsdMayaJobExportArgsTokens->staticSingleSample.GetText(),
         MSyntax::kBoolean);
 
+    syntax.addFlag(
+        kCustomLayerData,
+        UsdMayaJobExportArgsTokens->customLayerData.GetText(),
+        MSyntax::kString,
+        MSyntax::kString,
+        MSyntax::kString);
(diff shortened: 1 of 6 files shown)


dgovil:
@wtelford for visibility, this is a draft PR/proposal for the discussion regarding the custom layer data

dgovil:
I'm just removing the draft status on this, since we have it in our internal fork now.
I think the big question is how this would work with different layer specifiers, if the export plugin was to be expanded out to support multiple layers at once.

kxl-adsk:
@dgovil I'm wondering why we need something so specific to layer custom data...why we couldn't for example go with the chaser approach and achieve the same thing?

dgovil:
@kxl-adsk Do you mean something like writing a chaser to achieve this? That's what we did internally originally, but we had several varying requests from people internal/external about embedding layer metadata. Developing + building a chaser for it would be outside the abilities of most of the teams we spoke to, where they instead were relying on post processing to open+write the metadata.

This was an attempt to try and reduce the friction point for including metadata in the USD layer files by just providing export flags.

dgovil:
Per https://github.com/Autodesk/maya-usd/pull/1136 , closing this PR out since it is preferred right now to go with a chaser or callback route. This PR should still be viewable when closed as supporting material for the discussion in https://github.com/Autodesk/maya-usd/discussions/896#discussioncomment-284792

PR #1082, opened 2021-01-15
MAYA-109190 : Fix camera frame for instance proxies.

HamedSabri-adsk:
This PR fixes the regression with camera framing of instance proxies. The issue was related to the new conditional check that was added for "instance proxies" in UsdTransform3dMayaXformStackHandler::createTransform3d().  

This was a wrong place to have this condition since the original intention was to avoid "authoring edit transforms" for instance proxies. When the camera frame happens, we would end up calculating the ufe BBox which would go through transform3d() and eventually createTransform3d().


ppt-adsk (approved):

kxl-adsk:
Results are in the report - https://github.com/Autodesk/maya-usd/runs/1710502357?check_suite_focus=true

PR #1086, opened 2021-01-17
Fixed anonymous layer import

kxl-adsk:
Reported in https://github.com/Autodesk/maya-usd/discussions/1069. We didn't have symmetry between import and export behavior for anonymous layers. 

It may be confusing that we are using a file path to feed layer identifier, but since the exporter and the underlying job already handles it we filter out anonymous layers when validating the path provided.


mattyjams (approved):
Looks good to me!

PR #1094, opened 2021-01-20
author kind=group on the UsdPrim created by the group command when its parent is in the model hierarchy

mattyjams:
Authoring the kind metadata on the group `Xform` prim ensures that the model hierarchy remains contiguous between the parent and children of the new group prim. Prims underneath the group prim may still be in the model hierarchy if they also have kind authored.

If the parent prim of the group prim is not part of the model hierarchy, no kind is authored on the group prim.

More info about model hierarchy can be found here:
https://graphics.pixar.com/usd/docs/USD-Glossary.html#USDGlossary-ModelHierarchy

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoSetKindCommand.h
+++ b/lib/mayaUsd/ufe/UsdUndoSetKindCommand.h
@@ -1,0 +1,66 @@
+//
+// Copyright 2021 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef MAYAUSD_UFE_USD_UNDO_SET_KIND_COMMAND_H
+#define MAYAUSD_UFE_USD_UNDO_SET_KIND_COMMAND_H
+
+#include <mayaUsd/base/api.h>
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/ufe/UsdUndoSetKindCommand.h
```diff
@@ -0,0 +1,66 @@
...
+public:
+    using Ptr = std::shared_ptr<UsdUndoSetKindCommand>;
+
+    UsdUndoSetKindCommand(const PXR_NS::UsdPrim& prim, const PXR_NS::TfToken& kind);
+    ~UsdUndoSetKindCommand() override = default;
+
+    UsdUndoSetKindCommand(const UsdUndoSetKindCommand&) = delete;
+    UsdUndoSetKindCommand& operator=(const UsdUndoSetKindCommand&) = delete;
+    UsdUndoSetKindCommand(UsdUndoSetKindCommand&&) = delete;
+    UsdUndoSetKindCommand& operator=(UsdUndoSetKindCommand&&) = delete;
```
kxl-adsk, line 44: I see we copy this pattern around making the declaration of this trivial class busier than it has to be.

The base class `Ufe::UndoableCommand` already explicitly deletes the copy constructor. This will make the compiler implicitly delete any derived class copy constructors. So line 41 is not needed.
Lack of a copy constructor will also make the compiler implicitly delete the move constructor. So line 43 is not needed.

Finally, we should do a better job in the base class and delete the copy assignment operator, this would make the compiler implicitly delete the copy assignment operator and move the assignment operator from the derived class. This is not currently done, so lines 42 and 44 have to stay.

HamedSabri-adsk (approved):
Looking good and thank you for adding the test.

lib/mayaUsd/ufe/UsdUndoSetKindCommand.h
HamedSabri-adsk: That's a good point. I really don't have a strong opinion here. I personally am not a huge fan of implicit philosophy whether it is C++ or any other languages.

PR #1100, opened 2021-01-22
MAYA-109316 Propagate state of dirty draw items

JGamache-autodesk:
when they are about to be redrawn after getting out of sync.

This fixes drawing multiple Repr in sequence after dirtying the RPrim.
The first draw will clean the RPrim and the Reprs that were selected
to be drawn. Reprs that were not drawn will stay stale and need to
signal the Sync engine that they need to be refreshed on the next
draw cycle that will draw them.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
@@ -674,11 +674,11 @@ HdDirtyBits HdVP2Mesh::_PropagateDirtyBits(HdDirtyBits bits) const
             const HdReprSharedPtr& repr = pair.second;
             const auto&            items = repr->GetDrawItems();
 #if HD_API_VERSION < 35
-            for (HdDrawItem* item : items) {
-                if (HdVP2DrawItem* drawItem = static_cast<HdVP2DrawItem*>(item)) {
+            for (const HdDrawItem* item : items) {
+                if (const HdVP2DrawItem* drawItem = static_cast<const HdVP2DrawItem*>(item)) {
 #else
             for (const HdRepr::DrawItemUniquePtr& item : items) {
-                if (HdVP2DrawItem* const drawItem = static_cast<HdVP2DrawItem*>(item.get())) {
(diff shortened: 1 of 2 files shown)


williamkrick (approved):
It looks good. More duplicated code means refactoring these and adding a common parent class more and more important. We've talked about that a lot in the past, and I just logged MAYA-109380 so we can hopefully put it on the schedule.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -648,18 +648,39 @@ HdDirtyBits HdVP2Mesh::_PropagateDirtyBits(HdDirtyBits bits) const
...
+            for (const HdRepr::DrawItemUniquePtr& item : items) {
+                if (HdVP2DrawItem* const drawItem = static_cast<HdVP2DrawItem*>(item.get())) {
 #endif
-                drawItem->SetDirtyBits(bits);
+                    drawItem->SetDirtyBits(bits);
+                }
+            }
+        }
+    } else {
+        // RPrim is clean, find out if any drawItem about to be shown is dirty:
```
kxl-adsk, line 668: What's the performance impact of this change?

kxl-adsk:
FYI. This change will require fixes to compile with USD 20.02 (and 19.11 since we haven't yet cleaned it up)
```
E:\usd\maya-usd_pr\lib\mayaUsd\render\vp2RenderDelegate\mesh.cpp(746): error C2440: 'static_cast': cannot convert from 'const pxrInternal_v0_19__pxrReserved__::HdDrawItem *' to 'pxrInternal_v0_19__pxrReserved__::HdVP2DrawItem *'
```

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
JGamache-autodesk: No performance impact. I checked and vertex buffers are correctly shared and are updated only once when changing a sphere radius. The geometries get updated for each displayed render item, but only as necessary to clean the dirty bits found in the reprs.

kxl-adsk (approved):

PR #1104, opened 2021-01-22
Support import chaser plugins

ysiewappl:
Hi:

This PR adds symmetric functionality for import chaser plugins to complement the existing export chaser plugins. This allows sites to perform post-import operations that are specific to their pipeline more flexibly without having to deal with attempting to juggle incoming updates to the Maya-USD codebase from upstream, and makes prototyping new features to the importer a little easier.

A sample import chaser, `infoImportChaser`, has been provided to demonstrate possible intended usage and how to work with undo/redo operations of the normal `mayaUSDImport` command. You can run it via:

```
cmds.mayaUSDImport(
         file='/tmp/test.usda',
         chaser=['info']
     )
```

I ran into an issue that I haven't yet been able to find time to dig into yet, which is that it appears that the `pxrUsd` plugin _must_ be loaded first before the `mayaUsdPlugin`, or else neither export nor import chasers will be able to register when being called from the respective `mayaUSDImport` and `mayaUSDExport` commands (the error `Failed to create import chaser` is shown because somehow the macro that registers the chaser is never executed; this also happens even for the default Alembic export chaser plugin). If someone from ADSK or elsewhere could help investigate, I'd appreciate it.


kxl-adsk:
@ysiewappl Thank you for another contribution. Before starting the review process I would like us to focus on documenting this proposal. This will not only help collect feedback but as well serve as documentation once we move forward with the proposal.

While we haven't yet moved and updated the documentation for import/export commands, I propose you start writing one that would follow what is done for the `export chaser` - https://github.com/Autodesk/maya-usd/tree/dev/plugin/pxr/doc#export-chasers-advanced

ysiewappl:
> While we haven't yet moved and updated the documentation for import/export commands, I propose you start writing one that would follow what is done for the export chaser - https://github.com/Autodesk/maya-usd/tree/dev/plugin/pxr/doc#export-chasers-advanced
Hi, not sure if this got missed but I've gone ahead and updated the document in `82c847d` with the relevant info. Do let me know if this is suitable or if there's different info you'd like to see in there.

Thanks!

kxl-adsk:
Thank you, @ysiewappl for following up on this. This didn't get missed, I just wanted to collect a bit more feedback internally first. A few internal priorities got in the way, but I'm on it now.

kxl-adsk:
The concept and symmetry makes sense to me. I raised this internally as well and didn’t get any concerns reported. 

We have some milestones coming up that makes us slower in responding, but I will start the review process on this one.

lib/mayaUsd/fileio/chaser/CMakeLists.txt
```diff
@@ -1,25 +1,29 @@
 # -----------------------------------------------------------------------------
 # sources
 # -----------------------------------------------------------------------------
-target_sources(${PROJECT_NAME} 
+target_sources(${PROJECT_NAME}
     PRIVATE
         chaser.cpp
         chaserRegistry.cpp
+        importChaserRegistry.cpp
```
kxl-adsk, line 8: In translation framework code, we rarely use import or export in the names. Let's be consistent and use writer / reader convention.

lib/mayaUsd/fileio/jobs/readJob.cpp
```diff
@@ -271,10 +277,35 @@ bool UsdMaya_ReadJob::Read(std::vector<MDagPath>* addedDagPaths)
         if (TfMapLookup(mNewNodeRegistry, key, &obj)) {
             if (obj.hasFn(MFn::kDagNode)) {
                 addedDagPaths->push_back(MDagPath::getAPathTo(obj));
+                currentAddedDagPaths.append(MDagPath::getAPathTo(obj));
```
kxl-adsk, line 280: Can you explain the difference between `addedDagPaths` and `currentAddedDagPaths`?

lib/mayaUsd/fileio/jobs/readJob.h
```diff
@@ -128,9 +129,10 @@ class UsdMaya_ReadJob
     double _setTimeSampleMultiplierFrom(const double layerFPS);
 
     // Data
-    MDagModifier mDagModifierUndo;
-    bool         mDagModifierSeeded;
-    double       mTimeSampleMultiplier;
+    MDagModifier                    mDagModifierUndo;
+    bool                            mDagModifierSeeded;
+    double                          mTimeSampleMultiplier;
+    UsdMayaImportChaserRefPtrVector mImportChasers;
```
kxl-adsk, line 135: Please document the purpose and lifetime of this member (don't hesitate to add documentation above as well).

plugin/pxr/doc/README.md
```diff
@@ -32,6 +32,8 @@ The plugin creates two commands: `usdImport` and `usdExport`, and will also regi
 | ---------- | --------- | ---- | ------- | ----------- |
 | `-api` | `-apiSchema` | string (multi)| none | Imports the given API schemas' attributes as Maya custom attributes. This only recognizes API schemas that have been applied to prims on the stage. The attributes will properly round-trip if you re-export back to USD. |
 | `-ar` | `-assemblyRep` | string | `Collapsed` | If the import results in the creation of assembly nodes, this value specifies the assembly representation that will be activated after creation. If empty, no representation will be activated. Valid values are: `Collapsed`, `Expanded`, `Full`, `Import`, `(empty)`. `Import`: No USD reference assembly nodes will be created, and the geometry will be imported directly. `(empty)`: The assembly is created but is left unloaded after creation. See "Importing as Assemblies" below for more detail on assembly node creation. |
+|`-chr` | `-chaser` | string(multi) | none | Specify the import chasers to execute as part of the export. See "Import Chasers" below.
```
kxl-adsk, line 35: We will no longer update this document. Please add this documentation to https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/commands/Readme.md

plugin/pxr/maya/plugin/pxrUsd/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ pxr_plugin(${PXR_PACKAGE}
 
     CPPFILES
         alembicChaser.cpp
+        infoImportChaser.cpp
```
kxl-adsk, line 22: You added the test to the common location but made it dependent on building the Pixar plugin. This dependency we want to avoid and to achieve it, I propose to move infoImportChaser to https://github.com/Autodesk/maya-usd/tree/dev/test/lib/usd/plugin

kxl-adsk (commented):
@ysiewappl I didn't find anything major but left a few comments that would be worth addressing in this PR.

lib/mayaUsd/fileio/chaser/CMakeLists.txt
mattyjams: I'm less bothered by using "import" and "export" here since they're somewhat more tied to the import and export commands than to the translation framework, but whichever we choose, I think we should rename the existing `chaser.cpp` and `chaserRegistry.cpp` to include the "writer" or "export" terminology.

lib/mayaUsd/fileio/chaser/importChaserRegistry.cpp
```diff
@@ -0,0 +1,107 @@
...
+{
+    return this->_jobArgs;
+}
+
+TF_INSTANTIATE_SINGLETON(UsdMayaImportChaserRegistry);
+
+std::map<std::string, UsdMayaImportChaserRegistry::FactoryFn> _factoryImportRegistry;
+
+UsdMayaImportChaserRegistry::UsdMayaImportChaserRegistry() { }
+UsdMayaImportChaserRegistry::~UsdMayaImportChaserRegistry() { }
```
mattyjams, line 67: Maybe `= default;` these in the header instead?

lib/mayaUsd/fileio/chaser/importChaserRegistry.h
```diff
@@ -0,0 +1,130 @@
...
+        const UsdMayaJobImportArgs& GetImportJobArgs() const;
+
+    private:
+        UsdStagePtr                 _stage;
+        const MDagPathArray&        _dagPaths;
+        const SdfPathVector&        _sdfPaths;
+        const UsdMayaJobImportArgs& _jobArgs;
+    };
+
+    typedef std::function<UsdMayaImportChaser*(const FactoryContext&)> FactoryFn;
```
mattyjams, line 85: We've been favoring `using` over `typedef` in new code.

lib/mayaUsd/fileio/jobs/jobArgs.h
```diff
@@ -282,6 +284,10 @@ struct UsdMayaJobImportArgs
     /// non-empty interval.
     const GfInterval timeInterval;
 
+    typedef std::map<std::string, std::string> ChaserArgs;
```
mattyjams, line 287: `using` vs. `typedef`

mattyjams (approved):
Sorry, I missed that I was tagged as a reviewer on this one. No major comments beyond what @kxl-adsk already pointed out. Otherwise makes sense to me!

lib/mayaUsd/fileio/chaser/CMakeLists.txt
ysiewappl: I can certainly change the names as requested, but my concern is that the precedent is set in that the export chasers are already called that; it might be confusing to call these "reader" chasers instead afterwards (especially since the `readJob` calls methods such as `DoImport` as well, anyway, and the import chaser code runs at the end of that) that imply they do something other than symmetric functionality for the existing export chasers.

kxl-adsk: @ysiewappl just to make sure I understood your proposal - you suggest we keep the current naming convention in the PR or you suggest to use ‘import’ and ‘export’ terminology in names?

lib/mayaUsd/fileio/jobs/readJob.cpp
ysiewappl: Originally, it looked to me like `addedDagPaths` was being used for purposes other than just acting as a cache of the DAG paths added during the single import job; now, after looking closely, I see that there is no appreciable difference other than the fact that one is `std::vector<MDagPath>`, and the other a `MDagPathArray`. Should import chasers take a parameter with the signature of the former, or latter (since they're not really equivalent under the hood?)

lib/mayaUsd/fileio/chaser/CMakeLists.txt
ysiewappl: @kxl-adsk Yes, I'm suggesting we continue to use the terms "import" and "export" chasers and match that with the current naming convention in the PR (IMHO it's less confusing). But of course I can certainly rename these to be "reader" chasers as well; just as long as we decide one way or another and be consistent I have no big qualms with either.

kxl-adsk: I see we all agree with the need for consistency. I won't generate more work than required, since both terminologies are already in use, i.e. I will give up on the switch to `reader`/`writer` but please rename `chaser.cpp` and `chaseRegistry.cpp` to include `export`.

lib/mayaUsd/fileio/jobs/readJob.cpp
kxl-adsk: Since we are using Maya types here, let's stick to MDagPathArray.

plugin/pxr/doc/README.md
ysiewappl: This is addressed in 3c48ae3.

lib/mayaUsd/fileio/jobs/readJob.h
ysiewappl: This is addressed in 50c5396.

lib/mayaUsd/fileio/chaser/importChaserRegistry.cpp
ysiewappl: I realize that just letting the compiler generate default constructors instead is probably easier. Addressed in e8efa0c.

lib/mayaUsd/fileio/chaser/importChaserRegistry.h
ysiewappl: This is addressed in e8efa0c.

lib/mayaUsd/fileio/jobs/jobArgs.h
ysiewappl: This is addressed in e8efa0c.

kxl-adsk: You changed export, but not import.

ysiewappl: Apologies, you're correct. This is now addressed in ef5450c.

lib/mayaUsd/fileio/chaser/CMakeLists.txt
ysiewappl: This is now addressed in ef5450c specifically as a separate commit, which should hopefully make it easier to revert if it causes a regression.

plugin/pxr/maya/plugin/pxrUsd/CMakeLists.txt
ysiewappl: I'd like to avoid it as well, but the problem is that if I move it, I run into the same problem that I described in the PR description:

> I ran into an issue that I haven't yet been able to find time to dig into yet, which is that it appears that the pxrUsd plugin must be loaded first before the mayaUsdPlugin, or else neither export nor import chasers will be able to register when being called from the respective mayaUSDImport and mayaUSDExport commands (the error Failed to create import chaser is shown because somehow the macro that registers the chaser is never executed; this also happens even for the default Alembic export chaser plugin). If someone from ADSK or elsewhere could help investigate, I'd appreciate it.

Internally, we do not register/build our chasers the same way (they are separate DLLs altogether), so we do not have this issue. But for the sake of having a working example that anyone can compile and use, I left it in the same location as the Alembic chaser for now.

kxl-adsk: What I was suggesting is making your test chaser a new plugin, so this very much resembles your setup with separate DLLs. Have a look at what's in https://github.com/Autodesk/maya-usd/tree/dev/test/lib/usd/plugin.

ysiewappl: This request has been accommodated in 271c742040c5e2c3feeebd96831fefffc3e8eb2a.

kxl-adsk: I see you added the chaser to one of the existing plugins. I must have misguided you, but my hope was you will add a new target library to https://github.com/Autodesk/maya-usd/blob/dev/test/lib/usd/plugin/CMakeLists.txt that will only have your test chaser plugin.

As for test setup, see how `MAYA_PLUG_IN_PATH` is passed when registering a test https://github.com/Autodesk/maya-usd/blob/dev/test/lib/usd/translators/CMakeLists.txt#L216

ysiewappl: > I see you added the chaser to one of the existing plugins. I must have misguided you, but my hope was you will add a new target library to https://github.com/Autodesk/maya-usd/blob/dev/test/lib/usd/plugin/CMakeLists.txt that will only have your test chaser plugin.
Ah, I see. I have accommodated for this request in 57d630d1e. Hopefully that's correct now, though I must admit this setup is a little confusing to me, so I might still be misunderstanding. Please let me know if things look appropriate now.

> As for test setup, see how MAYA_PLUG_IN_PATH is passed when registering a test https://github.com/Autodesk/maya-usd/blob/dev/test/lib/usd/translators/CMakeLists.txt#L216
I did glance at that file, but I don't see immediately how it adds the plugins in `maya-usd/build/test/lib/usd/plugin/Debug` to the `MAYA_PLUG_IN_PATH`; the fact that I cannot run the existing test `testUsdMayaListShadingModesCommand` either with the same plugin import issue implies to me that it's not just the new test I'm adding that has the issue.

kxl-adsk: Have a look at the attached diff - this is how you configure test env when a particular test has special dependencies.
[fix_test_import_chaser.diff.txt](https://github.com/Autodesk/maya-usd/files/6021570/fix_test_import_chaser.diff.txt)

ysiewappl: Hmm, it still doesn't work for me. I think that's because when I build and test, I'm always doing so in `Debug` configuration, which means my plugins get deployed here:

`maya-usd/build/test/lib/usd/translators/../plugin/Debug/`

I've gone ahead and applied your patch in `ca1d981bc`, regardless.

ysiewappl:
I saw that CI encountered a failure, but I honestly can't make sense of how any code in this PR could have affected it. Is this actually related?

```
      File "S:\jenkins\workspace\ecg-mayausd-branch-preflight-2022-windows\ecg-maya-usd\maya-usd\build.py", line 607, in <module>
        RunTests(context, context.ctestArgs)
      File "S:\jenkins\workspace\ecg-mayausd-branch-preflight-2022-windows\ecg-maya-usd\maya-usd\build.py", line 390, in RunTests
        RunCTest(context,extraArgs)
      File "S:\jenkins\workspace\ecg-mayausd-branch-preflight-2022-windows\ecg-maya-usd\maya-usd\build.py", line 301, in RunCTest
        extraArgs=(" ".join(extraArgs) if extraArgs else "")))
      File "S:\jenkins\workspace\ecg-mayausd-branch-preflight-2022-windows\ecg-maya-usd\maya-usd\build.py", line 165, in Run
        .format(cmd=cmd, log=os.path.abspath(context.logFileLocation)))
    RuntimeError: Failed to run 'ctest --output-on-failure --timeout 500 -C RelWithDebInfo -E testUsdReferenceAssemblySelection|testUsdMayaAppDir|testUsdExportAsClip -LE interactive --parallel 16 '
    See W:\build\RelWithDebInfo\build_log.txt for more details.
build.py: error: Command failed with error code: 1
```

When I run the same test locally, it passes, which is the stranger bit to this.

test/lib/usd/plugin/CMakeLists.txt
```diff
@@ -127,13 +127,54 @@ if(IS_MACOSX OR IS_LINUX)
...
+        plugin.cpp
+        infoImportChaser.cpp
+)
+mayaUsd_compile_config(${TARGET_NAME})
+
 # -----------------------------------------------------------------------------
-# Plug Plug-in files
+# link libraries
 # -----------------------------------------------------------------------------
-configure_file("${CMAKE_CURRENT_SOURCE_DIR}/mayaUsdPlugInfo.json"
```
kxl-adsk, line 133: Here is the reason why the version check test failed. This code was removed by accident. Apply this patch to fix the test
[fix_code_removal.patch.txt](https://github.com/Autodesk/maya-usd/files/6031486/fix_code_removal.patch.txt)

kxl-adsk:
@ysiewappl for failing test - see my https://github.com/Autodesk/maya-usd/pull/1104#discussion_r581341974

kxl-adsk:
@ysiewappl PF passed. I still don’t understand why the test is not passing for you locally. How are you running tests?

kxl-adsk (approved):

ysiewappl:
> @ysiewappl PF passed. I still don’t understand why the test is not passing for you locally. How are you running tests?

I think I was having issues because I usually build in Debug mode only, and I was running tests under Release mode, which ended up looking at the wrong paths (For some of the other Python tests that didn't rely on loading any other plugins, this didn't matter previously, so I never ran into the problem). Once I set my test config appropriately, things worked again. 

Thanks for following up so promptly on this!

PR #1105, opened 2021-01-23
Maya fallback + additional op no longer crashes.

ppt-adsk:


lib/mayaUsd/ufe/UsdTransform3dFallbackMayaXformStack.cpp
```diff
@@ -80,7 +80,12 @@ bool MatchingSubstack(const std::vector<UsdGeomXformOp>& ops)
     // current op's index, since each index can appear only once.
     int orderedNdx = UsdTransform3dMayaXformStack::NdxTranslate;
     for (const auto& op : ops) {
-        auto opNdx = gOpNameToNdx.at(op.GetOpName());
+        auto found = gOpNameToNdx.find(op.GetOpName());
+        if (found == gOpNameToNdx.end()) {
+            // Found an op that doesn't match a Maya fallback stack op.
```
ppt-adsk, line 85: The Maya fallback transform stack must be the last set of transform ops in a prim's transform stack.  If there is already a Maya fallback transform stack at the tail of a prim's transform stack, any op that does not match means the Maya fallback transform stack has been "corrupted" in some way by adding or inserting an op that shouldn't be there.

If that happens, we don't create a fallback to the fallback: we detect the erroneous transform op, and return a nullptr Transform3d interface.

test/lib/ufe/testComboCmd.py
```diff
@@ -612,6 +612,67 @@ def testFallbackCases(self):
                 "xformOp:rotateZ", "!invert!xformOp:translate:pivot",
                 "xformOp:rotateXYZ:maya_fallback")))
 
+    def testBrokenFallback(self):
```
ppt-adsk, line 615: Written in TDD style: test written first, failed, code fixed, test passes.

test/lib/ufe/testComboCmd.py
```diff
@@ -612,6 +612,67 @@ def testFallbackCases(self):
...
+        cmds.file(new=True, force=True)
+
+        import mayaUsd_createStageWithNewLayer
+        mayaUsd_createStageWithNewLayer.createStageWithNewLayer()
+        proxyShapePath = ufe.PathString.path('|stage1|stageShape1')
+        proxyShapeItem = ufe.Hierarchy.createItem(proxyShapePath)
+        proxyShapeContextOps = ufe.ContextOps.contextOps(proxyShapeItem)
+        proxyShapeContextOps.doOp(['Add New Prim', 'Capsule'])
+
+        capsulePath = ufe.PathString.path('|stage1|stageShape1,/Capsule1')
```
ppt-adsk, line 634: Steps in test taken from ADSK JIRA bug description, and adapted.

test/lib/ufe/testComboCmd.py
```diff
@@ -612,6 +612,67 @@ def testFallbackCases(self):
...
+        # Add another transform op to break the Maya fallback stack.
+        capsuleXformable.AddRotateZOp()
+
+        self.assertEqual(
+            capsuleXformable.GetXformOpOrderAttr().Get(), Vt.TokenArray((
+                "xformOp:rotateX", "xformOp:rotateY",
+                "xformOp:rotateXYZ:maya_fallback", "xformOp:rotateZ")))
+
+        # Do any transform editing with Maya.
+        cmds.rotate(0, 0, 30, r=True, os=True, fo=True)
```
ppt-adsk, line 671: This line will crash if we don't have a Maya with the fix to MAYA-109343, which is fairly recent.  Will need to figure out how to identify this version of Maya and skip the test if necessary; this needs to work across Maya branches.  Suggestions welcome.

HamedSabri-adsk: @ppt-adsk Let's have a chat offline today.

HamedSabri-adsk (approved):

HamedSabri-adsk (approved):

ppt-adsk:
Only pre-flight failure is macOS Python 2, so indeed ready for merge.

PR #1107, opened 2021-01-25
Ufe: Remove unused include(s).

HamedSabri-adsk:


lib/mayaUsd/ufe/ProxyShapeHierarchyHandler.h
```diff
@@ -16,7 +16,6 @@
 #pragma once
 
 #include <mayaUsd/base/api.h>
-#include <mayaUsd/ufe/UfeVersionCompat.h>
```
seando-adsk, line 19: This header is definitely needed for the UFE v1 case. I just added this last week. Also are you sure about some of the other UFE headers you've removed? They could be needed by some of the ifdef's.

seando-adsk (changes requested):

lib/mayaUsd/ufe/UsdUndoSetKindCommand.cpp
```diff
@@ -17,9 +17,7 @@
 
 #include <mayaUsd/undo/UsdUndoBlock.h>
 
-#include <pxr/base/tf/token.h>
 #include <pxr/usd/usd/modelAPI.h>
-#include <pxr/usd/usd/prim.h>
```
mattyjams, line 22: What defines "unused" here? `TfToken` and `UsdPrim` are most definitely used below.

lib/mayaUsd/ufe/ProxyShapeHierarchyHandler.h
HamedSabri-adsk: > This header is definitely needed for the UFE v1 case. 

I am on UFE 2.x. We should wrap such ifdef so it is clear. I would have never known if it is needed for UFE v1.

> Also are you sure about some of the other UFE headers you've removed? They could be needed by some of the ifdef's.

I am getting some help from a tool to detect these. It is not perfect but nonetheless very useful. We need to be be careful about blindly adding these headers since it can affect compile time and we do have a lot them.

lib/mayaUsd/ufe/UsdUndoSetKindCommand.cpp
HamedSabri-adsk: The includes for TfToken and UsdPrim are already in the header.  The ones in the *.cpp are unnecessary.

https://github.com/Autodesk/maya-usd/blob/1fd0c0f3f8b643456b7fe808f4fdefd06d734c43/lib/mayaUsd/ufe/UsdUndoSetKindCommand.h#L23

lib/mayaUsd/ufe/ProxyShapeHierarchyHandler.h
seando-adsk: Normally I would agree, but this header it is all about UFE version compatibility so I don't think it should ever be conditionally included.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -19,6 +19,7 @@
 
 #include <mayaUsd/nodes/proxyShapeBase.h>
 #include <mayaUsd/ufe/ProxyShapeHandler.h>
+#include <mayaUsd/ufe/UfeVersionCompat.h>
```
HamedSabri-adsk: @seando-adsk I had a closer look at the logic in UfeVersionCompat.h and agree that it doesn't have to be wrapped by ifdef.

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -21,9 +21,16 @@
...
+#ifndef UFE_V2_FEATURES_AVAILABLE
+#include <pxr/usd/sdf/copyUtils.h>
+#endif
+
 #include <ufe/log.h>
 #include <ufe/pathComponent.h>
 #include <ufe/scene.h>
+#ifndef UFE_V2_FEATURES_AVAILABLE
+#include <ufe/sceneNotification.h>
+#endif
```
HamedSabri-adsk, line 33: These headers need to be wrapped.

lib/mayaUsd/ufe/UsdTRSUndoableCommandBase.cpp
```diff
@@ -19,6 +19,9 @@
 #include "private/Utils.h"
 
 #include <ufe/scene.h>
+#ifndef UFE_V2_FEATURES_AVAILABLE
+#include <ufe/sceneNotification.h>
+#endif
```
HamedSabri-adsk, line 24: same here

lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
```diff
@@ -26,6 +26,11 @@
 
 #include <pxr/usd/sdf/copyUtils.h>
 
+#ifndef UFE_V2_FEATURES_AVAILABLE
+#include <ufe/log.h>
+#include <ufe/sceneNotification.h>
+#endif
+
```
HamedSabri-adsk, line 33: same here

lib/mayaUsd/ufe/private/UfeNotifGuard.cpp
```diff
@@ -15,7 +15,6 @@
 //
 #include "UfeNotifGuard.h"
 
-#include <ufe/path.h>
```
seando-adsk, line 18: I noticed a few places where you have removed path.h - can you explain why? It is clearly used in this file - for example line 21. And path.h is **not** included by UfeNotifGuard.h. So it needed by this .cpp. It just so happens that it ends up being included down the line by <ufe/transform3d.h> but that is not something we should rely on. If a .cpp needs something it should include the header file.

HamedSabri-adsk: Yes I do realize that. I missed the class forwarding in UfeNotifGuard.h. The awkward formatting makes it hard to see it with a first glance. It is a few of them. 

I like the practice of going with forward declaration where it is appropriate.

seando-adsk: Yes I agree - forward declaration is the way to go. But then you still need to include <ufe.path.h> in this cpp.

PR #1108, opened 2021-01-25
make use of PXR_OVERRIDE_PLUGINPATH_NAME everywhere

pmolodo:
PXR_OVERRIDE_PLUGINPATH_NAME is already used in some places, but it's coverage is spotty - make use of it everywhere, instead of hard-coding PXR_PLUGINPATH_NAME


test/lib/usd/translators/testUsdExportCustomConverter.template.py
```diff
@@ -32,7 +32,7 @@ class testUsdExportCustomConverter(unittest.TestCase):
     def setUpClass(cls):
         # In case tests are run in parallel, make sure we export to separate
         # directories:
-        plugin_path = os.getenv("PXR_PLUGINPATH_NAME")
+        plugin_path = os.getenv("@PXR_OVERRIDE_PLUGINPATH_NAME@")
```
seando-adsk, line 35: Thats a neat trick using the configure_file

seando-adsk (approved):
Everything looks good to me.

mattyjams (approved):
Looks good to me!

test/lib/usd/translators/testUsdExportCustomConverter.template.py
pmolodo: Honestly, I wanted to find a way that DIDN'T have to use `configure_file`... but I couldn't think of any alternative.  At least this works!

PR #1110, opened 2021-01-25
[maya] error if unrecognized shadingMode on export

pmolodo:
We had a situation where shader exports were failing, but not being picked up as such, due to a mistyped shadingMode.

This will make the command error / fail if a bad shadingMode is requested


mattyjams (approved):
Looks good to me!

@JGamache-autodesk, you may want to take a quick peek too since you were working in this area not too long ago.

JGamache-autodesk (approved):

PR #1111, opened 2021-01-26
[usdMaya] add a root arg to usdExport

pmolodo:
The root arg enables exporting sub trees of the maya scene without also including all the parent nodes.

ie, if you have this Maya path:

```
|top|mid|plant
```

...and wish to export starting at `plant`, so we end up with just

```
/plant
```

in USD, then you would set `root="|top|mid|plant"` at export.

### DIFF
--- a/lib/mayaUsd/commands/baseExportCommand.cpp
+++ b/lib/mayaUsd/commands/baseExportCommand.cpp
@@ -115,6 +115,7 @@ MSyntax MayaUSDExportCommand::createSyntax()
         kExportReferenceObjectsFlag,
         UsdMayaJobExportArgsTokens->exportReferenceObjects.GetText(),
         MSyntax::kBoolean);
+    syntax.addFlag(kExportRootFlag, UsdMayaJobExportArgsTokens->root.GetText(), MSyntax::kString);
     syntax.addFlag(
         kExportSkelsFlag, UsdMayaJobExportArgsTokens->exportSkels.GetText(), MSyntax::kString);
     syntax.addFlag(
@@ -232,6 +233,22 @@ MStatus MayaUSDExportCommand::doIt(const MArgList& args)
             argData, UsdMayaJobExportArgs::GetDefaultDictionary());
 
         // Now read all of the other args that are specific to this command.
+        if (argData.isFlagSet("root")) {
+            MString stringVal;
+            argData.getFlagArgument("root", 0, stringVal);
+            std::string rootPath = stringVal.asChar();
+
+            if (!rootPath.empty()) {
+                MDagPath rootDagPath;
+                UsdMayaUtil::GetDagPathByName(rootPath, rootDagPath);
+                if (!rootDagPath.isValid()) {
+                    MGlobal::displayError(
+                        MString("Invalid dag path provided for root: ") + stringVal);
+                    return MS::kFailure;
+                }
+            }
+        }
+
         bool        append = false;
         std::string fileName;
 
--- a/lib/mayaUsd/commands/baseExportCommand.h
+++ b/lib/mayaUsd/commands/baseExportCommand.h
@@ -53,6 +53,7 @@ public:
     static constexpr auto kMergeTransformAndShapeFlag = "mt";
     static constexpr auto kStripNamespacesFlag = "sn";
     static constexpr auto kExportRefsAsInstanceableFlag = "eri";
+    static constexpr auto kExportRootFlag = "rt";
     static constexpr auto kExportDisplayColorFlag = "dsp";
     static constexpr auto kShadingModeFlag = "shd";
     static constexpr auto kConvertMaterialsToFlag = "cmt";
--- a/lib/mayaUsd/fileio/jobs/jobArgs.cpp
+++ b/lib/mayaUsd/fileio/jobs/jobArgs.cpp
@@ -314,6 +314,7 @@ UsdMayaJobExportArgs::UsdMayaJobExportArgs(
     , exportReferenceObjects(_Boolean(userArgs, UsdMayaJobExportArgsTokens->exportReferenceObjects))
     , exportRefsAsInstanceable(
           _Boolean(userArgs, UsdMayaJobExportArgsTokens->exportRefsAsInstanceable))
+    , exportRootPath(_String(userArgs, UsdMayaJobExportArgsTokens->root))
     , exportSkels(_Token(
           userArgs,
           UsdMayaJobExportArgsTokens->exportSkels,
@@ -391,6 +392,7 @@ std::ostream& operator<<(std::ostream& out, const UsdMayaJobExportArgs& exportAr
         << "exportNurbsExplicitUV: " << TfStringify(exportArgs.exportNurbsExplicitUV) << std::endl
         << "exportRefsAsInstanceable: " << TfStringify(exportArgs.exportRefsAsInstanceable)
         << std::endl
+        << "exportRootPath: " << exportArgs.exportRootPath << std::endl
         << "exportSkels: " << TfStringify(exportArgs.exportSkels) << std::endl
         << "exportSkin: " << TfStringify(exportArgs.exportSkin) << std::endl
         << "exportBlendShapes: " << TfStringify(exportArgs.exportBlendShapes) << std::endl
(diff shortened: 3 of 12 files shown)


pmolodo:
Pushed an update with `clang-format` fixes

kxl-adsk:
@pmolodo Few things we were wondering about this change:
1. How does it play with other flags on the exporter command, e.g. `selection` flag?
2. What's the behavior that the user will get when using the `export selection` menu option?
3. What's the world space after the export? (are we trying to preserve the world space or export object space)

pmolodo:
> 1. How does it play with other flags on the exporter command, e.g. `selection` flag?

It acts as an intersection filter - objects will only be exported if they are both selected and under the export root node.

> 2. What's the behavior that the user will get when using the `export selection` menu option?

Well, we haven't added gui support for the options menu... so currently, it can't be accessed through the menu option.  However, it does work if added to the options flag when using the file translator, ie:

```
cmds.file('export_with_root.usda', exportSelected=True, type='USD Export', options='root=|Top')
```

...and in that case, it works the same as 1) - it's an intersection.

If we need / want the export-options-GUI suport, I can add that pretty easily - wasn't sure if you want want to consider this a more "advanced" option or not.

> 3. What's the world space after the export? (are we trying to preserve the world space or export object space)

Things are exported in the local space of the root node - ie, the transforms of the root node an any of it's parents are not written out or applied.  This was the easiest to implement, and worked well for our use cases.

kxl-adsk:
@pmolodo After running this by folks internally it turns out that we have a design for a very similar feature. We haven't prioritized working on it, but I hope we can align this pull request with the direction we are planning to take.

In your pull-request, the root flag is taking a single string and acts as a filter. We thought about taking a slightly different approach:
* introduce `includeAncestors` boolean flag on the export command; by default this flag is true
* when exporting selection, `includeAncestors` will decide if upstream (in DAG hierarchy) from selected nodes should be exported
* multiple nodes can be selected and should be exported as "roots" when `includeAncestors` is set to false and nodes are not part of each other's downstream hierarchy

In our designs, we also have an option to choose between preserving the world space or exporting object space. We agree that starting with object space is a good default. At some other point in time, an additional mode can be added to preserve the world space.

pmolodo:
They're definitely related, but they both are able to handle situations that the other is not.  The `includeAncestors` proposal is obviously able to create multiple independent roots, which `exportRoot` is not able to do.  However, `exportRoot` allows you to selectively prune pieces both above *and* below a target, while preserving relative hierarchy and positioning... which  `includeAncestors` does not.

As an example, suppose you have these maya objects:

```
|A|B|E
|A|B|F|X
|A|B|F|Y
|A|B|G
```

...and the only portions that you wish to export are `|A|B|E` and `|A|B|F|Y` - but you want to retain their relative postioning - both in the hierarchy, and transformation-wise.  Further, you know that `|A` is an extra level of hierarchy introduced for maya-convenience only.  So when exporting to USD, what you want to end up with is:

```
/B/E
/B/F/Y
```

Export root makes this easy, by selecting `["|A|B|E", "|A|B|F|Y"]`, and doing `exportRoot="|A|B"`.  You can't really get the same behavior with `includeAncestors`, because if you select the same objects, you would end up with

```
/E
/Y
```

...and if you select `|A|B`, you would include unwanted prims like `/B/F/X` and `/B/G`.

So... given that they do different things, which should we prefer? (Because I do agree that they overlap enough that implementing both could be confusing - both for end-users trying to figure out which to use when, and for developers, figuring out the right thing to do if both options are used together.)

To me, these questions stand out:

- How do you handle xforms?
  You brought up object-space vs world-space, but there's another factor to consider, which is preserving vs discarding relative transformations - `exportRoot` will preserve relative positioning of all selected objects, while `includeAncestors` will generally discard it.  I say "generally" because if you have a world-space xform option, then their relative positioning will remain intact - but at the cost of potentially including unwanted extra levels of xform above your selected objects.  My gut tells that what would be wanted "most often" would be to preserve the relative positioning of objects, but discard the xform information of non-exported parent objects.  `exportRoot`, in local-space mode, allows us to do that by default.
- Which scenario is going to be desired more frequently?
  Before we started working a lot with USD, I might not have had a strong opinion.  However, now that we have, we've realized that the fact that references can only target a single root-level prim means we strongly prefer exports  with only one root level prim - particularly when exporting component-level prims, which is what the maya-usd exporter currently specializes in.  (We even an extra flag (parentScope) to enforce and control this).  Also, our own workflows have shown us that it's useful to be able to have extra levels of maya-only hierarchy to organize things in your maya scene, that you can easily "discard" on USD export... but then still be able to selectively pick out objects below to export (depending on export options, scene setup, etc).  So far, we haven't come up with any use cases where we've wanted to have the sort of scenarios allowed by only `includeAncestors`.  And if they did come up, the next question would be...
- How hard is it to get the functionality of the "other option"?
  - If you start with `exportRoot`, and want to implement behavior similar to `includeAncestors`:
    This is relatively easy - you just do a loop over all the items you would have selected for `includeAncestors`, calling `mayaUSDExport` for each, and setting the `exportRoot` to be the same as the (single) selected item for each.  Honestly, given that maya is generally used for component-level export, and `includeAncestors` throws out relative xform information anyway, that's probably all you'd need to do - you'd then just rely on whatever higher-level USD scaffolding you generally use to assemble those components as desired.  However, if for whatever reason you really wanted them assembled in one layer, it would be easy to make a layer which references (or sub-layers) all those just-exported items into a single new `.usd`.  And, if for some reason you really wanted them in one FLAT layer, it would be pretty easy to use `usdcat` (or one of the other options for flattening), or just go straight to `SdfCopySpec`.
  - If you start with `includeAncestors`, and want to implement behavior similar to `exportRoot`:
    I can think of two ways to do this:
    - Proceed in a loop, with multiple calls to `mayaUSDExport` like described above.  However, you'd then need to recreate the missing levels of relative hierarchy somehow - either with some sort of chaser, or via a post-process where you create a new layer and use referencing.  In either situation, though, you're having to recreate some of the maya export machinery, since at minimum you'd need to recreate the xform information, and ideally you'd want to export ALL the same information that the "normal" maya exporter would make.
    - First traverse the maya hierarchy to calculate all the unwanted bits; then, since there's not currently a way to blacklist maya paths on export (only whitelist, via selection), you'd have to delete all the those unwanted bits, then select the desired `exportRoot`, and export with `includeAncestors`.  Obviously, having to alter the maya scene (and especially making such destructive changes - ones that I probably wouldn't "trust" undo to always get right) is a downside.
- Which will be faster to implement?
  Ok, that's a bit of an unfair advantage, since one is already implemented, but hey, it IS a consideration...


Having said all that... the ideal solution is, of course, some option which would allow us to handle either scenario with a single, relatively uncomplicated new option.  However, barring that inspiration - my (admittedly, biased) opinion is that `exportRoot` is ultimately a more flexible option, which will give more end-users what they want with less effort.

wtelford:
Reading these proposals, I have a thought on how we can combine the concepts, but keep the number of flags to a minimum.

1. We get rid of the includeAncestors flag.
2. Instead of exportRoot, we have an exportRoots or exportFromRoots (the key here is that its plural) allowing users to specify multiple roots. 
3. exportRoots should take an empty string '', the default, which exports each selected object as its own root (the current export behavior we inherited from Pixar)
4. exportRoots should take a wildcard string * which means All roots. This would perform the includeAncestors behavior defined above.
5. The UI will only expose a boolean option to switch between '' and *, which switches you between our current export behavior (Pixar style) and the incudeAncestor behavior (AL style) exports.
6. From the export command a user would be able to pass a comma separated list of roots to accomplish the behavior Paul is proposing.

Thoughts?

pmolodo:
That sounds like a good solution to me!  For some reason I thought the situation was reversed (that the Pixar plugin preserved parents by default, and AL didn't), but I could be wrong, and it doesn't really matter for this discussion, anyway.  But yeah, this seems like a good way to have maximum flexibility...

wtelford:
It’s very possible I have my Pixar and AL behavior flipped.

Ennosch:
Hi, 
I wanted to mention that I am very much looking forward to test this :)

Do you believe it could be included in the next mayaUSD release ?
Or available for the mayaUSD plugin that comes with  maya 2022  ?

daviddeacon:
I really could do with the exportRoot arg, as it replicates what I'm currently doing with our alembic pipeline, so it would be really useful as I migrate to USD.

Is there an ETA when this will be included in a public release of Maya USD plugin?

Thanks,
David.

kxl-adsk:
Given that there was no progress on this change and that we are getting customer requests and internal requests to support it we opened https://github.com/Autodesk/maya-usd/pull/1592 to deliver the feature and address the requested changes.

PR #1112, opened 2021-01-26
MAYA-105827 - Compute extents for modified GPrims

JGamache-autodesk:
The UsdGeomBoundable requirements state that extents must be explicitly authored if any GPrim attribute that affect extents is authored.
This PR tries to keep the extents in sync by recomputing them at notification time.


lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -540,7 +540,10 @@ void HdVP2Mesh::Sync(
     }
 
     if (HdChangeTracker::IsExtentDirty(*dirtyBits, id)) {
-        _sharedData.bounds.SetRange(delegate->GetExtent(id));
+        // Do not trust GPrim extents until we add code to restore the integrity of the data model
```
HdC-adsk, line 543: Looks good. Please also test the bounding box display mode in the 3d view.

HdC-adsk: Maybe better to add a doc here saying VP2 will compute the bounds when none is provided.

HdC-adsk: Second thought: bounding box display mode might be broken by this change.

HdC-adsk: If it is broken, we can keep storing the extent in the shared data, but each draw item can opt in/out on using it.

williamkrick (changes requested):
I'll leave this in request changed until Huidong's question gets answered.

williamkrick (approved):

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
JGamache-autodesk: Bounding box was indeed broken (as well as selection when switching to wireframe using wrong BB). This should fix it.

williamkrick (approved):

HdC-adsk (approved):

HdC-adsk (approved):

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -1187,6 +1190,58 @@ void MayaUsdProxyShapeBase::_OnStageContentsChanged(const UsdNotice::StageConten
...
+        const UsdPrim&   changedPrim = stage->GetPrimAtPath(changedPrimPath);
+        UsdGeomBoundable boundableObj = UsdGeomBoundable(changedPrim);
+        if (!boundableObj) {
+            continue;
+        }
+
+        // If the attribute is not part of the primitive schema, it does not affect extents
+        const UsdPrimDefinition& primDefinition = changedPrim.GetPrimDefinition();
+        if (!primDefinition.GetSchemaAttributeSpec(changedPropertyToken)) {
+            continue;
```
JGamache-autodesk, line 1224: Filters out the transform notifications when manipulating USD objects.

williamkrick (approved):

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -1187,6 +1190,58 @@ void MayaUsdProxyShapeBase::_OnStageContentsChanged(const UsdNotice::StageConten
...
+
+    // Recompute the extents of any UsdGeomBoundable that has authored extents
+    const auto& stage = notice.GetStage();
+    if (stage != getUsdStage()) {
+        TF_CODING_ERROR("We shouldn't be receiving notification for other stages than one "
+                        "returned by stage provider");
+        return;
+    }
+
+    UsdEditContext editContext(stage, stage->GetSessionLayer());
```
kxl-adsk, line 1202: Here is a bit of a problem. The session layer is not serialized together with the root layer. 

I would assume that this notification will happen right after the authoring operation and for that, we already have the right edit target set. What's the reason for changing the edit target?

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -1187,6 +1190,58 @@ void MayaUsdProxyShapeBase::_OnStageContentsChanged(const UsdNotice::StageConten
...
+            UsdGeomGprim::GetSchemaAttributeNames(true).cend());
+        if (ignoredAttributes.count(changedPropertyToken) > 0) {
+            continue;
+        }
+
+        UsdAttribute extentsAttr = boundableObj.GetExtentAttr();
+        if (extentsAttr && extentsAttr.HasValue()) {
+            VtVec3fArray extent(2);
+            if (UsdGeomBoundable::ComputeExtentFromPlugins(
+                    boundableObj, UsdTimeCode::Default(), &extent)) {
```
kxl-adsk, line 1240: Here is another challenging place. What if the extent is animated?

lib/mayaUsd/nodes/proxyShapeBase.cpp
JGamache-autodesk: Yes, this line introduces bugs. I will remove it.

lib/mayaUsd/nodes/proxyShapeBase.cpp
JGamache-autodesk: I must say that reacting by notification is probably not appropriate. There is not enough context to know if the values is/was animated which forces writing a single non-animated extent value. This makes the extent completely wrong if another untouched parameter is kept animated (radius and length of a capsule, for example). There is also no information stored for Undo, and no mechanism to store the undo data of an animated extent (same issue appears in Ufe Undo handler for `TypedUsdAttribute::Set`).

We could build a set of affected boundables and wait until the Compute function of the shape node is called. At this time we might have a better overall view of the state. We still end up with animation issues: we are expected to compute the extent at the current frame, but might be forced to recompute for the whole clip duration since we have absolutely no idea about the final animated state of the extent, and more specifically, which attributes affect the extent.

kxl-adsk (approved):
We should request more testing to validate that there is no impact on performance for interactive manipulation.

kxl-adsk:
@JGamache-autodesk I know we are still checking the performance impact, but I took this change for a spin. We still have some work. Compilation fails with USD 20.02 due to change done in the prim (GetPrimDefinition is returning a different type in 20.02) and in sdfpath (GetAsString doesn't exist).

It would also be great to add a regression test since we don't have it tested by anything existing.

kxl-adsk (approved):

PR #1113, opened 2021-01-26
MAYA-109414 workaround for material consolidation update issue

HdC-adsk:
Workaround for a material consolidation update issue in VP2. Before USD 0.20.11, a Rprim will be recreated if its material has any change, so everything gets refreshed and the update issue gets masked. Once the update issue is fixed in VP2 we will disable this workaround.


williamkrick (approved):

HdC-adsk:
The PF failure seems to be relevant to OSX Python 2 mentioned by @kxl-adsk in today's meeting. So we can ignore it.

kxl-adsk (approved):

PR #1114, opened 2021-01-26
Spinell/maya 107554/revert to file

spinell-adsk:
MAYA-107554 - As a user, I'd like to see 'Revert to File' instead of 'Discard Edits'

- Rename context menu  'Discard Edits' to 'Revert to File'
- 'Revert to File' menu always enable for layer on disc no matter is the layer has edit or not.
- 'Revert to File' menu not available for anonynous layer.
- Make 'Clear' option available for anonynous layer


spinell-adsk:
close and replaced by https://github.com/Autodesk/maya-usd/pull/1115

Issue #1103, opened 2021-01-22
g_StageMap.isDirty() error when saving layer to disk first time


cfmoore007:
**Describe the bug**
After creating a simple scene setup, while saving the layer to disk an error is returned.
If I save the maya file after this error, and re-open the file, everything is still preserved.
Unsure if the error has any other consequences ...

**Steps to reproduce**
1. Create stage from New Layer
2. Create Xform
3. Create Def
4. Add Prim Sphere ( note this also occurs if you add reference )
5. Go to USD Layer Editor
6. Save Layer
7. Witness Error :

select -r |stage1|stageShape1,/Xform1 ;
select -r |stage1|stageShape1,/Xform1/Def1 ;
setAttr -type "string" |stage1|stageShape1.filePath "F:/Documents/maya/projects/default/scenes/anonymousLayer1_MLtsUA.usd";
// Error: Failed verification: ' g_StageMap.isDirty() ' -- Coding Error in onStageSet at line 370 of F:\Repos\maya-usd\lib\mayaUsd\ufe\StagesSubject.cpp // 

**Expected behavior**
No error returned and successful save to disk ( though it appears it still wrote out the file )

**Attachments**
![LayerSaveError](https://user-images.githubusercontent.com/2152250/105538664-bf213480-5cc1-11eb-9ebc-731ff1e8db2a.png)

**Specs (if applicable):**
 - OS & version [Windows 10]
 - Compiler & version [VS2017]
 - Maya version [e.g. Maya Beta (latest)]
 - Maya USD commit SHA [e.g. dev at 0.7.0]
 - Pixar USD commit SHA [e.g. dev at 21.02]

**Additional context**

cfmoore007:
Part 2 of this is :

If you do not save the maya file after this error, and you do a file -new, the following error is returned :

// Clearing USD Stage Cache // 
// Result: untitled // 
# Error: '|stage1|stageShape1' is not a path to a proxy shape node. -- Coding Error in proxyShapeHandle at line 36 of F:\Repos\maya-usd\lib\mayaUsd\ufe\UsdStageMap.cpp #

HamedSabri-adsk:
> Unsure if the error has any other consequences

Hi @cfmoore007 we discovered this issue earlier this week and have an internal jira ticket against it. 

As for the consequences, One thing I discovered was that the transformation authoring in the stronger layer ( e.g session layer ) is busted now.

[...]

PR #1116, opened 2021-01-26
Fixed crash caused by out of date stage map

kxl-adsk:
When using the `onStageInvalidate` notification, extra care has to be taken to NOT trigger any notification that would result in pulling on stage map. This notification is sent during dirty propagation, and we can't trigger compute from it (number one DG rule).
We didn't fully follow this rule and started to send `Ufe::SubtreeInvalidate`. One of the listeners, i.e. Outliner, was trying to refresh the tree by pulling on the list of stages. This caused a lack of refresh and what's more important, the potential use of released memory in future calls.

The fix is to delay this notification till later when we know the stage map has been already populated.

### DIFF
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -392,9 +392,9 @@ void StagesSubject::onStageSet(const MayaUsdProxyStageSetNotice& notice)
                 Ufe::Scene::instance().notify(Ufe::SubtreeInvalidate(sceneItem));
             }
         }
+#endif
 
         fInvalidStages.clear();
-#endif
 
         stageSetGuardCount = false;
     }
@@ -404,14 +404,12 @@ void StagesSubject::onStageInvalidate(const MayaUsdProxyStageInvalidateNotice& n
 {
     afterOpen();
 
-#ifdef UFE_V2_FEATURES_AVAILABLE
     auto p = notice.GetProxyShape().ufePath();
     if (!p.empty()) {
         // We can't send notification to clients from dirty propagation.
         // Delay it till the new stage is actually set during compute.
         fInvalidStages.insert(p);
     }
-#endif
 }
 
 #ifdef UFE_V2_FEATURES_AVAILABLE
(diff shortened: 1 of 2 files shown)


kxl-adsk:
Linking with #1103

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -391,10 +407,9 @@ void StagesSubject::onStageInvalidate(const MayaUsdProxyStageInvalidateNotice& n
...
     auto p = notice.GetProxyShape().ufePath();
     if (!p.empty()) {
-        Ufe::SceneItem::Ptr sceneItem = Ufe::Hierarchy::createItem(p);
-        if (sceneItem) {
-            Ufe::Scene::instance().notify(Ufe::SubtreeInvalidate(sceneItem));
-        }
+        // We can't send notification to clients from dirty propagation.
+        // Delay it till the new stage is actually set during compute.
+        fInvalidStages.insert(p);
     }
```
seando-adsk, line 413: I see you didn't put fInvalidStages inside the ifdef V2 block. So I think that either all the code that uses this var should be inside an ifdef or remove the ifdef from this block. The block above should be ifdef protected because subtreeinvalidate is V2 only. But then line 396 should be moved outside the block.

kxl-adsk: It was mostly to leave the header file cleaner since code that is not guarded has no real performance impact (very low-frequency operation on empty set). It also builds with all flavors of the build, because the entire `ufe` folder is not included when building without UFE.

Let me know if you feel strongly about it and I can add the guards.

seando-adsk: My one concern is that the population of fInvalidStages is done inside an ifdef V2 block. So if someone later adds code to use this fInvalidStages var and that code is not inside a V2 block, the new code will not work since the set will never have anything in it.

kxl-adsk: Addressed in ea81f449b09455330246e2d5a90cd918ef66b281

seando-adsk (approved):

PR #1119, opened 2021-01-26
Implement viewport selection highlighting of point instances and point instances pick modes

mattyjams:
Building on #1027, this PR adds support for selection highlighting of individual point instances when using the Viewport 2.0 render delegate. It also implements all of the "Prims", "Models", "Instances", and "Prototypes" pick modes that are available in the "View -> Pick mode" menu in `usdview`. A "Kind" pick mode unique to the render delegate represents the existing behavior and is the default. Similar to selection kind, the pick mode can be changed using the `UsdPickMode` `optionVar`.


kxl-adsk:
@mattyjams Can you provide a video to demo your new proposed additions? I would like to loop in the design here before starting the code review process.

mattyjams:
> @mattyjams Can you provide a video to demo your new proposed additions? I would like to loop in the design here before starting the code review process.

@kxl-adsk: I'm not sure a video would demonstrate it all that clearly really. Like selection kind, the pick mode is only adjustable using an `optionVar` right now. I think it's probably most obvious to look at the PickModes test and check out the baseline image and the UFE path selected for each pick mode to see what they do. I'll try to distill that from the source files below.

The scene is a single `PointInstancer` prim that uses seven different prototypes. Each prototype is just a cube with a different displayColor and material. The `PointInstancer` then creates two instances of each prototype. The test simulates a click in the center of the viewport, hitting the lower left of the two green cubes.

Scene file:
https://github.com/Autodesk/maya-usd/blob/9a01aa58c303458c7da5e0fdf270e62b2881fbde/test/testSamples/pointInstances/PointInstancer_Grid_14.ma

The default behavior when the `UsdPickMode` `optionVar` is not set, or is set to something other than one of `Prims`, `Models`, `Instances`, or `Prototypes`, is to use what I'm calling the `Kind` pick mode, which is exactly the same pick behavior from before these changes. If you don't have anything set for `UsdSelectionKind`, then the `Kind` and `Prototypes` pick modes are exactly the same.

Pick result in `Kind` pick mode with no `UsdSelectionKind` set (same result as in `Prototypes` pick mode):
Pick result in Kind pick mode

Selected UFE path:
```
|UsdProxy|UsdProxyShape,/PointInstancerGrid/PointInstancer/prototypes/GreenCube/Geom/Cube
```

---

Pick result in `Prims` pick mode:
Pick result in Prims pick mode

Selected UFE path:
```
|UsdProxy|UsdProxyShape,/PointInstancerGrid/PointInstancer
```

---

Pick result in `Models` pick mode:
Pick result in Models pick mode

Selected UFE path:
```
|UsdProxy|UsdProxyShape,/PointInstancerGrid/PointInstancer/prototypes/GreenCube
```

---

Pick result in `Instances` pick mode:
Pick result in Instances pick mode

Selected UFE path:
```
|UsdProxy|UsdProxyShape,/PointInstancerGrid/PointInstancer/3
```

---

Again, the goal with these pick modes was to mirror their counterparts in `usdview`.

ppt-adsk:
FWIW, my non-product design, single developer opinion is that this picking behavior seems reasonable to me.  Of course, the people actually responsible for this kind of thing on our end will be weighing in with a real opinion :)

Not sure about the use of "Kind" to label the pick mode where the prototype geometry is picked, but if that's the USD terminology for it then that's fine.

mattyjams:
> Not sure about the use of "Kind" to label the pick mode where the prototype geometry is picked, but if that's the USD terminology for it then that's fine.

Thanks @ppt-adsk! And I hear you about the `Kind` label. I just invented that because there is no such corresponding pick mode on the USD side. I needed to keep it distinct from the `Prototypes` pick mode, since in USD `Prototypes` mode will never consider kind, so I needed some term to use for the original pick resolution behavior. Definitely open to suggestions if you all think there's a better way to handle that.

mattyjams:
We discussed offline how these two optionVars control pick resolution and decided to fully separate from the `usdview` behavior.

The optionVar that was previously `UsdPickMode` is now `UsdPointInstancesPickMode` with the possible values of `PointInstancer`, `Instances`, or `Prototypes`. If the object picked in the viewport is a point instance, this optionVar controls what that pick resolves to (the `PointInstancer` prim, a specific point instance scene item, or the prototype prim being instanced, respectively). When we're in `Instances` mode and we resolve to a point instance, that's the end of the pick resolution process. Otherwise if there is a `UsdSelectionKind` specified, we start searching up the hierarchy from whatever we resolved to for a kind match. This results in a more flexible way to use both the point instances pick mode and the selection kind in combinations that `usdview` doesn't support while still allowing us to express the pick modes that `usdview` does support.

Updating my comment from above, here is the summary based on the new changes:

---

The scene is a single `PointInstancer` prim that uses seven different prototypes. Each prototype is just a cube with a different displayColor and material. The `PointInstancer` then creates two instances of each prototype. The test simulates a click in the center of the viewport, hitting the lower left of the two green cubes.

Scene file:
https://github.com/Autodesk/maya-usd/blob/dev/test/testSamples/pointInstances/PointInstancer_Grid_14.ma

The default behavior when the `UsdPointInstancesPickMode` `optionVar` is not set or is invalid is to use `PointInstancer`. Note that this a change versus the current pick resolution which is effectively `Prototypes`.

Pick result in `PointInstancer` point instances pick mode (the default if `UsdPointInstancesPickMode` is unset):
Pick result in PointInstancer point instances pick mode


Selected UFE path:
```
|UsdProxy|UsdProxyShape,/PointInstancerGrid/PointInstancer
```

---

Pick result in `Instances` point instances pick mode:
Pick result in Instances point instances pick mode

Selected UFE path:
```
|UsdProxy|UsdProxyShape,/PointInstancerGrid/PointInstancer/3
```

---

Pick result in `Prototypes` point instances pick mode:
Pick result in Prototypes point instances pick mode

Selected UFE path:
```
|UsdProxy|UsdProxyShape,/PointInstancerGrid/PointInstancer/prototypes/GreenCube/Geom/Cube
```

---

Pick result when configured to match `usdview`'s `Models` pick mode (`UsdPointInstancesPickMode=Prototypes`, `UsdSelectionKind=model`):
usdview's Models pick mode

Selected UFE path:
```
|UsdProxy|UsdProxyShape,/PointInstancerGrid/PointInstancer/prototypes/GreenCube
```

---

Pick result with `UsdPointInstancesPickMode=PointInstancer` and `UsdSelectionKind=model`, a configuration not directly supported by `usdview`. In the test scene's USD, the root prim has kind `component`, which `IsA` `model`, whereas the `PointInstancer` prim has kind `subcomponent`, which is **not** `model`:
Non-instance models

Selected UFE path:
```
|UsdProxy|UsdProxyShape,/PointInstancerGrid
```

kxl-adsk:
Thank you @mattyjams, I started the review process.

We also try to clean-up optionVars and make them use a consistent `mayaUsd_` prefix and ease the discovery by locating them all in [one place](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/base/tokens.h#L33). Do you mind moving and renaming:
UsdPointInstancesPickMode -> mayaUsd_PointInstancesPickMode 
UsdSelectionKind -> mayaUsd_SelectionKind

HdC-adsk:
@mattyjams Thank you Matt.

From first glance, USDView only has a single control on the selection behavior, its `Pick mode` menu has `Prims`, `Models`, `Instances`, and `Prototypes`. Maybe you've explained it in the above design doc but I missed it, could you elaborate why we need to introduce another control in addition to the current `UsdSelectionKind`? Is there a chance that we use only one control to support all the modes like USDView?

mattyjams:
> Thank you @mattyjams, I started the review process.
> 
> We also try to clean-up optionVars and make them use a consistent `mayaUsd_` prefix and ease the discovery by locating them all in [one place](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/base/tokens.h#L33). Do you mind moving and renaming:
> UsdPointInstancesPickMode -> mayaUsd_PointInstancesPickMode
> UsdSelectionKind -> mayaUsd_SelectionKind

@kxl-adsk: Sounds good. I can move and rename the optionVars. Let me take care of that in a rebase before you all get too deep into the review though, since that will cascade across a bunch of the commits here.

HdC-adsk:
> @mattyjams Thank you Matt.
> 
> From first glance, USDView only has a single control on the selection behavior, its `Pick mode` menu has `Prims`, `Models`, `Instances`, and `Prototypes`. Maybe you've explained it in the above design doc but I missed it, could you elaborate why we need to introduce another control in addition to the current `UsdSelectionKind`? Is there a chance that we use only one control to support all the modes like USDView?

Please ignore this question. I just received the offline email thread which exactly answers my question.

mattyjams:
Ok, should be ready for review now. I moved both optionVar tokens to `mayaUsd/base/tokens.h`, and I also wrapped those tokens to Python for use in the tests.

lib/mayaUsd/base/tokens.h
```diff
@@ -28,9 +28,30 @@ PXR_NAMESPACE_OPEN_SCOPE
...
 // Tokens that are used as optionVars in MayaUSD
 //
+// clang-format off
 #define MAYA_USD_OPTIONVAR_TOKENS                        \
     /* Always target a session layer on a mayaUsdProxy*/ \
-    (mayaUsd_ProxyTargetsSessionLayerOnOpen)
+    ((ProxyTargetsSessionLayerOnOpen, "mayaUsd_ProxyTargetsSessionLayerOnOpen")) \
+    /* The kind to be selected when viewport picking. */ \
+    /* After resolving the picked prim, a search from */ \
+    /* that prim up the USD namespace hierarchy will  */ \
```
HdC-adsk, line 37: USD namespace -> USD scene ?

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -568,9 +568,8 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
                 if (SdfLayerRefPtr rootLayer = SdfLayer::FindOrOpen(fileString)) {
                     SdfLayerRefPtr sessionLayer = computeSessionLayer(dataBlock);
 
-                    bool targetSession
-                        = MGlobal::optionVarIntValue(UsdMayaUtil::convert(
-                              MayaUsdOptionVars->mayaUsd_ProxyTargetsSessionLayerOnOpen))
+                    bool targetSession = MGlobal::optionVarIntValue(UsdMayaUtil::convert(
+                                             MayaUsdOptionVars->ProxyTargetsSessionLayerOnOpen))
                         == 1;
```
HdC-adsk, line 573: TBH, both styles forced by clang-format look a bit weird. Maybe adding a parenthesis pair to the right hand side of the assignment could make it better?

HdC-adsk: Or declare another variable to hold the return value of UsdMayaUtil::convert()?

lib/mayaUsd/nodes/stageNode.cpp
```diff
@@ -112,7 +112,7 @@ MStatus UsdMayaStageNode::compute(const MPlug& plug, MDataBlock& dataBlock)
             UsdStageCacheContext ctx(UsdMayaStageCache::Get(loadAll));
 
             bool targetSession = MGlobal::optionVarIntValue(UsdMayaUtil::convert(
-                                     MayaUsdOptionVars->mayaUsd_ProxyTargetsSessionLayerOnOpen))
+                                     MayaUsdOptionVars->ProxyTargetsSessionLayerOnOpen))
                 == 1;
```
HdC-adsk, line 116: Same concern about the style.

lib/mayaUsd/python/module.cpp
```diff
@@ -31,8 +31,9 @@ TF_WRAP_MODULE
     TF_WRAP(ReadUtil);
     TF_WRAP(RoundTripUtil);
     TF_WRAP(StageCache);
+    TF_WRAP(Tokens);
+    TF_WRAP(UsdUndoManager);
     TF_WRAP(UserTaggedAttribute);
     TF_WRAP(WriteUtil);
     TF_WRAP(XformStack);
-    TF_WRAP(UsdUndoManager);
```
HdC-adsk, line 37: 👍

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -62,6 +70,29 @@
...
+    // UsdSceneItems with UFE paths to a PointInstancer prim and a non-negative
+    // instanceIndex for the specific point instance. In this mode, any setting
+    // for selection kind is ignored.
+    Instances,
+    // The prototype being instanced by the point instance is picked. If a
+    // selection kind is specified, the traversal up the hierarchy looking
+    // for a kind match will begin at the prototype prim.
+    Prototypes
+};
+
```
HdC-adsk, line 95: Can you confirm the doc here can be recognized by doxygen? If not, please replace `//` with `//!` .

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -120,6 +151,57 @@ TfToken GetSelectionKind()
...
+TF_DEFINE_PRIVATE_TOKENS(
+    _pointInstancesPickModeTokens,
+
+    (PointInstancer)
+    (Instances)
+    (Prototypes)
+);
+// clang-format on
+
+UsdPointInstancesPickMode GetPointInstancesPickMode()
```
HdC-adsk, line 164: Please add doc for the method.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -120,6 +151,57 @@ TfToken GetSelectionKind()
...
+//
+// If neither the prim itself nor any of its ancestors above it in the
+// namespace hierarchy have an authored kind that matches, an invalid null
+// prim is returned.
+UsdPrim GetPrimOrAncestorWithKind(const UsdPrim& prim, const TfToken& kind)
+{
+    UsdPrim iterPrim = prim;
+
+    while (iterPrim) {
+        TfToken primKind;
```
HdC-adsk, line 194: If you don't mind please move the primKind declaration out of the loop.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -138,14 +220,15 @@ void PopulateSelection(
         return;
     }
 
-    SdfPath usdPath = usdItem->prim().GetPath();
+    SdfPath   usdPath = usdItem->prim().GetPath();
+    const int instanceIndex = usdItem->instanceIndex();
```
HdC-adsk, line 224: Having the instanceIndex declared here might cause unused variable warning on older USD Imaging API version. Please move it down into the relevant conditional compilation block.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -589,9 +672,11 @@ void ProxyRenderDelegate::_Execute(const MHWRender::MFrameContext& frameContext)
     if (inSelectionPass && !inPointSnapping) {
         _globalListAdjustment = GetListAdjustment();
         _selectionKind = GetSelectionKind();
+        _pointInstancesPickMode = GetPointInstancesPickMode();
     } else {
         _globalListAdjustment = MGlobal::kReplaceList;
         _selectionKind = TfToken();
+        _pointInstancesPickMode = UsdPointInstancesPickMode::PointInstancer;
```
HdC-adsk, line 679: Did you test the behavior of point snapping on point instances? What is the expected behavior?

HdC-adsk: I check the getInstancedSelectionPath() implementation. This value shouldn't matter for point snapping. What matters for point snapping is the hit points which will be calculated by VP2.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -751,36 +847,94 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
         indexPath = _sceneDelegate->GetPathForInstanceIndex(rprimId, 0, nullptr);
         usdPath = _sceneDelegate->ConvertIndexPathToCachePath(indexPath);
     }
+
+    // The "Instances" point instances pick mode is not supported for
+    // USD_IMAGING_API_VERSION < 13 (core USD versions earlier than 20.05), so
+    // no setting of topLevelPath or topLevelInstanceIndex here.
```
HdC-adsk, line 853: What behavior is expected in this scenario?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -75,6 +75,11 @@ enum HdVP2SelectionStatus
     kFullyLead          //!< A Rprim is lead (meaning fully lead for instanced Rprims)
 };
 
+// Forward declaration of UsdPointInstancesPickMode so that it can be added as
+// a private member of ProxyRenderDelegate. The values and usage of this enum
+// are otherwise an implementation detail.
+enum class UsdPointInstancesPickMode;
```
HdC-adsk, line 81: We are trying to put doc in the headers instead of cpp files. I think it is probably better to have both definition and doc in the header for this case.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -751,36 +847,94 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
...
-        UsdPrim prim = _proxyShapeData->UsdStage()->GetPrimAtPath(usdPath);
-        while (prim) {
-            TfToken kind;
-            if (UsdModelAPI(prim).GetKind(&kind) && KindRegistry::IsA(kind, selectionKind)) {
-                usdPath = prim.GetPath();
-                break;
+    UsdPrim prim = _proxyShapeData->UsdStage()->GetPrimAtPath(usdPath);
+    UsdPrim topLevelPrim = topLevelPath.IsEmpty()
+        ? UsdPrim()
+        : _proxyShapeData->UsdStage()->GetPrimAtPath(topLevelPath);
```
HdC-adsk, line 875: Are you trying to make the validity check explicit? Otherwise, the GetPrimAtPath call can handle the validity check itself and return an invalid prim when the path is empty.

HdC-adsk: If it is not going to be changed later, `topLevelPrim` should be a const object.

HdC-adsk (changes requested):
Some general notes:

1) Did you test the viewport selection performance for massive objects and instances? Does it have a major impact to non-point-instancer assets?

2) A question for @kxl-adsk How do we manage image data in the repo? Do we encourage use of `git lfs`?

kxl-adsk:
@HdC-adsk There used to be some limitations around Git Large File Storage and the build infrastructure that we are using. As long as images are small, we should be ok without `lfs`.

ppt-adsk (approved):
Great tests!

lib/mayaUsd/base/tokens.h
mattyjams: Thanks @ppt-adsk, yeah, I think "namespace" is a little more precise here. See also some discussion on this in the USD object model docs here:
https://graphics.pixar.com/usd/docs/api/_usd__page__object_model.html#Usd_OM_UsdPrim

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
mattyjams: Admittedly, no, I did not test point snapping against a point instancer. I'm not sure whether I'm interpreting your suggestion correctly though. Are you saying we don't need to set `_pointInstancesPickMode` in the `else` case here and this should be removed?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
mattyjams: You definitely will not be able to pick individual point instances for USD 20.02. If you're in `PointInstancer` or `Instances` pick mode, we'll still be able to identify if you've picked an instance proxy and can traverse up to the nearest non-proxy prim. In `Prototypes` pick mode, the behavior should be no different from the current behavior. We'll also still apply selection kind filtering afterwards in all cases.

This was less about trying to support a specific behavior for USD 20.02 and just trying to make sure the functionality degrades somewhat gracefully.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
mattyjams: Addressed in ac6bf39fd219df491dd995fdb6eb0367c97ca46f.

ppt-adsk (approved):

lib/mayaUsd/nodes/proxyShapeBase.cpp
mattyjams: Addressed in 991bdb68dc8ca260b2174f33b705f8a0675a4b17.

lib/mayaUsd/nodes/stageNode.cpp
mattyjams: Addressed in 991bdb68dc8ca260b2174f33b705f8a0675a4b17.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
mattyjams: Addressed with the move to the header file in 33c80b7ec3e77cb693bab4a7c8cab853741a0f16.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
mattyjams: Addressed in 07a506398e7d60af6e5d1486c47f3a859894e272.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
mattyjams: Addressed in 33c80b7ec3e77cb693bab4a7c8cab853741a0f16.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
mattyjams: Addressed in ca2d4cdee6afa7d5293dbd01d269664651836dbc.

mattyjams:
> 1. Did you test the viewport selection performance for massive objects and instances? Does it have a major impact to non-point-instancer assets?

At least in my testing so far, I haven't seen the pick mode handling incurring significant overhead and impacting performance, though I suppose it may depend on your definition of "massive". Do you have an asset handy that I should test with?

I would guess that in the case of single selections, the overhead of any particular point instances pick mode would likely be dwarfed by the rest of the selection machinery. For large marquee selections that result in many scene items getting selected, it could possibly add up though? I'm happy to take a performance optimization pass on this later if we see this becoming a significant bottleneck.

HdC-adsk (approved):

PR #1120, opened 2021-01-26
Adding tools to help with clang-format to repo

pmolodo:
This adds two things:

- an executable to easily run clang-format on the whole repo (or a specific subset of it)
- adds a configuration for use with the [pre-commit tool](https://pre-commit.com/), to make it easy to install a hook to run clang-format before commit

I started [a discussion about this](https://github.com/Autodesk/maya-usd/discussions/1118), to see if there was interest in this, but decided I was going to use it myself anyway, so I might as well just make the PR, and people can decide.

If you haven't used [pre-commit](https://pre-commit.com/), before it's very easy:

```
pip install pre-commit # to download/install the tool itself
cd $MAYA_USD_REPO_ROOT
pre-commit install   # adds the pre-commit hook to this repo
```

One nice thing about [pre-commit](https://pre-commit.com/) is that if you want the hook, its' very easy to install, but if you don't, you can simply ignore it - ie, all hooks are just installed per local repo.


kxl-adsk:
One thing we should consider reviewing here are the locations of files. We may be at the point where it would be good to group some of the git related files in a folder (assuming there are no constraints related to integrations). Additionally, the `test` folder holds everything that is related to testing the code base.

pmolodo:
Yeah, I wasn't really crazy about locating in `test/bin` either - but I also wasn't sure I wanted to make a new top-level directory just for this one thing, and I wasn't sure any other location was better.  Happy to place anywhere else, though...

kxl-adsk:
@pmolodo Sorry for the long delay with this one. The silly thing is that we only need a decision on where to put these files because I'm not aware of any reason for not accepting this contribution.

I agree with your comment about yet another top-level directory. Does anyone happen to know of a reason why we couldn't put both files in the `.github` folder?

pmolodo:
I believe `.pre-commit-config.yaml` needs to be in the root of the project, as that's where the `pre-commit` tool looks for it.

.github makes sense for the utility script though...

kxl-adsk:
I see. OK, then let's just move the utility script and be done here.

pmolodo:
So done!

kxl-adsk:
Thank you, @pmolodo. I will do some testing on Windows and macOS to make sure the script works on all platforms (I'm assuming you tested Linux).

.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
+repos:
+-   repo: local
+    hooks:
+
+    - id: clang-format
+      name: clang-format
+      description: Format files with clang-format
+      entry: test/bin/run-clang-format.py
```
kxl-adsk, line 8: You may want to update this path

pmolodo: Doh, right, forgot about that...

kxl-adsk:
@pmolodo I have not debugged it, but the first attempt to run the clang-format tool results in
```
user@BOSPC1CKS7L maya-usd_pr % ./.github/run-clang-format.py 
  File "./.github/run-clang-format.py", line 203, in main
    commit=args.commit)
  File "./.github/run-clang-format.py", line 95, in run_clang_format
    include_regex = regex_from_file(include_path)
  File "./.github/run-clang-format.py", line 57, in regex_from_file
    with io.open(path, 'r') as f:
IOError: [Errno 2] No such file or directory: '/Users/user/Dev/usd/.clang-format-include'
```

.github/run-clang-format.py
```diff
@@ -0,0 +1,214 @@
...
+import re
+import stat
+import subprocess
+import sys
+import time
+
+
+THIS_FILE = os.path.normpath(os.path.abspath(inspect.getsourcefile(lambda: None)))
+THIS_DIR = os.path.dirname(THIS_FILE)
+# THIS_DIR = REPO_ROOT/tests/bin
```
kxl-adsk, line 24: You will need to adjust this to the new location of the file

kxl-adsk (changes requested):

.github/run-clang-format.py
pmolodo: Done!

pmolodo:
Fixed the pathing issue pointed out, and also noticed that it wasn't working with python-3, so added a fix for that as well

kxl-adsk (approved):

PR #1121, opened 2021-01-27
MAYA-104394 - As a user, I'd like to be warned when saving a layer th…

spinell-adsk:
…at's already on disk

Add a confirm dialog box when saving a layer already on disk.

### DIFF
--- a/lib/usd/ui/layerEditor/layerTreeItem.cpp
+++ b/lib/usd/ui/layerEditor/layerTreeItem.cpp
@@ -248,10 +248,10 @@ void LayerTreeItem::saveEdits()
 
         QString okButtonText = StringResources::getAsQString(StringResources::kSave);
         shouldSaveEdits = confirmDialog(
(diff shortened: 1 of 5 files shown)


luceric-adsk (approved):

spinell-adsk (commented):
At first I added the confirmation box in LayerTreeItem::saveEdits() but it's cause a double confirmation when saving a stage with dirty layers. So I moved the confirmation box in
MayaLayerEditorWindow::saveEdits() to avoid that.

spinell-adsk:
After talking with @luceric-adsk we conclude that it's will be better to keep the confirmation box in  LayerTreeItem as I did on the first try. To avoid the double confirmation issue when saving a stage, I added a new function LayerTreeItem::saveEditsNoPrompt().

luceric-adsk (approved):

seando-adsk (approved):

fowlertADSK (approved):
LGTM

PR #1124, opened 2021-01-28
T bailp/maya 105624/fix bad xform no anim data

pierrebai-adsk:
There are two private functions that reads transforms: _pushUSDXformOpToMayaXform() and _pushUSDXformToMayaXform().

The first was using the earliest-time time code when no time interval were used for importing. The other was using default-time.

Using default-time prevented importing the transform since their time code did not match and the imported mesh would have bad transform (they would all have identity matrix).

This change make both code paths use earliest-time. This fixes bad import of geometry when not importing animation data in Maya.


lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -378,14 +378,16 @@ static bool _pushUSDXformToMayaXform(
     const UsdMayaPrimReaderArgs&    args,
     const UsdMayaPrimReaderContext* context)
 {
+    MTime::Unit timeUnit = MTime::uiUnit();
+    double timeSampleMultiplier = (context != nullptr) ? context->GetTimeSampleMultiplier() : 1.0;
+
```
pierrebai-adsk, line 383: Re-ordered the lines to be the same as in the other function (_pushUSDXformOpToMayaXform).

lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -397,9 +399,9 @@ static bool _pushUSDXformToMayaXform(
 
         timeArray.setLength(timeCodes.size());
     } else {
-        // If there were no time samples, we'll just use the default time and
+        // If there were no time samples, pick the first available sample or default and
         // leave the MTimeArray empty.
-        timeCodes.push_back(UsdTimeCode::Default());
+        timeCodes.push_back(UsdTimeCode::EarliestTime());
```
pierrebai-adsk, line 404: This is the actual fix. _pushUSDXformOpToMayaXform() uses EarliestTime. (As do other translators, like for meshes.)

kxl-adsk (approved):

kxl-adsk:
@pierrebai-adsk I believe you branched out of master (based on the commit history). Do you mind rebasing this on dev to have a clean commit in the review?

PR #1127, opened 2021-01-28
MAYA105624 make xform reading more uniform

pierrebai-adsk:
There are two private functions that reads transforms: _pushUSDXformOpToMayaXform() and _pushUSDXformToMayaXform().

The first was using the earliest-time time code when no time interval were used for importing. The other was using default-time.

Using default-time prevented importing the transform since their time code did not match and the iimportaed mesh would have bad transform (they would all have identity matrix).

This change make both code paths use earliest-time. This fixes bad import of geometry when not importing animation data in Maya.


kxl-adsk (approved):

kxl-adsk:
@pierrebai-adsk Would you mind adding a unit test to this PR?

mattyjams (approved):
Looks good to me too, but yeah, a small unit test would be great.

Thanks!

PR #1128, opened 2021-01-28
MAYA-107607 - As a user, I'd like to not save a layer when a sublayer…

spinell-adsk:
… is anonymous

- Avoid saving a layer that contains sub anonymous layer (Add warning dialog box)


seando-adsk (approved):

kxl-adsk:
The single failure in `2020 Windows python2` is to ignore. We still continue to have issues with PF on windows. As soon as the review is completed, this should be ready to go.

luceric-adsk (approved):

PR #1130, opened 2021-01-28
Fixes related to export of skeleton rest-xforms

pmolodo:
Some bugfixes for edge cases, and made it so that we export restXforms whenever we export bindXforms (ie, the restXforms are calculated from the bindXforms if present).

We were getting some errors in other tools if rest xforms weren't present (as well as an annoying warning/error from usdview, IIRC).

### DIFF
--- a/lib/usd/translators/jointWriter.cpp
+++ b/lib/usd/translators/jointWriter.cpp
@@ -259,6 +259,18 @@ bool _GetLocalTransformForDagPoseMember(
     MStatus status;
 
     MPlug xformMatrixPlug = dagPoseDep.findPlug("xformMatrix");
+    if (TfDebug::IsEnabled(PXRUSDMAYA_TRANSLATORS)) {
+        // As an extra debug sanity check, make sure that the logicalIndex
+        // already exists
+        MIntArray allIndices;
+        xformMatrixPlug.getExistingArrayAttributeIndices(allIndices);
+        if (std::find(allIndices.cbegin(), allIndices.cend(), logicalIndex) == allIndices.cend()) {
+            TfDebug::Helper().Msg(
+                "Warning - attempting to retrieve %s[%u], but that index did not exist yet",
+                xformMatrixPlug.name().asChar(),
+                logicalIndex);
+        }
+    }
     MPlug xformPlug = xformMatrixPlug.elementByLogicalIndex(logicalIndex, &status);
     CHECK_MSTATUS_AND_RETURN(status, false);
 
--- a/lib/usd/translators/jointWriter.cpp
+++ b/lib/usd/translators/jointWriter.cpp
@@ -241,10 +241,9 @@ static bool _FindDagPoseMembers(
     for (size_t i = 0; i < visitedIndices.size(); ++i) {
         uint8_t visited = visitedIndices[i];
         if (visited != 1) {
-            unsigned int index = (*indices)[i];
             TF_WARN(
                 "Node '%s' is not a member of dagPose '%s'.",
-                MFnDependencyNode(dagPaths[index].node()).name().asChar(),
+                MFnDependencyNode(dagPaths[i].node()).name().asChar(),
                 dagPoseDep.name().asChar());
             return false;
         }
--- a/lib/usd/translators/jointWriter.cpp
+++ b/lib/usd/translators/jointWriter.cpp
@@ -136,7 +136,6 @@ static GfMatrix4d _GetJointWorldBindTransform(const MDagPath& dagPath)
         CHECK_MSTATUS_AND_RETURN(status, GfMatrix4d(1));
         MPlug plgWorldMatrices = fnNode.findPlug("worldMatrix", false, &status);
         CHECK_MSTATUS_AND_RETURN(status, GfMatrix4d(1));
-        TF_VERIFY(membersIdx < plgWorldMatrices.numElements());
         MPlug         plgWorldMatrix = plgWorldMatrices.elementByLogicalIndex(membersIdx);
         MObject       plgWorldMatrixData = plgWorldMatrix.asMObject();
         MFnMatrixData fnMatrixData(plgWorldMatrixData, &status);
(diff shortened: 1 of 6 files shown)


pmolodo:
@dgovil recently touched some of this code, if he wants to take a look over...

dgovil:
Also CC'ing @ysiewappl since he touched it too. (we treat my personal fork as a submission staging area)

lib/usd/translators/jointWriter.cpp
```diff
@@ -222,16 +221,16 @@ static bool _FindDagPoseMembers(
     indices->resize(numDagPaths);
 
     std::vector<uint8_t> visitedIndices(numDagPaths, 0);
-    for (unsigned int i = 0; i < membersPlug.numElements(); ++i) {
+    for (unsigned int i = 0; i < membersPlug.numConnectedElements(); ++i) {
```
ysiewappl, line 225: I don't mind so much because I think this makes sense (this connection needs to exist anyway to be meaningful so the change will work), but what catches my eye is the reasoning; your comment seems suspicious that somehow `connectedTo` created a new element. Might be worth trying using `evaluateNumElements` to see if this is somehow an evaluation caching issue and if so worth reporting as a separate Maya bug.

ysiewappl (approved):
I think this makes sense to fallback on the bind transforms, and might solve problems we're seeing here as well.

lib/usd/translators/jointWriter.cpp
```diff
@@ -254,25 +252,23 @@ static bool _FindDagPoseMembers(
...
-    unsigned int             index,
+    unsigned int             logicalIndex,
     GfMatrix4d*              xform)
 {
     MStatus status;
 
     MPlug xformMatrixPlug = dagPoseDep.findPlug("xformMatrix");
-    if (index < xformMatrixPlug.numElements()) {
-        MPlug xformPlug = xformMatrixPlug[index];
+    MPlug xformPlug = xformMatrixPlug.elementByLogicalIndex(logicalIndex, &status);
```
ysiewappl, line 262: Right now I know there's only one call site and the only indices stuffed in here should the logical ones, but _just in case_, might be worth having debug assert to check if the logical index actually exists already via `getExistingArrayAttributeIndices` or similar.

lib/usd/translators/jointWriter.cpp
pmolodo: Yeah, the thing where it was somehow creating an element was odd.  Unfortunately, I didn't have time to fully investigate at the time, and I don't know if I'd be able to replicate now.  I guess I also didn't feel too motivated too, since I feel like using `.numConnectedElements()` is conceptually better anyway.
If you want, I can edit the commit message, and remove the bit about the crash / element being created, or throw a bunch more caveats around it.

ysiewappl: Not that big of a deal; if it's not something that you cannot reliably reproduce already, this code is fine as is (IMHO).

lib/usd/translators/jointWriter.cpp
pmolodo: My understanding is that if somehow that logical index doesn't exist yet, it will be created... which, as a fallback, seems ok to me.
Though I guess it might make sense to just put in a debug-only check / warning message, since it WOULD be an odd/unexpected occurence, which might warrent further investigation...

lib/usd/translators/jointWriter.cpp
pmolodo: To be clear - does the thumbs up mean you're ok with this as is? Or that you'd like me to edit the commit message?

pmolodo:
@kxl-adsk - are PF builds a little wonky again? I checked one of the failed builds, and the tests all passed - just some sort of permission error running a groovy script afterwards?

lib/usd/translators/jointWriter.cpp
pmolodo: Ok, added a check...

ppt-adsk:
> 
> 
> @kxl-adsk - are PF builds a little wonky again? I checked one of the failed builds, and the tests all passed - just some sort of permission error running a groovy script afterwards?

Yes, I got the same and reported it internally.  Stay tuned.

lib/usd/translators/jointWriter.cpp
```diff
@@ -136,7 +137,6 @@ static GfMatrix4d _GetJointWorldBindTransform(const MDagPath& dagPath)
         CHECK_MSTATUS_AND_RETURN(status, GfMatrix4d(1));
         MPlug plgWorldMatrices = fnNode.findPlug("worldMatrix", false, &status);
         CHECK_MSTATUS_AND_RETURN(status, GfMatrix4d(1));
-        TF_VERIFY(membersIdx < plgWorldMatrices.numElements());
```
kxl-adsk, line 139: Can you comment on why this check wasn't good?

lib/usd/translators/jointWriter.cpp
```diff
@@ -286,11 +295,12 @@ static bool _GetJointLocalRestTransformsFromDagPose(
     // Use whatever bindPose the root joint is a member of.
     MObject bindPose = _FindBindPose(rootJoint);
     if (bindPose.isNull()) {
-        TF_WARN(
-            "%s -- Could not find a dagPose node holding a bind pose: "
-            "The Skeleton's 'restTransforms' property will not be "
-            "authored.",
-            skelPath.GetText());
+        TF_DEBUG(PXRUSDMAYA_TRANSLATORS)
```
kxl-adsk, line 298: Interested to know why we want to switch this warning to a debug trace? (and same for the one below)

test/lib/usd/translators/testUsdExportSkeleton.py
```diff
@@ -205,6 +207,49 @@ def testSkelWithoutBindPose(self):
             animSource.GetTranslationsAttr().Get(5.0),
             Vt.Vec3fArray([Gf.Vec3f(5.0, 5.0, 0.0)]))
 
+    def testSkelRestXformsWithNoDagPose(self):
+        """
+        Tests export of of rest xforms when there is no dagPose node at all.
```
kxl-adsk, line 212: typo: `of of`

lib/usd/translators/jointWriter.cpp
pmolodo: It's because `membersIdx` is a logical index, while `numElements()` gives the number of physical plugs... so the two aren't related.

lib/usd/translators/jointWriter.cpp
pmolodo: Because we now have an additional fallback - we set the `restTransforms` from the bindPose - whereas before this would mean that we would "give up", and not author any `restTransforms`.

I don't mind changing either this or the next one back to `TF_WARN` if you want though.

Actually - on looking again, I think I'll definitely change the next one back to `TF_WARN`, because the inability to read the local transform shouldn't happen.  (Whereas with this one, it merely indicates that there was no dagPose... which, while not standard, is still kosher.)

test/lib/usd/translators/testUsdExportSkeleton.py
pmolodo: Fixed in https://github.com/Autodesk/maya-usd/pull/1130/commits/afaea5505a8aadef0d02bff165e72cc36e5c8a36

lib/usd/translators/jointWriter.cpp
pmolodo: Changed the second to `TF_WARN`, left this one as `TF_DEBUG` for now - let me know if you want this one back to `TF_WARN` as well:
https://github.com/Autodesk/maya-usd/pull/1130/commits/afaea5505a8aadef0d02bff165e72cc36e5c8a36

kxl-adsk (approved):

PR #1132, opened 2021-01-29
support for UV mirroring on UsdUVTexture

pmolodo:
...when exporting / importing.

Also does a bit of code unification for U+V cases

### DIFF
--- a/lib/usd/translators/shading/usdUVTextureReader.cpp
+++ b/lib/usd/translators/shading/usdUVTextureReader.cpp
@@ -106,6 +106,7 @@ TF_DEFINE_PRIVATE_TOKENS(
 
     // Values for wrapS and wrapT
     (black)
+    (mirror)
     (repeat)
 
     // UsdUVTexture Output Names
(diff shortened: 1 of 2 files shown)


JGamache-autodesk (approved):

JGamache-autodesk:
The Windows build has failed:
maya-usd\lib\usd\translators\shading\usdFileTextureWriter.cpp(445): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax
maya-usd\lib\usd\translators\shading\usdFileTextureWriter.cpp(445): fatal error C1001: An internal error has occurred in the compiler.
It was nice to see this elegant initializer syntax in the loops, but it seems VS does not understand it.

lib/usd/translators/shading/usdUVTextureReader.cpp
```diff
@@ -311,25 +312,36 @@ bool PxrMayaUsdUVTexture_Reader::Read(UsdMayaPrimReaderContext* context)
...
-    // Wrap V
-    usdInput = shaderSchema.GetInput(_tokens->wrapT);
-    mayaAttr = uvDepFn.findPlug(_tokens->wrapV.GetText(), true, &status);
-    if (usdInput && status == MS::kSuccess) {
-        if (usdInput.Get(&val) && val.IsHolding<TfToken>()) {
-            TfToken wrapT = val.UncheckedGet<TfToken>();
-            val = wrapT == _tokens->repeat;
-            UsdMayaReadUtil::SetMayaAttr(mayaAttr, val);
+    // Wrap U/V
+    for (auto wrapMirrorTriple :
```
kxl-adsk, line 316: We are still having trouble with windows nodes, but in this case PF actually found something real.

```S:\jenkins\workspace\ecg-mayausd-branch-preflight-2020-windows\ecg-maya-usd\maya-usd\lib\usd\translators\shading\usdUVTextureReader.cpp(316): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax```

**Review by kxl-adsk — CHANGES_REQUESTED** [review]
I commented in one place where PF caught error - please check logs if you need more details.

BTW. We still have issues with some windows nodes, so apologies for the inconvenience.

File: lib/usd/translators/shading/usdUVTextureReader.cpp (hunk 1 above)
Comment (pmolodo) [reply lib/usd/translators/shading/usdUVTextureReader.cpp]: Hmm, sorry about that. I guess the initializer syntax I used was not strict c++11, but a gcc extension:

```
(const TfToken[2][3]){ ... }
```

I just ended up declaring a variable to hold the array.  Alternatively, I could apparently have declared a type alias, since it seems that

```
T{ ... }
```

is valid, but only if `T` is a "simple-type-specifier".  So I could have done:

```
using Token23 = const TfToken[2][3];
for (auto wrapMirrorTriple : Token23{ ... }) {
```

...but I decided at that point it was just simpler to make the variable...

(Thanks to this post for the info about simple-type-specifiers: https://stackoverflow.com/a/24390563/920545)

kxl-adsk (approved):

PR #1133, opened 2021-01-29
MAYA-109519 - Maya USD Export menu displays animation option after co…

spinell-adsk:
…llapsing and expanding

- Fix animation data section visibility when animation frame layout is expanded.


fowlertADSK (approved):
LGTM
Glad you fixed this one, it was confusing when we saw it pop up in a demo the other day.

PR #1134, opened 2021-01-29
Use data model undo framework for attribute undo.

ppt-adsk:
UsdAttribute derived classes were not overriding the Ufe::Attribute::setCmd() virtual.  This meant that we were using the implementation from UFE, rather than our new, data model correct undo system.  This was bad in two ways:
- The UFE implementation only does value undo, not data model undo
- It uses the Ufe::Attribute to set the previous value, and in our implementation this Ufe::Attribute stores a UsdPrim that can become stale, as per the Autodesk MAYA-109522 internal bug.
Now fixed, with an automated test to validate the fix.

### DIFF
--- a/lib/mayaUsd/ufe/UsdAttribute.cpp
+++ b/lib/mayaUsd/ufe/UsdAttribute.cpp
@@ -15,6 +15,9 @@
 //
 #include "UsdAttribute.h"
 
+#include <mayaUsd/undo/UsdUndoBlock.h>
+#include <mayaUsd/undo/UsdUndoableItem.h>
+
 #include <pxr/base/tf/token.h>
 #include <pxr/base/vt/value.h>
 #include <pxr/usd/sdf/attributeSpec.h>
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -140,6 +143,31 @@ void setUsdAttributeVectorFromUfe(
     UFE_ASSERT_MSG(b, kErrorMsgFailedSet);
 }
 
+template <typename T, typename A = MayaUsd::ufe::TypedUsdAttribute<T>>
+class SetUndoableCommand : public Ufe::UndoableCommand
```
ppt-adsk, line 147: Simple class that uses the new UsdUndoableItem support.

test/lib/ufe/testAttribute.py
```diff
@@ -557,3 +557,50 @@ def testObservation(self):
         self.assertEqual(ball34Obs.notifications, 3)
         self.assertEqual(ball35Obs.notifications, 3)
         self.assertEqual(globalObs.notifications, 7)
+
+    # Run last to avoid file new disturbing other tests.
+    def testZAttrChangeRedoAfterPrimCreateRedo(self):
```
ppt-adsk, line 562: Test-Driven Development: wrote test first, it failed (crashed), coded fix, test now passes.

kxl-adsk (approved):

ppt-adsk:
All Windows Python 2 builds failed because of lack of disk space errors, but other builds succeeded.  I therefore believe this pull request is good to go.

PR #1136, opened 2021-01-29
[usdMaya] Exporting certain color sets as motion vectors.

pmolodo:
This just hardcodes certain color-set names to export as motion vectors... not sure if we need/want an option to control this.

### DIFF
--- a/lib/usd/translators/meshWriter.cpp
+++ b/lib/usd/translators/meshWriter.cpp
@@ -51,6 +51,7 @@
 #include <maya/MGlobal.h>
 #include <maya/MIntArray.h>
 #include <maya/MItDependencyGraph.h>
+#include <maya/MItMeshVertex.h>
 #include <maya/MPlug.h>
 #include <maya/MPlugArray.h>
 #include <maya/MStatus.h>
@@ -274,6 +275,9 @@ TF_DEFINE_PRIVATE_TOKENS(
     ((skelGeomBindTransform, "skel:geomBindTransform"))
 );
 // clang-format on
+
+const std::vector<std::string> PxrUsdTranslators_MeshWriter::motionVectorNames
+    = { "velocityPV", "velocity", "v" };
 
 PxrUsdTranslators_MeshWriter::PxrUsdTranslators_MeshWriter(
     const MFnDependencyNode& depNodeFn,
@@ -646,6 +650,11 @@ bool PxrUsdTranslators_MeshWriter::writeMeshAttrs(
         if (_excludeColorSets.count(colorSetName) > 0)
             continue;
 
+        if (std::find(motionVectorNames.begin(), motionVectorNames.end(), colorSetName)
+            != motionVectorNames.end()) {
+            writeMotionVectors(primSchema, usdTime, finalMesh, colorSetName);
+            continue;
+        }
         bool isDisplayColor = false;
 
         if (colorSetName == UsdMayaMeshPrimvarTokens->DisplayColorColorSetName.GetString()) {
@@ -782,6 +791,53 @@ bool PxrUsdTranslators_MeshWriter::writeMeshAttrs(
     }
 
     return true;
+}
+
+void PxrUsdTranslators_MeshWriter::writeMotionVectors(
+    UsdGeomMesh&       primSchema,
+    const UsdTimeCode& usdTime,
+    MFnMesh&           mesh,
+    const std::string& colorSetName)
+{
+    const MString    colorSetNameMS(colorSetName.c_str());
+    VtArray<GfVec3f> motionVectors;
+    motionVectors.resize(static_cast<size_t>(mesh.numVertices()));
+
+    // We need to average the color out here, because we need per vertex velocity.
+    MObject   meshObject = mesh.object();
+    MIntArray faces;
+
+    for (MItMeshVertex itVertex(meshObject); !itVertex.isDone(); itVertex.next()) {
+        faces.clear();
+        itVertex.getConnectedFaces(faces);
+
+        const auto flen = faces.length();
+        if (flen > 0) {
+            const auto scale = 1.0f / static_cast<float>(flen);
+            const auto id = itVertex.index();
+
+            auto& v = motionVectors[id];
+            v[0] = 0.0f;
+            v[1] = 0.0f;
+            v[2] = 0.0f;
+
+            for (auto f = decltype(flen) { 0 }; f < flen; ++f) {
+                MColor col;
+                if (itVertex.getColor(col, faces[f], &colorSetNameMS)) {
+                    v[0] += col[0];
+                    v[1] += col[1];
+                    v[2] += col[2];
+                }
+            }
+
+            v[0] *= scale;
+            v[1] *= scale;
+            v[2] *= scale;
+        }
+    }
+
+    UsdMayaWriteUtil::SetAttribute(
+        primSchema.GetVelocitiesAttr(), motionVectors, usdTime, _GetSparseValueWriter());
 }
 
 bool PxrUsdTranslators_MeshWriter::ExportsGprims() const { return true; }
--- a/lib/usd/translators/meshWriter.h
+++ b/lib/usd/translators/meshWriter.h
@@ -62,6 +62,12 @@ public:
 private:
     bool writeMeshAttrs(const UsdTimeCode& usdTime, UsdGeomMesh& primSchema);
 
+    void writeMotionVectors(
+        UsdGeomMesh&       primSchema,
+        const UsdTimeCode& usdTime,
+        MFnMesh&           mesh,
+        const std::string& colorSetName);
+
     /// Cleans up any extra data authored by SetPrimvar().
     void cleanupPrimvars();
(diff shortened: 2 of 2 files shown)


kxl-adsk:
I would like to better understand how generic and useful this is. As well, what would prevent you from adding this support with an export chaser plugin?

pmolodo:
As is, probably not too generic, since the colorSets that are interpreted as vertex-velocities is hard coded: 

```
"velocityPV", "velocity", "v"
```

It could be made more generic if we added an arg to control velocity vector export - say,

```
mayaUsdExport(meshVelocityColorSet="colorSetName", ...)
```

...which would both enable authoring of the velocity attr, and provide the color set to pull the data from.  As for useful - we certainly found it useful to have vertex velocity information (I think mainly for various fx passes, IIRC), but I suppose I can't speak directly to how many in the industry would want this.  We could also expand the usability by also allowing for export of accelerations via a similar mechanism:

```
mayaUsdExport(meshAccelerationColorSet="colorSetName", ...)
```

...in which case you could use it to generate motion blur for geo with varying topology.

As for the export chaser - this would be doable, but it's certainly more convenient to have a flag.  At minimum, you'd also have to remember to delete the generic primvar that got created, or else disable colorSet export altogether.

Perhaps we can take take a quick straw poll at one of the next meetups to see if others would find this useful?

kxl-adsk:
I think it would be worth having a stand on how we `protect` export flags from going out of control. I understand it's convenient, but I would prefer us to find a way to separate core features from studio-specific extensions. This would help as well with similar proposal #1081 for extending core flags driven by the convenience of use.

FYI. @dgovil

dgovil:
@kxl-adsk I think it's fair to be concerned about explosion of export flags and I definitely share your concern in general.

I would posit that the layer data one will not be studio specific since I've had to share post processing scripts with multiple partners to add those in to their files after export, which gets tricky when it comes to usdz files (well if they didn't have usdzip on their system), and it's something I'd have used in studio use personally.

Chasers are great but a little limited to post processing, plus require rebuilding every time a new plugin is released in our experience. It also makes it harder to share compiled chasers with other teams/studios/partners/tds who may not have the technical savvy to compile C++ code, whereas when it's in the mayaUsd plugin, they can just use it straight out of Maya easily.

So with those things in mind here were some thoughts:

* Yes I think it's wise to be frugal about the export flags. I'm perhaps biased and would still advocate for the custom layer data ones to go in, but I can also support arguments against general flag bloat.
* Perhaps Python based chasers can help alleviate the technical burden of maintaining chasers for third parties. It's something we're looking into after we prepare our next round of contributions. I certainly think they'd help the average pipeline TD in a production scenario.
* Maybe C++/Python chasers could be contributed to this repo in general and built along side the main plugin as a secondary tier of functionality? That would at least reduce flag bloat, though wouldn't curtail code/feature bloat which would also be a legit concern.
* With regards to this specific PR, I wonder if it makes sense to do a slightly different approach and have a flag like `colorSetPurpose` which could be used like `colorSetPurpose=[("velocity", "velocity"), ("foo", "displayColor")]` where the first value is the name of the maya color set, and the secondary value is the name of the purpose of the color set. So if it's velocity, it gets used as motion vectors etc...

kxl-adsk:
Python bindings for translation framework (including chasers) is the option I was hoping can help with the entry cost for translation framework customization.

In the meantime, what I just learned about are MEL and Python callbacks for post export notifications - https://github.com/Autodesk/maya-usd/blob/e7b56854a44af7173ead4c125956a50cae4d8c8f/lib/mayaUsd/fileio/jobs/writeJob.cpp#L760. I never noticed them before (and not sure how much use they get, since MEL ones were broken), but given that we know the destination file name, there should be a way from the callback to retrieve the layer that you export to and write any additional data you need.

Assuming the above works:
@dgovil would this work for your metadata case?
@pmolodo same for motion vectors, you should be able to tweak the stage just before save

dgovil:
Yep, I think we could provide metadata integration as a callback (didn't
know those were there) till the Python bindings come online. I think the
future Python bindings for chasers would be optimal since it looks like
callbacks don't have space for passing in arguments, so would need some
level of trickery to customize per export. That metadata PR was more to
serve as supporting material for the discussion on custom data in general
in the USD, using our internal changes as a reference, but I can close it
out since the PR can still be accessible when closed.

pmolodo:
In general, we've found the MEL / python postCallbacks not very useful, because they don't have access to the stage.  It's true that you can write directly to the layer, but we generally avoid layer edits unless there's no other option, because they're more cumbersome and fragile - ie, there's no composition, you can't use methods on schemas or things like the primvar API, etc.  Metadata is simple enough that I think it does work well, but for anything more complex, it's not the best solution.  I've noticed that on usd-interest, Spiff often tries to try to steer people away from direct SdfSpec editing when possible, as well.

I actually really like @dgovil's earlier suggestion: making this PR more broadly useful, and providing a way to specify uses for color sets in general.  In addition to specifying the display color, velocity, and accelerations color sets to use, it could also specify colorSets to skip, by giving them use as 'none'  (or 'skip' / the empty string / etc).  Plus it provides some room for future customizability, so it provides a certain measure of protection against bloat in the future.

dgovil:
Ah in the case that they don't receive the stage, then you're right. they don't serve as good alternatives to a chaser or direct integration. It would be cool if they could receive the stage pointer.

kxl-adsk:
I'm still exploring options to better understand what falls into which category. So using this PR only as an example, you could still create a stage to edit the root layer with USD level interfaces from python callback?

pmolodo:
You know... I feel silly for not thinking of that.  I'm so used to thinking of stages conceptually as something that deals with the arcane wizardry of composition, and so two stages with the same root layer are not equivalent... but in this case, where we can assume that the root layer is flat, and there aren't many composition arcs to worry about, creating a new stage would probably work.

It still makes me a bit itchy - for instance, one composition arc the exporter DOES support is variant sets, and I'd need to think about that a bit to make sure that wouldn't create wrinkles (other than making iterating over all the prims more complicated)... and there's the possibility that we start supporting more advanced compositions in the future, and I have a vague worry that there might be some options on stage creation, or some sort of stage state I'm not thinking of that could affect things... but I think you're right, and that we probably could get something along those lines working.

For this PR specifically, though - I still feel like, with the colorSet-purpose-generalization, there's enough general utility that it would still be worth including, since it would:

- increase the utility of already supported features:
   - avoid hard-coding a colorSet to use for the displayColor
   - allow black-listing of colorSets NOT to export
   - ...or white-listing, if we allow, say, '*' as a colorSet-name-placeholder
- add the ability to export velocities and accelerations
- give us room for future customization, to avoid some requests for more special-case flags

Perhaps as a compromise we can bring this up at the next Charter meeting, and see if others would find this useful?  If the general consensus is it wouldn't be useful, I'll drop the request.

Another alternative - perhaps another level of chaser might be useful - ie, we can already use chasers to get callbacks after default values, after each frame, and after the entire stage as a whole is done.... but what if we had a way to get callbacks per-prim (or per primWriter)? This would ensure that we were writing in the same context as maya's normal writers.  It could be done during the same loop that the `primWriter->PostExport()` calls are made... and perhaps we could even have a way to register callbacks just for specific types of primWriters.  (The primWriter-type registration might be important if we're allowing python chasers of this sort, to avoid doing the "do I care about this node?" checks in python...)

If we had something like that, I think it would create a LOT of extra flexibility... it would provide an easy way to modify prims, in the exact same context as maya's "normal" export writers, and without having to do our own extra iteration over all prims again.    If I was fairly sure something like that was coming eventually (and ideally with python support), I'd happily close this PR in anticipation of that.

kxl-adsk:
@pmolodo 
re `colorSet-purpose-generalization` - I floated the idea internally and this seems to be resonating with people. We will come back to you once we have a bit more time to validate the different use cases.

re per prim callbacks - we could move this to a discussion. Without thinking too much about it, I have two concerns:
1. Performance
2. Consistency of the data model, i.e. in the callback you can make a change that should affect more than a single prim, but translators running after may not be prepared to react to such changes.

kxl-adsk:
@pmolodo Spent some more time with folks internally and `colorSet-purpose-generalization` is the direction we would like to take. 

Will you work towards that goal in this PR or it's NOT something you were planning to work on? If it's the latter, we should log an enhancement issue to capture the outcome of this discussion and close this PR.

pmolodo:
> @pmolodo Spent some more time with folks internally and `colorSet-purpose-generalization` is the direction we would like to take.

Awesome!

> Will you work towards that goal in this PR or it's NOT something you were planning to work on? If it's the latter, we should log an enhancement issue to capture the outcome of this discussion and close this PR.

Well, it WAS something I was planning to work on... but I've recently gotten a new high-prio task.  I may have time to at least take a first pass at this, but probably won't have time to get it fully polished up (ie, working test-suite, etc).  If I don't have time in the next few days, though, I'll make the issue + close this PR.

pmolodo:
Ok, made this issue:

https://github.com/Autodesk/maya-usd/issues/1210

...and closing this PR...

PR #1137, opened 2021-01-30
[usdmaya] basic usdlux export - only pointLight / spotLight for now

pmolodo:
Putting this up as a PR because I figure some light export is better than none... and it provides a template that can be expanded for other light types.

Possibly just a jumping off or reference point for more fleshed-out usdLux support, if we don't want to provide support for just point + spot.

### DIFF
--- a/lib/usd/translators/lightWriter.cpp
+++ b/lib/usd/translators/lightWriter.cpp
@@ -13,10 +13,81 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+#include "lightWriter.h"
+
 #include <mayaUsd/fileio/primWriterRegistry.h>
 #include <mayaUsd/fileio/translators/translatorRfMLight.h>
+#include <mayaUsd/fileio/utils/adaptor.h>
+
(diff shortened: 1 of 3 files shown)


kxl-adsk:
Not sure if you have an answer to this question: Is there a set of generic utilities/classes we could create based on this PR and existing RenderMan light support? We would do this to help other renderers, like Arnold, when implementing renderer specific light translation.

HamedSabri-adsk:
I agree with Krystian's comment. My first impression when I saw this PR was that it would makes more sense if we have these changes as part of "lib/mayaUsd/fileio/translators" where others can leverage upon.

pmolodo:
The code already makes uses of some genericized templates to do some basic setup stuff, so I definitely think it wouldn't be hard to put the guts in a utility lib in `fileio/translators`.

However... we're not currently using this code in production, and I probably don't have time to clean this up properly... so I think maybe I'll just close this PR, and it can just be used as a potential jumping off point or reference if we want to add better UsdLux support in the future.

PR #1138, opened 2021-01-30
move minimum core usd version to v20.02

mattyjams:
Following up on #1101, these changes set core USD version 20.02 as the minimum supported version of core USD. This version is specified as the minimum in the `find_package()` call in the top-level `CMakeLists.txt` so that the build will quickly bail out on any attempt to use an older USD version.

All conditionally compiled code that was active for versions of USD prior to 20.02 was removed. The bulk of the removed code was in `hdMaya`/`mayaToHydra`.

Our environment has moved on such that I can't actually build core USD 20.02 anymore, but I was able to verify this against 20.05 on Linux using Maya PR 121. A few minor adjustments were needed though (see 1281cbb374daba6ab96c38e7bdd6cd574e5ef045).

I also took this opportunity to mark the current USD dev branch commit (https://github.com/PixarAnimationStudios/USD/commit/82dd40ad31bcdecfb18c2f23e39a3ddf68ef10d4) as supported in `doc/build.md` as well (c6625f1abf80dda0d4a27ad760e47af58676abfd).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,7 +99,7 @@ include(cmake/utils.cmake)
 include(cmake/jinja.cmake)
 
 find_package(Maya 2018 REQUIRED)
-find_package(USD REQUIRED)
+find_package(USD 0.20.02 REQUIRED)
 include(cmake/usd.cmake)
 include(${USD_CONFIG_FILE})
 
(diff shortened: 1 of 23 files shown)


kxl-adsk:
Thank you, @mattyjams for making this cleanup. I run checks and can confirm:
:white_check_mark:  build stops when using an older version than 20.02
:white_check_mark: compilation and regression tests pass on Windows for 20.02, 20.05, 20.08, 20.11 for various Maya releases.

I restarted PF, but there seem to have been a few issues that require attention from another team.

kxl-adsk (approved):

PR #1139, opened 2021-02-01
MAYA-105623 better support for USD leftHanded mesh.

pierrebai-adsk:
The  main problem is that the 'leftHanded' orientation is USD changes the interpretation of the vertices order of faces *and* the direction of generated normals.

On the other hand, the 'opposite' flag in Maya flips the faces directions, but does not flip the normals.

So the behaviour differs. To better suport USD import of externally-generated USD files, we will no longer set the 'opposite' flag when importing ledftHanded USD files. We will instead change the vertices orders to be right-handed.

Note that we will still export 'opposite' geometries from Maya as by setting the 'leftHanded' flag. This is done for two reasons: first, to keep the code change minimal, second because USD does not support different handedness for faces and normasl anyway.

### DIFF
--- a/test/lib/usd/translators/UsdImportMeshTest/Mesh.usda
+++ b/test/lib/usd/translators/UsdImportMeshTest/Mesh.usda
@@ -31,18 +31,69 @@ def Xform "World"
         int[] primvars:st:indices = [0, 1, 2, 3, 3, 2, 4, 5, 5, 4, 6, 7, 7, 6, 8, 9, 1, 10, 11, 2, 12, 0, 3, 13]
     }
 
+    def Mesh "LeftHandedSubdivMesh" (
+        kind = "component"
+    )
+    {
+        uniform token orientation = "leftHanded"
+        float3[] extent = [(0.5, 0.5, 0.5), (1.5, 1.5, 1.5)]
+        token faceVaryingLinearInterpolation = "all"
+        int[] faceVertexCounts = [4, 4, 4, 4, 4, 4]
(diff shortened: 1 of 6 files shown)


JGamache-autodesk (changes requested):
The code looks good. I am a bit unsure about algorithms that could potentially use adjacent triangles like OpenSubDiv, but if subdivided left and right handed cubes look good then things should be fine.

Can you add a unit test? There is a `\test\lib\usd\translators\testUsdImportMesh.py` that could be expanded to include mesh handedness tests.

pierrebai-adsk:
OK, I'll look at the test.

test/lib/usd/translators/UsdImportMeshTest/Mesh.usda
```diff
@@ -31,18 +31,69 @@ def Xform "World"
...
+            interpolation = "faceVarying"
+        )
+        int[] primvars:st:indices = [0, 1, 2, 3, 3, 2, 4, 5, 5, 4, 6, 7, 7, 6, 8, 9, 1, 10, 11, 2, 12, 0, 3, 13]
+    }
+
     def Mesh "PolyMesh" (
         kind = "component"
     )
     {
-        float3[] extent = [(-0.5, -0.5, -0.5), (0.5, 0.5, 0.5)]
```
pierrebai-adsk, line 38: I changed coordinates so that the meshes and subdiv don't overload, so it can more easily be manually viewed in Maya or others.

JGamache-autodesk (approved):

JGamache-autodesk (approved):

HamedSabri-adsk (approved):
Looking good. I like the consistence behavior of 'opposite' in relation to normals directions.

PR #1140, opened 2021-02-01
MAYA-106503 load all prims in hierarchy view.

pierrebai-adsk:


seando-adsk (approved):
I'm pretty sure that the LoadNone was intentional as initially we didn't want theses prims. But its fine to change it now if the specs changed.

pierrebai-adsk:
IDK if the specs, changed, just that a bug was filed about this issue... if the LoadNone was intentional, the other option would be to close the bug as Äs Designed" instead. I don't think I should be the one to decide on design...

seando-adsk:
Sorry I wasn't implying that you needed to make any decision. I was just pointing out that design team might have changed the spec. It was a long time ago I wrote this, so I don't recall. The change is fine with me.

PR #1141, opened 2021-02-02
Mute layer and variant switch preserve selection.

ppt-adsk:


luceric-adsk (approved):

lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include "layerEditorCommand.h"
 
+#include <mayaUsd/ufe/Utils.h>
```
seando-adsk, line 19: As we discussed yesterday all the changes in this file need to be UFE protected, with something like:
#if defined(WANT_UFE_BUILD)

seando-adsk (changes requested):

lib/mayaUsd/commands/layerEditorCommand.cpp
ppt-adsk: Yeah, I'm on it.

seando-adsk: As per our coding guidelines all conditional includes should come last (after the system includes).

seando-adsk (approved):

Issue #528, opened 2020-05-26
proxyShape does not hold a strong pointer to it's UsdStage through MayaUsdStageData


williamkrick:
**Describe the bug**
In `MayaUsdProxyShapeBase::computeInStageDataCached` we open or create a new `UsdStage`. The stage is saved into a MayaUsdStageData and stored in the node. `MayaUsdStageData` doesn't keep a strong reference to the stage (`UsdStageRefPtr`), instead it only holds `UsdStageWeakPtr` and relies on the stage cache keeping a strong reference to keep the `UsdStage` alive.

If something clears the `UsdStageCache` then all the proxyShapes will lose their stage data and break.

**Steps to reproduce**
1. In `MayaUsdProxyShapeBase::computeInStageDataCached` remove the calls which create `UsdStageCacheContext`. This disables the `UsdStageCache`.
2. Build and try to use the proxyShape. Nothing works.

**Expected behavior**
The proxyShape should not rely on the `UsdStageCache` to keep it's data alive.

**Additional context**
There's a big comment in `MayaUsdStageData` explaining the reasoning behind using a weak pointer.

> 
>         // The original Pixar code was the following:
>         //
>         // UsdStageRefPtr stage;
>         //
>         // Now using a weak pointer instead of a referencing pointer.  The AL
>         // plugin originally used a referencing pointer, but ran into problems.
>         // From a Rob Bateman @ AL e-mail, 7-Apr-2019:
>         // 
>         // "The reason for the weak pointer was that we found that Maya seemed
>         // to have a memory leak (as in, the MPxData derived objects were being
>         // created, but never deleted when we expected them to be - possibly by
>         // design?). We had originally used a direct mirror of the Pixar data
>         // object, but because the data objects were never freed (even after
>         // file new), the original stage would be retained. This meant the
>         // internal SdfLayerCache in USD would keep hold of the previously
>         // loaded layers. So we found we had this problem: 
>         // 
>         // 1. Import some.usda file into a proxy shape.
>         // 2. Make some modifications
>         // 3. File New
>         // 4. Import the same "some.usda" file into a proxy shape.
>         // 
>         // At this point, USD would essentially hand you back the stage
>         // composed of the modified layers, rather than a clean stage composed
>         // from the files on disk. Switching the type from a shared_ptr to a
>         // weak_ptr worked around this issue. Basically if we use a shared
>         // pointer here, the only way to reload a scene, would be to restart
>         // Maya."

[...]

Issue #164, opened 2019-12-06
refactoring_sandbox : VP2 Delegate : Prims in Duplicated ProxyShapes Move Together


cfmoore007:
**Issue :**

While testing PR#162, after I duplicated the proxyShape, I tried to move one of the sphere prims within one of the shapes.

When moving the prim in the one proxyShape, the prim with the same name is also moving in the other shape.  

![duplicateMovesSameTime](https://user-images.githubusercontent.com/2152250/70359378-c54c1b80-1849-11ea-8fd8-e3bde3da91d7.png)

This is not expected.

This occurs when there are session edits ( or not ).

**Steps to re-produce :
[3_Spheres.zip](https://github.com/Autodesk/maya-usd/files/3934157/3_Spheres.zip)
**

open maya ( UFE does not need to be set, just the envar for VP2_RENDER_DELEGATE_PROXY )
2) import an AL proxyShape :

import maya.cmds as cmds
cmds.loadPlugin ("AL_USDMayaPlugin")

```
# import proxyShape
cmds.AL_usdmaya_ProxyShapeImport(file="3_Spheres.usda")

# select and duplicate
cmds.select("AL_usdmaya_Proxy",r=True)
cmds.duplicate()

# select a prim and move it
```

santosg87:
I can reproduce this on our side as well. this might be related to the session layer when modifying the objects, as they both reference in the same usd file. 

I can also reproduce this behavior by creating 2 stages using the same usd file (instead of doing a duplicate)

cfmoore007:
I just tested this against Maya USD 0.6.0 w/ PR121 and the problem still exists.

bfloch:
This is not a viable solution, but we found as temporary workaround - if you duplicate special and set to duplicate the input graph it will duplicate the StageDataNode, which eliminates the single stage/layer problem.
Edits will be unique again.

kxl-adsk:
Linking with #1142 . 

@cfmoore007 and/or @bfloch - do you think you could give the above pull request a try?

cfmoore007:
@kxl-adsk I just tried it and still see the issue as described in the ticket.  Note : this time, I'm using the Maya ProxyShape.

cfmoore007:
After Kristian's note, when the edit target is not set to a file backed layer ( i.e : the session layer ), it works :

![duplicateFixed](https://user-images.githubusercontent.com/2152250/106791151-8d459180-6622-11eb-972a-a8a48d5d869f.png)

[...]

PR #1142, opened 2021-02-02
MAYA-108318: Add support for duplicating a proxyShape.

HamedSabri-adsk:
This PR provides the proper infrastructure to handle duplicating proxy shapes. 

Upon duplication, we determine if the Layer(s)/Sublayer(s) need to transfer content (deep copied) or shared. For instance, if we are duplicating a stage with an anonymous layer, the Layer(s)/Sublayer(s) shouldn't be shared and gets divorced from its source.
On the other hand, if the we are duplicating a file backed stage, the Layer(s)/Sublayer(s) remain shared.


test/lib/mayaUsd/nodes/testProxyShapeBase.py
```diff
@@ -44,6 +48,134 @@ def testBoundingBox(self):
...
+        self.assertEqual(duplStage.GetEditTarget().GetLayer(), duplStage.GetRootLayer())
+        self.assertEqual(True, "-session" in duplStage.GetSessionLayer().identifier)
+        self.assertEqual(True, "tree.usd" in duplStage.GetRootLayer().identifier)
+
+        # delete "/TreeBase"
+        cmds.delete('|Tree_usd1|Tree_usd1Shape,/TreeBase')
+
+        # confirm that edits are shared. Both source and duplicated proxyShapes have no children now.
+        self.assertEqual(0, len(ufe.Hierarchy.hierarchy(duplProxyShapeItem).children()))
+        self.assertEqual(0, len(ufe.Hierarchy.hierarchy(treebaseItem).children()))
```
HamedSabri-adsk, line 179: Added new tests cases to back up the logics in ProxyShapeBase.

lib/mayaUsd/nodes/stageData.h
```diff
@@ -75,39 +75,9 @@ class MayaUsdStageData : public MPxGeometryData
...
-    // from the files on disk. Switching the type from a shared_ptr to a
-    // weak_ptr worked around this issue. Basically if we use a shared
-    // pointer here, the only way to reload a scene, would be to restart
-    // Maya."
-    //
-    // Logged as https://github.com/Autodesk/maya-usd/issues/528
-
-    UsdStageWeakPtr stage;
-    SdfPath         primPath;
+    UsdStageRefPtr stage;
```
HamedSabri-adsk, line 78: Thanks to @kxl-adsk  for finding this out.

In the new code in copyInternalData(), we could be dealing with a file backed stage where we do not transfer the content and rather point to the same stage.  Having a weak pointer to the stage in the data block would be problematic.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -768,6 +819,99 @@ UsdTimeCode MayaUsdProxyShapeBase::GetOutputTime(MDataBlock dataBlock) const
...
+    if (isSessionLayer) {
+        newUsdStage->SetEditTarget(newUsdStage->GetSessionLayer());
+    } else if (!isSessionLayer && !isAnonymous) {
+        newUsdStage->SetEditTarget(srcCurrentTargetLayer);
+    }
+
+    // recursively create new anon Sublayer(s) for session and root layers
+    createNewAnonSubLayerRecursive(newUsdStage, srcCurrentTargetLayer, sessionLayer);
+    createNewAnonSubLayerRecursive(newUsdStage, srcCurrentTargetLayer, rootLayer);
+
```
HamedSabri-adsk, line 897: We could have nested Sublayer hierarchy under both "Session" and "Root" layers.

We recursively need to create a deep copy Sublayer(s) if the parent layer is AnonymousLayer.

cfmoore007:
As mentioned in the linked issue, I just tried this PR and still see the issue.

( Just to make sure I'm not doing something silly here .. )
git fetch origin refs/pull/1142/head:add_support_duplicate_proxyshape
git checkout dev
git merge add_support_duplicate_proxyshape --no-commit
( built that )

1. open maya
2. create stage from file ( using my 3_Spheres example )
3. duplicate proxy shape
4. move one of the spheres, it still moves the original as well as the sphere in the duplicated shape

I have also tried the above with editing one of the spheres prior to the duplication of the proxy shape.  Same results.

Thanks!

kxl-adsk:
@cfmoore007 By default we set as edit target the root layer when creating a proxy from a file. File backed layers have to remain shared, so if you continue modifying it, then you will see the same behavior. 

When the original ticket #164 was created, we have been setting the session layer as the edit target. If you do your steps again, but switch the edit target between steps 2 and 3 to the session layer, you should get the expected behavior.

cfmoore007:
Thanks for that detail.  Makes sense!

I tried again, and now it's working as expected :

duplicateFixed

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -134,6 +135,56 @@ MObject MayaUsdProxyShapeBase::outTimeAttr;
...
+    std::size_t found = identifier.find_last_of(":");
+    return identifier.substr(found + 1);
+}
+
+// recursive function to create new anonymous Sublayer(s)
+// and set the edit target accordingly.
+void createNewAnonSubLayerRecursive(
+    UsdStageRefPtr newUsdStage,
+    SdfLayerRefPtr targetLayer,
+    SdfLayerRefPtr parentLayer)
```
ppt-adsk, line 154: Minor: shouldn't we pass these pointers in as const-ref, to avoid creating the temporary argument pointers and therefore avoiding an inc / dec of the ref count?

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -768,6 +819,99 @@ UsdTimeCode MayaUsdProxyShapeBase::GetOutputTime(MDataBlock dataBlock) const
     return _GetTime(dataBlock);
 }
 
+void MayaUsdProxyShapeBase::copyInternalData(MPxNode* srcNode)
+{
+    MStatus retValue { MS::kSuccess };
+
+    // get the srource data block
```
ppt-adsk, line 826: "srource" --> "source".

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -768,6 +819,99 @@ UsdTimeCode MayaUsdProxyShapeBase::GetOutputTime(MDataBlock dataBlock) const
...
+    // ---------------------------------------------------------------------------------
+
+    // first, read the input value from "outStageDataAttr". outStageDataAttr gets computed when
+    // we get the stage on the proxy. If there is no incoming data, we return right away.
+    MDataHandle srcInDataCachedHandle = srcDataBlock.inputValue(outStageDataAttr, &retValue);
+    if (srcInDataCachedHandle.data().isNull()) {
+        return;
+    }
+
+    // query form the destination block to make sure inStageDataCachedAttr attribute is clean. If it
```
ppt-adsk, line 841: "form" --> "from".

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -768,6 +819,99 @@ UsdTimeCode MayaUsdProxyShapeBase::GetOutputTime(MDataBlock dataBlock) const
...
+    UsdStageRefPtr srcUsdStage = srcInData->stage;
+
+    // transfer session layer
+    // session layer is never shared so transfer its content always.
+    SdfLayerRefPtr sessionLayer
+        = SdfLayer::CreateAnonymous(kAnonymousLayerName + kSessionLayerPostfix + ".usda");
+    sessionLayer->TransferContent(srcUsdStage->GetSessionLayer());
+
+    // decide if the root layer needs to be shared or deep copied.
+    SdfLayerRefPtr rootLayer = SdfLayer::CreateAnonymous(kAnonymousLayerName);
```
ppt-adsk, line 870: Shouldn't we move the CreateAnonymous() call inside the conditional, just before line 872?

lib/mayaUsd/nodes/stageData.h
ppt-adsk: If this is no longer a problem, then we should re-test
https://github.com/Autodesk/maya-usd/issues/528
and close it, if it's no longer reproducible after your fix.

test/lib/mayaUsd/nodes/testProxyShapeBase.py
```diff
@@ -44,6 +48,134 @@ def testBoundingBox(self):
...
+        proxyShapeItem = ufe.Hierarchy.createItem(proxyShapePath)
+        proxyShapeContextOps = ufe.ContextOps.contextOps(proxyShapeItem)
+        proxyShapeContextOps.doOp(['Add New Prim', 'Capsule'])
+
+        # proxy shape is expected to have one child
+        proxyShapeHier = ufe.Hierarchy.hierarchy(proxyShapeItem)
+        self.assertEqual(1, len(proxyShapeHier.children()))
+
+        # child name is expected to be Capsule1
+        self.assertEqual(str(proxyShapeHier.children()[0].path().back()), "Capsule1")
```
ppt-adsk, line 70: Minor: can be just proxyShapeHier.children()[0].nodeName().  In a couple of other places as well.

test/lib/mayaUsd/nodes/testProxyShapeBase.py
```diff
@@ -44,6 +48,134 @@ def testBoundingBox(self):
...
+
+        # duplicated stage name is expected to be incremented correctly
+        self.assertEqual(str(duplStageItem.path().back()), "stage2")
+
+        # duplicated proxyShape name is expected to be incremented correctly
+        duplProxyShapeHier = ufe.Hierarchy.hierarchy(duplStageItem)
+        duplProxyShapeItem = duplProxyShapeHier.children()[0]
+        self.assertEqual(str(duplProxyShapeItem.path().back()), "stageShape2")
+
+        # duplicated ProxyShapeItem should have exactly one children
```
ppt-adsk, line 97: Nit-pick: "children" --> "child".

test/lib/mayaUsd/nodes/testProxyShapeBase.py
ppt-adsk: Thanks for the test!

ppt-adsk (commented):
Only some minor tweaks, but the core looks good to me!

lib/mayaUsd/nodes/proxyShapeBase.cpp
HamedSabri-adsk: Thanks @ppt-adsk. All addressed: ad23ef814853ca48d0f7b3be0970cde9231fe6bc

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -721,7 +721,7 @@ MStatus MayaUsdProxyShapeBase::computeOutStageData(MDataBlock& dataBlock)
 
     // Create the output outData
     MFnPluginData pluginDataFn;
-    MObject       stageDataObj = pluginDataFn.create(MayaUsdStageData::mayaTypeId, &retValue);
```
HamedSabri-adsk, line 673: stageDataObj  is created and not used any where.

lib/mayaUsd/nodes/stageData.h
HamedSabri-adsk: Good point. Looking into it.

HamedSabri-adsk: @ppt-adsk addressed in 4b8947b158e47d0e2f00802d49e56e3faf9d3fa2

We no longer need the workaround.

kxl-adsk: @HamedSabri-adsk @ppt-adsk I haven't yet reviewed the code, but this is change is outside of the scope for the duplicate implementation. Let's take it out of this PR and open a new one after.

HamedSabri-adsk: Ok, I added the change here since it was relevant. I will revert the change and open a separate PR if it makes thing easier.

kxl-adsk: Thank you, Hamed.

kxl-adsk (approved):

ppt-adsk (approved):

HamedSabri-adsk:
@kxl-adsk @ppt-adsk  I was just going through the PF logs and noticed that **2020-osx** had an error related no finding ufeScripts module which is very strange. 

```
      2/122 Test  #10: testProxyShapeBase ..............................***Failed   14.41 sec
      File "<string>", line 8, in <module>
      File "testProxyShapeBase.py", line 26, in <module>
        import usdUtils, mayaUtils
      File "/local/S/jenkins/workspace/ecg-mayausd-branch-preflight-2020-osx/build/RelWithDebInfo/test/python/usdUtils.py", line 26, in <module>
        import ufeUtils
      File "/local/S/jenkins/workspace/ecg-mayausd-branch-preflight-2020-osx/build/RelWithDebInfo/test/python/ufeUtils.py", line 28, in <module>
        from ufeScripts import ufeSelectCmd
    ImportError: No module named ufeScripts
```

test/lib/mayaUsd/nodes/testProxyShapeBase.py
```diff
@@ -23,7 +23,7 @@
 import os
 import unittest
 
-import usdUtils, mayaUtils
+import usdUtils, mayaUtils, ufeUtils
```
HamedSabri-adsk, line 26: Maya 2019 and 2020 don't have ufeSupport plugin, so we need to use a fallback.

https://github.com/Autodesk/maya-usd/blob/0b7a931b8149afacfafd1027cb655158ff400ebe/test/testUtils/ufeUtils.py#L28

ppt-adsk (approved):

lib/mayaUsd/nodes/stageData.h
kxl-adsk: FYI. @pmolodo @mattyjams @fabal

test/lib/mayaUsd/nodes/testProxyShapeBase.py
```diff
@@ -113,6 +115,8 @@ def testDuplicateProxyStageAnonymous(self):
         self.assertEqual(0, len(ufe.Hierarchy.hierarchy(duplProxyShapeItem).children()))
         self.assertEqual(1, len(ufe.Hierarchy.hierarchy(proxyShapeItem).children()))
 
+    # protect against Maya flavor: Maya PR 121 now has support for UFE path string.
+    @unittest.skipUnless(((ufeUtils.ufeFeatureSetVersion() >= 2) and (mayaUtils.previewReleaseVersion() >= 121)), 'testDuplicateProxyStageFileBacked is only available in Maya Preview Release 121 or later.')    
     def testDuplicateProxyStageFileBacked(self):
```
HamedSabri-adsk, line 120: - moved "ufeSelectCmd" under "testUtils" so it can be used in "test/lib/mayaUsd"
- added protection against Maya flavor: Maya PR 121 now has support for UFE path string.

ppt-adsk: Path strings have been in UFE v2 for a long time...  Why isn't ufeFeatureSetVersion() >= 2 sufficient?

HamedSabri-adsk: I was following what we had in testSelection. Should we clean it up too?

https://github.com/Autodesk/maya-usd/blob/80cddd12c85cf0678b758484e98d5ffb0ea58f4e/test/lib/ufe/testSelection.py#L140

ppt-adsk: I think in that case it was to protect for availability of the Maya select command improvements to support selection with path strings.  Since we're now at preview release 122 (soon 123), we should go back and clean all those obsolete preview release conditional test executions, in a cleanup pull request, no doubt.

PR #1143, opened 2021-02-03
use accurate MAYA_API_VERSION in guard for MRenderItem::setObjectTypeExclusionFlag()

mattyjams:
Maya PR 121 reports its API version as 20210000 while PR 122 reports its API version as 20220000. The newly added `MRenderItem::setObjectTypeExclusionFlag()` function whose use was added in pull request #1008 is only available in PR 122, not in PR 121. This means that compilation against PR 121 fails because the API version check guarding the function's usage is incorrect.

While I concede that we don't intend to support all previous preview/beta releases of Maya, this has been causing sufficient headache for us as we try to get PR 122 licensed that I was hoping taking these changes would be acceptable. The updated `MAYA_API_VERSION` also more accurately reflects the version in which the new API was added.


HdC-adsk (approved):
@mattyjams Thank you for making the change.

PR #1144, opened 2021-02-03
make tests in test/lib/ufe directly runnable and use fixturesUtils

mattyjams:
This change syncs the tests in `test/lib/ufe` with tests elsewhere in the source tree so that the script files are directly runnable (with the addition of a `__name__ == "__main__"` block at the bottom) rather than only as a module. They also now use `fixturesUtils` to ensure that Maya standalone is initialized and uninitialized in that case.

I also cleaned up and reordered the module imports at the top of each test script.

### DIFF
--- a/test/lib/ufe/testAttribute.py
+++ b/test/lib/ufe/testAttribute.py
@@ -16,17 +16,23 @@
 # limitations under the License.
 #
 
-import os
+import fixturesUtils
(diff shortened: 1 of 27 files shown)


ppt-adsk (approved):
Looks good to me!  Thanks for doing this cleanup @mattyjams.

PR #1145, opened 2021-02-03
MAYA-109438 - Need to remove attributes from AE stage view

seando-adsk:
MAYA-109438 - Need to remove attributes from AE stage view
* Temp suppress two non-functional attributes.

### DIFF
--- a/plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
+++ b/plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
@@ -38,12 +38,16 @@ global proc AEmayaUsdProxyShapeTemplate( string $nodeName )
         editorTemplate -suppress "stageCacheId";
         editorTemplate -suppress "outStageCacheId";
 
+        // MAYA-109438 - temp suppress these non-functional attributes.
+        editorTemplate -suppress "primPath";
+        editorTemplate -suppress "loadPayloads";
+
         editorTemplate -beginLayout "Stage" -collapse 0;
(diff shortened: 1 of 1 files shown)


seando-adsk:
@kxl-adsk As you can see the changes are trivial. I just suppress a couple of (non-functional) attributes so they won't appear in the AE. I don't think a preflight is needed here as it won't test anything. So please merge whenever you are ready.

plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
```diff
@@ -38,12 +38,16 @@ global proc AEmayaUsdProxyShapeTemplate( string $nodeName )
...
+        editorTemplate -suppress "primPath";
+        editorTemplate -suppress "loadPayloads";
+
         editorTemplate -beginLayout "Stage" -collapse 0;
             // We aren't displaying the "inStageData" attribute in the ProxyShapeInfo. We simply
             // pass it in order to get the current nodename. Cannot use the input one here as it
             // changes based on which proxy shape the AE is viewing.
             editorTemplate -callCustom "AEmayaUsdProxyShapeInfoNew" "AEmayaUsdProxyShapeInfoReplace" inStageData;
-            editorTemplate -addControl "loadPayloads";
+            //editorTemplate -addControl "loadPayloads";
```
kxl-adsk, line 50: I'm guessing we leave this dead code only because we will bring it back soon.

kxl-adsk (approved):

PR #1146, opened 2021-02-03
[MAYA-104104] Add import/export cmd documentation

pierrebai-adsk:


pierrebai-adsk:
Note that I have not removed the assembly bits from the base command (so I've also left the entries the tables).

lib/mayaUsd/commands/Readme.md
```diff
@@ -0,0 +1,440 @@
...
+further `kind`/model inference optional. The current behavior is:
+
+* We initially author the kinds on prims based on the `-kind` flag
+  in `MayaUSDExportCommand` (see the `MayaUSDExportCommand` flags above) or the
+  `USD_kind` attribute on individual Maya nodes (see the "Maya
+  Custom Attributes" table below).
+
+* For each root prim in the scene, we *validate* the kind if it
+  has been specified. Otherwise, we *compute* a kind for the root prim:
+  * If the root prim is specified to be an `assembly` (or type derived
```
pierrebai-adsk, line 152: I did not touch this section. Here there is a mention of assembly, which I was not sure of its correctness.

lib/mayaUsd/commands/Readme.md
```diff
@@ -0,0 +1,440 @@
...
+
+
+## `MayaUSDImportCommand`
+
+### Command Flags
+
+| Long flag              | Short flag | Type           | Default                           | Description |
+| ---------------------- | ---------- | -------------- | --------------------------------- | ----------- |
+| `-apiSchema`           | `-api`     | string (multi) | none                              | Imports the given API schemas' attributes as Maya custom attributes. This only recognizes API schemas that have been applied to prims on the stage. The attributes will properly round-trip if you re-export back to USD. |
+| `-assemblyRep`         | `-ar`      | string         | `Collapsed`                       | If the import results in the creation of assembly nodes, this value specifies the assembly representation that will be activated after creation. If empty, no representation will be activated. Valid values are: `Collapsed`, `Expanded`, `Full`, `Import`, `(empty)`. `Import`: No USD reference assembly nodes will be created, and the geometry will be imported directly. `(empty)`: The assembly is created but is left unloaded after creation. See "Importing as Assemblies" below for more detail on assembly node creation. |
```
mattyjams, line 33: "Assembly" here refers to Maya's `MFnAssembly` which is only relevant for the `pxrUsd` plugin, so I would assume we probably want to remove `-assemblyRep` here.

lib/mayaUsd/commands/Readme.md
mattyjams: This should all still be relevant, since "assembly" here refers to the USD kind and **not** to Maya's MFnAssembly.

lib/mayaUsd/commands/Readme.md
```diff
@@ -0,0 +1,440 @@
...
+exports). Chasers are registered with a particular name and can be
+passed argument name/value pairs in an invocation of `MayaUSDExportCommand`.
+There is no "plugin discovery" method here – the developer/user is
+responsible for making sure the chaser is registered.
+
+We provide one such chaser plugin called `AlembicChaser` to try to make
+integrating USD into Alembic-heavy pipelines easier. One of its features
+is that it can export all Maya attributes whose name matches a particular
+prefix (e.g. `ABC_`) as USD attributes by using its `attrprefix` argument.
+Here is an example of what that call to `mayaUSDExport` might look like:
```
mattyjams, line 236: The mechanism for registering chasers is in the core of mayaUsd, but the Alembic chaser was left behind in the `pxrUsd` plugin:
https://github.com/Autodesk/maya-usd/blob/dev/plugin/pxr/maya/plugin/pxrUsd/alembicChaser.cpp

So all of the Alembic-related sections below should probably be removed.

It might be nice though to keep a generic description and example here of what an invocation of `mayaUSDExport()` looks like when you're using a chaser that takes argument.

lib/mayaUsd/commands/Readme.md
pierrebai-adsk: Yes, it is planned for another PR.

pierrebai-adsk: The reason being that assembly-related command flags are still on the base class, probably by mistake we think, so it will require a code refactor.

lib/mayaUsd/commands/Readme.md
pierrebai-adsk: Would a simple mention that the alembic chaser is given as an example and is actually part of the pxr plugin be enough?

pierrebai-adsk: Ah, but if the alembic chaser is part of pxr, but the example mention the mayaUSDExport command, that would be a bit weird...

pierrebai-adsk: (OTOH, maybe any plugin can use the chaser of any other plugin? Then the example would at least work...)

lib/mayaUsd/commands/Readme.md
pierrebai-adsk: Finally I did the rework to move the assemblyRep flag in this PR.

lib/mayaUsd/commands/Readme.md
```diff
@@ -0,0 +1,441 @@
...
+| `-materialCollectionsPath`       | `-mcp`     | string           | none                | Path to the prim where material collections must be exported. |
+| `-mergeTransformAndShape`        | `-mt`      | bool             | true                | Combine Maya transform and shape into a single USD prim that has transform and geometry, for all "geometric primitives" (gprims). This results in smaller and faster scenes. Gprims will be "unpacked" back into transform and shape nodes when imported into Maya from USD. |
+| `-normalizeNurbs`                | `-nnu`     | bool             | false               | When setm the UV coordinates of nurbs are normalized to be between zero and one. |
+| `-pythonPerFrameCallback`        | `-pfc`     | string           | none                | Python function called after each frame is exported |
```
kxl-adsk, line 91: Ha, I just learned something new!

lib/mayaUsd/commands/Readme.md
```diff
@@ -0,0 +1,441 @@
...
+    model-hierarchy rules.
+  * Else, if model is an `assembly`, ensure that all the intermediate
+    prims between the root and the locations at which other assets
+    are referenced in get `kind=group` so that there is a contiguous
+    hierarchy of models from the root down to the "leaf" model references.
+
+
+#### UV Name Translation: `map1` becomes `st`
+
+Currently, for Mesh export (and similarly for NurbsPatch, also),
```
kxl-adsk, line 181: @JGamache-autodesk can you please review this section.

lib/mayaUsd/commands/Readme.md
```diff
@@ -0,0 +1,441 @@
...
+of the highest prim(s) imported. This is generally the fullDagPath
+that corresponds to the imported primPath but could be multiple
+paths if primPath="/".
+
+
+## `MayaUSDExportCommand`
+
+### Command Flags
+
+| Long flag                        | Short flag | Type             | Default             | Description |
```
kxl-adsk, line 59: There are few flags missing here:
-dsp -exportDisplayColor             on
-fst -frameStride                    Float
-mfc -melPerFrameCallback           
-mpc -melPostCallback               
-msn -materialsScopeName             String

kxl-adsk (changes requested):
Thank you! This starts to look very good. I left a comment about missing flags that I discovered while reviewing.

There is as well a question about the existing document in pxr plugin - @mattyjams what would your recommendation be?

lib/mayaUsd/commands/Readme.md
pierrebai-adsk: Will look it up, I thought I had compared carefully with the help from within Maya, but I must have missed them somehow.

pierrebai-adsk: I don't understand the mel callbacks: they take no argument... yet the call back seem that it should be a string?

kxl-adsk (approved):

lib/mayaUsd/commands/Readme.md
JGamache-autodesk: Oups, sorry about that. This section should have been updated with the UV set fixes of #876.

The current behavior will preserve uvSet names.
To get back the old renaming behavior requires setting environment variables:
on export, renaming requires MAYAUSD_EXPORT_MAP1_AS_PRIMARY_UV_SET to be set.
on import, renaming depends on MAYAUSD_IMPORT_PRIMARY_UV_SET_AS_MAP1.

lib/mayaUsd/commands/Readme.md
```diff
@@ -0,0 +1,445 @@
...
+
+The purpose of this command is to translate between nice-names, internal
+names and annotations for various elements passed to the other commands.
+
+### Command Flags
+
+| Long flag              | Short flag | Type           | Description |
+| ---------------------- | ---------- | -------------- | ----------- |
+| `-export`              | `-ex`      | noarg          | Retrieve the list of export shading mode nice names. |
+| `-exportOptions`       | `-eo`      | string         | Retrieve the real name of the shading mode given its nice name. These can be used in the `shadingMode` option of the export command. |
```
JGamache-autodesk, line 374: Slightly more than that. Specifying a material export mode is done with two parameters:
 -shadingMode
 -convertMaterialsTo
So `exportOptions` will return the real values for both these parameters to get the expected material export.

lib/mayaUsd/commands/Readme.md
```diff
@@ -0,0 +1,445 @@
...
+| `-exportAnnotation`    | `-ea`      | string         | Retrieve the description of the export shading mode option |
+| `-findExportName`      | `-fen`     | string         | Retrieve the nice name of an export shading mode |
+| `-findImportName`      | `-fin`     | string         | Retrieve the nice name of an import shading mode |
+| `-import`              | `-im`      | noarg          | Retrieve the list of import shading mode nice names. |
+| `-importOptions`       | `-io`      | string         | Retrieve the real name of the shading mode given its nice name. These can be used in the `shadingMode` option of the export command. |
```
JGamache-autodesk, line 379: Returns a pair of strings that completely defines a `shadingMode` entry, as used by the USD import options.

lib/mayaUsd/commands/Readme.md
mattyjams: Yeah, in the short term, if you wanted to add something that points to the Alembic chaser as an example, and includes a mention that you would need to `cmds.loadPlugin('pxrUsd')` for the example to work, that might be enough for now.

Longer term, I think we'll want to decide what to do with the Alembic chaser. We (Pixar) never used it internally. It was written really early on when we open sourced the plugin to make it a bit easier for external users to try out USD. If people are still making use of it, it should probably get moved to the adsk plugin or some other new alembicUsdExportPlugin or whatever. If it stays in pxrUsd, it'll probably end up going away with the rest of the stuff in there.

mattyjams:
> There is as well a question about the existing document in pxr plugin - @mattyjams what would your recommendation be?

Sorry I missed this. What's the question, exactly? :)

I typically suggest using the adsk plugin for anyone considering new pipeline that takes advantage of this project, so I think the goal would be that the `pxrUsd` plugin and ultimately everything under `plugin/pxr` eventually get retired and removed. We are not actively updating the pxr import and export commands anymore, nor are we ensuring that the pxr-specific doc stays up-to-date with new options being added to the core.

mattyjams (approved):
Looks good to me!

There's a lot of surface area to cover, so I'm sure we'll find little things here and there as we keep going, but this seems like a great start.

lib/mayaUsd/commands/Readme.md
pierrebai-adsk: done

JGamache-autodesk: The review comment was for the line above. This means the fix you did to "-importAnnotation" should have been done to "-importOptions" instead.

pierrebai-adsk: Done.

JGamache-autodesk (approved):

PR #1149, opened 2021-02-04
Support empty blendshape targets

ysiewappl:
Hi:

This PR supplements the previous added support for exporting blendshapes with some minor changes to the names of the targets exported, and also adds support for exporting empty blendshape targets, gated by the `ignoreWarnings` flag.

### DIFF
--- a/lib/usd/translators/meshWriter_BlendShapes.cpp
+++ b/lib/usd/translators/meshWriter_BlendShapes.cpp
@@ -311,9 +311,21 @@ MStatus mayaGetBlendShapeInfosForMesh(
         MObject curBlendShape = itDg.currentItem();
         TF_VERIFY(curBlendShape.hasFn(MFn::kBlendShape));
 
-        MPlug outputGeomPlug = itDg.thisPlug();
-        TF_VERIFY(outputGeomPlug.isElement() == true);
-        unsigned int outputGeomPlugIdx = outputGeomPlug.logicalIndex();
+        MPlug outputGeomElemPlug = itDg.thisPlug();
+        // NOTE: (yliangsiew) Because this can end up grabbing the wrong plug (i.e.
+        // blendShape.weight[1]), we double-check here and advance the iterator if necessary.
+        if (!outputGeomElemPlug.isElement()) {
+            itDg.next();
+            continue;
+        }
+        MPlug outputGeomPlug = outputGeomElemPlug.array(&stat);
+        CHECK_MSTATUS_AND_RETURN_IT(stat);
+        MString outputGeomPlugName = outputGeomPlug.partialName(0, 0, 0, 0, 0, 1);
+        if (outputGeomPlugName != "outputGeometry") {
+            itDg.next();
+            continue;
+        }
+        unsigned int outputGeomPlugIdx = outputGeomElemPlug.logicalIndex();
 
         // NOTE: (yliangsiew) Because we can have multiple output
         // deformed meshes from a single blendshape deformer, we have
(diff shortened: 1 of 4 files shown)


lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -636,8 +674,13 @@ MObject PxrUsdTranslators_MeshWriter::writeBlendShapeData(UsdGeomMesh& primSchem
                         CHECK_MSTATUS_AND_RETURN(stat, MObject::kNullObj);
                         // NOTE: (yliangsiew) Because a single weight can drive multiple targets, we
                         // have to put a numeric suffix in the target name.
```
mattyjams, line 676: Is this comment still accurate now? Or maybe it's only accurate in a certain condition (`k != 0`)?

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -804,12 +849,20 @@ MObject PxrUsdTranslators_MeshWriter::writeBlendShapeData(UsdGeomMesh& primSchem
...
```
mattyjams, line 851: Similar here. These could maybe just move into the `else` blocks?

mattyjams (approved):
I'm still not particularly well-versed in blend shapes, but seems reasonable to me!

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -505,19 +527,31 @@ void findUnionAndProcessArrays(
...
         const VtVec3fArray& origOffsetsArray = offsetsArrays[i];
         const size_t        numOrigOffsets = origOffsetsArray.size();
-        TF_VERIFY(numOrigOffsets != 0);
-        VtVec3fArray& newOffsetsArray = unionOffsetsArrays[i];
+        VtVec3fArray&       newOffsetsArray = unionOffsetsArrays[i];
         newOffsetsArray.assign(numUnionIndices, GfVec3f(0.0f));
 
         const VtVec3fArray& origNormalsArray = normalsArrays[i];
+        const size_t        numOrigNormals = origNormalsArray.size();
 #ifdef _DEBUG
```
williamkrick, line 535: I think we don't need the _DEBUG preprocessor macro anymore.

williamkrick (approved):
It makes sense to me. Do you have a lot of scenes with no-op blend shape targets? I don't understand the circumstances which would generate Maya scenes that cause this code path to be used. Clearly it is happening because you've found you need it, I just don't have the imagination to figure out why.

Can you share the workflow that produces this type of blend shape?

ysiewappl:
> It makes sense to me. Do you have a lot of scenes with no-op blend shape targets? I don't understand the circumstances which would generate Maya scenes that cause this code path to be used. Clearly it is happening because you've found you need it, I just don't have the imagination to figure out why.
> 
> Can you share the workflow that produces this type of blend shape?

We have a need by our internal artists that requires "placeholder" blendshapes be present, so that client code in other applications used for triggering various shapes across different USD files is able to at least find an empty delta to trigger (even if the shape isn't explicitly authored). It's arguable that this should be on client-side parsing code to handle...but Maya itself also allows for "empty" targets to be created via the GUI itself by the average non-technical end-user (i.e. duplicate a cube twice and apply the second cube as a blendshape target of the first without making any topology changes), so we thought it was best to kill two birds with one stone in this case and support both use cases.

lib/usd/translators/meshWriter_BlendShapes.cpp
ysiewappl: Resolved in 522b6b0.

lib/usd/translators/meshWriter_BlendShapes.cpp
ysiewappl: Resolved in 522b6b0.

lib/usd/translators/meshWriter_BlendShapes.cpp
ysiewappl: Resolved in 522b6b0.

kxl-adsk:
@ysiewappl I was about to merge this change, but how about we add a test to cover the improvement you made?

ysiewappl:
> @ysiewappl I was about to merge this change, but how about we add a test to cover the improvement you made?

This is addressed in 824510a.

PR #1150, opened 2021-02-04
MAYA-109683: Remove workaround hack around not having a strong reference to the stage

HamedSabri-adsk:
See the logged issue for more infomration: https://github.com/Autodesk/maya-usd/issues/528

As part of PR https://github.com/Autodesk/maya-usd/pull/1142, stageData no longer holds a weak pointer to the stage and therefore we no longer this workaround hack.


HamedSabri-adsk:
@kxl-adsk based on our conversation in PR 1142, I created this PR to remove the workaround hack. 

This PR can wait for now until PR1142 is merged to dev first.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -605,41 +605,31 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
                 loadSet = UsdStage::InitialLoadSet::LoadNone;
             }
 
-            {
-                // When opening or creating stages we must have an active UsdStageCache.
-                // The stage cache is the only one who holds a strong reference to the
-                // UsdStage. See https://github.com/Autodesk/maya-usd/issues/528 for
-                // more information.
-                UsdStageCacheContext ctx(
```
kxl-adsk, line 613: A stage cache is needed for the re-use of stages across proxy shapes. Issue #528 was logged when we worked on in-memory stage workflows, and here leveraging stage cache doesn't make much sense, since we are not going to re-use that stage in another proxy.

Your code should look like what was there before https://github.com/Autodesk/maya-usd/pull/519/commits/4af6c19ba9a4e4984730ee096792d48d196b7b24, i.e. UsdStageCacheContext is used when we have a layer from file and need to create a stage for it.

Having said all of that, I know we have an internal ticket (MAYA-109039) that will require changes in the same place. How about we close this PR and make the change as part of MAYA-109039?

kxl-adsk (changes requested):

PR #1152, opened 2021-02-05
Maya 105566/serialize usd edits

fowlertADSK:
This is a continuation of an earlier start on saving Usd edits that was posted as a draft review here:  https://github.com/Autodesk/maya-usd/pull/871

This PR contains the option to save edits into the Maya file, or to push edits back to the .usd files themselves.  Saving back to Usd will mean choosing files for all anonymous layers other than the Session layer.  This can be done automatically by Maya or through a new UI contained in this PR.

**New UI Included in this PR**

The UI to choose which option to use has been integrated into the File Save options dialog box.  This UI is contained in this PR but additional hooks were required to be added to Maya as well, so this UI will only appear in Maya PR’s starting now.
image

Before a Maya Save/Export, if the user has not chosen “Don’t ask me again”, a dialog will appear to allow the user to make the choice.
image

If there are anonymous layers in the scene being saved the user will be given an opportunity to choose their new file names/locations.  The list of file-backed layers that will also be saved will be listed as well.
image

This same batch save dialog will also now be used in the Layer Editor for saving individual stages.  Previously a user would have to first save their anonymous layers individually before the save operation would work on all dirty file-backed layers.  Now this same dialog will be posted, giving the user the ability to save both their anonymous and file-backed layers at the same time. 

In the previous save layer workflow, the option UI to save .usd files as either binary or ascii was embedded in the file dialog itself.  This has now been moved to the Layer Editor options menu.
image

**Option to save back to Usd files**

Overview of how it works:

1. Iterate over all proxy shapes, and their UsdStages, looking for types that have been registered with the Layer Manager as ones that are supported.  This is currently only the ASDK ProxyShape node, both the Animal Logic and Pixar proxy shape nodes will be skipped.
2. Recursively iterate through all UsdStage layer stacks looking for any that have dirty layers.
3. Keep a list of the layers that need to be saved as well as the parent layers that may have them as a contained sublayer.
4. The Layer Manager can have a callback registered with it that is called at this point with the list of UsdStages.  If no callback is registered, the LM will continue to save edits depending on the current save option.  If it needs to save anonymous layers to disk then it will automatically choose names for them.
5. Currently the Layer Editor will register a callback with the Layer Manager, so in step 4 the LM will end up calling the LE and this is where the batch save dialog UI is posted.
 
Note: The Maya callback that starts all of this is a “Check” callback, so canceling in any of the dialogs will abort the file IO operation.  

**Option to save edits to Maya files**

The behavior to save edits into Maya files is mostly unchanged from the previous draft review.  
1. The Layer Manager node is created in a pre-save callback if there are dirty layers to write to the Maya file.
2. In a post-read callback the Usd layers saved into the file are recreated and the LM node is deleted. A map of original and new Usd identifiers to the newly created layers is kept so the new layer can be looked up using either the old or new identifier.
3. After creating the layers, we edit the sub layer paths of any layer whose identifier has changed. This is what allows us to save/restore all anonymous layers.
4. The base proxy shape uses the saved map to query what session/root layers to use when first evaluating its stage.
5. LM uses a stage loaded callback to determine when to remove layers from its map. After each proxy shape has loaded its stage the LM map should be empty and not be holding any references to the layers it created.
6. If there are dirty layers that need to be written out, the LM will write an entry for every loaded layer whether it was dirty or not, but only the dirty layers will be exported to a string and have their data stored in the Maya file.


lib/usd/ui/layerEditor/batchSaveLayersUIDelegate.h
```diff
@@ -0,0 +1,36 @@
...
+#include <pxr/usd/usd/common.h>
+
+#include <vector>
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+namespace UsdLayerEditor {
+
+MAYAUSD_UI_PUBLIC
+bool batchSaveLayersUIDelegate(const std::vector<UsdStageRefPtr>&);
```
fowlertADSK, line 32: This callback is registered with the Layer Manager to provide a UI to choose file names.  The core library does not depend on Qt and can't be the one to create UI.

lib/usd/ui/layerEditor/layerEditorWidget.cpp
```diff
@@ -112,14 +120,48 @@ void setupDefaultMenu(SessionState* in_sessionState, QMainWindow* in_parent)
...
+        auto formatArgsOption
+            = UsdMayaUtil::convert(MayaUsdOptionVars->mayaUsd_SaveLayerFormatArgBinaryOption);
+        if (MGlobal::optionVarExists(formatArgsOption)) {
+            isBinary = MGlobal::optionVarIntValue(formatArgsOption) != 0;
+        }
+        isBinary ? formatBinary->setChecked(true) : formatAscii->setChecked(true);
+        auto onFileFormatChanged = [=](QAction* action) {
+            MGlobal::setOptionVarValue(formatArgsOption, action->data().toInt());
+        };
+        QObject::connect(formatGroup, &QActionGroup::triggered, in_parent, onFileFormatChanged);
```
fowlertADSK, line 163: This change is to add the binary or ascii option for .usd files to the Layer Editor menu.  It was previously in the file dialog itself.

lib/usd/ui/layerEditor/layerTreeModel.cpp
```diff
@@ -410,81 +410,31 @@ LayerTreeModel::getAllAnonymousLayers(const LayerTreeItem* item /* = nullptr*/)
 
 void LayerTreeModel::saveStage(QWidget* in_parent)
 {
-    QString dialogTitle = StringResources::getAsQString(StringResources::kSaveStage);
```
fowlertADSK, line 413: We originally only displayed an error when trying to save while there were anonymous layers.  Then we changed to a dialog for anonymous layers and a message dialog when you only had file-backed to save.  Now we are removing those two separate cases and instead showing the dialog that has two sections, one for anonymous layers and the other to list the file-backed.

lib/usd/ui/layerEditor/mayaSessionState.cpp
```diff
@@ -197,19 +197,24 @@ void MayaSessionState::nodeRenamedCB(MObject& obj, const MString& oldName, void*
...
+    // this does not work:
+    //        if OpenMaya.MFnDependencyNode(obj).typeName == PROXY_NODE_TYPE
+    if (obj.hasFn(MFn::kShape)) {
+        MDagPath dagPath;
+        MFnDagNode(obj).getPath(dagPath);
+        auto shapePath = dagPath.fullPathName();
+
+        StageEntry entry;
+        if (getStageEntry(&entry, shapePath)) {
+            Q_EMIT stageRenamedSignal(entry._displayName, entry._stage);
```
fowlertADSK, line 217: This "node renamed" callback can be called during file IO but did not have an "Idle" implementation like other callbacks here.  Now nodeNamedCB just uses a QTimer to run nodeRenamedCBOnIdle at a later time.

lib/usd/ui/layerEditor/mayaSessionState.cpp
```diff
@@ -226,35 +231,7 @@ bool MayaSessionState::saveLayerUI(
...
-
-    // unambiguous formats
-    if (extension == "usda" || extension == "usdc") {
-        *out_pFormat = extension.toStdString();
-    } else {
-        *out_pFormat = binary ? "usdc" : "usda";
-    }
-
-    return true;
+    return SaveLayersDialog::saveLayerFilePathUI(*out_filePath, *out_pFormat);
```
fowlertADSK, line 234: Using the new dialog instead of the MEL scripts now.

lib/usd/ui/layerEditor/saveLayersDialog.h
```diff
@@ -1,25 +1,43 @@
...
-        QWidget*                           in_parent);
+    typedef std::unordered_multimap<SdfLayerRefPtr, std::string, TfHash> stageLayerMap;
+
+    // Create dialog using single stage (from session state).
+    SaveLayersDialog(SessionState* in_sessionState, QWidget* in_parent);
+
+#if defined(WANT_UFE_BUILD)
+    // Create dialog for bulk save using all stages.
+    SaveLayersDialog(QWidget* in_parent, const std::vector<UsdStageRefPtr>& stages);
+#endif
```
fowlertADSK, line 34: Previously this dialog was only used from the Layer Editor itself, but it can now be displayed from a File -> Save operation.  The first constructor is used from the LE, and the dialog gets most of the information it needs from the SessionState.  From the File -> Save callback function we have already iterated over all of the proxies and their stages and have the list of stages we need to save, so it uses this second constructor.  Saving of Usd edits does require UFE so you will see some ifdef's for that.

plugin/adsk/plugin/ProxyShape.cpp
```diff
@@ -69,4 +74,18 @@ void ProxyShape::postConstructor()
...
+    return LayerManager::findLayer(UsdMayaUtil::convert(rootLayerName));
+}
+
+SdfLayerRefPtr ProxyShape::computeSessionLayer(MDataBlock& dataBlock)
+{
+    auto sessionLayerName = dataBlock.inputValue(sessionLayerNameAttr).asString();
+    return LayerManager::findLayer(UsdMayaUtil::convert(sessionLayerName));
+}
+#endif
+
```
fowlertADSK, line 90: As in the previous PR, this is how a supported proxy shape can respond to the compute in the base proxy with the serialized layers.

plugin/adsk/plugin/plugin.cpp
```diff
@@ -315,6 +320,13 @@ MStatus initializePlugin(MObject obj)
         }
     }
 
+#if defined(WANT_UFE_BUILD)
+    MayaUsd::LayerManager::addSupportForNodeType(MAYAUSD_NS::ProxyShape::typeId);
+#if defined(WANT_QT_BUILD)
+    MayaUsd::LayerManager::SetBatchSaveDelegate(UsdLayerEditor::batchSaveLayersUIDelegate);
+#endif
```
fowlertADSK, line 327: Hooks up the callback to provide a UI to choose names/locations for anonymous layers during save.

lib/mayaUsd/nodes/layerManager.h
```diff
@@ -0,0 +1,117 @@
...
+    type that have UsdStages with dirty Layers in them.  If stages are found, there are currently
+    three options for how to handle the Usd edits:
+
+    1. Save back to .usd files.
+    There are three steps to handling this option.
+    a) All anonymous layers must be saved to disk.  If a batch save delegate has been installed,
+    and Maya is running in interactive mode, then a UI dialog can be displayed to provide a choice
+    of file names and locations for all anonymous layers.  If Maya is not running in interactive
+    mode, or there is no installed delegate to handle it, then Maya will automatically choose name
+    and locations for all anonymous layers.
```
aloysbaillet, line 54: I think there's a case to be made for a new "Save back to .usd files and keep anonymous layers" case which would save all external files, but not force people to choose file locations for any anonymous layer in the scene. This is a use case we have currently where if a USD file is supposed to be saved on disk we create the file on disk, but for data that we prefer not having to choose a file location then we use anonymous layers. This early decision shouldn't have to be compromised at save time by forcing users to choose between saving their unsaved USD files on disk and giving up on anonymous layers and having to create new files for all.

lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -0,0 +1,909 @@
...
+                    continue;
+                }
+            }
+
+            LayerDatabase::instance().addLayer(layer, identifierVal);
+            createdLayers.push_back(layer);
+        }
+    }
+
+    removeManagerNode(lm);
```
aloysbaillet, line 674: It would be great to mark the maya scene as clean once the layer manager node is removed so that the "unsaved scene" dialog doesn't pop up after opening a scene that contains a layer manager node and opening a new scene (or closing maya) right after.

aloysbaillet (commented):
Thanks @fowlertADSK !
The overall feature set looks great, my main comment would be to add a fourth mode which is to save file-backed USD files but keep anonymous layers anonymous. I believe that's what we would use by default.
Also it would be great to check that this all works fine with autosave, as we just noticed some events such as kBeforeSave are not sent by maya during autosave...

aloysbaillet:
I also just thought that it would be great to have the dag path of the proxy shape next to each anonymous layer (and maybe the named ones as well!) in the save dialog as it would be quite tricky to know which layer comes from which shape if you have more than one in the scene...

seando-adsk:
@aloysbaillet There are tooltips on both the anonymous and file backed layers that show the proxy shape(s) they are from. Does that work?

aloysbaillet:
Oh that's great! Thanks Sean I didn't see that. I will try to build that branch and give it a go in maya over the next few days.

lib/mayaUsd/base/tokens.h
```diff
@@ -28,9 +28,17 @@ PXR_NAMESPACE_OPEN_SCOPE
...
+// mayaUsd_ProxyTargetsSessionLayerOnOpen: Always target a session layer on a mayaUsdProxy
+// mayaUsd_SaveLayerFormatArgBinaryOption: When saving as .usd, should the format be binary
+// mayaUsd_SerializedUsdEditsLocation: Option for what to do with Usd edits when the current
+// Maya scene is about to be saved.  optionVar values are:
+//    1: save all edits back to usd files.
+//    2: export the dirty usd layers to Maya string attributes to be serialized to the Maya file.
+//    3: ignore all Usd edits.
+// mayaUsd_SerializedUsdEditsLocationPrompt: optionVar to force a prompt on every save
+#define MAYA_USD_OPTIONVAR_TOKENS                                                     \
+    (mayaUsd_ProxyTargetsSessionLayerOnOpen)(mayaUsd_SaveLayerFormatArgBinaryOption)( \
```
pmolodo, line 40: Minor, but I'd like to stick the the convention of using `// clang-format off` for token defintions - ie, like in [jobArgs.h](https://github.com/Autodesk/maya-usd/blob/1ac15c974d7f57f5e0c524508779cad5abbcdc15/lib/mayaUsd/fileio/jobs/jobArgs.h#L37).  I know, I was one of the big proponents of `clang-format`, but I think this is an exception where it looks much nicer manually formatted, and makes for much cleaner diffs if we need to add/remove tokens later...

plugin/adsk/plugin/plugin.cpp
```diff
@@ -315,6 +320,13 @@ MStatus initializePlugin(MObject obj)
         }
     }
 
+#if defined(WANT_UFE_BUILD)
+    MayaUsd::LayerManager::addSupportForNodeType(MAYAUSD_NS::ProxyShape::typeId);
```
pmolodo, line 324: It looks like right now we're only using the new LayerManager for the autodesk proxy-shape.  The AL one already has it's own version of this, which you didn't want to worry about unifying... but what about the Pixar proxy shape?  Are there any reasons you know of why the Pixar proxy shape wouldn't work, or were you just deferring that decision to Pixar?

@mattyjams, any thoughts on whether you think this would work for the Pixar proxy?  Ideally, I like bringing all the proxy shapes closer together, and adding this to the Pixar too would help with that larger goal...

mattyjams: `pxrUsdProxyShape` effectively does **not** support UFE. It's only valid use anymore as far as I'm concerned is for anyone continuing to use `pxrUsdReferenceAssembly`, where activating the "Collapsed" or "Expanded" representation of the  assembly node brings in transient `pxrUsdProxyShape` nodes. The assembly node (poorly) supports edits using the Maya assembly framework instead of UFE, so I wouldn't really consider any workflow involving either of the pxr nodes UFE-compatible.

We're actively working towards deprecating use of our own nodes (and indeed everything under `plugin/pxr`) and switching over to `mayaUsdProxyShape` instead, and I'd advocate for others to do the same.

lib/mayaUsd/nodes/layerManager.h
pmolodo: I think this is a good point - both because it makes a lot of sense as default when running in batch mode, as @aloysbaillet points out, and also because I thin it "conceptually fits" with anonymous layers vs file-backed layers.  That is, I think of anonymous layers as more for temporary storage, which might make sense saved out to the maya scene, whereas file-backed layers are generally for more "permanent" changes which you might want to, ie, share with others, publish downstream, etc.

However - my next thought was that, since the session layer is already treated in a special manner (ie, always saved to maya), and the session layer is the only anonymous layer we use at Luma - and likely the only anonymous layer 90%+ of the time - we would effectively already be getting the behavior @aloysbaillet describes, because of the special session-layer handling.

@aloysbaillet - I'm curious, does Animal Logic make a lot of use of non-session-layer anonymous layers? Or were you thinking more of "what if" and "just in case"?

pmolodo (approved):
Mostly looks good!  Looking forward to getting these changes in..

plugin/adsk/plugin/plugin.cpp
pmolodo: Ok, makes sense!

lib/mayaUsd/base/tokens.h
fowlertADSK: Yes!  Sean actually mentioned that to me when I complained that it kept formatting the change like this, but I forgot to go back and add that.

lib/mayaUsd/nodes/layerManager.cpp
fowlertADSK: Good point, I'll look into that.  Can't remember off the top of my head if there's a MGlobal/MFileIO call to easily do that or not...

lib/mayaUsd/nodes/layerManager.h
fowlertADSK: Just to be clear, are we talking about putting all anonymous layers into the Maya file (and saving file backed layers to disk) or are you suggesting an option where we ignore anonymous layers entirely (because you consider them kind of temporary)?

lib/mayaUsd/base/tokens.h
mattyjams: Just wanted to flag also that there was a bit of clean-up in this area from #1119 that probably created a merge conflict that needs resolving. I put the `mayaUsd_` prefix in the values only:

```C++
...
    ((SaveLayerFormatArgBinaryOption, "mayaUsd_SaveLayerFormatArgBinaryOption")) \
...
```

lib/mayaUsd/nodes/layerManager.h
fowlertADSK: @aloysbaillet  we talked some internally about your suggestion and I think we’d like to get this merged with the existing options for testing before moving on to additional changes.  Right now the UI does implement what was requested by design so we have completed the task.  I think you also had some questions about overriding some of the file naming/saving behavior, which is what we were starting to think about with the delegate function that the plugin registers to have the UI lib help, so maybe we can work to get this in and then start up a new conversation about next steps.  What do you think?

aloysbaillet: Yes I confirm I would like an option to save all external files, but keep anonymous layers anonymously saved inside the maya file. And I'm happy to wait for that option!
The tension between simplicity and flexibility is quite apparent here, adding another option will make things more confusing for users, and the layer manager can be used to save all external files first, then the option of saving all unsaved layers in the maya scene would work for us.

PR #1153, opened 2021-02-05
MAYA-107222 - Clear bounding box cache when editing USD stage

JGamache-autodesk:


kxl-adsk (approved):

kxl-adsk:
@JGamache-autodesk Please add a UFE manipulation test to cover the invalidation and bounding box computation.

kxl-adsk (approved):

PR #1156, opened 2021-02-08
MAYA-109647 - Target becomes out of synch when an ancestor of a targe…

spinell-adsk:
…t layer is being removed

Set the root layer as the edittarget when the current edit target is removed

### DIFF
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -161,18 +161,16 @@ public:
             // ignore : Optional layer used has the root of a hierarchy that
             //          we don't want to check in.
             // ignoreSubPath : Optional subpath used whith ignore layer.
-            auto isInHierarchy = [this](
-                                     const SdfLayerHandle& rootLayer,
-                                     const SdfLayerHandle& layer,
-                                     const SdfLayerHandle* ignore = nullptr,
-                                     const std::string*    ignoreSubPath = nullptr) {
+            auto isInHierarchy = [](const SdfLayerHandle& rootLayer,
+                                    const SdfLayerHandle& layer,
+                                    const SdfLayerHandle* ignore = nullptr,
+                                    const std::string*    ignoreSubPath = nullptr) {
                 // Impl used for recursive call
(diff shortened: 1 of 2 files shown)


fowlertADSK (commented):
Just wondering if instead of recursively trying to figure everything out we could just check if the edit target got lost after the operation?  If we do correctly catch the situation then we’re just going to go with the root layer anyways right?

e.g. logic kind of like…
auto originalEditTarget = stage->GetEditTarget();
… do work …
auto currentEditTarget = stage->GetEditTarget();

if (nullptr != originalEditTarget && nullptr == currentEditTarget)
{
   // The edit target was set to something but now it isn’t set to anything.
   // Must have removed a parent layer?
   stage->SetEditTarget(stage->GetRootLayer());
}

fowlertADSK:
Replying to my own comment...  Took another look at this and I think I missed something when I first looked at it.  Is it the case that you are not trying to figure out AFTER the fact that the edit target got removed, you are trying to figure out BEFORE you attempt to remove the layer so that the edit is stored in something that will still be valid after the operation.

I first read the diff as trying to clean up after the fact and I was going to suggest using something like stage->HasLocalLayer(currentTarget.GetLayer()) to figure out of the layer is still valid after the operation, but now that I look closer this is all happening before the removal.

fowlertADSK (approved):

spinell-adsk:
@fowlertADSK what you were proposing about querying stage->GetEditTarget() after removing the layer it's exactly what I first try. There is a few issue with this way.

When removing an anonymous layer, the command keep a reference to the layer for the undo operation. So USD never delete the layer and stage->GetEditTarget() return the same layer after the remove operation. 

If the layer to remove is based on a file and the LayerEditor windows is open, USD will not delete the layer when calling RemoveSubLayerPath() because the UI still have a reference on the layer. The layer will be delete on th next UI refresh which happen after the command end and tage->GetEditTarget() return the same layer after the remove operation.

If the layer to remove is based on a file and the LayerEditor windows is not open, USD will delete the layer when calling RemoveSubLayerPath() and and tage->GetEditTarget() return a invalid layer.

PR #1157, opened 2021-02-08
MAYA-104490 geom subset material bind support for meshes

williamkrick:
This commit adds support for geom subset material binding for meshes.

It includes a large refactor of behavior in HdVP2Mesh which I'll explain here.

The addition of geom subsets adds another layer of complexity to HdVP2Mesh, and how it evaluates and stores data. The classes are organized like so:

- HdVP2Mesh::Sync for a given reprToken (smoohHull, wire, bbox)
  - One or more repr descriptions for the given token (Hull, Edge, etc)
    - One or more render items per description (multiple hull items to represent subset material bindings)

The code changes move the responsibility for update & storage of the various pieces of data to the highest level possible, so that the data can be shared among many render items.


lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -71,10 +71,6 @@ struct CommitState
 
     //! If valid, new index buffer data to commit
     int* _indexBufferData { nullptr };
-    //! If valid, new color buffer data to commit
```
williamkrick, line 74: The changes to basisCurves are mirroring part of the refactor in did in mesh. It's all centered around using _CommitVertexBuffer to create a small main thread task that commits one buffer, vs. saving all the committing for later.

In basisCurves.cpp I didn't go as far with the refactor as I did in mesh.cpp, but I believe the same types of changes I was making there could be made here.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -697,7 +710,7 @@ void HdVP2BasisCurves::_UpdateDrawItem(
 
     HdDirtyBits itemDirtyBits = drawItem->GetDirtyBits();
 
-    CommitState                    stateToCommit(*drawItem);
```
williamkrick, line 700: Minor signature change. We're commit data for a render item, so the CommitState doesn't get access to the drawItem anymore.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
```diff
@@ -25,21 +25,20 @@ PXR_NAMESPACE_OPEN_SCOPE
...
 */
-HdVP2DrawItem::HdVP2DrawItem(HdVP2RenderDelegate* delegate, const HdRprimSharedData* sharedData)
+HdVP2DrawItem::HdVP2DrawItem(
+    HdVP2RenderDelegate*     delegate,
+    const HdRprimSharedData* sharedData)
     : HdDrawItem(sharedData)
     , _delegate(delegate)
 {
     // In the case of instancing, the ID of a proto has an attribute at the end,
-    // we keep this info in _renderItemName so if needed we can extract proto ID
```
williamkrick, line 33: A draw item now has multiple render items, so this member was renamed to _drawItemName

lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
```diff
@@ -25,21 +25,20 @@ PXR_NAMESPACE_OPEN_SCOPE
...
+    // we keep this info in _drawItemName so if needed we can extract proto ID
     // and use it to figure out Rprim path for each instance. For example:
     //
     //   "/Proxy/TreePatch/Tree_1.proto_leaves_id0"
     //
-    _renderItemName = GetRprimID().GetText();
-    _renderItemName += TfStringPrintf("/DrawItem_%p", this).c_str();
-
-    _renderItemData._indexBuffer.reset(
-        new MHWRender::MIndexBuffer(MHWRender::MGeometry::kUnsignedInt32));
```
williamkrick, line 42: Each render item has it's own index buffer, so the index buffer is not store on the draw item anymore.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.h
```diff
@@ -38,19 +39,17 @@ class HdVP2RenderDelegate;
...
 {
 public:
-    //! A primvar vertex buffer map indexed by primvar name.
-    using PrimvarBufferMap = std::
-        unordered_map<TfToken, std::unique_ptr<MHWRender::MVertexBuffer>, TfToken::HashFunctor>;
-
     //! Helper struct providing storage for render item data
     struct RenderItemData
     {
-        //! Render item color buffer - use when updating data
```
williamkrick, line 48: RenderItem share buffers, so this data moved up to the shape.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.h
```diff
@@ -89,21 +91,44 @@ class HdVP2DrawItem final : public HdDrawItem
...
 
+    void AddRenderItem(MHWRender::MRenderItem* item, const HdGeomSubset* geomSubset = nullptr);
+
+    using RenderItemDataVector = std::vector<RenderItemData>;
+    RenderItemDataVector& GetRenderItems() { return _renderItems; }
+
     /*! \brief  Get access to render item data.
      */
-    RenderItemData& GetRenderItemData() { return _renderItemData; }
+    RenderItemData& GetRenderItemData()
```
williamkrick, line 101: I left all of these convenience methods around so that I wouldn't have to make as many changes to basisCurves.h

lib/mayaUsd/render/vp2RenderDelegate/draw_item.h
```diff
@@ -134,19 +163,27 @@ class HdVP2DrawItem final : public HdDrawItem
...
-
-    uint32_t    _renderItemUsage { kRegular }; //!< What is the render item created for
-    HdDirtyBits _dirtyBits {
-        HdChangeTracker::AllDirty
-    }; //!< Dirty bits to control data update of render item
+    /*
+        Data stored directly on the HdVP2DrawItem can be shared across all the MRenderItems
+        *this represents.
+
+        2021-01-29 TODO: Share vertex buffers within a draw item so that geom subset items
```
williamkrick, line 170: I did this, i'll fix the comment.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -66,14 +66,10 @@ using PrimvarBufferDataMap = std::unordered_map<TfToken, void*, TfToken::HashFun
 //!         (such commit task will be executed on main-thread)
 struct CommitState
 {
-    HdVP2DrawItem::RenderItemData& _drawItemData;
+    HdVP2DrawItem::RenderItemData& _renderItemData;
```
williamkrick, line 69: It is render item specific, not draw item specific, so rename the member.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -327,6 +323,74 @@ void setWantConsolidation(MHWRender::MRenderItem& renderItem, bool state)
     renderItem.setWantSubSceneConsolidation(state);
 #endif
 }
+
+PrimvarInfo* _getInfo(const PrimvarInfoMap& infoMap, const TfToken& token)
```
williamkrick, line 327: I eliminated the special data members to specific buffers. In their place I search for the buffers in the map when necessary.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -437,29 +814,80 @@ void HdVP2Mesh::Sync(
 #endif
     }
```
HdC-adsk, line 816: We will also need to apply the material consolidation update workaround to geom subset scenario. Here is the reference code:

https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/4d66c9264641c0f68a964f128eb14819a7d88558/pxr/imaging/plugin/hdRpr/mesh.cpp#L288

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -367,6 +428,322 @@ HdVP2Mesh::HdVP2Mesh(HdVP2RenderDelegate* delegate, const SdfPath& id, const Sdf
...
+            HdVP2RenderDelegate::sProfilerCategory,
+            MProfiler::kColorC_L2,
+            rprimId.asChar(),
+            "CommitBuffer"); // TODO: buffer usage so we know it is positions normals etc
+
+        buffer->commit(bufferData);
+    });
+}
+
+void HdVP2Mesh::_PrepareSharedVertexBuffers(
```
williamkrick, line 446: All this code was extracted from various different parts of HdVP2Mesh and gathered into one place. This is where we go from primvar data we've pulled from the scene delegate to buffer data ready to go into VP2.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -367,6 +428,322 @@ HdVP2Mesh::HdVP2Mesh(HdVP2RenderDelegate* delegate, const SdfPath& id, const Sdf
...
+                        = std::make_unique<PrimvarInfo>(
+                            PrimvarSource(normals, HdInterpolationVertex, true), nullptr);
+                } else {
+                    normalsInfo->_source.data = normals;
+                    normalsInfo->_source.interpolation = HdInterpolationVertex;
+                }
+            }
+        }
+
+        if (requireFlatNormals && (rprimDirtyBits & DirtyFlatNormals)) {
```
williamkrick, line 523: In the original code there are these checks for flat normals, but nothing can actually produce the flat normals.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -367,6 +428,322 @@ HdVP2Mesh::HdVP2Mesh(HdVP2RenderDelegate* delegate, const SdfPath& id, const Sdf
...
+                adjacencyComputation->Resolve();
+
+                // Only the points referenced by the topology are used to compute
+                // smooth normals.
+                VtValue normals(Hd_SmoothNormals::ComputeSmoothNormals(
+                    adjacency.get(),
+                    _points(_meshSharedData->_primvarInfo).size(),
+                    _points(_meshSharedData->_primvarInfo).cdata()));
+
+                if (!normalsInfo) {
```
williamkrick, line 512: The computed normals get stored in _primvarInfo, so later code can move the data to an MVertexBuffer without anything extra here.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -367,6 +428,322 @@ HdVP2Mesh::HdVP2Mesh(HdVP2RenderDelegate* delegate, const SdfPath& id, const Sdf
...
+            = _getInfo(_meshSharedData->_primvarInfo, HdVP2Tokens->displayColorAndOpacity);
+        if (!colorAndOpacityInfo) {
+            _meshSharedData->_primvarInfo[HdVP2Tokens->displayColorAndOpacity]
+                = std::make_unique<PrimvarInfo>(
+                    PrimvarSource(VtValue(), HdInterpolationConstant, true), nullptr);
+            colorAndOpacityInfo
+                = _getInfo(_meshSharedData->_primvarInfo, HdVP2Tokens->displayColorAndOpacity);
+        }
+
+        if (!colorAndOpacityInfo->_buffer) {
```
williamkrick, line 550: For color and opacity I get the buffer data & I set up the MVertexBuffer. This case is a little more complicated because color and opacity are two different primvar sources, but we merge them together into a single MVertexBuffer.

We could make the system a bit more powerful by storing more data with the primvarInfo. If we stored buffer descriptors like stride & offset then I could have two primvar sources that share an MVertexBuffer and write to different parts of it. Something to think about to clean things up further.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -367,6 +428,322 @@ HdVP2Mesh::HdVP2Mesh(HdVP2RenderDelegate* delegate, const SdfPath& id, const Sdf
...
+        HdInterpolation colorInterp = HdInterpolationConstant;
+        HdInterpolation alphaInterp = HdInterpolationConstant;
+        VtVec3fArray    colorArray;
+        VtFloatArray    alphaArray;
+
+        _getColorData(_meshSharedData->_primvarInfo, colorArray, colorInterp);
+        _getOpacityData(_meshSharedData->_primvarInfo, alphaArray, alphaInterp);
+
+        PrimvarInfo* colorAndOpacityInfo
+            = _getInfo(_meshSharedData->_primvarInfo, HdVP2Tokens->displayColorAndOpacity);
```
williamkrick, line 541: This is a new token I created. If we took the ideas from my comment directly below this one, we wouldn't need this token anymore.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -871,6 +1266,86 @@ void HdVP2Mesh::_InitRepr(const TfToken& reprToken, HdDirtyBits* dirtyBits)
     }
 }
 
+void HdVP2Mesh::_CreateSmoothHullRenderItems(HdVP2DrawItem& drawItem)
+{
+    // 2021-01-29: Changing topology is not tested
```
williamkrick, line 1271: Do we have workflows where topology changes?

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
williamkrick: Okay so I need to call SubscribeForMaterialUpdates & UnsubscribeForMaterialUpdates in reasonable places? I will figure out where to add it.

JGamache-autodesk (approved):

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
HdC-adsk: Ok. And I assume there won't be a bigger difference which would make it more difficult for us to unify the Rprim logic (such as a base class for HdVP2Rprim) after this change?

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -680,6 +676,23 @@ void HdVP2BasisCurves::Sync(
...
+    void*                           bufferData) const
+{
+    const MString& rprimId = _rprimId;
+
+    _delegate->GetVP2ResourceRegistry().EnqueueCommit([buffer, bufferData, rprimId]() {
+        MProfilingScope profilingScope(
+            HdVP2RenderDelegate::sProfilerCategory,
+            MProfiler::kColorC_L2,
+            rprimId.asChar(),
+            "CommitBuffer"); // TODO: buffer usage so we know it is positions normals etc
```
HdC-adsk, line 690: We have had many code in VP2RenderDelegate that swaps the order of the two string arguments for `MProfilingScope`: `eventName` vs. `description`. But I've recently started to write code with the correct order without changing other existing codes. Maybe we can all start to do the same to avoid inconsistency. Let me know how you think.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
```diff
@@ -54,4 +51,20 @@ HdVP2DrawItem::~HdVP2DrawItem()
     }
 }
```
HdC-adsk, line 52: Do we need to update the dtor to release render items for all geom subsets?

lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
```diff
@@ -54,4 +51,20 @@ HdVP2DrawItem::~HdVP2DrawItem()
...
 
+void HdVP2DrawItem::AddRenderItem(MHWRender::MRenderItem* item, const HdGeomSubset* geomSubset)
+{
+    _renderItems.push_back(RenderItemData());
+    RenderItemData& renderItemData = _renderItems.back();
+
+    renderItemData._renderItem = item;
+    renderItemData._renderItemName = _drawItemName;
+    if (geomSubset) {
+        renderItemData._geomSubset = *geomSubset;
```
HdC-adsk, line 62: Is there a reference-counting approach to hold the geomSubset object which can avoid the assignment?

lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
```diff
@@ -54,4 +51,20 @@ HdVP2DrawItem::~HdVP2DrawItem()
     }
 }
 
+void HdVP2DrawItem::AddRenderItem(MHWRender::MRenderItem* item, const HdGeomSubset* geomSubset)
+{
+    _renderItems.push_back(RenderItemData());
```
HdC-adsk, line 56: We can use emplace_back instead.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
```diff
@@ -54,4 +51,20 @@ HdVP2DrawItem::~HdVP2DrawItem()
...
+    RenderItemData& renderItemData = _renderItems.back();
+
+    renderItemData._renderItem = item;
+    renderItemData._renderItemName = _drawItemName;
+    if (geomSubset) {
+        renderItemData._geomSubset = *geomSubset;
+        renderItemData._renderItemName += geomSubset->id.GetString().c_str();
```
HdC-adsk, line 63: We are combining two SdfPaths in a string. Do we need a separator to make it readable and easier for debugging?

HdC-adsk: FWIW, GetText() or data() serves the same purpose of GetString().c_str().

HdC-adsk: Did you test selection on geom subsets? We rely on the render item name to find the index path so I think we need to update the selection interpretation code in `ProxyRenderDelegate::getInstancedSelectionPath` so we can extract the index path from the new name properly?

lib/mayaUsd/render/vp2RenderDelegate/draw_item.h
```diff
@@ -64,8 +64,13 @@ class HdVP2DrawItem final : public HdDrawItem
...
         bool _enabled { true };
 
+        //! Whether or not the render item is transparent
+        bool _transparent { false };
+
         //! Primitive type of the render item
-        MHWRender::MGeometry::Primitive _primitiveType { MHWRender::MGeometry::kInvalidPrimitive };
+        MHWRender::MGeometry::Primitive _primitiveType {
+            MHWRender::MGeometry::kInvalidPrimitive
+        }; // TODO: this is stored on the MRenderItem, do we need it here too?
```
HdC-adsk, line 73: Yea, maybe we can remove the duplicate data here because we have the render item pointer already which we can use to query the state and test whether it needs to be updated.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.h
```diff
@@ -38,19 +39,18 @@ class HdVP2RenderDelegate;
...
-        //! Render item primvar buffers - use when updating data
-        PrimvarBufferMap _primvarBuffers;
+        //! Unique name for easier debugging and profiling.
+        MString _renderItemName;
+        //! Pointer of the render item for fast access. No ownership is held.
+        MHWRender::MRenderItem* _renderItem { nullptr };
+
+        //! The geom subset this render item represents. _geomSubset.id is StdPath::EmptyPath() if
+        //! there is no geom subset.
+        HdGeomSubset _geomSubset;
```
HdC-adsk, line 52: Do we have to maintain another instance here? If the geom subset is stored in the topology, is it safe for us to just store a pointer here?

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
HdC-adsk: This has been fixed by new commits.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
HdC-adsk: I think it should be fine, but it would be nice to confirm whether it can impact performance on animated assets.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -327,6 +323,75 @@ void setWantConsolidation(MHWRender::MRenderItem& renderItem, bool state)
...
+        infoMap[HdTokens->displayOpacity] = std::make_unique<PrimvarInfo>(
+            PrimvarSource(VtValue(opacityArray), interpolation, PrimvarSource::CPUCompute),
+            nullptr);
+    }
+}
+
+//! Access the points
+VtVec3fArray _points(PrimvarInfoMap& infoMap)
+{
+    if (PrimvarInfo* info = _getInfo(infoMap, HdTokens->points)) {
```
HdC-adsk, line 386: Looks like we never expect the points to be unavailable. Maybe we should add a TF_VERIFY here.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
HdC-adsk: I commented about the swapping of the two string arguments in basisCurves.cpp. It is the same here.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -367,6 +429,324 @@ HdVP2Mesh::HdVP2Mesh(HdVP2RenderDelegate* delegate, const SdfPath& id, const Sdf
...
+void HdVP2Mesh::_CommitMVertexBuffer(MHWRender::MVertexBuffer* const buffer, void* bufferData) const
+{
+    const MString& rprimId = _rprimId;
+
+    _delegate->GetVP2ResourceRegistry().EnqueueCommit([buffer, bufferData, rprimId]() {
+        MProfilingScope profilingScope(
+            HdVP2RenderDelegate::sProfilerCategory,
+            MProfiler::kColorC_L2,
+            rprimId.asChar(),
+            "CommitBuffer"); // TODO: buffer usage so we know it is positions normals etc
```
HdC-adsk, line 441: I commented about the swapping of the two string arguments in basisCurves.cpp. It is the same here.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -404,7 +786,46 @@ void HdVP2Mesh::Sync(
...
+                        geomSubset.materialId);
+                HdVP2Material* material = static_cast<HdVP2Material*>(
+                    renderIndex.GetSprim(HdPrimTypeTokens->material, materialId));
+
+                if (material) {
+                    material->SubscribeForMaterialUpdates(id);
+                }
+            }
+        }
+#endif
```
HdC-adsk, line 827: Thanks for adding the workaround.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -437,29 +858,69 @@ void HdVP2Mesh::Sync(
...
+                }
+            }
+        };
+
+        // there is a chance that the geom subsets cover all the faces of the
+        // mesh and that the overall material id is unused. I don't figure that
+        // out until much later, so for now just accept that we might pull unnecessary
+        // primvars required by the overall material but not by any of the geom subset
+        // materials.
+        addRequiredPrimvars(GetMaterialId());
```
HdC-adsk, line 884: Shouldn't there be a loop to ask each material of the geom subset for the required primvars?

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -437,29 +858,69 @@ void HdVP2Mesh::Sync(
...
+
+        // For each relevant draw item, update dirty buffer sources.
+        bool hasHullGeomStyle = false;
+        for (size_t descIdx = 0; descIdx < reprDescs.size(); ++descIdx) {
+            const HdMeshReprDesc& desc = reprDescs[descIdx];
+            if (desc.geomStyle == HdMeshGeomStyleHull) {
+                hasHullGeomStyle = true;
+                break;
+            }
+        }
```
HdC-adsk, line 898: Can Hydra invalidate primvars when we switch from wireframe mode to shaded mode in the viewport? If not, we will have to always add primvar requirements for each bound material regardless of the state of `hasHullGeomStyle`.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
HdC-adsk: Currently with the consolidation update workaround, if we modify a shader attribute value, we will send DirtyMaterialId message to the Rprim side which will then propagate the dirty bits to topology.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -885,35 +1407,41 @@ void HdVP2Mesh::_UpdateRepr(HdSceneDelegate* sceneDelegate, const TfToken& reprT
...
+        if (desc.geomStyle == HdMeshGeomStyleHull) {
+            // it is possible we haven't created MRenderItems for this HdDrawItem yet.
+            // if there are no MRenderItems, create them.
+            if (drawItem->GetRenderItems().size() == 0) {
+                _CreateSmoothHullRenderItems(*drawItem);
+
+                auto* const param = static_cast<HdVP2RenderParam*>(_delegate->GetRenderParam());
+                MSubSceneContainer* subSceneContainer = param->GetContainer();
+                if (ARCH_UNLIKELY(!subSceneContainer))
+                    return;
```
HdC-adsk, line 1429: Please move these lines out of the loop.

HdC-adsk (changes requested):
I've left my requests for changes in relevant files.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
williamkrick: I think that this change doesn't make that refactor any more complicated.

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
williamkrick: Looking at MProfilingScope the 3rd argument is meant to be the eventName and the 4th argument is the description. So I think in this case I will switch "CommitBuffer" to be the eventName, and have the rprimId be the description? Is that the more correct way of doing it?

HdC-adsk: I believe so, the new order will make more sense for the text box when we hover the mouse over the event in Profiler.

@kxl-adsk What do you think?

lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
williamkrick: No there is no reference counting for [HdGeomSubset](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/imaging/hd/geomSubset.h). The structure itself is made up of two SdfPaths and a VtIntArray, both of which have efficient operator equals, so the cost of making the copy is low.

Making the copy also substantially simplifies the code in the absence of ref counting, because I don't have to keep on going back to the topology for sub geom values.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
williamkrick: Right now the output looks like:

`/Proxy_Fender_stratocaster_usdShape_0000023131BC8120/fender_stratocaster/StratocasterStand/DrawItem_000002314B068E60/fender_stratocaster/StratocasterStand/pxrUsdPreviewSurface10SG
`

The material path starts with a "/" so it ends up looking reasonable. We could add another character in between the object and material to make it clearer.

I thought selection was working but looking more carefully it is simply selecting the top level proxy shape, not the specific rprim. The code in proxyRenderDelegate is failing to convert the name into a valid rprim id as you have guessed.

I think what I will do is modify the render item name so that it starts with the rprimId, then include a separator which is not USD_UFE_SEPARATOR, then include extra information like instancing information or geom subset material id. That will make it simple to find that unique separator in getInstancedSelectionPath so we can select things correctly.

kxl-adsk: Make sure to run interactive tests to confirm everything works for the recently contributed point instancer selection.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.h
williamkrick: It is not 100% clear if a pointer is safe. The real question is when the topology gets dirtied how much of it changes? Are all the HdGeomSubset objects destroyed and re-created? When topology changes will we try to re-use existing HdGeomSubset items or destroy them all and start again from scratch?

I don't think we have workflows that modify topology, so all I have a questions and not answers.

The cost of copying the HdGeomSubset is low because it holds only SdfPath and VtIntArray objects, both of which have fast copy operations.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
williamkrick: If someone calls _points and they don't exist in the infoMap they'll get back an empty VtVec3fArray.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
williamkrick: I tried it out and you are correct. If I load the scene in wireframe mode & switch over to shaded then VP2 complains about missing buffers. I guess the primvars are not dirty and so we don't re-run this code. We'd have to always compare the old geometry requirements & new geometry requirements to see if they change and then force extraction.

I will switch it to always include all the requirements for now and we can choose to optimize wireframe by supporting fewer requirements later.

williamkrick: We add the base material id primvars already anyway, so this is not very different.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
williamkrick: I haven't seen a performance impact on the animation scenes we have.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
williamkrick: yes!

lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
```diff
@@ -46,21 +49,25 @@ HdVP2DrawItem::~HdVP2DrawItem()
...
+    _renderItems.emplace_back();
     RenderItemData& renderItemData = _renderItems.back();
 
     renderItemData._renderItem = item;
     renderItemData._renderItemName = _drawItemName;
     if (geomSubset) {
         renderItemData._geomSubset = *geomSubset;
+        renderItemData._renderItemName += std::string(1, VP2_RENDER_DELEGATE_SEPARATOR).c_str();
         renderItemData._renderItemName += geomSubset->id.GetString().c_str();
+        fprintf(stderr, "%s\n", renderItemData._renderItemName.asChar());
```
HdC-adsk, line 70: Please remove `fprintf`.

HdC-adsk (approved):
LGTM except one `fprintf` left in the code.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
williamkrick: point instancer selection is working as long as I have the changes for https://github.com/Autodesk/maya-usd/pull/1186 in place, so wait for that to get merged before this one.

williamkrick:
@wtelford found an issue with this one where disabled objects remain in the viewport, I will try to address that today before this gets merged.

williamkrick:
The new commit fixes the issue with objects not disappearing when the prim becomes deactivated.

lib/mayaUsd/nodes/proxyShapeBase.h
```diff
@@ -50,6 +50,7 @@ constexpr int  MAYA_UFE_RUNTIME_ID = 1;
 constexpr char MAYA_UFE_SEPARATOR = '|';
 constexpr int  USD_UFE_RUNTIME_ID = 2;
 constexpr char USD_UFE_SEPARATOR = '/';
+constexpr char VP2_RENDER_DELEGATE_SEPARATOR = ';';
```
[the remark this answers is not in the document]
kxl-adsk: Additionally, I wonder why it even is in the proxy shape? We only use it in VP2RenderDelegate - so my suggestion would be not only to move it outside of conditional compilation on UFE but as well move it to VP2RenderDelegate itself.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
kxl-adsk: I'm assuming this is now outdated information. Resolving this conversation.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.h
```diff
@@ -89,21 +94,46 @@ class HdVP2DrawItem final : public HdDrawItem
...
+    {
+        TF_VERIFY(_renderItems.size() == 1);
+        return _renderItems[0];
+    }
+
+    /*! \brief Get acces to render item data.
+     */
+    const RenderItemData& GetRenderItemData() const
+    {
+        TF_VERIFY(_renderItems.size() == 1);
```
kxl-adsk, line 114: I'm not sure I understand this check. We allow the addition of render items, so why verify that it's always 1?

kxl-adsk (changes requested):
Since the change was already reviewed by two reviewers, I only went over it at a high level.

Besides the requested change to address compilation error without UFE, two things I would suggest:
* start documenting in one place in each rprim type what are the commit tasks that we enqueue and dependencies between them. I'm assuming they currently have to run in a specific order, so let's capture it somewhere.
* consider adding in a separate PR testing coverage for GeomSubset support. It could be as a well good idea to add a test for normal generation, so we have something validating Viewport Computation.

lib/mayaUsd/render/vp2RenderDelegate/draw_item.h
williamkrick: The function and the other similar functions that access `RenderItemData` without specifying which render item assume that the `HdVP2DrawItem` is functioning the old way where only one render item was possible. I did this to minimize the the impact of this change on the basicCurve code. The TF_VERIFY is to protect against future changes misusing these functions and using single render item access in situations where there are multiple render items per draw item.

Eventually, we need to refactor basisCurve & mesh to unify how they work and all the code to manage render item access is going to be in the parent class.

kxl-adsk (approved):

PR #1159, opened 2021-02-09
perrauo/MAYA-109675/fix-crash-after-load-unload-and-loading-new

perrauo-adsk:
* Remove lingering Ufe::GlobalSelection callback observer when the renderOverride is destroyed.

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
@@ -473,18 +473,10 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext)
         drawContext.getRenderTargetSize(width, height);
 
         GfVec4d viewport(originX, originY, width, height);
-#if USD_VERSION_NUM >= 1910
         _taskController->SetFreeCameraMatrices(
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
```diff
@@ -264,6 +264,12 @@ MtohRenderOverride::MtohRenderOverride(const MtohRendererDescription& desc)
 
 MtohRenderOverride::~MtohRenderOverride()
 {
+    const Ufe::GlobalSelection::Ptr& ufeSelection = Ufe::GlobalSelection::get();
```
kxl-adsk, line 267: You will need to protect this change with the `WANT_UFE_BUILD` flag - see what we do when we register the observer:
https://github.com/Autodesk/maya-usd/blob/b733ccf51d7942ee73ecfab2a1b1ac68de4076a0/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp#L256

kxl-adsk (changes requested):

kxl-adsk (approved):

pmolodo (approved):
LGTM!

PR #1160, opened 2021-02-09
We will need to adopt new OCIOv2 in mtoh tests, but for now switch it off

kxl-adsk:
We will have to detect if OCIOv2 is running and set it up differently. The CM default settings have changed, including the default render color space (instead of "scene-linear Rec 709/sRGB"), and the former view transform "sRGB gamma" is now a pair "ACES 1.0 SDR-video (sRGB). The end result is to see the viewport is a little bit darker.

Also, the outputTransformName "Raw" was renamed to "Raw (sRGB)".

Above will require changes, including new base images. For now, we switch back to legacy mode.


mattyjams (approved):
Seems reasonable to me.

PR #1166, opened 2021-02-10
MAYA-108709 preliminary code refactor

pierrebai-adsk:
Code originally written by Pierre Tremblay to refactor common code related to retrieving local transforms.

I slightly modified it since some code paths used prim + op, others already had the ops vector and iterator.

### DIFF
--- a/lib/mayaUsd/ufe/XformOpUtils.cpp
+++ b/lib/mayaUsd/ufe/XformOpUtils.cpp
@@ -1,0 +1,99 @@
+//
+// Copyright 2020 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include "XformOpUtils.h"
+
+#include <pxr/usd/usdGeom/xformable.h>
+
+namespace MAYAUSD_NS_DEF {
+namespace ufe {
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/ufe/XformOpUtils.cpp
```diff
@@ -0,0 +1,92 @@
...
+    if (!UsdGeomXformable::GetLocalTransformation(&m, argOps, time)) {
+        TF_FATAL_ERROR("Local transformation computation failed.");
+    }
+
+    return m;
+}
+
+template <bool INCLUSIVE>
+GfMatrix4d
+computePrimLocalTransform(const UsdPrim& prim, const UsdGeomXformOp& op, const UsdTimeCode& time)
```
ppt-adsk, line 51: I would still call this computeLocalTransform(), and rely on function overloading to resolve to the right call.  To me, computePrimLocalTransform() has the meaning that we'll compute the local transform for the whole prim, which is not what we're doing.

pierrebai-adsk: The reason I gave it a different name is that the using declaration below cannot work for overloaded templates. And, even it the compiler did find them ambiguous, those using declaration would have to have different names since you cannot overload a using declaration.

OTOH, if you have a better suggestion for the name I can change it. I named it Prim because it acts on a prim. (I also don't understand the comment about the whole prim. The XForm are taken from the prim, are they not affecting the prim?)

pierrebai-adsk: What could be done is make the names that are declared in the "using" statements be real functions, which could then be overloaded, and only declare the templates, with different names, inside the cpp file, hidden from view.

ppt-adsk: computeLocalTransform() in this case is computing the local transform up to (and optionally including) a given transform op in the prim's transform stack, not the local transform of the complete transform stack.

The template is an implementation detail to avoid code duplication, so as long as you find a way to keep that idea, I'm fine with whatever works.

ppt-adsk (approved):

PR #1168, opened 2021-02-10
Cleanup + simplify some of the selection logic in the AL plugin

pmolodo:
I noticed that some bits of the AL selection code can be simplified / removed, some of which led to cascades of other bits being removed. In particular:

- I removed the custom code for getting nearest hit to camera, and instead rely on built-in resolveNearestToCamera
- Removed the need for `*SelectionHelper`, since we no longer use the hits we tracked with it
- Removed the need for custom `HitInfo` struct, as we never we no longer used the instanceIndex info, just the world space position

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Engine.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Engine.cpp
@@ -54,7 +54,6 @@ bool Engine::TestIntersectionBatch(
     UsdImagingGLRenderParams params,
     const TfToken&           resolveMode,
     unsigned int             pickResolution,
-    PathTranslatorCallback   pathTranslator,
     HitBatch*                outHit)
 {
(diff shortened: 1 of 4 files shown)


fabal (approved):
Thanks for this @pmolodo :+1:

kxl-adsk:
@pmolodo Please check the PF log files. Looks like your change doesn't compile with USD 21.02 (and newer).

kxl-adsk (changes requested):
Just updating the pr status to reflect that more work is needed to fix compilation with supported version of USD:
```
    /home/S/jenkins/workspace/ecg-mayausd-branch-preflight-python3-linux/ecg-maya-usd/maya-usd/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeUI.cpp:342:58: error: ‘HdxPickTokens’ was not declared in this scope
         TfToken resolveMode = selectInfo.singleSelection() ? HdxPickTokens->resolveNearestToCamera
```

kxl-adsk (approved):

PR #1170, opened 2021-02-11
Fix bbox computation for purposes, add test.

ppt-adsk:


lib/mayaUsd/ufe/UsdObject3d.cpp
```diff
@@ -75,7 +75,9 @@ Ufe::BBox3d UsdObject3d::boundingBox() const
     // as the time.
 
     auto bbox = UsdGeomImageable(fPrim).ComputeUntransformedBound(
-        getTime(sceneItem()->path()), UsdGeomTokens->default_);
+        getTime(sceneItem()->path()), UsdGeomTokens->default_,
+        UsdGeomTokens->guide, UsdGeomTokens->proxy, UsdGeomTokens->render
```
kxl-adsk, line 79: On the proxy shape, we have toggles that control what will be rendered. I wonder if we shouldn't be respecting it here because purpose helps you manage complexity and including everything when computing bounding box may be simply too expensive.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -333,6 +334,54 @@ UsdTimeCode getTime(const Ufe::Path& path)
     return proxyShape->getTime();
 }
 
+TfTokenVector getProxyShapePurposes(const Ufe::Path& path)
```
ppt-adsk, line 337: Copy-pasted and adapted from getTime().  Arguably, getTime() itself has duplication from the stage map, and all calls to nameToDagPath() in the existing code base should be examined for performance and centralized.  This is out of scope for the current branch.

test/testUtils/testUtils.py
```diff
@@ -30,6 +30,10 @@ def assertVectorAlmostEqual(testCase, a, b, places=7):
     for va, vb in zip(a, b):
         testCase.assertAlmostEqual(va, vb, places)
 
+def assertVectorNotAlmostEqual(testCase, a, b, places=7):
```
ppt-adsk, line 33: Thought I would need this, ended up not using it, but left it in, as it's useful.

kxl-adsk (approved):

seando-adsk (approved):

PR #1171, opened 2021-02-11
Fix UFEv2 test failures with USD 20.02

kxl-adsk:


test/lib/ufe/testSceneItem.py
```diff
@@ -69,7 +71,11 @@ def testNodeTypes(self):
         # This node type should be the first item on the ancestor node type list
         # and it should have other items.
         ball35AncestorNodeTypes = ball35Item.ancestorNodeTypes()
-        self.assertEqual(ball35NodeType, ball35AncestorNodeTypes[0])
+        if Usd.GetVersion() > (0, 20, 2):
+            self.assertEqual(ball35NodeType, ball35AncestorNodeTypes[0])
+        else:
+            self.assertEqual("UsdGeomXform", ball35AncestorNodeTypes[0])
```
seando-adsk, line 77: Should this test be >= (0, 20, 5) to be consistent with the cpp:
https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/ufe/UsdSceneItem.cpp#L79

kxl-adsk: The end result is exactly the same, so I don't believe we need to make a change here.

PR #1173, opened 2021-02-11
Fixing interactive tests to run in python 3.

seando-adsk:


test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeLiveSurface.py
```diff
@@ -71,7 +71,7 @@ def setUp(self):
         # Get the viewport widget.
         self._view = OMUI.M3dView()
         OMUI.M3dView.getM3dViewFromModelPanel(self._panel, self._view)
-        self._viewWidget = wrapInstance(long(self._view.widget()), QWidget)
+        self._viewWidget = wrapInstance(int(self._view.widget()), QWidget)
```
seando-adsk, line 74: In py3 long no longer exists. This same change was made when we ported Maya to py3.

test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegatePointInstanceSelection.py
```diff
@@ -134,7 +134,7 @@ def _RunTest(self):
         newSelection.clear()
 
         # Select the back half of the instances.
-        for instanceIndex in range(self._numInstances / 2, self._numInstances):
+        for instanceIndex in range(self._numInstances // 2, self._numInstances):
```
seando-adsk, line 137: In py2, dividing two integer s uses integer division, also known as floor division (apply the floor after division). In py3 this was deprecated. So use the floor division operator "//" avail in python 2.2 and later.

mattyjams (approved):
Looks good to me! Thanks @seando-adsk!

seando-adsk:
@kxl-adsk I see no point in running the preflight as all these changes are in the interactive tests which aren't run during preflight. Krystian I tested this on my local build using py3 and the two render delegate tests are now passing. The two proxy shape ones still fail, but for other reasons which I haven't figured out yet.

Can you test this on your py2 setup and merge. I'll continue investigation on the two remaining test failures.

PR #1175, opened 2021-02-12
maintain UV connectivity across export and re-import round trips

mattyjams:
Some of our artists have discovered that the re-indexing and compression that we're doing when we export UV sets is actually messing with the connectivity information that Maya computes from the values and indices arrays. The result is that when they export Maya meshes to USD and then re-import them back into Maya, they're ending up with disconnected and hard to work with UV sets.

With the changes here, we now translate the Maya mesh's UV values and assignment indices more or less directly, and we no longer apply any compression. This offers the best chance of maintaining the UV set's connectivity across round-trips. The net result in USD is the same (i.e. the array of UVs produced by doing `primvar.ComputeFlattened()` is the same as before), but the array values for the UV set primvar and its accompanying indices attribute may be different.

The real meat of the changes are in 82ae536ae6617b306ed9f8199c52a27777f60b64 while the other commits are from clean-up, simplification, and testing.

### DIFF
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
@@ -30,6 +30,7 @@
 #include <pxr/base/tf/staticTokens.h>
 #include <pxr/base/tf/token.h>
 #include <pxr/base/vt/array.h>
+#include <pxr/pxr.h>
 #include <pxr/usd/usdGeom/mesh.h>
 #include <pxr/usd/usdGeom/pointBased.h>
 #include <pxr/usd/usdGeom/primvar.h>
@@ -873,7 +874,12 @@ bool UsdMayaMeshWriteUtils::getMeshUVSetData(
     uvArray->clear();
     uvArray->reserve(static_cast<size_t>(uArray.length()));
     for (unsigned int uvId = 0u; uvId < uArray.length(); ++uvId) {
+#if PXR_VERSION >= 2011
         uvArray->emplace_back(uArray[uvId], vArray[uvId]);
+#else
+        GfVec2f value(uArray[uvId], vArray[uvId]);
+        uvArray->push_back(value);
+#endif
     }
 
     // Now iterate through all the face vertices and fill in the faceVarying
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
@@ -846,39 +846,46 @@ bool UsdMayaMeshWriteUtils::getMeshUVSetData(
     TfToken*       interpolation,
     VtIntArray*    assignmentIndices)
 {
-    MStatus status { MS::kSuccess };
-
-    // Sanity check first to make sure this UV set even has assigned values
-    // before we attempt to do anything with the data.
+    // Check first to make sure this UV set even has assigned values before we
+    // attempt to do anything with the data. We cannot directly use this data
+    // otherwise though since we need a uvId for every face vertex, and the
+    // returned uvIds MIntArray may be shorter than that if there are unmapped
+    // faces.
     MIntArray uvCounts, uvIds;
-    status = mesh.getAssignedUVs(uvCounts, uvIds, &uvSetName);
-    if (status != MS::kSuccess) {
-        return false;
-    }
-    if (uvCounts.length() == 0 || uvIds.length() == 0) {
+    MStatus   status = mesh.getAssignedUVs(uvCounts, uvIds, &uvSetName);
+    CHECK_MSTATUS_AND_RETURN(status, false);
+
+    if (uvCounts.length() == 0u || uvIds.length() == 0u) {
         return false;
     }
 
-    // using itFV.getUV() does not always give us the right answer, so
-    // instead, we have to use itFV.getUVIndex() and use that to index into the
-    // UV set.
+    // Transfer the UV values directly to USD, in the same order as they are in
+    // the Maya mesh.
     MFloatArray uArray;
     MFloatArray vArray;
-    mesh.getUVs(uArray, vArray, &uvSetName);
+    status = mesh.getUVs(uArray, vArray, &uvSetName);
+    CHECK_MSTATUS_AND_RETURN(status, false);
+
     if (uArray.length() != vArray.length()) {
         return false;
     }
 
-    // We'll populate the assignment indices for every face vertex, but we'll
-    // only push values into the data if the face vertex has a value. All face
-    // vertices are initially unassigned/unauthored.
-    const unsigned int numFaceVertices = mesh.numFaceVertices(&status);
     uvArray->clear();
-    assignmentIndices->assign((size_t)numFaceVertices, -1);
+    uvArray->reserve(static_cast<size_t>(uArray.length()));
+    for (unsigned int uvId = 0u; uvId < uArray.length(); ++uvId) {
+        uvArray->emplace_back(uArray[uvId], vArray[uvId]);
+    }
+
+    // Now iterate through all the face vertices and fill in the faceVarying
+    // assignmentIndices array, again in the same order as in the Maya mesh.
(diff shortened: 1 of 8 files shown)


test/lib/usd/translators/UsdExportUVSetsFloatTest/UsdExportUVSetsTest_Float.ma
```diff
@@ -436,12 +352,10 @@ createNode mesh -n "SharedFacesCubeShape" -p "SharedFacesCube";
...
 	setAttr ".uvst[1].uvsn" -type "string" "PairedFacesSet";
-	setAttr -s 23 ".uvst[1].uvsp[0:22]" -type "float2" 0 0 0.5 0 0.5 0.5
-		 0 0.5 0.5 0.5 1 0.5 1 1 0.5 1 0 0 0.5 0 0.5 0.5 0 0.5 0.5 0.5 1 0.5 1 1 0.5 1 0 0
-		 0.5 0 0.5 0.5 0 0.5 0.5 0.5 1 0.5 1 1;
+	setAttr -s 7 ".uvst[1].uvsp[0:6]" -type "float2" 0 0 0.5 0 0.5 0.5
+		 0 0.5 1 0.5 1 1 0.5 1;
 	setAttr ".uvst[2].uvsn" -type "string" "AllFacesSharedSet";
-	setAttr -s 24 ".uvst[2].uvsp[0:23]" -type "float2" 0 0 1 0 1 1 0 1 0
-		 0 1 0 1 1 0 1 0 0 1 0 1 1 0 1 0 0 1 0 1 1 0 1 0 0 1 0 1 1 0 1 0 0 1 0 1 1 0 1;
+	setAttr -s 4 ".uvst[2].uvsp[0:3]" -type "float2" 0 0 1 0 1 1 0 1;
```
mattyjams, line 358: I would expect that these changes and the indexing changes just below would result in shorter arrays of UV values, but that doesn't seem to be the case. See more in testUsdExportUVSets.py below.

test/lib/usd/translators/testUsdExportUVSets.py
```diff
@@ -84,6 +87,51 @@ def setUpClass(cls):
         cmds.select("box.map[0:299]", r=True)
         cmds.polyEditUV(u=1.0, v=1.0)
 
+        # XXX: Although the UV sets on the "SharedFacesCubeShape" are stored in
+        # the Maya scene with a minimal number of UV values and UV shells, they
+        # seem to be expanded when the file is opened such that we end up with
+        # a UV value per face vertex rather than these smaller arrays of UV
+        # values with only the indices being per face vertex. As a result, we
+        # have to reassign the UVs just before we export.
```
mattyjams, line 95: Does this ring any bells with anyone at Autodesk?

I attempted to author the UVs in the way I'd like to see them exported in 25e721b14927af1f8ae9528fb9ac793e1fa24d9b, but found that I needed to explicitly re-author them after opening the file in the test.

If you were to open the Maya scene file, run the `clear()`, `setUVs`(), and `assignUVs()` commands below and then save the file, there is no change to the mesh in the Maya scene file.

Any thoughts?

test/lib/usd/translators/testUsdExportUVSets.py
```diff
@@ -374,7 +363,8 @@ def testExportUvVersusUvIndexFromIterator(self):
         else:
             stPrimvar = brokenBoxMesh.GetPrimvar("map1").ComputeFlattened()
 
-        self.assertEqual(stPrimvar[0], Gf.Vec2f(1.0, 1.0))
+        self.assertEqual(stPrimvar[0], Gf.Vec2f(1.0, 2.0))
+
```
mattyjams, line 367: This UV set was previously getting compressed to `vertex` interpolation. With the compression now removed, it remains as `faceVarying`, and the first UV in the flattened array is now a different value.

test/lib/usd/translators/UsdExportUVSetsFloatTest/UsdExportUVSetsTest_Float.ma
JGamache-autodesk: From my experiments in Maya it looks like you should have removed some of the "mu" lines to simulate faces assigne to only one UV set.
- Create grid
- Copy map1 uv set
- Select and delete UVs in the UV editor
- Freeze history
When saved, a face assigned to 2 UV sets will look like:
```		f 4 128 131 -150 -130
		mu 0 4 67 68 79 78
		mu 1 4 1 2 8 7
```
While a single UV set face will be:
```		f 4 123 125 -145 -125
		mu 0 4 64 65 76 75
```

JGamache-autodesk: What I see from experimenting in Maya is that there is one `mu` row per UV assigned to the face. The first number in a `mu` row being the UV set index. You might want to edit in Maya, freeze the edits, and then replicate the deleted rows you see in the edited file.

JGamache-autodesk (approved):
Code is good. Might want to revisit the Maya file used in the unit test to see if the contents correspond to what Maya would save for partial UV sets.

test/lib/usd/translators/testUsdExportUVSets.py
JGamache-autodesk: Might be a missing freeze history. Are you seeing extra operators added to the mesh construction history in the save file?

mattyjams: Thanks @JGamache-autodesk. There were some `polyMapDel` nodes in the Maya scene file, but those applied to the `OneMissingFaceCubeShape` and `OneAssignedFaceCubeShape`. I deleted all history in the scene and that cleaned up those shapes, so I pushed that in b94de4fe8de195b963dbdedc6879dcbacaf3d7f3, but it unfortunately didn't seem to affect `SharedFacesCubeShape`, so I couldn't take this workaround out of the test script.

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -846,39 +846,46 @@ bool UsdMayaMeshWriteUtils::getMeshUVSetData(
...
 
-    // We'll populate the assignment indices for every face vertex, but we'll
-    // only push values into the data if the face vertex has a value. All face
-    // vertices are initially unassigned/unauthored.
-    const unsigned int numFaceVertices = mesh.numFaceVertices(&status);
     uvArray->clear();
-    assignmentIndices->assign((size_t)numFaceVertices, -1);
+    uvArray->reserve(static_cast<size_t>(uArray.length()));
+    for (unsigned int uvId = 0u; uvId < uArray.length(); ++uvId) {
+        uvArray->emplace_back(uArray[uvId], vArray[uvId]);
```
kxl-adsk, line 876: Looks like this doesn't compile with older versions of USD. Here is an error with 20.02
```
E:\maya-usd_pr\lib\mayaUsd\fileio\utils\meshWriteUtils.cpp(876): error C2039: 'emplace_back': is not a member of 'pxrInternal_v0_20__pxrReserved__::VtArray<pxrInternal_v0_20__pxrReserved__::GfVec2f>'
```

kxl-adsk (changes requested):
Please fix the compilation error with older versions of USD (20.02)

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
mattyjams: Ah, sorry about that. I failed to notice that `VtArray::emplace_back()`is only available from 20.11 on. Should be addressed by 6780fda93b047a6eb91ead50cbe2ca05c638bcb9.

I couldn't go as far back as 20.02, but I did verify it against 20.05 where this was also problematic.

kxl-adsk (approved):

kxl-adsk:
@mattyjams Thank you. PF failures are unrelated to the change, this is now ready for merge.

PR #1179, opened 2021-02-15
Support duplicate blendshape target names

ysiewappl:
Hi:

This PR fixes some bugs introduced with the newly-added blendshape export functionality, related to exporting multiple targets across multiple meshes.

The following are of significance in this PR:

- Previously, exporting multiple targets across multiple meshes resulted in some targets not having their animation written out at all.

- Exporting targets that share the same name across multiple meshes is now supported; we now namespace each target name (while leaving the prim name untouched to act as a "nice" name) to ensure that there is no risk of collision.

- We also now write out the default weights for blendshape targets, if they exist. This (unfortunately) only applies when animation export is enabled for now.

### DIFF
--- a/lib/mayaUsd/fileio/writeJobContext.h
+++ b/lib/mayaUsd/fileio/writeJobContext.h
@@ -135,7 +135,6 @@ public:
         const VtVec3fArray& bbox,
         const UsdTimeCode&  timeSample);
 
-
     /// Used to cache the animated blend shape weight plugs that need to be sampled per-frame.
     /// This is cached at the writeJob-level because the state needs to persist across instances of
     /// the meshWriter->Write() function.
--- a/lib/mayaUsd/fileio/writeJobContext.h
+++ b/lib/mayaUsd/fileio/writeJobContext.h
@@ -135,11 +135,6 @@ public:
         const VtVec3fArray& bbox,
         const UsdTimeCode&  timeSample);
 
-    /// Used to cache the animated blend shape weight plugs that need to be sampled per-frame.
-    /// This is cached at the writeJob-level because the state needs to persist across instances of
-    /// the meshWriter->Write() function.
-    MPlugArray mBlendShapesAnimWeightPlugs;
-
 protected:
     /// Opens the stage with the given \p filename for writing.
     /// If \p append is \c true, the file must already exist.
--- a/lib/mayaUsd/fileio/writeJobContext.h
+++ b/lib/mayaUsd/fileio/writeJobContext.h
@@ -138,7 +138,6 @@ public:
     /// Used to cache the animated blend shape weight plugs that need to be sampled per-frame.
     /// This is cached at the writeJob-level because the state needs to persist across instances of
     /// the meshWriter->Write() function.
-    MAYAUSD_CORE_PUBLIC
     MPlugArray mBlendShapesAnimWeightPlugs;
 
 protected:
--- a/lib/mayaUsd/fileio/writeJobContext.h
+++ b/lib/mayaUsd/fileio/writeJobContext.h
@@ -27,6 +27,7 @@
 #include <maya/MDagPath.h>
 #include <maya/MFnDependencyNode.h>
 #include <maya/MObjectHandle.h>
+#include <maya/MPlugArray.h>
 
 #include <memory>
 #include <vector>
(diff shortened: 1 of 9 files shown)


lib/mayaUsd/fileio/writeJobContext.h
```diff
@@ -134,6 +135,12 @@ class UsdMayaWriteJobContext
         const VtVec3fArray& bbox,
         const UsdTimeCode&  timeSample);
 
+    /// Used to cache the animated blend shape weight plugs that need to be sampled per-frame.
+    /// This is cached at the writeJob-level because the state needs to persist across instances of
+    /// the meshWriter->Write() function.
+    MAYAUSD_CORE_PUBLIC
+    MPlugArray mBlendShapesAnimWeightPlugs;
```
kxl-adsk, line 142: It's rather surprising to see translator-specific data in this generic context. What do others think? Should we make it easier to attach any blind data to the context instead of hardcoding something specific each time a translator needs it?

ysiewappl: Yes; I foresee a lot of future needs for this where a translator will need to retain state between invocations or similar, especially for certain types of Maya nodes.

williamkrick (approved):

HamedSabri-adsk (changes requested):
@ysiewappl  I am trying to build this PR on my Windows machine and getting C2071 error:

`maya-usd/lib/mayaUsd/fileio/writeJobContext.h(163): error C2071: 'pxrInternal_v0_21__pxrReserved__::UsdMayaWriteJobContext::mBlendShapesAnimWeightPlugs': illegal storage class`

ysiewappl:
> @ysiewappl I am trying to build this PR on my Windows machine and getting C2071 error:
> 
> `maya-usd/lib/mayaUsd/fileio/writeJobContext.h(163): error C2071: 'pxrInternal_v0_21__pxrReserved__::UsdMayaWriteJobContext::mBlendShapesAnimWeightPlugs': illegal storage class`

Hi @HamedSabri-adsk: That's...odd. For one, the code is just:

```
    MAYAUSD_CORE_PUBLIC
    MPlugArray mBlendShapesAnimWeightPlugs;
```

Which on w32 should just expand to `__declspec(dllexport)`. Second...the code should exist on line 142, not 163. Did I miss something here that's causing these issues?

HamedSabri-adsk:
@ysiewappl I was able to build the PR by removing the "MAYAUSD_CORE_PUBLIC" before your public member variable. 

I didn't have had time to investigate why VS compiler is not happy.  I also don't see anything that jumps up under "Selective Member Import/Export" criteria:

https://docs.microsoft.com/en-us/cpp/cpp/using-dllimport-and-dllexport-in-cpp-classes?view=msvc-160

ysiewappl:
> @ysiewappl I was able to build the PR by removing the "MAYAUSD_CORE_PUBLIC" before your public member variable.
> 
> I haven't had time to investigate why VS compiler is not happy. I also don't see anything that jumps up under "Selective Member Import/Export" criteria:
> 
> https://docs.microsoft.com/en-us/cpp/cpp/using-dllimport-and-dllexport-in-cpp-classes?view=msvc-160

Ah, I think I know what the issue is; I haven't defined the specific symbol. Clang doesn't seem to treat this as an error, oddly enough; I think MSVC is making the right call on this one.

> If you define a member as dllexport but do not include it in the class definition, a compiler error is generated. You must define the member in the class header.

There is no reason why this member needs to be made an exportable symbol, so I've gone ahead and removed that macro prefix. Previously I had just added in there in keeping with the surrounding code.

lib/mayaUsd/fileio/writeJobContext.h
ysiewappl: @kxl-adsk Based on our discussion, I've gone ahead and made the change to move this into the meshWriter translator instead, along with updating the unit test as also discussed. Please let me know if there are other questions/concerns.

kxl-adsk: Sounds good. I re-requested the review from @williamkrick and dismissed the stale review since it was no longer applicable based on what we discussed. I also started a new PF.

lib/usd/translators/meshWriter.h
```diff
@@ -75,12 +74,12 @@ class PxrUsdTranslators_MeshWriter : public UsdMayaPrimWriter
     bool    writeBlendShapeAnimation(const UsdTimeCode& usdTime);
     bool    writeAnimatedMeshExtents(const MObject& deformedMesh, const UsdTimeCode& usdTime);
 
+    /// Used to cache the animated blend shape weight plugs that need to be sampled per-frame.
```
williamkrick, line 77: We should add the explanation here of why this needs to be a static member.

williamkrick (changes requested):
Looks good to me. Just a bit more explanation of why the static member is needed right where it is declared and I will approve it.

lib/usd/translators/meshWriter.h
ysiewappl: This request has been accommodated for in `2ab03a5`.

williamkrick (approved):

PR #1180, opened 2021-02-15
T bailp/108709/undo manager for transform

pierrebai-adsk:
- Implemented template undo/redo base class for transforms.
- Move common handling of the undo/redo state and USD undo block in a base class.
- Derived classes only need to override handleSet() and optionally set().
- The base class handles the weird transitions coming from UFE.
- Implement MayaXFormStack, matrix-op and common API to use new base class.


pierrebai-adsk:
Note: this merges into my other branch. Once that branch is merged in dev, this will automatically update to merge in dev.

lib/mayaUsd/ufe/UsdUndoableCommand.h
```diff
@@ -0,0 +1,197 @@
...
+#include <pxr/usd/usdGeom/xformOp.h>
+
+#include <ufe/baseUndoableCommands.h>
+#include <ufe/path.h>
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+namespace MAYAUSD_NS_DEF {
+namespace ufe {
+
```
pierrebai-adsk, line 32: This is the starting point for the refactor, the top-most new base class.

Its goals, as stated are:
- Implement the UFE undo/redo/execute protocol/virtual functions.
- Hide the peculiarity of the undo/redo protocol, for example that some commands actually call set() instead of redo(), or that undo(0 may be called before execute().
- Provide the simplest interface for sub-classes. Basically, only handleSet(0 and optionally set(0 are needed.

lib/mayaUsd/ufe/UsdUndoableCommand.h
```diff
@@ -0,0 +1,197 @@
...
+    void undo() override;
+    void redo() override;
+
+    UsdTimeCode readTime() const;
+    UsdTimeCode writeTime() const;
+
+protected:
+    // Update the new value that will be set by execute().
+    void setNewValue(const VtValue& v);
+
```
pierrebai-adsk, line 69: While the previous design of teh code used State inner classes, and initially it was thought that not all sub-classes would need it, it turns out UFE does not always call the function you would expect and sometimes makes extra call you would also not expect.

So the states were simplified to 3, and mostly it only switches between Done and Undone, really. So between all classes needing to be isolated from the UFE details and the small number of states, it seemed simplir to only have a enum state with a very few if in undo/redo/execute.

lib/mayaUsd/ufe/UsdTransform3dCommonAPI.cpp
```diff
@@ -24,167 +25,160 @@ namespace ufe {
...
 
-private:
-    const UsdTimeCode     _time;
+protected:
     UsdGeomXformCommonAPI _commonAPI;
-    GfVec3d               _prevT, _t;
 };
 
-class UsdRotateUndoableCmd : public Ufe::RotateUndoableCommand
+class UsdTranslateUndoableCmd : public UsdTRSUndoableCmdBase
```
pierrebai-adsk, line 121: All other changes are basically the same refactor of replacing the old base class and code logic to use the new base classes that handle most of the work. All that is kept is implementing handleSet() and sometimes set().

PR #1181, opened 2021-02-16
replace usage of USD_VERSION_NUM with PXR_VERSION

mattyjams:
maya-usd currently recomputes the `PXR_VERSION` provided by core USD as `USD_VERSION_NUM`. With the minimum USD version now set at 20.02, all supported versions of core USD provide the `PXR_VERSION` symbol in `pxr/pxr.h`, so we no longer need to recompute its value.

All usages of `USD_VERSION_NUM` in C++ were replaced with `PXR_VERSION`, and `#include`s of `pxr/pxr.h` were added where it was used.

For versions of core USD before 21.05, we still need to recompute `USD_VERSION_NUM` as a CMake variable, but we now use the `PXR_VERSION` name for consistency with C++. Starting with core USD 21.05, `PXR_VERSION` along with `PXR_MAJOR_VERSION`, `PXR_MINOR_VERSION`, and `PXR_PATCH_VERSION` will all be set as CMake variables in `pxrConfig.cmake` (see https://github.com/PixarAnimationStudios/USD/commit/30df5e9c8af0a3b3afd1a393adf92363b4042ab9). We'll then no longer need to extract them from the `pxr/pxr.h` header.

### DIFF
--- a/cmake/modules/FindUSD.cmake
+++ b/cmake/modules/FindUSD.cmake
@@ -92,7 +92,14 @@ if(listlen GREATER 1)
     get_filename_component(PXR_USD_LOCATION "${USD_CONFIG_FILE}" DIRECTORY)
 endif()
 
-if(USD_INCLUDE_DIR AND EXISTS "${USD_INCLUDE_DIR}/pxr/pxr.h")
+if(DEFINED PXR_VERSION)
+    # Starting in core USD 21.05, pxrConfig.cmake provides the various USD
+    # version numbers as CMake variables, in which case PXR_VERSION should have
+    # been defined, along with the major, minor, and patch version numbers, so
+    # there is no need to extract them from the pxr/pxr.h header file anymore.
+    # The only thing we need to do is assemble the USD_VERSION version string.
+    set(USD_VERSION ${PXR_MAJOR_VERSION}.${PXR_MINOR_VERSION}.${PXR_PATCH_VERSION})
+elseif(USD_INCLUDE_DIR AND EXISTS "${USD_INCLUDE_DIR}/pxr/pxr.h")
     foreach(_usd_comp MAJOR MINOR PATCH)
         file(STRINGS
             "${USD_INCLUDE_DIR}/pxr/pxr.h"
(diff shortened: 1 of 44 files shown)


kxl-adsk (approved):

cmake/modules/FindUSD.cmake
```diff
@@ -90,7 +92,14 @@ if(listlen GREATER 1)
...
 endif()
 
-if(USD_INCLUDE_DIR AND EXISTS "${USD_INCLUDE_DIR}/pxr/pxr.h")
+if(DEFINED PXR_VERSION)
+    # Starting in core USD 21.05, pxrConfig.cmake provides the various USD
+    # version numbers as CMake variables, in which case PXR_VERSION should have
+    # been defined, along with the major, minor, and patch version numbers, so
+    # there is no need to extract them from the pxr/pxr.h header file anymore.
+    # The only thing we need to do is assemble the USD_VERSION version string.
+    set(USD_VERSION ${PXR_MAJOR_VERSION}.${PXR_MINOR_VERSION}.${PXR_PATCH_VERSION})
```
seando-adsk, line 101: Great.

lib/mayaUsd/CMakeLists.txt
```diff
@@ -33,7 +33,6 @@ find_package(Boost COMPONENTS filesystem system REQUIRED)
 # -----------------------------------------------------------------------------
 target_compile_definitions(${PROJECT_NAME}
     PUBLIC
-        USD_VERSION_NUM=${USD_VERSION_NUM}
```
seando-adsk, line 36: So I assume by all the change below that pxr.h has the _#define PXR_VERSION xxx_ set correctly in all the USD versions we support.

seando-adsk (approved):
Everything looks good to me. This makes it cleaner without that extra var.

lib/mayaUsd/CMakeLists.txt
mattyjams: Yup, `PXR_VERSION` goes back at least as far as 20.02:
https://github.com/PixarAnimationStudios/USD/blob/v20.02/pxr/pxr.h.in#L33

It first appeared as a usable value in 19.11 thanks to @pmolodo:
https://github.com/PixarAnimationStudios/USD/commit/f75a464309daf895a317b1142861d39e47a47904

PR #1184, opened 2021-02-16
MAYA-109928: Remove dependency on Boost chrono and date_time.

HamedSabri-adsk:
This PR removes the dependency on Boost chrono and date_time in AL plugin.

When Building USD, `--with-date_time` no longer needs to be passed.


kxl-adsk (approved):
The change makes sense to me, but since it impacts the AL plugin, let's give @fabal chance to comment on it.

fabal (approved):
Thanks @HamedSabri-adsk :+1:

kxl-adsk:
@HamedSabri-adsk please resolve merge conflicts and mark for merge when ready.

PR #1185, opened 2021-02-16
Remove the old formatTag option and anything that might still call it

fowlertADSK:
There was still one function call we were making that looked at the old binary/ascii option.  Fixing that was a one-liner, to just leave the function parameter as an empty string which gets the utils function to check the optionVar for us, but I went ahead and removed the member variable in the SaveLayerPathRow class that tracked the old option and any use of it.

All code should now use the option that is in the Layer Editor menu for ascii or binary when saving to a .usd file.

### DIFF
--- a/lib/usd/ui/layerEditor/saveLayersDialog.cpp
+++ b/lib/usd/ui/layerEditor/saveLayersDialog.cpp
@@ -103,9 +103,8 @@ public:
 
     QString absolutePath() const;
     QString pathToSaveAs() const;
-    QString usdFormatTag() const;
 
-    void setAbsolutePath(const std::string& path, const std::string& tag);
+    void setAbsolutePath(const std::string& path);
 
 protected:
     void onOpenBrowser();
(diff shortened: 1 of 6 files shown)


lib/usd/ui/layerEditor/saveLayersDialog.cpp
```diff
@@ -501,8 +495,8 @@ void SaveLayersDialog::onSaveAll()
                 auto qFileName = row->absolutePath();
                 auto sFileName = qFileName.toStdString();
 
-                auto newLayer = UsdMayaSerialization::saveAnonymousLayer(
-                    sdfLayer, sFileName, parentLayer, row->usdFormatTag().toStdString());
+                auto newLayer
+                    = UsdMayaSerialization::saveAnonymousLayer(sdfLayer, sFileName, parentLayer);
```
fowlertADSK, line 499: row->usdFormatTag() was still tracking the option, but without any UI to change it now it is stuck at binary.  This is really the only change needed to fix the defect, everything else is removing code that shouldn't be active now that we have the new optionVar + menu item.

seando-adsk (approved):
Looks good, thanks.

fowlertADSK:
Leaving a comment for QA:  I got all the way through fixing and code review and then realized that I'd typed in the wrong Jira number as part of the branch name.  The correct Jira number for this issue is MAYA-109929.

kxl-adsk:
@fowlertADSK FYI. PF failures look to be related to the network issue and unrelated to this change. This change is ready for merge.

PR #1186, opened 2021-02-16
When some instances are hidden there is a difference between the Scene

williamkrick:
Delegate instance Id and the Maya instance Id. Store a mapping buffer from the Scene Delegate Id to the Maya Id so we can correctly draw selection highlights.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -1130,11 +1130,22 @@ void HdVP2BasisCurves::_UpdateDrawItem(
                 // Assign with the index to the dormant wireframe color by default.
                 colorIndices.resize(instanceCount, 0);
 
+                // The active & lead selection information tells us the scene delegate instance
+                // Id. We need to convert that to the Vp2 instance id used in the colorIndices
+                // vector.
+                VtIntArray delegateToVp2Map = static_cast<HdVP2Instancer*>(instancer)
+                                                  ->ComputeInstanceIdToVp2InstanceIdArray(id);
+
                 // Assign with the index to the active selection highlight color.
                 if (auto state = drawScene.GetActiveSelectionState(id)) {
                     for (const auto& indexArray : state->instanceIndices) {
                         for (const auto index : indexArray) {
-                            colorIndices[index] = 1;
+                            if (index < delegateToVp2Map.size()) {
+                                int vp2InstanceIndex = delegateToVp2Map[index];
+                                if (vp2InstanceIndex >= 0) {
+                                    colorIndices[vp2InstanceIndex] = 1;
+                                }
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1488,11 +1488,24 @@ void HdVP2Mesh::_UpdateDrawItem(
...
+
+            bool visibleInstance = false;
+
             // Assign with the index to the active selection highlight color.
             if (const auto state = drawScene.GetActiveSelectionState(id)) {
                 for (const auto& indexArray : state->instanceIndices) {
                     for (const auto index : indexArray) {
-                        colorIndices[index] = 1;
+                        if (index < delegateToVp2Map.size()) {
+                            int vp2InstanceIndex = delegateToVp2Map[index];
```
williamkrick, line 1503: We need to convert the scene delegate instance id into the Maya instance id to know which instance to draw.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1501,25 +1514,35 @@ void HdVP2Mesh::_UpdateDrawItem(
...
-                const MColor& color = colors[colorIndex];
-                for (unsigned int j = 0; j < kNumColorChannels; j++) {
-                    stateToCommit._instanceColors.append(color[j]);
+                    const MColor& color = colors[colorIndex];
+                    for (unsigned int j = 0; j < kNumColorChannels; j++) {
+                        stateToCommit._instanceColors.append(color[j]);
+                    }
                 }
+            } else {
+                instancerWithNoInstances = true;
```
williamkrick, line 1545: This is also necessary to not draw something at the origin when only hidden instances are selected.

HdC-adsk:
I guess we also have to mirror the change to basisCurves?

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -1130,11 +1130,22 @@ void HdVP2BasisCurves::_UpdateDrawItem(
                 // Assign with the index to the dormant wireframe color by default.
                 colorIndices.resize(instanceCount, 0);
 
+                // The active & lead selection information tells us the scene delegate instance
+                // Id. We need to convert that to the Vp2 instance id used in the colorIndices
+                // vector.
+                VtIntArray delegateToVp2Map = static_cast<HdVP2Instancer*>(instancer)
+                                                  ->ComputeInstanceIdToVp2InstanceIdArray(id);
+
```
williamkrick, line 1138: It is very similar to the mesh version. The only difference is I don't have to track is any instances are visible here. In the basis curve case I know some instances are visible because of the earlier check, and we re-use a single render item to draw the dormant curves as well as the lead & active items. The only thing we have to get right is which items do we change the color of, and that is what the new changes do.

HdC-adsk (approved):

kxl-adsk (changes requested):
Following tests are failing with this change
```
54 - testVP2RenderDelegatePointInstanceSelection (Failed)
55 - testVP2RenderDelegatePointInstancesPickMode (Failed)
```

williamkrick:
Logged https://github.com/PixarAnimationStudios/USD/issues/1461. Will revisit once we have an answer.

PR #1187, opened 2021-02-17
MAYA-109948 - Crash saving USD without opening Layer Editor

seando-adsk:
#### MAYA-109948 - Crash saving USD without opening Layer Editor
* Moved the MayaQtUtils init to the plugin.


fowlertADSK (approved):
LGTM

PR #1189, opened 2021-02-17
MAYA-107276 - As a user, on the stage AE template I'd like to use Pri…

spinell-adsk:
…m Path

-Add Prim Path field in attribute editor
-Use the prime path to render with VP2RenderDelegate
-Change prim path tooltip


plugin/adsk/scripts/AEmayaUsdProxyShapeTemplate.mel
```diff
@@ -39,7 +39,7 @@ global proc AEmayaUsdProxyShapeTemplate( string $nodeName )
         editorTemplate -suppress "outStageCacheId";
 
         // MAYA-109438 - temp suppress these non-functional attributes.
-        editorTemplate -suppress "primPath";
+        //editorTemplate -suppress "primPath";
```
seando-adsk, line 42: Let's just remove this line.

seando-adsk (approved):

spinell-adsk:
Please do not merge yet. I found 2 issues that I need to fix.
The first one is the primPath feature does not work as expected when the stage is create in memory.
The second one is when using primPath, the transform matrix use for rendering is wrong.

spinell-adsk (commented):
@mattyjams I moved the dependency between  primPathAttr and inStageDataCachedAttr from ProxyShapeBase to ProxyShape
because we are currently have issue when the primPath changed. When we are re-computing the stage cache, if the stage is not found in a file (stage created in memory) we are creating a new in memory stage which override the current one.

Do you have any concern about this change ?

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -387,8 +387,6 @@ MStatus MayaUsdProxyShapeBase::initialize()
     retValue = attributeAffects(filePathAttr, outStageCacheIdAttr);
     CHECK_MSTATUS_AND_RETURN_IT(retValue);
 
-    retValue = attributeAffects(primPathAttr, inStageDataCachedAttr);
```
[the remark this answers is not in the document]
kxl-adsk: It's worth mentioning why this dependency doesn't belong to the base class - when computing inStageDataCachedAttr, there is nothing that reads primPathAttr unless we are calling into computeSessionLayer implementation of pxr proxy shape - https://github.com/Autodesk/maya-usd/blob/dev/plugin/pxr/maya/lib/usdMaya/proxyShape.cpp#L189

There is still the unwanted effect of clearing the stage on a change to the prim path with pxr proxy shape, but assembly workflows were not working with in-memory stages.

kxl-adsk (changes requested):
Mostly looks good to me. I think it would be worth saving on matrix multiplication when prim path is not the default.
Additionally, this is something we would like to have a regression test for and make sure we don't regress - please check https://github.com/Autodesk/maya-usd/blob/dev/test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegatePointInstanceSelection.py for an example of such a test.

mattyjams:
> @mattyjams I moved the dependency between primPathAttr and inStageDataCachedAttr from ProxyShapeBase to ProxyShape
> because we are currently have issue when the primPath changed. When we are re-computing the stage cache, if the stage is not found in a file (stage created in memory) we are creating a new in memory stage which override the current one.
> 
> Do you have any concern about this change ?

I *think* we should be fine with this change, since the dependency is still intact from the point of the `pxrUsdProxyShape`.

I'm hoping that the performance impact of computing the local transform in `_Sync()` in `pxrUsdMayaGL` will be relatively minimal, since our typical use case is to fill the `primPath` attribute with a root prim (e.g. `/Asset`) that has no transform.

lib/mayaUsd/render/pxrUsdMayaGL/usdProxyShapeAdapter.cpp
```diff
@@ -171,10 +171,23 @@ bool PxrMayaHdUsdProxyShapeAdapter::_Sync(
         _renderTags.push_back(HdRenderTagTokens->guide);
     }
 
+    // Update the root transform used to render by the delagate.
+    // USD considare that the root prim transform is always the Identity matrix so that means
+    // the root transform define the root prim transform. When the real stage root is used to
+    // render this is not a issue because the root transform will be the maya transform.
+    // The problem is when using a primPath as the root prim, we are losesing
+    // the prim path world transform. So we need to set the root transform as the world
+    // transform of the prim used for rendering.
```
mattyjams, line 180: Just a few typos in here:
considare -> considers
losesing -> losing

And looks like they're repeated in the render delegate below.

mattyjams (approved):
Seems reasonable to me, though I'd agree with @kxl-adsk that the transform computation can be avoided when `usdPrim()` yields the psuedoroot or an invalid prim.

kxl-adsk (approved):

lib/mayaUsd/render/pxrUsdMayaGL/usdProxyShapeAdapter.cpp
```diff
@@ -171,10 +171,26 @@ bool PxrMayaHdUsdProxyShapeAdapter::_Sync(
...
+    // render this is not a issue because the root transform will be the maya transform.
+    // The problem is when using a primPath as the root prim, we are losing
+    // the prim path world transform. So we need to set the root transform as the world
+    // transform of the prim used for rendering.
     MStatus       status;
     const MMatrix transform = GetDagPath().inclusiveMatrix(&status);
     if (status == MS::kSuccess) {
         _rootXform = GfMatrix4d(transform.matrix);
+
+        if (!usdProxyShape->usdPrim().GetPath().IsAbsoluteRootPath()) {
```
kxl-adsk, line 186: This doesn't compile with the minimum supported version of USD, i.e. 20.02.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -604,8 +603,24 @@ void ProxyRenderDelegate::_UpdateSceneDelegate()
...
+    // USD considers that the root prim transform is always the Identity matrix so that means
+    // the root transform define the root prim transform. When the real stage root is used to
+    // render this is not a issue because the root transform will be the maya transform.
+    // The problem is when using a primPath as the root prim, we are losing
+    // the prim path world transform. So we need to set the root transform as the world
+    // transform of the prim used for rendering.
+    const MMatrix inclusiveMatrix = _proxyShapeData->ProxyDagPath().inclusiveMatrix();
+    GfMatrix4d    transform(inclusiveMatrix.matrix);
+
+    if (!_proxyShapeData->ProxyShape()->usdPrim().GetPath().IsAbsoluteRootPath()) {
```
kxl-adsk, line 616: This doesn't compile with the minimum supported version of USD, i.e. 20.02.

lib/mayaUsd/render/pxrUsdMayaGL/usdProxyShapeAdapter.cpp
kxl-adsk:
```suggestion
        if (usdProxyShape->usdPrim().GetPath() != SdfPath::AbsoluteRootPath()) {
```

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
kxl-adsk:
```suggestion
    if (_proxyShapeData->ProxyShape()->usdPrim().GetPath() != SdfPath::AbsoluteRootPath()) {
```

kxl-adsk (changes requested):
We have to support older versions of USD. Please see my comments below with suggestions on how to fix compilation errors for all supported versions of USD.

kxl-adsk (approved):

PR #1192, opened 2021-02-18
al: notifyProxyShapeOfRedraw: fix for check if xform has changed

pmolodo:
Bugfix - in [line 2065](https://github.com/Autodesk/maya-usd/blob/21d410cf9cfcce4231b554cf715a5124f54c4b06/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/TransformationMatrix.cpp#L2065) we get the old transform, but then never use it... then when actually comparing the value, we [used to get the current value twice](https://github.com/Autodesk/maya-usd/pull/1192/files#diff-4918577650a36d85c51a468a604f5b99cf93dcd92521951e7dae78aeda422155L1722), and compare against each other (which is always equal).


kxl-adsk (approved):
LGTM! I will let @fabal review this change as well to make sure there isn't anything I missed.

fabal (approved):
Thanks @pmolodo :+1:

PR #1193, opened 2021-02-18
AL: fix proxy shape time-scalar export

pmolodo:
If exporting a ProxyShape as a reference, make sure we export the timeScalar settings we've changed.

NOTE - the interpretation of the timeScalar is OPPOSITE that as used by USD.  Not sure if this was intentional, but I invert the values on export, to make it work in USD as it did in Maya.


fabal (approved):
Thanks @pmolodo :+1:

PR #1194, opened 2021-02-18
new signal/slot to reset a stage/proxy in the Layer Editor

fowlertADSK:
I have some other changes to how we handle anonymous root layers, but this fixes an issue that showed up in a demo last week where a proxy that has a new stage set still has the old stage visible in the Layer Editor if it wasn't the visible proxy when the reset happens.

The easiest way to test this right now is to copy/paste in a path to a new .usd file for the filePath attribute of a proxy that is not being displayed in the Layer Editor.  Switching to that proxy after making the change to filePath will still show the old stage.  This defect can also be hit as part of File -> Save, but this change only deals with the Layer Editor.  Follow up changes are going to improve how we handle anonymous root layers as part of saving.

### DIFF
--- a/lib/usd/ui/layerEditor/mayaSessionState.cpp
+++ b/lib/usd/ui/layerEditor/mayaSessionState.cpp
@@ -156,9 +156,23 @@ void MayaSessionState::unregisterNotifications()
 void MayaSessionState::mayaUsdStageReset(const MayaUsdProxyStageSetNotice& notice)
 {
     auto shapePath = notice.GetShapePath();
+    auto stage = notice.GetStage();
     if (shapePath == _currentProxyShapePath) {
-        auto stage = notice.GetStage();
         QTimer::singleShot(0, this, [this, stage]() { setStage(stage); });
+    } else {
+        auto shapePath = notice.GetShapePath();
+        QTimer::singleShot(
+            0, this, [this, shapePath, stage]() { mayaUsdStageResetCBOnIdle(shapePath, stage); });
+    }
+}
+
+void MayaSessionState::mayaUsdStageResetCBOnIdle(
+    const std::string&            shapePath,
+    PXR_NS::UsdStageRefPtr const& stage)
+{
+    StageEntry entry;
+    if (getStageEntry(&entry, shapePath.c_str())) {
+        Q_EMIT stageResetSignal(entry._proxyShapePath, stage);
     }
 }
 
--- a/lib/usd/ui/layerEditor/stageSelectorWidget.cpp
+++ b/lib/usd/ui/layerEditor/stageSelectorWidget.cpp
@@ -25,9 +25,7 @@
 #include <QtWidgets/QHBoxLayout>
 #include <QtWidgets/QLabel>
 
-PXR_NAMESPACE_USING_DIRECTIVE
-
-Q_DECLARE_METATYPE(UsdStageRefPtr);
+Q_DECLARE_METATYPE(PXR_NS::UsdStageRefPtr);
 
 namespace UsdLayerEditor {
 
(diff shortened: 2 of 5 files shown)


lib/usd/ui/layerEditor/stageSelectorWidget.cpp
```diff
@@ -58,28 +61,29 @@ void StageSelectorWidget::setSessionState(SessionState* in_sessionState)
         &SessionState::stageListChangedSignal,
         this,
         &StageSelectorWidget::updateFromSessionState);
-    updateFromSessionState();
```
fowlertADSK, line 61: Looked like we had a copy/paste error and ended up with multiple calls to updateFromSessionState

lib/usd/ui/layerEditor/stageSelectorWidget.cpp
```diff
@@ -18,13 +18,16 @@
...
 
+#include <pxr/pxr.h>
 #include <pxr/usd/usd/common.h>
 
 #include <QtCore/QSignalBlocker>
 #include <QtWidgets/QHBoxLayout>
 #include <QtWidgets/QLabel>
 
-Q_DECLARE_METATYPE(PXR_NS::UsdStageRefPtr);
+PXR_NAMESPACE_USING_DIRECTIVE
```
fowlertADSK, line 28: I added the using directive and deleted all the explicit PXR_NS calls, seems to be the pattern we're following in most places.

lib/usd/ui/layerEditor/mayaSessionState.cpp
```diff
@@ -156,9 +156,23 @@ void MayaSessionState::unregisterNotifications()
 void MayaSessionState::mayaUsdStageReset(const MayaUsdProxyStageSetNotice& notice)
 {
     auto shapePath = notice.GetShapePath();
+    auto stage = notice.GetStage();
     if (shapePath == _currentProxyShapePath) {
```
fowlertADSK, line 160: Main problem was that we only ever attempted to modify the proxy that was currently being displayed, but the Layer Editor does have UsdStage ptr's for all the proxy's in its dropdown list so we do need to update that.

I did initially try to reuse some existing signals that request the entire list be rebuilt, but I was finding that it wasn't a great experience.  By resetting everything, the Layer Editor gets emptied and the entire UI rebuilt which ends up being pretty annoying even if it doesn't take all that long.

lib/usd/ui/layerEditor/stageSelectorWidget.cpp
seando-adsk: Huh, I thought internally we just talked about this and decided that we would stop using the using directive and always use PXR_NS.

seando-adsk (changes requested):
I'm not sure about the change to remove the PXR_NS

lib/usd/ui/layerEditor/stageSelectorWidget.cpp
fowlertADSK: I typed in that comment and posted the review and THEN the internal message was sent out.  I think that sentence was what prompted it.

fowlertADSK: I don't care either way.  I made the change after someone pointed out how in one of the files we had the using directive but then there was a mix of using PXR_NS or not and it wasn't consistent.  Here I just added it and removed because that's normally what I see in code, but if we want to go the other way then that's fine with me.  It does sound like we might make a sweep of the entire repo and get things consistent though, whatever we decide to do, so not sure if we necessarily need to go back and remove it all from this change.

seando-adsk: I would kind of prefer if we left this code alone for now. Once we make the decision about how we want to proceed via PXR_NS we can have a dedicated PR to fix that.

seando-adsk: So I'm suggesting you put back the PXR_NS to not confuse this PR with changes not related to the actual fix.

seando-adsk (approved):

PR #1195, opened 2021-02-19
[mtoh] if using the VP2 MPxSubSceneOverride, don't update if we're also using mtoh

pmolodo:
These apparently still get called even when using MtoH's RenderOverride;
however, since MtoH has it's own proxyShape handling, they're unneeded, but
were still populating their own renderIndex.  Skip this to save time + memory.

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/utils.h
+++ b/lib/mayaUsd/render/mayaToHydra/utils.h
@@ -54,7 +54,7 @@ using MtohRendererSettings
 inline bool IsMtohRenderOverrideName(const MString& overrideName)
 {
     // See if the override is an mtoh one - ie, it starts with the right prefix
-    
+
     // VS2017 msvc didn't accept this as a constexpr
     const auto prefixLen = strlen(MTOH_RENDER_OVERRIDE_PREFIX);
     if (overrideName.length() < prefixLen) {
--- a/lib/mayaUsd/render/mayaToHydra/utils.h
+++ b/lib/mayaUsd/render/mayaToHydra/utils.h
@@ -55,12 +55,6 @@ inline bool IsMtohRenderOverrideName(const MString& overrideName)
 {
     // See if the override is an mtoh one - ie, it starts with the right prefix
     constexpr auto prefixLen = strlen(MTOH_RENDER_OVERRIDE_PREFIX);
-
-    TF_WARN(
-        "%s: %s an mtoh renderer",
-        overrideName.asChar(),
-        overrideName.substring(0, prefixLen - 1) == MTOH_RENDER_OVERRIDE_PREFIX ? "is" : "is not");
-
     if (overrideName.length() < prefixLen) {
         return false;
     }
--- a/lib/mayaUsd/render/mayaToHydra/utils.h
+++ b/lib/mayaUsd/render/mayaToHydra/utils.h
@@ -50,7 +50,7 @@ using MtohRendererDescriptionVector = std::vector<MtohRendererDescription>;
 using MtohRendererSettings
     = std::unordered_map<TfToken, HdRenderSettingDescriptorList, TfToken::HashFunctor>;
 
-// Defining these in header so don't need to link to use... template is
+// Defining these in header so don't need to link to use
 inline bool IsMtohRenderOverrideName(const MString& overrideName)
 {
     // See if the override is an mtoh one - ie, it starts with the right prefix
--- a/lib/mayaUsd/render/mayaToHydra/utils.h
+++ b/lib/mayaUsd/render/mayaToHydra/utils.h
@@ -20,10 +20,15 @@
 #include <pxr/imaging/hd/renderDelegate.h>
 #include <pxr/pxr.h>
 
+#include <maya/MFrameContext.h>
+#include <maya/MString.h>
+
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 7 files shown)


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -348,11 +352,23 @@ bool ProxyRenderDelegate::enableUpdateForSelection() const { return true; }
...
+    std::cout << "ProxyRenderDelegate::update - BUILD_HDMAYA - checking if IsMtohRenderOverride"
+              << std::endl;
+
+    // If the current viewport renderer is an mtoh one, skip this update, as
+    // mtoh already has special handling for proxy shapes, and we don't want to
+    // build out a render index we don't need
+    if (IsMtohRenderOverride(frameContext)) {
+        return false;
+    }
+    std::cout << "ProxyRenderDelegate::requiresSpdate - was not mtoh" << std::endl;
```
HdC-adsk, line 370: Could you remove the prints at Line 361 and Line 370 which will get shown for every refresh? Or at least protect it with a `TF_DEBUG` option.

HdC-adsk: Or change to print some message when it is using mtoh render override (with TF_DEBUG condition).

lib/mayaUsd/render/mayaToHydra/utils.h
```diff
@@ -45,6 +50,31 @@ using MtohRendererDescriptionVector = std::vector<MtohRendererDescription>;
 using MtohRendererSettings
     = std::unordered_map<TfToken, HdRenderSettingDescriptorList, TfToken::HashFunctor>;
 
+// Defining these in header so don't need to link to use... template is
+inline bool IsMtohRenderOverrideName(const MString& overrideName)
+{
+    // See if the override is an mtoh one - ie, it starts with the right prefix
+    constexpr auto prefixLen = strlen(MTOH_RENDER_OVERRIDE_PREFIX);
+
+    TF_WARN(
```
HdC-adsk, line 59: Please remove the warning as it would print the message for every refresh.

lib/mayaUsd/render/mayaToHydra/utils.h
```diff
@@ -45,6 +50,31 @@ using MtohRendererDescriptionVector = std::vector<MtohRendererDescription>;
...
+        "%s: %s an mtoh renderer",
+        overrideName.asChar(),
+        overrideName.substring(0, prefixLen - 1) == MTOH_RENDER_OVERRIDE_PREFIX ? "is" : "is not");
+
+    if (overrideName.length() < prefixLen) {
+        return false;
+    }
```
HdC-adsk, line 66: If the view is not using any render override, do we return from this check?

HdC-adsk (changes requested):
Requests of change left in files.

HdC-adsk:
Can you test multiple views with one using mtoh and one using VP2? I think it should work but would like to confirm.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
pmolodo: Oops - yes, absolutely - accidentally had left some debug prints in.

Removed in https://github.com/Autodesk/maya-usd/pull/1195/commits/03890ff59a4c347bb5107445bb710e8f6c4eb01d

lib/mayaUsd/render/mayaToHydra/utils.h
pmolodo: Oops - yes, absolutely - accidentally had left some debug prints in.

Removed in https://github.com/Autodesk/maya-usd/pull/1195/commits/03890ff59a4c347bb5107445bb710e8f6c4eb01d

lib/mayaUsd/render/mayaToHydra/utils.h
pmolodo: Yes - `overrideName` is the empty string in that case

pmolodo:
Fixed those debug print statements - sorry!  Also, I confirmed that it works fine with multiple viewports, with a mix of mtoh/non-mtoh.

HdC-adsk (approved):

HdC-adsk:
Adding @HamedSabri-adsk to review the cmake change and @mattyjams to review the `pxrUsdMayaGL ` change.

HamedSabri-adsk (changes requested):
With `add_definitions` all the targets are now contaminated which is not good.  Use `target_compile_definitions` in targets that really need this def.

pmolodo:
> With `add_definitions` all the targets are now contaminated which is not good. Use `target_compile_definitions` in targets that really need this def.

Good point - should be fixed in https://github.com/Autodesk/maya-usd/pull/1195/commits/23b7735f030e6528668e3f015c1c706f746b4892

kxl-adsk:
@pmolodo Please fix the clang-format issue, so we can get the PF run.

pmolodo:
> @pmolodo Please fix the clang-format issue, so we can get the PF run.

Oops - sorry about that, should be fixed now.  Was hopping around branches to a point where I didn't have the clang-format-tools stuff merged in, and had turned off the pre-commit check... and forgot to re-enable it!

PR #1201, opened 2021-02-22
Fallback, multi-matrix parent absolute.

ppt-adsk:


lib/mayaUsd/ufe/UsdTransform3dSetObjectMatrix.cpp
```diff
@@ -0,0 +1,85 @@
...
+Ufe::Vector3d UsdTransform3dSetObjectMatrix::scale() const
+{
+    TF_CODING_ERROR("Illegal call to unimplemented UsdTransform3dSetObjectMatrix::scale()");
+    return Ufe::Vector3d(1, 1, 1);
+}
+
+Ufe::SetMatrix4dUndoableCommand::Ptr
+UsdTransform3dSetObjectMatrix::setMatrixCmd(const Ufe::Matrix4d& m)
+{
+    return _wrapped->setMatrixCmd(mw(m));
```
ppt-adsk, line 65: This is the key line in the entire review: use the wrapper UsdTransform3dSetObjectMatrix to compute an appropriate matrix when editing part of the transform op stack, i.e. a transform stack with a Maya fallback, or a single matrix op in a stack of multiple matrix ops.  It simply calls the mw() method to compute this matrix, using the inverse of the less-local (Ml) and inverse of the more-local (Mr) aggregate transformations.

lib/mayaUsd/ufe/UsdTransform3dSetObjectMatrix.h
```diff
@@ -0,0 +1,95 @@
...
+// for a fallback Maya transform stack, respectively.
+//
+// In both cases the complete transform stack will have additional transform
+// ops that define the complete 3D transformation for the whole object, while
+// only part of the complete transform stack (a single matrix and the fallback
+// Maya transform stack) is editable.
+//
+// Certain Maya operations (such as parent -absolute) require the capability to
+// edit the matrix transform of the whole object.  This class wraps a
+// UsdTransform3dMatrixOp or UsdTransform3dFallbackMayaXformStack and converts
```
ppt-adsk, line 41: For UsdTransform3dMatrixOp targeting a single matrix op in a multi-matrix stack, and unconditionally for UsdTransform3dFallbackMayaXformStack, use this new UsdTransform3dSetObjectMatrix object to set the local transform of the object (e.g. for parent -absolute).

lib/mayaUsd/ufe/wrapUtils.cpp
```diff
@@ -213,6 +225,7 @@ void wrapUtils()
     def("usdPathToUfePathSegment",
         usdPathToUfePathSegment,
         (arg("usdPath"), arg("instanceIndex") = UsdImagingDelegate::ALL_INSTANCES));
+    def("getTime", getTime);
```
ppt-adsk, line 228: Reading transform op values directly through the USD interface requires time (whereas the UFE interface uses proxy time).  Expose the time accessor in Python.

test/lib/ufe/testComboCmd.py
```diff
@@ -706,8 +706,15 @@ def testFallback(self):
         mayaSphereItem        = ufe.Hierarchy.createItem(mayaSpherePath)
         usdSphereItem         = ufe.Hierarchy.createItem(usdSpherePath)
         usdFallbackSphereItem = ufe.Hierarchy.createItem(usdFallbackSpherePath)
-        usdSphere3d         = ufe.Transform3d.transform3d(usdSphereItem)
-        usdFallbackSphere3d = ufe.Transform3d.transform3d(usdFallbackSphereItem)
+        # For scene items with fallback transform ops, the transform3d()
```
ppt-adsk, line 709: Fix latent bug exposed with new transform3d() implementation for Maya fallback stacks.  It is inappropriate to use transform3d() for reading and writing fallback transform ops; editTransform3d() must be used.

test/lib/ufe/testParentCmd.py
```diff
@@ -386,6 +386,277 @@ def testParentAbsoluteSingleMatrixOp(self):
...
+            capsuleXformable.GetXformOpOrderAttr().Get(), Vt.TokenArray((
+                "xformOp:translate", "xformOp:rotateXYZ", "xformOp:rotateX")))
+
+        # Redo: capsule still hasn't moved, Maya fallback ops are back.
+        cmds.redo()
+
+        checkParentDone()
+
+    @unittest.skipUnless(mayaUtils.previewReleaseVersion() >= 123, 'Requires Maya fixes only available in Maya Preview Release 123 or later.') 
+    def testZParentAbsoluteMultiMatrixOp(self):
```
ppt-adsk, line 511: Evil test order dependence is back.  testParentAbsoluteMultiMatrixOp() causes testParentAbsoluteSingleMatrixOp() to fail if the former is run first, so rename it.

test/lib/ufe/testParentCmd.py
```diff
@@ -386,6 +386,277 @@ def testParentAbsoluteSingleMatrixOp(self):
...
+            xformXformable.GetXformOpOrderAttr().Get(), Vt.TokenArray((
+                "xformOp:translate", "xformOp:rotateXYZ")))
+
+        sn.clear()
+
+        capsulePath = ufe.Path([proxyShapePathSegment, 
+                                usdUtils.createUfePathSegment('/Capsule1')])
+        capsuleItem = ufe.Hierarchy.createItem(capsulePath)
+
+        # Add 3 matrix transform ops to the capsule.
```
ppt-adsk, line 551: Create a stack of 3 matrices, and test editing of each one, in turn.

test/lib/ufe/testParentCmd.py
```diff
@@ -386,6 +386,280 @@ def testParentAbsoluteSingleMatrixOp(self):
...
+            # Redo: capsule still hasn't moved.
+            cmds.redo()
+
+            checkParentDone()
+
+            # Go back to initial conditions for next iteration of loop.
+            cmds.undo()
+
+        # Restore initial conditions.
+        del os.environ['MAYA_USD_MATRIX_XFORM_OP_NAME']
```
kxl-adsk, line 661: How about we spare other developers debugging and put an assert in setUp to detect when another test left this env variable set?

lib/mayaUsd/ufe/UsdTransform3dSetObjectMatrix.h
kxl-adsk: You wrote a lot of great comments and documentation...but it still took me some time to figure out what exactly maps to Ml, Mw, Mr. Now looking at it back, what would help me to understand the code quicker is an example of a xform stack and how you break it down into these three transformations given a specific op we want to edit. It's all in the code, but having it described in one place (including pull request description) would help a lot.

kxl-adsk (commented):
Left bunch of comments...mostly requests for improvements to the documentation.

lib/mayaUsd/ufe/UsdTransform3dFallbackMayaXformStack.cpp
```diff
@@ -143,7 +144,10 @@ void setXformOpOrder(const UsdGeomXformable& xformable)
     xformable.SetXformOpOrder(newOrder, resetsXformStack);
 }
 
-Ufe::Transform3d::Ptr createTransform3d(const Ufe::SceneItem::Ptr& item)
+Ufe::Transform3d::Ptr createEditTransform3dImp(
```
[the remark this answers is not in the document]
ppt-adsk: Will add documentation.  As for input versus output, I usually stick to
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f17-for-in-out-parameters-pass-by-reference-to-non-const
so that non-const ref parameters will be modified inside the function, but I will add that information in the documentation.

lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
```diff
@@ -466,6 +469,9 @@ Ufe::Vector3d UsdTransform3dMayaXformStack::rotation() const
     }
     UsdGeomXformOp r = getOp(NdxRotate);
     TF_AXIOM(r);
+    if (!r.GetAttr().HasValue()) {
```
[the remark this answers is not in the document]
ppt-adsk: Here are my findings:
- UsdTransform3d.cpp: HasAttribute() checks in place.
- UsdTransform3dBase.cpp: default implementations only, no attribute access.
- UsdTransform3dCommonAPI.cpp: no direct attribute access, only accessing
      through UsdGeomXformCommonAPI object, which returns the identity matrix
      if no value has been authored, which is fine.
- UsdTransform3dFallbackMayaXformStack.cpp: no value attribute access.
- UsdTransform3dMatrixOp.cpp: getValue() utility is a problem, it ignores the
      UsdAttribute::Get() boolean return value, which may be false.  Entered as MAYA-110061, assigned to me.
 
      Otherwise, uses UsdGeomXformOp::GetOpTransform(), which returns the
      identity matrix if no value has been authored, which is fine.
- UsdTransform3dMayaXformStack.cpp: fixed in this pull request.
- UsdTransform3dSetObjectMatrix.cpp: no direct attribute access.

lib/mayaUsd/ufe/UsdTransform3dFallbackMayaXformStack.cpp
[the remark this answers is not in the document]
kxl-adsk: I would prefer to see more usage of Doxygen formatted documentation with proper usage of direction `[in]` and `[out]` to make this clear - https://www.doxygen.nl/manual/commands.html#cmdparam. I won't block the review on it.

lib/mayaUsd/ufe/UsdTransform3dSetObjectMatrix.h
```diff
@@ -58,6 +58,36 @@ namespace ufe {
 // matrices inv(Ml) and inv(Mr).  In the case of the Maya fallback transform
 // stack, Mr is the identity by definition, as the Maya fallback transform
 // stack must be the last group of transform ops in the transform stack.
+//
+// Here is an example given a Maya fallback transform stack:
```
kxl-adsk, line 62: Awesome!

test/lib/ufe/testParentCmd.py
```diff
@@ -84,6 +84,21 @@ def setUp(self):
...
         cmds.select(clear=True)
 
+        # Save current 'MAYA_USD_MATRIX_XFORM_OP_NAME' env var, if present.
+        self.mayaUsdMatrixXformOpName = os.environ.get(
+            'MAYA_USD_MATRIX_XFORM_OP_NAME')
+
+    def tearDown(self):
+        # If there was no 'MAYA_USD_MATRIX_XFORM_OP_NAME' environment variable,
+        # make sure there is none at the end of the test.
+        if self.mayaUsdMatrixXformOpName is None:
```
kxl-adsk, line 94: Thank you

kxl-adsk (approved):

PR #1204, opened 2021-02-23
MAYA-108709 add a command base class to handle undo item

pierrebai-adsk:
- Add UsdUndoableCommandBase that hanels the UsdUndoableItem, undo/redo and the distinction between set() being called during execute(0 or redo().
- Adapt existing transform commands to use this base class.


lib/mayaUsd/ufe/UsdUndoableCommandBase.h
```diff
@@ -0,0 +1,143 @@
...
+    //
+    // Note: unfortunately, we need to track the initial/done/undo
+    //       state ourselves because UFE does *not* call redo() to
+    //       redo, but instead calls set() again.
+    //
+    //       So, if we want to use the USD undo system, we have to
+    //       track the state of the undo /redo ourselves so that when
+    //       set() is called when we're in a undone state, we do
+    //       redo() instead...
+    enum class State
```
pierrebai-adsk, line 72: Note: I had to keep the state variable because some transform command previously always set the value on the attribute, even during redo(). Now that we are using undo-item for redo, we need to distinguish between real set() and redo(), early undo or redo.

lib/mayaUsd/ufe/UsdUndoableCommandBase.h
```diff
@@ -0,0 +1,143 @@
...
+    // executed "within" a UsdUndoBlock to capture undo data,
+    // to be implemented by the sub-class.
+    virtual void executeImpl() = 0;
+
+private:
+    UsdUndoableItem _undoableItem;
+    State           _state = State::Initial;
+};
+
+// Implementation of the command base class.
```
ppt-adsk, line 89: Because this is all inline, in a single file, and the implementation is really short, I would put the implementation inside the class declaration.  Shorter, less repetition.

lib/mayaUsd/ufe/UsdUndoableCommandBase.h
```diff
@@ -0,0 +1,143 @@
...
+
+template <typename Cmd>
+#ifdef UFE_V2_FEATURES_AVAILABLE
```
ppt-adsk, line 92: I'm hoping we can avoid this...  Do we really need your new class in code that supports UFE v1?

ppt-adsk (changes requested):

lib/mayaUsd/ufe/UsdUndoableCommandBase.h
pierrebai-adsk: I only wrote it because some of the the classes using it were doing this. I can remove it.

ppt-adsk: I know, it's a compromise between helping out old code that is UFE v1 dependent, versus making new code UFE v1 dependent and thus more difficult to read, understand and maintain by adding UFE v1 conditional compilation to it...  If it doesn't save a massive amount of UFE v1 code, then I would vote for not adding UFE v1 to your new code.

lib/mayaUsd/ufe/UsdTransform3dCommonAPI.cpp
```diff
@@ -42,32 +43,31 @@ class UsdTranslateUndoableCmd : public Ufe::TranslateUndoableCommand
...
 
-    void undo() override { _commonAPI.SetTranslate(_prevT, _time); }
-    void redo() override { _commonAPI.SetTranslate(_t, _time); }
-
     // Executes the command by setting the translation onto the transform op.
     bool set(double x, double y, double z) override
     {
         _t = GfVec3d(x, y, z);
-        redo();
+        execute();
```
ppt-adsk, line 53: I'm perplexed by the fact that set() calls execute(), which will call executeUndoBlock().  set() is called during the manipulation; at that point whatever attributes needed to be added to the prim to receive the in-manipulation values should have been added.  If you call execute() (and thus executeUndoBlock()) isn't it the case that you will replace the contents of the UsdUndoableItem with the new value you've set?  See here:
https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/undo/UsdUndoManager.cpp#L56
This is insufficient to do a complete undo, e.g. if you've added a new translate xform op to the prim, it should be removed.

lib/mayaUsd/ufe/UsdTransform3dMatrixOp.cpp
```diff
@@ -93,13 +93,12 @@ GfMatrix4d xformInv(
     return m.GetInverse();
 }
 
-// Class for setMatrixCmd() implementation.  UsdUndoBlock data member and
-// undo() / redo() should be factored out into a future command base class.
-class UsdSetMatrix4dUndoableCmd : public Ufe::SetMatrix4dUndoableCommand
+// Class for setMatrixCmd() implementation.
+class UsdSetMatrix4dUndoableCmd : public UsdUndoableCommand<Ufe::SetMatrix4dUndoableCommand>
```
ppt-adsk, line 97: Excellent, thanks.

lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
```diff
@@ -162,13 +161,12 @@ createTransform3d(const Ufe::SceneItem::Ptr& item, NextTransform3dFn nextTransfo
     return stackOps.empty() ? nextTransform3dFn() : UsdTransform3dMayaXformStack::create(usdItem);
 }
 
-// Class for setMatrixCmd() implementation.  UsdUndoBlock data member and
-// undo() / redo() should be factored out into a future command base class.
-class UsdSetMatrix4dUndoableCmd : public Ufe::SetMatrix4dUndoableCommand
+// Class for setMatrixCmd() implementation.
```
ppt-adsk, line 164: Excellent.

lib/mayaUsd/ufe/UsdTransform3dCommonAPI.cpp
pierrebai-adsk: But we do need to capture the new value in the UsdUndoableItem so that if the user does undo then redo, the redo puts the latest value that was set, right?

I'm not sure how that would work. If we only call set, then the UsdUndoableItem will not contain the latest value? But OTOH, as you say, if we call execute, then if the attribute had been created, the newly generated UsdUndoableItem will not capture that.

I'll have to think about how to make both cases work.

pierrebai-adsk: I'll take a look at the API of UsdUndoableItem to see if we can update its content without creating a new instance from scratch.

pierrebai-adsk: Ah, I see that doInvert() declares an UsdUndoBlock, so that will capture the latest value once undo() gets called. I had assume the undo item was capture both the old and new value. So I'll change the code not to call execute.

pierrebai-adsk: Looking at the code and running the unit-tests, I saw that the correct fix was to keep calling execute() but having execute() detects that it had already been executed and not refill the UsdUndoableItem, instead skipping it and directly calling the implementation (executeUndoBlock).

So, for short, for these interactive commands, set(0 can call execute() which will take care of capturing the changes within a UsdUndoBlock + UsdUndoableItem or not, depending on the fact that execute() has already been called or not, which was conveniently already captured in the state variable.

PR #1205, opened 2021-02-23
MAYA-106780 - Editing USD should mark scene as dirty

seando-adsk:
#### MAYA-106780 - Editing USD should mark scene as dirty
* New regression test for checking scene dirty.

### DIFF
--- a/plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
+++ b/plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
@@ -115,7 +115,7 @@ global proc stageFromFile_UICommit(string $parent) {
         (`checkBoxGrp -q -value1 loadPayloadsCheckBox`);
 }
 
-proc doCreateStage(string $fileName) {
+proc string doCreateStage(string $fileName) {
     // actually load the file
     string $baseName = capitalizeString(`basenameEx $fileName`);
     if( ! isValidObjectName($baseName) )
@@ -132,6 +132,8 @@ proc doCreateStage(string $fileName) {
     setAttr ($shapeNode+".loadPayloads") $loadp;
     connectAttr time1.outTime ($shapeNode+".time");
     select -r $shapeNode;
+    string $fullPath[] = `ls -l $shapeNode`;
+    return $fullPath[0];
 }
 
 global proc createStageFromExistingCallback(string $parent, int $doIt)
@@ -239,7 +241,7 @@ global proc mayaUsd_createStageFromFile() {
     }
 }
 
-global proc mayaUsd_createStageFromFilePath(string $usdFile)
+global proc string mayaUsd_createStageFromFilePath(string $usdFile)
 {
-    doCreateStage($usdFile);
+    return doCreateStage($usdFile);
 }
--- a/plugin/adsk/scripts/mayaUsd_createStageWithNewLayer.py
+++ b/plugin/adsk/scripts/mayaUsd_createStageWithNewLayer.py
@@ -32,3 +32,5 @@ def createStageWithNewLayer():
     shapeNode = cmds.createNode('mayaUsdProxyShape', skipSelect=True, name='stageShape1')
     cmds.connectAttr('time1.outTime', shapeNode+'.time')
     cmds.select(shapeNode, replace=True)
+    fullPath = cmds.ls(shapeNode, long=True)
+    return fullPath[0]
--- a/test/lib/CMakeLists.txt
+++ b/test/lib/CMakeLists.txt
@@ -11,6 +11,7 @@ set(TEST_SCRIPT_FILES
 
 if (UFE_FOUND AND MAYA_APP_VERSION VERSION_GREATER 2020)
     list(APPEND TEST_SCRIPT_FILES
+        testMayaUsdDirtyScene.py
         testMayaUsdProxyAccessor.py
     )
 endif()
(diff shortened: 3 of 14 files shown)


test/lib/CMakeLists.txt
```diff
@@ -15,6 +15,12 @@ if (UFE_FOUND AND MAYA_APP_VERSION VERSION_GREATER 2020)
     )
 endif()
 
+if (UFE_FOUND AND MAYA_APP_VERSION VERSION_GREATER_EQUAL 2022)
+    list(APPEND TEST_SCRIPT_FILES
+        testMayaUsdDirtyScene.py
+    )
+endif()
```
seando-adsk, line 22: The UFE scene dirty was only added in Maya 2022, so this test can only be done there.

test/lib/testMayaUsdDirtyScene.py
```diff
@@ -0,0 +1,163 @@
...
+def getMayaStage():
+    proxyShapes = cmds.ls(type=PROXY_NODE_TYPE, long=True)
+    shapePath = proxyShapes[0]
+    stage = mayaUsd.lib.GetPrim(shapePath).GetStage()
+    return shapePath, stage
+
+def getCleanMayaStage():
+    """ gets a stage that only has an anon layer """
+    cmds.file(new=True, force=True)
+    mayaUsd_createStageWithNewLayer.createStageWithNewLayer()
```
ppt-adsk, line 39: Kinda makes you wish that createStageWithNewLayer() would return the stage...  Out of scope for this pull request, of course, but this would be a compatible change we could make.

test/lib/testMayaUsdDirtyScene.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def testDefaultPrim(self):
+        mayaUtils.openTopLayerScene()
+        shapePath, stage = getMayaStage()
+
+        # Change default prim -> dirties scene
+        ufePath = ufe.PathString.path('|transform1|proxyShape1,/Room_set/Props/Ball_35')
+        item = ufe.Hierarchy.createItem(ufePath)
+        raw = item.getRawAddress()
+        prim = mayaUsd.ufe.getPrimFromRawItem(item.getRawAddress())
```
ppt-adsk, line 134: If you want to limit the use of getRawAddress(), could use mayaUsd.ufe.ufePathToPrim() with your line 131 path string above.

ppt-adsk (approved):
Nice test!

test/lib/testMayaUsdDirtyScene.py
seando-adsk: I like that idea. It wasn't that hard to do and I've cleaned up all the other tests.

test/lib/testMayaUsdDirtyScene.py
seando-adsk: Right, good suggestion.

plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
```diff
@@ -239,7 +241,7 @@ global proc mayaUsd_createStageFromFile() {
     }
 }
 
-global proc mayaUsd_createStageFromFilePath(string $usdFile)
+global proc string mayaUsd_createStageFromFilePath(string $usdFile)
 {
-    doCreateStage($usdFile);
+    return doCreateStage($usdFile);
```
seando-adsk, line 246: Modified so it returns the full path of the proxy shape node it created.

plugin/adsk/scripts/mayaUsd_createStageWithNewLayer.py
```diff
@@ -32,3 +32,5 @@ def createStageWithNewLayer():
     shapeNode = cmds.createNode('mayaUsdProxyShape', skipSelect=True, name='stageShape1')
     cmds.connectAttr('time1.outTime', shapeNode+'.time')
     cmds.select(shapeNode, replace=True)
+    fullPath = cmds.ls(shapeNode, long=True)
+    return fullPath[0]
```
seando-adsk, line 36: Modified so it returns the full path of the proxy shape node it created.

test/lib/mayaUsd/nodes/testProxyShapeBase.py
```diff
@@ -57,8 +57,8 @@ def testDuplicateProxyStageAnonymous(self):
 
         # create a proxy shape and add a Capsule prim
         import mayaUsd_createStageWithNewLayer
-        mayaUsd_createStageWithNewLayer.createStageWithNewLayer()
-        proxyShapePath = ufe.PathString.path('|stage1|stageShape1')
+        proxyShape = mayaUsd_createStageWithNewLayer.createStageWithNewLayer()
+        proxyShapePath = ufe.PathString.path(proxyShape)
```
seando-adsk, line 61: Modified all tests that use either of these two helpers (above) to use the new return value.

test/lib/ufe/testAttribute.py
```diff
@@ -575,14 +575,14 @@ def testZAttrChangeRedoAfterPrimCreateRedo(self):
...
-        proxyShapePath = ufe.PathString.path('|stage1|stageShape1')
+        proxyShape = mayaUsd_createStageWithNewLayer.createStageWithNewLayer()
+        proxyShapePath = ufe.PathString.path(proxyShape)
         proxyShapeItem = ufe.Hierarchy.createItem(proxyShapePath)
         proxyShapeContextOps = ufe.ContextOps.contextOps(proxyShapeItem)
         cmd = proxyShapeContextOps.doOpCmd(['Add New Prim', 'Capsule'])
         ufeCmd.execute(cmd)
 
-        capsulePath = ufe.PathString.path('|stage1|stageShape1,/Capsule1')
+        capsulePath = ufe.PathString.path('%s,/Capsule1' % proxyShape)
```
seando-adsk, line 585: A few other changes like this - replacing the hard coded string (of proxy shape) with return value from create helper.

ppt-adsk (approved):
Thanks for making the createStage() change, and doing the cleanup!

Issue #951, opened 2020-11-30
[MAYA-109889] Incorrect viewport selection for "bounds" drawMode


csyshing:
**Describe the bug**
Hi,

We encountered a viewport (**UFE**/**VP2**) selection issue that when a prim was set to draw "bounds", the prim can still being selected even if it's behind other prims, looks like a depth sorting issue in `vp2renderdelegate`?

**Steps to reproduce**
1. Launch Maya and load `maya-usd` plugin;
2. Import attached simple USD file *spheres.usda*:
    - The USD file has two spheres, notice the first one, *ball_01* is set to **"bounds"**;
    - A side note for the metadata **prepend apiSchemas = ["GeomModelAPI"]** : it is for USD-20.11, it can be ignored if USD version < 20.11 (https://github.com/PixarAnimationStudios/USD/issues/1401) 
3. Change the viewport a bit to let the second sphere displayed in front of the first one (bbox);
4. Select the the overlapping area on the sphere, but the bbox behind will be selected;
5. Create a Maya native cube, repeat steps 3 and 4, the problem can be reproduced as well.

![incorrect_viewport_selection_for_bounds](https://user-images.githubusercontent.com/1593576/100596826-b90d7800-3350-11eb-96e8-bb400c8ced3a.gif)

**Expected behavior**
Expected to have correct selection from viewport.

**Attachments**
[spheres.zip](https://github.com/Autodesk/maya-usd/files/5615237/spheres.zip)

**Specs (if applicable):**
 - OS & version: Centos 7.4
 - Compiler & version: gcc 6.3.1
 - Maya version: Maya 2020
 - Maya USD commit SHA: v0.5.0 [dev at 62037f3] **UFE** build
 - Pixar USD commit SHA: official tagged version 19.11 and 20.11

santosg87:
thank you for logging this. 

I am able to reproduce this internally as well.

testing a bit further, it seems to be the case with the origin draw mode as well, but not with actual geometry.

csyshing:
Thanks @santosg87 to confirm. May I ask if there is any plan how to fix it? Should I expect to see a fix in `maya-usd` ?

HdC-adsk:
[...]

PR #1208, opened 2021-02-24
Use the nurbs curve selection mask for basis curves. 

williamkrick:
kSelectCurves includes kSelectSurfaceEdge which has a higher priority that causes issues.

This addresses https://github.com/Autodesk/maya-usd/issues/951


HdC-adsk:
LGTM. Thanks for fixing the issue.

Maybe one addition test: can you check off `Select curve objects` button in the Maya tools and see whether we are still able to use that setting to control whether we want to select USD curves or not?

image

HdC-adsk (approved):

williamkrick:
With `Select curve objects` disabled I am not able to pick basisCurves in the viewport, which seems like the correct behavior.

HdC-adsk:
Cool. Thanks for confirming.

PR #1212, opened 2021-02-24
null and anonymous layer checks

fowlertADSK:


lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -504,7 +504,9 @@ bool LayerDatabase::saveUsdToUsdFiles()
 
             SdfLayerHandleVector allLayers = stage->GetLayerStack(false);
             for (auto layer : allLayers) {
-                layer->Save();
+                if (!layer->IsAnonymous()) {
+                    layer->Save();
+                }
```
fowlertADSK, line 509: There is a larger change coming where we handle anonymous root layers better, but this avoids the error that is printed to the script editor even though everything appears to work correctly.  The problem with an anonymous root layer is that we Export it and then do a setAttr on the filePath attribute of the proxy to update it, but right here after we've done the Export we are still holding on to the old root layer.  All we need to do here is skip the anonymous layer, and it will be replaced shortly after when the proxy computes again.

lib/mayaUsd/utils/utilSerialization.cpp
```diff
@@ -212,7 +212,7 @@ SdfLayerRefPtr UsdMayaSerialization::saveAnonymousLayer(
 
     SdfLayerRefPtr newLayer = SdfLayer::FindOrOpen(path);
 
-    if (parentLayer) {
+    if (newLayer && parentLayer) {
```
fowlertADSK, line 215: Missed the null check somehow when copying this over from the Save Layers UI.

lib/mayaUsd/nodes/layerManager.cpp
mattyjams: You may want to consider using `SdfLayer::PermissionToSave()` instead which will catch the anonymous case as well as others that might be problematic, such as saving to a muted layer:

https://github.com/PixarAnimationStudios/USD/blob/dc710925675f7f58f7a37f6c18d046a687b09271/pxr/usd/sdf/layer.cpp#L2011

fowlertADSK: Good catch.  I'm remember this one while I continue cleaning up some of the serialization code.

seando-adsk (approved):

PR #1213, opened 2021-02-25
Fix crash on failure to initialize a renderer, and only try once.

marsupial:
If a render-delegate fails to initialize itself **mtoh** will crash.
The checks here are a bit overly-cautious, but the function is called once (after switching renderers) so shouldn't affect perf.

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
@@ -348,7 +348,7 @@ std::vector<MString> MtohRenderOverride::AllActiveRendererNames()
 
     std::lock_guard<std::mutex> lock(_allInstancesMutex);
     for (auto* instance : _allInstances) {
-        if (instance->_initializedViewport) {
+        if (instance->_initializationSucceeded) {
             renderers.push_back(instance->_rendererDesc.rendererName.GetText());
         }
     }
@@ -525,7 +525,7 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext)
             _lastRenderTime = std::chrono::system_clock::now();
         }
     };
-    if (_initializedViewport && !_taskController) {
+    if (_initializationAttempted && !_initializationSucceeded) {
         // Initialization must have failed already, stop trying.
         return MStatus::kFailure;
     }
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
```diff
@@ -533,6 +537,13 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext)
 
     if (!_initializedViewport) {
         _InitHydraResources();
+
+        if (!_taskController) {
+            // No task controller means something has gone wrong, use this as a signal so we
+            // aren't endlessly failing to initialize.
```
pmolodo, line 543: As you say, we're essentially using `_taskController` as a indicator for whether initialization succeeded; however, I think it would be better if we had a new dedicated member variable, say `_initializationSucceeded`, for this purpose - both for clearer naming, and because there's a decent chunk of stuff done AFTER the `_taskController` is created.  We could then also rename `_initializatedViewport` to `_initializationAttempted` - since that's effectively how it's being used now - and unconditionally set it as soon as `_InitHydraResources()` starts.  Again, this is effectively how it's already functioning, since it's set to true whether it succeeds or fails, but it's obscured by the fact that it's set to `true` in two different places.

Between the naming and placement of where these are set (at the very start and end of `_InitHydraResources()`), I think the intent will be a lot clearer, without the need for explanation.

pmolodo (changes requested):
The idea and makes sense, and the implementation is functional, but since I think we're now co-opting some variables for multiple uses, and slightly changing how others are used, I think the code could be clarified.  Made comments in the code.

pmolodo:
I realized that it would probably have been simpler to just make the changes rather than try to describe them, so I did so here:

https://github.com/LumaPictures/maya-usd/tree/pr/mtoh-init-crash
https://github.com/LumaPictures/maya-usd/commit/fca7e01536492520d8c95c02bd1e95e3fb100673

Let me know if that makes sense for you...

(Note - still haven't confirmed my changes work / build, since it's only about 1/3 of the way done on my laptop, but will update when it finishes!)

pmolodo (approved):
Looks good!

PR #1218, opened 2021-02-26
MAYA-107424 - On stage AE template load/reload a file as stage source

seando-adsk:
MAYA-107424 - On the stage AE template I'd like to load/reload a file as the stage source

* Added debugging messages to template helpers (controlled from var).
* New template helper method to deal with file path being changed from text field and file browser button.
* New template helper method to refresh file path attribute.
* Replace file path attribute with custom control containing: label, text field (for path), file browser button, file refresh button.

### DIFF
--- a/plugin/adsk/scripts/AETemplateHelpers.py
+++ b/plugin/adsk/scripts/AETemplateHelpers.py
@@ -1,6 +1,14 @@
+import os.path
+import maya.cmds as cmds
 import ufe
 import mayaUsd.ufe
 import maya.internal.ufeSupport.ufeCmdWrapper as ufeCmd
+import maya.mel as mel
+
+def debugMessage(msg):
+    DEBUG = False
+    if DEBUG:
+        print(msg)
 
 def GetDefaultPrimName(proxyShape):
     try:
@@ -9,7 +17,8 @@ def GetDefaultPrimName(proxyShape):
             defPrim = proxyStage.GetDefaultPrim()
             if defPrim:
                 return defPrim.GetName()
-    except:
+    except Exception as e:
+        debugMessage('GetDefaultPrimName() - Error: %s' % str(e))
         pass
     return ''
 
@@ -20,7 +29,8 @@ def GetRootLayerName(proxyShape):
             rootLayer = proxyStage.GetRootLayer()
             if rootLayer:
                 return rootLayer.GetDisplayName() if rootLayer.anonymous else rootLayer.realPath
-    except:
+    except Exception as e:
+        debugMessage('GetRootLayerName() - Error: %s' % str(e))
         pass
     return ''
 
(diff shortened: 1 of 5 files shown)


plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -35,4 +45,162 @@ def ProxyShapePayloads(proxyShape, loadAll):
...
+        elif newFilePath is not None:
+            # Instead of opening a file dialog to get the USD file, simply
+            # use the one provided as input.
+            # Note: it is important to check "not None" since the new file path can be
+            #       an empty string.
+            debugMessage('  New file to load provided, setting filepath="%s"' % newFilePath)
+            cmds.setAttr(filePathAttr, newFilePath, type='string')
+            return True
+    except Exception as e:
+        debugMessage('ProxyShapeFilePathChanged() - Error: %s' % str(e))
```
seando-adsk, line 158: You'll see I added this in the other functions above that had try/except blocks already. During my dev work whenever there was an error during the script it would end up here, but I didn't know why until I added the exception error message.

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -35,4 +45,162 @@ def ProxyShapePayloads(proxyShape, loadAll):
...
+        if dirtyStack:
+            kTitleFormat = mel.eval('getMayaUsdString("kDiscardStageEditsTitle")')
+            kMsgFormat = mel.eval('getMayaUsdString("kDiscardStageEditsLoadMsg")')
+            kYes = mel.eval('getMayaUsdString("kButtonYes")')
+            kNo = mel.eval('getMayaUsdString("kButtonNo")')
+            kTitle = cmds.format(kTitleFormat, stringArg=stageName)
+            # Our title is a little long, and the confirmDialog command is not making the
+            # dialog wide enough to show it all. So by telling the message string to not
+            # word-wrap, the dialog is forced wider.
+            kMsg = "<p style='white-space:pre'>" + cmds.format(kMsgFormat, stringArg=stageName)
```
seando-adsk, line 125: Trick to make the dialog wider (by not wrapping the text).

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -35,4 +45,162 @@ def ProxyShapePayloads(proxyShape, loadAll):
...
+            kMsgFormat = mel.eval('getMayaUsdString("kDiscardStageEditsReloadMsg")')
+            kYes = mel.eval('getMayaUsdString("kButtonYes")')
+            kNo = mel.eval('getMayaUsdString("kButtonNo")')
+            kTitle = cmds.format(kTitleFormat, stringArg=stageName)
+            # Our title is a little long, and the confirmDialog command is not making the
+            # dialog wide enough to show it all. So by telling the message string to not
+            # word-wrap, the dialog is forced wider.
+            filepath = cmds.getAttr(filePathAttr)
+            if filepath:
+                filename = os.path.basename(filepath)
```
seando-adsk, line 195: Does our proxy shape handle the asset resolver and if so would I need a special case here when its not actually a file path?

plugin/adsk/scripts/AEmayaUsdProxyShapeBaseTemplate.mel
```diff
@@ -56,6 +56,75 @@ global proc AEmayaUsdProxyShapePayloadsReplace(string $loadPayloads)
...
+global proc AEmayaUsdProxyShapeFilePathBrowser(string $filePathAttr)
+{
+    python("import AETemplateHelpers; AETemplateHelpers.ProxyShapeFilePathChanged('" + $filePathAttr + "')");
+}
+
+// Called when the user clicks the file path refresh button.
+global proc AEmayaUsdProxyShapeFilePathRefresh(string $filePathAttr)
+{
+    python("import AETemplateHelpers; AETemplateHelpers.ProxyShapeFilePathRefresh('" + $filePathAttr + "')");
+}
```
seando-adsk, line 116: Wrapped these in functions (instead of attaching the python code directly to the buttons) so that the user will see a better name, rather than obscure python code.

plugin/adsk/scripts/AEmayaUsdProxyShapeBaseTemplate.mel
```diff
@@ -56,6 +56,75 @@ global proc AEmayaUsdProxyShapePayloadsReplace(string $loadPayloads)
...
+
+// Called in response to attributeChange notif on filePath attribute.
+global proc AEmayaUsdProxyShapeFilePathChanged(string $filePathAttr)
+{
+    // The Proxy Shape file path attribute was changed outside so we need to set
+    // the new value in our field and enable/disable the refresh button.
+    string $filePath = `getAttr $filePathAttr`;
+    textField -edit -fileName $filePath mayaUsdProxyShapeFilePath;
```
seando-adsk, line 124: Possibly another place to handle strings that aren't paths, but assets?

seando-adsk (commented):
There was only one minor change (to a string in the reload dialog) since my demo yesterday.

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -35,4 +45,162 @@ def ProxyShapePayloads(proxyShape, loadAll):
...
+        debugMessage('ProxyShapePayloads() - error during function')
+        pass
+
+def IsProxyShapeLayerStackDirty(proxyStage):
+    # Helper method which returns True if any layer (skipping SessionLayer)
+    # in the layer stack is dirty.
+    if proxyStage:
+        # Is any layer in the layerstack dirty?
+        for layer in proxyStage.GetLayerStack(includeSessionLayers=False):
+            debugMessage('  Proessing layer: %s' % layer.GetDisplayName())
```
ppt-adsk, line 57: "Proessing" --> "Processing".

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -35,4 +45,162 @@ def ProxyShapePayloads(proxyShape, loadAll):
...
+                debugMessage('    Found dirty layer')
+                return True
+    return False
+
+def GetStageFromProxyShapeAttr(attr):
+    # Helper method which returns the proxy shape name and Usd stage
+    # from the input attribute on the proxy shape.
+
+    # First get the stage name from the input attribute.
+    stageName = mel.eval('plugNode("%s")' % attr)
```
ppt-adsk, line 68: Only thing plugNode.mel does is this:
tokenize($plug, ".", $buffer);
return $buffer[0];
Feels like we can do this pretty easily using Python string manipulation.

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -35,4 +45,162 @@ def ProxyShapePayloads(proxyShape, loadAll):
...
+            rootLayer = proxyStage.GetRootLayer()
+            if rootLayer and not rootLayer.empty:
+                dirtyStack = True
+            elif newFilePath is None:
+                debugMessage('  Empty stage, opening file dialog...')
+                openFileDialog = True
+
+        # If we have a dirty layer stack, display dialog to confirm new loading.
+        if dirtyStack:
+            kTitleFormat = mel.eval('getMayaUsdString("kDiscardStageEditsTitle")')
```
ppt-adsk, line 117: Pretty awkward that there is no way to do this from Python.  Certainly getMayaUsdString() isn't doing much, and 
https://git.autodesk.com/maya3d/maya/blob/master/Maya/src/OpenMaya/Commands/getPluginResource.mel
isn't doing much either...  I guess we can just leave it as you have it, not a big deal.

plugin/adsk/scripts/AETemplateHelpers.py
ppt-adsk: No idea.

ppt-adsk (approved):

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -35,4 +45,162 @@ def ProxyShapePayloads(proxyShape, loadAll):
...
+                openFileDialog = True
+            else:
+                # User said NO to update, so don't proceed any farther.
+                return False
+
+        if openFileDialog and newFilePath is None:
+            # Pop the file open dialog for user to load new usd file.
+            title = mel.eval('getMayaUsdString("kLoadUSDFile")')
+            okCaption = mel.eval('getMayaUsdString("kLoad")')
+            fileFilter = mel.eval('getMayaUsdString("kAllUsdFiles")') + ' (*.usd *.usda *.usdc *.usdz );;*.usd;;*.usda;;*.usdc;;*.usdz';
```
fowlertADSK, line 141: Do we have the filter tokens available in Python?  e.g. UsdMayaTranslatorTokens is what we use in c++ when we need filters like this.

plugin/adsk/scripts/AETemplateHelpers.py
fowlertADSK: I don't think we do anything special with the actual string attribute, it gets passed to USD to do something with it and it's there that things get resolved.

fowlertADSK (approved):

plugin/adsk/scripts/AETemplateHelpers.py
seando-adsk: thx - fixed

plugin/adsk/scripts/AETemplateHelpers.py
seando-adsk: Right, good idea.

plugin/adsk/scripts/AETemplateHelpers.py
seando-adsk: Great.

plugin/adsk/scripts/AETemplateHelpers.py
seando-adsk: I see those in the C++ code, but I don't see anywhere they are wrapped into python. I also checked the USD repo and in their python code they just use the actual strings.

plugin/adsk/scripts/AETemplateHelpers.py
seando-adsk: I looked at the MEL procs for register/get string resources and unfortunately they don't have python equivalents. But they are trivial, so I've implemented a python version of them in our plugin and now no longer need the mel.eval.

plugin/adsk/scripts/mayaUSDRegisterStrings.py
```diff
@@ -0,0 +1,70 @@
...
+#
+
+import maya.cmds as cmds
+import maya.mel as mel
+
+def register(key, value):
+    registerPluginResource('mayaUsdPlugin', key, value)
+
+def getMayaUsdString(key):
+    return getPluginResource('mayaUsdPlugin', key)
```
seando-adsk, line 23: Identical to the ones in the MEL file (of the same name).

plugin/adsk/scripts/mayaUSDRegisterStrings.py
```diff
@@ -0,0 +1,70 @@
...
+    # This function is called from the equivalent MEL proc
+    # with the same name. The strings registered here and all the
+    # ones registered from the MEL proc can be used in either
+    # MEL or python.
+    register("kButtonYes", "Yes")
+    register("kButtonNo", "No")
+    register("kDiscardStageEditsTitle", "Discard Edits on ^1s's Layers")
+    register("kDiscardStageEditsLoadMsg", "Are you sure you want to load in a new file as the stage source?\n\nAll edits on your layers in ^1s will be discarded.")
+    register("kDiscardStageEditsReloadMsg", "Are you sure you want to reload ^1s as the stage source?\n\nAll edits on your layers (except the session layer) in ^2s will be discarded.")
+    register("kLoadUSDFile", "Load USD File")
```
seando-adsk, line 35: These are the strings that are used exclusively from python. But they can also be loaded by MEL.

plugin/adsk/scripts/mayaUSDRegisterStrings.py
```diff
@@ -0,0 +1,70 @@
...
+    # Form full id string.
+    # Plugin string id's are prefixed with "p_".
+    fullId = 'p_%s.%s' % (pluginId, stringId)
+    if cmds.displayString(fullId, exists=True):
+        dispStr = cmds.displayString(fullId, query=True, value=True)
+        return dispStr
+    else:
+        msgFormat = mel.eval('uiRes("m_getPluginResource.kLookupFailed")')
+        msg = cmds.format(msgFormat, stringArg=(pluginId, stringId))
+        cmds.error(msg)
```
seando-adsk, line 70: Python ports of the MEL procs from Maya.

ppt-adsk (approved):

fowlertADSK (commented):
LGTM

PR #1219, opened 2021-02-26
MAYA-106484 add support for 16-bits textures in VP2 renderer delegate.

pierrebai-adsk:


pierrebai-adsk:
I added Krystian as reviewer mostly so he can be aware of the progress on this task.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -486,6 +486,43 @@ _LoadTexture(const std::string& path, bool& isColorSpaceSRGB, MFloatArray& uvSca
...
+                texels[t * bpp_8 + 0] = storage[t * bpp + 0];
+                texels[t * bpp_8 + 1] = storage[t * bpp + 1];
+                texels[t * bpp_8 + 2] = storage[t * bpp + 2];
+                texels[t * bpp_8 + 3] = storage[t * bpp + 3];
+                texels[t * bpp_8 + 4] = storage[t * bpp + 4];
+                texels[t * bpp_8 + 5] = storage[t * bpp + 5];
+                texels[t * bpp_8 + 6] = lowAlpha;
+                texels[t * bpp_8 + 7] = highAlpha;
+            }
+        }
```
HdC-adsk, line 516: For scenes with lots of large textures (e.g. 4k textures), this loop could be a hotspot and can be converted into a concurrent loop.

HdC-adsk: @kxl-adsk Are we required to support USD 20.08 and before? If so, the other conditional compilation block needs to have the same change as well.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -486,6 +486,43 @@ _LoadTexture(const std::string& path, bool& isColorSpaceSRGB, MFloatArray& uvSca
...
+
+        texture = textureMgr->acquireTexture(path.c_str(), desc, texels.data());
+        isColorSpaceSRGB = image->IsColorSpaceSRGB();
```
HdC-adsk, line 519: I thought sRGB color space was never applied to floating point textures. Is it required to check the state from the image? Below is what I found from wikipedia.

https://en.wikipedia.org/wiki/SRGB

> It is often the "default" color space for images that contain no color space information, especially if the images' pixels are stored in 8-bit integers per color channel.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -523,7 +560,9 @@ _LoadTexture(const std::string& path, bool& isColorSpaceSRGB, MFloatArray& uvSca
         isColorSpaceSRGB = image->IsColorSpaceSRGB();
         texture = textureMgr->acquireTexture(path.c_str(), desc, spec.data);
         break;
-    default: break;
+    default:
+        TF_WARN("VP2 renderer delegate: unsuported pixel format: %d.", (int)spec.format);
```
HdC-adsk, line 564: There is a typo: `unsuported` -> `unsupported`

Maybe you can print the file path as well for easier lookup/debugging?

HdC-adsk (changes requested):

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
pierrebai-adsk: Honestly, given that I would not be able to test it (since Maya does not currently use that old USD version), I would rather not modify code that I would be modifying blindly.

HdC-adsk: We can build with a custom USD version on our own. But let's wait for @kxl-adsk to confirm the requirement first.

kxl-adsk: Yes. Our minimal supported version is 20.02.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -486,6 +488,44 @@ _LoadTexture(const std::string& path, bool& isColorSpaceSRGB, MFloatArray& uvSca
...
+        desc.fBytesPerSlice = desc.fBytesPerRow * spec.height;
+
+        GfHalf               opaqueAlpha(1.0f);
+        const unsigned short alphaBits = opaqueAlpha.bits();
+        const unsigned char  lowAlpha = reinterpret_cast<const unsigned char*>(&alphaBits)[0];
+        const unsigned char  highAlpha = reinterpret_cast<const unsigned char*>(&alphaBits)[1];
+
+        std::vector<unsigned char> texels(desc.fBytesPerSlice);
+
+        tbb::parallel_for(tbb::blocked_range<int>(0, spec.height), [&](tbb::blocked_range<int> r) {
```
HdC-adsk, line 506: The default grain size is 1. It might be too small depending on the width. Could it be better to use `blocked_range2d`?

pierrebai-adsk: I thought it would be better to only parallelize in the Y direction due to data access, dividing in X would reduce locality. (And there are enough values in height to allow parallel work.)

The default partitioner (auto_partitioner) implement the automatic grainsize heuristic, so I don't think we need to specify a grain size. Given the height of the image will be much larger than the number of CPU,

HdC-adsk: Ok, makes sense. Thanks for elaboration. Did you profile the loop performance before and after the change?

pierrebai-adsk: No.

HdC-adsk: Could you profile it for a large texture and a small texture, just to be sure about the change?

pierrebai-adsk: I'm currently rebuilding USD 20.02, I'll look at profiling when I've done the old GL versions of conversion.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
pierrebai-adsk: I just realized that our version of USD 20.02 is *not* built with openimageio... (after all, adding it to USD is what this task is about...) so it cannot possibly load EXR textures.

HdC-adsk: @seando-adsk Do you have any suggestion for Pierre?

pierrebai-adsk: Oh, I ended-up rebuilding USD 20.02 by hand, enabling oiio and openEXR. It just took me the afternoon to find the correct recipe to get it to compile successfully.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
pierrebai-adsk: I profiled with a 18MB texture. The parallel version is slower, but in both cases, the cost of conversion is minimal vs the I/O and rendering, etc. So It does not matter much either way. IDK if 18MB is considered large or not, maybe the parallel version is more "future proof".

HdC-adsk: What is the resolution of the 18MB texture?

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -42,7 +42,14 @@
 #include <maya/MUintArray.h>
 #include <maya/MViewport2Renderer.h>
 
+#if PXR_VERSION > 2008
+#else
```
HdC-adsk, line 46: It is better to remove the empty conditional compilation branch whose PXR_VERSION is larger than 2008.

pierrebai-adsk: The reason I wrote it this way was to be identical to the ifdef used in the _LoadImage() function, to avoid subtle issue if I were to inverse the logic. (for example comparing to <= 2008 instead.) But I can change it.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
pierrebai-adsk: 2048x2048

HdC-adsk: 2kx2k is common. Maybe try 4kx4k and we can see some performance gain?

HdC-adsk: I do agree with you the major overhead here would be I/O overhead and VRAM transfer overhead. So it might not make a big difference overall.

HdC-adsk (approved):

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
kxl-adsk: I like @HdC-adsk suggestions to do a scalability test with a few different sizes of the texture. If we can't prove with the data that there is a value in doing the parallel version, why use up all the resources.

Additionally, is _LoadTexture called from main-thread on it is part of multithreaded sync execution?

HdC-adsk: @kxl-adsk Currently all sprim syncs (from which _LoadTexture is called) are serial on main thread, however, in the future it might be changed to multithreaded sync according to the post below:

https://groups.google.com/g/usd-interest/c/WH6DgLDwNvE/m/7ngJnkYoCAAJ

kxl-adsk: Thx, so we have additional resources currently available (since there is nothing else running at that time). Still would be good to do the scalability test. @pierrebai-adsk do you agree?

pierrebai-adsk: I can try more textures. Do we have bigger EXR files I can test with? Otherwise I'll try to search.

Honestly, this may all fall into the "premature optimization" bucket, we could revert to the simple loop until we do have data showing it is a bottleneck. (The only data point so far with that 18MB texture pointing toward being detrimental.)

HdC-adsk: The image content doesn't matter, so I guess we can easily create some white images with different resolutions and convert them to EXR.

If all resolutions cannot benefit from the concurrent loop, I would agree with the simple loop.

pierrebai-adsk: 👍

pierrebai-adsk: I converted a 4800x6500 image in EXR 16-bits, resulting in a 186MB image. It was faster without parallel for, In all cases, the total time to load / convert /etc was 8% with 5% being spent simply on I/O (in image->Read(spec) at line 461), on a reasonable SSD. So, the I/O is still the biggest factor anyway.

kxl-adsk (approved):

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
kxl-adsk: Perfect. Thank you for providing this data.

HdC-adsk (approved):

Issue #327, opened 2020-03-04
[MAYA-109887] VP2RenderDelegate: support per-instance data via inheritance


cfmoore007:
**Describe the bug**
It appears that inherited displayColors are not supported currently with the VP2 Delegate, while HdSt/Hydra ( mtoh ) seems to support this fine.

Attached is an example asset that uses PxrSurface shaders, but still has displayColor defined.  Another file inheritedDisplayColor_noPxrMtls.usda is included to rule out any influence from the PxrSurface definition.

We tested this w/ an AL_usdmaya_ProxyShape, but it also appears with the Maya proxyShape.

**Steps to reproduce**
1. load included usda file
2. witness blue vp2 fallback shader

**Expected behavior**
displayColor is reflected properly in the viewport that matches Hydra.

**Attachments**
All files :
- [inheritedDisplayColor.tar.gz](https://github.com/Autodesk/maya-usd/files/4288461/inheritedDisplayColor.tar.gz)

**Inherits Display Color (PxrSurface)**
inheritedDisplayColor_pxrSurface.usda
![inheritsDisplayColor](https://user-images.githubusercontent.com/2152250/75901239-35981800-5e0c-11ea-98df-b05592e49543.png)

**VP2 Delegate Result**
![blueApple](https://user-images.githubusercontent.com/2152250/75901272-40eb4380-5e0c-11ea-889b-4b8f91c89549.png)

Mtoh / Hydra Result
![properAppleColor](https://user-images.githubusercontent.com/2152250/75901311-4f395f80-5e0c-11ea-94c9-cfb7e3434911.png)

**Material definition removed - only displayColor**
inheritedDisplayColor_noPxrMtls.usda
![grayApple](https://user-images.githubusercontent.com/2152250/75901356-61b39900-5e0c-11ea-9519-617e54ffc00a.png)

**Specs (if applicable):**
 - OS & version [e.g. Linux]
 - Compiler & version [e.g. gcc 6.3.1]
 - Maya version [e.g. Maya 2020]
 - Maya USD commit SHA [e.g. dev at 44e597f]
 - Pixar USD commit SHA [e.g. 19.11]

**Additional context**

HdC-adsk:
@cfmoore007 Thanks for reporting the issue. A quick look reveals two issues.

* The "payload" node is instanceable and in this case the instancer is where we can fetch the color data from, but not the mesh node. This is not implemented yet in HdVP2Instancer.

[...]

PR #1220, opened 2021-02-26
Add per instance inherited data support to Vp2RenderDelegate.

williamkrick:
This pull request fixes https://github.com/Autodesk/maya-usd/issues/327


williamkrick:
I'm planning to add a test for this as well.

HdC-adsk:
I guess basisCurves has the same requirement, no?

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1832,118 +1871,145 @@ void HdVP2Mesh::_UpdateDrawItem(
...
-            if (stateToCommit._shader != nullptr) {
-                renderItem->setShader(stateToCommit._shader);
-                renderItem->setTreatAsTransparent(stateToCommit._isTransparent);
-            }
+    _delegate->GetVP2ResourceRegistry().EnqueueCommit([stateToCommit,
+                                                       param,
+                                                       primvarInfo,
+                                                       indexBuffer,
+                                                       isBBoxItem,
+                                                       &sharedBBoxGeom]() {
```
HdC-adsk, line 1879: What happens here? I like the new format. Is it done automatically by clang-format and is it upgraded?

lib/mayaUsd/render/vp2RenderDelegate/mesh.h
```diff
@@ -70,6 +70,7 @@ struct PrimvarInfo
 
     PrimvarSource                             _source;
     std::unique_ptr<MHWRender::MVertexBuffer> _buffer;
+    MFloatArray                               _extraInstanceData;
```
HdC-adsk, line 73: It is color specific currently. I don't see we are going to have more other data coming in. So maybe rename it to `_extraInstanceColor` so its name is more informative?

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -544,42 +544,77 @@ void HdVP2Mesh::_PrepareSharedVertexBuffers(
...
+                float* alpha
+                    = reinterpret_cast<float*>(&static_cast<GfVec4f*>(bufferData)[instance])
+                    + alphaChannelOffset;
+
+                if (colorInterp == HdInterpolationInstance) {
+                    *color = colorArray[index];
+                } else if (colorInterp == HdInterpolationConstant) {
+                    *color = colorArray[0];
+                } else {
+                    // can't handle instanced color sets
```
HdC-adsk, line 569: Do you think we need a warning or assertion here?

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -544,42 +544,77 @@ void HdVP2Mesh::_PrepareSharedVertexBuffers(
...
+            void* bufferData = &colorAndOpacityInfo->_extraInstanceData[0];
+
+            size_t alphaChannelOffset = 3;
+            for (size_t instance = 0; instance < numInstances; instance++) {
+                int      index = instanceIndices[instance];
+                GfVec3f* color = reinterpret_cast<GfVec3f*>(
+                    reinterpret_cast<float*>(&static_cast<GfVec4f*>(bufferData)[instance]));
+                float* alpha
+                    = reinterpret_cast<float*>(&static_cast<GfVec4f*>(bufferData)[instance])
+                    + alphaChannelOffset;
```
HdC-adsk, line 562: Yuck! The clang-format looks ugly here, causing different formats for different lines...

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
williamkrick: Yes the indentation changed because of clang-format. I'm not sure why it decided to do things differently. github can ignore whitespace differences from the "gear" menu under the pull request name in the top left of the browser.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1832,118 +1871,145 @@ void HdVP2Mesh::_UpdateDrawItem(
...
+                for (unsigned int i = 0; i < newInstanceCount; i++) {
+                    // VP2 defines instance ID of the first instance to be 1.
+                    drawScene.updateInstanceTransform(
+                        *renderItem, i + 1, stateToCommit._instanceTransforms[i]);
+                }
+            } else {
+                drawScene.setInstanceTransformArray(*renderItem, stateToCommit._instanceTransforms);
+            }
+
+            // upload any extra instance data
```
williamkrick, line 1953: This is the real difference in this code block. I make additional calls to MPxSubSceneOverride::setExtraInstanceData().

I don't have support for generic additional instanced streams because the USD name of the stream is not necessarily the name of the stream in the shader. If we had some way to handle that cleanly then we wouldn't have to make display color a special case.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1832,118 +1871,145 @@ void HdVP2Mesh::_UpdateDrawItem(
...
 #endif
-                drawScene.setInstanceTransformArray(*renderItem, stateToCommit._instanceTransforms);
+            drawScene.setInstanceTransformArray(*renderItem, stateToCommit._instanceTransforms);
+
+            if (stateToCommit._instanceColors.length() == newInstanceCount * kNumColorChannels) {
+                drawScene.setExtraInstanceData(
+                    *renderItem, kSolidColorStr, stateToCommit._instanceColors);
+            }
+
+            // upload any extra instance data
```
williamkrick, line 1991: Also need the upload in this case. The duplicated code here is a bit awkward but I didn't want to refactor this and risk breaking something I don't understand fully. This code path happens on the first frame when _usingInstancedDraw is not yet set.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
williamkrick: I guess there is maybe a performance optimization here where we update the values one by one. @HdC-adsk do you know if calling that method in a loop is faster then updating the whole buffer in one shot? I see the code for updating transforms does it one by one if possible.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1591,11 +1631,15 @@ void HdVP2Mesh::_UpdateDrawItem(
...
-                    && alphaInterp == HdInterpolationConstant) {
+                if ((colorInterp == HdInterpolationConstant
+                     || colorInterp == HdInterpolationInstance)
+                    && (alphaInterp == HdInterpolationConstant
+                        || alphaInterp == HdInterpolationInstance)) {
                     const GfVec3f& clr3f = UsdMayaColorSpace::ConvertLinearToMaya(colorArray[0]);
                     const MColor   color(clr3f[0], clr3f[1], clr3f[2], alphaArray[0]);
-                    shader = _delegate->GetFallbackShader(color);
+                    shader = _delegate->GetFallbackShader(
+                        color); // The color of the fallback shader is ignored when the
```
HdC-adsk, line 1641: Having the comment after the code breaks the line into two. I think it is better to have the comment in a separate line.

lib/mayaUsd/render/vp2RenderDelegate/mesh.h
williamkrick: I don't like that because then if the primvar source is points it'll have a data member _extraInstanceColor with it.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
HdC-adsk: Actually the two branches have been there before I joined the USD project. Maybe @kxl-adsk remembers the history?

My guess is that if in the future we can have fine-grain update for instance transforms (e.g. one moving instance with all the other being static), the loop could be a better option. For now I don't know how it can be faster.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
HdC-adsk: Ok. I think the duplication was due to a limitation of pre-2022 releases in which consolidation and GPU instancing cannot be active for a render item at the same time. For 2022, we can make the branch simpler. But yes, it is better not to refactor that part in the pull request.

HdC-adsk (approved):

williamkrick:
@HdC-adsk For basisCurves I think you are correct, we need to make changes there too. I'm not sure if I would rather do the common parent class for mesh & basisCurves first or not. I'll spend some time this afternoon and decided to either make the common parent class or do basisCurves specific changes.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
williamkrick: I guess we can make it look better if we reduce the indentation on the left side, so it doesn't have to split lines. Maybe some of this can be moved into a separate function to make it more readable?

HdC-adsk (approved):

williamkrick:
There is still an outstanding issue. I'm working with USD 21.02 locally and the behavior of the tests in USD 20.05 and 20.02 is different. I need to figure out which version is producing desirable results (hopefully 21.02 is correct) and then decide what to do about it.

williamkrick:
@kxl-adsk found an issue with Maya 2020 + USD 20.05, I will investigate.

Issue #1215, opened 2021-02-26
cbegin/cend are only available in 2019 and later, making this fairly minor sanity check a breaking change


eoffermann:
https://github.com/Autodesk/maya-usd/blob/c820064d3ff4014aa3a0c81e9a15b720a6e1f64f/lib/usd/translators/jointWriter.cpp#L267-L272

As stated in the subject: MIntArray.cbegin/MIntArray.cend were introduced to the API in Maya2019 - if building against Maya2018, this simple debug code prevents the plugin from building. 

It would be a shame to halt support for 2018 for this.

kxl-adsk:
You discovered something that I was planning to post a message on Discussions. We indeed are stopping official support for Maya 2018. I will go ahead and make this post right away.

kxl-adsk:
The PR is open to fix Maya 2018 compilation. Past this, see https://github.com/Autodesk/maya-usd/discussions/1217 discussion.

PR #1225, opened 2021-03-01
Fix Maya 2018 compilation

kxl-adsk:
We are going to drop support for Maya 2018 as mentioned in https://github.com/Autodesk/maya-usd/discussions/1217. This PR fixes compilation before we officially remove the support from the https://github.com/Autodesk/maya-usd/blob/dev/doc/build.md documentation.


kxl-adsk:
This also fixes https://github.com/Autodesk/maya-usd/issues/1215

PR #1226, opened 2021-03-01
Separate callbacks for Save and Export

fowlertADSK:
The function that gathers the stages that might need to be written out has logic in it to check if we are doing Export vs. Export Selected, but the check for that state is incorrect.  The MFileIO::exportType check only tells us what option the last Export ran with and isn't changed when doing a Save.  We never want to check if something is selected during a Save, but it doesn't look like there is an explicit state kept by Maya so here we just create separate callbacks so we can keep track.

### DIFF
--- a/lib/mayaUsd/nodes/layerManager.cpp
+++ b/lib/mayaUsd/nodes/layerManager.cpp
@@ -194,12 +194,14 @@ public:
 
     static LayerDatabase& instance();
     static void           setBatchSaveDelegate(BatchSaveDelegate delegate);
-    static void           prepareForWriteCheck(bool*, void*);
+    static void           prepareForSaveCheck(bool*, void*);
+    static void           prepareForExportCheck(bool*, void*);
+    static void           prepareForWriteCheck(bool*, bool);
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -343,10 +355,10 @@ void LayerDatabase::prepareForWriteCheck(bool* retCode, void*)
     }
 }
 
-bool LayerDatabase::getStagesToSave()
+bool LayerDatabase::getStagesToSave(bool isExport)
 {
     auto allStages = MayaUsd::ufe::getAllStages();
-    bool checkSelection = (MFileIO::kExportTypeSelected == MFileIO::exportType());
+    bool checkSelection = isExport && (MFileIO::kExportTypeSelected == MFileIO::exportType());
```
fowlertADSK, line 361: We only want to check selection if we are in Export and the export mode is Export Selected.

seando-adsk (approved):
looks good

seando-adsk:
@fowlertADSK I restarted your preflight since the other one timed out and failed.

PR #1228, opened 2021-03-02
MAYA-110196 - Multiple ProxyShapes with Same Stage (InStageData) Bug

alsoussi:
We can have a situation where we have multiple proxy shapes that point to the same exact stage, using the InStageData port. Unfortunately this causes an issue in the layer editor since the combo-box is keeping track of the stage only, so two entries with the same stage is ambiguous. 

The fix was to move away from storing the Pixar Stage Reference as the data in the combo box and move to the StageEntry which encapsulates both stage and the proxy shape path.

### DIFF
--- a/lib/usd/ui/layerEditor/mayaSessionState.cpp
+++ b/lib/usd/ui/layerEditor/mayaSessionState.cpp
@@ -32,6 +32,7 @@
 #include <QtCore/QTimer>
 #include <QtWidgets/QMenu>
 
+
 #ifdef THIS
 #undef THIS
 #endif
@@ -59,19 +60,11 @@ MayaSessionState::~MayaSessionState()
     //
 }
 
-void MayaSessionState::setStage(PXR_NS::UsdStageRefPtr const& in_stage)
+void MayaSessionState::setStageEntry(StageEntry const& inEntry)
 {
-    PARENT_CLASS::setStage(in_stage);
-    if (in_stage) {
-        auto stageList = allStages();
-        for (auto const& entry : stageList) {
-            if (entry._stage == in_stage) {
-                _currentProxyShapePath = entry._proxyShapePath;
-                break;
-            }
-        }
-    } else {
-        _currentProxyShapePath.clear();
+    PARENT_CLASS::setStageEntry(inEntry);
+    if (!inEntry._stage) {
+        _currentStageEntry.clear();
     }
 }
 
@@ -155,25 +148,23 @@ void MayaSessionState::unregisterNotifications()
 
 void MayaSessionState::mayaUsdStageReset(const MayaUsdProxyStageSetNotice& notice)
 {
-    auto shapePath = notice.GetShapePath();
-    auto stage = notice.GetStage();
-    if (shapePath == _currentProxyShapePath) {
-        QTimer::singleShot(0, this, [this, stage]() { setStage(stage); });
-    } else {
-        auto shapePath = notice.GetShapePath();
-        QTimer::singleShot(
-            0, this, [this, shapePath, stage]() { mayaUsdStageResetCBOnIdle(shapePath, stage); });
+    auto       shapePath = notice.GetShapePath();
+    StageEntry entry;
+    if (getStageEntry(&entry, shapePath.c_str())) {
+        if (entry._proxyShapePath == _currentStageEntry._proxyShapePath) {
+            QTimer::singleShot(0, this, [this, entry]() {
+                mayaUsdStageResetCBOnIdle(entry);
+                setStageEntry(entry);
+            });
+        } else {
+            QTimer::singleShot(0, this, [this, entry]() { mayaUsdStageResetCBOnIdle(entry); });
+        }
     }
 }
 
-void MayaSessionState::mayaUsdStageResetCBOnIdle(
-    const std::string&            shapePath,
-    PXR_NS::UsdStageRefPtr const& stage)
+void MayaSessionState::mayaUsdStageResetCBOnIdle(StageEntry const& entry)
 {
-    StageEntry entry;
-    if (getStageEntry(&entry, shapePath.c_str())) {
-        Q_EMIT stageResetSignal(entry._proxyShapePath, stage);
-    }
+    Q_EMIT stageResetSignal(entry);
 }
 
 /* static */
@@ -195,7 +186,7 @@ void MayaSessionState::proxyShapeAddedCBOnIdle(const MObject& obj)
     auto       shapePath = dagPath.fullPathName();
     StageEntry entry;
     if (getStageEntry(&entry, shapePath)) {
-        Q_EMIT stageListChangedSignal(entry._stage);
+        Q_EMIT stageListChangedSignal(entry);
     }
 }
 
@@ -213,11 +204,12 @@ void MayaSessionState::nodeRenamedCB(MObject& obj, const MString& oldName, void*
         auto THIS = static_cast<MayaSessionState*>(clientData);
 
         // doing it on idle give time to the Load Stage to set a file name
-        QTimer::singleShot(0, [THIS, obj]() { THIS->nodeRenamedCBOnIdle(obj); });
+        QTimer::singleShot(
+            0, [THIS, obj, oldName]() { THIS->nodeRenamedCBOnIdle(oldName.asChar(), obj); });
     }
 }
 
-void MayaSessionState::nodeRenamedCBOnIdle(const MObject& obj)
+void MayaSessionState::nodeRenamedCBOnIdle(std::string const& oldName, const MObject& obj)
 {
     // this does not work:
     //        if OpenMaya.MFnDependencyNode(obj).typeName == PROXY_NODE_TYPE
@@ -228,7 +220,13 @@ void MayaSessionState::nodeRenamedCBOnIdle(const MObject& obj)
 
         StageEntry entry;
         if (getStageEntry(&entry, shapePath)) {
-            Q_EMIT stageRenamedSignal(entry._displayName, entry._stage);
+            // Need to update the current Entry also
+            if (_currentStageEntry._displayName == oldName
+                && _currentStageEntry._stage == entry._stage) {
+                _currentStageEntry = entry;
+            }
+
+            Q_EMIT stageRenamedSignal(oldName, entry);
         }
     }
 }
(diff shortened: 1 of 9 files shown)


lib/usd/ui/layerEditor/stageSelectorWidget.cpp
```diff
@@ -112,25 +112,40 @@ void StageSelectorWidget::selectedIndexChanged(int index)
 void StageSelectorWidget::sessionStageChanged()
 {
     if (!_internalChange) {
-        auto index = _dropDown->findData(QVariant::fromValue(_sessionState->stage()));
+        size_t index = -1;
+        for (size_t i = 0; i < _dropDown->count(); i++) {
```
alsoussi, line 116: This seems silly, ideally I should just do _dropdown->findData, but that doesn't work for some reason, I'm not a Qt expert, but i searched online and with structs it "should" be ok, but it doesn't seem to be.

luceric-adsk: hello. What makes the find work for the UsdStage is the macro at line 17 of this file
Q_DECLARE_METATYPE(pxr::UsdStageRefPtr);

you simply need to change that for the a StageEntry pointer instead

lib/usd/ui/layerEditor/mayaSessionState.cpp
```diff
@@ -59,19 +60,11 @@ MayaSessionState::~MayaSessionState()
     //
 }
 
-void MayaSessionState::setStage(PXR_NS::UsdStageRefPtr const& in_stage)
+void MayaSessionState::setStageEntry(StageEntry const& inEntry)
```
luceric-adsk, line 63: I would like this method to continue being called setStage, it's still the current stage what we're setting. The "stage entry' is one in the stage list. Here it's how we're going to specify the stage, but we're not really setting 'the stage entry'.

lib/usd/ui/layerEditor/sessionState.cpp
```diff
@@ -23,18 +23,19 @@ void SessionState::setAutoHideSessionLayer(bool hideIt)
     Q_EMIT autoHideSessionLayerSignal(_autoHideSessionLayer);
 }
 
-void SessionState::setStage(PXR_NS::UsdStageRefPtr const& in_stage)
+void SessionState::setStageEntry(StageEntry const& in_entry)
```
luceric-adsk, line 26: I would like this method to continue to be named setStage instead of set"StageEntry"

lib/usd/ui/layerEditor/sessionState.h
```diff
@@ -41,21 +41,70 @@ class SessionState : public QObject
...
+            std::string const&            proxyShapePath)
+        {
+            _stage = stage;
+            _displayName = displayName;
+            _proxyShapePath = proxyShapePath;
+        }
+
+        StageEntry(const StageEntry& entry)
+        {
+            _stage = entry._stage;
```
luceric-adsk, line 71: did you need to declare all of these? We get all the copy contructors and equality operators for free when we don't define any.  It's a lot of code to maintain if we don't need them

lib/usd/ui/layerEditor/stageSelectorWidget.cpp
luceric-adsk: sorry, I see you already tried that

lib/usd/ui/layerEditor/stageSelectorWidget.cpp
```diff
@@ -112,25 +112,40 @@ void StageSelectorWidget::selectedIndexChanged(int index)
...
 }
 
-void StageSelectorWidget::stageRenamed(std::string const& name, PXR_NS::UsdStageRefPtr const& stage)
+void StageSelectorWidget::stageRenamed(
+    std::string const&              oldName,
+    SessionState::StageEntry const& renamedEntry)
 {
-    auto index = _dropDown->findData(QVariant::fromValue(stage));
+    size_t index = -1;
+    for (size_t i = 0; i < _dropDown->count(); i++) {
```
luceric-adsk, line 135: QComboBox::count() return int and not size_t, so all of these would be wrong.  (in fact, size_t is unsigned, so comparision with -1 might build break on some platform)

lib/usd/ui/layerEditor/mayaSessionState.cpp
```diff
@@ -228,7 +220,13 @@ void MayaSessionState::nodeRenamedCBOnIdle(const MObject& obj)
 
         StageEntry entry;
         if (getStageEntry(&entry, shapePath)) {
-            Q_EMIT stageRenamedSignal(entry._displayName, entry._stage);
+            // Need to update the current Entry also
+            if (_currentStageEntry._displayName == oldName
```
luceric-adsk, line 224: There is no garantee that the _displayName is the name we're going to get from Maya from the notifications. 
Did you test this?  What about when a proxy is in a Namespace I don't think you should be doing anything with 'old name', we already have the MObject pointer from which we can get the UsdStage, and that's how we can make sure a StageEntry is for the Proxy that's been renamed.  This is why the previous code ignores the old name and does not send it in the signal. I don't think you should send it down the signal and use it to identify the Stage Entry later, it's error-prone

lib/usd/ui/layerEditor/stageSelectorWidget.cpp
```diff
@@ -112,25 +112,40 @@ void StageSelectorWidget::selectedIndexChanged(int index)
...
+        auto const& data = _dropDown->itemData(i);
+        auto        entry = data.value<SessionState::StageEntry>();
+        if (entry._displayName == oldName && entry._stage == renamedEntry._stage) {
```
luceric-adsk, line 138: I don't think you should compare the _displayName to the 'old name'.  See comment in stageRenamedSignal
Please only compare the stage. Maybe send down the proxyShapePath instead if you need it identify the StageEntry when more than one proxy use the stage stage (which is the bug you're fixing I guess). The reason is that the display name is just that, something I trimmed to fit in a combo

luceric-adsk (changes requested):

lib/usd/ui/layerEditor/mayaSessionState.cpp
alsoussi: I'm confused on what you mean. The class logic changed, instead of holding a pointer to stage, which is ambiguous in the case where you have two proxy shapes with the same stage, we are holding a stageEntry which has the proxyshape path and the stage (and the name). So yes we are setting the CurrentStageEntry, naming it just setStage when clearly its setting the Whole struct seems more confusing in my opinion.

lib/usd/ui/layerEditor/stageSelectorWidget.cpp
alsoussi: Good point! Nice catch!

lib/usd/ui/layerEditor/sessionState.h
alsoussi: Yup normally I agree with you, but I needed the equality ones since Qt wasn't happy with the default. 

The only extra one's are the copy and other constructor, that was just habit and convenience, but I can remove them.

lib/usd/ui/layerEditor/mayaSessionState.cpp
alsoussi: Yes of course I tested it and its working. The stage is not enough as a unique identifier, that's the whole point of this change. The previous code could ignore it since it doesn't matter. Since data which is stored in the combo box technically doesn't need updating (was just the stage), but in this new case it does, since the data in the combo box is the stage entry which also contains the name and the path.

Maybe an example could help. Lets assume you have the following proxyshapes (which are connected via the in stage data):
ProxyShape1 - Stage1
ProxyShape2 - Stage1

So lets assume to Rename "ProxyShape2" to "NewProxyShape2".

When you get to the part of the code which needs to update you will get two things:
oldName: ProxyShape2
newEntry: ("NewProxyShape2", "NewProxyShape2Path", Stage1)

So at this point I have to figure out which entry was changed and set the new data for it, since that data contains the name as well. And to find which entry in the list I need to know what the name use to be since if I just had the new Entry I would not be able to find it in the list. 

So since you are telling me that the displayname is not enough the other way to do this, but I'm not a fan of, is to use the proxyshapepath and ASSUME that the index of allStages maps to the dropBox list and just find the index of the NEW entry in that new list and just use that to update. I can make this change

lib/usd/ui/layerEditor/stageSelectorWidget.cpp
alsoussi: Fair enough, I responded there.

lib/usd/ui/layerEditor/sessionState.cpp
alsoussi: I guess its a style thing, but I think keeping the same name is confusing.

lib/usd/ui/layerEditor/mayaSessionState.cpp
alsoussi: So I got rid of the old name downstream, but couldn't get rid of it here. Let me know if you have any ideas.

fowlertADSK: To fix a different issue I was about to add another utility function in lib/mayaUsd/ufe/Utils.h to return the MDagPath of any given stage, that Utils class has access to the UsdStageMap we maintain.  A utility function like that might replace a lot of this wouldn't it?  The UsdStageMap should be able to correctly identify the dag object you need, but while testing my other issue I hadn't been thinking about multiple proxy shapes pointing to the same stage.  If the stage map doesn't solve your problem here then it also might not solve what I'm looking at in this situation.

lib/usd/ui/layerEditor/sessionState.cpp
luceric-adsk: it's setting the current stage, the StageEntry is just the 'how'. Like previously, it wasn't called "setStageRefPtr"

lib/usd/ui/layerEditor/mayaSessionState.cpp
luceric-adsk: Since I'm not certain of MdagPath rules and unicity, they way I handled this in Bifrost is I would use the node's Muuid and store that as a string identifier.  you get it with MFnDependencyNode::uuid and then can call asString on the MUIid.  So the StageEntry could have also have an ProxyShape ID field that's the MUuid, or its string representation, and that's a reliable way to id that two objects are the same.  So you could pass down that ID in your rename signal instead of the previous name

luceric-adsk: A second suggestion:  in this method, MayaSessionState::nodeRenamedCBOnIdle, you have the MObject, and if indeed you get MDagPath from it, you can then call MayaSessionState::getStageEntry to get a StageEntry for it which will run the same logic to generate the displayName, and then it should be much safer to compare the displayName

lib/usd/ui/layerEditor/sessionState.cpp
alsoussi: Yea but its not just setting the current stage, its setting the current stage, the proxy shape id the path and the display name, so calling it setStage implies its only doing one of those things.

lib/usd/ui/layerEditor/mayaSessionState.cpp
alsoussi: I updated to use the UUID as you suggested, feels much more reliable now, great suggestion!

luceric-adsk (approved):

seando-adsk (approved):

PR #1229, opened 2021-03-02
identify clang-format version 10.0.0 as the "standard" version for the project

mattyjams:
I realized that we discussed this standardization offline but didn't actually have it documented anywhere.


kxl-adsk (approved):

kxl-adsk:
Thank you, @mattyjams !

PR #1230, opened 2021-03-02
PXR_NAMESPACE_USING_DIRECTIVE removal from lib/mayaUsd/ufe.

ppt-adsk:
This pull request removes PXR_NAMESPACE_USING_DIRECTIVE statements from all files in lib/mayaUsd/ufe, in favor of using the explicit PXR_NS namespace for Pixar USD data types.  This ensures that headers don't propagate the using directive where they are themselves included.

### DIFF
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
@@ -34,6 +34,8 @@
 #include <mayaUsd/ufe/UsdUndoReorderCommand.h>
 #endif
 
+PXR_NAMESPACE_USING_DIRECTIVE
+
 namespace {
 UsdPrimSiblingRange getUSDFilteredChildren(
     const UsdPrim&               prim,
--- a/lib/mayaUsd/ufe/UsdSceneItemOps.h
+++ b/lib/mayaUsd/ufe/UsdSceneItemOps.h
@@ -23,8 +23,6 @@
 #include <ufe/path.h>
 #include <ufe/sceneItemOps.h>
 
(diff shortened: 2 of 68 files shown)


lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
```diff
@@ -34,6 +34,8 @@
 #include <mayaUsd/ufe/UsdUndoReorderCommand.h>
 #endif
 
+PXR_NAMESPACE_USING_DIRECTIVE
```
ppt-adsk, line 37: Simpler and more readable to keep the using directive in the implementation, where the number of changes is high.

lib/mayaUsd/ufe/UsdSceneItemOps.h
```diff
@@ -48,8 +46,9 @@ class MAYAUSD_CORE_PUBLIC UsdSceneItemOps : public Ufe::SceneItemOps
 
     void             setItem(const UsdSceneItem::Ptr& item);
     const Ufe::Path& path() const;
-    inline UsdPrim   prim() const
+    inline PXR_NS::UsdPrim prim() const
     {
+        PXR_NAMESPACE_USING_DIRECTIVE
```
ppt-adsk, line 51: Macros like TF_AXIOM have no namespace, so forced to use the using directive here.

ppt-adsk:
Pre-flight failure is master OSX python3 only, with the unfortunately familiar
hudson.remoting.ChannelClosedException: channel is already closed
error.  Should be good to go.

HamedSabri-adsk (approved):

PR #1233, opened 2021-03-04
MAYA-108770 - When an Xformable is added to a USD Prim, the Attribute…

spinell-adsk:
… Editor doesn't update automatically

-Fix AE refresh issue when transform changed on a prim

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
@@ -1,11 +1,38 @@
 import fnmatch
 import ufe
+import maya.mel as mel
+import maya.cmds as cmds
 import mayaUsd.ufe
 import maya.internal.common.ufe_ae.template as ufeAeTemplate
 
 # We manually import all the classes which have a 'GetSchemaAttributeNames'
 # method so we have access to it and the 'pythonClass' method.
 from pxr import Usd, UsdGeom, UsdLux, UsdRender, UsdRi, UsdShade, UsdSkel, UsdUI, UsdVol
+
+
+# Custom control, but does not have any UI. Instead we use
+# this control to be notified from UFE when any attribute has changed
+# so we can update the AE. This is to fix refresh issue
+# when transform is added to a prim.
+class UfeAttributesObserver(ufe.Observer):
+    def __init__(self, item):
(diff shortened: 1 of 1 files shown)


spinell-adsk:
Add a custom widget with no UI into the UFE attribute editor template.
The custom widget is a UFE attribute observer used to refresh the AE when the prim transform changed.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -1,12 +1,39 @@
...
+        super(UfeAttributesObserver, self).__init__()
+        self._item = item
+
+    def __del__(self):
+        ufe.Attributes.removeObserver(self)
+
+    def __call__(self, notification):
+        if isinstance(notification, ufe.AttributeValueChanged):
+            if notification.name() == "xformOpOrder":
+                mel.eval("evalDeferred(\"AEbuildControls\");")
```
spinell-adsk, line 28: using updateAE(node) force the prim tab to become the active one.
AEbuildControls Look to be a better choice to force the template to be rebuild

seando-adsk (approved):
Looks good to me.

seando-adsk:
@spinell-adsk / @kxl-adsk - I marked this as ready for merge. The only change is to the attribute editor template file which is not used in any test, so I see no need to run a preflight.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -1,12 +1,39 @@
...
+# Custom control, but does not have any UI. Instead we use
+# this control to be notified from UFE when any attribute has changed
+# so we can update the AE. This is to fix refresh issue
+# when transform is added to a prim.
+class UfeAttributesObserver(ufe.Observer):
+    def __init__(self, item):
+        super(UfeAttributesObserver, self).__init__()
+        self._item = item
+
+    def __del__(self):
```
ppt-adsk, line 22: There is nothing wrong with this, but it's not needed, strictly speaking.  The Subject keeps a weak ptr to its observers, and cleans up dead weak ptrs on notification, so that when an observer dies, the Subject will not notify it, and will end up removing it.

As a matter of fact, simply to avoid future developers copy-pasting this code, we should just remove it.

PR #1234, opened 2021-03-04
MAYA-109969 - No warning icon in USD Save Overwrite section.

spinell-adsk:
Add warning icon in USD Save Overwrite section


fowlertADSK (approved):
LGTM

PR #1239, opened 2021-03-08
Added namespace section to coding guidelines.

ppt-adsk:


HamedSabri-adsk (changes requested):
Thanks Pierre. Looking good. 

One small thing though, having "PXR_NAMESPACE_USING_DIRECTIVE" inside a function definition can be confusing. Not only that, there could be many instances of these macros in the same cpp file and having that macro before each one could make a lot of noise.

I would vote to put the directive outside of the function definition and call it just once. 

```
// this directive is needed for Pixar macros, which are not namespaced
PXR_NAMESPACE_USING_DIRECTIVE

inline PXR_NS::UsdPrim prim() const
{
    TF_AXIOM(fItem != nullptr);
    return fItem->prim();
}
```

ppt-adsk:
> 
> 
> Thanks Pierre. Looking good.
> 
> One small thing though, having "PXR_NAMESPACE_USING_DIRECTIVE" inside a function definition can be confusing. Not only that, there could be many instances of these macros in the same cpp file and having that macro before each one could make a lot of noise.
> 
> I would vote to put the directive outside of the function definition and call it just once.
> 
> ```
> // this directive is needed for Pixar macros, which are not namespaced
> PXR_NAMESPACE_USING_DIRECTIVE
> 
> inline PXR_NS::UsdPrim prim() const
> {
>     TF_AXIOM(fItem != nullptr);
>     return fItem->prim();
> }
> ```

This is in a .h file, so you can't...  Let me know if there's any way for me to make the explanation clearer.

HamedSabri-adsk (approved):
Got it!! You clearly say "In header files".  Looking good.

kxl-adsk (approved):

dj-mcg (approved):
Looks good to me, thanks Pierre!

pmolodo (approved):
Sounds good.  Only suggestion I might have would be regarding formatting - might help (for instance, with Hamed's confusion about .h vs .cpp) to have separate sections, and clearly make required vs recommended.  ie, something like:

```

### Namespaces

#### In header files (e.g. .h)

* **Required:** to use fully qualified namespace names. Global scope using directives are not allowed.  Inline code can use using directives in implementations, within a scope, when there is no other choice (e.g. when using macros, which are not namespaced).

    ```cpp
    // In aFile.h
    inline PXR_NS::UsdPrim prim() const
    {
        PXR_NAMESPACE_USING_DIRECTIVE
        TF_AXIOM(fItem != nullptr);
        return fItem->prim();
    }
    ```

#### In implementation files (e.g. .cpp)

* **Recommended:** to use fully qualified namespace names, unless clarity or readability is degraded by use of explicit namespaces, in which case a using directive is acceptable.
* **Recommended:** to use the existing namespace style, and not make gratuitous changes: if the file is using explicit namespaces, new code should follow this style, unless the changes are so significant that clarity or readability is degraded.  If the file has one or more using directives, new code should follow this style.
```

I'm not wedded to that, though, so I'm also happy to accept this as is.

ppt-adsk:
Two days have gone by, assuming that no further feedback is forthcoming.  Went with Paul's suggestion, ready for merge.

PR #1241, opened 2021-03-09
add support for manipulating point instances via UFE

mattyjams:
I've been having trouble finding the time to root out the remaining issues with point instance manipulation, so I wanted to just post what I've got so far. Feel free to review (read: ridicule) whenever you have a chance. :)

At a high level, this probably most closely aligns with the `XformCommonAPI` `Transform3d` implementation where instead of authoring translation, rotation, and scale values to xformOps dedicated to those operations, we author the values at the appropriate index in the `PointInstancer` prim's `positions`, `orientations`, and `scales` array attributes instead.

At the bottom, there is a `UsdPointInstanceModifierBase` class that handles most of the work of getting and setting individual element values in a `PointInstancer` prim's array attributes. This is then specialized into concrete subclasses for position, orientation, and scale modifiers.

Above that, there are a set of undoable commands that use the modifiers for redo'ing and undo'ing their respective operations.

There's then a `UsdTransform3dPointInstance` subclass of `Ufe::Transform3d` that creates commands of the appropriate type, and manages getting a point instance's current translation, rotation, and scale. It also supports querying the inclusive and exclusive matrices of an instance, and setting the instance's matrix (non-interactively).

One wrinkle is that for rotation, Maya/UFE speaks in terms of Euler angles while  the `PointInstancer` schema stores the `orientations` attribute as quaternions (`GfQuath`). I have an (admittedly ham-handed) conversion for that in the orientation modifier, but it makes for some pretty jittery manipulation as you drag the handles, and I don't think it's actually working correctly when multiple manipulations are strung together.

The scale manipulator seems to work at least for the initial manipulation, but I'm seeing an odd issue when pulling on one handle then pulling on another where the second pull often scales along *both* axes rather than only along the second handle's axis.

The position manipulator luckily seems to be working correctly though, so at least that seems to be a win.

I'm hopeful that the things that aren't working are just small mistakes in my implementation stemming from unfamiliarity with the interface. Curious to hear what you all think!

### DIFF
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -26,6 +26,9 @@
 #include <mayaUsd/undo/UsdUndoManager.h>
 #endif
 
+#include <pxr/pxr.h>
+#include <pxr/usd/usd/prim.h>
+#include <pxr/usd/usdGeom/pointInstancer.h>
 #include <pxr/usd/usdGeom/tokens.h>
 #include <pxr/usd/usdGeom/xformOp.h>
 
@@ -38,6 +41,7 @@
 #include <ufe/transform3d.h>
 
 #include <atomic>
+#include <limits>
 #include <vector>
 
 #ifdef UFE_V2_FEATURES_AVAILABLE
@@ -309,10 +313,59 @@ void StagesSubject::stageChanged(
 
         if (!InTransform3dChange::inTransform3dChange()) {
             // Is the change a Transform3d change?
+            const UsdPrim prim = stage->GetPrimAtPath(changedPath.GetPrimPath());
             const TfToken nameToken = changedPath.GetNameToken();
             if (nameToken == UsdGeomTokens->xformOpOrder || UsdGeomXformOp::IsXformOp(nameToken)) {
                 Ufe::Transform3d::notify(ufePath);
                 UFE_V2(sendValueChangedFallback = false;)
+            } else if (prim && prim.IsA<UsdGeomPointInstancer>()) {
+                // If the prim at the changed path is a PointInstancer, check
+                // whether the modified path is one of the attributes authored
+                // by point instance manipulation.
+                if (nameToken == UsdGeomTokens->orientations
+                    || nameToken == UsdGeomTokens->positions
+                    || nameToken == UsdGeomTokens->scales) {
+                    // This USD change represents a Transform3d change to a
+                    // PointInstancer prim.
+                    // Unfortunately though, there is no way for us to know
+                    // which point instance indices were actually affected by
+                    // this change. As a result, we must assume that they *all*
+                    // may have been affected, so we construct UFE paths for
+                    // every instance and issue a notification for each one.
+                    const UsdGeomPointInstancer pointInstancer(prim);
+
+#if PXR_VERSION >= 2011
+                    const size_t numInstances
+                        = bool(pointInstancer) ? pointInstancer.GetInstanceCount() : 0u;
+#else
+                    VtIntArray protoIndices;
+                    if (pointInstancer) {
+                        const UsdAttribute protoIndicesAttr = pointInstancer.GetProtoIndicesAttr();
+                        if (protoIndicesAttr) {
+                            protoIndicesAttr.Get(&protoIndices);
+                        }
+                    }
+                    const size_t numInstances = protoIndices.size();
+#endif
+
+                    // The PointInstancer schema can theoretically support as
+                    // as many instances as can be addressed by size_t, but
+                    // Hydra currently only represents the instanceIndex of
+                    // instances using int. We clamp the number of instance
+                    // indices to the largest possible int to ensure that we
+                    // don't overflow.
+                    const int numIndices
+                        = (numInstances <= static_cast<size_t>(std::numeric_limits<int>::max()))
+                        ? static_cast<int>(numInstances)
+                        : std::numeric_limits<int>::max();
+
+                    for (int instanceIndex = 0; instanceIndex < numIndices; ++instanceIndex) {
+                        const Ufe::Path instanceUfePath = stagePath(sender)
+                            + usdPathToUfePathSegment(changedPath.GetPrimPath(), instanceIndex);
+                        Ufe::Transform3d::notify(instanceUfePath);
+                    }
+                    UFE_V2(sendValueChangedFallback = false;)
+                }
             }
         }
 
--- a/lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
+++ b/lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
@@ -1,0 +1,281 @@
+//
+// Copyright 2021 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef MAYAUSD_UFE_USD_POINT_INSTANCE_MODIFIER_BASE_H
+#define MAYAUSD_UFE_USD_POINT_INSTANCE_MODIFIER_BASE_H
+
+#include <mayaUsd/base/api.h>
+
+#include <pxr/base/tf/diagnostic.h>
+#include <pxr/base/vt/array.h>
+#include <pxr/base/vt/types.h>
+#include <pxr/pxr.h>
+#include <pxr/usd/usd/attribute.h>
+#include <pxr/usd/usd/prim.h>
+#include <pxr/usd/usd/timeCode.h>
+#include <pxr/usd/usdGeom/pointInstancer.h>
+
+namespace MAYAUSD_NS_DEF {
+namespace ufe {
+
+/// Abstract utility class for accessing and modifying attributes of USD point
+/// instances.
+///
+/// In USD, a PointInstancer prim generates point instances based on data
+/// encoded as arrays in its attributes. The instance index of each point
+/// instance is used to index into these arrays to access the point instance's
+/// data.
+///
+/// This class provides an interface for getting and setting a point instance's
+/// attribute values in either the native USD type or in its equivalent UFE
+/// type. Derived classes of this base class must implement the functions to
+/// directly get and create the USD attribute on the PointInstancer (typically
+/// with a call to UsdGeomPointInstancer::Get...Attr() and
+/// UsdGeomPointInstancer::Create...Attr(), respectively), functions to convert
+/// values between USD and UFE types, and a function to retrieve the default
+/// value in the USD type.
+///
+template <class UfeValueType, class UsdValueType>
+class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
+{
+public:
+    UsdPointInstanceModifierBase()
+        : _prim()
+        , _instanceIndex(-1)
+    {
+    }
+
+    UsdPointInstanceModifierBase(const PXR_NS::UsdPrim prim, int instanceIndex)
+    {
+        setPrimAndInstanceIndex(prim, instanceIndex);
+    }
+
+    virtual ~UsdPointInstanceModifierBase() = default;
+
+    bool setPrimAndInstanceIndex(const PXR_NS::UsdPrim prim, int instanceIndex)
+    {
+        _prim = PXR_NS::UsdPrim();
+        _instanceIndex = -1;
+
+        if (!PXR_NS::UsdGeomPointInstancer(prim) || instanceIndex < 0) {
+            return false;
+        }
+
+        _prim = prim;
+        _instanceIndex = instanceIndex;
+
+        return true;
+    }
+
+    PXR_NS::UsdPrim prim() const { return _prim; };
+
+    PXR_NS::UsdGeomPointInstancer getPointInstancer() const
+    {
+        return PXR_NS::UsdGeomPointInstancer(_prim);
+    };
+
+    UfeValueType getUfeValue(PXR_NS::UsdTimeCode usdTime = PXR_NS::UsdTimeCode::Default()) const
+    {
+        const UsdValueType usdValue = getUsdValue(usdTime);
+        return convertValueToUfe(usdValue);
+    }
+
+    UsdValueType getUsdValue(PXR_NS::UsdTimeCode usdTime = PXR_NS::UsdTimeCode::Default()) const
+    {
+        UsdValueType usdValue = this->getDefaultUsdValue();
+
+        PXR_NS::UsdGeomPointInstancer pointInstancer = getPointInstancer();
+        if (!pointInstancer) {
+            return usdValue;
+        }
+
+        if (_instanceIndex < 0) {
+            return usdValue;
+        }
+
+        const size_t instanceIndex = static_cast<size_t>(_instanceIndex);
+
+        PXR_NS::UsdAttribute usdAttr = _getAttribute();
+        if (!usdAttr) {
+            return usdValue;
+        }
+
+        PXR_NS::VtArray<UsdValueType> usdValues;
+        if (!usdAttr.Get(&usdValues, usdTime)) {
+            return usdValue;
+        }
+
+        if (instanceIndex >= usdValues.size()) {
+            return usdValue;
+        }
+
+        // Avoid triggering a copy-on-write by making sure that we invoke
+        // operator[] on a const reference to the array.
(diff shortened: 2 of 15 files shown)


lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+namespace {
+
+using namespace MayaUsd;
+using namespace MayaUsd::ufe;
+
+// Class for setMatrixCmd() implementation.
+class UsdPointInstanceSetMatrix4dUndoableCmd
```
ppt-adsk, line 55: I know you're trying to implement the interface as completely as possible, but to my knowledge the setMatrixCmd() is called only by parent -absolute, and your instances cannot be reparented, can they?

I can't really think of another use case for this code --- you would need to have a workflow which would need an undoable command to set the instance's matrix, and currently I can't think of one; let me know if you think otherwise.

If that's the case, then this is essentially dead code --- and therefore, nice as it is :), we should probably take it out.

lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+/* override */
+Ufe::Vector3d UsdTransform3dPointInstance::scale() const { return _scaleModifier.getUfeValue(); }
+
+/* override */
+Ufe::TranslateUndoableCommand::Ptr
+UsdTransform3dPointInstance::translateCmd(double x, double y, double z)
+{
+    return std::make_shared<UsdPointInstanceTranslateUndoableCommand>(
+#ifdef UFE_V2_FEATURES_AVAILABLE
```
ppt-adsk, line 150: You added UsdTransform3dPointInstance.cpp to the UFE_V2 part of the CMakeLists.txt file, so either that's incorrect, or you can remove the UFE_V2_FEATURES_AVAILABLE ifdefs here.

lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp
```diff
@@ -0,0 +1,270 @@
...
+        usdSceneItem(),
+#endif
+        UsdTimeCode::Default());
+}
+
+/* override */
+Ufe::SetMatrix4dUndoableCommand::Ptr
+UsdTransform3dPointInstance::setMatrixCmd(const Ufe::Matrix4d& m)
+{
+    return std::make_shared<UsdPointInstanceSetMatrix4dUndoableCmd>(path(), m);
```
ppt-adsk, line 186: See above, where I question the need for this in practice.  If you feel there's a good use case for it, then you're good, otherwise I suggest some kind of TF_CODING_ERROR and returning nullptr.

lib/mayaUsd/ufe/UsdPointInstanceUndoableCommands.h
```diff
@@ -0,0 +1,216 @@
...
+
+/// Templated base class for undoable commands that manipulate USD point instances.
+///
+template <class BaseUfeUndoableCommand, class PointInstanceModifierType, class UsdValueType>
+class MAYAUSD_CORE_PUBLIC UsdPointInstanceUndoableCommandBase : public BaseUfeUndoableCommand
+{
+public:
+    using Ptr = std::shared_ptr<UsdPointInstanceUndoableCommandBase>;
+
+#ifdef UFE_V2_FEATURES_AVAILABLE
```
ppt-adsk, line 47: If UsdTransform3dPointInstance depends on UFE_V2, does it make sense to conditionally compile in UsdPointInstanceUndoableCommands for UFE_V1?

ppt-adsk (changes requested):
Looks really good to me from a code structure point of view!  Left a few comments / questions, and there is UFE_V2 conditional compilation work that needs a bit of clarification, but this is close to being ready, in my opinion.

I cannot tell at a glance why rotation and scaling manipulation are causing trouble.  I have had trouble in the past using matrix decomposition code from Pixar, and had to change to Maya API matrix decomposition to get proper manipulation --- see
https://github.com/Autodesk/maya-usd/blob/c56819050bed19e73370f31740ac6c51f7a5987c/lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp#L175
Possibly something to try later on; I don't feel we should delay converting this from a draft pull request to a pull request because of this issue.

lib/mayaUsd/ufe/UsdPointInstanceUndoableCommands.h
mattyjams: Good catch. I admittedly have only been working with this code against PR 122, so there's an excellent chance this doesn't even build successfully against Maya 2020 as it is right now.

I think my intent was to gracefully ignore attempts to manipulate point instances for UFE v1. The way things are currently setup, it should be possible in UFE v1 to have UFE selections that include scene items representing point instances, even though point instance selection in the viewport is UFE v2 only. In that case, I was aiming for requests to manipulate point instance items in UFE v1 basically becoming no-ops. That was why I had the modifiers and the commands included in the CMakeLists.txt for *both* UFE v1 and UFE v2.

So I think what I actually have to do is ensure that all of the new files here are compiled for both UFE versions. Does all of this sound right/reasonable to you?

lib/mayaUsd/ufe/UsdTransform3dHandler.cpp
```diff
@@ -46,6 +47,10 @@ Ufe::Transform3d::Ptr UsdTransform3dHandler::transform3d(const Ufe::SceneItem::P
     assert(usdItem);
 #endif
 
+    if (usdItem->isPointInstance()) {
```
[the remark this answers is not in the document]
mattyjams: Right, yeah. My main concern here was if you have multiple point instances for the same `PointInstancer` selected, we don't want to fall through to the "usual" `Transform3d` which may result in many objects that all try to treat their item as a discrete prim and attempt to create the same xformOps on the `PointInstancer`.

Do you think it's clearer or more safe to just not support point instances in this old handler? Maybe we're planning to get rid of it soon anyway?

lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp
mattyjams: Ahh, that explains some things actually, because I could figure out a way to get Maya to invoke this command. :)

That's true that it's not possible to re-parent point instances. I was thinking with this that maybe if you selected a point instance and then ran a `cmds.xform()` command to try to directly set the point instance's matrix that it might bottom out in this command, but that didn't seem to be the case.

I'll just remove this then if there's no way we can actually end up calling this command.

lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp
```diff
@@ -0,0 +1,270 @@
...
+    {
+        UsdSceneItem::Ptr usdItem = downcast(sceneItem());
+
+        if (!usdItem || !usdItem->isPointInstance()) {
+            return;
+        }
+
+        const UsdPrim& usdPrim = usdItem->prim();
+        const int      instanceIndex = usdItem->instanceIndex();
+        _positionModifier.setPrimAndInstanceIndex(usdPrim, instanceIndex);
```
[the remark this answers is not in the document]
mattyjams: Yeah, there's a bit of wishful thinking baked in here.

I'm hoping that at some point we might have some additional API to make these single-element-of-an-array type reads and writes more efficient. One idea was that if we could make the modifiers more active observers of their attribute, the modifier could actually just directly hold an array value. In that case, the modifier would just update its own array and then use that to `Set()` the `UsdAttribute`. There would be a bunch more work needed though to make sure that the modifier's copy of the array value stays in sync with what's in USD, especially when other operations might manipulate the attribute in USD.

The other thing I was considering was whether some kind of manipulation "batching" API might be desirable in UFE at some point, where we could create a single `Transform3d` that represents the same manipulation operation applied to multiple scene items. That could be another case where the modifier holds onto a whole array, the batched `Transform3d` could make a whole bunch of updates to it, and those could then be "committed" to the `UsdAttribute` when the manipulation is done.

I'm not sure whether any of those are good reasons for it to stay the way it is now since none of that currently exists, but hopefully helps explain how I ended up here.

lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp
mattyjams: Yup, thanks for this. As noted above, I think the former is likely incorrect and `UsdTransform3dPointInstance` should get compiled in for both versions of UFE.

lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp
mattyjams: Doesn't sound like there is a good use case currently, so I'll do that, thanks!

mattyjams (commented):
Thanks for the feedback, @ppt-adsk! All great notes.

I'll see if I can "finish" this by addressing what you've noted so far, and add at least a few simple tests for the new functionality. I'll then convert it to a non-draft PR.

I'm a little concerned that the unit tests wouldn't actually cover the issues I was seeing when interactively pulling on the manipulator handles, since just doing `cmds.move()`, `cmds.rotate()`, etc. seem to work. Thanks for noting the issues you saw with the USD-side `Decompose()` though. I can give the Maya-side versions of that a try and see  whether that makes a difference.

I'm also still a little concerned about how performance scales with having a `Transform3d` per point instance scene item and the the single-element modifications that each of those are doing. With the API that's currently available on both the USD side and the UFE side, I'm not sure there's much of a way around that at the moment.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -0,0 +1,279 @@
...
+
+        const size_t instanceIndex = static_cast<size_t>(_instanceIndex);
+
+        PXR_NS::UsdAttribute usdAttr = _getAttribute();
+        if (!usdAttr) {
+            return usdValue;
+        }
+
+        PXR_NS::VtArray<UsdValueType> usdValues;
+        if (!usdAttr.Get(&usdValues, usdTime)) {
```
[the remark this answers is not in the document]
mattyjams: Addressed this here:
https://github.com/Autodesk/maya-usd/pull/1241/commits/ee4b1f2d401e3a7dd45d1f3cff917ecd64e34656#diff-949fb6c03eacb42b0f81ca6117be20a35ac50525a233783622059ad8281c3ce2R126

lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp
mattyjams: Removed in rebased commit ac913871efe412ad175fa1f1adfc2850b7ff56f2.

lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp
mattyjams: With the command class removed, this now issues a coding error and returns `nullptr` instead:
https://github.com/Autodesk/maya-usd/pull/1241/commits/ac913871efe412ad175fa1f1adfc2850b7ff56f2#diff-80faaef9cd24f227b2c4ada0deab537b01871c846af3c0cfa1aeb66d5b9dafd5R124

mattyjams:
I got some great notes on this offline from @kxl-adsk and @ppt-adsk, so the newly pushed changes here seem to resolve all of the issues I was previously seeing. There were two important changes necessary:

- The UFE `StagesSubject` needed to monitor for change notifications from USD when paths to `PointInstancer` schema `positions`, `orientations`, or `scales` attributes were modified. It was already doing something similar for `xformOpOrder` and specific xformOp attributes for the other transform types. That addition is represented in 15ab1dee0754b35c6185b43a54dbd1e6113e1944. The somewhat unsavory aspect of that though is that we can't tell which specific instances were updated, so we have to issue notifications for *all* of them.
- The value conversion functions for the orientations modifier were rewritten to use Maya's rotation and transform handling utilities rather than USD/`Gf`'s. The manipulation is much smoother now when dragging the handles in the viewport.

I still need to validate this against UFE v1 and then finish up the unit tests, but it looks to be in much better shape now. Should hopefully be ready to convert this to a regular pull request soon.

lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp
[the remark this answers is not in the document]
ppt-adsk: Yes, I think batching the write on manipulation end is the way to go.  How we get there is unclear, so let's keep what you have for now, as we don't have any definite proposals to improve upon it.

lib/mayaUsd/ufe/UsdPointInstanceUndoableCommands.h
ppt-adsk: If UFE v1 is intended to be a no-op, then it seems to me that you could just return null pointers when you're asked for commands.  That way, you wouldn't need to compile commands against UFE v1 --- and if the commands are the only clients of the modifiers, then the modifiers wouldn't need to be compiled against UFE v1 either.  Does this align with your thinking?

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -309,10 +312,58 @@ void StagesSubject::stageChanged(
...
+            } else if (prim && prim.IsA<UsdGeomPointInstancer>()) {
+                // If the prim at the changed path is a PointInstancer, check
+                // whether the modified path is one of the attributes authored
+                // by point instance manipulation.
+                if (nameToken == UsdGeomTokens->orientations
+                    || nameToken == UsdGeomTokens->positions
+                    || nameToken == UsdGeomTokens->scales) {
+                    // This USD change represents a Transform3d change to a
+                    // PointInstancer prim.
+                    // Unfortunately though, there is no way for us to know
```
ppt-adsk, line 329: Ouch.  We have to find a way to fix this.  Certainly the Transform3d interface knows which point instance index it's dealing with.  There has to be a way to leverage that.

ppt-adsk (commented):
Notification scheme is the more troublesome part, but at least we have correctness, if not performance :)  Perhaps given the fairly urgent time constraints this is sufficient, and we can move to creating a test and converting this draft to an actual pull request?

I'm still hopeful that we can provide a no-op implementation of the point instancer Transform3d for UFE v1 (i.e. Maya 2020) by returning null pointers instead of compiling actual code --- please set me straight on this if I've misunderstood.

lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp
mattyjams: As it turns out, `UsdTransform3dPointInstance` inherits from `UsdTransform3dBase`, which only gets compiled for UFE v2, so I updated the CMakeLists.txt to only build in `UsdTransform3dPointInstance` for UFE v2 as well.

lib/mayaUsd/ufe/UsdPointInstanceUndoableCommands.h
mattyjams: It does, thanks! In the most recent changes, `UsdPointInstanceUndoableCommands.h` is only included for UFE v2.

I ended up leaving the modifiers in for both versions of UFE, since strictly speaking there's nothing in them that requires v2, so they could theoretically get used for non-Transform3d authoring operations.

lib/mayaUsd/ufe/StagesSubject.cpp
mattyjams: I think the most recent changes largely address this. The point instance manipulation commands now block this handling from running using the existing UfeNotifGuard mechanism, and they issue the Transform3d notifications themselves during `undo()` and `redo()`:
https://github.com/Autodesk/maya-usd/pull/1241/commits/b6575883e4b9d76786970459eebc114ec4d850f6#diff-26f9c9bdd71eb3cbf8eb499ab811fe928d671f8f6038293209b8dd08fce76dc2R73

This all instances blast still needs to remain though for any "external" authoring that happens in USD.

mattyjams:
I converted this to a regular PR, as I believe what we've got now is functionally correct and the major performance concerns we can currently tackle have been addressed. Tests are still missing, but I'll append another commit for that soon.

Thanks again @ppt-adsk and @kxl-adsk for all of your help with this!

mattyjams:
And now with tests in e4d1e200f989edf1b1c6d04525544656d4cc1e70. Kicking off a preflight run.

Thanks again!

mattyjams:
There was one more remaining issue in the unit test when running with Python 3 which was addressed in d5045198934994270e45991d805bd1a926a0344d. Other than that, I think the remaining preflight failures are the spurious kind we've seen before.

kxl-adsk:
I also run the remaining set of tests. All that is left is the review and we can merge it.

kxl-adsk (approved):
Looks very good. Thank you @mattyjams !

ppt-adsk (approved):
Great work Matt!  Thanks for sticking with this and bringing it over the finish line.

ppt-adsk:
Agree with Matt, the single failing Maya 2020 Windows python 2 test run appears to be build pipeline related, all tests pass.

PR #1242, opened 2021-03-09
MAYA-107380 - On the stage AE template I'd like to load/unload all payloads

seando-adsk:
MAYA-107380 - On the stage AE template I'd like to load and unload all payloads

* Reworked solution to avoid using UFE v2 and ufeSupport plugin as neither of these are present in Maya 2020. Created simple stage load/unload Maya commands.


seando-adsk (commented):
@ppt-adsk As we discussed, I created two Maya commands to load all and unload all payloads. Then I call these commands directly from the Maya-Usd proxy shape template (MEL). I removed the previous python code I added using the ContextOps and ufeSupport.

PR #1243, opened 2021-03-09
Remove/replace all UFE_PREVIEW_VERSION_NUM

seando-adsk:
Remove/replace all UFE_PREVIEW_VERSION_NUM now that latest MayaPR has UFE v2.

### DIFF
--- a/test/lib/ufe/testObject3d.py
+++ b/test/lib/ufe/testObject3d.py
@@ -321,7 +321,7 @@ class Object3dTestCase(unittest.TestCase):
         self.assertFalse(ufe.Object3d.hasObserver(visObs))
         self.assertEqual(ufe.Object3d.nbObservers(), 0)
 
-    @unittest.skipIf(ufeUtils.ufeFeatureSetVersion() >= 2, 'testUndoVisibleCmd only available in UFE v2 or greater.')
+    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() >= 2, 'testUndoVisibleCmd only available in UFE v2 or greater.')
     def testUndoVisibleCmd(self):
 
         ''' Verify the token / attribute values for visibility after performing undo/redo '''
@@ -386,7 +386,7 @@ class Object3dTestCase(unittest.TestCase):
         # capsuleItem must be invisible now
         self.assertFalse(object3d.visibility())
 
-    @unittest.skipIf(ufeUtils.ufeFeatureSetVersion() >= 2, 'testMayaHideAndShowHiddenUndoCommands only available in UFE v2 or greater.')
+    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() >= 2, 'testMayaHideAndShowHiddenUndoCommands only available in UFE v2 or greater.')
     def testMayaHideAndShowHiddenUndoCommands(self):
         ''' Verify the token / attribute values for visibility via "hide", "showHidden" commands + Undo/Redo '''
 
--- a/test/lib/ufe/testObject3d.py
+++ b/test/lib/ufe/testObject3d.py
@@ -18,6 +18,7 @@
 
 import fixturesUtils
 import mayaUtils
+import ufeUtils
 from testUtils import assertVectorAlmostEqual
 from testUtils import assertVectorEqual
 import usdUtils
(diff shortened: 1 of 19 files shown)


seando-adsk (commented):
Now that the latest Maya PR has UFE v2, I removed and/or replaced all the UFE_PREVIEW_VERSION_NUM with UFE v2 checks.

mattyjams (approved):
For what it's worth, this looks great to me! Thanks for doing this, Sean!

I'm assuming we think we may still want/need to use `mayaUtils.previewReleaseVersion()`, so that's sticking around? There are a handful of those in the attribute editor code and in tests that check for PR 121 or earlier, so I was curious whether those could be either removed or replaced with UFE version-based checks instead.

test/lib/ufe/testObject3d.py
```diff
@@ -320,7 +321,7 @@ def testVisibility(self):
         self.assertFalse(ufe.Object3d.hasObserver(visObs))
         self.assertEqual(ufe.Object3d.nbObservers(), 0)
 
-    @unittest.skipIf(os.getenv('UFE_PREVIEW_VERSION_NUM', '0000') < '2034', 'testUndoVisibleCmd is only available in UFE preview version 0.2.34 and greater')
+    @unittest.skipIf(ufeUtils.ufeFeatureSetVersion() >= 2, 'testUndoVisibleCmd only available in UFE v2 or greater.')
```
ppt-adsk, line 324: Wrong way around --- should be skipUnless(), or < 2.

test/lib/ufe/testObject3d.py
```diff
@@ -385,7 +386,7 @@ def testUndoVisibleCmd(self):
         # capsuleItem must be invisible now
         self.assertFalse(object3d.visibility())
 
-    @unittest.skipIf(os.getenv('UFE_PREVIEW_VERSION_NUM', '0000') < '2034', 'testMayaHideAndShowHiddenUndoCommands is only available in UFE preview version 0.2.34 and greater')
+    @unittest.skipIf(ufeUtils.ufeFeatureSetVersion() >= 2, 'testMayaHideAndShowHiddenUndoCommands only available in UFE v2 or greater.')
```
ppt-adsk, line 389: Same here.

ppt-adsk (changes requested):

test/lib/ufe/testObject3d.py
seando-adsk: Ah yes the dreaded skipIf vs skipUnless. Thanks for catching this.

seando-adsk:
@mattyjams Pierre and I looked at the `mayaUtils.previewReleaseVersion()` function and yes it needs to stay. It will for sure be needed in the future. However some of these existing checks can be cleaned up and changed into a Maya major version check. I've added a new function for that. This will make it clearer.

test/testUtils/mayaUtils.py
```diff
@@ -232,7 +232,12 @@ def previewReleaseVersion():
...
         return 110
+    elif majorVersion == 2022:
+        return 122
 
     match = prRe.match(cmds.about(v=True))
 
     return int(match.group(1)) if match else sys.maxsize
+
+def mayaMajorVersion():
+    return int(cmds.about(majorVersion=True))
```
seando-adsk, line 243: Trivial - but new helper to return Maya major version. So all checks can go thru mayaUtils/ufeUtils.

seando-adsk:
FYI, I did not bother changing the maya pr check in the AE template files as these are all deprecated and will be removed (very soon in another PR by me).

ppt-adsk (approved):
Thanks for doing this work Sean.

test/testUtils/mayaUtils.py
```diff
@@ -232,7 +232,12 @@ def previewReleaseVersion():
         return 98
     elif majorVersion == 2020:
         return 110
+    elif majorVersion == 2022:
+        return 122
```
mattyjams, line 236: I'm curious about this, but maybe I'm misunderstanding.

For the next (hypothetical) preview releases 123, 124, 125, etc., wouldn't they all report their majorVersion as  `2022`, in which case this would return `122` for all of those? I can see how the special cases for 2019 and 2020 are needed since we don't expect there to be any preview releases with those major versions anymore.

mattyjams (approved):
There was one thing I was unsure about in `mayaUtils.py` where a `majorVersion == 2022` early return condition was added, but otherwise looks great to me.

Thanks again, Sean!

test/testUtils/mayaUtils.py
seando-adsk: Yes you are correct, I should not have added these lines. I also jumped the gun on removing the PR123 checks. I realized that PR 123 isn't released yet. I've reverted some of my changes.

mattyjams (approved):
Looks good to me! Thanks again, @seando-adsk!

ppt-adsk (approved):

PR #1245, opened 2021-03-10
Minimize name-based lookup of proxy shape nodes.

ppt-adsk:
The UsdStageMap was performing name-based lookup of proxy shape nodes on every access.  This has been replaced with a scheme that uses an unordered map index by Ufe::Path, which are hashable.

### DIFF
--- a/lib/mayaUsd/ufe/UsdStageMap.cpp
+++ b/lib/mayaUsd/ufe/UsdStageMap.cpp
@@ -23,10 +23,6 @@
 
 #include <maya/MFnDagNode.h>
 
-#ifdef UFE_V2_FEATURES_AVAILABLE
-#include <ufe/pathString.h>
-#endif
-
 #include <cassert>
 
 PXR_NAMESPACE_USING_DIRECTIVE
@@ -195,11 +191,7 @@ void UsdStageMap::rebuildIfDirty()
 
     for (const auto& psn : ProxyShapeHandler::getAllNames()) {
         addItem(
-#ifdef UFE_V2_FEATURES_AVAILABLE
-            Ufe::PathString::path(psn));
-#else
             Ufe::Path(Ufe::PathSegment("|world" + psn, getMayaRunTimeId(), '|')));
-#endif
     }
     fDirty = false;
 }
--- a/lib/mayaUsd/ufe/UsdStageMap.cpp
+++ b/lib/mayaUsd/ufe/UsdStageMap.cpp
@@ -190,8 +190,7 @@ void UsdStageMap::rebuildIfDirty()
         return;
 
     for (const auto& psn : ProxyShapeHandler::getAllNames()) {
-        addItem(
-            Ufe::Path(Ufe::PathSegment("|world" + psn, getMayaRunTimeId(), '|')));
+        addItem(Ufe::Path(Ufe::PathSegment("|world" + psn, getMayaRunTimeId(), '|')));
     }
     fDirty = false;
 }
--- a/lib/mayaUsd/ufe/UsdStageMap.cpp
+++ b/lib/mayaUsd/ufe/UsdStageMap.cpp
@@ -112,16 +112,22 @@ void UsdStageMap::addItem(const Ufe::Path& path)
         return;
     }
 
-    // We've just done a name-based lookup of the proxy shape, so the stage
-    // cannot be null.
-    auto stage = objToStage(proxyShape.object());
+    // Non-const MObject& requires an lvalue.  We've just done a name-based
+    // lookup of the proxy shape, so the stage cannot be null.
+    auto obj = proxyShape.object();
+    auto stage = objToStage(obj);
     TF_AXIOM(stage);
 
     fPathToObject[path] = proxyShape;
     fStageToObject[stage] = proxyShape;
 }
 
-UsdStageWeakPtr UsdStageMap::stage(const Ufe::Path& path) { return objToStage(proxyShape(path)); }
+UsdStageWeakPtr UsdStageMap::stage(const Ufe::Path& path)
+{
+    // Non-const MObject& requires an lvalue.
+    auto obj = proxyShape(path);
+    return objToStage(obj);
+}
 
 MObject UsdStageMap::proxyShape(const Ufe::Path& path)
 {
--- a/lib/mayaUsd/ufe/UsdStageMap.cpp
+++ b/lib/mayaUsd/ufe/UsdStageMap.cpp
@@ -131,6 +131,8 @@ UsdStageWeakPtr UsdStageMap::stage(const Ufe::Path& path)
 
 MObject UsdStageMap::proxyShape(const Ufe::Path& path)
 {
+    rebuildIfDirty();
+
     const auto& singleSegmentPath
         = nbPathSegments(path) == 1 ? path : Ufe::Path(path.getSegments()[0]);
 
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/ufe/UsdStageMap.cpp
```diff
@@ -85,32 +110,53 @@ void UsdStageMap::addItem(const Ufe::Path& path, UsdStageWeakPtr stage)
...
+    TF_AXIOM(stage);
+
+    fPathToObject[path] = proxyShape;
     fStageToObject[stage] = proxyShape;
 }
 
-UsdStageWeakPtr UsdStageMap::stage(const Ufe::Path& path)
+UsdStageWeakPtr UsdStageMap::stage(const Ufe::Path& path) { return objToStage(proxyShape(path)); }
+
+MObject UsdStageMap::proxyShape(const Ufe::Path& path)
```
ppt-adsk, line 130: This is the engine method for the stage() and proxyShapeNode() methods.  It indexes into the unordered map of MObjectHandle's to proxy shapes using the Ufe::Path hash.  This is how we avoid name-based lookups.

lib/mayaUsd/ufe/UsdStageMap.cpp
```diff
@@ -85,32 +110,53 @@ void UsdStageMap::addItem(const Ufe::Path& path, UsdStageWeakPtr stage)
...
 {
+    const auto& singleSegmentPath
+        = nbPathSegments(path) == 1 ? path : Ufe::Path(path.getSegments()[0]);
+
+    auto iter = fPathToObject.find(singleSegmentPath);
+    if (iter != std::end(fPathToObject)) {
+        return iter->second.object();
+    }
+
+    // Did not find object in cache, either because the cache is stale, or
```
ppt-adsk, line 140: If the lookup fails, the cache might be stale.  Rebuild and try again.

lib/mayaUsd/ufe/UsdStageMap.cpp
```diff
@@ -85,32 +110,53 @@ void UsdStageMap::addItem(const Ufe::Path& path, UsdStageWeakPtr stage)
...
 
-    auto proxyShape = proxyShapeHandle(path);
-    if (!proxyShape.isValid()) {
+    // At this point the cache is rebuilt, so lookup failure means the object
+    // doesn't exist.
+    iter = fPathToObject.find(singleSegmentPath);
+    return iter == std::end(fPathToObject) ? MObject() : iter->second.object();
+}
+
+MayaUsdProxyShapeBase* UsdStageMap::proxyShapeNode(const Ufe::Path& path)
```
ppt-adsk, line 151: New accessor to the cache for the proxy shape node pointer.  Allows us to retrieve data from the proxy shape node, such as time, or enabled purposes.

lib/mayaUsd/ufe/UsdStageMap.cpp
```diff
@@ -147,12 +193,13 @@ void UsdStageMap::rebuildIfDirty()
...
         return;
 
-    auto proxyShapeNames = ProxyShapeHandler::getAllNames();
-    for (const auto& psn : proxyShapeNames) {
-        MDagPath  dag = UsdMayaUtil::nameToDagPath(psn);
-        Ufe::Path ufePath = dagPathToUfe(dag);
-        auto      stage = ProxyShapeHandler::dagPathToStage(psn);
-        addItem(ufePath, stage);
+    for (const auto& psn : ProxyShapeHandler::getAllNames()) {
+        addItem(
```
ppt-adsk, line 197: Rebuilding is now simpler, name-based lookup now done by addItem.

lib/mayaUsd/ufe/UsdStageMap.h
```diff
@@ -27,33 +27,31 @@
...
-        order of notification of Ufe observers.  An earlier implementation with
-        rename observation had the Maya Outliner (which observes rename) access the
-        UsdStageMap on rename before the UsdStageMap had been updated.
+    Two-way map of proxy shape UFE path to corresponding stage.
+
+    We will assume that	a USD proxy shape will not be instanced (even though
+    nothing in the data model prevents it).  To generalized access to the
+    underlying node, we store an MObjectHandle in the maps.
+
+    The cache is refreshed on access to a stage given a path which cannot be
```
ppt-adsk, line 47: Still wish to avoid having the cache observe the data model, to avoid observation cost as well as order of notification problems.  Rebuild on access failure is simpler to code, and access failure through incorrect input is most likely not a high-frequency code path.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -306,33 +306,12 @@ UsdTimeCode getTime(const Ufe::Path& path)
...
-    } else {
-        // Not found in the cache, or no longer valid.  Get the proxy shape
-        // MObject from its path, and put it in the cache.  Pop the head of the
-        // UFE path to get rid of "|world", which is implicit in Maya.
-        auto proxyShapeDagPath = UsdMayaUtil::nameToDagPath(proxyShapePath.popHead().string());
-        TF_VERIFY(proxyShapeDagPath.isValid());
-        proxyShapeObj = proxyShapeDagPath.node();
-        cache = std::pair<Ufe::Path, MObjectHandle>(proxyShapePath, MObjectHandle(proxyShapeObj));
+    // Proxy shape node should not be null.
+    auto proxyShape = g_StageMap.proxyShapeNode(path);
```
ppt-adsk, line 310: Removed homebrew, single node caching in favor of using the global proxy shape cache.

lib/mayaUsd/ufe/UsdStageMap.cpp
kxl-adsk: This is surprising. We should have proper invalidation of the cache on data model changes. Maybe this refactoring changed something and now we rebuild the cache too early (i.e. cause compute during dirty propagation) - this should be considered as a bug.

 Did you happen to find a case where this explicit setDirty was needed?

ppt-adsk: The only invalidation on data model change we do is for file new, which is done here:
https://github.com/Autodesk/maya-usd/blob/9a9642151be2d5d50e647abc3a5a8bc85c48b2b2/lib/mayaUsd/ufe/StagesSubject.cpp#L195
The version of the stage cache I am changing does not do scene observation for reparenting and renaming --- and the reason it doesn't is that it always does name-based lookup.  Once it does that, it obtains an MObjectHandle, and uses that as a key to index into a map to retrieve the stage.  However, this "caching" is not very effective: it is trivial to retrieve the stage once you have the MObject.

The version of the cache I am proposing does not observe the data model for rename and reparent either, and the rationale is explained here:
https://github.com/Autodesk/maya-usd/blob/538b360ba7f6be6df30ab64cb189750e493d92ee/lib/mayaUsd/ufe/UsdStageMap.h#L47
Essentially, we completely avoid order of notification problems with the update on demand scheme I am proposing, and it's much simpler as well.

kxl-adsk: The name may be here confusing, but we actually observe the data model and invalidate when stage gets invalidated during dirty propagation (or parallel evaluation) - https://github.com/Autodesk/maya-usd/blob/9a9642151be2d5d50e647abc3a5a8bc85c48b2b2/lib/mayaUsd/ufe/StagesSubject.cpp#L450

ppt-adsk: Ah, thanks for the tip, yes the naming is indeed a bit confusing.  Does this handle renaming and reparenting?  My intuition is that it does not, because this is purely a DG invalidation mechanism, in which case I believe my stale cache handling is still required.

kxl-adsk (approved):

ppt-adsk:
Will have QA try this branch out quickly before we merge.

ppt-adsk:
David Santos tested this branch manually, and ran QATA tests, so good to go.

PR #1248, opened 2021-03-12
MAYA-106484 add single-channel half-float EXR support.

pierrebai-adsk:
Handle teh case where we receive a single-channel half-float texture. The support in Maya already existeed, only the correspondence in the switch/case was missing.


HdC-adsk (changes requested):
Please mirror the change to older USD versions in the other conditional compilation block.

pierrebai-adsk:
It was already done, that is why I did not need to add it.

HdC-adsk (approved):

PR #1252, opened 2021-03-15
MAYA-110332 Add a camera test which relies on unimplemented methods.

williamkrick:
Adds a test which uses unimplemented methods on UsdTransform3dSetObjectMatrix & fails because the methods are not implemented.


ppt-adsk (approved):
Test looks good to me.

williamkrick (commented):
It looks good to me, I think we can merge it.

PR #1254, opened 2021-03-15
MAYA-110410 Author a default value for color & opacity if none are present.

williamkrick:


HdC-adsk:
LGTM.

@kxl-adsk FYI. This fix seems to have some connection to the issue Luc and you found last week when testing a scene without authored color?

kxl-adsk:
@HdC-adsk I don't think it will solve the problem. The solution was to author display color...but we haven't author opacity in the regression test.

williamkrick:
@HdC-adsk does that other issue block this one? Maybe you just forgot to approve the review?

HdC-adsk (approved):
Not blocking. I was just waiting for Krystian to confirm.

PR #1255, opened 2021-03-15
MAYA-104490 Add a basic test for Vp2RenderDelegate geom subset material binding support.

williamkrick:


HdC-adsk:
It looks to me the test scene can be added to RTS without having to add the test script separately since you only capture the image of unselected case. Why do you choose to do it this way?

williamkrick:
> 
> 
> It looks to me the test scene can be added to RTS without having to add the test script separately since you only capture the image of unselected case. Why do you choose to do it this way?

I like the test here for two reasons: it is in a public place so our partners can run it when validating their own changes and because these tests can run before a change is merged, while RTS runs after. The sooner we get the results from the test the better!

Right now the interactive tests (including this one) are not run as a part of the preflight before a merge, but we're hoping to get that working eventually.

HdC-adsk:
That is convincing. Thanks for clarification.

HdC-adsk (approved):

kxl-adsk:
And just to bring one more argument to the mix. Regression tests are run across multiple USD versions and Maya versions. RTS doesn't have this setup.

PR #1258, opened 2021-03-15
Part 1: Block attribute(s) authoring in weaker layer(s)

HamedSabri-adsk:
**Part 1:** Initial PR to add support for blocking "Transformation attribute authoring via manipulators" and "Single attribute authoring" in weaker layers.

e.g Single Attribute:

single_attr_block

e.g Transformation via manipulators:

manipulation_attr_block

### DIFF
--- a/lib/mayaUsd/ufe/UsdAttribute.cpp
+++ b/lib/mayaUsd/ufe/UsdAttribute.cpp
@@ -43,7 +43,6 @@
 //       in USD.
 // PXR_NAMESPACE_USING_DIRECTIVE
 
-static constexpr char kErrorMsgFailedSet[] = "Failed to set USD attribute with new value";
 static constexpr char kErrorMsgFailedConvertToString[]
     = "Could not convert the attribute to a string";
 static constexpr char kErrorMsgInvalidType[]
--- a/lib/mayaUsd/ufe/UsdAttribute.cpp
+++ b/lib/mayaUsd/ufe/UsdAttribute.cpp
@@ -82,9 +82,9 @@ template <typename T> bool setUsdAttr(const PXR_NS::UsdAttribute& attr, const T&
     // our own in the StagesSubject, which we invoke here, so that only a
     // single UFE attribute changed notification is generated.
     MayaUsd::ufe::AttributeChangedNotificationGuard guard;
-    std::string errMsg;
+    std::string                                     errMsg;
     bool isSetAttrAllowed = MayaUsd::ufe::isAttributeEditAllowed(attr, &errMsg);
-    if (!isSetAttrAllowed){
+    if (!isSetAttrAllowed) {
         MGlobal::displayError(errMsg.c_str());
         return false;
     }
--- a/lib/mayaUsd/ufe/UsdAttribute.cpp
+++ b/lib/mayaUsd/ufe/UsdAttribute.cpp
@@ -82,8 +82,14 @@ template <typename T> bool setUsdAttr(const PXR_NS::UsdAttribute& attr, const T&
     // our own in the StagesSubject, which we invoke here, so that only a
     // single UFE attribute changed notification is generated.
     MayaUsd::ufe::AttributeChangedNotificationGuard guard;
-    bool isSetAttrAllowed = MayaUsd::ufe::isAttributeEditAllowed(attr);
-    return isSetAttrAllowed ? attr.Set<T>(value) : false;
+    std::string errMsg;
+    bool isSetAttrAllowed = MayaUsd::ufe::isAttributeEditAllowed(attr, &errMsg);
+    if (!isSetAttrAllowed){
+        MGlobal::displayError(errMsg.c_str());
+        return false;
+    }
+
+    return attr.Set<T>(value);
 }
 
 PXR_NS::UsdTimeCode getCurrentTime(const Ufe::SceneItem::Ptr& item)
--- a/lib/mayaUsd/ufe/UsdAttribute.cpp
+++ b/lib/mayaUsd/ufe/UsdAttribute.cpp
@@ -15,6 +15,8 @@
 //
 #include "UsdAttribute.h"
 
+#include "private/Utils.h"
+
 #include <mayaUsd/undo/UsdUndoBlock.h>
 #include <mayaUsd/undo/UsdUndoableItem.h>
 
@@ -37,8 +39,8 @@
 #include <ufe/ufeAssert.h>
 
 // Note: normally we would use this using directive, but here we cannot because
-//		 our class is called UsdAttribute which is exactly the same as the one
-//		 in USD.
+//       our class is called UsdAttribute which is exactly the same as the one
+//       in USD.
 // PXR_NAMESPACE_USING_DIRECTIVE
 
 static constexpr char kErrorMsgFailedSet[] = "Failed to set USD attribute with new value";
@@ -81,7 +83,8 @@ template <typename T> bool setUsdAttr(const PXR_NS::UsdAttribute& attr, const T&
     // our own in the StagesSubject, which we invoke here, so that only a
     // single UFE attribute changed notification is generated.
     MayaUsd::ufe::AttributeChangedNotificationGuard guard;
-    return attr.Set<T>(value);
+    bool isSetAttrAllowed = MayaUsd::ufe::isAttributeEditAllowed(attr);
+    return isSetAttrAllowed ? attr.Set<T>(value) : false;
 }
 
 PXR_NS::UsdTimeCode getCurrentTime(const Ufe::SceneItem::Ptr& item)
@@ -140,8 +143,7 @@ void setUsdAttributeVectorFromUfe(
     T vec;
     UFE_ASSERT_MSG(attr.Get<T>(&vec, time), kErrorMsgInvalidType);
     vec.Set(value.x(), value.y(), value.z());
-    bool b = setUsdAttr<T>(attr, vec);
-    UFE_ASSERT_MSG(b, kErrorMsgFailedSet);
+    setUsdAttr<T>(attr, vec);
 }
 
 template <typename T, typename A = MayaUsd::ufe::TypedUsdAttribute<T>>
@@ -273,8 +275,7 @@ void UsdAttributeEnumString::set(const std::string& value)
     UFE_ASSERT_MSG(
         fUsdAttr.Get<PXR_NS::TfToken>(&dummy, getCurrentTime(sceneItem())), kErrorMsgInvalidType);
     PXR_NS::TfToken tok(value);
-    bool            b = setUsdAttr<PXR_NS::TfToken>(fUsdAttr, tok);
-    UFE_ASSERT_MSG(b, kErrorMsgFailedSet);
+    setUsdAttr<PXR_NS::TfToken>(fUsdAttr, tok);
 }
 
 Ufe::UndoableCommand::Ptr UsdAttributeEnumString::setCmd(const std::string& value)
@@ -355,8 +356,7 @@ template <> void TypedUsdAttribute<std::string>::set(const std::string& value)
         std::string dummy;
         UFE_ASSERT_MSG(
             fUsdAttr.Get<std::string>(&dummy, getCurrentTime(sceneItem())), kErrorMsgInvalidType);
-        bool b = setUsdAttr<std::string>(fUsdAttr, value);
-        UFE_ASSERT_MSG(b, kErrorMsgFailedSet);
+        setUsdAttr<std::string>(fUsdAttr, value);
         return;
     } else if (typeName.GetHash() == SdfValueTypeNames->Token.GetHash()) {
         PXR_NS::TfToken dummy;
@@ -364,8 +364,7 @@ template <> void TypedUsdAttribute<std::string>::set(const std::string& value)
             fUsdAttr.Get<PXR_NS::TfToken>(&dummy, getCurrentTime(sceneItem())),
             kErrorMsgInvalidType);
         PXR_NS::TfToken tok(value);
-        bool            b = setUsdAttr<PXR_NS::TfToken>(fUsdAttr, tok);
-        UFE_ASSERT_MSG(b, kErrorMsgFailedSet);
+        setUsdAttr<PXR_NS::TfToken>(fUsdAttr, tok);
         return;
     }
 
@@ -384,8 +383,7 @@ template <> void TypedUsdAttribute<Ufe::Color3f>::set(const Ufe::Color3f& value)
     GfVec3f vec;
     UFE_ASSERT_MSG(fUsdAttr.Get<GfVec3f>(&vec, getCurrentTime(sceneItem())), kErrorMsgInvalidType);
     vec.Set(value.r(), value.g(), value.b());
-    bool b = setUsdAttr<GfVec3f>(fUsdAttr, vec);
-    UFE_ASSERT_MSG(b, kErrorMsgFailedSet);
+    setUsdAttr<GfVec3f>(fUsdAttr, vec);
 }
 
 template <> Ufe::Vector3i TypedUsdAttribute<Ufe::Vector3i>::get() const
@@ -443,8 +441,7 @@ template <typename T> void TypedUsdAttribute<T>::set(const T& value)
     T dummy;
     UFE_ASSERT_MSG(
         fUsdAttr.Get<T>(&dummy, getCurrentTime(Ufe::Attribute::sceneItem())), kErrorMsgInvalidType);
-    bool b = setUsdAttr<T>(fUsdAttr, value);
-    UFE_ASSERT_MSG(b, kErrorMsgFailedSet);
+    setUsdAttr<T>(fUsdAttr, value);
 }
 
 //------------------------------------------------------------------------------
@@ -557,23 +554,23 @@ UsdAttributeDouble3::create(const UsdSceneItem::Ptr& item, const PXR_NS::UsdAttr
 
 #if 0
 // Note: if we were to implement generic attribute setting (via string) this
-//		 would be the way it could be done.
+//       would be the way it could be done.
 bool UsdAttribute::setValue(const std::string& value)
 {
-	// Put the input string into a VtValue so we can cast it to the proper type.
-	PXR_NS::VtValue val(value.c_str());
+    // Put the input string into a VtValue so we can cast it to the proper type.
+    PXR_NS::VtValue val(value.c_str());
 
-	// Attempt to cast the value to what we want.  Get a default value for this
-	// attribute's type name.
-	PXR_NS::VtValue defVal = fUsdAttr.GetTypeName().GetDefaultValue();
+    // Attempt to cast the value to what we want.  Get a default value for this
+    // attribute's type name.
+    PXR_NS::VtValue defVal = fUsdAttr.GetTypeName().GetDefaultValue();
 
-	// Attempt to cast the given string to the default value's type.
-	// If casting fails, attempt to continue with the given value.
-	PXR_NS::VtValue cast = PXR_NS::VtValue::CastToTypeOf(val, defVal);
-	if (!cast.IsEmpty())
-		cast.Swap(val);
+    // Attempt to cast the given string to the default value's type.
+    // If casting fails, attempt to continue with the given value.
+    PXR_NS::VtValue cast = PXR_NS::VtValue::CastToTypeOf(val, defVal);
+    if (!cast.IsEmpty())
+        cast.Swap(val);
 
-	return setUsdAttr<PXR_NS::VtValue>(fUsdAttr, val);
+    return setUsdAttr<PXR_NS::VtValue>(fUsdAttr, val);
 }
 #endif
 
--- a/lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
+++ b/lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
@@ -506,7 +506,9 @@ UsdTransform3dMayaXformStack::rotateCmd(double x, double y, double z)
               auto attr = usdSceneItem->prim().GetAttribute(attrName);
 
               // return an invalid XformOp if the attribute edit is not allowed.
-              if (!MayaUsd::ufe::isAttributeEditAllowed(attr)) {
+              std::string errMsg;
+              if (!MayaUsd::ufe::isAttributeEditAllowed(attr, &errMsg)) {
+                  MGlobal::displayError(errMsg.c_str());
                   return UsdGeomXformOp();
               }
 
@@ -524,7 +526,9 @@ UsdTransform3dMayaXformStack::rotateCmd(double x, double y, double z)
               // At this point we already know that writing to attribute is going
               // to succeed but we do not know if writing to "transform op order" succeed since
               // we could be a weaker layer.
-              if (!MayaUsd::ufe::isAttributeEditAllowed(xformable.GetXformOpOrderAttr())) {
+              std::string errMsg;
+              if (!MayaUsd::ufe::isAttributeEditAllowed(xformable.GetXformOpOrderAttr(), &errMsg)) {
+                  MGlobal::displayError(errMsg.c_str());
                   return UsdGeomXformOp();
               }
 
@@ -558,7 +562,9 @@ Ufe::ScaleUndoableCommand::Ptr UsdTransform3dMayaXformStack::scaleCmd(double x, 
               auto attr = usdSceneItem->prim().GetAttribute(attrName);
 
               // return an invalid XformOp if the attribute edit is not allowed.
-              if (!MayaUsd::ufe::isAttributeEditAllowed(attr)) {
+              std::string errMsg;
+              if (!MayaUsd::ufe::isAttributeEditAllowed(attr, &errMsg)) {
+                  MGlobal::displayError(errMsg.c_str());
                   return UsdGeomXformOp();
               }
 
@@ -574,7 +580,9 @@ Ufe::ScaleUndoableCommand::Ptr UsdTransform3dMayaXformStack::scaleCmd(double x, 
               // At this point we already know that writing to attribute is going
               // to succeed but we do not know if writing to "transform op order" succeed since
               // we could be a weaker layer.
-              if (!MayaUsd::ufe::isAttributeEditAllowed(xformable.GetXformOpOrderAttr())) {
+              std::string errMsg;
+              if (!MayaUsd::ufe::isAttributeEditAllowed(xformable.GetXformOpOrderAttr(), &errMsg)) {
+                  MGlobal::displayError(errMsg.c_str());
                   return UsdGeomXformOp();
               }
 
@@ -675,7 +683,9 @@ Ufe::SetVector3dUndoableCommand::Ptr UsdTransform3dMayaXformStack::setVector3dCm
               auto attr = usdSceneItem->prim().GetAttribute(attrName);
 
               // return an invalid XformOp if the attribute edit is not allowed.
-              if (!MayaUsd::ufe::isAttributeEditAllowed(attr)) {
+              std::string errMsg;
+              if (!MayaUsd::ufe::isAttributeEditAllowed(attr, &errMsg)) {
+                  MGlobal::displayError(errMsg.c_str());
                   return UsdGeomXformOp();
               }
 
@@ -696,7 +706,9 @@ Ufe::SetVector3dUndoableCommand::Ptr UsdTransform3dMayaXformStack::setVector3dCm
                 // At this point we already know that writing to attribute is going
                 // to succeed but we do not know if writing to "transform op order" succeed
                 // since we could be a weaker layer.
-                if (!MayaUsd::ufe::isAttributeEditAllowed(xformable.GetXformOpOrderAttr())) {
+                std::string errMsg;
+                if (!MayaUsd::ufe::isAttributeEditAllowed(xformable.GetXformOpOrderAttr(), &errMsg)) {
+                    MGlobal::displayError(errMsg.c_str());
                     return UsdGeomXformOp();
                 }
 
@@ -727,7 +739,9 @@ UsdTransform3dMayaXformStack::pivotCmd(const TfToken& pvtOpSuffix, double x, dou
               auto attr = usdSceneItem->prim().GetAttribute(pvtAttrName);
 
               // return an invalid XformOp if the attribute edit is not allowed.
-              if (!MayaUsd::ufe::isAttributeEditAllowed(attr)) {
+              std::string errMsg;
+              if (!MayaUsd::ufe::isAttributeEditAllowed(attr, &errMsg)) {
+                  MGlobal::displayError(errMsg.c_str());
                   return UsdGeomXformOp();
               }
 
@@ -751,7 +765,9 @@ UsdTransform3dMayaXformStack::pivotCmd(const TfToken& pvtOpSuffix, double x, dou
               // At this point we already know that writing to attribute is going
               // to succeed but we do not know if writing to "transform op order" succeed since
               // we could be a weaker layer.
-              if (!MayaUsd::ufe::isAttributeEditAllowed(xformable.GetXformOpOrderAttr())) {
+              std::string errMsg;
+              if (!MayaUsd::ufe::isAttributeEditAllowed(xformable.GetXformOpOrderAttr(), &errMsg)) {
+                  MGlobal::displayError(errMsg.c_str());
                   return UsdGeomXformOp();
               }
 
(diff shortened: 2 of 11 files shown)


lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -191,6 +216,53 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
...
+    // opinions for this property.
+    const auto& propertyStack = attr.GetPropertyStack();
+
+    // HS March 15th,2021
+    // TODO: This code works as long as existing opinions are in the stage’s local LayerStack.
+    // Relying on the "expanded primIndex" would be a better way to detect the opinions across
+    // composition arc(s). Some more details can be found:
+    // https://groups.google.com/g/usd-interest/c/xTxFYQA_bRs/m/lX_WqNLoBAAJ
+    SdfLayerHandle strongLayer;
+    for (const auto& spec : propertyStack) {
```
HamedSabri-adsk, line 239: @ppt-adsk See the comments in code that I left here.

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -191,6 +216,53 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
...
+        }
+    }
+
+    if (strongLayer) {
+        std::string err = TfStringPrintf(
+            "Cannot edit [%s] attribute because there is a stronger opinion in [%s].",
+            attr.GetBaseName().GetText(),
+            strongLayer->GetDisplayName().c_str());
+
+        MGlobal::displayError(err.c_str());
```
HamedSabri-adsk, line 259: I originally intended to just "throw std::runtime_error" or "throw" but that won't work considering the very long
error message that we would get when the exceptions are thrown.

`# Error: RuntimeError: file <string> line 2: RuntimeError: file C:\Users\sabrih\maya\builds\master\maya\build\RelWithDebInfo\runTime\Python\lib\site-packages\maya\internal\ufeSupport\ufeCmdWrapper.py line 32: Cannot edit [height] attribute because there is a stronger opinion in [anonymousLayer4].`

Instead this function returns a bool and simply shows an error message using MGlobal::displayError

lib/mayaUsd/ufe/wrapUtils.cpp
```diff
@@ -194,6 +217,47 @@ int ufePathToInstanceIndex(const std::string& ufePathString)
...
+            strongLayer = spec->GetLayer();
+            break;
+        }
+    }
+
+    if (strongLayer) {
+        return false;
+    }
+
+    return true;
```
HamedSabri-adsk, line 258: I duplicated the isAttributeEditAllowed() routine here to make it work with the python binding. 
I realize it is a bit wacky but having a python binding for "Ufe/Private/Utils" turned out to be more work. We can chat offline if you think there is a better way to go about this.

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -21,13 +21,37 @@
...
 
 PXR_NAMESPACE_USING_DIRECTIVE
 
+namespace {
+// find position index for a given layer in the local LayerStack ( strong-to-weak order )
+uint32_t findLayerIndex(const PXR_NS::UsdAttribute& attr, const PXR_NS::SdfLayerHandle& layer)
+{
+    const auto& prim = attr.GetPrim();
+    const auto& stage = prim.GetStage();
+    const auto& layerStack = stage->GetLayerStack();
```
ppt-adsk, line 38: This is all fine, of course --- but I wonder performance-wise.  Unless the layer stack changes (and I think there are USD events to detect this --- maybe this one
https://graphics.pixar.com/usd/docs/api/class_sdf_notice_1_1_layers_did_change.html
but the docs appear broken), you can compute the index once and cache it.  Maybe we can wait for this optimization until we figure out what Pixar will be giving us (or when they will give it to us :) ).

I can see something like
std::unordered_map<SdfLayerHandle, int> layerIndices;

lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -81,7 +82,8 @@ template <typename T> bool setUsdAttr(const PXR_NS::UsdAttribute& attr, const T&
     // our own in the StagesSubject, which we invoke here, so that only a
     // single UFE attribute changed notification is generated.
     MayaUsd::ufe::AttributeChangedNotificationGuard guard;
-    return attr.Set<T>(value);
+    bool isSetAttrAllowed = MayaUsd::ufe::isAttributeEditAllowed(attr);
```
ppt-adsk, line 85: As per my proposal, here I'd pass in a pointer to string to receive an eventual error message, and displayError() if the return value is false.

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -21,13 +21,37 @@
...
+
+    uint32_t position { 0 };
+
+    auto iter = std::find_if(
+        std::begin(layerStack), std::end(layerStack), [&](const PXR_NS::SdfLayerHandle& l) -> bool {
+            return layer == l;
+        });
+
+    if (iter != layerStack.end()) {
```
ppt-adsk, line 47: Bit awkward that if iter == layerStack.end() we'll return position == 0.  Should we use TF_AXIOM(iter != layerStack.end())?

test/lib/ufe/testAttribute.py
```diff
@@ -217,13 +220,17 @@ def testAttributeBool(self):
         # Compare the initial UFE value to that directly from USD.
         self.assertEqual(ufeAttr.get(), usdAttr.Get())
 
-        # Set the attribute in UFE with the opposite boolean value.
-        ufeAttr.set(not ufeAttr.get())
+        # check to see if the attribute edit is allowed
+        isAttrEditAllowed = mayaUsdUfe.isAttributeEditAllowed(usdAttr)
+        self.assertFalse(isAttrEditAllowed, False)
```
ppt-adsk, line 225: This is weird...  assertFalse() takes a single argument...  And if you're expecting no edits allowed, then line 226 is a bit pointless :)  Same in a bunch of places.

test/lib/ufe/testAttribute.py
```diff
@@ -613,9 +644,139 @@ def testZAttrChangeRedoAfterPrimCreateRedo(self):
...
+        proxyShapePath = proxyShapes[0]
+        proxyShapeItem = ufe.Hierarchy.createItem(ufe.PathString.path(proxyShapePath))
+        proxyShapeContextOps = ufe.ContextOps.contextOps(proxyShapeItem)
+
+        stage = mayaUsd.lib.GetPrim(proxyShapePath).GetStage()
+
+        # create a prim. This creates the primSpec in the root layer.
+        proxyShapeContextOps.doOp(['Add New Prim', 'Capsule'])
+
+        # create a USD prim
```
ppt-adsk, line 667: More like "Get the USD prim from the scene item."

test/lib/ufe/testAttribute.py
```diff
@@ -613,9 +644,139 @@ def testZAttrChangeRedoAfterPrimCreateRedo(self):
...
+        capsulePath = ufe.PathString.path('|stage1|stageShape1,/Capsule1')
+        capsuleItem = ufe.Hierarchy.createItem(capsulePath)
+        capsulePrim = mayaUsd.ufe.ufePathToPrim(ufe.PathString.string(capsulePath))
+
+        # author a new radius value
+        self.assertTrue(capsulePrim.HasAttribute('radius'))
+        radiusAttrUsd = capsulePrim.GetAttribute('radius')
+
+        # authoring new attribute edit is expected to be allowed.
+        self.assertTrue(mayaUsdUfe.isAttributeEditAllowed(radiusAttrUsd), True)
```
ppt-adsk, line 677: assertTrue() takes a single argument.

test/lib/ufe/testAttribute.py
```diff
@@ -613,9 +644,139 @@ def testZAttrChangeRedoAfterPrimCreateRedo(self):
...
+        # authoring new transformation edit is expected to be allowed.
+        self.assertTrue(mayaUsdUfe.isAttributeEditAllowed(translateAttr), True)
+        sphereT3d.translate(5.0, 6.0, 7.0)
+        self.assertEqual(translateAttr.Get(), Gf.Vec3d(5.0, 6.0, 7.0))
+
+        # set the edit target to a weaker layer (LayerC)
+        cmds.mayaUsdEditTarget(proxyShapePath, edit=True, editTarget=subLayerC)
+
+        # authoring new transformation edit is not allowed.
+        self.assertFalse(mayaUsdUfe.isAttributeEditAllowed(translateAttr), False)
```
ppt-adsk, line 769: Might want to write a value different from the existing one and confirm that the write did not happen.

lib/mayaUsd/ufe/wrapUtils.cpp
ppt-adsk: Yeah, the code duplication is unfortunate.  Would it help if you moved it out of private/Utils and put it somewhere else?  Because now, your Python code isn't testing your C++ code...

ppt-adsk (changes requested):

lib/mayaUsd/ufe/wrapUtils.cpp
HamedSabri-adsk: @ppt-adsk I got rid of the wacky duplicate by moving out the code one level up into "Ufe/Utils" which would work better.

Addressed in : fe8131c1c837ebf7bf5405e7c594df7b5b7b1db4

lib/mayaUsd/ufe/private/Utils.cpp
HamedSabri-adsk: @ppt-adsk Good point. addressed in fe8131c1c837ebf7bf5405e7c594df7b5b7b1db4

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -191,6 +216,53 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
...
+    // TODO: This code works as long as existing opinions are in the stage’s local LayerStack.
+    // Relying on the "expanded primIndex" would be a better way to detect the opinions across
+    // composition arc(s). Some more details can be found:
+    // https://groups.google.com/g/usd-interest/c/xTxFYQA_bRs/m/lX_WqNLoBAAJ
+    SdfLayerHandle strongLayer;
+    for (const auto& spec : propertyStack) {
+
+        // skip if the edit target layer is stronger than the propSpec layer
+        const auto propSpecLayerIndex = findLayerIndex(attr, spec->GetLayer());
+        if (targetLayerIndex <= propSpecLayerIndex) {
```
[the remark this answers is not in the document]
HamedSabri-adsk: @ppt-adsk I kind of see what you mean by eliminating the need to iterate the PropertyStack() but I think the algorithm is more obvious by using it. We can discuss it if my comment is not clear. Correct, a strong layer have a lower index than a weak layer. 

The PropertyStack() would eventually go away if we decide to rely on the "site" and iterating PcpPrimIndex.  

PcpPrimIndex is an index of the all sites of scene description that contribute opinions to a specific prim, under composition semantics.

https://graphics.pixar.com/usd/docs/api/class_pcp_prim_index.html#details

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -21,13 +21,37 @@
...
+// find position index for a given layer in the local LayerStack ( strong-to-weak order )
+uint32_t findLayerIndex(const PXR_NS::UsdAttribute& attr, const PXR_NS::SdfLayerHandle& layer)
+{
+    const auto& prim = attr.GetPrim();
+    const auto& stage = prim.GetStage();
+    const auto& layerStack = stage->GetLayerStack();
+
+    uint32_t position { 0 };
+
+    auto iter = std::find_if(
```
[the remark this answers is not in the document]
HamedSabri-adsk: I believe std::find_if is what I wanted to use there since "std::find_if" takes a unary predicate and checks whether the predicate returns true or false to a given element. However you are right, the conditional find could be replace by this:

auto iter = std::find(std::begin(layerStack), std::end(layerStack), layer);

Both ways would work. Addressed in fe8131c1c837ebf7bf5405e7c594df7b5b7b1db4

lib/mayaUsd/ufe/private/Utils.cpp
HamedSabri-adsk: @ppt-adsk caching the index compute should be trivial. We can add it at any-times if performance becomes a bottleneck. 

Honestly, how many layers are we talking about? Even with 500s layers which is a lot of layers to me we should have a very fast look up. 

My hope with Pixar is that the client(s) of such APIs shouldn't even think about such details. It should all happen behind the scene.

lib/mayaUsd/ufe/UsdAttribute.cpp
HamedSabri-adsk: Addressed in fe8131c1c837ebf7bf5405e7c594df7b5b7b1db4

test/lib/ufe/testAttribute.py
HamedSabri-adsk: @ppt-adsk  Good catch. Copy-paste damage  :) 

Addressed in 7be71c51668f70493e0533da23f91b1368e73bad

I agree that the check after is useless by my thinking was the code was still valuable to have around. I can remove them if it makes things more clear.

lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
```diff
@@ -101,7 +102,7 @@ namespace ufe {
 
 namespace {
```
HamedSabri-adsk, line 103: @ppt-adsk  Unrelated to this PR but I believe having an anonymous namespace inside MAYAUSD_NS_DEF::ufe namespace is very confusing. Considering there is already an anonymous namespace starting at line 40 and ending at line 98.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -80,6 +80,24 @@ bool stringBeginsWithDigit(const std::string& inputString)
     return false;
 }
 
+// find position index for a given layer in the local LayerStack ( strong-to-weak order )
+uint32_t findLayerIndex(const PXR_NS::UsdAttribute& attr, const PXR_NS::SdfLayerHandle& layer)
```
ppt-adsk, line 84: Just noticed this (sorry...), and not a big deal because it's in the unnamed namespace, but interface is awkward: you require an attr, but in fact you really want just a stage, because you use the attr only to get the stage.  If we ever factor this out, I would change this to accept a stage as the first argument.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -384,6 +402,54 @@ TfTokenVector getProxyShapePurposes(const Ufe::Path& path)
...
+    const auto& editTarget = stage->GetEditTarget();
+
+    // get the index to edit target layer
+    const auto targetLayerIndex = findLayerIndex(attr, editTarget.GetLayer());
+
+    // get the strength-ordered ( strong-to-weak order ) list of property specs that provide
+    // opinions for this property.
+    const auto& propertyStack = attr.GetPropertyStack();
+
+    // HS March 15th,2021
```
ppt-adsk, line 419: As I mentioned I think the loop is an over-elaboration.  What you want is to compare the layer at the top of the property stack to the target layer index, something like this:

SdfLayerHandle topLayer = propertyStack.front().GetLayer();
if (findLayerIndex(attr, topLayer) < targetLayerIndex) {
    std::string err = TfStringPrintf( // etc.

lib/mayaUsd/ufe/private/Utils.cpp
ppt-adsk: I'm less concerned about the number of layers than about the number of attributes to check...  Which might be hundreds of thousands.  But you're right, let's put that on the side for now, we can revisit this later.

test/lib/ufe/testAttribute.py
ppt-adsk: As mentioned in the code, I'm concerned that we've lost code coverage because of this.  Would be nice to rework the test file perhaps, so that we can keep the tests?

lib/mayaUsd/ufe/wrapUtils.cpp
```diff
@@ -219,43 +196,8 @@ int ufePathToInstanceIndex(const std::string& ufePathString)
...
-        }
-    }
-
-    if (strongLayer) {
-        return false;
-    }
-
-    return true;
+    std::string dummyStr;
+    return ufe::isAttributeEditAllowed(attr, &dummyStr);
```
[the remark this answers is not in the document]
HamedSabri-adsk: That's what I tried first but boost python binding doesn't like it. 

```
 Stack trace:
15:   mayaUsd.dll!std::basic_string<char,std::char_traits<char>,std::allocator<char> >::assign Line 2659
15:   mayaUsd.dll!MayaUsd_v0::ufe::isAttributeEditAllowed Line 435
15:   _ufe.pyd!boost::python::objects::caller_py_function_impl<boost::python::detail::caller<bool (__cdecl*)(pxrInternal_v0_21__pxrReserved__::UsdAttribute const &),boost::python::default_call_policies,boost:
```

test/lib/ufe/testAttribute.py
```diff
@@ -222,7 +222,7 @@ def testAttributeBool(self):
 
         # check to see if the attribute edit is allowed
         isAttrEditAllowed = mayaUsdUfe.isAttributeEditAllowed(usdAttr)
-        self.assertFalse(isAttrEditAllowed, False)
+        self.assertFalse(isAttrEditAllowed)
```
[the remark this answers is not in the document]
HamedSabri-adsk: Correct, the logic in the "if isAttrEditAllowed:" conditions are dead code. 
When these test were written, we didn't have such restriction to block attribute edits but now we have. In the case of top_layer.ma example, the top layer "Ball.maya.usd" has already set certain attributes ( e.g doubleSided, etc... ) and setting them in the weaker layers are blocked now and therefore the following tests after "if isAttrEditAllowed:" condition results in error.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -80,6 +80,24 @@ bool stringBeginsWithDigit(const std::string& inputString)
...
+{
+    const auto& prim = attr.GetPrim();
+    const auto& stage = prim.GetStage();
+    const auto& layerStack = stage->GetLayerStack();
+
+    uint32_t position { 0 };
+
+    auto iter = std::find(std::begin(layerStack), std::end(layerStack), layer);
+
+    if (iter != layerStack.end()) {
```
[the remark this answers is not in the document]
HamedSabri-adsk: addressed ac2bb4dc732e9862957b25a17eb9f6f7719adcf3

lib/mayaUsd/ufe/Utils.cpp
HamedSabri-adsk: addressed ac2bb4dc732e9862957b25a17eb9f6f7719adcf3

lib/mayaUsd/ufe/Utils.cpp
HamedSabri-adsk: I replaced the loop if it makes logics easier to follow.

addressed ac2bb4dc732e9862957b25a17eb9f6f7719adcf3

lib/mayaUsd/ufe/Utils.cpp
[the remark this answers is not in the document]
ppt-adsk: Think we're settling on TF_ debugging facilities, e.g. TF_AXIOM(iter != layerStack.end())

lib/mayaUsd/ufe/Utils.cpp
ppt-adsk: Thanks, makes more sense to me at least :)

test/lib/ufe/testAttribute.py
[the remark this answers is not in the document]
ppt-adsk: Right --- and my suggestion here, to keep the test code coverage for the UFE attributes support, is to set the attributes in the stronger layer, otherwise we won't be testing the Ufe::Attribute::set() interfaces any more.  You've already added two new tests to test attribute blocking, so we don't need additional testing for that, but we do need to keep the test coverage for Ufe::Attribute::set().

lib/mayaUsd/ufe/Utils.cpp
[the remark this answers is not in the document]
HamedSabri-adsk: Yup that's what I tried first and spent 5 mins last night scratching my head why boost python doesn't like it:

```
15:   KERNELBASE.dll!DebugBreak
15: 00007FFA24C99E10Result: C:/Users/sabrih/Desktop/workspace/build/RelWithDebInfo/test/testSamples/ballset/StandaloneScene/top_layer.ma
15: Fatal Error. Attempting to save in C:/Users/sabrih/AppData/Local/Temp/sabrih.20210318.1311.ma
```
I tried to search USD code base and I can't find any examples where TF_AXIOM is being used in any functions that are exposed to Boost python. 

I believe the good old assert() should do fine here.

test/lib/ufe/testAttribute.py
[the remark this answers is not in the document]
HamedSabri-adsk: Fair enough, let me have a look.

lib/mayaUsd/ufe/wrapUtils.cpp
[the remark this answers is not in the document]
ppt-adsk: That's very strange, I think the fact you have Boost in the call stack is simply because it's being called from Python.  This should work, let me know if you'd like us to debug this together.

lib/mayaUsd/ufe/Utils.cpp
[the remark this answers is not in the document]
ppt-adsk: Again, Python shouldn't have anything to do this this...  Is the TF_AXIOM failing then?

test/lib/ufe/testAttribute.py
```diff
@@ -417,19 +395,15 @@ def testAttributeFloat3(self):
...
-                             ufe.Vector3f(vec.x()+1.0, vec.y()+2.0, vec.z()+3.0))
+        # Set the attribute in UFE with some random values.
+        vec = ufe.Vector3f(random.random(), random.random(), random.random())
+        ufeAttr.set(vec)
+
+        # Then make sure that new UFE value matches what it in USD.
+        self.assertVectorAlmostEqual(ufeAttr.get(), usdAttr.Get())
+
+        self.runUndoRedo(ufeAttr,
+                         ufe.Vector3f(vec.x()+1.0, vec.y()+2.0, vec.z()+3.0))
```
HamedSabri-adsk: There is no more dead code and all my silly isAttrEditAllowed() checks are removed.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -410,29 +423,36 @@ bool isAttributeEditAllowed(const PXR_NS::UsdAttribute& attr, std::string* errMs
...
     const auto& propertyStack = attr.GetPropertyStack();
 
     if (!propertyStack.empty()) {
+        // get the strongest layer that has the attr.
+        auto strongestLayer = attr.GetPropertyStack().front()->GetLayer();
 
-        auto topLayer = attr.GetPropertyStack().front()->GetLayer();
+        // compare the calculated index between the "attr" and "edit target" layers.
+        if (findLayerIndex(prim, strongestLayer) < targetLayerIndex) {
+            if (errMsg) {
```
HamedSabri-adsk, line 447: Thank you Pierre. Indeed we needed to check for the null here....

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -410,29 +423,36 @@ bool isAttributeEditAllowed(const PXR_NS::UsdAttribute& attr, std::string* errMs
...
+    // Read on Value Clips here:
+    // https://graphics.pixar.com/usd/docs/api/_usd__page__value_clips.html
+
     // get the strength-ordered ( strong-to-weak order ) list of property specs that provide
     // opinions for this property.
     const auto& propertyStack = attr.GetPropertyStack();
 
     if (!propertyStack.empty()) {
+        // get the strongest layer that has the attr.
+        auto strongestLayer = attr.GetPropertyStack().front()->GetLayer();
```
HamedSabri-adsk, line 443: renamed topLayer with strongestLayer

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -80,19 +82,30 @@ bool stringBeginsWithDigit(const std::string& inputString)
...
-    if (iter != layerStack.end()) {
-        return std::distance(layerStack.begin(), iter);
+        // iterate through the "local" Layer stack for each site
+        // to find the layer
+        for (SdfLayerRefPtr const& l : layerStack->GetLayers()) {
+            if (l == layer) {
+                return position;
+            }
+            ++position;
+        }
```
HamedSabri-adsk, line 108: The original algorithm that I wrote here was very naive and didn't handle the cases where the attribute came in across composition arcs. By walking the expanded primIndex, we would be able to visit all the "**sites**" and their "**local layer stack**".

I am also thinking of another way to write this code but believe for the pass should be good. 

Thanks to Brett Levin and Spiff to help me better understand the "primindex" and "site" concepts.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -25,6 +25,8 @@
 
 #include <pxr/base/tf/hashset.h>
 #include <pxr/base/tf/stringUtils.h>
+#include <pxr/usd/pcp/layerStack.h>
+#include <pxr/usd/pcp/site.h>
```
HamedSabri-adsk, line 29: bring in "pcp/site" and "pcp/layerStack".

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -410,29 +423,36 @@ bool isAttributeEditAllowed(const PXR_NS::UsdAttribute& attr, std::string* errMs
...
 
-    // HS March 15th,2021
-    // TODO: This code works as long as existing opinions are in the stage’s local LayerStack.
-    // Relying on the "expanded primIndex" would be a better way to detect the opinions across
-    // composition arc(s). Some more details can be found:
+    // HS March 22th,2021
+    // TODO: "Value Clips" are UsdStage-level feature, unknown to Pcp.So if the attribute in
+    // question is affected by Value Clips, we would will likely get the wrong answer. See Spiff
+    // comment for more information :
     // https://groups.google.com/g/usd-interest/c/xTxFYQA_bRs/m/lX_WqNLoBAAJ
```
HamedSabri-adsk, line 432: See my comment in regard to "Value Clips".

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -80,19 +82,30 @@ bool stringBeginsWithDigit(const std::string& inputString)
...
 {
-    const auto& layerStack = stage->GetLayerStack();
-
     uint32_t position { 0 };
 
-    auto iter = std::find(std::begin(layerStack), std::end(layerStack), layer);
+    const PXR_NS::PcpPrimIndex& primIndex = prim.GetPrimIndex();
+
+    // iterate through the expanded primIndex
+    for (const PcpNodeRef& node : primIndex.GetNodeRange()) {
```
HamedSabri-adsk, line 94: PF results back and I am hit with `-Wrange-loop-analysis` with Clang on MacOS:

```
    /local/S/jenkins/workspace/ecg-mayausd-branch-preflight-2022-python3-osx/ecg-maya-usd/maya-usd/lib/mayaUsd/ufe/Utils.cpp:94:28: error: loop variable 'node' is always a copy because the range of type 'pxrInternal_v0_20__pxrReserved__::PcpNodeRange' (aka 'pair<pxrInternal_v0_20__pxrReserved__::PcpNodeIterator, pxrInternal_v0_20__pxrReserved__::PcpNodeIterator>') does not return a reference [-Werror,-Wrange-loop-analysis]
        for (const PcpNodeRef& node : primIndex.GetNodeRange()) {
```

I believe I have previously commented on this flag before in https://github.com/Autodesk/maya-usd/pull/930/ 

My original intention here is that I don't want a copy in my range loop and find it hard this iterator object is always copy. 

`for(const PcpNodeRef &node: primIndex.GetNodeRange()) `

Not to mention that this is pattern that is used everywhere in USD code base:

e.g https://github.com/PixarAnimationStudios/USD/blob/dc710925675f7f58f7a37f6c18d046a687b09271/pxr/usd/usdShade/material.cpp#L439

ppt-adsk: Did you try const auto& node?

lib/mayaUsd/ufe/Utils.cpp
ppt-adsk: So this linearizes the complete layer stack and expands out each site, such that if there were m node sites each with n layers, then the value of position at the end of the loop would be (m x n) - 1, right?

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -80,19 +82,30 @@ bool stringBeginsWithDigit(const std::string& inputString)
     return false;
 }
 
-// find position index for a given layer in the local LayerStack ( strong-to-weak order )
-uint32_t findLayerIndex(const UsdStageWeakPtr& stage, const PXR_NS::SdfLayerHandle& layer)
+// This function calculates the position index for a given layer across all
+// the site's local LayerStack.
```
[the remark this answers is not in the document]
HamedSabri-adsk: @ppt-adsk  Yes it is. Every Site ( simply a PcpNode in the DAG ) has a "local layer stack" and "scene path".

"Local" just means a site's "root layer and all its sublayers". That's the sense in which is a "local" layer stack: it has no "global" knowledge.

To make it easier to see things, I make a change in Pcp DAG graph representation, where I show the size of
each layer stack in each site. 

PrimIndex_Concept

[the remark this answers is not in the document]
ppt-adsk: Fantastic explanation, thanks!  Therefore, shouldn't the comment read "across all the site's local LayerStacks" plural layer stacks?

HamedSabri-adsk:
> So this linearizes the complete layer stack and expands out each site, such that if there were m node sites each with n layers, then the value of position at the end of the loop would be (m x n) - 1, right?

Yes and no. The number of iteration is (M x N)

where 

M = total number of sites
N = total number of layers 
Position = N - 1

It is because N -1 because position value is incremented after the check.

So let's use the DAG graph above as an example. By the way, there is a tiny typo is my above image. **Site 5** should be replaced by **Site 6**. Apologizes.

M = 6 
N = 10

If I am searching for **@/Ball.maya.usd@</Ball/mesh>**, the position would be **9**

Site 1
-------
top_layer-session.usda
top_layer.usda
Assembly_room_set.usda
edits.usda
edits.usda

Site 2
-------
Ball.usd

Site 3
-------
Ball_payload.usd

Site 4
-------
Ball.shadingVariants.usda

Site 5
-------
Ball.shadingVariants.usda

Site 6
-------
Ball.maya.usd

lib/mayaUsd/ufe/Utils.cpp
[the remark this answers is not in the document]
HamedSabri-adsk: Good point, I see. Yes should be plural.

HamedSabri-adsk:
> Did you try const auto& node?

@ppt-adsk I haven't tried but one shouldn't have to...  

As I reported before `range-loop-analysis` has a tendency to give false positive results. We don't have this flag in Maya and I also haven't seen it enforced in other projects that I know.

**Avoid Wrange-loop-analysis false positives**

https://reviews.llvm.org/rG318677e78def0023d210a29f4b3cf648e02f9fdc

https://quuxplusone.github.io/blog/2020/08/26/wrange-loop-analysis/

lib/mayaUsd/ufe/Utils.cpp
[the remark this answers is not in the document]
HamedSabri-adsk: Addressed in 818e642effab378839cf042abf4af9b6462df5a6

ppt-adsk (commented):
Will leave it up to you as to how to fix the compiler warning as error on macOS, no particular guidance I could give on that one.

HamedSabri-adsk:
@ppt-adsk per our internal discussion. 

I reverted the clang flag change to minimize the impact on others who may need it:

b89e2b295bc3277d7af2238f799e21064c35edad

Dropped `const&` for PcpNodeRef and use it by value since it is just a pointer.  We could also use the `auto&& ` pattern but going by value should be more obvious for some. Added my Godbolt example here for reference:

https://gcc.godbolt.org/z/cK3bWPT4d
  

b2356d4f806b6ffa02c35278059cb252b7e84656

ppt-adsk (approved):

PR #1260, opened 2021-03-16
MAYA-105877 - As a user, I would like to be able to select usd object…

spinell-adsk:
…s based on kind from the Select menu

Add new menu "Select Mode" under the maya "Select" menu to allow usd selection by kind


plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -113,6 +148,58 @@ proc initRuntimeCommands() {
         }
     }
 
+    //
```
spinell-adsk, line 151: add runtime command for all usd built-in kinds

spinell-adsk (commented):
Add new menu under the maya's select menu to allow user to select by kind.
The new menu will display the built-in kind first, than it's will display all custom kinds in alphabetic order

spinell-adsk:
Only changed in mel files, so I did not run the preflight

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
```diff
@@ -90,9 +92,25 @@ global proc mayaUSDRegisterStrings() {
     register("kSaveOptionNoPromptAnn", "You can re-enable this prompt under File|Save Scene Options.");
     register("kSaveOptionTitle", "Save USD Options");
     register("kSaveOptionUnsavedEdits", "You have unsaved USD edits. How would you like to proceed?");
+    register("kUSDSelectionMode", "USD Selection Mode");
+    register("kUSDSelectionModeAnn", "Assign selection categories to prims via the Attribute Editor to activate selection modes in this menu. Choose a selection mode to reflect changes when you select prims in the Viewport and Outliner. Note: By default, the fallback selection mode is by prim. If a selection mode is applicable, it is applied to your scene but otherwise defaults to selecting by prim.");
```
seando-adsk, line 96: In the JIRA item I see that the text "Attribute Editor" in this line is supposed to be bolded. You can do that with <b>text to bold</b>. See kSaveOption2GBWarning below.

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
```diff
@@ -90,9 +92,25 @@ global proc mayaUSDRegisterStrings() {
...
+    register("kUSDSelectionModeModel", "Model");
+    register("kUSDSelectionModeModelAnn", "Selection mode for prims in the model hierarchy. This includes prims set to group, assembly and component kind.");
+    register("kUSDSelectionModeNone", "(none)");
+    register("kUSDSelectionModeNoneAnn", "Selection mode for prims that have no kind set.");
+    register("kUSDSelectionModeSubComponent", "Subcomponent");
+    register("kUSDSelectionModeSubComponentAnn", "Selection mode for prims set to subcomponent kind. Tip: Set subcomponent kind in the Attribute Editor to prims that are an individual asset.");
     register("kTimeAnn", "Edits the current time value of a stage, which corresponds to the animation frame drawn in the viewport. By default, this value connects to Maya's global time node.");
     register("kTipYouCanChooseMultipleFiles", "<b>Tip:</b>  You can choose multiple files.");
     register("kUnloadAll", "Unload All");
+    register("kUniversalSceneDescription", "Universal Scene Description");
```
seando-adsk, line 113: This string already exists with "(USD)" at the end and is used where we add the stages menu. See tag kMenuStageSubMenu above. They are probably supposed to be the same. Please check with Natalia to see if they should be the same and if so we don't need this extra tag (just use one above).

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
seando-adsk: I also think you missed the second paragraph of this text. You can add the break with <br>

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
```diff
@@ -90,9 +92,25 @@ global proc mayaUSDRegisterStrings() {
     register("kSaveOptionNoPromptAnn", "You can re-enable this prompt under File|Save Scene Options.");
     register("kSaveOptionTitle", "Save USD Options");
     register("kSaveOptionUnsavedEdits", "You have unsaved USD edits. How would you like to proceed?");
+    register("kUSDSelectionMode", "USD Selection Mode");
+    register("kUSDSelectionModeAnn", "Assign selection categories to prims via the Attribute Editor to activate selection modes in this menu. Choose a selection mode to reflect changes when you select prims in the Viewport and Outliner. Note: By default, the fallback selection mode is by prim. If a selection mode is applicable, it is applied to your scene but otherwise defaults to selecting by prim.");
+    register("kUSDSelectionModeAssembly", "Assembly");
+    register("kUSDSelectionModeAssemblyAnn", "Selection mode for prims set to assembly kind. Tip: Set assembly kind in the Attribute Editor to prims that are part of an important group.");
+    register("kUSDSelectionModeComponent", "Component");
```
seando-adsk, line 99: I wonder if these menu entries should be localized. If we show a localized string in the menu, then it won't match the actual kind metadata value stored in the prim. The metadata value will never be localized so I don't think we should localize these either. Ask Nat what she thinks.

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -69,6 +71,39 @@ proc removeMenuCallback(string $menuName, string $cmd) {
 
 }
 
+///////////////////////////////////////////////////////////////////////////////
+// mayaUSD_setSelectionKind
+// Helper function to set the selection kind mode
+global proc mayaUSD_setSelectionKind(string $king) {
+    optionVar -stringValue "mayaUsd_SelectionKind" $king;
```
seando-adsk, line 78: Typo "king" -> "kind"

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -69,6 +71,39 @@ proc removeMenuCallback(string $menuName, string $cmd) {
...
+    }
+    return $selectionKind;
+}
+
+///////////////////////////////////////////////////////////////////////////////
+// mayaUSD_isSelectionKindValid
+// return true if the current selection kind is valid
+proc int mayaUSD_isSelectionKindValid() {
+    python "from pxr import Kind";
+    $allKinds = `python "Kind.Registry.GetAllKinds()"`;
```
seando-adsk, line 97: This can be combined into a single line like so:

```$allKinds = python("from pxr import Kind; Kind.Registry.GetAllKinds()");```

Comment (seando-adsk) [reply plugin/adsk/scripts/mayaUsdMenu.mel:97]: I also see you didn't declare the type of $allKinds or $currentKind.

File: plugin/adsk/scripts/mayaUSDRegisterStrings.mel (hunk 4 above)
Comment (seando-adsk) [reply plugin/adsk/scripts/mayaUSDRegisterStrings.mel:99]: Okay I see below that you are using the USD tokens, such as "Kind.Tokens.model" to get the string. So these should definitely not be localized here as they won't match when using other languages.

File: plugin/adsk/scripts/mayaUsdMenu.mel [hunk 7]
```diff
@@ -154,6 +241,116 @@ global proc mayaUsdMenu_createMenuCallback() {
...
+        menuItem -radioButton on  -runTimeCommand ("mayaUsdSelectKindNone") selectKindNone;
+        menuItem -radioButton off -runTimeCommand ("mayaUsdSelectKindModel") selectKindModel;
+        menuItem -radioButton off -runTimeCommand ("mayaUsdSelectKindGroup") selectKindGroup;
+        menuItem -radioButton off -runTimeCommand ("mayaUsdSelectKindAssembly") selectKindAssembly;
+        menuItem -radioButton off -runTimeCommand ("mayaUsdSelectKindComponent") selectKindComponent;
+        menuItem -radioButton off -runTimeCommand ("mayaUsdSelectKindSubComponent") selectKindSubcomponent;
+      
+        // add the custom select kind menu items
+        python "from pxr import Kind";
+        string $builtInKinds[] = `python "[Kind.Tokens.model, Kind.Tokens.group, Kind.Tokens.assembly, Kind.Tokens.component, Kind.Tokens.subcomponent]"`;
```
seando-adsk, line 316: These can be combined into a single call to "python"

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -154,6 +241,116 @@ global proc mayaUsdMenu_createMenuCallback() {
...
+        menuItem -divider true -dividerLabel $label;
+        radioMenuItemCollection;
+
+        // add the default (built-in) select kind menu items
+        menuItem -radioButton on  -runTimeCommand ("mayaUsdSelectKindNone") selectKindNone;
+        menuItem -radioButton off -runTimeCommand ("mayaUsdSelectKindModel") selectKindModel;
+        menuItem -radioButton off -runTimeCommand ("mayaUsdSelectKindGroup") selectKindGroup;
+        menuItem -radioButton off -runTimeCommand ("mayaUsdSelectKindAssembly") selectKindAssembly;
+        menuItem -radioButton off -runTimeCommand ("mayaUsdSelectKindComponent") selectKindComponent;
+        menuItem -radioButton off -runTimeCommand ("mayaUsdSelectKindSubComponent") selectKindSubcomponent;
```
seando-adsk, line 312: What is the purpose of creating/having all these runtime commands? Why not just call the MEL proc mayaUSD_setSelectionKind with string just like you do below on line 330 for the custom kinds?

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -154,6 +241,116 @@ global proc mayaUsdMenu_createMenuCallback() {
...
+            break;
+        case "assembly":
+            menuItem -e -radioButton on selectKindAssembly;
+            break;
+        case "component":
+            menuItem -e -radioButton on selectKindComponent;
+            break;
+        case "subcomponent":
+            menuItem -e -radioButton on selectKindSubcomponent;
+            break;
```
seando-adsk, line 267: Do we need special cases for all these known types? Why not just let the loop below (lines 273-279) handle all the types? The only special case is the empty one.

seando-adsk (changes requested):

plugin/adsk/scripts/mayaUsdMenu.mel
spinell-adsk: They are built-in kind ... so many user will want to bind a shortcut  key ? I think in maya to be able to bind a shortcut key we need to use runtime command

seando-adsk: To bind to a hotkey you need a **nameCommand** and for that you can set any command, but normally in Maya we use runtimeCommand for these. I see you are following the pattern of other menu items in this file so I'm good with it.

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
spinell-adsk: Those texts are annotations on menuItem, so they will be display inside maya status bar which does not support bold text....

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
spinell-adsk: According to the user story, the built-in kind should start with a capital letter. This is why I had a string for them.

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
spinell-adsk: I spoke with Natalia and built-in kind will not start with capital letter. So I will remove the localization.

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
spinell-adsk: I spoke with Natalia and built-in kind will not start with capital letter. So I will remove the localization.

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
seando-adsk: Oops sorry I missed that. This one is very long, does it all fit on the status line?

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
seando-adsk: Okay. I would suggest using the token from USD for the menu string.

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
spinell-adsk: It's depend of the maya UI layout you have. With my current layout yes. But the other part of the string that I did not add will not be visible. This is why I did not add it.

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
```diff
@@ -35,6 +35,8 @@ global proc mayaUSDRegisterStrings() {
     register("kExcludePrimPathsAnn", "Specify the path of a prim to exclude it from the viewport display. Multiple prim paths must be separated by a comma.");
     register("kExcludePrimPathsSbm", "Specify the path of a prim to exclude it from the viewport display.");
     register("kFileAnn", "Load in a file as the stage source.");
+    register("kInvalidSelectionKind", "Invalid Maya Usd selection kind !");
```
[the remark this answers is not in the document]
spinell-adsk: Right I will remove the space.

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
spinell-adsk: I spoke with Natalia and she will think about it. So I think for now that should be fine

plugin/adsk/scripts/mayaUsdMenu.mel
spinell-adsk: Now that the build-in will not have a capital letter I will only use a special case for the empty one.

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
seando-adsk: Maybe just mention to Anisha where these strings will appear. I bet she though they were tooltips, and thus the length. She might reword them knowing they are on the statusbar. Also tell her, bold is not supported.

spinell-adsk: Help text updated

seando-adsk (approved):
It all looks good now. Thanks.

dj-mcg:
A few questions, which I can write up in separate issues if that makes the most sense for you:

1. Users also mentioned that the default USD Selection Mode of (none) was probably going to be unhelpful for everyone. They suggested perhaps a default of 'model' or 'subcomponent'
2. Is there a plan to add a Selection Mode for Point Instancer types (Instance/Instancer/Prototype)? We cobbled together a shelf item but it's a bit unsatisfying.

Thanks!

kxl-adsk:
@dj-mcg Let's have these conversations in github discussions - https://github.com/Autodesk/maya-usd/discussions.

For your questions:

1. Selection by kind is an optionVar, so you can configure it and have persistent between Maya sessions.
2. Please log an issue for this one. Changes will pretty much follow what's in this PR so don't hesitate if you want to take a stab at it and retire your shelf menu.

PR #1261, opened 2021-03-16
Explicitly case Ar path string to ArResolvedPath

dj-mcg:
Passing a path string to ArResolver::OpenAsset throws an error when built against the latest version of USD. This parameter is explicitly cast to ArResolvedPath elsewhere in the USD code, so updating this callsite to follow that pattern.

### DIFF
--- a/lib/usd/translators/shading/usdUVTextureReader.cpp
+++ b/lib/usd/translators/shading/usdUVTextureReader.cpp
@@ -266,7 +266,8 @@ bool PxrMayaUsdUVTexture_Reader::Read(UsdMayaPrimReaderContext* context)
             && !filePath.empty() && ArIsPackageRelativePath(filePath)) {
             // NOTE: (yliangsiew) Package-relatve path means that we are inside of a USDZ file.
             ArResolver& arResolver = ArGetResolver(); // NOTE: (yliangsiew) This is cached.
-            std::shared_ptr<ArAsset> assetPtr = arResolver.OpenAsset(filePath);
+            std::shared_ptr<ArAsset> assetPtr = arResolver.OpenAsset(
+                ArResolvedPath(filePath));
             if (assetPtr == nullptr) {
                 TF_WARN(
                     "The file: %s could not be found within the USDZ archive for extraction.",
--- a/lib/usd/translators/shading/usdUVTextureReader.cpp
+++ b/lib/usd/translators/shading/usdUVTextureReader.cpp
@@ -266,8 +266,7 @@ bool PxrMayaUsdUVTexture_Reader::Read(UsdMayaPrimReaderContext* context)
             && !filePath.empty() && ArIsPackageRelativePath(filePath)) {
             // NOTE: (yliangsiew) Package-relatve path means that we are inside of a USDZ file.
             ArResolver& arResolver = ArGetResolver(); // NOTE: (yliangsiew) This is cached.
-            std::shared_ptr<ArAsset> assetPtr = arResolver.OpenAsset(
-                ArResolvedPath(filePath));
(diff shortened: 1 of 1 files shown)


kxl-adsk:
FYI @ysiewappl

JGamache-autodesk (approved):
LGTM

lib/usd/translators/shading/usdUVTextureReader.cpp
```diff
@@ -266,7 +266,7 @@ bool PxrMayaUsdUVTexture_Reader::Read(UsdMayaPrimReaderContext* context)
             && !filePath.empty() && ArIsPackageRelativePath(filePath)) {
             // NOTE: (yliangsiew) Package-relatve path means that we are inside of a USDZ file.
             ArResolver& arResolver = ArGetResolver(); // NOTE: (yliangsiew) This is cached.
-            std::shared_ptr<ArAsset> assetPtr = arResolver.OpenAsset(filePath);
+            std::shared_ptr<ArAsset> assetPtr = arResolver.OpenAsset(ArResolvedPath(filePath));
```
kxl-adsk, line 269: @dj-mcg Please check PF results - https://github.com/Autodesk/maya-usd/pull/1261/checks?check_run_id=2125284262. Looks like this wasn't existing in USD 20.11. We currently support USD versions down to 20.02.

kxl-adsk (changes requested):

lib/usd/translators/shading/usdUVTextureReader.cpp
dj-mcg: Thanks @kxl-adsk - put in a PXR_VERSION check and running preflight now

kxl-adsk (approved):

JGamache-autodesk (approved):

kxl-adsk:
We had some slow network issues...that slowed down the PF enough to cause the timeout on GitHub Action side. The internal job (job/ecg-mayausd-branch-preflight-master-public/1128/) finished with success. No need to re-trigger the PF.

ysiewappl:
Sorry, something was bugging me and I just double-checked with @dgovil: I am already building against USD 21.02 here and the code builds fine without this check.

From what I can see:

https://graphics.pixar.com/usd/docs/668045551.html#AssetResolution(Ar)2.0-RolloutandTransition

> Ar 2.0 will substantially modify the ArResolver interface and other related classes. To maintain backwards compatibility with existing ArResolver implementations and client code sites, users will be able to specify whether they want to enable Ar 2.0 at build time. If enabled, the AR_VERSION flag will be defined and to 2 and the Ar 2.0 changes will be built and installed, causing old code to break if not updated to accommodate those changes. If disabled, AR_VERSION will be set to 1 and Ar will revert back to the code that exists today, ensuring no change in behavior or build breakages. 

> For its initial release, Ar 2.0 will be disabled by default – the current Ar implementation will remain the default for backwards compatibility, but will be deprecated. Users will be able to enable Ar 2.0 using the build flag to begin testing and transitioning their resolver implementations and client code to the new interface. Ar 2.0 will be enabled by default in the subsequent release, and the old Ar implementation will be removed at that time.

Should the macro be changed to check against this instead rather than the USD version itself, assuming people want to use the old resolver and haven't built their libUSD with this flag enabled?

kxl-adsk:
I'm not sure about the connection to Ar 2.0 here. We are internally building **without** Ar 2.0 which means PF is validating against an artifact that doesn't have it. The problem was that this change originally only worked for USD 21.02 and above. It was not compiling with older versions (like 20.11 down to 20.02). 

Based on that, why should the check be changed?

PR #1262, opened 2021-03-17
Add "/permissive-" flag for VS compiler

HamedSabri-adsk:
This PR, enables `/permissive-` flag for Visual Studio compiler. Read more here:

https://docs.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=msvc-160

I recently wrote a code where I bind a non-const reference to a temporary which was rightfully detected as error with GCC but surprisingly not with either VS compiler 2017/2019.

```
void foo(std::string* errMsg)
{
    std::string message{"potato_message"};
    *errMsg = message;
}

int main()
{
    // bad code: I am binding a temporary to a reference and by doing 
    // so I am creating a reference to an object that will be 
    // destroyed at the end of the expression.
    foo(&std::string());
}
```
godbolt pseudo-code:

https://godbolt.org/z/3KzhnM

The "/permissive-" option is supported in Visual Studio 2017 and later and it would be great to have it ON to detect cases like:

- two-phase name lookup
- binding a non-const reference to a temporary
- allowing multiple user-defined conversions in initialization
- treating copy init as direct init

You can read on "Two-phase name lookup support comes to MSVC" here:

https://devblogs.microsoft.com/cppblog/two-phase-name-lookup-support-comes-to-msvc/

### DIFF
--- a/cmake/compiler_config.cmake
+++ b/cmake/compiler_config.cmake
@@ -24,7 +24,7 @@ set(MSVC_FLAGS
     /W3
     $<$<BOOL:${BUILD_STRICT_MODE}>:/WX>
     # enable two-phase name lookup and other strict checks (binding a non-const reference to a temporary, etc..)
-    /permissive-
+    $<$<BOOL:$<VERSION_GREATER:${PXR_VERSION},2002>>:/permissive->
     # enable pdb generation.
     /Zi
     # standards compliant.
--- a/cmake/compiler_config.cmake
+++ b/cmake/compiler_config.cmake
@@ -23,6 +23,8 @@ set(MSVC_FLAGS
     # we want to be as strict as possible
     /W3
     $<$<BOOL:${BUILD_STRICT_MODE}>:/WX>
+    # enable two-phase name lookup and other strict checks (binding a non-const reference to a temporary, etc..)
+    /permissive-
     # enable pdb generation.
     /Zi
     # standards compliant.
--- a/cmake/compiler_config.cmake
+++ b/cmake/compiler_config.cmake
@@ -24,7 +24,7 @@ set(MSVC_FLAGS
     /W3
     $<$<BOOL:${BUILD_STRICT_MODE}>:/WX>
     # enable two-phase name lookup and other strict checks (binding a non-const reference to a temporary, etc..)
-    $<$<BOOL:$<VERSION_GREATER:${PXR_VERSION},2002>>:/permissive->
+    $<$<BOOL:$<VERSION_GREATER:${USD_BOOST_VERSION},106600>>:/permissive->
     # enable pdb generation.
     /Zi
     # standards compliant.
--- a/cmake/modules/FindUSD.cmake
+++ b/cmake/modules/FindUSD.cmake
@@ -101,9 +101,29 @@ elseif(USD_INCLUDE_DIR AND EXISTS "${USD_INCLUDE_DIR}/pxr/pxr.h")
     math(EXPR PXR_VERSION "${USD_MAJOR_VERSION} * 10000 + ${USD_MINOR_VERSION} * 100 + ${USD_PATCH_VERSION}")
 endif()
 
+# Get the boost version from the one built with USD
+if(USD_INCLUDE_DIR)
+    file(GLOB _USD_VERSION_HPP_FILE "${USD_INCLUDE_DIR}/boost-*/boost/version.hpp")
+    list(LENGTH _USD_VERSION_HPP_FILE found_one)
+    if(${found_one} STREQUAL "1")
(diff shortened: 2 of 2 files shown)


ppt-adsk (approved):

kxl-adsk (changes requested):
I really like this change and that we are trying to reduce the differences in strictness level between platforms. Unfortunately, as the change is right now, we can't merge it because older versions are failing to compile. Please try with USD 20.02.

After looking for 2 seconds into it, the issue is more related to the boost version, so we probably need to have some conditional here to check if the right version is used.

ppt-adsk (approved):

cmake/compiler_config.cmake
```diff
@@ -23,6 +23,8 @@ set(MSVC_FLAGS
     # we want to be as strict as possible
     /W3
     $<$<BOOL:${BUILD_STRICT_MODE}>:/WX>
+    # enable two-phase name lookup and other strict checks (binding a non-const reference to a temporary, etc..)
+    $<$<BOOL:$<VERSION_GREATER:${PXR_VERSION},2002>>:/permissive->
```
kxl-adsk, line 27: By checking the USD version here, you make assumptions that nobody changes the default boost version used to build the USD library. I wouldn't make this assumption and instead, add a check that validates the boost version since this is the culprit to the problem.

HamedSabri-adsk: Not quite sure if I follow what you mean by "changing the default boost version". 

We can't and shouldn't rely on Boost version at maya-usd top level cmake.

`${Boost_VERSION} ` is not populated unless you call `find_package(Boost REQUIRED)`. We tried so hard to eliminate such calls and bringing it back again doesn't make much sense.

Second, it makes more sense to rely on USD version because moving forward with USD 20.05, Pixar uses boost-1_70 and it would be highly unlikely that they decide to switch to an old version of boost for any future releases. 

https://github.com/PixarAnimationStudios/USD/blob/ebac0a8b6703f4fa1c27115f1f013bb9819662f4/build_scripts/build_usd.py#L631

I agree with you that the culprit is boost. If it helps I can add a comment to clarify it.

kxl-adsk: Not everyone builds USD with the default configuration and you exactly pointed to the place where one can change the boost version. This means you can build older versions of USD with a newer boost version (or vice versa).

seando-adsk: I see and understand both your points. The only component that relies on Boost in the repo is "plugin/al" (that is the only place "find_package(Boost)" is called, thus here in this top-level compiler_config.cmake we don't know the boost version. I agree with Hamed that we should not add a find_package(Boost) to the top-level cmake. So the only thing we can rely on here is the USD version. Unless you have another idea Krystian?

kxl-adsk: The comment in 400532bb1b0531cf83c9d556a2bdffd942d6f6fb talks about boost python - we do use boost python for core module python bindings. What am I missing that prevents us from getting the boost version?

seando-adsk: It is a bit confusing. Yes we use boost python in the core, which we get via inheritance from linking with USD libs such as Tf. From USD we get the include path to boost (inside the USD artifact) and the boost lib(s). Using boost in this fashion does **not** give us the boost version. 

One way to get the boost version would be to read it from the boost file _version.hpp_. I suppose we modify our FindUSD.cmake to do this just like it gets the USD_VERSION from prx.h.

HamedSabri-adsk: @kxl-adsk The boost python dependency is propagated as soon as we link against ANY of the Pixar USD libraries ( e.g usd, usdImaging, etc... ).

image

If you recall, such dependencies are leaking from Pixar libraries. So even if you have a target that has nothing to do with Boost or python, you still need to make sure that target links against such libraries.

kxl-adsk: I see, this makes much more sense now. Thank you. I'm guessing what Sean proposed is still pretty much doable?

> modify our FindUSD.cmake to do this just like it gets the USD_VERSION from prx.h

HamedSabri-adsk: I would prefer to not make this change for the sake of this PR. We could revisit this at any time **if** any problems rises as a result of this change.

kxl-adsk: The change is FindUSD.cmake looks straight forward and we already have a pattern to copy from. I spoke with @seando-adsk and he will help with it.

seando-adsk (approved):
Since it was my suggestion to get the boost version from the boost file in the USD build, I added the code for it.

ppt-adsk (approved):

kxl-adsk (approved):
Awesome job! Looks to be working great.

PR #1267, opened 2021-03-19
MAYA-108220 - As a user, on the prim AE template I'd like to see arra…

spinell-adsk:
…y attributes suppressed

Add a new option to hide/display array attributes in AE

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
@@ -229,8 +229,11 @@ class AETemplate(object):
 
         # Get the UFE Attributes interface for this scene item.
         self.attrS = ufe.Attributes.attributes(self.item)
+        self.suppressedAttrs = []
 
-        self.showArrayAttributes = cmds.optionVar(query="mayaUSD_AEShowArrayAttributes")
+        self.showArrayAttributes = False
+        if cmds.optionVar(exists="mayaUSD_AEShowArrayAttributes"):
+            self.showArrayAttributes = cmds.optionVar(query="mayaUSD_AEShowArrayAttributes")
 
         cmds.editorTemplate(beginScrollLayout=True)
         self.buildUI()
@@ -238,21 +241,15 @@ class AETemplate(object):
         self.createMetadataSection()
         cmds.editorTemplate(endScrollLayout=True)
 
-        # Suppress all array attributes not used
-        if not self.showArrayAttributes:
-            for attrName in self.attrS.attributeNames:
-                if self.isArrayAttribute(attrName):
-                    self.suppress(attrName)
 
-
-    @staticmethod
-    def addControls(controls):
+    def addControls(self, controls):
         for c in controls:
-            cmds.editorTemplate(addControl=[c])
+            if c not in self.suppressedAttrs:
+                cmds.editorTemplate(addControl=[c])
 
-    @staticmethod
-    def suppress(control):
+    def suppress(self, control):
         cmds.editorTemplate(suppress=control)
+        self.suppressedAttrs.append(control)
 
     @staticmethod
     def defineCustom(customObj, attrs=[]):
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -230,17 +230,30 @@ def __init__(self, ufeSceneItem):
...
         self.buildUI()
         cmds.editorTemplate(addExtraControls=True)
         self.createMetadataSection()
         cmds.editorTemplate(endScrollLayout=True)
 
+        # Suppress all array attributes not used
+        if not self.showArrayAttributes:
+            for attrName in self.attrS.attributeNames:
+                if self.isArrayAttribute(attrName):
+                    self.suppress(attrName)
```
seando-adsk, line 245: Maya code will issue a warning if you "add then suppress" or "suppress and then add" an attribute. You are only supposed to call one or the other. So you should rework this somehow to avoid calling the addControls() on an item that you suppressed here.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -230,17 +230,30 @@ def __init__(self, ufeSceneItem):
         # Get the UFE Attributes interface for this scene item.
         self.attrS = ufe.Attributes.attributes(self.item)
 
+        self.showArrayAttributes = cmds.optionVar(query="mayaUSD_AEShowArrayAttributes")
```
seando-adsk, line 233: There is no default value for this optionVar, so if the user never opens the show menu it will not be set. Therefore you should account for that here and first set your class var to a good default (false), then test if the optionVar exists and if so query it.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -296,11 +309,16 @@ def createTransformAttributesSection(self, sectionName, attrsToAdd):
         geomX = UsdGeom.Xformable(self.prim)
         xformOps = geomX.GetOrderedXformOps()
         xformOpOrderNames = [op.GetOpName() for op in xformOps]
-        xformOpOrderNames.append(UsdGeom.Tokens.xformOpOrder)
+
+        if self.showArrayAttributes:
+            xformOpOrderNames.append(UsdGeom.Tokens.xformOpOrder)
```
seando-adsk, line 314: Are you sure about this? Even though attribute is an array type, I think we always want to show not matter what the value of the optionVar is.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -335,8 +353,19 @@ def buildUI(self):
             if schemaType.pythonClass:
                 attrsToAdd = schemaType.pythonClass.GetSchemaAttributeNames(False)
 
+                # remove all array attriutes from the attrsToAdd list
+                if not self.showArrayAttributes:
+                    attrsToAdd = [x for x in attrsToAdd if not self.isArrayAttribute(x)]
```
seando-adsk, line 358: Typo in your comment "attributes".
Here you are calling isArrayAttribute() again on ones that you already called on (from the __init__). We should only call this method once on each attribute.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
spinell-adsk: According the spec, the token[] should be hidden. I will double check with Nat if we want a exception for this one

seando-adsk (changes requested):
Okay except for a rework of calling only add or suppress on each attribute.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
spinell-adsk: this is to avoid calling addControl on the attribute. The one in init is called for suppress the attribute. Otherwise the attribute will be in the "Extra" section

seando-adsk: Understood, but calling isArrayAttribute twice is a little overkill. How about if where you first call it (to suppress) you maintain a list. Then here or maybe even better in the addControls you can check the list.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
spinell-adsk: Ok I spoke with Nat and xformOpOrder will be an exception

seando-adsk (approved):
Great. It all looks good now.

spinell-adsk:
Note: Only mel and python files have changed, so I did not run the preflight.

PR #1274, opened 2021-03-24
Spinell/maya 110157/clearing mutiple layer crash

spinell-adsk:
- Remove warning comfirmation dialog box when clearing a layer. (That cause the Layer editor to refresh when clearing multiple layer which can cause a dangling pointer and crash)
- Suspend Usd Notices until all layers are cleared.

When clearing multiple layer at the same time, maya crash because of a dangling pointer. This happen when in the selected layers, there is a sub layer of another layer. When clearing layers, we are first getting a list of all selected layers than loop over the list and clear layer one by one (Same pattern as remove layer). The problem here is after the first clear, we are notified by USD that a layer changed and we are queuing a Layer Editor refresh on idle. On the next loop iteration, items can become a dangling pointer to an object that have been deleted if the Layer Editor refresh already happen (During the refresh all the items are deleted are recreated). 

Let suppose you have the fallowing hierarchy:
&nbsp; layer1
&nbsp;&nbsp; layer2
&nbsp;&nbsp;&nbsp; layer3
&nbsp;&nbsp;&nbsp;&nbsp; layer4 

If all layers are selected and layer1 is cleared first, sub layer2 is removed from layer1 and a Layer Editor refresh callback is set on idle. If the refresh happen before we finish to iterate on the selected items, the selection list will contains dangling pointer.

To be sure the Layer Editor is not refreshing before we finish to clear all layers, I'm disabling USD notification and I'm forcing a refresh after all layer are cleared.

Note : 
Before clearing each layer, a warning dialog box is displayed and maya become idle so the Layer Editor get refresh and the selection list become invalidate. I spoke with Natalia and we conclude that displayed a warning dialog box when clearing a layer does not really make sense because this operation is undo-able and we don't display a warning when removing a layer. So I removed the confirmation box.

### DIFF
--- a/lib/usd/ui/layerEditor/layerTreeItem.cpp
+++ b/lib/usd/ui/layerEditor/layerTreeItem.cpp
@@ -396,19 +396,7 @@ void LayerTreeItem::printLayer()
 
 void LayerTreeItem::clearLayer()
 {
-    MString title;
-    title.format(
-        StringResources::getAsMString(StringResources::kClearLayerTitle),
-        MQtUtil::toMString(text()));
-
-    MString desc;
-    desc.format(
-        StringResources::getAsMString(StringResources::kClearLayerConfirmMessage),
-        MQtUtil::toMString(text()));
-
-    if (confirmDialog(MQtUtil::toQString(title), MQtUtil::toQString(desc))) {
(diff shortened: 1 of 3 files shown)


seando-adsk (approved):
Looks good to me.

PR #1278, opened 2021-03-26
MAYA-109387 add more camera tests to check that animated cameras

williamkrick:
and modified cameras update correctly in VP2.


ppt-adsk (approved):

kxl-adsk (changes requested):
`testVP2RenderDelegateUsdCamera` is failing (tested with Maya 2022 Gold and USD 21.02). Please investigate.

williamkrick:
Somehow I got into a state where I have two pull requests for similar changes, this one and #1298 . I will close this one and finish the work there.

The test fails in 2022 gold because it relies on some additional Maya changes for correct behavior. I will figure out how to conditionally run the parts of the test which depend on the updated Maya behavior so that the test can pass in 2022 gold & test the corrected behavior in future versions of Maya.

PR #1280, opened 2021-03-26
Don't write fStop when depthOfField is disabled.

marsupial:
Maya has an explicit attribute to enable/disable depth of field, but **fStop=0** does that in USD.
Currently USD import/export is a direct translation of the USD fStop attribute, which changes intent/meaning.

TODO: Might be good to export the maya fStop curve in a namespaced custom property for loseless data **Maya->USD->Maya**.


lib/mayaUsd/fileio/translators/translatorCamera.cpp
```diff
@@ -470,9 +471,17 @@ bool _ReadToCamera(
     }
 
     usdAttr = usdCamera.GetFStopAttr();
-    plugName = _tokens->MayaCameraAttrNameFStop;
-    if (!_TranslateUsdAttributeToPlug(usdAttr, cameraFn, plugName, args, context)) {
-        return false;
+    if (usdAttr.IsAuthored()) {
```
williamkrick, line 474: Is the IsAuthored() test enough to catch all cases where fStop is zero? Looking at the docs it says it'll return true if there any any opinions on any layer, so if there is an opinion which sets it zero I think we'll get into this code block which represents a non-zero fStop.

lib/mayaUsd/fileio/translators/translatorCamera.cpp
```diff
@@ -470,9 +471,17 @@ bool _ReadToCamera(
...
+        plugName = _tokens->MayaCameraAttrNameFStop;
+        if (!_TranslateUsdAttributeToPlug(usdAttr, cameraFn, plugName, args, context)) {
+            return false;
+        }
+        // Enable depthOfField
+        plugName = _tokens->MayaCameraAttrNameDepthOfField;
+        MPlug plug = cameraFn.findPlug(plugName.GetText(), true, &status);
```
williamkrick, line 481: I think we should have an else clause which sets Maya's depth of field attribute false when fStop is zero.

williamkrick (changes requested):

lib/mayaUsd/fileio/translators/translatorCamera.cpp
marsupial: Yes, thought I added a **FIXME** in there to that effect, but I guess not.

The reason I didn't tackle it was that _really_ checking for zero could incur iteration over all time samples, and I thought that part would best best be left to phase 2 if Maya import/export had to write a custom property for keyframes.

If we're not going to be worrying about time sampled data for now (which for our case isn't currently relevant) then I'll go ahead and add the read-value.

lib/mayaUsd/fileio/translators/translatorCamera.cpp
marsupial: Can the import/export re-attach to existing cameras ? Otherwise dof is off by default in Maya.

JGamache-autodesk (changes requested):
Thanks for the fix. I would have expected that at least one of the exported cameras in the unit tests would have a disabled depth of field on export to test that the USD stream ends up having the right values, and it would be nice to have an import test with an fStop of zero to exercise the new behavior fully.

williamkrick (approved):

marsupial:
Added the test..

JGamache-autodesk (approved):

kxl-adsk:
@marsupial please apply clang-format so we can trigger PF for this change.

kxl-adsk:
The two failing MacOS are known issues unrelated to this change. This change is now ready for merge.

PR #1281, opened 2021-03-26
[mtoh] Fix Hydra prim invalidation for hidden lights that are never population

marsupial:
Lights translation/population doesn't occur when the light is invisible, so subsequent attempt to invalidate the prim issue errors as it doesn't exists.


pmolodo:
Pretty straightforward fix, thanks!

pmolodo (approved):
Pretty straightforward fix, thanks!

PR #1282, opened 2021-03-26
[mtoh] Add camera adapter to support physical camera parameters and configurable motion blur and depth-of-field.

marsupial:
_Running the test-suite locally to verify nothing has been broken, but as the change-set is quite large posting it now_

Currently mtoh does motion-blur with two hard-coded samples **t=frame** and **t=frame+1**. We'd like this to be configurable, as these values/ranges are rarely used by our camera's, which are usually centered around the current frame.  Additionally, reliance in the Hydra-projection matrix only for the camera means depth-of-field isn't currently supported.

These changes shouldn't affect Storm or other delegates that only use the Hydra render-pass view matrix.
 
- Removes **mtohEnableMotionSamples**, in favor of explicit shutter start and end settings. Setting **start=0**, **end=1** with a camera's **shutterAngle=360** is the equivalent of the old hard-coded mode.
- **HdMayaSceneDelegate::InsertDag** is refactored and templatized for the common pattern in adapter creation.
- **HdMayaMeshAdapter::SamplePrimvar** and **HdMayaDagAdapter::SampleTransform** call through to a templatized  **HdMayaDelegate::SampleValues** method.
- Hydra **HdCameraTokens->windowPolicy** is exported as **CameraUtilMatchHorizontally** or **CameraUtilMatchVertically** to avoid any potential conversion issues with Maya's **Fill** or **Overscan** modes leaking out to HdRenderDelagtes.
- **HdMayaDagAdapter** matrix cache is reduced from two matrices to one.

**TODO:** Need a way to use the Maya viewport camera in a UsdProxyShape.  Waiting until we can actually get the USD camera path when selected in the viewport to see what else can be done....The only way I can think of doing this right now is to build up a camera in the UsdProxyShape's stage, which has some downsides.

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
@@ -30,6 +30,7 @@
 #include <pxr/base/tf/instantiateSingleton.h>
 #include <pxr/base/vt/value.h>
 #include <pxr/imaging/glf/contextCaps.h>
+#include <pxr/imaging/hd/camera.h>
 #include <pxr/imaging/hd/rendererPluginRegistry.h>
 #include <pxr/imaging/hd/rprim.h>
 #include <pxr/imaging/hdx/pickTask.h>
@@ -468,18 +469,6 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext)
     // }
     TF_DEBUG(HDMAYA_RENDEROVERRIDE_RENDER).Msg("MtohRenderOverride::Render()\n");
     auto renderFrame = [&](bool markTime = false) {
-        const auto originX = 0;
-        const auto originY = 0;
-        int        width = 0;
-        int        height = 0;
-        drawContext.getRenderTargetSize(width, height);
-
-        GfVec4d viewport(originX, originY, width, height);
-        _taskController->SetFreeCameraMatrices(
-            GetGfMatrixFromMaya(drawContext.getMatrix(MHWRender::MFrameContext::kViewMtx)),
-            GetGfMatrixFromMaya(drawContext.getMatrix(MHWRender::MFrameContext::kProjectionMtx)));
-        _taskController->SetRenderViewport(viewport);
-
         HdTaskSharedPtrVector tasks = _taskController->GetRenderingTasks();
 
         // For playblasting, a glReadPixels is going to occur sometime after we return.
@@ -583,7 +572,52 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext)
 
     params.cullStyle = HdCullStyleBackUnlessDoubleSided;
 
+    int width = 0;
+    int height = 0;
+    drawContext.getRenderTargetSize(width, height);
+
+    bool vpDirty;
+    if ((vpDirty = (width != _viewport[2] || height != _viewport[3]))) {
+        _viewport = GfVec4d(0, 0, width, height);
+        _taskController->SetRenderViewport(_viewport);
+    }
+
+    _taskController->SetFreeCameraMatrices(
+        GetGfMatrixFromMaya(drawContext.getMatrix(MHWRender::MFrameContext::kViewMtx)),
+        GetGfMatrixFromMaya(drawContext.getMatrix(MHWRender::MFrameContext::kProjectionMtx)));
+
+    if (delegateParams.motionSamplesEnabled()) {
+        MStatus  status;
+        MDagPath camPath = getFrameContext()->getCurrentCameraPath(&status);
+        if (status == MStatus::kSuccess) {
+            // FIXME: This is what a USD camera selected in the viewport returns.
+            static const MString defaultUfeProxyCameraShape(
+                "|defaultUfeProxyCameraTransformParent|defaultUfeProxyCameraTransform|"
+                "defaultUfeProxyCameraShape");
+            if (defaultUfeProxyCameraShape != camPath.fullPathName()) {
+                for (auto& delegate : _delegates) {
+                    if (HdMayaSceneDelegate* mayaScene
+                        = dynamic_cast<HdMayaSceneDelegate*>(delegate.get())) {
+                        params.camera = mayaScene->SetCameraViewport(camPath, _viewport);
+                        if (vpDirty)
+                            mayaScene->GetChangeTracker().MarkSprimDirty(
+                                params.camera, HdCamera::DirtyParams | HdCamera::DirtyProjMatrix);
+                        break;
+                    }
+                }
+            }
+        } else {
+            TF_WARN(
+                "MFrameContext::getCurrentCameraPath failure (%d): '%s'"
+                "\nUsing viewport matrices.",
+                int(status.statusCode()),
+                status.errorString().asChar());
+        }
+    }
+
     _taskController->SetRenderParams(params);
+    if (!params.camera.IsEmpty())
+        _taskController->SetCameraPath(params.camera);
 
     // Default color in usdview.
     _taskController->SetSelectionColor(_globals.colorSelectionHighlightColor);
@@ -787,6 +821,7 @@ void MtohRenderOverride::ClearHydraResources()
         _rendererPlugin = nullptr;
     }
 
+    _viewport = GfVec4d(0, 0, 0, 0);
     _initializationSucceeded = false;
     _initializationAttempted = false;
     SelectionChanged();
--- a/lib/usd/hdMaya/adapters/cameraAdapter.h
+++ b/lib/usd/hdMaya/adapters/cameraAdapter.h
@@ -1,0 +1,76 @@
+//
+// Copyright 2021 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef HDMAYA_CAMERA_ADAPTER_H
+#define HDMAYA_CAMERA_ADAPTER_H
+
+#include <hdMaya/adapters/dagAdapter.h>
+#include <hdMaya/adapters/shapeAdapter.h>
+
+#include <pxr/pxr.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+class HdMayaCameraAdapter : public HdMayaShapeAdapter
+{
+public:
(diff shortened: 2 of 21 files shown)


lib/usd/hdMaya/adapters/cameraAdapter.h
```diff
@@ -0,0 +1,73 @@
...
+    HDMAYA_API
+    void CreateCallbacks() override;
+
+    HDMAYA_API
+    void SetViewport(const GfVec4d& viewport);
+
+protected:
+    static TfToken CameraType();
+
+    std::unique_ptr<GfVec4d> _viewport;
```
pmolodo, line 66: Can you add a quick note explaining that this is a ptr so that we can detect if the viewport has been set yet (and provide default behavior if it hasn't)?

lib/usd/hdMaya/adapters/dagAdapter.h
```diff
@@ -57,7 +57,7 @@ class HdMayaDagAdapter : public HdMayaAdapter
     HDMAYA_API
     virtual void RemovePrim() override;
     HDMAYA_API
-    const GfMatrix4d& GetTransform();
+    GfMatrix4d GetTransform();
```
pmolodo, line 60: I'm guessing this was changed from `const GfMatrix4d &` to `GfMatrix4d` just to make it more consistent with `HdSceneDelegate::GetTransform`?

lib/usd/hdMaya/delegates/delegate.h
```diff
@@ -124,6 +128,55 @@ class HdMayaDelegate
...
+    GfInterval GetCurrentTimeSamplingInterval() const;
+
+    /// Common function to return templated sample types 
+    template <typename T, typename Getter> HDMAYA_API
+    size_t SampleValues(size_t maxSampleCount, float* times, T* samples, Getter getValue)
+    {
+        if (ARCH_UNLIKELY(maxSampleCount == 0)) {
+            return 0;
+        }
+        if (maxSampleCount == 1 || !GetParams().enableMotionSamples()) {
```
pmolodo, line 149: Feel like this is technically wrong behavior for `maxSampleCount == 1` if, say, `startTime = endTime = .25`.  Instead,  I think we should just set `tStep = 0` if `maxSampleCount == 1`, and otherwise I think the "normal" code branch should work.

lib/usd/hdMaya/delegates/sceneDelegate.cpp
```diff
@@ -503,6 +503,35 @@ HdMayaMaterialAdapterPtr HdMayaSceneDelegate::GetMaterialAdapter(const SdfPath&
...
+    const std::function<AdapterPtr(HdMayaDelegateCtx*, const MDagPath&)>& adapterCreator,
+    Map& adapterMap, bool isSprim)
+{
+    if (!adapterCreator) {
+        return {};
+    }
+
+    TF_DEBUG(HDMAYA_DELEGATE_INSERTDAG)
+        .Msg(
+            "HdMayaSceneDelegate::InsertDag::"
```
pmolodo, line 517: Assuming this should be changed to `"HdMayaSceneDelegate::Create::"`?

pmolodo (changes requested):
Overall makes sense to me - just a few small questions / suggestions.

lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
```diff
@@ -583,7 +572,46 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext)
...
+    _taskController->SetFreeCameraMatrices(
+        GetGfMatrixFromMaya(drawContext.getMatrix(MHWRender::MFrameContext::kViewMtx)),
+        GetGfMatrixFromMaya(drawContext.getMatrix(MHWRender::MFrameContext::kProjectionMtx)));
+
+    if (delegateParams.enableMotionSamples()) {
+        MStatus status;
+        MDagPath camPath = getFrameContext()->getCurrentCameraPath(&status);
+        if (status == MStatus::kSuccess) {
+            // FIXME: This is what a USD camera selected in the viewport returns.
+            static const MString defaultUfeProxyCameraShape("|defaultUfeProxyCameraTransformParent|defaultUfeProxyCameraTransform|defaultUfeProxyCameraShape");
```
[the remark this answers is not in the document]
kxl-adsk: That will be @williamkrick. I believe the answer is: no, not at the moment. I count on William to correct me.

lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
```diff
@@ -583,7 +572,46 @@ MStatus MtohRenderOverride::Render(const MHWRender::MDrawContext& drawContext)
...
 
+    int        width = 0;
+    int        height = 0;
+    drawContext.getRenderTargetSize(width, height);
+
+    bool       vpDirty;
+    if ((vpDirty = (width != _viewport[2] || height != _viewport[3]))) {
+        _viewport = GfVec4d(0, 0, width, height);
+        _taskController->SetRenderViewport(_viewport);
+    }
```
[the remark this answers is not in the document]
marsupial: Yeah, vpDirty is what I want for the camera-prim invalidation.
MFrameContext::kViewMtx would be a transform change already picked up via the dag callback for a camera.

I'll double check the following is correct, but viewport changes dirties projection because of "fill", "overscan", "horizontal", "fitting" and how it can relate to the **fov** not of the camera, but inorder to to match the OpenGL viewport of the camera.  i.e. nothing has changed parameter wise, but we still want the render delegate to pull a new value for the **projectionMatrix** or **focalLength** to match OpenGL.

lib/usd/hdMaya/adapters/dagAdapter.h
marsupial: Pretty much.  The reference was always copied anyway.

lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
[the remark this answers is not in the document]
williamkrick: We have a lot of work to do to fully support USD cameras in Maya and we don't have a way to access the Ufe::Path to the USD camera being used to in the viewport yet.

Once we do have functionality to handle USD cameras better I believe the code written here will continue to function as intended. MFrameContext::getCurrentCameraPath() will still exist and continue to return the DAG path to the proxy camera shape when the view is looking through a Ufe camera.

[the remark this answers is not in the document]
ppt-adsk: This fixed string makes me a bit nervous...  This should be an implementation detail that no one needs to care about.  So yes, let's definitely circle back when we have something better to use.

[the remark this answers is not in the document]
pmolodo: Definitely... but, since it sounds like nobody has a better solution for now, I'm gonna mark this as ok for this PR...

lib/usd/hdMaya/delegates/delegate.h
marsupial: I'm fine with that, but perhaps @kxl-adsk can clarify, is there any overhead to the `MDGContextGuard` for the case where we can just be querying maya value's directly ?
i.e Should we keep an optimized case of 1 sample where `startTime == 0` vs the below
```
MDGContextGuard guard(<CURRENT_MAYA_TIME>);
...Query Maya Values...
```

lib/usd/hdMaya/delegates/sceneDelegate.cpp
marsupial: Sure can add it.
I am curious if anyone is using the debug-message logging features?
I never do, and personally find it a bit harder to read the code with them...

pmolodo: Oh, I definitely find them very useful!  The targeted feedback is great - what you get if you just do `TF_DEBUG=*` can be overwhelming (though occasionally still useful too).

marsupial:
Think I covered it all. Closed out the -easy- issues, but left the more complicated ones open for you to do the same!

lib/usd/hdMaya/delegates/delegate.h
marsupial: Keeping the _fast path_ for one sample where `sampleTime=currentFrame` for now, should be easy to delete the lines if they proove to be irrelevant.

Also in the case of one-sample, I'm using the center of the shutter-range so 1 sample for [-0.25, -0.25] and [-1, 1] would both be **t=0**.  It seems a bit more intuitive to me.

pmolodo (approved):
All the changes look good!

williamkrick (approved):

kxl-adsk:
@marsupial Can you please apply clang-format so we can get PF results. Thank you.

lib/usd/hdMaya/delegates/delegate.h
kxl-adsk: >  is there any overhead to the MDGContextGuard for the case where we can just be querying maya value's directly ?

Each node has it state and each state is associated with a context. When you start a new context computation, you will have to compute the state for it. We don't assume that what you need are results from normal context (current time). Reading values from the current time should happen within a normal context.

marsupial:
Hopefully latest will pass pre-flight.
Squashed everything down as it just seemed growing with too many correction-commits.

kxl-adsk:
@marsupial Clang-format worked...but there are some build failures on windows.

lib/usd/hdMaya/delegates/delegate.h
marsupial: Thanks.  The fast-path of 1 sample at t=0 is still there.

marsupial:
I took a wild guess at why MSVC is complaining about my *_perfectly legal_* code.
If it doesn't work, it'll probably be a lot faster to build the PR on a Windows machine (which I don't have access to) than me taking wild stabs.

-Thanks

kxl-adsk:
@marsupial 
There looks to be some sort of symbol clash on Windows. I won't have time to investigate, but here is a working patch that makes your changes successfully build on Windows 
[cameraAdapter.patch.txt](https://github.com/Autodesk/maya-usd/files/6273859/cameraAdapter.patch.txt)

There is one additional point that makes me slightly nervous about this change - there is no regression test added to cover the changes you made. I wouldn't want us to rely on manual testing to validate that things continue to work since that's not going to work with all permutations of versions that we must support.

marsupial:
Thanks for the patch...woulda taken me forever to figure that out change-submit-change-submit style.

I agree about the testing, but sadly Storm (and your VP-2 delegate) doesn't support _HdCamera_ or motion-blur, best I could offer is some images proving our delegate matches the GL-Viewport in all fit-modes, and editing shutter start, end, and angle gives the expected results.  Nothing was changed in the delegate for this patch, which matches what _HdKarma_ does in Houdini.

This was written so we can start doing more advanced play-blasting from Maya, so it should be getting exercised and any breakage will be reported to me pretty quickly.

marsupial:
Should be passing now, just a pref-name change.

kxl-adsk (approved):
LGTM, thank you.

PR #1284, opened 2021-03-27
Fix referenced material non-default attributes not getting exported to USD

### LINKED ISSUE #796: [MAYA-107837] USD Exporter doesn't write out non-default material values coming from a referenced file
**Describe the bug**
When a Maya file references another Maya file, the usd exporter treats all values on materials as being defaults, even if they aren't. This results in a usd file that doesn't correctly represent the scene.

**Steps to reproduce**
1. Create a Maya scene with a usd preview surface
2. Modify the shader values
3. Export a usda
4. Create a new Maya scene and reference the first one
5. Export a usda

**Expected behavior**
I expect to see material values written out that reflect the values in the scene like so

```
def Shader "AssetMat"
            {
                uniform token info:id = "UsdPreviewSurface"
                float inputs:clearcoat = 0.41216215
                color3f inputs:diffuseColor.connect = </AssetGroup/Looks/pxrUsdPreviewSurface1SG/file1.outputs:rgb>
                float inputs:metallic = 0.49324325
                float inputs:roughness = 0.79075
                token outputs:displacement
                token outputs:surface
            }
```

instead I get no values like this

```
def Shader "asset_AssetMat"
            {
                uniform token info:id = "UsdPreviewSurface"
                color3f inputs:diffuseColor.connect = </AssetGroup/Looks/pxrUsdPreviewSurface1SG/asset_file1.outputs:rgb>
                token outputs:displacement
                token outputs:surface
            }
```

**Attachments**
The attached file has two maya files, one that references the other (with a `_ref` suffix), and a similar set of usda exports. You can see the difference in the Shader values between the non-ref and ref files
[bugreports.zip](https://github.com/Autodesk/maya-usd/files/5272360/bugreports.zip)

**Specs (if applicable):**
 - macOS Catalina and Big Sur
 - Compiler version: Apple clang version 12.0.0 (clang-1200.0.32.5) (reproducible on older clang versions too)
 - Maya 2018.5 and 2020.0
 - Maya USD commit SHA 4365cc6675f666ee8ec23bc21c12ef68b3a5d5c7
 - Pixar USD commit SHA 71b4baace2044ea4400ba802e91667f9ebe342f0

**Additional context**
We are still using `pxrUSD.bundle` to do our exporting.

ysiewappl:
Hi:

This PR fixes #796, where referenced material attributes at non-default values were not getting exported to USD, causing a discrepancy in what was shown in Maya compared to what was getting exported.

### DIFF
--- a/lib/mayaUsd/utils/util.cpp
+++ b/lib/mayaUsd/utils/util.cpp
@@ -1211,7 +1211,7 @@ bool UsdMayaUtil::IsPlugDefaultValue(const MPlug& plug)
     // the source of a connection or is not connected at all, it's
     // authored-ness only depends on its own value, which is checked below.
     if (plug.isDestination(&status)) {
-        return true;
+        return false;
     }
 
     if (plug.isDefaultValue(true, &status)) {
--- a/lib/mayaUsd/utils/util.cpp
+++ b/lib/mayaUsd/utils/util.cpp
@@ -1198,6 +1198,29 @@ bool UsdMayaUtil::IsAuthored(const MPlug& plug)
     return false;
 }
 
+bool UsdMayaUtil::IsPlugDefaultValue(const MPlug& plug)
+{
+    MStatus status;
+
+    if (plug.isNull(&status) || status != MS::kSuccess) {
+        return false;
+    }
+
+    // Plugs that are the destination of a connection are considered authored,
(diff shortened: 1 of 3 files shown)


JGamache-autodesk (approved):
This should fix the issue for now. Thanks!

PR #1288, opened 2021-03-29
MAYA-110566 - No tooltip for Prim Path in Attribute Editor

spinell-adsk:
add missing tooltips


spinell-adsk:
no preflight ! I only change mel file.

seando-adsk (approved):

PR #1298, opened 2021-04-01
MAYA-109387 test usd camera animated params and editing params

williamkrick:
Test the Ufe::Camera python interface.
Test Maya rendering using Ufe cameras and UsdCamera, including animated camera transforms and camera parameters.


test/lib/ufe/testCamera.py
```diff
@@ -0,0 +1,245 @@
...
+        self.assertAlmostEqual(ufeCamera.nearClipPlane(), 1)
+        self.assertAlmostEqual(ufeCamera.farClipPlane(), 50)
+
+    def _TestHorizontalAperture(self, ufeCamera, usdCamera):
+        usdAttr = usdCamera.GetAttribute('horizontalAperture')
+        horizontalAperture = usdAttr.Get()
+
+        # the USD schema specifics that horizontal aperture is authored in mm
+        # the ufeCamera gives us inches
+        self.assertAlmostEqual(self.mmToCm * horizontalAperture, self.inchesToCm * ufeCamera.horizontalAperture())
```
williamkrick, line 147: I don't have a good pattern here for doing the unit conversions.

test/lib/ufe/testCamera.py
```diff
@@ -0,0 +1,245 @@
...
+
+        # setting camera values through Ufe is not yet implemented in MayaUSD
+
+        usdAttr.Set(0.5)
```
williamkrick, line 151: USD doesn't do any conversion, so the Set(0.5) isn't setting the horizontal aperture to 0.5 scene units, but to 0.5mm, the defined linear unit for horizontal aperture.

lib/mayaUsd/ufe/UsdCamera.cpp
```diff
@@ -181,7 +181,13 @@ float UsdCamera::fStop() const
         stageUnits = UsdGeomGetStageMetersPerUnit(stage);
     }
 
+
```
williamkrick, line 184: In 2022 Gold we set this value in Maya without sanitizing the input and a value of 0 (USD's default value) asserts.

williamkrick:
I believe the test will pass on 2022 gold now, which is where it had an issue before.

ppt-adsk (approved):

ppt-adsk (approved):

williamkrick:
Okay now I just need to wait on master getting changes merged for MAYA-109387 and this will be ready.

ppt-adsk (approved):

PR #1312, opened 2021-04-06
Fix unused variable compiler warning

seando-adsk:


lib/mayaUsd/ufe/UsdSceneItem.cpp
```diff
@@ -73,13 +73,13 @@ std::vector<std::string> UsdSceneItem::ancestorNodeTypes() const
...
-    const auto&         schemaReg = UsdSchemaRegistry::GetInstance();
     std::vector<TfType> tfAncestorTypes;
     schemaType.GetAllAncestorTypes(&tfAncestorTypes);
     for (const TfType& ty : tfAncestorTypes) {
         // If there is a concrete schema type name, we'll return that since it is what
         // is used/shown in the UI (ex: 'Xform' vs 'UsdGeomXform').
 #if PXR_VERSION >= 2005
+        const auto& schemaReg = UsdSchemaRegistry::GetInstance();
         strAncestorTypes.emplace_back(
             schemaReg.IsConcrete(ty) ? schemaReg.GetSchemaTypeName(ty) : ty.GetTypeName());
```
seando-adsk, line 84: The variable schemaReg is only used here (line 83), so I moved it inside this #if to remove an "unused variable" compiler warning on Linux (that gets promoted to error since we have strict mode on).

kxl-adsk (approved):

PR #1315, opened 2021-04-07
MAYA-109387 Send Ufe::CameraChanged when camera attributes are modified.

williamkrick:


lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -69,7 +71,36 @@ bool inAttributeChangedNotificationGuard()
     return attributeChangedNotificationGuardCount.load() > 0;
 }
 
-std::unordered_map<Ufe::Path, std::string> pendingAttributeChangedNotifications;
+std::unordered_map<Ufe::Path, TfToken> pendingAttributeChangedNotifications;
```
williamkrick, line 74: Should this be an unordered_multimap? If the same path is modified twice with different tokens then the second token will overwrite the first one and the first token will never be in an AttributeValueChanged notification.

ppt-adsk: I think you're right!  You've uncovered a latent bug, as far as I can tell.  An SdfPath can encode the path to an attribute, but a Ufe::Path cannot.  Therefore, if multiple attributes change on a UsdPrim, we'd lose data as you describe.

Do you mind making the change in this changeset?  It would minimize overhead, and it feels fairly direct.

Good catch!

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -69,7 +71,36 @@ bool inAttributeChangedNotificationGuard()
...
+
+void sendValueChanged(const Ufe::Path& ufePath, const TfToken& changedToken)
+{
+    Ufe::AttributeValueChanged vc(ufePath, changedToken.GetString());
+    Ufe::Attributes::notify(vc);
+
+    if (changedToken == HdCameraTokens->horizontalAperture
```
ppt-adsk, line 81: Style nit-pick: I would have made a static set of tokens and used find on that:
static std::unordered_set<TfToken, TfToken::HashFunctor> cameraTokens{HdCameraTokens->horizontalAperture, HdCameraTokens->verticalAperture, // ...  etc. };
if (cameraTokens.count(changedToken)) {
    Ufe::Camera::notify(ufePath);
}
Not sure which is faster --- unordered_set is amortized O(1), the chain of possibilities is O(n), but that's just complexity, not real-life performance.  Your pick.

ppt-adsk (changes requested):

ppt-adsk (approved):

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -69,7 +70,26 @@ bool inAttributeChangedNotificationGuard()
...
 
-std::unordered_map<Ufe::Path, std::string> pendingAttributeChangedNotifications;
+std::unordered_multimap<Ufe::Path, TfToken> pendingAttributeChangedNotifications;
+
+void sendValueChanged(const Ufe::Path& ufePath, const TfToken& changedToken)
+{
+    Ufe::AttributeValueChanged vc(ufePath, changedToken.GetString());
+    Ufe::Attributes::notify(vc);
+
+    if (MAYAUSD_NS_DEF::ufe::UsdCamera::isCameraToken(changedToken)) {
```
seando-adsk, line 80: The macro MAYAUSD_NS_DEF is only meant to be used when declaring a namespace. Here you should just use "MayaUsd::"

lib/mayaUsd/ufe/StagesSubject.cpp
williamkrick: I made this change and now I'm having a test failure:

```
    ======================================================================
    FAIL: testObservation (testAttribute.AttributeTestCase)
    Test Attributes observation interface.
    ----------------------------------------------------------------------
      File "testAttribute.py", line 512, in testObservation
        self.assertEqual(ball34Obs.notifications, 1)
    AssertionError: 2 != 1
```

Looking at the test code it seems like there should only be one notification for ball34:

```
        self.assertEqual(ball34Obs.notifications, 0)

        ufeCmd.execute(ball34XlateAttr.setCmd(ufe.Vector3d(1, 2, 3)))

        self.assertEqual(ball34Obs.notifications, 1)
```

I added a debug statement to show what the notifications are:

```
15: Sending Attribute Value Changed |world|transform1|proxyShape1/Room_set Room_set
15: Sending Attribute Value Changed |world|transform1|proxyShape1/Room_set/Props Props
15: Sending Attribute Value Changed |world|transform1|proxyShape1/Room_set/Props/Ball_34 Ball_34
15: Sending Attribute Value Changed |world|transform1|proxyShape1/Room_set/Props/Ball_34 xformOp:translate
```

So exactly the pattern that is changed with multimap is what is causing the test failure.

What I don't know is are the first 3 notifications correct? I'm not really sure what they are trying to tell us having their token be the same as the last piece of the path.

seando-adsk: With your change we will now be getting extra (in this case 2) attribute value changes notifs (with different tokens) - which is now the correct behavior. So I think you just need to update the test to account for this. Note: you'll have to update the values on all subsequent lines since it does undo/redo and verifies the count after each.

williamkrick: Pierre and I were not sure about notifying attribute changed for tokens which are not attributes (the first 3 notifications). I logged MAYA-110878 to investigate those extra notifications, and I'm going to switch back to unordered_set for this pull request.

ppt-adsk (approved):

PR #1321, opened 2021-04-08
MAYA-110850 fix a point snapping issue on USD instances

HdC-adsk:
This change will fix the bug when trying to point snap to a USD instance with camera-based selection off.

There is still another issue affecting point snapping with camera-based selection on. We will fix it separately.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -794,19 +794,16 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
     MDagPath&                       dagPath) const
 {
 #if defined(WANT_UFE_BUILD)
-    if (_proxyShapeData->ProxyShape() == nullptr) {
-        return false;
+    // When point snapping, only the point position matters, so return the DAG path and avoid the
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -794,20 +794,17 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
...
+        return true;
     }
 
-    if (!_proxyShapeData->ProxyShape()->isUfeSelectionEnabled()) {
+    if (!_proxyShapeData->ProxyShape() || !_proxyShapeData->ProxyShape()->isUfeSelectionEnabled()) {
         return false;
     }
 
-    // When point snapping, only the point position matters, so return false
-    // to use the DAG path from the default implementation and avoid the UFE
```
HdC-adsk, line 806: The default implementation in VP2 assumes that the instance id of the hit item can be mapped to the instance index of the instanced DAG path. But in maya-usd we don't support instanced proxy shape at least for now, so we just override the default behavior and we don't really care about the mapping at all.

JGamache-autodesk (approved):

williamkrick (approved):

HdC-adsk:
@kxl-adsk Please merge. Thanks.

PR #1326, opened 2021-04-12
Adding a Maya:index custom attribute to uv sets. This allows easier re-use of materials as the name of the uv set can be derived instead of hard coded in the primvar reader

koen-v:
Adding a Maya:index custom attribute to uv sets. These allow us to link back the order of the uv sets in the usd to the order of the uv sets in maya.

### DIFF
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
@@ -957,7 +957,7 @@ bool UsdMayaMeshWriteUtils::writeUVSetsAsVec2fPrimvars(
 
         // create UV PrimVar
         createUVPrimVar(
-            primSchema, i, setName, usdTime, uvValues, interpolation, assignmentIndices, valueWriter);
+            primSchema, setName, i, usdTime, uvValues, interpolation, assignmentIndices, valueWriter);
     }
 
     return true;
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
@@ -95,6 +95,10 @@ const GfVec4f UnauthoredColorSetRGBA = GfVec4f(
     UnauthoredColorSetRGB[2],
     UnauthoredColorAlpha);
 
+/// The key in the custom data dict we use to store the index of the uv
+/// attributes in maya
+const TfToken UvIndexCustomKey("Maya:index");
+
 // XXX: Note that this function is not exposed publicly since the USD schema
 // has been updated to conform to OpenSubdiv 3. We still look for this attribute
 // on Maya nodes specifying this value from OpenSubdiv 2, but we translate the
@@ -246,6 +250,7 @@ void setPrimvar(
 void createUVPrimVar(
     UsdGeomGprim&              primSchema,
     const TfToken&             name,
+    const int&                 index,
     const UsdTimeCode&         usdTime,
     const VtArray<GfVec2f>&    data,
     const TfToken&             interpolation,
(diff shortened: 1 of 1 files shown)


kxl-adsk:
@koen-v Can you elaborate a bit on how this custom attribute would make material re-use simpler? Maybe you could provide USD files to illustrate it?

koen-v:
@kxl-adsk Of course, I'll attach the file I added in the discussion section here as well. The grid and the box are using the same material, but have different uv names. So if you lookin in **unitTest.usd**, you will see def Material "**blinn2SG**" for this material. This gets assigned to the grid.

Now when assigning the same material to the box, it does not work, as in theblinn2SG material there is: 
token inputs:file1:varname = "blah4"

which is the name of the uv set we want to use on the grid. The box does not have this, so we need to specialize the material:
```
def Material "blinn2SG_map1

" (
            specializes = </Blah/materials/blinn2SG>
        )
  {
      token inputs:file1:varname = "map1"
  }

```

and this one we can assign to the box.  

This example with the specialization is an export from maya. Now imagine both the box and the grid would use "st" for the uv name; I would be able to directly assign the material to both the grid and the box by setting the varname to "st". ( currently we can do this with an environment variable, but imagine a material with 2 uv sets, that is currently not possible).

Next imagine an artist clicking random materials from a library to see how they look. If we can rely on the uv names, it is a simple override on the materials binding, if it's not, I have to every time create a temp material, specialize the uv set, assign that material and when the user clicks the next, delete the old one and do this again. Even when assigning materials to objects in scenes, its nice to just bind the material without having to inspect the uv sets and guess which one you need to hook up.

In our pipeline, we actually assign materials later on, so if I have an index, I have an opportunity to rename the uv sets in the model before assigning materials. I have a bigger change in the current plugin we use, where I rename the uv sets to st, st1, st2, st3 etc and I add meta data for the original maya name, and I restore this on import. This change will change the behavior of the plugin.

I thought storing the index would be a nice compromise between giving pipelines the flexibility they might need without changing current behavior.

Thanks,
Koen

[unitTest.zip](https://github.com/Autodesk/maya-usd/files/6312258/unitTest.zip)

koen-v:
Hello @kxl-adsk,

Did you get a chance to look at this? Any concerns?

Thanks,
Koen

kxl-adsk:
I didn't get as much time to look at this as I would like. Regardless, the part that I'm trying to understand is your index based workflow. USD is binding UVs by name and all the magic has to be happening in your tooling around USD.

>Next imagine an artist clicking random materials from a library to see how they look. If we can rely on the uv names, it is a simple override on the materials binding
I'm wondering what's the actual implementation of this that leverages Maya:index - I'm assuming it has to involve renaming of primvars? How will it work with assets produced by other DCCs (they won't have Maya:index attribute).

koen-v:
Indeed, I rename the uv's I would love a flag that says "rename all maps to st, st1, etc"  That's what I am doing in my version of the plugin. I am trying to move to the default plugin and use a python script after export to fix up the file. This is what I do with the uv's: ( so yes, all the uv sets are renamed, and I need the indices to do it. ) 

>In our pipeline, we actually assign materials later on, so if I have an index, I have an opportunity to rename the uv sets in the >model before assigning materials. I have a bigger change in the current plugin we use, where I rename the uv sets to st, st1, >st2, st3 etc and I add meta data for the original maya name, and I restore this on import. This change will change the behavior >of the plugin.

I feel like I explained the same thing 5 times now. If you ever want to do a call, please let me know. I gave up yesterday on using the default plugin anyway (because of this and the material names). If you dont see value  in this, please feel free to close the pull request.

koen

koen-v:
I just came across this material in a usd exported from maya, it illustrates some if the issues with the uv maps :)
 
```
def Material "shaderball_map1_map1_map1_map1_map1_map1_map1_map1_map1_map1" (
            specializes = </rigged/materials/shaderball>
        )
        {
            token inputs:env_map:varname = "map1"
            token inputs:f_base_c:varname = "map1"
            token inputs:f_base_g:varname = "map1"
            token inputs:f_base_n:varname = "map1"
            token inputs:f_envBrdfLut_GGX:varname = "map1"
        }
```

kxl-adsk:
@koen-v We finally got to implement the solution we discussed some time ago. Please have a look at https://github.com/Autodesk/maya-usd/pull/1528.

PR #1329, opened 2021-04-13
MAYA-110917 Increase depth priority for point snapping items

HdC-adsk:
After turning on camera-based selection and zooming out far enough
to a USD object, we could only snap to silhouette points. After
zooming in closer, the issue will become better. This is a preexisting
depth priority issue in VP2RenderDelegate so it has nothing to do with
the recent point snapping refactoring inside Maya (MAYA-109518).


JGamache-autodesk (approved):

HdC-adsk:
@kxl-adsk Please merge. Thanks.

HdC-adsk:
The pf failed because osx platforms failed to launch the remote build.

@kxl-adsk I think it is safe to merge.

HdC-adsk:
@seando-adsk Is the above pf issue known to you?

PR #1331, opened 2021-04-13
MAYA-110361 - add PF jobs to cover what's currently tested manually

seando-adsk:
#### MAYA-110361 - add PF jobs to cover what's currently tested manually
* Fixing linux compilation error when building with USD v20.02.


kxl-adsk (approved):
LGTM

PR #1339, opened 2021-04-14
MAYA-108563 - Support default material shading using new SDK APIs

JGamache-autodesk:
Draws correctly, but selection fails to update in default material repr.


JGamache-autodesk:
Incomplete. One item remaining: selection does not update when dealing with the default material repr.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1597,7 +1642,8 @@ void HdVP2Mesh::_UpdateDrawItem(
     }
 #endif
 
-    if (desc.geomStyle == HdMeshGeomStyleHull) {
+    if (desc.geomStyle == HdMeshGeomStyleHull
+        && desc.shadingTerminal == HdMeshReprDescTokens->surfaceShader) {
```
JGamache-autodesk, line 1646: defaultMaterial repr has a different shading terminal and will ship the shading update.
Regular smoothMesh has "surfaceShader" as terminal, so this addition does not require version check.

JGamache-autodesk:
Done and working, but can not be run thru preflight until some Maya updates are integrated.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1280,15 +1280,61 @@ void HdVP2Mesh::_InitRepr(const TfToken& reprToken, HdDirtyBits* dirtyBits)
...
+            break;
         case HdMeshGeomStyleHullEdgeOnly:
-            // The smoothHull repr uses the wireframe item for selection
-            // highlight only.
+            // The smoothHull repr uses the wireframe item for selection highlight only.
+#ifdef HAS_DEFAULT_MATERIAL_SUPPORT_API
+            if (reprToken == HdReprTokens->smoothHull
+                || reprToken == HdVP2ReprTokens->defaultMaterial) {
+                // Share selection highlight render item between smoothHull and defaultMaterial:
+                bool                        foundShared = false;
```
williamkrick, line 1301: If we end up using this pattern to share more render items then this seems like a good candidate to be refactored into a method.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
williamkrick: If we modify the real material while in default material mode does it update correctly when we turn off "use default material"? IIRC somewhere we clear all the dirty flags, so we'd forget about the dirty material id and then not update.

williamkrick (changes requested):
I put request changes until we clarify the question about material updates while in "use default material" mode.

kxl-adsk (approved):

williamkrick (approved):

PR #1342, opened 2021-04-16
Configure resolver to generate resolver context

kxl-adsk:
Address issue reported by https://github.com/Autodesk/maya-usd/discussions/1236. FWIW. We are still planning on exploring Ar 2.0 so things may change here.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -657,6 +657,8 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
             }
 
             {
+                PXR_NS::ArGetResolver().ConfigureResolverForAsset(fileString);
+
                 // When opening or creating stages we must have an active UsdStageCache.
                 // The stage cache is the only one who holds a strong reference to the
                 // UsdStage. See https://github.com/Autodesk/maya-usd/issues/528 for
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -684,10 +684,15 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
                         if (!sessionLayer)
                             sessionLayer = SdfLayer::CreateAnonymous();
                         usdStage = UsdStage::Open(
-                            rootLayer, sessionLayer, ArGetResolver().GetCurrentContext(), loadSet);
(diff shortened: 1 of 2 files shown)


kxl-adsk:
@dj-mcg @fabal @pmolodo Please review and raise concerns.

dolivares-spinvfx:
Might this be another relevant location?

https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/utils/utilFileSystem.cpp#L57

dj-mcg (approved):
Looks good, thanks!

fabal (approved):
That's a good thing to have :+1: 

Did you consider using `ResolverContext` objects too?
Typically in places like [this](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/nodes/proxyShapeBase.cpp#L688) where a call like `ArGetResolver().CreateDefaultContextForAsset(fileString)` has become the default standard in many places (usdview, solaris).

A step further would be to be able to apply arbitrary resolver contexts to proxyShapes.

kxl-adsk:
@fabal Thanks for bringing this up. I'm catching up on docs around this, it's unclear to me why we need this extra `ArGetResolver().CreateDefaultContextForAsset(fileString)` . A call to ConfigureResolverForAsset will update the [default context](https://github.com/PixarAnimationStudios/USD/blob/dc710925675f7f58f7a37f6c18d046a687b09271/pxr/usd/ar/defaultResolver_v1.cpp#L92). Why do we have to create it again explicitly instead of querying what was already configured?

Regardless, I updated the change to include suggestions raised in this review and matched the code to what is [here](https://github.com/PixarAnimationStudios/USD/blob/090ef0d849ced875b3b99a15e454b69148ccd8e1/pxr/usdImaging/usdviewq/__init__.py#L287).

@dolivares-spinvfx please give it a try since you raised this issue originally.

fabal:
Admittedly, that explicit call to `CreateDefaultContextForAsset` is not needed in that particular case as `Stage::Open` does this by default but I guess it could be a good start towards explicit resolver contexts.
Over time, we steered clear of the default resolver context as it can lead to asset resolution "leaks" in complex scenarios (e.g. several proxyShapes, using dedicated resolver context). By "leaks", I don't mean a USD bug, I mean resolution overrides of one asset leaking into another one.
In relation to that, we also ended up guarding most of our calls to `Sdf::Layer::FindOrOpen` with a resolver context binder.
These comments are mainly for complex asset resolution use cases though, however I believe it is worth mentioning.

dolivares-spinvfx:
Was just able to test this today due to a disk failure.  Seems like asset resolution is working with this change.  Thanks!

Issue #506, opened 2020-05-19
Material names on exported mesh based on shader group instead of material


stevkalinowski:
**Describe the bug**
When I export a mesh to .usd and then import the file back in to Maya, the material names are incorrect.  The shader group names have been used instead and on top of that there is a suffix _lambert one added to the end.

**Steps to reproduce**
* create a mesh sphere
* assign a new blinn material, name it "test_material"
* File > Export All
* File > New
* File > Import, choose the exported usd file
* Notice material name is now "blinn1SG_lambert"

**Expected behavior**
I would expect that the material be named the same as when it was exported.

**Specs (if applicable):**
Windows 10
Maya PR114

santosg87:
Thank you for logging this @stevkalinowski 

I can reproduce this on our side as well.

JGamache-autodesk:
Not only that, but the material is now a Lambert instead of a Blinn, and all values and connections are lost.

I am looking into improving that, and insights about how much information you expect to transfer losslessly would help. I want, at minimum, to get back a Blinn and having file connections restored.

stevkalinowski:
For us maintaining as many file node connections as possible would be expected.  The attribute values would be nice too.

Maybe a stretch, but maybe also the color space setting of the file nodes as well?  sRGB or Raw for example.

stevkalinowski:
For clarification, "attribute values" I refer to values alongside the texture inputs, such as eccentricity, diffuse, reflectivity, etc.  Anything in "Common" and "Specular" sections of the attribute editor.  We use lambert, blinn, and phong.

stevkalinowski:
If it helps, I've uploaded a sample scene "character_export.zip" to Activision's sharepoint drive.

koen-v:
Hello all, was this addressed? If so is there perhaps a CL this was addressed in? In my internal build, I am still getting this, and I was curious where to best start to address it.

Thanks,
Koen

JGamache-autodesk:
I would probably start in `UsdMayaShadingModeExportContext::MakeStandardMaterialPrim()` by replacing:
```
        MFnDependencyNode seDepNode(_shadingEngine, &status);
        if (!status) {
            return ret;
        }
        materialName = seDepNode.name().asChar();
```
with
```
        MFnDependencyNode surfDepNode(GetSurfaceShader(), &status);
        if (!status) {
            return ret;
        }
        materialName = surfDepNode.name().asChar();
```
I suspect that is all that would be needed to get the proper material name.

[...]

PR #1344, opened 2021-04-16
Remove namespace on material import

JGamache-autodesk:
Fixes material roundtripping issues and #506 (which has roundtripping implications)


dj-mcg (approved):
Admittedly, my experience is thin here, but the templatedRegex check looks like it may be a little fragile if the name generation code changes from under this code - not sure if it's worth splitting that logic out to a place where both the generation code and the callsites that check for it can use. But if this is all pretty stable under the hood, this change fixes up round tripping and seems good to me!

kxl-adsk:
@koen-v,  @stevkalinowski  Since this PR addressed as well issue https://github.com/Autodesk/maya-usd/issues/506. Would you mind giving it a try before we merge?

koen-v:
Thanks I'll try to build this, but I'll probably need a few days to get to this, as I am in the middle of stabilizing an internal build. Please feel free to merge, but I will test as soon as I can.

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -387,13 +394,28 @@ UsdPrim UsdMayaShadingModeExportContext::MakeStandardMaterialPrim(
...
+        if (fnDepNode.setObject(_shadingEngine) == MS::kSuccess) {
+            sgName = fnDepNode.name().asChar();
+        } else {
             return ret;
         }
-        MString seName = seDepNode.name();
-        materialName = seName.asChar();
+
+        std::smatch match;
+        if (std::regex_match(sgName, match, _templatedRegex)) {
```
kxl-adsk, line 407: Would it work if we compare the SG name with the surface shader name? If they only differ by `SG` suffix, we have an original name.

test/lib/usd/translators/testUsdExportImportRoundtripPreviewSurface.py
```diff
@@ -116,37 +120,37 @@ def testUsdPreviewSurfaceRoundtrip(self):
         # Check the new sphere is in the new shading group:
         self.assertTrue(cmds.sets(
             "pSphere1Shape",
-            isMember="USD_Materials:usdPreviewSurface1SG"))
+            isMember=material_sg))
 
         # Check that we have no spurious "Looks" transform
-        expectedTr = set(['front', 'persp', 'side', 'top', 'pSphere1', 'pSphere2'])
+        expectedTr = set(['front', 'persp', 'side', 'top', 'pSphere1'])
```
kxl-adsk, line 126: What happened with pSphere2?

kxl-adsk (approved):

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
JGamache-autodesk: That is an optimization that could be done: If both names differ only by the presence of a terminal "SG" the SG name should be taken as is since it is already in sync with the surface. If the DFA generated by std::regex is slow that could save some time.

test/lib/usd/translators/testUsdExportImportRoundtripPreviewSurface.py
JGamache-autodesk: I added a file->new to make sure no renaming would happen on import. This means that after the roundtrip there is only one sphere in the final scene and it needs to have the same name it had at export time.

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
kxl-adsk: It's not really optimization why I mentioned it. Currently, you hardcoded the node types in template regex. For any shader type that isn't captured in there, we will fail to use the surface shader name. So it's more robustness than optimization that I'm talking about.

kxl-adsk (approved):

kxl-adsk:
Removing ready-for-merge since PF is failing.

PR #1352, opened 2021-04-19
Point instance interactive batched position, orientation and scale ch…

ppt-adsk:
…anges.

Point instance manipulation performance is greatly enhanced by batching reads and writes for the position, orientation and scale attributes, which are on the point instancer prim.

### DIFF
--- a/lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
+++ b/lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
@@ -49,8 +49,12 @@ template <class UsdValueType> struct MAYAUSD_CORE_PUBLIC UsdPointInstanceBatch
     inline bool isWriter() const { return ((count + 1) % nbInstances) == 0; }
 
     PXR_NS::VtArray<UsdValueType> usdValues;
-    int                           nbInstances { 0 };
-    int                           count { 0 };
+    // Number of instances in the batch.  Incremented by
+    // UsdPointInstanceModifierBase::joinBatch().
+    unsigned int nbInstances { 0 };
+    // Running count of execution of the batch.  Incremented by
+    // UsdPointInstanceModifierBase::setValue().
+    unsigned int count { 0 };
 };
 
 /// Abstract utility class for accessing and modifying attributes of USD point
@@ -74,12 +78,11 @@ template <class UfeValueType, class UsdValueType>
 class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
 {
 public:
-    typedef std::shared_ptr<UsdPointInstanceBatch<UsdValueType>> Batch;
-    typedef std::unordered_map<Ufe::Path, Batch>                 Batches;
+    using Batch = std::shared_ptr<UsdPointInstanceBatch<UsdValueType>>;
+    using Batches = std::unordered_map<Ufe::Path, Batch>;
 
     UsdPointInstanceModifierBase()
         : _prim()
-        , _instanceIndex(-1)
     {
     }
 
@@ -110,8 +113,6 @@ public:
     const Ufe::Path& path() const { return _path; };
 
     Ufe::Path pointInstancerPath() const { return _path.pop(); };
-
-    virtual Batches& batches() = 0;
 
     PXR_NS::UsdGeomPointInstancer getPointInstancer() const
     {
@@ -218,6 +219,12 @@ public:
         return writer ? usdAttr.Set(_batch->usdValues, usdTime) : true;
     }
 
+    // Join a point instancer batch.  Because objects of
+    // UsdPointInstanceModifierBase derived types are used to read from point
+    // instancers (e.g. in UsdTransform3dPointInstance), joining a batch cannot
+    // be done automatically when calling setSceneItem().  It must be done
+    // deliberately by calling joinBatch() on a modifier used to actually write
+    // to the point instancer.
     void joinBatch()
     {
         // If we've already joined a point instance batch, nothing to do.
@@ -246,6 +253,16 @@ public:
     virtual UsdValueType getDefaultUsdValue() const = 0;
 
 protected:
+    // Retrieve the active batches (one per point instancer path) for the
+    // derived attribute type (i.e. position, orientation, or scale).  We keep
+    // one active batch map per attribute type (3 batch maps), rather than a
+    // single global active batch map, as the modifier interface allows
+    // creating a single command that would write in batches to a single point
+    // instancer for more than one attribute type (e.g. a command that would
+    // write both position and orientation).  At time of writing (26-Apr-2021)
+    // no such command exists.
+    virtual Batches& batches() = 0;
+
     virtual PXR_NS::UsdAttribute _getAttribute() const = 0;
 
     virtual PXR_NS::UsdAttribute _createAttribute() = 0;
@@ -359,10 +376,10 @@ protected:
     // An alternative to storing these three data members would be to simply
     // store a UsdSceneItem.  PPT, 12-Apr-2021.
     PXR_NS::UsdPrim _prim;
-    int             _instanceIndex;
-    Ufe::Path       _path {};
+    int             _instanceIndex { -1 };
+    Ufe::Path       _path;
 
-    Batch _batch { nullptr };
+    Batch _batch;
 };
 
 } // namespace ufe
--- a/lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
+++ b/lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
@@ -230,7 +230,7 @@ public:
         auto       found = b.find(instancerPath);
         // If we're the first to join the batch, create it.
         if (found == b.end()) {
-            auto inserted = b.insert(Batches::value_type(
+            auto inserted = b.insert(typename Batches::value_type(
                 instancerPath, std::make_shared<UsdPointInstanceBatch<UsdValueType>>()));
             found = inserted.first;
         }
--- a/lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
+++ b/lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
@@ -17,6 +17,7 @@
 #define MAYAUSD_UFE_USD_POINT_INSTANCE_MODIFIER_BASE_H
 
 #include <mayaUsd/base/api.h>
+#include <mayaUsd/ufe/UsdSceneItem.h>
 
 #include <pxr/base/tf/diagnostic.h>
 #include <pxr/base/vt/array.h>
@@ -29,6 +30,28 @@
 
 namespace MAYAUSD_NS_DEF {
 namespace ufe {
+
+//! \brief Shared data for USD point instance read and write batching.
+//
+// This data structure can be used by UsdPointInstanceModifierBase when
+// modifying the position, orientation, or scale of multiple point instances
+// from a single point instancer.  These changes can be batched such that the
+// first modifier to make a change will read from the point instancer attribute
+// into the usdValues array, and the last modifier to make a change will write
+// the usdValues array back to the point instancer attribute.  All modifiers in
+// the batch will write to and read from the usdValues array, a much less
+// expensive operation than writing to or reading from the point instancer
+// attribute.
+//
+template <class UsdValueType> struct MAYAUSD_CORE_PUBLIC UsdPointInstanceBatch
+{
+    inline bool isReader() const { return (count % nbInstances) == 0; }
+    inline bool isWriter() const { return ((count + 1) % nbInstances) == 0; }
+
+    PXR_NS::VtArray<UsdValueType> usdValues;
+    int                           nbInstances { 0 };
+    int                           count { 0 };
+};
 
 /// Abstract utility class for accessing and modifying attributes of USD point
 /// instances.
@@ -51,23 +74,25 @@ template <class UfeValueType, class UsdValueType>
 class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
 {
 public:
+    typedef std::shared_ptr<UsdPointInstanceBatch<UsdValueType>> Batch;
+    typedef std::unordered_map<Ufe::Path, Batch>                 Batches;
+
     UsdPointInstanceModifierBase()
         : _prim()
         , _instanceIndex(-1)
     {
     }
 
-    UsdPointInstanceModifierBase(const PXR_NS::UsdPrim prim, int instanceIndex)
-    {
-        setPrimAndInstanceIndex(prim, instanceIndex);
-    }
-
     virtual ~UsdPointInstanceModifierBase() = default;
 
-    bool setPrimAndInstanceIndex(const PXR_NS::UsdPrim prim, int instanceIndex)
+    bool setSceneItem(const UsdSceneItem::Ptr& sceneItem)
     {
         _prim = PXR_NS::UsdPrim();
         _instanceIndex = -1;
+        _path = Ufe::Path();
+
+        auto prim = sceneItem->prim();
+        auto instanceIndex = sceneItem->instanceIndex();
 
         if (!PXR_NS::UsdGeomPointInstancer(prim) || instanceIndex < 0) {
             return false;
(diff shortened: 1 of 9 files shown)


lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -51,36 +74,45 @@ template <class UfeValueType, class UsdValueType>
...
+    typedef std::shared_ptr<UsdPointInstanceBatch<UsdValueType>> Batch;
+    typedef std::unordered_map<Ufe::Path, Batch>                 Batches;
+
     UsdPointInstanceModifierBase()
         : _prim()
         , _instanceIndex(-1)
     {
     }
 
-    UsdPointInstanceModifierBase(const PXR_NS::UsdPrim prim, int instanceIndex)
```
ppt-adsk, line 60: This constructor was unused, removed.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -51,36 +74,45 @@ template <class UfeValueType, class UsdValueType>
...
+        _path = sceneItem->path();
 
         return true;
     }
 
     PXR_NS::UsdPrim prim() const { return _prim; };
 
+    const Ufe::Path& path() const { return _path; };
+
+    Ufe::Path pointInstancerPath() const { return _path.pop(); };
```
ppt-adsk, line 112: By definition, the point instancer is the parent of its point instances.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -153,23 +185,58 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
 
         const size_t instanceIndex = static_cast<size_t>(_instanceIndex);
 
-        PXR_NS::UsdAttribute usdAttr = _getOrCreateAttribute();
-        if (!usdAttr) {
-            return false;
+        // Once setValue() is called, any point instance batch under
+        // construction is closed.  If no batch was created, make a trivial,
+        // unshared batch of a single point instance.
+        _closeBatch();
```
ppt-adsk, line 191: The first time you set a value means that batch creation is over for that command, for that point instancer.  All point instance modifiers that share the batch will now use it.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -153,23 +185,58 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
...
+            usdAttr = _getOrCreateAttribute();
+            if (!usdAttr) {
+                return false;
+            }
         }
 
-        PXR_NS::VtArray<UsdValueType> usdValues;
-        if (!usdAttr.Get(&usdValues, usdTime)) {
-            return false;
+        if (reader) {
```
ppt-adsk, line 206: Point instance modifier in the batch that is first to set a value will read into the batch's VtArray.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -153,23 +185,58 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
...
-        if (instanceIndex >= usdValues.size()) {
+        if (instanceIndex >= _batch->usdValues.size()) {
             return false;
         }
 
-        usdValues[instanceIndex] = usdValue;
+        _batch->usdValues[instanceIndex] = usdValue;
 
-        return usdAttr.Set(usdValues, usdTime);
+        return writer ? usdAttr.Set(_batch->usdValues, usdTime) : true;
```
ppt-adsk, line 218: Last modifier to write will write back to the point instancer attribute.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -153,23 +185,58 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
...
+
+    void joinBatch()
+    {
+        // If we've already joined a point instance batch, nothing to do.
+        if (_batch) {
+            return;
+        }
+
+        const auto instancerPath = pointInstancerPath();
+        auto&      b = batches();
```
ppt-adsk, line 229: Access the map of current batches under construction, at most one batch per instancer path.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -153,23 +185,58 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
...
+        auto       found = b.find(instancerPath);
+        // If we're the first to join the batch, create it.
+        if (found == b.end()) {
+            auto inserted = b.insert(Batches::value_type(
+                instancerPath, std::make_shared<UsdPointInstanceBatch<UsdValueType>>()));
+            found = inserted.first;
+        }
+        found->second->nbInstances++;
```
ppt-adsk, line 237: Increment the number of point instances in the batch.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -153,23 +185,58 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
...
+
+        _batch = found->second;
```
ppt-adsk, line 239: Join the batch (by shared pointer).

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -275,8 +342,27 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
...
 
+    void _closeBatch()
+    {
+        // To close the batch we simply remove it from the map of batches under
+        // construction, and thus it can no longer be joined.  If a batch was
+        // created, erase() will harmlessly fail for all modifiers except the
+        // first one.  If no batch was created, make a trivial, unshared batch
+        // just for this modifier.
+        batches().erase(pointInstancerPath());
+        if (!_batch) {
```
ppt-adsk, line 353: When closing the batch there might not have been any created.  In that case create a trivial batch with ourselves as the single point instance in the batch.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -275,8 +342,27 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
...
+            _batch = std::make_shared<UsdPointInstanceBatch<UsdValueType>>();
+            _batch->nbInstances = 1;
+        }
+    }
+
+    // An alternative to storing these three data members would be to simply
```
ppt-adsk, line 359: I really had to hold myself back with both hands to not change these three data members to a single UsdSceneItem, because that's exactly what a UsdSceneItem is: a prim, a path, and an instance index.  However, this kind of software engineering minor improvement has nothing to do with performance.

lib/mayaUsd/ufe/UsdPointInstanceOrientationModifier.h
```diff
@@ -45,11 +45,6 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceOrientationModifier
     {
     }
 
-    UsdPointInstanceOrientationModifier(const PXR_NS::UsdPrim prim, int instanceIndex)
```
ppt-adsk, line 48: Unused constructor, removed.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -30,6 +31,28 @@
...
+// expensive operation than writing to or reading from the point instancer
+// attribute.
+//
+template <class UsdValueType> struct MAYAUSD_CORE_PUBLIC UsdPointInstanceBatch
+{
+    inline bool isReader() const { return (count % nbInstances) == 0; }
+    inline bool isWriter() const { return ((count + 1) % nbInstances) == 0; }
+
+    PXR_NS::VtArray<UsdValueType> usdValues;
+    int                           nbInstances { 0 };
```
kxl-adsk, line 52: This can never be negative, consider switching to one of the unsigned types.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -30,6 +31,28 @@
...
+    int                           count { 0 };
```
kxl-adsk, line 53: Same here.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -51,36 +74,45 @@ template <class UfeValueType, class UsdValueType>
 class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
 {
 public:
+    typedef std::shared_ptr<UsdPointInstanceBatch<UsdValueType>> Batch;
```
kxl-adsk, line 77: Based on modern C++ guidelines, we should switch to alias-declaration (`using` keyword) instead of `typedef`.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -153,23 +185,58 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
...
+        PXR_NS::UsdAttribute usdAttr;
+
+        const bool reader = _batch->isReader();
+        const bool writer = _batch->isWriter();
+        _batch->count++;
+
+        if (reader || writer) {
+            usdAttr = _getOrCreateAttribute();
+            if (!usdAttr) {
+                return false;
```
kxl-adsk, line 202: Something is really wrong if we enter here with the writer index. Deserves a coding error?

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
kxl-adsk: It would be useful as well to document here how the number of instances is discovered.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -275,8 +342,27 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
...
+            _batch = std::make_shared<UsdPointInstanceBatch<UsdValueType>>();
+            _batch->nbInstances = 1;
+        }
+    }
+
+    // An alternative to storing these three data members would be to simply
+    // store a UsdSceneItem.  PPT, 12-Apr-2021.
     PXR_NS::UsdPrim _prim;
     int             _instanceIndex;
+    Ufe::Path       _path {};
```
kxl-adsk, line 363: There shouldn't be any need to initialize the path explicitly. BUT it is a good practice for trivial types like int (just above).

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -275,8 +342,27 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceModifierBase
...
+
+    Batch _batch { nullptr };
```
kxl-adsk, line 365: Same comment here. By default, it should be an invalid shared pointer.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
```diff
@@ -51,36 +74,45 @@ template <class UfeValueType, class UsdValueType>
...
         return true;
     }
 
     PXR_NS::UsdPrim prim() const { return _prim; };
 
+    const Ufe::Path& path() const { return _path; };
+
+    Ufe::Path pointInstancerPath() const { return _path.pop(); };
+
+    virtual Batches& batches() = 0;
```
kxl-adsk, line 114: Doesn't feel very kosher to give access to this to the outside world. I will probably learn later why it's needed, but for now, it feels like detail like this should be protected.

lib/mayaUsd/ufe/UsdPointInstanceOrientationModifier.cpp
```diff
@@ -101,5 +106,10 @@ UsdAttribute UsdPointInstanceOrientationModifier::_createAttribute()
     return pointInstancer.CreateOrientationsAttr();
 }
 
+UsdPointInstanceOrientationModifier::Batches& UsdPointInstanceOrientationModifier::batches()
+{
+    return sBatches;
```
kxl-adsk, line 111: Why do we need a batch map per modifier type?

lib/mayaUsd/ufe/UsdPointInstanceUndoableCommands.h
```diff
@@ -59,7 +59,10 @@ class MAYAUSD_CORE_PUBLIC UsdPointInstanceUndoableCommandBase : public BaseUfeUn
             return;
         }
 
-        _modifier.setPrimAndInstanceIndex(item->prim(), item->instanceIndex());
+        _modifier.setSceneItem(item);
+        // We're using a modifier to change a point instancer attribute, so
+        // batch the reads and writes, for efficiency.
+        _modifier.joinBatch();
```
kxl-adsk, line 65: Why we can't join the batch when setting a scene item on a modifier?

kxl-adsk (commented):
Very good job @ppt-adsk. I left some comments and questions in the code. One additional question - what's the granularity of edits during undo/redo? Are we restoring entire attribute value or operating on individual elements and will need batching as well?

JGamache-autodesk (approved):
Nice. Setting the instance array once per move instead of once per instance per move will help a lot.
Since each member of the batch read/writes to a different location in the instance array, can the process be made parallel when the number of selected instances becomes massive?

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
ppt-adsk: Because of problems with mixed integral type arithmetic
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-mix
I usually don't use signed integers (John Lakos has a minor design rule on this, unfortunately don't have his book handy).

However, in this case, because I'm using modular arithmetic, I think your point is well taken.

ppt-adsk:
> Since each member of the batch read/writes to a different location in the instance array, can the process be made parallel when the number of selected instances becomes massive?

I don't really know how we could do parallelization.  As currently called by Maya, UFE commands are sequential.  We would need to parallelize write translation / rotation / scale, and this would need to be a special case just for point instancers, as USD is geared to single writer / multiple reader.  Unless we measure it to be a performance blocker, I won't investigate this.

lib/mayaUsd/ufe/UsdPointInstanceUndoableCommands.h
ppt-adsk: That's the way I had it originally, but it's incorrect, as modifiers are used by UsdTransform3dPointInstance to read:
https://github.com/Autodesk/maya-usd/blob/0e13ab44a2f10050ccd257607596539c830980db/lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp#L49
It's really only when you're using a modifier to write to the attribute that you want to join a batch.  I will add that as a code comment.

ppt-adsk:
> 
> 
> Very good job @ppt-adsk. I left some comments and questions in the code. One additional question - what's the granularity of edits during undo/redo? Are we restoring entire attribute value or operating on individual elements and will need batching as well?

The new undo / redo system should restore the complete attribute, and thus will be batched by virtue of the attribute write having been batched.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
ppt-adsk: Didn't get that one --- what is really wrong?  We need the attribute to read and write.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
ppt-adsk: Excellent point, done.

lib/mayaUsd/ufe/UsdPointInstanceModifierBase.h
kxl-adsk: Does it make sense to set values on indexes when there is no attribute? Should we stop it before we get to the last point instance?

kxl-adsk (approved):

PR #1354, opened 2021-04-20
Clean up runtimeCommands created by the plugin

fowlertADSK:
Plugin was creating runtimeCommands on load but not cleaning them up when unloading.  This change cleans those up on unload and also associates the commands with the plugin, which will mean Maya can load the plugin automatically if someone calls one of the commands when it's not yet loaded.

Few things for QA to check with this change:
1. Shut down Maya with the plugin loaded and set to autoload.  On restart the plugin shouldn't need to re-create the commands since they are stored in the pref's.  Just want to confirm that we don't spit out any errors about commands already existing.
2. Shut down Maya with the plugin loaded but NOT set to autoload.  On restart try calling one of the runtime commands, e.g. mayaUsdOpenUsdLayerEditor should load the plugin for you and open the Layer Editor.
3. Shut down Maya with the plugin NOT loaded and NOT set to autoload.  On restart if you try calling one of the commands then it should print an error that it doesn't know what that command is.

### DIFF
--- a/plugin/adsk/scripts/mayaUsdMenu.mel
+++ b/plugin/adsk/scripts/mayaUsdMenu.mel
@@ -128,7 +128,7 @@ proc int mayaUSD_isSelectionKindValid() {
 // create all the runtime commands we'll use and the user can map to hotkeys
 proc initRuntimeCommands() {
     if (!`runTimeCommand -exists mayaUsdCreateStageWithNewLayer`) {
-        runTimeCommand -plugin "mayaUsdPlugin"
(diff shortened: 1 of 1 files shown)


fowlertADSK:
After discussing this, we decided to move back to a default runtime command, which means the commands will not be saved to pref's, and to not bother removing them for now.  Specifying the plugin that created the commands is good enough.  We can filter for them correctly now and calling one of them will autoload the plugin.

luceric-adsk (approved):

Issue #933, opened 2020-11-19
IBL Support for USD Preview Surface Shader


dgovil:
**Is your feature request related to a problem? Please describe.**
The current USD Preview Surface shader doesn't allow for IBL's to contribute to the look of things with lighting and reflection.
This makes it somewhat cumbersome to accurately assess what a shader would look like when using Maya for lookdev with the Usd Preview Surface Shader

**Describe the solution you'd like**
It would be nice if there was a slot for an IBL texture similar to the way the Stingray shader was setup to gauge lighting and reflection

**Describe alternatives you've considered**
N/A

**Additional context**
Our Maya users like to lookdev in Maya, and having a better IBL representation would help them better match their assets to an environment or handle reflections etc...

JGamache-autodesk:
Hi @dgovil
have you tried using an Arnold Skydome Light from the MtoA plugin to light assets with UsdPreviewSurface shading?

dgovil:
Hi @JGamache-autodesk , sorry I missed your reply earlier.
The Skydome works pretty well for the lighting aspect of the IBL. 

However, it doesn't seem to handle reflection like Stingray PBR did though in the shader. I've updated my (in hindsight poorly worded) issue to reflect that the IBL would benefit both lighting and reflection.

dgovil:
@JGamache-autodesk so I discovered something today, the USDPreviewSurface does actually support reflection of the aiSkyDomeLight **IF** `specularColor>0`. 

![image](https://user-images.githubusercontent.com/4443090/115310332-7e02ba00-a122-11eb-83d7-054de6534971.png)

There seems to be a related bug that this doesn't depend on specular workflow being enabled or not. As long as specular color is set to something you get reflections of the SkyDome Light as an IBL.

dgovil:
I'm finalizing a PR internally to submit to get support for IBL reflection with the metallic workflow as well

PR #1356, opened 2021-04-20
Metallic IBL Reflection

dgovil:
The USD Preview Surface shader setup only reflected IBLs (represented by an Arnold SkyDomeLight with a texture) if the specular color was not black. It also did this when the specular workflow was not enabled.

This PR fixes that (and addresses my needs in https://github.com/Autodesk/maya-usd/issues/933), so that metallic workflows also are capable of reflecting the IBL.

I also took the liberty of removing the autogeneration comments because, as far as we could discern, these are no longer direct results of the auto generation.


dgovil:
@mattyjams and @JGamache-autodesk , the logic I'm applying here is pretty simplistic. I'm not sure if the metallic+roughness logic that is in `usdPreviewSurfaceLighting.xml` should also be represented here or if this is sufficient, since it comes after that particular fragment.

For reference, this is the code from the lighting fragment.
```
if (!useSpecularWorkflow) {
            float R = (1.0 - ior) / (1.0 + ior);
            float3 specColor = mix(float3(1.0, 1.0, 1.0), diffuseColor, metallic);
            F0  = R * R * specColor;
            F90 = specColor;

            // For metallic workflows, pure metals have no diffuse
            d *= 1.0 - metallic;
        }
```

At the very  least, it allows for a basic reflection estimation in the viewport now
image

JGamache-autodesk (changes requested):

lib/mayaUsd/render/vp2ShaderFragments/usdPreviewSurfaceCombiner.xml
```diff
@@ -141,12 +148,16 @@ usdPreviewSurfaceCombiner(
...
 
     if (!mayaAlphaCut) {
-        result.outColor *= saturate(1.0 - transparency); 
+        result.outColor *= saturate(1.0 - transparency);
     }
 
     // Specular
     result.outColor += specularIrradianceIn;
-    result.outColor += specularColor * specularEnv;
+    if (useSpecularWorkflow) {
```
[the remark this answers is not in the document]
dgovil: Thanks! I'll take a look at integrating those (unless someone else beats me to it)

dgovil:
We've tried to adapt the specular formula into the combiner with some modifications as we believe pertains to reflectivity.
@ysiewappl and I think this should be "close" but without a reference renderer to compare against exactly, we're unsure and would like someone more well versed in the UsdPreviewSurface like @mattyjams to verify when they can.
We are trying to make sure however that we do not modify the unlit mode in any undue way with our changes.

Below you can see a comparison of the before and after of the test sphere with our test IBL, with before and afters of this PR. I include the lit and unlit versions to discern how much things differ. Lit is forward/colorful, unlit is behind and greyscale since the sphere itself has no color. The shader setup in Maya is that the sphere is a light grey, with a black/white checkerboard in the specular/metalness parameters. The IBL is a validation room with a checkerboard pattern and bright text.

metallic

Similarly here is a comparison with the gramaphone asset in the Apple USDZ gallery. Again, each section includes the lit and unlit version of each asset. The IBL is a basic HDR room.

metallic_gramaphone

In both images you can see that the unlit versions remain unaffected as we hope.
Prior to this PR, you can see that the metallic workflow has no IBL reflection since the shader fragments assumed only specular would reflect. 
However after this PR, the lit version of the assets are ever so slightly lifted. We're not sure how to avoid that without diverging from our understanding of the USD Preview Surface.

Note: I have not attempted to update the clear coat logic that @JGamache-autodesk mentioned yet. Perhaps that should be left to a secondary PR, once this one is sorted.

lib/mayaUsd/render/vp2ShaderFragments/usdPreviewSurfaceCombiner.xml
```diff
@@ -87,12 +101,31 @@ usdPreviewSurfaceCombiner(
     result.outColor += diffuseColor * irradianceEnv;
```
JGamache-autodesk, line 101: Can you try:
```
if (useSpecularWorkflow) {
    result.outColor += diffuseIrradianceIn;
    result.outColor += diffuseColor * irradianceEnv;
} else {
    // For metallic workflows, pure metals have no diffuse
    result.outColor += diffuseIrradianceIn * (1.0 - metallic);;
    result.outColor += diffuseColor * irradianceEnv * (1.0 - metallic);
}
```
To see if it helps? Might remove the need to modulate by `length(irradianceEnv)` later.
Ref: https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usdImaging/plugin/usdShaders/shaders/previewSurface.glslfx#L316

lib/mayaUsd/render/vp2ShaderFragments/usdPreviewSurfaceCombiner.xml
```diff
@@ -87,12 +101,31 @@ usdPreviewSurfaceCombiner(
...
+        //   If metallic is 0, then Albedo is ignored in the calculation of F0 and F90; F0 is derived from ior via ( (1-ior)/(1+ior) )^2 and F90 is white. In between, we interpolate.
+
+        // So let's calculate F0 as a function of metalness...
+        float3 F0 = mix(float3(Rsq, Rsq, Rsq), specularEnv, metallic);
+        // ...and calculate F90 as a function of metalness....
+        float3 F90 = mix(float3(1.0, 1.0, 1.0), specularEnv, metallic);
+        // ... finally get the incidence angle to find the resulting outcolor
+        //     and multiply by the intensity of the env irradiance to make sure
+        //     we aren't lifting colors more than the lighting in the scene
+        float incidence = acos(dot(Lw, Nw))/90.0;
```
JGamache-autodesk, line 126: Should be `SchlickFresnel(EdotH)`, but that would require functions and values that are given to  usdPreviewSurfaceLighting.xml but not to the combiner... I start to see how difficult this is going to be.

JGamache-autodesk (approved):
Already looks a lot better for metallic workflows. I have a small thing that might help if you want to try it, otherwise the code is a big step in the right direction.

JGamache-autodesk:
There is a nice discussion on metallic looks here if you are interested: https://github.com/PixarAnimationStudios/USD/issues/1174
Using the provided Damaged_Helmet_usda.zip asset in Maya with USD/pxr/imaging/hdx/textures/StinsonBeach.exr as the environment file should allow comparing the Maya shader against usdView.

lib/mayaUsd/render/vp2ShaderFragments/usdPreviewSurfaceCombiner.xml
ysiewappl: > Should be SchlickFresnel(EdotH), but that would require functions and values that are given to usdPreviewSurfaceLighting.xml but not to the combiner... I start to see how difficult getting this perfectly aligned is going to be.

Hi:

There is this language specifically in the UsdPreviewSurface proposal that confuses me when thinking about how to lerp between F0/F90:

> useSpecularWorkflow = 1: (Specular workflow ) 
> specularColor - color3f - (0.0, 0.0, 0.0)
> Specular color to be used. This is the color at 0 incidence. Edge color is assumed white. Transition between the two colors according to Schlick fresnel approximation.
> useSpecularWorkflow = 0:   (Metalness workflow ) 
> metallic - float - 0.0
> Use 1 for metallic surfaces and 0 for non-metallic.
> - If metallic is 1, then both F0 (reflectivity at 0 degree incidence) and edge F90 reflectivity will simply be the Albedo.
> - If metallic is 0, then Albedo is ignored in the calculation of F0 and F90; F0 is derived from ior via ( (1-ior)/(1+ior) )^2 and F90 is white. *In between, we interpolate*.

It seems to imply that Schlick's fresnel should be used only when specular workflow is observed, and when in a metalness one, the word "interpolate" seemed ambiguous (I am not familiar with the math described in the metalness workflow), so we elected to try and follow the language (a.k.a. lerp between F0/F90 based on angle of incidence) in that case. Is this an incorrect assumption and we should still be using Schlick's fresnel for the interpolant term?

I found this quote from https://learnopengl.com/PBR/Theory which supports the statement above (i.e. use Schlick's fresnel for metalness too instead of using direct angle of incidence), but not sure if it applies.

> There are a few subtleties involved with the Fresnel equation. One is that the Fresnel-Schlick approximation is only really defined for dielectric or non-metal surfaces. For conductor surfaces (metals), calculating the base reflectivity with indices of refraction doesn't properly hold and we need to use a different Fresnel equation for conductors altogether. As this is inconvenient, we further approximate by pre-computing the surface's response at normal incidence (F0) at a 0 degree angle as if looking directly onto a surface. We interpolate this value based on the view angle, as per the Fresnel-Schlick approximation, such that we can use the same equation for both metals and non-metals.

dgovil: I'm trying to use the SchlickFresnel...but I'm stumped by `EdotH`. I don't see it declared as a property in `usdPreviewSurfaceLighting.xml` and I don't see it calculated anywhere.

I do however see a `VdotH` that is never used. My guess is that `EdotH` is a typo and it should be `VdotH` and the properties just get passed in positional order rather than any sort of name ? I just wanted to confirm though before making that assumption.

JGamache-autodesk: Not a typo. If you look at the declaration section of the fragment you will see that the VdotH fragment input provided by Maya becomes the named parameter EdotH in the function call. Name mismatch are allowed as long as the parameter order and type matches.
V is for "view" and E is for "eye".

dgovil: Thanks for clarifying. That makes sense. I'll add a comment in the file for future clarity.

dgovil: Switched to an inline Schlick fresnel copy and it looks good to my eyes at least.

lib/mayaUsd/render/vp2ShaderFragments/usdPreviewSurfaceCombiner.xml
dgovil: I think this definitely helps. For comparison, the before and after. The metallic portions look a lot better on the assets I've tested with.

image
image

lib/mayaUsd/render/vp2ShaderFragments/usdPreviewSurfaceCombiner.xml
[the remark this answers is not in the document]
dgovil: I've copied the shading logic (with adaptations from the Pixar version) for specular.
I would like to leave clear coat for a future time just to keep things separate.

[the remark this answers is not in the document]
JGamache-autodesk: Good idea.

dgovil:
I think your suggestions really helped.

Here's a new before and after. 
The specular workflow remains unchanged.
Metallic now modulates its specular color by F0/F90 as a function of diffuse color, which means that specular highlights appropriately pick up the nature of the base color for metallic elements.

Let me know if you think there's any other changes I should try here, but otherwise I feel like this is in a good state for me personally.

metallic_gramaphone
metallic

JGamache-autodesk:
Looks nice indeed. Thanks.
Does not require PF since all changes are in non-compiler VP2 fragment XML files.

kxl-adsk:
Looking great indeed. Great job!
How about we get an image base test to make sure results of this hard work are not lost? Here is one [example](https://github.com/Autodesk/maya-usd/blob/dev/test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateGeomSubset.py).

dgovil:
@kxl-adsk I'll look into that however I had some questions:

1. The PF doesn't support Arnold which is required for the IBL map. I could spoof it with just colored lights though? Or I could write it to use the Arnold SkyDome and it could be disabled for the PF here.
2. If we do use the skydome light for the test, is there a specific IBL/HDR etc that your project prefers to use?  The gramophone asset is MIT, or the sphere  would be a new contribution so either would be fine to add to the test data set. However I do not have a license compatible IBL to add here as part of the test.

kxl-adsk:
We are adding support for interactive tests in PF, but we won't have the ability to add them with additional dependencies (like MtoA). So Arnold SkyDome is not something we can use in tests.

@JGamache-autodesk do you have any recommendations on what would make sense to do in a test without Arnold dependency?

JGamache-autodesk:
A double metallic/roughness ramp with two directional lights would probably be enough to make sure this does not regress:
```
from maya import cmds

for i in range(5):
	for j in range(5):
		sphere_xform = cmds.polySphere()[0]
		cmds.move( 2.5*(i-2), 2.5*(j-2), 0, sphere_xform, absolute=True )

		material_node = cmds.shadingNode("usdPreviewSurface", asShader=True)
		material_sg = cmds.sets(renderable=True, noSurfaceShader=True, empty=True, name=material_node+"SG")
		cmds.connectAttr(material_node+".outColor", material_sg+".surfaceShader", force=True)
		cmds.sets(sphere_xform, e=True, forceElement=material_sg)

		cmds.setAttr(material_node + ".diffuseColor", 0.944, 0.776, 0.373, type="double3")
		cmds.setAttr(material_node + ".useSpecularWorkflow", False)
		cmds.setAttr(material_node + ".metallic", i * 0.25)
		roughness = j * 0.25
		if j == 0:
			roughness = 0.001
		cmds.setAttr(material_node + ".roughness", roughness)
```
With a sky blue directional light rotated -45 in X and a yellow directional light rotated at +45 in X you get this incredibly ugly looking scene that yet makes sure all the metallic features are there. Of course it looks better with an IBL:
image
image

dgovil:
Thanks! I'll get that set up as part of this PR.

dgovil:
@kxl-adsk  I've added a test to this PR. I also made some modifications to the utility libraries to handle changing cameras and differing color management environments (it was erroring in our environment where we had recently enabled color management workflows)

JGamache-autodesk (approved):

PR #1359, opened 2021-04-21
MAYA-108265 Disable the material consolidation update workaround if Maya has the 

williamkrick:
fix in it. Also add a test to make sure modifying materials correctly updates the viewport when using vp2RenderDelegate.


JGamache-autodesk (approved):

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1622,7 +1623,7 @@ void HdVP2Mesh::_UpdateDrawItem(
 
             // Use fallback shader if there is no material binding or we failed to create a shader
             // instance for the material.
-            if (!drawItemData._shader && _PrimvarIsRequired(HdTokens->displayColor)) {
+            if ((!drawItemData._shader || _meshSharedData->_fallbackColorDirty)&& _PrimvarIsRequired(HdTokens->displayColor)) {
```
JGamache-autodesk, line 1626: I suspect the lack of space before the && will trigger the clang-format check.

JGamache-autodesk (approved):

lib/mayaUsd/render/vp2RenderDelegate/mesh.h
```diff
@@ -84,6 +84,9 @@ struct HdVP2MeshSharedData
 #ifdef HDVP2_ENABLE_GPU_COMPUTE
     MSharedPtr<MeshViewportCompute> _viewportCompute;
 #endif
+
+    //! Fallback color changed
+    bool _fallbackColorDirty { false };
```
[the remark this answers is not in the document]
kxl-adsk: One way would be to use dirty bits - add a custom one here https://github.com/Autodesk/maya-usd/blob/600c31abcd68ef3fe0047b28fb5f33f6edd0872f/lib/mayaUsd/render/vp2RenderDelegate/mesh.h#L171

Then you don't have to deal with clearing it that much since they are all cleared at the end of the repr update.

test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateConsolidation.py
```diff
@@ -0,0 +1,97 @@
...
+from maya import cmds
+
+from pxr import Vt
+
+import ufe
+
+import os
+
+
+class testVP2RenderDelegateConsolidation(imageUtils.ImageDiffingTestCase):
```
kxl-adsk, line 36: Just double-checking, but this test is valid for any Maya version with UFE >= 2 ?

williamkrick: Yes I believe so. There is only a relatively minor dependency on UFE to call getStage. I use USD to change the colors of the materials.

kxl-adsk (changes requested):
testVP2RenderDelegatePerInstanceInheritedData is failing with this change on Maya 2022 and USD 21.02 (Windows python2)

williamkrick:
I found the issue which was causing the problem on 2022 Gold with testVP2RenderDelegatePerInstanceInheritedData. This new version should work correctly now. I will mark "Ready for Merge" if the pre-flight finishes successfully.

kxl-adsk (approved):

PR #1360, opened 2021-04-21
Add support for the parentScope to be an xform

koen-v:
Add support for the parentScope to be an xform so it can be converted into a SkelRoot. This solves the problem of maya scenes where the model and the joints dont share a parent. The meshes used the parentScopePath already, so just converting the parentScope prim to an xform would put the meses in the right plave but the dag path to sdfpath for skeletons did not consider this, so we have to pass the path around a bit.

This is just as an idea. for an improvement,  it works in my case, but might need more design for a generic solution.


barbalt:
Closing this as it has been addressed by another PR #3729

PR #1367, opened 2021-04-27
MAYA-108221 - As a user, on the prim AE template I'd like to see usef…

spinell-adsk:
…ul widgets for arrays

Add Custom widget for array


ppt-adsk (approved):

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -184,6 +184,33 @@ def _onInstanceableChanged(self, value):
         with mayaUsdLib.UsdUndoBlock():
             self.prim.SetInstanceable(value)
 
+# Custom control for all array attribute.
+class ArrayCustomControl(aecustom.CustomControl):
```
seando-adsk, line 188: I think maybe we shouldn't derive from this CustomControl anymore. It is meant for Maya custom controls as the base class tries to get a plug name (which UFE items don't have). I have a local change to remove this from the other custom control in this file.

Just make your class derive from 'object', remove the args/kwargs from the __init__ and change buildControlUI to (you also need a OnReplace).

```
def onCreate(self, *args):
...
def onReplace(self, *args):
    pass
```

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -334,6 +367,13 @@ def createMetadataSection(self):
             self.defineCustom(metaDataControl)
             self.defineCustom(usdNoticeControl)
 
+    def createCustomExtraAttrs(self):
+        # We are not using the maya default "Extra Attributes" section
+        # because we are using custom widget for array type and it's not
+        # possible to inject our widget inside the maya "Extra Attributes" section.
+        extraAttrs = [attr for attr in self.attrS.attributeNames if attr not in self.addedAttrs]
```
seando-adsk, line 374: Do we also need to ignore any attributes in self.suppressedAttrs? It is possible that someone can call suppress() on any attribute and we don't want it ending up in extra attributes.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -334,6 +367,13 @@ def createMetadataSection(self):
...
+        self.createSection("Extra Attributes", extraAttrs, True)
```
seando-adsk, line 375: Its a shame because this section label won't be translated anymore. Is there anyway we can get the translated string from Maya?

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
spinell-adsk: Suppressing attribute will not be an issue as long as we are using the suppress function from the template. The suppress function will add the attribute to suppress into self.suppressedAttrs array which is used by addControls(). So extraAttrs will contain suppressed attributes but they will be ignored.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
spinell-adsk: I will fix it with 
mel.eval("uiRes(\"s_TPStemplateStrings.rExtraAttributes\");")

spinell-adsk:
@seando-adsk , I realized that I forgot to add the attribute documentation as annotation on the widget.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
seando-adsk: Ah right. Can you add a comment here explaining that.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
seando-adsk: Perfect.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
spinell-adsk: done

seando-adsk (approved):

spinell-adsk:
@kxl-adsk no preflight. I only changed python file.

PR #1368, opened 2021-04-27
Fix missing fixturesUtils.runTests in testVP2RenderDelegateUSDPreviewSurface

dgovil:
Had a missing entry point at the end that failed running the test properly via ctest (but works fine interactively otherwise)


dgovil:
One sec, I need to update the baseline image too because the ctest environment has slightly different color results than when running Maya otherwise.

dgovil:
Okay baseline image updated, and tests pass when run via ctest.
Note (to self or if someone else hits this) : If the test is run in a "normal" Maya interactive session (not via the ctest environment) , the colors are very slightly different. Seems to be a color management issue that.

kxl-adsk (approved):

kxl-adsk:
@dgovil Thank you for a very quick fix. This is now making interactive tests pass again. For mac failures, they are to ignore and should be gone in future PFs.

PR #1372, opened 2021-04-28
MAYA-110489 - Prim AE template: categories persist their expand/colla…

seando-adsk:
#### MAYA-110489 - Prim AE template: categories persist their expand/collapse state
* Set new forceRebuild flag true to keep our prim template always destroy/recreate itself. This is needed because
  we are dynmaically changing the transform section and we do not know the attributes to show and their order.
* Fixed bug with xformOpOrder change where copied AE tabs were not freshing to show/remove updated transforms.
* Fixed a bug in UsdAttribute where setting a value (that had no previous value) would throw error.

#### MAYA-110215 - On AE template I'd like to see some categories collapsed
* Simple logic to collapse (by default) certain categories.

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
@@ -20,7 +20,6 @@ import maya.cmds as cmds
 import mayaUsd.ufe as mayaUsdUfe
 import mayaUsd.lib as mayaUsdLib
 import maya.internal.common.ufe_ae.template as ufeAeTemplate
-import maya.internal.common.ae.custom as aecustom
 from maya.common.ui import LayoutManager
 from maya.common.ui import setClipboardData
 
@@ -62,7 +61,7 @@ class UfeAttributesObserver(ufe.Observer):
     def __call__(self, notification):
         if isinstance(notification, ufe.AttributeValueChanged):
             if notification.name() == UsdGeom.Tokens.xformOpOrder:
-                mel.eval("evalDeferred(\"AEbuildControls\");")
+                mel.eval("evalDeferred -low \"refreshEditorTemplates\";")
 
     def onCreate(self, *args):
         ufe.Attributes.addObserver(self._item, self)
@@ -70,17 +69,16 @@ class UfeAttributesObserver(ufe.Observer):
     def onReplace(self, *args):
         pass
 
-class MetaDataCustomControl(aecustom.CustomControl):
+class MetaDataCustomControl(object):
     # Custom control for all prim metadata we want to display.
-    def __init__(self, prim, *args, **kwargs):
+    def __init__(self, prim):
         self.prim = prim
-        super(MetaDataCustomControl, self).__init__(args, kwargs)
 
         # There are four metadata that we always show: primPath, kind, active, instanceable
         # We use a dictionary to store the various other metadata that this prim contains.
         self.extraMetadata = dict()
 
-    def buildControlUI(self):
+    def onCreate(self, *args):
         # Should we display nice names in AE?
         useNiceName = True
         if cmds.optionVar(exists='attrEditorIsLongName'):
@@ -147,6 +145,9 @@ class MetaDataCustomControl(aecustom.CustomControl):
 
         # Update all metadata values.
         self.refresh()
+
+    def onReplace(self, *args):
+        pass
 
     def refresh(self):
         # PrimPath
@@ -276,6 +277,16 @@ class AETemplate(object):
         self.createMetadataSection()
         cmds.editorTemplate(endScrollLayout=True)
 
+        if int(cmds.about(majorVersion=True)) > 2022:
+            # Because of how we dynamically build the Transform attribute section,
+            # we need this template to rebuild each time it is needed. This will
+            # also restore the collapse/expand state of the sections.
+            # Note: in Maya 2022 all UFE templates were forcefully rebuilt, but
+            #       no restore of section states.
+            try:
+                cmds.editorTemplate(forceRebuild=True)
+            except:
+                pass
 
     def addControls(self, controls):
         for c in controls:
@@ -473,7 +484,10 @@ class AETemplate(object):
                 if schemaTypeName == 'UsdGeomXformable':
                     self.createTransformAttributesSection(sectionName, attrsToAdd)
                 else:
-                    self.createSection(sectionName, attrsToAdd)
+                    sectionsToCollapse = ['Curves', 'Point Based', 'Geometric Prim', 'Boundable',
+                                          'Imageable', 'Field Asset', 'Light']
+                    collapse = sectionName in sectionsToCollapse
+                    self.createSection(sectionName, attrsToAdd, collapse)
 
     def suppressArrayAttribute(self):
         # Suppress all array attributes except UsdGeom.Tokens.xformOpOrder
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -61,25 +60,24 @@ def __del__(self):
     def __call__(self, notification):
         if isinstance(notification, ufe.AttributeValueChanged):
             if notification.name() == UsdGeom.Tokens.xformOpOrder:
-                mel.eval("evalDeferred(\"AEbuildControls\");")
+                mel.eval("evalDeferred -low \"refreshEditorTemplates\";")
```
seando-adsk, line 63: I found this Maya command which does a AEbuildControls and hen refreshed the copied layouts. So this fixes a bug where a copied tab would not update correctly after adding a new transform op.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -61,25 +60,24 @@ def __del__(self):
...
 
     def onCreate(self, *args):
         ufe.Attributes.addObserver(self._item, self)
 
     def onReplace(self, *args):
         pass
 
-class MetaDataCustomControl(aecustom.CustomControl):
+class MetaDataCustomControl(object):
```
seando-adsk, line 71: I decided that we should not base our custom controls on the Maya one since it will try and get the plug from attribute (which is UFE here, so no plug). The base class was simple anyways - we don't need it.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -472,7 +483,10 @@ def buildUI(self):
                 if schemaTypeName == 'UsdGeomXformable':
                     self.createTransformAttributesSection(sectionName, attrsToAdd)
                 else:
-                    self.createSection(sectionName, attrsToAdd)
+                    sectionsToCollapse = ['Curves', 'Point Based', 'Geometric Prim', 'Boundable',
+                                          'Imageable', 'Field Asset', 'Light']
+                    collapse = sectionName in sectionsToCollapse
+                    self.createSection(sectionName, attrsToAdd, collapse)
```
seando-adsk, line 489: This is a simple fix for another JIRA item (to collapse certain sections by default).

lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -118,7 +117,11 @@ getUsdAttributeValueAsString(const PXR_NS::UsdAttribute& attr, const PXR_NS::Usd
         return os.str();
     }
 
-    UFE_ASSERT_MSG(false, kErrorMsgFailedConvertToString);
+    try {
+        UFE_ASSERT_MSG(false, kErrorMsgFailedConvertToString);
+    } catch (const std::exception&) {
+        // noop
+    }
```
seando-adsk, line 124: As we discussed we should be trapping all these asserts since they will throw. I also removed any extraneous ones about get first before set to verify the type. We'll let USD handle that.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -147,6 +145,9 @@ def buildControlUI(self):
         # Update all metadata values.
         self.refresh()
 
+    def onReplace(self, *args):
+        pass
```
ppt-adsk, line 149: Should document why this is unused.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
ppt-adsk: I guess the hard-coded list is O.K. because it won't change much?

lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -118,7 +117,11 @@ getUsdAttributeValueAsString(const PXR_NS::UsdAttribute& attr, const PXR_NS::Usd
         return os.str();
     }
 
-    UFE_ASSERT_MSG(false, kErrorMsgFailedConvertToString);
+    try {
+        UFE_ASSERT_MSG(false, kErrorMsgFailedConvertToString);
```
ppt-adsk, line 121: Because this asserts on false, it will always assert --- and you never want it to assert.  Just remove the line and return an empty string.

lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -263,30 +264,29 @@ UsdAttributeEnumString::create(const UsdSceneItem::Ptr& item, const PXR_NS::UsdA
...
-    UFE_ASSERT_MSG(
-        fUsdAttr.Get<PXR_NS::TfToken>(&dummy, getCurrentTime(sceneItem())), kErrorMsgInvalidType);
     PXR_NS::TfToken tok(value);
     setUsdAttr<PXR_NS::TfToken>(fUsdAttr, tok);
 }
 
 Ufe::UndoableCommand::Ptr UsdAttributeEnumString::setCmd(const std::string& value)
 {
     auto self = std::dynamic_pointer_cast<UsdAttributeEnumString>(shared_from_this());
-    UFE_ASSERT_MSG(self, kErrorMsgInvalidType);
```
ppt-adsk, line 289: If self is false just return a nullptr.

lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -358,23 +357,20 @@ template <> void TypedUsdAttribute<std::string>::set(const std::string& value)
...
-        UFE_ASSERT_MSG(
-            fUsdAttr.Get<PXR_NS::TfToken>(&dummy, getCurrentTime(sceneItem())),
-            kErrorMsgInvalidType);
         PXR_NS::TfToken tok(value);
         setUsdAttr<PXR_NS::TfToken>(fUsdAttr, tok);
         return;
     }
 
     // If we get here it means the USDAttribute type wasn't TfToken or string.
-    UFE_ASSERT_MSG(false, kErrorMsgInvalidType);
```
ppt-adsk, line 377: Just remove the line.

ppt-adsk (changes requested):

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
seando-adsk: Right it was a list given to me in the JIRA item. It would only change if Nat wants a change.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
seando-adsk: Okay done.

lib/mayaUsd/ufe/UsdAttribute.cpp
seando-adsk: I actually left this one here on purpose. Asserting here would mean that for some reason we were unable to get the attribute value as a string. So I wanted this to fire so that we would see it and possibly make a fix.

seando-adsk:
@ppt-adsk As discussed I removed all the UFE_ASSERT_MSG and replaced them with TF_XXX (from USD).

ppt-adsk (approved):

ppt-adsk (approved):

PR #1374, opened 2021-04-28
MAYA-108221 - As a user, on the prim AE template I'd like to see usef…

spinell-adsk:
…ul widgets for arrays

-Add missing import for setClipboardData()


spinell-adsk:
@kxl-adsk not sure what happened but I realized there is a missing python import for
https://github.com/Autodesk/maya-usd/pull/1367

kxl-adsk (approved):

PR #1380, opened 2021-04-30
MAYA-111346 : AE: Array attribute does not support Nice/Long name.

spinell-adsk:
Add support for nice/long name for array attribute type.


spinell-adsk:
@seando-adsk I'm querying the attrEditorIsLongName optionvar in the AETemplate and I'm passing the result to the Metadata control and each array control

seando-adsk (approved):
Looks good. Thanks.

seando-adsk:
@kxl-adsk Single python file change - no preflight should be required.

PR #1386, opened 2021-05-04
MAYA-108586: Keep the TF_ERROR message format consistence with TF_WARN and TF_STATUS 

HamedSabri-adsk:
This PR keeps the TF_ERROR message format consistence with TF_WARN and TF_STATUS by displaying the by only displaying the **commentary** string.

Example below shows the case where an image resource is failed to created which results in a vague message to the user.

Before:
```
// Error: Failed verification: ' image ' -- Coding Error in _LoadTexture at line 422 of S:\jenkins\workspace\ecg-mayausd-full-2022-python3-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\render\vp2RenderDelegate\material.cpp // 
// Error: Failed verification: ' image ' -- Coding Error in _LoadTexture at line 422 of S:\jenkins\workspace\ecg-mayausd-full-2022-python3-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\render\vp2RenderDelegate\material.cpp // 
// Error: Failed verification: ' image ' -- Coding Error in _LoadTexture at line 422 of S:\jenkins\workspace\ecg-mayausd-full-2022-python3-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\render\vp2RenderDelegate\material.cpp // 
// Error: Failed verification: ' image ' -- Coding Error in _LoadTexture at line 422 of S:\jenkins\workspace\ecg-mayausd-full-2022-python3-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\render\vp2RenderDelegate\material.cpp // 
// Error: Failed verification: ' image ' -- Coding Error in _LoadTexture at line 422 of S:\jenkins\workspace\ecg-mayausd-full-2022-python3-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\render\vp2RenderDelegate\material.cpp // 
// Error: Failed verification: ' image ' -- Coding Error in _LoadTexture at line 422 of S:\jenkins\workspace\ecg-mayausd-full-2022-python3-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\render\vp2RenderDelegate\material.cpp // 
// Error: Failed verification: ' image ' -- Coding Error in _LoadTexture at line 422 of S:\jenkins\workspace\ecg-mayausd-full-2022-python3-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\render\vp2RenderDelegate\material.cpp // 
// Error: Failed verification: ' image ' -- Coding Error in _LoadTexture at line 422 of S:\jenkins\workspace\ecg-mayausd-full-2022-python3-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\render\vp2RenderDelegate\material.cpp // 
```

After the fix with a more meaningful message:
```
// Error: Failed verification: ' image ' -- Unable to create an image from RetroTVBody_Albedo.png // 
// Error: Failed verification: ' image ' -- Unable to create an image from RetroTVBody_Emissive.png // 
// Error: Failed verification: ' image ' -- Unable to create an image from RetroTVBody_Metallic.png // 
// Error: Failed verification: ' image ' -- Unable to create an image from RetroTVBody_Normal.png // 
// Error: Failed verification: ' image ' -- Unable to create an image from RetroTVBody_AO.png // 
// Error: Failed verification: ' image ' -- Unable to create an image from RetroTVBody_Roughness.png // 
// Error: Failed verification: ' image ' -- Unable to create an image from RetroTVScreen_Albedo.png // 
// Error: Failed verification: ' image ' -- Unable to create an image from RetroTVScreen_Roughness.png // 
```

### DIFF
--- a/lib/mayaUsd/utils/diagnosticDelegate.cpp
+++ b/lib/mayaUsd/utils/diagnosticDelegate.cpp
@@ -107,7 +107,7 @@ void UsdMayaDiagnosticDelegate::IssueError(const TfError& err)
     // In addition, always display the full call site for errors by going
     // through _FormatDiagnostic.
     if (ArchIsMainThread()) {
-        MGlobal::displayError(_FormatDiagnostic(err));
+        MGlobal::displayError(err.GetCommentary().c_str());
     } else {
         std::cerr << _FormatDiagnostic(err) << std::endl;
     }
--- a/lib/mayaUsd/utils/diagnosticDelegate.cpp
+++ b/lib/mayaUsd/utils/diagnosticDelegate.cpp
@@ -23,7 +23,7 @@
 
 #include <maya/MGlobal.h>
 
-#include <iostream>
+#include <ghc/filesystem.hpp>
 
 PXR_NAMESPACE_OPEN_SCOPE
 
@@ -67,17 +67,13 @@ class _WarningOnlyDelegate : public UsdUtilsCoalescingDiagnosticDelegate
 
 static MString _FormatDiagnostic(const TfDiagnosticBase& d)
 {
-    // strip out all the characters before "maya-usd"
-    std::string filterFilePath { d.GetContext().GetFile() };
-    filterFilePath = filterFilePath.substr(filterFilePath.find("maya-usd"));
-
     const std::string msg = TfStringPrintf(
         "%s -- %s in %s at line %zu of %s",
         d.GetCommentary().c_str(),
         TfDiagnosticMgr::GetCodeName(d.GetDiagnosticCode()).c_str(),
         d.GetContext().GetFunction(),
         d.GetContext().GetLine(),
-        filterFilePath.c_str());
+        ghc::filesystem::path(d.GetContext().GetFile()).relative_path().string().c_str());
 
     return TfGetEnvSetting(MAYAUSD_SHOW_FULL_DIAGNOSTICS) ? msg.c_str() : d.GetCommentary().c_str();
 }
--- a/lib/mayaUsd/utils/diagnosticDelegate.cpp
+++ b/lib/mayaUsd/utils/diagnosticDelegate.cpp
@@ -67,15 +67,18 @@ class _WarningOnlyDelegate : public UsdUtilsCoalescingDiagnosticDelegate
 
 static MString _FormatDiagnostic(const TfDiagnosticBase& d)
 {
-    const std::string msg = TfStringPrintf(
-        "%s -- %s in %s at line %zu of %s",
-        d.GetCommentary().c_str(),
-        TfDiagnosticMgr::GetCodeName(d.GetDiagnosticCode()).c_str(),
-        d.GetContext().GetFunction(),
-        d.GetContext().GetLine(),
-        ghc::filesystem::path(d.GetContext().GetFile()).relative_path().string().c_str());
-
-    return TfGetEnvSetting(MAYAUSD_SHOW_FULL_DIAGNOSTICS) ? msg.c_str() : d.GetCommentary().c_str();
+    if (!TfGetEnvSetting(MAYAUSD_SHOW_FULL_DIAGNOSTICS)) {
+        return d.GetCommentary().c_str();
+    } else {
+        const std::string msg = TfStringPrintf(
+            "%s -- %s in %s at line %zu of %s",
+            d.GetCommentary().c_str(),
+            TfDiagnosticMgr::GetCodeName(d.GetDiagnosticCode()).c_str(),
+            d.GetContext().GetFunction(),
+            d.GetContext().GetLine(),
+            ghc::filesystem::path(d.GetContext().GetFile()).relative_path().string().c_str());
+        return msg.c_str();
+    }
 }
 
 static MString _FormatCoalescedDiagnostic(const UsdUtilsCoalescingDiagnosticDelegateItem& item)
@@ -115,7 +118,7 @@ void UsdMayaDiagnosticDelegate::IssueError(const TfError& err)
     const auto diagnosticMessage = _FormatDiagnostic(err);
 
     if (ArchIsMainThread()) {
-        MGlobal::displayInfo(diagnosticMessage);
+        MGlobal::displayError(diagnosticMessage);
     } else {
         std::cerr << diagnosticMessage << std::endl;
     }
@@ -145,7 +148,7 @@ void UsdMayaDiagnosticDelegate::IssueWarning(const TfWarning& warning)
     const auto diagnosticMessage = _FormatDiagnostic(warning);
 
     if (ArchIsMainThread()) {
-        MGlobal::displayInfo(diagnosticMessage);
+        MGlobal::displayWarning(diagnosticMessage);
     } else {
         std::cerr << diagnosticMessage << std::endl;
     }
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/utils/diagnosticDelegate.cpp
```diff
@@ -107,7 +107,7 @@ void UsdMayaDiagnosticDelegate::IssueError(const TfError& err)
     // In addition, always display the full call site for errors by going
     // through _FormatDiagnostic.
     if (ArchIsMainThread()) {
-        MGlobal::displayError(_FormatDiagnostic(err));
+        MGlobal::displayError(err.GetCommentary().c_str());
```
seando-adsk, line 110: Why not do a fix inside of _FormatDiagnostic? If we fall, into the else then it will still print all the info (function/line/file) for error/status/warning. How about inside of _FormatDiagnostic you use if DEBUG check to print all info and in release only the commentary?

hamedsabri: @seando-adsk Yeah, I thought about doing it that way but then not all the developers work in DEBUG. :) I myself most of the time work in RelWithDebInfo and barely ever bother with Debug.

The only other option would be to introduce an ENV variable to control this behaviour but again this could be controversial.

seando-adsk: But at least having it in DEBUG would provide an option to show all the info (file/function/line). With your change the only way to get that is to not be in the main thread, which I don't know how/why we wouldn't be.

hamedsabri: Sure, I don't have a strong preference but do people really need this? Why stop at "Errors"? What about Warnings? 

For example, I see in many places in our code based people tend to use TF_WARN when checking for object validation.

```
if (shaderInstance == nullptr) {
    TF_WARN("Failed to create shader instance for %s", nodeId.asChar());
    break;
}
```

lib/mayaUsd/utils/diagnosticDelegate.cpp
```diff
@@ -68,7 +74,8 @@ static MString _FormatDiagnostic(const TfDiagnosticBase& d)
         d.GetContext().GetFunction(),
         d.GetContext().GetLine(),
         d.GetContext().GetFile());
-    return msg.c_str();
+
+    return TfGetEnvSetting(MAYAUSD_SHOW_FULL_DIAGNOSTICS) ? msg.c_str() : d.GetCommentary().c_str();
```
kxl-adsk, line 78: Most of the time we will not be in full diagnostics mode. With the current code structure, we will always generate a full string. String operations are costly, how about we insert if/else condition and only generate msg string when needed?

lib/mayaUsd/utils/stageCache.cpp
```diff
@@ -50,7 +50,6 @@ struct _OnSceneResetListener : public TfWeakBase
 
     void OnSceneReset(const UsdMayaSceneResetNotice& notice)
     {
-        TF_STATUS("Clearing USD Stage Cache");
```
kxl-adsk, line 53: Thank you!

lib/mayaUsd/utils/diagnosticDelegate.cpp
seando-adsk: Yes agreed. I also wonder about the need to check this new env var on every call for _FormatDiagnostic(). Do we need to be able to modify this env var during a Maya run? How about just getting this var once into a static variable?

lib/mayaUsd/utils/diagnosticDelegate.cpp
```diff
@@ -104,12 +111,13 @@ void UsdMayaDiagnosticDelegate::IssueError(const TfError& err)
...
     // Errors are never batched. They should be rare, and in those cases, we
     // want to see them separately.
-    // In addition, always display the full call site for errors by going
-    // through _FormatDiagnostic.
+
+    const auto diagnosticMessage = _FormatDiagnostic(err);
+
     if (ArchIsMainThread()) {
-        MGlobal::displayError(_FormatDiagnostic(err));
+        MGlobal::displayInfo(diagnosticMessage);
```
seando-adsk, line 118: Was it an intention change (or a copy/paste mistake) that you always call displayInfo() even for the error/warning?

lib/mayaUsd/utils/diagnosticDelegate.cpp
```diff
@@ -68,7 +74,8 @@ static MString _FormatDiagnostic(const TfDiagnosticBase& d)
         d.GetContext().GetFunction(),
         d.GetContext().GetLine(),
         d.GetContext().GetFile());
```
seando-adsk, line 76: I wonder if we should strip off the path from this and leave only the filename? Isn't part of the problem that we are leaking our paths from the build VMs. I would again suggest that in Debug you leave it alone, but in RelWithDebInfo you strip it. Our Debug builds are never shipped.

lib/mayaUsd/utils/diagnosticDelegate.cpp
HamedSabri-adsk: @kxl-adsk @seando-adsk I agree with you that std::string allocation can be costly and am very well aware of them. But the cost here is super minimal. I honestly don't believe we don't need to worry about operation costs in here. We can discuss this internally if needed.

lib/mayaUsd/utils/diagnosticDelegate.cpp
seando-adsk: okay.

lib/mayaUsd/utils/diagnosticDelegate.cpp
seando-adsk: But a user can enable the new env var and then they will see all the paths from our VMs.

ppt-adsk: Can we not create a simple regex to strip everything before maya-usd/src?

lib/mayaUsd/utils/diagnosticDelegate.cpp
```diff
@@ -67,13 +67,17 @@ class _WarningOnlyDelegate : public UsdUtilsCoalescingDiagnosticDelegate
...
+    filterFilePath = filterFilePath.substr(filterFilePath.find("maya-usd"));
+
     const std::string msg = TfStringPrintf(
         "%s -- %s in %s at line %zu of %s",
         d.GetCommentary().c_str(),
         TfDiagnosticMgr::GetCodeName(d.GetDiagnosticCode()).c_str(),
         d.GetContext().GetFunction(),
         d.GetContext().GetLine(),
-        d.GetContext().GetFile());
+        filterFilePath.c_str());
```
HamedSabri-adsk, line 80: filter the file path so it discards everything before the project name (`maya-usd`).

e.g
` Error: Failed verification: ' image ' -- Coding Error in _LoadTexture at line 422 of maya-usd\lib\mayaUsd\render\vp2RenderDelegate\material.cpp`

lib/mayaUsd/utils/diagnosticDelegate.cpp
HamedSabri-adsk: I would be happy to see the profiler results to understand where we hit this issue.

seando-adsk (approved):
Looks good. Thanks for changing the msg if/else.

kxl-adsk (approved):

PR #1404, opened 2021-05-10
MAYA-110958 Store render item dirty bits with each render item, so hidden items

williamkrick:
which are dirtied update correctly when shown.


williamkrick:
The pre flight run failed on Maya 2022 [USD 0.20.02, Python 2, Interactive] testVP2RenderDelegatePerInstanceInheritedData with image inheritedDisplayColor_noPxrMtls_unselected.png. I ran this exact same test locally and it passed so I am not sure what is going on.

PR #1405, opened 2021-05-10
Updating visual baseline of PxrUsdPreviewSurfaceDraw test

dj-mcg:
Due to recent USDPreviewSurface improvements, one of our draw baselines changed. This doesn't seem to cause mayaUsd open source tests to fail as they do not seem to compare baselines for this test, but this does fix an internal pixar test


kxl-adsk:
There is no connection between the test failure reported by PF and this change. We don't clear temp folders at each run, so some tests can hit conflicts. Something to investigate but not a blocker for merge.

PR #1406, opened 2021-05-11
MAYA-110517 - Crash on redo of creation and manipulation of prim

spinell-adsk:
- Add python test (The fix is in maya).

### DIFF
--- a/test/lib/ufe/testSelection.py
+++ b/test/lib/ufe/testSelection.py
@@ -445,7 +445,7 @@ class SelectTestCase(unittest.TestCase):
 
         self.assertTrue(sn.empty())
 
-    @unittest.skipUnless(mayaUtils.previewReleaseVersion() >= 127, 'Requires Maya fixes only available in Maya Preview Release 127 or later.')
+    @unittest.skipUnless(mayaUtils.previewReleaseVersion() >= 126, 'Requires Maya fixes only available in Maya Preview Release 126 or later.')
     def testMayaSelectUndoPrimCreation(self):
         '''Test if the SceneItem's prim is still valid on selection after the prim creation is undone then redone'''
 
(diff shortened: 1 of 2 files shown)


test/lib/ufe/testSelection.py
```diff
@@ -445,5 +445,53 @@ def testMayaSelectSwitchVariant(self):
 
         self.assertTrue(sn.empty())
 
+    @unittest.skipUnless(mayaUtils.previewReleaseVersion() >= 125, 'Requires Maya fixes only available in Maya Preview Release 125 or later.')
```
ppt-adsk, line 448: Shouldn't this be 126 or later?  Customers will have Preview Release 125, and it doesn't have your fix.

test/lib/ufe/testSelection.py
```diff
@@ -445,5 +445,53 @@ def testMayaSelectSwitchVariant(self):
...
+    def testMayaSelectUndoPrimCreation(self):
+        '''Test if the SceneItem's prim is still valid on selection after the prim creation is undo than redo'''
```
ppt-adsk, line 450: "undo than redo" --> "undone then redone".

test/lib/ufe/testSelection.py
```diff
@@ -445,5 +445,53 @@ def testMayaSelectSwitchVariant(self):
...
+
+        # helper function to check if the current
+        # selected SceneItem's prim is valid.
+        def checkSelectedSceneItemPrim():
```
ppt-adsk, line 454: Why not pass in the scene item that you expect will be the selected item, then compare it with globalSn.front()?  Would make for an even stronger test.

ppt-adsk (approved):

ppt-adsk (approved):

ppt-adsk (approved):

PR #1407, opened 2021-05-11
MAYA-111184 - Title for Shaping API shows up incorrect in the Attribu…

spinell-adsk:
…te Editor

Fix attribute editor applied schema section name when name contains multiple "api" string


seando-adsk (approved):
Cool trick for a one-time reverse replace.

PR #1408, opened 2021-05-11
MAYA-111084 - LE save icon missing when scaling is 125%

seando-adsk:
#### MAYA-111084 - LE save icon missing when scaling is 125%
* We don't provide icons at 125% so make sure to account for that.

### DIFF
--- a/lib/usd/ui/layerEditor/layerEditorWidget.cpp
+++ b/lib/usd/ui/layerEditor/layerEditorWidget.cpp
@@ -54,8 +54,16 @@ using namespace UsdLayerEditor;
 // properly support high dpi with style sheets
 QString getDPIPixmapName(QString baseName)
 {
-    QString suffix(QString::number(IS_MAC_OS ? 100 : DPIScale(100)));
-    return baseName + "_" + suffix + ".png";
+#ifdef Q_OS_DARWIN
+    return baseName + "_100.png";
+#else
+    const auto scale = utils->dpiScale();
+    if (scale >= 2.0)
+        return baseName + "_200.png";
+    else if (scale >= 1.5)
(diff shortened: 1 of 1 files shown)


lib/usd/ui/layerEditor/layerEditorWidget.cpp
```diff
@@ -54,8 +54,16 @@ using namespace UsdLayerEditor;
...
+#ifdef Q_OS_DARWIN
+    return baseName + "_100.png";
+#else
+    const auto scale = utils->dpiScale();
+    if (scale >= 2.0)
+        return baseName + "_200.png";
+    else if (scale >= 1.5)
+        return baseName + "_150.png";
+    return baseName + "_100.png";
+#endif
```
seando-adsk, line 66: When running at custom scaling 125%, the base name would become "<basename>_125.png" - but we do not provide at 125% icon file. So reworked the code to make sure we always return either _100/_150/_100 - which are the three sizes we support.

lib/usd/ui/layerEditor/layerEditorWidget.cpp
```diff
@@ -71,6 +79,8 @@ static void setupButtonWithHIGBitmaps(QPushButton* button, QString baseName)
     QPushButton {
         padding : %1px;
         background-image: url(%2);
+        background-position: center center;
+        background-repeat: no-repeat;
```
seando-adsk, line 83: At 125 the icon used is _100 so there was extra space and the icon was centered and actually repeated. So this centers it (vert/horz) and turns off repeat. I checked all supported scaling factors (100/125/150/200).

spinell-adsk (approved):
look good to me.

PR #1412, opened 2021-05-12
MAYA-110368 send ufe notifications on undo and redo

williamkrick:
The only thing that is not working fully consistently is when we set the value we get one Ufe::AttributeValueChanged for the height changing. On undo and redo we get two Ufe::AttributeValueChanged: one for the height and one for the bounds. I would prefer if execute, undo and redo all sent exactly the same notifications.


test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateInteractiveWorkflows.py
```diff
@@ -0,0 +1,139 @@
...
+    """
+    Tests imaging using the Viewport 2.0 render delegate when using per-instance
+    inherited data on instances.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        # The test USD data is authored Z-up, so make sure Maya is configured
+        # that way too.
+        # cmds.upAxis(axis='z')
```
ppt-adsk, line 49: Commented out?

test/lib/mayaUsd/render/vp2RenderDelegate/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ set(TEST_SCRIPT_FILES "")
 
 list(APPEND TEST_SCRIPT_FILES
 	testVP2RenderDelegateGeomSubset.py
+	testVP2RenderDelegateInteractiveWorkflows.py #probably ufe v2 only
```
ppt-adsk, line 8: The Attributes interface is UFE v2, so this needs to be moved.

ppt-adsk (changes requested):

ppt-adsk (approved):

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -235,18 +235,18 @@ void StagesSubject::stageChanged(
             // We need to send some notifs so Maya can update (such as on undo
```
williamkrick, line 235: Just above here, I could modify the loop over GetResyncedPaths to get access to the SdfChangeList::entry objects and access the same flags I get in the changedInfoOnlyPaths loop.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -406,7 +408,20 @@ void StagesSubject::stageChanged(
...
         if (sendValueChangedFallback) {
-            valueChanged(ufePath, changedPath.GetNameToken());
+
+            // check to see if there is an entry which Ufe should notify about.
+            std::vector<const SdfChangeList::Entry*> entries = it.base()->second;
+            for (const auto& entry : entries)
+            {
+                // Adding an inert prim means we created a primSpec for an ancestor of
+                // a prim which has a real change to it.
+                if (entry->flags.didAddInertPrim || entry->flags.didRemoveInertPrim)
```
williamkrick, line 418: This is the new filtering that catches the over primSpec creation & deletion. There are loads of other flags which could be relevant to check here.

ppt-adsk (approved):
Great work William, thanks for getting to the bottom of this!

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -235,18 +235,18 @@ void StagesSubject::stageChanged(
             // We need to send some notifs so Maya can update (such as on undo
             // to move the transform manipulator back to original position).
             const TfToken nameToken = changedPath.GetNameToken();
+            auto          usdPrimPathStr = changedPath.GetPrimPath().GetString();
+            auto ufePath = stagePath(sender) + Ufe::PathSegment(usdPrimPathStr, g_USDRtid, '/');
```
seando-adsk, line 239: I have a feeling this will not pass the clang-format. Did you run it locally?

seando-adsk (approved):
Great work. Thanks for adding the comments to the existing attribute test.

ppt-adsk (approved):

seando-adsk (approved):

PR #1416, opened 2021-05-14
MAYA-109893 Integrate 21.05 UsdPreviewSurface shader into mayaUsd plugin

JGamache-autodesk:
- Leverages recent code in Maya that simplifies Light integration.
- Original shading fragment still present for older versions of Maya.


JGamache-autodesk:
Brings the Helmet discussed in https://github.com/PixarAnimationStudios/USD/issues/1174 closer to Pixar's usdView version.
Helmet using original UsdPreviewSurface.xml shader fragment:
HelmetBefore
Using latest Pixar shader code from UsdPreviewSurfaceNG.xml fragment:
helmetNG Dx11

JGamache-autodesk:
Still missing: Normal maps, as is plainly visible on the air intake tubes.

JGamache-autodesk:
Here is the gramophone, for comparison:
image

williamkrick (commented):
To what extent should I be reviewing usdPreviewSurfaceNGLighting.xml? There is a lot of code there which looks reasonable at first glance, but I didn't try to delve too deeply into it.

The rest of the files are all fine.

williamkrick:
> 
> 
> To what extent should I be reviewing usdPreviewSurfaceNGLighting.xml? There is a lot of code there which looks reasonable at first glance, but I didn't try to delve too deeply into it.
> 
> The rest of the files are all fine.

I did some comparisons between the Pixar glslfx file USD/pxr/usdImaging/plugin/usdShaders/shaders/previewSurface.glslfx and the Maya version usdPriviewSurfaceNGLighting.xml and they seem aligned with one another.

williamkrick (approved):

kxl-adsk (commented):
@JGamache-autodesk I see you updated `testMetallicResponse.png` baseline. I would assume this baseline will be different depending on the Maya version?

JGamache-autodesk:
@kxl-adsk The result will be different based on the Maya version. I will add a check otherwise the preflight will not work.

cmake/modules/FindMaya.cmake
```diff
@@ -301,13 +301,16 @@ find_program(MAYA_PY_EXECUTABLE
 set(MAYA_LIGHTAPI_VERSION 1)
 if(IS_MACOSX)
     set(MAYA_DSO_SUFFIX ".dylib")
+    sef(MAYA_DSO_PREFIX "lib")
```
dgovil, line 304: Typo? `sef` -> `set`

kxl-adsk (approved):

PR #1422, opened 2021-05-18
MAYA-105710: fix USD files fail to load on mapped mounted volume.

HamedSabri-adsk:
This PR fixes the issue with loading usd from mapped drivers.

#### Steps to reproduce the error:
```
    Plugin and external drive or USB Drive to your computer.
    Go to Disk Manager.
    Right Click on your connected drive.
    Select "Change Drive Letter and Path..."
    Select "Add".
    Use "Mount in the following empty NTFS folder:"
    Locate the folder you want to mount to.
    Extract the attached folder to that location.
    Open Maya and then open the *.ma file that contains usd with relative path
    Notice that the USD file doesn't load.
```
Note: Opening the same file if saved on a local drive or a network drive works without issues.

#### Description:
There is a bug in `gulark's filesystem::canonical` which fails to properly resolve relative path. C++17 std::filesystem::canonical doesn't have this issue and properly resolves the relative path.

**ghc::filesystem::canonical**
```
auto fullPath = ghc::filesystem::path(currentFileDir).append(relativeFilePath); 
// C:\\MOUNT_ME\\RTS\\RTS\\master_files\\..\\referenced_files\\USD\\barrelTests\\Library\\Assets\\Barrels\\Barrels.usd
auto path = ghc::filesystem::canonical(fullPath, errorCode); // error 
```

**std::filesystem::canonical**
```
auto fullPath = std::filesystem::path(currentFileDir).append(relativeFilePath);
//"C:\\MOUNT_ME\\RTS\\RTS\\master_files\\../referenced_files/USD/barrelTests/Library/Assets/Barrels/Barrels.usd"
auto path = std::filesystem::canonical(fullPath, errorCode); // Ok, good 
```

#### Workaround:

To work around this, I removed the need to call `filesystem::canonical` all together since the given path to `ArGetResolver().ConfigureResolverForAsset` is resolved properly. 

https://github.com/Autodesk/maya-usd/blob/092f52057f46c51c3ed725d9e25422c52e0bdd0c/lib/mayaUsd/nodes/proxyShapeBase.cpp#L660

e.g 
`C:\\MOUNT_ME\\RTS\\RTS\\master_files\\..\\referenced_files\\USD\\barrelTests\\Library\\Assets\\Barrels\\Barrels.usd`

### DIFF
--- a/cmake/gulark.cmake
+++ b/cmake/gulark.cmake
@@ -38,7 +38,7 @@ else()
     FetchContent_Declare(
         ${CONTENT_NAME}
         GIT_REPOSITORY https://github.com/gulrak/filesystem.git
-        GIT_TAG        3d3c02ce35dcc68b5ebb34f21cb1fc507be9a66e
+        GIT_TAG        4e21ab305794f5309a1454b4ae82ab9a0f5e0d25
         USES_TERMINAL_DOWNLOAD TRUE
         GIT_CONFIG     advice.detachedHead=false
     )
--- a/lib/mayaUsd/utils/utilFileSystem.cpp
+++ b/lib/mayaUsd/utils/utilFileSystem.cpp
@@ -157,10 +157,12 @@ std::string UsdMayaUtilFileSystem::resolveRelativePathWithinMayaContext(
         return relativeFilePath;
     }
 
-    auto path = ghc::filesystem::path(currentFileDir).append(relativeFilePath);
+    std::error_code errorCode;
+    auto            path = ghc::filesystem::canonical(
+        ghc::filesystem::path(currentFileDir).append(relativeFilePath), errorCode);
 
-    // if file does not exist
-    if (path.empty()) {
+    if (errorCode) {
+        // file does not exist
         return std::string();
     }
 
--- a/lib/mayaUsd/utils/utilFileSystem.cpp
+++ b/lib/mayaUsd/utils/utilFileSystem.cpp
@@ -143,23 +143,24 @@ std::string UsdMayaUtilFileSystem::resolveRelativePathWithinMayaContext(
     const MObject&     proxyShape,
     const std::string& relativeFilePath)
 {
-    if (relativeFilePath.length() < 3)
+    if (relativeFilePath.length() < 3) {
         return relativeFilePath;
+    }
 
     std::string currentFileDir = getMayaReferencedFileDir(proxyShape);
 
-    if (currentFileDir.empty())
+    if (currentFileDir.empty()) {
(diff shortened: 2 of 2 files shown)


JGamache-autodesk (approved):

JGamache-autodesk:
You will most likely need to update some unit tests that have hardcoded canonical paths in them.

kxl-adsk (approved):
Great find @HamedSabri-adsk !

We are changing what `resolveRelativePathWithinMayaContext` is returning, i.e. used to be the canonical version of the path. This path is consumed then by a few different methods inside of proxy shape compute, including opening/finding a layer for a given path. I was wondering how it will work for layer identifiers and whether or not they will include non-canonical paths...but everything seems to be correct there.

I'm fine with the change, but two things would be good considering doing:
1. Creating a defect in gulrak
2. Check with @fabal if this change will work for AL as well. I remember we refactored this method out of the AL plugin so there may be additional dependencies we don't know about that require returned path to be canonical.

HamedSabri-adsk:
@kxl-adsk  Great points. I will log a defect against gulark sometimes today or tomorrow.

@fabal I would be happy to hear your thoughts on this change ? Thanks

cmake/gulark.cmake
```diff
@@ -38,7 +38,7 @@ else()
     FetchContent_Declare(
         ${CONTENT_NAME}
         GIT_REPOSITORY https://github.com/gulrak/filesystem.git
-        GIT_TAG        3d3c02ce35dcc68b5ebb34f21cb1fc507be9a66e
+        GIT_TAG        4e21ab305794f5309a1454b4ae82ab9a0f5e0d25
```
HamedSabri-adsk, line 41: Switching to v 1.5.6 which fixed the bug in ghc::filesystem::canonical

https://github.com/gulrak/filesystem/issues/124

lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -157,10 +157,12 @@ std::string UsdMayaUtilFileSystem::resolveRelativePathWithinMayaContext(
         return relativeFilePath;
     }
 
-    auto path = ghc::filesystem::path(currentFileDir).append(relativeFilePath);
+    std::error_code errorCode;
+    auto            path = ghc::filesystem::canonical(
+        ghc::filesystem::path(currentFileDir).append(relativeFilePath), errorCode);
```
HamedSabri-adsk: revert my workaround now that bug is fixed.

kxl-adsk (approved):

PR #1426, opened 2021-05-19
Fixed bug in reading paired xformOp operations which would result in incorrect transform applied in Maya

ysiewappl:
Hi:

This PR fixes an issue we are encountering where import of USD files that have the pivot translated paired with its corresponding `!invert!` operation results in Maya-USD ignoring the inverse operation, resulting in an incorrect transform applied on import. This should also fix multiple xformOp ops not being applied correctly in sequence as well since the previous code essentially overwrote one xformOp with another.

An example of a basic Maya scene setup which demonstrates the issue is attached:
[unit_test.zip](https://github.com/Autodesk/maya-usd/files/6510476/unit_test.zip)

Preview of the scene:
image

Exporting the scene to USD and re-importing it in a new Maya session, without the fix:
image

With the fix, the inverse xformOp operation is no longer ignored and the result is correct on import.
image

### DIFF
--- a/lib/mayaUsd/fileio/translators/translatorXformable.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorXformable.cpp
@@ -22,6 +22,7 @@
 #include <pxr/base/gf/math.h>
 #include <pxr/base/gf/matrix4d.h>
 #include <pxr/base/gf/vec3d.h>
+#include <pxr/base/tf/diagnostic.h>
 #include <pxr/base/tf/token.h>
 #include <pxr/pxr.h>
 #include <pxr/usd/usd/stage.h>
@@ -30,6 +31,7 @@
 #include <pxr/usd/usdGeom/xformCommonAPI.h>
 #include <pxr/usd/usdGeom/xformable.h>
 
+#include <maya/MAnimUtil.h>
 #include <maya/MDagModifier.h>
 #include <maya/MEulerRotation.h>
 #include <maya/MFnAnimCurve.h>
@@ -122,8 +124,8 @@ _getXformOpAsVec3d(const UsdGeomXformOp& xformOp, GfVec3d& value, const UsdTimeC
 // Sets the animation curve (a knot per frame) for a given plug/attribute
 static void _setAnimPlugData(
     MPlug                           plg,
-    std::vector<double>&            value,
-    MTimeArray&                     timeArray,
+    const std::vector<double>&      value,
+    const MTimeArray&               timeArray,
     const UsdMayaPrimReaderContext* context)
 {
     MStatus      status;
@@ -135,7 +137,7 @@ static void _setAnimPlugData(
     MObject animObj = animFn.create(plg, nullptr, &status);
     if (status == MS::kSuccess) {
         MDoubleArray valueArray(&value[0], value.size());
-        animFn.addKeys(&timeArray, &valueArray);
+        animFn.addKeys(&(const_cast<MTimeArray&>(timeArray)), &valueArray);
         if (context) {
             context->RegisterNewMayaNode(animFn.name().asChar(), animObj);
         }
@@ -146,8 +148,72 @@ static void _setAnimPlugData(
     }
 }
 
+static void _applyAnimPlugData(
+    MPlug                           plg,
+    const std::vector<double>&      value,
+    const MTimeArray&               timeArray,
+    const UsdMayaPrimReaderContext* context)
+{
+    MStatus      status;
+    MFnAnimCurve animFn;
+    // Make the plug keyable before attaching an anim curve
+    if (!plg.isKeyable()) {
+        plg.setKeyable(true);
+    }
+
+    MObjectArray animCurves;
+    MAnimUtil::findAnimation(plg, animCurves);
+    // NOTE: (yliangsiew) If there are no existing animCurves, then just set the values as-is.
+    // Otherwise, we need to additively apply the values to the existing keyframes.
+    if (animCurves.length() == 0) {
+        MObject animObj = animFn.create(plg, nullptr, &status);
+        if (status == MS::kSuccess) {
+            MDoubleArray valueArray(&value[0], value.size());
+            animFn.addKeys(&(const_cast<MTimeArray&>(timeArray)), &valueArray);
+            if (context) {
+                context->RegisterNewMayaNode(animFn.name().asChar(), animObj);
+            }
+        } else {
+            MString mayaPlgName = plg.partialName(true, true, true, false, true, true, &status);
+            TF_RUNTIME_ERROR(
+                "Failed to create animation object for attribute: %s", mayaPlgName.asChar());
+        }
+    } else {
+        // NOTE: (yliangsiew) We will assume that we will apply all our new values to the first
+        // animCurve affecting the plug.
+        MObject animCurve = animCurves[0];
+        status = animFn.setObject(animCurve);
+        if (status != MS::kSuccess) {
+            TF_RUNTIME_ERROR(
+                "Failed to create animation object for attribute: %s",
+                plg.partialName(true, true, true, false, true, true).asChar());
+            return;
+        }
+        unsigned int numKeys = animFn.numKeys();
+        MDoubleArray valueArray(&value[0], value.size());
+        if (numKeys == 0) {
+            animFn.addKeys(&(const_cast<MTimeArray&>(timeArray)), &valueArray);
+        } else {
+            for (unsigned int i = 0; i < timeArray.length(); ++i) {
+                MTime        curTime = timeArray[i];
+                unsigned int keyIdx = 0;
+                bool         hasKey = animFn.find(curTime, keyIdx, &status);
+                TF_VERIFY(status == MStatus::kSuccess, "Mismatch between times and key indices.");
+                if (!hasKey) {
+                    animFn.addKey(curTime, valueArray[i]);
+                } else {
+                    double curVal = animFn.value(keyIdx, &status);
+                    TF_VERIFY(status == MStatus::kSuccess, "Unable to retrieve value for key.");
+                    status = animFn.setValue(keyIdx, curVal + valueArray[i]);
+                    TF_VERIFY(status == MStatus::kSuccess, "Unable to set value for key.");
+                }
+            }
+        }
+    }
+}
+
 // Returns true if the array is not constant
-static bool _isArrayVarying(std::vector<double>& value)
+static bool _isArrayVarying(const std::vector<double>& value)
 {
     bool isVarying = false;
     for (unsigned int i = 1; i < value.size(); i++) {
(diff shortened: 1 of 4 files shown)


ysiewappl:
Ok, I've taken a closer look at the specific unit test that failed and it's indicating that I'm missing something in my logic here and how Maya pivot transformations work. It looks like simply not ignoring the inverse transform operations and applying them as well is insufficient. However, I'm having trouble trying to figure out what the correct mapping of USD xformOps to Maya rotate/scale pivot/pivotTranslation should be, though; I haven't quite figured out how exactly pivotTranslation in Maya works. I'd appreciate it if someone from Autodesk could review this and advise. Thanks!

ysiewappl:
Ok, after reading a bit on the `UsdGeomXformCommonAPI`, I've pushed a modification which I think is reasonable; for paired pivot ops that have no other transformation operations, the code _will_ apply the inverse of the pivot (thus resulting in no moved pivot) as the USD suggests; for anything else it will continue to respect the ignoring of the pivot's inverse operation.

This is causing one of the existing unit tests (`testImportMayaXformVariations`) to fail, which boils down to the following as far as I can find:

```
#usda 1.0
(
    defaultPrim = "topPrim"
    metersPerUnit = 0.01
    upAxis = "Y"
)

def Xform "topPrim" (
    kind = "assembly"
)
{
    def Xform "mayaXform_100001000000"
    {
        float3 xformOp:translate:rotatePivot = (1.6550425, 3.952048, 1.5850885)
        uniform token[] xformOpOrder = ["xformOp:translate:rotatePivot", "!invert!xformOp:translate:rotatePivot"]
        # I've added implicit cubes so it's easier to verify with something to visualize
        def Cube "local"
        {}
    }

    def Cube "world" {}
}
```

The unit test expects that the final pivot of `mayaXform_100001000000` should be that of `(1.6550425, 3.952048, 1.5850885)` when imported into Maya. This is related to what's causing the original issue being described in this PR, which seems to be at odds at what this unit test expects. Could someone help clarify what exactly the pivot in Maya _should_ be in this case? From a layman's perspective I would expect the pivot to be zeroed-out in this specific case, but I might not fully understand what the constraints are here.

HamedSabri-adsk:
@ysiewappl  Apologies for late reply.  I had a look at !invert!xformOp:translate:rotatePivot and to me it makes sense to end up with identity matrix. I haven't had a chance to fully look at this.  

@ppt-adsk is an expert in pivot calculation. Could you help out with this?

ppt-adsk:
> 
> 
> @ysiewappl Apologies for late reply. I had a look at !invert!xformOp:translate:rotatePivot and to me it makes sense to end up with identity matrix. I haven't had a chance to fully look at this.
> 
> @ppt-adsk is an expert in pivot calculation. Could you help out with this?

The Maya rotate pivot should be (1.6550425, 3.952048, 1.5850885).  See
https://graphics.pixar.com/usd/docs/api/class_usd_geom_xformable.html
for details on paired inverted ops.  Here, the rotate pivot will be applied to no rotation at all, but that's fine: you can still specify its value.

ysiewappl:
Hi @HamedSabri-adsk and @ppt-adsk, thanks for helping take a look at this.

> The Maya rotate pivot should be (1.6550425, 3.952048, 1.5850885). See
> https://graphics.pixar.com/usd/docs/api/class_usd_geom_xformable.html
> for details on paired inverted ops. Here, the rotate pivot will be applied to no rotation at all, but that's fine: you can still specify its value.

I did take a look at the `UsdGeomXformable` documentation as part of working on this; but I'm not sure I understand why it would mean the translation would persist. Referring to the section called **Paired "Inverted" Ops** (emphasis mine):

> The "paired op" behavior is encoded as an "!invert!" prefix in xformOpOrder, as the result of an AddXformOp(isInverseOp=True) call. 
> The xformOpOrder for the last example would look like:
> 
> uniform token[] xformOpOrder = [ "xformOp:translate", "xformOp:rotateXYZ", "xformOp:translate:scalePivot", "xformOp:scale", "!invert!xformOp:translate:scalePivot" ]
> **When asked for its value via UsdGeomXformOp::GetOpTransform(), an "inverted" Op (i.e. the "inverted" half of a set of paired Ops) will fetch the value of its paired attribute and return its negation**. This works for all op types - an error will be issued if a "transform" type op is singular and cannot be inverted. When getting the authored value of an inverted op via UsdGeomXformOp::Get(), the raw, uninverted value of the associated attribute is returned.

This language seems to imply that the `scalePivot` in the above example would end up being back at identity translation.

Thus, for an example of `uniform token[] xformOpOrder = ["xformOp:translate:rotatePivot", "!invert!xformOp:translate:rotatePivot"]`, shouldn't this result in the `rotatePivot`'s translation being zeroed out? If someone could be kind enough to explain further, I'd appreciate it.

I'd be happy to implement an alternative solution for the problem described in the PR, but from what I can tell, the pivot being in the wrong place is directly responsible for causing the issue originally described.

ppt-adsk:
> Thus, for an example of `uniform token[] xformOpOrder = ["xformOp:translate:rotatePivot", "!invert!xformOp:translate:rotatePivot"]`, shouldn't this result in the `rotatePivot`'s translation being zeroed out? If someone could be kind enough to explain further, I'd appreciate it.

Hi @ysiewappl, once multiplied out into the final local transformation, in your example, yes, absolutely, the resulting translation will be zero.  However, both in the USD common transform API and in Maya, we store the rotate pivot, as it is non-null.  The order of application of the transformations for the USD common transform API is described here:
https://graphics.pixar.com/usd/docs/api/class_usd_geom_xform_common_a_p_i.html
And the order of application of the transformations for the Maya transform API is described here:
https://help.autodesk.com/view/MAYAUL/2022/ENU/?guid=Maya_SDK_cpp_ref_class_m_fn_transform_html
Note how in both cases the pivot inverses (in the Maya case, the rotate pivot inverse and the scale pivot inverse) are applied by simply using the inverse of the forward pivot transforms.  Hope this helps; if not, let us know!

lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -146,8 +148,72 @@ static void _setAnimPlugData(
     }
 }
 
+static void _applyAnimPlugData(
```
HamedSabri-adsk, line 151: Please add a brief description for what _applyAnimPlugData()  does.

lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -201,6 +267,61 @@ static void _setMayaAttribute(
...
+                plg.setDouble(zVal[0] * curValue);
+            } else {
+                plg.setDouble(zVal[0] + curValue);
+            }
+            if (zVal.size() > 1 && _isArrayVarying(zVal))
+                _applyAnimPlugData(plg, zVal, timeArray, context);
+        }
+    }
+    return;
+}
```
HamedSabri-adsk, line 323: Your new _applyToMayaAttribute() and _setMayaAttribute() has very similar function signature with minor changes in logics. I would consolidate these two routines and pass a flag to separate the logics.

lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -550,16 +671,18 @@ void UsdMayaTranslatorXformable::Read(
     MFnDagNode MdagNode(mayaNode);
     if (!stackOps.empty()) {
         // make sure stackIndices.size() == xformops.size()
-        for (unsigned int i = 0; i < stackOps.size(); i++) {
+        size_t numStackOps = stackOps.size();
```
HamedSabri-adsk, line 674: Nitpick: `const size_t numStackOps = stackOps.size();`

HamedSabri-adsk (changes requested):
@ysiewappl  Added some minor changes request while trying to read the code and wrapping my head around the logics here.

lib/mayaUsd/fileio/translators/translatorXformable.cpp
HamedSabri-adsk: Certainly for another day and PR, but I had a hard time following the logics with all these `static `and `_ `stuff all over the place. All of these helper functions can go under the namespace.

```
namespace
{
    bool getXformOpAsVec3d(...)
    {
    ...
    }
    void setAnimPlugData(...)
    {
    ...
    }
    void applyAnimPlugData(...)
    {
    ...
    }
    bool isArrayVarying(...)
    {
    ...
    }

    void setMayaAttribute(...)
    {
    ...
    }

    void applyToMayaAttribute(...)
    {
    ...
    }

    bool pushUSDXformOpToMayaXform(...)
    {
    ...
    }
    
    bool isIdentityMatrix(...)
    {
    ...
    }

    bool pushUSDXformToMayaXform(...)
    {
    ...
    }
} // namespace
```

HamedSabri-adsk: unused variable at line 649

https://github.com/Autodesk/maya-usd/blob/f531c371d04e9c4910d36ece17c17558f3d0ff2a/lib/mayaUsd/fileio/translators/translatorXformable.cpp#L649

test/lib/usd/translators/testUsdImportXformOps.py
```diff
@@ -0,0 +1,63 @@
...
+        cmds.mayaUSDImport(file=self.scene_path)
+        selList = om.MSelectionList()
+        selList.add("pCube1")
+        cubeDagPath = om.MDagPath()
+        selList.getDagPath(0, cubeDagPath)
+        transform = cubeDagPath.inclusiveMatrix()
+        expectedTranslation = Gf.Vec3f(0.0, -1.3681414, 0.0)
+        actualTranslation = om.MTransformationMatrix(transform).getTranslation(om.MSpace.kTransform)  # NOTE: (yliangsiew) world space
+        actualTranslation = Gf.Vec3f(actualTranslation[0], actualTranslation[1], actualTranslation[2])
+        print actualTranslation
```
HamedSabri-adsk, line 59: No need to print actualTranslation

HamedSabri-adsk:
@ysiewappl 

I did a round trip using unit_test.zip that you provided without this PR changes and can't reproduce the issue.  Please see my video. 

How are you exporting the scene to USD? Are you passing any specific flags?

https://user-images.githubusercontent.com/48299423/119984282-37a94200-bf8f-11eb-842a-71aa715e3a05.mp4

ysiewappl:
Hi @HamedSabri-adsk:

I just built off `092f52057f46c51c3ed725d9e25422c52e0bdd0c` (May 14th) to re-test (I don't think there is much that differs WRT this issue from this commit VS the one you demonstrated in the video; I can build against latest ToT `dev` later to test if needed). Exporting the USD using the following command:

```
mayaUSDExport -v -sl -f "/Users/yliangsiew/tmp/test_output.usda" -ignoreWarnings true  -skl "auto" -skn "auto"; 
```

And re-importing it with the following command:

```
mayaUSDImport -v -f "/Users/yliangsiew/tmp/test_output.usda";
```

Should present the issue described above. I also attempted exporting using the GUI like you did, with the following options set:

```
file -force -options ";exportUVs=1;exportSkels=auto;exportSkin=auto;exportBlendShapes=0;exportColorSets=1;defaultMeshScheme=catmullClark;defaultUSDFormat=usda;animation=0;eulerFilter=0;staticSingleSample=0;startTime=1;endTime=1;frameStride=1;frameSample=0.0;parentScope=;exportDisplayColor=0;shadingMode=none;exportInstances=1;exportVisibility=1;mergeTransformAndShape=1;stripNamespaces=0" -typ "USD Export" -pr -es "/Users/yliangsiew/tmp/test_output_2.usda";
```

And re-importing with the following options as well:

```
file -import -type "USD Import"  -ignoreVersion -ra true -mergeNamespacesOnClash false -namespace "test_output_2" -options "apiSchema=[];assemblyRep=Collapsed;excludePrimvar=[];importInstances=1;metadata=[hidden,instanceable,kind];preferredMaterial=none;shadingMode=[[useRegistry,UsdPreviewSurface]];useAsAnimationCache=0;readAnimData=0;useCustomFrameRange=0"  -pr  -importFrameRate true  -importTimeRange "override" "/Users/yliangsiew/tmp/test_output_2.usda";
```

This ends up presenting the same issue.

HamedSabri-adsk:
@ysiewappl Apologies for late reply. I tried building your branch again yesterday and was able to reproduce the issue on the round-trip.  Your testUsdImportXformOps.py also fails to pass for me.

I was wondering if you got a chance to see Pierre's comment?
https://github.com/Autodesk/maya-usd/pull/1426#issuecomment-849842579

kxl-adsk:
Looks like more work is required. Closing the PR until someone is ready to pick up the work.

PR #1428, opened 2021-05-19
Use Maya matrix decomposition for matrix op TRS accessors.

ppt-adsk:
Using USD APIs to perform matrix decomposition does not produce results that are the inverse of matrix composition.  Using the Maya APIs does, so change UsdTransform3dMatrixOp to use Maya code (e.g. as does UsdPointInstanceOrientationModifier.cpp).


test/lib/ufe/testMatrices.py
```diff
@@ -161,6 +163,27 @@ def testMatrices(self):
             cylInclMat.matrix, [[1, 0, 0, 0], [0, cx60, sx60, 0],
                                 [0, -sx60, cx60, 0], [0, 10, 0, 1]])
 
+    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() >= 2, 'testTransform3dMatrixOpAccessors only available in UFE v2 or greater.')
+    def testTransform3dMatrixOpAccessors(self):
```
ppt-adsk, line 167: Code written with Test Driven Development (TDD): test written first, test failed, code fix implemented, test passes.

test/testSamples/transform3d/TranslateRotate_vs_xform.usda
```diff
@@ -0,0 +1,98 @@
+#usda 1.0
```
ppt-adsk, line 1: Test file taken from Maya bug report.

PR #1429, opened 2021-05-21
Update baselines for new storm wireframe rendering

dj-mcg:


kxl-adsk (approved):

dj-mcg:
FYI - the Storm change forcing these baseline updated also appears to affect testMtohBasicRender. While image compares for these Pixar tests aren't part of the open source test suite, testMtohBasicRender's image compare is, so that test will require a baseline update once the next USD release is rolled out.

PR #1430, opened 2021-05-21
Replace callsites into deprecated "Master" API

dj-mcg:


lib/mayaUsd/fileio/jobs/readJob.cpp
```diff
@@ -56,7 +56,7 @@
 #include <maya/MStatus.h>
 #include <maya/MTime.h>
 
-#include <ghc/filesystem.hpp>
+#include <boost/filesystem.hpp>
```
HamedSabri-adsk, line 59: @dj-mcg Could you clarify why you needed to make this change here? In fact, This PR should not compile at all on any platforms since we no longer support boost::filesystem.

lib/mayaUsd/fileio/jobs/readJob.cpp
```diff
@@ -486,25 +491,29 @@ bool UsdMaya_ReadJob::_DoImport(UsdPrimRange& rootRange, const UsdPrim& usdRootP
...
+                    while (prototypeNode.childCount()) {
+                        prototypeNode.removeChildAt(prototypeNode.childCount() - 1);
                     }
                 }
-                deleteMasterMod.deleteNode(masterObject);
+                deletePrototypeMod.deleteNode(prototypeObject);
             }
         }
-        deleteMasterMod.doIt();
+        deletePrototypeMod.doIt();
```
HamedSabri-adsk, line 516: I found both masterObject and prototypeObject confusing. Are they not just "instance objects"? I think having a more generic name would be better.

plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/TransformIterator.cpp
```diff
@@ -179,7 +184,7 @@ UsdPrim TransformIterator::parentPrim() const
     UsdPrim parentPrim = m_stage->GetPseudoRoot();
     if (m_primStack.size() > 1) {
         parentPrim = (m_primStack.end() - 2)->m_prim;
-        if (parentPrim.IsMaster()) {
+        if (parentPrim.IsPrototype()) {
```
HamedSabri-adsk, line 187: There is no guard against PXR_VERSION < 2011?

HamedSabri-adsk (changes requested):
Please see my comments.

dj-mcg:
Oh thanks - looks like the USD folks sent me these changes, but they include some that shouldn't be pushed. Let me clean them up a bit and push again.

dj-mcg:
Thank you for catching these, Hamed. I had assumed the author had tested these changes against the open source project but it was only tested internally. I'll make sure to more carefully review the changes that touch mayaUsd before opening a pull request.

lib/mayaUsd/fileio/jobs/readJob.cpp
dj-mcg: This was an oversight on my part. We change these internally since we still use boost, so this didn't show up in the diff I was looking at. I've corrected it

lib/mayaUsd/fileio/jobs/readJob.cpp
dj-mcg: I've made the other changes as it breaks functionality, but I've left this for now since it's a style preference. If you feel strongly about changing the var names, I can raise it with the USD team. Please let me know.

plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/TransformIterator.cpp
dj-mcg: Nice catch, thank you, this was my fault for not testing thoroughly enough before pushing. I'll make sure to start building and testing against both USD's release version and latest version prior to opening a pull request.

HamedSabri-adsk (approved):
Looking good. Thanks.

Regarding the variable name changes, it would be nice to clean them up but I don't think it is urgent.

It would be also nice to have a simple unit-test to back up the new changes.

PR #1432, opened 2021-05-25
MAYA-111760 Invisibile prims don't get notifications on _PropagateDirtyBits or

williamkrick:
Sync, so when a prim turns invisible make sure all reprs can draw correctly.


JGamache-autodesk (approved):
Code is very good. Can you check if the BasisCurves also require this fix?

PR #1433, opened 2021-05-25
MAYA-111897 - Show USD MaterialX in VP2 render delegate

JGamache-autodesk:
- Translate UsdShade network back to MaterialX document using Pixar APIs
- Use Maya contributions to MaterialX shadergen to create VP2 fragment code

Limitiation:
- Texture color space metadata from MaterialX not transmitted thru Hydra (because it is metadata, not data).


lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -684,6 +677,33 @@ HdVP2Material::HdVP2Material(HdVP2RenderDelegate* renderDelegate, const SdfPath&
     : HdMaterial(id)
     , _renderDelegate(renderDelegate)
 {
+#ifdef WANT_MATERIALX_BUILD
```
williamkrick, line 680: This looks like one time initialization. If it is we can wrap it in a std::once to make sure it only happens on time.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1142,16 +1352,32 @@ void HdVP2Material::_UpdateShaderInstance(const HdMaterialNetwork& mat)
...
 
-        // The simplified path has only one token which is the node name.
-        const SdfPath& nodePath = it->second;
-        const MString  nodeName(nodePath != _surfaceShaderId ? nodePath.GetText() : "");
+            // The simplified path has only one token which is the node name.
+            const SdfPath& nodePath = it->second;
+            if (nodePath != _surfaceShaderId) {
+                nodeName = nodePath.GetText();
+            }
+#ifdef WANT_MATERIALX_BUILD
```
williamkrick, line 1378: Rather than have an extra #ifdef just to end the scope you could move the open brace outside the above #ifdef and have that code in a scope.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1237,6 +1473,17 @@ void HdVP2Material::_UpdateShaderInstance(const HdMaterialNetwork& mat)
                 MFloatMatrix matrix;
                 value.UncheckedGet<GfMatrix4f>().Get(matrix.matrix);
                 status = _surfaceShader->setParameter(paramName, matrix);
+            } else if (value.IsHolding<std::string>()) {
```
williamkrick, line 1476: I think I would prefer that the else if be inside the WANT_MATERIALX_BUILD macro. It is just an empty clause now which looks strange to me.

williamkrick (approved):
Overall it looks good, just some minor style changes.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
kxl-adsk: +1

lib/mayaUsd/render/vp2RenderDelegate/material.h
```diff
@@ -123,7 +136,11 @@ class HdVP2Material final : public HdMaterial
     SdfPath              _surfaceShaderId;  //!< Path of the surface shader
     HdVP2TextureMap      _textureMap;       //!< Textures used by this material
     TfTokenVector        _requiredPrimvars; //!< primvars required by this material
-
+#ifdef WANT_MATERIALX_BUILD
+    MaterialX::FileSearchPath _mtlxSearchPath; //!< MaterialX library search path
```
kxl-adsk, line 140: I'm on the fence here...some part of me would like to hide these details inside of a cpp file and not expose them in the header file.

lib/mayaUsd/render/vp2ShaderFragments/CMakeLists.txt
```diff
@@ -24,6 +24,8 @@ list(APPEND SHADERFRAGMENTS_XMLS
     usdPreviewSurfaceCombiner.xml
     usdPreviewSurfaceLightingAPI1.xml
     usdPreviewSurfaceLightingAPI2.xml
+    # MaterialX
+    materialXTw.xml
```
kxl-adsk, line 28: Why not guard this with a macro?

kxl-adsk (approved):
Left few minor comments I would like you to respond to but otherwise looks good!

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
JGamache-autodesk: No. Definitely one-time. Will also move it to the C++ file as requested.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -942,6 +1002,156 @@ void HdVP2Material::_ApplyVP2Fixes(HdMaterialNetwork& outNet, const HdMaterialNe
...
+            }
+        }
+
+        _isMtlx = true;
+
+        // Add automatic tangent generation:
+        shaderInstance->addInputFragment("materialXTw", "Tw", "Tw");
+
+    } catch (mx::Exception e) {
+        TF_RUNTIME_ERROR(
```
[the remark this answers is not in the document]
JGamache-autodesk: The error handling style of the MaterialX library is exception based, so any issues with shader generation will throw a survivable exception, as long as there is a handler active.

lib/mayaUsd/render/vp2RenderDelegate/material.h
JGamache-autodesk: Agreed.

kxl-adsk (approved):
Looks good. Thank you for hiding MaterialX dependency and cleaning up the material header.

williamkrick (approved):

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1097,6 +1469,272 @@ void HdVP2Material::_ApplyVP2Fixes(HdMaterialNetwork& outNet, const HdMaterialNe
     }
 }
 
+#ifdef WANT_MATERIALX_BUILD
+
+void HdVP2Material::_ApplyMtlxVP2Fixes(HdMaterialNetwork2& outNet, const HdMaterialNetwork2& inNet)
```
williamkrick, line 1474: Can we share any of the code in _ApplyMtlxVP2Fixes and _ApplyVP2Fixes? Maybe we could make some helper functions that replace the names or something? If we have to make any changes to that system for re-using shaders we'll have to remember to update both code paths and they are far enough apart now that it seems like we might forget.

williamkrick (approved):

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
JGamache-autodesk: The main issue is that the UsdPreviewSurface code is using the old HdMaterialNetwork, while the MaterialX version uses the more recent HdMaterialNetwork2.
These two functions will have a lot of shareable code once we update everything to the latest material API.

lib/mayaUsd/render/vp2RenderDelegate/render_delegate.h
```diff
@@ -142,6 +146,10 @@ class HdVP2RenderDelegate final : public HdRenderDelegate
 
     MHWRender::MShaderInstance* GetShaderFromCache(const TfToken& id);
```
kxl-adsk, line 147: Since we are now having one global cache:
* why wouldn't we switch these methods to static?
* is render delegate still the right place to access the cache?

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -195,17 +201,148 @@ bool _IsMaterialX(const HdMaterialNode& node)
...
+// Returns true if s ends with suffix
+inline bool _EndsWith(const std::string& s, const std::string& suffix)
+{
+    return s.size() >= suffix.size()
+        && s.compare(s.size() - suffix.size(), std::string::npos, suffix) == 0;
+}
+
+//! Return true if that node parameter has topological impact on the generated code.
+//
+// Swizzle and geompropvalue nodes are currently the only ones that have an attribute that affects
```
kxl-adsk, line 231: Great comment!

kxl-adsk (approved):

PR #1434, opened 2021-05-26
MAYA-111521: Fix error messages showing up when opening the "Applied Schema" category

HamedSabri-adsk:
### Description:
Accessing certain AE tabs (e.g Applied Schema, Cylinder Light ) results in "propName.IsEmpty()" error to spit out by USD. We currently allow creating attributes with an empty strings which are not accepted by USD. This doesn't only happen when accessing "Applied Schema" tabs but also in other attribute tabs "Cylinder Light". 

In USD land At Line 523 ,  the property name can't be empty. Hence the error reported.

https://github.com/PixarAnimationStudios/USD/blob/090ef0d849ced875b3b99a15e454b69148ccd8e1/pxr/usd/usd/stage.cpp#L4523

image

### DIFF
--- a/lib/mayaUsd/ufe/UsdAttributes.cpp
+++ b/lib/mayaUsd/ufe/UsdAttributes.cpp
@@ -66,6 +66,11 @@ Ufe::Attribute::Type UsdAttributes::attributeType(const std::string& name)
 
 Ufe::Attribute::Ptr UsdAttributes::attribute(const std::string& name)
 {
+    // early return if name is empty.
+    if (name.empty()) {
+        return nullptr;
+    }
+
     // If we've already created an attribute for this name, just return it.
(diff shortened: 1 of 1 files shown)


kxl-adsk:
It looks simple and safe, but there is one thing I don't know. @seando-adsk Is it expected to have attributes with empty names? How are they created?

seando-adsk:
No it is not expected to have empty attribute names, but there is no harm in protecting against a bad caller passing in an empty string. That is what happens in this case from some bad code in Maya that is trying to optimize boolean attributes onto a single line.

kxl-adsk:
Had an offline discussion about this. Looks like we have an additional place to fix in Maya, but this fix is still valuable to merge.

kxl-adsk (approved):

PR #1444, opened 2021-05-28
MAYA-110682 - On drag and drop of a layer with an editTarget, the edi…

spinell-adsk:
…tTarget moves to the root layer

Add a new flag (moveSubPath) to the mayaUsdLayerEditor command

When drag and drop a layer, we are currently doing 2 steps.
We are first removing the layer from it's current parent and after we are adding the layer to it's new parent.
If the layer we are moving is the edit target, the layer become not the edit target anymore because the command
for removing a layer will reset the edit target.

So to fix this issue I added a new flag to the  mayaUsdLayerEditor command to move layer into another layer.
I feel this solution it's better instead of doing tricky stuff to keep the edit target unmodified.

### DIFF
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -649,7 +649,12 @@ MSyntax LayerEditorCommand::createSyntax()
     syntax.makeFlagMultiUse(kRemoveSubPathFlag);
     syntax.addFlag(kReplaceSubPathFlag, kReplaceSubPathFlagL, MSyntax::kString, MSyntax::kString);
     syntax.makeFlagMultiUse(kReplaceSubPathFlag);
-    syntax.addFlag(kMoveSubPathFlag, kMoveSubPathFlagL, MSyntax::kString, MSyntax::kString, MSyntax::kUnsigned);
+    syntax.addFlag(
+        kMoveSubPathFlag,
+        kMoveSubPathFlagL,
+        MSyntax::kString,    // path to move
+        MSyntax::kString,    // new parent layer
+        MSyntax::kUnsigned); // layer index inside the new parent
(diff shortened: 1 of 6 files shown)


ppt-adsk (approved):
Looks good to me!  I agree with your assessment that a dedicated move is better than delete + add.  Thanks for the test.

lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -301,6 +304,104 @@ class RemoveSubPath : public InsertRemoveSubPathBase
...
+            newParentLayer->RemoveSubLayerPath(_newIndex);
+            layer->InsertSubLayerPath(_path, _oldIndex);
+        }
+
+        return true;
+    }
+
+    std::string  _path;
+    std::string  _newParentLayer;
+    unsigned int _newIndex;
```
seando-adsk, line 399: Why not make these private and change the constructor to take them as input?

lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -630,6 +737,24 @@ MStatus LayerEditorCommand::parseArgs(const MArgList& argList)
...
+
+            MString newParentLayer;
+            argParser.getFlagArgument(kMoveSubPathFlag, 1, newParentLayer);
+
+            unsigned int index { 0 };
+            argParser.getFlagArgument(kMoveSubPathFlag, 2, index);
+
+            cmd->_path = subPath.asUTF8();
+            cmd->_newParentLayer = newParentLayer.asUTF8();
+            cmd->_newIndex = index;
```
seando-adsk, line 754: Can we move line 741 here and send these as input to constructor?

seando-adsk (changes requested):

seando-adsk (approved):

PR #1455, opened 2021-06-03
MAYA-112039 - As a user, I'd like to have an unshareable stage

alsoussi:
1) ProxyShape Changes
    - Adding New unsharable workflow
       - Main idea is we create a new (unshared) root and put the incoming source as a sublayer, thereby having the ability to sandbox your changes
       - Exposed new attribute for the unshared composition (needed for layer management)
       - There's an unshared SDF layer on the proxy (private member) which stores the data so that the user can toggle on and off without losing their data
       - Since we have the ability to create new layers under this new unshared root, we need to keep track of the "unshared layers" and the source layers. That data (layer list) is stored as metadata on the unshared composition root.
       - Saving on disk follows the same mechanism that it normally would, but as you will see in the save portion we have added some special handling for unshared comp layers (saved in maya file)
       - Also added a bunch of helper functions to Util save/load layer metadata and have a shared token

    - Adding Ability to know if the Stage or Layer is incoming
        - By incoming I mean the stage is coming in from a connection, for example the instage data or the cache Id
        - This is needed since we want to have special handling for these stages. For example incoming stages should NOT be saved and show up differently in the layer edtior
        - Also added a bunch of helper functions to Util to be able to get all the layers recursively so we know if a layer is incoming.

2) Save Changes
    - Moving some of the shared code between saveToUSD and SaveToMayaFile into separate local functions
    - Refactor the getStagesToSave to Actually be getProxiesToSave, other than the naming, Tim had already started this, but what I did here is instead of walking the stages to determine which ones are dirty, I walk the proxies and determine which ones need saving, also I don't walk the proxies again when we do the savetousd or savetomaya, which should be a performance improvement (insignificant probably)
    - We skip saving incoming Layers, here you will see I added the concept of Stub, so I save empty layer data and this is needed so that I can keep track of where the user moves these "incoming layers" in the hierarchy.
    - Unshared comp layers are saved to maya file ALWAYS. Similar to the session layer the unshared comp layers always get saved into the maya file, so I put a special case to not prompt the user for changes when the changes are only in a layer that gets saved to maya file. NOTE: I did NOT fix this for the session layer.

3) Layer Editor Changes
    - Adding the concept of ReadOnly Layers
    - For now the read only layers are the incoming layers in the stage
    - For UI I'm using the same UI as the mute
    - These Readonly layers cannot be moved unless you are moving the parent readonly layer in the tree that contains both readonly and non-readonly layers
    - Exposing new functions on the command hook to ask the proxy if the stage is coming from an incoming source and if its also using the unshared composition workflow


ppt-adsk (approved):
Not quite as familiar as I'd like to be with the workflow and how it connects to the data model and implementation, so my review doesn't have enough depth to stand on its own.  Otherwise looks good to me.

lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -358,28 +375,45 @@ void LayerDatabase::prepareForWriteCheck(bool* retCode, bool isExport)
...
+                MStatus status = shareStagePlug.getValue(sharedStage);
+                if ((status == MS::kSuccess && !sharedStage) || !pShape->isStageIncoming()) {
                     SdfLayerHandleVector allLayers = stage->GetLayerStack(true);
                     for (auto layer : allLayers) {
                         if (layer->IsDirty()) {
-                            _proxiesToSave.append(proxyDagPath);
+                            MDagPath proxyDagPath;
+                            MDagPath::getAPathTo(mobj, proxyDagPath);
+                            // We handle unshared compostion internally in Maya USD file
+                            if ((status == MS::kSuccess && sharedStage) || status != MS::kSuccess) {
```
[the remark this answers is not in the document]
alsoussi: @ppt-adsk Yea that looks a bit confusing... 

Line 405: This predicate is "Should we save the stage?". 

1) shareStagePlug doesn't exist (older maya that doesn't have this) : In this case we case we only care about saving if the input is not an incoming connection, since in that case the node that "owns" the stage (upstream node) is responsible for saving. For example if you have multiple proxy shapes daisy chained one's out_stage feeding the other's in_stage. In that case only one proxy is responsible for saving (the first one). 

2) shareStagePlug exists: Similarly but in this case if the stage is unshared, it means we are responsible for saving the root layer (and stubs for the sublayers so we can put them back in the same spot). So doesn't matter if its incoming or not, we need to save. 

Line 412: This is more "Where should we save the stage?". The reason we have this distinction now is that some Layers are special case layers that should be saved to the maya file only. There are two examples currently of this, the session layer (was there before) and my newly added unshared root layer. So since we have batchSave and a delegate which handles saving externally, we need to manage some proxies ourselves and control where they save.. Ideally we should move all the code in one place but that's a bigger refactor probably. 

I've discussed some of these changes with @fowlertADSK

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -595,15 +685,60 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
...
+                 ++it) {
+                auto layer = LayerManager::findLayer(*it);
+                if (layer) {
+                    updatedReferences.push_back(layer->GetIdentifier());
+                }
+            }
+            setReferencedLayers(_unsharedStageRootLayer, updatedReferences);
+        }
+    }
+#endif
```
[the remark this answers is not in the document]
alsoussi: Yea agree its probably unneeded, I think it was for computeRootLayer, I remember some test failing then I put that... but from the look of it that shouldn't happen.

lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -414,40 +450,142 @@ bool LayerDatabase::saveUsd(bool isExport)
...
+    anonHandle.setBool(isAnon);
+
+    std::string temp;
+    if (!stubOnly && ((exportOnlyIfDitry && layer->IsDirty()) || !exportOnlyIfDitry)) {
+        if (!layer->ExportToString(&temp)) {
+            status = MS::kFailure;
+        }
+    }
+
+    serializedHandle.setString(UsdMayaUtil::convert(temp));
```
seando-adsk, line 501: Should we be setting this when status == MS:kFailure (from line 497). Won't temp be bad?

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -595,15 +685,60 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
...
+        // Once an anon layer is loaded the identifier changes
+        _unsharedStageRootLayer = computeRootLayer(dataBlock, "");
+        if (_unsharedStageRootLayer) {
+            // Anon layers when loaded will have difference identifiers, remap them
+            auto referencedLayers = getReferencedLayers(_unsharedStageRootLayer);
+            std::vector<std::string> updatedReferences;
+            for (VtArray<std::string>::const_iterator it = referencedLayers.cbegin(),
+                                                      end = referencedLayers.cend();
+                 it != end;
+                 ++it) {
```
seando-adsk, line 713: Can you use the newer way: 
for ( const auto& layer : referencedLayers )

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -1529,4 +1794,4 @@ Ufe::Path MayaUsdProxyShapeBase::ufePath() const
 }
 #endif
 
-PXR_NAMESPACE_CLOSE_SCOPE
+PXR_NAMESPACE_CLOSE_SCOPE
```
seando-adsk, line 1797: Did you remove the newline here?

lib/mayaUsd/utils/customLayerData.h
```diff
@@ -0,0 +1,65 @@
...
+namespace CustomLayerData {
+
+// Token for referenced layers
+const std::string kReferencedLayersToken = "maya_shared_layers";
+
+// Token for export file path
+const std::string kExportFilePathToken = "maya_export_file_path";
+
+/*! \brief Get the String Array custom data on the layer
+ */
```
seando-adsk, line 36: I think you add more doc regarding the input params and the return value.

lib/mayaUsd/utils/util.h
```diff
@@ -656,6 +656,29 @@ MString GetCurrentMayaWorkspacePath();
 MAYAUSD_CORE_PUBLIC
 MString GetCurrentSceneFilePath();
 
+/**
+ * Returns all the sublayers recursively for a given layer
+ *
+ * @param layer   The layer to start from and get the sublayers
```
seando-adsk, line 662: I noticed you are mixing the kinds of doxygen tags you are using. Here you use @ and above you use \

lib/mayaUsd/utils/utilFileSystem.h
```diff
@@ -75,6 +75,17 @@ std::string resolveRelativePathWithinMayaContext(
 MAYAUSD_CORE_PUBLIC
 bool pathAppendPath(std::string& a, const std::string& b);
 
+/**
+ * Appends `b` to the path `a` and returns directory separators as necessary.
+ *
+ * @param a         A string that respresents the first path
+ * @param b         A string that respresents the second path
+ *
+ * @return         the two paths joinied by a seperator
```
seando-adsk, line 84: Typo "joined"

lib/mayaUsd/utils/utilFileSystem.h
```diff
@@ -75,6 +75,17 @@ std::string resolveRelativePathWithinMayaContext(
 MAYAUSD_CORE_PUBLIC
 bool pathAppendPath(std::string& a, const std::string& b);
 
+/**
+ * Appends `b` to the path `a` and returns directory separators as necessary.
```
seando-adsk, line 79: This doesn't really "return directory separators" - it returns a path (by appending two input paths).

lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -203,16 +235,41 @@ bool LayerTreeItem::appearsMuted() const
...
+    return false;
+}
+
+bool LayerTreeItem::isReadOnly() const
+{
+    if (_isSharedLayer || !_layer->PermissionToEdit()) {
+        return true;
+    }
+
+    return false;
```
seando-adsk, line 257: Can't this just be:
return (_isSharedLayer || !_layer->PermissionToEdit());

lib/usd/ui/layerEditor/layerTreeItem.h
```diff
@@ -96,12 +99,16 @@ class LayerTreeItem : public QStandardItem
     bool isMuted() const;
     // check if this layer is muted, or any of its parent
     bool appearsMuted() const;
+    // check if this layer is readonly
+    bool isReadOnly() const;
     // true if dirty, but look at needsSaving for UI feedback
-    bool isDirty() const { return _layer ? _layer->IsDirty() : false; }
+    bool isDirty() const { return _layer ? _layer->IsDirty() && !isReadOnly() : false; }
```
seando-adsk, line 105: You've modified the behavior of this method. It is no longer "isDirty" Now it is "isDirty and not ReadOnly". Why? I think the code should have to call both these methods (isDirty/isReadOnly) if that is the answer they want.

lib/usd/ui/layerEditor/mayaCommandHook.cpp
```diff
@@ -205,4 +205,41 @@ ufeSelectCmd.replaceWith(sn)
...
+    }
+    std::string cmd = "getAttr " + GetProxyShapeName(proxyShapePath) + "." + attributeName;
+    int         result = 0;
+    auto        status = MGlobal::executeCommand(MString(cmd.c_str()), result, false, false);
+    if (status != MS::kSuccess || result == 0) {
+        return false;
+    }
+
+    return true;
+}
```
seando-adsk, line 233: These two functions should be in the unnamed namespace above and be lowerCamelCase.

lib/usd/ui/layerEditor/mayaCommandHook.h
```diff
@@ -69,6 +69,14 @@ class MayaCommandHook : public AbstractCommandHook
     // this method is used to select the prims with spec in a layer
     void selectPrimsWithSpec(UsdLayer usdLayer) override;
 
+    // this method is used to check if the stage in the proxy shape is from
+    // an incoming connection (using instage data or cache id for example)
+    bool isProxyShapeStageIncoming(std::string proxyShapePath) override;
```
seando-adsk, line 74: const std::string&

lib/usd/ui/layerEditor/stringResources.h
```diff
@@ -47,61 +47,61 @@ MStringResourceId create(const char* key, const char* value);
...
 const auto kSaveAnonymousConfirmOverwrite { create("kSaveAnonymousConfirmOverwrite", "^1s file(s) already exist and will be overwritten.  Do you want to continue?")};
 
-const auto kSaveLayerUsdFileFormatAnn{ create("kSaveLayerUsdFileFormatAnn", "Select whether the .usd file is written out in binary or ASCII. You can save a file in .usdc (binary) or .usda (ASCII) format. Manually entering a file name with an extension overrides the selection in this drop-down menu.") };
-const auto kSaveLayerUsdFileFormatSbm{ create("kSaveLayerUsdFileFormatSbm", "Select whether the .usd file is written out in binary or ASCII") };
+const auto kSaveLayerUsdFileFormatAnn    { create("kSaveLayerUsdFileFormatAnn", "Select whether the .usd file is written out in binary or ASCII. You can save a file in .usdc (binary) or .usda (ASCII) format. Manually entering a file name with an extension overrides the selection in this drop-down menu.") };
+const auto kSaveLayerUsdFileFormatSbm    { create("kSaveLayerUsdFileFormatSbm", "Select whether the .usd file is written out in binary or ASCII") };
```
seando-adsk, line 105: Did you actually change anything in this file? It looks like all whitespace changes to me. If so was it you or because of clang-format?

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -585,21 +586,41 @@ global proc mayaUsdMenu_layerEditorContextMenu(string $panelName) {
     $enabled = 1;
     menuItem -label $label -enable $enabled -c $cmd;
 
+    if($isIncoming) {
+        $incoming = `listConnections -s true -d false -c true $proxyShape`;
+        $incomingNode = "";
```
seando-adsk, line 591:
```suggestion
        string $incomingNode = "";
```

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -585,21 +586,41 @@ global proc mayaUsdMenu_layerEditorContextMenu(string $panelName) {
...
+        for($i=0;$i<size($incoming);$i+=2) {
+            $j = $i+1;
+            if($incoming[$i] == $proxyShape+".inStageData") {
+                $incomingNode = $incoming[$j];
+            }
+            else if($incomingNode == "" && $incoming[$i] ==  $proxyShape+".stageCacheId") {
+                $incomingNode = $incoming[$j];
+            }
```
seando-adsk, line 599: Can you add a comment as to what this code is doing?

seando-adsk (changes requested):
Overall very good job. Just a few comments and suggestions.

lib/mayaUsd/nodes/layerManager.cpp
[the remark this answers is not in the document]
ppt-adsk: O.K., sounds good, it would be great if you could convert a condensed version of your explanation above into a code comment.

kxl-adsk (commented):
Still have ProxyShape changes to review, but posting everything else that I have reviewed already.

lib/mayaUsd/nodes/layerManager.cpp
alsoussi: Yes, but this is the old behavior, I didn't want to change it.. But I agree with you

alsoussi: Let me know if I should change it

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -539,29 +584,74 @@ SdfLayerRefPtr MayaUsdProxyShapeBase::computeSessionLayer(MDataBlock&) { return
 
 #endif
 
-MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
+namespace {
+
+VtArray<std::string> getReferencedLayers(const SdfLayerRefPtr& layer)
```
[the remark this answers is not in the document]
alsoussi: Yea  I tried to clean up a little more.

lib/mayaUsd/utils/util.h
alsoussi: Honestly I try to follow the style of the file I'm in

lib/usd/ui/layerEditor/layerTreeItem.h
alsoussi: Yea your right!

lib/usd/ui/layerEditor/stringResources.h
alsoussi: Oh, I had some changes but they were removed, its only clang format I guess

lib/usd/ui/layerEditor/mayaCommandHook.cpp
```diff
@@ -205,4 +205,41 @@ ufeSelectCmd.replaceWith(sn)
...
+
+bool GetBooleanAttributeOnProxyShape(
+    const std::string& proxyShapePath,
+    const std::string& attributeName)
+{
+    if (proxyShapePath.empty()) {
+        return false;
+    }
+    std::string cmd = "getAttr " + GetProxyShapeName(proxyShapePath) + "." + attributeName;
```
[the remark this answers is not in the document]
alsoussi: I mean this whole file is doing that, just following convention.

lib/mayaUsd/utils/util.cpp
```diff
@@ -2434,3 +2434,42 @@ MString UsdMayaUtil::GetCurrentSceneFilePath()
...
+{
+    std::set<std::string>      allSublayers;
+    std::deque<SdfLayerRefPtr> processing;
+    processing.push_back(layer);
+    while (!processing.empty()) {
+        auto layerToProcess = processing.front();
+        processing.pop_front();
+        SdfSubLayerProxy sublayerPaths = layerToProcess->GetSubLayerPaths();
+        for (auto path : sublayerPaths) {
+            allSublayers.insert(path);
```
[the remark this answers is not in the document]
alsoussi: Good point, I shouldn't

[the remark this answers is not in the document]
alsoussi: Ah I think it was to maintain the list that was provided even if they are not loaded, but I think maybe I should load them. I agree the behavior is undefined in that case

plugin/adsk/scripts/mayaUsdMenu.mel
seando-adsk: General question: I saw that you resolved this conversation, but I don't see any new comment. I would prefer if the conversations were left open and resolved by the person who created them.

alsoussi: The resolution means I addressed in my code, it would have been marked automatically anyway because the code is outdated. For me its a way to keep track of what I resolved or not

alsoussi: I haven't pushed my code since I didnt want to push many small changes, I wanted to address all the comments in one change.

lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -414,40 +450,142 @@ bool LayerDatabase::saveUsd(bool isExport)
...
+
+    return status;
+}
+
+MStatus setValueForAttr(const MObject& node, const MObject& attribute, const std::string& value)
```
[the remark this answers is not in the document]
alsoussi: No, its using the MObject not the attribute name. But nice one! Would have used it!

test/lib/mayaUsd/nodes/testProxyShapeBase.py
```diff
@@ -179,5 +179,93 @@ def testDuplicateProxyStageFileBacked(self):
         self.assertEqual(0, len(ufe.Hierarchy.hierarchy(duplProxyShapeItem).children()))
         self.assertEqual(0, len(ufe.Hierarchy.hierarchy(treebaseItem).children()))
 
+    def testShareStage(self):
```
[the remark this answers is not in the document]
alsoussi: Added more tests!

alsoussi:
@ppt-adsk @seando-adsk @kxl-adsk I updated with regards to your comments, apologies for resolving them on my side, it was my way of keeping track of them but moving forward I will let the person who puts the comment resolve it. 

Let me know if you have any questions.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -539,29 +584,56 @@ SdfLayerRefPtr MayaUsdProxyShapeBase::computeSessionLayer(MDataBlock&) { return
...
-    // If inData has an incoming connection, then use it. Otherwise generate stage from the filepath
-    if (!inDataHandle.data().isNull()) {
-        //
-        // Propagate inData -> inDataCached
-        //
-        MDataHandle inDataCachedHandle = dataBlock.outputValue(inStageDataCachedAttr, &retValue);
-        CHECK_MSTATUS_AND_RETURN_IT(retValue);
+    if (!remappedLayers.empty()) {
+        for (auto sublayer : sublayers) {
+            remapSublayerRecursive(sublayer, remappedLayers);
```
kxl-adsk, line 621: During iteration, you can discover that all layers were remapped. So you can save both searching for layers, vector insert and calls to remapSublayerRecursive if you do something like:
```
if (remappedLayers.empty())
    return;
for(const auto& sublayerPath : sublayerPaths) {
    SdfLayerRefPtr subLayer = SdfLayer::Find(sublayerPath);
    remapSublayerRecursive(sublayer, remappedLayers);
    if (remappedLayers.empty())
        return;
}
```

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -705,32 +833,114 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
...
+        // When we switch out of unshared we need to save this data so when the user toggles back
+        // they get the same state they were in, but in order to this we have to keep the layers in
+        // the heirharchy alive since the stage is gone and so they will get removed
+        if (_unsharedStageRootLayer) {
+            auto subLayerIds = UsdMayaUtil::getAllSublayers(_unsharedStageRootLayer);
+            _unsharedStageRootSublayers.clear();
+            for (const auto& identifier : subLayerIds) {
+                SdfLayerRefPtr sublayer = SdfLayer::Find(identifier);
+                if (sublayer) {
+                    _unsharedStageRootSublayers.push_back(sublayer);
```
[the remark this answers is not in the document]
alsoussi: Its simpler to keep pointers to them, putting them in the cache I will have to keep track of the id and remove them once i'm done with them or we leak (grow the cache).

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -705,32 +833,114 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
...
+    if (isIncomingStage) {
+        std::vector<std::string> incomingLayers { usdStage->GetRootLayer()->GetIdentifier() };
+        _incomingLayers = UsdMayaUtil::getAllSublayers(incomingLayers, true);
+    } else {
+        _incomingLayers.clear();
+    }
+
+    // Share the stage
+    if (sharableStage) {
+        if (isIncomingStage) {
```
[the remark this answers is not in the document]
alsoussi: Done!

lib/mayaUsd/nodes/proxyShapeBase.h
```diff
@@ -126,6 +129,8 @@ class MayaUsdProxyShapeBase
     static MObject outStageDataAttr;
     MAYAUSD_CORE_PUBLIC
     static MObject outStageCacheIdAttr;
+    MAYAUSD_CORE_PUBLIC
+    static MObject outIsStageIncomingAttr;
```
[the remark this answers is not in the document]
alsoussi: OK no worries I removed it. I had it for the test, but now I also removed those tests.

lib/mayaUsd/nodes/proxyShapeBase.h
```diff
@@ -358,6 +373,19 @@ class MayaUsdProxyShapeBase
...
+
+    // We need to keep track of unshared sublayers (otherwise they get removed)
+    std::vector<SdfLayerRefPtr> _unsharedStageRootSublayers;
+
+    // Keep track of the incoming layers
+    std::set<std::string> _incomingLayers;
+
+    // Keep track of source stage's root identifier (when we change the source we can swap it
+    // in-place in the layer heirarchy)
+    std::string _sourceRootIdentifier;
```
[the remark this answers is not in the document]
alsoussi: This is because I need to keep track of the source. On the proxy shape the source can be one of 4 things:

1. Stage Data In port
2. Stage Cache ID In Port
3. File Path (Valid)
4. New Anon Stage (if all else fails)

So when you are in unshared mode that source above is greyed out but you can still move it in the hierarchy. Meaning you can have

UnsharedRootLayer
|---> Some User Created Middle Layer
   |---> The source (referenced above, buried in a complex hierarchy)

So now when we change the file or change the source we want to maintain that hierarchy, so we need to keep track of what the source was and what it will be. 

After our discussion yesterday I was trying to access the old stage from the out port, but its hard for me to figure out which layer is which, the code gets really complicated if you don't want to keep track of the source layer. Simplest thing is to keep it as is.

lib/mayaUsd/nodes/proxyShapeBase.cpp
alsoussi: Done

lib/mayaUsd/nodes/proxyShapeBase.h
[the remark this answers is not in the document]
alsoussi: Done

alsoussi:
@kxl-adsk done!

kxl-adsk (approved):
Thank you. LGTM.

lib/usd/ui/layerEditor/mayaCommandHook.cpp
```diff
@@ -34,6 +36,9 @@
 #define STR(x) std::string(x)
 
 namespace {
+
+PXR_NAMESPACE_USING_DIRECTIVE;
```
seando-adsk, line 40: Why did you add this? According to our coding standards we should use the fully qualified name. So for USD code that would be "PXR_NS::xxx"

test/lib/CMakeLists.txt
```diff
@@ -37,6 +42,30 @@ foreach(script ${TEST_SCRIPT_FILES})
...
+            "LD_LIBRARY_PATH=${ADDITIONAL_LD_LIBRARY_PATH}"
+            "MAYA_LIGHTAPI_VERSION=${MAYA_LIGHTAPI_VERSION}"
+
+            # Maya uses a very old version of GLEW, so we need support for
+            # pre-loading a newer version from elsewhere.
+            "LD_PRELOAD=${ADDITIONAL_LD_PRELOAD}"
+
+            # Fallback to old color management. We will have to investigate
+            # and introduce OCIOv2 compatible version of these tests.
+            "MAYA_COLOR_MANAGEMENT_SYNCOLOR=1"
```
seando-adsk, line 62: Do you need any of this? I think not for your simple test.

seando-adsk (changes requested):
Overall looks good - just a couple of minor changes.

lib/usd/ui/layerEditor/mayaCommandHook.cpp
alsoussi: Mostly because trhe UsdMayaUtil is under the PXR_NS, which is weird to me, so if someone fixes that then its one spot to change. But anyhow I put PXR_NS:: before my UsdMayaUtil calls so its not an issue. There are two of them.

lib/mayaUsd/utils/customLayerData.h
```diff
@@ -32,26 +32,48 @@ const std::string kReferencedLayersToken = "maya_shared_layers";
...
  */
 MAYAUSD_CORE_PUBLIC
 void setStringArray(
     const PXR_NS::VtArray<std::string>& data,
     const PXR_NS::SdfLayerRefPtr&       layer,
     const std::string&                  token);
 
-/*! \brief Get the String custom data on the layer
+/**
+ * Get the String custom data on the laye
```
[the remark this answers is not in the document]
alsoussi: Done, nice catch!

lib/usd/ui/layerEditor/mayaCommandHook.cpp
```diff
@@ -54,7 +59,40 @@ MString executeMel(const std::string& commandString)
 // maya doesn't support spaces in undo chunk names...
 MString cleanChunkName(QString name) { return quote(name.replace(" ", "_").toStdString()).c_str(); }
 
+std::string GetProxyShapeName(const std::string& proxyShapePath)
```
seando-adsk, line 62: This function and the one below should be "lowerCamelCase" so "get" not "Get"

test/lib/CMakeLists.txt
alsoussi: Done

alsoussi:
@seando-adsk Done!

lib/usd/ui/layerEditor/mayaCommandHook.cpp
alsoussi: Done!

seando-adsk (approved):

PR #1456, opened 2021-06-03
Refactoring private tokens to reduce proliferation

JGamache-autodesk:


JGamache-autodesk:
No functional change.
Since there are a lot of tokens shared between import<->export we end up with duplication.
This will get worse as I introduce MaterialX to the mix, so let's centralize the tokens.

JGamache-autodesk:
The preflight failure is known: testGroupCmd everything else runs fine.

JGamache-autodesk:
Closing. This has been integrated in PR #1478.

PR #1457, opened 2021-06-03
MAYA-111835: fix regression test group command

HamedSabri-adsk:


kxl-adsk:
We went with revert (#1459) and will investigate the issue more. Closing this PR.

PR #1458, opened 2021-06-03
Clean up logics in Group command.

HamedSabri-adsk:


lib/mayaUsd/ufe/UsdUndoCreateGroupCommand.cpp
```diff
@@ -81,10 +81,10 @@ void UsdUndoCreateGroupCommand::execute()
 
     auto newParentHierarchy = Ufe::Hierarchy::hierarchy(_group);
     if (newParentHierarchy) {
-        for (auto child : _selection) {
+        for (const auto& child : _selection) {
             auto parentCmd = newParentHierarchy->appendChildCmd(child);
-            parentCmd->execute();
             append(parentCmd);
+            parentCmd->execute();
```
HamedSabri-adsk, line 87: @ppt-adsk this logics seems a bit awkward.  I believe the order matters here. We want to append the commands to the CmdList before running the execute().

lib/mayaUsd/ufe/UsdUndoCreateGroupCommand.cpp
```diff
@@ -81,10 +81,10 @@ void UsdUndoCreateGroupCommand::execute()
 
     auto newParentHierarchy = Ufe::Hierarchy::hierarchy(_group);
     if (newParentHierarchy) {
-        for (auto child : _selection) {
+        for (const auto& child : _selection) {
```
HamedSabri-adsk, line 84: avoid the unnecessary copy.

HamedSabri-adsk:
@ppt-adsk I am closing this since I believe there are some work that needs to be done here. 

The insertion part in UsdUndoCreateGroupCommand::execute() is not really a composition.  In every loop we are executing the parent command and then append it to CompositeUndoableCommand::CmdList which I understand is used for the purposes of undo/redo.

It would make more sense to me to create the parent command once, then append it command list and execute it after the loop.

 ```
Ufe::InsertChildCommand::Ptr parentCmd{nullptr};
 for (const auto& child : _selection) {
      parentCmd = newParentHierarchy->appendChildCmd(child);
      append(parentCmd);
 }
```

PR #1468, opened 2021-06-07
Handling Custom Attributes, and Hierarchy on Import/Export

BSalem:
…ser Attributes And the Hierarchyduring the I/O operations, the aim of these features is to simplify and reduce the amount of work for the migration from Alembic to USD pipeline.

Features:
---------
* Added -root/-rt multi-flag that copies the behavior of AbcExport which limits the roots under which the primitives will be written in the USD rather than translating all the parents up till scene root.

Enhancements:
-------------

* -root/-rt (multi-flag) and -sl flag together gives full control on which part of the hierarchy to be written to the USD and what exactly under the given roots will be exported, for example:
if we have the following hierarchy:

Assets
|---- chr_model_ast
|----|---- body
|----|---- aux_nodes
|----chr_garment_ast
|----|---- cloth
|----|---- accessories
|----|---- garment_curves_grp
|----|---- sim_garment

select -r body cloth accessories;
usdExport -root "chr_model_ast" -root "chr_garment_ast" -userattr "userProperties:ast" -userattr "userProperties:version" -sl -file "/path/to/usd.usda";

this will write a USD that have the following hierarchy (with the custom attributes exported):

/
chr_model_ast
|--- body
chr_garment_ast
|--- cloth
|--- accessories

### DIFF
--- a/lib/mayaUsd/commands/baseExportCommand.cpp
+++ b/lib/mayaUsd/commands/baseExportCommand.cpp
@@ -180,6 +180,12 @@ MSyntax MayaUSDExportCommand::createSyntax()
     syntax.addFlag(kFileFlag, kFileFlagLong, MSyntax::kString);
     syntax.addFlag(kSelectionFlag, kSelectionFlagLong, MSyntax::kNoArg);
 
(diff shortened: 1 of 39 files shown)


BSalem:
having some issues with git/diffs

PR #1469, opened 2021-06-07
MAYA-111943: Add support for ungroup command.

HamedSabri-adsk:
Ungroup_Command

This PR adds support for ungroup command.  See ungroup page:

https://download.autodesk.com/us/maya/2009help/Commands/ungroup.html

Known Issues:
- parent flag is not currently implemented. This shouldn't be a blocker since one can use the existing parent command. I will add this flag later if it is really requested. 

- Children are not selected after ungrouping.  **MAYA-112126**

- There is a bug in our selection where hierarchy item created from group item falsely report that children don't exist. See `testUngroupAfterUndoRedo` case in `testUngroup.py`  **MAYA-112510**

### DIFF
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
@@ -248,12 +248,6 @@ ProxyShapeHierarchy::reorderCmd(const Ufe::SceneItemList& orderedList) const
     return UsdUndoReorderCommand::create(getUsdRootPrim(), orderedTokens);
 }
 
-Ufe::UndoableCommand::Ptr ProxyShapeHierarchy::ungroupCmd() const
-{
-    // pseudo root can not be ungrouped.
-    return nullptr;
-}
-
 Ufe::SceneItem::Ptr ProxyShapeHierarchy::defaultParent() const
 {
     // Maya shape nodes cannot be unparented.
@@ -261,6 +255,14 @@ Ufe::SceneItem::Ptr ProxyShapeHierarchy::defaultParent() const
 }
 
 #endif // UFE_V2_FEATURES_AVAILABLE
+
+#ifdef UFE_V3_FEATURES_AVAILABLE
+Ufe::UndoableCommand::Ptr ProxyShapeHierarchy::ungroupCmd() const
+{
+    // pseudo root can not be ungrouped.
+    return nullptr;
+}
+#endif // UFE_V3_FEATURES_AVAILABLE
 
 } // namespace ufe
 } // namespace MAYAUSD_NS_DEF
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
@@ -32,7 +32,6 @@
 #include <mayaUsd/ufe/UsdUndoCreateGroupCommand.h>
 #include <mayaUsd/ufe/UsdUndoInsertChildCommand.h>
 #include <mayaUsd/ufe/UsdUndoReorderCommand.h>
-#include <mayaUsd/ufe/UsdUndoUngroupCommand.h>
 #endif
 
 PXR_NAMESPACE_USING_DIRECTIVE
@@ -251,9 +250,8 @@ ProxyShapeHierarchy::reorderCmd(const Ufe::SceneItemList& orderedList) const
 
 Ufe::UndoableCommand::Ptr ProxyShapeHierarchy::ungroupCmd() const
 {
-    auto usdItem = UsdSceneItem::create(sceneItem()->path(), getUsdRootPrim());
-
-    return UsdUndoUngroupCommand::create(usdItem);
+    // pseudo root can not be ungrouped.
+    return nullptr;
 }
 
 Ufe::SceneItem::Ptr ProxyShapeHierarchy::defaultParent() const
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
@@ -32,6 +32,7 @@
 #include <mayaUsd/ufe/UsdUndoCreateGroupCommand.h>
 #include <mayaUsd/ufe/UsdUndoInsertChildCommand.h>
 #include <mayaUsd/ufe/UsdUndoReorderCommand.h>
+#include <mayaUsd/ufe/UsdUndoUngroupCommand.h>
 #endif
 
 PXR_NAMESPACE_USING_DIRECTIVE
@@ -246,6 +247,13 @@ ProxyShapeHierarchy::reorderCmd(const Ufe::SceneItemList& orderedList) const
 
     // create a reorder command and pass in the parent and its ordered children list
     return UsdUndoReorderCommand::create(getUsdRootPrim(), orderedTokens);
+}
+
+Ufe::UndoableCommand::Ptr ProxyShapeHierarchy::ungroupCmd() const
+{
+    auto usdItem = UsdSceneItem::create(sceneItem()->path(), getUsdRootPrim());
+
+    return UsdUndoUngroupCommand::create(usdItem);
 }
 
 Ufe::SceneItem::Ptr ProxyShapeHierarchy::defaultParent() const
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.h
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.h
@@ -77,6 +77,8 @@ public:
     insertChildCmd(const Ufe::SceneItem::Ptr& child, const Ufe::SceneItem::Ptr& pos) override;
 
     Ufe::UndoableCommand::Ptr reorderCmd(const Ufe::SceneItemList& orderedList) const override;
+
+    Ufe::UndoableCommand::Ptr ungroupCmd() const override;
 #endif
 
 private:
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.h
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.h
@@ -77,7 +77,9 @@ public:
     insertChildCmd(const Ufe::SceneItem::Ptr& child, const Ufe::SceneItem::Ptr& pos) override;
 
     Ufe::UndoableCommand::Ptr reorderCmd(const Ufe::SceneItemList& orderedList) const override;
+#endif
 
+#ifdef UFE_V3_FEATURES_AVAILABLE
     Ufe::UndoableCommand::Ptr ungroupCmd() const override;
 #endif
 
--- a/lib/mayaUsd/ufe/UsdHierarchy.cpp
+++ b/lib/mayaUsd/ufe/UsdHierarchy.cpp
@@ -42,8 +42,10 @@
 #include <mayaUsd/ufe/UsdUndoCreateGroupCommand.h>
 #include <mayaUsd/ufe/UsdUndoInsertChildCommand.h>
 #include <mayaUsd/ufe/UsdUndoReorderCommand.h>
-#include <mayaUsd/ufe/UsdUndoUnGroupCommand.h>
+#endif
 
+#ifdef UFE_V3_FEATURES_AVAILABLE
+#include <mayaUsd/ufe/UsdUndoUnGroupCommand.h>
 #endif
 
 PXR_NAMESPACE_USING_DIRECTIVE
@@ -262,13 +264,14 @@ Ufe::UndoableCommand::Ptr UsdHierarchy::reorderCmd(const Ufe::SceneItemList& ord
     // create a reorder command and pass in the parent and its reordered children list
     return UsdUndoReorderCommand::create(downcast(sceneItem())->prim(), orderedTokens);
 }
+#endif // UFE_V2_FEATURES_AVAILABLE
 
+#ifdef UFE_V3_FEATURES_AVAILABLE
 Ufe::UndoableCommand::Ptr UsdHierarchy::ungroupCmd() const
 {
     return UsdUndoUngroupCommand::create(fItem);
 }
-
-#endif // UFE_V2_FEATURES_AVAILABLE
+#endif // UFE_V3_FEATURES_AVAILABLE
 
 } // namespace ufe
 } // namespace MAYAUSD_NS_DEF
--- a/lib/mayaUsd/ufe/UsdHierarchy.cpp
+++ b/lib/mayaUsd/ufe/UsdHierarchy.cpp
@@ -42,6 +42,8 @@
 #include <mayaUsd/ufe/UsdUndoCreateGroupCommand.h>
 #include <mayaUsd/ufe/UsdUndoInsertChildCommand.h>
 #include <mayaUsd/ufe/UsdUndoReorderCommand.h>
+#include <mayaUsd/ufe/UsdUndoUnGroupCommand.h>
+
 #endif
 
 PXR_NAMESPACE_USING_DIRECTIVE
@@ -259,6 +261,11 @@ Ufe::UndoableCommand::Ptr UsdHierarchy::reorderCmd(const Ufe::SceneItemList& ord
 
     // create a reorder command and pass in the parent and its reordered children list
     return UsdUndoReorderCommand::create(downcast(sceneItem())->prim(), orderedTokens);
+}
+
+Ufe::UndoableCommand::Ptr UsdHierarchy::ungroupCmd() const
+{
+    return UsdUndoUngroupCommand::create(fItem);
 }
 
 #endif // UFE_V2_FEATURES_AVAILABLE
--- a/lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
@@ -19,9 +19,7 @@
 
 #include <mayaUsd/ufe/Utils.h>
 #include <mayaUsd/undo/UsdUndoBlock.h>
-#include <mayaUsdUtils/util.h>
 
-#include <pxr/usd/sdf/changeBlock.h>
 #include <pxr/usd/usd/editContext.h>
 
 #include <ufe/globalSelection.h>
@@ -37,7 +35,6 @@ namespace ufe {
 UsdUndoUngroupCommand::UsdUndoUngroupCommand(const UsdSceneItem::Ptr& groupItem)
     : Ufe::UndoableCommand()
     , _groupItem(groupItem)
-    , _compositeInsertCmd(std::make_shared<Ufe::CompositeUndoableCommand>())
 {
 }
 
@@ -53,15 +50,9 @@ void UsdUndoUngroupCommand::execute()
     // "Ungrouping" means moving group's children up a level in hierarchy
     // followed by group node getting removed.
 
-    // move group's children one level up
-    auto groupHier = Ufe::Hierarchy::hierarchy(_groupItem);
-    auto levelUpHier = Ufe::Hierarchy::hierarchy(groupHier->parent());
-    for (auto& child : groupHier->children()) {
-        auto insertChildCmd = levelUpHier->appendChildCmd(child);
-        _compositeInsertCmd->append(insertChildCmd);
-    }
-
-    _compositeInsertCmd->execute();
+    // Handling insertion (a.k.a move ) is best to be done on Maya side to cover
+    // all possible flags ( absolute, relative, world, parent ).
+    // For now, the prim removal must still happen in the plugin side. HS, June17, 2021
 
     // remove group prim
     MayaUsd::ufe::InAddOrDeleteOperation ad;
@@ -78,16 +69,9 @@ void UsdUndoUngroupCommand::undo()
     MayaUsd::ufe::InAddOrDeleteOperation ad;
 
     _undoableItem.undo();
-
-    _compositeInsertCmd->undo();
 }
 
-void UsdUndoUngroupCommand::redo()
-{
-    _compositeInsertCmd->redo();
-
-    _undoableItem.redo();
-}
+void UsdUndoUngroupCommand::redo() { _undoableItem.redo(); }
 
 } // namespace ufe
 } // namespace MAYAUSD_NS_DEF
--- a/lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
@@ -1,0 +1,128 @@
+//
+// Copyright 2021 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "UsdUndoUngroupCommand.h"
+
+#include "private/UfeNotifGuard.h"
+
+#include <mayaUsd/ufe/Utils.h>
+#include <mayaUsd/undo/UsdUndoBlock.h>
+#include <mayaUsdUtils/util.h>
+
+#include <pxr/usd/usd/editContext.h>
+
+#include <ufe/globalSelection.h>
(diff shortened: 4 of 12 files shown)


lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -261,6 +263,11 @@ Ufe::UndoableCommand::Ptr UsdHierarchy::reorderCmd(const Ufe::SceneItemList& ord
     return UsdUndoReorderCommand::create(downcast(sceneItem())->prim(), orderedTokens);
 }
 
+Ufe::UndoableCommand::Ptr UsdHierarchy::ungroupCmd() const
+{
+    return UsdUndoUngroupCommand::create(fItem);
+}
+
```
HamedSabri-adsk, line 270: Pass in fItem ( a.k.a group node ) to UsdUndoUngroupCommand

lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
```diff
@@ -0,0 +1,128 @@
...
+        auto insertChildCmd = levelUpHier->appendChildCmd(child);
+        _compositeInsertCmd->append(insertChildCmd);
+    }
+
+    _compositeInsertCmd->execute();
+
+    // remove group prim
+    auto status = removePrimGroup();
+    TF_VERIFY(status, "Removing group node failed!");
+}
```
HamedSabri-adsk, line 68: Ungrouping is a two-tier operation which involves:

1- moving group's children up a level in hierarchy

line 56-63 does the insertion part.  Insertion part is done via composition ( _compositeInsertCmd )

2- removing the group node itself.

This part is a bit tricky since we really don't have a remove command and USD is not so friendly with such operation. However with a little bit of trickery we can workaround it. 

First, we have to make sure the undoBlock captures the _groupItem->prim() removal. After this _groupItem->prim() is stale and should not be accessed anywhere afterwards.

```
    MayaUsd::ufe::InAddOrDeleteOperation ad;
    UsdUndoBlock                         undoBlock(&_undoableItem);

    auto           stage = _groupItem->prim().GetStage();
    UsdEditContext ctx(stage, stage->GetEditTarget().GetLayer());
    return stage->RemovePrim(_groupItem->prim().GetPath());
```

lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
```diff
@@ -0,0 +1,128 @@
...
+    // Make sure to add the newly created _group (a.k.a parent) to selection. This matches native
+    // Maya behavior and also prevents the crash on grouping a prim twice.
+    Ufe::Selection groupSelect;
+    groupSelect.append(_groupItem);
+    Ufe::GlobalSelection::get()->replaceWith(groupSelect);
+
+    TF_VERIFY(
+        Ufe::GlobalSelection::get()->size() == 1,
+        "_group node should be in the global selection now. \n");
+}
```
HamedSabri-adsk, line 105: On undo these happens:

1- calling _undoableItem.undo() will brings back the removed _groupItem->prim(). Since this prim is stale, we can not use it. However, thanks to UFE path, we can construct a new prim off of _groupItem->path(). 

2- At this point, the newly created prim has lost it's Specifier and TypeName. Therefore line 83-87 explicitly adjust it. The better way would be that I query the original specifier beforehand but I was lazy. Should be an easy change if really needed.

3- With a valid prim, now I construct a new group scene item. Line 90

4- undo the composite inserted children. Line 94

5- and finally make sure to select the new group scene item. 

and we are done here.

lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+void UsdUndoUngroupCommand::redo()
+{
+    // redo the inserted children first
+    _compositeInsertCmd->redo();
+
+    // remove group prim
+    auto status = removePrimGroup();
+    TF_VERIFY(status, "Removing group node failed!");
+}
```
HamedSabri-adsk, line 115: On redo these happens:

1- first we redo the inserted children.
2- redo the remove.  

Note: line 113 should be _undoableItem.redo();  I will fix it in next commit.

and that's it.

test/lib/ufe/testUngroupCmd.py
```diff
@@ -0,0 +1,133 @@
...
+
+        ungroupCmd.undo()
+
+        # check parentHierarchy children size
+        self.assertEqual(len(parentHierarchy.children()), 5)
+
+        ungroupCmd.redo()
+
+        self.assertEqual(len(parentHierarchy.children()), 6)
+
```
HamedSabri-adsk, line 131: A simple test to show the basic ungroup operation. I will add a better one with more meat later.

lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
```diff
@@ -85,11 +86,15 @@ void UsdUndoUngroupCommand::undo()
...
+        SdfChangeBlock changeBlock;
 
-    primSpec->SetSpecifier(SdfSpecifierDef);
-    primSpec->SetTypeName("Xform");
+        auto primSpec = MayaUsdUtils::getPrimSpecAtEditTarget(prim);
+        TF_AXIOM(primSpec);
+
+        primSpec->SetSpecifier(SdfSpecifierDef);
+        primSpec->SetTypeName("Xform");
+    }
```
HamedSabri-adsk: TODO: I am going to refactor this part and replace it with stage::DefinePrim()

lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
```diff
@@ -81,23 +81,15 @@ void UsdUndoUngroupCommand::undo()
...
-        primSpec->SetSpecifier(SdfSpecifierDef);
-        primSpec->SetTypeName("Xform");
-    }
+    auto groupUsdPath = getPath(_groupItem->path());
+    auto groupPrim = stage->DefinePrim(groupUsdPath, TfToken("Xform"));
 
     // create a new group scene item
-    _groupItem = UsdSceneItem::create(_groupItem->path(), prim);
+    _groupItem = UsdSceneItem::create(_groupItem->path(), groupPrim);
     TF_AXIOM(_groupItem);
```
HamedSabri-adsk, line 93: refactoring the prim creation.

lib/mayaUsd/ufe/Utils.h
```diff
@@ -50,14 +50,18 @@ namespace ufe {
 // Helper functions
 //------------------------------------------------------------------------------
 
-//! Get USD stage corresponding to argument Maya Dag path.
+//! Get USD stage corresponding to argument UFE path.
```
HamedSabri-adsk, line 53: Fixed a typo in comment.

lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    _compositeInsertCmd->execute();
+
+    // remove group prim
+    MayaUsd::ufe::InAddOrDeleteOperation ad;
+    UsdUndoBlock                         undoBlock(&_undoableItem);
+
+    auto           stage = _groupItem->prim().GetStage();
+    UsdEditContext ctx(stage, stage->GetEditTarget().GetLayer());
```
ppt-adsk, line 71: What if the group prim is defined in a layer other than the edit target layer?  Will RemovePrim() fail?

lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
```diff
@@ -0,0 +1,117 @@
...
+    auto           retVal = stage->RemovePrim(_groupItem->prim().GetPath());
+    TF_VERIFY(retVal, "Failed to remove '%s'", _groupItem->prim().GetPath().GetText());
```
ppt-adsk, line 73: If retVal is false, we should undo _compositeInsertCmd, I think.

lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
```diff
@@ -0,0 +1,117 @@
...
+}
+
+void UsdUndoUngroupCommand::undo()
+{
+    MayaUsd::ufe::InAddOrDeleteOperation ad;
+
+    _undoableItem.undo();
+
+    // After undo, the group prim loses it's specifier and type.
```
ppt-adsk, line 82: "it's" --> "its".

lib/mayaUsd/ufe/Utils.h
```diff
@@ -50,14 +50,18 @@ namespace ufe {
...
 MAYAUSD_CORE_PUBLIC
 PXR_NS::UsdStageWeakPtr getStage(const Ufe::Path& path);
 
 //! Return the ProxyShape node UFE path for the argument stage.
 MAYAUSD_CORE_PUBLIC
 Ufe::Path stagePath(PXR_NS::UsdStageWeakPtr stage);
 
+//! Get USD path corresponding to argument UFE path.
+MAYAUSD_CORE_PUBLIC
+PXR_NS::SdfPath getPath(const Ufe::Path& path);
```
ppt-adsk, line 63: Would be better named as getSdfPath(), or similar.  getPath() is too generic.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -135,6 +135,11 @@ UsdStageWeakPtr getStage(const Ufe::Path& path) { return g_StageMap.stage(path);
 
 Ufe::Path stagePath(UsdStageWeakPtr stage) { return g_StageMap.path(stage); }
 
+PXR_NS::SdfPath getPath(const Ufe::Path& path)
+{
+    return SdfPath(path.getSegments().back().string());
```
ppt-adsk, line 140: We should check that the final segment's runTimeId is USD, otherwise we should return an invalid SdfPath, or throw.

lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
```diff
@@ -248,6 +249,13 @@ ProxyShapeHierarchy::reorderCmd(const Ufe::SceneItemList& orderedList) const
     return UsdUndoReorderCommand::create(getUsdRootPrim(), orderedTokens);
 }
 
+Ufe::UndoableCommand::Ptr ProxyShapeHierarchy::ungroupCmd() const
```
ppt-adsk, line 252: I think this should return a nullptr.  You can't ungroup the pseudo-root: it doesn't have a parent.

ppt-adsk (changes requested):
A few things to look at, but overall looking good!  I think my main concern is performing the test with the Maya ungroup command (from maya.cmds), rather than the UFE commands used to implement the Maya ungroup command.

lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
```diff
@@ -0,0 +1,117 @@
...
+    // At this point _groupItem->prim() is stale and should not be used.
+    // Define a new prim from ufe path to get back the original concrete xfrom prim.
+    auto stage = getStage(_groupItem->path());
+    TF_AXIOM(stage);
+
+    auto groupUsdPath = getPath(_groupItem->path());
+    auto groupPrim = stage->DefinePrim(groupUsdPath, TfToken("Xform"));
```
ppt-adsk, line 89: On thinking about this one a bit more, this line has me worried.  What if the prim we removed is not of type "Xform", but rather some funky plugin type?  What if it had custom metadata, or animations, or property values?  E.g. what if the original grouping Xform had an animated transform?  Here, you're just bringing back a brand new, vanilla Xform, so on undo, the prim won't be brought back to its original state.  I hope there's a way to use the UsdStateDelegate to bring back the full state of the original prim.

test/lib/ufe/testUngroupCmd.py
HamedSabri-adsk: @ppt-adsk  regarding the unit tests, I was planning to add more cases to cover exactly what you mentioned in ^ comments but was sidetracked with other works. 

Also, I found a crash in outliner when trying to **undo multiple ungroups**. It looks like something to do with outliner draw and notifications. I haven't had a chance to look it in details but I have it in my TODO list for tomorrow.

I will have 2-3 more test cases here soon.

lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
HamedSabri-adsk: @ppt-adsk  Very good question.... That's what I was wondering earlier. Currently, the grouping does not happen in a layer other than the edit target layer. 

See **Parent_Video2.mp4** in **MAYA-112251**  :) 

I need to think about this part a little bit more.

lib/mayaUsd/ufe/UsdUndoUngroupCommand.cpp
HamedSabri-adsk: Hmmm... I was under the impression that the UsdStateDelegate would bring back the properties but now that is not the case after doing some tests.  That's a shame....

Ok, I need to think about this.  If somehow I can store the property specs for the removed prim on execute(), then I can transfer it back to the new defined prim on undo(). Something similar to SdfCopySpec.

test/lib/ufe/testUngroupCmd.py
```diff
@@ -0,0 +1,415 @@
...
+        pass
+
+    def testUngroupProxyShape(self):
+        '''Verify ungrouping of the proxyShape.'''
+
+        # select proxyShape
+        cmds.select(ufe.PathString.string(self.proxyShapeItem.path()))
+
+        # create a group
+        cmds.group()
```
ppt-adsk, line 332: This test is grouping the proxy shape, then ungrouping.  Not sure how interesting this is: it's pretty standard Maya functionality.

What I would like to confirm is that trying to ungroup the proxy shape itself (which groups the pseudo-root's children) is not possible.  If it's prevented by virtue of the proxy shape being a shape, and thus incapable of being a parent to other Maya nodes, that's enough for me --- you can adjust this test to demonstrate that.

ppt-adsk (changes requested):
Great test!  A few changes to make here and there, and as I understand your test comment we still have a problem in testUngroupAfterUndoRedo().

lib/mayaUsd/undo/UsdUndoStateDelegate.cpp
```diff
@@ -22,6 +22,42 @@
 
 PXR_NAMESPACE_USING_DIRECTIVE
 
+namespace {
+void copySpecAtPath(const SdfAbstractData& src, SdfAbstractData* dst, const SdfPath& path)
+{
+    dst->CreateSpec(path, src.GetSpecType(path));
```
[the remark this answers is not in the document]
HamedSabri-adsk: I don't think we need it since _dst is basically a pointer to a layer's internal data but added it anyways.

```
auto layerDataPtr = get_pointer(_GetLayerData());
TF_AXIOM(layerDataPtr);
SpecCopier specCopier(layerDataPtr);
```

_GetLayerData function definition:
```
SdfAbstractDataPtr 
SdfLayerStateDelegateBase::_GetLayerData() const
{
    return _layer ? SdfAbstractDataPtr(_layer->_data) : SdfAbstractDataPtr();
}
```

lib/mayaUsd/undo/UsdUndoStateDelegate.cpp
```diff
@@ -22,6 +22,42 @@
...
+
+    const std::vector<TfToken>& tokens = src.List(path);
+
+    for (const auto& token : tokens) {
+        dst->Set(path, token, src.Get(path, token));
+    }
+}
+
+class SpecCopier : public SdfAbstractDataSpecVisitor
```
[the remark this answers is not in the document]
HamedSabri-adsk: I added a basic class description based on my brief inspection and understanding of underlying SdfAbstractDataSpecVisitor mechanics. 

As I mentioned earlier on usd group interest, these details should be handled on the USD API side and clients never have to deal with them directly.

test/lib/ufe/testUngroupCmd.py
```diff
@@ -24,17 +24,55 @@
...
+        self.gen = self.__generate(num)
+        self.num = num
+        self.contextOp = contextOp
+        self.proxyShapePathStr = proxyShapePathStr
+
+    def createSphere(self):
+        return next(self.gen)
+
+    def __addPrimSphere(self, incrment):
+        self.contextOp.doOp(['Add New Prim', 'Sphere'])
```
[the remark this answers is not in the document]
HamedSabri-adsk: You are right, UsdStage::DefinePrim() would do the same but at the time I wasn't thinking about it. I would like to leave it as since it does the work.

test/lib/ufe/testUngroupCmd.py
```diff
@@ -24,17 +24,55 @@
...
+        return ufe.PathString.path('{},/Sphere{}'.format(self.proxyShapePathStr, incrment))
+
+    def __generate(self, num):
+        incrment = 0
+        while incrment < self.num:
+            incrment += 1
+            yield self.__addPrimSphere(incrment)
+
+def joinPathSegments(ufePath):
```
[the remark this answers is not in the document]
HamedSabri-adsk: Thanks. Good point.  3911bc85ea7d6f8357d851bba8c97c54a77d8503

lib/mayaUsd/undo/UsdUndoStateDelegate.cpp
```diff
@@ -327,7 +367,15 @@ void UsdUndoStateDelegate::_OnDeleteSpec(const SdfPath& path, bool inert)
         return;
     }
 
-    SdfDataRefPtr     deletedData = TfCreateRefPtr(new SdfData());
+    SdfDataRefPtr deletedData = TfCreateRefPtr(new SdfData());
+
+    SdfLayer::TraversalFunction copyFunc = std::bind(
```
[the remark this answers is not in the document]
HamedSabri-adsk: Good point. I always prefer lambda over std::bind to avoid unnecessary overheads and also found it easier to read/write the code. I learned this from Scott's talk long time ago and never had any issues with it.

Also, there are still some ugly std::bind left in UsdUndoStateDelegate.cpp but I don't want to do it now to make more noise.  

addressed ce535e5fbaa0f4389f2469d5caa99345b4f0cebd

test/lib/ufe/testUngroupCmd.py
```diff
@@ -0,0 +1,415 @@
...
+        for _ in range(4):
+            pathList.append(sphereGen.createSphere())
+
+        # group /Sphere2, /Sphere3, /Sphere4
+        cmds.group(joinPathSegments(pathList[1]),
+                   joinPathSegments(pathList[2]),
+                   joinPathSegments(pathList[3]))
+
+        # verify the paths after grouping
+        assert ([x for x in self.stage.Traverse()] ==
```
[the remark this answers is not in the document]
HamedSabri-adsk: addressed 3911bc85ea7d6f8357d851bba8c97c54a77d8503

ppt-adsk (commented):
Thanks for the changes.  Only a couple of minor things left in UsdUndoUngroupCommand.cpp and UsdUndoStateDelegate.cpp.  What about the problem that remains in testUngroupCmd.py, lines 402 to 409?  Is this for a later branch, or do you plan on fixing this now?

lib/mayaUsd/undo/UsdUndoStateDelegate.cpp
```diff
@@ -23,17 +23,23 @@
...
+    // get the list of tokens
     const std::vector<TfToken>& tokens = src.List(path);
 
+    // set the value of dst at the given path and a fieldName
     for (const auto& token : tokens) {
         dst->Set(path, token, src.Get(path, token));
     }
 }
 
+// This class is used to copy specs from source SdfAbstractData contrainer
```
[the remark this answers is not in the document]
HamedSabri-adsk: @ppt-adsk  I have `testungroup.py ` and `UsdUndoUngroupCmd[*.h][*.cpp] ` wrapped in `CMAKE_UFE_V3_FEATURES_AVAILABLE` since these are UFE3 features.  See 91e77bc1664925763db75b4dad165bded244c179

HamedSabri-adsk:
> Thanks for the changes. Only a couple of minor things left in UsdUndoUngroupCommand.cpp and UsdUndoStateDelegate.cpp. What about the problem that remains in testUngroupCmd.py, lines 402 to 409? Is this for a later branch, or do you plan on fixing this now?

I haven't looked into it just yet. I will probably have a separate PR to address this with the fixes for parent, ungroup item selection .
I will add a note in the PR description for known issues.

ppt-adsk (commented):
Still a few unresolved conversations in there. Ping me when you feel it's ready for another review round.

test/lib/ufe/testUngroupCmd.py
```diff
@@ -0,0 +1,415 @@
...
+
+        # TODO: after ungroup all children must be selected per Maya native behavior.
+        # HS, June 21, 2021 this is not yet implemented
+        # self.assertEqual(len(self.globalSn), 2)
+
+        # undo
+        cmds.undo();
+
+        # verify that pseudoroot has 1 child (group1)
+        self.assertEqual(len(self.stage.GetPseudoRoot().GetChildren()), 1)
```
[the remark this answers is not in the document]
HamedSabri-adsk: addressed 6ec37a5bfa64a3046017f90480a7c72cc34929e6

test/lib/ufe/testUngroupCmd.py
```diff
@@ -320,38 +322,18 @@ def testUngroupParent(self):
...
 
-            dagItr.next()
+        # create 5 spheres 
+        for _ in range(5):
+            sphereGen.createSphere()
 
-        self.assertEqual(groupFound, False)
+        # ungrouping the proxy shape with existing USD children should fail
+        with self.assertRaises(RuntimeError):
+            cmds.ungroup(ufe.PathString.string(self.proxyShapeItem.path()))
```
HamedSabri-adsk, line 336: ungrouping the proxy shape should fail 

6ec37a5

lib/mayaUsd/ufe/UsdUndoUngroupCommand.h
```diff
@@ -0,0 +1,58 @@
...
+#include <mayaUsd/ufe/UsdSceneItem.h>
+#include <mayaUsd/undo/UsdUndoableItem.h>
+
+#include <ufe/undoableCommand.h>
+
+namespace MAYAUSD_NS_DEF {
+namespace ufe {
+
+//! \brief UsdUndoUngroupCommand
+class MAYAUSD_CORE_PUBLIC UsdUndoUngroupCommand : public Ufe::UndoableCommand
```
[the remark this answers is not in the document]
ppt-adsk: After discussion, this would be a premature rename.  We will wait until it's needed.

test/lib/ufe/testUngroupCmd.py
```diff
@@ -0,0 +1,415 @@
...
+        groupItem = self.globalSn.front()
+        self.assertEqual(str(groupItem.path().back()), "group1")
+
+        # verify that groupItem has 2 children
+        groupHierarchy = ufe.Hierarchy.hierarchy(groupItem)
+        self.assertEqual(len(groupHierarchy.children()), 2)
+
+        # remove group1 from the hierarchy. What should remain
+        # is /Sphere1, /Sphere2.
+        cmds.ungroup("{},/group1".format(self.proxyShapePathStr), absolute=True)
```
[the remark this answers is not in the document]
ppt-adsk: As discussed, this requires a change to group, scheduled for later work, so this suggestion cannot be implemented now.

ppt-adsk (approved):

lib/mayaUsd/ufe/ProxyShapeHierarchy.h
```diff
@@ -77,6 +77,8 @@ class MAYAUSD_CORE_PUBLIC ProxyShapeHierarchy : public Ufe::Hierarchy
     insertChildCmd(const Ufe::SceneItem::Ptr& child, const Ufe::SceneItem::Ptr& pos) override;
 
     Ufe::UndoableCommand::Ptr reorderCmd(const Ufe::SceneItemList& orderedList) const override;
+
+    Ufe::UndoableCommand::Ptr ungroupCmd() const override;
```
seando-adsk, line 81: I think this needs to be #ifdef UFE_V3

ppt-adsk (approved):

lib/mayaUsd/ufe/ProxyShapeHierarchy.h
HamedSabri-adsk: I already have it protected. See line 82?

ppt-adsk (approved):

seando-adsk (approved):

PR #1471, opened 2021-06-08
Custom User Attributes Import/Export & Custom Export Hierarchy (-userattr, -root)

BSalem:
Compatibility with the way how Alembic used to handle the IO of the User Attributes
And the Hierarchy during the I/O operations, the aim of these features is to simplify and reduce the amount of work for the migration from Alembic to USD pipeline.

Features:
---------

* Added -root/-rt multi-flag that copies the behavior of AbcExport which limits the roots under which the primitives will be written in the USD rather than translating all the parents up till scene root.

Enhancements:
-------------

* -root/-rt (multi-flag) and -sl flag together gives full control on which part of the hierarchy to be written to the USD and what exactly under the given roots will be exported, for example:
if we have the following hierarchy:

    Assets
    |---- chr_model_ast
    |----|---- body
    |----|---- aux_nodes
    |----chr_garment_ast
    |----|---- cloth
    |----|---- accessories
    |----|---- garment_curves_grp
    |----|---- sim_garment

select -r body cloth accessories::

    usdExport -root "chr_model_ast" -root "chr_garment_ast" -userattr "primvars:varyColor" -userattr "userProperties:ast" -userattr "userProperties:version" -sl -file "/path/to/usd.usda";

this will write a USD that have the following hierarchy (with the custom attributes exported):

/
chr_model_ast
|--- body
chr_garment_ast
|--- cloth
|--- accessories

### DIFF
--- a/lib/mayaUsd/commands/baseExportCommand.cpp
+++ b/lib/mayaUsd/commands/baseExportCommand.cpp
@@ -180,6 +180,12 @@ MSyntax MayaUSDExportCommand::createSyntax()
     syntax.addFlag(kFileFlag, kFileFlagLong, MSyntax::kString);
     syntax.addFlag(kSelectionFlag, kSelectionFlagLong, MSyntax::kNoArg);
 
+    syntax.addFlag(kUserAttrFlag, kUserAttrFlagLong, MSyntax::kString);
+    syntax.makeFlagMultiUse(kUserAttrFlag);
+
+    syntax.addFlag(kRootFlag, kkRootFlagLong, MSyntax::kString);
+    syntax.makeFlagMultiUse(kRootFlag);
+
     syntax.addFlag(kFilterTypesFlag, kFilterTypesFlagLong, MSyntax::kString);
     syntax.makeFlagMultiUse(kFilterTypesFlag);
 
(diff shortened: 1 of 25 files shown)


kxl-adsk:
@BSalem Thank you for your PR. Before digging deeper into the code I wanted to check with you if you happen to see the conversation in a similar PR to yours - #1111? Especially the part where we outlined the requirements for the final implementation - https://github.com/Autodesk/maya-usd/pull/1111#issuecomment-804408492

BSalem:
Yeah... I did, what you planned for is somehow what I was looking for...
I believe my implementation here can achieve the same requirements as your method (although, I might have followed a different approach according to my humble understanding to the code tree and the API) - I treated empty root somehow like "*" that when no -root flag is set.

In case we want each selected object to be a root as well, then we can simply pass the selection list to the root arg, something like this::

    usdExport(file="/path/to/scene.usda", sl=True, root=ls(sl=True))

or in case we want to select/export using wildcard:

    select(ls('*_geo'), r=True)
    usdExport(file="/path/to/scene.usda", sl=True, root=ls('*_model_ast'))

Anyway, you can say, I couldn't wait until I see such root feature there, both the user custom attributes and the hierarchy are super important for us in Seneferu's pipeline and applications, we rely intensively on -sl and -root flags, by copying a somehow similar behavior of what AbcExport does, the migration from Alembic to USD becomes as easy as replacing AbcExport with usdExport while keeping everything else as is, then things go as expected (at least at this phase and until we completely migrate to USD).

BSalem:
What worth mentioning about my method, that:
1.By using root(s) and sl, then selected objects that are not under one of the given roots won't be exported (in such case, we have to also add that object to the roots list passed to the export command - in addition to that it's selected).
2. Ignored parents that were above the given roots, will have their transformations ignored as well (in future - we can add a -w/-world flag that can be used to consider the parents transforms).

kxl-adsk:
@BSalem Thank you for these extra details.

You have joined a couple of changes together in this PR but to move it forward we will have to focus on them separately. I would propose we start with "export roots" functionality because this is something we have already discussed in the past and several other users raised interest/need for it. 

Here are the next steps that will have to be taken to move things forward:

- We need a signed CLA before we can merge any contributions. Please refer to https://github.com/Autodesk/maya-usd/blob/dev/doc/CONTRIBUTING.md
- Open a separate PR that only focuses on "export roots" functionality
- Write unit tests to cover what original PR #1111 intended to support (there is a unit test included in the PR) and workflows mentioned in the PRs thread
- Iterate with us on how to support exporting of all selected objects as roots. Some questions that were raised internally: how does this work in MEL? how can we avoid passing the entire selection list? (maybe special empty string?) 

Is the above plan something you are ready to take on?

BSalem:
Yeah... sure, I thought my old CLA on PixarUSD would work, I signed now the one you shared.
- I will separate the PRs, that's not a big deal (they were two commits locally already).
- I will add a unit test to the new PR.

Regarding your questions:
1. Since the -root flag is a multiflag, then it should be given multiple times once per each root:

     `usdExport -root "chr_model_ast" -root "chr_garment_ast" -sl -file "/path/to/usd.usda";`

in Python, the root (as usual in this case in maya with multi-flags), it can take a string or list:

`usdExport(file="/path/to/scene.usda", sl=True, root=["chr_model_ast", "/path/to/usd.usda"])`


2. For the case when we want each selected object to be its own root without passing the whole selection list to the root arg, I would like to add some suggestions for this together besides yours, then please let me know what do you think:
     a. adding another flag called '-noroot' (to mean that don't look for parents/roots)
     b. passing "" to the root flag (to mean no roots are required or the root is None/empty).
     c. passing "*" to the root flag (to mean that "all" selected objects are roots)
     **d. The option that I highly recommend, the "sl/exportSelected" flag should do it by default without the root flag at all, I mean, to let 'sl' act as exactly what it says, "exportSelected" and nothing else unless it's requested, accordingly, if the 'sl' flag is passed and the '-root' flag was not, then each selected object will go as is without its parents, if it happens and we want the parents, we can request them by the root flag, if we don't want the parent(s) but we want to respect their transformations of the exported object, we can add a -w/-world flag that writes the objects in world space**

What do you think?

kxl-adsk:
I run this by our team and we would like to proceed with option B, i.e. passing empty string. Your proposed solution D would change the default behavior and it's not something we have a taste of doing.

We should probably discuss how empty string will work when passed together with it another root flag, but feel free to propose an implementation for it.

BSalem:
cool... I actually did both option D and the option with empty string (two PRs are there for them).
maybe we can discuss them on their PRs.

kxl-adsk:
I'm closing this PR given it is now split into multiple (#1491, #1497, #1498).

PR #1450, opened 2021-06-02
MAYA-111901 MaterialX export of simple standard surface setups

JGamache-autodesk:
Exports standard surface and file nodes to a MaterialX UsdShade network.

PR #1478, opened 2021-06-08
MaterialX import export

JGamache-autodesk:
This PR is built on top of PR #1450 and PR #1456.

So we get the previously submitted MaterialX export and the token refactoring.

In addition, we now get:
- Import of the MaterialX exported from Maya back to Maya
- Support for preview surface (in MaterialX)
- Support for normal mapping in standard surface


lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -597,7 +597,11 @@ class _UVMappingManager
                     TfStringPrintf("%s:%s", itNode->GetText(), _tokens->varname.GetText()));
                 UsdShadeInput materialInput = material.GetInput(inputName);
                 TF_VERIFY(itName != largestSet.cend());
-                materialInput.Set(*itName);
```
JGamache-autodesk, line 600: MaterialX uses std::string while UsdPreviewSurface uses TfToken.

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -512,9 +516,11 @@ class UseRegistryShadingModeImporter
         const auto                   iter = _shaderReaderMap.find(shaderPath);
         UsdMayaShaderReaderSharedPtr shaderReader;
         MObject                      sourceObj;
+        MPlug                        sourcePlug;
         if (iter != _shaderReaderMap.end()) {
             shaderReader = iter->second;
-            if (!_context->GetCreatedObject(shaderSchema.GetPrim(), &sourceObj)) {
```
JGamache-autodesk, line 517: Ask the shader reader in case it is a converter.

lib/usd/translators/shading/mtlxPreviewSurfaceWriter.cpp
```diff
@@ -0,0 +1,202 @@
...
+            // Done with inputs.
+            break;
+        }
+
+        MPlug attrPlug = depNodeFn.findPlug(mayaAttrName.GetText(), true);
+
+        // Keep our authoring sparse by ignoring attributes with no values set
+        // and no connections. We know that the default value of base and base
+        // color diverged between Maya and MaterialX in version 1.38.
+        if (!UsdMayaUtil::IsAuthored(attrPlug) && !attrPlug.isConnected()) {
```
kxl-adsk, line 111: Thank you for the comment. This is good to have captured in here.

test/lib/usd/translators/CMakeLists.txt
```diff
@@ -231,6 +231,23 @@ if (MAYA_APP_VERSION VERSION_GREATER_EQUAL 2020)
     endforeach()
 endif()
 
+# Tests using MaterialX, which was introduced in Maya 2022.1 with USD > 21.05.
+if (MAYA_APP_VERSION VERSION_GREATER 2022 AND PXR_VERSION VERSION_GREATER 2102)
```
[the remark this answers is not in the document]
JGamache-autodesk: The import/export code does not require a recent USD since we write directly in UsdShade the expected MaterialX graph that represents the Maya nodes we support.
Using 2102 allows the unit test to be run by some of our preflight configurations even though the Vp2 render delegate would be incapable of rendering the results. This is why the unit test is in `test/lib/usd/translators` instead of `test\lib\mayaUsd\render\vp2RenderDelegate`.
I will clarify the comment then to separate the export/import, which can be done in 21.02, from the rendering part, which absolutely requires >21.05.

lib/usd/translators/shading/usdStandardSurfaceReader.cpp
```diff
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+#include "shadingTokens.h"
```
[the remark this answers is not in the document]
JGamache-autodesk: Yes once the local _tokens are replaced with TrMayaTokens.

test/lib/usd/translators/UsdImportMaterialX/UsdImportMaterialX.usda
```diff
@@ -0,0 +1,178 @@
...
+{
+    rel material:binding = </MaterialX/Materials/Gold>
+    double3 xformOp:translate = (8, 0, 0)
+    uniform token[] xformOpOrder = ["xformOp:translate"]
+}
+
+over "MaterialX"
+(
+    references = [@./standard_surface_jade.usda@</MaterialX>,
+                  @./standard_surface_gold.mtlx@</MaterialX>]
```
JGamache-autodesk, line 46: Will exercise the SdfFileReference auto-conversion to USD, but requires the UsdMtlx plugin which is not in the USD we currently use for tests.

test/lib/usd/translators/testUsdImportMaterialX.py
```diff
@@ -0,0 +1,119 @@
...
+
+try:
+    from pxr import UsdMtlx
+    USD_HAS_MATERIALX_PLUGIN=True
+except ImportError:
+    USD_HAS_MATERIALX_PLUGIN=False
+
+
+class testUsdImportMaterialX(unittest.TestCase):
+    """Test importing an asset with MaterialX shading:
```
kxl-adsk, line 34: Excellent test, @JGamache-autodesk! The only thing that would be good to add (but it's not blocking for this merge) is a test that performs export and import. I know part of the import file is actually having content exported from Maya, but I would want to avoid situations where USD test file gets updated but not the actual export code.

kxl-adsk (approved):

PR #1479, opened 2021-06-08
Fixed bug with visibility determination

ysiewappl:
Hi all:

This small PR fixes an issue where if you have the Draw Overrides option enabled on a DAG node set to template instead of using the one in the Display rollout, the utility function in Maya-USD would report it to be renderable instead of not. This was causing issues for us internally in our own internal chaser plugins when determining visibility state.

### DIFF
--- a/lib/mayaUsd/utils/util.cpp
+++ b/lib/mayaUsd/utils/util.cpp
@@ -627,7 +627,8 @@ bool UsdMayaUtil::isRenderable(const MObject& object)
     }
 
     const MDAGDrawOverrideInfo drawOverrideInfo = dagPath.getDrawOverrideInfo();
-    if (drawOverrideInfo.fOverrideEnabled && drawOverrideInfo.fDisplayType == MDAGDrawOverrideInfo::kDisplayTypeTemplate) {
+    if (drawOverrideInfo.fOverrideEnabled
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/utils/util.cpp
```diff
@@ -619,6 +619,18 @@ bool UsdMayaUtil::isRenderable(const MObject& object)
...
+    stat = mFn.getPath(dagPath);
+    CHECK_MSTATUS_AND_RETURN(stat, false);
+    if (!dagPath.isVisible()) {
+        return false;
+    }
+
+    const MDAGDrawOverrideInfo drawOverrideInfo = dagPath.getDrawOverrideInfo();
+    if (drawOverrideInfo.fOverrideEnabled && drawOverrideInfo.fDisplayType == MDAGDrawOverrideInfo::kDisplayTypeTemplate) {
+        return false;
+    }
```
HamedSabri-adsk, line 632: Lgtm! So to be clear, previously this function didn't return the state correctly when the draw override's display type was set to "Template".

image

HamedSabri-adsk (approved):

kxl-adsk:
@ysiewappl Please fix clang-format so we can start PF for this change.

PR #1482, opened 2021-06-11
Added instruction for installing devtoolset-6 on CentOS.

HamedSabri-adsk:


doc/build.md
```diff
@@ -235,7 +235,26 @@ To allow your tests to run, you can inject LD_LIBRARY_PATH into any of the mayaU
 
 There is a related ADDITIONAL_PXR_PLUGINPATH_NAME cmake var which can be used if schemas are installed in a non-standard location
 
+##### Devtoolset-6:
 
+Devtoolset-6, which includes GCC 6 on CentOS has been retired from the main repo and moved into the vault. Follow below instruction for installing devtoolset-6 on CentOS:
```
ppt-adsk, line 240: "Follow below instruction" --> "Follow instructions below".

ppt-adsk (commented):
Minor fix, but otherwise thanks for doing this!  Maybe at some point if platform-specific instructions become too long we can move them into a separate platformBuild.md page (say) and provide a link in build.md to it.

HamedSabri-adsk:
> Minor fix, but otherwise thanks for doing this! Maybe at some point if platform-specific instructions become too long we can move them into a separate platformBuild.md page (say) and provide a link in build.md to it.

Good point! For now I leave it as is but if more platform specific instruction are required then having a separate page makes sense to have.

ppt-adsk (approved):

PR #1489, opened 2021-06-12
-root/-rt flag (limit export USD parents level - empty string means all selected are roots)

BSalem:

Enhancement:
------------

- if empty string "" is passed to root, each selected object/transform will be treated as a root by its own.

      usdExport(file='/path/to/usd.usda', sl=True, root='')


BSalem:
Now this method will also support mixing empty string and actual roots in the root arg:

`usdExport(file='/path/to/scene.usda', sl=True, root=['SomeRoot1', 'SomeRoot2', ''])`

the method above will export each selected object as a root by itself, if one of the actual passed root names is a parent of any of the selected objects, then this parent becomes the root instead

BSalem:
Closing temporarily! Was able to produce a situation where this method needs to be enhanced

PR #1490, opened 2021-06-12
-root/-rt flag (sl flag doesn't export parents to top unless "|" is passed to the 'root' flag)

BSalem:
New Feature
-----------

Enhancement:
------------

exportSelected flag now will export only the selected objects and their children, all parents above the selected objects will be ignored unless requested via the 'root' flag.

To achieve the old behavior the 'exportSelected' flag used to do before the implementation of the "root" feature, a pipe "|" can be passed to the root flag, this will export the selected objects under all their parents up till the scene top root::

    `usdExport(file='/path/to/usd.usda', sl=True, root='|')`


BSalem:
The advantage of this method that we can easily mix/pass additional roots besides the roots that were internally made of the selection list.

BSalem:
Closing temporarily! Was able to produce a situation where this method needs to be enhanced

PR #1491, opened 2021-06-12
Creating shape/transform user custom attributes on USD Import (with anim support)

BSalem:
Importing/Exporting custom user attributes

Features:
---------

Enhancements:
-------------

Known Limitations:
------------------

- Animation of animated custom Array Attributes will not be animated, the values of the attribute array will be the value at the last timesSample of the importing frame range.

### DIFF
--- a/lib/mayaUsd/commands/baseExportCommand.cpp
+++ b/lib/mayaUsd/commands/baseExportCommand.cpp
@@ -180,6 +180,9 @@ MSyntax MayaUSDExportCommand::createSyntax()
     syntax.addFlag(kFileFlag, kFileFlagLong, MSyntax::kString);
     syntax.addFlag(kSelectionFlag, kSelectionFlagLong, MSyntax::kNoArg);
 
+    syntax.addFlag(kUserAttrFlag, kUserAttrFlagLong, MSyntax::kString);
+    syntax.makeFlagMultiUse(kUserAttrFlag);
+
     syntax.addFlag(kFilterTypesFlag, kFilterTypesFlagLong, MSyntax::kString);
     syntax.makeFlagMultiUse(kFilterTypesFlag);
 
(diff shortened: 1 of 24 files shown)


BSalem:
Is there any concerns regarding this PR?

kxl-adsk:
We haven't got to this PR yet. I wonder if you explored `apiSchema` flag on import? This would allow you to author custom schema attributes easily on Maya's data model with adaptor class and roundtrip them after with mentioned flag. Have a look at regression tests for some examples.

BSalem:
No honestly I didn't test it, it wasn't clear to me what does it do... I will look at it.

kxl-adsk:
I haven't seen any update since the last comment. Closing it for now. Feel free to start a new discussion if existing options are unable to fit in your workflow.

BSalem:
seems I forget to update on it!
When I tried to use apiSchema I couldn't get the attributes imported.

Can the apiSchema flag import string attr userProperties:tex_dir (or animated vector userProperties:astPos) as custom attribute somehow?

BSalem:
Is there a way to reopen this PR? or discuss it?
I'm not sure if you can see my comment on the closed ticket.

kxl-adsk:
Yes, we get notifications regardless of the status of the PR. I wanted to find some time to write examples but here it goes.

The options that you are having:
1. Go with an import chaser and bring the custom attributes. No API schema is required and with the latest job context addition, you can configure this for artists without modifying the core classes. Here is a test that could get you started - https://github.com/Autodesk/maya-usd/blob/83887106d8ff21c68607fdf528b97e4506491a99/test/lib/mayaUsd/fileio/testImportChaser.py
2. Define an API schema for your attributes and implement API schema adapters to have attributes discovered on export and on import. We just added the flexibility to customize how API schema attributes are stored in Maya's data model. See https://github.com/Autodesk/maya-usd/blob/572ee57a619b20359e7fb5766c39e45581494901/test/lib/usd/translators/testUsdExportSchemaApi.py#L117 as a start. We are also working on a white paper to explain better the concepts. Prior to this work, API schema attributes would also be interop, but authoring needed to happen via adaptors.

oumad:
I reached this issue after a lot of search failing to find a way to round trip custom attributes with maya import/export. 
based on the doc https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/commands/Readme.md
UJsing "USD_UserExportedAttributesJson" attribute containing a json, we're able to keep our maya custom attributes in a USD export, but once we re-import that USD file we lose those attributes. There doesn't seem to be a flag in USD import to get these attributes back. Do we really have to setup some "chaser" or an API schema just to get our attributes to round trip ?

BSalem:
To be honest, I myself haven't succeeded to get all the custom attributes imported back into maya too, knowing the names of the custom attributes you want to import is not an option in lots of cases, so we are still relying on this PR to have all the custom attributes imported as the default behavior.

In my opinion, we should not decide to ignore things, when we import a USD, we are expecting to have everything inside it just available for use, if we don't want them, we wouldn't export them at the first place (this is the current behavior of other applications and it's how alembic used to be as well)

**So, I suggest that we import everything by default with no exceptions, while adding a flag (like -ignoreCustomAttrs) for those who don't want the custom attributes get imported.**

PR #1497, opened 2021-06-15
-root/-rt flag (exportSelected - are roots until a "|" is passed to the root arg)

BSalem:
New Feature
----------------

- exportSelected flag now will export only the selected objects and their children (where each selected object is treated as root by its own), while all parents above the selected objects will be ignored unless requested via the 'root' flag, this way, we don't have to pass all the selected objects to the root arg to make them roots as well (but they will be like that by default).

**However,** to achieve the old behavior the 'exportSelected' flag used to do before the implementation of the "root" feature, a pipe "|" can be passed to the root flag, this will export the selected objects under all their parents up till the scene top root ("|"):

    `usdExport(file='/path/to/usd.usda', sl=True, root='|')`

**Note:**
    Currently there is no unit test, however, the python script found in test/lib/usd/translators/UsdExportRootTest can help testing almost all the possible different export scenarios 

`$ cd test/lib/usd/translators/UsdExportRootTest`
`$ mayapy -c "import test_usdExport"`


BSalem:
Since we already implemented another option for passing selected objects as roots by passing empty string to represent all selected objects that is not under a direct given root, I'm closing this option.

PR #1498, opened 2021-06-15
-root/-rt flag (exportSelected - are treated as roots by their own if empty string is passed to the root arg)

BSalem:

- passing empty string to the root arg, will treat each selected node as root by its own, this will help writing the selected objects on the USD root without having to pass the whole selection list to the root arg in the export command.
- Mixing actual root(s) and empty string "" in the list given to the root arg (or passing -root "" in combination with -root "something" in case with mel), then all selected and falling under the actual given root will go under that root, everything else falling outside the given root will be treated as independent root.

**Note:**
    Currently there is no unit test, however, the python script found in test/lib/usd/translators/UsdExportRootTest can help testing almost all the possible different export scenarios 

`$ cd test/lib/usd/translators/UsdExportRootTest`
`$ mayapy -c "import test_usdExport"`


kxl-adsk:
@BSalem Thank you for splitting the PR. As you might have noticed, I closed the original PR https://github.com/Autodesk/maya-usd/pull/1471. 

I know you proposed multiple options to how we could approach exportRoots, but we would like to invest your and our time only in the solution that uses empty string (as per https://github.com/Autodesk/maya-usd/pull/1471#issuecomment-859850285). Please close https://github.com/Autodesk/maya-usd/pull/1497.

As you remember, a very similar pull request was open in the past (#1111) and a number of users expressed interest in it. It's never an easy situation when we have conflicting PRs. Your PR seems to match our requirements more closely, but I would like to make sure to cover all user cases from PR #1111. The easiest to do so will be to leverage the regression test available in that PR and include an adapted version to this PR -  https://github.com/Autodesk/maya-usd/blob/51d39616f280a8d3c92803c4cfa80c5906066018/test/lib/usd/translators/testUsdExportRoot.py.

 Finally, I have asked @JGamache-autodesk and @nxkb to help with the code review.

BSalem:
Sure Krystian... I closed the PR of the other option so we focus on this one, however, this option with the empty string required that we do fullDagPath string comparisons for all each selected object against each given root besides the empty string to tell which selected object should be represented by the empty string and which one goes under a directly given root... in case you think that this is a performance issue especially on heavy scenes, then you may consider that the other option didn't require such string comparisons.

Regarding the unit test, I will implement soon.

**There are two more things to discuss:**
1. Exporting with the objects as roots, then the objects looks are also on the USD root unless the a parent group was set as root (for example: in case Top->Group1->Cube1, the looks of Cube1 will come in the USD under "looks" scope if the Top group or Group1 is the export root for selected Cube1, but when Cube1 is the selected and the root, then the looks will be written to the root of the USD directly (with the "looks" scope)... so, **is this fine? or we should force creating the "looks" scope on such case?**
2. When writing the USD with multiple roots, empty string, or mix of the two, the primitives in the USD file are written in the same order of which the roots were given, which means that changing the order of the given roots will change the order of the objects inside written USD, **is this OK? or should we do alphabetical sort before the writing loop?**

nxkb:
Nice work BSalem, I rebased our #1111 on the latest dev in case it helps to grab the tests.

> 1. Exporting with the objects as roots, then the objects looks are also on the USD root unless the a parent group was set as root (for example: in case Top->Group1->Cube1, the looks of Cube1 will come in the USD under "looks" scope if the Top group or Group1 is the export root for selected Cube1, but when Cube1 is the selected and the root, then the looks will be written to the root of the USD directly (with the "looks" scope)... so, **is this fine? or we should force creating the "looks" scope on such case?**

If Cube1 was the root and selected: provided the assignments were to Cube or lower... wouldn't it end up with "/Cube1/Looks"? That seems desirable from a referencing point of view, because then the Looks would come along for the ride. If there's no "common ancestor" of material assignment targets, it looks like we do author Looks at the root. So if there is a case you can't quite handle and that is the result, then the savvy user could use a "parentScope" arg to get everything under a single root, but I think we want to make that unnecessary whenever possible.

> 2. When writing the USD with multiple roots, empty string, or mix of the two, the primitives in the USD file are written in the same order of which the roots were given, which means that changing the order of the given roots will change the order of the objects inside written USD, **is this OK? or should we do alphabetical sort before the writing loop?**

I think preserving the order is the lesser of two evils, just in case someone does want the ordering. I don't have a use case for that yet, but prims do stay ordered in usd files so keeping the maya DAG order or the arg order sounds best to me.

BSalem:
Thanks @nxkb, I will then take your test and tweak for this here.

Regarding the "Looks" scope, the way the exporting commands used to work was to create a Look scope for the whole stage under the Top root that has all shading engines... so, in case when there are multiple roots, then we have two options here:
Option A: writing a "Looks" scope under each root hierarchy for the prims related to this root.
Option B: writing all the shading engines for the whole stage under one "Looks" scope under the stage root (which I'm currently doing - although, the method that writes the shaders doesn't seem to be creating a "Looks" scope to hold the shading engines when it detects that the looks parent is the stage root - didn't discover where this happens yet)

nxkb:
I think B makes more sense to me, as it will avoid repetitive scene description if there is material reuse.

> although, the method that writes the shaders doesn't seem to be creating a "Looks" scope to hold the shading engines when it detects that the looks parent is the stage root - didn't discover where this happens yet)

Yeah the behavior in `_GetMaterialParent()` strikes me as a bit odd. If you pass in a looks scope as a parameter to the export, it should be used even if there is no common ancestor and it defaults to the stage root. However the way that it defaults to a defaultMaterialsScopeName makes it hard to know whether the user passed it in or not. I'd probably vote for the scope (or default material scope) to be authored even at the root for consistency, but I don't know if anyone's relying on the current behavior. Its borderline out of the PR scope but I can see how it would influence your decision with the roots. @kxl-adsk any thoughts on that?

BSalem:
I agree with you, but what happens when merging/referencing USDs later, does USD treat the material path as relative to the loaded stage hierarchy? Or is it treated as absolute path where we may experience paths/names clashes in such case? (I don't know how this is currently being handled)

lib/mayaUsd/fileio/jobs/writeJob.cpp
```diff
@@ -296,85 +371,142 @@ bool UsdMaya_WriteJob::_BeginWriting(const std::string& fileName, bool append)
...
             continue;
         }
         curDagPathIsValid = curDagPath.isValid(&status);
 
         while (status == MS::kSuccess && curDagPathIsValid) {
             curDagPathStr = curDagPath.partialPathName(&status).asChar();
+
             if (status != MS::kSuccess) {
-                break;
+                goto ctn;
```
JGamache-autodesk, line 390: That `goto` is going to make a lot of people uncomfortable. The use is correct, though recommended only when exiting nested loops. Please review http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-goto and potentially the next recommendation regarding breaking out of loops.
In doubt, just leave the code as is.

test/lib/usd/translators/UsdExportRootTest/test_usdExport.py
```diff
@@ -0,0 +1,49 @@
...
+def main():
+    tmpdir = tempfile.gettempdir()
+    scn_dir = os.path.dirname(os.path.realpath(__file__))
+    
+    cmds.loadPlugin('mayaUsdPlugin')
+    cmds.loadPlugin('pxrUsd')
+    cmds.file('{}/UsdExportRootTest.ma'.format(scn_dir), type="mayaAscii", o=True)
+   
+    select(['pCone1', 'Cube1'], r=True)
+    cmds.usdExport(file='{}/only_set1_to_top.usda'.format(tmpdir), sl=True, root="Top")
```
JGamache-autodesk, line 15: Can you please open the stages afterwards and make sure the paths are as expected?

JGamache-autodesk (changes requested):

nxkb:
> I agree with you, but what happens when merging/referencing USDs later, does USD treat the material path as relative to the loaded stage hierarchy? Or is it treated as absolute path where we may experience paths/names clashes in such case? (I don't know how this is currently being handled)

In general, I'd say it is "relative" to the usd files hierarchy. Most of the time you'd use a reference arc to add an asset to a stage and then the materials would come in if they were under the default/referenced prim. It's hard to answer absolutely because the possibilities of composition and loading are vast. You could probably get an "absolute" path behavior using sublayers but I think you'd regret that setup as you try to assemble a shot from different assets.

kxl-adsk:
@BSalem Based on review comments - there is more work required to get this PR to a mergeable state. This feature is important for other customers and internally we are as well going to need it. Do you still need any feedback from our side? Is it still something you are hoping to contribute back?

kxl-adsk:
Given that there was no progress on this change and that we are getting customer requests and internal requests to support it we opened https://github.com/Autodesk/maya-usd/pull/1592 to deliver the feature and address the requested changes.

BSalem:
sorry, was super busy to address the notes.
So, is #1592 now supports exporting multiple roots as well as considering empty string for writing selections as roots? Or should we still address the given notes and reopen the PR?

PR #1499, opened 2021-06-16
Add geomSidedness flag to exporter

dgovil:
This PR adds a new flag to the exporter to override the sidedness of the mesh.
Options are:

* **derived** : Same behaviour as right now where it's taken from the double Sided attribute
*  **single** : All gprims are exported as single sided
* **double** : All gprims are exported as double sided

I've also updated the mesh tests and updated the command documentation.

### DIFF
--- a/lib/mayaUsd/fileio/translators/translatorGprim.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorGprim.cpp
@@ -15,8 +15,6 @@
 //
 #include "translatorGprim.h"
 
-#include "../writeJobContext.h"
-
 #include <mayaUsd/utils/util.h>
 
 #include <maya/MFnDagNode.h>
@@ -44,16 +42,13 @@ void UsdMayaTranslatorGprim::Write(
 {
     MFnDependencyNode depFn(mayaNode);
 
-    bool doubleSided = false;
-    if (sidedness == GeomSidedness::Derived
-        && UsdMayaUtil::getPlugValue(depFn, "doubleSided", &doubleSided) && doubleSided) {
-        doubleSided = true;
-    } else if (sidedness == GeomSidedness::Double) {
-        doubleSided = true;
-    }
-
-    if (doubleSided) {
-        gprim.CreateDoubleSidedAttr(VtValue(doubleSided), doubleSided);
+    bool doubleSided = sidedness == GeomSidedness::Double;
+    if (sidedness == GeomSidedness::Derived) {
+        if (UsdMayaUtil::getPlugValue(depFn, "doubleSided", &doubleSided)) {
+            gprim.CreateDoubleSidedAttr(VtValue(doubleSided), true);
+        }
+    } else {
+        gprim.CreateDoubleSidedAttr(VtValue(doubleSided), true);
     }
 
     bool opposite = false;
--- a/lib/mayaUsd/fileio/translators/translatorGprim.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorGprim.cpp
@@ -42,8 +42,12 @@ void UsdMayaTranslatorGprim::Write(
 {
     MFnDependencyNode depFn(mayaNode);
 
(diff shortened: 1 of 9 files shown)


kxl-adsk:
Thank you @dgovil, this matches the design discussion we had with @womanyoyoyo-adsk. We will start the review.

dgovil:
I just pushed a minor update to clean up the code a bit, and remove an unnecessary include. Also removed my note about the bug in the initial PR comment because it turned out to be something I introduced internally that was resolved by this change, so wasn't of consequence here.

test/lib/usd/translators/testUsdExportMesh.py
```diff
@@ -45,6 +45,7 @@ def _AssertVec3fArrayAlmostEqual(self, arr1, arr2):
                 self.assertAlmostEqual(arr1[i][j], arr2[i][j], places=3)
 
     def testExportAsCatmullClark(self):
+        cmds.file(self.meshTestFile, force=True, open=True)
```
kxl-adsk, line 48: Why did you have to make this change?

kxl-adsk (approved):
Looks very good. Thank you.

test/lib/usd/translators/testUsdExportMesh.py
dgovil: Otherwise it was opening up the file at the setup of the class, and AFAIK running of the tests is non deterministic. So if my test ran in between an existing test and the older ones, I'd have broken them. This made it slower, but safer.

dgovil:
@kxl-adsk I see one test is failing, but I do not believe it's related to my change. Looks to be the usdz packaging complaining about temp files. I assume I can ignore that and it's just some Ci problems?

lib/mayaUsd/fileio/translators/translatorGprim.cpp
```diff
@@ -37,12 +37,17 @@ void UsdMayaTranslatorGprim::Read(
...
+    UsdMayaPrimWriterContext*,
+    GeomSidedness sidedness)
 {
     MFnDependencyNode depFn(mayaNode);
 
-    bool doubleSided = false;
-    if (UsdMayaUtil::getPlugValue(depFn, "doubleSided", &doubleSided)) {
+    bool doubleSided = sidedness == GeomSidedness::Double;
+    if (sidedness == GeomSidedness::Derived) {
+        if (UsdMayaUtil::getPlugValue(depFn, "doubleSided", &doubleSided)) {
```
ppt-adsk, line 47: Am I missing something here?  UsdMayaUtil::getPlugValue() returns true for a successful operation.  The value of the plug is in doubleSided.  Therefore, shouldn't this read
if (UsdMayaUtil::getPlugValue(depFn, "doubleSided", &doubleSided) && doubleSided) 
?

kxl-adsk:
@davidgyu Indeed `testUsdExportPackage` has been showing up randomly lately. We weren't able to isolate the source of it.

lib/mayaUsd/fileio/translators/translatorGprim.cpp
dgovil: I'm trying to preserve the existing behavior where it always writes the attribute, regardless of if it's true or false.
I could change it to only write it if it's true, but there shouldn't be a downside to it one way or another.

ppt-adsk: Thanks for the explanation; it would make a great code comment.

ppt-adsk (approved):

lib/mayaUsd/fileio/translators/translatorGprim.cpp
dgovil: Added a comment to that block

dgovil:
Addressed the notes in https://github.com/Autodesk/maya-usd/pull/1499/commits/af51b7abae853af6d4627169a1f3159db9cdfcb8

ppt-adsk (approved):

PR #1502, opened 2021-06-16
MAYA-111686 Add kSelectPointsForGravity to all selected render items when we should

williamkrick:
be snapping to active objects.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/mayaPrimCommon.h
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mayaPrimCommon.h
@@ -1,0 +1,36 @@
+//
+// Copyright 2021 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef HD_VP2_MAYA_PRIM_COMMON
+#define HD_VP2_MAYA_PRIM_COMMON
+
+#include "pxr/imaging/hd/types.h"
+#include "pxr/imaging/hd/changeTracker.h"
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+struct MayaPrimCommon {
+    enum DirtyBits : HdDirtyBits
+    {
+        DirtySelection = HdChangeTracker::CustomBitsBegin,
+        DirtySelectionHighlight = (DirtySelection << 1),
+        DirtySelectionMode = (DirtySelectionHighlight << 1),
+        DirtyBitLast = DirtySelectionMode
+    };
+};
+
+PXR_NAMESPACE_CLOSE_SCOPE
+
+#endif // HD_VP2_MAYA_PRIM_COMMON
--- a/lib/mayaUsd/render/vp2RenderDelegate/mayaPrimCommon.h
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mayaPrimCommon.h
@@ -16,12 +16,13 @@
 #ifndef HD_VP2_MAYA_PRIM_COMMON
 #define HD_VP2_MAYA_PRIM_COMMON
 
-#include "pxr/imaging/hd/types.h"
 #include "pxr/imaging/hd/changeTracker.h"
+#include "pxr/imaging/hd/types.h"
 
 PXR_NAMESPACE_OPEN_SCOPE
 
-struct MayaPrimCommon {
+struct MayaPrimCommon
+{
     enum DirtyBits : HdDirtyBits
     {
         DirtySelection = HdChangeTracker::CustomBitsBegin,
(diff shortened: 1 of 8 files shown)


lib/mayaUsd/render/vp2RenderDelegate/mayaPrimCommon.h
```diff
@@ -0,0 +1,36 @@
...
+#ifndef HD_VP2_MAYA_PRIM_COMMON
+#define HD_VP2_MAYA_PRIM_COMMON
+
+#include "pxr/imaging/hd/types.h"
+#include "pxr/imaging/hd/changeTracker.h"
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+struct MayaPrimCommon {
+    enum DirtyBits : HdDirtyBits
```
williamkrick, line 25: I need to be able to mark the selection mode dirty for a prim from code where I don't know the prim's type, so make a common place for Vp2RenderDelegate prims to define dirty flags that they all share.

This can be the beginning the refactoring we are hoping to do for HdVP2Mesh and HdVP2BasisCurve.

I'm not sure MayaPrimCommon is the best name, maybe Vp2RenderDelegatePrimCommon? Any other suggestions?

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1159,11 +1159,6 @@ HdDirtyBits HdVP2Mesh::_PropagateDirtyBits(HdDirtyBits bits) const
         bits |= _customDirtyBitsInUse & (DirtySmoothNormals | DirtyFlatNormals);
     }
 
-    // If the topology is dirty, recompute custom indices resources.
```
williamkrick, line 1162: These bits are never used so remove them.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -2014,13 +2009,19 @@ void HdVP2Mesh::_UpdateDrawItem(
 
 #ifdef MAYA_NEW_POINT_SNAPPING_SUPPORT
     if (!isBBoxItem && !isDedicatedSelectionHighlightItem
-        && (itemDirtyBits & DirtySelectionHighlight)) {
+        && (itemDirtyBits & (DirtySelectionHighlight | DirtySelectionMode))) {
         MSelectionMask selectionMask(MSelectionMask::kSelectMeshes);
 
+#ifdef MAYA_SNAP_TO_SELECTED_OBJECTS_SUPPORT
+        if (_selectionStatus == kUnselected || drawScene.SnapToSelectedObjects()) {
```
williamkrick, line 2016: Updating this expression fixes the issue so that when moving a rotate pivot we can point snap to the selected objects points.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -728,9 +729,10 @@ void ProxyRenderDelegate::_Execute(const MHWRender::MFrameContext& frameContext)
         }
 #endif
     } else {
-        if (_selectionChanged) {
+        if (_selectionChanged || _selectionModeChanged) {
```
williamkrick, line 732: The _selectionModeChanged work is deliberately not protected by the preprocessor macro. I'm planning to use it in my next change as well and that one doesn't have any Maya version dependencies.

lib/mayaUsd/render/vp2RenderDelegate/mayaPrimCommon.h
JGamache-autodesk: I think Vp2RenderDelegatePrimCommon would be too long without adding any important info. I like MayaPrimCommon, but since you ask for suggestions, I will dare going to the short extreme: Vp2Prim.

JGamache-autodesk (approved):

williamkrick:
I made a big enough change that I'd like you to look at it again.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -41,9 +42,9 @@
 #endif
 
 // The new Maya point snapping support doesn't require point snapping items any more.
-#if MAYA_API_VERSION >= 20230000
-// The new Maya point snapping support has some known issues. Disable it for now.
-// #define MAYA_NEW_POINT_SNAPPING_SUPPORT
+#if (MAYA_API_VERSION >= 20220200 && MAYA_API_VERSION < 20230000) || (MAYA_API_VERSION >= 20230000 && MAYA_PREVIEW_RELEASE_VERSION >= 127)
+#define MAYA_NEW_POINT_SNAPPING_SUPPORT
+#define MAYA_SNAP_TO_SELECTED_OBJECTS_SUPPORT
```
williamkrick, line 47: The expression here ended up being the same for MAYA_NEW_POINT_SNAPPING_SUPPORT and MAYA_SNAP_TO_SELECTED_OBJECTS_SUPPORT. I was thinking of removing the snap to selected one and simplifying things a bit. I don't see any downside to that, do you?

JGamache-autodesk: Yes, they could be merged.

JGamache-autodesk: Looks like the conditional will require an update in the future. Do you want to future-proof it with:
```
(MAYA_PREVIEW_RELEASE_VERSION >= 127 || MAYA_PREVIEW_RELEASE_VERSION == 0)
```
instead?

JGamache-autodesk (approved):

williamkrick:
Re-review for changes to detection of the new point snapping API function.

JGamache-autodesk (approved):

PR #1511, opened 2021-06-17
[al] ProxyShape: add pauseUpdate attr to pause evaluation

nxkb:
Adds an attribute to the proxy shape to delay any costly updates to the maya viewport representation. This allows users or tools to make a series of usd edits efficiently and only pay once for the final result of all the changes.

We use this internally to pause evaluation as our users select variants from the available options and then only display the end result once the stage is fully configured. ( As a result, for now, this PR only pauses updates due to variant switches...)

Would like to make it pause updates due to other usd changes (not too hard) as well as changes in file path or prim path, too, but that is trickier. We'd be curious to know if Autodesk may have other methods for pausing evaluation that we can tap into to solve this problem. Otherwise, please consider the way we've approached it and how we might get this merged in for others.

#### Example Workflow that this enables:
1.) Open a scene with a shot stage in a proxy shape
2.) Open a pipeline specific tool and point it at the proxy shapes stage
3.) set "pauseUpdates" = True
4.) Change top level variants on a prim (which might reveals nested variants options, sub assets, new prim metadata, etc). This can be very slow if the stage is still pushing updates to the viewport or maya nodes.
5.) Use new info in your pipeline specific tool to continue making variant selections.
6.) close tool, set "pauseUpdates" = False, and allow maya + viewport to catch up and make a SINGLE sparse update to the FINAL state of the Stage.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.h
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.h
@@ -342,6 +342,9 @@ public:
     /// when stage is opened
     AL_DECL_ATTRIBUTE(assetResolverConfig);
 
+    /// Don't update the proxy shape when updates to the usd stage are made
+    AL_DECL_ATTRIBUTE(pauseUpdates);
+
     /// Variant fallbacks if stage was opened and/reopened a Maya scene with custom variants
     /// fallbacks
     AL_DECL_ATTRIBUTE(variantFallbacks);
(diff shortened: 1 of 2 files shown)


kxl-adsk:
Given this change is only modifying AL proxy, I will rely on a review from @fabal and PF.

kxl-adsk:
I missed your last comment where you indicate that you would like to apply this approach to others (proxy shapes). The base proxy shape doesn't have the same features as AL and we are not heading in the direction of having to synchronize Maya's data model with USD. It would be beneficial to profile the same workflows with the Adsk plugin and confirm the bottlenecks.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.h
```diff
@@ -749,6 +762,9 @@ class ProxyShape
         return m_unloadedProxyShapes;
     }
 
+    AL_USDMAYA_PUBLIC
+    static void loadUnloadedProxyShapes();
```
fabal, line 766: The definition is missing.

fabal (commented):
Thanks @nxkb , it looks good to me. If you can just provide the missing method's definition so I can approve.

nxkb:
as discussed we'll just support this behavior for al proxy for now.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.h
nxkb: sorry about that, it looks like it was just an accidental add part of another old feature that is no longer needed. Removed!

fabal (approved):
Thanks @nxkb 👍

PR #1515, opened 2021-06-21
Use SdfComputeAssetPathRelativeToLayer to compute sublayer paths

dj-mcg:
This replaces custom logic that tried to anchor sublayer
paths to the parent layer using direct calls to ArResolver.
USD uses the Sdf function during composition, so using
it will match USD's behavior. This also allows us to
remove the use of ArResolver API that is deprecated
under Ar 2.0.

### DIFF
--- a/lib/mayaUsd/utils/utilSerialization.cpp
+++ b/lib/mayaUsd/utils/utilSerialization.cpp
@@ -20,7 +20,7 @@
 #include <mayaUsd/utils/utilFileSystem.h>
 
 #include <pxr/base/tf/stringUtils.h>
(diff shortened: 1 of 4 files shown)


seando-adsk (approved):
Everything looks good. Just to confirm, this USD layer utils function you are using is available in all our supported USD versions? I looked at the USD repo and I think the answer is yes.

dj-mcg:
> Everything looks good. Just to confirm, this USD layer utils function you are using is available in all our supported USD versions? I looked at the USD repo and I think the answer is yes.

Yep! SdfFindOrOpenRelativeToLayer was added in 2016.

Issue #1417, opened 2021-05-16
[MAYA-112059] USD stage displaying textures incorrectly.


daviddeacon:
Hi,

I'm testing out converting Maya assets to USD, I'm using the standard surface material. I'm keeping it simple and just connecting a sRGB base colour map. I then export from Maya to USD, with the material's option set to 'USD Preview Surface'

When reading in the USD via a USD stage, the colour map displays incorrectly, it looks like it might just be a colour space issue. I tried to match the USD source's texture colour space to what I was seeing on the stage but nothing matched.

I also tried exporting with basic Lambert material assigned, it also had the same issue with being too dark with a red hue.

Viewport screenshot - 
<img width="1852" alt="previewSurfaceColourSpace" src="https://user-images.githubusercontent.com/60041784/118398012-a06af300-b64e-11eb-951d-68abccf5260e.png">
[https://www.dropbox.com/s/pzv7fdx5ku12le8/previewSurfaceColourSpace.png?dl=0](url)

Scene file - 
[https://www.dropbox.com/sh/nxhy4s2h0ud64rf/AACMz1wpW9SnDQfNPEHN6i3ya?dl=0](url)

macOS Big Sur, 11.2.3
Maya USD 0.9.0
Maya 2022

JGamache-autodesk:
The USD stage uses USD's image loader and tries to color correct the texture data from sRGB to ACEScg using the following code:
https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/render/vp2ShaderFragments/UsdUVTexture.xml#L63
Deactivating that code gives back the equivalent of "raw" in Maya, so this is where the redshift happens.
I will have to consult with our color management experts to see if a better transform exists that would replicate what we get with the Maya image handlers.

JGamache-autodesk:
Logged internally as MAYA-112059

PR #1516, opened 2021-06-21
MAYA-112059 - UsdUVTexture works with Maya color management prefs

JGamache-autodesk:
Maya versions before 2022 had a default color management rendering space of "scene-linear Rec 709/sRGB", which means converting a sRGB texture required only the gamma correction part,
With Maya 2022, we now use the "ACEScg" rendering color space, so an extra multiplication by a color correction matrix is required to move from linrec709 to ACEScg.
This is a user preference, though, and one that is saved with the scene. So we need to query the current rendering space and use the correct color correction.

This is the list of supported rendering color spaces:
- scene-linear Rec 709/sRGB
- ACEScg
- ACES2065-1
- scene-linear DCI-P3 D65
- scene-linear Rec.2020

Note that the VP2 render delegate does not register for notifications when the rendering color space is changed, so if you change the rendering color space, you should save the scene and reload to force the USD stages materials to be updated with the correct color space transformation.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -873,6 +873,10 @@ void HdVP2Material::_ApplyVP2Fixes(HdMaterialNetwork& outNet, const HdMaterialNe
     // Get the shader registry so I can look up the real names of shading nodes.
     SdrRegistry& shaderReg = SdrRegistry::GetInstance();
 
+    // We might need to query the working color space of Maya if we hit texture nodes. Delay
+    // the query until necessary.
+    MString mayaWorkingColorSpace;
+
     // Replace the authored node paths with simplified paths in the form of "node#". By doing so
     // we will be able to reuse shader effects among material networks which have the same node
     // identifiers and relationships but different node paths, reduce shader compilation overhead
@@ -889,7 +893,19 @@ void HdVP2Material::_ApplyVP2Fixes(HdMaterialNetwork& outNet, const HdMaterialNe
         // everything to use the SdrShaderNode name.
         SdrShaderNodeConstPtr sdrNode
             = shaderReg.GetShaderNodeByIdentifierAndType(outNode.identifier, _tokens->glslfx);
-        outNode.identifier = TfToken(sdrNode->GetName());
+
+        if (_IsUsdUVTexture(node)) {
+            // We need to rename according to the Maya color working space pref:
+            if (!mayaWorkingColorSpace.length()) {
+                // Query the user pref:
+                mayaWorkingColorSpace = MGlobal::executeCommandStringResult(
+                    "colorManagementPrefs -q -renderingSpaceName");
+            }
+            outNode.identifier = TfToken(
+                HdVP2ShaderFragments::getUsdUVTextureFragmentName(mayaWorkingColorSpace).asChar());
+        } else {
+            outNode.identifier = TfToken(sdrNode->GetName());
+        }
 
         if (_IsUsdDrawModeNode(outNode)) {
             // I can't easily name a Maya fragment something with a '.' in it, so pick a different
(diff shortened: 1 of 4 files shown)


JGamache-autodesk:
@hodoulp, could you review this one, please!
Fixes #1417

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1069,7 +1074,21 @@ void HdVP2Material::_ApplyVP2Fixes(HdMaterialNetwork& outNet, const HdMaterialNe
         }
 
         addPredecessorNodes(node);
-        outNet.nodes.push_back(node);
+        if (_IsUsdUVTexture(node)) {
```
williamkrick, line 1077: I think this code fits better with the code around line 880. That code is already renaming some other nodes, so renaming the texture nodes fits better there then it does here.

lib/mayaUsd/render/vp2ShaderFragments/shaderFragments.cpp
```diff
@@ -330,6 +344,94 @@ MStatus HdVP2ShaderFragments::registerFragments()
...
+
+        const std::regex RE_FRAG("UsdUVTexture");
+        const std::regex RE_GLSL("TO_MAYA_COLOR_SPACE_GLSL");
+        const std::regex RE_HLSL("TO_MAYA_COLOR_SPACE_HLSL");
+        const std::regex RE_CG("TO_MAYA_COLOR_SPACE_CG");
+
+        auto registerSpace = [&](const std::string& spaceName,
+                                 const std::string& fragName,
+                                 const float*       convMatrix) {
+            if (!fragmentManager->hasFragment(fragName.c_str())) {
```
williamkrick, line 367: Please add a comment here which shows what the generated code looks like.

williamkrick (changes requested):

williamkrick (approved):

lib/mayaUsd/render/vp2ShaderFragments/shaderFragments.cpp
```diff
@@ -330,6 +344,121 @@ MStatus HdVP2ShaderFragments::registerFragments()
...
+                std::string xmlFinal = std::regex_replace(xmlString, RE_FRAG, fragName.c_str());
+                if (convMatrix) {
+                    std::stringstream op_glsl, op_hlsl, op_cg;
+                    op_glsl << "outColor = mat4(";
+                    op_hlsl << "outColor = mul(outColor, float4x4(";
+                    op_cg << "outColor = mul(float4x4(";
```
hodoulp, line 387: I have some concerns about the matrix implementation which is different from the OpenColorIO one (refer to 
https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/src/OpenColorIO/GpuShaderUtils.cpp) where the transpose is done for HLSL only!

When I did some search on the Matrix access for GLSL, HLSL and Cg, I find that HLSL is column first, and GLSL & Cg are raw first!

hodoulp (changes requested):

hodoulp:
[Global Comment] As discussed this implementation is a short-term solution to fix the current problem, but it has some major limitations:
1. It will fail if the config file is not one of the Maya default. That's a common use case for big studios.
2. The long-term approach must be to use OpenColorIO directly to benefit from a complete Color Management support.

lib/mayaUsd/render/vp2ShaderFragments/shaderFragments.cpp
JGamache-autodesk: Yes, this is a very complex issue. You need also to consider that the OpenGL code does mat4 * vec4 while the HLSL code does vec4 * mat4.
When in doubt, the best option is to see what code OCIO generates and replicate that exactly, so yes, I do admit I cheated and ran Maya thru a debugger to see the exact shader code it generated for each shading language and replicated it exactly (with an eye for transposed data).
To test: use the most asymmetric transform "scene-linear DCI-P3 D65" and see what happens in Maya. The color will either be correct, or completely off.

hodoulp: Good point. You can copy & paste OCIO for now. You already have the link in the comment.

PR #1519, opened 2021-06-23
MAYA-110878 - handle multiple deferred notifs for single path

seando-adsk:
#### MAYA-110878 - MayaUSD sends some UFE attribute changed notifications for attributes which don't exist, and drops some notifications

* Handle multiple deferred notifications for a single path (keeping the original notif order).

### DIFF
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -72,10 +72,12 @@ bool inAttributeChangedNotificationGuard()
     return attributeChangedNotificationGuardCount.load() > 0;
 }
 
-// TODO: This should be an unordered_multimap to prevent notifications from
-// overwriting earlier recorded notifications for the same attribute. See
-// MAYA-110878 for more information on why this change isn't already made.
-std::unordered_map<Ufe::Path, TfToken> pendingAttributeChangedNotifications;
+// Keep an array of the pending attribute change notifications. Using a vector
+// for two main reasons:
+// 1) Order of notifs must be maintained.
+// 2) Allow notif with same path, but different token. At worst the check is linear
+//    in the size of the vector (which is the same as an unordered_multimap).
+std::vector<std::pair<Ufe::Path, TfToken>> pendingAttributeChangedNotifications;
 
 void sendValueChanged(const Ufe::Path& ufePath, const TfToken& changedToken)
 {
@@ -90,7 +92,15 @@ void sendValueChanged(const Ufe::Path& ufePath, const TfToken& changedToken)
 void valueChanged(const Ufe::Path& ufePath, const TfToken& changedToken)
 {
     if (inAttributeChangedNotificationGuard()) {
-        pendingAttributeChangedNotifications[ufePath] = changedToken;
+        // Don't add pending notif if one already exists with same path/token.
+        auto p = std::make_pair(ufePath, changedToken);
+        if (std::find(
+                pendingAttributeChangedNotifications.begin(),
+                pendingAttributeChangedNotifications.end(),
+                p)
+            == pendingAttributeChangedNotifications.end()) {
+            pendingAttributeChangedNotifications.emplace_back(p);
+        }
     } else {
         sendValueChanged(ufePath, changedToken);
     }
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -72,10 +72,12 @@ bool inAttributeChangedNotificationGuard()
...
-// TODO: This should be an unordered_multimap to prevent notifications from
-// overwriting earlier recorded notifications for the same attribute. See
-// MAYA-110878 for more information on why this change isn't already made.
-std::unordered_map<Ufe::Path, TfToken> pendingAttributeChangedNotifications;
+// Keep an array of the pending attribute change notifications. Using a vector
+// for two main reasons:
+// 1) Order of notifs must be maintained.
+// 2) Allow notif with same path, but different token. At worst the check is linear
+//    in the size of the vector (which is the same as an unordered_multimap).
+std::vector<std::pair<Ufe::Path, TfToken>> pendingAttributeChangedNotifications;
```
seando-adsk, line 80: Change is pretty simple. I just changed the container to vector (for reasons give here). I ran all the UFE tests and they all passed with no test changes required. I believe this is due to the previous changes around line 420 which ignore inert prim changes.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -242,11 +252,7 @@ void StagesSubject::stageChanged(
                     Ufe::Transform3d::notify(ufePath);
                 }
             }
-#ifdef UFE_V2_FEATURES_AVAILABLE
-            if (!inAttributeChangedNotificationGuard()) {
-                sendValueChanged(ufePath, changedPath.GetNameToken());
-            }
-#endif
+            UFE_V2(valueChanged(ufePath, changedPath.GetNameToken());)
```
seando-adsk, line 255: Use the helper func which checks the notif guard. The previous code would also not add this value changed to the deferred notifs if we were in a guard.

lib/mayaUsd/ufe/StagesSubject.cpp
williamkrick: I think that vector will not be fast enough for the duplicate notification elimination at line 95. While the worst case is the same, the average case or unordered_map is constant while the average case the vector is linear in the size of the vector.

I think to do better you can have a second container which is a std::unordered_set< std::pair<Ufe::Path, TfToken> >. The idea is you use the set to test if the std::pair<Ufe::Path, TfToken> has already been saved in the vector in constant time, and you use the vector to maintain ordering. We trade off a little constant time & some space, but we avoid the linear search.

williamkrick (changes requested):

lib/mayaUsd/ufe/StagesSubject.cpp
seando-adsk: I do see your point, but I wonder if its really needed. I added some tf_debug statements to print out when a notif gets delayed. I ran all the "UFE" tests and the largest size the pendingAttributeChangedNotifications vector got was only 2. Granted the tests aren't real-world data, but I think that when normally running the notif guard will expire and send the pending notifs long before they ever have a chance to build up. @ppt-adsk What do you think?

ppt-adsk: Hard to say, I tend to think @seando-adsk you're right, but it's a code complexity vs. performance guarantee tradeoff.  Can we perhaps put up a message if the vector size grows above some threshold?

williamkrick: The case I'm worried about is someone writing a script that iterates over everything in the scene and changes each one a little bit, say changing the purpose of each rprim. That is going to create N messages and have O(N^2) complexity because it is doing a linear search.

If you really don't like the unordered_set we can leave it as-is and ask QA to play around with some heavy scenes and see if they hit any performance issues.

seando-adsk: If you look [here](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/ufe/UsdAttribute.cpp#L54) you can see where we use the notif guard - basically when setting a single usd attribute. In your scenario on each prim we change a little bit we will enter/leave a notif guard. So I don't ever see this notif guard growing beyond just a few notifs.

williamkrick: Okay sounds good to me!

williamkrick (approved):

PR #1520, opened 2021-06-23
Updates from code review of PR 1516.

JGamache-autodesk:
Adds recommendation to integrate OCIO in a folow-up task.
Adds old synColor names to support Maya < 2022.
Simplifies the code and adds proper message if scene working space is not supported.


JGamache-autodesk:
@hodoulp here are the updates you requested in PR #1516

lib/mayaUsd/render/vp2ShaderFragments/shaderFragments.cpp
```diff
@@ -345,6 +350,15 @@ MStatus HdVP2ShaderFragments::registerFragments()
     }
 
     {
+        // This is a temporary fix for Maya 2022 that is quite fragile and will need to be revisited
+        // in the near future. It will not handle the custom color space some large scale clients
+        // use. It will also not handle any custom OCIO config that could change the color
+        // interpolation algorithm.
```
hodoulp, line 356: Below is my suggestion for the comment to add:

Note: Because the supported color spaces are hard-coded (i.e names and color transformation implementation) it can only support few color spaces from the Maya default configs and it will definitively fail on any custom configs. So, this is a temporary fix for Maya that is quite fragile and will need to be revisited in the near future.

lib/mayaUsd/render/vp2ShaderFragments/shaderFragments.cpp
```diff
@@ -375,28 +389,30 @@ MStatus HdVP2ShaderFragments::registerFragments()
...
-                        op_cg << convMatrix[i] << ", " << convMatrix[3 + i] << ", "
-                              << convMatrix[6 + i] << ", " << 0.0 << ", ";
+                        op_cg << convMatrix[i] << ", " << convMatrix[4 + i] << ", "
+                              << convMatrix[8 + i] << ", " << convMatrix[12 + i]
+                              << ((i < 3) ? ", " : "");
                     }
-                    op_glsl << 0.0 << ", " << 0.0 << ", " << 0.0 << ", " << 1.0 << ") * outColor;";
-                    op_hlsl << 0.0 << ", " << 0.0 << ", " << 0.0 << ", " << 1.0 << "));";
-                    op_cg << 0.0 << ", " << 0.0 << ", " << 0.0 << ", " << 1.0 << "), outColor);";
+                    op_cg << "), outColor);";
```
hodoulp, line 415: I still have investigations to perform about the matrix layout. But you can use the OpenColorIO implementation to at least have the same 'result' than OpenColorIO v2.
Refer to https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/src/OpenColorIO/GpuShaderUtils.cpp#L671

hodoulp (changes requested):

lib/mayaUsd/render/vp2ShaderFragments/shaderFragments.cpp
JGamache-autodesk: The result is the same because the color transformation matrices used here are already pre-transposed to Dx11/OpenGL order. 

This allows making Cg the odd one out requiring back-transposition, which is OK since that language was deprecated nine years ago.

hodoulp: > The result is the same because the color transformation matrices used here are already pre-transposed to Dx11/OpenGL order.

Could you add a comment to mention that matrices are already transposed because that's really not clear in the code?

hodoulp:
LGTM, waiting for a last comment before approving.

lib/mayaUsd/render/vp2ShaderFragments/shaderFragments.cpp
JGamache-autodesk: Done.

williamkrick (approved):

PR #1526, opened 2021-06-29
MAYA-112226 - Stage to be created without the proxyShape being capitalized

seando-adsk:
#### MAYA-112226 - Stage to be created without the proxyShape being capitalized

* [GitHub] #1379
* Simple fix to remove the capitalization.

### DIFF
--- a/plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
+++ b/plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
@@ -117,7 +117,7 @@ global proc stageFromFile_UICommit(string $parent) {
 
 proc string doCreateStage(string $fileName) {
     // actually load the file
-    string $baseName = capitalizeString(`basenameEx $fileName`);
+    string $baseName = basenameEx($fileName);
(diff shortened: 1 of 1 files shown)


plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
```diff
@@ -117,7 +117,7 @@ global proc stageFromFile_UICommit(string $parent) {
 
 proc string doCreateStage(string $fileName) {
     // actually load the file
-    string $baseName = capitalizeString(`basenameEx $fileName`);
+    string $baseName = basenameEx($fileName);
```
seando-adsk, line 120: I removed the capitalization from the mayaUsd plugin only. The Pixar plugin still has the capital code.

kxl-adsk (approved):
Do we have a regression test that validates this Mel procedure? If not, would be good to add one.

seando-adsk:
Yes we have a test: testMayaUsdCreateStageCommands.py. That is why I'll run the preflight now.

PR #1527, opened 2021-06-29
MAYA-111478 - [I] I'd like to see categories persist their expand/collapse state

seando-adsk:
#### MAYA-111478 - [I] I'd like to see categories persist their expand/collapse state

* Original change integrated back to Maya 2022 update.

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
@@ -358,7 +358,7 @@ class AETemplate(object):
         self.createMetadataSection()
         cmds.editorTemplate(endScrollLayout=True)
 
-        if int(cmds.about(majorVersion=True)) > 2022:
+        if ('%s.%s' % (cmds.about(majorVersion=True), cmds.about(minorVersion=True))) > '2022.1':
             # Because of how we dynamically build the Transform attribute section,
             # we need this template to rebuild each time it is needed. This will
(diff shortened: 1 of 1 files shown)


seando-adsk:
@kxl-adsk This change is needed to go with some changes that were backported in the Maya update branch. Once approved please merge. No preflight needed as this change is in the prim template file that currently has no test(s).

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -358,7 +358,7 @@ def __init__(self, ufeSceneItem):
         self.createMetadataSection()
         cmds.editorTemplate(endScrollLayout=True)
 
-        if int(cmds.about(majorVersion=True)) > 2022:
+        if ('%s.%s' % (cmds.about(majorVersion=True), cmds.about(minorVersion=True))) > '2022.1':
```
kxl-adsk, line 361: I don’t understand. You are making greater than operation on strings?

seando-adsk: Yes because I need to test the Maya update (minor) version. Why? Do you think there is a case where it won't work correctly?

kxl-adsk: Wasn't expecting that to work...but looks like it's working without issues in python. Learned something new.

kxl-adsk (approved):

PR #1528, opened 2021-06-29
MAYA-112227 - UV set renamed to st, st1, st2

JGamache-autodesk:
Implements automatic UV set renaming.
- The sets will be renamed st, st1, st2, st3,... in the order they are found in the mesh.
- The original name will be stored in custom data for roundtripping
- On import, the names will always be restored if the roundtripping information is found
- On import, the UV sets are restored in the same order as found in the USD primitive
- The environment variables MAYAUSD_IMPORT_PRIMARY_UV_SET_AS_MAP1 and  MAYAUSD_EXPORT_MAP1_AS_PRIMARY_UV_SET were removed as their functionnality is subsumed in the global renaming.

Caveat:
If you have a mesh with multiple UV sets, and a UV set named "map1" is not the first UV set, then it will not automatically be renamed "st" as was previously the case using MAYAUSD_EXPORT_MAP1_AS_PRIMARY_UV_SET. We expect these cases to be quite rare, but if you have post-processing scripts that are expecting "st" and scenes where "map1" is not the first UV sets; then the results at the other end of the post-processing pipeline will be different.

### DIFF
--- a/test/lib/usd/translators/UsdExportUVSetMappingsTest/UsdExportUVSetMappingsTest.ma
+++ b/test/lib/usd/translators/UsdExportUVSetMappingsTest/UsdExportUVSetMappingsTest.ma
@@ -1,6 +1,6 @@
 //Maya ASCII 2018 scene
 //Name: UsdExportUVSetMappingsTest.ma
-//Last modified: Wed, Nov 04, 2020 02:45:47 PM
+//Last modified: Tue, Jun 29, 2021 11:29:45 AM
 //Codeset: 1252
 requires maya "2018";
 currentUnit -l centimeter -a degree -t film;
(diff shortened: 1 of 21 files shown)


test/lib/usd/translators/UsdExportUVSetMappingsTest/UsdExportUVSetMappingsTest.ma
```diff
@@ -148,7 +173,7 @@ createNode transform -n "pPlane3";
 	rename -uid "A9516DAE-46FB-C555-7036-F5BFA863699B";
 	setAttr ".t" -type "double3" 2.4 0 0 ;
 	setAttr ".r" -type "double3" 89.999999999999986 0 0 ;
-createNode mesh -n "pPlaneShape3" -p "pPlane3";
+createNode mesh -n "pPlane3" -p "|pPlane3";
```
JGamache-autodesk, line 176: @dgovil , @ysiewappl This makes the shape name identical to the transform name and makes sure the code in _UVMappingManager is still correctly handling that situation.

test/lib/usd/translators/testUsdExportUVSetMappings.py
```diff
@@ -67,24 +64,15 @@ def testExportUVSetMappings(self):
...
         '''
         expected = [
-            ("/pPlane1", "/blinn1SG_map1_map1_map1", "map1", "map1", "map1"),
-            ("/pPlane2", "/blinn1SG", "st1", "st1", "st1"),
-            ("/pPlane3", "/blinn1SG", "st1", "st1", "st1"),
-            ("/pPlane4", "/blinn1SG_st2_st2_st2", "st2", "st2", "st2"),
-            ("/pPlane5", "/blinn1SG_p5a_p5b_p5c", "p5a", "p5b", "p5c"),
-            ("/pPlane6", "/blinn1SG_p62_p63_p61", "p62", "p63", "p61"),
-            ("/pPlane7", "/blinn1SG_p7r_p7p_p7q", "p7r", "p7p", "p7q"),
+            ("/pPlane1", "/blinn1SG", "st", "st", "st"),
```
JGamache-autodesk, line 67: The automatic renaming does reduce the number of specializations created.

kxl-adsk:
@dj-mcg Can you help review this change and confirm that it works for Pixar as well?

kxl-adsk (approved):
LGTM!

dj-mcg:
This looks good to me, and I believe it allows me to delete one of our internal patches. I'll fix up whatever internal tests are affected, so lgtm!

dj-mcg (approved):

PR #1530, opened 2021-06-29
Pr/remove calls to configure resolver for asset under ar2

dj-mcg:
This method has been deprecated and is being removed for Ar 2.0.
Among other reasons, the resolver is a global resource and a
method for configuring it for a single asset is misleading
since the resolver may be used by many different assets at
the same time. If a resolver has asset-specific resolution
behaviors, they should be encoded in the resolver context
created via ArResolver::CreateDefaultContextForAsset.

The AL_USDMaya plugin had previously subverted this function
to configure its resolver by passing in an arbitrary string
to ConfigureResolverForAsset. This has been removed as
part of this change, but this could be replaced in the
future with ArResolver::CreateContextFromString, which
allow a resolver to create a resolver context from an
arbitrary string and was introduced in Ar 2.0.

### DIFF
--- a/lib/mayaUsd/utils/utilFileSystem.cpp
+++ b/lib/mayaUsd/utils/utilFileSystem.cpp
@@ -27,6 +27,28 @@
 
 #include <ghc/filesystem.hpp>
 
+#include <random>
+#include <system_error>
+namespace {
+std::string generateUniqueName()
+{
+    const auto  len { 6 };
+    std::string uniqueName;
+    uniqueName.reserve(len);
+
+    const std::string alphaNum { "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz" };
+
+    std::random_device              rd;
+    std::mt19937                    gen(rd());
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -27,35 +27,14 @@
 
 #include <ghc/filesystem.hpp>
 
-#include <random>
-#include <system_error>
-
-namespace {
-std::string generateUniqueName()
```
kxl-adsk, line 34: @dj-mcg How this change is related to AR 2.0?

dj-mcg: Yep - good catch - an internal patched change snuck in with my push. I'll be pushing a new version momentarily

kxl-adsk (approved):
I see this matches changes to https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usdImaging/usdviewq/__init__.py#L308. No concerns from my side but I will let @fabal review changes to AL proxy.

fabal (approved):
Thanks @dj-mcg 👍

PR #1550, opened 2021-07-05
MAYA-112372: Add restriction for grouping.

HamedSabri-adsk:
group_blocking

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoCreateGroupCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoCreateGroupCommand.cpp
@@ -79,24 +79,36 @@ void UsdUndoCreateGroupCommand::execute()
         setKindCmd->execute();
     }
 
-    auto newParentHierarchy = Ufe::Hierarchy::hierarchy(_group);
-    if (newParentHierarchy) {
-        for (auto child : _selection) {
-            auto parentCmd = newParentHierarchy->appendChildCmd(child);
-            parentCmd->execute();
-            append(parentCmd);
+    // Make sure to handle the exception if the parenting operation fails.
+    // This scenario happens if a user tries to group prim(s) in a layer
+    // other than the one where they were defined. In this case, the group creation itself
+    // will succeed, however the re-parenting is expected to throw an exception. We also need to
+    // make sure to undo the previous command ( AddNewPrimCommand ) when this happens.
+    try {
+        auto newParentHierarchy = Ufe::Hierarchy::hierarchy(_group);
+        if (newParentHierarchy) {
+            for (auto child : _selection) {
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/ufe/UsdUndoCreateGroupCommand.cpp
```diff
@@ -79,24 +79,36 @@ void UsdUndoCreateGroupCommand::execute()
...
+    } catch (...) {
+        // undo previous AddNewPrimCommand
+        undo();
 
-    TF_VERIFY(
-        Ufe::GlobalSelection::get()->size() == 1,
-        "_group node should be in the global selection now. \n");
+        throw; // re-throw the same exception
+    }
 }
```
HamedSabri-adsk, line 113: @seando-adsk 
Grouping in a two tier operation:

1- create a group node
2- insert selected item(s) under the group node.

There could be a case where creating a group node would be Ok by the restriction rules but not the insertion part.
For example, this scenario could happen if a user tries to group prim(s) in a layer other than the one where they were defined. In this case, the group creation itself will succeed but not the insertion part. In this case, we should throw an exception and undo the previous group command. 

Also keep in mind that the way we are performing the insertion here makes it impossible to support grouping flags (e.g absolute, relative, etc...) . Eventually, the insertion logics needs to be moved in Maya so we can support those flags.

HamedSabri-adsk:
@seando-adsk for Maya's side of changes see MAYA-112372

test/lib/ufe/testGroupCmd.py
```diff
@@ -241,6 +241,75 @@ def testGroupKind(self):
         self.assertEqual(Usd.ModelAPI(newGroupPrim).GetKind(), "")
         self.assertFalse(newGroupPrim.IsModel())
 
+    @unittest.skipUnless(mayaUtils.mayaMajorVersion() >= 2022, 'Grouping restriction is only available in Maya 2022 or greater.')
```
seando-adsk, line 244: Does this really on Maya version, or is does it need UFE v3?

HamedSabri-adsk: There is no UFE v3 required for changes in this PR.   The Maya side of changes which captures this exception is now in both master and release/2022 branch.

HamedSabri-adsk: Also as a note, merge order doesn't really matter in the case of this PR.

seando-adsk (approved):

HamedSabri-adsk:
@kxl-adsk  I am marking this PR as ready to merge. 

I ran the PF before and it all passed but one of the interactive fail which is not relevant here.

https://github.com/Autodesk/maya-usd/actions/runs/1006156994

PR #1552, opened 2021-07-06
MAYA-107463 Make sure we leave default material mode when the test ends. Check fo…

williamkrick:
…r UFE version correctly.

Discovered this while working on MAYA-107463, decided to split out the useful changes because that work is held up.


lib/mayaUsd/CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ if (MAYA_API_VERSION VERSION_GREATER_EQUAL 20220100 AND MAYA_API_VERSION VERSION
     )
 endif()
 
-if (MAYA_API_VERSION VERSION_GREATER_EQUAL 20230000 AND UFE_VERSION VERSION_GREATER_EQUAL 2.0.3)
+if (MAYA_API_VERSION VERSION_GREATER_EQUAL 20230000 AND UFE_VERSION VERSION_GREATER_EQUAL 0.3.0)
```
williamkrick, line 72: Building with master I discovered this expression was wrong.

test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegatePerInstanceInheritedData.py
```diff
@@ -55,6 +55,11 @@ def setUpClass(cls):
 
         cls._testDir = os.path.abspath('.')
 
+    @classmethod
+    def tearDownClass(cls):
+        panel = mayaUtils.activeModelPanel()
+        cmds.modelEditor(panel, edit=True, useDefaultMaterial=False)
+
```
williamkrick, line 62: If testInstanceDefaultMaterial fails the image test while useDefaultMaterial is enabled the rest of the script doesn't run to turn it off. I have to set useDefaultMaterial false in the tearDownClass so other scripts will not have it on, and also at the start of the other test runs within this script to ensure they don't have it on.

JGamache-autodesk (approved):

PR #1553, opened 2021-07-06
MAYA-111902 Fix MaterialX import errors found by testing

JGamache-autodesk:
Fixes roughness not exporting correctly for UsdPreviewSurface node in MaterialX
Fixes UsdPreviewSurface not importing from MaterialX
Adds MaterialX mode to UsdPreviewSurface rountrip unit test.


kxl-adsk (approved):

lib/mayaUsd/CMakeLists.txt
```diff
@@ -62,14 +62,7 @@ if(DEFINED UFE_PREVIEW_VERSION_NUM)
...
 
-if (MAYA_API_VERSION VERSION_GREATER_EQUAL 20220100 AND MAYA_API_VERSION VERSION_LESS 20230000)
-    target_compile_definitions(${PROJECT_NAME}
-        PRIVATE
-            HAS_DEFAULT_MATERIAL_SUPPORT_API=1
-    )
-endif()
-
-if (MAYA_API_VERSION VERSION_GREATER_EQUAL 20230000 AND UFE_VERSION VERSION_GREATER_EQUAL 2.0.3)
+if (MAYA_DEFAULT_MATERIAL_API EQUAL 1)
```
JGamache-autodesk, line 65: Some recent build breaks have cast doubt on these combinations of versions. FindMaya.cmake now makes sure the API is correctly detected.

cmake/modules/FindMaya.cmake
```diff
@@ -331,6 +332,30 @@ if (MAYA_OGSDEVICES_LIBRARY)
...
+        "Maya's ${MAYA_LIB} library path"
+    # NO_CMAKE_SYSTEM_PATH needed to avoid conflicts between
+    # Maya's Foundation library and OSX's framework.
+    NO_CMAKE_SYSTEM_PATH
+)
+if (MAYA_OPENMAYARENDER_LIBRARY)
+    file(STRINGS ${MAYA_OPENMAYARENDER_LIBRARY} HAS_DEFAULT_MATERIAL_API REGEX "setDefaultMaterialHandling")
+    if (HAS_DEFAULT_MATERIAL_API)
+        set(MAYA_DEFAULT_MATERIAL_API 1)
+        message(STATUS "This version of Maya has the default material API")
```
williamkrick, line 355: I always like the log to print the name of the env var as well, so I can ctrl+f in the log to check the value that was used.

cmake/modules/FindMaya.cmake
```diff
@@ -331,6 +332,30 @@ if (MAYA_OGSDEVICES_LIBRARY)
...
+        lib/
+        bin/
+    DOC
+        "Maya's ${MAYA_LIB} library path"
+    # NO_CMAKE_SYSTEM_PATH needed to avoid conflicts between
+    # Maya's Foundation library and OSX's framework.
+    NO_CMAKE_SYSTEM_PATH
+)
+if (MAYA_OPENMAYARENDER_LIBRARY)
+    file(STRINGS ${MAYA_OPENMAYARENDER_LIBRARY} HAS_DEFAULT_MATERIAL_API REGEX "setDefaultMaterialHandling")
```
williamkrick, line 352: This is a good idea and I'm wondering if we should use it more generally. Does it take much time to run the search?

williamkrick (approved):

cmake/modules/FindMaya.cmake
JGamache-autodesk: I added a bit more output at the two key detection moments.

cmake/modules/FindMaya.cmake
JGamache-autodesk: It was quite fast, but it will be even faster (and more precise) to look directly in the header file. Fixed that.

williamkrick (approved):

williamkrick (approved):

PR #1554, opened 2021-07-06
MAYA-111834 - Crash when transforming multiple objects with duplicate stage

seando-adsk:
… stage

* Replaced TF_AXIOM (which abort when fail) with TF_VERIFY to allow Add{Scale/Translate/Rotate} calls that already exist to return without crashing.

### DIFF
--- a/lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
+++ b/lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
@@ -433,7 +433,6 @@ UsdTransform3dMayaXformStack::translateCmd(double x, double y, double z)
 Ufe::RotateUndoableCommand::Ptr
 UsdTransform3dMayaXformStack::rotateCmd(double x, double y, double z)
 {
-    GfVec3f        v(x, y, z);
     UsdGeomXformOp op;
     TfToken        attrName;
     const bool     hasRotate = hasOp(NdxRotate);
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
```diff
@@ -657,8 +667,20 @@ Ufe::SetVector3dUndoableCommand::Ptr UsdTransform3dMayaXformStack::setVector3dCm
...
+                // error and return an empty UsdGeomXformOp (since a TranslateOp already exists).
+                // Therefore we detect (and allow) this case.
+                // TODO - What is the expected behavior in this case? Should the object move twice?
                 auto op = xformable.AddTranslateOp(OpPrecision<V>::precision, opSuffix);
-                TF_AXIOM(op);
+                if (!TF_VERIFY(
+                        op,
+                        "TranslateOp already exists for %s",
+                        cmd.sceneItem()->path().string().c_str()))
+                    return UsdGeomXformOp();
```
seando-adsk, line 683: You'll see on lines 631-633 it checks for the existence of the attribute and then creates a OpFunc (from lambda) for either using existing attribute op, or creating one first (this second lamdba) and then using the newly created one. The AddTranslateOp will output an error message if one already exists and return an empty UsdGeomXformOp. The previous test at line 661 would then abort because of the empty op. I changed it to a TF_VERIFY which just prints a message and then returns empty op. So crash is avoided.

lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
```diff
@@ -476,8 +476,13 @@ UsdTransform3dMayaXformStack::rotateCmd(double x, double y, double z)
...
               }
 
+              // See comment in setVector3dCmd().
               auto r = xformable.AddRotateXYZOp(UsdGeomXformOp::PrecisionFloat, opSuffix);
-              TF_AXIOM(r);
+              if (!TF_VERIFY(
+                      r,
+                      "RotateXYZOp already exists for %s",
+                      usdSceneItem->path().string().c_str()))
+                  return UsdGeomXformOp();
```
seando-adsk, line 485: See my comment down below. Same change made for Scale/Rotate/Translate/Pivot.

lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
kxl-adsk: We will fall into this situation only on the first manipulation when ops are missing. Every consecutive manipulation will perform double edit. 

We can make the behavior consistent and re-query the op when the addition failed. Why did you choose to issue a warning instead?

seando-adsk: Because I didn't know what the expected behavior is supposed to be. You are right about the consecutive manips will be double - I'm not sure this is expected either.

seando-adsk: @ppt-adsk What do you think?

ppt-adsk: @kxl-adsk I am not sure I understand your comment.  The core issue is the fact that two UFE paths refer to the same underlying USD object.  If you create n UFE scene items for a Maya command that have this property, then you will get n edits (2 in the case of a duplicated stage).  It is difficult for this mayaUsd code to identify this situation --- we would be violating separation of concerns, IMO.  That is for the double edit part of the discussion.

For the add / don't add a transform op part of the discussion, I'm not sure what consistent behavior is.  On command creation we determined there was no transform op, and on command execution a transform op appeared.  For unique objects, this would be a bug caused by invalid operation of the software.  In the current case, with multiple UFE scene items pointing to the same USD prim, we can indeed deal with the problem by assuming all is well and querying the op --- and the result will be a double edit.  I believe users will find this unexpected.

ppt-adsk: After discussion with Krystian we will separate out the behavior from preventing the crash.  To simplify debugging the behavior in the future, we'll make the create transform op failure code path simply try accessing the transform op, per Krystian's suggestion.

lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
```diff
@@ -808,6 +771,22 @@ UsdTransform3dMayaXformStack::getCvtRotXYZToAttrFn(const TfToken& opName) const
...
+    if (attr && !MayaUsd::ufe::isAttributeEditAllowed(attr, &errMsg)) {
+        return false;
+    } else if (!attr) {
+        UsdGeomXformable xformable(prim());
+        if (!MayaUsd::ufe::isAttributeEditAllowed(xformable.GetXformOpOrderAttr(), &errMsg)) {
+            return false;
+        }
+    }
+    return true;
+}
```
seando-adsk, line 788: Helper to factor out some common error checking code.

lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
```diff
@@ -433,111 +433,97 @@ UsdTransform3dMayaXformStack::translateCmd(double x, double y, double z)
...
+
+                auto r = xformable.AddRotateXYZOp(UsdGeomXformOp::PrecisionFloat, opSuffix);
+                TF_AXIOM(r);
+                r.Set(v);
+                auto result = setXformOpOrderFn(xformable);
+                TF_AXIOM(result);
+
+                return r;
+            }
+        });
```
seando-adsk, line 479: Combined two lambda OpFunc into one. First check if the prim already has the attribute. If yes, then use it to create UsdGeomXformOp. Otherwise first add the attribute and then use the newly added one. Same changes made in all the command methods.

seando-adsk (commented):
I made the changes as discussed with Pierre.

kxl-adsk (approved):

ppt-adsk (approved):
Much cleaner, thanks for doing this!

ppt-adsk (approved):

PR #1560, opened 2021-07-07
MAYA-111902 Review all standard surface default values

JGamache-autodesk:
And always export to MaterialX any attribute with a default value that changed since standard surface was first introduced.


kxl-adsk:
@JGamache-autodesk can you check the PF, quite a lot of tests failed

kxl-adsk (approved):

PR #1563, opened 2021-07-09
Don't normalize point instancer quaternions

dj-mcg:
Various clients of UsdGeomPointInstancer were handling per-instance
rotation quaternions differently: some renderers would normalize while
others wouldn't, and extents were generally calculated from normalized
quaternions. In an effort to make this consistent, we now never
normalize these quaternions, relying on the incoming data being normal
instead.

Note: The mechanism of normalization was a bit sneaky: the GfRotation
constructor automatically normalizes quaternions, so by converting from
vec4 to rotation to quat we would normalize; in other situations where
we went straight to GfQuat, we wouldn't normalize.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/instancer.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/instancer.cpp
@@ -181,8 +181,7 @@ VtMatrix4dArray HdVP2Instancer::ComputeInstanceTransforms(SdfPath const& prototy
             GfVec4f quat;
             if (sampler.Sample(instanceIndices[i], &quat)) {
                 GfMatrix4d rotateMat(1);
-                rotateMat.SetRotate(
-                    GfRotation(GfQuaternion(quat[0], GfVec3d(quat[1], quat[2], quat[3]))));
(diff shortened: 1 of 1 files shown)


kxl-adsk:
When you say that you rely on incoming data to be normalized - who is responsible for doing this? Can there be assets out there that will break after this change?

dj-mcg:
> When you say that you rely on incoming data to be normalized - who is responsible for doing this? Can there be assets out there that will break after this change?

Short version: yes, it's possible there will be some differences, but because of the inconsistency this addresses, standardizing either way (either always normalizing or never normalizing) would cause something to break.

OpenGL never normalized these quats, but some of the extents computations (and Renderman renders) did. If clients are using Hydra for GL rendering, nothing will change except possibly camera framing due to extents. We're expecting most folks in the Open Source would is probably just worried about GL, so they all should be ok.

Even for those folks using Renderman or some other custom renderer, all the USD files we've seen had normalized quats except for a few that people wrote by hand for tests, so we still don't expect much fallout.

kxl-adsk (approved):

Issue #1571, opened 2021-07-13
[AL] Hard crash when loading explicit prim path from a file that does not exist


ika-rporter:
**Describe the bug**
The AL USD Maya ProxyShape node hard crashes Maya when loading explicit prim path from a file that does not exist

**Steps to reproduce**
1. Open Maya and load the AL USD Maya plugin.
1. Execute this MEL: `AL_usdmaya_ProxyShapeImport -f "/does_not_exist.usd" -pp "/any";`

**Expected behavior**
The AL USD Maya ProxyShape node should not hard crash Maya

**Attachments**
If applicable, add screenshots, sample files, etc to help explain your problem.

**Specs (if applicable):**
 - CentOS 7
 - Maya 2020.4
 - Maya USD tag v0.8.0

**Additional context**
This also occurs if you edit an existing Proxy, or try to open a file with an existing Proxy, such that `.primPath` is not an empty string and `.filePath` cites a file that does not exist.

PR #1572, opened 2021-07-13
Check that stage ptr is valid before getting prim path

ika-rporter:
When you create or modify an `AL_usdmaya_ProxyShape` node such that both:
 - The `.filePath` attribute refers to a file that does not exist.
 - The `.primPath` attribute has any value.

Maya will instantly hard crash. This request adds a guard against the null pointer when attempting to get the prim at the specified path.

Resolves [1571](https://github.com/Autodesk/maya-usd/issues/1571)


fabal (approved):
Thanks @ika-rporter 👍

PR #1577, opened 2021-07-14
MAYA-112777 - Add support for light translators

sebastienblor:
This PR adds support for maya lights translators (writer and reader), inspired from previous PR #1137 .
It supports maya's : 
 - directionalLight
 - pointLight
 - spotLight
 - areaLight

There were already lights readers & writers but they were used for renderman lights, in `lib/usd/translators/lightReader.cpp` & lightWriter.cpp. I moved their code to lightRfMReader & lightRfMWriter. For imports, when a usd builtin light is found we used to import it to a renderman light. We should now be importing to a maya light, but this can be toggled with a define `MAYA_USD_IMPORT_PXR_LIGHTS`  to restore the previous behaviour.

As requested in #1137 comments, I'm moving most of the translation code in `lib/mayaUsd/fileio/translators/translatorLight`
I needed to support import of animated parameters, which didn't seem to be in any utility function. I took inspiration from the code in `translatorCamera`  but added it as a util `ReadUsdAttribute` in `fileio/utils/readUtil.h` . 

There are some slight differences in the translation as compared to #1137. I'll list them here and explain why I did it differently, but it's of course open to discussion : 
-  In order to export/import UsdLuxShadows `shadow:enable`, I'm using maya's attribute `useRaytracedShadows`. There is another parameter in maya `useDepthMapShadows` but I'm ignoring it as it refers to a different set of parameters (lightAngle, lightRadius is supported and meant for raytraced shadows). It's also better to translate 1 maya attribute into 1 usd attribute, otherwise the import becomes tricky.
- UsdLuxLight has [0,1] float values for `diffuse` and `specular`, but maya just has checkboxes `emitDiffuse`, `emitSpecular`. Instead of ignoring this parameter during the translation, I'm writing it as 0 or 1 depending on the bool value. The reader enables them if the float value is > 0.
- UsdLuxLight prims all have a parameter `normalize` (off by default). In maya it just exists in the areaLight. For other lights types, the previous PR was forcing them to true. Here I'm just leaving it untouched, so that it matches the USD default.
- Note that maya's directional and spot/points lights have a parameter under Shadows tab called respectively `lightAngle` / `lightRadius`. For maya, they just refer to the shadows, but I'm still translating them as UsdLuxDistantLight `angle` and UsdSphereLight `radius` , which are supposed to be for the actual light illumination and not just the shadows. While the parameters aren't fully equivalent, this seems like the best compromise to translate data.

I added 2 tests `testUsdExportLight` and `testUsdImporLight` that exercise the writer and reader with static and animated parameters


sebastienblor:
Note that the import of animated parameters that were added in `readUtils` is meant to be extended/improved. It's currently supporting the use cases needed for lights, but it should be improved when needed

test/lib/usd/translators/testUsdExportLight.py
```diff
@@ -0,0 +1,187 @@
...
+        (translateOp,rotateOp,scaleOp) = rectLight.GetOrderedXformOps()
+        self.assertEqual(translateOp.GetOpType(), UsdGeom.XformOp.TypeTranslate)
+        self.assertEqual(rotateOp.GetOpType(), UsdGeom.XformOp.TypeRotateXYZ)
+        self.assertEqual(scaleOp.GetOpType(), UsdGeom.XformOp.TypeScale)
+        self.assertTrue(UsdGeom.XformCommonAPI(rectLight))
+        self.assertTrue(Gf.IsClose(translateOp.Get(1), Gf.Vec3d(8, 0, 10), 1e-6))
+        self.assertTrue(Gf.IsClose(scaleOp.Get(1), Gf.Vec3f(4,3,2), 1e-6))
+        self.assertTrue(Gf.IsClose(rotateOp.Get(1), Gf.Vec3f(0,23,0), 1e-6))
+        
+    def testExportLights(self):
```
kxl-adsk, line 176: How about adding a roundtrip test?

lib/usd/translators/lightWriter.cpp
```diff
@@ -13,54 +13,205 @@
...
+    UsdLuxDistantLight primSchema(_usdPrim);
+    MFnDirectionalLight lightFn(GetDagPath(), &status);
+    if (!status) {
+        // Do this just to get the print
+        CHECK_MSTATUS(status);
+        return;
+    }
+    
+    // First write the base light attributes
+	UsdMayaTranslatorLight::WriteLightAttrs(usdTime, primSchema, lightFn, _GetSparseValueWriter());
```
kxl-adsk, line 69: FYI. You will probably get a clang-format lint error in PF - I see a tab in here.

lib/usd/translators/lightWriter.cpp
```diff
@@ -13,54 +13,205 @@
...
+PXRUSDMAYA_REGISTER_ADAPTOR_SCHEMA(pointLight, UsdLuxSphereLight);
+
+
+PxrUsdTranslators_PointLightWriter::PxrUsdTranslators_PointLightWriter(
+        const MFnDependencyNode& depNodeFn,
+        const SdfPath&           usdPath,
+        UsdMayaWriteJobContext&  jobCtx)  : UsdMayaPrimWriter(depNodeFn, usdPath, jobCtx)
 {
-    return UsdMayaTranslatorRfMLight::Write(args, context);
+    // TODO : do we want to move the creation to fileio/translators ?
```
kxl-adsk, line 84: Keeping it separate will allow reusing utilities for prims that are already created in memory. Not a current use-case for us IIRC, but nothing hurts us keeping it in this way.

kxl-adsk (commented):
At the higher level, this all looks very good @sebastienblor! I can't comment on the actual mapping of attributes but hopefully, Jerry can help here.

kxl-adsk:
@nxkb - I wonder if you would be interested in testing this change? Luma originally opened https://github.com/Autodesk/maya-usd/pull/1137 but closed the PR after.

@dj-mcg - You mentioned having support for native lights is something interesting for your artists. Would you like to help to validate this change?

lib/usd/translators/lightReader.cpp
```diff
@@ -28,49 +28,24 @@
...
-    return UsdMayaTranslatorRfMLight::Read(args, context);
-}
-
-PXRUSDMAYA_DEFINE_READER(UsdLuxDiskLight, args, context)
-{
-    return UsdMayaTranslatorRfMLight::Read(args, context);
-}
+// Build variable used to import usd builtin
+// lights as maya lights
+#ifndef MAYA_USD_IMPORT_PXR_LIGHTS
```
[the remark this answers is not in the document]
sebastienblor: We could do this for lightReader indeed, but not for lightRfMReader. 
The renderman reader also includes reading of custom renderman schemas that we don't want to disable. Currently the ifdef only surrounds the import of usd builtin lights, and always leaves the other ones. 
So either we keep 2 ifdefs with the same logic, one in each file, or we use a ifdef in the renderman reader and test this variable in CMake for the new one. Personally I would tend to prefer if the same logic applies to both, or I think I could get confused, but I'm fine either way.

test/lib/usd/translators/testUsdExportLight.py
sebastienblor: It's already more or less what is being done, as the reader test uses the output file I got from the export test. Since I'm testing all the parameters, in theory anything that breaks should fail either the reader or the writer test, so I'm not sure an additional roundtrip test would actually test more that what is already done. But yes, we can add another test if desired

lib/usd/translators/lightWriter.cpp
sebastienblor: ok, I removed these comments

lib/usd/translators/lightWriter.cpp
sebastienblor: fixed, there were several tabs indeed

lib/mayaUsd/fileio/utils/readUtil.cpp
```diff
@@ -889,4 +891,181 @@ size_t UsdMayaReadUtil::ReadSchemaAttributesFromPrim(
...
+            return false;
+        }
+        values[i] = attrValue;
+    }
+    return true;
+}
+
+// Check if this usd attribute is animated and eventually connect an animation
+// curve if needed. Return true if the animation was imported properly.
+static bool _TranslateAnimatedUsdAttribute(
```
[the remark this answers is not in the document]
sebastienblor: done

sebastienblor:
the second commit addresses some of the comments above (I marked the most obvious ones as resolved)

lib/mayaUsd/fileio/translators/translatorLight.cpp
```diff
@@ -0,0 +1,503 @@
...
+                     _tokens->DropoffPlugName,
+                     args, context);
+
+    float UsdConeAngle = 1.f;
+    shapingAPI.GetShapingConeAngleAttr().Get(&UsdConeAngle);
+    float coneSoftness = 0.f;
+    shapingAPI.GetShapingConeSoftnessAttr().Get(&coneSoftness);
+
+    float penumbraAngle = UsdConeAngle * coneSoftness;
+    float mayaConeAngle = 2.f * (UsdConeAngle - penumbraAngle);
```
JGamache-autodesk, line 360: A negative penumbra angle becomes positive on a roundtrip and this will adjust the cone angle accordingly, resulting in the exact same illumination. This is perfectly fine, but should probably be documented.

JGamache-autodesk (approved):
Looks very good. Thanks!

lib/usd/translators/lightReader.cpp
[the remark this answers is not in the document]
kxl-adsk: OK. Ignore my comment.

test/lib/usd/translators/testUsdExportLight.py
kxl-adsk: I don't know how this applies to lights since they probably won't change that much...but with time USD files may get out of date with what exporter would generate and we may miss roundtripping issues.

sebastienblor:
The 3rd commit addresses more comments. It's adding a roundtrip test, replaces `_ReportError` by `TF_RUNTIME_ERROR` , and adds comments about penumbra.
While doing the roundtrip test, I found an issue that I found a bit surprising. When an attribute has a single explicit time sample, then doing `UsdAttribute::Get(VtValue *)` will return a value that is completely wrong. We need to specify a time, and then whatever time value we choose, the value is then correct.

Because of that, I need to use differently the function `SetMayaAttr` from `readUtils.cpp` and instead get the VtValue myself at the min interval time. I also found a few places in the lights reader where I was relying on a `UsdAttribute::Get` without specifying a time, and the same issues were also happening. By specifying a time everywhere, the problems get solved. I commited these changes, but I feel like there's something else going on here, that might have to be fixed

lib/mayaUsd/fileio/translators/translatorLight.cpp
sebastienblor: I guess you were referring to online documentation, but I added a comment about this in the code

lib/mayaUsd/fileio/translators/translatorLight.cpp
```diff
@@ -0,0 +1,503 @@
...
+    ((DropoffPlugName, "dropoff"))
+    ((PenumbraAnglePlugName, "penumbraAngle"))
+    ((ConeAnglePlugName, "coneAngle"))
+    ((LightRadiusPlugName, "lightRadius"))
+);
+// clang-format on
+
+static bool _ReportError(const std::string& msg, const SdfPath& primPath = SdfPath())
+{
+    TF_RUNTIME_ERROR(
```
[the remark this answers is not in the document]
sebastienblor: I started doing this, and then I realized the function `_ReportError` wasn't doing much more than calling TF_RUNTIME_ERROR so I removed it and I'm now calling TF_RUNTIME_ERROR directly in the code

test/lib/usd/translators/testUsdExportLight.py
sebastienblor: Added a roundtrip test in the last commit

sebastienblor:
This 4th commit extends the function 
```
static bool SetMayaAttr(
        MPlug&              attrPlug,
        const UsdAttribute& usdAttr,
        const bool          unlinearizeColors = true)
```
that isn't currently used anywhere in the maya-usd code, and adds an optional attribute 
`UsdTimeCode         time = UsdTimeCode::Default()`
that allows to specify a time for `UsdAttribute::Get`.
By doing this change, I can now reuse this function as in the first commit, and give it a time equal to `UsdTimeCode::EarliestTime()`.

JGamache-autodesk (approved):

kxl-adsk (approved):

sebastienblor:
The "import light" test wasn't compatible with older versions of USD, as there have been incompatible changes in the last year.
To fix this, I'm using 3 different usd input files :
- one for versions < 21.02 where none of the lights attributes have an "inputs" namespace
- one for version 21.02 where lights have an "inputs" namespace, but not UsdLuxShadowsAPI / UsdLuxShapingAPI
- one for versions > 21.02 where all lights parameters (including shadows and shaping) have the namespace

the test loads the appropriate file depending on the current usd version. In my local tests, this worked for 20.5, 20.11, 21.02, 21.05

There seems to be also an issue with the maya scene from the "export light" test, as it was done with a recent Maya cut which stores data that can't load with older maya versions. I removed the data that seemed to cause the problem from the .ma scene (about color management)

PR #1582, opened 2021-07-16
MAYA-111068 - Handle UDIM in MaterialX shaders

JGamache-autodesk:
Properly set the texture scale/offset attributes that adjust ST for
UDIM support.


williamkrick (approved):

JGamache-autodesk:
Pre-flight failed on a build machine with a full drive. Every other pre-flight succeeded. Fix is trivial enough to not require another one.

PR #1583, opened 2021-07-16
Update P4 filetype to Binary to match crate format

dj-mcg:
The P4 format of this file was text although it is a crate file. This
was causing issues with some of our internal development software. This
changes the p4 format to binary


dj-mcg:
Actually, another pull request will subsume this one

PR #1589, opened 2021-07-20
MAYA-112854 - Fix warnings from USDZ texture import

### LINKED ISSUE #1588: [MAYA-112854] mayaUSDImport throws USDZ related error when importing other file types
**Describe the bug**
`mayaUSDImport` throws the following error when importing any form of USD file:

`# Error: The directory specified for USDZ texture imports: P:\current\project\sourceimages is not valid. # `

It is accurate that this folder does not exist, but I get this same error when importing a USDA with no textures. Setting `importUSDZTextures=False` has no effect. I can suppress the error by pointing `importUSDZTexturesFilePath` to some folder that exists (the temp folder for example), but I don't believe I should be seeing this error in the first place.

**Steps to reproduce**
1. Set the project to a location that does not have a "sourceimages" subfolder
2. Run `cmds.mayaUSDImport(file="path/to/any/usd.usda")`

**Expected behavior**
The USD file should be imported without error

**Specs (if applicable):**
 - Windows 10
 - Maya 2022
 - Python 3
 - maya-usd 0.10.0

JGamache-autodesk:
- Do not check folders if not importing USDZ textures
- Create sourceImages if no path was given, and there is an active project, and the sourceImage folder is not yet created in that project.

Fixes #1588

### DIFF
--- a/test/lib/usd/translators/testUsdExportImportRoundtripPreviewSurface.py
+++ b/test/lib/usd/translators/testUsdExportImportRoundtripPreviewSurface.py
@@ -17,6 +17,7 @@
 
 from pxr import Gf
 from pxr import Sdf
+from pxr import Tf
 from pxr import Usd
 from pxr import UsdShade
 from pxr import UsdUtils
@@ -65,6 +66,10 @@ class testUsdExportImportRoundtripPreviewSurface(unittest.TestCase):
         """
         Tests that a usdPreviewSurface exports and imports correctly.
         """
+        mark = Tf.Error.Mark()
+        mark.SetMark()
+        self.assertTrue(mark.IsClean())
+
         mayaUsdPluginName = "mayaUsdPlugin"
         if not cmds.pluginInfo(mayaUsdPluginName, query=True, loaded=True):
             cmds.loadPlugin(mayaUsdPluginName)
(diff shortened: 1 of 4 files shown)


kxl-adsk (commented):
Looks great, Jerry. Do you mind adding a regression test?

test/lib/usd/translators/testUsdExportImportRoundtripPreviewSurface.py
```diff
@@ -65,6 +66,10 @@ def __testUsdPreviewSurfaceRoundtrip(self,
         """
         Tests that a usdPreviewSurface exports and imports correctly.
         """
+        mark = Tf.Error.Mark()
+        mark.SetMark()
+        self.assertTrue(mark.IsClean())
```
JGamache-autodesk, line 71: This unit test used to report the "sourceImage" error 10 times. Adding TfMarks makes sure the unit test runs clean and proves the issue is fixed.

test/lib/usd/translators/testUsdImportUSDZTextures.py
```diff
@@ -40,12 +42,37 @@ def tearDownClass(cls):
...
+    def testUSDZImportIntoProject(self):
+        mark = Tf.Error.Mark()
+        mark.SetMark()
+        self.assertTrue(mark.IsClean())
+
+        project_path = os.path.normpath(cmds.workspace(q=True, rd=True))
+        image_path = os.path.join(project_path, cmds.workspace(fre="sourceImages"))
+        # Purposefully erase the sourceimage directory:
+        if os.path.isdir(image_path):
+            shutil.rmtree(image_path)
```
JGamache-autodesk, line 66: Specifically testing import into a missing sourceImages folder:

test/lib/usd/translators/testUsdExportImportRoundtripPreviewSurface.py
JGamache-autodesk: (would be a good idea to generalize the use of these marks to make sure we never ignore a USD error)

kxl-adsk: Great find! And agreed!

kxl-adsk (approved):

JGamache-autodesk:
Got the expected license failure on OS/X. So ready for merge.

PR #1593, opened 2021-07-21
MAYA-112867 MayaUSD : Fix unit tests to run in Debug builds

boudrey:
Adding missing environment variable necessary for Debug builds
Fix asserts that were breaking the tests

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
@@ -287,6 +287,7 @@ MtohRenderOverride::~MtohRenderOverride()
         MMessage::removeCallbacks(panelAndCallbacks.second);
     }
 
+    if(!_allInstances.empty())
     {
         std::lock_guard<std::mutex> lock(_allInstancesMutex);
         _allInstances.erase(
--- a/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
@@ -287,8 +287,7 @@ MtohRenderOverride::~MtohRenderOverride()
         MMessage::removeCallbacks(panelAndCallbacks.second);
     }
 
-    if(!_allInstances.empty())
-    {
+    if (!_allInstances.empty()) {
         std::lock_guard<std::mutex> lock(_allInstancesMutex);
         _allInstances.erase(
             std::remove(_allInstances.begin(), _allInstances.end(), this), _allInstances.end());
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.cpp
@@ -498,7 +498,7 @@ void TranslatorContext::preRemoveEntry(
     // (which will guarentee the the itemsToRemove will be ordered such that the child prims will be
     // destroyed before their parents).
     auto iter = range_end;
-    itemsToRemove.reserve(itemsToRemove.size() + range_end - range_begin);
+    itemsToRemove.reserve(itemsToRemove.size() + (range_end - range_begin));
     while (iter != range_begin) {
         --iter;
         PrimLookup& node = *iter;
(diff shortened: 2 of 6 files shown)


lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
```diff
@@ -287,6 +287,7 @@ MtohRenderOverride::~MtohRenderOverride()
         MMessage::removeCallbacks(panelAndCallbacks.second);
     }
 
+    if(!_allInstances.empty())
     {
         std::lock_guard<std::mutex> lock(_allInstancesMutex);
         _allInstances.erase(
```
boudrey, line 293: Trying to erase an empty range was asserting in Debug.

plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.cpp
```diff
@@ -498,7 +498,7 @@ void TranslatorContext::preRemoveEntry(
     // (which will guarentee the the itemsToRemove will be ordered such that the child prims will be
     // destroyed before their parents).
     auto iter = range_end;
-    itemsToRemove.reserve(itemsToRemove.size() + range_end - range_begin);
+    itemsToRemove.reserve(itemsToRemove.size() + (range_end - range_begin));
```
boudrey, line 501: Asserting in Debug because of invalid offset when range is empty; forcing (range_end - range_begin) to compute first solved the issue.

plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.cpp
```diff
@@ -550,7 +551,7 @@ void TranslatorContext::removeEntries(const SdfPathVector& itemsToRemove)
         }
 
         // The item might already have been removed by a translator...
-        if (node != m_primMapping.end() && node->path() == path) {
+        if (primMappingSize == m_primMapping.size()) {
```
boudrey, line 554: node was not valid when removed by translator.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
```diff
@@ -241,16 +241,16 @@ inline bool ProxyShape::TransformReference::decRef(const TransformReason reason)
             m_required);
     switch (reason) {
     case kSelection:
-        assert(m_selected);
```
boudrey, line 244: This function needs to be able to exit in Debug, the returned value is tested properly and this is why it is working in release. Also making sure not to set negative values.

lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
kxl-adsk: Great catch.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
kxl-adsk: Not sure I understand this one completely. It feels like the intention was to never get into a situation where increment and decrement pattern is not respected.

kxl-adsk (approved):

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
kxl-adsk: @fabal Can you review and approve this change to the AL plugin?

boudrey: The starting point of the case that was popping the assert: maya-usd\plugin\al\plugin\AL_USDMayaTestPlugin\AL\usdmaya\nodes\test_ProxyShape.cpp (1007) :
// construct a chain of transform nodes
            proxy->removeUsdTransforms(
                kneePrim, modifier3, AL::usdmaya::nodes::ProxyShape::kRequested);
            EXPECT_EQ(MStatus(MS::kSuccess), modifier3.doIt());

fabal (approved):
Thanks @boudrey

PR #1596, opened 2021-07-22
Fix crash on scene item creation.

ppt-adsk:


lib/mayaUsd/ufe/UsdTransform3dHandler.cpp
```diff
@@ -51,6 +51,10 @@ Ufe::Transform3d::Ptr UsdTransform3dHandler::transform3d(const Ufe::SceneItem::P
     assert(usdItem);
 #endif
 
+    if (!usdItem) {
```
ppt-adsk, line 54: Trivial null pointer check.  Bug description creates a scene item in Python without going through Ufe::Hierarchy::createItem().  This creates a scene item that is not a UsdSceneItem.  Although this is not useful, it should not cause a crash.

seando-adsk (approved):

ppt-adsk:
Only the usual master Python3 macOS license failure, ready for merge.

PR #1597, opened 2021-07-23
MAYA-112834 Make sure that the render item name stored in HdVP2DrawItem exactly …

williamkrick:
matches the name of the MRenderItem.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
@@ -48,7 +48,8 @@ HdVP2DrawItem::~HdVP2DrawItem()
         MSubSceneContainer* subSceneContainer = param ? param->GetContainer() : nullptr;
         if (subSceneContainer) {
             for (const auto& renderItemData : _renderItems) {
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
```diff
@@ -61,11 +61,9 @@ HdVP2DrawItem::AddRenderItem(MHWRender::MRenderItem* item, const HdGeomSubset* g
     RenderItemData& renderItemData = _renderItems.back();
 
     renderItemData._renderItem = item;
-    renderItemData._renderItemName = _drawItemName;
+    renderItemData._renderItemName = item->name();
```
williamkrick, line 64: The name stored in the _renderItemData must exactly match the name of the MRenderItem. We use the name to remove the item from the subscene container. If the names do not match we'll leak items into the container and forget about them.

williamkrick: Actually I changed the code a bit more so that we don't rely on _renderItemName. That member is meant for debugging only.

JGamache-autodesk (approved):

PR #1599, opened 2021-07-23
MAYA-112917: Avoid TF_Verify message when USD attribute is empty.

HamedSabri-adsk:


ppt-adsk (approved):
Thanks for investigating this Hamed!.

seando-adsk (approved):

PR #1609, opened 2021-07-28
Maya 2022.1 has the Light loop API V2 and is compatible with MaterialX

JGamache-autodesk:
- Updated MaterialX build how-to.
- Transparency support
- Updated list of topo nodes

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -237,6 +237,24 @@ TF_DEFINE_PRIVATE_TOKENS(
     (ND_swizzle_vector4_vector2)
     (ND_swizzle_vector4_vector3)
     (ND_swizzle_vector4_vector4)
+    // Conversion nodes:
+    (ND_convert_float_color3)
+    (ND_convert_float_color4)
+    (ND_convert_float_vector2)
+    (ND_convert_float_vector3)
+    (ND_convert_float_vector4)
+    (ND_convert_vector2_vector3)
+    (ND_convert_vector3_color3)
+    (ND_convert_vector3_vector2)
+    (ND_convert_vector3_vector4)
+    (ND_convert_vector4_color4)
+    (ND_convert_vector4_vector3)
+    (ND_convert_color3_vector3)
+    (ND_convert_color4_vector4)
+    (ND_convert_color3_color4)
+    (ND_convert_color4_color3)
+    (ND_convert_boolean_float)
+    (ND_convert_integer_float)
 );
 // clang-format on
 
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -237,6 +237,24 @@ TF_DEFINE_PRIVATE_TOKENS(
...
+    (ND_convert_vector3_vector2)
+    (ND_convert_vector3_vector4)
+    (ND_convert_vector4_color4)
+    (ND_convert_vector4_vector3)
+    (ND_convert_color3_vector3)
+    (ND_convert_color4_vector4)
+    (ND_convert_color3_color4)
+    (ND_convert_color4_color3)
+    (ND_convert_boolean_float)
+    (ND_convert_integer_float)
```
JGamache-autodesk, line 257: These nodes also affect topology, as found while trying the brick procedural MaterialX demo file.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1877,6 +1895,8 @@ MHWRender::MShaderInstance* HdVP2Material::_CreateMaterialXShaderInstance(
         // Add automatic tangent generation:
         shaderInstance->addInputFragment("materialXTw", "Tw", "Tw");
 
+        shaderInstance->setIsTransparent(ogsFragment.isTransparent());
```
JGamache-autodesk, line 1898: This requires getting at least https://github.com/autodesk-forks/MaterialX/pull/1285 so the transparency is correctly computed.

kxl-adsk: It sounds like we should update the build instructions in the doc with the above info as well?

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
kxl-adsk: where are these tokens use?

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
JGamache-autodesk: Done.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
JGamache-autodesk: They are used to recognize nodes that affect the topology of the shader graph when we compute the graph key for shader re-use. I used a MaterialX API to simplify the list in the latest commit.

kxl-adsk (approved):
Much cleaner with the switch to use category. LGTM.

PR #1611, opened 2021-08-02
MAYA-112438 export opacity of standardSurface to usdPreviewSurface and back.

### LINKED ISSUE #1507: [MAYA-112438] StandardSurface shader opacity/transparency not supported in Maya>USD export
When exporting a USD from Maya, neither the StandardSurface's transparency or opacity channel is exported. Conversely, Lamberts, Blinns and USD proxy surface all export transparency correctly.

This is a big problem because without transparency/opacity, the StandardSurface is only partially useful for USD exports.

Tested with maya-usd v0.10.0 (and earlier versions), Maya 2022 with Windows x64.

JGamache-autodesk:
Fixes #1507

As described in the solution proposed in the issue.

Results are shown in the following image:
OpacityRoundtrip
 - First column is implicit. Monochrome texture connected via outColor results in the "r" channel being connected.
 - Second column is implicit. Monochrome texture with alpha channel connected via outColor results in the "a" channel being connected.
 - Third column generates a warning since we have no clue which channel of this RGB texture to use. Still defaults to using the "r" channel.
 - Fourth column is implicit. RGBA texture connected via outColor results in the "a" channel being connected.
 - Fifth and sixth columns are explicit connections from outColorG and outAlpha of an RGBA texture.
 - Seventh column has no connections, only RGB values which are exported via a luminance formula.
You can see that explicit connections are used on import.
The file used here can be found in the test area.


JGamache-autodesk:
@mallochj, @ysiewappl, is this as you would expect?

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -271,17 +271,16 @@ class UseRegistryShadingModeExporter : public UsdMayaShadingModeExporter
...
-
                 const TfToken dstPlugName
                     = TfToken(UsdMayaShadingUtil::GetStandardAttrName(dstPlug, false));
                 UsdAttribute dstAttribute = dstShaderInfo->GetShadingAttributeForMayaAttrName(
                     dstPlugName, MayaUsd::Converter::getUsdTypeName(srcPlug));
 
+                const TfToken srcPlugName
+                    = TfToken(UsdMayaShadingUtil::GetStandardAttrName(srcPlug, false));
+                UsdAttribute srcAttribute = srcShaderInfo->GetShadingAttributeForMayaAttrName(
+                    srcPlugName, dstAttribute.GetTypeName());
```
JGamache-autodesk, line 282: The src will often create an output to match the requested input type, so be extremely precise by returning the expected USD attribute type instead of the Maya type of the connection. This allows converting an outColor connection into a connection to a subcomponent.

lib/usd/translators/shading/fileTextureUtil.cpp
```diff
@@ -0,0 +1,86 @@
...
+// Match UDIM pattern, from 1001 to 1999
+const std::regex
+    _udimRegex(".*[^\\d](1(?:[0-9][0-9][1-9]|[1-9][1-9]0|0[1-9]0|[1-9]00))(?:[^\\d].*|$)");
+} // namespace
+
+void FileTextureUtil::MakeUsdTextureFileName(
+    std::string&       fileTextureName,
+    const std::string& usdFileName,
+    bool               isUDIM)
+{
```
JGamache-autodesk, line 44: Refactoring functions common between usdFileTextureWriter and mtlxFileTextureWriter.

lib/usd/translators/shading/usdFileTextureWriter.cpp
```diff
@@ -608,15 +577,49 @@ void PxrUsdTranslators_FileTextureWriter::WriteTransform2dNode(
...
     SdfValueTypeName usdTypeName = SdfValueTypeNames->Float;
 
     if (mayaAttrName == TrMayaTokens->outColor) {
-        usdAttrName = TrUsdTokens->RGBOutputName;
-        usdTypeName = SdfValueTypeNames->Float3;
+        if (typeName == SdfValueTypeNames->Color3f) {
+            usdAttrName = TrUsdTokens->RGBOutputName;
+            usdTypeName = SdfValueTypeNames->Float3;
+        } else {
+            // Float input detected. Happens when connecting outColor to opacity and requires an
```
JGamache-autodesk, line 593: Implicit conversion from outColor to a subcomponent happens here.

lib/usd/translators/shading/usdStandardSurfaceReader.cpp
```diff
@@ -115,20 +163,41 @@ void PxrUsdTranslators_StandardSurfaceReader::_OnBeforeReadAttribute(
...
+        }
         surfaceFn.setBaseColor(color);
         surfaceFn.setBase(1.0f);
     } else if (mayaAttrName == TrMayaTokens->emissionColor) {
         MColor color(surfaceFn.emissionColor());
         float  scale(surfaceFn.emission());
-        color /= scale;
+        if (scale != 0.0f) {
+            color /= scale;
+        }
```
JGamache-autodesk, line 176: Emission will often be zero. Prevent a division by zero here.

ysiewappl:
> @mallochj, @ysiewappl, is this as you would expect?

The logic itself described in the diagram sounds good to me; I might not have time to review the code changes themselves right now. Thanks for making this change!

lib/usd/translators/shading/fileTextureUtil.h
```diff
@@ -0,0 +1,45 @@
...
+/// \file
+/// Generic utilities shared between all file texture exporters and importers
+
+#include <pxr/pxr.h>
+
+#include <string>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+struct FileTextureUtil
```
kxl-adsk, line 28: I'm not sure how much we expect this file to grow, but we already have https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/fileio/utils/shadingUtil.h that could be a good place for these.

lib/usd/translators/shading/usdStandardSurfaceReader.cpp
```diff
@@ -82,6 +86,59 @@ PxrUsdTranslators_StandardSurfaceReader::PxrUsdTranslators_StandardSurfaceReader
 {
 }
 
+PxrUsdTranslators_StandardSurfaceReader::~PxrUsdTranslators_StandardSurfaceReader()
+{
+    // Expand Opacity from R to RGB if necessary.
+    if (!_standardSurfaceObj.isNull()) {
```
kxl-adsk, line 92: Did we do something like this before? It doesn't look kosher to rely on the lifetime of an object to fix up the connectivity.

lib/usd/translators/shading/usdStandardSurfaceWriter.cpp
```diff
@@ -155,18 +162,28 @@ void PxrUsdTranslators_StandardSurfaceWriter::Write(const UsdTimeCode& usdTime)
...
+            PxrMayaUsdPreviewSurfaceTokens->OpacityAttrName, SdfValueTypeNames->Float);
+
+        if (!opacityPlug.numConnectedChildren()) {
+            VtValue opacityValue = UsdMayaWriteUtil::GetVtValue(
+                opacityPlug,
+                SdfValueTypeNames->Color3f,
+                /* linearizeColors = */ false);
+
+            // Need the luminance since we have a single float to populate on the USD side.
+            // This is the ACEScg luminance formula. Should we infer lin_rec709 instead?
```
kxl-adsk, line 182: I was actually wondering about a similar thing - should this query color space from Maya?

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -271,17 +271,19 @@ class UseRegistryShadingModeExporter : public UsdMayaShadingModeExporter
...
-                UsdAttribute srcAttribute = srcShaderInfo->GetShadingAttributeForMayaAttrName(
-                    srcPlugName, MayaUsd::Converter::getUsdTypeName(dstPlug));
-
                 const TfToken dstPlugName
                     = TfToken(UsdMayaShadingUtil::GetStandardAttrName(dstPlug, false));
                 UsdAttribute dstAttribute = dstShaderInfo->GetShadingAttributeForMayaAttrName(
                     dstPlugName, MayaUsd::Converter::getUsdTypeName(srcPlug));
 
+                UsdAttribute srcAttribute;
+                if (dstAttribute) {
```
kxl-adsk, line 280: Why was this change necessary?

lib/usd/translators/shading/usdStandardSurfaceWriter.cpp
JGamache-autodesk: Added an OCIO v2 todo item.

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
JGamache-autodesk: Because the connection on the Maya side is to a float3 opacity, while the connection USD side is to a float opacity. We need this to get the proper destination type.

lib/usd/translators/shading/fileTextureUtil.h
JGamache-autodesk: Moved.

lib/usd/translators/shading/usdStandardSurfaceReader.cpp
JGamache-autodesk: Made everything explicit.

kxl-adsk (approved):

JGamache-autodesk:
Jenkins preflight 1476 passed after adding support for both USD 20.08 and 20.11.

PR #1613, opened 2021-08-03
MAYA-112790 USD: StagesSubject Failed verification fStageListeners.em…

boudrey:
…pty()

Enabling the test that is not failing anymore.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -512,9 +512,7 @@ MStatus MayaUsdProxyShapeBase::compute(const MPlug& plug, MDataBlock& dataBlock)
         ProxyAccessor::compute(_usdAccessor, plug, dataBlock);
         return retStatus;
     } else if (plug == outStageDataAttr) {
-        in_compute++;
         auto ret = computeOutStageData(dataBlock);
-        in_compute--;
         return ret;
     } else if (plug == outStageCacheIdAttr) {
         return computeOutStageCacheId(dataBlock);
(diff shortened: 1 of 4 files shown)


ppt-adsk (approved):

boudrey:
@ppt-adsk can you have a look at the new change please.

lib/mayaUsd/nodes/proxyShapeBase.h
```diff
@@ -379,6 +379,10 @@ class MayaUsdProxyShapeBase
 
     // Keep track of the incoming layers
     std::set<std::string> _incomingLayers;
+
+public:
+    // Counter for the number of times compute is re-entered
+    static std::atomic<int> in_compute;
```
ppt-adsk, line 385: Because this member is static, it will be used by all proxy shape nodes.  What if there are more than one proxy shape node in the scene?  Can it happen that both would enter compute() in parallel, and that we could drop a notification for one or more stages?

boudrey: Since in_compute is atomic, parallel calls to compute() should not be an issue. g_StageMap and g_StagesSubject are also global so I think the important is that onStageSet work be done only once.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -512,7 +512,10 @@ MStatus MayaUsdProxyShapeBase::compute(const MPlug& plug, MDataBlock& dataBlock)
         ProxyAccessor::compute(_usdAccessor, plug, dataBlock);
         return retStatus;
     } else if (plug == outStageDataAttr) {
-        return computeOutStageData(dataBlock);
+        in_compute++;
```
ppt-adsk, line 515: As discussed, for better encapsulation, future proofing (even though currently there is just one call site), and exception safeness (even though I don't think any exceptions can be thrown in computeOutStageData()), we can move the increment / decrement inside computeOutStageData(), with a guard object to take care of the decrement.

ppt-adsk (approved):

PR #1614, opened 2021-08-03
Ensure only parent is selected by test.

ppt-adsk:


test/lib/ufe/testParentCmd.py
```diff
@@ -192,7 +192,6 @@ def testParentRelative(self):
         cylChildren = cylHier.children()
         self.assertEqual(len(cylChildren), 1)
 
-    @unittest.skipIf(mayaUtils.previewReleaseVersion() == 122, 'Test broken in Maya Preview Release 122.')
```
ppt-adsk, line 195: Preview release 122 is far away in the rear view mirror.

test/lib/ufe/testParentCmd.py
```diff
@@ -258,8 +257,10 @@ def testParentAbsolute(self):
         # Cube world y coordinate is currently 0.
         self.assertAlmostEqual(cubeWorld.matrix[3][1], 0)
 
-        # Move the parent
-        ufe.GlobalSelection.get().append(cylinderItem)
+        # Move only the parent.
```
ppt-adsk, line 260: Test was failing when run with sabrih/MAYA-112126/parent_selection because this new change properly sets the selection onto the parented child item, rather than on nothing (current behavior).  Using append instead of replace meant that both the parent and the child were moved, which is not what the test intended to do.  This change is safe with the current Maya master, and will allow sabrih/MAYA-112126/parent_selection to be merged in Maya.

pierrebai-adsk (approved):

PR #1616, opened 2021-08-04
MAYA-112871 MayaUSD: VS Code setup

boudrey:
Fix canonical path for script run-clang-format.py (fix only on Windows and when using Python 3.6 or higher).
  - reported file paths reported by Git and paths inclusion had different case making the test fail finding any files.
Fix for forcing the use of old ABI on Linux same as done in Maya so that Maya-usd can be compiled with GCC 9.x or higher without error.

### DIFF
--- a/cmake/compiler_config.cmake
+++ b/cmake/compiler_config.cmake
@@ -99,6 +99,12 @@ function(mayaUsd_compile_config TARGET)
             PRIVATE
                 ${GNU_CLANG_FLAGS}
         )
+        if(IS_LINUX)
+            target_compile_definitions(${TARGET}
+                PRIVATE
+                    _GLIBCXX_USE_CXX11_ABI=0 # USD is built with old ABI
+            )
+        endif()
     elseif(IS_MSVC)
         target_compile_options(${TARGET} 
             PRIVATE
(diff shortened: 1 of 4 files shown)


cmake/compiler_config.cmake
```diff
@@ -99,6 +99,12 @@ function(mayaUsd_compile_config TARGET)
             PRIVATE
                 ${GNU_CLANG_FLAGS}
         )
+        if(IS_LINUX)
+            target_compile_definitions(${TARGET}
+                PRIVATE
+                    _GLIBCXX_USE_CXX11_ABI=0 # USD is built with old ABI
```
kxl-adsk, line 105: Is it true for every USD artifact or there can be USD artifacts (think of a custom build at a studio) that build with a new ABI?

kxl-adsk (commented):
Please review https://github.com/Autodesk/maya-usd/blob/dev/doc/build.md to make sure it's reflecting what's currently supported.

cmake/compiler_config.cmake
boudrey: Maya is using old ABI and has the same define defined in its CMake files.

kxl-adsk (approved):

kxl-adsk:
Validated by internal PF /job/ecg-mayausd-branch-preflight-master-public/1479/

PR #1618, opened 2021-08-09
MAYA-113148 MayaUSD: Pass UsdMayaPrimReaderContext as reference inste…

boudrey:
…ad of pointer

### DIFF
--- a/lib/mayaUsd/fileio/primReader.h
+++ b/lib/mayaUsd/fileio/primReader.h
@@ -33,10 +33,9 @@ public:
 
     /// Reads the USD prim given by the prim reader args into a Maya shape,
     /// modifying the prim reader context as a result.
-    /// Callers must ensure \p context is non-null.
     /// Returns true if successful.
     MAYAUSD_CORE_PUBLIC
-    virtual bool Read(UsdMayaPrimReaderContext* context) = 0;
+    virtual bool Read(UsdMayaPrimReaderContext& context) = 0;
 
     /// Whether this prim reader specifies a PostReadSubtree step.
     MAYAUSD_CORE_PUBLIC
@@ -44,13 +43,12 @@ public:
 
     /// An additional import step that runs after all descendants of this prim
     /// have been processed.
-    /// Callers must ensure \p context is non-null.
     /// For example, if we have prims /A, /A/B, and /C, then the import steps
     /// are run in the order:
     /// (1) Read A (2) Read B (3) PostReadSubtree B (4) PostReadSubtree A,
     /// (5) Read C (6) PostReadSubtree C
     MAYAUSD_CORE_PUBLIC
-    virtual void PostReadSubtree(UsdMayaPrimReaderContext* context);
+    virtual void PostReadSubtree(UsdMayaPrimReaderContext& context);
 
 protected:
     /// Input arguments. Read data about the input USD prim from here.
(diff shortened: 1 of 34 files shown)


lib/mayaUsd/fileio/primReader.h
```diff
@@ -33,24 +33,22 @@ class UsdMayaPrimReader
 
     /// Reads the USD prim given by the prim reader args into a Maya shape,
     /// modifying the prim reader context as a result.
-    /// Callers must ensure \p context is non-null.
```
boudrey, line 36: Not necessary anymore.

lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -602,7 +602,8 @@ class UseRegistryShadingModeImporter
     {
         // UsdMayaPrimReader::Read is a function that works by indirect effect. It will return
         // "true" on success, and the resulting changes will be found in the _context object.
-        if (!shaderReader.Read(_context->GetPrimReaderContext())) {
```
boudrey, line 605: This is the only tricky place, I tried to reproduce the logic.

lib/mayaUsd/fileio/translators/translatorLight.cpp
```diff
@@ -465,12 +465,12 @@ bool UsdMayaTranslatorLight::Read(
     }
 
     // Find which maya node needs to be our light's parent
-    MObject parentNode = context->GetMayaNode(lightSchema.GetPath().GetParentPath(), false);
+    MObject parentNode = context.GetMayaNode(lightSchema.GetPath().GetParentPath(), false);
```
boudrey, line 468: Many places like here were using the pointer without testing it.

lib/usd/translators/meshReader.cpp
```diff
@@ -67,12 +67,8 @@ TF_REGISTRY_FUNCTION_WITH_TAG(UsdMayaPrimReaderRegistry, UsdGeomMesh)
     });
 }
 
-bool MayaUsdPrimReaderMesh::Read(UsdMayaPrimReaderContext* context)
+bool MayaUsdPrimReaderMesh::Read(UsdMayaPrimReaderContext& context)
 {
-    if (!context) {
```
boudrey, line 72: Not necessary anymore.

kxl-adsk:
@dj-mcg Can you help review this change?

JGamache-autodesk (approved):
Looks good to me. Any plans to propagate that change to callees like UsdMayaReadUtil::ReadUsdAttribute?

dj-mcg:
Internally, Pixar likes to use pointers for output parameters to allow readers to immediately identify them. This isn't really an output parameter per se, and I wasn't able to find a site that might actually reassign the pointer itself, but it looks like that parameter can be mutated in the logic. I think this should be fine.

dj-mcg (approved):
Complies with Pixar's output-params-as-pointers as it's not exactly an output param. I think it looks good

boudrey:
I looked at UsdMayaReadUtil::ReadUsdAttribute but decided not to extend the change to this function because of a condition in _ReadAnimatedUsdAttribute. "(context) ? context->GetTimeSampleMultiplier() : 1.0", context could be NULL here. I didn't know if this condition is used.

PR #41, opened 2019-09-24
Changed the way reference nodes are named to fix duplicate named proxy problem

elenadBlueSky:
Changed the way reference nodes are named in order to fix the problem where
proxy nodes with duplicate names all created/attached to the same reference
node, making it impossible to save separate reference edits for them.
Also changed how code checks whether a reference has previously been created
for a prim (so it can be reused).  This allowed me to get rid of having to
create an attribute on the reference node, which wasn't serving a useful
purpose in the new scheme of things.

PR #1620, opened 2021-08-10
[al] fix MayaReference tolerance for stage hierarchy changes

nxkb:
This PR tweaks the logic that is used to reconnect maya references to the nodes that load them when they become disconnected.

#### Problem

The existing behavior can fail in two cases when references are disconnected:
a.) prim is moved or renamed and the stage hierarchy changes
b.) proxyShape is moved or renamed in the maya scene

In our pipeline, a.) happens often, especially when assemblies are being worked on early on in a project. b.) won't happen often and we've locked that node, so everyone knows to leave it alone. This PR addresses a.) and builds off [pr41](https://github.com/Autodesk/maya-usd/pull/41) to allow for flexible and reliable reconnections.

#### Fix

We take the following approach to keep MayaReference prim <-> ref connections alive in various workflows.

1.) default (current behavior)
 - auto assign maya ns based on full prim path
 - auto name reference node with proxy shape full path + full prim path.
 - do match based on maya heirarchy + prim path == reference node name
 - this rigid matching ensures that you’ll never have 2 prims fighting over the same reference

2.) specify maya namespace on prim (according to current MayaReference spec):
 - reference nodes will be tagged with specified ns
 - if attr exists, do match based on RN attribute matching specified ns and proxy shape fullpath matching start of reference node name
 - pipelines/users are in charge of ensuring these ns are unique per stage
 - use this approach for a more flexible workflow where hierarchies can change in maya or usd without breaking the connection between prim and reference node
 - if a prim (with ns attr) is duplicated in maya or the same asset is referenced on a stage, users are able to make a simple usd attribute edit and modify the prim to assign a new namespace. This should be even easier with adsk plugin and ufe. They can also clear the attribute to revert matching to default behavior.
 
#### Impact

 This way pipelines can decide whether they value safe guaranteed unique connections (1) OR more adaptable and reliable connections (2). The one meaningful change would be for pipelines that are using the custom NS but would prefer to have rigid path based connection logic. I suspect that most pipelines would be able to use the flexible logic, but if that is a concern we can think that one through.

This effectively allows pipelines to revert back to the time-tested logic before pr41 but keeps the support added by pr 41 for multi proxy shape workflows.

### DIFF
--- a/lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
@@ -137,14 +137,16 @@ const MObject getMessageAttr()
 }
 } // namespace
 
-const TfToken UsdMayaTranslatorMayaReference::m_namespaceName = TfToken("mayaNamespace");
-const TfToken UsdMayaTranslatorMayaReference::m_referenceName = TfToken("mayaReference");
+const TfToken     UsdMayaTranslatorMayaReference::m_namespaceName = TfToken("mayaNamespace");
+const TfToken     UsdMayaTranslatorMayaReference::m_referenceName = TfToken("mayaReference");
+const char* const UsdMayaTranslatorMayaReference::m_primNSAttr = "usdPrimNamespace";
 
 MStatus UsdMayaTranslatorMayaReference::LoadMayaReference(
     const UsdPrim& prim,
     MObject&       parent,
     MString&       mayaReferencePath,
-    MString&       rigNamespaceM)
+    MString&       rigNamespaceM,
+    const bool     createNsAttr)
 {
     TF_DEBUG(PXRUSDMAYA_TRANSLATORS)
         .Msg("MayaReferenceLogic::LoadMayaReference prim=%s\n", prim.GetPath().GetText());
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -457,36 +518,56 @@ MStatus UsdMayaTranslatorMayaReference::update(const UsdPrim& prim, MObject pare
...
-                                    + refNamespace + "rigNs: " + rigNamespace.c_str() + ": ")
-                                + mayaReferencePath);
-                        }
+                            command.asChar());
+                    if (!MGlobal::executeCommand(command)) {
+                        MGlobal::displayError(
+                            MString(
+                                "Failed to update reference with new namespace. refNS:"
+                                + refNamespace + "rigNs: " + rigNamespace.c_str() + ": ")
+                            + mayaReferencePath);
```
nxkb, line 548: Note that the de-indentation for scope here is just a bonus fix: It makes it so that an outdated namespace is fixed even in the case where the file path is also being updated.

seando-adsk:
@nxkb The clang-format check failed. You'll need to make a change for that. Thanks, Sean

nxkb:
> @nxkb The clang-format check failed. You'll need to make a change for that. Thanks, Sean

This formatting should be fixed now!

csyshing (dismissed):

plugin/al/translators/tests/testTranslators.py
```diff
@@ -132,6 +134,52 @@ def assertUsingUsdVariant():
...
+        Tests that connection between MayaReference prim and maya reference does not
+        break on hierarchy changes if mayaNamespace UsdAttribute is used.
+        """
+        import AL.usdmaya
+
+        mc.file(new=1, f=1)
+
+        # copy usd file A to temporary location
+        tempPath = tempfile.NamedTemporaryFile(suffix=".usda", prefix="tempTestStage", delete=True)
+        shutil.copy('./testMayaRefStageA.usda', tempPath.name)
```
nxkb, line 148: note: causing a Permissions error on windows... will fix soon.
```
IOError: [Errno 13] Permission denied: 'c:\\temp\\tempTestStagetuuicr.usda'
```

nxkb: OK found some info about the windows discrepancy [here](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) and pushed up a fix for the situation. I'm hopeful that the test will pass now.

nxkb: @kxl-adsk I'm not sure how to trigger the workflow so that I can see if the windows test is fixed. Is that something you have to do?

kxl-adsk:
Yes, you have the `write` permission and can trigger the PF workflow. All you have to do is assign (or re-assign) the PR to `ecp-devops-maya-adsk` user.
Screen Shot 2021-12-03 at 9 43 39 AM

I will trigger a new one for you now.

seando-adsk:
@nxkb I re-ran your preflight and it has test failures (TestAdditionalTranslators) on Windows in 2020 & 2022. Can you have a look please?

nxkb:
closing because this was rolled into: https://github.com/Autodesk/maya-usd/pull/2124 which is merged.

PR #1626, opened 2021-08-16
MAYA-113212 - Revert to using shading with the V1 lighting API

JGamache-autodesk:
- The shading for V2 light API is still present and will be fixed shortly. It can be enabled using the MAYAUSD_VP2_ENABLE_V2_LIGHTING_SHADER env var.

Fixes a regression found with production scenes.


JGamache-autodesk:
Preflight succeeded.

seando-adsk (approved):

PR #1629, opened 2021-08-16
Support for Hydra's new per-draw-item material tag

dj-mcg:
Hydra is changing material tags to be per-drawitem rather than per-rprim
in order to allow meshes to support geom subsets with material bindings
using different material tags. This future-proofs the vp2 render
delegate callsite


williamkrick (commented):
The one thing I don't understand is where does the per-drawitem filtering now occur? Is it handled by the HdPrimGather call?

dj-mcg:
Hey @williamkrick - the per-drawitem filtering process within HdRenderIndex now also handles the material tag filtering. Before, that was done per-rprim using HdPrimGather, but that's been changed since material tags are now per-drawitem

JGamache-autodesk (approved):

williamkrick (approved):

Issue #1624, opened 2021-08-13
[MAYA-113230] ALab loading speed bottleneck


Xaia:
**Describe the bug**
Hi there,

I believe there is some kind of bottleneck while loading ALab scene with MayaUSD plugin. Tested on both 0.10.0 and 0.11.0
I feel like Proxy and Render purpose is loading in the same time so the time to first "pixel" is stretched to 38 seconds. This is way too long for such simple scene.

**Steps to reproduce**
1. Load the ALab scene.

**Expected behavior**
In my humble opinion Render purpose should load only on "demand" or during actual Render time. 
Proxy version is low poly and has no textures assigned so it should load in less than a second. 
For instance model with 2000 separate objects and 50M polys loads in 4 seconds.

Video with comparison https://www.youtube.com/watch?v=lHjzs4_hENE

**Specs (if applicable):**
 - Windows 10 20H2
 - MayaUSD plugin 0.11.0
 - Maya PR127

Xaia:
Just for the comparison, here is how Houdini is loading this stage. Geo loads in an instant and then textures are quickly loaded in. 
Also the Alab is rendering with Karma without any issues. Arnold is not rendering the Alab at all. 
https://user-images.githubusercontent.com/60515654/129478474-0d7c4d7e-e6ad-4b7d-bd74-854c34984bfe.mp4

wtelford:
@santosg87 can you link this to internal issue MAYA-113230 ?

williamkrick:
@Xaia the long loading time is caused by texture loading. USD only synchronizes rprims whose purpose is necessary to draw the scene, but it synchronizes _all_ the sprims. When the scene loads in with only proxy purpose displayed the time taken is all the time to load the proxy geometry & materials plus all the time to load the render materials. And the render materials take up most of the time!

I reproduce a similar long load time in UsdView. There if I disable every render purpose and load the scene, it still takes me about 40s to load. I logged an issue against USD, https://github.com/PixarAnimationStudios/USD/issues/1597, to track this.

I suspect that Houdini is fast because their texture loading is asynchronous. During the actual USD load step their sprim sync calls are all fast, and so the scene appears right away without textures. Then, as textures load in the background they appear in the viewport.

Xaia:
@williamkrick thank you for checking what is exactly causing this issue. Would be a bummer to lose all the magnificent USD speed just because of some puny textures ;).

williamkrick:
[...]

PR #1633, opened 2021-08-17
MAYA-113230 Minor performance improvement for loading scenes into Vp2RenderDelegate

williamkrick:
When render tags change (or on file load) set the task controller render tags to the minimum set of tags rather than all tags. This saves time syncing rprims which are not drawn.


williamkrick:
This pull request does not solve the linked issue #1624 so do not close that issue when this is merged.

JGamache-autodesk (approved):

PR #1636, opened 2021-08-18
Improperly created scene item must not crash move tool.

ppt-adsk:
https://github.com/Autodesk/maya-usd/pull/1596
fixed crashing on improperly created scene items for UFE v1.  This pull request does more of the same, for UFE v2 this time.

### DIFF
--- a/lib/mayaUsd/ufe/UsdTransform3dCommonAPI.cpp
+++ b/lib/mayaUsd/ufe/UsdTransform3dCommonAPI.cpp
@@ -278,9 +278,10 @@ Ufe::Transform3d::Ptr
 UsdTransform3dCommonAPIHandler::transform3d(const Ufe::SceneItem::Ptr& item) const
 {
     UsdSceneItem::Ptr usdItem = std::dynamic_pointer_cast<UsdSceneItem>(item);
-#if !defined(NDEBUG)
-    assert(usdItem);
-#endif
+
+    if (!usdItem) {
+        return nullptr;
+    }
 
     // If the prim supports the common transform API, create a common API
     // interface for it, otherwise delegate to the next handler in the chain of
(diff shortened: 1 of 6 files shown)


test/lib/ufe/testMoveCmd.py
```diff
@@ -394,5 +394,37 @@ def createCommonAPI(testCase, sphereXformable):
 
         self.runTestOpUndo(createCommonAPI, 'xformOp:translate')
 
+    def testBadSceneItem(self):
```
ppt-adsk, line 397: Test Driven Development: wrote the test first, it crashed and failed, wrote the fix, test passes.

lib/mayaUsd/ufe/UsdTransform3dCommonAPI.cpp
```diff
@@ -278,9 +278,10 @@ Ufe::Transform3d::Ptr
 UsdTransform3dCommonAPIHandler::transform3d(const Ufe::SceneItem::Ptr& item) const
 {
     UsdSceneItem::Ptr usdItem = std::dynamic_pointer_cast<UsdSceneItem>(item);
-#if !defined(NDEBUG)
-    assert(usdItem);
-#endif
+
+    if (!usdItem) {
```
ppt-adsk, line 282: All files have the same fix.  Strictly speaking, only the first handler in the Transform3d chain of responsibility needs the fix, but more robust to fix all Transform3d handlers.

Transform3d handlers for UFE v2 are different from the UFE v1 handler (there is no chain of responsibility in UFE v1).

test/lib/ufe/testMoveCmd.py
```diff
@@ -394,5 +394,37 @@ def createCommonAPI(testCase, sphereXformable):
...
+        spherePath = ufe.PathString.path('%s,/Sphere1' % proxyShape)
+        # The proper way to create a scene item is the following:
+        #
+        # sphereItem = ufe.Hierarchy.createItem(spherePath)
+        #
+        # A naive user can create a scene item as the following.  The resulting
+        # scene item is not a USD scene item: it is a Python base class scene
+        # item, which has a path but nothing else.  This should not a crash
+        # when using the move command.
+        sphereItem = ufe.SceneItem(spherePath)
```
ppt-adsk, line 421: This is how you create a bad scene item: just use the Python constructor.

seando-adsk (approved):
Looks good.

PR #1637, opened 2021-08-18
MAYA-113008 Relative paths in the USD Layer Editor are set relative t…

boudrey:
…o Maya.exe not the Project.

If the specified filepath is a relative path, compute the absolute path from the scene folder otherwise, USD will use a path relative the current working directory.

### DIFF
--- a/lib/usd/ui/layerEditor/saveLayersDialog.cpp
+++ b/lib/usd/ui/layerEditor/saveLayersDialog.cpp
@@ -111,10 +111,8 @@ public:
 protected:
     void onOpenBrowser();
     void onTextChanged(const QString& text);
-    void onRelativeButtonChecked(bool checked);
 
 public:
-    QString                                                _initialStartFolder;
     QString                                                _absolutePath;
     SaveLayersDialog*                                      _parent { nullptr };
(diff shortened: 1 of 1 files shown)


lib/usd/ui/layerEditor/saveLayersDialog.cpp
```diff
@@ -189,20 +186,6 @@ void SaveLayerPathRow::onOpenBrowser()
...
-        QDir dir(_initialStartFolder);
-
-        QString relativePath = dir.relativeFilePath(_absolutePath);
-        _pathEdit->setText(relativePath);
-        _pathEdit->setEnabled(false);
-    } else {
-        _pathEdit->setEnabled(true);
-        _pathEdit->setText(_absolutePath);
-    }
-}
```
seando-adsk, line 204: Just to confirm - this looks like a method that was never actually connected to any button. So it was dead code you removed?

seando-adsk (approved):

lib/usd/ui/layerEditor/saveLayersDialog.cpp
boudrey: Yes, probably copied originally from LayerPathRow::onRelativeButtonChecked and never cleaned up.

PR #1638, opened 2021-08-19
MAYA-113320 - useSpecularWorkflow now works

JGamache-autodesk:
The value used to be ignored for USD stages. It will now take effect.


lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -2220,7 +2222,12 @@ void HdVP2Material::_UpdateShaderInstance(const HdMaterialNetwork& mat)
                 }
             } else if (value.IsHolding<int>()) {
                 const int& val = value.UncheckedGet<int>();
-                status = _surfaceShader->setParameter(paramName, val);
+                if (node.identifier == UsdImagingTokens->UsdPreviewSurface
+                    && token == _tokens->useSpecularWorkflow) {
+                    status = _surfaceShader->setParameter(paramName, val != 0);
```
williamkrick, line 2227: Is this because the type of useSpecularWorkflow is different in USD and Maya? I suppose changing the type in Maya to match theirs is a lot of work?

williamkrick (approved):

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
williamkrick: Do you think we should log an issue vs. Pixar USD for it? Maybe they will elaborate or fix it.

JGamache-autodesk: The int has been there since the V2.0 specification of UsdPreviewSurface introduced in [June 2018](https://github.com/PixarAnimationStudios/USD/commit/b9dc969d502d2263aab66520a5902c83f4e93deb). We can always ask if anyone remembers why it is an int.

PR #1644, opened 2021-08-23
MAYA-113009 prevent saving layers with identical names.

pierrebai-adsk:
If one or more layers have identical file names, prevent saving and tell the user about the identical names.

### DIFF
--- a/lib/usd/ui/layerEditor/saveLayersDialog.cpp
+++ b/lib/usd/ui/layerEditor/saveLayersDialog.cpp
@@ -513,10 +513,10 @@ void SaveLayersDialog::onCancel() { reject(); }
 
 bool SaveLayersDialog::okToSave()
 {
-    // Files can have the same file names in complicated ways, with one files having two copies,
+    // Files can have the same file names in complicated ways, with one file having two copies,
     // another three, so we keep the exact number of copies per file path.
-    std::map<QString, int> alreadySeenPaths;
-    QStringList            existingFiles;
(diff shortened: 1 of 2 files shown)


lib/usd/ui/layerEditor/saveLayersDialog.cpp
```diff
@@ -513,19 +513,26 @@ void SaveLayersDialog::onCancel() { reject(); }
 
 bool SaveLayersDialog::okToSave()
 {
-    int         i, count;
-    QStringList existingFiles;
+    // Files can have the same file names in complicated ways, with one files having two copies,
```
seando-adsk, line 516: Minor typo "files" -> "file"

lib/usd/ui/layerEditor/saveLayersDialog.cpp
```diff
@@ -513,19 +513,26 @@ void SaveLayersDialog::onCancel() { reject(); }
...
+    // another three, so we keep the exact number of copies per file path.
+    std::map<QString, int> alreadySeenPaths;
```
seando-adsk, line 518: I probably would have used a QMap here just to stay in Qt (vs std). The Qt containers are better optimized for handling Qt classes.

seando-adsk (approved):

seando-adsk (approved):

pierrebai-adsk:
Sorry, I forget that we need to manually trigger the build on github...

PR #1646, opened 2021-08-24
Update maya-usd for USD v0.21.08

seando-adsk:
#### MAYA-112845 - Update maya-usd for USD v0.21.08
* Update python init files for changes to USD.
* Add v21.08 to build doc.
* Fix two failing tests.

### DIFF
--- a/lib/mayaUsd/python/__init__.py
+++ b/lib/mayaUsd/python/__init__.py
@@ -1,4 +1,8 @@
-from . import _mayaUsd
 from pxr import Tf
-Tf.PrepareModule(_mayaUsd, locals())
-del _mayaUsd, Tf
+if hasattr(Tf, 'PreparePythonModule'):
+    Tf.PreparePythonModule('_mayaUsd')
+else:
+    from . import _mayaUsd
+    Tf.PrepareModule(_mayaUsd, locals())
+    del _mayaUsd
+del Tf
--- a/plugin/al/usdtransaction/AL/usd/transaction/__init__.py
+++ b/plugin/al/usdtransaction/AL/usd/transaction/__init__.py
@@ -22,7 +22,7 @@ except Exception:
 class ScopedTransaction(object):
 
     def __init__(self, stage, layer):
-        self.transaction = _AL_USDTransaction.Transaction(stage, layer)
+        self.transaction = Transaction(stage, layer)
 
     def __enter__(self):
         return self.transaction.Open()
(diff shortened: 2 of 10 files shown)


lib/mayaUsd/python/__init__.py
```diff
@@ -1,4 +1,8 @@
...
 from pxr import Tf
-Tf.PrepareModule(_mayaUsd, locals())
-del _mayaUsd, Tf
+if hasattr(Tf, 'PreparePythonModule'):
+    Tf.PreparePythonModule('_mayaUsd')
+else:
+    from . import _mayaUsd
+    Tf.PrepareModule(_mayaUsd, locals())
+    del _mayaUsd
+del Tf
```
seando-adsk, line 8: You'll see this same change in a few places. Basically I adapted the files to handle a change made in USD v21.08 where they introduced a new python function `PreparePythonModule`. So first check if that one exists and if so call it. Otherwise for previous versions of USD use the current PrepareModule.

plugin/al/usdtransaction/AL/usd/transaction/__init__.py
```diff
@@ -19,7 +22,7 @@
 class ScopedTransaction(object):
 
     def __init__(self, stage, layer):
-        self.transaction = _AL_USDTransaction.Transaction(stage, layer)
+        self.transaction = Transaction(stage, layer)
```
seando-adsk, line 25: Removed the unneeded `_AL_USDTransaction` which causes a problem for the new case since we don't import it. This fixes the test "Python:AL_USDTransactionTests" which was failing.

seando-adsk:
@ppt-adsk  / @dj-mcg - Pierre & Dan, I've added both of you as reviewers for this change to update the code to USD v21.08.

test/lib/ufe/testPythonWrappers.py
```diff
@@ -112,6 +112,11 @@ def testWrappers(self):
         self.assertTrue(usdRtid > 0)
         self.assertNotEqual(mayaRtid, usdRtid)
 
+        # Added this File|New to command to fix a crash when this test
+        # exists (after upgrading to USD v21.08). The crash is coming
```
ppt-adsk, line 116: Typo: "exists" --> "exits".

dj-mcg (approved):
This looks good to me, thanks @seando-adsk!

PR #1668, opened 2021-08-31
MAYA-113002 - Create USD and Maya-USD plugin artifacts built with Pyt…

seando-adsk:
#### MAYA-113002 - Create USD and Maya-USD plugin artifacts built with Python 3.9

* Added support for python 3.9.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,7 @@ option(BUILD_TESTS "Build tests." ON)
 option(BUILD_STRICT_MODE "Enforce all warnings as errors." ON)
 option(BUILD_SHARED_LIBS "Build libraries as shared or static." ON)
 option(BUILD_WITH_PYTHON_3 "Build with python 3." OFF)
+set(BUILD_WITH_PYTHON_3_VERSION 3.7 CACHE STRING "The version of Python 3 to build with")
 option(CMAKE_WANT_UFE_BUILD "Enable building with UFE (if found)." ON)
 option(CMAKE_WANT_MATERIALX_BUILD "Enable building with MaterialX (experimental)." OFF)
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -79,7 +79,11 @@ if (DEFINED PYTHON_INCLUDE_DIR AND DEFINED PYTHON_LIBRARIES AND DEFINED Python_E
     SET(PYTHONLIBS_FOUND TRUE)
     # Use the Python module to find the python lib.
     if(BUILD_WITH_PYTHON_3)
-        find_package(Python 3.7 EXACT REQUIRED COMPONENTS Interpreter)
+        # We support two versions of python 3 (3.7/3.9).
+        find_package(Python 3.9 EXACT QUIET COMPONENTS Interpreter)
+        if (NOT Python_FOUND)
+            find_package(Python 3.7 EXACT REQUIRED COMPONENTS Interpreter)
+        endif()
```
seando-adsk, line 86: First search for python 3.9 with quiet option so if not found, no message or error will be output. Then if python was not found we search again for 3.7 with required (will error if not found). This case is for our internal build system where we send in the exact python to use via the variables (line 77).

cmake/python.cmake
```diff
@@ -57,7 +57,11 @@ endif()
 
 # Use the Python module to find the python lib.
 if(BUILD_WITH_PYTHON_3)
-    find_package(Python 3.7 EXACT REQUIRED COMPONENTS Interpreter)
+    # We support two versions of python 3 (3.7/3.9).
+    find_package(Python 3.9 EXACT QUIET COMPONENTS Interpreter)
+    if (NOT Python_FOUND)
+        find_package(Python 3.7 EXACT REQUIRED COMPONENTS Interpreter)
+    endif()
```
seando-adsk, line 64: Same as above, but this code is only used when no python (via vars) was sent in. Called from CMakeLists.txt (above) line 95.

kxl-adsk (approved):

CMakeLists.txt
```diff
@@ -79,7 +80,7 @@ if (DEFINED PYTHON_INCLUDE_DIR AND DEFINED PYTHON_LIBRARIES AND DEFINED Python_E
     SET(PYTHONLIBS_FOUND TRUE)
     # Use the Python module to find the python lib.
     if(BUILD_WITH_PYTHON_3)
-        find_package(Python 3.7 EXACT REQUIRED COMPONENTS Interpreter)
+        find_package(Python ${BUILD_WITH_PYTHON_3_VERSION} EXACT REQUIRED COMPONENTS Interpreter)
```
seando-adsk, line 83: I reworked my solution as the preflight failed. I realized that you cannot call find_package twice with the same component (in this case Python) because it caches internal variables and actually overwrites the output ones. In my previous case when trying (and failing) to find py 3.9 it would set "Python_EXECUTABLE" to NOT-FOUND. And thus the call to find py 3.7 would also fail. So now I set a cache variable with the py 3 version to find (default is 3.7). Then from our build when we update to 3.9 we can set this.

kxl-adsk (approved):

PR #1669, opened 2021-08-31
Update primvars with instance interpolation when instancing is dirty.

williamkrick:
Add new test images to reflect the new behavior of USD 21.08.


test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegatePerInstanceInheritedData.py
```diff
@@ -106,27 +113,26 @@ def testPerInstanceInheritedData(self):
...
+        if usdVersion < (0, 21, 5):
             return
 
-        # There was a bugfix in USD version 21.08 that changes the results of these
-        # tests. Disable the tests while I work to update them.
-        if Usd.GetVersion() >= (0, 21, 8):
-            return
+        imageVersion = None
+        if usdVersion < (0, 21, 8):
+            imageVersion = 'pre-21_08'
```
williamkrick, line 121: There are now two sets of images for the tests that follow this code. The main set in the regular folder is what the most up-to-date results are. The older images are in the pre-21_08 folder. In the future when we don't support any USD versions older than 21.08 we can remove those old images.

kxl-adsk:
More question than a concern - do you know which change made this difference? Would be good to document.

kxl-adsk (approved):

williamkrick:
Yes it was the changes Pixar made for https://github.com/PixarAnimationStudios/USD/issues/1515

PR #1675, opened 2021-09-01
MAYA-113559 prevent crash when accessing a stale attribute from Python

pierrebai-adsk:
User (and ours) Python scripts can keep variable holding references to invalid USD attributes. Detect when an invalid USD attribute is accessed and prevent getting and setting its value rather than using invalid pointers and crash.

This was triggered in our own UI code when receiving notifications of attribute value change when two scenes with prim with identical UFE paths were manipulated in successsion in a Python script.

A typical example was a Python script that created a new Maya scene, a new USD stage, and a cone within that stage and setting a value on a property. Re-running the script within the same Maya session would crash.

### DIFF
--- a/lib/mayaUsd/ufe/UsdAttribute.cpp
+++ b/lib/mayaUsd/ufe/UsdAttribute.cpp
@@ -48,6 +48,11 @@ static constexpr char kErrorMsgInvalidType[]
 
 namespace {
 
+bool isAttributeValid(const PXR_NS::UsdAttribute& attr)
+{
+    return attr.GetPrim() && attr.GetStage();
+}
+
 template <typename T> bool setUsdAttr(const PXR_NS::UsdAttribute& attr, const T& value)
 {
     // USD Attribute Notification doubling problem:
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -48,6 +48,11 @@ static constexpr char kErrorMsgInvalidType[]
 
 namespace {
 
+bool isAttributeValid(const PXR_NS::UsdAttribute& attr)
+{
+    return attr.GetPrim() && attr.GetStage();
```
ppt-adsk, line 53: Just curious why we need the attr.GetStage() test.  Shouldn't the attr.GetPrim() test be sufficient?  Is this saying we could have a valid prim that doesn't have a stage?

ppt-adsk (commented):
Good catch!

lib/mayaUsd/ufe/UsdAttribute.cpp
pierrebai-adsk: Honestly, I started by writing the check for GetStage() and it was not enough so I added the other. I'll try without it!

ppt-adsk (approved):

PR #1683, opened 2021-09-03
Verify that the DAG path of the returned proxy shape object matches

williamkrick:
its key value in the cache.

### DIFF
--- a/lib/mayaUsd/ufe/UsdStageMap.cpp
+++ b/lib/mayaUsd/ufe/UsdStageMap.cpp
@@ -45,7 +45,7 @@ MObjectHandle nameLookup(const Ufe::Path& path)
 Ufe::Path firstPath(const MObject& object)
 {
     MDagPath dagPath;
-    auto status = MFnDagNode(object).getPath(dagPath);
+    auto     status = MFnDagNode(object).getPath(dagPath);
     CHECK_MSTATUS(status);
     return MayaUsd::ufe::dagPathToUfe(dagPath);
 }
@@ -153,7 +153,7 @@ MObject UsdStageMap::proxyShape(const Ufe::Path& path)
         = nbPathSegments(path) == 1 ? path : Ufe::Path(path.getSegments()[0]);
 
     auto iter = fPathToObject.find(singleSegmentPath);
-    
+
     if (iter == std::end(fPathToObject)) {
         // When we don't find an entry in the cache then we are in scenerio 1.
         // MObjects stay valid even when re-parented or re-named, so we can
--- a/lib/mayaUsd/ufe/UsdStageMap.cpp
+++ b/lib/mayaUsd/ufe/UsdStageMap.cpp
@@ -137,43 +137,63 @@ MObject UsdStageMap::proxyShape(const Ufe::Path& path)
 {
     rebuildIfDirty();
 
+    // In additional to the explicit dirty system it is possible that
+    // the cache is in an invalid state and needs to be refreshed. See
+    // the class comment in UsdStageMap.h for details.
+    // There are two scenerios which signal a cache refresh is required:
+    // 1. A path is searched for which cannot be found. This indicates that
+    //    the stage has been reparented and the new path has been used to search
+    //    for the stage and the stage is not present in the cache.
+    // 2. The DAG path of a cached MObject doesn't match the key path used to
+    //    find that MObject. This indicates that the stage has been reparented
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/UsdStageMap.cpp
```diff
@@ -138,7 +142,12 @@ MObject UsdStageMap::proxyShape(const Ufe::Path& path)
 
     auto iter = fPathToObject.find(singleSegmentPath);
     if (iter != std::end(fPathToObject)) {
-        return iter->second.object();
+        // If the proxyShape has been re-parented in Maya & the stage lookup
```
williamkrick, line 145: This fixes the issue I am hitting in internal code review 253390 where I can get a Ufe::SceneItem for a prim which had its proxyShape node re-parented using a Ufe::Path to the old proxy shape.

Does this lose the performance benefit of the cache? I'm not sure what exactly the cache was meant to speed up, but now the code is getting the DAG path of the object and converting that into a Ufe::Path every time we hit the cache.

ppt-adsk (commented):
Needs more investigation.

lib/mayaUsd/ufe/UsdStageMap.cpp
williamkrick: The comment on line 153 is correct. UsdStageMap is not listening for re-parent notifications for when the proxy shape is re-parented, so when a proxy shape is re-parented the cache remains valid and will respond to queries.

Following a re-parent operation the code will detect the invalid cache state when a query is made for the new proxy shape path which is not in the cache. That triggers the code at line 160 to execute a rebuild the cache.

The problem is if no queries have come in for the new proxy shape path then the old cache is still intact and handling queries. If a query comes in for the _old_ proxy shape path then there will be a cache hit & a proxy shape returned for a DAG path which doesn't exist! The cached MObject is valid, but it's path is not the path that was searched for.

The new code handles this case by verifying that the path of the MObject stored in the cache matches the cache key used to look up the MObject. These two paths will be different when the MObject has been re-parented & the cache hasn't been rebuilt yet.

seando-adsk: I think the assumption is that this cache was never meant to be called with the "old" name (after a reparent or rename). Is the performance impact of this change? I wonder if this UsdStageMap should be listening for Maya rename/reparent and if on any shape we've cache'd we invalidate cache.

williamkrick: The performance impact is every time we hit the change we do two operations whose complexity is on the order of the length of a path to a proxy shape. So the cost is not very high, but it is not zero.

I think UsdStageMap originally had a callback based design but it was changed to the passive error detection approach for correctness. In the comments in the file it talks about other things such as outliner listening for the callbacks & trying to use UsdStageMap and there being a "race" between which callback is invoked first (because there is no ordering).

Even the current design is not 100% correct. If there was a series of re-parenting & renaming operations that swap the paths of two proxy shapes without allowing any queries in between. In this case the cache would never get rebuilt because there is always a cache hit, and the cache would return the wrong proxy shape to the user!

For example if we have two proxy shapes `|stage1|stageShape1` which has stage A and `|stage2|stageShape2` that has stage B. Then if I execute some python like:

```
import maya.cmds as cmds
cmds.rename('|stage1|stageShape1', '|stage1|intermediate')
cmds.rename('|stage2|stageShape2', '|stage2|stageShape1')
cmds.reparent('stage2|stageShape1', |stage1', relative=True)
cmds.reparent('stage1|intermediate', |stage2', relative=True)
cmds.rename('|stage2|intermediate', |stage2|stageShape2)
```

Then the cache will have `|stage1|stageShape1` return stage B and `|stage2|stageShape2` return stage A!

So really I think we need the code I added to handle the callback race case as best as we can, but we also need to listen to the paths again so that we can prevent errors from occurring.

seando-adsk: @williamkrick I'm on vacation for 1.5 weeks. I would suggest putting Krystian as reviewer in my place.

williamkrick:
I had a chat with @ppt-adsk and I have a plan for more changes so don't review until you see another commit!

williamkrick:
I've updated this PR with the results of the discussion @ppt-adsk and I had on Friday, so this is now ready for review!

ppt-adsk (approved):

PR #1684, opened 2021-09-03
MAYA-109415 Fix warning on blendShape export

JGamache-autodesk:
Got "Unrecognized node" error on blendShape export. The node was the
fCurve driving the weight.

Looked in mayaFindOrigMeshFromBlendShapeTarget() and found out that the
traversal would start on a random plug in the checked branch. Sometimes
the input geom would be found first, sometimes the fCurve would be
traversed as well.

Fixed the code to make sure the traversal starts at the inputGeom plug
that would have been used to compute fnGeoFilter.inputShapeAtIndex(...)
in the ignoreWarning branch.

### DIFF
--- a/lib/usd/translators/meshWriter.cpp
+++ b/lib/usd/translators/meshWriter.cpp
@@ -124,30 +124,26 @@ MObject mayaFindOrigMeshFromBlendShapeTarget(const MObject& mesh, MObjectArray* 
             MPlug outputGeomPlug = itDg.thisPlug();
             TF_VERIFY(outputGeomPlug.isElement() == true);
 
+            // Find the corresponding "inputGeometry" plug:
+            MFnDependencyNode blendShapeNode(curBlendShape);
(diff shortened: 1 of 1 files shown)


JGamache-autodesk:
@ysiewappl , since you wrote the original code, I would appreciate if you could have a look. Thanks!

ysiewappl:
> Looked in mayaFindOrigMeshFromBlendShapeTarget() and found out that the
> traversal would start on a random plug in the checked branch. Sometimes
> the input geom would be found first, sometimes the fCurve would be
> traversed as well.

Yea, I think your fix of starting right from the corresponding element plug for
`inputGeom` rather than the entire blendshape node itself makes sense to avoid
cases like these where there are other paths that lead to a mesh. Thanks for
helping to fix this!

ysiewappl (approved):

boudrey (approved):

PR #1689, opened 2021-09-08
Apply fallback value to UsdUVTexture if needed

fabal:
This change is to handle networks with "empty" UsdUVTexture nodes and apply any authored fallback.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -105,6 +105,7 @@ namespace {
 TF_DEFINE_PRIVATE_TOKENS(
     _tokens,
 
+    (fallback)
     (file)
     (opacity)
     (useSpecularWorkflow)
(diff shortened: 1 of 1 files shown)


JGamache-autodesk (changes requested):
Very good, but can you provide HLSL and Cg versions as well? There is an equivalent `file.GetDimensions()` call that you could use.

JGamache-autodesk:
After a bit of digging I found that setting the MHWRender::MSamplerStateDesc::borderColor struct member with the fallback color is probably what you are looking for. This would mean the [_GetSamplerStateDesc()](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/render/vp2RenderDelegate/material.cpp#L637) function would need to be updated to lookup the fallback color and set it in the sampler state.

fabal:
I went ahead with your suggestion @JGamache-autodesk , however, it doesn't give me the result I want.
My understanding is that by setting `borderColor`, changes in the shader itself are not needed anymore. Did I miss something?

JGamache-autodesk:
I see. Works in Maya because it will generate a 1x1 texture of the fallback value when the file is not found. This can be replicated in the mayaUsdPlugin. Can you try #1692 to see if the result is what you want?

fabal:
Closing this in favour of #1692

PR #1695, opened 2021-09-10
Made filteredTypeIds const in export job arguments

JGamache-autodesk:
It was one of the very remaining few non-const member.

### DIFF
--- a/lib/mayaUsd/fileio/jobs/jobArgs.cpp
+++ b/lib/mayaUsd/fileio/jobs/jobArgs.cpp
@@ -326,6 +326,55 @@ static PcpMapFunction::PathMap _ExportRootsMap(
     return pathMap;
 }
 
+static void _AddFilteredTypeName(const MString& typeName, std::set<unsigned int>& filteredTypeIds)
+{
+    MNodeClass   cls(typeName);
+    unsigned int id = cls.typeId().id();
+    if (id == 0) {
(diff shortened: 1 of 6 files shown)


lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -326,6 +326,55 @@ static PcpMapFunction::PathMap _ExportRootsMap(
     return pathMap;
 }
 
+static void _AddFilteredTypeName(const MString& typeName, std::set<unsigned int>& filteredTypeIds)
```
JGamache-autodesk, line 329: Moved from below and made static private.

lib/mayaUsd/fileio/jobs/jobArgs.h
```diff
@@ -240,33 +245,16 @@ struct UsdMayaJobExportArgs
     MAYAUSD_CORE_PUBLIC
     static const VtDictionary& GetDefaultDictionary();
 
-    /// Adds type name to filter out during export. This will also add all
-    /// inherited types (so if you exclude "constraint", it will also exclude
-    /// "parentConstraint")
-    MAYAUSD_CORE_PUBLIC
-    void AddFilteredTypeName(const MString& typeName);
```
JGamache-autodesk, line 247: There were no clients modifying this variable mid-export. The list was always built as the arguments were parsed and left untouched for the duration of the export.

boudrey (approved):

PR #1696, opened 2021-09-13
Fix MaterialX code generation

JGamache-autodesk:
Also fix MStringArray not being 100% STL compatible for VS 2019.

### DIFF
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -53,17 +53,6 @@
 
 PXR_NAMESPACE_USING_DIRECTIVE
 
-#ifndef MAYA_MSTRINGARRAY_ITERATOR_CATEGORY
-// MStringArray::Iterator is not standard-compliant in Maya 2019, needs the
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -313,8 +302,7 @@ bool isAGatewayType(const std::string& mayaNodeType)
     cmd.format("nodeType -inherited -isTypeName ^1s", mayaNodeType.c_str());
     if (MS::kSuccess == MGlobal::executeCommand(cmd, inherited)) {
         MString gatewayNodeType(ProxyShapeHandler::gatewayNodeType().c_str());
-        auto    iter2 = std::find(inherited.begin(), inherited.end(), gatewayNodeType);
-        isInherited = (iter2 != inherited.end());
+        isInherited = inherited.indexOf(gatewayNodeType) != -1;
```
JGamache-autodesk, line 305: Fails with VS2019 because of missing type traits "value_type" and "reference".

ppt-adsk (approved):

kxl-adsk (approved):

PR #1697, opened 2021-09-13
T bailp/maya 112770/property editability

pierrebai-adsk:


lib/mayaUsd/utils/editability.cpp
```diff
@@ -0,0 +1,52 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "editability.h"
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+/*! \brief  The tokens used in the lock metadata.
+ */
+TfToken Editability::MetadataToken("mayaLock");
```
kxl-adsk, line 22: It's time to consolidate all the metadata token into a single file. Similar to what's was done with optionvars - https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/base/tokens.h#L32. Please create another set of tokens in that file and consolidate selectability, editability, [custom layer data](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/utils/customLayerData.h#L30), and look for others in the mayaUsd library.

kxl-adsk (changes requested):

lib/mayaUsd/utils/editability.cpp
pierrebai-adsk: Oh right, I forgot that, will do.

pierrebai-adsk:
I'll close the pull request while I move things around to avoid noise.

PR #1698, opened 2021-09-13
Fix python 3.9 AL TransactionManager crash

JGamache-autodesk:
Makes sure it does not hold stale Python objects on exit.


fabal (approved):
Thanks @JGamache-autodesk 👍

PR #1708, opened 2021-09-16
[base] usdExport: bugfix for exporting instances that have namespaces

nxkb:
This PR is a simple bugfix for exporting instances that have namespaces or other use invalid characters. 

Without it, an instance master like "...|namespace:pCube1" causes a blank `SdfPath('')` to be returned as the exportPath and then the master + instances are skipped and not included in the export.

It copies the same clean-up logic that is used for the full path.


boudrey (approved):

JGamache-autodesk (dismissed):

JGamache-autodesk:
Thanks @nxkb for the fix. It would be nice if it was accompanied by a unit test to ensure namespaced instances continues exporting correctly as the code evolves.
One solution would be to add namespaced instances to the [scene](https://github.com/Autodesk/maya-usd/blob/dev/test/lib/usd/translators/UsdExportInstancesTest/UsdExportInstancesTest.ma) used by the current [instance unit test](https://github.com/Autodesk/maya-usd/blob/dev/test/lib/usd/translators/testUsdExportInstances.py). This would be pretty trivial to add if you have access to Maya 2019 (since unit tests need to run with the oldest supported Maya version).
Another solution is to add a separate test in the Python file that programmatically builds the scenario you want to export. Would something like 
```
cmds.namespace(add="ns")
cmds.namespace(set="ns")
cubeXform = cmds.polyCube()[0]
cubeInstance = cmds.instance(cubeXform)
cmds.move(cubeInstance, r=(0,2,0))
cmds.usdExport( ...... )
```
Be sufficient to reproduce the issue you had?

nxkb:
> Thanks @nxkb for the fix. It would be nice if it was accompanied by a unit test to ensure namespaced instances continues exporting correctly as the code evolves. One solution would be to add namespaced instances to the [scene](https://github.com/Autodesk/maya-usd/blob/dev/test/lib/usd/translators/UsdExportInstancesTest/UsdExportInstancesTest.ma) used by the current [instance unit test](https://github.com/Autodesk/maya-usd/blob/dev/test/lib/usd/translators/testUsdExportInstances.py). This would be pretty trivial to add if you have access to Maya 2019 (since unit tests need to run with the oldest supported Maya version). 

Test added!

> Be sufficient to reproduce the issue you had?

Yep, I believe anything where the instance source had a namespace was going to trip up the exporter.

boudrey (dismissed):

JGamache-autodesk (dismissed):

JGamache-autodesk (approved):

PR #1709, opened 2021-09-17
MAYA-113821 MayaUSD : Python bindings un-registration of registered t…

boudrey:
…ranslators

Register python to atexit, similar to how it is done for dlls.

### DIFF
--- a/lib/mayaUsd/fileio/chaser/exportChaserRegistry.cpp
+++ b/lib/mayaUsd/fileio/chaser/exportChaserRegistry.cpp
@@ -51,12 +51,15 @@ TF_INSTANTIATE_SINGLETON(UsdMayaExportChaserRegistry);
 
 std::map<std::string, UsdMayaExportChaserRegistry::FactoryFn> _factoryRegistry;
 
-bool UsdMayaExportChaserRegistry::RegisterFactory(const std::string& name, FactoryFn fn)
+bool UsdMayaExportChaserRegistry::RegisterFactory(
+    const std::string& name,
+    FactoryFn          fn,
+    bool               fromPython)
 {
     TF_DEBUG(PXRUSDMAYA_REGISTRY).Msg("Registering chaser '%s'.\n", name.c_str());
     auto ret = _factoryRegistry.insert(std::make_pair(name, fn));
     if (ret.second) {
-        UsdMaya_RegistryHelper::AddUnloader([name]() { _factoryRegistry.erase(name); });
+        UsdMaya_RegistryHelper::AddUnloader([name]() { _factoryRegistry.erase(name); }, fromPython);
     }
     return ret.second;
 }
--- a/lib/mayaUsd/fileio/registryHelper.cpp
+++ b/lib/mayaUsd/fileio/registryHelper.cpp
@@ -282,6 +282,7 @@ VtDictionary UsdMaya_RegistryHelper::GetComposedInfoDictionary(const std::vector
     return result;
 }
 
+// Vector of Python unloaders
 std::vector<std::function<void()>> g_pythonUnloaders;
 
 static void PythonUnload(size_t unloaderIndex)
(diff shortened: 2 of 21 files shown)


lib/mayaUsd/fileio/chaser/exportChaserRegistry.cpp
```diff
@@ -51,12 +51,15 @@ TF_INSTANTIATE_SINGLETON(UsdMayaExportChaserRegistry);
 
 std::map<std::string, UsdMayaExportChaserRegistry::FactoryFn> _factoryRegistry;
 
-bool UsdMayaExportChaserRegistry::RegisterFactory(const std::string& name, FactoryFn fn)
```
boudrey, line 54: Same parameter fromPython added at every Registration to determine if registration is done in c++ (within a .dll) or from Python.

lib/mayaUsd/fileio/registryHelper.cpp
```diff
@@ -280,11 +282,28 @@ VtDictionary UsdMaya_RegistryHelper::GetComposedInfoDictionary(const std::vector
...
+
 /* static */
-void UsdMaya_RegistryHelper::AddUnloader(const std::function<void()>& func)
+void UsdMaya_RegistryHelper::AddUnloader(const std::function<void()>& func, bool fromPython)
 {
-    if (g_pythonRegistry)
+    if (fromPython) {
+        g_pythonUnloaders.emplace_back(func);
+        if (boost::python::import("atexit")
+                .attr("register")(&PythonUnload, g_pythonUnloaders.size() - 1)
```
boudrey, line 301: The magic is here, registering to atexit with static c++ function and g_pythonUnloaders current last element.

JGamache-autodesk (approved):

JGamache-autodesk:
Fixes unregistering all Python entries before the interpreter gets destroyed. The atexit logic should make sure all items registered while Python was live will be run before the interpreter exits, but I wonder if registering for onMayaExit MMessage would be safer.

One interesting question is how does this work if someone unloads the mayaUsdPlugin after an export, then reloads it. Will the callback be correctly unregistered when the module is unloaded, and will they correctly be re-registered when the plugin is reloaded? Is this a case where we should prevent unloading the plugin to preserve the sanity of the registries?

JGamache-autodesk:
Seems like having any entry in the registries should disallow unloading the maya plugin.

To test:
  - Create mesh sphere
  - Assign red standard surface
  - loadPlugin mayaUsdPlugin
  - Export all to USD (notice all the registrations with TF_DEBUG=PXRUSDMAYA_REGISTRY)
  - unloadPlugin mayaUsdPlugin
  - loadPlugin mayaUsdPlugin
  - Export all to USD
BUG: Second export fails with
`Found loaded usdMaya plugin mayaUsd_Translators: UsdMaya/PrimWriter = mesh.  No maya plugin.
No usdMaya writer plugin for maya type mesh. No maya plugin found.
No usdMaya writer plugin for maya type surfaceShape. No maya plugin found.
No usdMaya writer plugin for maya type controlPoint. No maya plugin found.
No usdMaya writer plugin for maya type deformableShape. No maya plugin found.
No usdMaya writer plugin for maya type geometryShape. No maya plugin found.
No usdMaya writer plugin for maya type shape. No maya plugin found.
No usdMaya writer plugin for maya type dagNode. No maya plugin found.
No usdMaya writer plugin for maya type entity. No maya plugin found.
No usdMaya writer plugin for maya type containerBase. No maya plugin found.`

boudrey:
This was testing the registration/un-registration of c++ plugin not python. Right? If so, I can create a new JIRA for it and adding that the Python registration/un-registration should also be validated.

As for using onMayaExit MMessage, atexit callback will be called before, which is what I prefer. What do you mean by safer?

JGamache-autodesk:
Yes, that test script was for the existing C++ part which is currently broken. The solution will have to be good enough to support Python as well.

I have re-read how atexit works in Python and it is exactly the right function to use. Please disregard the mayaExit comment.

boudrey:
Logged as MAYA-113849 Failing export after unloading/re-loading mayaUsdPlugin (https://jira.autodesk.com/browse/MAYA-113849)

PR #1714, opened 2021-09-20
Made anonymous sublayers serialisable

J-Mo63:
Currently loading a scene serialised with anonymous sublayers results in stale references to previous sublayers' IDs. Although they exist in the registry on load, their IDs no longer match those referenced. This PR adds support for the serialisation and deserialisation of anonymous sublayers to Maya files by the layer manager.


J-Mo63:
This change has been reopened with https://github.com/Autodesk/maya-usd/pull/1872

PR #1715, opened 2021-09-20
MAYA-112768 - Channel box for USD prims

seando-adsk:
#### MAYA-112768 - Channel box for USD prims
* Set default USD attribute display filter in Channel Box.
* Created new global UI Ufe Observer which connects to global attribute changes to update Channel Box.

### DIFF
--- a/lib/mayaUsd/ufe/CMakeLists.txt
+++ b/lib/mayaUsd/ufe/CMakeLists.txt
@@ -56,6 +56,7 @@ if(CMAKE_UFE_V2_FEATURES_AVAILABLE)
             UsdTransform3dSetObjectMatrix.cpp
             UsdTransform3dUndoableCommands.cpp
             UsdUIInfoHandler.cpp
+            UsdUIUfeObserver.cpp
             UsdUndoAddNewPrimCommand.cpp
             UsdUndoCreateGroupCommand.cpp
             UsdUndoInsertChildCommand.cpp
@@ -129,6 +130,7 @@ if(CMAKE_UFE_V2_FEATURES_AVAILABLE)
         UsdTransform3dSetObjectMatrix.h
         UsdTransform3dUndoableCommands.h
         UsdUIInfoHandler.h
+        UsdUIUfeObserver.h
         UsdUndoableCommand.h
         UsdUndoAddNewPrimCommand.h
         UsdUndoCreateGroupCommand.h
--- a/lib/mayaUsd/ufe/Global.cpp
+++ b/lib/mayaUsd/ufe/Global.cpp
@@ -36,6 +36,7 @@
 #include <mayaUsd/ufe/UsdTransform3dMayaXformStack.h>
 #include <mayaUsd/ufe/UsdTransform3dPointInstance.h>
 #include <mayaUsd/ufe/UsdUIInfoHandler.h>
+#include <mayaUsd/ufe/UsdUIUfeObserver.h>
 #endif
 
 #include <ufe/hierarchyHandler.h>
@@ -150,6 +151,7 @@ MStatus initialize()
     handlers.transform3dHandler = pointInstanceHandler;
 
     g_USDRtid = Ufe::RunTimeMgr::instance().register_(kUSDRunTimeName, handlers);
+    MayaUsd::ufe::UsdUIUfeObserver::create();
 #else
     auto usdHierHandler = UsdHierarchyHandler::create();
     auto usdTrans3dHandler = UsdTransform3dHandler::create();
@@ -185,6 +187,8 @@ MStatus finalize()
     if (g_MayaContextOpsHandler)
         Ufe::RunTimeMgr::instance().setContextOpsHandler(g_MayaRtid, g_MayaContextOpsHandler);
     g_MayaContextOpsHandler.reset();
+
+    MayaUsd::ufe::UsdUIUfeObserver::destroy();
 #endif
     Ufe::RunTimeMgr::instance().unregister(g_USDRtid);
     g_MayaHierarchyHandler.reset();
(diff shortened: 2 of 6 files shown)


lib/mayaUsd/ufe/CMakeLists.txt
```diff
@@ -56,6 +56,7 @@ if(CMAKE_UFE_V2_FEATURES_AVAILABLE)
             UsdTransform3dSetObjectMatrix.cpp
             UsdTransform3dUndoableCommands.cpp
             UsdUIInfoHandler.cpp
+            UsdUIUfeObserver.cpp
```
seando-adsk, line 59: I put both files in the V2 block since they don't rely on anything new from UFE.

lib/mayaUsd/ufe/Global.cpp
```diff
@@ -150,6 +151,7 @@ MStatus initialize()
     handlers.transform3dHandler = pointInstanceHandler;
 
     g_USDRtid = Ufe::RunTimeMgr::instance().register_(kUSDRunTimeName, handlers);
+    MayaUsd::ufe::UsdUIUfeObserver::create();
```
seando-adsk, line 154: Again in the V2 block I create/destroy this new observer.

lib/mayaUsd/ufe/UsdUIUfeObserver.cpp
```diff
@@ -0,0 +1,90 @@
...
+{
+}
+
+/*static*/
+void UsdUIUfeObserver::create()
+{
+    TF_VERIFY(!ufeObserver);
+    if (!ufeObserver) {
+        ufeObserver = std::make_shared<UsdUIUfeObserver>();
+        Ufe::Attributes::addObserver(ufeObserver);
```
seando-adsk, line 53: For now listen to global attribute changes.

lib/mayaUsd/ufe/UsdUIUfeObserver.cpp
```diff
@@ -0,0 +1,90 @@
...
+    if (auto ac = dynamic_cast<const Ufe::AttributeValueChanged*>(&notification)) {
+        if (ac->name() == UsdGeomTokens->xformOpOrder) {
+            MString cmd("if (`channelBox -exists mainChannelBox`) channelBox -q -mainObjectList "
+                        "mainChannelBox;");
+            MStringArray paths;
+            if (MGlobal::executeCommand(cmd, paths) && (paths.length() > 0)) {
+                auto ufePath = Ufe::PathString::path(paths[0].asChar());
+                if (ufePath.startsWith(ac->path())) {
+                    cmd = "channelBox -e -update mainChannelBox;";
+                    MGlobal::executeCommand(cmd);
```
seando-adsk, line 82: If we get an attribute change on an xformOpOrder attribute, check if the channel box is displaying the prim this notif came from. If yes, tell the channel box to update (which will show the newly added xform attribute).

seando-adsk: I thought about removing the code in our [AE template](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py#L78) that refreshes it for the same xformOpOrder change, but there is no way to query the AE for the displayed node. Plus you can have multiple AE tabs and windows opened. As far as I know there is only a single global var "gAECurrentTab" that stores the name/ufepathstring of the object in the current tab of the AE.

plugin/adsk/scripts/mayaUsd_pluginUICreation.mel
```diff
@@ -7,6 +7,14 @@ global proc mayaUsd_pluginUICreation()
 
     source "mayaUsd_fileOptions.mel";
 
+    // Set the default USD attributes we want to display
+    // in the Channel Box.
+    global string $gChannelBoxName;
+    if (`channelBox -exists $gChannelBoxName`) {
+        // This flag only exists in recent versions of Maya.
+        catchQuiet(`channelBox -e -ufeFixedAttrList "USD" {"xformOp:translate*", "xformOp:rotate*", "xformOp:scale*", "xformOp:*", "visibility"} $gChannelBoxName`);
```
seando-adsk, line 15: Use catchQuiet so I don't need any Maya version check here. That way if backported this code will just work.

test/lib/ufe/testAttribute.py
```diff
@@ -460,7 +460,9 @@ def testObservation(self):
         (ball34Obs, ball35Obs, globalObs) = [TestObserver() for i in range(3)]
 
         # Maya registers a single global observer on startup.
-        self.assertEqual(ufe.Attributes.nbObservers(), 1)
+        # Maya-Usd lib registers a single global observer when it is initialized.
+        kNbGlobalObs = 2
+        self.assertEqual(ufe.Attributes.nbObservers(), kNbGlobalObs)
```
seando-adsk, line 465: Updated this test because I added a new global attribute observer.

lib/mayaUsd/ufe/UsdUIUfeObserver.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+//------------------------------------------------------------------------------
+// Ufe::Observer overrides
+//------------------------------------------------------------------------------
+
+void UsdUIUfeObserver::operator()(const Ufe::Notification& notification)
+{
+    if (auto ac = dynamic_cast<const Ufe::AttributeValueChanged*>(&notification)) {
+        if (ac->name() == UsdGeomTokens->xformOpOrder) {
+            MString cmd("if (`channelBox -exists mainChannelBox`) channelBox -q -mainObjectList "
```
ppt-adsk, line 75: Might consider making the MString commands constants somehow, either by making them static or moving them to file scope.

ppt-adsk (approved):

lib/mayaUsd/ufe/UsdUIUfeObserver.cpp
seando-adsk: Done

ppt-adsk (approved):

PR #1717, opened 2021-09-21
Use PXR_NAMESPACE_USING_DIRECTIVE to fix internal Pixar builds

dj-mcg:


kxl-adsk:
@dj-mcg Sorry about that and thx for fixing!

dj-mcg:
np!

PR #1723, opened 2021-09-27
Fix compilation issue for schema, header and UFE linking

csyshing:
- 1st commit `73c880` is to fix a leftover from previous [refactoring PR](https://github.com/Autodesk/maya-usd/pull/655) ([commit](https://github.com/Autodesk/maya-usd/commit/7db124ef06b4bb2df2259cee2ca57a5217090894))
- 2nd commit `e066d02` is needed due to our internal build system builds `lib/mayaUsd` and individual Maya plugins separately
- 3rd commit `bd800b8` is to fix an issue from previous [PR](https://github.com/Autodesk/maya-usd/pull/1138) ([commit](https://github.com/Autodesk/maya-usd/commit/1281cbb374daba6ab96c38e7bdd6cd574e5ef045)) that OSD includes was in `HDVP2_ENABLE_GPU_COMPUTE` macro but has been moved out after refactoring
- 4th commit `a096043` is fix the linking issue of PXR_USDMaya, our build system compiles PXR_USDMaya separately, we found that we need to explicitly link against UFE.


lib/mayaUsd/render/vp2RenderDelegate/meshViewportCompute.cpp
```diff
@@ -22,14 +22,15 @@
...
-#include <pxr/imaging/pxOsd/refinerFactory.h>
-#include <pxr/imaging/pxOsd/tokens.h>
 #include <pxr/pxr.h>
 
 #include <maya/MProfiler.h>
 
 #ifdef HDVP2_ENABLE_GPU_COMPUTE
 
+#include <pxr/imaging/pxOsd/refinerFactory.h>
+#include <pxr/imaging/pxOsd/tokens.h>
```
seando-adsk, line 32: I thought moving these includes here would break the clang-format (regarding the rules about include order). But the preflight passed. Is ```HDVP2_ENABLE_GPU_COMPUTE``` removing the entire file? Can that line (29) just be moved up after ```#include "meshViewportCompute.h"``` and leave these two headers where they were?

seando-adsk (approved):

lib/mayaUsd/render/vp2RenderDelegate/meshViewportCompute.cpp
csyshing: I have updated the code: [`6d27610`](https://github.com/Autodesk/maya-usd/pull/1723/commits/6b27610d5ada2b757461678b26c4e22e0999aea5)

seando-adsk: Great thanks.

PR #1724, opened 2021-09-27
Removed AL custom profiling code and added maya profiling

csyshing:
Our internal repo has removed the custom profiling code (namely `CodeTiming`) and updated to native Maya profiling.

Please notice that almost all the changes are in `plugin/al` **except one file** `lib/mayaUsd/nodes/proxyShapeBase.cpp`.


kxl-adsk (approved):
You have added quite a lot of new categories. This isn't a typical usage pattern, we suggest grouping events into bigger categories, like translators, proxy, drawing. It's not a show stopper for the merge of changes since new categories are only added to the AL plugin.

PR #1725, opened 2021-09-27
Make TranslatePrim update behaviour consistent with Proxy Shape update behaviour

csyshing:
Changes:
* call ProxyShapePostLoadProcess update/connect code when update requested via TranslatePrim
* TranslatePrim: remove double import/update/teardown
* added unit test

When an onObjectsChanged callback causes prims with schema translators to be called, if a translator plugin supports update, the following occurs:
 - update() is called on the translator
 - postImport() is called on the translator
 - generateUniqueKey() is called on the translator by context->updateUniqueKey()
 - any extraData plugins have their update and postImport methods called
 
We import any prims which weren't update-able because they didn't already exist in the scene, we now call the same code from TranslatePrim, so the behaviours are now consistent; have also removed a doulbe update/import/teardown call which was happening in TranslatePrim.

### DIFF
--- a/plugin/al/plugin/AL_USDMayaTestPlugin/py/testTranslators.py
+++ b/plugin/al/plugin/AL_USDMayaTestPlugin/py/testTranslators.py
@@ -181,6 +181,48 @@ class DeleteParentNodeOnPostImport(usdmaya.TranslatorBase):
     def exportObject(self, stage, path, usdPath, params):
         return
 
+class UpdateableTranslator(usdmaya.TranslatorBase):
+
+    actions = []
+
+    def initialize(self):
+        return True
+
+    def getTranslatedType(self):
+        return Tf.Type.Unknown
+
+    def needsTransformParent(self):
+        return True
+
+    def supportsUpdate(self):
+        return True
+
+    def importableByDefault(self):
+        return True
+
+    def importObject(self, prim, parent=None):
+        self.__class__.actions.append('import' + str(prim.GetPath()))
+        return True
+
+    def postImport(self, prim):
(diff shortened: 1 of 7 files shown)


plugin/al/plugin/AL_USDMayaTestPlugin/py/testTranslators.py
```diff
@@ -417,6 +458,20 @@ def test_deletion_of_parent_node_by_translator_does_not_crash_Maya(self):
...
+        '''
+
+        usdmaya.TranslatorBase.registerTranslator(UpdateableTranslator(), 'test')
+
+        stage = Usd.Stage.Open("../test_data/translator_update_postimport.usda")
+        stageCache = UsdUtils.StageCache.Get()
+        stageCache.Insert(stage)
+        stageId = stageCache.GetId(stage)
+        shapeName = 'updateProxyShape'
+        cmds.AL_usdmaya_ProxyShapeImport(stageId=stageId.ToLongInt(), name=shapeName)
```
kxl-adsk, line 473: How does this test work? I'm not sure I understand how we detect errors with this test? I don't see anything thrown or collected by the UpdateableTranslator.

kxl-adsk (approved):

kxl-adsk:
FYI. We have an artifactory server issue today and it caused PF failures. Will trigger another one tomorrow.

kxl-adsk:
@csyshing Please review PF results. Looks like one unit test failed on all windows builds.

csyshing:
@kxl-adsk , I might have found out why the test failed on Windows, the fix will be pushed to external soon!

PR #1726, opened 2021-09-27
Fix excluded prim paths

csyshing:
Ensure excluded prims (that do not need to be rendered) handled properly so that we don't get double up geo on viewport.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
@@ -318,20 +318,7 @@ void ProxyShape::constructGLImagingEngine()
 
             // delete previous instance
             destroyGLImagingEngine();
-
-            const auto& translatedGeo = m_context->excludedGeometry();
-
-            // combine the excluded paths
-            SdfPathVector excludedGeometryPaths;
-            excludedGeometryPaths.reserve(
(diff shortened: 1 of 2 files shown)


kxl-adsk (approved):

csyshing:
I checked the unit test log, the message is about schema generation error during configuration, on Windows + Maya 2019 only:
```
      File "P:/shared/Artifactory/Windows/USD/4ae4ff5/bin/usdGenSchema", line 48, in <module>
        from pxr import Plug, Sdf, Usd, Vt, Tf
      File "P:\shared\Artifactory\Windows\USD\4ae4ff5\lib\python\pxr\Plug\__init__.py", line 29, in <module>
        Tf.PreparePythonModule()
      File "P:\shared\Artifactory\Windows\USD\4ae4ff5\lib\python\pxr\Tf\__init__.py", line 89, in PreparePythonModule
        "." + moduleName, f_locals["__name__"])
      File "P:\shared\Artifactory\Windows\rundev-2019\202003131251\runtime\155eb96\runTime\bin\python27.zip\importlib\__init__.py", line 37, in import_module
    ImportError: DLL load failed: The specified procedure could not be found.
    CMake Error at lib/usd/schemas/CMakeLists.txt:139 (message):
      Schemas generation failed
```
All other build variants passed file.

I am not quite sure why this PR is causing such an error on Windows + Maya 2019 ?

seando-adsk:
@csyshing The 2019 build problem is a fault with our internal builds. I'm fixing it now, then we'll merge your PR.

PR #1727, opened 2021-09-27
Avoid opening duplicated usd stages on Maya file load

csyshing:
This is to avoid duplicated usd stages on Maya scene reopen.

It happens in cases when we have nested proxyShapes, where the children proxyShapes got loaded twice:
- Reopen Maya file, postFileRead callback would be run;
- Root proxyShape starts to load the stage;
- Root proxyShape's outStageData plug would be computed, which will request the children proxysShapes' outStageData being computed as well;
- Children proxyShapes will be requested to load their stages, thus postFileRead callback would be triggered again.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
@@ -1224,7 +1224,7 @@ void ProxyShape::loadStage()
         std::string fileString = TfStringTrimRight(file.asChar());
 
         TF_DEBUG(ALUSDMAYA_TRANSLATORS)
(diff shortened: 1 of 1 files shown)


kxl-adsk (approved):
Do you happen to know if this issue can happen with the Adsk plugin?

csyshing:
> Do you happen to know if this issue can happen with the Adsk plugin?

It might be - according to the original commit info from Miguel Gao, it happens when nested proxyShape is invoked, that means we cannot verify that until we fully migrate the "asset type" based Python translators workflow to ADSK plugin.
I will keep in mind to double check later.

seando-adsk:
@csyshing The 2019 build problem is a fault with our internal builds. I'm fixing it now, then we'll merge your PR.

PR #1728, opened 2021-09-27
Fill in missing export translator implementation

csyshing:
Python translator export never worked properly, but the mismatch between the wrapping and the Python implementation is causing errors that we would like to eliminate as a potential source of problems.

Where no export is supported, `canExport()` and `exportObject()` methods are not expected to be implemented in Python translators - leaving dummy methods which don't return the corret type will cause errors.

If implementing the method but return the incorrect type, you will get a runtime error like this:
```
TF_PYTHON_EXCEPTION: in TfPyConvertPythonExceptionToTfErrors at line 114 of ../pxr/base/lib/tf/pyError.cpp -- Tf Python Exception
TypeError: No registered converter was able to produce a C++ rvalue of type AL::usdmaya::fileio::translators::ExportFlag from this Python object of type bool
```

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/wrapTranslatorBase.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/wrapTranslatorBase.cpp
@@ -177,7 +177,12 @@ public:
     {
         MFnDependencyNode fn(obj);
         std::string       name(fn.name().asChar());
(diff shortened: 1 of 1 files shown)


kxl-adsk (approved):

csyshing:
The unit test failed for the same reason as #1726 and #1727 on Windows + Maya 2019, seems like a problem on the Jenkins machine? Any idea?

seando-adsk:
I'll have a look at the failing build. Yesterday we updated all builds to use USD v21.08.

PR #1729, opened 2021-09-27
Fix translator context on nested proxy shapes

csyshing:
This PR is related to loading nested proxyShape https://github.com/Autodesk/maya-usd/pull/1727 that the Python translator context can be maintained in nested proxyShape.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorBase.h
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorBase.h
@@ -31,6 +31,7 @@
 
 #include <functional>
 #include <iostream>
+#include <stack>
 #include <unordered_map>
 
 namespace AL {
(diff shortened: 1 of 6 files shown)


plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorBase.h
```diff
@@ -460,11 +466,30 @@ class TranslatorManufacture
...
+//----------------------------------------------------------------------------------------------------------------------
+/// \brief  used to prepare python translator context on creation and push it into stack, and pop
+/// from stack on destruction. \ingroup   translators
+//----------------------------------------------------------------------------------------------------------------------
+class TranslatorContextSetterCtx
+{
+public:
+    TranslatorContextSetterCtx(TranslatorContext::RefPtr context)
+    {
+        TranslatorManufacture::preparePythonTranslators(context);
```
kxl-adsk, line 490: I'm assuming this code is designed to run single-threaded? I'm wondering if we have somewhere code available to assert that we execute on main-thread...

kxl-adsk (approved):

plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorBase.h
csyshing: Yes, I think so. We have used this code for a long time internally but no threading issue / race condition problem so far.

PR #1731, opened 2021-09-28
Added USD pick mode fixes for VP2

jonjondev:
This PR adds support for pick mode settings in VP2 to toggle selection level between models, prims, and instances.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyDrawOverride.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyDrawOverride.cpp
@@ -58,10 +58,42 @@
 #include <pxr/base/arch/env.h>
 #endif
 
+#include <pxr/usd/usd/modelAPI.h>
+
 namespace AL {
 namespace usdmaya {
 namespace nodes {
 namespace {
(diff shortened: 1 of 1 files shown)


kxl-adsk:
We already have a selection by kind available in VP2RenderDelegate and in the UI
Screen Shot 2021-10-01 at 1 45 11 PM

The code responsible for this is in https://github.com/Autodesk/maya-usd/blob/302eba01a412fc856e69ed408d4e55a13ec0282c/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp#L1104

We hope you can adopt this workflow or let us know what's missing.

J-Mo63:
Hi @kxl-adsk, thanks for taking a look!

At the moment, the AL team is in the process of realigning our internal changes with this repo, and we want to preserve the history of the original committers where possible. This means that some of the changes being PR'd now will likely be overwritten by later changes as soon as these get merged.

Let me know if that's ok, but yes, we absolutely aim to share as much code as feasible with the ADSK plugin to smooth that migration process later down the road.

kxl-adsk (approved):
Apologies @J-Mo63, for some reason I thought this change was done in VP2RenderDelegate. I'm starting PF and once passed it should be good to go.

PR #1732, opened 2021-09-28
T tindj/maya 113606/validate xform token

renderRa:
Add Xform token validation for UsdTransform3dMayaXformStack


BranchHistory.md
```diff
@@ -0,0 +1,52 @@
+# Build Branch
```
ppt-adsk, line 1: Are you sure this file should be merged in?

ppt-adsk (changes requested):
Really unsure about all those commits, and the BranchHistory.md file.  We typically don't squash and merge, so all those commits will end up in the branch history, which seems undesirable.

BranchHistory.md
seando-adsk: @renderRa It looks like you created your branch from our internal **build** branch instead of dev. Please close this PR and fix your branch and then create a new PR.

renderRa: yes ill get that cleaned up. thanks.

renderRa:
closed request for cleaning up.

PR #1734, opened 2021-09-29
Added caching to LayerManager

J-Mo63:
This PR adds caching to the `LayerManager` node to avoid some very expensive Maya queries being run when editing begins.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/LayerManager.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/LayerManager.cpp
@@ -29,6 +29,7 @@
 #include <maya/MFnDependencyNode.h>
 #include <maya/MGlobal.h>
 #include <maya/MItDependencyNodes.h>
+#include <maya/MObjectHandle.h>
 #include <maya/MPlugArray.h>
 #include <maya/MProfiler.h>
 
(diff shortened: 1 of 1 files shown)


kxl-adsk:
I saw you have a number of fixes around LayerManager. Some of them were reasons for us to create a new copy of LayerManager in the core library (e.g. we needed all anonymous layers to be serialized to Maya file and not only session layer). Do you think with the upcoming changes you could actually adopt core [LayerManager](https://github.com/Autodesk/maya-usd/blob/302eba01a412fc856e69ed408d4e55a13ec0282c/lib/mayaUsd/nodes/layerManager.h)? The big benefit of going this path would help with migration from AL to Adsk plugin, fixes will benefit both plugins...and ultimately remove code duplication.

J-Mo63:
Hi @kxl-adsk!

Again, this PR is part of AL's alignment process, and will likely have further changes applied on top soon. I think that adopting the shared `LayerManager` is moving the right direction, and I'll bring it up with the team. Similarly to https://github.com/Autodesk/maya-usd/pull/1731, let me know if there's anything concerning otherwise, but this purely reflects the effort to catch up with the latest internal changes in a somewhat linear manner.

J-Mo63:
Can do @kxl-adsk. Will make the appropriate changes to the shared `layerManager` and add it to the PR.

PR #1735, opened 2021-09-29
T bailp/maya 112773/load on duplicate

pierrebai-adsk:


lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
```diff
@@ -70,8 +70,33 @@ void UsdUndoDuplicateCommand::execute()
...
+        auto policy = rules.IsLoadedWithAllDescendants(path)
+            ? UsdLoadPolicy::UsdLoadWithDescendants
+            : UsdLoadPolicy::UsdLoadWithoutDescendants;
+        stage->Load(_usdDstPath, policy);
+    } else if (primRule == UsdStageLoadRules::NoneRule) {
+        stage->Unload(_usdDstPath);
+    }
+
+    auto layer = stage->GetEditTarget().GetLayer();
+    bool retVal = PXR_NS::SdfCopySpec(layer, path, layer, _usdDstPath);
```
kxl-adsk, line 99: This makes a lot of sense. The only thing I wonder is why you setup the `_usdDstPath` prior to doing a copy? Does order matter?

pierrebai-adsk: Otherwise the copy might end-up loading the payload. If the source was not loaded, we do not want the copy to be loaded. Given how the notification system works, we could have ended-up with the payloaded loaded before the function returned and set the exclusion rule.

kxl-adsk (approved):
LGTM. As discussed offline, having a unit test for this change would be great.

kxl-adsk (approved):

PR #1736, opened 2021-09-30
Added USD Preview Surface for AL plugin

J-Mo63:
This PR adds a `usdPreviewSurface` for the AL plugin. Rather than the `usdPreviewSurface` node being registered by the first plugin to load, and then deregistered by that same plugin (such as with the `MayaUsdProxyShapePlugin` wrapper), the AL plugin registers its own, distinct preview surface node using the same implementation under the name of `alUsdPreviewSurface`.

Maya seems to treat them as separate nodes and even lists them as separate requirements in `.ma` files:

```ma
requires -nodeType "alUsdPreviewSurface" -dataType "pxrUsdStageData" "AL_USDMayaPlugin" "2.5.0";
requires -nodeType "usdPreviewSurface" "mayaUsdPlugin" "0.10.0";
```

Loading and unloading also seems to function well.

image

The AE template from the shared library is also copied across at build time and formatted accordingly.


JGamache-autodesk (approved):
The registration code follows exactly what was done in the Pixar and Autodesk plugins for UsdPreviewSurface, so the behavior will be the same. Thanks.

kxl-adsk:
One of our goals is is to consolidate and get to one of "each". I know that we already have two preview surface nodes and it was mostly caused by the fact that we wanted to rename the existing node and remove `pxr` prefix.

I would love to explore ways with you that will not require a new node. The main motivation would be to avoid (potential) future migration problems from AL to Adsk plugin. For example, there is no registered translator that will know how to export AL preview surface with a core translation framework.

One possibility mentioned by @JGamache-autodesk would be to make core UsdPreviewSurface a plugin on its own. We haven't fully explored this option, but would that work for your use case?

fabal:
Having it as a separate plugin would work for us. When do you think it could be available?

PR #1737, opened 2021-09-30
MAYA-113944 Enable primvar filtering to match HdSt behavior.

williamkrick:
This gives a substantial performance improvement by avoiding calls to _ComputeAndMergePrimvar made from UsdImagingGprimAdapter::UpdateForTime.


williamkrick:
After this change to filtering the performance of the animated scene improves and the next bottleneck is normal calculation performance.

kxl-adsk (approved):

JGamache-autodesk (approved):

PR #1755, opened 2021-10-13
MAYA-113077: componentTag IO to files

pierrebai-adsk:
Code for component tag support on behalf of Hans Rijpmeka.

### DIFF
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
@@ -1383,7 +1383,7 @@ MStatus UsdMayaMeshWriteUtils::exportComponentTags(UsdGeomMesh& primSchema, MObj
 
     const MFnDependencyNode depNodeFn(obj, &status);
     CHECK_MSTATUS_AND_RETURN_IT(status);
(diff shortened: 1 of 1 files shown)


plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -50,6 +50,9 @@ MStatus UsdMayaExportTranslator::writer(
     const MString&                    optionsString,
     MPxFileTranslator::FileAccessMode mode)
 {
+    MStatus s = MPxFileTranslator::writer(file, optionsString, mode);
```
ppt-adsk, line 53: Don't get this...  You're calling the base class writer(), but the Maya docs say this fails unconditionally, and I just looked at the code, and it does fail unconditionally...  And you're checking for not failure, so this is all a no-op.  I must be missing something.

lib/usd/translators/meshWriter.cpp
```diff
@@ -799,6 +799,10 @@ bool PxrUsdTranslators_MeshWriter::writeMeshAttrs(
             _GetSparseValueWriter());
     }
 
+    if (true /*exportArgs.exportComponentTags*/) {
```
ppt-adsk, line 802: Do we want to fix this before we merge?

Issue #640, opened 2020-07-09
Support attribute on MayaReference prim to toggle -mergeNamespacesOnClash flag


ika-rporter:
**Is your feature request related to a problem? Please describe.**
The  `UsdMayaTranslatorMayaReference::LoadMayaReference` function is hard coded to call `-mergeNamespacesOnClash false`. This means that if you create an `AL_usdmaya_ProxyShape` node in a namespace (eg, `my_asset_1:proxy`), that is proxy for an asset, it is not possible for the rig to load in the same namespace. To avoid an off-by-one error (eg, `my_asset_2:rig`), you have to salt the namespace (eg, `_my_asset_1:rig`) or nest the namespace (eg, `rigs:my_asset_1:rig)`. Either option introduces an undesired complexity in naming conventions. 

**Describe the solution you'd like**
The `UsdMayaTranslatorMayaReference::LoadMayaReference` function searches for an attribute on the MayaReference prim (eg, `maya_mergeNamespaceOnClash`), whose value is used to set the  `-mergeNamespacesOnClash` flag as `true/false`. 

The default is `false` to maintain legacy behavior.

**Additional context**
The reason for loading the proxy asset in a namespace is so that the rig can load in the same namespace. If the asset is named "my_asset_1" in the root namespace, then it is not possible to have the rig use "my_asset_1" as a namespace since namespaces must be globally unique.

PR #1759, opened 2021-10-16
Support MayaReference prim behavior in mayaUsdProxyShape

yantor3d:
This PR adds support for the `MayaReference` prim to the `mayaUsdProxyShape` node using the `UsdMayaPrimUpdaterRegistry`.

# Problem
Most studios have are still using the ALUSD Proxy for their Maya based shot workflows because the MayaUSD Proxy does not yet support the MayaReference prim. 

My understanding of the problem, based on a conversation with @wtelford, is that the existing behavior in the ALUSD Proxy could not be matched because one cannot parent a DAG node to the UFE Prim representation. 

I don't feel that should be a blocking issue, though. When using any of the USD export commands, you will get extra prims in the layer because of the proxy node that will have to be removed. Or you can use the `-exportRoots` flag, in which case you get a layer with the `/Geom` prim as the root. In either case, it does not matter if the reference is parented into the "correct" position in the hierarchy. 

# Changes
 - Extend mayaUsdProxyShape node to use prim updater registry for managing custom prim schema (eg, MayaReference)
 - Modify MayaReference translator (UsdMayaTranslatorMayaReference) reference node naming algorithm. 
   - Before it would use the full DAG path of the parent node. This worked for the ALUSD Proxy because it would create a transform hierarchy that managed the prim path. This would not work for the MayaUSD Proxy without involve UFE. The algorithm now uses the full DAG path of the parent transform of the proxy shape node, plus the full prim path of the MayaReference prim. The prim path is only appended if necessary to prevent it from appearing twice in the name of a reference node associated with an ALUSD Proxy.
  
# Tests
 - Added tests coeravege for  `mayaUsdProxyShape` MayaReference behavior 
 - Tests for `AL_usdmaya_ProxyShape` MayaReference behavior still pass.


kxl-adsk:
@yantor3d Addressing the issue you are mentioning is on our radar and we will be pushing new changes **shortly**. It's possible that some of the changes may conflict, but we will review your PR once our internal changes are visible to the public.

yantor3d:
Hi @kxl-adsk - I've looked at #1789 - it does remove Push, Pull, & Clear from the `UsdMayaPrimUpdater`, but as I mentioned in my comment, those were unused in `dev` and my PR will just put them back.

I do have a concern about the magic strings in [PxrUsdTranslators_MayaReferenceUpdater::PushCopySpecs](https://github.com/Autodesk/maya-usd/pull/1789/files?diff=split&w=0#diff-9baafe756fd44b2f8f2f1aa6a9ca229bc5bca5827dda2c1355ac4fbef875317cR71) lines 71-80. The variantSet "animVariant" and the variant names "cache" and "rig" aren't referenced anywhere else in the PR, and only appear in two other places in the entire module - the AL End to End Tutorial [prep_anim_shot](https://github.com/Autodesk/maya-usd/blob/600c31abcd68ef3fe0047b28fb5f33f6edd0872f/plugin/al/tutorials/endToEndMaya/tutorial_scripts/prep_anim_shot_s00_01.py) script and the [README](https://github.com/Autodesk/maya-usd/blob/600c31abcd68ef3fe0047b28fb5f33f6edd0872f/plugin/al/tutorials/endToEndMaya/README.md). It feels to me like this is some studio specific naming conventions. But it's shouldn't interfere with my PR so I wasn't sure if I should raise it there.

yantor3d:
I've merged the latest commits to `dev` into my branch. There was a minor collision in `mayaReferenceUpdater.h/cpp`, but nothing terrible. 

However, it seems the `dev` branch is not in a buildable state - `lib\mayaUsd\ufe\UsdPathMappingHandler.h` includes the `ufe/pathMappingHandler.h` header, which does not exist in the public repo. I am unable to verify that the changes made in this PR still work. In addition, it seems the new Copy | Push/Pull data flow may supersede my changes. Is this an accurate assessment of the situation?

ppt-adsk:
> However, it seems the `dev` branch is not in a buildable state - `lib\mayaUsd\ufe\UsdPathMappingHandler.h` includes the `ufe/pathMappingHandler.h` header, which does not exist in the public repo. I am unable to verify that the changes made in this PR still work.   [...]

Apologies for that, will fix ASAP.

ppt-adsk:
Hi @yantor3d, after a bit of digging the UFE support for pathMappingHandler.h went into Maya Preview Release 128, and we are currently are on Preview Release 129.  Could it be that you're trying to build maya-usd against a fairly old Maya Preview Release?  Typically we only strive to support the current Maya Preview Release.

yantor3d:
>  Could it be that you're trying to build maya-usd against a fairly old Maya Preview Release?

Ah! That makes sense. I've been building against 2022.1, I think. I'll download the latest PR and give that a go.

yantor3d:
I was able to successfully compile against PR129. 

Upon doing so I discovered that recent commits already introduced support for the Maya Reference prim to the mayaUsdProxyShape, with the referenced DAG root(s) faux "parented" to the prim as part of the new push/pull/edit workflow. 

I feel that this obsoletes this PR. 

However, as I understand the new workflow it will require Maya 2023 and UFE 3.0. Do we want to try and repurpose the work I have done for use in Maya 2020/2022, since my changes did not require UFE at all?

kxl-adsk:
It is definitely an option to consider. Another option to consider is to loosen the UFE 3.0 dependency and have the key components (Updaters & UpdaterManager) available in Maya 2020/2022. We may have slightly affected workflows (marking menus and outliner display) but the core functionality would remain the same.

Do you know any reason why the latter option would not be acceptable?

yantor3d:
> Do you know any reason why the latter option would not be acceptable?

I do not.

kxl-adsk:
Once the foundation will stabilize, we will look into removing some of the UFE v3 restrictions. This will give you access to updater and updater manager. For now, let's close this PR.

Having said that, I would like to make sure we are not ignoring any of the changes you made to the MayaReference translator and details about your workflow with rigs. Things like PR #1620 would be good for us to learn about.

yantor3d:
Thanks Krystian - I'll keep an eye out for future developments, and I'll definitely put in a PR for a feature request I filed a while ago - [#640](https://github.com/Autodesk/maya-usd/issues/640) that relates to rigs and a multi-proxy workflow.

PR #1760, opened 2021-10-18
Updated LayerManager to purge cached self on scene reset

jonjondev:
This change to the AL plugin's `LayerManager` was authored to solve a multi-threaded-related crash when using a somehow "valid" and "alive" `MObjectHandle` for the `LayerManager` cache. The change ensures that the cache is cleared on each new scene, resolving the issue.

I have also gone ahead and made similar changes to the shared `LayerManager` as requested in PR https://github.com/Autodesk/maya-usd/pull/1734. Let me know if there are any changes you think would be appropriate for the shared implementation - I've tried to base it on what was already there.

### DIFF
--- a/lib/mayaUsd/nodes/layerManager.cpp
+++ b/lib/mayaUsd/nodes/layerManager.cpp
@@ -58,6 +58,7 @@
 
 namespace {
 static std::recursive_mutex findNodeMutex;
+static MObjectHandle        layerManagerHandle;
 
 // Utility func to disconnect an array plug, and all it's element plugs, and all
 // their child plugs.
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -137,14 +138,17 @@ getFileFormatForLayer(const std::string& identifierVal, const std::string& seria
...
-    MFnDependencyNode    fn;
-    static MObjectHandle layerManagerHandle { MObject::kNullObj };
+    MFnDependencyNode fn;
     if (layerManagerHandle.isValid() && layerManagerHandle.isAlive()) {
         MObject mobj { layerManagerHandle.object() };
         if (!mobj.isNull()) {
             fn.setObject(mobj);
             return static_cast<MayaUsd::LayerManager*>(fn.userNode());
+        } else {
+            MGlobal::displayInfo("LayerManager::findNode cache got null MObject");
```
kxl-adsk, line 148: These traces doesn't tell much to the end-user, please remove them.

kxl-adsk (changes requested):
Left one request for a change in comments. Additionally, I'm curious about this multi-threaded-related crash that you were getting. Do you happen to have any more info? Call-stack maybe?

kxl-adsk (approved):

PR #1774, opened 2021-10-20
Pr/hydra updates to pxr hd renderer

dj-mcg:
Use physical parameters rather than OpenGL matrices for the HdCamera


dj-mcg:
Oops, the hydra code this depends on is old but not old enough! More API guarding incoming...

PR #1776, opened 2021-10-20
MAYA-114170 - Maya freeze on playback when using a invalid stageCacheId

spinell-adsk:
Make a copy of the _unsharedStageRootLayer to avoid callstack overflow when maya caching is enable caused
by "StageChange" event.


alsoussi:
@kxl-adsk I think its a stack overflow issue, the stack might not be so useful. I helped Luc on this. The main issue is when we call the remapSublayerRecursive, this usually works fine without caching because the stages get removed and recreated everytime, but with caching we are keeping the stages for all timeframes, so when we call the function "remapSublayerRecursive" on the layer (which is unsharedRoot, and part of some cached stages) it notifies a change, which then calls some function that asks the node for the stage which in turn executes the same loop. Granted that part of the stack might be useful so @spinell-adsk maybe you can include that. 

In any case my suggestion was to simply copy the layer everytime, its not the most efficient, but guarantees safety. And since the unsharedRoot layer is the only one being copied, and we (kinda) control it, and its pretty small I feel its an ok fix. Let me know your thoughts or if you have another idea.

Issue #1542, opened 2021-07-02
[MAYA-112773] Duplicating a model should bring it in loaded


dj-mcg:
Our users noticed that when duplicating a model by Ctrl+D in the Outliner, the new model brought into the scene unloaded. They would prefer to have them loaded.

After discussions with Will & Krystian we suspect this is just a bug.

wtelford:
@santosg87 Please log internally if you can reproduce this.

santosg87:
hi @dj-mcg i am having trouble reproducing this. i have tried a few different instances and workflows, but when i duplicate, i always get my models loaded. would you be able to add a bit more details on the behavior they are seeing? are they duplicating prims that are not loaded to begin with? are the unloaded prims still visible?

dj-mcg:
Hi @santosg87 - I just tried creating a stage on a new layer, adding a prim, and duplicating. Sure enough, this comes in loaded. It must be specific to duplicating referenced models. Let me try to collect more details surrounding the issue.

It's also worth noting I'm testing this in base Maya 2022, so if there have been any fixes only available in PRs, I haven't picked those up. I've been waiting for folks to iron out licensing issues but it seems like I should just circumvent all of that and set up a local install.

dj-mcg:
As far as I can tell, the model/prim you are duplicating must have a payload (ie. can be loaded/unloaded) and must be parented under an Xform prim. The duplicated model will come loaded only if you've already done "Load with Descendants" on the parent. The load state of the source model won't matter - you can duplicate an unloaded prim, and if the parent has been loaded w/ descendants, the duplicated model comes in loaded.

If you don't "Load With Descendants" on the parent prim, and instead only "Load" the source model itself, the duplicated model will come in unloaded.

wtelford:
Hey @dj-mcg,
I'd like to clarify your comment "The expectation of our users is that the duplicated model always comes in loaded."

My thought is:
When duplicating a prim, the duplicated prim should have the same load state as the source prim we are duplicating, not always be loaded, correct?

If source is unloaded, the duplicate should be unloaded.
If source is loaded, then the duplicate should be loaded. (regardless of ancestor load state)

If I'm an artist who has unloaded a prim because it is too heavy, I wouldn't want to take the hit for loading that object just because I duplicated it, but if I am working with a prim that is already loaded and I duplicate it, I would fully expect to see that prim to be loaded as well.

dj-mcg:
[...]

PR #1778, opened 2021-10-22
MAYA-112773 fix load on duplicate

pierrebai-adsk:
Now duplicate all rules that controlled loading starting at the source path.


dj-mcg:
Looks like this fixed the behavior observed here:
https://github.com/Autodesk/maya-usd/issues/1542

Code looks fine to me, but I'll defer to others that may have more state on the code

PR #1792, opened 2021-10-29
Added file format attribute to LayerManager and improved format detection for serialised layers

J-Mo63:
This PR allows for the extension to now be retrieved using USD's API instead of string parsing, and adds the layer's file format as a serialisation attribute (along with the current identifier and serialised content).

This is to help deserialisation in the case that the identifier does not resolve without a proper context (which is the case with the layer manager).

### DIFF
--- a/lib/mayaUsd/nodes/layerManager.cpp
+++ b/lib/mayaUsd/nodes/layerManager.cpp
@@ -498,11 +498,15 @@ MStatus addLayerToBuilder(
     MDataHandle layersElemHandle = builder.addLast(&status);
     CHECK_MSTATUS_AND_RETURN_IT(status);
     MDataHandle idHandle = layersElemHandle.child(lm->identifier);
+    MDataHandle fileFormatIdHandle = layersElemHandle.child(lm->fileFormatId);
     MDataHandle serializedHandle = layersElemHandle.child(lm->serialized);
     MDataHandle anonHandle = layersElemHandle.child(lm->anonymous);
 
     idHandle.setString(UsdMayaUtil::convert(layer->GetIdentifier()));
     anonHandle.setBool(isAnon);
+
+    auto fileFormatIdToken = layer->GetFileFormat()->GetFormatId();
+    fileFormatIdHandle.setString(UsdMayaUtil::convert(fileFormatIdToken.GetString()));
 
     std::string temp;
     if (!stubOnly && ((exportOnlyIfDirty && layer->IsDirty()) || !exportOnlyIfDirty)) {
@@ -752,9 +756,11 @@ void LayerDatabase::loadLayersPostRead(void*)
     MPlug                       allLayersPlug(lm->thisMObject(), lm->layers);
     MPlug                       singleLayerPlug;
     MPlug                       idPlug;
+    MPlug                       fileFormatIdPlug;
     MPlug                       anonymousPlug;
     MPlug                       serializedPlug;
     std::string                 identifierVal;
+    std::string                 fileFormatIdVal;
     std::string                 serializedVal;
     SdfLayerRefPtr              layer;
     std::vector<SdfLayerRefPtr> createdLayers;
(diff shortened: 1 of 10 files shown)


J-Mo63:
Hi @kxl-adsk, just saw a missing Clang format blocked the preflight builds. It's been formatted now if you're happy to rerun the CI.

plugin/al/plugin/AL_USDMayaTestPlugin/py/testLayerManager.py
```diff
@@ -143,5 +143,75 @@ def test_sessionLayerSerialisation(self):
         os.remove(_tmpMayafile.name)
 
 
+    def test_multipleFormatsSerialisation(self):
+        """Tests multiple dirty layers with various formats can be saved and restored.
+
+        """
+        _, rootLayerPath = tempfile.mkstemp(suffix=".usda")
+        _, tmpMayafilePath = tempfile.mkstemp(suffix=".ma")
```
kxl-adsk, line 151: We have moved away from using temp files, because:
* it's harder to debug
* required periodic temp folder cleanup

For this reason, we use explicit names and local to the test area within the build folder. Generated files are easy to discover and the build folder is more often cleared than the temp folder.

Here is an example - https://github.com/Autodesk/maya-usd/blob/dev/test/lib/usd/translators/testUsdExportMaterialX.py#L43

kxl-adsk (changes requested):
Left small request for a change in the test. You will also need to look at the PF results...there is an error in the test.

plugin/al/plugin/AL_USDMayaTestPlugin/py/testLayerManager.py
J-Mo63: Hi @kxl-adsk, looks like PF checks are now passing, and I've updated the test to avoid tempfiles (we'll keep this in mind going forward). Let me know if there's anything else needed.

plugin/al/plugin/AL_USDMayaTestPlugin/py/testLayerManager.py
```diff
@@ -17,37 +17,50 @@
...
 
 import os
-import tempfile
+import shutil
 import unittest
 
 from maya import cmds
 from AL.usdmaya import ProxyShape
 
-import fixturesUtils
```
kxl-adsk, line 26: We use `fixturesUtils` all over the place. What was the reason to remove it from AL tests?

lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -1027,6 +1026,16 @@ MStatus LayerManager::initialize()
         stat = addAttribute(identifier);
         CHECK_MSTATUS_AND_RETURN_IT(stat);
 
+        fileFormatId
```
kxl-adsk, line 1029: Since we don't share the same implementation of LayerManager, it would be good to add a test for this new addition to validate the core implementation as well. I won't make it a blocker for this change, but please open a new PR with it.

kxl-adsk (approved):

plugin/al/plugin/AL_USDMayaTestPlugin/py/testLayerManager.py
J-Mo63: Internally we have not set up our testing to take advantage of the module, however that transition is underway and a PR will be up for it at some point. I'll also be removing tempfile usage across AL tests.

kxl-adsk: That’s great to hear. Looking forward to this PR.

kxl-adsk: @J-Mo63 Looks like I didn't notice something in your change. All our tests are writing files to the build folder. I just run the new tests locally and you write the source folder...this means that now I always have changes in the source tree after running tests, e.g. 
```
git status
HEAD detached at origin/dev
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	plugin/al/plugin/AL_USDMayaTestPlugin/py/testLayerManagerOutput/
```
 Please adjust it to match what other tests are doing.

J-Mo63: Ah yes that's no good at all. The switch to `fixturesUtils` is on the way.

PR #1794, opened 2021-10-29
MAYA-114519 Make Shader reader/writer test self contained

boudrey:
Copying the MaterialX tests files so that the test is independent.


dj-mcg (approved):
Looks good to me, thanks so much for the change!

PR #1796, opened 2021-10-29
Adding definition for "kCreateStage" string constant

Vsevolod1983:
"kCreate" string definition is missing in mayaUSDRegisterStrings.mel file which causes the following runtime error upon click on "Universal Scene Descriptor(USD)" -> "Stage From File" menu item:

// Error: file: C:/Program Files/Autodesk/Maya2022/scripts/others/getPluginResource.mel line 66: Plug-in resource lookup failed, resource ("mayaUsdPlugin","kCreate") is not registered

### DIFF
--- a/plugin/adsk/scripts/mayaUSDRegisterStrings.mel
+++ b/plugin/adsk/scripts/mayaUSDRegisterStrings.mel
@@ -29,7 +29,7 @@ global proc mayaUSDRegisterStrings() {
     register("kCreateUsdStageFromFile", "Create USD Stage from File");
     register("kCreateUsdStageFromFileOptions", "Create USD Stage from File Options");
     register("kCreateStageFromFile", "Create Stage from File");
-    register("kCreate", "Ok");
+    register("kCreateStage", "Create");
     register("kDefaultPrim", "Default Prim");
     register("kDefaultPrimAnn", "As part of its metadata, each stage can identify a default prim. This is the primitive that is referenced in if you reference in a file.");
     register("kExcludePrimPaths", "Exclude Prim Paths:");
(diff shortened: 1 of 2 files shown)


plugin/adsk/scripts/mayaUSDRegisterStrings.mel
```diff
@@ -29,6 +29,7 @@ global proc mayaUSDRegisterStrings() {
     register("kCreateUsdStageFromFile", "Create USD Stage from File");
     register("kCreateUsdStageFromFileOptions", "Create USD Stage from File Options");
     register("kCreateStageFromFile", "Create Stage from File");
+    register("kCreate", "Ok");
```
seando-adsk, line 32: This **kCreate** ID is already being registered by the Layer Editor here: https://github.com/Autodesk/maya-usd/blob/dev/lib/usd/ui/layerEditor/stringResources.h#L58

I think the problem you are having is because you aren't building with Qt, and thus no layer editor. Which is a supported build option.

However this ID is being used by the plugin here: https://github.com/Autodesk/maya-usd/blob/dev/plugin/adsk/scripts/mayaUsd_createStageFromFile.mel#L224

That is the error you are seeing. However note the string should be "Create" not "Ok".  So what I would suggest is to change this to:
```suggestion
    register("kCreateStage", "Create");
```
And then change line 224 in `mayaUsd_createStageFromFile.mel` to use kCreateStage.

Vsevolod1983: Yes, thank you.
Changed.

seando-adsk (approved):

kxl-adsk:
@Vsevolod1983 Thank you for contributing a fix. Your PR is ready for merge, but we don't have a signed CLA to cover your contribution. Please follow https://github.com/Autodesk/maya-usd/blob/dev/doc/CONTRIBUTING.md and provide us with a signed copy of CLA.

Vsevolod1983:
Hello @kxl-adsk !
I've just sent signed CLA to usdformaya.contributor.agreements@autodesk.com
Should I send it to you as well? If yes, which email address I should use ?
Thank you!

kxl-adsk:
Already replied...and merged the change. Thank you for this contribution!

PR #1798, opened 2021-11-01
MAYA-114428 Compile mayapy on Linux and OSX to avoid search path issues.

boudrey:
Set Frameworks to force the test to take Python from the correct path.


boudrey:
This is a required fix for OSX issues. The issue in PF can only be fixed by modifying build.py to exclude the test "testProxyShapeLiveSurface" that has been broken by a change in Maya in the master branch.

kxl-adsk (approved):

PR #1800, opened 2021-11-02
Fix textures missing on OS/X with V2 Lighting

JGamache-autodesk:
The OS/X shader compilator support only literals for const initializers.


JGamache-autodesk:
Logged internally as MAYA-114588.

williamkrick (approved):

JGamache-autodesk:
Ready for merge. The failed preflight is a separate issue currently under investigation.

PR #1803, opened 2021-11-04
Fix OpenGL initialization to restore GPU Compute functionality.

williamkrick:


lib/mayaUsd/render/vp2RenderDelegate/meshViewportCompute.cpp
```diff
@@ -1303,6 +1304,11 @@ bool MeshViewportCompute::execute(
 
     prepareAdjacencyBuffer();
 
+    if (!hasOpenGL()) {
```
williamkrick, line 1307: Fix the expression, and move the initialization code before any OpenGL calls are made. I think in earlier versions of USD OpenGL was initialized for us, and so the old code never did anything useful. Now that OpenGL is not initialized it's obvious the old version was wrong.

lib/mayaUsd/render/vp2RenderDelegate/meshViewportCompute.cpp
```diff
@@ -277,6 +277,12 @@ void MeshViewportCompute::createConsolidatedTopology(TopologyAccessor getTopolog
 
             HdMeshTopology& sourceTopology = getTopology(sourceMeshSharedData);
 
+            if (sourceIndex == 0) {
```
williamkrick, line 280: This is to fix the assertions that follow from here. In some scenes I get different subdivision schemes than Catmull Clarke, so instead of hard coding the scheme make the scheme of the consolidated topology the same as the scheme of the first consolidated source, and then verify that all the other items have the same scheme.

lib/mayaUsd/render/vp2RenderDelegate/meshViewportCompute.cpp
```diff
@@ -1324,11 +1330,15 @@ bool MeshViewportCompute::canConsolidate(const MPxViewportComputeItem& other) co
     // If the compute has executed then the data to be consolidated will already
     // be smoothed. Smoothed items can only consolidate with other smoothed items.
     return hasExecuted() == otherMeshViewportCompute->hasExecuted()
+        && _meshSharedData->_renderingTopology.GetScheme()
```
williamkrick, line 1333: Only allow items that share the same values for scheme, orientation and refine level to consolidate together.

kxl-adsk (approved):

PR #1804, opened 2021-11-04
T bailp/maya 114111/diff merge prim over pull push 01 v2

pierrebai-adsk:
Implement minimal merge of modified data when merging back to USD after editing in Maya.

The algorithm is separated into layers to make it more reusable and easier to test. Each subsequent layer is built from the preceeding layer. The layers are:
- Low-level diff'ing of raw data, like array of floating-point values, for example.
- Diff'ing of USD data, in increasing level of complexities: VtValue, VtDictionary, UsdAttribute, UsdRelationship, etc all the way up to UsdPrim.
- Minimal merging of modified prims into baseline (that is, merge target) prims. This takes advantage of the existing SdfCopySpec() function with custom callbacks.

The diff'ing is slightly more sophisticated than what is used presently during merging: it can detect more than merely that values have changed, but can return such results as subset, super-set, reordered, etc. These will be used for more sophisticated list editing merge in a subsequent work.

All layers have extensive unit testing.

Areas to be later improved:
- Merging can leave entirely empty "over". We could later add a clean-up pass to remove empty overs.
- Lists are always entirely re-written. We could use later prepend and append in the future.
- Attributes absent from the modified source prim are removed from the target prim. This is correct behaviour as far as the logic of the Merge-to-USD goes, but leads to potential loss of attributes that are foreign to Maya, that will never be created by Maya and will thus not appear in the prim created by Maya. Fixing this would require being able to detect that an absent attribute is unknown to Maya. This requires having an extensive list of all possible attributes that can be generated by Maya and avoiding removing attributes that are unknown.
- Merging to an intermediary layer currently ignores upper layers. The code support ignoring upper layer or not, but this option is not made accessible to the user. We could later add a merge-to-USD user-accessible option to control the bahaviour.
- Merging to an intermediary layer while ignoring upper layers could have a performance impact since it trigger composition arc reconsolidation. No performance testing has been done to verify if there is a performance cost.


pierrebai-adsk:
build!

pierrebai-adsk:
build!

lib/mayaUsd/CMakeLists.txt
```diff
@@ -144,6 +144,7 @@ target_include_directories(${PROJECT_NAME}
         ${PXR_INCLUDE_DIRS}
         $<$<BOOL:${UFE_FOUND}>:${UFE_INCLUDE_DIR}>
         ${CMAKE_BINARY_DIR}/include
+        ${PROJECT_SOURCE_DIR}/..
```
seando-adsk, line 147: Why did you add this? You should be able to access your new headers by following the same include standard as other files from the same project. For example:
```
#include <mayaUsdUtils/MergePrims.h>
```

test/lib/usd/utils/test_DiffValuesHalfVecs.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+#include <gtest/gtest.h>
+
+using namespace PXR_NS;
+using namespace MayaUsdUtils;
+
+//----------------------------------------------------------------------------------------------------------------------
+TEST(DiffValuesHalfVecs, vec2hCompareValuesSame)
+{
+    VtValue    baselineValue(GfVec2h(1, 12));
```
ppt-adsk, line 26: Probably doesn't matter at all, but I was surprised to see that in the half float vector test the only test values you have are integers :)

ppt-adsk (changes requested):
A very impressive amount of work, and many thanks for the thorough testing.  Only minor comments, otherwise looks good to me.

lib/mayaUsd/CMakeLists.txt
pierrebai-adsk: I'll retry, as I added this early on and don't remember what the error was.

seando-adsk (approved):
I only looked at the cmake changes and they look good now.

ppt-adsk (approved):

PR #1810, opened 2021-11-08
Fix hierarchy handler chain of responsibility child filter bug.

ppt-adsk:


ppt-adsk:
Fix done with Test Driven Development (TDD): test written first, test failed, fix implemented, test passes.

lib/mayaUsd/ufe/PulledObjectHierarchyHandler.cpp
```diff
@@ -69,7 +69,10 @@ Ufe::SceneItem::Ptr PulledObjectHierarchyHandler::createItem(const Ufe::Path& pa
     return _mayaHierarchyHandler->createItem(path);
 }
 
-Ufe::Hierarchy::ChildFilter PulledObjectHierarchyHandler::childFilter() const { return {}; }
+Ufe::Hierarchy::ChildFilter PulledObjectHierarchyHandler::childFilter() const
+{
+    return _mayaHierarchyHandler->childFilter();
```
ppt-adsk, line 74: In the chain of responsibility we cannot assume that Maya handlers have no child filters: the proxy shape hierarchy handler adds a child filter flag for inactive prims, since its children are USD prims.

seando-adsk: Looks good.

PR #1814, opened 2021-11-08
MAYA-112228 Use Adaptors for custom schema import-export

JGamache-autodesk:
Adds custom API schema importers and exporter capabilities to existing UsdMayaAdaptor class. This can be done two ways:
- For simple attribute-to-attribute mapping we offer a way to Apply/Remove/Edit schemas using custom generated UsdMayaAttributeAdaptor
- For more complex work, we offer a copyToPrim and copyFromPrim callbacks that allow doing all the import/export work in a single function call
The schema API export and import call have been augmented to support animation, and an apiSchemas parameter was added to import job the replicate the export workflow.

Added the concept of jobContext to import and export code paths.
These contexts can be registered at will and will add/modify import and export job context parameters

Added Python wrappers for these new concepts
An example of use can be found in the test\lib\usd\translators\testUsdExportSchemaApi.py unit test.


lib/mayaUsd/fileio/primReaderContext.h
```diff
@@ -90,6 +102,9 @@ class UsdMayaPrimReaderContext
     // used to keep track of prims that are created.
     // for undo/redo
     ObjectRegistry* _pathNodeMap;
+
+    // Tracks new nodes:
+    std::shared_ptr<ObjectRegistry> _trackedNewMayaNodes;
```
kxl-adsk, line 107: Can we switch to a small vector of objects instead of a map? We know that this container will not have many elements and would be great to leverage this information.

lib/mayaUsd/fileio/primReaderContext.h
```diff
@@ -90,6 +102,9 @@ class UsdMayaPrimReaderContext
     // used to keep track of prims that are created.
     // for undo/redo
     ObjectRegistry* _pathNodeMap;
+
+    // Tracks new nodes:
```
kxl-adsk, line 106: Please leave a comment explaining why we are using a shared pointer.

lib/mayaUsd/fileio/utils/adaptor.cpp
```diff
@@ -58,8 +88,26 @@ UsdMayaAdaptor::UsdMayaAdaptor(const MObject& obj)
...
+    const UsdMayaPrimReaderArgs& jobImportArgs,
+    UsdMayaPrimReaderContext&    context)
+    : _jobImportArgs(&jobImportArgs)
+    , _jobImportContext(&context)
+{
+}
+
 UsdMayaAdaptor::operator bool() const
 {
+    if (_jobImportArgs && _jobImportContext) {
```
kxl-adsk, line 107: Can you document logic in here? I will take a few cycles to understand all the modes adaptor can be in.

lib/mayaUsd/fileio/utils/adaptor.h
```diff
@@ -124,203 +364,14 @@ PXR_NAMESPACE_OPEN_SCOPE
...
-        const UsdPrimDefinition* GetSchemaDefinition() const;
-
-    private:
-        /// Gets the name of the adapted Maya attribute for the given attribute
-        /// name. The name may come from the registered aliases if one
-        /// exists and is already present on the node.
-        std::string _GetMayaAttrNameOrAlias(const TfToken& attrName) const;
-    };
+    MAYAUSD_CORE_PUBLIC
+    UsdMayaAdaptor(const MObject& obj, const UsdMayaJobExportArgs* jobExportArgs);
```
kxl-adsk, line 371: Would be great to document these two new constructors.

kxl-adsk (commented):
Great job, Jerry. My last comment will be only about documentation - we will need to help developers discover how to work with API schema adaptor plugins. A white paper that describes the unit test could be a great start.

kxl-adsk (approved):

PR #1823, opened 2021-11-10
Fix UV export test

pierrebai-adsk:
Fix the test to not expect hard-coded UV set sizes.

### DIFF
--- a/test/lib/usd/translators/testUsdExportUVSets.py
+++ b/test/lib/usd/translators/testUsdExportUVSets.py
@@ -119,7 +119,7 @@ class testUsdExportUVSets(unittest.TestCase):
                 uvSetName)
             meshFn.assignUVs([4, 4, 4, 4, 4, 4], [0, 1, 2, 3] * 6, uvSetName)
             (numUVShells, shellIndices) = meshFn.getUvShellsIds(uvSetName)
-        self.assertEquals(numUVShells, 1)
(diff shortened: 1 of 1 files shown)


test/lib/usd/translators/testUsdExportUVSets.py
```diff
@@ -97,38 +105,36 @@ def setUpClass(cls):
...
-        assert(numUVShells == 1)
+        if len(uArray) != 4 or numUVShells != 1:
+            meshFn.clearUVs(uvSetName)
+            meshFn.setUVs(
+                [0.0, 1.0, 1.0, 0.0],
+                [0.0, 0.0, 1.0, 1.0],
+                uvSetName)
+            meshFn.assignUVs([4, 4, 4, 4, 4, 4], [0, 1, 2, 3] * 6, uvSetName)
+            (numUVShells, shellIndices) = meshFn.getUvShellsIds(uvSetName)
+        self.assertEquals(numUVShells, 1)
```
JGamache-autodesk, line 122: assertEquals is [deprecated](https://docs.python.org/3/library/unittest.html#deprecated-aliases). Can you use assertEqual instead?

JGamache-autodesk (approved):

pierrebai-adsk:
Single failure in preflight: comm channel was cut-off for some reason.

JGamache-autodesk (approved):

Issue #1351, opened 2021-04-19
Maya 2022 links to missing Python on Big Sur, causing CMake builds to fail


dgovil:
Maya 2022 symlinks to Python 3.7 on Big Sur. I believe Maya 2022 is built on Catalina where this is possible, but this is not the case on Big Sur.

The [python.cmake](https://github.com/Autodesk/maya-usd/blob/dev/cmake/python.cmake) file tries to find the python3 file at this path `Maya.app/Contents/Frameworks/Python.framework/Versions/Current/bin/python3` and errors out during build because it can't execute it further down.

Setting this to point to `mayapy` doesn't seem to satisfy CMake either, because the default CMake `FindPython.cmake` does not respect `mayapy` as a valid interpreter name.

A simple fix on our end has been to modify `python.cmake` in our internal fork of this repo and adding this at the head, and passing in `-DMAYAPY_PATH=path/to/mayapy` to CMake. This short circuits the logic in the file and lets us work through the builds successfully.

```cmake
if(DEFINED MAYAPY_PATH)
    MESSAGE(STATUS "Setting Python_EXECUTABLE to MayaPy ${MAYAPY_PATH}")
    set(Python_EXECUTABLE ${MAYAPY_PATH})
    set(Python_Interpreter_FOUND TRUE)
    return()
endif()
```

While this unblocks us internally, I do think it's perhaps an issue that could be wider addressed in the main CMake definition for this project to simply accept mayapy as the project. I also think this will be problematic long term (in general) for Maya's python libraries to assume that the OS bundles Python anymore to symlink against, since that's not a safe assumption anymore.

wayne-arnold-adsk:
the embedded Python 3.7.7 in Maya is from python.org  I had thought that the internal library dependencies had been changed to make the embedded relocatable.  stills its not.

a workaround would be to install Python 3.7.7 from python.org to satisfy the dependencies.   just a workaround.   the issue has been entered to look at what can be done on the Maya side.  

Maya itself doesn't show the issue as use of DYLD_LIBRARY_PATH overrides where the libraries are located

PR #1828, opened 2021-11-11
Use mayapy as Python interpreter on Mac

dgovil:
This PR adds a workaround to compilation issues on Macs where the Python binary included with Maya points to a nonexistent path. See https://github.com/Autodesk/maya-usd/issues/1351 

I'd forgotten to clean up and update our internal workaround, so here it is.
This changes it so the CMake script will try and find Maya earlier in the generation, which then allows the `mayapy` path to be used as the Python interpreter directly.

This also seems to work fine with our Maya 2020 builds as well, so I've only left this behind a platform check and not a version check.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,9 @@ endif()
 #------------------------------------------------------------------------------
 # modules and definitions
 #------------------------------------------------------------------------------
+include(cmake/utils.cmake)
+find_package(Maya 2018 REQUIRED)
+
 if (BUILD_MAYAUSD_LIBRARY)
     include(cmake/mayausd_version.info)
     set(MAYAUSD_VERSION "${MAYAUSD_MAJOR_VERSION}.${MAYAUSD_MINOR_VERSION}.${MAYAUSD_PATCH_LEVEL}")
@@ -96,11 +99,8 @@ message(STATUS "   PYTHON_INCLUDE_DIRS = ${PYTHON_INCLUDE_DIRS}")
 message(STATUS "   PYTHON_LIBRARIES    = ${PYTHON_LIBRARIES}")
 message(STATUS "   Python_EXECUTABLE   = ${Python_EXECUTABLE}")
 
-include(cmake/utils.cmake)
-
 include(cmake/jinja.cmake)
 
-find_package(Maya 2018 REQUIRED)
 if (CMAKE_WANT_MATERIALX_BUILD)
     # Requires USD_SHA1 > 891eef161aeed582b9831584ab7f31ebe4bc52e8 for hdMtlx module
     find_package(USD 0.21.08 REQUIRED)
--- a/cmake/python.cmake
+++ b/cmake/python.cmake
@@ -55,6 +55,14 @@ if(PYTHONLIBS_FOUND AND PYTHON_MODULE_EXTENSION)
     return()
 endif()
 
+# On Mac, with Maya 2022+, the Python binaries link against a non-existent path
+# maya2022/Maya.app/Contents/Frameworks/Python.framework/Versions/Current/bin/python -> /Library/Frameworks/Python.framework/Versions/3.7/Python
+# However, just using the mayapy executable is good enough for the build system
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    MESSAGE(STATUS "Setting Python_EXECUTABLE to MayaPy ${MAYA_PY_EXECUTABLE}")
+    set(Python_EXECUTABLE ${MAYA_PY_EXECUTABLE})
+endif()
+
 # Use the Python module to find the python lib.
 if(BUILD_WITH_PYTHON_3)
     find_package(Python ${BUILD_WITH_PYTHON_3_VERSION} EXACT REQUIRED COMPONENTS Interpreter)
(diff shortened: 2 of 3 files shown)


dgovil:
Just made a minor update where mayapy wasn't finding the right LIBDIR on Mac.
It seems like this can happen on other platforms too, so I just have CMake check if the path exists, and otherwise I just go relative to the include directory, which is a pretty safe bet

CMakeLists.txt
```diff
@@ -70,6 +70,9 @@ endif()
 #------------------------------------------------------------------------------
 # modules and definitions
 #------------------------------------------------------------------------------
+include(cmake/utils.cmake)
```
seando-adsk, line 73: I'm not sure that this utils.cmake can be include here (before the python stuff from lines 81-100). Because in utils.cmake it uses the var "Python_EXECUTABLE" at [line 70](https://github.com/Autodesk/maya-usd/blob/dev/cmake/utils.cmake#L70). I'm not enough of a cmake expert to know if that var is replaced during the include of utils.cmake or only when the function _mayaUsd_find_python_module_ is called? If the latter then we should be okay.

cmake/python.cmake
```diff
@@ -55,8 +55,17 @@ if(PYTHONLIBS_FOUND AND PYTHON_MODULE_EXTENSION)
...
+# However, just using the mayapy executable is good enough for the build system
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+
+    MESSAGE(STATUS "Setting Python_EXECUTABLE to MayaPy ${MAYA_PY_EXECUTABLE}")
+    set(Python_EXECUTABLE ${MAYA_PY_EXECUTABLE})
+    set(Python_Interpreter_FOUND TRUE)
+
 # Use the Python module to find the python lib.
-if(BUILD_WITH_PYTHON_3)
+elseif(BUILD_WITH_PYTHON_3)
```
seando-adsk, line 68: I understand what you are doing here, but I'm a little wary about the _elseif_ here which means that it now skips the find_package(Python). Above you are only setting the Python_EXECUTABLE var. Does the find_package(Python) set other stuff that will be skipped now?

cmake/python.cmake
```diff
@@ -121,11 +130,18 @@ message(STATUS "PYTHON_MODULE_EXTENSION: ${PYTHON_MODULE_EXTENSION}")
...
+# so in the case that libdir is not able to be found, use a relative path from the include_dir
+IF(NOT EXISTS ${PYTHON_LIBDIR})
+    get_filename_component(_PYTHON_LIBDIR ${PYTHON_INCLUDE_DIR} PATH)
+    get_filename_component(_PYTHON_LIBDIR ${_PYTHON_LIBDIR} PATH)
+    set(PYTHON_LIBDIR "${_PYTHON_LIBDIR}/lib")
+endif()
+message(STATUS "PYTHON_LIBDIR: ${PYTHON_LIBDIR}")
 
-# Make sure the Python has the same pointer-size as the chosen compiler
+    # Make sure the Python has the same pointer-size as the chosen compiler
```
seando-adsk, line 144: Did you accidentally change the spacing on this line?

CMakeLists.txt
dgovil: It'll only do the lookup when the function is called.  So as long as the Python executable is defined before Line 102 (`include(cmake/jinja.cmake)`) it should be fine.

cmake/python.cmake
dgovil: From my searching through the current CMake files, I couldn't find anything else that gets used from the standard FindPython.cmake 

However I do see that if I only set the Python_EXECUTABLE and not make it supersede the find_package calls, it'll still work. So I've updated the PR with that, so it's less invasive, in case anyone else adds stuff in the future that find_package depends on

cmake/python.cmake
dgovil: Yes, sorry, removed it in the latest update that I pushed just now

CMakeLists.txt
seando-adsk: I wonder if we should add a message(FATAL_ERROR, ...) to confirm that Python_EXECUTABLE is valid in that function?

cmake/python.cmake
seando-adsk: Great thanks.

seando-adsk (approved):

seando-adsk (approved):
Looks good, thanks for adding the python exe check.

kxl-adsk:
@seando-adsk @dgovil I'm seeing a test failure with the following log details:
```
07:03:41     qt.qpa.clipboard: QXcbClipboard::setMimeData: Cannot set X11 selection owner
07:03:41     Fqt.qpa.clipboard: QXcbClipboard::setMimeData: Cannot set X11 selection owner
07:03:41     qt.qpa.clipboard: QXcbClipboard::setMimeData: Cannot set X11 selection owner
07:03:41     .
07:03:41     ======================================================================
07:03:41     FAIL: testInstancedIsolateSelect (__main__.testVP2RenderDelegateIsolateSelect)
07:03:41     ----------------------------------------------------------------------
```
It only happens with this change - can you comment on it?

kxl-adsk:
Actually...scratch that...I'm now seeing it in another PF.

dgovil:
Hmm odd. I don't think I know what coulf cause those off the top of my head, but let me see what I can replicate

PR #1831, opened 2021-11-11
MAYA-114707 test passing array to Maya commands

pierrebai-adsk:
Test passing array of Maya objects, UFE objects or a mix of both to Python `select` command and Mel `select` command.

Verified that the test fail in Maya 2022 and work with the new fix in Maya 2023.

### DIFF
--- a/test/lib/ufe/testSelectionByArray.py
+++ b/test/lib/ufe/testSelectionByArray.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 #
-# Copyright 2020 Autodesk
+# Copyright 2021 Autodesk
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -86,7 +86,6 @@ class SelectByArrayTestCase(unittest.TestCase):
         # Clear selection to start off
         cmds.select(clear=True)
 
-    @unittest.skipUnless(mayaUtils.mayaMajorVersion() >= 2023, 'testSelectByArray only works with fixes available in Maya 2023.')
     def testSelectMayaPathInMel(self):
         """
         Select multiple Maya items by passing them in an array to a mel command.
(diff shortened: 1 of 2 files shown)


test/lib/ufe/testSelectionByArray.py
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env python
+
+#
+# Copyright 2020 Autodesk
```
ppt-adsk, line 4: Nit-pick: copyright year.

test/lib/ufe/testSelectionByArray.py
```diff
@@ -0,0 +1,167 @@
...
+            return ufe.Path([shapeSegment, usdUtils.createUfePathSegment(name)])
+
+        ufeNames = ["/cubeXform", "/cylinderXform", "/sphereXform"]
+        self.usdPaths = [ufe.PathString.string(makeUsdPath(name)) for name in ufeNames]
+        self.mayaPaths = ["pCube1", "pCylinder1", "pSphere1"]
+
+        # Clear selection to start off
+        cmds.select(clear=True)
+
+    @unittest.skipUnless(mayaUtils.mayaMajorVersion() >= 2023, 'testSelectByArray only works with fixes available in Maya 2023.')
```
ppt-adsk, line 89: Copy-paste damage in skipUnless message (here and elsewhere).

test/lib/ufe/testSelectionByArray.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def melSelect(paths):
+    """
+    Call the mel select command using a single array as argument.
+    """
+    mel.eval(melSelectCmd(paths))
+
+    
+class SelectByArrayTestCase(unittest.TestCase):
```
ppt-adsk, line 50: You can actually skip whole classes --- would be appropriate here.  See
https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures
Either that, or make the test conditional in the CMakeLists.txt file.

ppt-adsk (commented):
Thanks for the test!  Minor tweaks you can look at.

test/lib/ufe/testSelectionByArray.py
pierrebai-adsk: Actually, the test for pure Maya objects can run in older versions, so I'll allow it.

ppt-adsk (dismissed):

ppt-adsk (approved):

ppt-adsk (approved):

PR #1833, opened 2021-11-12
Fix support for normal map channel connections

dgovil:
This PR fixes a regression our TD (Stella) had noticed from 73fe1f85 that caused our normal map connections to not export properly.
This patch resolves the issues for us by allowing normal maps to be connected as before, and includes tests to verify it.


test/lib/usd/translators/UsdExportUsdPreviewSurfaceTest/UsdExportUsdPreviewSurfaceTest.ma
```diff
@@ -0,0 +1,545 @@
+//Maya ASCII 2020 scene
```
JGamache-autodesk, line 1: Do you still have a copy of Maya 2019 you can use to generate the scene. We need to support that one as well until the next release is shipped.

This will most likely fail the preflight as we have a testsuite running on Maya 2019 to catch these issues.

JGamache-autodesk (dismissed):
LGTM. Thanks for the fix!

test/lib/usd/translators/UsdExportUsdPreviewSurfaceTest/UsdExportUsdPreviewSurfaceTest.ma
dgovil: Unfortunately we don't have anything below 2020 available at this time. Should I just modify the CMake test file  to only add the test if it's 2020 or higher?

I forget if there were any significant `.ma` changes from 2019 to 2020, otherwise I'd just do surgery on the file.

JGamache-autodesk: Yes, this is why test\lib\usd\translators\testUsdExportImportRoundtripPreviewSurface.py builds the scene using scripting to be compatible with Maya 2019. This is a bit involved, but feel free to copy the code if it makes sense.

You can also try this replicated scene from Maya 2019. Hopefully I got it sufficiently right that it would only require minor surgery to work with your unit test.
[UsdExportUsdPreviewSurfaceTest_2019.zip](https://github.com/Autodesk/maya-usd/files/7528994/UsdExportUsdPreviewSurfaceTest_2019.zip)

dgovil: Thanks for the 2019 file, I'll ask our TD to look into updating the tests to be compatible.

dgovil:
Also I see a build failure for Windows 2022, but I don't see anything in the logs saying what is failing. Any ideas?

kxl-adsk:
@dgovil I don't expect it to be related to this PR either, please ignore.

test/lib/usd/translators/UsdExportUsdPreviewSurfaceTest/UsdExportUsdPreviewSurfaceTest.ma
dgovil: We've updated the test to procedurally generate the files. We've tested in 2020 and 2022, but haven't verified in 2019

JGamache-autodesk (approved):

PR #1839, opened 2021-11-17
New duplicate wording, and unit test.

ppt-adsk:


pierrebai-adsk (approved):
I refactored this a bit in my branch to make it a command, but I'll deal with the conflict, no problem.

PR #1840, opened 2021-11-17
MAYA-113723 improve selection performance

williamkrick:
If there are no reprs selected then don't call execute! It causes extra dirty bits to get held on some rprims, which causes an actual update later when some reprs ARE selected, and that hurts performance. I spent a couple of days tracing a debugging this one and I still do not 100% understand how the dirty flags come back, but I know that doing an execute with no reprs is not useful so I can safely skip it.

The update with no reprs typically happens during the selection pass when we are not changing into or out of point snapping mode, which is pretty much every selection pass we do.


JGamache-autodesk (dismissed):

williamkrick:
Sorry this has some extra commits that I didn't intend to include with it. I'll close the PR and open a new one.

PR #1841, opened 2021-11-17
Allow prim deletion. Using namespace rules.

renderRa:
Allow prim deletion. Following namespace rules in the rename command. Allowing prims to be deleted when there is no stronger layer opinion.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
@@ -27,7 +27,7 @@
 #endif
 
 namespace {
-
+#ifdef MAYA_APP_VERSION >= 2020
 bool hasLayersMuted(const PXR_NS::UsdPrim& prim)
 {
     const PXR_NS::PcpPrimIndex& primIndex = prim.GetPrimIndex();
@@ -46,7 +46,7 @@ bool hasLayersMuted(const PXR_NS::UsdPrim& prim)
     }
     return false;
 }
-
+#endif
 } // anonymous namespace
 
 namespace MAYAUSD_NS_DEF {
@@ -77,7 +77,7 @@ void UsdUndoDeleteCommand::execute()
     auto        targetPrimSpec = stage->GetEditTarget().GetPrimSpecForScenePath(_prim.GetPath());
 
     if (hasLayersMuted(_prim)) {
-        TF_WARN("Cannot remove prim because there is muted layers.");
+        TF_WARN("Cannot remove prim because there are muted layers.");
         return;
     }
 
--- a/lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
@@ -18,8 +18,37 @@
 #include "private/UfeNotifGuard.h"
 
 #ifdef UFE_V2_FEATURES_AVAILABLE
+#include <mayaUsd/ufe/Utils.h>
 #include <mayaUsd/undo/UsdUndoBlock.h>
+
+#include <pxr/usd/pcp/layerStack.h>
+#include <pxr/usd/sdf/layer.h>
+#include <pxr/usd/usd/editContext.h>
+#include <pxr/usd/usd/primCompositionQuery.h>
 #endif
+
+namespace {
+
+bool hasLayersMuted(const PXR_NS::UsdPrim& prim)
+{
+    const PXR_NS::PcpPrimIndex& primIndex = prim.GetPrimIndex();
+
+    for (PcpNodeRef node : primIndex.GetNodeRange()) {
+
+        TF_AXIOM(node);
+
+        const PcpLayerStackSite&   site = node.GetSite();
+        const PcpLayerStackRefPtr& layerStack = site.layerStack;
+
+        const std::set<std::string>& mutedLayers = layerStack->GetMutedLayers();
+        if (mutedLayers.size() > 0) {
+            return true;
+        }
+    }
+    return false;
+}
+
+} // anonymous namespace
 
 namespace MAYAUSD_NS_DEF {
 namespace ufe {
@@ -27,6 +56,7 @@ namespace ufe {
 UsdUndoDeleteCommand::UsdUndoDeleteCommand(const PXR_NS::UsdPrim& prim)
     : Ufe::UndoableCommand()
     , _prim(prim)
+    , _stage(prim.GetStage())
 {
 }
 
@@ -37,13 +67,93 @@ UsdUndoDeleteCommand::Ptr UsdUndoDeleteCommand::create(const PXR_NS::UsdPrim& pr
     return std::make_shared<UsdUndoDeleteCommand>(prim);
 }
 
+bool isPrimDeletionAllowed(const PXR_NS::UsdPrim& prim, SdfPrimSpecHandle primSpec)
+{
+
+    auto primStack = prim.GetPrimStack();
+
+    std::string layerDisplayName;
+
+    // iterate over the prim stack, starting at the highest-priority layer.
+    for (const auto& spec : primStack) {
+        const auto& layerName = spec->GetLayer()->GetDisplayName();
+
+        // skip if there is no primSpec for the selected prim in the current stage's local layer.
+        if (!primSpec) {
+            // add "," separator for multiple layers
+            if (!layerDisplayName.empty()) {
+                layerDisplayName.append(",");
+            }
+            layerDisplayName.append("[" + layerName + "]");
+            continue;
+        }
+
+        // one reason for skipping the reference is to not clash
+        // with the over that may be created in the stage's sessionLayer.
+        // another reason is that one should be able to edit a referenced prim that
+        // either as over/def as long as it has a primSpec in the selected edit target layer.
+        if (spec->HasReferences()) {
+            break;
+        }
+
+        // if exists a def/over specs
+        if (spec->GetSpecifier() == SdfSpecifierDef || spec->GetSpecifier() == SdfSpecifierOver) {
+            // if spec exists in another layer ( e.g sessionLayer or layer other than stage's local
+            // layers ).
+            if (primSpec->GetLayer() != spec->GetLayer()) {
+                layerDisplayName.append("[" + layerName + "]");
+                TF_WARN(
+                    "Cannot delete [%s] because it is defined on a stronger layer %s.",
+                    prim.GetName().GetString().c_str(),
+                    layerDisplayName.c_str());
+                return false;
+            }
+            continue;
+        }
+    }
+    // Post a clear message to indicate that deleting a prim inside a variantset is not allowed.
+    // This restriction was already caught in the above loop.
+    UsdPrimCompositionQuery query(prim);
+    for (const auto& compQueryArc : query.GetCompositionArcs()) {
+        if (!primSpec && PcpArcTypeVariant == compQueryArc.GetArcType()) {
+            TF_WARN(
+                "Cannot delete [%s] because it is defined inside the variant composition arc %s.",
+                prim.GetName().GetString().c_str(),
+                layerDisplayName.c_str());
+            return false;
+        }
+    }
+
+    if (!layerDisplayName.empty()) {
+        TF_WARN(
+            "Cannot delete [%s]. Please set %s as the target layer to proceed.",
+            prim.GetName().GetString().c_str(),
+            layerDisplayName.c_str());
+        return false;
+    }
+    return true;
+}
+
 #ifdef UFE_V2_FEATURES_AVAILABLE
 void UsdUndoDeleteCommand::execute()
 {
     MayaUsd::ufe::InAddOrDeleteOperation ad;
 
     UsdUndoBlock undoBlock(&_undoableItem);
-    _prim.SetActive(false);
+
+    auto targetPrimSpec = _stage->GetEditTarget().GetPrimSpecForScenePath(_prim.GetPath());
+
+    if (hasLayersMuted(_prim)) {
+        TF_WARN("Cannot remove prim because there is muted layers.");
+        return;
+    }
+
+    if (isPrimDeletionAllowed(_prim, targetPrimSpec)) {
+        auto retVal = _stage->RemovePrim(_prim.GetPath());
+        if (!retVal) {
+            TF_VERIFY(retVal, "Failed to remove '%s'", _prim.GetPath().GetText());
+        }
+    }
 }
 
 void UsdUndoDeleteCommand::undo()
--- a/lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
@@ -105,7 +105,7 @@ void UsdUndoDeleteCommand::redo()
 void UsdUndoDeleteCommand::perform(bool state)
 {
     MayaUsd::ufe::InAddOrDeleteOperation ad;
- 
+
     const auto& stage = _prim.GetStage();
     auto        targetPrimSpec = stage->GetEditTarget().GetPrimSpecForScenePath(_prim.GetPath());
 
--- a/lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
@@ -27,7 +27,7 @@
 #endif
 
 namespace {
-#ifdef MAYA_APP_VERSION >= 2020
+#ifdef UFE_V2_FEATURES_AVAILABLE
 bool hasLayersMuted(const PXR_NS::UsdPrim& prim)
 {
     const PXR_NS::PcpPrimIndex& primIndex = prim.GetPrimIndex();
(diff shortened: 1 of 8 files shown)


lib/mayaUsd/ufe/UsdUndoDeleteCommand.h
```diff
@@ -53,7 +53,8 @@ class MAYAUSD_CORE_PUBLIC UsdUndoDeleteCommand : public Ufe::UndoableCommand
     void redo() override;
 
 private:
-    PXR_NS::UsdPrim _prim;
+    PXR_NS::UsdPrim         _prim;
+    PXR_NS::UsdStageWeakPtr _stage;
```
kxl-adsk, line 57: You need this stage in the `executeCommand`. There is no reason that I can find to store it as a member of the class. Simply instantiate it from prim (prim.GetStage()) prior to any delete operation.

lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
```diff
@@ -37,13 +67,93 @@ UsdUndoDeleteCommand::Ptr UsdUndoDeleteCommand::create(const PXR_NS::UsdPrim& pr
     return std::make_shared<UsdUndoDeleteCommand>(prim);
 }
 
+bool isPrimDeletionAllowed(const PXR_NS::UsdPrim& prim, SdfPrimSpecHandle primSpec)
+{
+
+    auto primStack = prim.GetPrimStack();
```
kxl-adsk, line 73: What was the reason to copy this code instead of refactoring [applyCommandRestriction](https://github.com/Autodesk/maya-usd/blob/600c31abcd68ef3fe0047b28fb5f33f6edd0872f/lib/mayaUsd/ufe/private/Utils.cpp#L122) to take the action name as an argument? If we have no good reasons, consider code sharing.

kxl-adsk (changes requested):
Left some comments. Additionally, we will need a unit test to validate this new behavior. Test cases when delete is possible and when it should error out. You have a good start for it in https://github.com/Autodesk/maya-usd/blob/dev/test/lib/ufe/testDeleteCmd.py (and I'm surprised that this test is not failing after your change now).

lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
renderRa: The delete command exception did not emit any warnings so there was no feedback when it failed. The path for exceptions was different and i thought it risky to change any behavior there. I can double check it. Point noted on the stage and Ill take care of it.

renderRa: using applyCommandRestriction() from private/Utils will require adding an exception handler catch in TdeleteCmd.cpp
while it works at first glance, im not sure of the ramification to the existing functionality.

<------------add--------------->
    catch (const std::exception& e) {
        wait.stop();
        TDGcommand::setCurrent(nullptr);

        Tstring fmt(rDeleteNonMayaObjectFailed), errMsg;
        Tstring what(e.what());
        errMsg.tFormat(&fmt, &path, &what);
        theCommandEngine->displayError(errMsg);
        return sFailure;
    }
<------------end add--------------->
    catch(...){
		wait.stop();
		TDGcommand::setCurrent( nullptr );
		return sFailure;
	}

kxl-adsk:
In that case, let's close this PR. Once the foundation is stable, we will look into removing some of the UFE v3 restrictions.

In the meantime, I would like to make sure we understand any changes that you had to do to MayaReference translator. For example, PR #1620 is proposing a change to how reference nodes are discovered. Did you have any workflow-specific requirements here?

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +321,240 @@ def testDeleteArgs(self):
...
+        usdPathSegment = usdUtils.createUfePathSegment('/apple/payload')
+        geoPath = ufe.Path([mayaPathSegment, usdPathSegment])
+        geoItem = ufe.Hierarchy.createItem(geoPath)
+
+        globalSn.append(geoItem)
+        self.assertEqual(len(ufe.GlobalSelection.get()), 1)
+
+        cmds.delete()
+        #deleted
+        self.assertEqual(ufeObs.nbDeleteNotif(), 1)
```
kxl-adsk, line 436: I'm not sure if you already have a test for it...but would be good to test undo and redo for USD objects deletion.

lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
```diff
@@ -43,7 +72,21 @@ void UsdUndoDeleteCommand::execute()
     MayaUsd::ufe::InAddOrDeleteOperation ad;
 
     UsdUndoBlock undoBlock(&_undoableItem);
-    _prim.SetActive(false);
+
+    const auto& stage = _prim.GetStage();
+    auto        targetPrimSpec = stage->GetEditTarget().GetPrimSpecForScenePath(_prim.GetPath());
+
+    if (hasLayersMuted(_prim)) {
+        TF_WARN("Cannot remove prim because there is muted layers.");
```
kxl-adsk, line 80: Typo: there **are** muted layers.

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -193,6 +194,18 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
     }
 }
 
+bool applyCommandRestrictionNoThrow(const UsdPrim& prim, const std::string& commandName)
```
kxl-adsk, line 197: The applyCommandRestriction method is rather surprising. It uses an exception mechanism to apply restrictions. It's not a pattern we would like to repeat but since this code was here before, let's consider it technical debt. Next time  we are here, let's swap the order and make the throwing version a wrapper on top of a method that explicitly returns a state.

kxl-adsk (approved):

kxl-adsk (approved):

kxl-adsk (approved):

test/lib/ufe/testDeleteCmd.py
```diff
@@ -203,6 +206,113 @@ def testDelete(self):
...
+
+        self.assertNotIn(sphereShapeName, sphereChildrenNames)
+        self.assertIn(ball35Name, propsChildrenNames)
+
+        # undo to restore state to original.
+        cmds.undo()
+
+        # After the undo we again added the sphere item back.
+        self.assertEqual(ufeObs.nbDeleteNotif(), 2)
+        self.assertEqual(ufeObs.nbAddNotif(), 2)
```
ppt-adsk, line 314: This test feels awkward, because in essence it's testing deleting a Maya object, which Maya already tests.  If the point of this test is to demonstrate we can't delete a referenced object, then we can test that, and forget about undo and redo.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -259,11 +369,12 @@ def testDeleteArgs(self):
...
 
         ufeObs.reset()
+        
         cmds.delete(
             ball35PathString, ball34PathString, "|pSphere1|pSphereShape1")
-
-        # We deleted 3 items.
-        self.assertEqual(ufeObs.nbDeleteNotif(), 3)
+        
+        # We deleted 1 items.
```
ppt-adsk, line 376: Again, here I believe that we're only testing the delete of a Maya object, which makes me uncomfortable.  We really have to test that USD objects are deleted.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +421,240 @@ def testDeleteArgs(self):
...
+        # delete the proxy shape node itself. 
+        cmds.delete('|transform1|proxyShape1')
+
+        self.assertEqual(ufeObs.nbDeleteNotif(), 1)
+        
+        #restore
+        cmds.undo()
+
+        ufeObs.reset()
+        self.assertEqual(len(cmds.ls('proxyShape1')), 1)
```
ppt-adsk, line 457: I would remove this test, because it's essentially testing that you can delete a Maya node, which Maya is already supposed to test.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +421,240 @@ def testDeleteArgs(self):
...
+
+        #cannot be deleted as it has a variant arc
+        cmds.delete(ball35PathStr)
```
ppt-adsk, line 460: This part is valuable, and should be kept.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +421,240 @@ def testDeleteArgs(self):
...
+        self.assertEqual(ufeObs.nbDeleteNotif(), 0)
+
+    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() >= 2, 'testDeleteProxyShape only available in UFE v2 or greater.')
```
ppt-adsk, line 463: Copy-paste damage on the skip message.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +421,240 @@ def testDeleteArgs(self):
...
+        
+        # check GetLayerStack behavior
+        self.assertEqual(stage.GetLayerStack()[0], stage.GetSessionLayer())
+        self.assertEqual(stage.GetEditTarget().GetLayer(), stage.GetRootLayer())
+
+        cmds.delete()    
+        
+        self.assertEqual(ufeObs.nbDeleteNotif(), 0)
+
+    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() >= 2, 'testDeleteProxyShape only available in UFE v2 or greater.')
```
ppt-adsk, line 493: Copy-paste.

test/lib/ufe/testDeleteCmd.py
ppt-adsk: Here I'd go beyond checking the number of notifications.  I would ask the parent if it still has this child.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +421,240 @@ def testDeleteArgs(self):
...
+        geoItem = ufe.Hierarchy.createItem(geoPath)
+
+        globalSn.append(geoItem)
+        self.assertEqual(len(ufe.GlobalSelection.get()), 1)
+
+        cmds.delete()
+        #deleted
+        self.assertEqual(ufeObs.nbDeleteNotif(), 1)
+
+    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() >= 2, 'testDeleteProxyShape only available in UFE v2 or greater.')
```
ppt-adsk, line 538: Copy-paste.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +421,240 @@ def testDeleteArgs(self):
...
+        # get the USD stage
+        stage = mayaUsd.ufe.getStage(str(mayaPathSegment))
+
+        # by default edit target is set to the Rootlayer.
+        self.assertEqual(stage.GetEditTarget().GetLayer(), stage.GetRootLayer())
+
+        # delete selected item `/TreeBase/trunk` `
+        cmds.delete()
+        # deleted two items
+        self.assertEqual(ufeObs.nbDeleteNotif() , 2)
```
ppt-adsk, line 582: Again I'd check beyond the number of notifications.  Also, this test feels like there's quite a bit of overlap with the last part of the testDeleteRestrictionHasSpecs test, unless I'm missing something --- if I am, might be good to add comments about what the difference is.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +421,240 @@ def testDeleteArgs(self):
...
+
+    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() >= 2, 'testDeleteProxyShape only available in UFE v2 or greater.')
```
ppt-adsk, line 584: Copy-paste.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +421,240 @@ def testDeleteArgs(self):
...
+        smallPotatoItem = ufe.Hierarchy.createItem(smallPotatoPath)
+        smallPotatoPrim = mayaUsd.ufe.ufePathToPrim(ufe.PathString.string(smallPotatoPath))
+
+        # add Geom to selection list
+        ufe.GlobalSelection.get().append(geomItem)
+
+        # get prim spec for Geom prim
+        primspecGeom = stage.GetEditTarget().GetPrimSpecForScenePath(geomPrim.GetPath());
+
+        # primSpec is expected to be None
```
ppt-adsk, line 623: Why is it expected to be None?

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +421,240 @@ def testDeleteArgs(self):
...
+
+        # get prim spec for Small_Potato prim
+        primspecSmallPotato = stage.GetEditTarget().GetPrimSpecForScenePath(smallPotatoPrim.GetPath())
+
+        # primSpec is expected to be None
+        self.assertIsNone(primspecSmallPotato)
+
+        # delete "/objects/Geom/Small_Potato"
+        # expect no delete
+        cmds.delete()
```
ppt-adsk, line 644: What's the difference between this test and the one at line 626?

ppt-adsk (changes requested):
Lots of great work in there, I appreciate all the effort you put into the tests!  A couple of overall points:

1. I'd make sure we really want delete as remove prim only in Maya 2022 and 2023, which is what you seem to have coded here.  Your change should work fine in Maya 2019 and Maya 2020, we don't need UFE v2 for this.
2. Your test needs more USD object undo / redo behavior testing, and I wouldn't only rely on counting the number of delete notifications to check delete behavior.
Keep up the good work!

test/lib/ufe/testDuplicateCmd.py
```diff
@@ -174,11 +174,11 @@ def childrenNames(children):
 
         cmds.select(clear=True)
 
-        # Delete the even numbered props:
+        # Deactivate the even numbered props:
         evenPropsChildrenPre = propsChildrenPre[0:35:2]
         for propChild in evenPropsChildrenPre:
             ufe.GlobalSelection.get().append(propChild)
-        cmds.delete()
```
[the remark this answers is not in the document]
renderRa: added back in the deactivation to preserve the original intent.

lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
```diff
@@ -16,17 +16,46 @@
...
 #ifdef UFE_V2_FEATURES_AVAILABLE
+#include <mayaUsd/ufe/Utils.h>
 #include <mayaUsd/undo/UsdUndoBlock.h>
+
+#include <pxr/usd/pcp/layerStack.h>
+#include <pxr/usd/sdf/layer.h>
+#endif
+
+namespace {
+#ifdef UFE_V2_FEATURES_AVAILABLE
```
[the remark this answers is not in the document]
renderRa: Resolved. Error in conditional includes

lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
```diff
@@ -43,7 +72,21 @@ void UsdUndoDeleteCommand::execute()
     MayaUsd::ufe::InAddOrDeleteOperation ad;
 
     UsdUndoBlock undoBlock(&_undoableItem);
-    _prim.SetActive(false);
+
+    const auto& stage = _prim.GetStage();
```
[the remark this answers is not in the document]
renderRa: Made available in all versions

test/lib/ufe/testDeleteCmd.py
```diff
@@ -98,7 +100,8 @@ def setUp(self):
         # Clear selection to start off
         cmds.select(clear=True)
 
-    def testDelete(self):
+    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() <= 1, 'only available in UFE v1 or less.')
```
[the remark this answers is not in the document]
renderRa: Made available in all versions

test/lib/ufe/testDeleteCmd.py
```diff
@@ -203,6 +206,113 @@ def testDelete(self):
...
+
+        self.assertIn(sphereShapeItem, sphereChildren)
+        self.assertIn(ball35Item, propsChildren)
+        self.assertIn(sphereShapeName, sphereChildrenNames)
+        self.assertIn(ball35Name, propsChildrenNames)
+
+        ufeObs.reset()
+        cmds.delete()
+
+        # We only deleted the maya object.
```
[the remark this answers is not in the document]
renderRa: Added a comment explaining its in a variant arc.

test/lib/ufe/testDeleteCmd.py
renderRa: good point, removed undo and redo test

test/lib/ufe/testDeleteCmd.py
renderRa: removed

test/lib/ufe/testDeleteCmd.py
renderRa: since it was a duplicate test removed and added comments

test/lib/ufe/testDeleteCmd.py
renderRa: these have been removed

test/lib/ufe/testDeleteCmd.py
renderRa: Its checking that the object under geom couldnt be removed as well

kxl-adsk:
Closing in favor of https://github.com/Autodesk/maya-usd/pull/1895.

PR #1842, opened 2021-11-17
MAYA-114297 MAYA-114300 Job context UI

JGamache-autodesk:
Add a job context dropdown called "Plug-in Configuration:" in both
import and export dialogs when in USD mode.

This dropdown will be populated with the available job contexts.

Selecting on the the job context will populate the UI with the settings
provided by the job context, and these settings will be made read-only.

### DIFF
--- a/lib/mayaUsd/commands/baseExportCommand.cpp
+++ b/lib/mayaUsd/commands/baseExportCommand.cpp
@@ -40,7 +40,7 @@ MSyntax MayaUSDExportCommand::createSyntax()
     MSyntax syntax;
 
     // These flags correspond to entries in
-    // UsdMayaJobExportArgs::GetDefaultDictionary.
+    // UsdMayaJobExportArgs::GetGuideDictionary.
     syntax.addFlag(
         kMergeTransformAndShapeFlag,
         UsdMayaJobExportArgsTokens->mergeTransformAndShape.GetText(),
@@ -245,7 +245,7 @@ MStatus MayaUSDExportCommand::doIt(const MArgList& args)
 
         // Read all of the dictionary args first.
         const VtDictionary userArgs = UsdMayaUtil::GetDictionaryFromArgDatabase(
-            argData, UsdMayaJobExportArgs::GetDefaultDictionary());
+            argData, UsdMayaJobExportArgs::GetGuideDictionary());
 
         // Now read all of the other args that are specific to this command.
         bool        append = false;
--- a/lib/mayaUsd/fileio/jobs/jobArgs.cpp
+++ b/lib/mayaUsd/fileio/jobs/jobArgs.cpp
@@ -785,6 +785,69 @@ const VtDictionary& UsdMayaJobExportArgs::GetDefaultDictionary()
     return d;
 }
 
+const VtDictionary& UsdMayaJobExportArgs::GetGuideDictionary()
+{
+    static VtDictionary   d;
+    static std::once_flag once;
+
+    std::call_once(once, []() {
+        // Common types:
+        const auto _boolean = VtValue(false);
+        const auto _string = VtValue(std::string());
+        const auto _stringVector = VtValue(std::vector<VtValue>({ _string }));
+        const auto _stringTriplet = VtValue(std::vector<VtValue>({ _string, _string, _string }));
+        const auto _stringTripletVector = VtValue(std::vector<VtValue>({ _stringTriplet }));
+
+        // Provide guide types for the parser:
+        d[UsdMayaJobExportArgsTokens->chaser] = _stringVector;
+        d[UsdMayaJobExportArgsTokens->chaserArgs] = _stringTripletVector;
+        d[UsdMayaJobExportArgsTokens->compatibility] = _string;
+        d[UsdMayaJobExportArgsTokens->defaultCameras] = _boolean;
+        d[UsdMayaJobExportArgsTokens->defaultMeshScheme] = _string;
+        d[UsdMayaJobExportArgsTokens->defaultUSDFormat] = _string;
+        d[UsdMayaJobExportArgsTokens->eulerFilter] = _boolean;
+        d[UsdMayaJobExportArgsTokens->exportCollectionBasedBindings] = _boolean;
+        d[UsdMayaJobExportArgsTokens->exportColorSets] = _boolean;
+        d[UsdMayaJobExportArgsTokens->exportDisplayColor] = _boolean;
+        d[UsdMayaJobExportArgsTokens->exportInstances] = _boolean;
+        d[UsdMayaJobExportArgsTokens->exportMaterialCollections] = _boolean;
+        d[UsdMayaJobExportArgsTokens->exportReferenceObjects] = _boolean;
+        d[UsdMayaJobExportArgsTokens->exportRefsAsInstanceable] = _boolean;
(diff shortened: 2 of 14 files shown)


lib/mayaUsd/commands/baseExportCommand.cpp
```diff
@@ -245,7 +245,7 @@ MStatus MayaUSDExportCommand::doIt(const MArgList& args)
 
         // Read all of the dictionary args first.
         const VtDictionary userArgs = UsdMayaUtil::GetDictionaryFromArgDatabase(
-            argData, UsdMayaJobExportArgs::GetDefaultDictionary());
+            argData, UsdMayaJobExportArgs::GetGuideDictionary());
```
JGamache-autodesk, line 248: This function works best if the dictionary passed contains fully typed values. The "Default" dictionary has a tendency to contain empty vectors for args that take a list, and this does not provide enough information about what goes into the list (especially chaserArgs where the contents are themselves vectors). The newly introduced "Guide" dictionary provides all the information for proper parsing.

lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -785,6 +785,69 @@ const VtDictionary& UsdMayaJobExportArgs::GetDefaultDictionary()
...
+    static VtDictionary   d;
+    static std::once_flag once;
+
+    std::call_once(once, []() {
+        // Common types:
+        const auto _boolean = VtValue(false);
+        const auto _string = VtValue(std::string());
+        const auto _stringVector = VtValue(std::vector<VtValue>({ _string }));
+        const auto _stringTriplet = VtValue(std::vector<VtValue>({ _string, _string, _string }));
+        const auto _stringTripletVector = VtValue(std::vector<VtValue>({ _stringTriplet }));
```
JGamache-autodesk, line 799: The value is not important. Only the type.

lib/mayaUsd/utils/util.cpp
```diff
@@ -1937,6 +1937,10 @@ std::pair<bool, std::string> UsdMayaUtil::ValueToArgument(const VtValue& value)
 {
     if (value.IsHolding<bool>()) {
         return std::make_pair(true, std::string(value.Get<bool>() ? "1" : "0"));
+    } else if (value.IsHolding<int>()) {
+        return std::make_pair(true, std::to_string(value.Get<int>()));
+    } else if (value.IsHolding<float>()) {
+        return std::make_pair(true, std::to_string(value.Get<float>()));
```
JGamache-autodesk, line 1943: We have int and float values in the args: startTime, endTime, frameStride, so we need a way to unparse them.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -15,25 +15,29 @@
 // limitations under the License.
 //
 
-proc mayaUsdTranslatorExport_SetCheckbox(string $arg, string $widget) {
+proc mayaUsdTranslatorExport_SetCheckbox(string $arg, int $enable, string $widget) {
```
JGamache-autodesk, line 18: All the setters take an extra argument to enable the field or not depending if the value was set from a regular argument, or from an argument provided by a job context.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -236,7 +359,7 @@ global proc int mayaUsdTranslatorExport (string $parent,
         string $skelsAnn = "Exports Maya joints as part of a USD skeleton.";
         string $skinClustersAnn = "Exports Maya skin clusters as part of a USD skeleton.";
         string $blendShapesAnn = "Exports Maya Blend Shapes as USD blendshapes. Requires skeletons to be exported as well.";
-        string $displayColorAnn = "If selected, exports the diffuse color of the geometry’s bound shader as a displayColor primvar on the USD mesh.";
+        string $displayColorAnn = "If selected, exports the diffuse color of the geometry's bound shader as a displayColor primvar on the USD mesh.";
```
JGamache-autodesk, line 362: MAYA-114878 - Use ASCII apostrophe instead of UNICODE one.

test/lib/usd/plugin/nullApiExporter.cpp
```diff
@@ -61,7 +61,29 @@ REGISTER_EXPORT_JOB_CONTEXT_FCT(Thierry, "Thierry", "Exports for Thierry rendere
 
 REGISTER_EXPORT_JOB_CONTEXT_FCT(SceneGrinder, "Scene Grinder", "Exports to Scene Grinder")
 {
-    return VtDictionary();
```
JGamache-autodesk, line 64: For testing interactively in Maya. Provide a few interesting values to make sure the fields are correctly updated and disabled.

test/lib/usd/translators/testUsdExportSchemaApi.py
```diff
@@ -59,12 +59,29 @@ def testExportJobContextLister(self):
...
 
+        # The job context used for UI debugging have int startTime and float frameStride values to convert:
+        self.assertEqual(set(cmds.mayaUSDListJobContexts(ig="Scene Grinder").strip(";").split(";")), 
+            set(['importUSDZTextures=1',
+                 'preferredMaterial=standardSurface',
+                 'readAnimData=1',
+                 'startTime=12',
+                 'useCustomFrameRange=1']))
+
+        self.assertEqual(set([i.rstrip("0")
```
JGamache-autodesk, line 76: Strip the trailing zeroes of the frameStride in case the precision differs per platform.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
seando-adsk: Do these strings appear in the UI? If so it would be nice if there were moved to the file mayaUSDRegisterStrings.mel.

plugin/adsk/scripts/mayaUsdTranslatorImport.mel
```diff
@@ -76,10 +170,21 @@ global proc int mayaUsdTranslatorImport (string $parent,
     int $bw = 125;
 
     if ($action == "post") {
+        string $jobContextAnn = "Select a loaded plug-in configuration to modify import options";
```
seando-adsk, line 173: Same comment for this string.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
seando-adsk: I realize you didn't add these strings, but if we can move them now it would be great.

seando-adsk (changes requested):

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -50,19 +54,19 @@ proc mayaUsdTranslatorExport_SetJobContextDropdown(string $arg) {
     }
 
     // None is the default.
-    optionMenuGrp -e -select 0 jobContextPopup;
+    optionMenuGrp -e -select 1 jobContextPopup;
```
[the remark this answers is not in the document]
JGamache-autodesk: Documentation of optionMenuGrp says: "Selects an item by index. The first item is 1."

seando-adsk (approved):
Awesome, thanks for doing the localization work.

PR #1844, opened 2021-11-17
MAYA-114374 Prevent parenting under an instance

JGamache-autodesk:


ppt-adsk (approved):
Thanks for the fix and the test!

PR #1855, opened 2021-11-23
Update AL_USDMaya Selection code to work with UFE Selection

J-Mo63:
This PR implements various changes additionally supporting UFE selection across the AL plugin.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -1108,6 +1108,9 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
         return false;
     }
 
+    if (_proxyShapeData->ProxyShape())
+        ((MayaUsdProxyShapeBase*)_proxyShapeData->ProxyShape())->notifyPreSelectionChanged();
+
 #ifdef UFE_V2_FEATURES_AVAILABLE
     TF_UNUSED(listAdjustment);
 
@@ -1137,6 +1140,9 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
 #else
     dagPath = _proxyShapeData->ProxyDagPath();
 #endif
+
+    if (_proxyShapeData->ProxyShape())
+        ((MayaUsdProxyShapeBase*)_proxyShapeData->ProxyShape())->notifyPostSelectionChanged();
 
     return true;
 }
--- a/lib/mayaUsd/ufe/Global.cpp
+++ b/lib/mayaUsd/ufe/Global.cpp
@@ -107,6 +107,12 @@ StagesSubject::Ptr g_StagesSubject;
 bool InPathChange::inGuard = false;
 bool InAddOrDeleteOperation::inGuard = false;
 
+void refreshStages()
+{
+    if (!g_StagesSubject) return;
+    g_StagesSubject->afterOpen();
+}
+
 //------------------------------------------------------------------------------
 // Functions
 //------------------------------------------------------------------------------
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
@@ -40,6 +40,8 @@
 
 #include <mayaUsd/listeners/proxyShapeNotice.h>
 #include <mayaUsd/nodes/stageData.h>
+#include <mayaUsd/ufe/Global.h>
+#include <mayaUsd/ufe/StagesSubject.h>
 #include <mayaUsd/utils/utilFileSystem.h>
 
 #include <pxr/usd/ar/resolver.h>
@@ -1448,6 +1450,8 @@ void ProxyShape::loadStage()
             MGlobal::displayWarning(MString("Failed to open usd file \"") + file + "\"");
         }
     }
+
+    MayaUsd::ufe::refreshStages();
 
     // Get the prim, if the stage is valid.
     // If no primPath string specified, then use the pseudo-root.
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
@@ -1098,11 +1098,9 @@ void SelectionUndoHelper::doIt()
     else {
         m_proxy->m_selectedPaths.clear();
         for (auto& it : m_newUFESelection) {
-            const auto& path = it->path();
-            const auto& pathStr = path.string();
-            const auto  index = pathStr.find_first_of('/');
-            if (index != std::string::npos) {
-                m_proxy->m_selectedPaths.insert(SdfPath(pathStr.c_str() + index));
+            auto segments = it->path().getSegments();
+            if (segments.size() > 1) {
+                m_proxy->m_selectedPaths.insert(SdfPath(segments[1].string()));
             } else {
                 // Presumably the root node has been selected, but it doesn't appear I
                 // can handle that edge case in UFE?
@@ -1170,11 +1168,9 @@ void SelectionUndoHelper::undoIt()
 
         m_proxy->m_selectedPaths.clear();
         for (auto& it : m_previousUFESelection) {
-            auto&       path = it->path();
-            std::string pathStr = path.string();
-            auto        index = pathStr.find_first_of('/');
-            if (index != std::string::npos) {
-                m_proxy->m_selectedPaths.insert(SdfPath(pathStr.c_str() + index));
+            auto segments = it->path().getSegments();
+            if (segments.size() > 1) {
+                m_proxy->m_selectedPaths.insert(SdfPath(segments[1].string()));
             } else {
                 // UHM. Not sure what to do here?
             }
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
@@ -19,12 +19,24 @@
 #include "AL/usdmaya/nodes/Scope.h"
 #include "AL/usdmaya/nodes/Transform.h"
 #include "AL/usdmaya/nodes/TransformationMatrix.h"
+#include "mayaUsd/nodes/proxyShapeBase.h"
+#include "mayaUsd/nodes/proxyShapePlugin.h"
+#include "mayaUsd/ufe/Global.h"
 
 #include <maya/MFnDagNode.h>
 #include <maya/MProfiler.h>
 #include <maya/MPxCommand.h>
 
 #include <cinttypes>
+
+#if defined(WANT_UFE_BUILD)
+#include <ufe/globalSelection.h>
+#include <ufe/observableSelection.h>
+#include <ufe/path.h>
+#include <ufe/runTimeMgr.h>
+#include <ufe/sceneItem.h>
+#include <ufe/selection.h>
+#endif
 
 namespace AL {
 namespace usdmaya {
@@ -869,7 +881,6 @@ void ProxyShape::removeUsdTransformChain_internal(
             }
             m_lockManager.setInherited(primPath);
         }
-
         parentPrim = parentPrim.GetParent();
     }
 }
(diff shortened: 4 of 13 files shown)


plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/commands/test_ProxyShapeSelect.cpp
```diff
@@ -27,8 +29,45 @@
...
+#include <ufe/globalSelection.h>
+#include <ufe/observableSelection.h>
+#include <ufe/path.h>
+#endif
+
 using AL::maya::test::buildTempPath;
 
+namespace {
+
+bool isAlive(AL::usdmaya::nodes::ProxyShape* proxy, SdfPath path)
```
ppt-adsk, line 42: Not sure I understand the difference between isAlive() and isSelected() --- both seem to look in the UFE selection for an SdfPath?

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/commands/test_ProxyShapeSelect.cpp
```diff
@@ -54,25 +93,43 @@ TEST(ProxyShapeSelect, selectNode1)
...
+                    if (pathName == paths[j].GetText()) {
+                        found = true;
+                        break;
+                    }
                 }
+                EXPECT_TRUE(found);
+            }
+        } else {
+            auto selection = Ufe::GlobalSelection::get();
+            for (auto it : *selection) {
```
ppt-adsk, line 120: I realize it's just a test, but it feels like there should be a way to create a UFE path and use Ufe::Selection searching, which is Trie-based and therefore more efficient.

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/commands/test_ProxyShapeSelect.cpp
```diff
@@ -107,27 +164,33 @@ TEST(ProxyShapeSelect, selectNode1)
     MSelectionList sl;
 
     EXPECT_EQ(1u, results.length());
-    EXPECT_EQ(MString("|transform1|root|hip1|knee1|ankle1|ltoe1"), results[0]);
+    if (!MayaUsdProxyShapePlugin::useVP2_NativeUSD_Rendering()) {
+        EXPECT_EQ(MString("|transform1|root|hip1|knee1|ankle1|ltoe1"), results[0]);
+    } else {
+        EXPECT_EQ(
+            MString("|world|transform1|AL_usdmaya_ProxyShape1/root/hip1/knee1/ankle1/ltoe1"),
```
ppt-adsk, line 171: I never like to see the |world UFE path component in a path literal, as it is arguably a UFE implementation detail --- should use Ufe::PathString::string() (and the reverse, Ufe::PathString::path()) wherever possible, which will remove this component.

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/commands/test_ProxyShapeSelect.cpp
```diff
@@ -288,25 +361,43 @@ TEST(ProxyShapeSelect, selectNode2)
...
                 }
+                EXPECT_TRUE(found);
+            }
+        } else {
+            auto selection = Ufe::GlobalSelection::get();
+            for (auto it : *selection) {
+                bool found = false;
+                for (uint32_t j = 0; j < paths.size(); ++j) {
+                    auto str = it->path().string();
+                    auto index = str.find_first_of('/');
```
ppt-adsk, line 392: Again, feels like you're only interested in the second segment of the path, so should use that directly, instead of converting to string and searching for the first '/'.

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/commands/test_ProxyShapeSelect.cpp
```diff
@@ -1526,204 +1746,221 @@ TEST(ProxyShapeSelect, pseudoRootSelect)
...
-    EXPECT_TRUE(proxy->isRequiredPath(SdfPath("/root/hip1/knee1/ankle1")));
-    EXPECT_TRUE(proxy->isRequiredPath(SdfPath("/root/hip1/knee1/ankle1/ltoe1")));
-    EXPECT_TRUE(proxy->isRequiredPath(SdfPath("/root/hip2")));
-    // Ensure the proxyShape is no longer selected
-    {
-        SCOPED_TRACE("");
-        compareNodes({ SdfPath("/root/hip1/knee1/ankle1/ltoe1"), SdfPath("/root/hip2") });
-    };
-}
+    // RB: Disabling this test for now as this is currently broken in UFE pending some feedback.
```
ppt-adsk, line 1758: Let me know how we can help here.

lib/mayaUsd/ufe/Global.cpp
```diff
@@ -107,6 +107,12 @@ StagesSubject::Ptr g_StagesSubject;
 bool InPathChange::inGuard = false;
 bool InAddOrDeleteOperation::inGuard = false;
 
+void refreshStages()
```
ppt-adsk, line 110: I'm surprised that this would need to be explicitly called.  This clears out USD observation, and causes the stage map to be rebuilt.  When is this needed outside of file open?

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
```diff
@@ -1449,6 +1451,8 @@ void ProxyShape::loadStage()
         }
     }
 
+    MayaUsd::ufe::refreshStages();
```
ppt-adsk, line 1454: The Autodesk plugin also loads stages, and doesn't explicitly force the revoking of USD listeners and rebuilding of the stage map.  What is different in the AL plugin?

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
```diff
@@ -1069,18 +1080,65 @@ void SelectionUndoHelper::doIt()
...
-    m_proxy->m_pleaseIgnoreSelection = false;
-    if (!MGlobal::optionVarIntValue("AL_usdmaya_ignoreLockPrims")) {
-        m_proxy->constructLockPrims();
+#if defined(WANT_UFE_BUILD)
+    else {
+        m_proxy->m_selectedPaths.clear();
+        for (auto& it : m_newUFESelection) {
+            const auto& path = it->path();
+            const auto& pathStr = path.string();
+            const auto  index = pathStr.find_first_of('/');
```
ppt-adsk, line 1103: Would be more effective to only deal with the USD segment, instead of hunting for the '/' character in a string version of the path.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
```diff
@@ -1069,18 +1080,65 @@ void SelectionUndoHelper::doIt()
...
+            if (index != std::string::npos) {
+                m_proxy->m_selectedPaths.insert(SdfPath(pathStr.c_str() + index));
+            } else {
+                // Presumably the root node has been selected, but it doesn't appear I
```
ppt-adsk, line 1107: Typically we use the proxy shape node itself as a proxy for the USD root, e.g. in the UFE hierarchy interface the proxy shape children are in fact the USD root children.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
```diff
@@ -1069,18 +1080,65 @@ void SelectionUndoHelper::doIt()
...
+                // can handle that edge case in UFE?
+            }
+        }
+        Ufe::GlobalSelection::get()->replaceWith(m_newUFESelection);
+        if (m_selectRoot) {
+            MSelectionList sl;
+            MGlobal::getActiveSelectionList(sl);
+            sl.add(m_proxy->thisMObject());
+
+            // UGH. This nukes the UFE selection, and it doesn't appear it's possible to create a
```
ppt-adsk, line 1117: Hopefully using the proxy shape node will be useful here.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
```diff
@@ -1069,18 +1080,65 @@ void SelectionUndoHelper::doIt()
...
+                MObject obj;
+                sl.getDependNode(i, obj);
+                if (obj == proxyObj) {
+                    sl.remove(i);
+                    break;
+                }
+            }
+
+            /// UGH! This clears the UFE selection :(
+            MGlobal::setActiveSelectionList(sl);
```
ppt-adsk, line 1136: The UFE selection should be a superset of the the Maya selection, as far as objects are concerned.  Therefore, following this call, the UFE selection should have all the Maya objects in sl.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
```diff
@@ -1218,271 +1318,400 @@ bool ProxyShape::doSelect(SelectionUndoHelper& helper, const SdfPathVector& orde
...
+                }
+            }
+        } break;
+
+        case MGlobal::kXORWithList: {
+            helper.m_newUFESelection = *selection;
+            for (auto& selectedPath : orderedPaths) {
+                // when selecting the root path, select the ProxyShape node using normal maya
+                // selection
+                if (selectedPath == root) {
```
ppt-adsk, line 1686: This difference between using the Maya selection and the UFE selection to deal with the pseudo-root feels tricky.  Why not just add the proxy shape to the UFE selection?  I'm far from being an expert in the AL plugin, so ignore this suggestion if it's not relevant.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -1108,6 +1108,9 @@ bool ProxyRenderDelegate::getInstancedSelectionPath(
         return false;
     }
 
+    if (_proxyShapeData->ProxyShape())
```
ppt-adsk, line 1111: Is this going to catch only marquee selections in the viewport?  What about selections made in the Outliner, or any other future UFE-enabled node viewer or editor?  How would they receive these pre selection and post selection changed notifications?  Is there no way to reformulate this as standard UFE selection changed notifications?  It feels awkward to go through the proxy shape to distribute the notifications, when the UFE global selection itself is already observable, and is not tied to viewport-based selection.

ppt-adsk (commented):
My lack of familiarity with the AL plugin most certainly means that some of the comments are off-base, but I'm perplexed by the change to the proxyShapeBase code to provide UFE selection observability.  Given that the UFE global selection is already observable, this might suggest that we are missing something to make the UFE global selection observation as useful as possible to its clients.

fabal:
Thanks for the review @ppt-adsk , we'll get back to you shortly.

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/commands/test_ProxyShapeSelect.cpp
J-Mo63: As far as I can tell, the distinction between the two is that `isAlive()` will also look for children of the specified path, i.e.
```cpp
return selection->contains(ufePath) || selection->containsDescendant(ufePath);
```
I have updated the method to use UFE selection instead of the string search anyhow.

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/commands/test_ProxyShapeSelect.cpp
J-Mo63: Unfortunately not for these specific tests since it's checking that all of the UFE selection items have *some* SdfPath, rather than the other way around. I have however gone ahead and split off the search code into a helper function to clarify that.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShapeSelection.cpp
J-Mo63: Is this in response to the comment?

fabal:
The code found in `ProxyShapeSelection.cpp` (and corresponding tests) is not supported anymore internally (at Animal Logic) which means that the changes you see in there are rather old (we still need to merge them though). 
To ease the review for our next PRs, we can highlight the code which doesn't need a review so you can save some time.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
fabal: We're looking into removing (or at least moving somewhere else) these notifications.

J-Mo63:
Thanks for the feedback @ppt-adsk, after some discussions within the team we're potentially looking to retire code related to proxy draw override and proxy shape selection as per discussion topic https://github.com/Autodesk/maya-usd/discussions/1879. I'll close the PR now, given that it's almost entirely about those two. Again, thank you for the suggestions, it's helped move things along quite a bit!

ppt-adsk:
Excellent, glad I could help!

PR #1789, opened 2021-10-28
Tremblp/maya 112824/pull push 01 to dev

ppt-adsk:
This pull request introduces the edit as Maya data / merge back to USD workflow (a.k.a pull and push, respectively).  This workflow is used to convert USD data to Maya data for editing, then back to USD once editing is done.  It leverages existing USD import support for edit as Maya (pull), and USD export support for merge back to USD (push).  In a similar way to the import and export prim readers and writers, the edit as Maya / merge back to USD uses prim updaters, created on a per type basis, to perform the conversions.

Once a USD hierarchy has been edited as Maya, it appears in the Outliner hierarchy as parented under its original USD parent.  However, in the Maya data model a separate, hidden Maya hierarchy is created to contain the pulled objects in their Maya representation.  Because the paths to these pulled objects are pure Maya paths, all existing Maya editors and tools can operate on pulled data, and they will appear in the Maya selection as normal Maya objects.

PR #1856, opened 2021-11-24
Fixed exiting callback failing to deregister on UFE global

J-Mo63:
This PR fixes an issue introduced in PR https://github.com/Autodesk/maya-usd/pull/1789, causing the exit callback in `lib/mayaUsd/ufe/Global.cpp` to be run multiple times as it would not get deregistered correctly, producing the following `TF_VERIFY` failure in `lib/mayaUsd/ufe/UsdUIUfeObserver.cpp` throughout tests:

```console
Coding Error: in destroy at line 60 of /fast/the_lot/maya-usd/lib/mayaUsd/ufe/UsdUIUfeObserver.cpp -- Failed verification: ' ufeObserver '
```

### DIFF
--- a/lib/mayaUsd/ufe/Global.cpp
+++ b/lib/mayaUsd/ufe/Global.cpp
@@ -66,7 +66,7 @@ void exitingCallback(void* /* unusedData */)
 
 int gRegistrationCount = 0;
 
-bool gExitingCbId = 0;
+MCallbackId gExitingCbId = 0;
 } // namespace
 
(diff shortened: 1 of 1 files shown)


kxl-adsk:
Thank you. That’s a good catch, @J-Mo63!

ppt-adsk (approved):
Ouch!  Good catch!  I thought we had compilers for that...  A bool is a kind of integer, I guess...

PR #1866, opened 2021-11-30
Added multi-format serialisation test for shared LayerManager

J-Mo63:
As requested by @kxl-adsk in PR https://github.com/Autodesk/maya-usd/pull/1792#discussion_r754490345, I've gone ahead and added a test for the updated `LayerManager` to test multi-format serialisation, similar to the one added [here](https://github.com/AnimalLogic/maya-usd/blob/f485f4b1a43c36d4d8d584d62c0988f2609fafcc/plugin/al/plugin/AL_USDMayaTestPlugin/py/testLayerManager.py#L143).

PR #1872, opened 2021-12-02
Made all dirty layers and anonymous sublayers serialisable

J-Mo63:
This PR incorporates two changes:

The proxy shape has been updated to make all dirty layers serialisable regardless of whether they have ever been set as an edit target or not when saving Maya scenes. As a result anonymous sublayer content can be serialised.

Loading a scene serialised with anonymous sublayers results in stale references to previous sublayers' IDs. Although they exist in the registry on load, their IDs no longer match those referenced. Support has been added for the serialisation and deserialisation of anonymous sublayers to Maya files by the layer manager.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/LayerManager.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/LayerManager.cpp
@@ -22,6 +22,7 @@
 
 #include <mayaUsd/listeners/notice.h>
 
+#include <pxr/base/tf/stringUtils.h>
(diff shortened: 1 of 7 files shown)


J-Mo63:
@kxl-adsk likewise this one is getting the same issue as PR https://github.com/Autodesk/maya-usd/pull/1866, where `testVP2RenderDelegateIsolateSelect` is failing for just `MayaUsd - Maya Master [USD max, Python 3, Interactive] (Linux)`.

kxl-adsk (approved):

kxl-adsk:
Yes. We have something to investigate...it's unclear to me which change caused it 😕

PR #1874, opened 2021-12-03
Make Enum UI control for TfToken type in case if usd schema contains list of possible values for the custom TfToken render setting

Vsevolod1983:
Please consider the following change:

In hdRPR render delegate we have bunch of custom render settings which we define using USD schema mechanism.
Some of these setting - are essentially attributes of TfEnum type. However, It looks like we can use only TfToken type for these attributes (we found it problematic to use TfEnum in usd schema).

What this PR does:
1. Adds support for TfToken within renderGlobals.cpp
2. We create Enum Maya attribute for TfToken type in case if we have list of possible values defined in USD schema for render setting for particular render delegate


nxkb:
Hi there, thanks for the PR.

I was initially surprised that TfEnum wouldn't be the way to express this in a schema. What kind of problems did you have?  Did you ever get some explanation for why you should do it this way?

Still, it looks like the schema's in USD itself use the TfToken + AllowedTokens approach so I think supporting it makes sense. 

- I would just add a comment saying "if a schema attribute has AllowedTokens set, we treat it as an enum instead". 
- You also need to run clang-format on your PR. There's a handy script in the repo called `.github/run-clang-format.py` which will get things formatted for you.

lib/mayaUsd/render/mayaToHydra/renderGlobals.cpp
```diff
@@ -920,11 +938,25 @@ MObject MtohRenderGlobals::CreateAttributes(const GlobalParams& params)
                     attr.defaultValue.UncheckedGet<GfVec4f>(),
                     userDefaults);
             } else if (attr.defaultValue.IsHolding<TfToken>()) {
-                _CreateStringAttribute(
-                    node,
-                    filter.mayaString(),
-                    attr.defaultValue.UncheckedGet<TfToken>().GetString(),
-                    userDefaults);
+				bool canCreateAsEnum = false;
```
nxkb, line 941: to consider: maybe "createdAsEnum" is more accurate to how this is used later.

lib/mayaUsd/render/mayaToHydra/renderGlobals.cpp
```diff
@@ -920,11 +938,25 @@ MObject MtohRenderGlobals::CreateAttributes(const GlobalParams& params)
                     attr.defaultValue.UncheckedGet<GfVec4f>(),
                     userDefaults);
             } else if (attr.defaultValue.IsHolding<TfToken>()) {
-                _CreateStringAttribute(
-                    node,
-                    filter.mayaString(),
-                    attr.defaultValue.UncheckedGet<TfToken>().GetString(),
-                    userDefaults);
```
nxkb, line 927: This is where I'd put something like:
```cpp
// If this attribute type has AllowedTokens set, we treat it as an enum instead, so that only valid values are available.```

nxkb (changes requested):

Vsevolod1983:
Hello, @nxkb !
I made two changes but have a question about using python script regarding clang format
Tried to run the script using my python 3.7.1 but got an error, saying that somea files are missing

FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:/Projects/maya-usdForked/maya-usd/.github/lib/mayausd/base/api.h' 

Looks like .github folder does not contain some files.
Could you please point to any instructions how to properly use the script?
Thank you!

Vsevolod1983:
Hopefully, managed to fix it manually.

seando-adsk:
@Vsevolod1983 Once the code review is completed by @nxkb I will merge this PR.

Vsevolod1983:
@seando-adsk 
Hi !
Could we approve the PR if there are no any questions on that?
Thank you!

kxl-adsk:
Hi @Vsevolod1983. Sorry about that. We will take it over. Can you please make sure to run clang-format? I'm sure PF will fail.

Vsevolod1983:
@kxl-adsk 
Hi!
I did not manage to run cmake format, I was getting: 
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:/Projects/maya-usdForked/maya-usd/.github/lib/mayausd/base/api.h'

So, either some files are missing in the repo or I'm doing something wrong.
Could you please point me how to run this?

Vsevolod1983:
@seando-adsk 
Ok, great. I managed to run this. Changes are committed.
Thank you!

kxl-adsk (approved):

PR #1875, opened 2021-12-03
MAYA-113917 Skip Ufe Identifier update when the identifiers are not in use.

williamkrick:
This change mitigates the high performance cost of reading the prim path of instanced objects from USD by only reading the path when it is necessary.


perrauo-adsk (dismissed):

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -46,6 +47,7 @@
 #include <pxr/usd/usd/prim.h>
 #include <pxr/usdImaging/usdImaging/delegate.h>
 
+#include <maya/M3dView.h>
```
JGamache-autodesk, line 50: Minor: Can you remove this include? I did not see it used.

JGamache-autodesk (approved):

JGamache-autodesk (approved):

PR #1877, opened 2021-12-03
MAYA-113820: [Github 1543] Unloaded prims should appear in the outliner

seando-adsk:
#### MAYA-113820: [Github 1543] Unloaded prims should appear in the outliner
* In default prim predicate remove IsLoaded (so we will always display both loaded and unloaded prims).

### DIFF
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
@@ -43,9 +43,15 @@
 PXR_NAMESPACE_USING_DIRECTIVE
 
 namespace {
+
+    // We want to display the unloaded prims, so removed UsdPrimIsLoaded from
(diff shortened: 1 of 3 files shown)


ppt-adsk (changes requested):
1. Needs a test :)
2. Great stuff, but there is also the UsdPrimDefaultPredicate in ProxyShapeHierarchy.cpp.  At some point we should take the time to factor out common code from both these classes...

seando-adsk:
Right thanks for catching the ProxyShapeHierarchy. I didn't do any code refactoring yet. Let's do that separately.
I've added a test for the unload to verify children count.

ppt-adsk:
> I didn't do any code refactoring yet. Let's do that separately.
Absolutely, it's out of scope for this pull request.  Thanks for the fix!

ppt-adsk (dismissed):

ppt-adsk (approved):

PR #1878, opened 2021-12-03
MAYA-114591 Convert UsdMayaPrimUpdater::isAnimated() use of Python to…

boudrey:
… C++

Actually the python part never worked so removing it.


ppt-adsk:
Since we're losing the original intent of the code, I'd want to have Krystian comment on whether it's still relevant, and if so if this is future work.  If it's no longer relevant, your change is fine.

ppt-adsk (approved):

kxl-adsk (approved):

PR #1883, opened 2021-12-06
t_bailp/MAYA-114109/undo redo push pull simplified

pierrebai-adsk:
Add undo / redo to Edit-as-Maya workflows.

Create commands:
- Add mayaUsdEditAsMaya command for edit as Maya.
- Add mayaUsdDuplicate command for duplicate as Maya and duplicate to USD.
- Add mayaUsdDiscardEdits command for discarding edits.
- Add mayaUsdMergeToUsd command to merge back to USD.
- This makes it easy to script and support undo/redo.
- Remove the old mel script for pull/push/discards/duplicate, replace them with these commands.
- Add unit tests for all new commands.

Create system to record undo items:
- Add OpUndoInfo to record all undoable steps for each commands.
- Add an instance to UsdUndoManager singleton to record undo items from anywhere.
- Add OpUndoItem to record a single undoable sub-operation step.
- Add OpUndoInfoMuting to muting recording undo items in block of code that handles undo differently.
- Add OpUndoInfoRecorder to automatically record all undo items in a block of scoped code.

Provide undo items for many common scenarios:
- Implement OpUndoItem for USD undo block (UsdUndoBlock / UsdUndoItem).
- Implement OpUndoItem for MDagModifier.
- Implement OpUndoItem for MDGModifier.
- Implement OpUndoItem for selection.
- Implement OpUndoItem for locking nodes.
- Implement OpUndoItem for pair of Python do/undo scripts.
- Implement OpUndoItem for any generic pair of do/undo C++ functions.

Various code adjustments:
- Remove the cached pull root from PrimUpdaterManager since it was interfering with undo/redo.
- Retrieving it for each command is not costly and avoid needing us to clear it.
- Use undo info muting in read job (import) class since it implements its own undo / redo.


lib/mayaUsd/undo/OpUndoInfo.h
```diff
@@ -0,0 +1,148 @@
...
+//------------------------------------------------------------------------------
+// OpUndoItem
+//------------------------------------------------------------------------------
+
+/// \class OpUndoItem
+/// \brief Record data needed to undo or redo a single undo sub-operation.
+///
+/// See OpUndoItems.h for concrete implementations.
+
+class OpUndoItem
```
pierrebai-adsk, line 40: The edit-as-maya nd merge-to-usd code needs to call commands, Maya API, Python code, etc. Some of these are not fully supported in MDGModifier. For example, it can run Python code, but we cannot provide the 'undo' Python code.

So this undo item interface gets implemented by classes in OpUndoItems.h to support all the need to the prim updater.

lib/mayaUsd/undo/OpUndoInfo.h
```diff
@@ -0,0 +1,148 @@
...
+};
+
+//------------------------------------------------------------------------------
+// OpUndoInfo
+//------------------------------------------------------------------------------
+
+/// \class OpUndoInfo
+/// \brief Record everything needed to undo or redo a complete operation or command.
+
+class OpUndoInfo
```
pierrebai-adsk, line 79: All OpUndoItem are added to a OpUndoInfo container. One is held in the UsdUndoManager singleton. The extract() method allows taking ownerships of all undo item into a command. That is what EditAsMayaCommand and others do to implement their undo / redo.

lib/mayaUsd/undo/OpUndoInfoMuting.h
```diff
@@ -0,0 +1,48 @@
...
+#include <mayaUsd/undo/UsdUndoManager.h>
+
+namespace MAYAUSD_NS_DEF {
+
+//! \brief Turn off undo info recording for a given scope.
+//
+// Useful if code implement their own undo/redo without using the undo info
+// but calls functions that generate undo info items that need to be ignored.
+
+class MAYAUSD_CORE_PUBLIC OpUndoInfoMuting
```
pierrebai-adsk, line 30: Since all OpUndoItem are added to the UsdUndoManager singleton, for code that don't want to be undoable, we need a way to clear the genarated undo items. That is what the OpUndoInfoMuting class does: it will clear all undo items generated while a muting instance is created.

lib/mayaUsd/undo/OpUndoInfoRecorder.h
```diff
@@ -0,0 +1,59 @@
...
+#include <mayaUsd/undo/UsdUndoManager.h>
+
+namespace MAYAUSD_NS_DEF {
+
+//! \brief Record and extract undo item in the scope where it is declared.
+//
+// Useful if code implement their undo/redo using the OpUndoInfo and need
+// to relieably extract the undo items from the UsdUndoManager.
+
+class MAYAUSD_CORE_PUBLIC OpUndoInfoRecorder
```
pierrebai-adsk, line 30: The OpUndoInfoRecorder is a convenience class to receive the undo items. It will transfer to the target OpUndoInfo all items generated while it exists. Meant to be used on the stack.

lib/mayaUsd/undo/OpUndoItems.h
```diff
@@ -0,0 +1,423 @@
...
+#include <maya/MGlobal.h>
+#include <maya/MSelectionList.h>
+
+#include <memory>
+#include <vector>
+
+namespace MAYAUSD_NS_DEF {
+
+//------------------------------------------------------------------------------
+// NodeDeletionUndoItem
```
pierrebai-adsk, line 36: Various implementations of OpUndoItem for the functionality needed for edit-as-maya, but the implementations are very generic.

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -589,15 +589,10 @@ global proc mayaUsdMenu_pushBackToUSD(string $obj)
         }
     }
     if (size($obj)) {
-        python("from mayaUsd.lib import PrimUpdaterManager; PrimUpdaterManager.mergeToUsd('" + $obj + "');");
```
pierrebai-adsk, line 592: In MEL, many places where there used to be Python code calling into C++ are replaced by proper commands.

lib/mayaUsd/fileio/jobs/readJob.cpp
```diff
@@ -97,6 +102,10 @@ UsdMaya_ReadJob::~UsdMaya_ReadJob() { }
 
 bool UsdMaya_ReadJob::Read(std::vector<MDagPath>* addedDagPaths)
 {
+    // Do not use the global undo info recording system.
+    // The read job Undo() / redo() functions will handling all operations.
+    OpUndoInfoMuting undoMuting;
```
pierrebai-adsk, line 107: The read job is an example of undo muting: it implements its own undo / redo so all undo items generated by the importer need to be discarded.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -284,61 +285,121 @@ PullImportPaths pullImport(
...
-    bool success = readJob.Read(&addedDagPaths);
+    bool success = readJob->Read(&addedDagPaths);
     if (!success || addedDagPaths.size() == 0) {
         return PullImportPaths({}, {});
     }
 
+    // Note: UsdMaya_ReadJob has explicit Read(), Undo() and Redo() functions,
+    //       and Read() has already been called, so create the function-undo item
+    //       but do not execute it.
+    FunctionUndoItem::create(
```
pierrebai-adsk, line 311: Most of the changes in the PrimUpdaterManager are to convert direct C++ calls to OpUndoItem creation. Some previously non-undoable stuff had to be converted to uodobale and some operations had to be refined or reordered to work with undo.

lib/mayaUsd/undo/OpUndoItems.h
```diff
@@ -0,0 +1,423 @@
...
+
+    /// \brief construct a Maya DG modifier recorder.
+    MAYAUSD_CORE_PUBLIC
+    NodeDeletionUndoItem(std::string name)
+        : OpUndoItem(std::move(name))
+    {
+    }
+
+    MAYAUSD_CORE_PUBLIC
+    ~NodeDeletionUndoItem() override;
```
ppt-adsk, line 63: Looks like your destructor doesn't do anything beyond what the base class does.  If that's the case, you may want to just leave it out altogether and let the compiler generate it for you.  I guess if yu really want to declare it you could say
~NodeDeletionUndoItem() override = default;
but then what's the point of overriding it...

lib/mayaUsd/undo/OpUndoItems.h
```diff
@@ -0,0 +1,423 @@
...
+    MSelectionList          _selection;
+    MSelectionList          _previousSelection;
+    MGlobal::ListAdjustment _selMode;
+};
+
+//------------------------------------------------------------------------------
+// LockNodesUndoItem
+//------------------------------------------------------------------------------
+
+/// \class LockNodesUndoItem
```
ppt-adsk, line 351: Would be nice to mention that this is for Maya nodes.

lib/mayaUsd/ufe/PullPushCommands.cpp
```diff
@@ -0,0 +1,371 @@
...
+    case MStatus::kUnknownParameter: reportError("Invalid parameter."); break;
+    case MStatus::kSuccess: break;
+    default: reportError("Command parsing error."); break;
+    }
+
+    return status;
+}
+
+// Parses a string as a UFE path, detecting invalid paths and empty paths.
+MStatus parseArgAsUfePath(const MString& arg, Ufe::Path& outputPath)
```
ppt-adsk, line 57: Do you really need to tell the difference between MS::kNotFound and MS::kInvalidParameter?  May as well return an empty Ufe::Path for the error case if you don't care.

lib/mayaUsd/ufe/PullPushCommands.cpp
```diff
@@ -0,0 +1,371 @@
...
+{
+    // Note: parsing can throw exceptions. Treat them as parsing error and return an empty path.
+    try {
+        outputPath = Ufe::PathString::path(arg.asChar());
+
+        if (outputPath.size() == 0)
```
ppt-adsk, line 63: Nit-pick: outputPath.empty()

lib/mayaUsd/ufe/PullPushCommands.cpp
```diff
@@ -0,0 +1,371 @@
...
+    syntax.enableEdit(false);
+
+    for (int i = 0; i < paramCount; ++i)
+        syntax.addArg(MSyntax::kString);
+
+    return syntax;
+}
+
+// Verifies if a UFE path corresponds to a valid USD prim.
+bool isPrimPath(const Ufe::Path& path) { return ufePathToPrim(path).IsValid(); }
```
ppt-adsk, line 87: Function name feels awkward, since it's checking prim validity as well.  Maybe isPathToValidPrim()?

lib/mayaUsd/ufe/PullPushCommands.cpp
```diff
@@ -0,0 +1,371 @@
...
+
+// Parse the indexed argument as a UFE path.
+MStatus parseUfePathArg(const MArgParser& argParser, int index, Ufe::Path& outputPath)
+{
+    MString text;
+    MStatus status = parseTextArg(argParser, index, text);
+    if (MS::kSuccess != status)
+        return status;
+
+    status = parseArgAsUfePath(text, outputPath);
```
ppt-adsk, line 107: From what I can tell here, you can just
return parseArgAsUfePath(...);

lib/mayaUsd/fileio/jobs/readJob.cpp
```diff
@@ -97,6 +102,10 @@ UsdMaya_ReadJob::~UsdMaya_ReadJob() { }
 
 bool UsdMaya_ReadJob::Read(std::vector<MDagPath>* addedDagPaths)
 {
+    // Do not use the global undo info recording system.
+    // The read job Undo() / redo() functions will handling all operations.
```
ppt-adsk, line 106: "will handle".

lib/mayaUsd/fileio/jobs/readJob.cpp
```diff
@@ -254,8 +263,9 @@ bool UsdMaya_ReadJob::Read(std::vector<MDagPath>* addedDagPaths)
         std::make_pair(rootPathToRegister.GetString(), mMayaRootDagPath.node()));
 
     if (mArgs.useAsAnimationCache) {
-        MDGModifier dgMod;
-        MObject     usdStageNode = dgMod.createNode(UsdMayaStageNode::typeId, &status);
+        auto&        undoInfo = UsdUndoManager::instance().getUndoInfo();
```
ppt-adsk, line 266: I'm a bit puzzled here.  Aren't we in the same method that has the undo muting guard?

lib/mayaUsd/fileio/utils/meshReadUtils.cpp
```diff
@@ -99,7 +99,7 @@ bool addCreaseSet(
     // .../lib/python2.7/site-packages/maya/app/general/creaseSetEditor.py
 
     MObject creasePartitionObj;
-    *statusOK = UsdMayaUtil::GetMObjectByName(":creasePartition", creasePartitionObj);
+    *statusOK = UsdMayaUtil::GetMObjectByName(MString(":creasePartition"), creasePartitionObj);
```
ppt-adsk, line 102: Hopefully we don't need this change.

lib/mayaUsd/fileio/primUpdater.cpp
```diff
@@ -73,15 +77,32 @@ bool UsdMayaPrimUpdater::editAsMaya(const UsdMayaPrimUpdaterContext& context) {
...
     }
-    return true;
+
+    return status == MS::kSuccess;
 }
 
 bool UsdMayaPrimUpdater::pushEnd(const UsdMayaPrimUpdaterContext& context)
 {
-    return discardEdits(context);
+    // Nothing. We will rely on the Maya scene cleaup code
```
ppt-adsk, line 103: What is the Maya scene cleanup code?

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -735,6 +820,22 @@ bool PrimUpdaterManager::mergeToUsd(const MFnDependencyNode& depNodeFn, const Uf
         return false;
     }
 
+    // Discard all pulled Maya nodes.
+    std::vector<MDagPath> toApplyOn = UsdMayaUtil::getDescendantsStartingWithChildren(mayaDagPath);
```
ppt-adsk, line 824: Doesn't this mean that prim updaters that pull and generate both Dag and DG nodes will have to delete their DG nodes, but not their Dag nodes, because the Dag node delete is being done here?  That feels awkward, if that's the case.

ppt-adsk (changes requested):
Amazing work!  The amount of changes that you needed to do is quite something.  If there are comments that require clarification, just reach out to me.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -76,7 +81,7 @@ namespace {
 const std::string kPullParentPathKey("Maya:Pull:ParentPath");
 
 // Set name that will be used to hold all pulled objects
-const std::string kPullSetName("pullStateSet");
```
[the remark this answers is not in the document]
pierrebai-adsk: The API being called using the name is taking a MString, so it made sense to keep the constant with the same type to avoid unnecessary copies when using it.

lib/mayaUsd/undo/OpUndoInfo.h
```diff
@@ -0,0 +1,148 @@
...
+    MAYAUSD_CORE_PUBLIC
+    bool isDeleted(const MObjectHandle obj) const;
+
+    /// \brief clear all undo/redo information contained here.
+    MAYAUSD_CORE_PUBLIC
+    void clear();
+
+    /// \brief extract all undo/redo information contained here and clear.
+    MAYAUSD_CORE_PUBLIC
+    OpUndoInfo extract();
```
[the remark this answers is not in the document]
pierrebai-adsk: Well, I did not want to leave the OpUndoInfo object in the invalid state that a move assignment / constructor would.  This is especially important because there is a permanent instance in the UsdUndoManager singleton.

lib/mayaUsd/undo/OpUndoInfo.h
```diff
@@ -0,0 +1,148 @@
...
+    MAYAUSD_CORE_PUBLIC
+    void addItem(OpUndoItem::Ptr&& item);
+
+    /// \brief add an object that was deleted, to help avoid deleting an object twice.
+    MAYAUSD_CORE_PUBLIC
+    void addDeleted(const MObjectHandle obj);
+
+    /// \brief verify if an object was already deleted, to help avoid deleting an object twice.
+    MAYAUSD_CORE_PUBLIC
+    bool isDeleted(const MObjectHandle obj) const;
```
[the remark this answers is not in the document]
pierrebai-adsk: I guess it was my inexperience with undo/redo showing. I'll try to use MObjectHandle and get rid of this deleted object tracker. IDK how it will interact with other code that keep MObject and not MObjectHandle, but i'll see when trying it out.

lib/mayaUsd/undo/OpUndoInfo.h
```diff
@@ -0,0 +1,148 @@
...
+
+    /// \brief redo a complete PrimUpdaterManager operation.
+    MAYAUSD_CORE_PUBLIC
+    bool redo();
+
+    /// \brief add a undo item. This takes ownership of the item.
+    MAYAUSD_CORE_PUBLIC
+    void addItem(OpUndoItem::Ptr&& item);
+
+    /// \brief add an object that was deleted, to help avoid deleting an object twice.
```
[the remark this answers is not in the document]
pierrebai-adsk: The double deletions only happened during a single command. After the command, the object being deleted prevent other code from interacting with it, that is why it did not need to be global. As discussed, I will try to get rid of it entirely.

lib/mayaUsd/undo/OpUndoInfo.h
```diff
@@ -0,0 +1,148 @@
...
+    MAYAUSD_CORE_PUBLIC
+    const std::string& getName() const { return _name; }
+
+private:
+    // Name the undo items to help debugging, tracing and logging.
+    std::string _name;
+};
+
+//------------------------------------------------------------------------------
+// OpUndoInfo
```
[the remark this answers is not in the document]
pierrebai-adsk: Initially I assumed more thing would be needed than a mere item list, but it turned out to be mostly that, so I'll rename it to OpUndoItemList.

lib/mayaUsd/undo/OpUndoInfo.h
```diff
@@ -0,0 +1,148 @@
...
+    MAYAUSD_CORE_PUBLIC
+    bool undo();
+
+    /// \brief redo a complete PrimUpdaterManager operation.
+    MAYAUSD_CORE_PUBLIC
+    bool redo();
+
+    /// \brief add a undo item. This takes ownership of the item.
+    MAYAUSD_CORE_PUBLIC
+    void addItem(OpUndoItem::Ptr&& item);
```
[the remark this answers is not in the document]
pierrebai-adsk: The OpUndoItem sub-class that handles the deletion adds it automatically, but as suggested this may very well go away entirely.

lib/mayaUsd/undo/OpUndoInfo.h
```diff
@@ -0,0 +1,148 @@
...
+    std::vector<OpUndoItem::Ptr> _undoItems;
+    DeletedObjectSet             _deletedMayaObjects;
+    bool                         _isUndone = false;
+};
+
+} // namespace MAYAUSD_NS_DEF
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+using OpUndoInfo = MAYAUSD_NS_DEF::OpUndoInfo;
```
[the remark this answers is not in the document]
pierrebai-adsk: I had written a lot of code when the OpUndoInfo switched from being under PXR ns to MAYAUSD and I had added this as a quick way to get the code compiling again and then forgot about it. I will get rid of this.

lib/mayaUsd/utils/util.cpp
```diff
@@ -1234,8 +1244,9 @@ MPlug UsdMayaUtil::GetConnected(const MPlug& plug)
 
 void UsdMayaUtil::Connect(const MPlug& srcPlug, const MPlug& dstPlug, const bool clearDstPlug)
```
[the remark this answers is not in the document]
pierrebai-adsk: Yes, that is a pattern used elsewhere, I'll do it.

lib/mayaUsd/undo/UsdUndoManager.h
```diff
@@ -55,10 +57,16 @@ class MAYAUSD_CORE_PUBLIC UsdUndoManager
     // tracks layer states by spawning a new UsdUndoStateDelegate
     void trackLayerStates(const SdfLayerHandle& layer);
 
+    // Retrieve the operation undo info, used to record undo items.
+    // The undo info can later be extracted into a command to implement
+    // its undo and redo. See: OpUndoInfo::extract()
+    OpUndoInfo& getUndoInfo() { return _undoInfo; }
```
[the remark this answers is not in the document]
pierrebai-adsk: NP, Krystian had suggested I put it here as they were both undo related, but I can create a new singleton.

lib/mayaUsd/undo/OpUndoInfo.cpp
```diff
@@ -0,0 +1,95 @@
...
+        for (auto iter = _undoItems.begin(); iter != end; ++iter)
+            iter->reset();
+    }
+
+    _undoItems.clear();
+    _deletedMayaObjects.clear();
+    _isUndone = false;
+}
+
+OpUndoInfo OpUndoInfo::extract()
```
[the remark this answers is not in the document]
pierrebai-adsk: Like I explained, C++ move operator leave the object in an unknown, but destructible state that must not be used. But we do keep OpUndoInfo permanently in the singleton, so we cannot leave it in this destructible-only state, unfortunately.

lib/mayaUsd/fileio/translators/translatorCamera.cpp
```diff
@@ -327,8 +331,9 @@ bool UsdMayaTranslatorCamera::Read(
     }
 
     // Create the camera shape node.
-    MDagModifier dagMod;
-    MObject      cameraObj
+    auto&         undoInfo = UsdUndoManager::instance().getUndoInfo();
+    MDagModifier& dagMod = MDagModifierUndoItem::create("Camera creation", undoInfo);
```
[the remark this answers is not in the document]
pierrebai-adsk: Originally, the OpUndoInfo was passed around, it was not a singleton. Now that it is a singleton, we could very well do this.

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
```diff
@@ -322,7 +323,7 @@ bool UsdMayaTranslatorMaterial::AssignMaterial(
         = UsdMayaTranslatorMaterial::Read(jobArguments, meshMaterial, primSchema, context);
 
     if (shadingEngine.isNull()) {
-        status = UsdMayaUtil::GetMObjectByName("initialShadingGroup", shadingEngine);
+        status = UsdMayaUtil::GetMObjectByName(MString("initialShadingGroup"), shadingEngine);
```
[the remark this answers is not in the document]
pierrebai-adsk: No, it's the opposite: there used to be only one version take a std::string and converting to MString. Now there is an overload for both MString and std::string, but that makes calling with a string literal ambiguous.

lib/mayaUsd/fileio/jobs/readJob.cpp
pierrebai-adsk: Ah yes, but the singleton and muting came after :) I'll revert this.

lib/mayaUsd/fileio/utils/meshReadUtils.cpp
pierrebai-adsk: I'll try to remove the overload taking a std::string and only keep the MString one, since that its the type that is needed.

lib/mayaUsd/fileio/primUpdater.cpp
pierrebai-adsk: Oh it's the code that goes through the Maya nodes in PrimUpdaterManager. I'll clarify the comment to say that instead.

lib/mayaUsd/fileio/primUpdaterManager.cpp
pierrebai-adsk: I think we have no choice. The Maya node have to be delete in the right order so that parent are not deleted before children, but individual updaters don't know what other updaters do. It seems safer to let the PrimUpdaterManager handle the deletions.

We might end-up having to manager the DG node deletions too. In fact, we will probably have to document very well what updater must do to be compatible with undo / redo.

lib/mayaUsd/fileio/utils/meshReadUtils.cpp
pierrebai-adsk: I tried removing the overload and the result is worse: many files hold std::string and would need to be modified.

lib/mayaUsd/commands/PullPushCommands.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+// Parse the indexed argument as a UFE path.
+MStatus parseUfePathArg(const MArgParser& argParser, int index, Ufe::Path& outputPath)
+{
+    MString text;
+    MStatus status = parseTextArg(argParser, index, text);
+    if (MS::kSuccess != status)
+        return status;
+
+    status = parseArgAsUfePath(text, outputPath);
```
ppt-adsk, line 105: See previous comment about just
return parseArgAsUfePath(text, outputPath);

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1133,13 +1109,13 @@ PrimUpdaterManager& PrimUpdaterManager::getInstance()
 MObject PrimUpdaterManager::findOrCreatePullRoot()
 {
     MObject pullRoot = findPullRoot();
-    if (!pullRoot.isNull())
+    if (!pullRoot.isNull()) {
+        _hasPulledPrims = true;
```
ppt-adsk, line 1113: Why is this line needed?  I thought that the commands below would set it correct.  Shouldn't this rather be
TF_AXIOM(_hasPulledPrims);
?

ppt-adsk (changes requested):

ppt-adsk:
Have a look at the unresolved conversations, would be nice to clear those out one way or another.  Just comment on those suggestions you feel aren't useful or don't work out, I can resolve those.

lib/mayaUsd/undo/OpUndoInfoRecorder.h
```diff
@@ -0,0 +1,59 @@
...
+    //! \brief Starts recording undo info in the given container.
+    OpUndoInfoRecorder(OpUndoInfo& undoInfo);
+
+    //! \brief Ends recording undo info.
+    ~OpUndoInfoRecorder();
+
+    //! \brief Starts recording undo info in the given container.
+    //
+    // Discard any previously recorded info.
+    void startUndoRecording();
```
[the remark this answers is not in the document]
pierrebai-adsk: Only if code would wat to re-use an instance, for example if they kept it as a member for some reason.

lib/mayaUsd/undo/OpUndoInfoRecorder.cpp
```diff
@@ -0,0 +1,55 @@
...
+OpUndoInfoRecorder::OpUndoInfoRecorder(OpUndoInfo& undoInfo)
+    : _isRecording(false)
+    , _undoInfo(undoInfo)
+{
+    startUndoRecording();
+}
+
+OpUndoInfoRecorder::~OpUndoInfoRecorder() { endUndoRecording(); }
+
+void OpUndoInfoRecorder::startUndoRecording()
```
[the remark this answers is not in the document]
pierrebai-adsk: Will cancel previously-recorded data if there are any. Like I mentioned above, I wrote it this way to allow re-using the object if needed.

lib/mayaUsd/undo/OpUndoItems.h
pierrebai-adsk: I often override destructors so to be able to put breakpoint in them when needed.

pierrebai-adsk:
I seem to have missed a bunch of your initial comments. I think github folds them when there are too many I I did not notice.

lib/mayaUsd/fileio/primUpdaterManager.cpp
pierrebai-adsk: Ah yes, I'll remove this.

lib/mayaUsd/ufe/PullPushCommands.h
```diff
@@ -0,0 +1,218 @@
...
+namespace MAYAUSD_NS_DEF {
+namespace ufe {
+
+//------------------------------------------------------------------------------
+// EditAsMayaCommand
+//------------------------------------------------------------------------------
+
+//! \brief Edit as maya undoable command.
+
+class EditAsMayaCommand : public MPxCommand
```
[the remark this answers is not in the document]
pierrebai-adsk: While many of these function look superficially the same, they all differ a bit: some parse the argument as UFE path, some as Dag path, some as text. Some syntax have 1 argument, some 2.

undoIt and redoIt and isUndoable can be shared in a base class.

lib/mayaUsd/undo/OpUndoItems.h
```diff
@@ -0,0 +1,423 @@
...
+//------------------------------------------------------------------------------
+// NodeDeletionUndoItem
+//------------------------------------------------------------------------------
+
+/// \class NodeDeletionUndoItem
+/// \brief Record data needed to undo or redo a Maya DG sub-operation.
+///
+/// For node deletion, use the specialized undo item that tracks which objects
+/// have already been deleted and avoid double-deletions.
+class NodeDeletionUndoItem : public OpUndoItem
```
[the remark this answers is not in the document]
pierrebai-adsk: I'm not convinced adding a templated base class to share three one-line functions is worth it. Especially since the code will have to live in the header instead.

pierrebai-adsk:
I think I've addressed all your comments this time, but it's easy to miss one due to github UI.

pierrebai-adsk:
Note: the undo / redo code break a new test that is in a commit ahead of my branch. I'll have to rebase again later. I'll wait for now to avoid losing the comments threads.

ppt-adsk (approved):
Fantastic work, a lot of effort went into this, thanks for pushing it through!

ppt-adsk:
> I think I've addressed all your comments this time, but it's easy to miss one due to github UI.

I know, the github workflow for conversations is really, really poor, and forces you to jump around file versions a lot, because of "obsolete" comments.  I go with unresolved conversations, until they're all resolved.

pierrebai-adsk:
I need to rebase so i'll close this PR and re-open after the rebase.

PR #1885, opened 2021-12-06
MAYA-115158 remove edit-as-maya menus when not available

pierrebai-adsk:


ppt-adsk (approved):

pierrebai-adsk:
One test failed in two of the dozens of PF: testVP2RenderDelegateIsolateSelect. IDK if it is known to sometimes fail and thus should not be blocking the PR?

plugin/adsk/scripts/USDMenuProc.mel
```diff
@@ -13,8 +13,22 @@
 // limitations under the License.
 //
 
+int $_hasPrimUpdater = -1;
```
seando-adsk, line 16: I think we use the Maya standard naming for globals  which would be "$gXXX"

ppt-adsk: Yeah, thought about that, but in this case we don't really want to expose this variable, it should not be accessed outside this source file.  I don't know if we have a naming convention for that.

plugin/adsk/scripts/mayaUsd_pluginUICreation.mel
```diff
@@ -2,6 +2,7 @@ global proc mayaUsd_pluginUICreation()
 // Callback registered with the MFnPlugin for when the plugin is loaded.
 // Used to initialize any UI we create.
 {
+    source "USDMenuProc.mel";
```
seando-adsk, line 5: I'm a little concerned about moving this up to the top. That file uses the ufeSupport plugin which is only loaded below. I know this file only contains proc which are called on RMB, but by using ufeSupport it implies that it will already be loaded.

I would suggest putting your new global function hasPrimUpdater in this file (at the top). That way it will be available to any UI function that needs it such as UsdMenuProc and mayaUsdMenu - as both of these sourced below.

ppt-adsk (approved):

seando-adsk (approved):

pierrebai-adsk:
The only test failure was the unrelated test, which also fails in my other unrelated branch with node code change in common, so I'm sure this is fine to be merged.

PR #1886, opened 2021-12-07
Updated AL Python tests to use fixturesUtils

J-Mo63:
This PR updates the AL plugin's Python tests to make use of `fixturesUtils`, and removes all uses of `tempfiles` as recommended in https://github.com/Autodesk/maya-usd/pull/1792#discussion_r754488444. Some minor modifications have been made to the utility script for compatibility purposes.


plugin/al/plugin/AL_USDMayaTestPlugin/py/testLayerManager.py
```diff
@@ -25,26 +25,24 @@
...
-        if os.path.exists(cls._outputPath):
-            shutil.rmtree(cls._outputPath)
-        os.mkdir(cls._outputPath)
-        cls._mayaFilePath = os.path.join(cls._outputPath, 'LayerManagerSerialisation.ma')
-        cls._usdFilePath = os.path.join(cls._outputPath, 'LayerManagerSerialisation.usda')
-
-        # Load the plugin
-        cmds.loadPlugin("AL_USDMayaPlugin", quiet=True)
+        # Load the plugin and setup test output
+        _ = fixturesUtils.setUpClass(__file__, pluginName="AL_USDMayaPlugin")
```
kxl-adsk, line 37: Why assign the return value to `_` if you don't use it? I would prefer either a descriptive name of the variable or simply do not store return value.

plugin/al/plugin/AL_USDMayaTestPlugin/py/testProxyShape.py
```diff
@@ -30,10 +29,21 @@
...
     """Test cases for static function: AL.usdmaya.ProxyShape.getUsdPrimFromMayaPath"""
 
+    @classmethod
+    def setUpClass(cls):
+        # Setup for tests for output
+        _ = fixturesUtils.setUpClass(__file__, loadPlugin=False)
+
+    @classmethod
+    def tearDownClass(cls):
+        _ = fixturesUtils.tearDownClass(unloadPlugin=False)
```
kxl-adsk, line 45: This one is repeated a few times...`tearDownClass` doesn't return any value.

kxl-adsk (changes requested):

plugin/al/plugin/AL_USDMayaTestPlugin/py/testLayerManager.py
J-Mo63: Yeah sorry about that. Bad habit I've imported from Ruby or Lua probably. I'll just remove it.

J-Mo63:
@kxl-adsk I've gone ahead and actioned your feedback.

kxl-adsk (approved):

PR #1887, opened 2021-12-07
MAYA-115158 fix the python code to a a one-liner.

pierrebai-adsk:


plugin/adsk/scripts/mayaUsd_pluginUICreation.mel
```diff
@@ -2,9 +2,16 @@ int $gHasPrimUpdater = -1;
...
     if ($gHasPrimUpdater < 0) {
-        $gHasPrimUpdater = `python("def checkPrimUp():\n  try:\n    from mayaUsd.lib import PrimUpdaterManager;\n    return 1;\n  except:\n    return 0;\ncheckPrimUp()")`;
+        // Note: we use a convoluted call to exec() in Python because
+        //       try block in Python cannot be one-liners and the MEL
+        //       python() function only returns the computed value when
+        //       the supplied Python code is a one-liner.
+        //
+        //       Putting the Python try block inside exec() allow us to
+        //       makethe Python code look like a one-liner to MEL.
+        $gHasPrimUpdater = `python("res = -1; exec('try:\\n    from mayaUsd.lib import PrimUpdaterManager;\\n    res = 1;\\nexcept:\\n    res = 0;\\n'); res")`;
```
pierrebai-adsk, line 14: Note: exec() uses the locals from the caller, so the res set within exec(0 is seen in the res variable outside it.

pierrebai-adsk:
I tested this in 2022 and 2023.

plugin/adsk/scripts/mayaUsd_pluginUICreation.mel
seando-adsk: In python 2.7, exec is a statement whereas in py3 its a function. Is this code expected to work in Maya 2022 when using py2?

pierrebai-adsk:
Note: the only PF failure is the known vp2 isolate select test failure on Linux.

plugin/adsk/scripts/mayaUsd_pluginUICreation.mel
pierrebai-adsk: But the calling convention is the same, isn't it? What would break?

pierrebai-adsk: I just tried it in a python 2.7 prompt and the code worked.

pierrebai-adsk: `>>> res = -1; exec('try:\n    from mayaUsd.lib import sys;\n    res = 1;\nexcept:\n    res = 0;\n'); print(res)
0
>>> res = -1; exec('try:\n    import sys;\n    res = 1;\nexcept:\n    res = 0;\n'); print(res)
1
>>>
`

seando-adsk: Excellent. If it works, then I have no concerns.

seando-adsk (approved):

PR #1895, opened 2021-12-08
Switch delete command behavior to delete prims

kxl-adsk:
Cleaned up version of https://github.com/Autodesk/maya-usd/pull/1841


ppt-adsk (dismissed):

kxl-adsk:
@seando-adsk and @ppt-adsk - please have a look at this change. I still want to clean up the actual test a bit, but the rest should be ready to comment on.

You will also notice that old deactivate behaviour was kept for UFE v1. Implementing the new behaviour can't be achieved without changes to UFE and Maya, so it's out of the scope.

ppt-adsk (dismissed):

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +305,342 @@ def testDeleteArgs(self):
...
+        '''Delete Maya and USD objects passed as command arguments.'''
+
+        # Create our UFE notification observer
+        ufeObs = TestObserver()
+
+        if ufeUtils.ufeFeatureSetVersion() < 2:
+            ufe.Scene.addObjectDeleteObserver(ufeObs)
+            ufe.Scene.addObjectAddObserver(ufeObs)
+        else:
+            ufe.Scene.addObserver(ufeObs)
```
seando-adsk, line 321: You are skipping this test its not Ufe v2 or greater. So why do we need this condition? Line 317 will never be true - right?

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +305,342 @@ def testDeleteArgs(self):
...
+        self.assertIn(sphereShapeName, sphereChildrenNames)
+        self.assertIn(ball35Name, propsChildrenNames)
+
+        ball34PathString = "|transform1|proxyShape1,/Room_set/Props/Ball_34"
+        # Test that "|world" prefix is optional for multi-segment paths.
+        ball35PathString = "|transform1|proxyShape1,/Room_set/Props/Ball_35"
+
+        ufeObs.reset()
+        
+        # Cannot delete Ball_34 and Ball_34 because they is defined inside the variant composition arc
```
seando-adsk, line 368: I think one of those "34" should be "35"

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +305,342 @@ def testDeleteArgs(self):
...
+        '''Restrict deleting USD node. Cannot delete a prim defined on another layer.'''
+        
+        # Create our UFE notification observer
+        ufeObs = TestObserver()
+
+        if ufeUtils.ufeFeatureSetVersion() < 2:
+            ufe.Scene.addObjectDeleteObserver(ufeObs)
+            ufe.Scene.addObjectAddObserver(ufeObs)
+        else:
+            ufe.Scene.addObserver(ufeObs)
```
seando-adsk, line 400: Same comment as above - this test only runs in Ufe v2 or greater.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +305,342 @@ def testDeleteArgs(self):
...
+
+        cmds.file(new=True, force=True)
+
+        # open appleBite.ma scene in testSamples
+        mayaUtils.openAppleBiteScene()
+
+        # Create our UFE notification observer
+        ufeObs = TestObserver()
+
+        if ufeUtils.ufeFeatureSetVersion() < 2:
```
seando-adsk, line 435: And again.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +305,342 @@ def testDeleteArgs(self):
...
+        # Create some extra Maya nodes
+        cmds.polySphere()
+
+        # clear selection to start off
+        cmds.select(clear=True)
+
+        # Create our UFE notification observer
+        ufeObs = TestObserver()
+
+        if ufeUtils.ufeFeatureSetVersion() < 2:
```
seando-adsk, line 512: And again.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -310,10 +305,342 @@ def testDeleteArgs(self):
...
+
+        cmds.file(new=True, force=True)
+
+        # open Variant.ma scene in testSamples
+        mayaUtils.openVariantSetScene()
+
+        # Create our UFE notification observer
+        ufeObs = TestObserver()
+
+        if ufeUtils.ufeFeatureSetVersion() < 2:
```
seando-adsk, line 567: One more.

seando-adsk (changes requested):

kxl-adsk:
@seando-adsk thx for the feedback. Yes, I'm going to rewrite the `testDeleteCmd` file since there is a ton of duplication and gaps in testing.

kxl-adsk:
Wrote new tests...and discovered a potential issue in a multi-delete scenario. I will take it off-line since a fix in Maya may be required.

ppt-adsk (approved):

seando-adsk (approved):

lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
```diff
@@ -47,6 +47,28 @@ bool hasLayersMuted(const PXR_NS::UsdPrim& prim)
     }
     return false;
 }
+
+void removePrimRecursive(const PXR_NS::UsdPrim& prim, const PXR_NS::UsdStageWeakPtr& stage)
```
ppt-adsk, line 51: I might be misunderstanding the problem this is trying to solve.  Why is RemovePrim() of the root of the sub-graph insufficient?

I thought the problem we were trying to solve was a list of prims passed to the delete command, either through the command line or through selection, where a parent appears in the list before a child.  In such a case, won't RemovePrim() of the child still fail, as it will be a stale prim from the moment where the parent has been RemovePrim()'ed?

test/lib/ufe/testDeleteCmd.py
```diff
@@ -267,7 +267,7 @@ def testDeleteHierarchy(self):
         # delete two USD prims and Maya's shape
         ufeObs.reset()
         cmds.delete('|Tree_usd|Tree_usdShape,/TreeBase')
-        self.assertEqual(ufeObs.nbDeleteNotif() , 1)
+        self.assertEqual(ufeObs.nbDeleteNotif() , 4)
```
ppt-adsk, line 270: This is quite a shame.  We should only notify once, at the root of the sub-graph, when deleting a single prim.  Otherwise, for a large scene, this will not be scalable.  An explicit deletion list, either through a command line argument, or a selection list, should generate one notification per prim, but not here.

Perhaps an SdfChangeBlock will help?

ppt-adsk (commented):
I think I need a deeper dive as to what we're trying to solve, and as to the number of notifications delete must generate.

lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
kxl-adsk: There were a few problems I was chasing to make this work...but I'm glad you asked since recursive prim removal wasn't finally necessary. It's removed now.

test/lib/ufe/testDeleteCmd.py
kxl-adsk: Yep, a change block would help and make it match what redo does...but now it doesn't matter since recursive removal is gone.

kxl-adsk:
@ppt-adsk Have a look at the revised code.

ppt-adsk (approved):

kxl-adsk:
Update on code changes - I introduced a new CMake variable to easily switch between new and legacy delete behaviours. Currently enabled only for mainline builds. Maya 2022 will require a fix for the order of undo calls in the delete command.

Update on PF - we have one expected failure, i.e. last PR doesn't yet have the above fix. We also have two Linux failures that are not expected. Both Windows and Mac are passing (locally and on jenkins) so it's kind of a surprise. Needs investigation.

kxl-adsk:
In commit de1b62f199b3fdb099ccce02fe2cdee6993ad3a2 I fixed test failures on Linux caused by compilation issue around the define. 

Now we should be down to a single PF failure which will go away once we bump the last PR to PR132.

test/lib/mayaUsd/undo/testUsdUndoManager.py
```diff
@@ -106,6 +107,7 @@ def testNestedUsdUndoBlock(self):
         # expect to have 2 items on the undo queue
         self.assertEqual(cmds.undoInfo(q=True), nbCmds+2)
 
+    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() >= 2, 'For unknown reason this test is failing on UFE v1')
```
ppt-adsk, line 110: Do we have a JIRA ticket to keep track of this?

ppt-adsk (approved):

test/lib/mayaUsd/undo/testUsdUndoManager.py
kxl-adsk: Not yet.

ppt-adsk (approved):

PR #1897, opened 2021-12-09
Work around Pixar USD issue #1516 by forcing populate selection to be called every update!

williamkrick:


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -766,14 +766,18 @@ void ProxyRenderDelegate::_Execute(const MHWRender::MFrameContext& frameContext)
 #endif
 #endif // defined(MAYA_ENABLE_UPDATE_FOR_SELECTION)
 
+    // Work around for USD issue #1516. We don't know if any instanced object has
+    // had it's instance index change, so re-populate selection every update to
+    // ensure correct selection highlighting of instanced objects.
+    bool instanceIndexChanged = true;
```
williamkrick, line 772: This change fixes the incorrect selection highlight display. Unfortunately there is not a simple way to know if the Instance Index of a selected rprim has changed, so we're forced to call populateSelection every time.

The performance of calling populateSelection every time is not terrible, but it is also not great. The operation is linear in the number of entries in the selection list. When adding/removing rprims on the selection list the populateSelection time is a tiny fraction of the overall time. When the selection list is not changing and we're now forced to call populateSelect the performance cost is real but small. We can't create really long selection lists right now due to other scalability problems, so I think the best approach is to leave it populating selection every frame and then re-visit when we re-examine long selection list performance as a part of MAYA-114875.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -2246,166 +2254,177 @@ void HdVP2Mesh::_UpdateDrawItem(
...
-                                                       &sharedBBoxGeom]() {
-        const HdVP2DrawItem::RenderItemData& drawItemData = stateToCommit._renderItemData;
-        MHWRender::MRenderItem*              renderItem = drawItemData._renderItem;
-        if (ARCH_UNLIKELY(!renderItem))
-            return;
+    // We can get an empty stateToCommit when viewport draw modes change. In this case every
+    // rprim is marked dirty to give any stale render items a chance to update. If there are
+    // no stale render items then stateToCommit can be empty!
+    if (!stateToCommit.Empty()) {
+        _delegate->GetVP2ResourceRegistry().EnqueueCommit([stateToCommit,
```
williamkrick, line 2261: The change here is not create the main thread task if there is nothing to do. That buys us a significant performance improvement when we have to do extra updates to get correctness. We can easily create enough empty tasks to cause a "long tail" of serial main thread tasks that do nothing.

The rest of this file is a white space change for the indentation.

JGamache-autodesk (approved):

PR #1904, opened 2021-12-13
MAYA-114281 - Query a USD attributes locked state via getAttr

seando-adsk:
#### MAYA-114281 - Query a USD attributes locked state via getAttr
* Added tests for new Ufe support of attributes using Maya getAttr command.

### DIFF
--- a/test/lib/ufe/testAttribute.py
+++ b/test/lib/ufe/testAttribute.py
@@ -61,6 +61,11 @@ class AttributeTestCase(unittest.TestCase):
 
     pluginsLoaded = False
 
+    # Note: this variable is computed in the setUpClass method. Therefore
+    #       you cannot use it in a test function decorator (as its value
+    #       has not been computed yet).
+    getAttrSupportsUfe = False
+
     @classmethod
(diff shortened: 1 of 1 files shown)


test/lib/ufe/testAttribute.py
```diff
@@ -70,6 +75,19 @@ def setUpClass(cls):
...
+        if cls.pluginsLoaded:
+            import mayaUsd_createStageWithNewLayer
+            psPathStr = mayaUsd_createStageWithNewLayer.createStageWithNewLayer()
+            stage = mayaUsd.lib.GetPrim(psPathStr).GetStage()
+            stage.DefinePrim('/A', 'Xform')
+            try:
+                cmds.getAttr('|stage1|stageShape1,/A.visibility')
+                cls.getAttrSupportsUfe = True
+            except:
+                pass
```
ppt-adsk, line 88: I guess to get rid of the bandit security warning you could write
cls.getAttrSupportsUfe = False
here, even though it's redundant.

ppt-adsk (approved):

PR #1910, opened 2021-12-14
Use the quaternion type directly rather than interpreting the values of a vec4.

williamkrick:


lib/mayaUsd/render/vp2RenderDelegate/instancer.cpp
```diff
@@ -178,10 +178,10 @@ VtMatrix4dArray HdVP2Instancer::ComputeInstanceTransforms(SdfPath const& prototy
     if (_primvarMap.count(_tokens->rotate) > 0) {
         HdVP2BufferSampler sampler(*_primvarMap[_tokens->rotate]);
         for (size_t i = 0; i < instanceIndices.size(); ++i) {
-            GfVec4h quath;
+            GfQuath quath;
```
williamkrick, line 181: We could guard the two different branches here using a PXR_VERSION preprocessor macro, but the code as is will work correctly in 21.11 and older versions.

seando-adsk (approved):

Issue #1834, opened 2021-11-14
[MAYA-114979] UsdPreviewSurface export has several correctness problems


spiffmon:
**Describe the bug**
We only recently became aware that the PreviewSurface export logic that Pixar added (but do not yet use internally, which is why we didn't notice) has three serious problems, two of which prevent materials with normal maps from rendering correctly anywhere except Apple products (because they chose to code around the problems, off-spec).

1. The normal map textures being used generally do not contain accurate profiles or colorSpace or gamma information, and therefore Hio texture reading logic believes they are sRGB, when non-floating-point-encoded normal maps are in fact universally linear.  Unless Maya has absolute control over the provenance of the textures, we therefore suggest that the exporter set `inputs:sourceColorSpace = "linear"` for the UsdUVTexture node that reads a normal map.
2. 8-bit normal map textures must be mapped into the [-1, 1] range for UsdPreviewSurface, which can only be accomplished by authoring `inputs:scale` and `inputs:bias` appropriately on the normal map reader.
3. Shader prims cannot be nested under other Shader prims.  Although this does not *currently* cause problems in UsdImaging, it is a violation of the UsdShade Object Model, and may cause problems in future.  De-nesting the texture-coord readers from the texture readers will also allow all texture readers to share the same one, producing more efficient shading.

**Steps to reproduce**
1. Export any geometry with UsdPreviewSurface shading, and a normal map
2. In 21.11 or dev branch, run `usdchecker` in the result.

**Expected behavior**
We should see no warnings from usdchecker, and the normals should look the same in usdview as they did in Maya.  In future releases, we would like to turn the three usdchecker warnings into errors.

**Additional context**
Pixar _may_ be able to help here, and hopes to at least provide a "fixup script" that will correct these problems in most cases of existing assets, but these patterns are unfortunately getting copied by others, so it would be great to get this addressed as soon as possible, and our resources are limited.

Thanks much and apologies for the trouble.

kxl-adsk:
We are doing a little bit of research before proposing a direction we would like to take and validating it with the community.

JGamache-autodesk:
For points 1 and 2, the proper values can be set in the attribute editor of the file node that is directly connected to the normal input of the UsdPreviewSurface node:
 - "Color Space" in the "File Attributes" section should be set to "Utility -> Raw"
 - "Color Gain" and "Color Offset" in the "Color Balance" section should be set to [2,2,2] and [-1,-1,-1] respectively
[...]

PR #1915, opened 2021-12-16
Unnest UV reader and optimize

JGamache-autodesk:
First item in to fix in Issue #1834

- The UsdUVTexture and UsdTransform2d are no longer nested under the
  file node
- Shared place2dTexture nodes result in shared UsdUVTexture
- If a file node does not have a place2DTexture node, then is shares a
  common UsdUVTexture

More to follow.

### DIFF
--- a/lib/usd/translators/shading/usdFileTextureWriter.cpp
+++ b/lib/usd/translators/shading/usdFileTextureWriter.cpp
@@ -67,7 +67,8 @@ public:
         const TfToken&          mayaAttrName,
         const SdfValueTypeName& typeName) override;
 
-    void WriteTransform2dNode(const UsdTimeCode& usdTime, const UsdShadeShader& texShaderSchema);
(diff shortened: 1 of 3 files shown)


lib/usd/translators/shading/usdFileTextureWriter.cpp
```diff
@@ -77,7 +78,7 @@ TF_DEFINE_PRIVATE_TOKENS(
     _tokens,
 
     // UsdPrimvarReader_float2 Prim Name
-    ((PrimvarReaderShaderName, "TexCoordReader"))
+    ((PrimvarReaderShaderName, "shared_TexCoordReader"))
```
boudrey, line 81: Could changing this name affects loading of existing files?

JGamache-autodesk: No. The name of the place2dTexture node is not restored when importing (and it was not previously written, as you can see).

boudrey (approved):

PR #1918, opened 2021-12-16
MAYA-114979 - Validate normal maps for correctness

JGamache-autodesk:
Also check monochromatic inputs.

Connecting a file texture node interactively or by script will
automatically set the color attributes on the file node according to the
destination of the connection.

@dgovil : Added an env var for automatic fixups on load.

Second PR for issue #1834

### DIFF
--- a/lib/usd/pxrUsdPreviewSurface/usdPreviewSurface.cpp
+++ b/lib/usd/pxrUsdPreviewSurface/usdPreviewSurface.cpp
@@ -15,12 +15,14 @@
 //
 #include "usdPreviewSurface.h"
 
+#include <pxr/base/tf/envSetting.h>
 #include <pxr/base/tf/staticTokens.h>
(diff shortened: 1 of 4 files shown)


dgovil:
Thanks for tagging me to keep me in the loop. 

I like the setup you've created for this. That env var will be very useful.

boudrey (approved):
LGTM

dgovil:
@JGamache-autodesk One point that had come up internally is that perhaps this should respect the `ignoreColorSpaceFileRules` setting in Maya to allow for manual control in those scenarios.

We can put in a PR if you like, but just wanted to get your thoughts first?

JGamache-autodesk:
@dgovil setting `ignoreColorSpaceFileRules` makes a lot of sense, otherwise a change at the global level will flip the settings. If you have a PR ready please post it. Thanks!

dgovil:
Cool, I'll prepare a PR

PR #1936, opened 2021-12-21
Medley of normals and a few other fixes

fabal:
Here are another batch of changes in AL's plugin.
The changes made around normals were aimed at restoring some coherence between the type of mesh (poly vs subdiv) and normals handling. Prior to this there was zero coupling between the which means that exporting subdiv with custom normals was perfectly valid.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Transform.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Transform.cpp
@@ -29,7 +29,6 @@
 #include <maya/MGlobal.h>
 #include <maya/MProfiler.h>
 #include <maya/MTime.h>
-#include <maya/MViewport2Renderer.h>
 
 namespace {
 const int _transformProfilerCategory = MProfiler::addCategory(
@@ -494,7 +493,6 @@ MStatus Transform::preEvaluation(const MDGContext& context, const MEvaluationNod
         UsdTimeCode usdTime(theTime.as(MTime::uiUnit()));
 
         getTransMatrix()->updateToTime(usdTime);
-        MHWRender::MRenderer::setGeometryDrawDirty(thisMObject());
     }
     return MPxTransform::preEvaluation(context, evaluationNode);
 }
@@ -515,7 +513,6 @@ MStatus Transform::postEvaluation(
     UsdTimeCode usdTime(theTime.as(MTime::uiUnit()));
 
     getTransMatrix()->updateToTime(usdTime);
-    MHWRender::MRenderer::setGeometryDrawDirty(thisMObject());
 
     return MPxTransform::postEvaluation(context, evaluationNode, evalType);
 }
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Transform.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Transform.cpp
@@ -25,9 +25,11 @@
 #include <pxr/usd/usdGeom/scope.h>
 
 #include <maya/MBoundingBox.h>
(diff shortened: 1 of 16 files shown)


fabal:
The error comes from `AL_USDTransaction` tests in which the python module can't be loaded. It is tricky to debug for me as our build/test setup is different here at Animal and I can't reproduce the error. I did change AL_USDTransaction cmake's config in this PR so there's obviously an issue with the public build. Just let me know what is missing.

Note that I'll be away until Jan 4th.

fabal:
Happy new year! Following up on that one, would you have any insight which would help me solve the remaining error?

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Transform.cpp
```diff
@@ -454,12 +461,65 @@ MStatus Transform::validateAndSetValue(
...
+        }
+
+        MTime time;
+        context.getTime(time);
+        MTime theTime = (time - inputTimeValue(dataBlock, m_timeOffset))
+            * inputDoubleValue(dataBlock, m_timeScalar);
+        UsdTimeCode usdTime(theTime.as(MTime::uiUnit()));
+
+        getTransMatrix()->updateToTime(usdTime);
+        MHWRender::MRenderer::setGeometryDrawDirty(thisMObject());
```
kxl-adsk, line 497: Notifying renderers about changes in a non-normal context is not allowed. Why this code is necessary?

plugin/al/usdtransaction/CMakeLists.txt
```diff
@@ -1,17 +1,8 @@
...
-  ${CMAKE_CURRENT_SOURCE_DIR}/AL/usd/__init__.py
-  ${CMAKE_CURRENT_SOURCE_DIR}/AL/usd/transaction/__init__.py
-)
-
 # copy to build location for module tests
-foreach(INPUT_FILE ${PY_INIT_FILES})
-  string(REPLACE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR} OUTPUT_FILE ${INPUT_FILE})
-  execute_process(COMMAND "${CMAKE_COMMAND}" -E copy "${INPUT_FILE}" "${OUTPUT_FILE}")
-endforeach()
+file(COPY AL DESTINATION ${CMAKE_CURRENT_BINARY_DIR} FILES_MATCHING PATTERN "*.py")
```
kxl-adsk, line 5: Why does this need to be changed? What are you fixing here?

fabal: It's a clean-up, improving readability.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/Transform.cpp
fabal: The context of this change was an issue we had with evaluation caching, it dates back to late 2020.
Let me dig into this further.

fabal: I can't answer why it was written that way but I can confirm evaluation caching works properly without it (in fact, it's not needed in post evaluation either).

fabal: Let me update this internally first, then I'll apply the changes here.
In the meantime, any help to solve the remaining failing test would be appreciated.

fabal: @kxl-adsk , code has been updated.

kxl-adsk (approved):

PR #1937, opened 2021-12-21
MAYA-114979 MAYA-114731 Fix correctness issues

### LINKED ISSUE #1834: [MAYA-114979] UsdPreviewSurface export has several correctness problems

JGamache-autodesk:
This fixes #1834 for the last correctness problems mentioned

- Use a separate varnameStr for MaterialX since the type is not compatible with the varname used in UsdPreviewSurface
- Shared place2dTexture nodes are now shared on MaterialX export
- NodeGraph is used instead of Scope to isolate shaders by type in multi-material export
- NodeGraph boundaries are not skipped anymore by shader connections
- Cleaned-up code producing empty NodeGraphs

### DIFF
--- a/lib/usd/translators/shading/mtlxFileTextureWriter.cpp
+++ b/lib/usd/translators/shading/mtlxFileTextureWriter.cpp
@@ -150,7 +150,8 @@ MtlxUsd_FileWriter::MtlxUsd_FileWriter(
     }
 
     // Now create a geompropvalue reader that the image shader will use.
-    SdfPath primvarReaderPath = _GetPlace2DTexturePath(depNodeFn);
+    SdfPath        primvarReaderPath = _GetPlace2DTexturePath(depNodeFn);
+    UsdShadeOutput primvarReaderOutput;
 
     if (!GetUsdStage()->GetPrimAtPath(primvarReaderPath)) {
         UsdShadeShader primvarReaderSchema
@@ -188,29 +189,20 @@ MtlxUsd_FileWriter::MtlxUsd_FileWriter(
             varnameInput.Set(UsdUtilsGetPrimaryUVSetName());
         }
 
-        UsdShadeOutput primvarReaderOutput
+        primvarReaderOutput
             = primvarReaderSchema.CreateOutput(TrMtlxTokens->out, SdfValueTypeNames->Float2);
-
-        // TODO: Handle UV SRT with a ND_place2d_vector2 node. Make sure the name derives from the
-        //       place2dTexture node if there is one (see usdFileTextureWriter for details)
-
-        // Connect the output of the primvar reader to the texture coordinate
-        // input of the UV texture.
-        texSchema.CreateInput(TrMtlxTokens->texcoord, SdfValueTypeNames->Float2)
-            .ConnectToSource(primvarReaderOutput);
     } else {
         // Re-using an existing primvar reader:
         UsdShadeShader primvarReaderShaderSchema(GetUsdStage()->GetPrimAtPath(primvarReaderPath));
-        UsdShadeOutput primvarReaderOutput = primvarReaderShaderSchema.GetOutput(TrMtlxTokens->out);
-
-        // TODO: Handle UV SRT with a ND_place2d_vector2 node. Make sure the name derives from the
-        //       place2dTexture node if there is one (see usdFileTextureWriter for details)
-
-        // Connect the output of the primvar reader to the texture coordinate
-        // input of the UV texture.
-        texSchema.CreateInput(TrMtlxTokens->texcoord, SdfValueTypeNames->Float2)
-            .ConnectToSource(primvarReaderOutput);
+        primvarReaderOutput = primvarReaderShaderSchema.GetOutput(TrMtlxTokens->out);
     }
+    // TODO: Handle UV SRT with a ND_place2d_vector2 node. Make sure the name derives from the
+    //       place2dTexture node if there is one (see usdFileTextureWriter for details)
+
+    // Connect the output of the primvar reader to the texture coordinate
+    // input of the UV texture.
+    texSchema.CreateInput(TrMtlxTokens->texcoord, SdfValueTypeNames->Float2)
+        .ConnectToSource(primvarReaderOutput);
 }
 
 /* virtual */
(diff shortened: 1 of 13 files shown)


lib/usd/translators/shading/mtlxPreviewSurfaceWriter.cpp
```diff
@@ -135,6 +129,15 @@ MtlxUsd_PreviewSurfaceWriter::MtlxUsd_PreviewSurfaceWriter(
 
         // All connections go directly to the node graph:
         if (attrPlug.isConnected()) {
+            if (!nodegraphSchema) {
```
JGamache-autodesk, line 132: Just-in-time creation to reduce number of empty NodeGraph generated.

lib/usd/translators/shading/mtlxFileTextureWriter.cpp
```diff
@@ -148,45 +150,67 @@ MtlxUsd_FileWriter::MtlxUsd_FileWriter(
...
-    // easy specialization based on UV mappings to geometries:
-    SdfPath          materialPath = GetUsdPath().GetParentPath();
-    UsdShadeMaterial materialSchema(GetUsdStage()->GetPrimAtPath(materialPath));
-    while (!materialSchema && !materialPath.IsEmpty()) {
-        materialPath = materialPath.GetParentPath();
-        materialSchema = UsdShadeMaterial(GetUsdStage()->GetPrimAtPath(materialPath));
-    }
+    SdfPath primvarReaderPath = _GetPlace2DTexturePath(depNodeFn);
+
+    if (!GetUsdStage()->GetPrimAtPath(primvarReaderPath)) {
```
JGamache-autodesk, line 155: Mostly indentation since we now detect shared place2dTexture nodes. Similar to what was done in usdFileTextureWriter.cpp in the first PR of the series.

lib/usd/translators/shading/mtlxFileTextureWriter.cpp
```diff
@@ -148,45 +150,67 @@ MtlxUsd_FileWriter::MtlxUsd_FileWriter(
...
-    } else {
-        varnameInput.Set(UsdUtilsGetPrimaryUVSetName());
-    }
+            TfStringPrintf("%s:%s", depNodeFn.name().asChar(), TrMtlxTokens->varnameStr.GetText()));
+
+        // We expose the primvar reader varnameStr attribute to the material to allow
+        // easy specialization based on UV mappings to geometries:
+        UsdPrim          materialPrim = primvarReaderSchema.GetPrim().GetParent();
+        UsdShadeMaterial materialSchema(materialPrim);
+        while (!materialSchema && materialPrim) {
```
JGamache-autodesk, line 169: Here we make sure to properly connect the NodeGraph.

lib/usd/translators/shading/usdFileTextureWriter.cpp
```diff
@@ -137,18 +137,27 @@ PxrUsdTranslators_FileTextureWriter::PxrUsdTranslators_FileTextureWriter(
...
         // We expose the primvar reader varname attribute to the material to allow
         // easy specialization based on UV mappings to geometries:
-        SdfPath          materialPath = GetUsdPath().GetParentPath();
-        UsdShadeMaterial materialSchema(GetUsdStage()->GetPrimAtPath(materialPath));
-        while (!materialSchema && !materialPath.IsEmpty()) {
-            materialPath = materialPath.GetParentPath();
-            materialSchema = UsdShadeMaterial(GetUsdStage()->GetPrimAtPath(materialPath));
+        UsdPrim          materialPrim = primvarReaderShaderSchema.GetPrim().GetParent();
+        UsdShadeMaterial materialSchema(materialPrim);
+        while (!materialSchema && materialPrim) {
```
JGamache-autodesk, line 147: Connect all intermediate NodeGraphs.

lib/usd/translators/shading/mtlxFileTextureWriter.cpp
```diff
@@ -148,45 +150,67 @@ MtlxUsd_FileWriter::MtlxUsd_FileWriter(
...
-    // input of the UV texture.
-    texSchema.CreateInput(TrMtlxTokens->texcoord, SdfValueTypeNames->Float2)
-        .ConnectToSource(primvarReaderOutput);
+        // TODO: Handle UV SRT with a ND_place2d_vector2 node. Make sure the name derives from the
+        //       place2dTexture node if there is one (see usdFileTextureWriter for details)
+
+        // Connect the output of the primvar reader to the texture coordinate
+        // input of the UV texture.
+        texSchema.CreateInput(TrMtlxTokens->texcoord, SdfValueTypeNames->Float2)
+            .ConnectToSource(primvarReaderOutput);
```
boudrey, line 212: Line 206-212 could be taken out of the if/else block instead of having a duplication of code.

boudrey (approved):

spiffmon:
Fantastic - thank you so much!

PR #1940, opened 2021-12-22
MAYA-114729 Use correct default for shading mode

JGamache-autodesk:
Especially on a clean set of user prefs. The correct default is *nothing* and this allows properly merging the list of shading modes later at import time if none was specified.

### DIFF
--- a/lib/mayaUsd/fileio/jobs/jobArgs.cpp
+++ b/lib/mayaUsd/fileio/jobs/jobArgs.cpp
@@ -936,9 +936,6 @@ const VtDictionary& UsdMayaJobImportArgs::GetDefaultDictionary()
             = std::vector<VtValue>({ VtValue(SdfFieldKeys->Hidden.GetString()),
                                      VtValue(SdfFieldKeys->Instanceable.GetString()),
                                      VtValue(SdfFieldKeys->Kind.GetString()) });
-        d[UsdMayaJobImportArgsTokens->shadingMode] = std::vector<VtValue> { VtValue(
-            std::vector<VtValue> { VtValue(UsdMayaShadingModeTokens->useRegistry.GetString()),
-                                   VtValue(UsdImagingTokens->UsdPreviewSurface.GetString()) }) };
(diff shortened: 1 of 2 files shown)


plugin/adsk/plugin/importTranslator.cpp
```diff
@@ -158,6 +158,10 @@ const std::string& UsdMayaImportTranslator::GetDefaultOptions()
         std::vector<std::string> entries;
         for (const std::pair<std::string, VtValue> keyValue :
              UsdMayaJobImportArgs::GetDefaultDictionary()) {
+            if (keyValue.first == UsdMayaJobImportArgsTokens->shadingMode.GetString()) {
+                // Shading mode varies with loaded plug-ins. Remove from defaults.
+                continue;
+            }
```
JGamache-autodesk, line 164: This is the value stored in Maya when the mayaUsdPlugin initializes. At this point in time, we have not yet finished loading all dependent USD plugins and have a very limited list of conversions to choose from.

lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -956,6 +953,40 @@ const VtDictionary& UsdMayaJobImportArgs::GetDefaultDictionary()
...
+        }
+    }
+    shadingModes.emplace_back(
+        std::vector<VtValue> { VtValue(UsdMayaShadingModeTokens->useRegistry.GetString()),
+                               VtValue(UsdImagingTokens->UsdPreviewSurface.GetString()) });
+    shadingModes.emplace_back(
+        std::vector<VtValue> { VtValue(UsdMayaShadingModeTokens->displayColor.GetString()),
+                               VtValue(UsdMayaShadingModeTokens->none.GetString()) });
+
+    d[UsdMayaJobImportArgsTokens->shadingMode] = { VtValue(shadingModes) };
```
JGamache-autodesk, line 988: This value will be used if the userArgs passed-in on import do not contain a shadingMode, which is what we want since the Maya-stored default does not have an opinion on this.

boudrey (approved):

PR #1945, opened 2022-01-04
MAYA-120860 Fix TestALUSDMayaPython_LayerManager random fails

boudrey:
Set TEMP and TMP environment variables for each tests registered with mayaUsd_add_test.

### DIFF
--- a/cmake/test.cmake
+++ b/cmake/test.cmake
@@ -330,7 +330,7 @@ finally:
             "${pathvar}=${MAYAUSD_VARNAME_${pathvar}}")
     endforeach()
     
-    # Set a temporary folder path for the MAYA_APP_DIR in which the
+    # Set a temporary folder path for the TMP,TEMP and MAYA_APP_DIR in which the
     # maya profile will be created.
     # Note: replace bad chars in test_name with _.
     string(REGEX REPLACE "[:<>\|]" "_" SANITIZED_TEST_NAME ${test_name})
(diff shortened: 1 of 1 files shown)


cmake/test.cmake
```diff
@@ -356,6 +356,15 @@ finally:
...
         "MAYA_DISABLE_CER=1")
 
+    # Set a temporary folder path for the test
+    # Note: replace bad chars in test_name with _.
+    string(REGEX REPLACE "[:<>\|]" "_" SANITIZED_TEST_NAME ${test_name})
+    set(TEST_TEMP_DIR "${CMAKE_BINARY_DIR}/test/Temporary/${SANITIZED_TEST_NAME}")
+    set_property(TEST "${test_name}" APPEND PROPERTY ENVIRONMENT
+        "TMP=${TEST_TEMP_DIR}"
+        "TEMP=${TEST_TEMP_DIR}")
+    file(MAKE_DIRECTORY ${TEST_TEMP_DIR})
```
seando-adsk, line 366: I would suggest just adding these two lines (with your comment) to the env var above on line 340.
```
        "TMP=${MAYA_APP_TEMP_DIR}"
        "TEMP=${MAYA_APP_TEMP_DIR}")
```

boudrey: I will, nice catch.

seando-adsk (approved):

PR #1973, opened 2022-01-11
Tremblp/maya 114982/maya reference 00

ppt-adsk:


lib/mayaUsd/fileio/primUpdater.h
```diff
@@ -55,10 +55,19 @@ class UsdMayaPrimUpdater
...
+        Failed,
+        Continue,
+        Prune
+    };
+
     // Copy the pushed prim from the temporary srcLayer where it has been
     // exported by push into the destination dstLayer which is in the scene.
     MAYAUSD_CORE_PUBLIC
-    virtual bool pushCopySpecs(
+    virtual PushCopySpecs pushCopySpecs(
```
ppt-adsk, line 70: pushCopySpecs() must indicate behavior beyond simple success and failure: pruning traversal is also success.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -575,33 +581,47 @@ SdfPath getDstSdfPath(const Ufe::Path& ufePulledPath, const SdfPath& srcSdfPath,
...
     const SdfLayerRefPtr&            srcLayer,
-    const SdfPath&                   primSpecPath,
+    const SdfPath&                   srcPath,
+    const SdfLayerRefPtr&            dstLayer,
+    const SdfPath&                   dstPath,
     const UsdMayaPrimUpdaterContext& context)
 {
-    // Get the primSpec from the src layer.
-    auto primSpec = srcLayer->GetPrimAtPath(primSpecPath);
+    // The root of the pulled hierarchy is crucial for determining push
```
ppt-adsk, line 591: Having a stable updater that performs both the pull and the push is crucial to success.  Ensure we get the same updater by using the original pulled path to create the updater for both pull and push.

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
```diff
@@ -0,0 +1,640 @@
+// Verbatim copy of fileOptions.mel:pushOptionsUITemplate(), which is not
```
ppt-adsk, line 1: Main file in the review.  Unfortunately mostly copy-pasted and adapted from Maya code, which was not designed to be extensible to use cases like Maya reference prim creation.

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.py
```diff
@@ -0,0 +1,26 @@
+import maya.api.OpenMaya as om
+from pxr import Sdf
+import mayaUsd
+import ufe
+
+# Adapted from testMayaUsdSchemasMayaReference.py.
+def createMayaReferencePrim(ufePathStr, mayaReferencePath, mayaNamespace):
```
ppt-adsk, line 7: Once all the UI dialog is done, perform the Maya reference prim creation.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -375,12 +377,12 @@ global proc string ClearAllUSDReferencesConfirm()
 ClearAllUSDReferencesConfirm();
 )";
 
-class AddReferenceUndoableCommand : public Ufe::UndoableCommand
+class AddUsdReferenceUndoableCommand : public Ufe::UndoableCommand
```
ppt-adsk, line 380: Renamed for clarity: existing code creates a USD reference, not a Maya reference prim.

lib/usd/translators/mayaReferenceUpdater.cpp
```diff
@@ -45,7 +45,11 @@ PXRUSDMAYA_REGISTER_UPDATER(
     reference,
     PxrUsdTranslators_MayaReferenceUpdater,
     (UsdMayaPrimUpdater::Supports::Push | UsdMayaPrimUpdater::Supports::Clear
-     | UsdMayaPrimUpdater::Supports::AutoPull));
+     // Disable AutoPull pending more granular support for it: MayaReference
```
ppt-adsk, line 48: Change is out of scope for this pull request: auto pull is not based on type, it's based on an instance of a type, i.e. because we need to ask the instance whether it can auto pull or not (and an instance without a file path should not be auto pulled).  We will need a virtual function to customize this.

modules/mayaUSD.mod.template
```diff
@@ -13,7 +13,6 @@ PXR_MTLX_STDLIB_SEARCH_PATHS+=${USD_INSTALL_LOCATION}/libraries
 icons: lib/icons
 plug-ins: 
 presets: 
-scripts:
```
ppt-adsk, line 16: scripts folder is no longer empty.

lib/mayaUsd/fileio/primUpdater.h
```diff
@@ -55,10 +55,19 @@ class UsdMayaPrimUpdater
         All = Push | Pull | Clear
     };
 
+    // pushCopySpecs result code.  Prune means success, but no further
+    // traversal should take place.
+    enum class PushCopySpecs
```
pierrebai-adsk, line 60: It feels weird to have a enum have the same name as a function. Maybe PushCopyResult would have been a better name? In any case, this is not a widely used type, so it is not very important.

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.py
```diff
@@ -0,0 +1,26 @@
...
+    '''Create a Maya reference prim parented to the argument path.'''
+
+    # Append reference name to parent Sdf path.
+    ufePath = ufe.PathString.path(ufePathStr)
+    parentPath = ufePath.segments[1]
```
pierrebai-adsk, line 12: Could you explain the hard-coded [1] here? If I understand it, the code assumes the second segment is a USD path and adds '/MayaReference'  to it? I found the comment is not entirely clear. I think the comment would be more helpful if it said something like:

Extract the USD path segment from the UFE path and adds the MayaReference under it

Or something similar.

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.py
```diff
@@ -0,0 +1,26 @@
...
+    primPath = Sdf.AssetPath(str(parentPath)+'/MayaReference')
+
+    stage = mayaUsd.ufe.getStage(ufePathStr)
+    prim = stage.DefinePrim(primPath.path, 'MayaReference')
+    if not prim.IsValid():
+        om.MGlobal.displayError("Could not create MayaReference prim under %s" % ufePathStr)
+        return
+
+    mayaReferenceAttr = prim.CreateAttribute('mayaReference', Sdf.ValueTypeNames.Asset)
+    mayaReferenceAttr.Set(mayaReferencePath)
```
pierrebai-adsk, line 22: We got a MayaReference path, a MayaReference custom prim type and mayaReference attribute. It's a bit confusing :) I suppose it cannot be avoided.

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
pierrebai-adsk: I'm honestly not very knowledgeable about MEL and always struggle to write it, so I'm not going to pretend to review it. This being mostly UI code, I'll approve it blindly.

pierrebai-adsk (changes requested):
Just waiting for your feedback on my suggestions.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -575,33 +581,47 @@ SdfPath getDstSdfPath(const Ufe::Path& ufePulledPath, const SdfPath& srcSdfPath,
...
+    // behavior.  When pulling, we may have created a Maya pull hierarchy root
+    // node whose type does not map to the same prim updater as the original
+    // USD prim, i.e. multiple USD prim types can map to the same pulled Maya
+    // node type (e.g. transform, which is the fallback Maya node type for many
+    // USD prim types).  Therefore, if we're at the root of the src hierarchy,
+    // use the prim at the pulled path to create the prim updater.
+    const bool usePulledPrim = (srcPath.GetPathElementCount() == 1);
```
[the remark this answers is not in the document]
ppt-adsk: Correct.  I'll adjust the comment to mention the temporary layer.

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.py
ppt-adsk: Those names were chosen by Animal Logic, I won't change them.
https://github.com/Autodesk/maya-usd/blob/600c31abcd68ef3fe0047b28fb5f33f6edd0872f/lib/usd/schemas/schema.usda.in#L22

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.py
ppt-adsk: O.K.

pierrebai-adsk (approved):

PR #1974, opened 2022-01-11
MAYA-114898 add merge options for missing attributes.

pierrebai-adsk:
Add MergePrimsOptions structure as options to control the merging of prims.

- The options have a constructor taking a VtDictionary to allow customizing them.
- Made the default for the 'ignore upper layers' to be off.
- Add merge missing options for each different fields.
- For example: propertie,s relationships, variants, etc.
- The options control the creation or deletion of entries in these fields.
- For example it control if an attribute missing from the source gets deleted in the destination.
- Other example: if an attributes missing in the destination gets created.

Modify code to allow passing the options:

- Add a VtDictionary userArgs to the PrimUpdaterManager in its editAsMaya, mergeToUsd and duplicate functions.
- Make sure the merge prims options are built with the default values.
- Use these user args in the prim updater manager code.

Add a new export command option -exportMaterials / -mat to control if materials are exported.

- Defaults to true.
- Added to the command and exporter job args.
- Implemented by make the shading mode exporter do nothing when off.


pierrebai-adsk:
@JGamache-autodesk : mostly for the material export options and code.

lib/mayaUsd/fileio/shading/shadingModeExporter.cpp
```diff
@@ -84,6 +84,9 @@ void UsdMayaShadingModeExporter::DoExport(
     const UsdMayaJobExportArgs& exportArgs = writeJobContext.GetArgs();
     const UsdStageRefPtr&       stage = writeJobContext.GetUsdStage();
 
+    if (!exportArgs.exportMaterials)
```
pierrebai-adsk, line 87: The material export is disabled here,

pierrebai-adsk: After talking with Jerry, I'll remove the material option as it is redundant. I'll re-open the review when done.

PR #1976, opened 2022-01-12
MAYA-114898 add merge options for missing attributes.

pierrebai-adsk:
Add MergePrimsOptions structure as options to control the merging of prims.

- The options have a constructor taking a VtDictionary to allow customizing them.
- Made the default for the 'ignore upper layers' to be off.
- Add merge missing options for each different fields.
- For example: properties, relationships, variants, etc.
- The options control the creation or deletion of entries in these fields.
- For example it control if an attribute missing from the source gets deleted in the destination.
- Other example: if an attributes missing in the destination gets created.

Modify code to allow passing the options:

- Add a VtDictionary userArgs to the PrimUpdaterManager in its editAsMaya, mergeToUsd and duplicate functions.
- Make sure the merge prims options are built with the default values.
- Use these user arguments in the prim updater manager code.

### DIFF
--- a/lib/usd/utils/MergePrims.cpp
+++ b/lib/usd/utils/MergePrims.cpp
@@ -36,16 +36,16 @@ namespace {
 // Data used for merging passed to all helper functions.
 struct MergeContext
 {
-    const MergeVerbosity  verbosity;
-    const bool            mergeChildren;
-    const UsdStageRefPtr& srcStage;
-    const SdfPath&        srcRootPath;
-    const UsdStageRefPtr& dstStage;
-    const SdfPath&        dstRootPath;
+    const MergePrimsOptions& options;
+    const UsdStageRefPtr&    srcStage;
+    const SdfPath&           srcRootPath;
+    const UsdStageRefPtr&    dstStage;
+    const SdfPath&           dstRootPath;
 };
 
 //----------------------------------------------------------------------------------------------------------------------
-/// Description of a merge location: layer, path and if the field already exists at that location.
+/// Description of a merge location: layer, path, field and if the field already exists at that
+/// location.
 struct MergeLocation
 {
     const SdfLayerHandle& layer;
@@ -63,7 +63,7 @@ void printAboutField(
     const char*          message,
     const char*          message2 = nullptr)
 {
-    if (!contains(printVerbosity, ctx.verbosity))
+    if (!contains(printVerbosity, ctx.options.verbosity))
         return;
 
     TF_STATUS(
@@ -94,7 +94,7 @@ void printAboutChildren(
     const char*                     message,
     const std::vector<std::string>& childrenNames)
 {
-    if (!contains(MergeVerbosity::Children, ctx.verbosity))
+    if (!contains(MergeVerbosity::Children, ctx.options.verbosity))
         return;
 
     const std::string allNames = TfStringJoin(childrenNames);
@@ -264,6 +264,7 @@ bool shouldMergeValue(
 template <class ChildPolicy>
 bool filterTypedChildren(
     const MergeContext&  ctx,
+    const MergeMissing   missingHandling,
     const MergeLocation& src,
     const MergeLocation& dst,
     VtValue&             srcChildrenValue,
@@ -294,19 +295,54 @@ bool filterTypedChildren(
     dstFilteredChildren.reserve(dstChildren.size());
 
     for (size_t i = 0; i < srcChildren.size(); ++i) {
-        if (srcChildren[i].IsEmpty() || dstChildren[i].IsEmpty()) {
-            printAboutFailure(ctx, src, "empty child. ");
+        const auto& srcChild = srcChildren[i];
+        const auto& dstChild = dstChildren[i];
+
+        // Special handling when the source or destination are empty.
+        if (srcChild.IsEmpty() || dstChild.IsEmpty()) {
+
+            // Both empty: should not happen, but let's be safe.
+            if (srcChild.IsEmpty() && dstChild.IsEmpty()) {
+                printAboutFailure(ctx, srcChild.IsEmpty() ? dst : src, "empty child. ");
+                continue;
+            }
+
+            // Check if we are  allowing missing attributes to be preserved.
+            if (srcChild.IsEmpty() && !contains(missingHandling, MergeMissing::Preserve)) {
+                printAboutFailure(ctx, dst, "not preserving child missing in source. ");
+                continue;
+            }
+
+            // Check if we are not allowing creation of attributes.
+            if (dstChild.IsEmpty() && !contains(missingHandling, MergeMissing::Create)) {
+                printAboutFailure(ctx, src, "not creating child missing in destination. ");
+                continue;
+            }
+
+            const SdfPath childPath = srcChild.IsEmpty()
+                ? ChildPolicy::GetChildPath(dst.path, dstChild)
+                : ChildPolicy::GetChildPath(src.path, srcChild);
+
+            const MergeLocation childLoc
+                = { srcChild.IsEmpty() ? dst.layer : src.layer, childPath, TfToken(), true };
+            printAboutField(ctx, childLoc, MergeVerbosity::Child, "empty child. ");
+
+            if (contains(ctx.options.verbosity, MergeVerbosity::Children))
+                childrenNames.emplace_back(childPath.GetName());
+
+            srcFilteredChildren.emplace_back(srcChild);
+            dstFilteredChildren.emplace_back(dstChild);
             continue;
         }
 
-        const SdfPath srcChildPath = ChildPolicy::GetChildPath(src.path, srcChildren[i]);
-        const SdfPath dstChildPath = ChildPolicy::GetChildPath(dst.path, dstChildren[i]);
+        const SdfPath srcChildPath = ChildPolicy::GetChildPath(src.path, srcChild);
+        const SdfPath dstChildPath = ChildPolicy::GetChildPath(dst.path, dstChild);
 
         // Note: don't use location's field, since we're in a child path, the children field is
         // irrelevant. We will assume the child exists, but we will actually verify it just
         // below with a call to HasSpec().
-        const MergeLocation childSrc = { src.layer, srcChildPath, TfToken(), true };
-        const MergeLocation childDst = { dst.layer, dstChildPath, TfToken(), true };
+        const MergeLocation childSrcLoc = { src.layer, srcChildPath, TfToken(), true };
+        const MergeLocation childDstLoc = { dst.layer, dstChildPath, TfToken(), true };
 
         // Note: we cannot drop a children that already has an opinion at the
         // destination, otherwise SdfCopySpec() will delete that opinion!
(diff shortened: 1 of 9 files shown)


pierrebai-adsk:
PF failure is a single unrelated test failure in the 2022 Windows build in testVP2RenderDelegatePerInstanceInheritedData.

lib/usd/utils/MergePrimsOptions.h
```diff
@@ -0,0 +1,190 @@
...
+
+inline bool contains(MergeMissing a, MergeMissing b) { return (a & b) != MergeMissing::None; }
+
+//----------------------------------------------------------------------------------------------------------------------
+// Options to control prims merging.
+
+struct MergePrimsOptions
+{
+    // How much logging is done during the merge.
+    MergeVerbosity verbosity = MergeVerbosity::Default;
```
ppt-adsk, line 96: Nit-pick: we should be using `{}` for initialization.  See
https://github.com/Autodesk/maya-usd/blob/dev/doc/codingGuidelines.md#modern-c

lib/usd/utils/MergePrimsOptions.h
```diff
@@ -0,0 +1,190 @@
...
+
+    // if true, merges children too, otherwise merge only the given prim.
+    bool mergeChildren = false;
```
ppt-adsk, line 99: Elsewhere, for structs initialized from dictionaries, we're making all the data members const, e.g.
https://github.com/Autodesk/maya-usd/blob/2bf98c7f2fae04d6237a43c8d060e6c065e589e1/lib/mayaUsd/fileio/jobs/jobArgs.h#L150
This certainly reinforces the notion that these objects are pure values that cannot be mutated.  Now I see that you have a public default constructor, which allows you to later on to initialize your choice of the non-const data members, so having them non-const is more flexible for initialization.  If you choose to leave them non-const, would be good to put in a single-line comment explaining why.

lib/usd/utils/MergePrims.h
```diff
@@ -71,7 +43,6 @@ bool mergePrims(
     const PXR_NS::UsdStageRefPtr& dstStage,
     const PXR_NS::SdfLayerRefPtr& dstLayer,
     const PXR_NS::SdfPath&        dstPath,
-    bool                          mergeChildren = false,
-    MergeVerbosity                verbosity = MergeVerbosity::Default);
+    const MergePrimsOptions&      options);
```
ppt-adsk, line 46: Looks like at least one call site is using default MergePrimOptions, so you could consider a default value here.

lib/usd/utils/MergePrimsOptions.cpp
```diff
@@ -0,0 +1,191 @@
...
+    static std::once_flag once;
+    std::call_once(once, []() {
+        const auto _token = VtValue(TfToken());
+        const auto _tokenVector = VtValue(std::vector<VtValue>({ _token }));
+
+        d[UsdMayaMergeOptionsTokens->verbosity]
+            = VtValue(std::vector<VtValue>({ VtValue(UsdMayaMergeOptionsTokens->Default) }));
+        d[UsdMayaMergeOptionsTokens->mergeChildren] = false;
+        d[UsdMayaMergeOptionsTokens->ignoreUpperLayerOpinions] = false;
+        d[UsdMayaMergeOptionsTokens->propertiesHandling]
```
ppt-adsk, line 124: Gotta be a way to turn this into a loop...
std::vector<TfToken> mergeOptionTokens{UsdMayaMergeOptionsTokens->propertiesHandling, UsdMayaMergeOptionsTokens->primsHandling, ...};
auto defaultValue = VtValue(std::vector<VtValue>({VtValue(UsdMayaMergeOptionsToken->All)}));
for (const auto& token : mergeOptionTokens) { ... };

lib/usd/utils/MergePrimsOptions.cpp
```diff
@@ -0,0 +1,191 @@
...
+const VtDictionary& MergePrimsOptions::getDefaultDictionary()
+{
+    static VtDictionary   d;
+    static std::once_flag once;
+    std::call_once(once, []() {
+        const auto _token = VtValue(TfToken());
+        const auto _tokenVector = VtValue(std::vector<VtValue>({ _token }));
+
+        d[UsdMayaMergeOptionsTokens->verbosity]
+            = VtValue(std::vector<VtValue>({ VtValue(UsdMayaMergeOptionsTokens->Default) }));
```
ppt-adsk, line 121: I'm confused as to why this is a VtValue that's a vector of VtValue's.

lib/usd/utils/MergePrimsOptions.cpp
```diff
@@ -0,0 +1,191 @@
...
+        d[UsdMayaMergeOptionsTokens->mappersHandling]
+            = VtValue(std::vector<VtValue>({ VtValue(UsdMayaMergeOptionsTokens->All) }));
+        d[UsdMayaMergeOptionsTokens->mapperArgsHandling]
+            = VtValue(std::vector<VtValue>({ VtValue(UsdMayaMergeOptionsTokens->All) }));
+    });
+
+    return d;
+}
+
+MergePrimsOptions::MergePrimsOptions(const VtDictionary& options)
```
ppt-adsk, line 147: To be able to make the data members const and initialize them from the constructor initializer list, you'll probably have to define a private constructor with a different signature, and delegate the construction to it:

MergePrimsOptions::MergePrimsOptions(const VtDictionary& filledOptions, bool dummy) : verbosity(parseVerbosity(filledOptions, UsdMayaMergeOptionsTokens->verbosity)), mergeChildren(...)

lib/usd/utils/MergePrims.cpp
```diff
@@ -357,6 +393,72 @@ bool filterTypedChildren(
     return shouldCopy;
 }
 
+template <class ChildPolicy>
+void unionChildren(
+    MergeMissing missingHandling,
+    VtValue&     srcChildrenValue,
+    VtValue&     dstChildrenValue)
+{
+    // If not preserving missing attributes then we don't need to claculate
```
ppt-adsk, line 402: Typos: "claculate", "destintation".

lib/usd/utils/MergePrims.cpp
```diff
@@ -357,6 +393,72 @@ bool filterTypedChildren(
...
+    ChildrenVector srcChildren;
+    if (!srcChildrenValue.IsEmpty())
+        srcChildren = srcChildrenValue.UncheckedGet<ChildrenVector>();
+
+    ChildrenVector dstChildren;
+    if (!dstChildrenValue.IsEmpty())
+        dstChildren = dstChildrenValue.UncheckedGet<ChildrenVector>();
+
+    // Create sets for fast comparison and return if both sets are equal,
+    // meaninig the source and destination already have the same children.
```
ppt-adsk, line 427: "meaninig".

ppt-adsk (commented):
Looks pretty good to me, nice job!  Enough nit-picks and questions that I didn't approve outright.

lib/usd/utils/MergePrimsOptions.cpp
pierrebai-adsk: It's the list of verbosity value, which can be or'ed together. So it must support more than a single token.

lib/usd/utils/MergePrimsOptions.h
pierrebai-adsk: Do enum have a default value? I don't see how it would know that Default is the default value.

ppt-adsk: Just change to
MergeVerbosity verbosity{MergeVerbosity::Default};

lib/usd/utils/MergePrimsOptions.h
pierrebai-adsk: I expected the option to be kept around, but I can see if making them const breaks anything.

lib/usd/utils/MergePrims.h
pierrebai-adsk: I expect that to change shortly, as we add the UI to control the pull and push options, so it does not seems worthwhile.

lib/usd/utils/MergePrimsOptions.cpp
ppt-adsk: So it can be built incrementally by appending to the vector, and must be OR'ed at read time?  That seems nice and flexible, but a bit complicated.

lib/usd/utils/MergePrimsOptions.h
ppt-adsk: Be careful with this, you've changed test_MergePrims.cpp to use non-constness for initialization purposes.  Don't know if it's worth your time, if your investigation takes you more than 10 minutes I'd just comment and move on.

lib/usd/utils/MergePrimsOptions.cpp
pierrebai-adsk: Well, I did not see a better way to express that is accept multiple tokens. The alternative would be to have all the possible combinations as tokens, but that creates 2 ^ N different tokens, which are all hard to read.

lib/usd/utils/MergePrimsOptions.h
pierrebai-adsk: Ah, OK, I misunderstood the suggestion!

lib/usd/utils/MergePrimsOptions.cpp
ppt-adsk: If it's got to be expressed as tokens, then yes you're right.  Dumb question time: why isn't it a plain integer, rather than a bunch of tokens?

lib/usd/utils/MergePrimsOptions.h
pierrebai-adsk: True, for testing having to pass through a dict just to set variable would be awkward.

lib/usd/utils/MergePrimsOptions.cpp
pierrebai-adsk: Would be obscure for programmers. I also expected these to be exposed, either in Python on in MEL when doing UI for the options, having dict with integer would obscure the code.

lib/usd/utils/MergePrimsOptions.h
```diff
@@ -89,45 +89,49 @@ inline bool contains(MergeMissing a, MergeMissing b) { return (a & b) != MergeMi
...
+//
+// To simplify the constructors and initialization in the unit test,
+// the individual member variables are not declared const, but the
+// whole structure is passed const to functions receiving it.
 
 struct MergePrimsOptions
 {
     // How much logging is done during the merge.
-    MergeVerbosity verbosity = MergeVerbosity::Default;
+    MergeVerbosity verbosity = { MergeVerbosity::Default };
```
ppt-adsk, line 100: Should just be
MergeVerbosity verbosity{MergeVerbosity::Default};

ppt-adsk (changes requested):

ppt-adsk (changes requested):
Just that one final change in MergePrimsOptions.h.

ppt-adsk (approved):

PR #1977, opened 2022-01-12
MAYA-114898 fix unit test assertions related to pull/push

pierrebai-adsk:
When using the direct python API to edit-as-Maya and merge-to-USD, the calls have to be wrapped in a OpUndoItemList block to capture the undo info, otherwise it will linger on. This is not a problem, except when Maya exit, the order of cleanup makes Maya assert that some deleted nodes are still present.

Fix the tests by using a block.

Also do a best effort to clean up the undo information that may have been left behind due to not using a OpUndoInfoList block in Python. On scene reset the undo info is cleared, although currently the assertions are hit before the scene-reset callback is triggered, so we still get assertions.

### DIFF
--- a/lib/mayaUsd/undo/OpUndoItemList.cpp
+++ b/lib/mayaUsd/undo/OpUndoItemList.cpp
@@ -22,7 +22,7 @@
 namespace MAYAUSD_NS_DEF {
 
 //------------------------------------------------------------------------------
-// OpUndoItemList automatic celar when scene in reset.
+// Notifier to automatically clear OpUndoItemList when the scene is reset.
 //------------------------------------------------------------------------------
 
 namespace {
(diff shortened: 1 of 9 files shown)


lib/mayaUsd/undo/OpUndoItemList.cpp
```diff
@@ -15,8 +15,34 @@
...
 #include "OpUndoItemList.h"
 
+#include <mayaUsd/listeners/notice.h>
+
+#include <pxr/base/tf/weakBase.h>
+
 namespace MAYAUSD_NS_DEF {
 
+//------------------------------------------------------------------------------
+// OpUndoItemList automatic celar when scene in reset.
```
boudrey, line 25: Type celar -> clear

boudrey (approved):

boudrey (approved):

PR #1979, opened 2022-01-12
MAYA-121112 Maya USD error and no UI elements

boudrey:
Moving the script to the expected script folder.

### DIFF
--- a/lib/mayaUsd/resources/scripts/CMakeLists.txt
+++ b/lib/mayaUsd/resources/scripts/CMakeLists.txt
@@ -4,5 +4,5 @@ list(APPEND scripts_src
 )
 
 install(FILES ${scripts_src}
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/scripts
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/plugin/adsk/scripts
 )


lib/mayaUsd/resources/scripts/CMakeLists.txt
```diff
@@ -4,5 +4,5 @@ list(APPEND scripts_src
 )
 
 install(FILES ${scripts_src}
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/scripts
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/plugin/adsk/scripts
```
ppt-adsk, line 7: Why are library base resources placed into a plugin-specific directory (the Autodesk plugin, in this case)?  This doesn't sound right.  The Animal Logic plugin won't get these resources.

seando-adsk: Pierre is right, we want these files installed into the lib "scripts" folder. He has already modified the mayaUsd.mod file to correctly use this path. I suspect a problem with the setup of the user who found this.

boudrey: I had to do this change locally for my "_InteractiveMaya" test to run properly.  I will find another way of fixing it and will let you look with David, what is wrong with its setup.

seando-adsk: If there is a test failing because of this then we need to add the lib scripts folder to the cmake test setup. Like its done [here](https://github.com/Autodesk/maya-usd/blob/dev/cmake/test.cmake#L230) for the plugin.

    list(APPEND MAYAUSD_VARNAME_PYTHONPATH
         "${CMAKE_INSTALL_PREFIX}/scripts")
    list(APPEND MAYAUSD_VARNAME_MAYA_SCRIPT_PATH
         "${CMAKE_INSTALL_PREFIX}/scripts")

PR #1980, opened 2022-01-13
Improved texture loading by running file reading asynchronously and only for active render purpose

csyshing:
This PR is an attempt to improve the user experience of loading USD scene with lots of textures file (https://github.com/Autodesk/maya-usd/issues/1624), the idea is to:
1. [Store necessary info](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-95070490bea9fe284d8a3c42c85ed209a5341007e11b85dac96e316dcec6e269R2483-R2484) ([TextureLoadingTask object](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-95070490bea9fe284d8a3c42c85ed209a5341007e11b85dac96e316dcec6e269R1215)) during sync-ing sprim but not do any actual reading, a [placeholder (grey texture)](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-95070490bea9fe284d8a3c42c85ed209a5341007e11b85dac96e316dcec6e269R2487-R2504) would be returned instead
2.  [Kick off reading textures](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-2cbdd2050dcfecd7df19538b8d42a716855a7c436caf309327c0c92258072468R938-R947) when sync-ing rprims: we can then figure out which materials are being used for particular rprims
3. For **used** materials, the actual reading will be run in [low priority via `MGlobal::executeTaskOnIdle()`](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-95070490bea9fe284d8a3c42c85ed209a5341007e11b85dac96e316dcec6e269R1239-R1246), when the task is finished, [inform Maya to refresh](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-95070490bea9fe284d8a3c42c85ed209a5341007e11b85dac96e316dcec6e269R2564) so that viewport can be updated; no matter the task finished or terminated, the temp `TextureLoadingTask object` will be deleted [in the same scope of idle callback](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-95070490bea9fe284d8a3c42c85ed209a5341007e11b85dac96e316dcec6e269R1244).
4. The temporary `TextureLoadingTask objects` for **unused** materials will be deleted in two scenarios:
  - At the destructor of [HdVP2Material](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-95070490bea9fe284d8a3c42c85ed209a5341007e11b85dac96e316dcec6e269R1301-R1305) for whatever reason;
  - Every time when [scene delegate is reset](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-e89b5be91b4c9930af19e1343239a832468afb52c0cfc841346c2e816c7899c7R529-R532) in VP2RenderDelegate, [`HdVP2Material::ClearPendingTasks()`](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-95070490bea9fe284d8a3c42c85ed209a5341007e11b85dac96e316dcec6e269R2564) will delete the temp objects if they are not used (not started)
5. There is an environment variable to toggle on/off the async loading: [MAYAUSD_DISABLE_ASYNC_TEXTURE_LOADING](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-95070490bea9fe284d8a3c42c85ed209a5341007e11b85dac96e316dcec6e269R104-R112), default is off

If this PR is useful (I hope so!), there are a few things I would like to discuss:
- `HdVP2RenderDelegate` is the bridge between VP2RenderDelegate and HdVP2Mesh(rprim) / HdVP2Material (sprim), I added an additional class variable, [_materialSprims](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-37107e5a67192eb54e4d8eaeae3dd27c57fdd29dd0af8c020e8a53cd305ae7a0R178), to keep tracking of created sprim, not sure if there is any better way to do so?
- Every time a texture is finished loading, a refresh command [will be triggered](https://github.com/Autodesk/maya-usd/compare/dev...AnimalLogic:csyshing/texture_loading_improvement#diff-95070490bea9fe284d8a3c42c85ed209a5341007e11b85dac96e316dcec6e269R2562-R2564), there might be other more efficient way to update viewport?

This is how it looks like, by delaying loading unused materials, the ALab can be loaded much faster by default (proxy), and once the "Render" is on, the textures will be updated one by one:

texture_loading_improvement


lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -2349,28 +2436,132 @@ void HdVP2Material::_UpdateShaderInstance(const HdMaterialNetwork& mat)
...
 /*! \brief  Acquires a texture for the given image path.
  */
-const HdVP2TextureInfo&
-HdVP2Material::_AcquireTexture(const std::string& path, const HdMaterialNode& node)
+const HdVP2TextureInfo& HdVP2Material::_AcquireTexture(
+    HdSceneDelegate*      sceneDelegate,
+    const std::string&    path,
+    const HdMaterialNode& node)
 {
     const auto it = _textureMap.find(path);
```
williamkrick, line 2444: This is a local cache of the textures used by this material. If the same texture is used my the material twice the async code will load it twice, and overwrite the first load in _textureMap. I'm not sure if that'll just hurt performance, or if overwriting the original texture info will cause a stability problem.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1199,6 +1212,72 @@ TfToken MayaDescriptorToToken(const MVertexBufferDescriptor& descriptor)
...
+
+private:
+    void _LoadOnIdle()
+    {
+        if (_terminated) {
+            return;
+        }
+        bool        isSRGB = false;
+        MFloatArray uvScaleOffset;
+        auto*       texture = _LoadTexture(_path, isSRGB, uvScaleOffset, _fallbackValue);
```
williamkrick, line 1266: _LoadTexture includes the call to see if the texture is already in the texture cache, so this idle loading approach is already working with texture caching and texture re-use.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -2349,28 +2436,132 @@ void HdVP2Material::_UpdateShaderInstance(const HdMaterialNetwork& mat)
...
+        fallbackValue = fallbackIt->second;
+    }
+
+    {
+        std::lock_guard<std::mutex> lock(_textureLoadingMutex);
+        _textureLoadingTasks.emplace(
+            path, new TextureLoadingTask(this, sceneDelegate, path, fallbackValue));
+    }
+
+    static HdVP2TextureInfo dummyInfo;
```
williamkrick, line 2487: This static variable is weird. We make a singleton dummyInfo which is grey, except if the first node that needed it happened to have a fallback color, in which case we use that nodes fallback color for everyone?

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1199,6 +1212,72 @@ TfToken MayaDescriptorToToken(const MVertexBufferDescriptor& descriptor)
...
+        : _parent(parent)
+        , _sceneDelegate(sceneDelegate)
+        , _path(path)
+        , _fallbackValue(fallbackValue)
+    {
+    }
+
+    ~TextureLoadingTask() = default;
+
+    void Load()
```
williamkrick, line 1232: I think this function should be renamed to something like EnqueueLoadOnIdle, at least to let the read know it's not doing the work right away.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -2349,28 +2436,132 @@ void HdVP2Material::_UpdateShaderInstance(const HdMaterialNetwork& mat)
...
+            }
+            dummyInfo._texture.reset(
+                _GenerateFallbackTexture(textureMgr, "dummy", VtValue(fallbackColor)));
+        }
+    }
+
+    return dummyInfo;
+}
+
+void HdVP2Material::LoadTextures()
```
williamkrick, line 2507: This function should be renamed something like enqueueLoadTextures or something so that the reader knows they will not be loaded immediately.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -2349,28 +2436,132 @@ void HdVP2Material::_UpdateShaderInstance(const HdMaterialNetwork& mat)
...
+{
+    std::lock_guard<std::mutex> lock(_textureLoadingMutex);
```
williamkrick, line 2509: I think we need a comment in here explaining why we need the lock.

We need it because multiple rprims could have this material assigned. Those rprims could be Sync in parallel, and so they could be requesting the textures load in parallel.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -2349,28 +2436,132 @@ void HdVP2Material::_UpdateShaderInstance(const HdMaterialNetwork& mat)
...
+    for (const auto& task : _textureLoadingTasks) {
+        task.second->Load();
+    }
+}
+
+void HdVP2Material::ClearPendingTasks()
+{
+    // Inform tasks that have not started or finished that this material object
+    // is no longer valid
+    std::lock_guard<std::mutex> lock(_textureLoadingMutex);
```
williamkrick, line 2519: I don't think this function can run in parallel with any other functions that operate on _textureLoadingTasks, and so it doesn't need to acquite the _textureLoadingMutex.

The callers are:

- ~HdVP2Material, which can't be called in parallel with anything else
- HdVP2RenderDelegate::cleanUpMaterials, which is called by clearInvalidData, which can't be called in parallel with itself or other places that acquire the _textureLoadingMutex

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -2349,28 +2436,132 @@ void HdVP2Material::_UpdateShaderInstance(const HdMaterialNetwork& mat)
...
+
+void HdVP2Material::_UpdateLoadedTexture(
+    HdSceneDelegate*     sceneDelegate,
+    const std::string&   path,
+    MHWRender::MTexture* texture,
+    bool                 isColorSpaceSRGB,
+    const MFloatArray&   uvScaleOffset)
+{
+    {
+        std::lock_guard<std::mutex> lock(_textureLoadingMutex);
```
williamkrick, line 2539: I think we also don't need to acquire the _textureLoadingMutex here. This is called from the idle task and nothing will be running in parallel.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -2349,28 +2436,132 @@ void HdVP2Material::_UpdateShaderInstance(const HdMaterialNetwork& mat)
...
+        // called directly from the task object method `loadOnIdle()`,
+        // we do not handle the deletion here, we will let the
+        // function on idle to delete the task object.
+        _textureLoadingTasks.erase(path);
     }
-    return info;
+
+    // TODO: it might triggers many refresh commands if USD has lots of textures
+    //       consider to consolidate/reduce the refresh
+    MGlobal::executeCommandOnIdle("refresh");
```
williamkrick, line 2564: Here we have to think about the user experience. We need the idle priority of the refreshes to be higher than the background loading priority so that we don't have to wait for all the loading to finish before the viewport updates, that's the whole point.

Given that the refresh priority is higher than the background loading priority, asking for refresh after every texture loads means we will refresh once per texture. The time the draw the viewport is probably going to be longer than the time to load one texture (depending on complexity and size), so the refreshes will limit the rate and which we can load textures. We definitely want fewer refreshes.

If the user is interacting with the scene they'll be triggering refreshes of their own, so we don't need to think about that case too much. If the user is not interacting and just watching the viewport, how many refreshes do we need? Maybe 1 per second or something in that range? We can experiment with the rate limit.

A possible implementation would be save the timestamp of the time the last idle refresh was requested, and only allow subsequent refreshes if X amount of time has past since then.

The other nice thing to do would be when all the idle tasks finish always trigger one refresh. That way if the user is waiting to see their scene complete it'll pop in immediately after finishing.

lib/mayaUsd/render/vp2RenderDelegate/render_delegate.cpp
```diff
@@ -1024,4 +1030,14 @@ HdVP2RenderDelegate::GetSamplerState(const MHWRender::MSamplerStateDesc& desc) c
  */
 const HdVP2BBoxGeom& HdVP2RenderDelegate::GetSharedBBoxGeom() const { return *sSharedBBoxGeom; }
 
+void HdVP2RenderDelegate::CleanupMaterials()
+{
+    std::lock_guard<std::mutex> guard(_renderDelegateMutex);
```
williamkrick, line 1035: I don't think this function can run in parallel with HdVP2RenderDelegate ctor or dtor so we don't need to acquire this mutex here.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1199,6 +1212,72 @@ TfToken MayaDescriptorToToken(const MVertexBufferDescriptor& descriptor)
...
+        , _path(path)
+        , _fallbackValue(fallbackValue)
+    {
+    }
+
+    ~TextureLoadingTask() = default;
+
+    void Load()
+    {
+        if (_started.load()) {
```
williamkrick, line 1234: This check still has races in it. If two rprims share a material they could both enter this code in parallel. The call to _started.load() here and _started = true below will have a single total order consistent in all threads, multiple threads could load started before any thread sets the value true, leading to multiple idle tasks loading the same texture.

I think it would probably be okay, because the 2nd task would hit the cache, but it's still better to fix this.

I believe the correct approach would be something like:

```
if (_started.exchange(true)) {
    return;
}
```

That way if the value was already true we'll return right away, and exactly one thread will get a return value of false and create the idle task.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1199,6 +1212,72 @@ TfToken MayaDescriptorToToken(const MVertexBufferDescriptor& descriptor)
...
+        }
+        _parent->_UpdateLoadedTexture(_sceneDelegate, _path, texture, isSRGB, uvScaleOffset);
+    }
+
+    HdVP2Material*    _parent;
+    HdSceneDelegate*  _sceneDelegate;
+    const std::string _path;
+    const VtValue     _fallbackValue;
+    std::atomic_bool  _started { false };
+    std::atomic_bool  _terminated { false };
```
williamkrick, line 1278: I think that _terminated doesn't need to be atomic. During the _loadOnIdle call none of the other bits of code which set _terminated true can be executing.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
williamkrick: Actually I think we don't need this lock if the _started atomic is switched to use exchange.

williamkrick (changes requested):
Overall it looks like a great approach, just a few wrinkles to iron out.

I think limiting the number of refreshes will greatly improve the user experience so we should definitely explore that before merging.

lib/mayaUsd/render/vp2RenderDelegate/render_delegate.h
```diff
@@ -173,6 +175,8 @@ class HdVP2RenderDelegate final : public HdRenderDelegate
     static HdResourceRegistrySharedPtr
         _resourceRegistry; //!< Shared and unused by VP2 resource registry
 
+    std::unordered_set<HdSprim*> _materialSprims;
```
williamkrick, line 178: Do we ever get sprims which are _not_ HdVP2Material? Having to do the dynamic cast is kind of a bummer. We only DO use this on clearing the scene so maybe the performance isn't a huge deal.

Is there no way for Usd to give us an iterator over all sprims?

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
williamkrick: Another thing to consider is using `M3dView::scheduleRefreshAllViews` rather than executeCommandOnIdle. I believe scheduleRefreshAllViews has lower overhead than creating the idle command.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -100,6 +101,16 @@ namespace mx = MaterialX;
 
 PXR_NAMESPACE_OPEN_SCOPE
 
+TF_DEFINE_ENV_SETTING(
+    MAYAUSD_DISABLE_ASYNC_TEXTURE_LOADING,
+    0,
+    "Disable asynchronous texture loading.");
+static bool _IsDisabledAsyncTextureLoading()
+{
+    static bool _v = TfGetEnvSetting(MAYAUSD_DISABLE_ASYNC_TEXTURE_LOADING) == 1;
```
williamkrick, line 110: We'll need a runtime switch to turn on and off the async loading so that test automation can control what is going on. Something that can be set through python would be good.

williamkrick:
We'll also need tests. Make sure the test includes:

- Materials that have more than 1 copy of the same texture.
- Many rpims sharing the same texture (different materials)
- Many rprims sharing the same material

We should have VP2RenderDelegate tests which do image comparisons to validate that the materials load correctly. We can use flushIdleQueue to force wait for all the texture loading to finish.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -820,12 +831,38 @@ _LoadUdimTexture(const std::string& path, bool& isColorSpaceSRGB, MFloatArray& u
...
+    desc.fWidth = 1;
+    desc.fHeight = 1;
+    desc.fFormat = MHWRender::kR8G8B8A8_UNORM;
+    desc.fBytesPerRow = 4;
+    desc.fBytesPerSlice = desc.fBytesPerRow;
+
+    const GfVec4f&             fallbackValue = value.UncheckedGet<GfVec4f>();
+    std::vector<unsigned char> texels(4);
+    for (size_t i = 0; i < 4; ++i) {
+        float texelValue = std::max(std::min(fallbackValue[i], 1.0f), 0.0f);
```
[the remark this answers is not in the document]
csyshing: This has been updated.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -2349,28 +2436,132 @@ void HdVP2Material::_UpdateShaderInstance(const HdMaterialNetwork& mat)
...
+            info._stScale.Set(
+                uvScaleOffset[0], uvScaleOffset[1]); // The first 2 elements are the scale
+            info._stOffset.Set(
+                uvScaleOffset[2], uvScaleOffset[3]); // The next two elements are the offset
+        }
+
+        return info;
+    }
+
+    VtValue fallbackValue;
```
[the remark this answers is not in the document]
csyshing: Updated.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
csyshing: I have added another check to avoid overwriting the texture:
https://github.com/Autodesk/maya-usd/pull/1980/files#diff-95070490bea9fe284d8a3c42c85ed209a5341007e11b85dac96e316dcec6e269R2584-R2588

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
csyshing: I changed this one a little bit: if there is no fallback color, all the materials would reuse the same "grey" `HdVP2TextureInfo`; otherwise create a `HdVP2TextureInfo` with their own fallback color respectively.
Hope that makes more sense!

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
csyshing: Done.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
csyshing: Done.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
csyshing: The lock has been removed.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
csyshing: The lock has been removed as well.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
csyshing: Removed lock.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
csyshing: `M3dView::scheduleRefreshAllViews` indeed improved the UI responsive, I have switched to use `M3dView::scheduleRefreshAllViews`; I also added a timestamp as you suggested, and for the last texture that has been finished loading, a refresh would be triggered regardless of the timestamp:
https://github.com/Autodesk/maya-usd/pull/1980/files#diff-95070490bea9fe284d8a3c42c85ed209a5341007e11b85dac96e316dcec6e269R2624-R2626

lib/mayaUsd/render/vp2RenderDelegate/render_delegate.cpp
csyshing: Lock removed.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
csyshing: Updated.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
csyshing: Updated as well, `_terminated` is now just a `bool`.

lib/mayaUsd/render/vp2RenderDelegate/render_delegate.h
csyshing: I can't find any API in SceneDelegate class that could return all created sprims :-(

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
csyshing: I changed the env var to be an Maya `optoinVar` called **mayaUsd_DisableAsyncTextureLoading** :
https://github.com/Autodesk/maya-usd/pull/1980/files#diff-79f656a2d9ef20b627475bb4041dad2cd7d14cf53ec94fafb7a18669fe228a6bR66-R69

Let me know if this is breaking any `maya-usd` development guideline?

csyshing:
Hi @williamkrick , sorry for the late update, was on another higher priority task. I have updated the code to address the comments, also added the unit test.

Have a look and let me know if there is any other feedback? Thanks!

kxl-adsk:
@csyshing Not related to the code changes, but in future I would recommend to keep the commits. Because you squashed them, it’s impossible to know the difference between the code that William reviewed and the new version.

csyshing:
Gotcha, sorry for the inconvenience, I will remember to follow the dev workflow in the future.

williamkrick (dismissed):
I built the changes locally and loaded ALab in less than a second! Great work @csyshing!

I think with the changes off by default they could be merged. After our QA reviews the changes I will switch the behavior to be enabled by default.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -100,6 +102,18 @@ namespace mx = MaterialX;
 
 PXR_NAMESPACE_OPEN_SCOPE
 
+static bool _IsDisabledAsyncTextureLoading()
+{
+    static const MString kOptionVarName(MayaUsdOptionVars->DisableAsyncTextureLoading.GetText());
+    if (MGlobal::optionVarExists(kOptionVarName)) {
+        return MGlobal::optionVarIntValue(kOptionVarName);
+    }
+    return false;
```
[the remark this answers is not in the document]
csyshing: No problem, I have set the async loading off by default in [c7ef3de](https://github.com/Autodesk/maya-usd/pull/1980/commits/c7ef3de4a89694fe5282ce3cb4ad82ede00a3914)

csyshing:
Hi @kxl-adsk , the unit test for this PR failed when comparing the image diff (very likely my local setup is different from the CI build), I don't think I can visualize the output image from CI, any tips?
```
    AssertionError: Images differed by 0.07098299254115226 (max allowed: 0.0001):
      /home/S/jenkins/workspace/ecg-mayausd-branch-preflight-cotd-python3-pxrusd-max-interactive-linux/ecg-maya-usd/maya-usd/test/lib/mayaUsd/render/vp2RenderDelegate/VP2RenderDelegateTextureLoadingTest/baseline/TextureLoading_Proxy_Async.png
      /home/S/jenkins/workspace/ecg-mayausd-branch-preflight-cotd-python3-pxrusd-max-interactive-linux/build/RelWithDebInfo/test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateTextureLoadingOutput/TextureLoading_Proxy_Async.png
    AssertionError: Images differed by 0.00029864515250544664 (max allowed: 0.0001):
      /home/S/jenkins/workspace/ecg-mayausd-branch-preflight-cotd-python3-pxrusd-max-interactive-linux/ecg-maya-usd/maya-usd/test/lib/mayaUsd/render/vp2RenderDelegate/VP2RenderDelegateTextureLoadingTest/baseline/TextureLoading_Proxy_Sync.png
      /home/S/jenkins/workspace/ecg-mayausd-branch-preflight-cotd-python3-pxrusd-max-interactive-linux/build/RelWithDebInfo/test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateTextureLoadingOutput/TextureLoading_Proxy_Sync.png
    AssertionError: Images differed by 0.0709829944323 (max allowed: 0.0001):
      S:\jenkins\workspace\ecg-mayausd-branch-preflight-2022-python2-pxrusd-min-interactive-windows\ecg-maya-usd\maya-usd\test\lib\mayaUsd\render\vp2RenderDelegate\VP2RenderDelegateTextureLoadingTest\baseline\TextureLoading_Proxy_Async.png
      W:\build\RelWithDebInfo\test\lib\mayaUsd\render\vp2RenderDelegate\testVP2RenderDelegateTextureLoadingOutput\TextureLoading_Proxy_Async.png
    AssertionError: Images differed by 0.00323438634713 (max allowed: 0.0001):
      S:\jenkins\workspace\ecg-mayausd-branch-preflight-2022-python2-pxrusd-min-interactive-windows\ecg-maya-usd\maya-usd\test\lib\mayaUsd\render\vp2RenderDelegate\VP2RenderDelegateTextureLoadingTest\baseline\TextureLoading_Render_Sync.png
      W:\build\RelWithDebInfo\test\lib\mayaUsd\render\vp2RenderDelegate\testVP2RenderDelegateTextureLoadingOutput\TextureLoading_Render_Sync.png
    
```

williamkrick:
@csyshing I get the same errors when I run the test locally. Here are the images I get in testVP2RenderDelegateTextureLoadingOutput
TextureLoading_Proxy_Async
TextureLoading_Proxy_Sync
:

The Async file is very different.

The difference in the sync file is tiny. In the baseline image the left most column of pixels is slightly blurry, and in my generated image the right most column of pixels is slightly blurry. It is as if in my run the object is shifted very slightly to the left. Maybe some precision difference? Possibly setting the camera / object position differently could make the test less sensitive to precision issues.

csyshing:
Thanks @williamkrick , let me double check locally

csyshing:
@williamkrick , I have updated the unit test for the sync case, but still having problem to get the async one pass: I don't think `cmds.flushIdleQueue()` can do what it supposed to do to flush the idle tasks? I debugged a bit and the queued idle tasks did not run at all.
Except modifying the code just for unit test, any other alternative API rather than `cmds.flushIdleQueue()`?

I encountered similar case before but gave up at the end (made tests in a sync approach), may I ask how do you guys usual do in unit test like that?

williamkrick:
@csyshing I spent some time on this today and I figured out the issue. When Maya starts up we suspend the idle queue to prevent interference with creating the UI and resume the idle queue later. However, when we run tests the idle events never get resumed. I added a call to `cmds.flushIdleQueue( resume=True )` in the setUp method of the class and that got the idle events running again. I also moved the call to `flushIdleQueue()` to be after the `createProxyFromFile` call because I think that is where the idle task will get created.

I also found that holding the fallback texture in a static variable is not safe. When Maya exits the MTexture could be deleted after Maya's renderer has been deleted and that crashes Maya. I've got a fix for Maya to skip destroying the texture in this case but we also need to fix it in this PR because those changes will take a while to propagate to all the difference versions of Maya. I think if we clear the static variable when the plugin is unloaded that should work but I didn't test it out.

csyshing:
I see, by adding `cmds.flushIdleQueue( resume=True )`, I can get the test running successfully locally!
I have also changed the texture to be non-static.

williamkrick:
The latest failure was testVP2RenderDelegateUSDPreviewSurface in two different preflight runs. In both cases the diffs were just barely larger than the threshold. The file is TestFallbackColor.png which sounds like it could reasonably be impacted by these changes vs. being a random failure.

csyshing:
@williamkrick the `testVP2RenderDelegateUSDPreviewSurface` test should be fixed now.
I have reverted the logic in `_LoadTexture()` to return null texture if no fallback color found, also reordered the arguments a bit to be like `(<input arguments> .... <output arguments>)`. Hope that looks better.

williamkrick:
@csyshing did you forget to push something? I don't see changes to return a nullptr texture when the fallback color is not found, and I'm still failing in testVP2RenderDelegateUSDPreviewSurface.

csyshing:
@williamkrick , I have sync-ed the changes now.

williamkrick (approved):

williamkrick:
@csyshing @santosd found some issues with the changes. If you have time to investigate them we'll log them as issues on github and assign them to you. Otherwise they'll go into our backlog and we'll prioritize them against the other work we have. What is your preference?

csyshing:
@williamkrick @santosg87 , I could have a look for that one, assign them to me please.

PR #1983, opened 2022-01-13
MAYA-121112: Maya USD error and no UI elements

seando-adsk:
#### MAYA-121112: Maya USD error and no UI elements
* Fix the path for the lib "scripts" folder.


lib/mayaUsd/resources/scripts/CMakeLists.txt
```diff
@@ -4,5 +4,5 @@ list(APPEND scripts_src
 )
 
 install(FILES ${scripts_src}
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/scripts
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/scripts
```
seando-adsk, line 7: The installer knew nothing abou this top-level "scripts" folder. It needs to be under "lib" where the other lib stuff is.

seando-adsk:
@boudrey I've fixed the script path problem and also added the path to the cmake test.

ppt-adsk (approved):

PR #1986, opened 2022-01-14
MAYA-121074 fix assertion during edit-as-maya undo / redo

pierrebai-adsk:
- The addition of the item to the pull set was not undone / redone.
- The import already has its own undo /redo but we were also using a UsdUndoBlock, which meant USD modifiaction were undone / redone twice.

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -153,13 +153,26 @@ bool writePullInformation(const Ufe::Path& ufePulledPath, const MDagPath& path)
     }
 
     // Add to a set, the set should already been created.
-    MObject pullSetObj;
-    auto    status = UsdMayaUtil::GetMObjectByName(kPullSetName, pullSetObj);
-    if (status != MStatus::kSuccess)
-        return false;
-
-    MFnSet fnPullSet(pullSetObj);
-    fnPullSet.addMember(path);
+    FunctionUndoItem::execute(
+        "Add edited item to pull set.",
+        [path]() {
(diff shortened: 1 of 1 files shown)


pierrebai-adsk:
In the end I was able to fix all assertions.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -172,6 +185,7 @@ bool writePullInformation(const Ufe::Path& ufePulledPath, const MDagPath& path)
     // Store medata on DG node
     auto              ufePathString = Ufe::PathString::string(ufePulledPath);
     MFnDependencyNode depNode(path.node());
+    MStatus           status;
```
pierrebai-adsk, line 188: FYI: the previous version of the code was re-suing a variable that is now inside the lambda.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -288,9 +302,6 @@ PullImportPaths pullImport(
         return PullImportPaths(addedDagPaths, pulledUfePaths);
     }
 
-    // Record all USD modifications in an undo block and item.
-    UsdUndoBlock undoBlock(&UsdUndoableItemUndoItem::create("Pull import USD data modifications"));
```
pierrebai-adsk, line 292: This is the redundant UsdUndoBlock that was improperly capturing what the import was doing. The importer has its own undo/redo capture system.

ppt-adsk (approved):
Thanks!  Bit of a shame that manual inspection of test output was required to find these, would have been nice to get outright test failure.

PR #1998, opened 2022-01-19
MAYA-110276 - USD: curves drawing color is the exact same as the maya viewport Gray background

vlasovi:
USD curves default color now matches Maya curves default color


lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -962,10 +962,21 @@ void HdVP2BasisCurves::_UpdateDrawItem(
                 }
             }
 
-            // If color/opacity is not found, the 18% gray color will be used
-            // to match the default color of Hydra Storm.
+            // If color/opacity is not found, the default color of Maya curves will be used
             if (colorArray.empty()) {
-                colorArray.push_back(GfVec3f(0.18f, 0.18f, 0.18f));
+                MDoubleArray curveColorResult;
+                MGlobal::executeCommand(
```
williamkrick, line 968: I think that this is not safe to call here. MGlobal::executeCommand calls into Maya's command engine which is not threadsafe. This code can be executed in parallel during the USD sync, so if multiple threads got in here together it could crash.

I think the approach should be to move the code that queries the colors over to `ProxyRenderDelegate::update()`. There we know Maya is executing serially so we should be able to safely call the command engine, save the values and then reference them here.

williamkrick (changes requested):

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
vlasovi: Right. This may be unsafe. I will make sure to address that.

vlasovi:
@williamkrick There is a call to MGlobal::executeCommandStringResult inside HdVP2Material::Sync that we should probably also fix in the same way that this one.

williamkrick (approved):

PR #2002, opened 2022-01-20
Prefer to use the PXR namespace "using" directive

dj-mcg:
This fixes Pixar internal builds


kxl-adsk (approved):
Sorry about that Dan. Thx for the fix.

dj-mcg:
No problem @kxl-adsk! Easy to find & easy to fix :smile:

PR #2004, opened 2022-01-20
MAYA-120831: Always use Namespaces when creating a Maya Reference

seando-adsk:
#### MAYA-120831: As a user, I always want to use Namespaces when creating a Maya Reference

* Make the "Use namespaces" always true.

### DIFF
--- a/lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
+++ b/lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
@@ -233,8 +233,13 @@ proc fileOptionsTabPage(string $tabLayout)
         -cc ("fo_changeReferenceNamespaceOption(\"" + $action + "\")")
         -annotation `uiRes("m_fileOptions.kNamespacesEnsureUniqueNodeNames")`
         useNamespacesGrp;
-    //All offlineMode is in namespace, so we set the useNamespacesGrp be true and gray
-    frameLayout -edit -expandCommand ("fo_NamespaceOptionExpand(\"" + $action + "\")") namespaceOptionFrame;
+
+    // For Usd Add Maya Reference we always want the "use namespaces" option to be true.
+    // So hide this checkbox.
+    // Note: we must leave all the controls for the non-useNamepsaces option as the
+    //       callbacks in Maya expect the controls to be there.
+    checkBoxGrp -edit -v1 true -manage false -visible false useNamespacesGrp;
+    frameLayout -edit -expandCommand ("fo_NamespaceOptionExpand(\"" + $action + "\"); checkBoxGrp -edit -v1 true -manage false -visible false useNamespacesGrp") namespaceOptionFrame;
 
     /*
         ------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
```diff
@@ -233,8 +233,13 @@ proc fileOptionsTabPage(string $tabLayout)
...
         useNamespacesGrp;
-    //All offlineMode is in namespace, so we set the useNamespacesGrp be true and gray
-    frameLayout -edit -expandCommand ("fo_NamespaceOptionExpand(\"" + $action + "\")") namespaceOptionFrame;
+
+    // For Usd Add Maya Reference we always want the "use namespaces" option to be true.
+    // So hide this checkbox.
+    // Note: we must leave all the controls for the non-useNamepsaces option as the
+    //       callbacks in Maya expect the controls to be there.
+    checkBoxGrp -edit -v1 true -manage false -visible false useNamespacesGrp;
+    frameLayout -edit -expandCommand ("fo_NamespaceOptionExpand(\"" + $action + "\"); checkBoxGrp -edit -v1 true -manage false -visible false useNamespacesGrp") namespaceOptionFrame;
```
seando-adsk, line 242: At first I had removed all the other controls (that are shown when use namespace is off), but that gave me all sorts of script errors from the fileOptions.mel in Maya since it refers to these controls.

So now I left all of them, but make sure that "use namespaces" is always checked (and hidden).

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
```diff
@@ -261,28 +266,9 @@ proc fileOptionsTabPage(string $tabLayout)
...
-    {
-        text -e -visible false prefixesObsoleteLabel;
-    }
-    else
-    {
-        text -e -visible true prefixesObsoleteLabel;
-    }
+
+    // Need this text field as it is referenced by Maya callbacks.
+    text -visible false prefixesObsoleteLabel;
```
seando-adsk, line 271: Here we only need the text field (not the actual text in it) since it will always be hidden.

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
```diff
@@ -233,8 +233,13 @@ proc fileOptionsTabPage(string $tabLayout)
         -cc ("fo_changeReferenceNamespaceOption(\"" + $action + "\")")
         -annotation `uiRes("m_fileOptions.kNamespacesEnsureUniqueNodeNames")`
         useNamespacesGrp;
-    //All offlineMode is in namespace, so we set the useNamespacesGrp be true and gray
-    frameLayout -edit -expandCommand ("fo_NamespaceOptionExpand(\"" + $action + "\")") namespaceOptionFrame;
+
+    // For Usd Add Maya Reference we always want the "use namespaces" option to be true.
+    // So hide this checkbox.
+    // Note: we must leave all the controls for the non-useNamepsaces option as the
```
ppt-adsk, line 239: "non-useNamepsaces"

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
```diff
@@ -854,7 +809,7 @@ global proc addMayaReferenceToUsdUiCb(string $parent, string $selectedFile)
 
     int $radioExist = `radioCollection -exists namespaceRadioCollection`;
     int $useNamePrefix  = 1;
-    if($useNS)
+    if (true) // Use namespaces
```
ppt-adsk, line 812: Can we just get rid of this?  Feels awkward to leave this dead code in.  Same at line 833.

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
seando-adsk: Fixed

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
seando-adsk: In this case (and the one below) I decided to leave the if here because of the variable and comment regarding namespaces. And in case we ever tried to compare this to the code from fileOptions.mel.

ppt-adsk (approved):

PR #2007, opened 2022-01-20
MAYA-106930 Pre-evaluate all the compute inputs so that any lazy-evaluated shared

williamkrick:
inputs are computed serially from sprim sync instead of in parallel in rprim sync.


lib/mayaUsd/render/vp2RenderDelegate/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ target_sources(${PROJECT_NAME}
         bboxGeom.cpp
         debugCodes.cpp
         draw_item.cpp
+		extComputation.cpp
```
JGamache-autodesk, line 10: can you convert these tabs to spaces?

JGamache-autodesk (dismissed):

JGamache-autodesk (approved):

PR #2012, opened 2022-01-21
MAYA-121242 fix flipped normals.

pierrebai-adsk:
Maya did not write the USD "orientation" attribute. In addition, Maya flipped the vertices order when editing a USD mesh with "leftHanded" orientation. On merge-to-USD, Maya would write right-handed vertices but not write the "orientation" property.

The result would be righ-handed verticeswith left-handed orientation attribute.

See: MAYA-105623 for the original reason we flip vertices orders. It has to do with the way "leftHanded" affect vertices order and normals in USD.

The fix is to explicitly write the "orientation" attribute on export.

A few improvements in the verbosity options and simplification of some code were also made whilte investigating the problem and are included.

### DIFF
--- a/lib/usd/utils/MergePrims.cpp
+++ b/lib/usd/utils/MergePrims.cpp
@@ -250,7 +250,7 @@ bool isLocalTransformModified(const UsdPrim& srcPrim, const UsdPrim& dstPrim)
 //
 // There are some metadata that we know the merge-to-USD temporary export-to-USD
 // will never produce. Others are generated only when needed, so their absence
-// should always means to remove them.
+// should always mean to remove them.
 //
 // isMetadataAlwaysPreserved() is for those metadata we know we never produce.
 //
--- a/lib/usd/utils/MergePrims.cpp
+++ b/lib/usd/utils/MergePrims.cpp
@@ -249,19 +249,35 @@ bool isLocalTransformModified(const UsdPrim& srcPrim, const UsdPrim& dstPrim)
 // Special metadata handling.
 //
 // There are some metadata that we know the merge-to-USD temporary export-to-USD
-// will never produce. List them here to be always preserved.
+// will never produce. Others are generated only when needed, so their absence
+// should always means to remove them.
+//
+// isMetadataAlwaysPreserved() is for those metadata we know we never produce.
+//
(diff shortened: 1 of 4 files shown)


lib/usd/utils/MergePrims.cpp
```diff
@@ -249,19 +249,35 @@ bool isLocalTransformModified(const UsdPrim& srcPrim, const UsdPrim& dstPrim)
 // Special metadata handling.
 //
 // There are some metadata that we know the merge-to-USD temporary export-to-USD
-// will never produce. List them here to be always preserved.
+// will never produce. Others are generated only when needed, so their absence
+// should always means to remove them.
```
ppt-adsk, line 253: "should always means" --> "should always mean".

ppt-adsk (approved):

dj-mcg:
Hi @pierrebai-adsk & @kxl-adsk - This broke one of our round trip tests I think. We start with a .ma file that contains a mesh with leftHanded orientation, which is exported to USD properly. Then we re-import that same exported USD file, then re-export. This re-exported USD file flips the verts and authors rightHanded orientation. 

Is this behavior expected? I assumed it should properly import (and re-export) the leftHanded USD mesh but I don't have the context that you mentioned regarding Maya flipping the vert order.

Thanks,
Dan

dj-mcg:
Also, is there a reason we don't write the orientation attribute sparsely? It looks like it's alwasy explicitly authoring opinions with the default value (rightHanded). I'm happy to make that change unless there's a reason not to?

pierrebai-adsk:
The problem was that the "opposite" flag in Maya flips the vertex order of the face *and* the normals, while the one in USD only flips the face order. So in Maya we cannot use "opposite" to only do one thing. So, when we encounter leftHanded in USD, we cannot use "opposite".

Maybe in the future we could undo the vertex re-ordering on export. We would need to keep around something that tells us to do so.

pierrebai-adsk:
I'll add that we recently improved the transformation round-trip by using a notion of "equivalent transform" that let us detect that two transforms might be build up differently but represent the same value. We could do something similar for vertices and normals, taking into consideration handed-ness when comparing the raw value arrays. This would at least detect that they have not been edited and avoid authoring opinions.

This is all speculative work though, we'd need to schedule and implement it.

PR #2013, opened 2022-01-21
MAYA-121245 fix failed edit-as-maya commands

pierrebai-adsk:
Properly undo semi-successful edit-as-Maya command and other related commands. Also provide helpful error messages when a command fails mid-way.

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -298,7 +298,7 @@ PullImportPaths pullImport(
 
     std::string mFileName = context.GetUsdStage()->GetRootLayer()->GetIdentifier();
     if (mFileName.empty()) {
-        TF_RUNTIME_ERROR("Empty file specified. Exiting.");
+        TF_WARN("Empty file specified. Exiting.");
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -298,7 +298,7 @@ PullImportPaths pullImport(
 
     std::string mFileName = context.GetUsdStage()->GetRootLayer()->GetIdentifier();
     if (mFileName.empty()) {
-        TF_RUNTIME_ERROR("Empty file specified. Exiting.");
+        TF_WARN("Empty file specified. Exiting.");
```
ppt-adsk, line 301: "Exiting" wording feels misleading, and so does talking about a file --- we're pulling on USD data that's in memory.

ppt-adsk (changes requested):
Great catch with the undo on failure.  Just the warning wording that needs tweaking.

ppt-adsk (approved):

pierrebai-adsk:
The code pass PF before and only I changed the wording of a warning. Feel free to wait for the PF or not before merging.

PR #2015, opened 2022-01-23
MayaReference auto-edit support

kxl-adsk:


lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
```diff
@@ -985,6 +997,8 @@ global proc string addMayaReferenceToUsd(string $ufePath)
         $qVarName = quoteString($tmpValue);
     }
 
+    string $autoEditAsMayaData = `optionVar -query usdMayaReferenceEditAsMayaDat`;
```
seando-adsk, line 1000: Your optionVar is an "int"

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.py
```diff
@@ -34,22 +35,26 @@ def defaultVariantSetName():
...
+def createPrimAndAttributes(stage, primPath, mayaReferencePath, mayaNamespace, mayaAutoEdit):
     prim = stage.DefinePrim(primPath.path, 'MayaReference')
 
     mayaReferenceAttr = prim.CreateAttribute('mayaReference', Sdf.ValueTypeNames.Asset)
     mayaReferenceAttr.Set(mayaReferencePath)
 
     mayaNamespaceAttr = prim.CreateAttribute('mayaNamespace', Sdf.ValueTypeNames.String)
     mayaNamespaceAttr.Set(mayaNamespace)
+    
+    mayaAutoEditAttr = prim.CreateAttribute('mayaAutoEdit', Sdf.ValueTypeNames.Bool)
```
seando-adsk, line 47: I saw below that there are tokens for these strings (mayaAutoEdit, mayaNamespace & mayaReference). It would be great it we used the tokens here in instead of hard-coding.

seando-adsk (changes requested):

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.py
kxl-adsk: There is a bigger cleanup to do around these. See other places, e.g. https://github.com/Autodesk/maya-usd/blob/70124fe61c82bbafc49393101c012a85c7b47ecc/lib/mayaUsd/fileio/translators/translatorMayaReference.cpp#L141 that requires cleanup. Not in scope for this PR.

seando-adsk (approved):

ppt-adsk (approved):

PR #2020, opened 2022-01-24
MAYA-121247 edit-as-maya whole selected object - not sub-part

pierrebai-adsk:
Use the entire selection instead of the item under the mouse when right-clicking in the viewport to edit-as-Maya or duplicate-as-Maya.

### DIFF
--- a/plugin/adsk/scripts/USDMenuProc.mel
+++ b/plugin/adsk/scripts/USDMenuProc.mel
@@ -15,22 +15,13 @@
 
 proc string expandToSelection(string $obj)
 {
-    // By default, for example if the obj is empty, we want to use the selection.
-    int $expandToSel = 1;
-
-    // If obj is non-empty, expand it to the selection if it is part of the selection.
     if (size($obj)) {
-        string $script = "import maya.internal.ufeSupport.utils as ufeUtils; selection = ufeUtils.getNonMayaSelectedItems(); inSel = ['" + $obj + "'.startswith(sel) for sel in selection]; selection and sum(inSel)";
-        $expandToSel = `python($script)`;
+        // If obj is non-empty, expand it to the selection if it is part of the selection.
+        return `python("import mayaUsd_selectionUtils; mayaUsd_selectionUtils.expandPathToSelection('" + $obj + "')")`;
+    } else {
+        // If obj is empty, use the first selected item non-Maya.
+        return `python("import maya.internal.ufeSupport.utils as ufeUtils; items = ufeUtils.getNonMayaSelectedItems(); items[0] if items else None")`;
     }
-
-    // If we expand to selection, use the UFE selection.
-    if ($expandToSel) {
-        string $nonMayaObjs[] = `python("import maya.internal.ufeSupport.utils as ufeUtils; ufeUtils.getNonMayaSelectedItems()")`;
-        $obj = $nonMayaObjs[0];
-    }
-
-    return $obj;
 }
 
 proc int canEditAsMaya(string $obj)
@@ -70,6 +61,7 @@ global proc USDMenuProc(string $parent, string $obj)
         return;
         
     $obj = expandToSelection($obj);
+    print($obj);
     if (size($obj) != 0)
     {
         popupMenu -e -dai $parent;
--- a/plugin/adsk/scripts/USDMenuProc.mel
+++ b/plugin/adsk/scripts/USDMenuProc.mel
@@ -13,19 +13,6 @@
 // limitations under the License.
 //
 
-proc int canEditAsMaya()
-{
-    if (!hasPrimUpdater())
-        return 0;
-
-    string $nonMayaObjs[] = `python("import maya.internal.ufeSupport.utils as ufeUtils; ufeUtils.getNonMayaSelectedItems()")`;
-    string $obj = $nonMayaObjs[0];
-    if (size($obj) != 0) {
-        return `python("from mayaUsd.lib import PrimUpdaterManager; PrimUpdaterManager.canEditAsMaya('" + $obj + "')")`;
-    }
-    return 0;
-}
-
 proc string expandToSelection(string $obj)
 {
     // By default, for example if the obj is empty, we want to use the selection.
@@ -33,7 +20,7 @@ proc string expandToSelection(string $obj)
 
     // If obj is non-empty, expand it to the selection if it is part of the selection.
     if (size($obj)) {
-        string $script = "import maya.internal.ufeSupport.utils as ufeUtils; selection = ufeUtils.getNonMayaSelectedItems(); inSel = ['" + $obj + "'.startswith(sel) for sel in selection]; not selection or sum(inSel)";
+        string $script = "import maya.internal.ufeSupport.utils as ufeUtils; selection = ufeUtils.getNonMayaSelectedItems(); inSel = ['" + $obj + "'.startswith(sel) for sel in selection]; selection and sum(inSel)";
         $expandToSel = `python($script)`;
     }
 
@@ -46,12 +33,22 @@ proc string expandToSelection(string $obj)
     return $obj;
 }
 
+proc int canEditAsMaya(string $obj)
+{
+    if (!hasPrimUpdater())
+        return 0;
+
+    if (size($obj) != 0) {
+        return `python("from mayaUsd.lib import PrimUpdaterManager; PrimUpdaterManager.canEditAsMaya('" + $obj + "')")`;
+    }
+    return 0;
+}
+
 global proc mayaUsdMenu_editAsMaya(string $obj)
 {
     if (!hasPrimUpdater())
         return;
 
-    $obj = expandToSelection($obj);
     if (size($obj) != 0) {
         mayaUsdEditAsMaya $obj;
     }
@@ -62,7 +59,6 @@ global proc mayaUsdMenu_duplicate(string $ufePath)
     if (!hasPrimUpdater())
         return;
         
-    $ufePath = expandToSelection($ufePath);
     if (size($ufePath)) {
         mayaUsdDuplicate $ufePath "|world";
     }
@@ -73,14 +69,14 @@ global proc USDMenuProc(string $parent, string $obj)
     if (!hasPrimUpdater())
         return;
         
-    int $haveNonMaya = `python("import maya.internal.ufeSupport.utils as ufeUtils; ufeUtils.hasNonMayaSelectedItems()")`;
-    if ($haveNonMaya)
+    $obj = expandToSelection($obj);
+    if (size($obj) != 0)
     {
         popupMenu -e -dai $parent;
         setParent -menu $parent;
 
         setParent -menu ..;
-        if (canEditAsMaya()) {
+        if (canEditAsMaya($obj)) {
             // Temporary - hide this context menu items behind an
             //             env var until it is completed.
             string $s = `getenv "MAYAUSD_ENABLE_EDIT_AS_MAYA_DATA"`;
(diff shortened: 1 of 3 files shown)


plugin/adsk/scripts/USDMenuProc.mel
```diff
@@ -40,16 +62,16 @@ global proc USDMenuProc(string $parent, string $obj)
...
             string $s = `getenv "MAYAUSD_ENABLE_EDIT_AS_MAYA_DATA"`;
             string $ls = tolower($s);
             if (`match $ls "true|yes|on|1"` !="")
             {
-                menuItem -label "Edit As Maya Data" -image "edit_as_Maya.png" -command ("mayaUsdEditAsMaya \"" + $obj + "\"");
+                menuItem -label "Edit As Maya Data" -image "edit_as_Maya.png" -command ("mayaUsdMenu_editAsMaya");
             }
         }
-        menuItem -label "Duplicate As Maya Data" -command ("mayaUsdDuplicate \"" + $obj + "\"" + "\"|world\"");
+        menuItem -label "Duplicate As Maya Data" -command ("mayaUsdMenu_duplicate");
```
seando-adsk, line 75: As a general rule the way these things normally work is if the input object is selected, then you operate on the entire selection, otherwise you operate on only the input object. Does that work in this case?

pierrebai-adsk: IIUC, if $obj is in the selection, use the code I wrote, else use only $obj?

My mel knowledge is low, how would I check if $obj is selected?

pierrebai-adsk: Found:
`proc int isObjectSelected(string $obj)
{
    if (size($obj)) {
        string $sel[] = `ls -sl`;
        return stringArrayContains($obj, $sel);
    }
    return 0;
}
`

pierrebai-adsk: I tried it, but the object is not part of the selection even though it is part of the selection. 😃 What I mean is that when you select the parent of geom in the outliner, the whole object hierarchy is selected in the viewport, but the selection as returned by `ls -sl -ufe` only contains the top node.

So, I cannot do as you suggested. The goal is to edit the UFE selection, even if the user right-click on a sub-part of the object in the viewport.

seando-adsk: What you are getting as the input $obj is the Ufe path string for the object. So you need the `-ufe` flag to `ls`. Something like this:
```
proc int isObjectSelected(string $obj) { if (size($obj)) { string $sel[] = `ls -sl -ufe`; return stringArrayContains($obj, $sel); } return 0; }
```

But it looks like this proc only gets called when there is a non-Maya item selected. In fact you don't even need to RMB over the non-Maya item (in the viewport) for this menu to be called. Simply RMB over empty space. So I guess the input object is really irrelevant. It would just be the first non-Maya item selected and not necessarily the one you RMB over.

So you can ignore my first comment.

plugin/adsk/scripts/USDMenuProc.mel
```diff
@@ -13,21 +13,43 @@
...
+global proc mayaUsdMenu_editAsMaya()
+{
+    if (!hasPrimUpdater())
+        return;
+
+    string $nonMayaObjs[] = `python("import maya.internal.ufeSupport.utils as ufeUtils; ufeUtils.getNonMayaSelectedItems()")`;
+    string $obj = $nonMayaObjs[0];
+    if (size($obj) != 0) {
+        mayaUsdEditAsMaya $obj;
+    }
```
seando-adsk, line 38: The title of this PR says "edit-as-maya **entire** maya selection". The code you have written does not do that, it only edits the **first** non-Maya selected item? How is that different than using the input $obj to `USDMenuProc`?

pierrebai-adsk: Well, I'm describing what the observable results of the use-case are. Before my changes, if a multi-part object was selected by selecting its root node, the entire object geometry would be highlighted in the viewport but right-clicking on it would only edit the sub-object under the mouse.

With my changes, Edit-as-Maya edits starting from the root node that was selected.

I'm not claiming anything about the multi-objects case, as we don't support invoking edit-as-Maya on multiple objects at once. Note that the old code was also only editing the first object.

A case could be made to issue multiple Edit-as-Maya commands if multiple objects are selected, but that would be a separate story.

pierrebai-adsk: Note: this is all based on the pre-requisite that USD selection is in sub-component mode. When in component mode, the old code would edit the whole component. In sub-component mode, it would edit only a ingle sub-component, even though the selection was on a parent component.

pierrebai-adsk: Note #2: if you got multiple objects selected, right-clicking does not show any menu, but with the old and new code.

pierrebai-adsk: I changed the PR title to better reflect the intent. The JIRA issue was that when an object with sub-component was selected, right-clicking on it would only edit a sub-part, the not the whole selected object.

seando-adsk: Let's have a discussion offline. I looked at the original bug logged and following the steps I don't need any sub-component mode to see the problem. I can also select an item and then RMB over a completely different non-selected item and that one will be the input $obj (even though its not selected).

pierrebai-adsk: I modified the code to use the object if it is not under the selection.

If it is under the selection, then I use the selection. (Well the first object in the selection as Edit-as-Maya only takes one object.)

pierrebai-adsk: I retested with selection, without, and object outside the selection:

- RMB on object outside the selection: edits the object, ignores selection.
- RMB on sub-object of the selection: edits the first object of the selection instead of the sub-object.
- RMB without a selection: edits the object.

For the last one, the old code would not even show the menu...

ppt-adsk: Not sure if this is useful or relevant, but Ufe::Selection has a containsAncestor() member.

pierrebai-adsk: In my case, I would need containsDescendant()

plugin/adsk/scripts/USDMenuProc.mel
```diff
@@ -13,28 +13,64 @@
...
 {
-    if (!hasPrimUpdater())
-        return 0;
+    // By default, for example if the obj is empty, we want to use the selection.
+    int $expandToSel = 1;
 
-    if (size($obj) == 0) {
+    // If obj is non-empty, expand it to the selection if it is part of the selection.
+    if (size($obj)) {
+        string $script = "import maya.internal.ufeSupport.utils as ufeUtils; selection = ufeUtils.getNonMayaSelectedItems(); inSel = ['" + $obj + "'.startswith(sel) for sel in selection]; selection and sum(inSel)";
```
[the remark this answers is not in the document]
ppt-adsk: This is O(n) for n items in the selection.  You should be able to use Ufe::Selection::containsAncestor(), which is amortized O(m), for a query path of m components.

plugin/adsk/scripts/USDMenuProc.mel
ppt-adsk: Which is there as well, also O(m) for a path of m components.

pierrebai-adsk: Ah unfortunately, given the discussion, I actually need to find *which* item in the selection is the ancestor. So while containsAccestorInclusive) did the right thing during my testing, there is no function to retrieve that ancestor, so I will have to loop over the items in the selection anyway.

ppt-adsk: I would do it differently: I would loop on the path components, calling Ufe::Selection::contains() for each path.  This will be better performing, because paths are typically short, and for large selections this will be a win.

pierrebai-adsk: What I did is first check if the selection is an ancestor (quick check). Then for each element of the selection, I check if this is the ancestor.

plugin/adsk/scripts/mayaUsd_selectionUtils.py
```diff
@@ -0,0 +1,17 @@
...
+    If there is no selection or no item in the selection is an ancestor of the item,
+    then the original path itself is returned.
+    """
+    selection = ufe.GlobalSelection.get()
+    ufePath = ufe.PathString.path(path)
+    if selection.containsAncestor(ufePath):
+        for sel in selection:
+            subSel = ufe.Selection()
+            subSel.append(sel)
+            if subSel.containsAncestor(ufePath):
```
ppt-adsk, line 15: Just use Ufe::Path::startsWith().

plugin/adsk/scripts/mayaUsd_selectionUtils.py
```diff
@@ -0,0 +1,17 @@
...
+
+def expandPathToSelection(path):
+    """
+    Retrieves the path of the selection item that contains the provided path.
+    If there is no selection or no item in the selection is an ancestor of the item,
+    then the original path itself is returned.
+    """
+    selection = ufe.GlobalSelection.get()
+    ufePath = ufe.PathString.path(path)
+    if selection.containsAncestor(ufePath):
```
ppt-adsk, line 11: Again, I would do this:
```
while not ufePath.empty():
    if selection.contains(ufePath):
        return ufe.PathString.string(ufePath)
    ufePath = ufePath.pop()
```

pierrebai-adsk: This is not exactly what we need. But it's a good starting point: I need to actually go up until I'm *no longer* in the selection and keep the one just before not being in the selection. (Your suggestion would return to original sub-part.)

ppt-adsk: But isn't it the case that if the original sub-part is in the selection, then we DO want to return it?

pierrebai-adsk: Yes. Look at the new version of the code. I stop when the parent would not be in the selection, thus finding the highest in-selection node that contained the object.

plugin/adsk/scripts/mayaUsd_selectionUtils.py
```diff
@@ -0,0 +1,18 @@
...
```
pierrebai-adsk, line 11: Ah, the if is redundant with the while... I will fix that. The code is getting shorter that is a good sign 😃

ppt-adsk (approved):
Looks good, thanks!

plugin/adsk/scripts/USDMenuProc.mel
pierrebai-adsk: I mostly implemented what you suggested, thanks for all the input, using the right function really simplified the code. (And is very probably faster with large selections.)

plugin/adsk/scripts/USDMenuProc.mel
```diff
@@ -13,28 +13,56 @@
...
+
 global proc USDMenuProc(string $parent, string $obj)
 {
     if (!hasPrimUpdater())
         return;
         
-    int $haveNonMaya = `python("import maya.internal.ufeSupport.utils as ufeUtils; ufeUtils.hasNonMayaSelectedItems()")`;
-    if ($haveNonMaya)
+    $obj = expandToSelection($obj);
+    print($obj);
```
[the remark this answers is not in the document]
pierrebai-adsk: Good catch

ppt-adsk (approved):

seando-adsk (approved):

PR #2031, opened 2022-01-26
MAYA-114709 As a user, I'd like to see a waiting cursor when pulling …

boudrey:
…or duplicating to Maya data


seando-adsk (approved):
The changes are good, but what do you think about creating a little helper class where constuctor/destructor toggle the cursor?

boudrey:
Still would not guaranty that cursor would actually be restore. The real fix would be to publish TwaitCursor to OpenMaya and create the Python binding for it. Yes, I can wrap it in a function for now.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -160,6 +160,25 @@ static const std::vector<std::string> kSchemaNiceNames = {
 // clang-format on
 #endif
 
+//! \brief Change the cursor to wait state when start() is called and restore it on destruction.
+struct MWaitCursor
+{
+    bool _started { false };
+
+    MWaitCursor() = default;
+    void start()
```
ppt-adsk, line 169: Why do we need the start() method?  Can't we just start in the constructor, and get rid of the _started data member?  Also not sure about the "M" prefix in the name.

plugin/adsk/scripts/USDMenuProc.mel
```diff
@@ -47,9 +47,9 @@ global proc USDMenuProc(string $parent, string $obj)
             string $ls = tolower($s);
             if (`match $ls "true|yes|on|1"` !="")
             {
-                menuItem -label "Edit As Maya Data" -image "edit_as_Maya.png" -command ("mayaUsdEditAsMaya \"" + $obj + "\"");
+                menuItem -label "Edit As Maya Data" -image "edit_as_Maya.png" -command ("{waitCursor -state 1; mayaUsdEditAsMaya \"" + $obj + "\"; waitCursor -state 0;}");
```
ppt-adsk, line 50: Hey!  Maybe you can do a guard object in MEL!

lib/mayaUsd/ufe/UsdContextOps.cpp
boudrey: This is how this class should look like if it was part of OpenMaya, and hopefully one day.

plugin/adsk/scripts/USDMenuProc.mel
boudrey: Please let me know if you know how to do it.

ppt-adsk: I would be very surprised if this would be possible :)  MEL is what it is.

ppt-adsk (approved):

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -160,6 +160,13 @@ static const std::vector<std::string> kSchemaNiceNames = {
 // clang-format on
 #endif
 
+//! \brief Change the cursor to wait state when start() is called and restore it on destruction.
```
seando-adsk, line 163: Minor - there is no more start method.

ppt-adsk (approved):

seando-adsk (approved):

PR #2032, opened 2022-01-26
Fix group command not finding unique names

JGamache-autodesk:
Delays checking for name unicity until execution starts.
Adds unit test for the issue.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoInsertChildCommand.h
+++ b/lib/mayaUsd/ufe/UsdUndoInsertChildCommand.h
@@ -64,9 +64,9 @@ private:
     bool insertChildUndo();
 
     UsdSceneItem::Ptr _ufeDstItem;
-    UsdSceneItem::Ptr _ufeParentItem;
 
     Ufe::Path _ufeSrcPath;
+    Ufe::Path _ufeParentPath;
     Ufe::Path _ufeDstPath;
 
     PXR_NS::SdfPath _usdSrcPath;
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/ufe/UsdUndoInsertChildCommand.h
```diff
@@ -64,6 +64,7 @@ class MAYAUSD_CORE_PUBLIC UsdUndoInsertChildCommand : public Ufe::InsertChildCom
     bool insertChildUndo();
 
     UsdSceneItem::Ptr _ufeDstItem;
+    UsdSceneItem::Ptr _ufeParentItem;
```
ppt-adsk, line 67: We only seem to use the _ufeParentItem's path, so we should probably store that as the data member.  Whenever mutating operations on the graph get involved (e.g. parenting), and you undo / redo, stale prims (i.e. stale scene items) are always a concern, so using the best practice of storing a Ufe::Path in a command applies here.

ppt-adsk (commented):
Thanks for the fix!  Great test!  Just a minor tweak and we'll be done.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.h
JGamache-autodesk: That is actually what I tried first. It failed the testParentCmd.testParentToProxyShape test because getting back the proxy shape from the parent path requires work. This is why I store the parent Ufe item instead to keep the code simple.

ppt-adsk (approved):

PR #2033, opened 2022-01-26
MAYA-106075 - USD selection highlighting doesn't take Color Settings into consideration

vlasovi:
Get up-to-date selection color settings from Maya. This includes 'lead' color as well as curve and surface active colors.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -1207,8 +1207,8 @@ void HdVP2BasisCurves::_UpdateDrawItem(
             // with one color vertex buffer.
             if (drawItem->ContainsUsage(HdVP2DrawItem::kSelectionHighlight)) {
                 const MColor colors[] = { drawScene.GetWireframeColor(),
-                                          drawScene.GetSelectionHighlightColor(false),
-                                          drawScene.GetSelectionHighlightColor(true) };
+                                          drawScene.GetSelectionHighlightColor("curve"),
+                                          drawScene.GetSelectionHighlightColor() };
 
                 // Store the indices to colors.
                 std::vector<unsigned char> colorIndices;
@@ -1264,9 +1264,9 @@ void HdVP2BasisCurves::_UpdateDrawItem(
                 int  primitiveStride = 0;
 
                 const MColor& color
-                    = (_selectionStatus != kUnselected
-                           ? drawScene.GetSelectionHighlightColor(_selectionStatus == kFullyLead)
-                           : drawScene.GetWireframeColor());
+                    = (_selectionStatus != kUnselected ? drawScene.GetSelectionHighlightColor(
+                           _selectionStatus == kFullyLead ? nullptr : "curve")
+                                                       : drawScene.GetWireframeColor());
 
                 if (desc.geomStyle == HdBasisCurvesGeomStylePatch) {
                     if (_selectionStatus != kUnselected) {
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -1206,9 +1206,10 @@ void HdVP2BasisCurves::_UpdateDrawItem(
             // it needs to display both wireframe color and selection highlight
             // with one color vertex buffer.
             if (drawItem->ContainsUsage(HdVP2DrawItem::kSelectionHighlight)) {
-                const MColor colors[] = { drawScene.GetWireframeColor(),
-                                          drawScene.GetSelectionHighlightColor("curve"),
-                                          drawScene.GetSelectionHighlightColor() };
+                const MColor colors[]
+                    = { drawScene.GetWireframeColor(),
+                        drawScene.GetSelectionHighlightColor(HdPrimTypeTokens->basisCurves),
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -1264,9 +1264,9 @@ void HdVP2BasisCurves::_UpdateDrawItem(
                 int  primitiveStride = 0;
 
                 const MColor& color
-                    = (_selectionStatus != kUnselected
-                           ? drawScene.GetSelectionHighlightColor(_selectionStatus == kFullyLead)
-                           : drawScene.GetWireframeColor());
+                    = (_selectionStatus != kUnselected ? drawScene.GetSelectionHighlightColor(
+                           _selectionStatus == kFullyLead ? nullptr : "curve")
+                                                       : drawScene.GetWireframeColor());
```
williamkrick, line 1269: Should this little bit of logic deciding which color to use based on the selection status move into a common location? I'm not sure if it makes sense on drawScene, maybe if we do a shared parentClass for rprims this logic could live there.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -1500,12 +1501,39 @@ GfVec3f ProxyRenderDelegate::GetCurveDefaultColor()
...
+        queryCommand += "\"`; colorIndex -q $index;";
+    } else {
+        queryCommand = "displayRGBColor -q \"lead\"";
+    }
+
+    // Query and return the selection color.
+    {
+        MDoubleArray                colorResult;
+        std::lock_guard<std::mutex> _(_mayaCommandEngineMutex);
+        MGlobal::executeCommand(queryCommand, colorResult);
```
williamkrick, line 1520: If a caller passes in a className which isn't valid do we get some kind of error we can warn about?

lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
vlasovi: Yes, I think I will address that as part of my future work on points.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -1500,12 +1501,39 @@ GfVec3f ProxyRenderDelegate::GetCurveDefaultColor()
...
+        queryCommand += className;
+        queryCommand += "\"`; colorIndex -q $index;";
+    } else {
+        queryCommand = "displayRGBColor -q \"lead\"";
+    }
+
+    // Query and return the selection color.
+    {
+        MDoubleArray                colorResult;
+        std::lock_guard<std::mutex> _(_mayaCommandEngineMutex);
```
williamkrick, line 1519: Can we rename the std::lock_guard local variable to something meaningful? I can guess at your thinking that nobody should be doing anything to it after it is created because of the RAII nature of the object so the name doesn't matter, but I don't like the style of naming it underscore. It just looks strange to me and then I spent some time thinking about it.

williamkrick (changes requested):
I think a lot of the testVP2RenderDelegate* tests will fail because the image diffs will pick up the different color of the selection.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
vlasovi: executeCommand runs with displayEnabled = false, so I guess it will not write any error info into log. I think it would be good to set displayEnabled to true here

williamkrick (approved):
Looks great!

kxl-adsk:
@vlasovi I run the PF and there is still one error to fix:
```
15:49:14     S:\jenkins\workspace\ecg-mayausd-branch-preflight-pr-python3-pxrusd-mid2-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\render\vp2RenderDelegate\proxyRenderDelegate.cpp(1553): error C2039: 'applyViewTransform': is not a member of 'Autodesk::Maya::OpenMaya20220000::MColorPickerUtilities'
15:49:14     P:\shared\Artifactory\Windows\rundev-master-debug\202104141515\runtime\0443d2a\runTime\include\maya/MColorPickerUtilities.h(72): note: see declaration of 'Autodesk::Maya::OpenMaya20220000::MColorPickerUtilities'
15:49:14     S:\jenkins\workspace\ecg-mayausd-branch-preflight-pr-python3-pxrusd-mid2-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\render\vp2RenderDelegate\proxyRenderDelegate.cpp(1554): error C2039: 'kInverse': is not a member of 'Autodesk::Maya::OpenMaya20220000::MColorPickerUtilities'
15:49:14     P:\shared\Artifactory\Windows\rundev-master-debug\202104141515\runtime\0443d2a\runTime\include\maya/MColorPickerUtilities.h(72): note: see declaration of 'Autodesk::Maya::OpenMaya20220000::MColorPickerUtilities'
```

vlasovi:
@kxl-adsk This error is due to the fact that our preflight is using an old version of Maya 2023 that doesn't have the API I use. The issue should be fixed by the preflight updated that @seando-adsk has in review now: https://git.autodesk.com/media-and-entertainment/ecg-maya-usd/pull/191

Issue #1991, opened 2022-01-18
[MAYA-121225] Performance issue when lots of models are selected


gracekumagai:
**Describe the bug**
When testing the performance of common set dresser workflows, we noticed that there is a very big slowdown when the user has the root of a heavy city set selected (or similarly, a group that contains lots of models). The set has thousands of models with varying complexity (hero buildings, hero props, etc). Unfortunately, the set we used to test is a production asset so I cannot share it to include steps to reproduce. 

We noticed that it typically takes ~10 seconds to perform common actions with a city set selected.
- Selecting/de-selecting the root in the outliner
- De-selecting in the viewport
- Rotating, panning, or zooming the camera (per update)
- Translating the set (per update)
- Framing the set ('F')

Since it takes so long to perform each update when moving the camera or translating the set, it is not interactive and feels unresponsive.

It also seems to take a very long time to switch between the channel box and attribute editor with the set selected (~ 25-30 seconds).

Thank you!

**Specs (if applicable):**
maya2022.update3

williamkrick:
Hi @gracekumagai,

I logged MAYA-121225 to track this one. I'm testing with a heavy scene internally which takes me ~3s to select the root in the outliner. I'm guessing that the bottlenecks you are seeing are similar to the ones I am, but without a sharable scene I can't be sure. I think my test is similar enough, but the possibility of a significant difference exists.

With that said, I'm reproducing slowness with workflows 1, 2, and 4.

With workflow 3 and 5 it is less clear. With 3 (Rotating, panning or zooming the camera) the first couple frames are slow, but then consolidated world kicks in and the subsequent frames are fast. Similarly, when consolidated world is working the framing the set is slow.

One difference between our scenes could be materials. VP2 render delegate has the ability to consolidate objects whose material compiles to the same shader if the only difference in the material parameters is the diffuseColor. If your materials compile to different shaders, or you have more parameters on your materials that vary from object to object then consolidated world doesn't work and VP2RenderDelegate's performance suffers. Can you share some more details about the materials you are using in the slow scene and how they vary from prim to prim?

What is happening here is modifying the selection is changing the consolidation status of the MRenderItems that VP2RenderDelegate uses to draw the wireframe objects. After a few frames those objects get consolidated and the performance jumps. That explains the bad performance translating the set as well. Setting the transform breaks the MRenderItems out of consolidation and triggers the slower drawing path.

[...]

PR #2034, opened 2022-01-26
MAYA-121408 Disable the workaround of populating selection every frame. It is too

williamkrick:
slow. Wait for a real fix.


kxl-adsk (approved):

williamkrick:
I think this change mostly alleviates https://github.com/Autodesk/maya-usd/issues/1991. I still have more changes for the selection workflows.

PR #2035, opened 2022-01-26
Disable multi-material options for legacy shading

JGamache-autodesk:
When a Plug-in context introduces a legacy shading mode, we disable the
multi-material export option since they only work for the new
useRegistry shading mode.

Fix harmless warning about variable redeclaration.

### DIFF
--- a/plugin/adsk/scripts/mayaUsdTranslatorExport.mel
+++ b/plugin/adsk/scripts/mayaUsdTranslatorExport.mel
@@ -192,6 +192,17 @@ proc mayaUsdTranslatorExport_SetConvertMaterialsToCheckboxes(string $arg, int $e
     }
 }
 
+proc mayaUsdTranslatorExport_DisableConvertMaterialsToCheckboxes() {
+    // If we are using a legacy material export mode, then disable the checkboxes
+    // since this material export mode is not compatible with multi-material export.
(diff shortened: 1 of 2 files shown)


JGamache-autodesk:
Does not require preflight.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -192,6 +192,17 @@ proc mayaUsdTranslatorExport_SetConvertMaterialsToCheckboxes(string $arg, int $e
...
 }
 
+proc mayaUsdTranslatorExport_DisableConvertMaterialsToCheckboxes() {
+    // If we are using a legacy material export mode, then disable the checkboxes
+    // since this material export mode is not compatible with multi-material export.
+    string $conversions[] = `mayaUSDListShadingModes -export -useRegistryOnly`;
+    for ($conversion in $conversions) {
+        string $opt = `mayaUSDListShadingModes -eo $conversion -useRegistryOnly`;
+        string $widget = $opt + "_ConvertMaterialsToCheckBox";
+        checkBoxGrp -e -v1 false -en false $widget;
```
seando-adsk, line 202: It is possible to get into a state where the string $widget will not be a valid checkBoxGrp? Should we add a if `checkbBoxGrp -exists $widget)`

seando-adsk (approved):

PR #2036, opened 2022-01-27
MAYA-115117 Deactivate prim on pull instead of using exclude feature

kxl-adsk:
Exclude prim is way too expensive for complex scenes since it repopulates the entire render index. We decided to switch to deactivation which reduces the cost to what changed.

### DIFF
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
@@ -46,8 +46,7 @@ namespace {
 
 // We want to display the unloaded prims, so removed UsdPrimIsLoaded from
 // the default UsdPrimDefaultPredicate.
-const Usd_PrimFlagsConjunction MayaUsdPrimDefaultPredicate
-    = UsdPrimIsActive && UsdPrimIsDefined && !UsdPrimIsAbstract;
+const Usd_PrimFlagsConjunction MayaUsdPrimDefaultPredicate = UsdPrimIsDefined && !UsdPrimIsAbstract;
 
 UsdPrimSiblingRange getUSDFilteredChildren(
     const UsdPrim&               prim,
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
```diff
@@ -46,8 +46,7 @@ namespace {
 
 // We want to display the unloaded prims, so removed UsdPrimIsLoaded from
 // the default UsdPrimDefaultPredicate.
-const Usd_PrimFlagsConjunction MayaUsdPrimDefaultPredicate
-    = UsdPrimIsActive && UsdPrimIsDefined && !UsdPrimIsAbstract;
+const Usd_PrimFlagsConjunction MayaUsdPrimDefaultPredicate = UsdPrimIsDefined && !UsdPrimIsAbstract;
```
kxl-adsk, line 49: I wish the predicate in USD be more flexible!

lib/mayaUsd/ufe/ProxyShapeHierarchy.h
```diff
@@ -98,7 +98,8 @@ class MAYAUSD_CORE_PUBLIC ProxyShapeHierarchy : public Ufe::Hierarchy
 
 private:
     const PXR_NS::UsdPrim& getUsdRootPrim() const;
-    Ufe::SceneItemList     createUFEChildList(const PXR_NS::UsdPrimSiblingRange& range) const;
+    Ufe::SceneItemList
+    createUFEChildList(const PXR_NS::UsdPrimSiblingRange& range, bool filterInactive) const;
```
pierrebai-adsk, line 102: Negative boolean tend to be confusing. I would prefere this to be `filter` with true meaning do filter, false meaning do not filter.

pierrebai-adsk: Oh, it mean filter-the-inactive-nodes and not make-filter-inactive... I mis-read the meaning of the variable.

pierrebai-adsk (approved):

ppt-adsk (approved):

PR #2037, opened 2022-01-27
MAYA-114898 Improve Python bindings

pierrebai-adsk:
- Add undo/redo to Python binding of OpUndoItemList.
- Add user args options to mergeToUsd and duplicate Python bindings.

### DIFF
--- a/lib/mayaUsd/python/wrapOpUndoItem.cpp
+++ b/lib/mayaUsd/python/wrapOpUndoItem.cpp
@@ -59,6 +59,18 @@ public:
         _undoItemList.reset();
     }
 
+    void undo()
+    {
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/python/wrapOpUndoItem.cpp
```diff
@@ -59,6 +59,18 @@ class PythonOpUndoItemList
         _undoItemList.reset();
     }
 
+    void undo()
+    {
+        if (_undoItemList)
```
ppt-adsk, line 64: In enter() and exit() you have this check as a TF_VERIFY, should do the same here, no?

ppt-adsk (approved):

PR #2038, opened 2022-01-27
MAYA-121408 Trigger the populate selection when a new instancer is added

williamkrick:
or removed. This covers some of the workflows that change the instance indexing, but not all.


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -506,10 +506,7 @@ void ProxyRenderDelegate::_ClearRenderDelegate()
 
     // reset any version ids or dirty information that doesn't make sense if we clear
     // the render index.
-    _renderTagVersion = 0;
-#ifdef ENABLE_RENDERTAG_VISIBILITY_WORKAROUND
-    _visibilityVersion = 0;
-#endif
+    _changeVersions.reset();
```
williamkrick, line 509: I refactored how these "versions" of things from USD are tracked into a single struct to keep it simpler.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -295,13 +295,49 @@ class ProxyRenderDelegate : public MHWRender::MPxSubSceneOverride
...
+        unsigned int _visibility { 0 };
+        //! The combined instancer index version and instance index change count the last time
+        //! _Execute was called
+        unsigned int _instanceIndex { 0 };
+
+        void sync(const HdChangeTracker& tracker)
+        {
+            _renderTag = tracker.GetRenderTagVersion();
+            _visibility = tracker.GetVisibilityChangeCount();
+            _instanceIndex = tracker.GetInstancerIndexVersion();
```
williamkrick, line 314: GetInstanceIndexVersion doesn't capture all the workflows we care about, but it captures some of them. So at least some things work correctly with this updated version of the code.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
kxl-adsk: Much nicer, thx!

kxl-adsk (approved):

williamkrick:
The failing test is testVP2RenderDelegatePerInstanceInheritedData in Maya 2022 with USD 20.05. That is a test I would expect to potentially fail with these changes, but the differing image is `inheritedDisplayColor_noPxrMtls_unselected.png` which I would not expect to fail. I tried running the test locally in Maya 2022 and USD 20.05, and it passed for me. I'll try running preflight again.

williamkrick:
These changes help with the performance issue in https://github.com/Autodesk/maya-usd/issues/1991.

vlasovi:
@williamkrick, based on the autotest png files, it looks like your PR changes what objects are selected. Why is that? Since your first preflight failed, I just want to make sure we are not introducing here some random behaviour.

williamkrick:
> @williamkrick, based on the autotest png files, it looks like your PR changes what objects are selected. Why is that? Since your first preflight failed, I just want to make sure we are not introducing here some random behaviour.

@vlasovi this change is intentionally changing the behavior of this test. What is happening is USD is telling the VP2RenderDelegate which instances of the instancer are selected. When that instancer is modified by adding or removing instances (by hiding them or causing them to not be drawn in some other way) USD doesn't update its internal information about which instances are selected. That causes the selection to "jump" from one instance to another.

We work around this issue of drawing the selection incorrectly by calling populateSelection again. Calling populateSelection gets the instance to update its internal state about which instance indices are selected. Calling populateSelection is slow, and we don't have complete dirty information telling us when we need to call it and when we don't need to call it.

Originally I had code in place always calling populate selection. Once we discovered it was too slow I created a PR to stop doing the work around (https://github.com/Autodesk/maya-usd/pull/2034). This change uses the incomplete dirty information to get the selection right some of the time, but not all of the time. We're still fast, and we're more correct than we were before. I also have a PR against USD to provide access to the additional dirty information we need to be correct all the time (https://github.com/PixarAnimationStudios/USD/pull/1754).

vlasovi:
> This change uses the incomplete dirty information to get the selection right some of the time, but not all of the time.

So probably that means that we may have some random behaviour here. But since we know about that and since we have a plan to fix that later, I think we can move forward with the change.

vlasovi (dismissed):

kxl-adsk (approved):

PR #2041, opened 2022-01-28
This test crashes! Don't merge becuase all our preflights will fail.

williamkrick:


kxl-adsk:
Converted this to draft. This way nobody can merge it.

kxl-adsk:
But maybe we don't need it anymore?

williamkrick:
Yes I'll close this PR.

I think it would be valuable to add some tests similar to this one for general stability testing. Manipulation of each rprim type we support, and eventually sprims like cameras and lights.

kxl-adsk:
100% agreed!

PR #2042, opened 2022-01-28
Fix order of Pixar's BatchRenderer members

dj-mcg:
This corrects the shutdown order and avoids potential shutdown/reset
crashes


kxl-adsk (approved):

kxl-adsk:
Why there is an image change associated with this code change?

dj-mcg:
This is a baseline update that was needed for recent USD/Hydra changes. It is exercised in the test that was crashing due to the member order issue. I can separate it out into its own PR if you prefer

kxl-adsk:
Thx for the info, I will merge as is.

PR #2043, opened 2022-01-28
MAYA-121316 Skip updating MRenderItem's that we're not going to draw anyway.

williamkrick:
This helps with the selection / de-selection performance issue mentioned in https://github.com/Autodesk/maya-usd/issues/1991. This change doesn't fully fix that issue.


vlasovi (approved):
Looks good to me. Would it make sense to apply the same logic to curves?

williamkrick:
> Looks good to me. Would it make sense to apply the same logic to curves?

@vlasovi in general yes applying the same logic to curves makes sense. However there won't be much of a performance difference there. When the curves change from selected to un-selected they are still drawn, just with a different color. So we still have to set the shader on them and hit the slow code path.

PR #2047, opened 2022-01-28
T bailp/maya 114898/merge to usd export options

pierrebai-adsk:
- Reworked the mayaUsdTranslatorExport function to support specifying the desired section names.
- Can include: specific section name to add it, "all", "none" or the section name prefixed with "!" to remove that section.
- Had to handle the UI updates too.
- Add unit test for export options.

### DIFF
--- a/lib/mayaUsd/commands/PullPushCommands.cpp
+++ b/lib/mayaUsd/commands/PullPushCommands.cpp
@@ -15,7 +15,9 @@
 //
 #include "PullPushCommands.h"
 
+#include <mayaUsd/fileio/jobs/jobArgs.h>
 #include <mayaUsd/fileio/primUpdaterManager.h>
+#include <mayaUsd/fileio/shading/shadingModeRegistry.h>
 #include <mayaUsd/ufe/Utils.h>
 #include <mayaUsd/undo/OpUndoItemRecorder.h>
 #include <mayaUsd/utils/util.h>
@@ -33,6 +35,9 @@ namespace MAYAUSD_NS_DEF {
 namespace ufe {
 
 namespace {
+
+static constexpr auto kExportOptionsFlag = "exo";
+static constexpr auto kExportOptionsFlagLong = "exportOptions";
 
 // Reports an error to the Maya scripting console.
 void reportError(const MString& errorString) { MGlobal::displayError(errorString); }
@@ -118,6 +123,14 @@ MStatus parseDagPathArg(const MArgParser& argParser, int index, MDagPath& output
         return status;
 
     return MDagPath::getAPathTo(obj, outputDagPath);
+}
+
+MString parseTextArg(const MArgDatabase& argData, const char* flag, MString defaultValue)
+{
+    MString value = defaultValue;
+    if (argData.isFlagSet(flag))
+        argData.getFlagArgument(flag, 0, value);
+    return value;
 }
 
 } // namespace
@@ -207,7 +220,12 @@ void* MergeToUsdCommand::creator()
 }
 
 // MPxCommand API to register the command syntax.
-MSyntax MergeToUsdCommand::createSyntax() { return createSyntaxWithUfeArgs(1); }
+MSyntax MergeToUsdCommand::createSyntax()
+{
+    MSyntax syntax = createSyntaxWithUfeArgs(1);
+    syntax.addFlag(kExportOptionsFlag, kExportOptionsFlagLong, MSyntax::kString);
+    return syntax;
+}
 
 // MPxCommand API to execute the command.
 MStatus MergeToUsdCommand::doIt(const MArgList& argList)
@@ -215,6 +233,8 @@ MStatus MergeToUsdCommand::doIt(const MArgList& argList)
     clearResult();
 
     setCommandString(commandName);
+
+    PXR_NS::VtDictionary userArgs;
 
     MStatus    status = MS::kSuccess;
     MArgParser argParser(syntax(), argList, &status);
@@ -236,12 +256,24 @@ MStatus MergeToUsdCommand::doIt(const MArgList& argList)
     if (!PXR_NS::PrimUpdaterManager::readPullInformation(dagPath, fPulledPath))
         return reportError(MS::kInvalidParameter);
 
+    MArgDatabase argData(syntax(), argList, &status);
+    if (status != MS::kSuccess)
+        return reportError(status);
+
+    const MString exportOptions = parseTextArg(argData, kExportOptionsFlag, "");
+    if (exportOptions.length() > 0) {
+        status = PXR_NS::UsdMayaJobExportArgs::GetDictionaryFromEncodedOptions(
+            exportOptions, &userArgs, nullptr);
+        if (status != MS::kSuccess)
+            return status;
+    }
+
     // Scope the undo item recording so we can undo on failure.
     {
         OpUndoItemRecorder undoRecorder(fUndoItemList);
 
         auto& manager = PXR_NS::PrimUpdaterManager::getInstance();
-        status = manager.mergeToUsd(fDagNode, fPulledPath) ? MS::kSuccess : MS::kFailure;
+        status = manager.mergeToUsd(fDagNode, fPulledPath, userArgs) ? MS::kSuccess : MS::kFailure;
     }
 
     // Undo potentially partially-made merge-to-USD on failure.
@@ -325,7 +357,12 @@ void* DuplicateCommand::creator()
 }
 
 // MPxCommand API to register the command syntax.
-MSyntax DuplicateCommand::createSyntax() { return createSyntaxWithUfeArgs(2); }
+MSyntax DuplicateCommand::createSyntax()
+{
+    MSyntax syntax = createSyntaxWithUfeArgs(2);
+    syntax.addFlag(kExportOptionsFlag, kExportOptionsFlagLong, MSyntax::kBoolean);
+    return syntax;
+}
 
 // MPxCommand API to execute the command.
 MStatus DuplicateCommand::doIt(const MArgList& argList)
@@ -333,6 +370,8 @@ MStatus DuplicateCommand::doIt(const MArgList& argList)
     clearResult();
 
     setCommandString(commandName);
+
+    PXR_NS::VtDictionary userArgs;
 
     MStatus    status = MS::kSuccess;
     MArgParser argParser(syntax(), argList, &status);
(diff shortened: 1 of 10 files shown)


lib/mayaUsd/commands/PullPushCommands.cpp
```diff
@@ -34,6 +36,9 @@ namespace ufe {
 
 namespace {
 
+static constexpr auto kExportOptionsFlag = "exo";
+static constexpr auto kExportOptionsFlagLong = "exportOptions";
```
pierrebai-adsk, line 40: This is the new option used in mayaUsdMergeToUsd and mayaUsdDuplicate commands to receive the export options in text form: "value1=x;value2=y"

lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -717,6 +718,133 @@ UsdMayaJobExportArgs UsdMayaJobExportArgs::CreateFromDictionary(
     return UsdMayaJobExportArgs(allUserArgs, dagPaths, timeSamples);
 }
 
+/* static */
+MStatus UsdMayaJobExportArgs::GetDictionaryFromEncodedOptions(
```
pierrebai-adsk, line 722: This whole function was moved from exportTranslator.cpp so that the logic can be shared among multiple callers.

It parses the "value1=x;value2=y" format and creates a VtDictionary from it and optionally the pair of start/end time samples.

plugin/adsk/plugin/exportTranslator.cpp
```diff
@@ -181,10 +66,15 @@ MStatus UsdMayaExportTranslator::writer(
         return MS::kSuccess;
     }
 
-    const std::vector<double> timeSamples
-        = UsdMayaWriteUtil::GetTimeSamples(timeInterval, frameSamples, frameStride);
-    PXR_NS::UsdMayaJobExportArgs jobArgs
-        = PXR_NS::UsdMayaJobExportArgs::CreateFromDictionary(userArgs, dagPaths, timeSamples);
+    VtDictionary        userArgs;
+    std::vector<double> timeSamples;
+    MStatus             status = UsdMayaJobExportArgs::GetDictionaryFromEncodedOptions(
```
pierrebai-adsk, line 71: Calls the new function, removed the moved code above.

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -621,10 +621,86 @@ global proc mayaUsdMenu_pushBackToUSD(string $obj)
...
     }
     if (size($obj)) {
-        mayaUsdMergeToUsd $obj;
+        mayaUsdMergeToUsd -exportOptions "animation=0;shadingMode=none;exportSkels=0;exportSkin=0;exportBlendShapes=0" $obj;
     }
 }
 
+global string $mayaUsdMergeToUsdOptions = "";
+
+global proc mayaUsdPushBackToUSDReceiveExportOptions(string $exportOptions)
```
pierrebai-adsk, line 630: From this point on it is the new merge-to-USD options dialog callbacks.

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -639,7 +715,8 @@ global proc mayaUsdMenu_duplicateToUSD( string $proxy, string $obj )
     string $objLong[] = `ls -l $obj`;
     string $proxyLong[] = `ls -l $proxy`;
     if (size($objLong) && size($proxyLong)) {
-        mayaUsdDuplicate $objLong[0] $proxyLong[0];
+        // TODO: show options.
```
pierrebai-adsk, line 718: The story is only to cover the merge-to-USD. I left a TODO if we want to do the same to the duplicate command.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -15,7 +15,12 @@
 // limitations under the License.
 //
 
+global string $mayaUsdTranslatorExport_SectionNames[];
+
 proc mayaUsdTranslatorExport_SetCheckbox(string $arg, int $enable, string $widget) {
+    if (`checkBoxGrp -q -exists $widget` == 0)
```
pierrebai-adsk, line 21: There are a lot of changes in this files to cope with the fact that some UI are not always created.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -336,10 +446,20 @@ global proc int mayaUsdTranslatorExport (string $parent,
...
+//
+//        In addition to the action, the list of UI sections that are desired can be added
+//        with the syntax "action=section1;section2;..." where the section names can be:
+//            context, output, geometry, materials, animation, advanced
+//
+//        The section name can also be "all", "none" or prefixed with "!":
+//            - When "all" is used, all sections are included.
+//            - When "none" is used, all sections are removed.
+//            - When a "!" prefix is used, the sectionis removed instead of added.
+//
```
pierrebai-adsk, line 462: This is how the section are selected. By default all section are present. Sections can be all removed or used with "none" and "all". Otherwise individual ones are added by name or removed by prefixing their name with ! For example:

- "none;geometry;animations" would keep only the Geometry and Animation sections.
- "all;!geometry" would remove geometry. So would the simpler "!geometry" since by default we keep all sections.

lib/mayaUsd/commands/PullPushCommands.cpp
```diff
@@ -120,6 +125,14 @@ MStatus parseDagPathArg(const MArgParser& argParser, int index, MDagPath& output
     return MDagPath::getAPathTo(obj, outputDagPath);
 }
 
+MString parseTextArg(const MArgDatabase& argData, const char* flag, MString defaultValue)
```
ppt-adsk, line 128: Nit-pick: defaultValue argument should be passed by const ref.

lib/mayaUsd/fileio/jobs/jobArgs.h
```diff
@@ -249,6 +249,16 @@ struct UsdMayaJobExportArgs
         const UsdMayaUtil::MDagPathSet& dagPaths,
         const std::vector<double>&      timeSamples = std::vector<double>());
 
+    /// Fills a VtDictionary from the given text-encoded options.
+    /// Issues runtime errors if some options contains values of the wrong format.
```
ppt-adsk, line 253: "if some options contains" --> "if some option contains", or "if some options contain".

test/lib/mayaUsd/fileio/testMergeToUsd.py
```diff
@@ -347,5 +347,40 @@ def testMergeToUnchangedCylinder(self):
...
 
+    def testMergeWithoutMaterials(self):
+        '''Merge edits on data and not merging the materials.'''
+
+        # Create a stage from a file.
+        testFile = getTestScene("cylinder", "cylinder.usda")
+        testDagPath, stage = mayaUtils.createProxyFromFile(testFile)
+        usdCylPathString = testDagPath + ",/pCylinder1"
+
+        # Verify the orignal scene does not have a look (material) prim.
```
ppt-adsk, line 358: "orignal" --> "original".

ppt-adsk (dismissed):
Nit-picks, but nothing to delay merging.  Clearly Jerry will know best, but otherwise LGTM.

ppt-adsk (approved):

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
seando-adsk: I don't know if there is a side-effect, to adding `-q`, but its not needed since the `-exists` flag is a **create** flag (not a query flag).

ppt-adsk (approved):

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -209,37 +272,52 @@ proc string mayaUsdTranslatorExport_AppendConvertMaterialsTo(string $currentOpti
...
-    optionMenuGrp -e -en 1 exportInstancesPopup;
-    optionMenuGrp -e -en 1 jobContextPopup;
-
-    intFieldGrp -e -en1 1 -en2 1 frameRangeFields;
-
-    floatFieldGrp -e -en1 1 frameStrideField;
-
-    textFieldGrp -e -en 1 frameSampleField;
-    textFieldGrp -e -en 1 parentScopeField;
+    if (stringArrayContains("geometry", $sectionNames)) {
```
seando-adsk, line 279: Are these strings (like "geometry") from a string resource? If yes, then you need to compare against the string from resource.

seando-adsk (changes requested):

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -621,10 +621,86 @@ global proc mayaUsdMenu_pushBackToUSD(string $obj)
...
+        string $sel[] = `ls -sl`;
+        if (size($sel)) {
+            $obj = $sel[0];
+        }
+    }
+    if (size($obj)) {
+        global string $mayaUsdMergeToUsdOptions = "";
+        string $dlgTitle = getMayaUsdString("kMergeToUSDOptionsTitle");
+        string $dlgReturn = `layoutDialog -ui "mayaUsdShowPushBackToUSDOptions" -title $dlgTitle`;
+        if ("Merge" == $dlgReturn) {
```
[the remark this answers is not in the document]
pierrebai-adsk: No. It is a string returned by the call to dismissDialog, it is not localized

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
pierrebai-adsk: No. They are internal strings to communicate what part of the UI to show or not. They are not localized.

plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -621,10 +621,86 @@ global proc mayaUsdMenu_pushBackToUSD(string $obj)
...
+global proc mayaUsdPushBackToUSDReceiveExportOptions(string $exportOptions)
+{
+    global string $mayaUsdMergeToUsdOptions = "";
+    $mayaUsdMergeToUsdOptions = $exportOptions;
+}
+
+global proc mayaUsdPushBackToUSDOptionsAccepted()
+{
+    mayaUsdTranslatorExport("", "query=all;!output", "", "mayaUsdPushBackToUSDReceiveExportOptions");
+    layoutDialog -dismiss "Merge";
```
pierrebai-adsk, line 639: Here is the dismiss with the "Merge"

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
ppt-adsk: I think I agree with Sean here.  The docs (including the docs to the very general "control" command) say that exists is a create flag.  I'd drop the -q, even though in current code it doesn't seem to matter, just to respect the spec and be future-proof.

plugin/adsk/scripts/mayaUsdMenu.mel
seando-adsk: Okay thanks.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
seando-adsk: okay thanks.

seando-adsk (changes requested):

ppt-adsk (approved):

seando-adsk (approved):

PR #2049, opened 2022-01-31
Fix display color not matching the source Lambert materials

JGamache-autodesk:


JGamache-autodesk:
To see where we come from, here is the added unit test without the fix:
image
Row1: 1 plane, 12 separate faces, one lambert per face.
Row2: The plane exported as displayColors, loaded as USD stage using CPV shader.
Row3: 12 planes, one lambert per plane.
Row4: 12 planes exported as displayColors, loaded as USD stage using solid color shader.
Here is the result after the fix:
image
This impacts all scenes that use display colors, which comprises a large number of unit tests, as can be seen in the diff.
Here is the kitchen set scene with a band of updated shading in the middle to give an idea on a known scene:
KitchenSet

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1933,9 +1937,12 @@ void HdVP2Mesh::_UpdateDrawItem(
             if ((colorInterp == HdInterpolationConstant || colorInterp == HdInterpolationInstance)
                 && (alphaInterp == HdInterpolationConstant
                     || alphaInterp == HdInterpolationInstance)) {
-                const GfVec3f& clr3f = MayaUsd::utils::ConvertLinearToMaya(colorArray[0]);
```
JGamache-autodesk, line 1936: Moved upstream. See line 376.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1933,9 +1937,12 @@ void HdVP2Mesh::_UpdateDrawItem(
...
+                const GfVec3f& clr3f = colorArray[0];
                 const MColor   color(clr3f[0], clr3f[1], clr3f[2], alphaArray[0]);
                 shader = _delegate->GetFallbackShader(color);
+                if (shader) {
+                    shader->setParameter("diffuse", 1.0f);
```
JGamache-autodesk, line 1944: We already multiply the exported display color by the `diffuse` coefficient of 0.8. We need not apply it a second time.

williamkrick (approved):

PR #2051, opened 2022-02-01
Remove explicit use of pxr namespace

dj-mcg:


kxl-adsk (approved):
Ehhh…how can we detect this at build time?

dj-mcg:
No worries, it's very easy for me to detect :smile:

PR #2053, opened 2022-02-01
MAYA-114898 fix merge UI option defaults.

pierrebai-adsk:
Also moved the re-activation of the edited prim before the merge in case deactivated prim would cause problems. (I'm frankly surprised it works when it was deactivated... it thought USD would ignore it.)

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -776,6 +776,16 @@ bool PrimUpdaterManager::mergeToUsd(
     if (!isCopy && TF_VERIFY(ufeMayaItem))
         scene.notify(Ufe::ObjectPreDelete(ufeMayaItem));
 
+    if (!isCopy) {
+        FunctionUndoItem::execute(
+            "Merge to Maya rendering inclusion",
+            [pulledPath]() {
+                removeExcludeFromRendering(pulledPath);
+                return true;
+            },
+            [pulledPath]() { return addExcludeFromRendering(pulledPath); });
+    }
+
     // Record all USD modifications in an undo block and item.
     UsdUndoBlock undoBlock(
         &UsdUndoableItemUndoItem::create("Merge to Maya USD data modifications"));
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -776,6 +776,16 @@ bool PrimUpdaterManager::mergeToUsd(
     if (!isCopy && TF_VERIFY(ufeMayaItem))
         scene.notify(Ufe::ObjectPreDelete(ufeMayaItem));
 
+    if (!isCopy) {
+        FunctionUndoItem::execute(
```
kxl-adsk, line 780: I don't understand this change...why are you putting it outside of undo block? How is it related to UI defaults?

pierrebai-adsk: I t has its own undo/redo mechanism (calling removeExclude / addExclude) so it does not need to be in the undo block.

I moved it before the merge becuase it worries me that we are merging data with a unactivated prim. Currently it does not seem to affect SdfCopySpecs nor our current diff code, but it may in the future, so Iit seemes ore natural to me to reactivate the prim before trying to merge data in it.

kxl-adsk: You are assuming that this method doesn't need the undo block. This is true today, doesn't have to be true tomorrow. If we don't have an actual issue with the code (reproduced with unit test would be ideal), then let's not do it.
If you can write a unit test to showcase an issue, then let's make this change but in a separate PR.

kxl-adsk (changes requested):
Take out changes to primUpdaterManager.cpp from this PR.

ppt-adsk (approved):

kxl-adsk (approved):

pierrebai-adsk:
Sorry, I discovered I was using 0 instead of none for two options/. It worked, but it caused warnings.

ppt-adsk (approved):

PR #2054, opened 2022-02-01
Cache to USD dialog for Maya reference prim.

ppt-adsk:


kxl-adsk (approved):

seando-adsk (approved):

lib/mayaUsd/fileio/primUpdater.h
```diff
@@ -115,6 +118,9 @@ class UsdMayaPrimUpdater
     MAYAUSD_CORE_PUBLIC
     UsdPrim getUsdPrim() const;
 
+    MAYAUSD_CORE_PUBLIC
+    const UsdMayaPrimUpdaterContext* getContext() const { return _context; }
```
pierrebai-adsk, line 122: Is there a reason why it is returned by pointer? It seems to me it should always be non-null. All callers will need to check for nullity.

ppt-adsk: Great point, could be a reference.  I don't think we should block the pull request on this, but I'll make a note for the follow-up.

ppt-adsk:
Only failing test is (__main__.testUtilsSelectabilityPointInstanceSelection) for configuration 2011_2022_python2_pxrusd-min_interactive_windows, which is a known, unrelated issue, therefore marking as ready to merge.

PR #2058, opened 2022-02-02
Skip instance transforms and colors update when the values haven't changed.

williamkrick:


lib/mayaUsd/render/vp2RenderDelegate/draw_item.h
```diff
@@ -59,6 +59,12 @@ class HdVP2DrawItem final : public HdDrawItem
         //! World matrix of the render item.
         MMatrix _worldMatrix;
 
+        //! Instance transforms for the render item
+        std::shared_ptr<MMatrixArray> _instanceTransforms;
+
+        //! Instance colors for the render item
+        std::shared_ptr<MFloatArray> _instanceColors;
+
```
williamkrick, line 67: Store these as shared_ptr so I can easily transfer ownership of the data from the commit state to the draw item data.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -121,9 +121,8 @@ struct CommitState
     bool Empty()
     {
         return _indexBufferData == nullptr && _shader == nullptr && _enabled == nullptr
-            && !_geometryDirty && _boundingBox == nullptr && !_renderItemData._usingInstancedDraw
-            && _instanceTransforms.length() == 0 && _ufeIdentifiers.length() == 0
-            && _worldMatrix == nullptr;
+            && !_geometryDirty && _boundingBox == nullptr && !_instanceTransforms
+            && !_instanceColors && _ufeIdentifiers.length() == 0 && _worldMatrix == nullptr;
```
williamkrick, line 125: Remove the _usingInstancedDraw check, because we no longer always update instancing.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1774,13 +1766,19 @@ void HdVP2Mesh::_UpdateDrawItem(
...
                 stateToCommit._enabled = &drawItemData._enabled;
             }
         }
     }
 
+    // The above checks have captured the cases where the enabled state of the item
+    // changes. If the item is still not enabled, skip the rest of the update. Further
+    // code can cause the item to be disabled, but it can't change the item from enabled
+    // to disabled.
+    if (!drawItemData._enabled)
```
williamkrick, line 1779: This fixes an assertion I was sometimes seeing at line 2541. Sometime an render item that is hidden could have other changes and remain hidden. renderItemData._enable would match the new enable value and so the early exit would be missed.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -2241,22 +2240,70 @@ void HdVP2Mesh::_UpdateDrawItem(
...
+        if (stateToCommit._instanceTransforms && drawItemData._instanceTransforms) {
+            instanceTransformsChanged
+                = (stateToCommit._instanceTransforms->length()
+                   != drawItemData._instanceTransforms->length());
+            for (unsigned int index = 0;
+                 index < stateToCommit._instanceTransforms->length() && !instanceTransformsChanged;
+                 index++) {
+                instanceTransformsChanged
+                    = ((*stateToCommit._instanceTransforms)[index]
+                       != (*drawItemData._instanceTransforms)[index]);
```
williamkrick, line 2273: As painfully slow as this is, setting the data redundantly into Maya is much slower because Maya rebuilding the instanced render item in VP2 (interacting with the graphics driver). Here we're doing work on the CPU, and this code is threaded so the scalability is still okay.

test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateSelection.py
```diff
@@ -130,12 +130,12 @@ def testSelection(self):
         cmds.modelEditor('modelPanel4', e=True, wireframeOnShaded=False, displayLights='default')
         self._selectionTest('', usdCube, usdCylinder, proxyDagPath, 'smoothShaded')
 
-        # Test smooth shaded
+        # Test wireframe on shaded
```
williamkrick, line 133: I noticed these comment typos when I was working on these changes getting the test to work correctly. The test behavior itself does not change.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
vlasovi: @williamkrick can we use memcmp to compare two buffers here? Theoretically it may be faster.

vlasovi (approved):

williamkrick:
This helps with https://github.com/Autodesk/maya-usd/issues/1991.

kxl-adsk:
@williamkrick can you please resolve conflicts?

williamkrick:
I need to re-evaluate this one. The conflicts are with the changes I had to skip updates for invisible render items and I'm not sure how well this code works without those changes. I will close this PR for now until I can circle back to it.

PR #2059, opened 2022-02-02
MAYA106075 - Introducing color cache to improve performance

vlasovi:
Introducing color cache to improve performance of ProxyRenderDelegate::GetSelectionHighlightColor and ProxyRenderDelegate::GetCurveDefaultColor.
Make sure the cache is accessed from inside the mutex to prevent races.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -1485,7 +1485,12 @@ const MColor& ProxyRenderDelegate::GetWireframeColor() const { return _wireframe
 
 GfVec3f ProxyRenderDelegate::GetCurveDefaultColor()
 {
-    // Enter the mutex and check the cache
+    // Check the cache. It is safe since _dormantCurveColorCache.second is atomic
+    if (_dormantCurveColorCache.second == _frameCounter) {
+        return _dormantCurveColorCache.first;
+    }
+
+    // Enter the mutex and check the cache again
     std::lock_guard<std::mutex> mutexGuard(_mayaCommandEngineMutex);
     if (_dormantCurveColorCache.second == _frameCounter) {
         return _dormantCurveColorCache.first;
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -1484,21 +1485,28 @@ const MColor& ProxyRenderDelegate::GetWireframeColor() const { return _wireframe
...
 {
-    MDoubleArray curveColorResult;
-    {
-        std::lock_guard<std::mutex> mutexGuard(_mayaCommandEngineMutex);
-        MGlobal::executeCommand(
-            "int $index = `displayColor -q -dormant \"curve\"`; colorIndex -q $index;",
-            curveColorResult);
+    // Enter the mutex and check the cache
+    std::lock_guard<std::mutex> mutexGuard(_mayaCommandEngineMutex);
+    if (_dormantCurveColorCache.second == _frameCounter) {
```
williamkrick, line 1490: I think if we make _dormantCurveColorCache.second a std::atomic<uint64_t> then we could move acquiring the mutex to be after this check to see if we already have already read the value this frame. That would prevent lots of mutex acquires which might be faster.

In the case where the test fails we'd have to acquire the mutex and re-test to avoid reading the value two or more times, but that is okay.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -1508,16 +1516,31 @@ MColor ProxyRenderDelegate::GetSelectionHighlightColor(const TfToken& className)
...
+        TF_WARN(
+            "ProxyRenderDelegate::GetSelectionHighlightColor - unsupported class: '%s'",
+            className.GetString().c_str());
+        return kDefaultActiveColor;
+    }
+
+    // Enter the mutex and check the cache
+    std::lock_guard<std::mutex> mutexGuard(_mayaCommandEngineMutex);
+    if (colorCache->second == _frameCounter) {
+        return colorCache->first;
```
williamkrick, line 1543: Could use a similar lock-free pattern here.

williamkrick (approved):
Did you test the performance of the changes?

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
vlasovi: The std::atomic<uint64_t> approach may be faster but not necessarily. We can try it if we still need to after this change.

vlasovi:
I haven't tested the performance on a large scene. I have none at the moment. I only tested my custom scenes and confirmed that the feature is working as expected.

williamkrick (approved):

PR #2060, opened 2022-02-02
MAYA-121514 fix leaking node due to undo items being left behind.

pierrebai-adsk:
- Use an OpUndoItemList to capture the undo items of mergeToUsd in the test.
- Add a new notice sent before scene change, including Maya exit.
- Clear the recorded global undo items that were not picked up by any command on exit.

### DIFF
--- a/lib/mayaUsd/listeners/notice.cpp
+++ b/lib/mayaUsd/listeners/notice.cpp
@@ -40,6 +40,13 @@ static void _OnMayaNewOrOpenSceneCallback(void* /*clientData*/)
     UsdMayaSceneResetNotice().Send();
 }
 
+static int _beforeSceneResetRegistrationCount = 0;
+
+static void _OnMayaBeforeResetCallback(void* /*clientData*/)
+{
+    UsdMayaBeforeSceneResetNotice().Send();
+}
+
 static int _exitRegistrationCount = 0;
 
 static void _OnMayaExitCallback(void* /*clientData*/)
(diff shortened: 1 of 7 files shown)


lib/mayaUsd/listeners/notice.cpp
```diff
@@ -29,28 +29,32 @@ namespace {
...
+
+static void _beforeNewSceneCallback(void* /*clientData*/)
 {
     // kBeforeFileRead messages are emitted when importing/referencing files,
     // which we don't consider a "scene reset".
     if (MFileIO::isImportingFile() || MFileIO::isReferencingFile()) {
         return;
     }
 
+    UsdMayaSceneBeforeResetNotice().Send();
```
ppt-adsk, line 42: I'm a bit puzzled by this notification.  Unless I'm mistaken, this is the only place where it's sent, and it's followed immediately by a UsdMayaSceneResetNotice notification.  However, in OpUndoItemList.cpp both notifications do exactly the same thing, which seems redundant: you'll just call OpUndoItemList::instance().clear() twice.  Am I missing something?

lib/mayaUsd/listeners/notice.cpp
```diff
@@ -29,28 +29,32 @@ namespace {
 // to be removed.
 static int _newOrOpenRegistrationCount = 0;
 
-static void _OnMayaNewOrOpenSceneCallback(void* /*clientData*/)
+static void _afterNewSceneCallback(void* /*clientData*/) { UsdMayaSceneResetNotice().Send(); }
```
ppt-adsk, line 32: Why do we need both a before new and after new callback?  What USD changes can occur between those two points in execution?

lib/mayaUsd/listeners/notice.cpp
pierrebai-adsk: No, but I did not want to change old user of the previously existing notification and wanted the new notification to cover all "before"cases. As you mention, there will be a 2nd call, but the implementation will clear a variable that is already clear, so the cost is minimal.

I did not want to tie the new "Before" notification to our specific use case and thought the double call innocuous.

lib/mayaUsd/listeners/notice.cpp
pierrebai-adsk: Unfortunately, for the undo items clearing, we need to clear them earlier than when the after-new-scene gets triggered. But we also need to clear them for file-load and new-scene. You are right that for the undo items clearing, we would not need the after-new-scene callback, but they are all tied together now. As I said previously, the redundant 2nd call that will happen is harmless.

kxl-adsk (commented):
I didn't review the initial undo/redo support for editAsMaya and I have some basic questions. I will take the conversation offline.

kxl-adsk (approved):

pierrebai-adsk:
Unfortunately, different code in the plugin need different timing on scene reset: undo clearing needs to be done before scene reset, but imaging delegate need to be after the scene reset. That is why we need to make both available.

lib/mayaUsd/undo/OpUndoItemList.cpp
```diff
@@ -33,12 +33,24 @@ struct OnSceneResetListener : public PXR_NS::TfWeakBase
     {
         PXR_NS::TfWeakPtr<OnSceneResetListener> self(this);
         PXR_NS::TfNotice::Register(self, &OnSceneResetListener::OnSceneReset);
+        PXR_NS::TfNotice::Register(self, &OnSceneResetListener::OnMayaAboutToExit);
+        PXR_NS::TfNotice::Register(self, &OnSceneResetListener::OnMayaAboutToResetScene);
```
kxl-adsk, line 37: Combining this registration into one call made sense to me in your original change.

kxl-adsk (approved):
Perfect...this is what we wanted to learn.

lib/mayaUsd/undo/OpUndoItemList.cpp
pierrebai-adsk: Yeah, OTOH, I can imagine some case where they would only need to do something on exit, so having them separate is not a big cost and provides the flexibility.

ppt-adsk (approved):

PR #2065, opened 2022-02-03
[al] Use quaternion form to compare rotation angles

csyshing:
The PR changed the rotation comparison to use quaternion instead of raw value (rx, ry, rz) in Euler space.

In USD, rotation ops, "xformOp:rotateXYZ" etc. are in Euler space, we had an issue in production that the rotation was twisted but in fact they should be treated as the same, e.g. `(180, -2.3171844, 180) == (0, 182.31718, 0)`, thus converting to quaternion for comparison.

### DIFF
--- a/plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/nodes/test_TransformMatrix.cpp
+++ b/plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/nodes/test_TransformMatrix.cpp
@@ -23,6 +23,7 @@
 #include <pxr/usd/sdf/types.h>
 #include <pxr/usd/usd/attribute.h>
 #include <pxr/usd/usd/stage.h>
+#include <pxr/usd/usdGeom/sphere.h>
 #include <pxr/usd/usdGeom/xform.h>
 #include <pxr/usd/usdGeom/xformCommonAPI.h>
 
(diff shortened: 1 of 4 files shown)


kxl-adsk (dismissed):

plugin/al/plugin/AL_USDMayaTestPlugin/AL/usdmaya/nodes/test_TransformMatrix.cpp
```diff
@@ -1344,3 +1345,148 @@ TEST(Transform, getTimeCode)
...
+        // Explicitly rotate X axis by 360 degree, there should be no "over" on USD
+        // Notice that we only set the X component here
+        MPlug(xformMobj, MPxTransform::rotateX).setValue(M_PI * 2);
+        ASSERT_TRUE(stage->GetEditTarget().GetLayer()->IsEmpty());
+    }
+
+    {
+        auto                                     xformPrim = stage->GetPrimAtPath(xformPath);
+        AL::usdmaya::nodes::TransformationMatrix tm;
+        tm.setPrim(xformPrim, nullptr);
```
kxl-adsk, line 1425: TransformationMatrix::setPrim will have to be exported (by adding AL_USDMAYA_PUBLIC) for Windows PF to pass.

csyshing: @kxl-adsk , I have updated the commit to export those symbols.

kxl-adsk (approved):

kxl-adsk (approved):

PR #2067, opened 2022-02-03
Export: read plug values at current time

JGamache-autodesk:
Used to read whatever was there, which might be stale data. This fixes invalid/incomplete animation export.

### DIFF
--- a/lib/mayaUsd/fileio/utils/writeUtil.cpp
+++ b/lib/mayaUsd/fileio/utils/writeUtil.cpp
@@ -41,6 +41,8 @@
 #include <pxr/usd/usdRi/statementsAPI.h>
 #include <pxr/usd/usdUtils/sparseValueWriter.h>
 
+#include <maya/MAnimControl.h>
+#include <maya/MDGContextGuard.h>
 #include <maya/MDoubleArray.h>
 #include <maya/MFnAttribute.h>
 #include <maya/MFnDependencyNode.h>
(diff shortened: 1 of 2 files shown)


test/lib/mayaUsd/fileio/testSchemaApiAdaptor.py
```diff
@@ -373,11 +373,11 @@ def testComplexAdaptation(self):
...
 
         # Modify the velocity so it can be exported to the RBD Physics schema.
-        cmds.setKeyframe(bulletPath, at="initialVelocityX", t=0, v=5.0)
+        cmds.setKeyframe(bulletPath, at="initialVelocityX", t=1, v=5.0)
         cmds.setKeyframe(bulletPath, at="initialVelocityX", t=10, v=50.0)
-        cmds.setKeyframe(bulletPath, at="initialVelocityY", t=0, v=6.0)
+        cmds.setKeyframe(bulletPath, at="initialVelocityY", t=1, v=6.0)
         cmds.setKeyframe(bulletPath, at="initialVelocityY", t=10, v=60.0)
-        cmds.setKeyframe(bulletPath, at="initialVelocityZ", t=0, v=7.0)
+        cmds.setKeyframe(bulletPath, at="initialVelocityZ", t=1, v=7.0)
```
JGamache-autodesk, line 380: There is a test at line 435 that the exported value is (5,6,7) at t=1. If we really pull at the current time, this test can only succeed if the keyframe is a t=1. It used to pass, showing that the export was reading stale values. With the fix I need to keyframe correctly to get the right result.

lib/mayaUsd/fileio/utils/writeUtil.cpp
```diff
@@ -647,7 +649,9 @@ bool UsdMayaWriteUtil::SetUsdAttr(
         return true;
     }
 
-    VtValue val = GetVtValue(attrPlug, usdAttr.GetTypeName());
+    MDGContext      dgContext(MAnimControl::currentTime());
```
kxl-adsk, line 652: Evaluating in the time context can be very expensive and in some cases incorrect. I'm not feeling comfortable changing this code.

PR #2068, opened 2022-02-03
Fix edit router parent test.

ppt-adsk:


test/lib/ufe/testParentCmd.py
```diff
@@ -1175,48 +1170,84 @@ def checkAfter(a, b, c, e, f):
...
+
+            # Here, prim is the parent prim.
+            prim = context.get('prim')
+            self.assertIsNot(prim, None)
+            self.assertFalse(len(prim.GetStage().GetRootLayer().subLayerPaths)==0)
+            layerId = prim.GetStage().GetRootLayer().subLayerPaths[0]
+            layer = Sdf.Layer.Find(layerId)
+            # Make sure the destination exists in the target layer, otherwise
+            # SdfCopySpec will error.
+            Sdf.JustCreatePrimInLayer(layer, prim.GetPath())
```
ppt-adsk, line 1216: This is the crucial change in the test: the destination layer is completely empty, but SdfCopySpec requires a destination prim.  Just create an empty prim in the destination layer.

test/lib/ufe/testParentCmd.py
```diff
@@ -1175,48 +1170,84 @@ def checkAfter(a, b, c, e, f):
...
+            routingData['layer'] = layerId
+
+        # Register our edit router which directs the parent edit to
+        # higher-priority layer B, which is not the edit target.
         mayaUsd.lib.registerEditRouter('parent', firstSubLayer)
 
+        # Check that layer B is empty.
```
ppt-adsk, line 1223: Various test improvements: ensure the destination layer is empty before parenting, that it has the expected over's after parenting, etc.

test/lib/ufe/testParentCmd.py
```diff
@@ -1175,48 +1170,84 @@ def checkAfter(a, b, c, e, f):
...
+        # item, as its path has changed.
+        b = ufe.Hierarchy.createItem(ufe.PathString.path(psPathStr+',/C/B'))
+        self.assertEqual(ufe.Hierarchy.hierarchy(b).parent(), c)
+
+        # Check that layer B now has the parent overs.
+        self.assertEqual(filterUsdStr(bSubLayer.ExportToString()),
+                         'over "C"\n{\n    def Xform "B"\n    {\n    }\n}')
+
+        # Restore default edit router.
+        mayaUsd.lib.registerDefaultEditRouter('parent')
```
ppt-adsk, line 1250: Restore default edit router on test exit, otherwise other parenting tests will fail.

lib/mayaUsd/utils/editRouter.h
```diff
@@ -84,7 +84,13 @@ EditRouter::Ptr getEditRouter(const PXR_NS::TfToken& operation);
 MAYAUSD_CORE_PUBLIC
 void registerEditRouter(const PXR_NS::TfToken& operation, const EditRouter::Ptr& editRouter);
 
+// Register the default edit router for the argument operation, overwriting
+// the currently-registered edit router.  Returns false if no such default
+// exists.
+MAYAUSD_CORE_PUBLIC
+bool registerDefaultEditRouter(const PXR_NS::TfToken& operation);
```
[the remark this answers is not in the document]
kxl-adsk: How about we call it restoreDefaultEditRouter? I needed to read the understand what this method is supposed to do.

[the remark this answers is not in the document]
ppt-adsk: Works for me.

kxl-adsk (dismissed):

lib/mayaUsd/python/wrapEditRouter.cpp
```diff
@@ -74,4 +74,9 @@ void wrapEditRouter()
             return MayaUsd::registerEditRouter(
                 operation, std::make_shared<PyEditRouter>(editRouter));
         });
+
+    def(
+        "restoreDefaultEditRouter", +[](const PXR_NS::TfToken& operation) {
```
boudrey, line 79: Why not just setting the address of the function to call instead of creating a lambda function that is calling the function directly?

boudrey (dismissed):

lib/mayaUsd/python/wrapEditRouter.cpp
ppt-adsk: Quite right!  I just mindlessly copy-pasted the code above, there's no need for a lambda in this case, good catch.

kxl-adsk (approved):

PR #2071, opened 2022-02-04
Revert "Skip updating MRenderItem's that we're not going to draw anyway."

williamkrick:
This reverts commit f814ca75764c8be1c5403bde541a941a485dda3c.


williamkrick:
These changes are causing a crash in Maya with consolidated world. I'm not sure exactly what is going wrong, but whatever the fixes are to this idea are probably going to be in Maya. Revert the changes for now.

kxl-adsk (approved):

PR #2075, opened 2022-02-08
[al] Removed tracking selectability and lock metadata

csyshing:
Previously, AL_USDMaya tracks selectability and locking metadata, but there are few issues for this approach:

 - selectability / locking status are easily becoming out of sync when user toggling the optionVar (from Maya menu);
 - setting selectability and locking metadata can affect each other
 - changes received from UsdTransation does not carry detail information about if metadata has been touched (see `ProxyShape::onTransactionNotice()`), thus also making the status out of sync easily
 
This PR is meant to change the logic to be "query selectability / locking when needed", it is the most straightforward and accurate way to track selectability and locking.

Notice that this PR affects the "transform chain" (parent transforms that are created automatically when selecting geometry prims, or for those prims that have been marked as "importByDefault" and "needsTransformParent" in schema translators or Python translators) regardless of the viewport mode.


kxl-adsk (approved):
LGTM. FWIW. I wonder if you had a chance to check what was done for selectability and attribute locking in base proxy shape?

csyshing:
> I wonder if you had a chance to check what was done for selectability and attribute locking in base proxy shape?

Yes @kxl-adsk , we are transitioning to use the features provided from the base class, instead of relying on AL_USDMaya. This PR is just yet another one to align internal and external code base.

PR #2092, opened 2022-02-10
MAYA-115079: USD: MayaUsd test failures with USD built with AssetResolver 2

seando-adsk:
#### MAYA-115079: USD: MayaUsd test failures with USD built with AssetResolver 2
* Code + test fixes for Usd compiled with AssetResolver v2.
* Minor test fixes when missing Pixar plugin.

### DIFF
--- a/test/lib/mayaUsd/fileio/testAddMayaReference.py
+++ b/test/lib/mayaUsd/fileio/testAddMayaReference.py
@@ -28,7 +28,7 @@ from maya import standalone
 import mayaUsdAddMayaReference
 import mayaUsdMayaReferenceUtils as mayaRefUtils
 
-import unittest
+import os, unittest
 
 class AddMayaReferenceTestCase(unittest.TestCase):
     '''Test Add Maya Reference.
@@ -135,7 +135,7 @@ class AddMayaReferenceTestCase(unittest.TestCase):
         # and that it has the expected metadata.
         attr = mayaRefPrim.GetAttribute('mayaReference')
         self.assertTrue(attr.IsValid())
-        self.assertEqual(attr.Get().resolvedPath, self.mayaSceneStr)
+        self.assertTrue(os.path.samefile(attr.Get().resolvedPath, self.mayaSceneStr))
         attr = mayaRefPrim.GetAttribute('mayaNamespace')
         self.assertTrue(attr.IsValid())
         self.assertEqual(attr.Get(), self.kDefaultNamespace)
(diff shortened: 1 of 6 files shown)


test/lib/mayaUsd/fileio/testAddMayaReference.py
```diff
@@ -135,7 +135,7 @@ def testDefineInVariant(self):
         # and that it has the expected metadata.
         attr = mayaRefPrim.GetAttribute('mayaReference')
         self.assertTrue(attr.IsValid())
-        self.assertEqual(attr.Get().resolvedPath, self.mayaSceneStr)
+        self.assertTrue(os.path.samefile(attr.Get().resolvedPath, self.mayaSceneStr))
```
seando-adsk, line 138: This test is failing with AssetResolver v2 because the driver letter is lower vs upper case. So the strings don't compare. Thus just added the ```os.path.samefile()``` condition.

test/lib/mayaUsd/fileio/testSchemaApiAdaptor.py
```diff
@@ -17,13 +17,20 @@
...
+from pxr import Tf, Gf, Usd
 
 from maya import cmds
 import maya.api.OpenMaya as om
 import maya.api.OpenMayaAnim as oma
 from maya import standalone
 
+HAS_USDMAYA = False
+try:
+    from pxr import UsdMaya
```
seando-adsk, line 29: This python module is not present if you don't build the Pixar plugin. It is used by **one** of the tests in this file (which exists in the maya-usd core). So skip that test if we don't have UsdMaya.

test/lib/usd/translators/CMakeLists.txt
```diff
@@ -91,6 +90,13 @@ if(BUILD_PXR_PLUGIN)
     list(APPEND TEST_SCRIPT_FILES
         testUsdExportUVTransforms.py
     )
+
+    # This test uses the file "asset.ma" which
+    # requires the plugin "pxrUsdPreviewSurface" that is built by the
+    # Pixar plugin.
+    list(APPEND TEST_SCRIPT_FILES
+        testUsdExportUsdzPackage.py
+    )
```
seando-adsk, line 99: Same situation as the test above.

test/lib/mayaUsd/fileio/testSchemaApiAdaptor.py
kxl-adsk: Why the issue came up with AR2.0? I can’t find the link?

test/lib/usd/translators/CMakeLists.txt
kxl-adsk: Same question as above…

kxl-adsk (approved):

PR #2094, opened 2022-02-11
Fixed recursive calls to bound Python translators under Python 3

J-Mo63:
Maya was spitting out `RecursionError`s when enabling or disabling certain Python translated objects. It was also noted that similar `RecursionError`s would be generated from a handful of other operations concerning the Python translators. It was later identified that the error would only occur in Python 3.

It was found that the fallback method (to be called in the event no Python override existed for it) was not properly wired up in such a way that it would short the method and still technically function correctly under Python 2:

```cpp
std::size_t generateUniqueKey(const UsdPrim& prim) const override
{
    if (Override o = GetOverride("generateUniqueKey")) {
        auto res = std::function<boost::python::object(const UsdPrim&)>(
            TfPyCall<boost::python::object>(o))(prim);
        if (!res) {
            return 0;
        }
        TfPyLock                            pyLock;
        boost::python::str                  strObj(res);
        boost::python::extract<std::string> strValue(strObj);
        if (strValue.check()) {
            return std::hash<std::string> {}(strValue);
        }
    }
    return 0;
}

...

.def(
    "generateUniqueKey",
    &TranslatorBase::generateUniqueKey,
    &TranslatorBaseWrapper::generateUniqueKey) // <- fallback method, calling wrapper override
```

Default methods were added to properly redirect to the base class as so:
```cpp
std::size_t default_generateUniqueKey(const UsdPrim& prim) const
{
    return base_t::generateUniqueKey(prim);
}

std::size_t generateUniqueKey(const UsdPrim& prim) const override

...

.def(
    "generateUniqueKey",
    &TranslatorBase::generateUniqueKey,
    &TranslatorBaseWrapper::default_generateUniqueKey) // <- fallback method, calling default method to base
```

This solution has been applied to each of the bound methods and seems to now be running without recursion errors (which were previously silenced) under both Python versions.

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/wrapTranslatorBase.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/wrapTranslatorBase.cpp
@@ -96,6 +96,11 @@ public:
         return this->CallPureVirtual<TfType>("getTranslatedType")();
     }
 
+    std::size_t default_generateUniqueKey(const UsdPrim& prim) const
+    {
+        return base_t::generateUniqueKey(prim);
+    }
+
     std::size_t generateUniqueKey(const UsdPrim& prim) const override
     {
         if (Override o = GetOverride("generateUniqueKey")) {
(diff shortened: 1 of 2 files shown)


kxl-adsk (approved):
FWIW. Your PR matches the pattern we have been using for core translators.

kxl-adsk (approved):

PR #2095, opened 2022-02-11
Support for orphaned objects

kxl-adsk:


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -932,7 +932,17 @@ bool PrimUpdaterManager::canEditAsMaya(const Ufe::Path& path) const
     return updater ? updater->canEditAsMaya() : false;
 }
 
-bool PrimUpdaterManager::discardEdits(const Ufe::Path& pulledPath)
+bool PrimUpdaterManager::discardEdits(const MDagPath& dagPath)
```
kxl-adsk, line 935: Orphaned objects (i.e. maya's objects that were created as a result of pull operation, but no longer have a valid prim associated with it) won't be able to provide a prim path. Switching to MDagPath to make it more evident since before we were expecting it always to be a two-segment path.

lib/mayaUsd/resources/scripts/mayaUsdMergeToUsd.py
```diff
@@ -58,14 +58,22 @@ def createMayaReferenceMenuItem(dagPath, precedingItem):
...
 
-    # If the pulled prim isn't a MayaReference, not our responsibility.
-    if prim.GetTypeName() != 'MayaReference':
-        return ''
-
-    cmd = partial(mayaUsdCacheMayaReference.cacheDialog, fullDagPath, prim)
-    cacheLabel = getMayaUsdLibString('kMenuCacheToUsd')
-    return cmds.menuItem(label=cacheLabel, insertAfter=precedingItem, 
-                         image="cache_to_USD.png", command=cmd)
+    # If the pulled prim doesn't exist anymore, we won't delegate the reponsibility to
```
kxl-adsk, line 61: This is something we will iterate a bit more with @womanyoyoyo-adsk. As the code is written now, we are relying on the prim type to decide on the menu items to display. It doesn't work for orphaned objects so we may need to change strategy. The priority of this work is undetermined at this point and this is why we left it outside of the scope for this PR and MAYA-121585.

lib/mayaUsd/resources/scripts/mayaUsdMergeToUsd.py
```diff
@@ -58,14 +58,22 @@ def createMayaReferenceMenuItem(dagPath, precedingItem):
...
+            return ''
+
+        cmd = partial(mayaUsdCacheMayaReference.cacheDialog, fullDagPath, prim)
+        cacheLabel = getMayaUsdLibString('kMenuCacheToUsd')
+        return cmds.menuItem(label=cacheLabel, insertAfter=precedingItem,
+                             image="cache_to_USD.png", command=cmd, enable=1)
+    else:
+        menuLabel = getMayaUsdLibString('kMenuMergeMayaEdits')
+        return cmds.menuItem(label=menuLabel, insertAfter=precedingItem,
+                             image="merge_to_USD.png", enable=0)
```
pierrebai-adsk, line 76: Is there a reason the icon is different when disabled? cache_to_USD vs merge_to_USD?

pierrebai-adsk (dismissed):
I only have that question about the icon and label on the disabled menu item.

lib/mayaUsd/resources/scripts/mayaUsdMergeToUsd.py
kxl-adsk: Yes, that's by the design...and that's what I was describing above https://github.com/Autodesk/maya-usd/pull/2095#discussion_r804535381

kxl-adsk:
Needed to merge the latest dev to resolve conflicts after UFE versioning cleanup. @pierrebai-adsk and @seando-adsk please re/approve.

seando-adsk (approved):

Issue #1688, opened 2021-09-07
[MAYA-114855] Crash when interacting with outliner after unloading payload


eddiehoyle:
**Describe the bug**
Maya crashes if user interacts with USD hierarchy in Maya's outliner if a prim with payloads is unloaded while expanded.

**Steps to reproduce**

**Note:** This repro has been manually copied and typed out (internal network restrictions, sorry) and is untested.

1. Save these layers to disk in the same folder:
   ```
   // entry.usda
   #usda 1.0
   (
       defaultPrim = "entry"
   )
   def "entry" (
       kind = "component"
       append payload = @./payload.usda@</entry>
   ) {}
   ```
   ```
   // payload.usda
   #usda 1.0
   over "entry" {
       def Cube "cube" {
       }
   }
   ```
2. In Maya, create proxy from file "entry.usda" with payloads enabled and expand the outliner hierarchy. You should see the following:
   ```
   proxy             # transform
       shape         # shape
           entry     # /entry
               cube  # /entry/cube
   ```
3. Unload payload prim
   ```python
   import mayaUsd
   from maya import cmds
   stage = mayaUsd.ufe.getStage("|world{}".format(cmds.ls(type="mayaUsdProxyShape", long=True)[0]))
   entry = stage.GetPrimAtPath("/entry")
   entry.Unload()
   ```
4. Try to select, or collapse USD hierarchy in Maya outliner
5. *crash*

Error message:
```
maya.bin crashed. FATAL ERROR: Used expired 'Cube' prim </entry/cube>
```

If the hierarchy is collapsed before calling `Unload`, no crash occurs. My guess the outliner needs to be refreshed.

**Expected behavior**
The outliner should refresh, display the correct USD structure (payload pruned) and should not crash.

**Specs:**
- Centos 7
- gcc 6.3.1
- Maya 2020
- Maya USD 0.10.0
- Usd 20.11

santosg87:
@eddiehoyle I am trying to reproduce this one for a bit of time now, but no luck. 

this seems to behave as expected on my side. are you able to test this on the latest mayaUSD build to confirm?

santosg87:
@eddiehoyle thanks for the info! this seems to be indeed something specific to 2020 only. I am able to repro this on 2020.4, but not on 2022 builds. 

given that this is something specific to 2020, my guess is that there is some UFE support missing in there.

santosg87:
[...]

PR #2103, opened 2022-02-14
MAYA-114855 - Ufe v1: Crash when interacting with Outliner after unloading payload

seando-adsk:
MAYA-114855 - [GIthub #1688] crashes when a user tries to interact with the outliner after unloading a payload.

* Use helpers to mimic some specific Ufe v2 notifs in Ufe v1.

### DIFF
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -386,8 +386,7 @@ void StagesSubject::stageChanged(
 #ifndef MAYA_ENABLE_NEW_PRIM_DELETE
             }
 #endif
-        }
-        else if (!prim.IsValid() && !InPathChange::inPathChange()) {
+        } else if (!prim.IsValid() && !InPathChange::inPathChange()) {
             Ufe::SceneItem::Ptr sceneItem = Ufe::Hierarchy::createItem(ufePath);
             if (!sceneItem || InAddOrDeleteOperation::inAddOrDeleteOperation()) {
                 sendObjectDestroyed(ufePath);


lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -133,6 +133,43 @@ void sendObjectPostDelete(const Ufe::SceneItem::Ptr& sceneItem)
...
+{
+#ifdef UFE_V2_FEATURES_AVAILABLE
+    Ufe::Scene::instance().notify(Ufe::SubtreeInvalidate(sceneItem));
+#else
+    // In Ufe v1 there was no subtree invalidate notif. So we mimic it by sending
+    // delete/add notifs.
+    sendObjectPostDelete(sceneItem);
+    sendObjectAdd(sceneItem);
+#endif
+}
```
seando-adsk, line 171: Two new helpers (like the Add/PostDelete above) that handle Ufe v1 case. Make the code below much simpler (no more ifdef).

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -133,6 +133,43 @@ void sendObjectPostDelete(const Ufe::SceneItem::Ptr& sceneItem)
...
+    Ufe::Scene::instance().notify(Ufe::ObjectDestroyed(ufePath));
+#else
+    // Unfortunately in Ufe v1 there was no object destroyed notif
+    // and the only delete notifs we have both take a scene item
+    // which we cannot create for the input Ufe path since that
+    // path is no longer valid (it has already been destroyed).
+    // So the only choice we have is to subtree invalidate the
+    // parent which will remove the destroyed item (keeping all
+    // the valid children).
+    auto sceneItem = Ufe::Hierarchy::createItem(ufePath);
```
ppt-adsk, line 148: As discussed, if we expect this to fail (because the UFE path is to a destroyed object), then we should skip directly to line 151 (creating the parent path), and TF_VERIFY'ing that this succeeds in the conditional at line 154.

ppt-adsk (changes requested):

ppt-adsk (approved):

seando-adsk:
@kxl-adsk Ready for merge. The stale review is just from a clang-format fix.

ppt-adsk (approved):

PR #2105, opened 2022-02-14
MAYA-114977 Updated documentation + unit test

### LINKED ISSUE #1849: Chaser documentation corrections
I'm looking at the "Common Plug-in Commands" readme, section "Setting Site-Specific Defaults". Judging by the source code, I believe the scope should be `"UsdMaya": { "UsdExport": {} }`, not `"UsdMaya": { "mayaUSDExport": {} }"`. Also, the "chaser" option is an array, so it should look something like this: `"chaser": ["alembic"]`.

JGamache-autodesk:
Fixes #1849

### DIFF
--- a/test/lib/usd/translators/testUsdImportExportSiteSpecificConfig.py
+++ b/test/lib/usd/translators/testUsdImportExportSiteSpecificConfig.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env mayapy
+#
+# Copyright 2021 Autodesk
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


test/lib/usd/translators/testUsdImportExportSiteSpecificConfig.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env mayapy
+#
+# Copyright 2021 Autodesk
```
boudrey, line 3: We should start putting 2022. :)

boudrey (dismissed):

PR #2106, opened 2022-02-15
Added export option to disable the model kinds processor

J-Mo63:
The ADSK plugin currently tags the root prim with a `kind` by default. There is an option,`-kind` (or internally, `rootKind`) to control the `kind` authored to the root prim which the docs for reads as follows:

| Long flag                        | Short flag | Type             | Default             | Description |
| -------------------------------- | ---------- | ---------------- | ------------------- | ----------- |
| `-kind`           | `-k`               | string           | none                | Specifies the required USD kind for *root prims* in the scene. (Does not affect kind for non-root prims.) If this flag is non-empty, then the specified kind will be set on any root prims in the scene without a `USD_kind` attribute (see the "Maya Custom Attributes" table below). Furthermore, if there are any root prims in the scene that do have a `USD_kind` attribute, then their `USD_kind` values will be validated to ensure they are derived from the kind specified by the `-kind` flag. For example, if the `-kind` flag is set to `group` and a root prim has `USD_kind=assembly`, then this is allowed because `assembly` derives from `group`. However, if the root prim has `USD_kind=subcomponent` instead, then `MayaUSDExportCommand` would stop with an error, since `subcomponent` does not derive from `group`. The validation behavior understands custom kinds that are registered using the USD kind registry, in addition to the built-in kinds. |

It states that the default is "none", however not supplying a `kind` will ultimately lead to it being authored as either a `component` or `assembly` depending on whether it detects exported gprims, and explicitly setting the option to "none" will result in `kind = "none"`. There is a need to disable this behaviour for pipelines that do not share this opinion.

The export option `-disableModelKindProcessor` (`-dmk`) has been added to provide the ability to disable this `kind` tagging behaviour by the ADSK plugin, and roundtrip of any imported `kind` will still be possible.


kxl-adsk:
How and when in the pipeline the kind will be authored on assets? What is the asset/model type that you are exporting from Maya? It would help if you could describe a bit the asset structure you are using.

fabal:
We use standalone proprietary libraries and tools (called LEAF, AssetWorkshop to name a few) to generate/edit our [Entity and Fragment](https://usd-alab.s3.amazonaws.com/documentation.html#file-asset-structure-entities-fragments) USD files in the first place. When the assets are loaded in Maya, kinds have already been set. We use the standard model hierarchy kinds to which we've added custom ones.
Describing the assets' model structure here would be fairly complex but it may be something we want to publish to ALab's documentation.

PR #2109, opened 2022-02-16
mayaUsdDuplicate selects duplicated object.

ppt-adsk:


lib/mayaUsd/undo/OpUndoItems.h
```diff
@@ -403,6 +404,46 @@ class SelectionUndoItem : public OpUndoItem
     MGlobal::ListAdjustment _selMode;
 };
 
+//------------------------------------------------------------------------------
+// UfeSelectionUndoItem
+//------------------------------------------------------------------------------
+
+/// \class UfeSelectionUndoItem
+/// \brief Record data needed to undo or redo select nodes sub-operations.
+class UfeSelectionUndoItem : public OpUndoItem
```
ppt-adsk, line 413: Added a new UfeSelectionUndoItem class.  Arguably the existing SelectionUndoItem should be transformed to hold a Ufe::Selection, because MSelectionList cannot hold Ufe::SceneItem's.  At that point we can simply remove UfeSelectionUndoItem.  This work is out of scope for this pull request.

lib/mayaUsd/commands/PullPushCommands.cpp
```diff
@@ -404,6 +408,25 @@ MStatus DuplicateCommand::doIt(const MArgList& argList)
...
         status = manager.duplicate(fSrcPath, fDstPath, userArgs) ? MS::kSuccess : MS::kFailure;
+
+        // If the duplicate src is Maya, the duplicate child of the destination
+        // will be USD, and vice-versa.  Construct an appropriate child path:
+        // - Maya duplicate to USD: we always duplicate directly under the
+        //   proxy shape, so add a single path component USD path segment.
+        // - USD duplicate to Maya: no path segment to add.
+        Ufe::Path childPath = (fSrcPath.runTimeId() == getMayaRunTimeId())
+            ?
+            // Maya duplicate to USD
```
pierrebai-adsk, line 419: Really small nit-pick, if the comment was before the '?' it would layout better.

lib/mayaUsd/undo/OpUndoItems.cpp
```diff
@@ -355,6 +357,53 @@ bool SelectionUndoItem::redo()
...
+bool UfeSelectionUndoItem::redo()
+{
+    invert();
+    return true;
+}
+
+void UfeSelectionUndoItem::invert()
+{
+    auto globalSn = Ufe::GlobalSelection::get();
+    auto previousSelection = *globalSn;
```
pierrebai-adsk, line 402: Can the returned value ever be null?

pierrebai-adsk (dismissed):
Except for those two minor comments, looks fine to me!

lib/mayaUsd/undo/OpUndoItems.cpp
ppt-adsk: No, the UFE global selection is initialized in Maya's model class, and we assert that the result is non-null.

pierrebai-adsk (approved):

PR #2110, opened 2022-02-16
MAYA-121303 Setting the instance colors or transforms into Maya is expensive. 

williamkrick:
If we discover they haven't changed then don't set the values into Maya.


vlasovi (approved):
Looks good. This will create a couple of conflicts with my refactoring changes so I will need to merge properly those changes with mine.

vlasovi (approved):

williamkrick:
The test failure was TestALUSDMayaPython_LayerManager which is unrelated to this change.

PR #2113, opened 2022-02-17
MAYA-121735: Add UV tangents directly in MaterialX

JGamache-autodesk:
- Removes the arbitrary tangent code previously used with MaterialX
  native code that can generate U-aligned tangents in the presence of a
  texcoord stream.

- Added stripped-down MaterialX graphs that provide compatible rendering
  of Lambert, Blinn, and Phong surfaces. The symmetric import export of
  these materials is lossless.


lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -387,18 +412,152 @@ MStatus _SetFAParameter(
 // name for UV at index zero.
 void _AddMissingTexcoordReaders(mx::DocumentPtr& mtlxDoc)
 {
-    // We expect only one node graph, but fixing them all is not an issue:
-    for (mx::NodeGraphPtr nodeGraph : mtlxDoc->getNodeGraphs()) {
-        if (nodeGraph->hasSourceUri()) {
```
JGamache-autodesk, line 392: Mostly whitespace since we are not looping on all NodeGraphs.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -387,18 +412,152 @@ MStatus _SetFAParameter(
...
+            if (retVal) {
+                break;
+            }
+        }
+    }
+    return retVal;
+}
+
+// USD does not provide tangents, so we need to build them from UV coordinates when possible:
+void _AddMissingTangents(mx::DocumentPtr& mtlxDoc)
```
JGamache-autodesk, line 515: New code to add tangent generators into the incoming graph.

test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateMaterialX.py
```diff
@@ -71,8 +69,15 @@ def _StartTest(self, testName):
...
            uses indexed UV streams"""
+        cmds.file(force=True, new=True)
+        cmds.move(2, -2, 1.5, 'persp')
         self._StartTest('MtlxUVStreamTest')
 
+    def testMayaSurfaces(self):
+        cmds.file(force=True, new=True)
+        cmds.move(6, -6, 6, 'persp')
+        cmds.rotate(60, 0, 45, 'persp')
+        self._StartTest('MayaSurfaces')
```
JGamache-autodesk, line 80: File includes a poly cube with standard surface MaterialX shading to show that the faces with Z-pointing normals render correctly now.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -407,60 +566,160 @@ void _AddMissingTexcoordReaders(mx::DocumentPtr& mtlxDoc)
...
-                        node->addInputFromNodeDef(input->getName());
-                        node->setConnectedNodeName(input->getName(), stReader->getName());
-                    }
+
+            if (!stReader && nodeDef->getNodeString() == _mtlxTokens->geompropvalue.GetString()
+                && node->getType() == _mtlxTokens->vector2.GetString()) {
+                // Grab the first st reader we can find. This will be the default one used for
+                // tangents unless we find something better.
+                stReader = node;
+                continue;
```
williamkrick, line 575: I'm not sure I understand why we can continue here. The first node def that matches will get the continue, the later if another nodeDef matches we execute the rest of the loop and maybe have a different behavior. I don't understand it in depth enough to know if it is a problem or not, it just looks a bit strange.

williamkrick (dismissed):

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
JGamache-autodesk: We can continue here because we know that a geompropvalue node never has a tangent input. Note that we do not continue in the next branch because normalmap nodes have tangent input.

vlasovi (approved):
Looks good to me.

williamkrick (approved):

PR #2123, opened 2022-02-18
MAYA-121545 persist cache-to-USD options

pierrebai-adsk:
- Add convertOptionsTextToDict function.
- Save and load the cache-to-USD options from an option var.
- Make the menu options values separate from the localized labels.
- Fill the cache-to-USD UI from the saved options.


ppt-adsk (dismissed):

ppt-adsk (approved):

pierrebai-adsk:
Only test that failed is the known flaky TestALUSDMayaPython_LayerManager

PR #1987, opened 2022-01-17
Support mergeNamespacesOnClash attribute on MayaReference prim

yantor3d:
# Premise
As explained in issue #640, the MayaReference prim is incompatible with a multi-proxy workflow where the proxy and a MayaReference in the proxy's stage (eg, a rig) share the same namespace.

# Changes
Add support for a `mergeNamespacesOnClash` boolean attribute on a MayaReference prim that controls the same flag on the Maya `file` MEL command.

# Tests
 - Added a test case to `testUsdImportMayaReference`
 - Tested in Maya under a mayaUsdProxyShape node(1). 
 
 ----
 
(1) Note that AutoPull was disabled for the MayaReference schema on 13/12/2021 by PPT - this behavior must be re-enabled to test the behavior under a mayaUsdProxyShape node.

PR #2124, opened 2022-02-21
MAYA-115146 As a user, I'd like my Maya Reference node to be named th…

boudrey:
…e regular Maya way

1. On Maya Reference behavior is modified.
2. Maya Reference Node is named using the mayaNamespace or filename if mayaNamespace is not set. And "RN" is added to the end.
3. On reconnect Maya Reference Node is rename to match any mayaNamespace or filename change.

+ modify the way Interactive Maya test is setup to use the .mod files.
+ adding another crash protection for ufe.

### DIFF
--- a/lib/mayaUsd/fileio/jobs/readJob.cpp
+++ b/lib/mayaUsd/fileio/jobs/readJob.cpp
@@ -141,7 +141,11 @@ bool UsdMaya_ReadJob::Read(std::vector<MDagPath>* addedDagPaths)
     } else {
         UsdStageCacheContext stageCacheContext(UsdMayaStageCache::Get(
             mImportData.stageInitialLoadSet() == UsdStage::InitialLoadSet::LoadAll));
-        stage = UsdStage::Open(rootLayer, sessionLayer, mImportData.stageInitialLoadSet());
+
+        if (mArgs.pullImport)
+            stage = UsdStage::Open(rootLayer, mImportData.stageInitialLoadSet());
+        else
+            stage = UsdStage::Open(rootLayer, sessionLayer, mImportData.stageInitialLoadSet());
     }
     if (!stage) {
         return false;
--- a/lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
@@ -48,6 +48,9 @@
 #include <maya/MNodeClass.h>
 #include <maya/MPlug.h>
 #include <maya/MSelectionList.h>
+#include <maya/MUuid.h>
+
+#include <ghc/filesystem.hpp>
 
 PXR_NAMESPACE_OPEN_SCOPE
 
@@ -137,12 +140,77 @@ const MObject getMessageAttr()
     static MObject messageAttr = MNodeClass("dagNode").attribute("message");
     return messageAttr;
 }
+
+const TfToken maya_associatedReferenceNode("maya_associatedReferenceNode"); // Didn't find any reference to this information, is it still used?
+
+const bool isMayaReference(const UsdPrim& prim)
+{
+//    return true;  // Force new behaviour for all prims.
+    const TfToken MayaReference("MayaReference");
+    return prim.GetTypeName() == MayaReference; // Only MayaReference prims are using the new behaviour.
+}
+
+const TfToken MayaReferenceAttribute("MayaReferenceAttribute");
+
 } // namespace
 
 const TfToken UsdMayaTranslatorMayaReference::m_namespaceName = TfToken("mayaNamespace");
 const TfToken UsdMayaTranslatorMayaReference::m_referenceName = TfToken("mayaReference");
 const TfToken UsdMayaTranslatorMayaReference::m_mergeNamespacesOnClash
     = TfToken("mergeNamespacesOnClash");
+
+// Get required namespace attribute from prim
+MString UsdMayaTranslatorMayaReference::namespaceFromPrim(const UsdPrim& prim)
+{
+    std::string ns;
+    if (UsdAttribute namespaceAttribute = prim.GetAttribute(m_namespaceName)) {
+        TF_DEBUG(PXRUSDMAYA_TRANSLATORS)
+            .Msg(
+                "MayaReferenceLogic::update Checking namespace on prim \"%s\".\n",
+                prim.GetPath().GetText());
+
+        if (!namespaceAttribute.Get<std::string>(&ns)) {
+            TF_DEBUG(PXRUSDMAYA_TRANSLATORS)
+                .Msg(
+                    "MayaReferenceLogic::update Missing namespace on prim \"%s\". Will create one "
+                    "from prim path.\n",
+                    prim.GetPath().GetText());
+            // Creating default namespace from prim path. Converts /a/b/c to a_b_c.
+            ns = prim.GetPath().GetString();
+            std::replace(ns.begin() + 1, ns.end(), '/', '_');
+        }
+    }
+    return MString(ns.c_str(), ns.size());
+}
+
+MString UsdMayaTranslatorMayaReference::getUniqueRefNodeName(const UsdPrim& prim, const MFnDagNode& parentDag, const MFnReference& refDependNode)
+{
+    MString uniqueRefNodeName;
+    if(isMayaReference(prim))
+    {
+        // Rename the reference node. Append "RN" to the end of filename, to indicate it's a reference
+        // node.
+        //
+        MString               filePath = refDependNode.fileName(false, false, false);
+        ghc::filesystem::path fsFilePath(filePath.asWChar());
+        fsFilePath.replace_extension(""); // Remove the extension
+        uniqueRefNodeName = namespaceFromPrim(prim);
+
+        if(uniqueRefNodeName.length()==0)
+        {
+            uniqueRefNodeName += fsFilePath.filename().wstring().c_str();
+        }
+        uniqueRefNodeName += L"RN";
+    } else {
+        // Rename the reference node.  We want a unique name so that multiple copies
+        // of a given prim can each have their own reference edits. We make a name
+        // from the full path to the prim for which the reference is being created
+        // (and append "_RN" to the end, to indicate it's a reference node, just
+        // because).
+        uniqueRefNodeName = refNameFromPath(parentDag) + "_RN";
+    }
+    return uniqueRefNodeName;
+}
 
 MStatus UsdMayaTranslatorMayaReference::LoadMayaReference(
     const UsdPrim& prim,
@@ -153,7 +221,6 @@ MStatus UsdMayaTranslatorMayaReference::LoadMayaReference(
 {
     TF_DEBUG(PXRUSDMAYA_TRANSLATORS)
         .Msg("MayaReferenceLogic::LoadMayaReference prim=%s\n", prim.GetPath().GetText());
-    const TfToken maya_associatedReferenceNode("maya_associatedReferenceNode");
     MStatus       status;
 
     MFnDagNode parentDag(parent, &status);
@@ -224,15 +291,17 @@ MStatus UsdMayaTranslatorMayaReference::LoadMayaReference(
     MFnReference refDependNode(referenceObject);
     connectReferenceAssociatedNode(parentDag, refDependNode);
 
-    // Rename the reference node.  We want a unique name so that multiple copies
-    // of a given prim can each have their own reference edits. We make a name
-    // from the full path to the prim for which the reference is being created
-    // (and append "_RN" to the end, to indicate it's a reference node, just
-    // because).
-    //
-    MString uniqueRefNodeName = refNameFromPath(parentDag) + "_RN";
+    MString uniqueRefNodeName = getUniqueRefNodeName(prim, parentDag, refDependNode);
     refDependNode.setName(uniqueRefNodeName);
 
+//  Always have to create the attribute, we are here because the prim didn't have it.
+//    if(!prim.HasAttribute(MayaReferenceAttribute)) 
+    {
+        UsdAttribute attr = prim.CreateAttribute(MayaReferenceAttribute, SdfValueTypeNames->String);
+        // Set the UUID of the reference node to prim
+        MUuid uuid = refDependNode.uuid();
+        attr.Set(uuid.asString().asChar());
+    }
     // Now load the reference to properly trigger the kAfterReferenceLoad callback
     MFileIO::loadReferenceByNode(referenceObject, &status);
     CHECK_MSTATUS_AND_RETURN_IT(status);
@@ -335,30 +404,12 @@ MStatus UsdMayaTranslatorMayaReference::update(const UsdPrim& prim, MObject pare
             prim.GetTypeName().GetText(),
             m_namespaceName.GetText());
 
-    // Get required namespace attribute from prim
-    std::string rigNamespace;
-    if (UsdAttribute rigNamespaceAttribute = prim.GetAttribute(m_namespaceName)) {
-        TF_DEBUG(PXRUSDMAYA_TRANSLATORS)
-            .Msg(
-                "MayaReferenceLogic::update Checking namespace on prim \"%s\".\n",
-                prim.GetPath().GetText());
-
-        if (!rigNamespaceAttribute.Get<std::string>(&rigNamespace)) {
-            TF_DEBUG(PXRUSDMAYA_TRANSLATORS)
-                .Msg(
-                    "MayaReferenceLogic::update Missing namespace on prim \"%s\". Will create one "
-                    "from prim path.\n",
-                    prim.GetPath().GetText());
-            // Creating default namespace from prim path. Converts /a/b/c to a_b_c.
-            rigNamespace = prim.GetPath().GetString();
-            std::replace(rigNamespace.begin() + 1, rigNamespace.end(), '/', '_');
-        }
-    }
-
     MFnDagNode parentDag(parent, &status);
     CHECK_MSTATUS_AND_RETURN_IT(status);
 
-    MString rigNamespaceM(rigNamespace.c_str(), rigNamespace.size());
+    // Get required namespace attribute from prim
+    MString rigNamespaceM = namespaceFromPrim(prim);
+    std::string rigNamespace = rigNamespaceM.asChar();
 
     MObject refNode;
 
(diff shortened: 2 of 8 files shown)


lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -138,6 +140,20 @@ const MObject getMessageAttr()
     return messageAttr;
 }
 
+const TfToken maya_associatedReferenceNode(
```
boudrey, line 143: Could not find any reference to use this value stored in the session layer.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -173,6 +189,60 @@ const TfToken UsdMayaTranslatorMayaReference::m_referenceName = TfToken("mayaRef
...
+        MString               filePath = refDependNode.fileName(false, false, false);
+        ghc::filesystem::path fsFilePath(filePath.asWChar());
+        fsFilePath.replace_extension(""); // Remove the extension
+        uniqueRefNodeName = namespaceFromPrim(prim);
+
+        if (uniqueRefNodeName.length() == 0) {
+            uniqueRefNodeName += fsFilePath.filename().wstring().c_str();
+        }
+        uniqueRefNodeName += L"RN";
+    } else {
```
ppt-adsk, line 235: For my benefit, can you explain when this else clause is taken?  For AL reference prim?  It might be a good code comment.

ppt-adsk: Why is there a difference between Maya reference prims and others?  Why can't the behavior be the same?  Is it for backward compatibility only?

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -254,15 +323,15 @@ MStatus UsdMayaTranslatorMayaReference::LoadMayaReference(
...
-    // from the full path to the prim for which the reference is being created
-    // (and append "_RN" to the end, to indicate it's a reference node, just
-    // because).
-    //
-    MString uniqueRefNodeName = refNameFromPath(parentDag) + "_RN";
+    MString uniqueRefNodeName = getUniqueRefNodeName(prim, parentDag, refDependNode);
     refDependNode.setName(uniqueRefNodeName);
 
+    //  Always have to create the attribute, we are here because the prim didn't have it.
+    //    if(!prim.HasAttribute(MayaReferenceAttribute))
```
ppt-adsk, line 330: Debugging code, perhaps remove?

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -424,29 +475,74 @@ MStatus UsdMayaTranslatorMayaReference::update(const UsdPrim& prim, MObject pare
...
+                                    attr.Set(refDependNodeName.asChar());
+                                }
+                                break;
+                            }
+                        }
+                    }
+                }
+            }
+        } else {
+            // Old behaviour
```
ppt-adsk, line 521: Old behavior seems a mis-nomer: it's current behavior for non-Maya references, right?

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -424,29 +475,74 @@ MStatus UsdMayaTranslatorMayaReference::update(const UsdPrim& prim, MObject pare
...
-                // If found a match, reconnect the reference node's `associatedNode`
-                // attr before loading it, since the previous connection may be gone.
-                //
-                if (refName == expectedRefName) {
-                    // Reconnect the reference node's `associatedNode` attr before
-                    // loading it, since the previous connection may be gone.
-                    connectReferenceAssociatedNode(parentDag, tempRefFn);
-                    refNode = tempRefNode;
-                    break;
+        if (isMayaReference(prim)) {
```
ppt-adsk, line 478: I would have appreciated a high-level description of the overall connection strategy, and the problems that may occur that it attempts to solve.

lib/mayaUsd/fileio/translators/translatorMayaReference.h
```diff
@@ -62,6 +62,11 @@ struct UsdMayaTranslatorMayaReference
     static MStatus update(const UsdPrim& prim, MObject parent);
 
 private:
+    static MString namespaceFromPrim(const UsdPrim& prim);
```
ppt-adsk, line 65: Nit-pick: if these methods are static and private, you may as well move them to the unnamed namespace in the .cpp file, to reduce recompilation and dependencies.  In a small project like ours, admittedly it doesn't make much difference, but it's still good practice.

lib/mayaUsd/fileio/jobs/readJob.cpp
```diff
@@ -141,7 +141,11 @@ bool UsdMaya_ReadJob::Read(std::vector<MDagPath>* addedDagPaths)
     } else {
         UsdStageCacheContext stageCacheContext(UsdMayaStageCache::Get(
             mImportData.stageInitialLoadSet() == UsdStage::InitialLoadSet::LoadAll));
-        stage = UsdStage::Open(rootLayer, sessionLayer, mImportData.stageInitialLoadSet());
+
+        if (mArgs.pullImport)
```
ppt-adsk, line 145: Can you add a code comment here describing why a pull import is different from a regular import?

ppt-adsk (commented):
Is there any way to provide an automated test specifically for this, or is your change covered by existing tests?

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
boudrey: Yes, we don't want to break existing pipelines.

lib/mayaUsd/fileio/translators/translatorMayaReference.h
boudrey: Those functions are using private static class members.

ppt-adsk (approved):

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -424,29 +472,74 @@ MStatus UsdMayaTranslatorMayaReference::update(const UsdPrim& prim, MObject pare
...
+
+                                // Update Maya Ref Node on reconnect
+                                MString uniqueRefNodeName
+                                    = getUniqueRefNodeName(prim, parentDag, tempRefFn);
+                                tempRefFn.setName(uniqueRefNodeName);
+
+                                MString refDependNodeName = tempRefFn.name();
+                                VtValue value(std::string(
+                                    refDependNodeName.asChar(), refDependNodeName.length()));
+                                prim.SetCustomDataByKey(maya_associatedReferenceNode, value);
```
nxkb, line 505: How is this different from the "MayaReferenceAttribute"?

It seems weird to set this only when re-connecting a maya reference but not when loading it in the first place. Maybe this is old code?

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -424,29 +472,74 @@ MStatus UsdMayaTranslatorMayaReference::update(const UsdPrim& prim, MObject pare
...
+                            if (expectedValue == tempRefFn.name()) {
+                                // Reconnect the reference node's `associatedNode` attr before
+                                // loading it, since the previous connection may be gone.
+                                connectReferenceAssociatedNode(parentDag, tempRefFn);
+                                refNode = tempRefNode;
+
+                                // Update Maya Ref Node on reconnect
+                                MString uniqueRefNodeName
+                                    = getUniqueRefNodeName(prim, parentDag, tempRefFn);
+                                tempRefFn.setName(uniqueRefNodeName);
```
nxkb, line 500: I believe that this will fail if the node is locked which it is sometimes.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -424,29 +472,74 @@ MStatus UsdMayaTranslatorMayaReference::update(const UsdPrim& prim, MObject pare
...
+                    // If found a match, reconnect the reference node's `associatedNode`
+                    // attr before loading it, since the previous connection may be gone.
+                    //
+                    if (refName == expectedRefName) {
+                        // Reconnect the reference node's `associatedNode` attr before
+                        // loading it, since the previous connection may be gone.
+                        connectReferenceAssociatedNode(parentDag, tempRefFn);
+                        refNode = tempRefNode;
+                        break;
+                    }
```
nxkb, line 542: I would sort of consider this old behavior to be "broken". I fixed it in this PR https://github.com/Autodesk/maya-usd/pull/1620 using the reference node name strategy, but I believe the "session" layer storage strategy should be even less-obtrusive. I think we can fix old and new for the same issues!

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -424,29 +472,74 @@ MStatus UsdMayaTranslatorMayaReference::update(const UsdPrim& prim, MObject pare
...
-                // attr before loading it, since the previous connection may be gone.
-                //
-                if (refName == expectedRefName) {
-                    // Reconnect the reference node's `associatedNode` attr before
-                    // loading it, since the previous connection may be gone.
-                    connectReferenceAssociatedNode(parentDag, tempRefFn);
-                    refNode = tempRefNode;
-                    break;
+        if (isMayaReference(prim)) {
+            // New behaviour for MayaReference
```
nxkb, line 476: Do you think you could adapt the test I made in #1620 to run on this code? Depending on exactly how you envision this working, all you would need is to update the stage file with the ref node name and see that it reconnects properly when the prim path changes.

nxkb, line 476: The one thing I worry about with the session layer storage, is how smooth the experience is for a user that wants to "reconnect" a reference. If a user wants to swap a reference in and connect it to a stage, is this how you see it working?

- existing ref (ref01_RN)
- bring in reference (named ref02_RN)
- set session layer data to point to ref02_RN (maybe via a tool)
- refresh prim... new ref is connected... and name is updated to ref01_RN but fails (hopefully it doesn't become ref03_RN)

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -424,29 +472,74 @@ MStatus UsdMayaTranslatorMayaReference::update(const UsdPrim& prim, MObject pare
...
+                                tempRefFn.setName(uniqueRefNodeName);
+
+                                MString refDependNodeName = tempRefFn.name();
+                                VtValue value(std::string(
+                                    refDependNodeName.asChar(), refDependNodeName.length()));
+                                prim.SetCustomDataByKey(maya_associatedReferenceNode, value);
+
+                                if (prim.HasAttribute(MayaReferenceAttribute)) {
+                                    UsdAttribute attr = prim.GetAttribute(MayaReferenceAttribute);
+                                    attr.Set(refDependNodeName.asChar());
```
nxkb, line 509: I like how this should get the updated name.

nxkb (commented):
Looks good, just wondering if we can get it working for the ALMayaReference too.

yantor3d:
@kxl-adsk 

> For Laika, I know you are working on the MayaReference workflow. I would love to make sure this proposal is aligned with what you have been doing.

Our MayaReference workflow is based on a multi-proxy workflow. Our MayaReference prims use the same namespace as the proxy itself - the inspiration for [PR 1987](https://github.com/Autodesk/maya-usd/pull/1987) - so if/when we need to get the referenceNode we do so using `cmds.referenceQuery` on a known node in the reference. 

This update should not affect us.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -424,29 +472,74 @@ MStatus UsdMayaTranslatorMayaReference::update(const UsdPrim& prim, MObject pare
...
+                                MString uniqueRefNodeName
+                                    = getUniqueRefNodeName(prim, parentDag, tempRefFn);
+                                tempRefFn.setName(uniqueRefNodeName);
+
+                                MString refDependNodeName = tempRefFn.name();
+                                VtValue value(std::string(
+                                    refDependNodeName.asChar(), refDependNodeName.length()));
+                                prim.SetCustomDataByKey(maya_associatedReferenceNode, value);
+
+                                if (prim.HasAttribute(MayaReferenceAttribute)) {
```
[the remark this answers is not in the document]
boudrey: Test has already been done earlier, I will remove it.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -254,15 +322,13 @@ MStatus UsdMayaTranslatorMayaReference::LoadMayaReference(
...
-    // of a given prim can each have their own reference edits. We make a name
-    // from the full path to the prim for which the reference is being created
-    // (and append "_RN" to the end, to indicate it's a reference node, just
-    // because).
-    //
-    MString uniqueRefNodeName = refNameFromPath(parentDag) + "_RN";
+    MString uniqueRefNodeName = getUniqueRefNodeName(prim, parentDag, refDependNode);
     refDependNode.setName(uniqueRefNodeName);
 
+    //  Always have to create the attribute, we are here because the prim didn't have it.
```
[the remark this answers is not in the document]
boudrey: Ref are locked so they cannot be removed or renamed.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
boudrey: It was already in the original implementation, just updating it here in case the value changed. It has the same value as the MayaReferenceAttribute. I didn't removed it not to break any possible existing workflow using it.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
boudrey: In the case I tested, the rename was working. Need to validate with more test cases.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
boudrey: This looks like the case in your test from https://github.com/Autodesk/maya-usd/pull/1620, let see how it behaves.

boudrey:
@nxkb have I covered all your concerns? Any other comments before it is merged?

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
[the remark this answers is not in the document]
nxkb: do you just mean the standard maya node locking? Our artists are definitely mischievous/savvy enough to delete one. 

Also couldn't it happen with a rename if an artist wants to rename and connect a different reference?

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
nxkb: awesome thanks for including it.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -424,29 +496,88 @@ MStatus UsdMayaTranslatorMayaReference::update(const UsdPrim& prim, MObject pare
...
+                        if (expectedValue == tempRefFn.name()) {
+                            // Reconnect the reference node's `associatedNode` attr before
+                            // loading it, since the previous connection may be gone.
+                            connectReferenceAssociatedNode(parentDag, tempRefFn);
+                            refNode = tempRefNode;
+
+                            // Update Maya Ref Node on reconnect
+                            MString uniqueRefNodeName
+                                = getUniqueRefNodeName(prim, parentDag, tempRefFn);
+                            tempRefFn.setName(uniqueRefNodeName);
```
nxkb, line 537: I still think you'll need to unlock this node before you rename, especially if something is "locking" ref nodes. If you haven't run into this yet, maybe it's possible that all the tests are creating new refs rather than updating existing ones.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -138,6 +142,20 @@ const MObject getMessageAttr()
...
+    if (TfGetenvBool("MAYAUSD_ENABLE_MAYA_REFERENCE_NEW_BEHAVIOUR", false)) {
+        return true; // Force new behaviour for all Maya Reference prims.
+    } else {
+        const TfToken MayaReference("MayaReference");
+        return prim.GetTypeName()
+            == MayaReference; // Only MayaReference prims are using the new behaviour.
+    }
+}
+
+const TfToken MayaReferenceAttribute("MayaReferenceAttribute");
```
nxkb, line 157: Naming thought: the "Attribute" suffix is not going to be very useful once it is literally a UsdAttribute. How about "NodeName" instead?

nxkb (commented):
Looking good, thanks for grabbing that test. Just a few more questions.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
boudrey: Node locking doesn't seems to be an issue. I tried modifying the "Maya Namespace" in AE and the rename is done without complaining about locking.

boudrey:
@ppt-adsk can you do one final review and merge this PR please.

ppt-adsk (approved):

lib/mayaUsd/fileio/jobs/readJob.cpp
nxkb: So does a pullImport bring everything in "fresh"? and create new reference nodes? always or just in cases where the target refNode isn't named exactly right? (if we ignore the session layer, we won't have the mapping between prim and maya reference node)

I wonder if adding a "why" it ignores the session layer here would help it be clear.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
nxkb: ^ I think this would be still be appreciated, or perhaps just updating the description above this which seems like it might be a little out of date now.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
nxkb: I do think this test would be useful but we can tackle it in a follow up MR if you wish.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
nxkb: ok great.

nxkb (commented):
final ideas for clarity in the code

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
nxkb: great, i think this works well.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -143,10 +145,13 @@ const MObject getMessageAttr()
 const TfToken maya_associatedReferenceNode("maya_associatedReferenceNode");
 const bool    isMayaReference(const UsdPrim& prim)
 {
-    //    return true;  // Force new behaviour for all Maya Reference prims.
-    const TfToken MayaReference("MayaReference");
-    return prim.GetTypeName()
-        == MayaReference; // Only MayaReference prims are using the new behaviour.
+    if (TfGetenvBool("MAYAUSD_ENABLE_MAYA_REFERENCE_NEW_BEHAVIOUR", false)) {
+        return true; // Force new behaviour for all Maya Reference prims.
```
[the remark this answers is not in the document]
nxkb: awesome, thanks for swapping.

lib/mayaUsd/fileio/jobs/readJob.cpp
boudrey: Stage search in cache doesn't affect how the mayaReferenceNodeName attribute is used in sessionLayer to match reference node. New reference node are created only if no match with mayaReferenceNodeName and if the expected name is not found in all the non-connected reference node.

PR #2128, opened 2022-02-22
Added mesh export options for interpolation type and reference object mode

J-Mo63:
In further efforts to bring the ADSK plugin's functionality to parity with the AL plugin, changes have been made to how `primvars:pref` is exported, adding two new export options: `-referenceObjectInterpolation` (`-roi`), and `-referenceObjectMode` (`-rom`).

The `referenceObjectInterpolation` option has been added as there is the assumption in the ADSK plugin that `primvars:pref` should be exported with `varying` as its interpolation type, whereas AL assumes it should be `vertex`. This options defaults to `varying` to maintain current functionality, but can be set by the user to any valid interpolation type offered by `UsdGeomTokens`.

_For context, this has been done because AL requires the authored interpolation of `primvars:pref` to always be `vertex` as so:_
```
point3f[] primvars:pref = [...] (
    interpolation = "vertex"
)
```

The `referenceObjectMode` option has been added, replacing `exportReferenceObjects` (a boolean), as when this options was set to `true`, it would only export `primvars:pref` in the scenario that the Maya mesh had a "referenceObject" connected to it, which only seems to be created if the mesh was imported with the ADSK plugin and already had this primvar authored. To negotiate this, it has been turned into a multi-select option offering `none` (equivalent to `false`), `attributeOnly` (equivalent to `true`), and `defaultToMesh`, which simply copies the mesh's points if nothing is connected to the "referenceObject" plug.

_If desired, we can revert the change to `exportReferenceObjects` and instead add a boolean for `exportMeshAsReferece` on top to preserve compatibility._


seando-adsk:
@J-Mo63 Just wanted to let you know that we are going to have a few internal discussions about the proposed changes and will get back to you.

J-Mo63:
> @J-Mo63 Just wanted to let you know that we are going to have a few internal discussions about the proposed changes and will get back to you.

Yeah no worries, @seando-adsk! Just want to flag that there will be a handful of PRs like this (and #2106) coming down the line as we elucidate workflow requirements and begin to migrate to the ADSK plugin for static import and export. I'll try to note what the requirement behind each change is as we go so we can simplify the implementation if you're just happy to adopt it (e.g. `primvars:pref` needs to have `vertex` interpolation type authored).

lib/mayaUsd/commands/Readme.md
```diff
@@ -129,7 +129,8 @@ their own purposes, similar to the Alembic export chaser example.
...
 | `-exportColorSets`               | `-cls`     | bool             | true                | Enable or disable the export of color sets |
 | `-exportInstances`               | `-ein`     | bool             | true                | Enable or disable the export of instances |
-| `-exportReferenceObjects`        | `-ero`     | bool             | false               | Whether to export reference objects for meshes. The reference object's points are exported as a primvar on the mesh object; the primvar name is determined by querying `UsdUtilsGetPrefName()`, which defaults to `pref`. |
+| `-referenceObjectMode`           | `-rom`     | string           | `none`              | Determines how to export reference objects for meshes. The reference object's points are exported as a primvar on the mesh object; the primvar name is determined by querying `UsdUtilsGetPrefName()`, which defaults to `pref`. Valid values are: `none` - No reference objects are exported, `attributeOnly` - Only meshes set with a valid "referenceObject" attached will be exported, `defaultToMesh` - Meshes with no "referenceObject" attached will export thier own points |
```
JGamache-autodesk, line 132: Typo: thier

lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -541,7 +541,19 @@ UsdMayaJobExportArgs::UsdMayaJobExportArgs(
...
+          UsdMayaJobExportArgsTokens->none,
+          { UsdMayaJobExportArgsTokens->attributeOnly, UsdMayaJobExportArgsTokens->defaultToMesh }))
+    , referenceObjectInterpolation(_Token(
+          userArgs,
+          UsdMayaJobExportArgsTokens->referenceObjectInterpolation,
+          UsdGeomTokens->varying,
+          { UsdGeomTokens->vertex,
+            UsdGeomTokens->constant,
+            UsdGeomTokens->faceVarying,
+            UsdGeomTokens->uniform }))
```
JGamache-autodesk, line 556: Could you add this to the `std::ostream& operator<<(std::ostream& out, const UsdMayaJobExportArgs& exportArgs)` function as well?

lib/mayaUsd/python/wrapPrimWriter.cpp
```diff
@@ -497,7 +497,7 @@ void wrapJobExportArgs()
         .def_readonly("exportMaterialCollections", &UsdMayaJobExportArgs::exportMaterialCollections)
         .def_readonly("exportMeshUVs", &UsdMayaJobExportArgs::exportMeshUVs)
         .def_readonly("exportNurbsExplicitUV", &UsdMayaJobExportArgs::exportNurbsExplicitUV)
-        .def_readonly("exportReferenceObjects", &UsdMayaJobExportArgs::exportReferenceObjects)
+        .def_readonly("referenceObjectMode", &UsdMayaJobExportArgs::referenceObjectMode)
```
JGamache-autodesk, line 500: Please add referenceObjectInterpolation

plugin/pxr/maya/lib/usdMaya/usdTranslatorExport.mel
```diff
@@ -153,7 +153,11 @@ global proc int usdTranslatorExport (string $parent,
         checkBox -l "Export UVs" exportUVsCheckBox;
 
         checkBox -l "Export ColorSets" exportColorSetsCheckBox;
-        checkBox -l "Export Reference Objects" exportReferenceObjectsCheckBox;
+
```
JGamache-autodesk, line 156: That's the Pixar plugin. Perhaps you meant to make these changes in `plugin\adsk\scripts\mayaUsdTranslatorExport.mel` instead?

JGamache-autodesk (changes requested):

plugin/pxr/maya/lib/usdMaya/usdTranslatorExport.mel
J-Mo63: No, we don't really need the option in the UI at this stage. Just missed file name and saw the same option name declared.

J-Mo63: I made the update because I got the impression that the option being changed in the shared translators seems to be referenced by the Pixar plugin as well. Is that correct?

J-Mo63:
@JGamache-autodesk after chatting with our rendering team, they believe that setting `primvars:pref` to `varying` [would not perform the correct interpolation for the attribute](https://graphics.pixar.com/opensubdiv/docs/subdivision_surfaces.html#vertex-and-varying-data), and anything other than `vertex` would be incorrect. If this is the case, it may be best to remove the added `referenceObjectInterpolation` option and just update the default to `vertex`. What are your thoughts?

JGamache-autodesk:
@J-Mo63 we arrived at the same conclusion here. The solution you propose would fix a bug.

J-Mo63:
> @J-Mo63 we arrived at the same conclusion here. The solution you propose would fix a bug.

@JGamache-autodesk I've gone ahead and updated the PR accordingly.

JGamache-autodesk (approved):

ppt-adsk:
Hi @J-Mo63, is this pull request ready for merge?  We at Autodesk believe it is.  If that's the case, then add the ready-for-merge label to it, and we'll merge it in.

J-Mo63:
> Hi @J-Mo63, is this pull request ready for merge? We at Autodesk believe it is. If that's the case, then add the ready-for-merge label to it, and we'll merge it in.

@ppt-adsk it's good to merge! Unfortunately I don't have permissions to add labels though.

ppt-adsk:
Merged!  Thanks for the contribution and the great work.

PR #2132, opened 2022-02-22
MAYA-121814 change label to cancel maya ref edit

pierrebai-adsk:
- Refactor code to detect if a pulled path is a maya reference.
- Use refactored code to detect maya reference when building the menu.
- Use different label for Maya reference when discarding edits.


seando-adsk (changes requested):
Your new python file needs the copyright block at the top.

seando-adsk (approved):

PR #2134, opened 2022-02-22
MAYA-120659 Get the display style of all viewports rather than just the current

williamkrick:
viewport to prevent thrashing display styles when multiple viewports are displayed.


vlasovi (commented):
Looks good to me.

vlasovi (approved):

PR #2136, opened 2022-02-23
:bug: change _GetLocalTransformForDagPoseMember check from maya api version to diffrence_type iterator trait support, since 2019's MArrayIteratorTemplate.h has missing iterator traits.

sun-frog:
Maya 2019 MArrayIteratorTemplate.h has missing iterator traits. This has been fix in Maya 2020 onwards.

### DIFF
--- a/cmake/modules/FindMaya.cmake
+++ b/cmake/modules/FindMaya.cmake
@@ -20,6 +20,7 @@
 # MAYA_UPDATE_UFE_IDENTIFIER_SUPPORT Presence of MPxSubSceneOverride::updateUfeIdentifier.
 # MAYA_ENABLE_NEW_PRIM_DELETE Enable new delete behaviour for delete command
 # MAYA_HAS_DISPLAY_STYLE_ALL_VIEWPORTS Presence of MFrameContext::getDisplayStyleOfAllViewports.
+# MAYA_ARRAY_ITERATOR_DIFFERENCE_TYPE_SUPPORT Presence of maya array iterator difference_type trait
 
 #=============================================================================
 # Copyright 2011-2012 Francisco Requena <frarees@gmail.com>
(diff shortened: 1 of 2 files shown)


williamkrick (dismissed):
@sun-frog I dug into this internally and found that Maya 2019.3 doesn't have the missing iterator traits but some later patch fixes available to some enterprise customers do have the traits. The change you've proposed here will disable the code for some versions that do have support for the iterators. In this case I'm okay with disabling the code in 2019 because it is a debug sanity check to validate behavior rather than some user-visible behavior.

For completeness I'll include how to correctly detect that Maya has the required API. In [FindMaya.cmake](https://github.com/Autodesk/maya-usd/blob/dev/cmake/modules/FindMaya.cmake) you can copy the pattern used for MAYA_UPDATE_UFE_IDENTIFIER_SUPPORT. This build code does a regular expression search for a specified string in a Maya API header file. In this case we'd need to search MArrayIteratorTemplate.h for "difference_type" and then set an new env var when it is present.

cmake/modules/FindMaya.cmake
```diff
@@ -407,6 +407,15 @@ if (MAYA_API_VERSION VERSION_GREATER_EQUAL 20230000)
     set(MAYA_ENABLE_NEW_PRIM_DELETE TRUE CACHE INTERNAL "enableNewPrimDelete")
 endif()
 
+set(MAYA_ARRAY_ITERATOR_DIFFERENCE_TYPE_SUPPORT FALSE CACHE INTERNAL "hasArrayIteratorDifferenceType")
+if(MAYA_INCLUDE_DIRS AND EXISTS "${MAYA_INCLUDE_DIR}/maya/MArrayIteratorTemplate.h")
+    file(STRINGS ${MAYA_INCLUDE_DIR}/maya/MArrayIteratorTemplate.h MAYA_HAS_API REGEX "hasArrayIteratorDifferenceType")
```
williamkrick, line 412: The regular expression here is not correct. This line is going to search MArrayIteratorTemplate.h for the string "hasArrayIteratorDifferenceType" and set MAYA_HAS_API if it does. However, "hasArrayIteratorDifferenceType" is not in MArrayIteratorTemplate.h and so we'll never set MAYA_HAS_API true.

I think the REGEX to search for should be "difference_type". That is a string which is in the updated file and not in the older file.

williamkrick (changes requested):

cmake/modules/FindMaya.cmake
sun-frog: You are correct, I accidentally uploaded this bugged and incomplete (I will add the new env variable at the top of the file comment) version of the code review change.

williamkrick (dismissed):

seando-adsk:
@sun-frog I was about to merge your change, but there is a conflict that needs to be resolved first.

williamkrick (approved):

PR #2137, opened 2022-02-23
Maya-121344 Support for HdPoints in VP2 render delegate. First iteration.

vlasovi:
- Adding new class HdVP2Points
- Using FatPoint shader to draw points (interim solution)
- Adding an autotest


lib/mayaUsd/render/vp2RenderDelegate/points.cpp
```diff
@@ -0,0 +1,1071 @@
...
+                const VtValue& value = it->second.data;
+                if (ARCH_LIKELY(value.IsHolding<VtVec3fArray>())) {
+                    normals = value.UncheckedGet<VtVec3fArray>();
+                }
+            }
+
+            // Using a zero vector to indicate requirement of camera-facing
+            // normals when there is no authored normals.
+            if (normals.empty()) {
+                normals.push_back(GfVec3f(0.0f, 0.0f, 0.0f));
```
williamkrick, line 269: What if the user has authored normals which are the zero vector for some reason?

lib/mayaUsd/render/vp2RenderDelegate/points.cpp
```diff
@@ -0,0 +1,1071 @@
...
+            const auto it = primvarSourceMap.find(HdTokens->widths);
+            if (it != primvarSourceMap.end()) {
+                const VtValue& value = it->second.data;
+                if (value.IsHolding<VtFloatArray>()) {
+                    widths = value.UncheckedGet<VtFloatArray>();
+                }
+            }
+
+            if (widths.empty()) {
+                widths.push_back(1.0f);
```
williamkrick, line 304: Does Storm have the same default value? Maybe we can name it in a constexpr to make it obvious?

lib/mayaUsd/render/vp2RenderDelegate/points.cpp
```diff
@@ -0,0 +1,1071 @@
...
+                                break;
+                            }
+                        }
+                    }
+                }
+            }
+
+            // If color/opacity is not found, the default color will be used
+            if (colorArray.empty()) {
+                colorArray.push_back(GfVec3f(0.247f, 0.137f, 0.122f));
```
williamkrick, line 400: Please name this in a constexpr.

lib/mayaUsd/render/vp2RenderDelegate/points.cpp
```diff
@@ -0,0 +1,1071 @@
...
+
+    // If the prim is instanced, create one new instance per transform.
+    // The current instancer invalidation tracking makes it hard for
+    // us to tell whether transforms will be dirty, so this code
+    // pulls them every time something changes.
+    // If the mesh is instanced but has 0 instance transforms remember that
+    // so the render item can be hidden.
+
+    bool instancerWithNoInstances = false;
+    if (!GetInstancerId().IsEmpty()) {
```
williamkrick, line 537: We should log another story to move this duplicated code to the parent class.

williamkrick (approved):
I think its good. Two things we should work on going forward:

1. you still had to duplicate a lot of code in points that is very similar to what is in mesh and probably curves. We should move that up to the parent when we have time.
2. A lot of our tests only test one prim type, meshes. We should log something to test basis curves and points too. For example the isolate select test doesn't do anything with curves (and now is also not testing points). That's a huge gap in our test coverage.

PR #2143, opened 2022-02-25
MAYA-113944 If USD reports that the topology is dirty compare the

williamkrick:
new topology to the old. Only do the topology updates if the old and new typologies are different. Save the adjacency calculation used to compute normals.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
@@ -528,15 +528,24 @@ void HdVP2Mesh::_PrepareSharedVertexBuffers(
             if (computeCPUNormals) {
                 // note: normals gets dirty when points are marked as dirty,
                 // at change tracker.
-                Hd_VertexAdjacencySharedPtr adjacency(new Hd_VertexAdjacency());
-                HdBufferSourceSharedPtr     adjacencyComputation
-                    = adjacency->GetSharedAdjacencyBuilderComputation(&_meshSharedData->_topology);
-                adjacencyComputation->Resolve();
+                if (!_meshSharedData->_adjacency) {
+                    _meshSharedData->_adjacency.reset(new Hd_VertexAdjacency());
+
+                    HdBufferSourceSharedPtr     adjacencyComputation
+                        = _meshSharedData->_adjacency->GetSharedAdjacencyBuilderComputation(&_meshSharedData->_topology);
+                    MProfilingScope profilingScope(
+                        HdVP2RenderDelegate::sProfilerCategory,
+                        MProfiler::kColorC_L2,
+                        _rprimId.asChar(),
+                        "HdVP2Mesh::computeAdjacency");
+                        
+                    adjacencyComputation->Resolve();
+                }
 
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -528,15 +528,25 @@ void HdVP2Mesh::_PrepareSharedVertexBuffers(
...
+                    HdBufferSourceSharedPtr adjacencyComputation
+                        = _meshSharedData->_adjacency->GetSharedAdjacencyBuilderComputation(
+                            &_meshSharedData->_topology);
+                    MProfilingScope profilingScope(
+                        HdVP2RenderDelegate::sProfilerCategory,
+                        MProfiler::kColorC_L2,
+                        _rprimId.asChar(),
+                        "HdVP2Mesh::computeAdjacency");
+
+                    adjacencyComputation->Resolve();
```
williamkrick, line 543: This call to resolve is slow, so save the result and only re-do it if the topology actually changes.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -919,7 +929,21 @@ void HdVP2Mesh::Sync(
                 MProfiler::kColorC_L2,
                 _rprimId.asChar(),
                 "HdVP2Mesh::GetMeshTopology");
-            _meshSharedData->_topology = GetMeshTopology(delegate);
+            HdMeshTopology newTopology = GetMeshTopology(delegate);
+
+            // Test to see if the topology actually changed. If not, we don't have to do anything!
+            // Don't test IsTopologyDirty anywhere below this because it is not accurate. Instead
+            // using the _indexBufferValid flag on render item data.
```
williamkrick, line 936: I need the separate flag because we've already lost dirty information by the time sync is called. In propagate dirty bits the old topology dirty flag saved the the render item was overwritten. If that item hasn't been getting updated (because it's in a repr that is not currently displayed) then we don't know if there was a ligitimate topology change before this frame that still needs to be handled for that item.

For that reason I need the separate _indexBufferValid flag per render item.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -2472,6 +2497,18 @@ void HdVP2Mesh::_UpdateDrawItem(
...
+        renderItemData._enabled = false;
+        _delegate->GetVP2ResourceRegistry().EnqueueCommit(
+            [&]() { renderItemData._renderItem->enable(false); });
+    };
+
+    _ForEachRenderItemInRepr(reprToken, hideDrawItem);
+}
+
+template <typename Func>
+void HdVP2Mesh::_ForEachRenderItemInRepr(const TfToken& reprToken, Func func)
```
williamkrick, line 2511: I think this could more to the parent class once the clean up gets merged.

vlasovi: Yeah. I moved this function to the parent class so this will have to be propely merged.

Issue #2142, opened 2022-02-24
Textures appear different when using Async Texture Loading in the Nvidia Attic Scene


santosd:
**Describe the bug**
When the Nvidia Attic Scene is loaded with Async Texture Loading enabled the textures appear different than when the stage is loaded without Async Texture Loading.

**Steps to reproduce**
1. Make sure that Async Texture Loading is enabled.
2. Create a stage from the Nvidia Attic Scene. Notice the shading on the beams how it is dark one side.
3. Create a new file in Maya.
4. Disable async texture loading.
5. Create a stage from the Nvidia Attic Scene. Notice the shading on the beams is dark on the opposite side.

**Expected behavior**
Loading with and without async texture loading should look the same.

**Attachments**
Async Disabled:
![Screenshot 2022-02-24 09:27:19](https://user-images.githubusercontent.com/29235905/155576467-5f84cf91-e847-4f35-a380-778eb67acd41.png)

Async Enabled
![Screenshot 2022-02-24 09:29:36](https://user-images.githubusercontent.com/29235905/155576488-65f97fa9-7068-471d-9375-f26be0956661.png)

Nvidia Attic Scene can be downloaded from the following link:
https://developer.nvidia.com/attic-nvidia

santosg87:
@santosd are you able to confirm the changes and close it? 

thank you! :)

santosd:
I retested this issue and this seems to be resolved so closing this issue for the time being.

PR #2145, opened 2022-02-25
MAYA-121882 Share the cached HdVP2TextureInfo between all materials so that isSRGB

williamkrick:
and uvScaleOffset and re-used correctly when hitting a texture in the cache.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -1318,7 +1318,7 @@ private:
 std::mutex                            HdVP2Material::_refreshMutex;
 std::chrono::steady_clock::time_point HdVP2Material::_startTime;
 std::atomic_size_t                    HdVP2Material::_runningTasksCounter;
-HdVP2TextureMap                       HdVP2Material::_textureMap;
+HdVP2GlobalTextureMap                 HdVP2Material::_globalTextureMap;
 
 /*! \brief  Releases the reference to the texture owned by a smart pointer.
  */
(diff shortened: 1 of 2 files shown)


williamkrick:
@csyshing I'm not sure if this impacting the issues with background material loading but it is in a similar enough area I thought you should be aware of the change.

@JGamache-autodesk we need to save the extra information that is HdVP2TextureInfo so that when another material hits the Maya texture cache we get these values right.

JGamache-autodesk (dismissed):
Looks good. Thanks!

JGamache-autodesk:
I strongly suspect this will help with #2142.

csyshing (dismissed):

csyshing:
@williamkrick @JGamache-autodesk I will test against #2142 to see how's it going.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -2580,20 +2592,23 @@ void HdVP2Material::_UpdateLoadedTexture(
     // function on idle to delete the task object.
     _textureLoadingTasks.erase(path);
 
-    // Check the local cache again, do not overwrite if same texture has
-    // been loaded asynchronously
-    if (_textureMap.find(path) != _textureMap.end()) {
-        return;
```
williamkrick, line 2586: The early exit here is wrong. It also stops marking the sprim dirty and triggering a refresh.

lib/mayaUsd/render/vp2RenderDelegate/material.h
```diff
@@ -147,10 +159,11 @@ class HdVP2Material final : public HdMaterial
 
     TfToken _surfaceNetworkToken; //!< Generated token to uniquely identify a material network
 
-    HdVP2ShaderUniquePtr _surfaceShader;    //!< VP2 surface shader instance
-    SdfPath              _surfaceShaderId;  //!< Path of the surface shader
-    HdVP2TextureMap      _textureMap;       //!< Textures used by this material
-    TfTokenVector        _requiredPrimvars; //!< primvars required by this material
+    HdVP2ShaderUniquePtr         _surfaceShader;    //!< VP2 surface shader instance
+    SdfPath                      _surfaceShaderId;  //!< Path of the surface shader
+    static HdVP2GlobalTextureMap _globalTextureMap; //!< Texture in use by all materials in MayaUSD
```
williamkrick, line 164: With a single global static structure we'd never delete the textures until the plugin unloads or Maya exits. At that time it is often not safe to try to destroy the texture, and textures were held in the cache beyond `file -new`.

This way we get the deletion at the right time.

JGamache-autodesk (dismissed):
Very good.

JGamache-autodesk (approved):

PR #2149, opened 2022-02-28
 MAYA-121626 Automatically set the animation option

pierrebai-adsk:
- Add animation, startTime, endTime, frameStride and frameSample to the users args dictionary.
- This allows passing these user options to merge-to-USD and cache-to-USD.
- Split translating text-format to a dictionary from extracting the time sample range.
- This allows passing only a dictionary for all user options instead of having additional separate options for the time samples.
- Adapt users of JobArgs to use the new pure-dictionary internal API.
- In particular, mergeToUsd and duplicate now use these dictionary options.
- Add animation data check-box in cache-to-USD options UI.
- Add animation data check-box in merge-to-USD options UI.
- Improve the text to dict conversion in Python to respect the historical format for options containing list of items.
- In particular, some use comma-separated values, others space-separated values.
- Add support for floating-point and integer values when parsing values when converting from text to VtDictionary in C++.
- Expose the isAnimated C++ function to Python.
- No longer always force the animation option on.
- Conditionally set it on if an object is animated.
- Properly handle the case where the animation option is set to more than 1 character when changing it.
- Prefill the animation range UI.

### DIFF
--- a/lib/mayaUsd/fileio/jobs/jobArgs.cpp
+++ b/lib/mayaUsd/fileio/jobs/jobArgs.cpp
@@ -91,14 +91,20 @@ bool _Boolean(const VtDictionary& userArgs, const TfToken& key)
 /// Extracts a double at \p key from \p userArgs, or defaultValue if it can't extract.
 double _Double(const VtDictionary& userArgs, const TfToken& key, double defaultValue)
 {
-    if (!VtDictionaryIsHolding<double>(userArgs, key)) {
-        TF_CODING_ERROR(
-            "Dictionary is missing required key '%s' or key is "
-            "not double type",
-            key.GetText());
-        return defaultValue;
-    }
-    return VtDictionaryGet<double>(userArgs, key);
+    if (VtDictionaryIsHolding<double>(userArgs, key))
+        return VtDictionaryGet<double>(userArgs, key);
+
+    // Since user dictionary can be provided from Python and in Python it is easy to
+    // mix int and double, especially since value literal will take the simplest value
+    // they can, for example 0 will be an int, support receiving the value as an integer.
+    if (VtDictionaryIsHolding<int>(userArgs, key))
+        return VtDictionaryGet<int>(userArgs, key);
+
+    TF_CODING_ERROR(
+        "Dictionary is missing required key '%s' or key is "
+        "not double type",
+        key.GetText());
+    return defaultValue;
 }
 
 /// Extracts a string at \p key from \p userArgs, or "" if it can't extract.
@@ -167,6 +173,11 @@ SdfPath _AbsolutePath(const VtDictionary& userArgs, const TfToken& key)
 template <typename T> std::vector<T> _Vector(const VtDictionary& userArgs, const TfToken& key)
 {
     // Check that vector exists.
+    if (VtDictionaryIsHolding<std::vector<T>>(userArgs, key)) {
+        std::vector<T> vals = VtDictionaryGet<std::vector<T>>(userArgs, key);
+        return vals;
+    }
+
     if (!VtDictionaryIsHolding<std::vector<VtValue>>(userArgs, key)) {
         TF_CODING_ERROR(
             "Dictionary is missing required key '%s' or key is "
--- a/lib/mayaUsd/fileio/jobs/jobArgs.cpp
+++ b/lib/mayaUsd/fileio/jobs/jobArgs.cpp
@@ -27,6 +27,7 @@
 #include <pxr/base/tf/fileUtils.h>
 #include <pxr/base/tf/staticTokens.h>
 #include <pxr/base/tf/token.h>
+#include <pxr/base/vt/array.h>
 #include <pxr/base/vt/dictionary.h>
 #include <pxr/usd/sdf/path.h>
 #include <pxr/usd/sdf/schema.h>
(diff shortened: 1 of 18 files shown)


lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -758,15 +765,7 @@ MStatus UsdMayaJobExportArgs::GetDictionaryFromEncodedOptions(
             }
 
             std::string argName(theOption[0].asChar());
-            if (argName == "animation") {
```
pierrebai-adsk, line 761: These are now normal options, found in the default and guide dictionaries.

lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -723,21 +737,14 @@ UsdMayaJobExportArgs UsdMayaJobExportArgs::CreateFromDictionary(
...
-    std::vector<double>* timeSamples)
+    const MString& optionsString,
+    VtDictionary*  toFill)
 {
     if (!toFill)
         return MS::kFailure;
 
     VtDictionary& userArgs = *toFill;
 
-    bool       exportAnimation = false;
```
pierrebai-adsk, line 735: The animation frame are now extracted in a separate function since the functionality needs to be used from multiple places.

lib/mayaUsd/python/wrapPrimUpdaterManager.cpp
```diff
@@ -35,6 +35,21 @@ PXR_NAMESPACE_USING_DIRECTIVE
 
 namespace {
 
+bool isAnimated(const std::string& nodeName)
+{
```
pierrebai-adsk, line 39: Animation detection now needs to be done in Python/MEL instead of C++, so it is exposed to Python.

lib/mayaUsd/resources/scripts/cacheToUsd.py
```diff
@@ -22,19 +22,11 @@
 import re
 
 
-def forceCacheOptions(textOptions):
-    """
```
pierrebai-adsk, line 26: Animation option is no longer forced true. The presence of animated node is used to set it on or off instead, which is done elsewhere.

lib/mayaUsd/resources/scripts/mayaUsdCacheMayaReference.py
```diff
@@ -36,7 +36,7 @@
 _cacheExportOptions = None
 
 # The options string that we pass to mayaUsdTranslatorExport.
-kTranslatorExportOptions = 'all;!output-parentscope;!animation-data'
+kTranslatorExportOptions = 'all;!output-parentscope'
```
pierrebai-adsk, line 39: We now show the animation check-box UI.

lib/mayaUsd/resources/scripts/mayaUsdOptions.py
```diff
@@ -41,10 +56,28 @@ def convertOptionsDictToText(optionsDict):
...
     """
-    optionsList = ['%s=%s' % (name, value) for name, value in optionsDict.items()]
+    optionsList = ['%s=%s' % (name, _convertValueToText(value)) for name, value in optionsDict.items()]
     return ';'.join(optionsList)
 
 
+def _convertValueToText(value):
+    # Unfortunately, for historical reasons, text lists have used
+    # comma-separated values, while floating-point used space-separated values.
+    if isinstance(value,list):
```
pierrebai-adsk, line 66: We need to support backward compatibility with existing user prefs or scripts, so unfortunately we need to support both space and comma as separator, for different data types.

lib/mayaUsd/utils/util.cpp
```diff
@@ -1886,6 +1886,10 @@ VtValue _ParseArgumentValue(const std::string& value, const VtValue& guideValue)
     // The export UI only has boolean and string parameters.
     if (guideValue.IsHolding<bool>()) {
         return VtValue(TfUnstringify<bool>(value));
+    } else if (guideValue.IsHolding<double>()) {
```
pierrebai-adsk, line 1889: This is code used by the text-to-VTDictionary converted.

lib/mayaUsd/resources/scripts/mayaUsdOptions.py
```diff
@@ -41,10 +56,28 @@ def convertOptionsDictToText(optionsDict):
...
+        # Note: empty list must create empty text.
+        if not len(value):
```
seando-adsk, line 68: What you have is okay, but you can also use 
```suggestion
        if not value:
```

seando-adsk (approved):
Everything looks good to me. Great job. My only comment would be is there any unit test that can be written to test some of these changes?

pierrebai-adsk:
The existing edit-as-maya and merge-to-usd already test some of it. The problem is that most of the changes are UI or there to support the UI for adding a Maya reference and merge options and those cannot really be tested in unit tests?

lib/mayaUsd/resources/scripts/mayaUsdOptions.py
pierrebai-adsk: True! OTOH, the code as written makes clear that its the length that we're diverging on, not any other falsy reason, like the variable being None. If there were other changes needed I'd do it, but I don't think it is worth a new batch of pre-flight.

pierrebai-adsk: OTOH, github let us commit the change right here, so if you feel like it, you can commit your suggestion before merging.

pierrebai-adsk:
I could add a unit test for merge-to-USD that at least passes some user arguments. I'll do that and also your suggested change.

seando-adsk (dismissed):
Great work on the tests. Thanks.

pierrebai-adsk:
Sorry, one of the test was actually not run because its name did not begin with 'test'... and of course it failed!

seando-adsk (approved):

PR #2151, opened 2022-03-01
MAYA-121937 - Renaming 2 prims w/illegal chars will crash Maya

seando-adsk:
MAYA-121937 - Renaming 2 usd items with illegal characters will cause Maya to crash
* Validate the name before making it unique.


ppt-adsk:
Can we add the crash repro steps to the automated test for this?

seando-adsk:
@ppt-adsk Ah yes good idea. I'll add a test.

pierrebai-adsk:
Good catch on the other cases of the same bug.

pierrebai-adsk (dismissed):

pierrebai-adsk (approved):

PR #2156, opened 2022-03-02
MAYA-121781 support duplicate to USD non-root

pierrebai-adsk:


pierrebai-adsk:
Note that duplicate to Maya also fails to support non-root destination, but there the process is done by the importer, which does not support specifying the destination, so the fix would be more complex. It might involve mass-moving object after import, and managing name conflicts. Beside, the ticket only mentioned the to-USD side, so this simple fix addresses that. If we want to address the to-Maya side, we should open a new ticket for that.

seando-adsk (approved):
Great work. Thanks for the test.

PR #2158, opened 2022-03-03
MAYA-107839 [GitHub 800] USD: exporting the attached maya file using …

boudrey:
…USD takes extremely long

I was able to cut in half the time to export going from about 10 min. to 5 min with an improved algorithm for shapes below dagpath.
The next bottleneck is here with 87% of the time spent: https://github.com/Autodesk/maya-usd/blob/dev/lib/usd/translators/nurbsCurveWriter.cpp#L61

### DIFF
--- a/lib/mayaUsd/fileio/writeJobContext.cpp
+++ b/lib/mayaUsd/fileio/writeJobContext.cpp
@@ -120,6 +120,8 @@ bool UsdMayaWriteJobContext::IsMergedTransform(const MDagPath& path) const
 
     // Any transform with multiple (non-intermediate) shapes below is
     // non-mergeable.
+    // Expanded out the implementation of MDagPath::numberOfShapesDirectlyBelow() to cut down
+    // the time by half.
     unsigned int numShapes = 0;
     const auto   childCount = path.childCount();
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/fileio/writeJobContext.cpp
```diff
@@ -120,9 +120,21 @@ bool UsdMayaWriteJobContext::IsMergedTransform(const MDagPath& path) const
 
     // Any transform with multiple (non-intermediate) shapes below is
     // non-mergeable.
-    unsigned int numberOfShapesDirectlyBelow = 0u;
-    path.numberOfShapesDirectlyBelow(numberOfShapesDirectlyBelow);
```
boudrey, line 124: I translated the code of numberOfShapesDirectlyBelow in OpenMaya code and stop if there is more than one shape instead of going through all the shapes to count them since it is not necessary saving 1/2 of the export time,

ppt-adsk: Thanks Yves.  Can you add in a comment saying that you expanded out the implementation of MDagPath::numberOfShapesDirectlyBelow() to cut down the time by half.  Otherwise, someone will come in and replace your code with numberOfShapesDirectlyBelow().

PR #2159, opened 2022-03-03
Maya-121344 - Implementing proper fallback point shader using a new geometry shader fragment

vlasovi:
- Implemented geometry shader fragment for points in PointsGeometry.xml
 - Implemented fallback shaders for points PointsFallbackShader.xml, PointsFallbackCPVShader.xml
 - Updated autotests


lib/mayaUsd/render/vp2RenderDelegate/points.cpp
```diff
@@ -266,7 +274,7 @@ void HdVP2Points::_UpdateDrawItem(
             // Using a zero vector to indicate requirement of camera-facing
```
williamkrick, line 274: This comment needs to be updated, it is not a zero vector anymore.

williamkrick (approved):

PR #2161, opened 2022-03-04
MAYA-121808 Add curve filtering option in UI

JGamache-autodesk:
Using the existing filtering mechanism.

### DIFF
--- a/plugin/adsk/scripts/mayaUsdTranslatorExport.mel
+++ b/plugin/adsk/scripts/mayaUsdTranslatorExport.mel
@@ -131,6 +131,17 @@ proc string mayaUsdTranslatorExport_AppendFromCheckbox(string $currentOptions, s
     }
 }
 
+proc string mayaUsdTranslatorExport_SetFromCheckbox(string $currentOptions, string $trueValue, string $falseValue, string $widget) {
+    if (`checkBoxGrp -exists $widget` == 0)
+        return $currentOptions;
+
+    if (`checkBoxGrp -q -v1 $widget` == 1) {
+        return $currentOptions + ";" + $trueValue;
+    } else {
+        return $currentOptions + ";" + $falseValue;
+    }
+}
+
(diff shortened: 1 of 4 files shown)


JGamache-autodesk:
Preflight succeeded. One Linux machine had GPU issues though.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -659,6 +688,8 @@ global proc int mayaUsdTranslatorExport (string $parent,
             $currentOptions = mayaUsdTranslatorExport_AppendFromPopup($currentOptions, "exportSkin", "skinClustersPopup");
             $currentOptions = mayaUsdTranslatorExport_AppendFromCheckbox($currentOptions, "exportBlendShapes", "exportBlendShapesCheckBox");
             $currentOptions = mayaUsdTranslatorExport_AppendFromCheckbox($currentOptions, "exportDisplayColor", "exportDisplayColorCheckBox");
+            // We currently have only one type to filter. In the future, we might need to build the array separately as more types are added.
+            $currentOptions = mayaUsdTranslatorExport_SetFromCheckbox($currentOptions, "", "filterTypes=nurbsCurve", "exportCurvesCheckBox");
```
seando-adsk, line 692: What you are saying here is that currently you cannot call mayaUsdTranslatorExport_SetFromCheckbox() again with a different filterType because then the option string would have two filterType args. Is that right. Instead we would need extra code here to build up the array of filter types?

seando-adsk (approved):
Great work. Thanks for the test.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
JGamache-autodesk: Exactly. In a future where we might add more prim filtering checkboxes, the filterType array will have to be built by reviewing all checkboxes before being assigned to the filterType export parameter.

PR #2164, opened 2022-03-07
MAYA-121886 Merge-to USD options help

pierrebai-adsk:
- Use the showHelp command.
- Use the content-ID UsdMergeMayaEditsToUsd
- Hide the menu if the content is not available.


pierrebai-adsk:
The only PF failure is the Linux interactive on GPU which is currently known to fail due to image comparison.

PR #2166, opened 2022-03-07
MAYA-122014 support duplicate to Maya to non root

pierrebai-adsk:
- Add isMayaRootPath helper function.
- Use the destination path when duplicating as Maya.
- Add a unit test.

### DIFF
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -372,6 +372,21 @@ MDagPath ufeToDagPath(const Ufe::Path& ufePath)
     );
 }
 
+bool isMayaRootPath(const Ufe::Path& ufePath)
+{
+    if (ufePath.runTimeId() != g_MayaRtid)
+        return false;
+
+    const auto& segments = ufePath.getSegments();
+    if (segments.size() != 1)
+        return false;
+
+    if (segments[0].components().size() != 1)
+        return false;
+
+    return true;
+}
+
 PXR_NS::MayaUsdProxyShapeBase* getProxyShape(const Ufe::Path& path)
 {
     // Path should not be empty.
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -372,19 +372,9 @@ MDagPath ufeToDagPath(const Ufe::Path& ufePath)
     );
 }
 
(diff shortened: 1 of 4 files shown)


pierrebai-adsk:
The only PF failure is the Linux interactive on GPU which is currently known to fail due to image comparison.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -372,6 +372,21 @@ MDagPath ufeToDagPath(const Ufe::Path& ufePath)
     );
 }
 
+bool isMayaRootPath(const Ufe::Path& ufePath)
```
ppt-adsk, line 375: Could be simply 

return (ufePath.runTimeId() == g_MayaRtid) && (ufePath.size() == 1);

lib/mayaUsd/ufe/Utils.h
```diff
@@ -125,6 +125,10 @@ Ufe::PathSegment dagPathToPathSegment(const MDagPath& dagPath);
 MAYAUSD_CORE_PUBLIC
 MDagPath ufeToDagPath(const Ufe::Path& ufePath);
 
+//! Verify if the UFE path is the Maya root.
```
ppt-adsk, line 128: Hmmm, wonder about the name.  In Maya the root node is called world, and it's called that way in Maya commands and documentation as well.  Perhaps call it isMayaWorldPath()?

test/lib/mayaUsd/fileio/testDuplicateAs.py
```diff
@@ -86,6 +86,39 @@ def testDuplicateAsMaya(self):
...
+        (_, _, aXlation, aUsdUfePathStr, aUsdUfePath, _, _, 
+               bXlation, bUsdUfePathStr, bUsdUfePath, _) = \
+            createSimpleXformScene()
+
+        xform = cmds.createNode('transform')
+        xformNames = cmds.ls(xform)
+        self.assertEqual(1, len(xformNames))
+        xformName = xformNames[0]
+
+        # Duplicate USD data as Maya data, placing it under the root.
```
ppt-adsk, line 101: Comment is incorrect.

ppt-adsk (changes requested):
Thanks for the fix!  Minor tweaks only.

ppt-adsk (approved):

pierrebai-adsk:
Again, image diff on linux error. We can ignore, unrelated to this PR.

PR #2171, opened 2022-03-08
MAYA-110239 remove load / unload payload from AE

pierrebai-adsk:


pierrebai-adsk:
Only the image comparison in the test testVP2RenderDelegateMaterialX on Linux fails. Image comparisons have randomly failed recently and is unrelated to my changes.

boudrey (approved):
LGTM

PR #2172, opened 2022-03-09
MAYA-121808 Fix the curve export tooltip

JGamache-autodesk:


JGamache-autodesk:
Does not require preflight. Only a tooltip update.

boudrey (approved):

PR #2173, opened 2022-03-09
T blais/maya 121277/investigate creating usd mesh node

simonblaisautodesk:
This is part from a long term goal to simplify the maya mesh file by embedding 'usd' or other format directly in the scene.
The performance angle is to be able to load on demand the geometry, so if the node is not visible it would not even load the file immediately. When this work we could have low res and high res version in the same Maya file without paying memory and load time cost for the high res, as long as the user work with the low res. But this pull request here simply add the possibility to prototype things further in maya.

### DIFF
--- a/lib/mayaUsd/fileio/jobs/meshDataReadJob.h
+++ b/lib/mayaUsd/fileio/jobs/meshDataReadJob.h
@@ -1,0 +1,62 @@
+//
+// Copyright 2022 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef PXRUSDMAYA_MESHDATAREAD_JOB_H
+#define PXRUSDMAYA_MESHDATAREAD_JOB_H
+
+#include <mayaUsd/fileio/jobs/readJob.h>
+
+#include <maya/MObject.h>
+
+#include <vector>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+// Custom class to prevent read job from creating new nodes and to access
+// geometric data directly.
+class UsdMaya_MeshDataReadJob : public UsdMaya_ReadJob
+{
+public:
+    MAYAUSD_CORE_PUBLIC
+    UsdMaya_MeshDataReadJob(
+        const MayaUsd::ImportData&  iImportData,
+        const UsdMayaJobImportArgs& iArgs);
+
+protected:
+    // Override prim reader to create a mesh without a node and to store
+    // the mesh then access it and return it in the plug.
+    bool OverridePrimReader(
+        const UsdPrim&               usdRootPrim,
+        const UsdPrim&               prim,
+        const UsdMayaPrimReaderArgs& args,
+        UsdMayaPrimReaderContext&    readCtx,
+        UsdPrimRange::iterator&      primIt) override;
+
+public:
+    struct Data
+    {
+        MObject geometry;
+        MObject matrix;
+    };
+
+    std::vector<Data> mMeshData;
+
+private:
+    std::map<SdfPath, GfMatrix4d> mTransforms;
+};
+
+PXR_NAMESPACE_CLOSE_SCOPE
+
+#endif
--- a/lib/mayaUsd/fileio/translators/translatorMesh.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorMesh.cpp
@@ -198,11 +198,15 @@ TranslatorMeshRead::TranslatorMeshRead(
     // set mesh name
     const auto& primName = prim.GetName().GetString();
     const auto  shapeName = TfStringPrintf("%sShape", primName.c_str());
-    meshFn.setName(MString(shapeName.c_str()), false, &stat);
 
-    if (!stat) {
-        *status = stat;
-        return;
+    // Set the mesh name if creating a maya mesh object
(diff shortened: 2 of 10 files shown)


lib/mayaUsd/fileio/jobs/meshDataReadJob.h
```diff
@@ -0,0 +1,62 @@
...
+{
+public:
+    MAYAUSD_CORE_PUBLIC
+    UsdMaya_MeshDataReadJob(
+        const MayaUsd::ImportData&  iImportData,
+        const UsdMayaJobImportArgs& iArgs);
+
+protected:
+    // Override prim reader to create a mesh without a node and to store
+    // the mesh then access it and return it in the plug.
```
pierrebai-adsk, line 39: The comment talks about "the plug" but there are no plug in sight. I assume it is talking about some plug of the caller? It would help the reader to c;larify this.

lib/mayaUsd/fileio/translators/translatorMesh.cpp
```diff
@@ -198,11 +198,15 @@ TranslatorMeshRead::TranslatorMeshRead(
     // set mesh name
     const auto& primName = prim.GetName().GetString();
     const auto  shapeName = TfStringPrintf("%sShape", primName.c_str());
-    meshFn.setName(MString(shapeName.c_str()), false, &stat);
 
-    if (!stat) {
-        *status = stat;
-        return;
+    // Set the mesh name if creating a maya mesh object
+    if (transformObj.isNull() || MFn::kMeshData != transformObj.apiType()) {
```
pierrebai-adsk, line 203: Maybe for someone knowing the code it is clear, but to me the comment talking about creating a mesh and the actual tests being done are not correlated, especially since the comment says to set the name of it is a mesh yet it sets the name if the type is **not** mesh-data.

I guess both these tests detect "a mesh object", but it is not obvious.

plugin/adsk/plugin/geomNode.cpp
```diff
@@ -0,0 +1,236 @@
...
+// Attributes
+MObject MayaUsdGeomNode::filePathAttr;
+MObject MayaUsdGeomNode::rootPrimAttr;
+
+// Output attributes
+MObject MayaUsdGeomNode::outGeomAttr;
+MObject MayaUsdGeomNode::outGeomMatrixAttr;
+
+namespace {
+class OutputArrayHandler
```
pierrebai-adsk, line 49: A comment explaining the purpose of the class would be nice. I get it that it's a handler for output array, but what does that do and why is it needed.

plugin/adsk/plugin/geomNode.h
```diff
@@ -0,0 +1,73 @@
...
+#include <maya/MPxNode.h>
+#include <maya/MStatus.h>
+
+#include <memory>
+#include <vector>
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+namespace MAYAUSD_NS_DEF {
+class MayaUsdGeomNode : public MPxNode
```
pierrebai-adsk, line 33: A comment for the purpose of the class. I know we're not always examplar in doing this ourselves... :)

plugin/adsk/plugin/geomNode.cpp
```diff
@@ -0,0 +1,236 @@
...
+    CHECK_MSTATUS_AND_RETURN_IT(retValue);
+
+    return retValue;
+}
+
+MayaUsdGeomNode::MayaUsdGeomNode()
+    : MPxNode()
+    , _cache(std::make_unique<CacheData>())
+{
+    TfRegistryManager::GetInstance().SubscribeTo<MayaUsdGeomNode>();
```
pierrebai-adsk, line 146: I work on USD and I don't even know what this does. A comment welcome if you know.

plugin/adsk/plugin/geomNode.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+        MDataHandle rootPrimHandle = dataBlock.inputValue(rootPrimAttr, &retValue);
+        CHECK_MSTATUS_AND_RETURN_IT(retValue);
+
+        const std::string fileName = TfStringTrim(filePathHandle.asString().asChar());
+        const std::string rootPrim = TfStringTrim(rootPrimHandle.asString().asChar());
+
+        if (!fileName.empty()) {
+            if (_cache->loadedFile.empty() || _cache->loadedFile != fileName
+                || _cache->rootPrim != rootPrim) {
```
pierrebai-adsk, line 169: Out of curiosity: will there be a mechanism to detect if the file is changed and reload, or is this under the user responsibility or is the cache considered private and should only be updated by Maya?

plugin/adsk/plugin/geomNode.cpp
```diff
@@ -0,0 +1,236 @@
...
+                MayaUsd::ImportData readData;
+                readData.setFilename(fileName);
+
+                if (!rootPrim.empty()) {
```
pierrebai-adsk, line 173: Could mention that if unset the default is "/".

plugin/adsk/plugin/geomNode.cpp
```diff
@@ -0,0 +1,236 @@
...
+            }
+
+            OutputArrayHandler geomOut(dataBlock, outGeomAttr, _cache->primitives.size(), retValue);
+            CHECK_MSTATUS_AND_RETURN_IT(retValue);
+
+            OutputArrayHandler matrixOut(
+                dataBlock, outGeomMatrixAttr, _cache->primitives.size(), retValue);
+            CHECK_MSTATUS_AND_RETURN_IT(retValue);
+
+            // MDataHandle outputGeomHandle = dataBlock.outputValue(outGeomAttr, &retValue);
```
pierrebai-adsk, line 203: Code in comments voluntary?

pierrebai-adsk (changes requested):
LGTM, except minor things.

plugin/adsk/plugin/geomNode.cpp
simonblaisautodesk: Oops, these are leftovers, I will remove them.

pierrebai-adsk (dismissed):

ppt-adsk (approved):

PR #2174, opened 2022-03-09
MAYA-121258 remove plugin menus

pierrebai-adsk:
The callback that created the menus were not removed.

### DIFF
--- a/plugin/adsk/scripts/mayaUsdMenu.mel
+++ b/plugin/adsk/scripts/mayaUsdMenu.mel
@@ -287,10 +287,9 @@ proc string getMayaMajorVersion() {
 // setup the items in Maya's "Create" menu
 global proc mayaUsdMenu_createMenuCallback() {
     global string $gMayaUsdCreateSubMenu;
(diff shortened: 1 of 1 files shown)


plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -287,10 +287,9 @@ proc string getMayaMajorVersion() {
 // setup the items in Maya's "Create" menu
 global proc mayaUsdMenu_createMenuCallback() {
     global string $gMayaUsdCreateSubMenu;
+    global string $gMainCreateMenu; // maya's create menu
 
     if ($gMayaUsdCreateSubMenu == "") {    
-        global string $gMainCreateMenu; // maya's create menu
-
```
boudrey, line 293: This move was not necessary, gMainCreateMenu is not used outside.

boudrey (changes requested):
Put back gMainCreateMenu where is was or use the variable outside.

boudrey (approved):

PR #2175, opened 2022-03-09
Locked attributes on Maya reference prim transform stand-in.

ppt-adsk:
Design wants all attributes except transform attributes to be locked on the Maya transform node stand-in which is created for the Maya reference prim on pull.  There might be other attributes that we will decide we don't want to lock, so this may evolve after we gain experience.

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -435,13 +435,13 @@ bool pullCustomize(const PullImportPaths& importedPaths, const UsdMayaPrimUpdate
         // If the Maya node holds USD type information (e.g. a dummy transform
         // node which is a stand-in for a non-transform USD prim type), use the
         // USD type instead.
-        auto              usdTypeNamePlug = dgNodeFn.findPlug("USD_typeName", true);
-        bool              useUsdType = !usdTypeNamePlug.isNull();
-        const std::string mayaTypeName(dgNodeFn.typeName().asChar());
+        auto       usdTypeNamePlug = dgNodeFn.findPlug("USD_typeName", true);
+        const bool useUsdType = !usdTypeNamePlug.isNull();
 
-        auto registryItem = useUsdType ? UsdMayaPrimUpdaterRegistry::FindOrFallback(
-                                TfToken(usdTypeNamePlug.asString().asChar()))
-                                       : UsdMayaPrimUpdaterRegistry::FindOrFallback(mayaTypeName);
+        auto registryItem = useUsdType
+            ? UsdMayaPrimUpdaterRegistry::FindOrFallback(
+                TfToken(usdTypeNamePlug.asString().asChar()))
+            : UsdMayaPrimUpdaterRegistry::FindOrFallback(std::string(dgNodeFn.typeName().asChar()));
         auto factory = std::get<UsdMayaPrimUpdaterRegistry::UpdaterFactoryFn>(registryItem);
         auto updater = factory(context, dgNodeFn, pulledUfePath);
 
@@ -543,6 +543,17 @@ SdfPath getDstSdfPath(const Ufe::Path& ufePulledPath, const SdfPath& srcSdfPath,
 // Create an updater for use with both pushCustomize() traversals /
 // customization points: pushCopySpec() and pushEnd().
 //
+// pushCopySpec() and pushEnd() must use the same updater type.  An earlier
+// version of this function tried to ensure this by using the pulled prim to
+// create the updater.  However, this prim cannot be relied on, as
+// pushCopySpec() has an edit router customization point that can remove the
+// pulled prim from the USD scene (e.g. by switching a variant set to a
+// different variant, such as what occurs when caching to a variant).  It is
+// more robust to use the USD primSpec type at srcPath, which is in the
+// srcLayer in the temporary stage.  If USD type round-tripping is set up
+// properly (see UsdMayaTranslatorUtil::CreateDummyTransformNode()), this
+// primSpec will have the type of the original pulled prim.
+//
 UsdMayaPrimUpdaterSharedPtr createUpdater(
     const SdfLayerRefPtr&            srcLayer,
     const SdfPath&                   srcPath,
@@ -605,8 +616,7 @@ bool pushCustomize(
     // Traverse the layer, creating a prim updater for each primSpec
     // along the way, and call PushCopySpec on the prim.
     auto pushCopySpecsFn
-        = [&context, &ufePulledPath, srcStage, srcLayer, dstLayer, dstRootParentPath](
-              const SdfPath& srcPath) {
+        = [&context, srcStage, srcLayer, dstLayer, dstRootParentPath](const SdfPath& srcPath) {
               // We can be called with a primSpec path that is not a prim path
               // (e.g. a property path like "/A.xformOp:translate").  This is not an
               // error, just prune the traversal.  FIXME Is this still true?  We
@@ -651,8 +661,7 @@ bool pushCustomize(
 
     // SdfLayer::TraversalFn does not return a status, so must report
     // failure through an exception.
-    auto pushEndFn = [&context, &ufePulledPath, srcLayer, dstLayer, dstRootParentPath](
-                         const SdfPath& srcPath) {
+    auto pushEndFn = [&context, srcLayer, dstLayer, dstRootParentPath](const SdfPath& srcPath) {
         // We can be called with a primSpec path that is not a prim path
         // (e.g. a property path like "/A.xformOp:translate").  This is not an
         // error, just a no-op.
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -58,7 +58,7 @@
 #include <tuple>
 
 using UpdaterFactoryFn = UsdMayaPrimUpdaterRegistry::UpdaterFactoryFn;
-using namespace MAYAUSD_NS_DEF;
+using namespace MayaUsd;
 
 // Allow for use of MObjectHandle with std::unordered_map.
 namespace std {
@@ -134,8 +134,8 @@ SdfPath makeDstPath(const SdfPath& dstRootParentPath, const SdfPath& srcPath)
 
 //------------------------------------------------------------------------------
 //
-// The UFE path and the prim refer to the same object: the prim is passed in as
-// an optimization to avoid an additional call to ufePathToPrim().
+// The UFE path is to the pulled prim, and the Dag path is the corresponding
+// Maya pulled object.
 bool writePullInformation(const Ufe::Path& ufePulledPath, const MDagPath& path)
 {
     auto pulledPrim = MayaUsd::ufe::ufePathToPrim(ufePulledPath);
@@ -432,9 +432,16 @@ bool pullCustomize(const PullImportPaths& importedPaths, const UsdMayaPrimUpdate
         const auto&       pulledUfePath = *ufePathIt;
         MFnDependencyNode dgNodeFn(dagPath.node());
 
+        // If the Maya node holds USD type information (e.g. a dummy transform
+        // node which is a stand-in for a non-transform USD prim type), use the
+        // USD type instead.
+        auto              usdTypeNamePlug = dgNodeFn.findPlug("USD_typeName", true);
+        bool              useUsdType = !usdTypeNamePlug.isNull();
         const std::string mayaTypeName(dgNodeFn.typeName().asChar());
 
-        auto registryItem = UsdMayaPrimUpdaterRegistry::FindOrFallback(mayaTypeName);
+        auto registryItem = useUsdType ? UsdMayaPrimUpdaterRegistry::FindOrFallback(
+                                TfToken(usdTypeNamePlug.asString().asChar()))
+                                       : UsdMayaPrimUpdaterRegistry::FindOrFallback(mayaTypeName);
         auto factory = std::get<UsdMayaPrimUpdaterRegistry::UpdaterFactoryFn>(registryItem);
         auto updater = factory(context, dgNodeFn, pulledUfePath);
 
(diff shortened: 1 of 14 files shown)


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -134,8 +134,8 @@ SdfPath makeDstPath(const SdfPath& dstRootParentPath, const SdfPath& srcPath)
 
 //------------------------------------------------------------------------------
 //
-// The UFE path and the prim refer to the same object: the prim is passed in as
-// an optimization to avoid an additional call to ufePathToPrim().
+// The UFE path is to the pulled prim, and the Dag path is the corresponding
```
ppt-adsk, line 137: Fixed comment while I was at it.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -432,9 +432,16 @@ bool pullCustomize(const PullImportPaths& importedPaths, const UsdMayaPrimUpdate
         const auto&       pulledUfePath = *ufePathIt;
         MFnDependencyNode dgNodeFn(dagPath.node());
 
+        // If the Maya node holds USD type information (e.g. a dummy transform
+        // node which is a stand-in for a non-transform USD prim type), use the
+        // USD type instead.
+        auto              usdTypeNamePlug = dgNodeFn.findPlug("USD_typeName", true);
```
ppt-adsk, line 438: This is a crucial improvement to create the proper updater: if USD type information was stored on the Maya node on import / pull (e.g. for dummy transform nodes used for Maya reference prims or scope prims), use the USD type information.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -533,31 +540,21 @@ SdfPath getDstSdfPath(const Ufe::Path& ufePulledPath, const SdfPath& srcSdfPath,
...
-    const bool usePulledPrim = (srcPath.GetPathElementCount() == 1);
-
     auto primSpec = srcLayer->GetPrimAtPath(srcPath);
     if (!TF_VERIFY(primSpec)) {
         return nullptr;
     }
 
-    TfToken typeName = usePulledPrim ? MayaUsd::ufe::ufePathToPrim(ufePulledPath).GetTypeName()
-                                     : primSpec->GetTypeName();
+    auto typeName = primSpec->GetTypeName();
```
ppt-adsk, line 557: Another crucial improvement: because the pull / push round trip is now USD type-preserving, we can now simply use the USD type when creating updaters during the pushCopySpecs() and pushEnd() temporary stage traversals.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -432,9 +432,16 @@ bool pullCustomize(const PullImportPaths& importedPaths, const UsdMayaPrimUpdate
         const auto&       pulledUfePath = *ufePathIt;
         MFnDependencyNode dgNodeFn(dagPath.node());
 
+        // If the Maya node holds USD type information (e.g. a dummy transform
```
pierrebai-adsk, line 435: I think a "no" is missing in the sentence: "If the Maya node holds *no* USD type informtation..."?

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -619,8 +616,7 @@ bool pushCustomize(
               }
 
               auto dstPath = makeDstPath(dstRootParentPath, srcPath);
-              auto updater
-                  = createUpdater(ufePulledPath, srcLayer, srcPath, dstLayer, dstPath, context);
+              auto updater = createUpdater(srcLayer, srcPath, dstPath, context);
```
ppt-adsk, line 619: Didn't need the ufePulledPath any more, and dstLayer was unused, so removed them.

lib/mayaUsd/fileio/translators/translatorMayaReference.cpp
```diff
@@ -38,10 +38,8 @@
 #include <maya/MDGModifier.h>
 #include <maya/MFileIO.h>
 #include <maya/MFileObject.h>
-#include <maya/MFnCamera.h>
```
ppt-adsk, line 41: Removed useless includes while I was at it (MFnTransform and MFnCamera for Maya reference translation?)

lib/mayaUsd/fileio/translators/translatorUtil.h
```diff
@@ -66,13 +72,15 @@ struct UsdMayaTranslatorUtil
     /// empty string, so a typeless def will be generated on export.
     MAYAUSD_CORE_PUBLIC
     static bool CreateDummyTransformNode(
```
ppt-adsk, line 74: When creating a dummy transform for Maya reference prims, the resulting Maya transform node stand-in must be transformable, so added a parameter to control transform locking on dummy transform node creation.

lib/mayaUsd/undo/OpUndoItems.h
```diff
@@ -474,6 +474,13 @@ class UfeSelectionUndoItem : public OpUndoItem
 
 /// \class LockNodesUndoItem
 /// \brief Record data needed to undo / redo the lock / unlock of Maya nodes.
+///
+/// The node at the Dag path root, and all its children, will be locked.
```
ppt-adsk, line 478: Added documentation while I was here.

lib/mayaUsd/utils/editRouter.cpp
```diff
@@ -97,7 +95,7 @@ void cacheMayaReference(const PXR_NS::VtDictionary& context, PXR_NS::VtDictionar
 
     auto createCachePrimFn = [stage, dstLayer, dstPrimPath](
                                  const PXR_NS::SdfPath& primPath, bool asReference, bool append) {
-        auto cachePrim = stage->DefinePrim(primPath);
+        auto cachePrim = stage->DefinePrim(primPath, PXR_NS::TfToken("Xform"));
```
ppt-adsk, line 98: The cache prim is immediately created with Xform type, rather than trying to set its type later.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -432,9 +432,16 @@ bool pullCustomize(const PullImportPaths& importedPaths, const UsdMayaPrimUpdate
         const auto&       pulledUfePath = *ufePathIt;
         MFnDependencyNode dgNodeFn(dagPath.node());
 
+        // If the Maya node holds USD type information (e.g. a dummy transform
+        // node which is a stand-in for a non-transform USD prim type), use the
+        // USD type instead.
+        auto              usdTypeNamePlug = dgNodeFn.findPlug("USD_typeName", true);
+        bool              useUsdType = !usdTypeNamePlug.isNull();
         const std::string mayaTypeName(dgNodeFn.typeName().asChar());
```
pierrebai-adsk, line 440: Really small opt: if we expect that usdTypeNamePlug will most often be valid, then moving the mayaTypeName code right inside the call to the other branch of the ?: op would avoid building that std::string most of the time.

lib/mayaUsd/utils/editRouter.cpp
```diff
@@ -116,13 +114,15 @@ void cacheMayaReference(const PXR_NS::VtDictionary& context, PXR_NS::VtDictionar
         context, "rn_payloadOrReference", PXR_NS::VtDefault = "");
     bool dstIsVariant
         = (PXR_NS::VtDictionaryGet<int>(context, "rn_defineInVariant", PXR_NS::VtDefault = 0) == 1);
+    const auto parentPath = pulledPath.GetParentPath();
```
ppt-adsk, line 117: Factored out common code.

lib/usd/translators/mayaReferenceReader.cpp
```diff
@@ -52,8 +52,19 @@ PXRUSDMAYA_DEFINE_READER(MayaUsd_SchemasMayaReference, args, context)
 
     MObject referenceParentNode;
 
-    UsdMayaTranslatorUtil::CreateTransformNode(
-        usdPrim, parentNode, args, &context, &status, &referenceParentNode);
+    // The stand-in for a Maya reference prim is a dummy transform node, which
+    // preserves the source USD prim type in attribute "USD_typeName".  To
+    // support workflows where this dummy transform node has its transform
+    // changed, we leave its transform unlocked.
+    UsdMayaTranslatorUtil::CreateDummyTransformNode(
```
ppt-adsk, line 59: Important change: translating a Maya reference prim now creates a dummy transform node, which is USD type-preserving.  Creating a plain Maya transform node does not preserve USD type information, which makes pull / push round-tripping very difficult, since we wish to use the same updater for push as was used for pull.

lib/usd/translators/mayaReferenceReader.cpp
```diff
@@ -52,8 +52,19 @@ PXRUSDMAYA_DEFINE_READER(MayaUsd_SchemasMayaReference, args, context)
...
+        usdPrim,
+        parentNode,
+        /*importTypeName*/ true,
+        args,
+        &context,
+        &status,
+        &referenceParentNode,
+        UsdMayaDummyTransformType::UnlockedTransform);
```
ppt-adsk, line 67: New formatting courtesy of clang-format.

lib/mayaUsd/fileio/primUpdaterManager.cpp
ppt-adsk: Trying to use the pulled prim was buggy, because in the pushEnd() traversal the Maya reference prim no longer exists when using the cache to variant workflow!  It has been replaced by the cache variant.

lib/usd/translators/mayaReferenceUpdater.cpp
```diff
@@ -108,6 +112,49 @@ bool PxrUsdTranslators_MayaReferenceUpdater::shouldAutoEdit() const
...
+    static std::set<std::string> xformAttrNames;
+    if (xformAttrNames.empty()) {
+        for (const auto& opClass : UsdMayaXformStack::MayaStack().GetOps()) {
+            if (!opClass.IsInvertedTwin()) {
+                xformAttrNames.insert(std::string(opClass.GetName().GetText()));
+            }
+        }
+    }
+
+    MFnDependencyNode fn(getMayaObject());
```
ppt-adsk, line 139: Core change: lock all attributes except transform attributes.  Re-used dummy transform creation algorithm to determine what is considered a transform attribute.

test/lib/mayaUsd/fileio/testCacheToUsd.py
```diff
@@ -45,6 +45,45 @@
 
 import os
 
+def createMayaRefPrimSiblingCache(testCase, cacheParent, cacheParentPathStr):
```
ppt-adsk, line 48: Factored out differences between sibling cache and variant set cache workflows into functions passed to the test engine function.

test/lib/mayaUsd/fileio/testCacheToUsd.py
```diff
@@ -69,17 +108,17 @@ def setUp(self):
         self.proxyShapePathStr = mayaUsd_createStageWithNewLayer.createStageWithNewLayer()
         self.stage = mayaUsd.lib.GetPrim(self.proxyShapePathStr).GetStage()
 
-    def testCacheToUsd(self):
-        '''Cache edits on a pulled Maya reference prim back to USD.'''
+    def runTestCacheToUsd(self, createMayaRefPrimFn, checkCacheParentFn):
```
ppt-adsk, line 111: Test engine function that is called to run both the sibling cache and variant set cache tests.

test/lib/mayaUsd/fileio/testCacheToUsd.py
```diff
@@ -195,5 +236,14 @@ def testCacheToUsd(self):
 
         self.assertTrue(foundPayload)
 
+        # Clean up
+        os.remove(cacheFile)
+
+    def testCacheToUsdSibling(self):
+        self.runTestCacheToUsd(createMayaRefPrimSiblingCache, checkSiblingCacheParent)
+
+    def testCacheToUsdVariant(self):
```
ppt-adsk, line 245: Added a test for variant set cache workflow.

lib/usd/translators/mayaReferenceReader.cpp
```diff
@@ -52,8 +52,19 @@ PXRUSDMAYA_DEFINE_READER(MayaUsd_SchemasMayaReference, args, context)
 
     MObject referenceParentNode;
 
-    UsdMayaTranslatorUtil::CreateTransformNode(
-        usdPrim, parentNode, args, &context, &status, &referenceParentNode);
+    // The stand-in for a Maya reference prim is a dummy transform node, which
+    // preserves the source USD prim type in attribute "USD_typeName".  To
```
pierrebai-adsk, line 56: I suppose a future extension would be to use this attribute for any node types that wish to clearly specify the USD node it wishes to come back as?

lib/usd/translators/mayaReferenceUpdater.cpp
```diff
@@ -108,6 +112,49 @@ bool PxrUsdTranslators_MayaReferenceUpdater::shouldAutoEdit() const
...
+        auto attr = fn.attribute(i);
+        auto plug = fn.findPlug(attr, true);
+        if (plug.isChild()) {
+            // Skip children of compounds, as their parent will be locked.
+            continue;
+        }
+        // If the attribute is not in our set of transform attributes, lock it.
+        MFnAttribute fnAttr(attr);
+        if (xformAttrNames.count(std::string(fnAttr.name().asChar())) == 0) {
+            plug.setLocked(true);
```
pierrebai-adsk, line 151: As I understand, this part is not undoable directly? (I'll keep reading, maybe the undo is handled manually later.)

pierrebai-adsk (approved):
Apart from the few small comments / questions, LGTM!

lib/mayaUsd/fileio/primUpdaterManager.cpp
ppt-adsk: No, the sentence is correct: we look at the Maya node, and if there is USD type information recorded in it, we will use the USD type to do updater lookup, rather than use the Maya type.  If you have a wording to suggest that would make this clearer, let me know!

lib/mayaUsd/fileio/primUpdaterManager.cpp
ppt-adsk: Makes perfect sense, done.

lib/usd/translators/mayaReferenceUpdater.cpp
ppt-adsk: Quite right...  An oversight on my part.  I don't see a way of performing this easily with OpenMaya; MDGModifier doesn't seem to have the required interface.  Suggestions welcome, otherwise I propose that we enter a ticket to do this work later.

lib/usd/translators/mayaReferenceReader.cpp
ppt-adsk: You bet.  I had to adjust the testCustomRig.py test to do this.

ppt-adsk (commented):
Thanks for the code review!

lib/mayaUsd/fileio/primUpdaterManager.cpp
pierrebai-adsk: Ah OK. Maybe "instead of the Maya type", because I had read "instead" as "instead of the USD type".

lib/usd/translators/mayaReferenceUpdater.cpp
pierrebai-adsk: I think opening a ticket might be good at this point as we are nearing the 0.17 code complete date.

pierrebai-adsk (approved):

pierrebai-adsk (approved):

PR #2177, opened 2022-03-10
MAYA-121344 - Materials support in VP2 render delegate points

vlasovi:
- Insert points geometry shader fragment into material shaders for VP2 point rendering
 - HdVP2 points now support arbitrary primvars (like texture coordinates etc.)
 - Points default color is now configurable (the same way it is done for curves)
 - Adding a small fix into MaterialX shader generation to allow insertion of custom geometry shaders (needed for points)
 - Merging dev branch in and resolve conflicts, particularly in CommitState structure


vlasovi:
This is a draft pull request (only for review) from the branch that has all the commits for points implementation as well as the initial refactoring. @williamkrick the commits you haven't looked at start with [65bd4f0] "Support for points shader fragment in materials" @JGamache-autodesk you may be interested in the following change: [3872b0e]"Make sure points geometry fragment can be inserted into MaterialX fragment grapth"

PR #2184, opened 2022-03-14
Update the MaterialNetwork conversion function

dj-mcg:
Changing the conversion function to return an HdMaterialNetwork2 for the
given HdMaterialNetworkMap. Doing so to hopefully avoid potential issues
if the caller passes in an existing HdMaterialNetwork2, in which it
would give an inaccurate final material network.


dj-mcg:
Just wondering if this waiting for anything on my end before it gets merged?

ppt-adsk:
Hi @dj-mcg, we've blocked merging because we're trying to pick a build for the next maya-usd release.  Should be getting back to normal soon.

PR #2195, opened 2022-03-16
MAYA-108102 Add unit test for select -addFirst

pierrebai-adsk:


pierrebai-adsk:
The build actually succeeded as can be seen here: https://github.com/ecp-maya-devops-adsk/log-transfer/blob/1708b509bfc39e4b90ba798c2d3d23dac4864c91/1998630364_2181_result.txt

The problem is the final transfer of the build logs.

PR #2200, opened 2022-03-18
MAYA-122134 bad variant cache with empty pref

pierrebai-adsk:
Fix that caching to USD fails when the Maya prefs are reset.
Also add word-wrap to the descriptions UI.


lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
```diff
@@ -116,7 +116,7 @@ proc fileOptionsTabPage(string $tabLayout)
     setParent $topForm;
     frameLayout -label `getMayaUsdLibString("kMayaRefDescription")`;
     columnLayout -adjustableColumn true;
-    text -align "left" -label `getMayaUsdLibString("kMayaRefAddToUSDDescription")`;
+    text -align "left" -wordWrap 1 -label `getMayaUsdLibString("kMayaRefAddToUSDDescription")`;
```
seando-adsk, line 119: Doesn't wordWrap flag take a boolean? Using a 1 makes it seems like its some kind of word wrap value choice.

seando-adsk (changes requested):

lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
ppt-adsk: https://help.autodesk.com/view/MAYAUL/2022/ENU/?guid=GUID-CCECE76B-A48A-45CF-A4C1-F121004E3E7A

pierrebai-adsk: No wonder it is confusing :) I guess it is all for historical reasons. Putting true here does cost much, so I did.

PR #2226, opened 2022-03-30
MAYA-114603 persist USD payload state

pierrebai-adsk:
Load rules conversions

- Add functions to convert the USD stage load rules into text and back.
- Add function to save/load load rules on a dynamic Maya attribute.
- Add unit tests for the text conversions.
- Only run load rules unit test on Windows.
- Other platforms have linker or run-time link problems with Cg and OpenMaya.

Load rules persistence

- Add a class to handle the persistence.
- It keeps a list of proxy shape that need to be updated before being saved
- Listen to kBeforeSave Maya message and transfers the load rules to the proxy stage.
- Each proxy shape registers with this class.
- The proxy shape applies the persisted load rules on load if present.
- Implemented the persistence using dynamic attributes.
- Since there won't be a ton of them, use MEL as it is more robust and simpler.

Others

- Fix a bug in UsdMayaSceneResetNotice.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeLoadRules.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeLoadRules.cpp
@@ -1,0 +1,95 @@
+//
+// Copyright 2016 Pixar
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "proxyShapeLoadRules.h"
+
+#include <mayaUsd/utils/loadRules.h>
+
+#include <maya/MSceneMessage.h>
+
+#include <set>
+
+namespace MAYAUSD_NS_DEF {
+
+namespace {
+
+using ProxyShapeSet = std::set<MayaUsdProxyShapeBase*>;
+
+MCallbackId beforeFileSaveCallbackId = 0;
+
+ProxyShapeSet& getTrackedProxyShapes()
+{
+    static ProxyShapeSet tracked;
+    return tracked;
+}
+
+void onMayaAboutToSave(void* /* unused */)
+{
+    ProxyShapeSet& tracked = getTrackedProxyShapes();
+    for (MayaUsdProxyShapeBase* proxyShape : tracked) {
+        if (!proxyShape)
+            continue;
+
+        auto stagePtr = proxyShape->getUsdStage();
+        if (!stagePtr)
+            continue;
+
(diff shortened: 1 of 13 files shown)


test/lib/mayaUsd/utils/CMakeLists.txt
```diff
@@ -39,3 +43,65 @@ foreach(script ${INTERACTIVE_TEST_SCRIPT_FILES})
...
+    mayaUsd_add_test(${TARGET_NAME}
+        COMMAND $<TARGET_FILE:${TARGET_NAME}>
+        ENV
+            "LD_LIBRARY_PATH=${ADDITIONAL_LD_LIBRARY_PATH}"
+            "MAYA_LOCATION=${MAYA_LOCATION}"
+    )
+endfunction()
+
+if(IS_WINDOWS)
+    # There are link problems on Linux and OSX with C++ test using USD + Maya,
```
pierrebai-adsk, line 99: I could not get the C++ unit test to link on Linux or run on OSX.

- On Linux 2020, the linker would fail to find OpenSubDiv
- On Linux 2022, execution ofthe test would fail to find libCg.so
- On OSX master, execution would fail to find libOpenMaya.dylib

But all other platforms / versions would work... so I only kept the test on Windows. It is enough to validate during pre-flight.

lib/mayaUsd/nodes/proxyShapeLoadRules.cpp
```diff
@@ -0,0 +1,95 @@
+//
+// Copyright 2016 Pixar
```
boudrey, line 2: Copyright should be Autodesk right?

lib/mayaUsd/utils/loadRulesAttribute.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+namespace {
+
+const char loadRulesAttrName[] = "usdStageLoadRules";
+
+bool hasLoadRulesAttribute(const MFnDependencyNode& depNode)
+{
+    MString nodeName = depNode.absoluteName();
+    MString cmd;
+    cmd.format("addAttr -query -exists \"^2s.^1s\"", loadRulesAttrName, nodeName);
```
boudrey, line 36: It is a shame to call MEL code in c++. How hard would it be to find the proper OpenMaya call to do the same?

pierrebai-adsk: I tried initially, but the code would just crash when accessing the dynamic attribute I added. Since I had prototyped it in MEl in Maya and it worked just fine, that MEL mostly never crashes even when things are not right and that I doubt people will create many stages in a single Maya file, I don't think it is worth trying to optimize it.

pierrebai-adsk: Plus, frankly, the docs in the C++ API related to dynamic attribute is... obscure or missing. Even searching the Maya code base yielded next to zero use case examples.

test/lib/mayaUsd/utils/CMakeLists.txt
boudrey: Please assign it to me after it is merged, I will give it a try.

boudrey (approved):

pierrebai-adsk:
The only failure is the known Linux interactive where image comparison unit tests now fail due to viewport changes.

PR #2227, opened 2022-03-30
Fix preflight test failures

williamkrick:
Set the material on the cube used for mtoh tests, update image results which use the default material.


lib/usd/hdMaya/adapters/materialNetworkConverter.cpp
```diff
@@ -577,7 +577,6 @@ void HdMayaMaterialNetworkConverter::initialize()
                 { HdMayaAdapterTokens->clearcoatRoughness, coatRoughnessConverter },
                 { HdMayaAdapterTokens->opacity, transmissionToOpacity },
                 { HdMayaAdapterTokens->metallic, fixedZeroFloat },
-                { HdMayaAdapterTokens->useSpecularWorkflow, fixedOneInt },
```
williamkrick, line 580: Should the other things that have useSpecularWorkflow also get removed?

JGamache-autodesk: Answer is probably yes. The reason they do not result in black is that we systematically do not set a specular color for these networks. If we were to try to set a specular for Blinn and Phong, they would turn black as well.

JGamache-autodesk: But if it is not broken, maybe we should not fix it?
The UsdPreviewSurface exporters for these nodes set it to 1, so if we decide to change that, we should keep the exporters in sync.

test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegatePerInstanceInheritedData.py
```diff
@@ -139,22 +142,28 @@ def testPerInstanceInheritedDataPartialOverride(self):
         self._StartTest('inheritedDisplayColor_pxrSurface')
 
     def testPerInstanceInheriedDataBasisCurves(self):
-        self._StartTest('basisCurveInstance')
+        imageVersion = None
+        if getMayaAPIVersion() <= 2023:
```
JGamache-autodesk, line 146: May I recommend you use the following?
```
import maya.mel
if maya.mel.eval("defaultShaderName") == "standardSurface":
    imageVersion == 'standardSurface'
```
This would be more robust.

JGamache-autodesk (changes requested):

JGamache-autodesk (approved):

test/lib/mayaUsd/render/mayaToHydra/testMtohBasicRender.py
```diff
@@ -47,17 +48,21 @@ def test_flat_orange(self):
...
     _file = __file__
 
     def test_cube(self):
+        imageVersion = None
+        if maya.mel.eval("defaultShaderName") != "standardSurface1":
+            imageVersion = 'lambertDefaultMaterial'
+
         self.makeCubeScene(camDist=6)
-        self.assertSnapshotClose("cube_unselected.png")
+        self.assertSnapshotClose("cube_unselected.png", imageVersion, 0.0002)
```
williamkrick, line 63: The preflight runs are generating a result slightly different than my local computer. The the images differ a little more so preflight works.

JGamache-autodesk (approved):

Issue #2139, opened 2022-02-23
[MAYA-122477] Textures disappear when manipulating a prim when Async Texture Loading is enabled.


santosd:
**Describe the bug**
If I have Async Texture Loading enabled and I create a stage from the ALab scene. After all of the prims and textures are loaded if I rotate the decor_helmet_head01_0001 prim the texture will disappear. It only happens on the first move, unless you undo the first move and move the object again.

**Steps to reproduce**
1. Make sure that Asyn Texture Loading is enabled:
`import maya.cmds as cmds
cmds.optionVar(iv=("mayaUsd_DisableAsyncTextureLoading", 0))`
2. Create a stage from the USD_Alab file.
3. Once prims and textures have loaded, select the decor_helmet_head01_0001 prim.
4. Rotate, scale, or translate it and notice that while the prim is being manipulated, the texture will disappear.

**Expected behavior**
After textures have loaded and transformation of the prims would show the texture.

**Attachments**

https://user-images.githubusercontent.com/29235905/155382653-27dd93d3-df7e-4db6-af6a-db6d16dbad0c.mp4

williamkrick:
https://github.com/Autodesk/maya-usd/pull/1980 is the change that introduces async material loading.

santosd:
Hi @williamkrick I retested this issue using the latest changes and this issue is still happening. Should we reassign this issue to try to get it fixed or set this in the backlog for the time being?

williamkrick:
Yes lets log something internally and see if we can get it scheduled.

santosd:
Logged it internally.

csyshing:
[...]

PR #2228, opened 2022-03-31
MAYA-122477 fix textures disappear when manipulating a prim when async texture loading is enabled

csyshing:
PR for https://github.com/Autodesk/maya-usd/issues/2139 .

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -1722,6 +1722,20 @@ HdVP2Material::~HdVP2Material()
 {
     // Tell pending tasks or running tasks (if any) to terminate
     ClearPendingTasks();
+
+    // USD `UsdImagingDelegate::ApplyPendingUpdates()` would request to
+    // remove the material then recreate, this is causing texture disappearing
+    // when user manipulating a prim (while holding mouse buttion).
+    // We hold a copy of the texture info reference, so that the texture will not
+    // get released immediately along with material removal.
+    // If the textures would have been requested to reload in `ApplyPendingUpdates()`,
+    // we could still reuse the loaded one from cache, otherwise the idle task can
+    // safely release the texture.
+    if (!_IsDisabledAsyncTextureLoading()) {
+        MGlobal::executeTaskOnIdle(
+            [](void* data) { delete static_cast<HdVP2LocalTextureMap*>(data); },
+            new HdVP2LocalTextureMap(_localTextureMap.begin(), _localTextureMap.end()));
+    }
 }
 
 /*! \brief  Synchronize VP2 state with scene delegate state based on dirty bits


lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1722,6 +1722,20 @@ HdVP2Material::~HdVP2Material()
...
+    // when user manipulating a prim (while holding mouse buttion).
+    // We hold a copy of the texture info reference, so that the texture will not
+    // get released immediately along with material removal.
+    // If the textures would have been requested to reload in `ApplyPendingUpdates()`,
+    // we could still reuse the loaded one from cache, otherwise the idle task can
+    // safely release the texture.
+    if (!_IsDisabledAsyncTextureLoading()) {
+        MGlobal::executeTaskOnIdle(
+            [](void* data) { delete static_cast<HdVP2LocalTextureMap*>(data); },
+            new HdVP2LocalTextureMap(_localTextureMap.begin(), _localTextureMap.end()));
```
williamkrick, line 1737: Does playing back animation have the same de-populate -> re-populate behavior as manipulation? In a case where a textured object is animated we could easily end up creating lots of lots of these idle tasks without ever running them, and then when playback finishes we'd delete them all. If when we copy the HdVP2LocalTextureMap we only copy the HdVP2TextureInfoSharedPtr when the reference count is 1 then we could avoid creating an excessive number of tasks.

williamkrick: Manipulation is not so dangerous because the user will stop dragging after not too long. But if someone leaves it running they could build up a huge queue of idle tasks.

williamkrick: One possible implementation would be to create an HdVP2LocalTextureMap on the stack, copy in the entries of _localTextureMap that have one reference, and then move-construct a new HdVP2LocalTextureMap as the parameter for the idle task.

Technically if this code was executing in parallel we could have a race on the shared_ptr use_count but the code appears to execute serially, so we should always find the material giving up the last reference to any given texture.

williamkrick (dismissed):

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
csyshing: > Does playing back animation have the same de-populate -> re-populate behavior as manipulation?

From my observation, USD only does this de-populate & re-populate if the scene delegate cache needs to be updated (e.g. the first time authoring a prim once loaded), in reality, playback **might** not trigger that for every single frame.

> But if someone leaves it running they could build up a huge queue of idle tasks.

I tweaked the code to create just one idle task for all pending removal textures, have a look at my new commit f2a6a3b ?

> Technically if this code was executing in parallel we could have a race on the shared_ptr use_count but the code appears to execute serially

Yes, USD imaging runs the removal serially, but my previous code relies on this implementation detail which is not ideal, I added a mutex to make sure copying shared pointer would not cause race condition.

Let me know if that looks better?

williamkrick (approved):

PR #2229, opened 2022-03-31
MAYA-121938 prevent recursive edit as Maya

pierrebai-adsk:
Make can-edit-as-Maya return false if the prim is already edited or an ancestor of an edited prim.

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -134,6 +134,36 @@ SdfPath makeDstPath(const SdfPath& dstRootParentPath, const SdfPath& srcPath)
 
 //------------------------------------------------------------------------------
 //
+// Verify if the given prim under the given UFE path is already edited or
+// is an ancestor of an already edited prim.
+bool isAlreadyEditedAsMaya(const Ufe::Path& ufePulledPath)
+{
+    MObject pullSetObj;
+    auto    status = UsdMayaUtil::GetMObjectByName(kPullSetName, pullSetObj);
+    if (status != MStatus::kSuccess)
+        return false;
+
+    MFnSet         fnPullSet(pullSetObj);
+    MSelectionList members;
+    const bool     flatten = true;
+    fnPullSet.getMembers(members, flatten);
+
+    for (unsigned int i = 0; i < members.length(); ++i) {
+        MDagPath itemPath;
+        members.getDagPath(i, itemPath);
+        Ufe::Path itemUfePath;
+        if (!PrimUpdaterManager::readPullInformation(itemPath, itemUfePath))
+            continue;
+
+        if (ufePulledPath.startsWith(itemUfePath) || itemUfePath.startsWith(ufePulledPath))
+            return true;
+    }
+
+    return false;
+}
+
+//------------------------------------------------------------------------------
+//
 // The UFE path is to the pulled prim, and the Dag path is the corresponding
 // Maya pulled object.
 bool writePullInformation(const Ufe::Path& ufePulledPath, const MDagPath& path)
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -132,6 +132,36 @@ SdfPath makeDstPath(const SdfPath& dstRootParentPath, const SdfPath& srcPath)
...
+    if (status != MStatus::kSuccess)
+        return false;
+
+    MFnSet         fnPullSet(pullSetObj);
+    MSelectionList members;
+    const bool     flatten = true;
+    fnPullSet.getMembers(members, flatten);
+
+    for (unsigned int i = 0; i < members.length(); ++i) {
+        MDagPath itemPath;
```
ppt-adsk, line 152: I know this is a frequent problem for us, because of the various paths we have to deal with (Maya, Sdf, UFE), but the variable naming here had me confused.  I would rename the variables like this:
ufePulledPath --> ufeQueryPath (it hasn't been pulled yet)
itemPath --> pulledDagPath
itemUfePath --> pulledUfePath

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -132,6 +132,36 @@ SdfPath makeDstPath(const SdfPath& dstRootParentPath, const SdfPath& srcPath)
...
+        members.getDagPath(i, itemPath);
+        Ufe::Path itemUfePath;
+        if (!PrimUpdaterManager::readPullInformation(itemPath, itemUfePath))
+            continue;
+
+        if (ufePulledPath.startsWith(itemUfePath) || itemUfePath.startsWith(ufePulledPath))
```
ppt-adsk, line 158: I'm confused about this.  How can a user edit as Maya a descendant of something that has already been pulled?  By definition, all descendants of a pulled prim are also pulled.  Therefore:

if (ufeQueryPath.startsWith(pulledUfePath) --> this means an already pulled UFE path is an ancestor of the query path.  Therefore, the query path is already edited as Maya.  This should never happen, and should be a TF_VERIFY, no?

|| pulledUfePath.startsWith(ufeQueryPath) --> this means the query path is an ancestor of a pulled path.  This does NOT mean the query path is already edited as Maya, which is what the function name says.  It just means the query argument is an ancestor of an already pulled path, and we want to prevent this query path from being edited.

I think the function should be renamed to hasEditedDescendant(), and we should assert that the query path is not a descendant of an already-pulled prim, which would in my opinion be demonstrating a programming error.

ppt-adsk (changes requested):
Thanks for the fix.  I think we should be testing this somehow.

lib/mayaUsd/fileio/primUpdaterManager.cpp
pierrebai-adsk: I wrote the function to be fool-proof without dependencies on what is possible now vs what could be possible in the future. I almost called it something like "wouldCauseDoubleEditions()" but I could not find a non-awkward name.

I can limit it to what is currently possible and follow your suggestions.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -132,6 +132,36 @@ SdfPath makeDstPath(const SdfPath& dstRootParentPath, const SdfPath& srcPath)
...
+    MFnSet         fnPullSet(pullSetObj);
+    MSelectionList members;
+    const bool     flatten = true;
+    fnPullSet.getMembers(members, flatten);
+
+    for (unsigned int i = 0; i < members.length(); ++i) {
+        MDagPath itemPath;
+        members.getDagPath(i, itemPath);
+        Ufe::Path itemUfePath;
+        if (!PrimUpdaterManager::readPullInformation(itemPath, itemUfePath))
```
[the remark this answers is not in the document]
pierrebai-adsk: Nothing prevent the user from adding stuff to the set. Just avoiding possible errors.

ppt-adsk (approved):

PR #2230, opened 2022-03-31
MAYA-122502 When instance colors is nullptr don't get the length.

williamkrick:


williamkrick:
> Actually, I remember a couple of weeks ago I got in touch with you @williamkrick to let you know that this place can crash. Good to see that it gets fixed :-)

Yes I remember and I thought that it wasn't possible to get in there, but clearly it is!

PR #2235, opened 2022-03-31
MAYA-122131 Maya Ref Auto-Edit Handling

pierrebai-adsk:
Turn off auto-edit when ref edit are discarded.

### DIFF
--- a/lib/mayaUsd/CMakeLists.txt
+++ b/lib/mayaUsd/CMakeLists.txt
@@ -200,7 +200,6 @@ target_link_libraries(${PROJECT_NAME}
         $<$<BOOL:${UFE_FOUND}>:${UFE_LIBRARY}>
         ${MAYA_LIBRARIES}
         mayaUsdUtils
-        mayaUsd_Schemas
     PRIVATE
         ghc_filesystem
         $<$<BOOL:${CMAKE_WANT_MATERIALX_BUILD}>:hdMtlx>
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/CMakeLists.txt
```diff
@@ -200,6 +200,7 @@ target_link_libraries(${PROJECT_NAME}
         $<$<BOOL:${UFE_FOUND}>:${UFE_LIBRARY}>
         ${MAYA_LIBRARIES}
         mayaUsdUtils
+        mayaUsd_Schemas
```
ppt-adsk, line 203: Don't think we should be adding this dependence.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1022,6 +1023,15 @@ bool PrimUpdaterManager::discardPrimEdits(const Ufe::Path& pulledPath)
         updater->discardEdits();
     }
 
+    // Reset the auto-edit when discarding the edit.
```
ppt-adsk, line 1026: Essentially, we are checking for a particular node type here --- and I don't think that's a good idea.  I would really hope we can implement an override of the discardEdits() virtual on MayaReferenceUpdater and do the work there.

ppt-adsk (changes requested):
Any way we can test this automatically?

lib/mayaUsd/CMakeLists.txt
pierrebai-adsk: It was necessary to access the schema attribute.

lib/usd/translators/mayaReferenceUpdater.cpp
```diff
@@ -42,6 +44,7 @@
 #include <pxr/usd/usdUtils/pipeline.h>
 
 #include <maya/MFnAttribute.h>
+#include <ufe/pathString.h>
```
ppt-adsk, line 47: Don't think we need this one.

ppt-adsk (approved):

PR #2238, opened 2022-04-01
MAYA-122124 avoid errors in marking menu

pierrebai-adsk:
The function used to try to convert a UFE path to a USD prim prints assertion when the path is not to a USD prim. Caller need to check if the prim is valid anyway, so don't cause spurious, unconditional errors to get printed.


boudrey (approved):

pierrebai-adsk:
Only failure is the known image comparison failure in Linux.

PR #2242, opened 2022-04-04
Fix for MAYA-122381 Target Layer moves and stays on Session Layer whe…

boudrey:
…n Editing As Maya Data

Stage is not passed as a parameter instead of trying to find the right one.
SessionLayer is only set temporarely asa the edit target instead of premanently.


boudrey:
This PR is bringing back the "[MAYA-115146](https://jira.autodesk.com/browse/MAYA-115146)
As a user, I'd like my Maya Reference node to be named the regular Maya way" that was reverted just before closing 0.17.0.

boudrey:
Only PF failure is the known Linux interactive.

ppt-adsk (commented):
LGTM, but is there a way we can provide an automated test to demonstrate that this works?

PR #2245, opened 2022-04-04
MAYA-122369 fix missing edit after merge

pierrebai-adsk:
When targeting a higher layer than where a prim is defined, the merge would fail to create the proper USD hierarchy. That was due to SdfCopySpec not creating automatically the missing parents.

### DIFF
--- a/lib/usd/utils/MergePrims.cpp
+++ b/lib/usd/utils/MergePrims.cpp
@@ -19,6 +19,7 @@
 
 #include <pxr/base/tf/stringUtils.h>
 #include <pxr/usd/sdf/copyUtils.h>
+#include <pxr/usd/usd/editContext.h>
 #include <pxr/usd/usdGeom/xformCommonAPI.h>
 
 #include <algorithm>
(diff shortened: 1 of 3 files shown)


pierrebai-adsk:
The preflight failure is due to the known image comparison failures on Linux due to Maya default lights intensity changes.

lib/usd/utils/MergePrims.cpp
```diff
@@ -789,6 +790,63 @@ bool mergeDiffPrims(
...
 }
 
+//----------------------------------------------------------------------------------------------------------------------
+/// Create any missing parents as "over". Parents may be missing because we are targeting a
+/// different layer than where the destination prim is authored. The SdfCopySpec function does not
+/// automatically create the missing parent, unlike other functions like UsdStage::CreatePrim.
+///
+/// Unfortunately, USD won't actually author "over" (or even "def") until we set at least one
+/// attribute on it. So we set the active attribute and once the merge is done we clear those
+/// attributes to avoid leaving behind opinions that the user ight not expect to exist. Otherwise
```
boudrey, line 800: ight->might

boudrey (dismissed):

lib/usd/utils/MergePrims.cpp
```diff
@@ -789,6 +790,63 @@ bool mergeDiffPrims(
     return SdfCopySpec(srcLayer, srcPath, dstLayer, dstPath, copyValue, copyChildren);
 }
 
+//----------------------------------------------------------------------------------------------------------------------
+/// Create any missing parents as "over". Parents may be missing because we are targeting a
+/// different layer than where the destination prim is authored. The SdfCopySpec function does not
+/// automatically create the missing parent, unlike other functions like UsdStage::CreatePrim.
+///
+/// Unfortunately, USD won't actually author "over" (or even "def") until we set at least one
```
ppt-adsk, line 798: That's not true.  See `SdfJustCreatePrimInLayer()`
https://graphics.pixar.com/usd/release/api/prim_spec_8h.html#af1812eef21037f6226a8b66765831542
This is what I think you should be using.

ppt-adsk (changes requested):

lib/usd/utils/MergePrims.cpp
pierrebai-adsk: That function would create a valid prim, and only its parents would be "over". We just need to create the parent over, The function is *almost* what we need, but not quite :)

ppt-adsk: `SdfJustCreatePrimInLayer` creates the prim as an over, with empty type.  Just use the parent of the path and you should be fine.

pierrebai-adsk: I'll try it. The description of the function is unclear if the prim is created as an over or only its parents.

pierrebai-adsk: Juts looked at the implementation code in USD and indeed, it creates even the final prim as an over

ppt-adsk (approved):

PR #2250, opened 2022-04-06
[al] Add support for default rgba values and color threshold for exporting

csyshing:
This PR mainly contains two features for exporting:
 - Default RGBA value (RGB = 0.18, A =1.0) so that the value is always meaningful
 - Optionally accept color "threshold", the UI is also extended to support custom precision to allow displaying high precision of floating point numbers, defaults to 1e-5 which should be enough for color.

### DIFF
--- a/plugin/al/mayautils/AL/maya/utils/PluginTranslatorOptions.cpp
+++ b/plugin/al/mayautils/AL/maya/utils/PluginTranslatorOptions.cpp
@@ -143,9 +143,25 @@ bool PluginTranslatorOptions::addInt(const char* optionName, int defaultValue)
 AL_MAYA_UTILS_PUBLIC
 bool PluginTranslatorOptions::addFloat(const char* optionName, float defaultValue)
 {
+    return addFloat(optionName, defaultValue, 1, "", true);
+}
+
+//----------------------------------------------------------------------------------------------------------------------
+bool PluginTranslatorOptions::addFloat(const char* optionName, float value, int precision)
+{
+    return addFloat(optionName, value, precision, "", true);
+}
+
+bool PluginTranslatorOptions::addFloat(
+    const char* optionName,
+    float       value,
+    int         precision,
+    const char* controller,
+    bool        state)
+{
     if (isOption(optionName))
         return false;
-    m_options.emplace_back(optionName, defaultValue);
+    m_options.emplace_back(optionName, value, precision, controller, state);
     return true;
 }
 
@@ -544,21 +560,54 @@ void PluginTranslatorOptions::generateIntGlobals(
 
 //----------------------------------------------------------------------------------------------------------------------
 void PluginTranslatorOptions::generateFloatGlobals(
-    const char* const prefix,
-    const MString&    niceName,
-    const MString&    optionName,
-    MString&          code,
-    float             value)
+    const char* const  prefix,
+    const MString&     niceName,
+    const MString&     optionName,
+    MString&           code,
+    float              value,
+    int                precision,
+    const std::string& controller,
+    bool               enableState)
 {
     MString valueStr;
     valueStr = value;
     MString controlName = MString(prefix) + "_" + makeName(optionName);
+    // Check and add on/off command if any
+    MString onOffCmd;
+    MString postUpdateCmd;
+    MString deferredPostUpdateCmd;
+    if (!controller.empty()) {
+        MString prefixedCtrl(MString(prefix) + "_" + makeName(controller.c_str()));
+        if (enableState) {
+            onOffCmd = "checkBox -e "
+                       "-onCommand \"floatFieldGrp -e -en 1 "
+                + controlName
+                + "\" "
+                  "-offCommand \"floatFieldGrp -e -en 0 "
+                + controlName + "\" " + prefixedCtrl + "; ";
+            // And update the default state according to the checkbox
+            postUpdateCmd = "floatFieldGrp -e -en `checkBox -q -v " + prefixedCtrl + "` "
+                + controlName + "; ";
+        } else {
+            onOffCmd = "checkBox -e "
+                       "-onCommand \"floatFieldGrp -e -en 0 "
+                + controlName
+                + "\" "
+                  "-offCommand \"floatFieldGrp -e -en 1 "
+                + controlName + "\" " + prefixedCtrl + "; ";
+            postUpdateCmd = "floatFieldGrp -e -en (`checkBox -q -v " + prefixedCtrl + "` ? 0: 1) "
+                + controlName + "; ";
+        }
+        onOffCmd += postUpdateCmd;
+        deferredPostUpdateCmd = MString() + "eval(\"" + postUpdateCmd + "\");";
+    }
+
     MString createCommand = MString("global proc create_") + controlName + "() {"
-        + MString("floatFieldGrp -l \"") + niceName + "\" -v1 " + valueStr + " " + controlName
-        + ";}\n";
+        + MString("floatFieldGrp -l \"") + niceName + "\" -v1 " + valueStr + " -pre " + precision
+        + " " + controlName + ";" + onOffCmd + "}\n";
     MString postCommand = MString("global proc post_") + controlName
-        + "(string $value){ eval (\"floatFieldGrp -e -v1 \" + $value + \" " + controlName
-        + "\");}\n";
+        + "(string $value){ eval (\"floatFieldGrp -e -v1 \" + $value + \" " + controlName + "\");"
+        + deferredPostUpdateCmd + "}\n";
     MString buildCommand = MString("global proc string build_") + controlName
         + "(){ string $str = \"" + makeName(optionName) + "=\" + `floatFieldGrp -q -v1 "
         + controlName + "` + \";\"; return $str;}\n";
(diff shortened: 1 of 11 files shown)


plugin/al/mayautils/AL/maya/utils/PluginTranslatorOptions.cpp
```diff
@@ -544,21 +560,54 @@ void PluginTranslatorOptions::generateIntGlobals(
...
-    const MString&    optionName,
-    MString&          code,
-    float             value)
+    const char* const  prefix,
+    const MString&     niceName,
+    const MString&     optionName,
+    MString&           code,
+    float              value,
+    int                precision,
+    const std::string& controller,
```
pierrebai-adsk, line 569: I would be nicer to pass as  MString, other parameters are passed using that type (except prefix for some reason). It gets converted to a MString in the function anwyay. Other wise the parameters would now take all three of: const char*, MString and std::string.

plugin/al/mayautils/AL/maya/utils/PluginTranslatorOptions.h
```diff
@@ -411,9 +437,17 @@ class PluginTranslatorOptions
         /// \brief  ctor
         /// \param  name the name of the option
         /// \param  defVal the default value
-        Option(const char* const name, const float& defVal)
+        Option(
+            const char* const name,
+            const float&      defVal,
+            int               defPre,
+            const char*       defController,
+            bool              defState)
```
pierrebai-adsk, line 445: Same here, I'd rename defState to enableState.

plugin/al/mayautils/AL/maya/utils/PluginTranslatorOptions.h
```diff
@@ -387,6 +410,9 @@ class PluginTranslatorOptions
         MString              defString;   ///< default string value
         std::vector<MString> enumStrings; ///< the text values for the enums
         OptionType           type;        ///< the type of the option
+        int                  precision { 1 };
+        std::string          controller {};
```
pierrebai-adsk, line 414: Same here: I'd use a MString instead of std::string.

plugin/al/plugin/AL_USDMayaTestPlugin/test_translators_MeshTranslator.cpp
```diff
@@ -1066,6 +1066,220 @@ TEST(translators_MeshTranslator, faceVaryingColourExport)
...
+        ASSERT_EQ(1u, received.size());
+
+        EXPECT_NEAR(0.001f, received[0][0], 1e-5f);
+        EXPECT_NEAR(0.4f, received[0][1], 1e-5f);
+        EXPECT_NEAR(0.5f, received[0][2], 1e-5f);
+        EXPECT_NEAR(1.0f, received[0][3], 1e-5f);
+    }
+}
+
+/// \brief  Test exporting mesh with color threshold value and compaction 1 (basic level)
```
pierrebai-adsk, line 1212: Comment does not match: claims test uses compaction 1 (basic level), but the test uses 3.

plugin/al/translators/CommonTranslatorOptions.h
```diff
@@ -22,7 +22,13 @@ static constexpr const char* const kMeshUvOnly = "Mesh UV Only"; ///< export mes
 static constexpr const char* const kMeshPointsAsPref
     = "Mesh Points as PRef"; ///< export mesh Points as PRef, duplicating "P"
 static constexpr const char* const kMeshColours = "Mesh Colours"; ///< export mesh Colour Sets
-static constexpr const char* const kMeshHoles = "Mesh Holes";     ///< export mesh face holes
+static constexpr const char* const kMeshDefaultColourRGB = "Default RGB"; ///< default rgb values
+static constexpr const char* const kMeshDefaultColourA = "Default Alpha"; ///< default alpha values
+static constexpr const char* const kCustomColourThreshold
+    = "Custom Colour Threshold"; ///< default alpha values
```
pierrebai-adsk, line 28: Comment here and on the next does not corresponds.

plugin/al/usdmayautils/AL/usdmaya/utils/DiffPrimVar.cpp
```diff
@@ -1674,6 +1715,78 @@ TfToken guessColourSetInterpolationTypeExtensive(
...
+
+    if (numElements <= 1 || isWithinThreshold(rgba, numElements, threshold)) {
+        return UsdGeomTokens->constant;
+    }
+
+    float absThreshold = std::abs(threshold);
+
+    // check for per-vertex assignment
+    std::vector<uint32_t> indicesMap;
+    indicesMap.resize(numPoints, -1);
```
pierrebai-adsk, line 1743: You are using -1 here but 0xFFFFFFFF below. I would be clearfer if they were both the same. (Also, passing a negative value to initialize an unsigned integer without a cast might trigger a warning on some compilers and we generally have the warning-as-error compiler option turned on.)

plugin/al/usdmayautils/AL/usdmaya/utils/DiffPrimVar.cpp
```diff
@@ -1674,6 +1715,78 @@ TfToken guessColourSetInterpolationTypeExtensive(
...
+    for (uint32_t pntInx = 0, n = pointIndices.length(); pntInx < n; ++pntInx) {
+        auto index = pointIndices[pntInx];
+        auto lastIndex = indicesMap[index];
```
pierrebai-adsk, line 1746: For robustness reason, I would suggest verifying that the index is within the indicesMap size, in case of data corruption.

plugin/al/usdmayautils/AL/usdmaya/utils/MeshUtils.cpp
```diff
@@ -1475,13 +1475,36 @@ void interleaveIndexedUvData(
 #endif
 }
 
+//----------------------------------------------------------------------------------------------------------------------
+// Loops through each Colour Set in the mesh writing out a set of non-indexed Colour Values in RGBA
```
pierrebai-adsk, line 1479: This comments does not match the function.

pierrebai-adsk (changes requested):
Only a few small improvements suggested.

plugin/al/usdmayautils/AL/usdmaya/utils/DiffPrimVar.cpp
csyshing: Done.

plugin/al/usdmayautils/AL/usdmaya/utils/DiffPrimVar.cpp
csyshing: I have updated the code to check against the visited index, it is less memory friendly but more robust, let me know if that looks better?

plugin/al/translators/CommonTranslatorOptions.h
csyshing: Fixed.

plugin/al/plugin/AL_USDMayaTestPlugin/test_translators_MeshTranslator.cpp
csyshing: Fixed.

plugin/al/mayautils/AL/maya/utils/PluginTranslatorOptions.h
csyshing: Updated.

plugin/al/mayautils/AL/maya/utils/PluginTranslatorOptions.h
csyshing: Updated.

plugin/al/mayautils/AL/maya/utils/PluginTranslatorOptions.cpp
csyshing: Done.

plugin/al/mayautils/AL/maya/utils/PluginTranslatorOptions.cpp
```diff
@@ -142,10 +142,26 @@ bool PluginTranslatorOptions::addInt(const char* optionName, int defaultValue)
...
+{
+    return addFloat(optionName, value, precision, "", true);
+}
+
+bool PluginTranslatorOptions::addFloat(
+    const char* optionName,
+    float       value,
+    int         precision,
+    const char* controller,
+    bool        state)
```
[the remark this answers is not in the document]
csyshing: Done.

plugin/al/usdmayautils/AL/usdmaya/utils/DiffPrimVar.cpp
pierrebai-adsk: This is in the AL plugin, if you feel the change would impact performance, you can revert it. The callers are from data you control, if you feel confident the index are always valid by construction, feel free to remove the check, but I'd advise to do profiling first to verify if it really is a bottleneck.

pierrebai-adsk (approved):

plugin/al/usdmayautils/AL/usdmaya/utils/DiffPrimVar.cpp
csyshing: I reverted the changes in `DiffPrimVar.cpp`.
I have to admit that it was a copy-n-paste from `guessColourSetInterpolationTypeExtensive()` and the code has been used in production for a long time, I would keep it untouched in that case.

pierrebai-adsk (approved):

PR #2251, opened 2022-04-06
[al] Mark outTime attribute dirty to force updating Maya bbox; fix incorrect time code for bbox

csyshing:
Changes for this PR:
 - Mark geometry dirty explicitly otherwise bbox won't move
 - Fix GetOutputTime() to return the correct time code

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
@@ -1605,6 +1605,7 @@ MStatus ProxyShape::compute(const MPlug& plug, MDataBlock& dataBlock)
     m_requestedRedraw = true;
     MTime currentTime;
     if (plug == outTime()) {
+        MHWRender::MRenderer::setGeometryDrawDirty(thisMObject());
         return computeOutputTime(plug, dataBlock, currentTime);
     } else if (plug == outStageData()) {
         MStatus status = computeOutputTime(MPlug(plug.node(), outTime()), dataBlock, currentTime);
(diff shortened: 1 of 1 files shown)


plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
```diff
@@ -1605,6 +1605,7 @@ MStatus ProxyShape::compute(const MPlug& plug, MDataBlock& dataBlock)
     m_requestedRedraw = true;
     MTime currentTime;
     if (plug == outTime()) {
+        MHWRender::MRenderer::setGeometryDrawDirty(thisMObject());
```
williamkrick, line 1608: I see that you don't call `MayaUsdProxyShapeBase` and that the implementation of the two `compute()` methods has drifted relatively far apart.

- I believe that `MayaUsdProxyShapeBase::compute` is already doing the right thing by testing if the computed plug is timeAttr, assuming that plug must be computed in order to compute outTimeAttr. Is my understanding correct?
- What workflow can be tested to verify that the Maya bbox is correctly updated? Can we write an automated test that will catch this?
- Is there anything in `MayaUsdProxyShapeBase::compute` that you should copy here? There is some code to support changing which purposes the proxy shape is displaying which could be relevant to you.

williamkrick (approved):

williamkrick:
@seando-adsk I have some questions but nothing that would prevent this from being merged.

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/ProxyShape.cpp
csyshing: @williamkrick , those are good questions, I am pretty sure there must be good reasons to be implemented in that way in the first place, however I don't think I am the right person to give you an answer :-(

PR #2252, opened 2022-04-06
[al] Add validation on custom transform node before accessing

csyshing:
This PR added another validation on the transform node via lookup uuid, otherwise accessing non-exist node would cause Maya to crash during serialization.


seando-adsk:
@pierrebai-adsk Can you please do a quick code review of this change from AL.

pierrebai-adsk (approved):

PR #2257, opened 2022-04-06
MAYA-121681 lieft restrictions when a prim has empty over or def.

pierrebai-adsk:
If a layer has an entry for a prim but that entry is entirely empty, do not block edition of that prim on a different layer. This happens when a prim is merely used to build the hierarchy leading to another prim.

### DIFF
--- a/lib/mayaUsd/ufe/private/Utils.cpp
+++ b/lib/mayaUsd/ufe/private/Utils.cpp
@@ -128,14 +128,20 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
         return;
     }
 
-    auto        primSpec = MayaUsdUtils::getPrimSpecAtEditTarget(prim);
-    auto        primStack = prim.GetPrimStack();
+    SdfPrimSpecHandle       primSpec = MayaUsdUtils::getPrimSpecAtEditTarget(prim);
+    SdfPrimSpecHandleVector primStack = prim.GetPrimStack();
     std::string layerDisplayName;
     std::string message { "It is defined on another layer" };
 
     // iterate over the prim stack, starting at the highest-priority layer.
-    for (const auto& spec : primStack) {
+    for (const SdfPrimSpecHandle& spec : primStack) {
         const auto& layerName = spec->GetLayer()->GetDisplayName();
+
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -128,15 +128,21 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
         return;
     }
 
-    auto        primSpec = MayaUsdUtils::getPrimSpecAtEditTarget(prim);
-    auto        primStack = prim.GetPrimStack();
+    SdfPrimSpecHandle       primSpec = MayaUsdUtils::getPrimSpecAtEditTarget(prim);
```
pierrebai-adsk, line 131: Changed to concrete types to make understanding the code easier when reading it.

lib/mayaUsd/ufe/private/Utils.cpp
```diff
@@ -128,15 +128,21 @@ void applyCommandRestriction(const UsdPrim& prim, const std::string& commandName
...
+    SdfPrimSpecHandleVector primStack = prim.GetPrimStack();
     std::string layerDisplayName;
     std::string message { "It is defined on another layer" };
 
     // iterate over the prim stack, starting at the highest-priority layer.
-    for (const auto& spec : primStack) {
+    for (const SdfPrimSpecHandle& spec : primStack) {
         const auto& layerName = spec->GetLayer()->GetDisplayName();
 
+        // Don't block edits if there are no authored properties (attributes, relationships, etc)
```
pierrebai-adsk, line 140: This is the fix: empty def and empty over should not block edit.

seando-adsk (dismissed):

lib/mayaUsd/ufe/private/Utils.cpp
ppt-adsk: Sounds like a reasonable fix, great job @pierrebai-adsk!

ppt-adsk (approved):

seando-adsk (approved):

seando-adsk:
@pierrebai-adsk What is the status of this PR? It has the label "do not merge".

PR #2259, opened 2022-04-06
MAYA-108102 test depends on Maya version after 2023.

pierrebai-adsk:


test/lib/ufe/testSelection.py
```diff
@@ -275,7 +275,7 @@ def testMayaSelectFlags(self):
         self.assertTrue(globalSn.contains(first.path()))
         self.assertEqual(globalSn.back(), first)
 
-    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() >= 2 and mayaUtils.mayaMajorMinorVersions() >= (2023, 1), 'testMayaSelectAddFirst only available in UFE v2 or greater and Maya 2023.1 or greater.')
+    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() >= 2 and mayaUtils.mayaMajorVersion() > 2023, 'testMayaSelectAddFirst only available in UFE v2 or greater and Maya greter than 2023.')
```
seando-adsk, line 278: Minor typo "greater". And I don't think we need the ufe check. Just the Maya check should be enough since we know the ufe version will be greater than 2.

seando-adsk (approved):

PR #2260, opened 2022-04-06
MAYA-114603 use correct way to query attribute

pierrebai-adsk:
Use the correct mel command to querythe existence of an attribute. The new code avoids errors being printed in teh MEL console when the attribute does not exists.

### DIFF
--- a/lib/mayaUsd/utils/loadRulesAttribute.cpp
+++ b/lib/mayaUsd/utils/loadRulesAttribute.cpp
@@ -33,7 +33,7 @@ bool hasLoadRulesAttribute(const MFnDependencyNode& depNode)
 {
     MString nodeName = depNode.absoluteName();
     MString cmd;
-    cmd.format("addAttr -query -exists \"^2s.^1s\"", loadRulesAttrName, nodeName);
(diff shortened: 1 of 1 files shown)


pierrebai-adsk:
TY Sean for seeing the issue and suggesting the fix! As a punishment, you review it. :)

lib/mayaUsd/utils/loadRulesAttribute.cpp
```diff
@@ -33,7 +33,7 @@ bool hasLoadRulesAttribute(const MFnDependencyNode& depNode)
 {
     MString nodeName = depNode.absoluteName();
     MString cmd;
-    cmd.format("addAttr -query -exists \"^2s.^1s\"", loadRulesAttrName, nodeName);
+    cmd.format("attributeQuery -exists -n \"^2s\" \"^1s\"", loadRulesAttrName, nodeName);
```
seando-adsk, line 36: Looks good. FYI, you don't really lines 39/40 as the default value for both these args is false.

seando-adsk (approved):

PR #2261, opened 2022-04-06
Adds optionVars for controlling specular lighting options

JGamache-autodesk:
MaterialX has multiple options for specular environment sampling. They
will be exposed as optionVars.

string MxMayaEnvironmentMethod (none, prefiltered, fis)
  - Defaults to fis in V3, prefiltered otherwise

int MxMayaEnvironmentSamples (any reasonable value, default 64)
  - only valid in FIS mode

string MxMayaEnvironmentAlbedoMethod (analytic, montecarlo)
  - only valid in FIS mode

### DIFF
--- a/lib/mayaUsd/render/MaterialXGenOgsXml/OgsFragment.cpp
+++ b/lib/mayaUsd/render/MaterialXGenOgsXml/OgsFragment.cpp
@@ -12,10 +12,70 @@
 #include <MaterialXCross/Cross.h>
 #endif
 
+#include <maya/MGlobal.h>
+#include <maya/MString.h>
+
 #include <iostream>
 
 namespace MaterialXMaya {
 namespace {
+
+/// String option var controlling the environment method. Valid values are "none", "prefiltered" and
+/// "fis". Default values are based on available light API:
+///
+///  API |  Default      | Options
+///   V1 |  prefiltered  | none, prefiltered
+///   V2 |  prefiltered  | none, prefiltered
+///   V3 |  fis          | none, prefiltered, fis
+const MString OPTVAR_ENVIRONMENT_METHOD = "MxMayaEnvironmentMethod";
+
+/// Int option var to control the number of FIS samples. Larger values will slow down GLSL
+/// rendering speed and can lead to TDR. Default is 64, expected range is 1 - 1024 by powers of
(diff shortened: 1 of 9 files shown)


lib/mayaUsd/render/MaterialXGenOgsXml/OgsFragment.cpp
```diff
@@ -223,7 +294,8 @@ std::string generateFragment(
     // MaterialX fragment).
     std::ostringstream nameStream;
     const size_t       sourceHash = std::hash<std::string> {}(fragmentSource);
-    nameStream << baseFragmentName << "__" << std::hex << sourceHash;
+    nameStream << baseFragmentName << "__" << std::hex << sourceHash
+               << OgsFragment::getSpecularEnvKey();
```
JGamache-autodesk, line 298: Shader code depends on environment settings, so we must make sure we add the key everywhere we want to cache shaders.

lib/mayaUsd/render/MaterialXGenOgsXml/libraries/mx_lighting_maya_v3.glsl
```diff
@@ -28,6 +18,10 @@ float mx_latlong_compute_lod(vec3 dir, float pdf, float maxMipLevel, int envSamp
 
 vec3 mx_environment_radiance(vec3 N, vec3 V, vec3 X, vec2 roughness, int distribution, FresnelData fd)
 {
+    if (mayaGetSpecularEnvironmentNumLOD() == 0) {
```
JGamache-autodesk, line 21: Bail out early if there is no dome light.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -279,6 +279,8 @@ struct _MaterialXData
         mx::loadLibraries({}, _mtlxSearchPath, _mtlxLibrary);
 
         _FixLibraryTangentInputs(_mtlxLibrary);
+
+        mx::OgsXmlGenerator::setUseLightAPI(MAYA_LIGHTAPI_VERSION_2);
```
JGamache-autodesk, line 283: Moved from below. Only needs to be called once.

williamkrick (approved):

PR #2096, opened 2022-02-11
Remove/replace all UFE_PREVIEW_VERSION_NUM now that latest MayaPR has UFE v3

seando-adsk:

PR #2265, opened 2022-04-07
MAYA-122218 showing dactivated children

pierrebai-adsk:
Some code paths were ignoring deactivated children when they should not because hasChildren() was ignoring deactivated children.

### DIFF
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
@@ -133,7 +133,7 @@ const UsdPrim& ProxyShapeHierarchy::getUsdRootPrim() const
 
 Ufe::SceneItem::Ptr ProxyShapeHierarchy::sceneItem() const { return fItem; }
 
-#ifdef UFE_V4_FEATURES_AVAILABLE
+#if defined(UFE_V4_FEATURES_AVAILABLE) && (UFE_MINOR_VERSION > 4 || UFE_PATCH_LEVEL > 3)
 
 bool ProxyShapeHierarchy::hasChildren() const
 {
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
@@ -133,7 +133,7 @@ const UsdPrim& ProxyShapeHierarchy::getUsdRootPrim() const
 
 Ufe::SceneItem::Ptr ProxyShapeHierarchy::sceneItem() const { return fItem; }
 
-#if defined(UFE_V4_FEATURES_AVAILABLE) && (UFE_MINOR_VERSION > 4 || UFE_PATCH_LEVEL > 3)
+#if (UFE_PREVIEW_VERSION_NUM >= 4004)
 
 bool ProxyShapeHierarchy::hasChildren() const
 {
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
@@ -133,6 +133,30 @@ const UsdPrim& ProxyShapeHierarchy::getUsdRootPrim() const
 
 Ufe::SceneItem::Ptr ProxyShapeHierarchy::sceneItem() const { return fItem; }
 
+#ifdef UFE_V4_FEATURES_AVAILABLE
+
+bool ProxyShapeHierarchy::hasChildren() const
+{
+    // We have an extra logic in createUFEChildList to remap and filter
+    // prims. Going this direction is more costly, but easier to maintain.
+    //
+    // I don't have data that proves we need to worry about performance in here,
+    // so going after maintainability.
+    return !children().empty();
+}
+
+bool ProxyShapeHierarchy::hasFilteredChildren(const ChildFilter& childFilter) const
+{
+    // We have an extra logic in createUFEChildList to remap and filter
+    // prims. Going this direction is more costly, but easier to maintain.
+    //
+    // I don't have data that proves we need to worry about performance in here,
+    // so going after maintainability.
+    return !filteredChildren(childFilter).empty();
+}
+
+#else
+
 bool ProxyShapeHierarchy::hasChildren() const
 {
     // Return children of the USD root.
@@ -147,6 +171,8 @@ bool ProxyShapeHierarchy::hasChildren() const
     // so going after maintainability.
     return !createUFEChildList(getUSDFilteredChildren(rootPrim), false /*filterInactive*/).empty();
 }
+
+#endif
 
 Ufe::SceneItemList ProxyShapeHierarchy::children() const
 {
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
@@ -135,12 +135,17 @@ Ufe::SceneItem::Ptr ProxyShapeHierarchy::sceneItem() const { return fItem; }
 
 bool ProxyShapeHierarchy::hasChildren() const
 {
+    // Return children of the USD root.
+    const UsdPrim& rootPrim = getUsdRootPrim();
+    if (!rootPrim.IsValid())
+        return false;
+
     // We have an extra logic in createUFEChildList to remap and filter
     // prims. Going this direction is more costly, but easier to maintain.
     //
     // I don't have data that proves we need to worry about performance in here,
     // so going after maintainability.
-    return !children().empty();
+    return !createUFEChildList(getUSDFilteredChildren(rootPrim), false /*filterInactive*/).empty();
 }
 
 Ufe::SceneItemList ProxyShapeHierarchy::children() const
--- a/lib/mayaUsd/ufe/PulledObjectHierarchy.cpp
+++ b/lib/mayaUsd/ufe/PulledObjectHierarchy.cpp
@@ -52,7 +52,7 @@ PulledObjectHierarchy::Ptr PulledObjectHierarchy::create(
 
 Ufe::SceneItem::Ptr PulledObjectHierarchy::sceneItem() const { return _mayaHierarchy->sceneItem(); }
 
-#if defined(UFE_V4_FEATURES_AVAILABLE) && (UFE_MINOR_VERSION > 4 || UFE_PATCH_LEVEL > 3)
+#if (UFE_PREVIEW_VERSION_NUM >= 4004)
 bool PulledObjectHierarchy::hasFilteredChildren(const ChildFilter& childFilter) const
 {
     return _mayaHierarchy->hasFilteredChildren(childFilter);
--- a/lib/mayaUsd/ufe/PulledObjectHierarchy.cpp
+++ b/lib/mayaUsd/ufe/PulledObjectHierarchy.cpp
@@ -52,7 +52,7 @@ PulledObjectHierarchy::Ptr PulledObjectHierarchy::create(
 
 Ufe::SceneItem::Ptr PulledObjectHierarchy::sceneItem() const { return _mayaHierarchy->sceneItem(); }
 
-#ifdef UFE_V4_FEATURES_AVAILABLE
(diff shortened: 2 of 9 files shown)


seando-adsk:
I'm not sure it is that simple. Did you look at the history of this file (and the pulledObjectHierarchy)? We (Krystian and I) have modified this code a few times to fix different bugs. Did you test that those scenarios (from prior bugs) are still fixed with these changes?

pierrebai-adsk:
I looked at the most recent changes through git file history, like MAYA 113820.

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -124,12 +124,14 @@ bool UsdHierarchy::hasChildren() const
     //
     // I don't have data that proves we need to worry about performance in here,
     // so going after maintainability.
-    return !children().empty();
+    const bool isFilteringInactive = false;
+    return !createUFEChildList(getUSDFilteredChildren(fItem), isFilteringInactive).empty();
```
seando-adsk, line 128: If we are going with this change then we don't need the comment above anymore. Also have you looked at the changes Krystian made in the past? He was the the one who changed it to use children().

And have you looked at PulledObjectHierarchy and ProxyShapeHierarhcy - they both have similar code.

pierrebai-adsk: I'll take a look at these classes, thanks.

pierrebai-adsk: I re-read the comment and it needs to stay: what it cays is that calling create-children-list has extra costs since it will go through all children, but it is simpler to avoid duplicating the logic that filters the items out.

pierrebai-adsk:
Only failure is the PF on OSX python 2 Maya 2020 failed to launch, but all others passed.

pierrebai-adsk:
Th eonyl PF failure is the OSX python build that constantly fails to start.

test/lib/ufe/testContextOps.py
```diff
@@ -362,8 +362,13 @@ def testAddNewPrimWithDelete(self):
         cmds.pickWalk(d='down')
         cmds.delete()
 
-        # The proxy shape should now have no UFE child items (since we skip inactive).
-        self.assertFalse(proxyShapehier.hasChildren())
+        # The proxy shape should now have no UFE child items (since we skip inactive)
+        # but hasChildren still reports true for inactive to allow the caller to then
+        # do conditional inactive filtering.
+        if mayaUtils.mayaMajorVersion() >= 2023:
```
seando-adsk, line 368: I'm confused why this check is against a Maya version. What change is there in Maya that is affecting this?

seando-adsk (dismissed):

lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
```diff
@@ -135,12 +135,17 @@ Ufe::SceneItem::Ptr ProxyShapeHierarchy::sceneItem() const { return fItem; }
...
+    if (!rootPrim.IsValid())
+        return false;
+
     // We have an extra logic in createUFEChildList to remap and filter
     // prims. Going this direction is more costly, but easier to maintain.
     //
     // I don't have data that proves we need to worry about performance in here,
     // so going after maintainability.
-    return !children().empty();
+    return !createUFEChildList(getUSDFilteredChildren(rootPrim), false /*filterInactive*/).empty();
```
ppt-adsk, line 148: I'm confused by this change.  Now hasChildren() is no longer a convenience for !children().empty(), and is creating a different filtered children list than what children() is doing.  Why is this correct?

test/lib/ufe/testContextOps.py
ppt-adsk: I'm confused as to why we changed the test at all.

pierrebai-adsk: The behaviour of the delete command vs prim changed from deactibvating to really deleting. When deactivated, hasChildren() returns true.

lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
pierrebai-adsk: It is now a convenience for filteredChildren(0 with inactive prim. Unlike children(), which has an alternative function taking a filter that allows specifying to filter deactivated children or not, hasChildren() has a single version. So hasChildren has to choose to report deactivated children or not. The old choice was causing problems because code relied on it even when showing deactivated prims.

If we don't fix it here, then all Maya versions will still have bad outliner behaviour. Changing it fixes the behaviour in all Maya versions that use UFE v2 when they use the latest MayaUSD plugin.

lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
```diff
@@ -133,6 +133,8 @@ const UsdPrim& ProxyShapeHierarchy::getUsdRootPrim() const
 
 Ufe::SceneItem::Ptr ProxyShapeHierarchy::sceneItem() const { return fItem; }
 
+#if defined(UFE_V4_FEATURES_AVAILABLE) && (UFE_MINOR_VERSION > 4 || UFE_PATCH_LEVEL > 3)
```
seando-adsk, line 136: This should just use a single test like this:
```suggestion
#if (UFE_PREVIEW_VERSION_NUM >= 4004)
```

lib/mayaUsd/ufe/PulledObjectHierarchy.h
```diff
@@ -58,7 +58,10 @@ class MAYAUSD_CORE_PUBLIC PulledObjectHierarchy : public Ufe::Hierarchy
     Ufe::SceneItem::Ptr sceneItem() const override;
     bool                hasChildren() const override;
     Ufe::SceneItemList  children() const override;
-    Ufe::SceneItemList  filteredChildren(const ChildFilter&) const override;
+#if defined(UFE_V4_FEATURES_AVAILABLE) && (UFE_MINOR_VERSION > 4 || UFE_PATCH_LEVEL > 3)
+    bool hasFilteredChildren(const ChildFilter&) const override;
+#endif
+    UFE_V2(Ufe::SceneItemList filteredChildren(const ChildFilter&) const override;)
```
seando-adsk, line 64: Why did you add the UFE_V2 here? It is not needed as this file is only included in Ufe v3

test/lib/ufe/testContextOps.py
seando-adsk: Okay then instead of using then the Maya test should also surround lines 363 (which does the delete) and the comment you wrote here should be added to the test. Then for Maya 2023 we should not use delete, but instead deactivate the prim.

test/lib/ufe/testChildFilter.py
```diff
@@ -92,7 +92,7 @@ def testFilteredChildren(self):
         ball3Prim = mayaUsd.ufe.ufePathToPrim(ball3PathStr)
         ball3Prim.SetActive(False)
 
-        # Props should now have 5 children and ball3 should not be one of them.
+        # Props children should have 5 children and ball3 should not be one of them.
```
seando-adsk, line 95: Why did you remove "now". The comment was correct. At line 84 we test that there are 6 children and now we are testing that there are 5.

seando-adsk (changes requested):

test/lib/ufe/testChildFilter.py
pierrebai-adsk: I took multiple passes to update the test and did not realized I dropped a word there.

lib/mayaUsd/ufe/PulledObjectHierarchy.cpp
```diff
@@ -52,14 +52,23 @@ PulledObjectHierarchy::Ptr PulledObjectHierarchy::create(
...
+{
+    return _mayaHierarchy->hasFilteredChildren(childFilter);
+}
+#endif
+
 bool PulledObjectHierarchy::hasChildren() const { return _mayaHierarchy->hasChildren(); }
 
 Ufe::SceneItemList PulledObjectHierarchy::children() const { return _mayaHierarchy->children(); }
 
+#ifdef UFE_V2_FEATURES_AVAILABLE
```
ppt-adsk, line 66: The entire file is UFE v3 (in CMakeLists.txt), so this ifdef is redundant.

test/lib/ufe/testContextOps.py
```diff
@@ -358,17 +358,20 @@ def testAddNewPrimWithDelete(self):
         self.assertEqual(len(proxyShapehier.children()), 1)
 
         # Using UFE, delete this new prim (which doesn't actually delete it but
-        # instead makes it inactive).
+        # instead makes it inactive). In Maya 2023 and vreater, delete really
```
ppt-adsk, line 361: "vreater" --> "greater"

test/lib/ufe/testContextOps.py
```diff
@@ -358,17 +358,20 @@ def testAddNewPrimWithDelete(self):
...
-        cmds.delete()
+        if mayaUtils.mayaMajorVersion() >= 2023:
+            ufeItem = ufe.GlobalSelection.get().front()
+            item = usdUtils.getPrimFromSceneItem(ufeItem)
+            item.SetActive(False)
+        else:
+            cmds.delete()
 
         # The proxy shape should now have no UFE child items (since we skip inactive)
         # but hasChildren still reports true for inactive to allow the caller to then
```
ppt-adsk, line 372: Comment and code don't match.

ppt-adsk (changes requested):

test/testUtils/ufeUtils.py
```diff
@@ -105,3 +105,26 @@ def ufeFeatureSetVersion():
     #   v0.2.20 (unreleased preview version 2).
     major = ufe.VersionInfo.getMajorVersion()
     return ufe.VersionInfo.getMinorVersion() if major == 0 else major
+
+def hasMinimumVersion(releaseVersions, unreleasedVersions):
```
ppt-adsk, line 109: You shouldn't need this.  See
https://github.com/Autodesk/maya-usd/pull/2096/files
on the approach we've chosen: we use the Maya preview release version and the UFE preview version to determine whether code should be compiled and tests run.  Then, when the next Maya preview release is released, we clean up ALL the preview release and UFE preview release conditions we're put in, and replace them with UFE feature set version checks.

This cleans up the code one preview release at a time so that we don't leave old unreleased version checks in the files.

ppt-adsk (changes requested):

test/testUtils/ufeUtils.py
pierrebai-adsk: In this case I till need to differentiate between 0.4.3 and 0.4.4 though, so I'll test that only intstead.

test/lib/ufe/testContextOps.py
```diff
@@ -358,12 +359,26 @@ def testAddNewPrimWithDelete(self):
...
+
+        # The proxy shape should now have no UFE child items (since we skip inactive
+        # when returning the children list) but hasChildren still reports true in
+        # UFE version before 0.4.4 for inactive to allow the caller to do conditional
+        # inactive filtering, so we test that hasChildren is true for those versions.
+        ufeVersion = ufeUtils.fullVersion()
+        if (0, 4, 4) <= ufeVersion < (1, 0, 0):
+            self.assertFalse(proxyShapehier.hasChildren())
+        else:
+            self.assertTrue(proxyShapehier.hasChildren())
```
seando-adsk, line 381: We have been doing this in a different way that allows us to find and remove these checks at a later date when a new Maya PR ships. Once the Maya PR ships with this new Ufe 0.4.x version we don't need a test on the specific Ufe version anymore - just a check against Ufe v4 is okay. So to find them we always use either the existing `ufeFeatureSetVersion` function or `UFE_PREVIEW_VERSION_NUM` and fix/remove them. So I think your test should be something like:
`if (os.getenv('UFE_PREVIEW_VERSION_NUM', '0000') > '4004'):`
This cmake variable is also set as an env var for all these Ufe tests.

pierrebai-adsk:
The pm;y failure was on Windows 2020 python 2 where a schema test timed out for unknown reason. The test is entirely unrelated to my changes.

ppt-adsk (approved):

seando-adsk (approved):

PR #2266, opened 2022-04-07
MAYA-122528 - MayaUSD should support UFE lights

vlasovi:
Create new class ufe::UsdLight inheriting from Ufe::Light to represent usd lights in UFE


lib/mayaUsd/ufe/UsdLight.cpp
```diff
@@ -0,0 +1,499 @@
...
+}
+
+std::shared_ptr<Ufe::Light::SphereInterface> UsdLight::sphereInterfaceImpl()
+{
+    return std::make_shared<UsdSphereInterface>(fItem);
+}
+
+std::shared_ptr<Ufe::Light::ConeInterface> UsdLight::coneInterfaceImpl()
+{
+    return std::make_shared<UsdConeInterface>(fItem);
```
williamkrick, line 490: I think in all of these methods we should assert that the interface is valid for the current light type.

test/lib/ufe/testLight.py
```diff
@@ -0,0 +1,173 @@
...
+    def testUsdLight(self):
+        self._StartTest('SimpleLight')
+        mayaPathSegment = mayaUtils.createUfePathSegment('|stage|stageShape')
+        lightUsdPathSegment = usdUtils.createUfePathSegment('/lights/simpleLight')
+        lightPath = ufe.Path([mayaPathSegment, lightUsdPathSegment])
+        lightItem = ufe.Hierarchy.createItem(lightPath)
+
+        ufeLight = ufe.Light.light(lightItem)
+        usdLight = usdUtils.getPrimFromSceneItem(lightItem)
+        self.assertEqual(ufeLight.type(), ufe.Light.Spot)
```
williamkrick, line 168: We need to test all the different light types we support. We should test setting values that don't exist on a given light type, expecting failure.

williamkrick (changes requested):

lib/mayaUsd/ufe/UsdLight.h
```diff
@@ -0,0 +1,161 @@
...
+    {
+        PXR_NAMESPACE_USING_DIRECTIVE
+        TF_AXIOM(fItem != nullptr);
+        return fItem->prim();
+    }
+
+    // Ufe::Light overrides
+    const Ufe::Path&    path() const override;
+    Ufe::SceneItem::Ptr sceneItem() const override;
+    Type                type() const override;
```
williamkrick, line 59: Something complicated possibly missing from the interface. Do we want to be able to change light types through Ufe?

If so does that mean removing any values that don't apply to the new type? In an undo-able way?

Maybe not being able to change the light type is a reasonable limitation for now, but not having the interface means adding that functionality is an ABI break.

lib/mayaUsd/ufe/UsdLight.cpp
vlasovi: All those functions are private and not called directly by the users. Instead the user calls methods Ufe::coneInterface etc. And in those methods we do the assertion as well as we make sure to not call coneInterfaceImpl if the type is not consistent with the interface

lib/mayaUsd/ufe/UsdLight.h
vlasovi: I don't see a usecase for that. You cant change type without recreating the object in USD, and you cannot do it in Maya as well.

vlasovi:
@ppt-adsk I wrapped the new code using UFE Light into UFE_V4_FEATURES_AVAILABLE macro, but the perflight still fails. I can also add a check against UFE_PREVIEW_VERSION_NUM, but this looks like a temporary solution that would require an additional PR in the future. How should I proceed to fix the preflights?

williamkrick (dismissed):

test/lib/ufe/testLight.py
```diff
@@ -0,0 +1,257 @@
...
+    def setUp(self):
+        ''' Called initially to set up the maya test environment '''
+        # Load plugins
+        self.assertTrue(self.pluginsLoaded)
+
+        # value from UsdGeomLinearUnits
+        self.inchesToCm = 2.54
+        self.mmToCm = 0.1
+
+    def _StartTest(self, testName):
```
ppt-adsk, line 73: Could be called in setupClass() or setUp().

test/lib/ufe/testLight.py
```diff
@@ -0,0 +1,257 @@
...
+        cmds.file(force=True, new=True)
+        mayaUtils.loadPlugin("mayaUsdPlugin")
```
ppt-adsk, line 75: Unnecessary, done by the (poorly-named) isMayaUsdPluginLoaded().

test/lib/ufe/testLight.py
```diff
@@ -0,0 +1,257 @@
+#!/usr/bin/env python
```
ppt-adsk, line 1: Would have been good to get some undo / redo testing done as well.  Perhaps next pull request.

lib/mayaUsd/ufe/UsdLight.cpp
```diff
@@ -0,0 +1,499 @@
...
+#include <pxr/usd/usdLux/rectLight.h>
+#include <pxr/usd/usdLux/shadowAPI.h>
+#include <pxr/usd/usdLux/shapingAPI.h>
+#include <pxr/usd/usdLux/sphereLight.h>
+
+namespace MAYAUSD_NS_DEF {
+namespace ufe {
+
+template <typename ValueTypeIn>
+class SetValueUndoableCommandImpl : public Ufe::SetValueUndoableCommand<ValueTypeIn>
```
ppt-adsk, line 38: Unfortunately we can't implement undo / redo in USD by setting the previous value on undo, because this is incomplete.  If you imagine a situation where the edit target is set to a higher-priority layer which has no current opinion, undo should completely remove the opinion, NOT set the new opinion to the previous value.

Fortunately we have helper classes for this which make it easy to do the right thing.  See the existing
https://github.com/Autodesk/maya-usd/blob/77c5b01b67a63b5c280bec4e1c8a30472c8431e2/lib/mayaUsd/ufe/UsdUndoableCommand.h#L36
and
https://github.com/Autodesk/maya-usd/blob/6e1c88af1bfb3652c7f6999aea41847c14fd1fee/lib/mayaUsd/ufe/UsdAttribute.cpp#L199
and their uses in the code base, and pick the one that's right for you.

lib/mayaUsd/ufe/UsdLight.cpp
```diff
@@ -0,0 +1,499 @@
...
+    } else if (usdPrim.IsA<UsdLuxSphereLight>()) {
+        const UsdLuxShapingAPI shapingAPI(usdPrim);
+        return shapingAPI.GetShapingConeAngleAttr().IsValid() ? Ufe::Light::Spot
+                                                              : Ufe::Light::Point;
+    }
+
+    return Ufe::Light::Invalid;
+}
+
+float getLightIntensity(const UsdPrim& prim)
```
ppt-adsk, line 142: Is it useful for these free functions to be visible in the MayaUsd::ufe namespace?  Or should they be private in the unnamed namespace for this translation unit?  No preference either way, just wondering.

ppt-adsk (changes requested):
Undo / redo implementation needs to use UsdUndoableItem / UsdUndoBlock, but you're really close.

ppt-adsk (approved):

PR #2268, opened 2022-04-08
MAYA-122640 fix crash with deactivate and console

pierrebai-adsk:
Fix a crash that happened when a prim was deactivated in the outliner while the MEL console was docked and MEl history options "show all commands" was on. It was due to recursive paint event corrupting the Qt internal paint engine.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
@@ -69,20 +69,21 @@ void addMetadataCount(
     }
 }
 
-const double* getInvisibleColor()
+std::vector<double> getInvisibleColor()
 {
-    auto colorInit = []() {
-        std::vector<double> rgb(3);
-        MDoubleArray        outlinerInvisibleColor;
+    static std::vector<double> rgb;
+    static bool                initialized = false;
+    if (!initialized) {
+        MDoubleArray outlinerInvisibleColor;
         if (MGlobal::executeCommand(
                 "displayRGBColor -q \"outlinerInvisibleColor\"", outlinerInvisibleColor)
             && (outlinerInvisibleColor.length() == 3)) {
+            rgb.resize(3);
             outlinerInvisibleColor.get(rgb.data());
+            initialized = true;
         }
-        return rgb;
-    };
-    static const std::vector<double> rgb = colorInit();
-    return rgb.data();
+    }
+    return rgb;
 }
 
 } // namespace
@@ -124,11 +125,15 @@ bool UsdUIInfoHandler::treeViewCellInfo(const Ufe::SceneItem::Ptr& item, Ufe::Ce
         if (!usdItem->prim().IsActive()) {
             changed = true;
             info.fontStrikeout = true;
-            const double* rgb = getInvisibleColor();
-            info.textFgColor.set(
-                static_cast<float>(rgb[0]), static_cast<float>(rgb[1]), static_cast<float>(rgb[2]));
-        } else {
-            info.textFgColor.set(0.403922f, 0.403922f, 0.403922f);
+            const std::vector<double> rgb = getInvisibleColor();
+            if (rgb.size() == 3) {
+                info.textFgColor.set(
+                    static_cast<float>(rgb[0]),
+                    static_cast<float>(rgb[1]),
+                    static_cast<float>(rgb[2]));
+            } else {
+                info.textFgColor.set(0.403922f, 0.403922f, 0.403922f);
+            }
         }
     }
 
--- a/lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
@@ -68,6 +68,23 @@ void addMetadataCount(
         addMetadataStrings(refs.size(), tooltip, needComma, singular, plural);
     }
 }
+
+const double* getInvisibleColor()
+{
+    auto colorInit = []() {
+        std::vector<double> rgb(3);
+        MDoubleArray        outlinerInvisibleColor;
+        if (MGlobal::executeCommand(
+                "displayRGBColor -q \"outlinerInvisibleColor\"", outlinerInvisibleColor)
+            && (outlinerInvisibleColor.length() == 3)) {
+            outlinerInvisibleColor.get(rgb.data());
+        }
+        return rgb;
+    };
+    static const std::vector<double> rgb = colorInit();
+    return rgb.data();
+}
+
 } // namespace
 
 namespace MAYAUSD_NS_DEF {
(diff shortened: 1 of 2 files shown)


pierrebai-adsk:
Only PF failure is the OSX master, which failed because the build failed to start. (Second times this happens today, the server seems flaky.)

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -68,6 +68,24 @@ void addMetadataCount(
...
 }
+
+std::vector<double> getInvisibleColor()
+{
+    static std::vector<double> rgb;
+    static bool                initialized = false;
+    if (!initialized) {
+        MDoubleArray outlinerInvisibleColor;
+        if (MGlobal::executeCommand(
+                "displayRGBColor -q \"outlinerInvisibleColor\"", outlinerInvisibleColor)
```
seando-adsk, line 79: Why is this command being echo'd in the Script Editor? The default value is `displayEnabled=False` so it should not echo. Are you saying that if echo all commands is on in the script editor then this will be echo'd even though we used false for display?

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -68,6 +68,24 @@ void addMetadataCount(
         addMetadataStrings(refs.size(), tooltip, needComma, singular, plural);
     }
 }
+
+std::vector<double> getInvisibleColor()
+{
+    static std::vector<double> rgb;
+    static bool                initialized = false;
+    if (!initialized) {
```
seando-adsk, line 76: We cannot do this once because the outlinerInvisibleColor can be modified while Maya is running and we should respect that.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
seando-adsk: If that is true then we need to find another way to get the color that doesn't involve executeCommand.

seando-adsk (changes requested):

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
pierrebai-adsk: IDK why it gets echoed even though echoing is off. Probably "echo all commands" overrides the flag?

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
pierrebai-adsk: Fixed it by registering a callback when the display colors change.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -68,6 +69,46 @@ void addMetadataCount(
         addMetadataStrings(refs.size(), tooltip, needComma, singular, plural);
     }
 }
+
+std::vector<double>& getInvisibleColor()
+{
+    static std::vector<double> rgb;
+    return rgb;
+}
```
seando-adsk, line 77: Great this is much better. Just one question. What is the reasoning for a static variable and helpers instead of a member var and private methods?

pierrebai-adsk: None. Just that the code evolved and this was the least amount of changes. OTOH, there is no reason to have the RGB value per-instance as the color is global, so that could be a post-hoc justification :).

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -68,6 +69,46 @@ void addMetadataCount(
...
+    MDoubleArray color;
+    MGlobal::executeCommand("displayRGBColor -q \"outlinerInvisibleColor\"", color);
+
+    // Note: support future possible alpha channel for invisible color.
+    if (color.length() >= 3) {
+        auto& rgb = getInvisibleColor();
+        rgb.resize(color.length());
+        color.get(rgb.data());
+    }
+}
```
seando-adsk, line 90: There is no need for this to be a vector. If you look at the command docs for `displayRGBColor` you need to add the `-alpha` flag to the query to get 4 floats. So I would suggest making this an std::array with a fixed size of 3.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
seando-adsk: There will always only be a **single** instance of this handler class. It is created and registered by Global.cpp and when someone asks for it, you get back the shared_ptr. So please use a member var to store the value and add private member functions to call the Maya command to get/update the value.

seando-adsk (changes requested):

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
pierrebai-adsk: I used to the size to detect that the array was initialized. I can use a separate variable for it.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -76,13 +77,55 @@ namespace ufe {
...
+    // a command in a painting event can cause a recursive paint
+    // event if commands echoing is on, which can corrupt the
+    // Qt paint internal which lead to a crash. Typical symptom
+    // is that the state variable of the Qt paint engine becomes
+    // null midway through the repaint.
+
+    MDoubleArray color;
+    MGlobal::executeCommand("displayRGBColor -q \"outlinerInvisibleColor\"", color);
+
+    if (color.length() >= 3) {
```
seando-adsk, line 113: Should this be (== 3) since the array is fixed to size 3?

lib/mayaUsd/ufe/UsdUIInfoHandler.h
```diff
@@ -47,6 +50,17 @@ class MAYAUSD_CORE_PUBLIC UsdUIInfoHandler : public Ufe::UIInfoHandler
...
+
+private:
+    void updateInvisibleColor();
+
+    // Note: the on-color-changed callback function is declared taking a void pointer
+    //       to be compatible with MMessage callback API.
+    static void onColorChanged(void*);
+
+    std::array<double, 3> fInvisibleColor;
+    bool                  fInvisibleColorValid = false;
```
seando-adsk, line 62: Aren't the RGB values between 0 and 1? So could we init this array to -1 values and use that as a valid check instead of using an extra boolean?

seando-adsk (changes requested):
Thanks for making the other changes. Just a couple of minor feedback now.

seando-adsk (approved):

pierrebai-adsk:
Manual PF passed except for the known Linux interactive image comparison due to the Maya default light intensity changes.

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
```diff
@@ -76,13 +77,59 @@ namespace ufe {
...
+    // a command in a painting event can cause a recursive paint
+    // event if commands echoing is on, which can corrupt the
+    // Qt paint internal which lead to a crash. Typical symptom
+    // is that the state variable of the Qt paint engine becomes
+    // null midway through the repaint.
+
+    MDoubleArray color;
+    MGlobal::executeCommand("displayRGBColor -q \"outlinerInvisibleColor\"", color);
+
+    if (color.length() == 3) {
```
ppt-adsk, line 118: Is this code correct?  color is an std::array, so its length will be 3 on construction.  Perhaps you wanted to check for the uninitialized value, which is -1?

ppt-adsk (changes requested):
Just a check that appears to be funky, but otherwise LGTM.  Actually, if the check always passes, wonder why it's there in the first place?

lib/mayaUsd/ufe/UsdUIInfoHandler.cpp
ppt-adsk: Ah, I see, confused fInvisibleColor with color, looks good then.

ppt-adsk (approved):

PR #2270, opened 2022-04-08
MAYA-114992 The true bindPose for a skin cluster comes from it's bindPreMatrix

williamkrick:
attribute. Use that rather than the dagPose node or the bindPose attribute of a joint.


vlasovi (approved):
Looks good

williamkrick:
Started preflight 2243.

vlasovi (approved):

Issue #171, opened 2019-12-11
dev: Non-UFE: VP2 Proxy - AL sub-item selection not working


pmolodo:
I tried building the latest dev branch recently, without UFE enabled, and selection of specific meshes within a USD with the AL proxy shape seems to be broken.

I was building against maya-2019.2.

Can anyone else confirm this?

elrond79:
So, I realized the problem was because I was using the mayaUsd.mod, which sets

VP2_RENDER_DELEGATE_PROXY=1

Once I disabled this, selection works

I renamed the ticket to reflect this new info...

elrond79:
Also, I think until this is resolved, we may want to NOT set VP2_RENDER_DELEGATE_PROXY by default in the .mod file, as item selection is a very important feature of the plugin...

pmolodo:
This is an old ticket, and selection has undergone major changes since then, so I'm going to close the ticket.  Can re-open if needed.

PR #2274, opened 2022-04-11
MAYA-105662 Maya USD does not build with gcc 9

boudrey:
Only fixed a few compiler warnings and I could build in Release with gcc 11.2.1


seando-adsk (approved):

boudrey:
I ran the PF 3 times, it failed 2 times on "MayaUsd [Python 2] - Branch - Maya 2020 - Preflight (Windows)" because of Time Out in 2 different tests!
157/157 Test #156: testUsdMayaListShadingModesCommand ..............***Timeout 500.03 sec
157/157 Test #171: GTest:AL_USDTransactionTests ....................***Timeout 500.01 sec
Pipeline seems very unstable, my change has nothing to do with this.

PR #2275, opened 2022-04-11
MAYA-122641 It is not safe to call setParameter from a worker thread because it

williamkrick:
triggers consolidation changes.


vlasovi (dismissed):

JGamache-autodesk (dismissed):

williamkrick:
@JGamache-autodesk the change to the location where I'm setting the diffuse value on the fallback material means that points and basisCurve prims also have the value set, which I think is fine?

JGamache-autodesk (approved):
The more I look, the more this seems to be the right place. This causes visual changes, but if they improve the match with usdview, I am OK with them.

vlasovi (approved):

PR #2277, opened 2022-04-11
MAYA-122628 The geomBindTransform should not be added to the mesh's transform.

williamkrick:


williamkrick:
@dj-mcg I found this working with importing/exporting some Maya rigged characters. I see many geometries flying away into space when I take a Maya scene, export it and then import it back into Maya.

I believe having the geomBindTransform added to the geometry transform is not correct and when I remove this code my rig imports correctly.

Do you know the original tests/scenes that required the geomBindTransform to get added to the meshes transform?

vlasovi (approved):
The code change looks good to me, but I don't know the specific details of the mesh's transform

pkanyuk:
Hi William, Dan McGarry pinged me about this change to give it a look over.  I was able to verify the problem you're seeing when importing an exported UsdSkel asset, but unfortunately I don't think your proposed change is correct.  By removing the geomBindXform handling, any UsdSkel asset with non-identity geomBindXform attrs will no longer import correctly.  If you try importing the open source UsdSkel from Pixar's website, you'll see the problem clearly on the character's shoes: https://graphics.pixar.com/usd/release/dl_usdskel_examples.html

I've attached images showing how this asset works well in the existing mayaUsd importer but fails with the proposed change.

However, that still leaves the problem of why round tripping fails.  It seems like the issue isn't with import, but rather just export: we're getting the mesh transforms written both on the transform attribute and the geomBindXform attribute.  Given that we want the exported asset to resemble the imported one, I suspect it's the transform, not the geomBindXform, that we want to exclude on export.

Thanks!
-- Paul

geom_bind_xform_fail2
geom_bind_xform_working

williamkrick:
I think I understand what is wrong. The goal of the import code is not to round trip a Maya rig through UsdSkel and back into a Maya rig without changes. The goal is to try to reproduce UsdSkel behavior in Maya.

The key bit of information is in the comment here: https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/fileio/translators/translatorSkel.cpp#L1044. To goal of this code is to prevent a double transformation from happening when transforming a node which is a parent of both the skeleton and the mesh. In Maya when a parent of the skeleton and mesh is transformed we do the double transform. In Usd when a parent of the skeleton and mesh is transformed it _doesn't_ do the double transform.

Maya is doing deformation in worldSpace so the transformation is applied to the mesh & the skeleton, so we see the transformation twice. UsdSkel is doing the deformation in skeleton space, then transforming the result into world space. The transformation is only applied once.

That means when we export from Maya to UsdSkel the rig behavior changes. Then when we re-import the code tries to match the UsdSkel behavior, and we end up with behavior that doesn't match the original Maya rig or the UsdSkel rig.

The import code tries to match UsdSkel by disabling the inherits transform attribute of the shape, and then setting the shape's transform to the "correct" world space transform. That way if the transform of a common ancestor DAG node is modified the shape's world space transform doesn't change and we avoid the double transform. That is _implemented_ by assuming the geomBindTransform is the correct world space transform for the shape. But that is not true. If the current world transform is different than the geomBindTransform then the transform that is written into the shape's transform node is not correct.

I'm not sure what exactly to do, but the geomBindTransform doesn't necessarily match the world space transform so it can't be used.

williamkrick:
@pkanyuk @dj-mcg have you had a chance to look over the new approach? I believe I now am now setting the skinned mesh's transform correctly so that the mesh will be placed correctly in world space when inherits transform is disabled. This reproduces the UsdSkel behavior as best as possible without doing any Maya changes.

dj-mcg:
Sure let me get sync'd up and we'll give it a try

pkanyuk:
Hi William,

Sorry for the slow reply, I just tested the change and I have good news and bad news.  The good news is that your new change does import the HumanFemale usd file correctly, but when I tested another asset, the results were no longer correct.  On closer inspection, the problem is that your new change assumes that the "skelSpacePrimTransform" is the same thing as as the geomBindTransform attribute.  For HumanFemale, that's true.  For other assets however, that's not the case.  We have assets where meshes have identity skel space transforms, for non-identity geomBindTransforms.  Sorry about the wrinkle, but I think the translator does need to consider geomBindTransform to be correct.

Thanks!
-- Paul

williamkrick:
@pkanyuk I have to say I'm pretty confused about UsdSkel's geomBindTransform.

In the [Usd Skel Info](https://graphics.pixar.com/usd/dev/api/_usd_skel__intro.html) page it says:
`Geom Bind Transform: The transform that positions a skinned primitive in its binding state in world space.`

In the API doc for [UsdSkelSkinPointsLBS ](https://graphics.pixar.com/usd/dev/api/group___usd_skel___skinning_utils.html#gaf4936087fc756233585198ed1858b24d) it says: `... the geomBindTransform provides the transform that transforms the initial points into the same skeleton space that the skinning transforms were computed in.`

And in [Usd Skel Schemas](https://graphics.pixar.com/usd/dev/api/_usd_skel__schemas.html) it says: `This primvar [GeomBindTransform] stores the world space transform of a skinned primitive at bind time.` But then it goes on to say: `...placing the result in either skeleton space or world space.`

How can the result be in either skeleton space or world space? How can I know which space the result is in? Is the geomBindTransform going from object space to world space or object space to skeleton space?

My reading up until now has been that primvars:skel:geomBindTransform is the world space transform of the skinned prim at bind time. If that reading is correct that setting that value into the imported mesh's transform is wrong. Any transformation of the root of the skeleton at bind time will be applied twice to the deformed mesh (once from the geomBindTransform and once from the skinned rig applying it through the deformation).

I guess from what you have been saying that geomBindTransform is actually the skeleton space transform of the geometry at bind time. If so then setting that value into the imported mesh's transform is correct. In this case our export code is wrong because we're exporting a world space geomBindTransform, not a skeleton space geomBindTransform.

Looking at the source for UsdSkel I don't see any code doing a correcting for any shared transform that the geomBindTransform and skelRoot have, which makes me think geomBindTransform really is the skeleton space transform.

Can you definitively confirm that geomBindTransform is skeleton space rather than world space?

williamkrick:
@pkanyuk @dj-mcg the code that was writing the transform back into Maya was doing it in a way that doesn't respect the pivots of the Maya node. The codes for [BindingAPI: GeomBindTransform](https://graphics.pixar.com/usd/dev/api/_usd_skel__schemas.html) say they ignore any transform on the prim. I'm assuming it is also ignoring pivots. When we export we still put the transform and pivots onto the skinned mesh prim, so when we import we bring back the pivots and transform that was there. Then the code in `_ConfigureSkinnedObjectTransform` was replacing the transform information with the geomBindTransform. BUT the pivots were still set, and the method of writing the decomposed transformation matrix into the TRS attributes of the transform thinks you want those new TRS values relative to the pivots.

The new code sets the full matrix into the Maya transform node using a method meant to take a world space transformation, so it computes the corrected TRS based on the pivots that are already set.

dj-mcg:
Hey @williamkrick - sorry for the delay on this. I'm trying to build with your latest changes but I'm hitting a compilation error:

`translatorSkel.cpp:1078:17: error: ‘class Autodesk::Maya::OpenMaya20220000::MFnTransform’ has no member named ‘resetTransformation’; did you mean ‘transformation’?
     transformFn.resetTransformation(gbt);
                 ^~~~~~~~~~~~~~~~~~~`

Is that 2023-only API?

williamkrick:
@dj-mcg apologies that is my mistake, I didn't realize that is a new API. I should be able to use a different slightly less convenient method and get the same result, expect an update soon!

dj-mcg:
Hey @williamkrick - I noticed you made a change to that fn call - are you ready for us to test the latest changes out?

williamkrick:
@dj-mcg Yes, I believe the new version should work in 2022 and 2023. Thanks for testing it again.

pkanyuk:
Hi @williamkrick , sorry for the slow response, the latest is update is working great!  I import->export->imported the usd test cases I had that exercised geomBindTransforms and the results look correct.  Thanks a ton for the thoroughness on this change, it looks great to me.

williamkrick:
@pkanyuk Thanks for double checking. I'm just glad we figured it out.

@seando-adsk can you merge this one? We have Pixar's go-ahead.

PR #2278, opened 2022-04-11
MAYA-122656 remove assertion with incorrect expression.

williamkrick:


lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1360,8 +1360,6 @@ void HdVP2Mesh::_CreateSmoothHullRenderItems(
             numFacesWithoutRenderItem = 0;
         }
     }
-
-    TF_VERIFY(numFacesWithoutRenderItem == 0);
```
williamkrick, line 1364: This was meant to check that every face has a render item associated with it, but I changed how the _meshSharedData->_faceIdToGeomSubsetId data is interpreted so that every face always has a render item.

vlasovi (approved):

Issue #1070, opened 2021-01-13
Allow maya-usd to build against Qt in devkit


dgovil:
**Is your feature request related to a problem? Please describe.**
The current build system requires specifying an external Qt location for building features that require Qt. (as described here: https://github.com/Autodesk/maya-usd/issues/1005). Since that issue is closed, I am opening this one to track this specific request. 

This is prohibitive for certain companies where Qt's licensing model and legalities push them towards recommending the commercial license out of caution for dealing with the Qt source. 

It would be great if we could piggy back off of the devkit supplied Qt version to avoid having to deal with an independent licensing system. It would also simplify the build process for people if that could happen.

**Describe the solution you'd like**
Modify the build system to allow working with the devkit supplied Qt libs

Thanks!

kxl-adsk:
Thank you for logging this one @dgovil. This is really a request for Maya and we have already raised it internally in the past. I will keep it open here for now and see if anything moved internally.

dgovil:
This has become a blocker for Maya 2024 unfortunately on arm64.
There's no arm64 version of Qt5 available, and it still isn't possible to build against the Qt libs with Maya.

Does Autodesk have instructions on building Qt from source that they can share so I can match your setup at least?
Otherwise this prevents all UFE use

dgovil:
So since this was a blocker for us adopting Maya 2024, I started reworking the CMake a bit.
Something about the way the Maya devkit Qt is laid out wasn't being conducive to how CMake's FindPackage expected it, so I just duplicated enough of what CMake does to make it work.

Here's a patch for anyone interested. It seems to work fine for Maya 2022 and above.  
[devkit.patch](https://github.com/Autodesk/maya-usd/files/11492158/devkit.patch)

I will leave this issue open though because I do think its something that ideally the project should solve. I don't however know enough details about the Linux and Windows side to submit my changes as a cleaned up patch.

rosenio:
Thanks @dgovil for that patch!
For thoes whot need it on Windows I've did a patch for Maya 2022 and above.

[devkit.patch](https://github.com/Autodesk/maya-usd/files/11533938/devkit.patch)

dgovil:
Thanks, Sean. I've been using your commits in our builds and it's been working great so far.

PR #2280, opened 2022-04-11
MAYA-122250 UFE 3 build as Universal Binary 2 (x86_64 + arm64)

boudrey:
+ Change a few compiler flags to be at par with Maya
+ Fix some easy Python 3 issues.

### DIFF
--- a/cmake/compiler_config.cmake
+++ b/cmake/compiler_config.cmake
@@ -5,8 +5,9 @@ set(GNU_CLANG_FLAGS
     # we want to be as strict as possible
     -Wall
     $<$<BOOL:${BUILD_STRICT_MODE}>:-Werror>
+    $<$<CONFIG:DEBUG>:-fstack-check>
     # optimization
-    -msse3
+    -msse4.2
     # disable warnings
     -Wno-deprecated
     -Wno-deprecated-declarations
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,9 @@ option(BUILD_TESTS "Build tests." ON)
 option(BUILD_STRICT_MODE "Enforce all warnings as errors." ON)
 option(BUILD_SHARED_LIBS "Build libraries as shared or static." ON)
 option(BUILD_WITH_PYTHON_3 "Build with python 3." OFF)
+if(APPLE)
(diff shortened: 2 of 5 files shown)


dgovil:
This will be very welcome :-)
Out of curiosity, since Qt5 doesn't have a Universal Binary 2 build, I assume you're patching it to build for both arches? Are you using the SideFX fork with the chromium backports?

A follow up question is, since Maya USD has historically required developers to provide a Qt5 install to be able to do the full build with the Qt tools, how do you see this working for devs who build maya-usd from source? Will Maya USD and the associated devkit be self sufficient for Qt in the future? (see https://github.com/Autodesk/maya-usd/issues/1070 for example of a prior request)

Anyway, this will be a very nice change to have overall! Thanks!

cmake/compiler_config.cmake
```diff
@@ -5,8 +5,9 @@ set(GNU_CLANG_FLAGS
     # we want to be as strict as possible
     -Wall
     $<$<BOOL:${BUILD_STRICT_MODE}>:-Werror>
+    $<$<CONFIG:DEBUG>:-fstack-check>
     # optimization
-    -msse3
+    -msse4.2
```
seando-adsk, line 10: These GNU_CLANG_FLAGS are used by both clang and gcc. Is this change okay for our Linux builds?

seando-adsk (approved):

cmake/compiler_config.cmake
seando-adsk: Yes, but current Maya is building with gcc 9, whereas we are still using gcc 6.3.1.

boudrey: This flag has been there for at least in Feb. 8, 2019 in Maya and it was merged from Peptide so it was in Peptide before that.

boudrey:
Hi @dgovil, maybe it would be for Qt6 only. Need to start providing UB2 across all our components if we want to finally get a version of Maya with UB2.

cmake/compiler_config.cmake
seando-adsk: Okay thanks for checking.

seando-adsk (approved):

PR #2282, opened 2022-04-11
LOOKDEVX-641: fix for latest Maya PR

seando-adsk:


cmake/test.cmake
```diff
@@ -235,6 +236,10 @@ finally:
          "${CMAKE_INSTALL_PREFIX}/plugin/adsk/scripts")
     list(APPEND MAYAUSD_VARNAME_MAYA_SCRIPT_PATH
          "${CMAKE_INSTALL_PREFIX}/plugin/adsk/scripts")
+    list(APPEND MAYAUSD_VARNAME_PXR_MTLX_STDLIB_SEARCH_PATHS
+         "${PXR_USD_LOCATION}/libraries")
+    list(APPEND MAYAUSD_VARNAME_PXR_MTLX_STDLIB_SEARCH_PATHS
+         "${CMAKE_INSTALL_PREFIX}/libraries")
```
seando-adsk, line 242: Add these two env vars here so that every test will have them. Jonathan's new testShaderDef requires them.

lib/mayaUsd/ufe/UsdShaderNodeDef.cpp
```diff
@@ -98,7 +98,6 @@ Ufe::NodeDefs UsdShaderNodeDef::definitions(const std::string& category)
         PXR_NS::SdrRegistry&        registry = PXR_NS::SdrRegistry::GetInstance();
         PXR_NS::SdrShaderNodePtrVec shaderNodeDefs = registry.GetShaderNodesByFamily();
         for (const PXR_NS::SdrShaderNodeConstPtr& shaderNodeDef : shaderNodeDefs) {
-            const std::string& type = shaderNodeDef->GetName();
```
seando-adsk, line 101: Fix warning (as error) on Linux/Mac - unused variable. Found when blessing the latest Maya since this is new code which requires an updated Ufe.

cmake/test.cmake
JGamache-autodesk: Looks good. Can you make this conditional on the presence of MaterialX (for older Maya/USD build combinations).

JGamache-autodesk (approved):

cmake/test.cmake
seando-adsk: I could wrap them inside of `CMAKE_WANT_MATERIALX_BUILD` - would that be good?

test/lib/mayaUsd/render/vp2RenderDelegate/CMakeLists.txt
```diff
@@ -67,8 +67,6 @@ if (CMAKE_WANT_MATERIALX_BUILD)
             "MAYA_PLUG_IN_PATH=${CMAKE_INSTALL_PREFIX}/lib/maya"
             "LD_LIBRARY_PATH=${ADDITIONAL_LD_LIBRARY_PATH}"
             "MAYA_LIGHTAPI_VERSION=${MAYA_LIGHTAPI_VERSION}"
-            "PXR_MTLX_STDLIB_SEARCH_PATHS=${PXR_USD_LOCATION}/libraries"
-            "PXR_MTLX_STDLIB_SEARCH_PATHS=${CMAKE_INSTALL_PREFIX}/libraries"
```
seando-adsk, line 71: This test is inside of `CMAKE_WANT_MATERIALX_BUILD`, but the one below is not. Is that a mistake or does it need these variables regardless of whether or not we have MaterialX build?

cmake/test.cmake
JGamache-autodesk: Yes. that would work since test.cmake is processed after we are done setting CMAKE_WANT_MATERIALX_BUILD.

test/lib/mayaUsd/render/vp2RenderDelegate/CMakeLists.txt
JGamache-autodesk: We don't need the variables but it does not hurt if they are there anyways. Probably means that we do not need to check for CMAKE_WANT_MATERIALX_BUILD in test.cmake, which means this PR is 100% correct and does not need modifications.

cmake/test.cmake
JGamache-autodesk: But having extra variable does not hurt. No need for an extra check here. Sorry for the noise.

JGamache-autodesk (approved):

seando-adsk:
I ran the preflight manually (as our internal build system is currently not working because of a GitHub API change) and it succeeded.

PR #2284, opened 2022-04-12
Refactored and simplified AL_USDMaya primvar interpolation type detection

J-Mo63:
This PR removes several unused primvar interpolation type detection functions from `AL_USDMaya` related to `Vec3` and `Vec4`, as well as bolstering the testing performed for UV and colour set interpolation type detection functions. The implementation of the used interpolation type detection methods has been preserved for now due to the ongoing migration effort to the Autodesk plugin. On a side note, these tests historically use a 31-vert, 8-faced shape (one face of 3 points) - keep in mind this is an open shape.


seando-adsk:
@williamkrick Can you have a quick look at this change in the AL plugin. Thanks.

williamkrick (approved):

seando-adsk:
@J-Mo63 We are having problems with our internal preflight system because of a recent API change on Github. So if this is not urgent can we wait until that is fixed before running the preflight and getting this merged in?

Sean

J-Mo63:
> So if this is not urgent can we wait until that is fixed before running the preflight and getting this merged in?

No worries @seando-adsk! How long do you suspect it will be down for? We have a couple of PRs we'll be sending up over the next week or so.

seando-adsk:
@J-Mo63 The github problem is fixed. I ran the preflight and there was a single test timeout error:
    157/157 Test #171: GTest:AL_USDTransactionTests ....................***Timeout 500.02 sec
I don't think that was affected by your changes. Can you confirm?

Thanks, Sean

J-Mo63:
> @J-Mo63 The github problem is fixed. I ran the preflight and there was a single test timeout error: 157/157 Test #171: GTest:AL_USDTransactionTests ....................***Timeout 500.02 sec I don't think that was affected by your changes. Can you confirm?
> 
> Thanks, Sean

Unfortunately we can't test for Py2/2020 on Windows here, but I would be quite surprised if the changelist on this PR would affect it. I'll check with the team on that though.

PR #2285, opened 2022-04-12
Implemented Outliner orphaned nodes, pulled node tooltips and italics.

ppt-adsk:
Provide visual feedback for Maya pulled nodes that are orphaned by a change in the USD hierarchy above them.

Also provide better tool tip feedback for pulled nodes.


lib/mayaUsd/ufe/MayaUIInfoHandler.cpp
```diff
@@ -0,0 +1,144 @@
...
+// the root of the pulled sub-hierarchy.
+//
+// Note that if it exists, the USD pulled ancestor prim is inactive, which is
+// done to avoid having that prim and its sub-hierarchy exist in the scene as
+// stale USD duplicates of Maya pulled nodes.  If the USD pulled ancestor does
+// not exist, the argument Maya node is orphaned.
+MayaUsd::ufe::UsdSceneItem::Ptr pulledUsdAncestorItem(const Ufe::SceneItem::Ptr& mayaItem)
+{
+    // This function requires a Maya item to compute its USD ancestor.
+    TF_AXIOM(mayaItem->runTimeId() == MayaUsd::ufe::getMayaRunTimeId());
```
seando-adsk, line 45: Isn't that pretty harsh to abort on this condition (and the one below)? Why not just return nullptr?

lib/mayaUsd/ufe/MayaUIInfoHandler.cpp
```diff
@@ -0,0 +1,144 @@
...
+//------------------------------------------------------------------------------
+// Ufe::UIInfoHandler overrides
+//------------------------------------------------------------------------------
+
+bool MayaUIInfoHandler::treeViewCellInfo(const Ufe::SceneItem::Ptr& mayaItem, Ufe::CellInfo& info)
+    const
+{
+    if (isOrphaned(mayaItem)) {
+        // If the Maya node is orphaned, give it the orphaned color.
+        info.textFgColor.set(0.403922f, 0.403922f, 0.403922f);
```
seando-adsk, line 101: Is this not a color we can query, rather than hard code it?

seando-adsk (approved):

PR #2288, opened 2022-04-12
Fix build for USD versions past 21.11

dj-mcg:
GetRenderTag is a function defined in HdRprim, but the callsite has been
moved from classes derived from HdRprim to a different parent class.
Pass the render tag into the _SyncSharedData from the HdRprim derived
classes to fix the build

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -503,7 +503,13 @@ void HdVP2BasisCurves::Sync(
         }
     }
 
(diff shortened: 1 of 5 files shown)


vlasovi (approved):
Looks good

williamkrick:
Because the preflight hook is not launching properly I've started a manual preflight run for this change. The build number of 2224. I created a copy of the changes in a branch local to Autodesk/maya-usd to achieve this: https://github.com/Autodesk/maya-usd/tree/krickw/Pass_RenderTag_To_Common_Sync_Fn which I will delete once this gets merged.

williamkrick:
I've started a new internal preflight 2234 to validate this one.

PR #2291, opened 2022-04-13
Set the default light intensity the new way. Also force it to be set 

williamkrick:
for testMtohVisibility.py which for some reason needs an explicit call.

### DIFF
--- a/test/testUtils/imageUtils.py
+++ b/test/testUtils/imageUtils.py
@@ -20,10 +20,14 @@ KNOWN_FORMATS = {
     'png': 32,
 }
 
-"""If the current Maya version supports setting the default light intensity,
-    then restore it to 1 so snapshots look equal across versions."""
-if maya.mel.eval("optionVar -exists defaultLightIntensity"):
-    maya.mel.eval("optionVar -fv defaultLightIntensity 1")
+def resetDefaultLightIntensity():
+    """If the current Maya version supports setting the default light intensity,
+        then restore it to 1 so snapshots look equal across versions."""
+    if maya.mel.eval("optionVar -exists defaultLightIntensity"):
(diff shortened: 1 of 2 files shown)


test/testUtils/imageUtils.py
```diff
@@ -20,10 +20,14 @@
...
-if maya.mel.eval("optionVar -exists defaultLightIntensity"):
-    maya.mel.eval("optionVar -fv defaultLightIntensity 1")
+def resetDefaultLightIntensity():
+    """If the current Maya version supports setting the default light intensity,
+        then restore it to 1 so snapshots look equal across versions."""
+    if maya.mel.eval("optionVar -exists defaultLightIntensity"):
+        maya.mel.eval("optionVar -fv defaultLightIntensity 1")
+    if cmds.attributeQuery('defaultLightIntensity', node='hardwareRenderingGlobals', exists=True):
+        cmds.setAttr('hardwareRenderingGlobals.defaultLightIntensity', 1.0)
+resetDefaultLightIntensity()
```
williamkrick, line 30: @JGamache-autodesk It seems like this top level code is not running in the test suite. I ended up wrapping it in a function and running it in testMtohVisibility.py and that works. But all the other tests that do screen shots also fail and don't seem to be running the top level code.

AFAIK importing the module should run the top level code, so I'm not sure what is going wrong.

JGamache-autodesk: If the values are stored in the scene, this means the value will be reset every time we open a scene, so it is no longer sufficient to run this only once. It must be run after each loadScene, so your call in `snapshot` is really what is now needed.

JGamache-autodesk (dismissed):

JGamache-autodesk (approved):

williamkrick:
Internal preflight 2232 for this change has passed.

PR #2307, opened 2022-04-22
MAYA-122104 ProxyShapeSceneSegmentHandler and UsdCameraHandler::findCameras.

williamkrick:


lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+Ufe::Selection ProxyShapeSceneSegmentHandler::findGatewayItems(const Ufe::Path& path) const
+{
+    Ufe::Selection result = fMayaSceneSegmentHandler ? fMayaSceneSegmentHandler->findGatewayItems(path) : Ufe::Selection();
+    for(const auto& stage : getAllStages()) {
+        Ufe::Path proxyShapePath = stagePath(stage);
+
+        // This handles gateway nodes that have a mix of children in other scene segments and children in the same
+        // segment as the gateway node, as well as gateway nodes with multiple descendent scene segments.
+        if (proxyShapePath == path) {
```
williamkrick, line 56: This extra `proxyShapePath == path` block is what is necessary to handle the special case of going into the child scene segments when passed a gateway item.

lib/mayaUsd/ufe/UsdCameraHandler.cpp
```diff
@@ -52,5 +56,25 @@ Ufe::Camera::Ptr UsdCameraHandler::camera(const Ufe::SceneItem::Ptr& item) const
     return UsdCamera::create(usdItem);
 }
 
+#if defined(UFE_V4_FEATURES_AVAILABLE) && (UFE_PREVIEW_VERSION_NUM >= 4008)
+Ufe::Selection UsdCameraHandler::findCamerasInSceneSegment(const Ufe::Path& path) const
```
williamkrick, line 60: Native USD search for cameras.

vlasovi (dismissed):
Looks good. I'm only wondering how much faster is this approach compared to just traversing Ufe::Hierarchy. In the end we still have to traverse USD hierarchy in search of cameras.

williamkrick:
> Looks good. I'm only wondering how much faster is this approach compared to just traversing Ufe::Hierarchy. In the end we still have to traverse USD hierarchy in search of cameras.

You are right in that the complexity of the algorithm is the same. The speed difference comes from eliminating the overhead of calls to Ufe during the traversal. Using Ufe::Hierarchy to traverse means creating many additional Ufe::SceneItems, Ufe::Paths and Ufe::Hierarchy objects, which means allocating a bunch of std::vector<>, std::shared_ptr<> etc. Doing the iteration with USD means we don't do any allocation when the object is not a gateway item or a camera (which is most objects).

Avoiding the overhead of Ufe when doing the hierarchy traversal ends up saving a significant amount of time. This version is something like 10x faster then the original code.

lib/mayaUsd/ufe/Global.cpp
```diff
@@ -246,6 +273,11 @@ MStatus finalize(bool exiting)
     Ufe::RunTimeMgr::instance().unregister(g_USDRtid);
     g_MayaHierarchyHandler.reset();
 
+#if defined(UFE_V4_FEATURES_AVAILABLE) && (UFE_PREVIEW_VERSION_NUM >= 4008)
```
ppt-adsk, line 276: Shouldn't we reset the g_MayaCameraHandler as well?

lib/mayaUsd/ufe/ProxyShapeCameraHandler.cpp
```diff
@@ -0,0 +1,74 @@
...
+// Ufe::CameraHandler overrides
+//------------------------------------------------------------------------------
+Ufe::Camera::Ptr ProxyShapeCameraHandler::camera(const Ufe::SceneItem::Ptr& item) const
+{
+    return nullptr;
+}
+
+Ufe::Selection ProxyShapeCameraHandler::find(const Ufe::Path& path) const
+{
+    // Allow the gateway item to also have children in the same runtime.
```
ppt-adsk, line 55: Do we really need to do this?  The ProxyShape is a shape node --- shape nodes only have children in limited circumstances, and can they really be cameras?

lib/mayaUsd/ufe/ProxyShapeCameraHandler.cpp
```diff
@@ -0,0 +1,74 @@
...
+    Ufe::Selection cameras = fMayaCameraHandler ? fMayaCameraHandler->find(path) : Ufe::Selection();
```
ppt-adsk, line 56: Shouldn't this be the else clause of the isAGatewayType()?  E.g. as per
https://github.com/Autodesk/maya-usd/blob/3bd67035ccee8936b68a2623e9defe0551204c0a/lib/mayaUsd/ufe/ProxyShapeHierarchyHandler.cpp#L50

lib/mayaUsd/ufe/ProxyShapeCameraHandler.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    Ufe::SceneItem::Ptr item = Ufe::Hierarchy::createItem(path);
+    if (isAGatewayType(item->nodeType()))
+    {
+        // path is a path to a proxyShape node.
+        // Get the UsdStage for this proxy shape node and search it for cameras
+        PXR_NS::UsdStageWeakPtr stage = getStage(path);
+        TF_VERIFY(stage);
+        Ufe::Selection usdCameras = UsdCameraHandler::find(path, path, stage->GetPseudoRoot());
```
ppt-adsk, line 65: If my hunch above is right, you can just return UsdCameraHandler::find(...);

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
```diff
@@ -0,0 +1,88 @@
...
+{
+    // Handle other gateway node types that MayaUSD is not aware of
+    bool result = fMayaSceneSegmentHandler
+        ? fMayaSceneSegmentHandler->isGateway_(path)
+        : false;
+
+    PXR_NS::UsdStageWeakPtr stage = getStage(path);
+
+
+    return result | !(!stage); // TfRefPtr only has operator ! to allow me to test if the stage is nullptr
```
ppt-adsk, line 84: Couldn't this just be
return stage ? true : fMayaSceneSegmentHandler ? fMayaSceneSegmentHandler->isGateway_(path) : false;

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.h
```diff
@@ -0,0 +1,60 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#pragma once
+
+#include <mayaUsd/base/api.h>
+#include <mayaUsd/ufe/UfeVersionCompat.h>
```
ppt-adsk, line 19: Do we need this include?

lib/mayaUsd/ufe/UsdCameraHandler.cpp
```diff
@@ -52,5 +59,31 @@ Ufe::Camera::Ptr UsdCameraHandler::camera(const Ufe::SceneItem::Ptr& item) const
...
+}
+
+/*static*/
+Ufe::Selection UsdCameraHandler::find(const Ufe::Path& stagePath, const Ufe::Path& searchPath, const PXR_NS::UsdPrim& prim)
+{
+    Ufe::Selection result;
+    if (prim.IsA<PXR_NS::UsdGeomCamera>()) {
+        result.append(Ufe::Hierarchy::createItem(searchPath));
+    }
+    PXR_NS::UsdPrimSubtreeRange range = prim.GetDescendants();
```
ppt-adsk, line 77: Nice!

ppt-adsk (changes requested):
Needs a test.

williamkrick:
These changes are already merged on top of @JGamache-autodesk work for https://github.com/Autodesk/maya-usd/pull/2315, so merge that one first.

The only thing I'm missing is a test for SceneSegmentHandler. I forgot the Ufe::RunTimeMgr Python API that I need to write the test, so it is going to come in another PR.

seando-adsk:
I added the label "do not merge yet" as a reminder to myself to merge #2315 first.

test/lib/ufe/testSceneSegment.py
```diff
@@ -0,0 +1,97 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        standalone.uninitialize()
+
+    def setUp(self):
+        ''' Called initially to set up the maya test environment '''
+        # Load plugins
+        self.assertTrue(self.pluginsLoaded)
+
+    def _StartTest(self, testName):
```
ppt-adsk, line 57: Why is this different from setUp()?

ppt-adsk (changes requested):
Thanks for the tests.  Would be great if you could adjust the test conditionals to follow the pattern we've done in the past, makes it easier to remove these obsolete conditionals at the next Maya preview release.

Was quite confusing to review your changes and Jerry's changes in the same pull request.  Can't we just merge in Jerry's changes first, then you can update from dev and rebase your commits on top?

lib/mayaUsd/ufe/CMakeLists.txt
```diff
@@ -94,6 +94,29 @@ if(CMAKE_UFE_V4_FEATURES_AVAILABLE)
...
+
+    if (${UFE_PREVIEW_VERSION_NUM} GREATER_EQUAL 4009)
+        target_sources(${PROJECT_NAME}
+            PRIVATE
+                ProxyShapeCameraHandler.cpp
+                ProxyShapeSceneSegmentHandler.cpp
+        )
+    endif()
+
+    if (${UFE_PREVIEW_VERSION_NUM} GREATER_EQUAL 4010)
```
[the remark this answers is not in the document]
williamkrick: We're in kind of a strange state because we've had a bunch of UFE preview versions go by without getting them into Maya. Everyone is building locally slightly differently. But no QA-testable build with Ufe 0.4.9 is ever going to exist, which means these could all be testing for 0.4.10 (or 0.4.11 if my extra python API changes are merged). That simplifies the cmake. But we don't know which version to check for until something gets merged into Maya, and we can't merge into Maya until this gets merged...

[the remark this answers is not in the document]
ppt-adsk: It seems to me we can merge this with GREATER_EQUAL 4010 (or 4011) at any time.  Local developer builds will have 0.4.10 (or 0.4.11, as the case may be), so maya-usd will work with the newer version, as well as with the older UFE version.  Am I missing something?

test/lib/ufe/testCamera.py
```diff
@@ -234,13 +235,70 @@ def testUsdCamera(self):
...
+        self.assertTrue(result.contains(camera1Path))
+        self.assertTrue(result.contains(camera2Path))
+        self.assertEqual(len(result), 2)
+
+        # searching the the parent of a gateway item searches the Maya scene segment
+        # for cameras and searches nested scene segments.
+        result = ufe.CameraHandler.findAll(proxyShapeParentItem)
+        self.assertTrue(result.contains(camera1Path))
+        self.assertTrue(result.contains(camera2Path))
+        self.assertEqual(len(result), 2)
```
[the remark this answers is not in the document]
williamkrick: The Maya camera handler is not implemented, and there are no cameras that are children of proxyShapeParentItem. If we had the Maya camera handler and we had searched the Maya root path then we'd find front, side, top and persp.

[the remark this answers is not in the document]
ppt-adsk: Right, so do we plan on leaving the Maya camera handler unimplemented, because there is no demonstrated need for it?  If so, would be nice to add a comment to that effect in the test.

test/lib/ufe/testSceneSegment.py
williamkrick: I'm not really sure. I copied most of the test from testCamera.py and it has this pattern too. Probably I copied it for testCamera from somewhere else.

I guess if all the tests are running with the same USD file (which for testCamera and testSceneSegment they are running with the same USD file, then the file could get loaded in setUp and each test could re-use it? That would make the test run faster. That should work relatively easily here because I'm not actually modifying the file at all.

I'll try it and see how it goes.

test/lib/ufe/testCamera.py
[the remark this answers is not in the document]
williamkrick: Sure I'll add a comment.

williamkrick:
The new changes were around the feedback provided by @ppt-adsk on the tests for CameraHandler and SceneSegementHandler. I also re-based on top of dev to eliminate merge conflicts.

test/lib/ufe/testCamera.py
```diff
@@ -231,16 +231,73 @@ def _TestProjection(self, ufeCamera, usdCamera):
...
         self._TestCameraAttributes(cam2Camera, cameraPrim)
 
+    @unittest.skipUnless(os.getenv('UFE_PREVIEW_VERSION_NUM', '0000') >= '4011', 'Requires Ufe Preview Version at least 2011.')
+    def testUsdCameraHandler(self):
+        mayaPathSegment = mayaUtils.createUfePathSegment('|stage|stageShape')
+        cam2UsdPathSegment = usdUtils.createUfePathSegment('/cameras/cam2/camera2')
+        camera2Path = ufe.Path([mayaPathSegment, cam2UsdPathSegment])
+
+        # Test that the camera handlers can find USD cameras in a scene segment
+        proxyShapeParentSegment = mayaUtils.createUfePathSegment('|stage')
```
ppt-adsk, line 250: I wonder about these single-use UFE path segments --- can't we just use ufe.PathString.path() to create them out of concatenated strings?  E.g.
mayaPathSegmentStr = '|stage|stageShape'
camerasParentPath = ufe.PathString.path(mayaPathSegmentStr+'/cameras/cam2')
I get that you're re-using the Maya path segment, and in real code there may be a performance benefit to doing this, but in a test it seems so verbose.

test/lib/ufe/testCamera.py
```diff
@@ -231,16 +231,73 @@ def _TestProjection(self, ufeCamera, usdCamera):
...
+        camera1PathSegment = usdUtils.createUfePathSegment('/cameras/cam1/camera1')
+        
+        proxyShapePath = ufe.Path([mayaPathSegment])
+        proxyShapeParentPath = ufe.Path([proxyShapeParentSegment])
+        camerasParentPath = ufe.Path([mayaPathSegment, camerasParentPathSegment])
+        camera2ParentPath = ufe.Path([mayaPathSegment, camera2ParentPathSegment])
+        geoPath = ufe.Path([mayaPathSegment, geoPathSegment])
+        camera1Path = ufe.Path([mayaPathSegment, camera1PathSegment])
+
+        proxyShapeItem = ufe.Hierarchy.createItem(proxyShapePath)
```
ppt-adsk, line 263: Have a look at ufeUtils.createItem().  It will create a ufe.SceneItem directly out of a path string, I think this is what you want, and will remove a lot of boiler-plate code from your test.

ppt-adsk (commented):
Not a big deal, but it feels like the tests could have some fat trimmed off, sorry for noticing this late.  Otherwise looks great!

williamkrick:
Updated the tests to create the required paths in a more concise way. Updated the Ufe preview version checks to the latest.

ppt-adsk (approved):

PR #2309, opened 2022-04-25
MAYA-122128 NURBS defaut to off when caching

pierrebai-adsk:
The defaults were not parsed properly when a list with a single item was given as teh default as it could not determine if it was a text or floating-point values.


pierrebai-adsk:
The only PF failure is the Windows 2020 test export where it fails to delete a tmp file.

ppt-adsk (approved):

seando-adsk (approved):

PR #2315, opened 2022-04-27
Ufe 0.4.8: Add classifications and metadata to def

JGamache-autodesk:
Ufe 0.4.8 has multiple changes to improve NodeDef discovery:

NodeDef:
- Expose read-only interface to query node classifications (allows
  sorting them in a node picker)
- Expose read-only metadata API (limited to known USD metadata tags at
  the moment)
- Remove internally kept data. Everything is built on-the-fly from the
  SdrShaderNode information
- Expanded the port API to allow name discovery and single port queries

AttributeDef:
- The base class in Ufe is now a full interface. Added an implementation
  for USD that returns the information directly from the associated
  SdrShaderProperty
- Expose read-only metadata API (limited to known USD tags at the
  moment)

### DIFF
--- a/lib/mayaUsd/ufe/UsdShaderAttributeDef.cpp
+++ b/lib/mayaUsd/ufe/UsdShaderAttributeDef.cpp
@@ -1,0 +1,102 @@
+//
+// Copyright 2022 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include "UsdShaderAttributeDef.h"
+
+#include "Global.h"
+#include "Utils.h"
+
+#include <pxr/base/tf/token.h>
+#include <pxr/usd/ndr/declare.h>
+#include <pxr/usd/sdr/shaderProperty.h>
+
+#include <map>
+#include <vector>
+
+namespace MAYAUSD_NS_DEF {
+namespace ufe {
+
+UsdShaderAttributeDef::UsdShaderAttributeDef(
+    const PXR_NS::SdrShaderPropertyConstPtr& shaderAttributeDef)
+    : Ufe::AttributeDef()
+    , fShaderAttributeDef(shaderAttributeDef)
+{
+    PXR_NAMESPACE_USING_DIRECTIVE
+    if (!TF_VERIFY(fShaderAttributeDef)) {
+        throw std::runtime_error("Invalid shader attribute definition");
+    }
+}
+
+UsdShaderAttributeDef::~UsdShaderAttributeDef() { }
+
+std::string UsdShaderAttributeDef::name() const
+{
+    return fShaderAttributeDef ? fShaderAttributeDef->GetName().GetString() : std::string();
+}
+
+std::string UsdShaderAttributeDef::type() const
+{
+    Ufe::Attribute::Type ufeType = Ufe::Attribute::kInvalid;
+    if (fShaderAttributeDef) {
+        const PXR_NS::SdfValueTypeName typeName = fShaderAttributeDef->GetTypeAsSdfType().first;
+        ufeType = usdTypeToUfe(typeName);
+    }
+    return ufeType;
+}
+
+std::string UsdShaderAttributeDef::defaultValue() const
+{
+    std::ostringstream defaultValue;
+    if (fShaderAttributeDef) {
+        defaultValue << fShaderAttributeDef->GetDefaultValue();
+    }
+    return defaultValue.str();
+}
+
+Ufe::AttributeDef::IOType UsdShaderAttributeDef::ioType() const
+{
+    return (fShaderAttributeDef && fShaderAttributeDef->IsOutput()) ? Ufe::AttributeDef::OUTPUT_ATTR
+                                                                    : Ufe::AttributeDef::INPUT_ATTR;
+}
+
+Ufe::Value UsdShaderAttributeDef::getMetadata(const std::string& key) const
+{
+    if (fShaderAttributeDef) {
+        const PXR_NS::NdrTokenMap& metadata = fShaderAttributeDef->GetMetadata();
+        auto                       it = metadata.find(PXR_NS::TfToken(key));
+        if (it != metadata.cend()) {
+            return Ufe::Value(it->second);
+        }
+    }
+    // TODO: Adapt UI metadata information found in SdrShaderProperty to Ufe standards
+    // TODO: Fix Mtlx parser in USD to populate UI metadata in SdrShaderProperty
+    return {};
+}
+
+bool UsdShaderAttributeDef::hasMetadata(const std::string& key) const
+{
+    if (fShaderAttributeDef) {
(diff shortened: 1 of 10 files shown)


lib/mayaUsd/ufe/UsdShaderNodeDef.cpp
```diff
@@ -81,12 +100,211 @@ UsdShaderNodeDef::UsdShaderNodeDef(const PXR_NS::SdrShaderNodeConstPtr& shaderNo
...
+
+#if (UFE_PREVIEW_VERSION_NUM >= 4010)
+Ufe::SceneItem::Ptr UsdShaderNodeDef::createNode(
+    const Ufe::SceneItem::Ptr& parent,
+    const Ufe::PathComponent&  name) const
+{
+    Ufe::SceneItem::Ptr createdItem = nullptr;
+
+    if (fShaderNodeDef) {
+        UsdSceneItem::Ptr parentItem = std::dynamic_pointer_cast<UsdSceneItem>(parent);
```
ppt-adsk, line 283: Shouldn't we be doing some kind of null ptr check?

lib/mayaUsd/ufe/UsdShaderNodeDef.cpp
```diff
@@ -81,12 +100,211 @@ UsdShaderNodeDef::UsdShaderNodeDef(const PXR_NS::SdrShaderNodeConstPtr& shaderNo
...
+        UsdUndoCreateFromNodeDefCommand::Ptr cmd
+            = UsdUndoCreateFromNodeDefCommand::create(fShaderNodeDef, parentItem, name.string());
+        if (cmd) {
+            cmd->execute();
```
ppt-adsk, line 287: I know we do this a lot, mostly as a convenience to avoid having to code command and non-command versions of the UFE API, but it does feel awkward that we can't directly call what UsdUndoCreateFromNodeDefCommand is calling, and have to create a command object and call execute() on it.  Couldn't there be some kind of static engine method on UsdUndoCreateFromNodeDefCommand that would do the work without a command object?

lib/mayaUsd/ufe/UsdShaderNodeDef.cpp
```diff
@@ -81,12 +100,211 @@ UsdShaderNodeDef::UsdShaderNodeDef(const PXR_NS::SdrShaderNodeConstPtr& shaderNo
...
+
+    return createdItem;
+}
+
+Ufe::InsertChildCommand::Ptr UsdShaderNodeDef::createNodeCmd(
+    const Ufe::SceneItem::Ptr& parent,
+    const Ufe::PathComponent&  name) const
+{
+    if (fShaderNodeDef) {
+        UsdSceneItem::Ptr parentItem = std::dynamic_pointer_cast<UsdSceneItem>(parent);
```
ppt-adsk, line 300: Null ptr check?

lib/mayaUsd/ufe/UsdShaderAttributeDef.cpp
```diff
@@ -0,0 +1,102 @@
...
+    if (!TF_VERIFY(fShaderAttributeDef)) {
+        throw std::runtime_error("Invalid shader attribute definition");
+    }
+}
+
+UsdShaderAttributeDef::~UsdShaderAttributeDef() { }
+
+std::string UsdShaderAttributeDef::name() const
+{
+    return fShaderAttributeDef ? fShaderAttributeDef->GetName().GetString() : std::string();
```
ppt-adsk, line 47: Given the exception raised in the constructor, do we need the null pointer checks in the other members?

lib/mayaUsd/ufe/UsdShaderAttributeDef.cpp
```diff
@@ -0,0 +1,102 @@
...
+    std::ostringstream defaultValue;
+    if (fShaderAttributeDef) {
+        defaultValue << fShaderAttributeDef->GetDefaultValue();
+    }
+    return defaultValue.str();
+}
+
+Ufe::AttributeDef::IOType UsdShaderAttributeDef::ioType() const
+{
+    return (fShaderAttributeDef && fShaderAttributeDef->IsOutput()) ? Ufe::AttributeDef::OUTPUT_ATTR
```
ppt-adsk, line 71: I would TF_AXIOM(fShaderAttributeDef);
and return the proper value.

ppt-adsk (changes requested):
Minor comments sprinkled in, but I am very hopeful we can get rid of 0.4.8 checks in favor of 0.4.10 checks, as UFE 0.4.8 and 0.4.9 will never be released publicly, unless I'm mistaken.

lib/mayaUsd/ufe/UsdShaderNodeDefHandler.cpp
```diff
@@ -51,16 +51,35 @@ Ufe::NodeDef::Ptr UsdShaderNodeDefHandler::definition(const Ufe::SceneItem::Ptr&
...
-    PXR_NS::TfToken        mxNodeType;
+    if (!shader) {
+        return nullptr;
+    }
+    PXR_NS::TfToken mxNodeType;
     shader.GetIdAttr().Get(&mxNodeType);
-    std::string type = mxNodeType.GetString();
-    return definition(type);
+
+    // Careful around name and identifier. They are not the same concept.
```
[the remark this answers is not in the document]
JGamache-autodesk: No. That would not be useful for clients. Famous last words, I know... we will cross that bridge when clients explicitly ask to specify which version of a node they want to use,

lib/mayaUsd/ufe/UsdShaderNodeDef.cpp
```diff
@@ -81,12 +100,211 @@ UsdShaderNodeDef::UsdShaderNodeDef(const PXR_NS::SdrShaderNodeConstPtr& shaderNo
...
 
 const Ufe::ConstAttributeDefs& UsdShaderNodeDef::outputs() const { return fOutputs; }
 
+#else
+
+std::string UsdShaderNodeDef::type() const
+{
+    // Returns the NodeDef name
+    // - Universal context: UsdPreviewSurface, UsdUVTexture, UsdPrimvarReader_float2
+    // - MaterialX context: ND_standard_surface_surfaceshader, ND_image_color3
```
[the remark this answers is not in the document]
JGamache-autodesk: Comment was removed.

lib/mayaUsd/ufe/UsdShaderNodeDef.cpp
```diff
@@ -81,12 +100,211 @@ UsdShaderNodeDef::UsdShaderNodeDef(const PXR_NS::SdrShaderNodeConstPtr& shaderNo
...
+    //      sourceType  glslfx                    mtlx              USD
+    //      category    -                         -                 -
+    //      identifier  UsdPrimvarReader_float2   ND_image_color3   PortalLight
+    //      label       -                         image             -
+    //      role        primvar                   texture           PortalLight
+
+    // UsdLux has 2 classification levels:
+    //     - Context
+    //     - SourceType
+    if (fShaderNodeDef->GetFamily().IsEmpty()) {
```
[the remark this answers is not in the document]
JGamache-autodesk: Role was always the same as identifier for UsdLux items.

lib/mayaUsd/ufe/UsdShaderNodeDef.cpp
JGamache-autodesk: How do we handle all those commands that have validation in the constructor? The create node command ends up using UsdUndoAddNewPrimCommand and will properly react if the prim can not be added.

lib/mayaUsd/ufe/UsdShaderAttributeDef.cpp
JGamache-autodesk: Requires PXR_NAMESPACE_USING_DIRECTIVE.

ppt-adsk (approved):

feldstj (approved):

PR #2316, opened 2022-04-27
MAYA-122362 cache-to-USD file type persistence

pierrebai-adsk:
- Refactor functions used to create the USD file format list.
- Provide function to retrieve different USD file format labels.
- Provide functions to get and set (load and save) the last-used USD file format used during caching.
- Make all USD dialogs use the same file formats list.

### DIFF
--- a/lib/mayaUsd/resources/scripts/mayaUsdUtils.py
+++ b/lib/mayaUsd/resources/scripts/mayaUsdUtils.py
@@ -84,17 +84,16 @@ def getMonoFormatFileFilterLabels(includeCompressed = True):
     to be join with ';;'. Call getUSDDialogFileFilters() instead as it does
     the joining.
     """
+    labelAndFilters = [
+        ("kUsdASCIIFiles", "*.usda"),
+        ("kUsdBinaryFiles", "*.usdc"),
+    ]
+
     if includeCompressed:
-        labelAndFilters = [
-            ("kUsdASCIIFiles", "*.usda"),
-            ("kUsdBinaryFiles", "*.usdc"),
-            ("kUsdCompressedFiles", "*.usdz"),
-        ]
-    else:
-        labelAndFilters = [
-            ("kUsdASCIIFiles", "*.usda"),
-            ("kUsdBinaryFiles", "*.usdc"),
-        ]
+        labelAndFilters.append(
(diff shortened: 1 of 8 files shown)


pierrebai-adsk:
@seando-adsk this PR might have gone under the radar?

lib/mayaUsd/resources/scripts/mayaUsdUtils.py
```diff
@@ -71,3 +75,80 @@ def isPulledMayaReference(dagPath):
...
+        labelAndFilters = [
+            ("kUsdASCIIFiles", "*.usda"),
+            ("kUsdBinaryFiles", "*.usdc"),
+            ("kUsdCompressedFiles", "*.usdz"),
+        ]
+    else:
+        labelAndFilters = [
+            ("kUsdASCIIFiles", "*.usda"),
+            ("kUsdBinaryFiles", "*.usdc"),
+        ]
```
seando-adsk, line 97: To avoid duplication here couldn't it initialize the labelAndFilters to usda/usdc and then if includeCompressed add the usdz?

lib/mayaUsd/resources/scripts/mayaUsdUtils.py
```diff
@@ -71,3 +75,80 @@ def isPulledMayaReference(dagPath):
...
+    if includeCompressed:
+        labelAndFilters = [
+            ("kAllUsdFiles", "(*.usd *.usda *.usdc *.usdz)"),
+            ("kUsdFiles", "*.usd"),
+        ]
+    else:
+        labelAndFilters = [
+            ("kAllUsdFiles", "(*.usd *.usda *.usdc)"),
+            ("kUsdFiles", "*.usd"),
+        ]
```
seando-adsk, line 118: Could we also do something to avoid duplication here? 
```suggestion
        labelAndFilters = [
            ("kAllUsdFiles", "(*.usd *.usda *.usdc{compressed})".format(compressed = " *.usdz" if includeCompressed else '')),
            ("kUsdFiles", "*.usd"),
        ]
```

seando-adsk (changes requested):
Looks good, just a couple of minor comments

seando-adsk (approved):

PR #2317, opened 2022-04-27
Can no longer pass ALL_INSTANCES into GetScenePrimPath.

williamkrick:


gracekumagai:
Thanks for the fix @williamkrick! I quickly tested creating a stage with the "kitchen_set_instanced" with the latest USD dev with your PR and it no longer crashes! Thanks!

dj-mcg (approved):
This looks good to me, thanks for the quick turnaround on this, and thanks for testing, Grace!

PR #2318, opened 2022-04-28
[al] Added support for importing animated bool and int type attributes

csyshing:
We have the function `DgNodeHelper::setFloatAttrAnim`() for float type attributes but missing similar functions for `int` and `bool` types, this PR is for that case.


ppt-adsk (approved):
Would have been nice to have some kind of automated test for this functionality, but I trust that you're measuring quality internally at AL in some form.

csyshing:
> Would have been nice to have some kind of automated test for this functionality, but I trust that you're measuring quality internally at AL in some form.

Yes, we do have unit tests on the client side that using these functions (not part of the AL_USDMaya code base).

PR #2319, opened 2022-04-28
Implement pivot accessors for wrapped matrix ops.

ppt-adsk:
Setting the rotate or scale pivot on a prim with a matrix (a.k.a. transform) transform op immediately creates a Maya fallback matrix stack, because pivot values are not supported by the single matrix data model.  When this occurs, a UsdTransform3dSetObjectMatrix wrapper Ufe::Transform3d interface is created to wrap the Maya fallback matrix stack.  This wrapper did not forward the pivot accessors to the wrapped Maya fallback stack, now fixed.

### DIFF
--- a/lib/mayaUsd/ufe/UsdTransform3dSetObjectMatrix.cpp
+++ b/lib/mayaUsd/ufe/UsdTransform3dSetObjectMatrix.cpp
@@ -64,6 +64,20 @@ Ufe::Vector3d UsdTransform3dSetObjectMatrix::scale() const
     return getScale(UsdTransform3dBase::matrix());
 }
 
+Ufe::Vector3d UsdTransform3dSetObjectMatrix::rotatePivot() const { return _wrapped->rotatePivot(); }
+
+Ufe::Vector3d UsdTransform3dSetObjectMatrix::scalePivot() const { return _wrapped->scalePivot(); }
+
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/ufe/UsdTransform3dSetObjectMatrix.cpp
```diff
@@ -64,6 +64,20 @@ Ufe::Vector3d UsdTransform3dSetObjectMatrix::scale() const
     return getScale(UsdTransform3dBase::matrix());
 }
 
+Ufe::Vector3d UsdTransform3dSetObjectMatrix::rotatePivot() const { return _wrapped->rotatePivot(); }
```
ppt-adsk, line 67: Forward pivot accessors to the wrapped Transform3d.

test/lib/ufe/testRotatePivot.py
```diff
@@ -228,6 +232,37 @@ def testRotatePivotCmd(self):
             sphereMatrix = om.MMatrix(t3d.inclusiveMatrix().matrix)
             self.checkPos(sphereMatrix, [20, 10, 0])
 
+    def testRotatePivotMatrixOp(self):
```
ppt-adsk, line 235: Test driven development: test written first, failed, implemented fix, test passes.

PR #2322, opened 2022-05-02
MAYA-122693 Use getMemberPaths for material export when it is available.

williamkrick:
It correctly finds all the materials.


vlasovi (approved):
LGTM

PR #2324, opened 2022-05-03
Fixed up shading mode export options and display colour fallback behaviour

J-Mo63:
The ADSK plugin currently provides a `-shadingMode` export option that can be used to control how shading data is exported, the docs of which read as follows:

| Long flag                        | Short flag | Type             | Default             | Description |
| -------------------------------- | ---------- | ---------------- | ------------------- | ----------- |
| `-shadingMode`                   | `-shd`     | string           | `displayColor`      | Set the shading schema to use. Valid values are: `none`: export no shading data to the USD, `displayColor`: unless there is a colorset named `displayColor` on a Mesh, export the diffuse color of its bound shader as `displayColor` primvar on the USD Mesh, `pxrRis`: export the authored Maya shading networks, applying the same translations applied by RenderMan for Maya to the shader types, `useRegistry`: Use a registry based to export the Maya shading network to an equivalent UsdShade network. |

It seems that while the default here is listed as `displayColor`, the default set in `UsdMayaJobExportArgs` is `none` (line 580), and the one used in its `GetDefaultDictionary` method is `useRegistry`. Unsure as to which was the originally intended, I've set it to `useRegistry` (though let me know if one of the others is more desirable).

In addition, currently using the `none` option will still result in the bound shader's result being baked to `primvars:displayColor` if there is no "displayColor" colour set currently authored. I've added a check to the mesh writer to prevent this behaviour unless the option is enabled.


seando-adsk:
@J-Mo63 We had an internal discussion on this PR. 
1. Setting default to useRegistry -> **very good**.
2. Regarding the mesh code change. There should be only one way to select display color, and it is the **-exportDisplayColor** boolean switch. Having the shading mode as either confirmation, or alternate way to select (change the && to ||) only adds confusion. It would be preferable to remove it from the documentation instead. Also, since this mode can co-exist with the other ones, keeping the boolean switch makes more sense since it can be activated in parallel with useRegistry modes like UsdPreviewSurface export.

J-Mo63:
> 2. Regarding the mesh code change. There should be only one way to select display color, and it is the **-exportDisplayColor** boolean switch. Having the shading mode as either confirmation, or alternate way to select (change the && to ||) only adds confusion. It would be preferable to remove it from the documentation instead. Also, since this mode can co-exist with the other ones, keeping the boolean switch makes more sense since it can be activated in parallel with useRegistry modes like UsdPreviewSurface export.

So the issue here is that we want to turn off the default baking of the shader data to the display colour. In reality, the option we're looking for is like `-doNotBakeDisplayColorFromShaderByDefault` (or something a bit more catchy) as `-exportDisplayColor` doesn't entirely describe the effects it has on the resulting export. Baking colour sets from the shader is usually something artists will do explicitly if necessary on our end. It was quite a while since I wrote up this PR so I think there were grander plans with what to do, but for the sake of expedience we can live with `-exportDisplayColor` for now and deal with it in a later PR. I'll remove everything expect the change to the export jobs/docs given that it seems we're on board with that as the default.

JGamache-autodesk (approved):

PR #2325, opened 2022-05-03
Fixed NURBS curve extents calculation to include widths

J-Mo63:
This PR aims to continue the alignment between the AL and ADSK plugins' export functionality, updating the AL plugin's NURBS curve export to perform the proper calculations on curve extents (authored as `float3[] extent`) by factoring in curve widths using the `UsdGeomCurves` implementation of `ComputeExtent` in lieu of `UsdGeomPointBased` (as it is in the ADSK plugin).

### DIFF
--- a/plugin/al/translators/NurbsCurve.cpp
+++ b/plugin/al/translators/NurbsCurve.cpp
@@ -143,7 +143,7 @@ UsdPrim NurbsCurve::exportObject(
     }
 
     if (params.getBool(GeometryExportOptions::kMeshExtents)) {
-        AL::usdmaya::utils::copyExtent(fnCurve, nurbs.GetExtentAttr(), params.m_timeCode);
+        AL::usdmaya::utils::copyExtent(fnCurve, nurbs, params.m_timeCode);
     }
 
     return nurbs.GetPrim();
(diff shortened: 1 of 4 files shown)


J-Mo63:
Hey @seando-adsk are we able to get feedback on this current set of changes some time today (#2325, #2326, and #2324)? I apologise for the late notice; these are extenuating circumstances.

J-Mo63:
Thanks @seando-adsk, appreciate it greatly!

plugin/al/translators/NurbsCurve.cpp
```diff
@@ -293,6 +272,28 @@ void NurbsCurve::writeEdits(
             nurbsCurvesPrim.GetWidthsAttr().Set(dataWidths);
         }
     }
+    if (diff_curves & AL::usdmaya::utils::kCurvePoints) {
```
pierrebai-adsk, line 275: Is there a reason why the block of code moved? it makes comparison harder.

pierrebai-adsk: Ah, the change requires the widths to be calculated

pierrebai-adsk (approved):

PR #2326, opened 2022-05-03
Updated NURBS curve export for widths and knots attributes

J-Mo63:
This PR aims to continue the alignment between the AL and ADSK plugins' export functionality, updating the ADSK plugin's NURBS curve export of `knots` to avoid adding extra knots in non-wrapping curves (i.e. do not copy the first and last knots in the exported vector). The PR also adds a simple implementation for the `widths` attribute as lifted from the AL plugin's NURBS curve export translator in favour of just authoring a constant `widths` vector of `[1]`.

### DIFF
--- a/lib/usd/translators/nurbsCurveWriter.cpp
+++ b/lib/usd/translators/nurbsCurveWriter.cpp
@@ -32,6 +32,9 @@
 
 #include <maya/MDoubleArray.h>
 #include <maya/MFnDependencyNode.h>
+#include <maya/MFnDoubleArrayData.h>
+#include <maya/MFnFloatArrayData.h>
+#include <maya/MFnNumericAttribute.h>
 #include <maya/MFnNurbsCurve.h>
 #include <maya/MPointArray.h>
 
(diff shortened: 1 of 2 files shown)


J-Mo63:
Looks like it's building and passing tests. macOS failing with:
```
/lib/usd/translators/nurbsCurveWriter.cpp:205:53
error: lambda capture 'mayaKnotsCount' is not used [-Werror,-Wunused-lambda-capture]
        auto copyKnots = [&mayaCurveKnots, &curveKnots, mayaKnotsCount](size_t from, size_t to) {
```

Seems that Clang is just being fussy about warnings. I'll have a fix up in the morning (Sydney time), but should be reviewable.

lib/usd/translators/nurbsCurveWriter.cpp
```diff
@@ -159,18 +200,26 @@ bool PxrUsdTranslators_NurbsCurveWriter::writeNurbsCurveAttrs(
...
+    auto           copyKnots = [&mayaCurveKnots, &curveKnots](size_t from, size_t to) {
+        memcpy(
+            ((double*)curveKnots.cdata()) + from,
+            (const double*)&mayaCurveKnots[0],
+            sizeof(double) * to);
+    };
     if (wrap) {
+        // Insert wrapping knots at either end of the vector
+        curveKnots.resize(mayaKnotsCount + 2);
+        copyKnots(1, mayaKnotsCount + 1);
```
williamkrick, line 214: I don't understand why this call to copyKnots needs to copy mayaKnotsCount + 1 knots. Doesn't that read beyond the end of the data in mayaCurveKnots?

The code does immediately overwrite whatever got copied into curveKnots, but reading that extra data is still incorrect.

If my interpretation is correct the copyKnots lambda doesn't need the to parameter, it could capture mayaKnotsCount because it should always copy the same amount of data.

williamkrick (changes requested):

lib/usd/translators/nurbsCurveWriter.cpp
J-Mo63: Sorry @williamkrick you are quite correct, must have mistaken that third param for an index. I'll fix that right up.

williamkrick (approved):

michaeljefferies:
The comment at the top says "updating the ADSK plugin's NURBS curve export of knots to avoid adding extra knots in non-wrapping curves (i.e. do not copy the first and last knots in the exported vector)", but according to the USD documentation there should be extra knots for both periodic and non-periodic curves: https://graphics.pixar.com/usd/dev/api/class_usd_geom_nurbs_curves.html#details

PR #2331, opened 2022-05-05
MAYA-122541 fix hydra crash with skydome

pierrebai-adsk:
Some resources used by the Hydra renderer are cached in the Hydra engine task context. The shutdown of the Maya-to-Hydra render override destroyed many Hydra-related objects, but the Hydra engine itself is kept by value and is never destroyed. We ended-up having data in teh task context that still referred to deleted objects and that led to a crsh if the Hydra render override was initialized and used again.

Arguably, the bug is in Hydra itself, where the texture handles (HdStTextureHandle) keep a raw pointer to the texture registry (HdSt_TextureHandleRegistry) instead of a smart pointer. That raw pointer is taken from the HdRenderIndex, which does keep its pointer to the registry as a shared_ptr.

OTOH, semi-shutting-down Hydra like our render override does is probably not correct. These low-level interactions and handling of shutdown and restart are rarely well documented.

In passing, I also fixed a bug in the UFE helper that assumed there always was at least one segment in teh given UFE path, which tured out to be wrong in some cases.

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
+++ b/lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
@@ -830,6 +830,15 @@ void MtohRenderOverride::ClearHydraResources()
     _delegates.clear();
     _defaultLightDelegate.reset();
 
+    // Cleanup internal context data that keep references to data that is now
+    // invalid.
+#if PXR_VERSION >= 2108
+    _engine.ClearTaskContextData();
+#else
+    for (const auto& token : HdxTokens->allTokens)
+        _engine.RemoveTaskContextData(token);
+#endif
+
     if (_taskController != nullptr) {
         delete _taskController;
         _taskController = nullptr;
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/render/mayaToHydra/renderOverride.cpp
```diff
@@ -830,6 +830,15 @@ void MtohRenderOverride::ClearHydraResources()
     _delegates.clear();
     _defaultLightDelegate.reset();
 
+    // Cleanup internal context data that keep references to data that is now
+    // invalid.
+#if PXR_VERSION >= 2108
+    _engine.ClearTaskContextData();
+#else
+    for (const auto& token : HdxTokens->allTokens)
```
pierrebai-adsk, line 838: Pre-21.08, there was no clear function and there is no way to iterate on the context data... so I just clear some of the potentially cached tokens, at least the ones that were causing this particular crash, related to the light and shadow system.

pierrebai-adsk:
@williamkrick fix for crash in mayaUsd

williamkrick (approved):

PR #2334, opened 2022-05-06
MAYA-123013 obey file-format choice in layer editor

pierrebai-adsk:
When re-saving a modified layer, follow the current ASCII / Binary format selected in the layer editor.

Notes:
- USD SdfLayer does not have a Save() function taking the desired format. The only option we have is to use Export().
- The choice of binary / ASCII is in a sub-menu and is global to all layers and all stages. That may not be convenient for the user: changing it will affect all stages and all layers. Beware!

### DIFF
--- a/lib/mayaUsd/utils/utilSerialization.cpp
+++ b/lib/mayaUsd/utils/utilSerialization.cpp
@@ -184,6 +184,20 @@ void setNewProxyPath(const MString& proxyNodeName, const MString& newValue)
         /*undo*/ false);
 }
 
+bool saveLayerWithFormat(
+    SdfLayerRefPtr     layer,
+    const std::string& potentialFilePath,
+    std::string        formatArg)
+{
+    const std::string& filePath
+        = potentialFilePath.empty() ? layer->GetRealPath() : potentialFilePath;
+
+    PXR_NS::SdfFileFormat::FileFormatArguments args;
+    args["format"] = formatArg.empty() ? usdFormatArgOption() : formatArg;
+
+    return layer->Export(filePath, "", args);
+}
+
 SdfLayerRefPtr saveAnonymousLayer(
     SdfLayerRefPtr     anonLayer,
     LayerParent        parent,
@@ -204,14 +218,7 @@ SdfLayerRefPtr saveAnonymousLayer(
         return nullptr;
     }
 
-    PXR_NS::SdfFileFormat::FileFormatArguments args;
-    if (!formatArg.empty()) {
-        args["format"] = formatArg;
-    } else {
-        args["format"] = usdFormatArgOption();
-    }
-
-    anonLayer->Export(path, "", args);
+    saveLayerWithFormat(anonLayer, path, formatArg);
 
     SdfLayerRefPtr newLayer = SdfLayer::FindOrOpen(path);
     if (newLayer) {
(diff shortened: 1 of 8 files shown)


lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -351,7 +351,9 @@ void LayerTreeItem::saveEditsNoPrompt()
         if (!isSessionLayer())
             saveAnonymousLayer();
     } else {
-        layer()->Save();
+        PXR_NS::SdfFileFormat::FileFormatArguments args;
+        args["format"] = MayaUsd::utils::usdFormatArgOption();
+        layer()->Export(layer()->GetRealPath(), "", args);
```
seando-adsk, line 356: Seems okay to me, but I'm wondering what are the implications of calling Export() vs Save()? @ppt-adsk Do you have any comments about this?

pierrebai-adsk: We use Export in many other place to save when we want to control the output format.

ppt-adsk: As per the docs, Save() saves to the previously-remembered file name, and fails if there is none.  Export() saves a copy of the layer, and does not affect the previously-remembered file name.  There is one important point: Save() will clear the layer's dirty state (i.e. SdfLayer::IsDirty() will return false), whereas Export() will not, so this change is introducing an important difference.

pierrebai-adsk: Export() calls _WriteToFile() which calls _MarkCurrentStateAsClean() if the filename did not change, which is our case since we're exporting to the same file (we pass GetRelPath()).

The differences I can see are:
- Save() does nothing if the layer is not dirty. We could add a call to IsDirty().
- Set modification time, which I don't think we care about.

Things that look different but are the same in the end:
- Save() check IsMuted() and IsAnonymous() directly, but Export() ends-up doing the same check because _WriteToFile() calls PermissionToSave(0 when the filename is the same and that does the same checks.
- Save(0 clears hints, but the hints are not used once the dirty state is reset.

pierrebai-adsk: ... unfortunately, there are two other places in the code where we call Save() directly on the layer... so they will not be aware of the changed file format. I'm taking a deeper look but we will either have to modify those too or abandon the idea of switching file formats for already-saved layers.

ppt-adsk (dismissed):

lib/mayaUsd/utils/utilSerialization.cpp
```diff
@@ -184,6 +184,20 @@ void setNewProxyPath(const MString& proxyNodeName, const MString& newValue)
...
+    const std::string& potentialFilePath,
+    std::string        formatArg)
+{
+    const std::string& filePath
+        = potentialFilePath.empty() ? layer->GetRealPath() : potentialFilePath;
+
+    PXR_NS::SdfFileFormat::FileFormatArguments args;
+    args["format"] = formatArg.empty() ? usdFormatArgOption() : formatArg;
+
+    return layer->Export(filePath, "", args);
```
seando-adsk, line 198: Is there a way here we can check the current format of the filePath and if it hasn't changed we just call Save()?

pierrebai-adsk: Done, but it was trickier than I anticipated.

seando-adsk (approved):
Great work, thanks.

PR #2350, opened 2022-05-11
T bailp/maya 115146/maya ref naming 2

pierrebai-adsk:
Based on Yves Boudreault's branch.

Simplified it a bit after merging, fixed issues with reconnecting to the wrong node.

From Yves's original code I kept:

- Add an environment variable MAYAUSD_ENABLE_MAYA_REFERENCE_OLD_BEHAVIOUR to use the old behaviour in AL.
- Keeps the Maya Ref node name in a custom attribute.
- Requires that the attribute be directly on the prim, not in a lower layer or indirect reference.
- This allows having a different Maya Ref Node for each prim even if they are built from the same referenced file and namespace.
- Adds more unit tests.

Improve the code to avoid having two Maya Ref that happen to refer to the same reference file use each other's Maya Ref Node by accident.

The subtle change is that we no longer search for the corresponding Maya Ref Node if the custom attribute is not present. When the attribute is not present, it means the Maya Ref was never edited and thus need to create its own Maya Ref Node, not search for an existing one.

I also refactored the code to avoid some code duplication and push out some code into their own functions for clarity.

- Make all ref-node-name generator always generate the full name, including any RN suffix.
- Renamed isMayaReference() to useNewMayaRefNaming() for clarity since that was the function's purpose.
- Added setMayaRefCustomAttribute() and getMayaRefCustomAttribute() for clarity, isolation and reuse.
- Got rid of getRefNodeNameFromPrim() since its only usage was to try to find a Maya Ref Node when the custom attribute was absent, which was incorrect and led to accidentally use the node from other Maya Ref.
- De-duplicate the Maya ref node searching code.
- Fixed the new test that simulated reparent a prim but did not simulate that the custom attribute would follow.


ppt-adsk (commented):
Nit-picks only, thanks for doing this!

ppt-adsk (approved):

ppt-adsk (approved):

pierrebai-adsk:
Given  that this is pretty much equivalent to what Yves had written and it was approved, I'm going to label this as ready for merge.

PR #2218, opened 2022-03-28
[al] Added export option to merge offset parent matrix

csyshing:
This PR added consideration of offset parent matrix being part of the 'local space matrix' of a Maya node, since it's only needed for anim cache, note that there is no round-tripping support to/from Maya <-> USD.

PR #2354, opened 2022-05-16
[al] Fixed incorrect computing local transformation matrix when exporting to USD

csyshing:
This PR is a follow up of https://github.com/Autodesk/maya-usd/pull/2218 which:
 - Fixed the computation of local transformation matrix (Maya is post-multiplied)
 - Added verification in unit test to check if the exported matrix can match Maya


seando-adsk:
@ppt-adsk Can you do a quick code review on this AL change.

ppt-adsk (approved):

PR #2355, opened 2022-05-16
[al] Fixed issue that resync prim would cause irrelevant translators being removed

csyshing:
USD hierarchy might become invalid after resync (managed by translators that are not in AL_USDMaya's scope), this PR fixed the issue that irrelevant translators were removed on resync, if the registered prim is not a child of requested resync path, keep the translator and do nothing (leaving it handled by its own translator).

### DIFF
--- a/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.h
+++ b/plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.h
@@ -261,8 +261,9 @@ public:
     /// \brief  this method is used after a variant switch to check to see if the prim types have
     /// changed in the
     ///         stage, and will update the internal state accordingly.
+    /// \param affectedPaths paths that being affected during translation
     AL_USDMAYA_PUBLIC
-    void updatePrimTypes();
+    void updatePrimTypes(const SdfPathSet& affectedPaths = SdfPathSet());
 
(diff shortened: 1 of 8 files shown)


seando-adsk:
@pierrebai-adsk Can you do a quick code review on this AL change.

plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.h
```diff
@@ -261,8 +261,9 @@ struct TranslatorContext
     /// \brief  this method is used after a variant switch to check to see if the prim types have
     /// changed in the
     ///         stage, and will update the internal state accordingly.
+    /// \param affectedPaths paths that being affected during translation
```
pierrebai-adsk, line 264: Small nit-pick: "that being affected" -> "that are being affected" or more simply "affected". The same typo is present in other modified header files.

pierrebai-adsk (dismissed):
LGTM, only that small comment typo.

plugin/al/lib/AL_USDMaya/AL/usdmaya/fileio/translators/TranslatorContext.h
csyshing: Should have been fixed in commit 939a550.

pierrebai-adsk (approved):

PR #2356, opened 2022-05-16
MAYA-122391 do not offer USDZ as a file format to cache to USD

pierrebai-adsk:
Also fix a few bugs:
- The file filters were not declared with teh correct syntax.
- A MEL file had a bad comment.


seando-adsk (dismissed):
Thanks for fixing that error with the comment in MEL file. I don't know how that got in.

pierrebai-adsk:
Sorry, I found a spurious error due to the load rules saving that I think is trivial to include at teh same time.

seando-adsk (approved):

PR #2358, opened 2022-05-16
MAYA-123016 catch exceptions during notifications

pierrebai-adsk:
Exception can be thrown by the notified objects or function. Do not let it out as it will crash Maya as nothing else above catches the exceptions.

### DIFF
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -68,6 +68,20 @@ bool isTransformChange(const TfToken& nameToken)
     return nameToken == UsdGeomTokens->xformOpOrder || UsdGeomXformOp::IsXformOp(nameToken);
 }
 
+// Prevent exception from the notifications from escaping and breaking USD/Maya.
+// USD does not wrap its notification in try/catch, so we need to do it ourselves.
+template <class RECEIVER> struct SafeNotification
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -68,6 +68,20 @@ bool isTransformChange(const TfToken& nameToken)
     return nameToken == UsdGeomTokens->xformOpOrder || UsdGeomXformOp::IsXformOp(nameToken);
 }
 
+// Prevent exception from the notifications from escaping and breaking USD/Maya.
+// USD does not wrap its notification in try/catch, so we need to do it ourselves.
+template <class RECEIVER> struct SafeNotification
```
ppt-adsk, line 73: Can be just a single templated function, doesn't need to be static, we're in the unnamed namespace already:
template<class R, class N> void notify(const N& n) {
    // try / catch stuff
    R::notify(n);
}
You can then explicitly specify a single template argument and the second template argument (the type of the function argument) will be deduced by the compiler.

pierrebai-adsk: Ah good, I thought that if we provided one template type, we had to provide both, that is why I use this trick to split them.

ppt-adsk (approved):

PR #2359, opened 2022-05-17
MAYA-122146 - Selecting a camera and framing it with "f" shows the origin instead of the camera

vlasovi:
Add Maya-specific camera extents to the USD UFE bounding box calculation


vlasovi:
The failed test in preflight is "testUsdExportPackage". It has nothing to do with my changes. This is just some permission problem on the build machine. The PR can be merged.

PR #2363, opened 2022-05-18
MAYA-123332 allow topological modifications

pierrebai-adsk:
When locking nodes, it also locks topological modifications. Explicitly re-enable it afterward.

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -731,8 +731,7 @@ bool pushCustomize(
         auto updater = createUpdater(srcLayer, srcPath, dstPath, context);
         if (!updater) {
             TF_WARN(
-                "Could not create a prim updater for path %s during PushEnd() traversal, "
-                "pruning "
+                "Could not create a prim updater for path %s during PushEnd() traversal, pruning "
                 "at that point.",
                 srcPath.GetText());
             return;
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -280,6 +280,25 @@ bool removeExcludeFromRendering(const Ufe::Path& ufePulledPath)
...
 
+//------------------------------------------------------------------------------
+//
+// Turn on the mesh flag to allow topological modifications.
+bool allowTopologyModifications(MDagPath& root)
+{
+    MDGModifier& dgMod = MDGModifierUndoItem::create("Allow topology modifications");
+
+    MItDag dagIt;
+    dagIt.reset(root, MItDag::kDepthFirst, MFn::kMesh);
```
ppt-adsk, line 291: Seems like a shame that LockNodesUndoItem does a traversal, then we do another traversal here.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -280,6 +280,25 @@ bool removeExcludeFromRendering(const Ufe::Path& ufePulledPath)
...
+    for (; !dagIt.isDone(); dagIt.next()) {
+        MString cmd;
+        if (!cmd.format("setAttr \"^1s.allowTopologyMod\" 1;", dagIt.fullPathName()))
```
ppt-adsk, line 294: Surely we can use MDGModifier::newPlugValueBool()?  Starting up the MEL interpreter to set a plug from C++ feels like a bit much.

pierrebai-adsk: I had looked to MDGModifier, but did not see setAttribute. I'll try your suggestion.

ppt-adsk (approved):

PR #2377, opened 2022-05-25
MAYA-122926 reset auto-edit in all variants

pierrebai-adsk:
- Add a utility function to apply a function to all variants of a prim.
- Add a utility class to automatically restore the current variant on exit.
- Use these to clear the auto-edit flag in all variants.
- Remove obsolete comments about type info not being correctly preserved and psuhEnd not being called (now it is called).

### DIFF
--- a/lib/usd/translators/mayaReferenceUpdater.cpp
+++ b/lib/usd/translators/mayaReferenceUpdater.cpp
@@ -25,6 +25,7 @@
 #include <mayaUsd/utils/editRouter.h>
 #include <mayaUsd/utils/util.h>
 #include <mayaUsd/utils/utilSerialization.h>
+#include <mayaUsd/utils/variants.h>
 #include <mayaUsdUtils/MergePrims.h>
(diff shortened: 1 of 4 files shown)


lib/usd/translators/mayaReferenceUpdater.cpp
```diff
@@ -48,14 +49,21 @@
...
 namespace {
 
+// Clear the auto-edit flag on a USD Maya Reference so that it does not
+// get edited immediately again. Clear in all variants, since each
+// variant has its own copy of the flag.
 void clearAutoEdit(const UsdPrim& prim)
 {
-    if (prim.IsValid()) {
+    UsdPrim parentPrim = prim.GetParent();
+    MAYAUSD_NS::applyToAllVariants(parentPrim, true, [prim]() {
```
seando-adsk, line 58: We normally just use the actual namespace name instead of the define. So just MayaUsd::

seando-adsk (approved):

PR #2380, opened 2022-05-25
Adding Maya 2022 Performance baselines for Pixar draw tests

dj-mcg:


dj-mcg:
Oh actually it looks like our baselines are not pushed - sorry about that!

PR #2385, opened 2022-05-27
Respect prim writer ShouldPruneChildren when shading mode is useRegistry

sshklifov:
Currently, there is no way to register a shader writer that handles all
connected plugs to the shader. For example, the writer might handle file
nodes connected to the shader, but useRegistry shading mode will invoke
FileTextureWriter unconditionally. It will write out a UsdUVTexture
shader, which useRegistry won't be able to connect to the shading network.

### DIFF
--- a/lib/mayaUsd/fileio/shaderWriterRegistry.cpp
+++ b/lib/mayaUsd/fileio/shaderWriterRegistry.cpp
@@ -18,6 +18,7 @@
 #include <mayaUsd/base/debugCodes.h>
 #include <mayaUsd/fileio/functorPrimWriter.h>
 #include <mayaUsd/fileio/registryHelper.h>
+#include <mayaUsd/fileio/shading/shadingModeRegistry.h>
 
 #include <pxr/base/tf/debug.h>
 #include <pxr/base/tf/diagnostic.h>
@@ -58,6 +59,10 @@ _Registry::const_iterator _Find(const TfToken& usdInfoId, const UsdMayaJobExport
 {
     using ContextSupport = UsdMayaShaderWriter::ContextSupport;
 
+    TfToken conversion = exportArgs.convertMaterialsTo;
+    const bool noFallback =
+        UsdMayaShadingModeRegistry::GetMaterialConversionInfo(conversion).noFallbackExport;
+
     _Registry::const_iterator ret = _reg.cend();
     _Registry::const_iterator first, last;
     std::tie(first, last) = _reg.equal_range(usdInfoId);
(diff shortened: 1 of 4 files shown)


seando-adsk:
@sshklifov Thank you for the PR. In order to proceed we will need a signed CLA from you. Please see [Contributing](https://github.com/Autodesk/maya-usd/blob/dev/doc/CONTRIBUTING.md)

sshklifov:
Hi @seando-adsk, sorry for the inconvenience. The CLA has been signed and sent to the appropriate destination.

seando-adsk:
@sshklifov No problem. We got the signed CLA and we can now proceed with your pull-request. I've assigned it to one of our devs for code review.

JGamache-autodesk (commented):
The code itself is OK, but I suspect it will act more as a workaround than a fix.

It is perfectly OK to challenge returning ContextSupport::Fallback in PxrUsdTranslators_FileTextureWriter::CanExport() when not exporting to UsdPreviewSurface materials. I do not feel that this code is doing the right thing anymore if it forces you to add explicit pruning to fight against it.
Maybe adding a "no-fallback" flag when you register a new useRegistry conversion mode would be better?

JGamache-autodesk:
Hi @sshklifov, adding a no-fallback argument to RegisterExportConversion looks like the best solution. If you could update the PR that would be really nice. Thanks!

sshklifov:
Hi, sorry for the delay, I updated the PR. I do have a question regarding ABI though. These changes will require recompilation of plugins that link to mayaUsd (struct ConversionInfo size has changed). It is however possible to maintain ABI compatibility if ConversionInfo is not modified.

Which approach is desired? As someone who is developing a plugin that uses mayaUsd, I think that breaking the ABI in general is undesirable. In this particular instance, we will have to recompile anyway in order to see the new symbol with the noFallback flag.

JGamache-autodesk (approved):
Looks good to me. Thanks for the update!

I am not worried with new members at the end of a struct that has no virtual members. That is allowed according to the very thorough [KDE binary compatibility guidelines](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B). Changing the signatures of existing functions could be an issue though if we want to preserve binary compatibility.

I think our guidelines allow binary incompatible changes between released versions of MayaUSD so I don't think this will be an issue. I will let @seando-adsk confirm.

seando-adsk:
There is no problem with binary breaking changes. That is the reason we are major version 0.

seando-adsk:
@sshklifov I ran the preflight build for you and unfortunately it failed during the initial clang-format step. Please access the checks tab and you'll see the clang-format errors along with the fixes. Or if you have access to clang-format you can run it locally. Please see our [coding guideslines](https://github.com/Autodesk/maya-usd/blob/dev/doc/codingGuidelines.md) for info about clang-format.

JGamache-autodesk (approved):

PR #2386, opened 2022-05-27
Boudrey/maya 123462/pullset remove

boudrey:


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1156,6 +1156,28 @@ bool PrimUpdaterManager::discardPrimEdits(const Ufe::Path& pulledPath)
         return false;
     }
 
+    // Discard of the pull set if it is empty.
```
pierrebai-adsk, line 1159: Should mergeToUsd also get rid of the set? IDK why the task only complained about discard... but you did as the task specified,  but maybe ask Natalia if the set shoudl also go away when the last edited  prim is merged, too.

pierrebai-adsk (approved):

lib/mayaUsd/fileio/primUpdaterManager.cpp
boudrey: Good idea, easy to do.

pierrebai-adsk: Maybe Natalia wanted to keep the set around for some reason. I don't think so, but it would be good to confirm.

pierrebai-adsk (approved):

boudrey:
Only failing test is unrelated to my change.

PR #2387, opened 2022-05-27
MAYA-123509 fix merging of normals

pierrebai-adsk:
Normals in USD can be kept in the "normals" attribute or the "primvars:normals" attribute. This breaks the semantic-agnostic merge algorithm as it did not relate the two attributes.

In addition, the "privars:normals" has higher priority.

The result was that the round-trip of Edit-as-Maya and Merge-to-USD would create the "normals" attribute attribute and the resulting prim would have both attributes, but the old unmodified primvats would win.

This was bad when the mesh was edited as the wrong normals were used, but catastrophic when the topology was edited as the number of normals was incorrect.

The fix is to add custom logic to handle normals merging. The new strategy is to rely on the USD exportdone during merge to always author the "normals" attribute and to delete the existing "primvars:normals" from the original USD data so that it does not take priority on the newly edited normals.

While fixing this, other small fixes were done:
- the messages printed during merging were improved to be easier to understand.
- Added support for some additional data-type during USD import: byte and unsigned int were not supported.
- Added a comments about the fact that we currently do no compare metadata to determine if a prim has changed.

### DIFF
--- a/lib/mayaUsd/fileio/utils/readUtil.cpp
+++ b/lib/mayaUsd/fileio/utils/readUtil.cpp
@@ -356,6 +356,12 @@ MObject UsdMayaReadUtil::FindOrCreateMayaAttr(
     } else if (type.IsA<int>()) {
         return _FindOrCreateMayaNumericAttr(
             attrName, attrNiceName, MFnNumericData::kInt, keyable, usedAsColor, depNode, modifier);
+    } else if (type.IsA<unsigned int>()) {
+        return _FindOrCreateMayaNumericAttr(
+            attrName, attrNiceName, MFnNumericData::kInt, keyable, usedAsColor, depNode, modifier);
+    } else if (type.IsA<unsigned char>()) {
+        return _FindOrCreateMayaNumericAttr(
+            attrName, attrNiceName, MFnNumericData::kByte, keyable, usedAsColor, depNode, modifier);
     } else if (type.IsA<GfVec2i>()) {
         return _FindOrCreateMayaNumericAttr(
             attrName, attrNiceName, MFnNumericData::k2Int, keyable, usedAsColor, depNode, modifier);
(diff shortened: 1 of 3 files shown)


pierrebai-adsk:
The only PF test failure is on Linux, the known problem with surface shader naming:

> AssertionError: 'ND_standard_surface_surfaceshader_100' != 'ND_standard_surface_surfaceshader'

lib/mayaUsd/fileio/utils/readUtil.cpp
```diff
@@ -356,6 +356,12 @@ MObject UsdMayaReadUtil::FindOrCreateMayaAttr(
     } else if (type.IsA<int>()) {
         return _FindOrCreateMayaNumericAttr(
             attrName, attrNiceName, MFnNumericData::kInt, keyable, usedAsColor, depNode, modifier);
+    } else if (type.IsA<unsigned int>()) {
```
pierrebai-adsk, line 359: These fixes are unrelated to the bug, but were causing warnings in the script editor log because the attribute type was not supported.

lib/usd/utils/MergePrims.cpp
```diff
@@ -173,7 +173,8 @@ template <class T> auto makeFuncWithContext(const MergeContext& ctx, T&& func)
 bool isTransformProperty(const UsdProperty& prop)
 {
     static const TfToken xformOpOrderToken("xformOpOrder");
-    return UsdGeomXformOp::IsXformOp(prop.GetName()) || prop.GetBaseName() == xformOpOrderToken;
+    return prop.IsValid()
```
pierrebai-adsk, line 176: This fixes is also unrelated but was causing exceptions during some tests in debug builds only when MAYAUSD_SHOW_FULL_DIAGNOSTICS is on, because USD would throw exception instead of merely logging a message.

boudrey: Not sure about this one, xformOpOrderToken is not used anymore and the function is not testing if the prop is an xfrom but just returning if it is valid. Is this intended?

pierrebai-adsk:
Only PF failure is a texture image comparison failure on Linux that is unrelated.

lib/usd/utils/MergePrims.cpp
```diff
@@ -246,6 +247,46 @@ bool isLocalTransformModified(const UsdPrim& srcPrim, const UsdPrim& dstPrim)
...
+
+bool isNormalsProperty(const TfToken& name)
+{
+    // Note: tokens are interned to be unique and thus we don't want
+    //       their constructor to be global in order to avoid the
+    //       unspecified order of global construction.
+    //
+    //       That is why we declare them as static inside the functions.
+    static const TfToken normalsToken(normalsTokenName);
+    static const TfToken pvNormalsToken(pvNormalsTokenName);
```
williamkrick, line 272: Normally we do this using the TF_DEFINE_PRIVATE_TOKENS macro. You can see an example: https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/render/vp2RenderDelegate/material.cpp#L119

williamkrick (dismissed):

williamkrick (approved):

PR #1873, opened 2021-12-02
Add action in the outliner's context menu to toggle the instanceable state of a prim

gracekumagai:
This PR adds an action to the Outliner's context menu to toggle the instanceable state of a prim. 

Following the design, it provides the options to "Mark as Instanceable" if the prim is not an instance and "Unmark as Instanceable" if the prim is an instance.

PR #2388, opened 2022-05-27
Fix issues between two versions of standard_surface

JGamache-autodesk:
Using GetShaderNodeByIdentifier resolves issues where there is more than
one node sharing the same name with two different versions.

See Pixar Issue #1873 about version parsing failure in usdMtlx.


JGamache-autodesk:
Pre-flight succeeded. Logs show channel disconnected while trying to report a successful run.

seando-adsk (approved):

PR #2390, opened 2022-05-30
Allowing a new prim spec to be written in edit target layers

NickWu:
This PR is for addressing the issue discussed in here https://github.com/Autodesk/maya-usd/discussions/2212. In short, at Animal Logic, we found `isAttributeEditAllowed` function in UFE a bit confusing and was blocking our pipeline so we updated its logic in our internal code base. This PR is to promote our internal solution.

### DIFF
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -78,7 +78,7 @@ bool stringBeginsWithDigit(const std::string& inputString)
 
 // This function calculates the position index for a given layer across all
 // the site's local LayerStacks
-uint32_t findLayerIndex(const UsdPrim& prim, const SdfLayerHandle& layer, bool& foundLayer)
+uint32_t findLayerIndex(const UsdPrim& prim, const SdfLayerHandle& layer)
 {
     uint32_t position { 0 };
 
@@ -96,7 +96,6 @@ uint32_t findLayerIndex(const UsdPrim& prim, const SdfLayerHandle& layer, bool& 
         // to find the layer
         for (SdfLayerRefPtr const& l : layerStack->GetLayers()) {
             if (l == layer) {
-                foundLayer = true;
                 return position;
             }
             ++position;
@@ -452,15 +451,8 @@ bool isAttributeEditAllowed(const PXR_NS::UsdAttribute& attr, std::string* errMs
         return false;
     }
 
-    bool foundLayer = false;
-
     // get the index to edit target layer
-    const auto targetLayerIndex = findLayerIndex(prim, editTarget.GetLayer(), foundLayer);
-
-    // If the edit target layer not found in the prim's layer stack, we return true to allow
-    // new future edits go to the edit target layer.
-    if (!foundLayer)
-        return true;
+    const auto targetLayerIndex = findLayerIndex(prim, editTarget.GetLayer());
 
     // HS March 22th,2021
     // TODO: "Value Clips" are UsdStage-level feature, unknown to Pcp.So if the attribute in
@@ -478,15 +470,14 @@ bool isAttributeEditAllowed(const PXR_NS::UsdAttribute& attr, std::string* errMs
     if (!propertyStack.empty()) {
         // get the strongest layer that has the attr.
         auto strongestLayer = attr.GetPropertyStack().front()->GetLayer();
-        foundLayer = false;
 
         // compare the calculated index between the "attr" and "edit target" layers.
-        if (findLayerIndex(prim, strongestLayer, foundLayer) < targetLayerIndex) {
+        if (findLayerIndex(prim, strongestLayer) < targetLayerIndex) {
             if (errMsg) {
                 *errMsg = TfStringPrintf(
                     "Cannot edit [%s] attribute because there is a stronger opinion in [%s].",
-                    attr.GetName().GetText(),
-                    strongestLayer->GetIdentifier().c_str());
+                    attr.GetBaseName().GetText(),
+                    strongestLayer->GetDisplayName().c_str());
             }
 
             return false;
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -78,7 +78,7 @@ bool stringBeginsWithDigit(const std::string& inputString)
 
 // This function calculates the position index for a given layer across all
 // the site's local LayerStacks
-uint32_t findLayerIndex(const UsdPrim& prim, const SdfLayerHandle& layer)
+uint32_t findLayerIndex(const UsdPrim& prim, const SdfLayerHandle& layer, bool& foundLayer)
 {
     uint32_t position { 0 };
 
@@ -96,6 +96,7 @@ uint32_t findLayerIndex(const UsdPrim& prim, const SdfLayerHandle& layer)
         // to find the layer
         for (SdfLayerRefPtr const& l : layerStack->GetLayers()) {
             if (l == layer) {
+                foundLayer = true;
                 return position;
             }
             ++position;
@@ -451,8 +452,15 @@ bool isAttributeEditAllowed(const PXR_NS::UsdAttribute& attr, std::string* errMs
         return false;
     }
 
+    bool foundLayer = false;
+
     // get the index to edit target layer
-    const auto targetLayerIndex = findLayerIndex(prim, editTarget.GetLayer());
+    const auto targetLayerIndex = findLayerIndex(prim, editTarget.GetLayer(), foundLayer);
+
+    // If the edit target layer not found in the prim's layer stack, we return true to allow
+    // new future edits go to the edit target layer.
+    if (!foundLayer)
+        return true;
 
     // HS March 22th,2021
     // TODO: "Value Clips" are UsdStage-level feature, unknown to Pcp.So if the attribute in
@@ -470,14 +478,15 @@ bool isAttributeEditAllowed(const PXR_NS::UsdAttribute& attr, std::string* errMs
     if (!propertyStack.empty()) {
         // get the strongest layer that has the attr.
         auto strongestLayer = attr.GetPropertyStack().front()->GetLayer();
+        foundLayer = false;
 
         // compare the calculated index between the "attr" and "edit target" layers.
-        if (findLayerIndex(prim, strongestLayer) < targetLayerIndex) {
+        if (findLayerIndex(prim, strongestLayer, foundLayer) < targetLayerIndex) {
             if (errMsg) {
                 *errMsg = TfStringPrintf(
                     "Cannot edit [%s] attribute because there is a stronger opinion in [%s].",
-                    attr.GetBaseName().GetText(),
-                    strongestLayer->GetDisplayName().c_str());
+                    attr.GetName().GetText(),
+                    strongestLayer->GetIdentifier().c_str());
             }
 
             return false;
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -82,7 +82,7 @@ uint32_t findLayerIndex(const UsdPrim& prim, const SdfLayerHandle& layer)
 {
     uint32_t position { 0 };
 
-    const PcpPrimIndex& primIndex = prim.GetPrimIndex();
+    const PcpPrimIndex& primIndex = prim.ComputeExpandedPrimIndex();
 
     // iterate through the expanded primIndex
     for (PcpNodeRef node : primIndex.GetNodeRange()) {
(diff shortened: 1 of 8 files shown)


pierrebai-adsk:
I will have to understand well the change. We had internal discussions about loosening the restrictions, but we don't have a design yet. Getting this wrong could break reparent / renaming for example, which could leave prim stranded if they are build from opinions from different layers and a single one gets renamed.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -96,6 +96,7 @@ uint32_t findLayerIndex(const UsdPrim& prim, const SdfLayerHandle& layer)
         // to find the layer
         for (SdfLayerRefPtr const& l : layerStack->GetLayers()) {
             if (l == layer) {
```
pierrebai-adsk, line 98: I would prefer an explicit set to false at line 105 when not found, for clarity.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -451,8 +452,15 @@ bool isAttributeEditAllowed(const PXR_NS::UsdAttribute& attr, std::string* errMs
...
 
+    bool foundLayer = false;
+
     // get the index to edit target layer
-    const auto targetLayerIndex = findLayerIndex(prim, editTarget.GetLayer());
+    const auto targetLayerIndex = findLayerIndex(prim, editTarget.GetLayer(), foundLayer);
+
+    // If the edit target layer not found in the prim's layer stack, we return true to allow
+    // new future edits go to the edit target layer.
+    if (!foundLayer)
```
pierrebai-adsk, line 462: This would allow authoring below the stronger opinion. Is the goal to allow new opinion on layers without opinion only above the strongest one or anywhere? Currently for us anywhere would go against the current design, so I'll have to talk to the designer and PO/PM about it.

lib/mayaUsd/ufe/Utils.cpp
pierrebai-adsk: This would also avoid the need for the explicit reset of the variable to false in the caller, when it tries to call this a 2nd time, below.

test/lib/ufe/testAttributeBlock.py
```diff
@@ -347,6 +349,83 @@ def testAttributeBlocking3dMatrixOps(self):
 
         # authoring new transformation edit is allowed.
         self.assertTrue(mayaUsdUfe.isAttributeEditAllowed(transformAttr))
+    
+    def testIsAttributeEditAllowedFunction(self):
+        rootLayerStr = """\
```
pierrebai-adsk, line 354: Please add a function doc explaining the goal of the test.

lib/mayaUsd/ufe/Utils.cpp
pierrebai-adsk: I just recreated your layer stack and tested it... and the edition of the foo attribute on prim B is already working without your changes. So it is the case that `primIndex.GetNodeRange()` return all layers, even those without opinions.

So, I think your test is not testing what you hoped it would? You can confirm this by commenting out the early return here and run your test and see if it indeed passes without your changes.

It also means your changes might not be doing what you think it would?

To start, I would need the answer to the first question: is the goal to only allow new opinions on stronger layers or on any layer, even weaker ones?

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -470,14 +478,15 @@ bool isAttributeEditAllowed(const PXR_NS::UsdAttribute& attr, std::string* errMs
...
         // compare the calculated index between the "attr" and "edit target" layers.
-        if (findLayerIndex(prim, strongestLayer) < targetLayerIndex) {
+        if (findLayerIndex(prim, strongestLayer, foundLayer) < targetLayerIndex) {
             if (errMsg) {
                 *errMsg = TfStringPrintf(
                     "Cannot edit [%s] attribute because there is a stronger opinion in [%s].",
-                    attr.GetBaseName().GetText(),
-                    strongestLayer->GetDisplayName().c_str());
+                    attr.GetName().GetText(),
+                    strongestLayer->GetIdentifier().c_str());
```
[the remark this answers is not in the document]
NickWu: `identifier` can be an a full URI also. I updated this message because we found displayName, which is the base filename of the identifier, was not helpful enough to troubleshoot issues for us.

test/lib/ufe/testAttributeBlock.py
NickWu: Will do, thanks.

lib/mayaUsd/ufe/Utils.cpp
NickWu: > I just recreated your layer stack and tested it... and the edition of the foo attribute on prim B is already working without your changes. So it is the case that `primIndex.GetNodeRange()` return all layers, even those without opinions.
> 
> So, I think your test is not testing what you hoped it would? You can confirm this by commenting out the early return here and run your test and see if it indeed passes without your changes.
> 
> It also means your changes might not be doing what you think it would?
> 


As I found that `isAttributeEditAllowed` was getting used in many places so my intention of my test was to confirm it's not breaking the previous behaviour and at the same time allowed our pipeline continue to work as it was in maya 2020. 


> To start, I would need the answer to the first question: is the goal to only allow new opinions on stronger layers or on any layer, even weaker ones?

The goal is to allow new opinions on stronger layers and this was not possible without the change.

NickWu: > If your use case is that you want to edit in a layer that currently has no opinion but is above the layer, then you could add a check when the layer is not found: search the top opinion layer, iterating through its ancestors and if you find the target layer then allow the edit.
> 
> The code already retrieve the strongestLayer a bit below, so move that variable up and use if when foundLayer is false and check that the target is above.

That's a good idea. I'll give it a try. Thanks!

NickWu (commented):
Thanks @pierrebai-adsk for taking a look. I'll be working on addressing your comments next.

lib/mayaUsd/ufe/Utils.cpp
NickWu: > > As I found that `isAttributeEditAllowed` was getting used in many places so my intention of my test was to confirm it's not breaking the previous behaviour and at the same time allowed our pipeline continue to work as it was in maya 2020.
> > The goal is to allow new opinions on stronger layers and this was not possible without the change.
> 
> Thank you for clarifying that the intent is to allow only stronger opinions. This aligns with what we want to do.
> 
> As I said, in my own testing with the current dev branch, creating a stage with layers and prims taken from your test, I was able to edit prim B's attribute, but not prim A's. I attached a zip containing the usda files I hand-created to be equivalent to your test, when staged and targeting layer B. (The file to stage is overall.usda)
> 
> That is why I'd like you to create a test that reproduces the blocking issue you were trying to address. The test should fail when run without your changes and passes with them.
> 
> The reason I'm asking is that the problem might not have been what you thought it was. The mayaUsd plugin sometimes authors opinions in the session layer and it might have been these, stronger opinions, that were really blocking your artists. If that is the case, then it would explain why your solution unblocked them: it allowed authoring below the session layer. Of course, this is speculation, but maybe after you change your solution to check the layer stack like I suggested, you will find that it no longer work for your artists.
> 
> In the meantime, thank you for your time and efforts!
> 
> [al-2390.zip](https://github.com/Autodesk/maya-usd/files/8814194/al-2390.zip) .

Sorry for the later reply. I did a quick check on our production case. I can safely say that It's not due to session layers for this issue. In our pipeline design, we pre-stack layers by departments from strong to weak in a root layer, i.e., Light, CFX, Animation, layout, assembly, etc. We'd have many def prim specs in assembly layers to reference assets and move  them to the right places for a set. When a prim is ready to be worked on in Layout, we'd set the edit target layer to Layout layer, and there'd be no over prim specs in the Layout layer at the time. With the current logic of `isAttributeEditAllowed`, it would prevent a prim's attributes (say translate, for instance) being authored in the edit target layer (which is the Layout layer in this case) because `isAttributeEditAllowed` couldn't find the layer in the prim's layer stack (but the layer is in the stage's layer stack). When `isAttributeEditAllowed` couldn't find the given edit target layer in its prim stack, it simply gives the layer the highest index number (which is not ideal in my mind and which was what I was proposing to change) and compare the number with the index number of the strongest layer in the prim's layer stack (which in our production case, it'd be the assembly layer) and `isAttributeEditAllowed` would return false to reject the prim spec authoring. 

I think my test doesn't reflect the issue happening in our production, which is why I think it might have been misleading you to the suspicion, so I'll be trying to update the test to bring it closer to the issue we resolved in our production, thanks.

pierrebai-adsk: Thank you for the detailed explanation!

pierrebai-adsk: Hi Nick,

We've discussed your issue internally and tried to reproduce your problem without success.

First, I' d like to mention that your characterization of the code is incorrect. When the no-authoring layer index is returned, because it has an index higher than any existing authoring layer, the code will **_return true_**. The relevant code is:

`if (findLayerIndex(prim, strongestLayer) < targetLayerIndex) {`

Second, I've attached a file to my comment of a simple Maya scene with a stage with three layers. The top layer would be your layout layer and the bottom layer would be your assembly layer. The assembly layer contains a translation attribute for the cube. When targeting a higher layer tha
[multi-layer-edit.zip](https://github.com/Autodesk/maya-usd/files/8922155/multi-layer-edit.zip)
t has no authoring of the cube yet, I can translate the cube.

So, we would be interested in a simple Maya scene or bundle of USD file that reproduces the problem you are seeing to see where our own understanding of the issue is wrong.

pierrebai-adsk: After tracing into the code, I see that with the example scene I gave above, the target layer is always found by findLayerIndex even when it does not have any authored opinions. So I'd be curious to have a simple scene that reproduces the issue. The layer structure must have something I don't understand.

Maybe it is when the original data that is edited was in a reference and that somehow affects findLayerIndex? I'll try that.

NickWu: Yes, our Layout assembly layer that is for position editing is nested and referenced by an entry departmental layout layer which is a sublayer to the root layer. I think layer referencing or nested referencing might cause findLayerIndex to not be able to find it. I'll try to update my test to mimic our issue in our production.

NickWu: I found another alternative solution other than using `foundLayer` boolean, which is to replace `prim.GetPrimIndex()` with `prim.ComputeExpandedPrimIndex()` at the following line:
https://github.com/Autodesk/maya-usd/blob/3fb772da6a507716cef7f47e8611f23ea67492c5/lib/mayaUsd/ufe/Utils.cpp#L85
From the doc of `GetPrimIndex` https://graphics.pixar.com/usd/release/api/class_usd_prim.html#ae6ae4259af3ac06b8a9898a787202a32. 
It says: 

> The prim index returned by this function is optimized and may not include sites that do not contribute opinions to this prim. Use [UsdPrim::ComputeExpandedPrimIndex](https://graphics.pixar.com/usd/release/api/class_usd_prim.html#ad642cd45c3326bb0a75c209fcb0ec9eb) to compute a prim index that includes all possible sites that could contribute opinions.

So by using `ComputeExpandedPrimIndex` function instead of `GetPrimIndex`, we'll able to return an index number from `findLayerIndex` function even if the edit target layer doesn't have opinions on the target prim yet (but might contribute opinions in future). 

Which one do you prefer - using `ComputeExpandedPrimIndex` or `foundLayer` boolean?

pierrebai-adsk: Your new suggestion seems like a clean answer. I would have a slight preference for it, but either ways would be fine.

At the risk of repeating myself, my main concern is having a unit unit test that replicates the problems you are seeing that would fail without your changes and pass with your changes applied.

NickWu: Ok, I'll update the code to use `ComputeExpandedPrimIndex` instead of `foundLayer` boolean as it seems to be cleaner and less code changes.  I'll also update the unit test when I get a chance. Thanks!

NickWu: Hey @pierrebai-adsk , sorry for the late reply. I finally got some time to update this PR. I have updated the unit tests to mimic the issue we were having in our production and use `ComputeExpandedPrimIndex` function instead of `foundLayer` boolean for the solution. The test should pass with my change and fail without my change. Please have a look, thanks!

pierrebai-adsk: Hi, I was on vacations until today. I will review your changes today or tomorrow when I'm fully back up to speed.

pierrebai-adsk (approved):

PR #2393, opened 2022-05-31
MAYA-123579 fix crash when deleting a stage

pierrebai-adsk:
The loop that retrieved all the stage was calling a function that could modify the container that was iterated over. Fix this by first caching the items we want to iterate.

It only crashed in debug build due to the luck that the meory was still present and accessible in release build.

Added a unit test that crashed in debug and now works with the fix.

### DIFF
--- a/lib/mayaUsd/ufe/UsdStageMap.cpp
+++ b/lib/mayaUsd/ufe/UsdStageMap.cpp
@@ -230,9 +230,21 @@ UsdStageMap::StageSet UsdStageMap::allStages()
 {
     rebuildIfDirty();
 
-    StageSet stages;
+    // Note: stage() calls proxyShape() which can modify fPathToObject,
+    //       so we cannot call stage within a loop on fPathToObject.
+    //
+    //       So, we first cache all UFE paths and then loop over them
+    //       calling stage().
+
+    std::vector<Ufe::Path> paths;
+    paths.reserve(fPathToObject.size());
     for (const auto& pair : fPathToObject) {
-        stages.insert(stage(pair.first));
+        paths.emplace_back(pair.first);
+    }
+
+    StageSet stages;
+    for (const auto& path : paths) {
+        stages.insert(stage(path));
     }
     return stages;
 }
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/UsdStageMap.cpp
```diff
@@ -230,9 +230,21 @@ UsdStageMap::StageSet UsdStageMap::allStages()
...
+    //       So, we first cache all UFE paths and then loop over them
+    //       calling stage().
+
+    std::vector<Ufe::Path> paths;
+    paths.reserve(fPathToObject.size());
     for (const auto& pair : fPathToObject) {
-        stages.insert(stage(pair.first));
+        paths.emplace_back(pair.first);
+    }
+
```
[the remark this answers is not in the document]
pierrebai-adsk: The source container is a std::map, so there are no copy constructor for its keys.

seando-adsk (dismissed):

pierrebai-adsk:
The only PF failure is on Maya 2020 Windows platform, a single test, testUsdImportDisplacement timed out, which is unrelated to the change.

seando-adsk (approved):

lib/mayaUsd/ufe/UsdStageMap.cpp
```diff
@@ -230,9 +230,21 @@ UsdStageMap::StageSet UsdStageMap::allStages()
...
+    StageSet stages;
+    for (const auto& path : paths) {
+        stages.insert(stage(path));
```
JGamache-autodesk, line 247: I think this code will append an empty/invalid Stage to the set when processing a deleted stage (but only once since all deleted stages will return the same hash). Would it be better to test that the stage is valid before adding it to the StageSet? The function is called `allStages`.

pierrebai-adsk: I had the same reflection, but the old code did not check the validity either. The stage map holds weak pointers, so the validity can change at any moment anyway...

PR #2397, opened 2022-06-03
MAYA-123462 usdEditAsMaya set doesn't disappear when cancelling editi…

boudrey:
…ng as Maya

Fix crash related to unsupported !invert!
+ remove duplicated declaration of Bool


lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -120,9 +132,11 @@ Ufe::Attribute::Ptr UsdAttributes::attribute(const std::string& name)
           };
 
 #undef ADD_UFE_USD_CTOR
-    auto ctorIt = ctorMap.find(newAttrType);
```
boudrey, line 123: Avoid crash when type is not found like it was before.

JGamache-autodesk (approved):
Looks good to me.

JGamache-autodesk (approved):

PR #2398, opened 2022-06-03
MAYA-123122 remember shared/unshared load state

pierrebai-adsk:
- Move and refactor code in proxy shape to avoid code duplication.
- (There was duplication between shared and unshared case.)
- Now unshared stage also applied the load state.
- Add functions to the proxy shape load rule helper class to trigger saving load rules.
- Save load rules when modifying the loaded state of prims.
- Refactor the load state commands to share mode code.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -693,6 +693,20 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
 
     bool isIncomingStage = false;
 
+    // Compute the laod set for the stage.
+    MDataHandle loadPayloadsHandle = dataBlock.inputValue(loadPayloadsAttr, &retValue);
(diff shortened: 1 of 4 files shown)


pierrebai-adsk:
The only PF failure is the OSX compiler crashing during the compilation of mayaShaderTranslation.cpp... only for 2022 python 3. Probably an unrelated random crash.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -693,6 +693,20 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
 
     bool isIncomingStage = false;
 
+    // Compute the laod set for the stage.
```
ppt-adsk, line 696: "load set"

ppt-adsk (commented):
Any way we can write a test for this?

ppt-adsk (dismissed):
Thanks for the test!

pierrebai-adsk:
Only PF failure is the known Linux interactive image comparison test problem.

Issue #1739, opened 2021-10-04
MayaUSDExportCommand - parentScope and exportRoots flags together raise error


bentoogood:
**Describe the bug**
We're unable to use both the _exportRoots_ and _parentScope_ flags together in the same mayaUSDExport command. Attempting to do so raises an error:
```
# Opening layer 'E:/tmp4.usda' for writing # 
# Error: RuntimeError: Given export root was neither a parent or child of any of the items to export; export aborting # 
```
We would like to use both to be able to pick out parts of scene for export but to arrange them within a different parent structure when generating the usd.

**Steps to reproduce**
1. Create empty Maya scene
2. Create a default sphere called pSphere1
3. Run this command to check exportRoots works:
```
pm.mayaUSDExport(file="C:/some_temp_location/roots.usda",
exportRoots=['|pSphere1'],
)
```
4. Then run this command to check parentScope works:
```
pm.mayaUSDExport(file="C:/some_temp_location/parent.usda",
parentScope='/geo_GRP',
)
```
5. Then attempt this command with both flags to see the error:
```
pm.mayaUSDExport(file="C:/some_temp_location/both.usda",
exportRoots=['|pSphere1'],
parentScope='/geo_GRP',
)
```

**Expected behavior**
In this simple case I would expect the result of step 4. and step 5. to be the same. And in a more complex case, I would expect to be able to export a branch of a hierarchy in Maya (specified with exportRoots) into a new parent in USD (specified with parentScope)

**Specs (if applicable):**
 - OS & version Windows 10
 - Maya version 2020 U4
 - Maya USD release 0.12.0_202108161213-d51aa00

cheers!

santosg87:
Hi @bentoogood I can see the behavior on my side as well. I have it marked as bug for now, but it is a bit unclear f this behavior is desired or not in terms of avoiding possible conflicts on the export. we are discussing this a bit more to decided on the next steps.

wtelford:
@santosg87 Can you log this internally?

santosg87:
logged as MAYA-114677

bentoogood:
Hi - just following up as we are testing the new 0.16.0 release in Maya 2022 and are experiencing this same error.

Are you able to confirm whether it is due to be addressed, or an immovable limitation?

thanks

bentoogood:
Just wanted to confirm that this works for us now in the new 0.18.0 release. Thanks for that.

PR #2400, opened 2022-06-06
MAYA-114678 [Github #1739] USDExport: using "parentScope" and "export…

boudrey:
…Roots" flags together returns an error

Add fix for legacy render layer mode.

### DIFF
--- a/lib/mayaUsd/fileio/writeJobContext.cpp
+++ b/lib/mayaUsd/fileio/writeJobContext.cpp
@@ -61,14 +61,18 @@ PXR_NAMESPACE_OPEN_SCOPE
 
 namespace {
 
-inline SdfPath _GetRootOverridePath(const UsdMayaJobExportArgs& args, const SdfPath& path)
+inline SdfPath _GetRootOverridePath(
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/fileio/writeJobContext.cpp
```diff
@@ -176,14 +180,14 @@ SdfPath UsdMayaWriteJobContext::ConvertDagToUsdPath(const MDagPath& dagPath) con
         path = path.GetParentPath();
     }
 
-    path = _GetRootOverridePath(mArgs, path);
+    path = _GetRootOverridePath(mArgs, path, false, true);
```
ppt-adsk, line 183: Can you comment each argument so we can figure out which is which at the call site?  E.g.
```suggestion
    path = _GetRootOverridePath(mArgs, path, /* modelRootOverride = */ false, /* rootMap = */ true);
```
Same at line 190.

ppt-adsk (approved):

dj-mcg (approved):
lgtm

ppt-adsk (approved):

PR #2406, opened 2022-06-09
MAYA-123507 - PluginLight needs to use the Light Icon instead of falling back to Xform icon

seando-adsk:
MAYA-123507 - PluginLight needs to use the Light Icon instead of falling back to the Xform Icon

* Added icon for PluginLight.
* Fix case where AL plugin is not built.
* Removed unneeded test property.


test/lib/ufe/testUIIcons.py
```diff
@@ -114,7 +112,7 @@ def testUIIcons(self):
             ('NurbsPatch',              'out_USD_NurbsPatch.png'),
             ('OpenVDBAsset',            'out_USD_UsdGeomXformable.png'),
             ('PackedJointAnimation',    'out_USD_SkelAnimation.png'),
-            ('PluginLight',             'out_USD_UsdGeomXformable.png' if usdVer >= (0, 21, 11) else 'out_USD_UsdLuxLight.png'),
+            ('PluginLight',             'out_USD_PluginLight.png' if usdVer >= (0, 21, 11) else 'out_USD_UsdLuxLight.png'),
```
seando-adsk, line 115: PluginLight now has an icon.

test/lib/ufe/testUIIcons.py
```diff
@@ -146,6 +144,14 @@ def testUIIcons(self):
                 ('PhysicsSphericalJoint',   'out_USD_UsdTyped.png')
             ])
 
+        # Special case for node types which are in an AL schema.
+        # They aren't available when compiling without the AL plugin.
+        if Usd.SchemaRegistry.IsConcrete(Tf.Type.FindByName('AL_usd_FrameRange')):
+            primTypes.extend([
+                ('ALExamplePolyCubeNode',   'out_USD_UsdTyped.png'),
+                ('ALFrameRange',            'out_USD_UsdTyped.png')
+        ])
```
seando-adsk, line 153: If you build locally without the AL plugin you won't have the schema that defines these two node types.

test/lib/ufe/testUIIcons.py
```diff
@@ -85,9 +85,7 @@ def testUIIcons(self):
         # All the prim types to test
         usdVer = Usd.GetVersion()
         primTypes = [
-                # Prim Type                # Icon file name
-            ('ALExamplePolyCubeNode',   'out_USD_UsdTyped.png'),
-            ('ALFrameRange',            'out_USD_UsdTyped.png'),
+            # Prim Type                 # Icon file name
             ('ALMayaReference',         'out_USD_MayaReference.png'),
```
boudrey, line 89: What about the ALMayaReference? Should it be moved too?

seando-adsk: The schema for that one is in `lib`, so its always there.

boudrey (approved):

boudrey (approved):

PR #2410, opened 2022-06-10
MAYA-123361 fix unshared stage toggling and loading

pierrebai-adsk:
- Add two new stage caches for unshared stages.
- These are necessary to be able to find unshared stages when toggling between shared and unshared.
- Do *not* pass an explicit sesssion layer when opening a stage unless unshared stage is explicitly desired.
- That is because the cache takes into consideration the session layer if given as an argument, which would unshare layers unintetionally.
- In particular, choosing to target the session layer (which is UI consideration) would indirectly make the stage unshared.
- Properly reload unshare stage and toggle between shared and unshared stage by transferring the content of the session layer.
- Clarify with comments the design and behaviour of the stage proxy shape.
- Adapted AL code to conform to the new API.
- Add unit test for session preservation

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -790,7 +790,7 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
                 if (rootLayer) {
                     // Note: computeSessionLayer will find a session layer *only* if the
                     //       Maya scene had been saved and thus serialized the session
-                    //       layer. Othersise it returns null which will mean to use
+                    //       layer. Otherwise it returns null which will mean to use
                     //       whatever session layer happens to be associated with the
                     //       stage we potentially find in the stage cache.
                     SdfLayerRefPtr sessionLayer = computeSessionLayer(dataBlock);
@@ -834,12 +834,9 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
                             loadSet);
                     }
 
-                    if (targetSession) {
-                        sessionLayer = sharedUsdStage->GetSessionLayer();
-                        sharedUsdStage->SetEditTarget(sessionLayer);
-                    } else {
-                        sharedUsdStage->SetEditTarget(sharedUsdStage->GetRootLayer());
-                    }
+                    sharedUsdStage->SetEditTarget(
+                        targetSession ? sharedUsdStage->GetSessionLayer()
+                                      : sharedUsdStage->GetRootLayer());
                 } else {
                     // Create a new stage in memory with an anonymous root layer.
                     sharedUsdStage = UsdStage::CreateInMemory(kAnonymousLayerName, loadSet);
@@ -958,7 +955,7 @@ UsdStageRefPtr MayaUsdProxyShapeBase::getUnsharedStage(UsdStage::InitialLoadSet 
     // attribute change, we want to find the same unshared stage, we don't want to lose
     // edits, in particular in its session layer.
     //
-    // We also need to be ble to find them when switching a stage between non-shared
+    // We also need to be able to find them when switching a stage between non-shared
     // and shared, so that we can transfer the content of the session layer.
     //
     // Fortunately, the USD stage cache matches stages using *all* arguments provided.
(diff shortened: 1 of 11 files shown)


lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -798,34 +805,49 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
                         = MGlobal::optionVarIntValue(kSessionLayerOptionVarName) == 1;
                     targetSession = targetSession || !rootLayer->PermissionToEdit();
 
-                    if (sessionLayer || targetSession) {
-                        if (!sessionLayer)
```
pierrebai-adsk, line 802: When targeting the session layer without a named session layer loaded from Maya scene, it would create a new session layer. This would prevent reusing a previously shared stage. Target session vs stage sharing are orthogonal. I looked at the original github issue and PR and the goal was to be able to target the session, with no discussion about affecting stage sharing.

I think the code was unwittingly written to have that side effect, so I fixed it.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -925,6 +951,83 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
...
+    UsdStageRefPtr           sharedUsdStage,
+    UsdStageRefPtr           unsharedUsdStage,
+    UsdStage::InitialLoadSet loadSet)
+{
+    // Based on the previous shared mode and current shared mode of the stage,
+    // transfer the content of the session layer from one to the other as needed.
+    const auto shareMode = isShareableStage() ? ShareMode::Shared : ShareMode::Unshared;
+    if (shareMode == _previousShareMode)
+        return;
+
```
pierrebai-adsk, line 986: On thing I was not sure of: do I need to reset the previous share mode in some cases? For example, if the user edits the filename, would we need to reset the flag? Where and when, through what mechanism would that be done?

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -925,6 +951,83 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
...
 }
 
+UsdStageRefPtr MayaUsdProxyShapeBase::getUnsharedStage(UsdStage::InitialLoadSet loadSet)
+{
+    // The unshared stages are *also* kept in a stage cache so that we can find them
+    // again when proxy shape attribute change. For example, if the 'loadPayloads'
+    // attribute change, we want to find the same unshared stage, we don't want to lose
+    // edits, in particular in its session layer.
+    //
+    // We also need to be ble to find them when switching a stage between non-shared
```
ppt-adsk, line 961: "ble" --> "able".

lib/mayaUsd/nodes/proxyShapeBase.cpp
ppt-adsk: Resetting the share mode on file (i.e. root layer) change is an interesting question.  I think you'll need to follow up with Design here so that we can schedule some time for this.

test/lib/mayaUsd/nodes/testProxyShapeBase.py
```diff
@@ -249,6 +249,48 @@ def testShareStage(self):
...
+        stage.DefinePrim("/dummy", "xform")
+
+        # verify that the prim exists.
+        def verifyPrim():
+            stage, _ = getStage()
+            self.assertTrue(stage.GetPrimAtPath("/dummy"))
+
+        verifyPrim()
+
+        # unshare the stage and verify the prim in the session laye still exists.
```
ppt-adsk, line 284: "session laye"

ppt-adsk (commented):
A few nit-picks, but nothing major.  Because you've changed Pixar plugin code (even though the changes are minor), I would feel more comfortable in having Pixar review this pull request, if possible.

lib/mayaUsd/nodes/proxyShapeBase.h
```diff
@@ -373,7 +395,12 @@ class MayaUsdProxyShapeBase
     // Whether or not the proxy shape has enabled UFE/subpath selection
     const bool _isUfeSelectionEnabled;
 
+    // Track the shared mode of the stage as seen in the last compute.
+    // Starts off as Unknown when the proxy shape is first created.
+    ShareMode _previousShareMode;
```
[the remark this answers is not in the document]
pierrebai-adsk: Well, when it is used in compute(), it represent the share mode as it was in the last compute, so it is the previous share mode, not the current one. The current one will be calculated during compute().

ppt-adsk (approved):

PR #2419, opened 2022-06-14
T bailp/maya 123815/reparent on deactivated

pierrebai-adsk:


seando-adsk (dismissed):
The change looks good to me. I've added PierreT as another review just to ensure I didn't miss anything.

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -214,7 +249,11 @@ Ufe::SceneItem::Ptr UsdHierarchy::parent() const
     // PointInstancer prim to be the "parent" of the point instance, even
     // though this isn't really true in the USD sense. This allows pick-walking
     // from point instances up to their PointInstancer.
-    return UsdSceneItem::create(fItem->path().pop(), prim().GetParent());
+    UsdPrim p = prim();
+    if (p.IsValid())
+        return UsdSceneItem::create(fItem->path().pop(), p.GetParent());
+    else
+        return Hierarchy::createItem(fItem->path().pop());
```
ppt-adsk, line 256: Not sure why this is correct.  If we have an invalid prim, shouldn't we return a nullptr as its parent?

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -270,12 +309,20 @@ Ufe::AppendedChild UsdHierarchy::appendChild(const Ufe::SceneItem::Ptr& child)
...
 
 Ufe::SceneItem::Ptr
 UsdHierarchy::insertChild(const Ufe::SceneItem::Ptr& child, const Ufe::SceneItem::Ptr& pos)
 {
+    // Changing the hierarchy of inactive items is not allowed.
+    if (!fItem->prim().IsValid())
+        return nullptr;
+
     auto insertChildCommand = insertChildCmd(child, pos);
     return insertChildCommand->insertedChild();
```
ppt-adsk, line 327: You're repeating the test, and now insertChildCmd() can return a nullptr, which you should check.  Yes, currently, since the test is the same, you're avoiding nullptr deref, but because there are two tests, they could get out of sync.  I would just remove the duplicate test here.
```suggestion
    auto insertChildCommand = insertChildCmd(child, pos);
    return insertChildCommand ? insertChildCommand->insertedChild() : nullptr;
```

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -284,6 +331,10 @@ UsdHierarchy::insertChild(const Ufe::SceneItem::Ptr& child, const Ufe::SceneItem
 #ifdef UFE_V3_FEATURES_AVAILABLE
 Ufe::SceneItem::Ptr UsdHierarchy::createGroup(const Ufe::PathComponent& name) const
 {
+    // Changing the hierarchy of invalid items is not allowed.
+    if (!fItem->prim().IsActive())
+        return nullptr;
+
```
ppt-adsk, line 337: Test is repeated a few places.  Would be best to put it inside UsdUndoCreateGroupCommand::create(), as the current code correctly deals with a nullptr cmd.

lib/mayaUsd/ufe/UsdHierarchy.cpp
```diff
@@ -338,6 +401,10 @@ Ufe::SceneItem::Ptr UsdHierarchy::defaultParent() const
 
 Ufe::UndoableCommand::Ptr UsdHierarchy::reorderCmd(const Ufe::SceneItemList& orderedList) const
 {
+    // Changing the hierarchy of inactive items is not allowed.
+    if (!fItem->prim().IsActive())
+        return nullptr;
+
```
ppt-adsk, line 407: You're saving a bit of work by putting the test before the loop, but it would still be safer to put the test in UsdUndoReorderCommand::create().

ppt-adsk (changes requested):
I'd move some of the checks to the command creation code.  Can you confirm that ProxyShapeHierarchy does not need any changes, because it corresponds to the pseudo-root, which cannot be inactivated?

lib/mayaUsd/ufe/UsdHierarchy.cpp
pierrebai-adsk: I hit this case while investigating. I did not write down the exact circumstances (I think you could use the repro steps of the Maya JIRA ticket with a breakpoint in this function without my code changes.) The path parent is valid, but not the path. It was breaking checking if a node had children, IIRC.

lib/mayaUsd/ufe/UsdHierarchy.cpp
pierrebai-adsk: Oh I failed to notice they called each other, I did not read the implementation, I just put the check in every function that was related to hierarchy changes. I'll try your changes.

lib/mayaUsd/ufe/UsdHierarchy.cpp
pierrebai-adsk: The function is an override of a virtual function exposed by UFE, so it could get called from elsewhere than UsdUndoCreateGroupCommand::create(). Would it not be better to leave the checks here?

lib/mayaUsd/ufe/UsdHierarchy.cpp
ppt-adsk: I can see that happening with the child of an inactive prim.  Attempting to use that prim would be invalid, its parent would not.  I suspect we are hiding a bug with this code...  We are not checking that the parent path is valid.  I would return a nullptr and see what breaks.

ppt-adsk (approved):
We can merge as-is, and you can investigate the code path where a parent is returned for an invalid prim whose parent path is valid later.

lib/mayaUsd/ufe/UsdHierarchy.cpp
pierrebai-adsk: The call was correct. It was trying to get the parent of the deactivated prim, IIRC. The deactivated prim has a parent, so we should not fail the call.

ppt-adsk: But the deactivated prim is still valid, so we should never get into that branch of the conditional.

pierrebai-adsk: Indeed, and yet I got into that situation, otherwise I'd not change the code :). IDK how I got there during my investigations, honestly. (and unfortunately, blame my bad memory)

pierrebai-adsk:
Only failure is the known Linux 2022 interactive test.

PR #2422, opened 2022-06-15
MAYA-123792 select undeleted stage's prim

pierrebai-adsk:
When a stage was deleted in the outliner and then the deletion undone, selection of a prim would cause errors. Also, the outliner content would be incorrect.

The root cause was that undo would not update the stage cache, which would break a lot of things.

The fix is to not try to catch every possible ways the stage might get invalid (the famous "cache invalidation is one of the hard problem of computer science") and instead just invalidate and refill it when a cache miss happens.

As the stage cache is not a LRU cache but a mapping between representations, its content is really rarely incorrect, except during transient period during the flurry of notification affecting its validity: deleting a stage node, renaming it, undoing these operation, etc. These operation would end-up invalidating the cache anyway, so rebuilding it on miss is equivalent. That is why refill on cache miss or cache inconsistency ends up being the correct thing to do.

Note that each approach has its pros and cons. Using explicit invalidation approach theoretically minimizes the number of rebuilds, but require to perfectly catch every situation that may make the cache invalid, which is error prone and costly to design and maintain. Using automatic rebuilding approach does not require any code maintenance, but could theoretically cause multiple rebuilds.

The "theoretically" are important there. In actual usage, automatic rebuild are efficient, because the transient period is short and in my testing the cache gets rebuilt only once. (i.e. some sub-system tries to access a stage that was undeleted, the cache is rebuilt and then the cache works for all subsequent callers.) Ironically, trying to explicitly invalidate the cache leads to multiple invalidations because the notifications sometimes come from multiple sources, so are triggered multiple times.


lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -307,14 +307,6 @@ void StagesSubject::afterOpen()
...
         });
     fStageListeners.clear();
-
-    // Set up our stage to proxy shape UFE path (and reverse)
-    // mapping.  We do this with the following steps:
-    // - get all proxyShape nodes in the scene.
-    // - get their Dag paths.
-    // - convert the Dag paths to UFE paths.
-    // - get their stage.
-    g_StageMap.setDirty();
```
seando-adsk, line 317: What happened with this map. I don't see anywhere that is kept up-to-date now, but it is still being referenced in Utils.cpp.

lib/mayaUsd/ufe/UsdStageMap.cpp
```diff
@@ -158,44 +170,27 @@ MObject UsdStageMap::proxyShape(const Ufe::Path& path)
...
-            // Get the UFE path from the map value.
-            auto newPath = firstPath(cachedObject);
-            if (newPath != cachedPath) {
-                // Key is stale.  Remove it from our cache, and add the new entry.
-                auto count = fPathToObject.erase(cachedPath);
-                TF_AXIOM(count);
-                fPathToObject[newPath] = cachedObject;
-            }
-        }
+        rebuildCache();
```
seando-adsk, line 173: Isn't it possible to get into a state where we rebuilt the cache (because it was dirty) on line 140. Then you are forcing it to rebuild again here.

lib/mayaUsd/ufe/StagesSubject.cpp
pierrebai-adsk: The map now gets updated automatically when it detects an attempt to access out-of--date data. But you are right I could remove the extern declaration at the top of the file. Also, I did not expect a review, I still need to write a unit test. :) But hey, you are very pro-active :)

lib/mayaUsd/ufe/UsdStageMap.cpp
pierrebai-adsk: Normally the 1st (conditional) rebuild would make the cache valid and not fail here, but there maybe cases were we rebuild twice. For example when accessing data that is really gone. In fact, currently nothing calls setDirty() in the whole code base... I think I could just get rid of the 1st conditional rebuild and let the code here do its thing, you are right.

lib/mayaUsd/ufe/StagesSubject.cpp
pierrebai-adsk: Oh, and Utils.cpp just provides function to access the content of the map, so that is normal that it still refers to it.

seando-adsk: Sorry never mind my comment. I didn't release at first this variable is our UsdStageMap class. I thought it was a Pixar type that we needed to add/remove items from.

ppt-adsk (commented):
When I last made a fix to this cache, I initially did as you did and rebuilt everything on invalidation.  Krystian had me change that here:
https://github.com/Autodesk/maya-usd/pull/1245/commits/5902c7d36841e7c96dae925aa4780690c17e6346
as his argument was that for a large cache, only one member might need updating, so rebuilding the whole cache was inefficient, e.g. for a scene with thousands of proxy shapes.  What is your answer to that concern?

pierrebai-adsk:
There are multiple points:

- We are unlikely to have thousands of proxy shapes, The typical case is one. I'd expect most case to fall into the less-than-five scenario.
- The current cases (and most cases I'd be inclined to think) where this gets triggered (stage deletion, renaming, etc) trigger USD scene invalidation notifications that would mark the cache dirty and thus rebuilding it anyway.
- This only happens when either deleting a stage or the stage is renamed, or a few other cases (a Maya ref containing a stage being unloaded, editing-as-maya a maya file that contains s stage) and other sub-systems have not yet receive the note that the stage is gone,

pierrebai-adsk:
When discussing with Pixar and AnimalLogic, they confirmed that a typical scene contains only one stage, so only one proxy shape.

ppt-adsk:
Yes, that's expected, but that's a sampling of only two customers.  Other customers have been using multiple proxy shapes.

seando-adsk:
@pierrebai-adsk / @ppt-adsk What is the status of this PR? Did we come to a conclusion on the number of proxy shapes and whether we need to maintain the code the way it is now or if dirtying the entire cache is okay.

PR #2428, opened 2022-06-17
MAYA-114376 preserve unsaved root layer

pierrebai-adsk:
When the proxy stage is recompute with an unsaved root layer, the attribute keeping the root layer name would be empty. A new root layer would be created each time, losing any previous content.

This attribute only gets set once the stage gets saved.

Fix this by re-using the same anonymous root layer when the stage has not been saved yet.

Also, as a bonus, this fix change teh code so that the same code path for newly created stage is used, which means that if the user ahas set the option var to target the session layer, it will also be honored for newly-created stages. It will also re-use any pre-existing stage, so we also avoid creating stages repeatedly for the same unsaved proxy stage node every time an attribute get sets.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -788,7 +788,10 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
                     rootLayer = SdfLayer::FindOrOpen(fileString);
 
                 if (nullptr == rootLayer) {
-                    // Create a new stage in memory with an anonymous root layer.
+                    // Create an empty in-memory root layer so that a new stage in memory
+                    // will be created below by the UsdStage::Open call. This happens when
+                    // a brand new stage with a new anonymous layer is requested to be
+                    // created by the user.
                     if (!_anonymousRootLayer)
                         _anonymousRootLayer = SdfLayer::CreateAnonymous(kAnonymousLayerName);
                     rootLayer = _anonymousRootLayer;
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -787,7 +787,14 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
                 if (nullptr == rootLayer)
                     rootLayer = SdfLayer::FindOrOpen(fileString);
 
-                if (rootLayer) {
+                if (nullptr == rootLayer) {
+                    // Create a new stage in memory with an anonymous root layer.
+                    if (!_anonymousRootLayer)
+                        _anonymousRootLayer = SdfLayer::CreateAnonymous(kAnonymousLayerName);
+                    rootLayer = _anonymousRootLayer;
+                }
+
+                {
                     // Note: computeSessionLayer will find a session layer *only* if the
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -787,7 +787,14 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
                 if (nullptr == rootLayer)
                     rootLayer = SdfLayer::FindOrOpen(fileString);
 
-                if (rootLayer) {
+                if (nullptr == rootLayer) {
```
pierrebai-adsk, line 790: This used to be in the else below, but by placing it here it can do the same extra processing existing stages do. There was no reason to exclude new stage from that processing. (Like targeting the session layer if requested.)

pierrebai-adsk:
The PF failure are unrelated: the known interactive image test and the last PR build fails on UFE, which is unrelated to my changes.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -787,7 +787,14 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
...
+                    // Create a new stage in memory with an anonymous root layer.
+                    if (!_anonymousRootLayer)
+                        _anonymousRootLayer = SdfLayer::CreateAnonymous(kAnonymousLayerName);
+                    rootLayer = _anonymousRootLayer;
```
seando-adsk, line 794: Your comment says "create a new stage in memory". Can you point me to the code in USD where that happens. I had a look at SdfLayer::CreateAnonymous() and don't see anywhere that it (or called methods) create a stage in memory.

And I don't fully understand why in the case where we just want an empty layer we would want to fall into the code below and call UsdStage::Open?

pierrebai-adsk: You are right, the comment should say create a new *layer* in memory!

As for falling down, in the end we want a stage, so for an empty layer, passing the newly-created layer to open will create a stage with that layer. The code is just so that all cases passes in common code paths, to avoid maintenance issues where different ways of creating the stage would not perform the same steps. (For example target the session vs root layer.)

pierrebai-adsk: Oh, no wait, the comment is correct, but anticipating what will happen: we create an anonymous layer to create an empty stage in memory in the Open call below.

ppt-adsk (commented):
Some of these files appear to come from another branch.  Am I missing something?

ppt-adsk:
> Which file? They are not coming from another branch.

UsdHierarchy.cpp and UsdUndoCreateGroupCommand.cpp.  What do these have to do with preserving an unsaved root layer?  For example, the UsdHierarchy.cpp changes are here
https://github.com/Autodesk/maya-usd/commit/610b5bb3593c9f9c92ff9d6d46301710092bbccb
which was merged here
https://github.com/Autodesk/maya-usd/pull/2419
which is your own pull request.

pierrebai-adsk:
I think I must have rebased at some point. I don't understand why github does not show that the file are unchanged if the change are already merged in another PR?

pierrebai-adsk:
I merged the dev branch into this one and it no longer shows unrelated changes.

ppt-adsk (approved):

seando-adsk (approved):

PR #2431, opened 2022-06-20
Provide explicit export options in tests

dj-mcg:
Provide explicit values for export options in open source tests even if
they match defaults. This allows tests to pass in environments which
run with site-specific export option overrides


dj-mcg:
Is there anything I need to do to address the Windows build failures or is that a preflight issue? Just wanted to make sure this wasn't waiting on me!

seando-adsk:
@dj-mcg The preflight errors that showed in the last run have been fixed in our internal system. I had started this preflight before the fix was updated in the build system. So I'll consider this preflight a pass and merge the PR.

PR #2435, opened 2022-06-20
Test Maya fixes for MAYA-122305 and MAYA-123312 (center pivot)

ppt-adsk:


test/lib/ufe/testComboCmd.py
```diff
@@ -890,6 +890,97 @@ def testCenterPivotUndo(self):
         assertVectorAlmostEqual(self, usdT3d.rotatePivot().vector, [0.0, 0.0, 0.0])
         assertVectorAlmostEqual(self, usdT3d.scalePivot().vector, [0.0, 0.0, 0.0])
 
+    @unittest.skipUnless(mayaUtils.ufeSupportFixLevel() > 0, "Requires center pivot Maya fix.")
```
ppt-adsk, line 893: Use UFE fix level support to avoid running unless fix is present in Maya.

pierrebai-adsk: No need to check major version?

ppt-adsk: Nope.  UFE fix level is always increasing, and always incremented in master.

PR #2437, opened 2022-06-21
MAYA-124064 fix unit test for Maya 2023

pierrebai-adsk:
- Diagnostic was unloading the plugin and for some reason reloading it failed on the CI.
- Making the test that unloads the plugin run last should fix the test.
- Note that I could not reproduce this failure, but comparing test logs showed that the failing CI had an extra error message in the log when trying to reload the plugin.
- Observable scene test was creating a path segment with an invalid runtime ID.
- On master, we catch exception during notification delivery, but not in 2023, so fix the test to use a valid run-time ID.


pierrebai-adsk:
I forgot to note: the reason it the scene observation test started to fail recently was that the exception is thrown by the new light handling code that was added recently in Maya.

pierrebai-adsk:
:TY: you were faster than me to set the label :)

PR #2441, opened 2022-06-23
LOOKDEVX-679 - Propagate MatX metadata info

JGamache-autodesk:
- Add detection code for new USD features that might appear earlier than next Pixar release
- Show shader parameters at the top of AE
- Implement grouping in AE if metadata is available
- Fix missing Ufe types from LOOKDEVX-670
- Add underscore support to UI label prettification code

### DIFF
--- a/lib/mayaUsd/python/wrapUtil.cpp
+++ b/lib/mayaUsd/python/wrapUtil.cpp
@@ -49,7 +49,7 @@ void wrapUtil()
 {
     scope s = class_<UsdMayaUtilScope>("Util", no_init)
                   .def("IsAuthored", UsdMayaUtil::IsAuthored)
-                  .def("PrettifyName", &UsdMayaUtil::PrettifyName)
-                  .staticmethod("PrettifyName")
+                  .def("prettifyName", &UsdMayaUtil::prettifyName)
+                  .staticmethod("prettifyName")
                   .def("getDictionaryFromEncodedOptions", getDictionaryFromEncodedOptions);
 }
--- a/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
@@ -146,7 +146,7 @@ class MetaDataCustomControl(object):
             for k in allMetadata:
                 # All extra metadata is for display purposes only - it is not editable, but we
                 # allow keyboard focus so you copy the value.
-                mdLabel = mayaUsdLib.Util.PrettifyName(k) if self.useNiceName else k
+                mdLabel = mayaUsdLib.Util.prettifyName(k) if self.useNiceName else k
                 self.extraMetadata[k] = cmds.textFieldGrp(label=mdLabel, editable=False, enableKeyboardFocus=True)
 
         # Update all metadata values.
@@ -243,7 +243,7 @@ class ArrayCustomControl(object):
             typeNameStr = str(typeName.scalarType)
             typeNameStr += ("[" + str(len(values)) + "]") if hasValue else "[]"
 
-            attrLabel = mayaUsdLib.Util.PrettifyName(self.attrName) if self.useNiceName else self.attrName
+            attrLabel = mayaUsdLib.Util.prettifyName(self.attrName) if self.useNiceName else self.attrName
             singleWidgetWidth = mel.eval('global int $gAttributeEditorTemplateSingleWidgetWidth; $gAttributeEditorTemplateSingleWidgetWidth += 0')
             with AEUITemplate():
                 # See comment in ConnectionsCustomControl below for why nc=5.
@@ -313,7 +313,7 @@ class ConnectionsCustomControl(object):
             attrLabel = str(ufeAttr.getMetadata("uiname"))
 
             if not attrLabel:
-                attrLabel = mayaUsdLib.Util.PrettifyName(self.attrName)
+                attrLabel = mayaUsdLib.Util.prettifyName(self.attrName)
         attrType = attr.GetMetadata('typeName')
 
         singleWidgetWidth = mel.eval('global int $gAttributeEditorTemplateSingleWidgetWidth; $gAttributeEditorTemplateSingleWidgetWidth += 0')
@@ -486,7 +486,7 @@ class AETemplate(object):
                 schemaTypeName = schemaTypeName.replace(p, r, 1)
                 break
 
-        schemaTypeName = mayaUsdLib.Util.PrettifyName(schemaTypeName)
+        schemaTypeName = mayaUsdLib.Util.prettifyName(schemaTypeName)
 
         # if the schema name ends with "api", replace it with
         # "API" and make sure to only replace the last occurence.
@@ -511,7 +511,7 @@ class AETemplate(object):
         # Hide all outputs:
         for name in nodeDef.GetOutputNames():
             self.suppress(UsdShade.Utils.GetFullName(name, UsdShade.AttributeType.Output))
-        with ufeAeTemplate.Layout(self, "Shader: " + mayaUsdLib.Util.PrettifyName(label)):
+        with ufeAeTemplate.Layout(self, "Shader: " + mayaUsdLib.Util.prettifyName(label)):
             pages = nodeDef.GetPages()
             if len(pages) == 1 and not pages[0]:
                 pages = []
@@ -528,7 +528,7 @@ class AETemplate(object):
                             name = UsdShade.Utils.GetFullName(name, UsdShade.AttributeType.Input)
                             attrsToAdd.append(name)
                     if attrsToAdd:
-                        with ufeAeTemplate.Layout(self, mayaUsdLib.Util.PrettifyName(pageLabel), collapse):
+                        with ufeAeTemplate.Layout(self, mayaUsdLib.Util.prettifyName(pageLabel), collapse):
                             self.addControls(attrsToAdd)
             else:
                 attrsToAdd = []
--- a/lib/mayaUsd/utils/util.cpp
+++ b/lib/mayaUsd/utils/util.cpp
@@ -724,7 +724,7 @@ std::string UsdMayaUtil::SanitizeName(const std::string& name)
     return TfStringReplace(name, ":", "_");
 }
 
-std::string UsdMayaUtil::PrettifyName(const std::string& name)
+std::string UsdMayaUtil::prettifyName(const std::string& name)
 {
     std::string prettyName(1, std::toupper(name[0]));
     size_t      nbChars = name.size();
--- a/lib/mayaUsd/ufe/UsdShaderAttributeDef.cpp
+++ b/lib/mayaUsd/ufe/UsdShaderAttributeDef.cpp
@@ -79,7 +79,7 @@ static const MetadataMap _metaMap = {
     { "uiname",
       [](const PXR_NS::SdrShaderProperty& p) {
           return !p.GetLabel().IsEmpty() ? p.GetLabel().GetString()
-                                         : UsdMayaUtil::PrettifyName(p.GetName().GetString());
+                                         : UsdMayaUtil::prettifyName(p.GetName().GetString());
       } },
     { "doc",
       [](const PXR_NS::SdrShaderProperty& p) {
--- a/lib/mayaUsd/ufe/UsdShaderNodeDef.cpp
+++ b/lib/mayaUsd/ufe/UsdShaderNodeDef.cpp
@@ -234,7 +234,7 @@ static const MetadataMap _metaMap = {
     { "uiname",
       [](const PXR_NS::SdrShaderNode& n) {
           return !n.GetLabel().IsEmpty() ? n.GetLabel().GetString()
-                                         : UsdMayaUtil::PrettifyName(n.GetFamily().GetString());
+                                         : UsdMayaUtil::prettifyName(n.GetFamily().GetString());
       } },
     { "doc",
       [](const PXR_NS::SdrShaderNode& n) {
--- a/lib/mayaUsd/utils/util.h
+++ b/lib/mayaUsd/utils/util.h
@@ -295,7 +295,7 @@ std::string SanitizeName(const std::string& name);
 /// Replace underscore by space and capitalize next letter
 /// Always capitalize first letter
 MAYAUSD_CORE_PUBLIC
-std::string PrettifyName(const std::string& name);
+std::string prettifyName(const std::string& name);
 
 // This to allow various pipeline to sanitize the colorset name for output
 MAYAUSD_CORE_PUBLIC
--- a/test/lib/ufe/testAttribute.py
+++ b/test/lib/ufe/testAttribute.py
@@ -1677,15 +1677,15 @@ class AttributeTestCase(unittest.TestCase):
 
     def testNamePrettification(self):
         '''Test the name prettification routine.'''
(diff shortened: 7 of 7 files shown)


lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -35,31 +35,12 @@
...
 # We manually import all the classes which have a 'GetSchemaAttributeNames'
 # method so we have access to it and the 'pythonClass' method.
-from pxr import Usd, UsdGeom, UsdLux, UsdRender, UsdRi, UsdShade, UsdSkel, UsdUI, UsdVol, Kind, Tf
+from pxr import Usd, UsdGeom, UsdLux, UsdRender, UsdRi, UsdShade, UsdSkel, UsdUI, UsdVol, Kind, Tf, Sdr
 
 nameTxt = 'nameTxt'
 attrValueFld = 'attrValueFld'
 attrTypeFld = 'attrTypeFld'
 
-def getPrettyName(name):
```
JGamache-autodesk, line 44: Moved to C++ function `UsdMayaUtil::prettifyName()` exported to Python as `mayaUsdLib.Util.prettifyName()`

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -507,6 +496,46 @@ def sectionNameFromSchema(self, schemaTypeName):
 
         return schemaTypeName
 
+    def createShaderAttributesSection(self):
```
JGamache-autodesk, line 499: Move all shader parameters to top of the AE. They used to reside in "Extra Attributes".

lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -177,7 +177,11 @@ U getUsdAttributeVectorAsUfe(
 {
     VtValue vt;
     if (!attr.isValid() || !attr.hasValue()) {
-        vt = MayaUsd::ufe::vtValueFromString(attr.typeName(), attr.defaultValue());
```
JGamache-autodesk, line 180: Do not crash if there is no default value. And do even report an error. These are problems coming from projects out of the control of Autodesk, so there is not much the end user can do about them.

lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -335,6 +347,12 @@ namespace ufe {
 UsdAttribute::UsdAttribute(const PXR_NS::UsdPrim& prim, const Ufe::AttributeDef::ConstPtr& attrDef)
     : fPrim(prim)
     , fAttrDef(attrDef)
+    , fUsdAttr(prim.GetAttribute(PXR_NS::UsdShadeUtils::GetFullName(
```
JGamache-autodesk, line 350: As was said in a previous review, the attribute can always be retrieved from the UsdPrim and the attribute name. Put that in application here to have one CTOR for Usd Attribute only, and another one for when we have an Attribute def. As was previously discussed, these two constructors should be in two classes, one derived from the other one, but that will not happen in this PR yet.

lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -532,14 +550,13 @@ Ufe::Value UsdAttribute::getMetadata(const std::string& key) const
                 return Ufe::Value(ss.str());
             }
         }
-        return Ufe::Value();
```
JGamache-autodesk, line 535: Do not return immediately here. There might be information coming from the AttributeDef.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -507,6 +496,46 @@ def sectionNameFromSchema(self, schemaTypeName):
...
+        """Use Sdr node information to populate the shader section"""
+        shader = UsdShade.Shader(self.prim)
+        nodeId = shader.GetIdAttr().Get()
+        if not nodeId:
```
feldstj, line 503: Should any form of error handling be done in these cases?

lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -734,8 +751,11 @@ UsdAttributeFilename::create(const UsdSceneItem::Ptr& item, const PXR_NS::UsdAtt
 
 std::string UsdAttributeFilename::get() const
 {
+    if (!hasValue())
```
JGamache-autodesk, line 754: Updated to use the same unauthored attribute logic as the other classes.

lib/mayaUsd/python/wrapUtil.cpp
```diff
@@ -50,5 +50,6 @@ void wrapUtil()
     scope s(class_<UsdMayaUtilScope>("Util"));
 
     def("IsAuthored", UsdMayaUtil::IsAuthored);
+    def("prettifyName", UsdMayaUtil::prettifyName);
```
feldstj, line 53: (minor) - why is IsAuthored capitalized while prettifyName is lower case?

lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -112,6 +112,11 @@ static PXR_NS::UsdAttribute _GetAttributeType(const PXR_NS::UsdPrim& prim, const
 
 Ufe::Attribute::Type UsdAttributes::attributeType(const std::string& name)
 {
+    // early return if name is empty.
```
JGamache-autodesk, line 115: Error handling. Required because Python code from Ufe checks the type before creating an attribute.

lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -124,22 +129,23 @@ Ufe::Attribute::Type UsdAttributes::attributeType(const std::string& name)
...
+        + fItem->path().string() + "\".";
+    throw std::invalid_argument(msg);
 }
 
 Ufe::Attribute::Ptr UsdAttributes::attribute(const std::string& name)
 {
     // early return if name is empty.
     if (name.empty()) {
-        return nullptr;
+        throw std::invalid_argument("Empty name.");
```
JGamache-autodesk, line 148: C++ consumers will get this error handling.

lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -181,10 +195,10 @@ Ufe::Attribute::Ptr UsdAttributes::attribute(const std::string& name)
                const PXR_NS::UsdPrim&             prim,                   \
                const Ufe::AttributeDef::ConstPtr& attrDef,                \
                const PXR_NS::UsdAttribute&        usdAttr) {                     \
-                if (usdAttr) {                                            \
```
JGamache-autodesk, line 184: The code was assuming that an existing USD attribute would not require an AttributeDef. This is incorrect since there is metadata information we might want to get. Changed to select a CTOR based on the presence of an AttributeDef.

lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -335,6 +347,12 @@ namespace ufe {
 UsdAttribute::UsdAttribute(const PXR_NS::UsdPrim& prim, const Ufe::AttributeDef::ConstPtr& attrDef)
```
feldstj, line 347: My recollection was that there were two UsdAttribute constructors, one that takes an attribute and the other that takes a prim and attribute definition. When the prim version is called there shouldn't be an attribute on the prim, no? In which case the GetAttribute(...) code below should fail. Is that not correct?

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -600,6 +600,9 @@ Ufe::Attribute::Type usdTypeToUfe(const PXR_NS::SdrShaderPropertyConstPtr& shade
                 { PXR_NS::SdrPropertyTypes->String, PXR_NS::SdfValueTypeNames->String },
                 { PXR_NS::SdrPropertyTypes->Float, PXR_NS::SdfValueTypeNames->Float },
                 { PXR_NS::SdrPropertyTypes->Color, PXR_NS::SdfValueTypeNames->Color3f },
+#if defined(USD_HAS_COLOR4_SDR_SUPPORT)
+                { PXR_NS::SdrPropertyTypes->Color4, PXR_NS::SdfValueTypeNames->Color4f },
+#endif
```
JGamache-autodesk, line 605: Without that patched USD with Color4 support, a MaterialX color4 will look and act as a Float4.

lib/mayaUsd/utils/util.cpp
```diff
@@ -722,6 +724,32 @@ std::string UsdMayaUtil::SanitizeName(const std::string& name)
     return TfStringReplace(name, ":", "_");
 }
 
+std::string UsdMayaUtil::prettifyName(const std::string& name)
+{
```
JGamache-autodesk, line 728: Moved from Python AE template code.

lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -112,6 +112,11 @@ static PXR_NS::UsdAttribute _GetAttributeType(const PXR_NS::UsdPrim& prim, const
 
 Ufe::Attribute::Type UsdAttributes::attributeType(const std::string& name)
 {
+    // early return if name is empty.
+    if (name.empty()) {
+        throw std::invalid_argument("Empty name.");
```
feldstj, line 117: I'm not really familiar with how MayaUsd exceptions are thrown. Is this enough information, or do you need to somehow include where this is coming from?

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -625,17 +634,26 @@ Ufe::Attribute::Type usdTypeToUfe(const PXR_NS::SdrShaderPropertyConstPtr& shade
...
+            { Ufe::Attribute::kFloat, PXR_NS::SdfValueTypeNames->Float },         // float
+            { Ufe::Attribute::kDouble, PXR_NS::SdfValueTypeNames->Double },       // double
+            { Ufe::Attribute::kString, PXR_NS::SdfValueTypeNames->String },       // std::string
+            { Ufe::Attribute::kEnumString, PXR_NS::SdfValueTypeNames->Token },    // TfToken
+            { Ufe::Attribute::kInt3, PXR_NS::SdfValueTypeNames->Int3 },           // GfVec3i
+            { Ufe::Attribute::kFloat3, PXR_NS::SdfValueTypeNames->Float3 },       // GfVec3f
+            { Ufe::Attribute::kDouble3, PXR_NS::SdfValueTypeNames->Double3 },     // GfVec3d
+            { Ufe::Attribute::kColorFloat3, PXR_NS::SdfValueTypeNames->Color3f }, // GfVec3f
+#if (UFE_PREVIEW_VERSION_NUM >= 4015)
+            { Ufe::Attribute::kFilename, PXR_NS::SdfValueTypeNames->Asset },
```
feldstj, line 650: Should you have comments equivalent to those above that show the data type? ie. // GfVec3d

feldstj (changes requested):

ppt-adsk (commented):
Just question about error management going from return value to exception.

lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -124,22 +129,23 @@ Ufe::Attribute::Type UsdAttributes::attributeType(const std::string& name)
...
+    std::string msg = "Can not find attribute \"" + name + "\" on scene item \""
+        + fItem->path().string() + "\".";
+    throw std::invalid_argument(msg);
 }
 
 Ufe::Attribute::Ptr UsdAttributes::attribute(const std::string& name)
 {
     // early return if name is empty.
     if (name.empty()) {
-        return nullptr;
```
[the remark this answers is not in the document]
JGamache-autodesk: The main Python call site is in UFE itself. It is one that I can actually point to on this public forum since it is installed in Maya.
Look for `/runTime/Python/Lib/site-packages/ufe/__init__.py`
When passing an empty string, or a name that does not exist, the code in `typedAttributeFactory()` calls `attrs.attributeType(attrName)`, receives `Unknown` then complains that this is not found in `typedAttributeDict`. I found that to be a bit too obscure, hence the exception. I would have preferred `KeyError` instead of `ValueError` on the Python side though.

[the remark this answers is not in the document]
JGamache-autodesk: I do have to agree that switching error handling on short notice is bug-prone though, so I would have no issue reverting that change.

lib/mayaUsd/python/wrapUtil.cpp
JGamache-autodesk: The coding standards requires CamelCase, I will update the code.

lib/mayaUsd/ufe/UsdAttribute.cpp
JGamache-autodesk: No. You can have both an Attribute and an AttributeDef. If that causes the Attribute-only CTOR to be called, then you lose all the rich information the AttributeDef provides.

I either had to add a 3rd CTOR for Attribute+AttributeDef, or simply leverage the fact that you don't even need the attribute to call the Prim+AttributeDef CTOR.

lib/mayaUsd/ufe/Utils.cpp
JGamache-autodesk: Are these comments adding value in this function?

lib/mayaUsd/ufe/UsdAttribute.cpp
feldstj: So I had a discussion with Patrick a little while ago about changing the attribute constructor to also take an attribute definition as well (nullptr for the definition if there isn't one). We scheduled work for this with a Jira ticket, but it hasn't been implemented yet. If you wanted to you could make that change.

lib/mayaUsd/ufe/Utils.cpp
feldstj: Not sure, but I would think for consistency they should either be removed or added.

feldstj: Also, isn't this a trivial change? If not, by all means ignore this.

lib/mayaUsd/ufe/UsdAttribute.cpp
ppt-adsk: The more we delay making interface changes that support the code in a more natural, better encapsulated way, with proper separation of concerns, the more we make client code hard to understand, full of unnecessary concepts and dependencies.

ppt-adsk (approved):

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
seando-adsk: Did you show this to Natalia?

seando-adsk (approved):

lib/mayaUsd/ufe/UsdAttributes.cpp
JGamache-autodesk: reverted.

seando-adsk (approved):

lib/mayaUsd/python/wrapUtil.cpp
JGamache-autodesk: Yes. I rechecked and it is indeed lowerCamelCase.

ppt-adsk (approved):

seando-adsk (approved):

PR #2447, opened 2022-06-29
MAYA-123876 preserve session layer on save

pierrebai-adsk:
When a new stage with an anonymous root layer is saved for the first time, the session layer was lost.

The reason was that when a new stage is saved, its root layer change ID. When the proxy shape was recomputed, it tried to find the root layer with the new ID in the stage cache. Since the just-saved stage was previously using an anonymous root layer, the cache did not find the just-saved stage. A new stage got created, which meant it got a brand new empty session layer.

This recomputation happened mid-save.

The fix is to avoid recomputing the proxy stage mid-save and to pre-fill the stage cache with the newly-saved root with its new ID and with the pre-existing session layer. This requires keeping more information when initially gathering the layers to be saved to avoid touching the proxy shape again during the save.

- Add StageSavingInfo to the layer manager to cache all info necessary during saving.
- Pass these info instead of only the DAG path to callbacks and various internal functions.
- Have the LayerDatabase fill this info when gathering layers to save.
- Avoid calling getProxiesToSave() mid-save.
- Add hasDirtyLayer() to support the previous point.
- Refactor clearing proxies to save into its own function to avoid code duplication.
- Add GetAllCaches() to UsdMayaStageCache to allow access to all caches without knowing the details of how caches are segregated.
- When saving a layer, prefill all stage caches that were holding the stages that were using the original layer to now cache a stage that uses the new layer and the session layer of the existing stage.
- This will allow the proxy shape to find these stage with the correct session layer when recomputed later.
- Update the batch save layer UI to the new API with StageSavingInfo.
- Add a unit test for session layer in save

In addition to that, the layer manager would not save the session layer of the stage in the Maya scene file when the layers were being saved to external USD files.

There were two problems: the session layer name was not being set on the stage proxy node and the layer manager was not saved to the Maya scene.

The layer manager node would be lost when saving layer to separate USD files. (Actually, this one but one possible case where the layer manager would be lost.) The reason was that the layer manager node would be deleted mid-save because some other code happened to access the stage while the stage was dirty, which would cause the DG to re-compute the stage node, which would access layers in the layer manager, which would wipe the layer manager node. And all this was dependent on the order some save callbacks were being called.

To avoid this, we now no longer wipe the layer manager node when in the middle of a save operation.

Also, improved the proxy shape test to cleanup after themselves, deleting their temp files.


pierrebai-adsk:
This PR, while large would solve multiple Maya ref bugs in one go (it also solves MAYA-124137), please review.

ppt-adsk (approved):

PR #2448, opened 2022-06-29
MAYA-124088: USD breaks callback registration for addRMBBakingMenuItems

seando-adsk:
#### MAYA-124088: USD breaks callback registration for addRMBBakingMenuItems
* Make sure to set parent menu after creating submenu in callback.


plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -713,6 +721,10 @@ global proc mayaUsdMenu_markingMenuCallback( string $obj )
             }
             menuItem -divider true;
             menuItem -label `getMayaUsdString("kMayaRefDuplicateAsUsdDataOptions")` -command ("mayaUsdMenu_duplicateAsUsdDataOptions " + $obj);
+
+            // Important: don't forget to reset the menu parent back to the context menu otherwise
+            //            another menu hook would add items to our menu.
+            setParent -menu ..;
```
seando-adsk, line 727: This is the actual reported bug. On line 717 we create a submenu which is set at the new parent menu (for subsequent menuItem calls). Before exiting we need to restore the menu parent (up one level) so if another callback is registered it will have the correct context menu as the parent.

PR #2462, opened 2022-07-06
Support Maya reference prim transform round trip.

ppt-adsk:


lib/usd/translators/mayaReferenceUpdater.cpp
```diff
@@ -205,30 +205,48 @@ UsdMayaPrimUpdater::PushCopySpecs PxrUsdTranslators_MayaReferenceUpdater::pushCo
     (*dstEditRouter)(routerContext, routingData);
 
     // Retrieve the destination layer and prim path from the routing data.
-    auto dstLayerStr = findValue(routingData, TfToken("layer"));
```
ppt-adsk, line 208: Renamed completely confusing "dst" prefixed variables.  The destination, passed in to this method as arguments, is the authored USD stage and its edit target layer.  The USD cache is a separate destination layer that gets referenced or payloaded in.

lib/usd/translators/mayaReferenceUpdater.cpp
```diff
@@ -205,30 +205,47 @@ UsdMayaPrimUpdater::PushCopySpecs PxrUsdTranslators_MayaReferenceUpdater::pushCo
...
+    auto srcPrim = srcStage->GetPrimAtPath(srcSdfPath);
+    if (TF_VERIFY(UsdGeomXformable(srcPrim))) {
+        auto dstPrim = dstStage->GetPrimAtPath(dstSdfPath);
+        if (TF_VERIFY(UsdGeomXformable(dstPrim))) {
+            // The Maya transform that corresponds to the Maya reference prim
+            // only has its transform attributes unlocked.  Bring any transform
+            // attribute edits over to the Maya reference prim.
+            MayaUsdUtils::MergePrimsOptions options;
+            options.ignoreUpperLayerOpinions = true;
+            MayaUsdUtils::mergePrims(
```
pierrebai-adsk, line 230: mergePrims returns a boolean for success. We might at least want to warn on failure. (IDK if the failure cases already log something or not, it might depend on the cause of failure.)

pierrebai-adsk (changes requested):
Great work, only minor adjustments needed.

pierrebai-adsk (approved):

PR #2463, opened 2022-07-06
MAYA-124253 - Rename/Reparent a prim and still have it on its Display…

seando-adsk:
#### MAYA-124253 - Rename/Reparent a prim and still have it on its Display Layer
* Update display layer unit test with more tests.


ppt-adsk (approved):
Nice test!

PR #2472, opened 2022-07-11
Fix attribute creation on connect

JGamache-autodesk:
- When unauthored attributes get connected, they must become authored.
 - When Material outputs are connected, we need to check the render context of the source node and use the right render context

### DIFF
--- a/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
@@ -141,6 +141,10 @@ bool UsdConnectionHandler::createConnection(
                 PXR_NS::SdrShaderNodeConstPtr srcShaderNodeDef
                     = registry.GetShaderNodeByIdentifier(srcInfoId);
                 if (!srcShaderNodeDef) {
+                    TF_RUNTIME_ERROR(
+                        "Could not find node definition '%s' for node '%s'.",
+                        srcInfoId.GetText(),
+                        Ufe::PathString::string(srcAttr->sceneItem()->path()).c_str());
                     return false;
                 }
                 TfToken renderContext = srcShaderNodeDef->GetSourceType() == "glslfx"
--- a/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
@@ -22,6 +22,7 @@
 #include <mayaUsd/ufe/UsdSceneItem.h>
 
 #include <pxr/base/tf/diagnostic.h>
+#include <pxr/usd/sdr/registry.h>
 
 #include <ufe/pathString.h>
 #include <ufe/scene.h>
@@ -125,8 +126,36 @@ bool UsdConnectionHandler::createConnection(
                 = dstApi.CreateInput(dstBaseName, dstUsdAttr->usdAttributeType());
             return UsdShadeConnectableAPI::ConnectToSource(dstInput, srcOutput);
         } else {
-            UsdShadeOutput dstOutput
-                = dstApi.CreateOutput(dstBaseName, dstUsdAttr->usdAttributeType());
+            UsdShadeOutput   dstOutput;
+            UsdShadeMaterial dstMaterial(dstUsdAttr->usdPrim());
+            // Special case when connecting to Material outputs:
+            if (dstMaterial
+                && (dstBaseName == UsdShadeTokens->surface || dstBaseName == UsdShadeTokens->volume
+                    || dstBaseName == UsdShadeTokens->displacement)) {
+                // Create the required output based on the type of the shader node we are trying to
+                // connect:
+                UsdShadeShader       srcShader(srcUsdAttr->usdPrim());
+                PXR_NS::SdrRegistry& registry = PXR_NS::SdrRegistry::GetInstance();
+                PXR_NS::TfToken      srcInfoId;
+                srcShader.GetIdAttr().Get(&srcInfoId);
+                PXR_NS::SdrShaderNodeConstPtr srcShaderNodeDef
+                    = registry.GetShaderNodeByIdentifier(srcInfoId);
+                if (!srcShaderNodeDef) {
+                    return false;
+                }
+                TfToken renderContext = srcShaderNodeDef->GetSourceType() == "glslfx"
+                    ? UsdShadeTokens->universalRenderContext
+                    : srcShaderNodeDef->GetSourceType();
+                if (dstBaseName == UsdShadeTokens->surface) {
+                    dstOutput = dstMaterial.CreateSurfaceOutput(renderContext);
+                } else if (dstBaseName == UsdShadeTokens->volume) {
+                    dstOutput = dstMaterial.CreateVolumeOutput(renderContext);
+                } else {
+                    dstOutput = dstMaterial.CreateDisplacementOutput(renderContext);
+                }
+            } else {
+                dstOutput = dstApi.CreateOutput(dstBaseName, dstUsdAttr->usdAttributeType());
+            }
             return UsdShadeConnectableAPI::ConnectToSource(dstOutput, srcOutput);
         }
     }
--- a/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
@@ -96,6 +96,12 @@ bool UsdConnectionHandler::createConnection(
         return false;
     }
 
+    // Use the UsdShadeConnectableAPI to create the connections and attributes to make sure the USD
+    // data model ends up in the right state.
+    //
+    // Using lower level APIs, like UsdPrim::CreateAttribute() tend leave the attributes marked as
+    // being custom instead of native.
+
     UsdShadeConnectableAPI srcApi(srcUsdAttr->usdPrim());
     TfToken                srcBaseName;
     UsdShadeAttributeType  srcAttrType;
--- a/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
@@ -33,22 +33,21 @@ namespace ufe {
 
 namespace {
 
-PXR_NS::UsdAttribute usdAttrFromUfeAttr(const Ufe::Attribute::Ptr& attr)
+UsdAttribute* usdAttrFromUfeAttr(const Ufe::Attribute::Ptr& attr)
 {
     if (!attr) {
         TF_RUNTIME_ERROR("Invalid attribute.");
-        return PXR_NS::UsdAttribute();
+        return nullptr;
     }
 
     if (attr->sceneItem()->runTimeId() != getUsdRunTimeId()) {
         TF_RUNTIME_ERROR(
             "Invalid runtime identifier for the attribute '" + attr->name() + "' in the node '"
             + Ufe::PathString::string(attr->sceneItem()->path()) + "'.");
-        return PXR_NS::UsdAttribute();
+        return nullptr;
     }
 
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/ufe/UsdConnectionHandler.cpp
```diff
@@ -88,30 +88,96 @@ bool UsdConnectionHandler::createConnection(
...
     const Ufe::Attribute::Ptr& dstAttr) const
 {
-    PXR_NS::UsdAttribute srcUsdAttr = usdAttrFromUfeAttr(srcAttr);
-    PXR_NS::UsdAttribute dstUsdAttr = usdAttrFromUfeAttr(dstAttr);
+    UsdAttribute* srcUsdAttr = usdAttrFromUfeAttr(srcAttr);
+    UsdAttribute* dstUsdAttr = usdAttrFromUfeAttr(dstAttr);
 
-    if (!isConnected(srcUsdAttr, dstUsdAttr)) {
-        return dstUsdAttr.AddConnection(srcUsdAttr.GetPath());
+    if (!srcUsdAttr || !dstUsdAttr
```
hodoulp, line 94: if `srcUsdAttr` or `dstUsdAttr` is null then the code should report an error e.g. `TF_RUNTIME_ERROR`

lib/mayaUsd/ufe/UsdConnectionHandler.cpp
```diff
@@ -88,30 +88,96 @@ bool UsdConnectionHandler::createConnection(
...
+                // Create the required output based on the type of the shader node we are trying to
+                // connect:
+                UsdShadeShader       srcShader(srcUsdAttr->usdPrim());
+                PXR_NS::SdrRegistry& registry = PXR_NS::SdrRegistry::GetInstance();
+                PXR_NS::TfToken      srcInfoId;
+                srcShader.GetIdAttr().Get(&srcInfoId);
+                PXR_NS::SdrShaderNodeConstPtr srcShaderNodeDef
+                    = registry.GetShaderNodeByIdentifier(srcInfoId);
+                if (!srcShaderNodeDef) {
+                    return false;
```
hodoulp, line 144: if that an error code path, the code must report an error e.g. TF_RUNTIME_ERROR

lib/mayaUsd/ufe/UsdConnectionHandler.cpp
```diff
@@ -88,30 +88,96 @@ bool UsdConnectionHandler::createConnection(
...
+            UsdShadeInput dstInput
+                = dstApi.CreateInput(dstBaseName, dstUsdAttr->usdAttributeType());
+            return UsdShadeConnectableAPI::ConnectToSource(dstInput, srcOutput);
+        } else {
+            UsdShadeOutput   dstOutput;
+            UsdShadeMaterial dstMaterial(dstUsdAttr->usdPrim());
+            // Special case when connecting to Material outputs:
+            if (dstMaterial
+                && (dstBaseName == UsdShadeTokens->surface || dstBaseName == UsdShadeTokens->volume
+                    || dstBaseName == UsdShadeTokens->displacement)) {
```
hodoulp, line 134: Can that list of `materials` change (e.g. new material type with MatX, Arnold, etc.)?

lib/mayaUsd/ufe/UsdConnectionHandler.cpp
```diff
@@ -88,30 +88,96 @@ bool UsdConnectionHandler::createConnection(
...
+        if (dstAttrType == UsdShadeAttributeType::Input) {
+            UsdShadeInput dstInput
+                = dstApi.CreateInput(dstBaseName, dstUsdAttr->usdAttributeType());
+            return UsdShadeConnectableAPI::ConnectToSource(dstInput, srcOutput);
+        } else {
+            UsdShadeOutput   dstOutput;
+            UsdShadeMaterial dstMaterial(dstUsdAttr->usdPrim());
+            // Special case when connecting to Material outputs:
+            if (dstMaterial
+                && (dstBaseName == UsdShadeTokens->surface || dstBaseName == UsdShadeTokens->volume
```
hodoulp, line 133: [Question] What is that global variable `UsdShadeTokens` ?

lib/mayaUsd/ufe/UsdConnectionHandler.cpp
```diff
@@ -88,30 +88,96 @@ bool UsdConnectionHandler::createConnection(
...
+                    dstOutput = dstMaterial.CreateSurfaceOutput(renderContext);
+                } else if (dstBaseName == UsdShadeTokens->volume) {
+                    dstOutput = dstMaterial.CreateVolumeOutput(renderContext);
+                } else {
+                    dstOutput = dstMaterial.CreateDisplacementOutput(renderContext);
+                }
+            } else {
+                dstOutput = dstApi.CreateOutput(dstBaseName, dstUsdAttr->usdAttributeType());
+            }
+            return UsdShadeConnectableAPI::ConnectToSource(dstOutput, srcOutput);
```
hodoulp, line 159: [Question] The previous code was using `dstUsdAttr.AddConnection()`, and the new implementation uses `UsdShadeConnectableAPI::ConnectToSource()`, why?

lib/mayaUsd/ufe/UsdConnectionHandler.cpp
```diff
@@ -88,30 +88,96 @@ bool UsdConnectionHandler::createConnection(
...
     const Ufe::Attribute::Ptr& dstAttr) const
 {
-    PXR_NS::UsdAttribute srcUsdAttr = usdAttrFromUfeAttr(srcAttr);
-    PXR_NS::UsdAttribute dstUsdAttr = usdAttrFromUfeAttr(dstAttr);
+    UsdAttribute* srcUsdAttr = usdAttrFromUfeAttr(srcAttr);
+    UsdAttribute* dstUsdAttr = usdAttrFromUfeAttr(dstAttr);
 
-    if (isConnected(srcUsdAttr, dstUsdAttr)) {
-        return dstUsdAttr.RemoveConnection(srcUsdAttr.GetPath());
+    if (!srcUsdAttr || !dstUsdAttr
```
hodoulp, line 174: if srcUsdAttr or dstUsdAttr is null then the code should report an error e.g. TF_RUNTIME_ERROR

test/lib/ufe/testConnections.py
```diff
@@ -421,5 +423,99 @@ def testConnectionsHandler(self):
...
+        #
+        shaderOutput = shaderAttrs.attribute("outputs:out")
+        materialOutput = materialAttrs.attribute("outputs:surface")
+
+        connectionHandler = ufe.RunTimeMgr.instance().connectionHandler(materialItem.runTimeId())
+
+        # The attributes are not created yet:
+        self.assertEqual(len(materialPrim.GetAuthoredProperties()), 0)
+        self.assertEqual(len(shaderPrim.GetAuthoredProperties()), 1)
+        self.assertIn("info:id", [i.GetName() for i in shaderPrim.GetAuthoredProperties()])
```
hodoulp, line 483: As the code previously checks that there is only one entry, this line can directly access the first (and only) entry i.e. avoid the loop.

test/lib/ufe/testConnections.py
```diff
@@ -421,5 +423,99 @@ def testConnectionsHandler(self):
...
+
+        connectionHandler.connect(shaderOutput, materialOutput)
```
hodoulp, line 485: The code must check the return code.

test/lib/ufe/testConnections.py
```diff
@@ -421,5 +423,99 @@ def testConnectionsHandler(self):
...
+        self.assertEqual(len(conns), 0)
+
+        # Not redirected since already on outputs:mtlx:surface:
+        connectionHandler.connect(shaderOutput, materialOutput)
+
+        connections = connectionHandler.sourceConnections(materialItem)
+        self.assertIsNotNone(connections)
+        conns = connections.allConnections()
+        self.assertEqual(len(conns), 1)
+
```
hodoulp, line 515: The test can now create a connection i.e. it should fail because the connection already exists.
Note that the same check can be done with the delete connection.

test/lib/ufe/testConnections.py
```diff
@@ -421,5 +423,99 @@ def testConnectionsHandler(self):
...
+        # The attributes directly used by the connection got created:
+        self.assertEqual(len(shaderPrim.GetAuthoredProperties()), 2)
+        self.assertIn("info:id", [i.GetName() for i in shaderPrim.GetAuthoredProperties()])
+        self.assertIn("outputs:out", [i.GetName() for i in shaderPrim.GetAuthoredProperties()])
+        # The connection of a MaterialX shader to the material got redirected to the proper render context.
+        self.assertEqual(len(materialPrim.GetAuthoredProperties()), 1)
+        self.assertEqual(materialPrim.GetAuthoredProperties()[0].GetName(), "outputs:mtlx:surface")
+
+        materialOutput = materialAttrs.attribute("outputs:mtlx:surface")
+        connectionHandler.disconnect(shaderOutput, materialOutput)
```
hodoulp, line 501: The code must check the return code.

test/lib/ufe/testConnections.py
```diff
@@ -421,5 +423,99 @@ def testConnectionsHandler(self):
...
+
+        connections = connectionHandler.sourceConnections(materialItem)
+        self.assertIsNotNone(connections)
+        conns = connections.allConnections()
+        self.assertEqual(len(conns), 0)
+
+        # Not redirected since already on outputs:mtlx:surface:
+        connectionHandler.connect(shaderOutput, materialOutput)
```
hodoulp, line 509: The code must check the return code.

hodoulp (changes requested):
Few comments and questions.

test/lib/ufe/testConnections.py
JGamache-autodesk: Can't do that. Function returns void/None, as defined in UFE:
```
class UFE_SDK_DECL ConnectionHandler
{
public:
    virtual void connect(
        const AttributeInfo &srcInfo, const AttributeInfo &dstInfo) const;
```

test/lib/ufe/testConnections.py
JGamache-autodesk: Same issue, disconnect return void/None.

lib/mayaUsd/ufe/UsdConnectionHandler.cpp
ppt-adsk: I think you should convert this review comment into a code comment.

lib/mayaUsd/ufe/UsdConnectionHandler.cpp
hodoulp: Discussed, and the answer is no.

test/lib/ufe/testConnections.py
```diff
@@ -421,5 +423,99 @@ def testConnectionsHandler(self):
...
+        import mayaUsd_createStageWithNewLayer
+        proxyShape = mayaUsd_createStageWithNewLayer.createStageWithNewLayer()
+
+        # Create a ContextOps interface for the proxy shape.
+        proxyPathSegment = mayaUtils.createUfePathSegment(proxyShape)
+        proxyShapePath = ufe.Path([proxyPathSegment])
+        proxyShapeItem = ufe.Hierarchy.createItem(proxyShapePath)
+        contextOps = ufe.ContextOps.contextOps(proxyShapeItem)
+
+        cmd = contextOps.doOpCmd(['Add New Prim', 'Capsule'])
```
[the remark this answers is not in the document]
hodoulp: Suggestion: This unit test purpose is not to test 'new prim' so, the code can call contextOps.doOp() directly.

hodoulp (approved):

ppt-adsk (approved):

PR #2478, opened 2022-07-12
LOOKDEVX-754 | Adds a nullptr check to UsdHierachyHandler::hierarchy(…

feldstj:
…...)

### DIFF
--- a/lib/mayaUsd/ufe/UsdHierarchyHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdHierarchyHandler.cpp
@@ -18,6 +18,8 @@
 #include <mayaUsd/ufe/UsdSceneItem.h>
 #include <mayaUsd/ufe/Utils.h>
 
+#include <pxr/base/tf/diagnostic.h>
+
 namespace MAYAUSD_NS_DEF {
 namespace ufe {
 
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/UsdHierarchyHandler.cpp
```diff
@@ -41,6 +41,9 @@ UsdHierarchyHandler::Ptr UsdHierarchyHandler::create()
 Ufe::Hierarchy::Ptr UsdHierarchyHandler::hierarchy(const Ufe::SceneItem::Ptr& item) const
 {
     UsdSceneItem::Ptr usdItem = std::dynamic_pointer_cast<UsdSceneItem>(item);
```
ppt-adsk, line 43: I would do this differently.  If you get a nullptr argument, then returning a nullptr hierarchy interface is fine.

However, if you get a non-null argument, and the dynamic cast fails, that's an error.  You should
```
if (!item || !TF_VERIFY(usdItem)) {
   return nullptr;
} 
```

ppt-adsk (changes requested):

ppt-adsk (approved):

PR #2491, opened 2022-07-19
MAYA-124207 prevent skeleton moving when underlying base mesh is moved

pierrebai-adsk:
When the base mesh of a skeleton is moved in Maya using the manipulator, the skeleton that is using that mesh should not move. We enforce this by detecting that the points of the prim were computed.


lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -2447,7 +2453,18 @@ void HdVP2Mesh::_UpdatePrimvarSources(
             updatePrimvarInfo(
                 primvarName, cpuComputation->GetOutputByIndex(outputIndex), HdInterpolationVertex);
         }
+
+        // Records that points primvar is computed.
+        if (primvarName == HdTokens->points) {
```
pierrebai-adsk, line 2458: IDK if there is a smarter way to do this?

williamkrick (approved):

vlasovi (approved):
Looks good

PR #2492, opened 2022-07-19
MAYA-124269 set a default prim when caching to USD

pierrebai-adsk:


ppt-adsk (changes requested):
Looks great, thanks for the fix and the quick turnaround!  It does need a test, though.

pierrebai-adsk:
Ah, yes, I was about to add check in an existing cache-to-USD test to see if a default prim is set.

ppt-adsk (approved):
Thanks!

PR #2495, opened 2022-07-20
MAYA-124458 - When caching to USD, 'Prim Name' is disabled when Define in Variant is selected

seando-adsk:
Removed the code which disabled this field. Just leave the field enabled all the time.


lib/mayaUsd/resources/scripts/mayaUsdCacheMayaReference.py
```diff
@@ -169,7 +169,6 @@ def variantOrNewPrim(buttonChecked):
     variantSelected = cmds.radioButtonGrp('variantRadioButton', query=True, select=1)
     cmds.rowLayout('usdCacheVariantSetRow', edit=True, enable=variantSelected)
     cmds.rowLayout('usdCacheVariantNameRow', edit=True, enable=variantSelected)
-    cmds.textFieldGrp('primNameText', edit=True, enable=not variantSelected)
```
seando-adsk, line 172: According to the JIRA item there was a mistake in the original design. This field should be enabled all the time.

boudrey (approved):

seando-adsk:
No preflight needed as the change is to a single python file that is not part of the build or any unit test.

PR #2497, opened 2022-07-21
MAYA-124418 copy transform to the correct variant.

pierrebai-adsk:
- Need to find and temporarily switch to the variant that contains the Maya Reference.
- Since variant set and variant name are under the user control, we need an explicit search.
- Copy only the transform and properly handle the case of the source and destination not being variants.

Second version added:
- Move the copying of the Maya Reference transform to the edit router.
- Go back to using mergePrim to do the work.
- Properly handle the case teh destination is in a variant.
- Cleanup the edit router code by refactor common code and common code patterns.
- Read all expected context arguments in one go, makes it easeier to understand what argument are expected.
- Move inner lambda function to a separate function for code clarity.
- Avoiding having such a function within a function makes the code more linear and easier to read.

### DIFF
--- a/lib/usd/translators/mayaReferenceUpdater.cpp
+++ b/lib/usd/translators/mayaReferenceUpdater.cpp
@@ -324,7 +324,7 @@ UsdMayaPrimUpdater::PushCopySpecs PxrUsdTranslators_MayaReferenceUpdater::pushCo
     // Maya reference prim.  The Maya transform node changes have already been
     // exported into the temporary layer as a transform prim, which is our
     // source.  The destination prim in the stage is the Maya reference prim.
-    auto srcPrim = srcStage->GetPrimAtPath(srcSdfPath);
+    auto             srcPrim = srcStage->GetPrimAtPath(srcSdfPath);
     UsdGeomXformable srcXform(srcPrim);
     if (TF_VERIFY(srcXform)) {
         std::unique_ptr<AutoVariantRestore> variantRestore;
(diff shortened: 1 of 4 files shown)


pierrebai-adsk:
pr-build!

lib/usd/translators/mayaReferenceUpdater.cpp
```diff
@@ -177,6 +176,19 @@ bool PxrUsdTranslators_MayaReferenceUpdater::editAsMaya()
     return true;
 }
 
+bool callEditRouter(
```
pierrebai-adsk, line 179: Will be useful if we change the edit router API to return a success/failure value or if we allow explicitly to throw exception on error, we will be able to catch the exception here and convert it to a success/failure value.

lib/usd/translators/mayaReferenceUpdater.cpp
```diff
@@ -191,67 +203,53 @@ UsdMayaPrimUpdater::PushCopySpecs PxrUsdTranslators_MayaReferenceUpdater::pushCo
...
 
-    // Use the edit router to find the destination layer and path.
-    auto dstEditRouter = MayaUsd::getEditRouter(TfToken("mayaReferencePush"));
-    if (!dstEditRouter) {
-        return PushCopySpecs::Failed;
-    }
-
     PXR_NS::VtDictionary routerContext = getContext()->GetUserArgs();
+
+    // Pass the source and destination stage, layer and path to routers.
```
pierrebai-adsk, line 208: These values are required for the copy-transform code. In general, an edit router might want to know these detail to make decision where to route.

ppt-adsk (approved):
Nice cleanup, thanks for the fix.

PR #2500, opened 2022-07-21
Changing Pxr image baseline

dj-mcg:
This baseline changed slightly after enabling light/materials features
in our environment


dj-mcg:
Hey folks - looks like the remote build is timing out. Is there something I need to do on my end?

seando-adsk:
@dj-mcg We had problems yesterday with our internal build system. Should be fixed now. I've relaunched the preflight for you.

seando-adsk (approved):

PR #2503, opened 2022-07-26
Undo of delete must update stage cache.

ppt-adsk:


pierrebai-adsk (approved):

lib/mayaUsd/ufe/UsdStageMap.cpp
```diff
@@ -83,6 +86,16 @@ inline Ufe::Path::Segments::size_type nbPathSegments(const Ufe::Path& path)
 #endif
 }
 
+inline Ufe::Path toPath(const std::string& pathString)
```
pierrebai-adsk, line 89: I almost feel like we should put this in some util function, it has come up a few times in the past... I know I've been bitten a lot by it in unit test (but that is in Python, irrelevant here.)

pierrebai-adsk (approved):

pierrebai-adsk:
Assigned the Jira ticket to you so you can resolve it to QA when the tests pass.

test/lib/mayaUsd/nodes/testProxyShapeBase.py
```diff
@@ -147,6 +147,53 @@ def testDeleteStage(self):
         # Request the now dead proxy shape. This should not crash.
         result = handler.findGatewayItems(proxyShapePath)
 
+    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() >= 2, 'testDeleteStageUndo uses PathString which is only available in UFE v2 or greater.')
```
ppt-adsk, line 150: Thanks to @pierrebai-adsk for the test.

pierrebai-adsk (approved):

PR #2508, opened 2022-07-27
MAYA-123143 - As a user, I'd like my prims to be templated in a Display Layer

vlasovi:
- Support for display layer's 'template' attribute
- Update render prims on color space change + proper color space conversion for "lead" selection color
- Fixing rendering issues in points


seando-adsk (approved):
Not my area of expertise, but the changes look fine to me.

PR #2509, opened 2022-07-27
MAYA-124578 - Hide on playback option doesn't work on re-open

vlasovi:
- Set HideOnPlayback flag to all new render items
- Minor refactoring


seando-adsk (approved):
Nice clean changes.

PR #2519, opened 2022-08-02
MAYA-123160 - As a user, I'd like my prims to be referenced in a Display Layer

vlasovi:
Support for 'reference' attribute of display layers


seando-adsk (approved):
Not my area of expertise, but the changes look good to me.

PR #2523, opened 2022-08-03
MAYA-108733 fix removing multiple layers at the same time

pierrebai-adsk:
- Handle adjusting indexes.
- Added unit test


pierrebai-adsk:
@seando-adsk please review when convenient.

lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -694,15 +729,23 @@ MStatus LayerEditorCommand::parseArgs(const MArgList& argList)
...
             auto count = argParser.numberOfFlagUses(kInsertSubPathFlag);
             for (unsigned i = 0; i < count; i++) {
                 auto cmd = std::make_shared<Impl::InsertSubPath>();
 
                 MArgList listOfArgs;
                 argParser.getFlagArgumentList(kInsertSubPathFlag, i, listOfArgs);
-                cmd->_index = listOfArgs.asInt(0);
+
+                const int originalIndex = listOfArgs.asInt(0);
+                const int adjustedIndex = indexAdjustments.getAdjustedIndex(originalIndex);
```
ppt-adsk, line 743: The methods on IndexAdjustments are fine for what they do, but the client always needs to call two methods, getAdjustedIndex(), then addInsertionAdjustment() or addRemovalAdjustment(), and both have the same argument (originalIndex).  Wouldn't it be nice to add a couple of additional methods to make the client's job less error-prone and simpler:

auto originalIndex = listOfAgs.asInt(0);
auto adjustedIndex = indexAdjustments.insertionAdjustment(originalIndex);

?  And similarly for removalAdjustment().  These two new convenience methods would simply call getAdjustedIndex() and addInsertionAdjustment() / addRemovalAdjustment().

ppt-adsk (changes requested):

ppt-adsk (approved):

lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -630,6 +630,59 @@ class MuteLayer : public BaseCmd
     PXR_NS::SdfLayerRefPtr _mutedLayer;
 };
 
+// We assume the indexes given to the command are the original indexes
+// of the layers. Since each command is executed individually and in
+// order, each one may affect the index of subsequent commands. We
+// records adjustements that must be applied to indexes in the map.
+// Removal of a layer creates a negative adjustment, inserttion of a
```
seando-adsk, line 637: Minor typo "insertion"

lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -630,6 +630,59 @@ class MuteLayer : public BaseCmd
...
+        // Apply all adjustment that were done on indexes lower or
+        // equal to the input index.
+        int adjustedIndex = index;
+        for (const auto& indexAndAdjustement : _indexAdjustments) {
+            if (indexAndAdjustement.first > index)
+                break;
+            adjustedIndex += indexAndAdjustement.second;
+        }
+        return adjustedIndex;
+    }
```
seando-adsk, line 662: I wonder if these 3 methods should be private and a caller should only have access to the two methods below. What do you two think?

ppt-adsk: Would be less error-prone, I agree.

test/lib/testMayaUsdLayerEditorCommands.py
```diff
@@ -362,6 +362,40 @@ def setAndRemoveEditTargetRecursive(layer, index, editLayer):
...
+
+        verifySubLayers(layerNames)
+
+        # Remove layers 0 and 1 and verify that the correct layers have been removed.
+        #
+        # Note: passing the same argument multiple times does not work in Python, so we call MEL.
+        #
+        # Note: when inserted in USD, layers are added at the top, so layer 0 and 1 are
+        #       layer5 and layer6.
+        mel.eval("""mayaUsdLayerEditor -removeSubPath 0 "%s" -removeSubPath 1 "%s" "%s";""" % (shapePath, shapePath, rootLayer.identifier))
```
seando-adsk, line 396: To call this directly from python try:
```suggestion
        cmds.mayaUsdLayerEditor(rootLayer.identifier, removeSubPath=((0, shapePath), (1, shapePath)))
```

seando-adsk (approved):

PR #2529, opened 2022-08-04
Flush all resource caches on exit

JGamache-autodesk:
This makes sure textures and shaders are freed before VP2 shuts down.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -125,7 +125,54 @@ namespace {
 // If the textures would have been requested to reload in `ApplyPendingUpdates()`,
 // we could still reuse the loaded one from cache, otherwise the idle task can
 // safely release the texture.
-std::unordered_set<HdVP2TextureInfoSharedPtr> pendingRemovalTextures;
+class _TransientTexturePreserver
+{
+public:
+    static _TransientTexturePreserver& GetInstance()
+    {
+        static _TransientTexturePreserver sInstance;
+        return sInstance;
+    }
+
+    void PreserveTextures(HdVP2LocalTextureMap& localTextureMap)
+    {
+        if (_isExiting) {
+            return;
+        }
+
+        // Locking to avoid race condition for insertion to pendingRemovalTextures
+        std::lock_guard<std::mutex> lock(_removalTaskMutex);
+
+        // Avoid creating multiple idle tasks if there is already one
+        bool hasRemovalTask = !_pendingRemovalTextures.empty();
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -116,6 +117,16 @@ static const std::size_t kRefreshDuration { 1000 };
...
 
+// USD `UsdImagingDelegate::ApplyPendingUpdates()` would request to
+// remove the material then recreate, this is causing texture disappearing
+// when user manipulating a prim (while holding mouse buttion).
+// We hold a copy of the texture info reference, so that the texture will not
+// get released immediately along with material removal.
+// If the textures would have been requested to reload in `ApplyPendingUpdates()`,
+// we could still reuse the loaded one from cache, otherwise the idle task can
+// safely release the texture.
+std::unordered_set<HdVP2TextureInfoSharedPtr> pendingRemovalTextures;
```
JGamache-autodesk, line 128: Moved from the destructor so it can be accessed by OnMayaExit.

seando-adsk (dismissed):

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1734,18 +1745,8 @@ HdVP2Material::~HdVP2Material()
...
-    // We hold a copy of the texture info reference, so that the texture will not
-    // get released immediately along with material removal.
-    // If the textures would have been requested to reload in `ApplyPendingUpdates()`,
-    // we could still reuse the loaded one from cache, otherwise the idle task can
-    // safely release the texture.
-    static std::unordered_set<HdVP2TextureInfoSharedPtr> pendingRemovalTextures;
-    static std::mutex                                    removalTaskMutex;
-
     if (!_IsDisabledAsyncTextureLoading() && !_localTextureMap.empty()) {
+        std::mutex removalTaskMutex;
```
williamkrick, line 1749: I thought that this mutex needs to be static so that all HdVP2Material will share the same mutex? With this version of the code the lock_guard does nothing and we could have unsafe parallel modification of pendingRemovalTextures.

Can we go back to the original changes and see if we can figure out the workflow that requires the lock?

williamkrick (changes requested):

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
JGamache-autodesk: Agreed. The mutex and the structure it protects should ideally be kept together.

williamkrick (approved):

Issue #2085, opened 2022-02-10
[MAYA-121636] As artists, we would like to be able to parent prims underneath 'scopes'


paulbaa:
Currently, we are not able to parent existing prims underneath Usd-scopes in Maya. This is desired behaviour, so we can author our existing scenography with scopes at any point. Creating new prims underneath a scope works.

Open Maya 2022.3
Create a new empty USD stage
Create a scope
Create a (sibling) xform
parent the xform underneath the scope, either by middle-mouse drag, 'P' or edit >parent
ERROR: 'Object ... does not support parenting and therefore cannot be a parent.'

As per the conversation with @kxl-adsk 

> The culprit, on our end, is that the scope prim is not transformable. The new child would like to have the parent transform (probably to position itself correctly in world space) and because the scope has no 3D transform, it refuses to be parented under it. The default for the Maya parent command is parent -absolute, which preserves the world space position of objects by changing their local transformation (parent -relative preserves the local transform). Since a scope doesn't have a 3D transform, we error out.

ppt-adsk:
Hi @paulbaa, we are working on this issue, and have a question for you.  What do you expect the behavior to be if we call parent -absolute (the default) on a USD child hierarchy that has a scope as its root?  The scope transform cannot be changed.  Two possibilities are:

1. Simply add the child hierarchy to its new parent, and don't worry about fixing up the child hierarchy world space position.  This would be the equivalent of parent -relative.
2. Traverse the child hierarchy looking for prims that are transformable, and transform the highest-level prims we find in the hierarchy.

Option (2) is the most correct, but also the most complex to code and test, and has potential scalability problems for large child hierarchies.  Therefore, we're wondering if it's a real use case, and the simpler option (1) would be a fine practical solution.  Wouldn't you simply author a child hierarchy with a transform at its root if you wanted its world space position to be maintained by parent -absolute, rather than have a scope at the root of the child hierarchy?  Thanks for any info!

paulbaa:
Apologies for the late reply with holidays, Siggraph and thanks for looking into this. 

I'd agree with 1) - in most/all cases our assets need to have a transform as it's root, to be positioned, parented etc. I could imagine if a scope is the root, you'd just get a warning saying 'root of asset is a scope and can't be transformed. Parenting relative' 

I also think this has the benefit of staying explicit, i.e. you don't get a possibly arbitrary transform down the child's hierarchy.

tanant:
[...]

PR #2530, opened 2022-08-04
MAYA121636 handle scope for read-only transform

pierrebai-adsk:
- Implement Transform3dRead for scope prim.
- Add a Transform3dHandler that creates that implementation for scopes.
- Add unit tests.


pierrebai-adsk:
Implementation of the UFE interface that went in UFE 0.4.25. (Maya is now at 0.4.30.)

lib/mayaUsd/ufe/UsdTransform3dRead.cpp
```diff
@@ -0,0 +1,128 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "UsdTransform3dRead.h"
+
+#include <mayaUsd/ufe/Utils.h>
+
+#include <pxr/usd/usdGeom/scope.h>
+#include <pxr/usd/usdGeom/xformCache.h>
+#include <pxr/usd/usdGeom/xformCommonAPI.h>
```
ppt-adsk, line 22: Unused, remove.

lib/mayaUsd/ufe/UsdTransform3dRead.h
```diff
@@ -0,0 +1,93 @@
...
+
+namespace MAYAUSD_NS_DEF {
+namespace ufe {
+
+//! \brief Read-only implementation for USD object 3D transform information.
+//
+// Note that all calls to specify time use the default time, but this
+// could be changed to use the current time, using getTime(path()).
+
+class MAYAUSD_CORE_PUBLIC UsdTransform3dRead : public Ufe::Transform3dRead
```
ppt-adsk, line 34: Is there any way we can convert this into a template, so that UsdTransform3dBase could publicly derive from
UsdTransform3dReadT<Ufe::Transform3d>, and share some code ?  You can then simply typedef UsdTransform3dReadT<Ufe::Transform3dRead> UsdTransform3dRead;
I realize that UsdTransform3dBase can't simply derive from UsdTransform3dRead, because it needs the full interface of Ufe::Transform3d, but using a template to remove code duplication might be nice.

lib/mayaUsd/ufe/UsdTransform3dRead.h
```diff
@@ -0,0 +1,93 @@
...
+    UsdSceneItem::Ptr fItem;
+    PXR_NS::UsdPrim   fPrim;
+
+}; // UsdTransform3dRead
+
+//! \brief Factory to create a UsdTransform3dRead interface object.
+class MAYAUSD_CORE_PUBLIC UsdTransform3dReadHandler : public Ufe::Transform3dHandler
+{
+public:
+    using Ptr = std::shared_ptr<UsdTransform3dReadHandler>;
```
ppt-adsk, line 73: Is this typedef really needed / used?

ppt-adsk (changes requested):
Can have a look at the latest discussion in the original issue:
https://github.com/Autodesk/maya-usd/issues/2085

lib/mayaUsd/ufe/UsdTransform3dRead.h
pierrebai-adsk: All classes providing the create() function declare their return value as a typedef. I guess it makes the code more consistent to not have to know the precise type. For example: MayaUIInfoHandler.h, ProxyShapeHierarchy.h, ProxyShapeHierarchyHandler.h, etc.

ppt-adsk: O.K., makes sense.

lib/mayaUsd/ufe/UsdTransform3dRead.h
pierrebai-adsk: I will instead write a non-template implementation class and use it in both? That will avoid having code in the header.

ppt-adsk: Up to you whichever is cleanest.  You can also do as per
https://git.autodesk.com/media-and-entertainment/ufe/blob/master/include/trie.imp.h
and put the template implementation into a separate file, only to be included by those files than instantiate the template.

lib/mayaUsd/ufe/UsdTransform3dRead.cpp
pierrebai-adsk: scope is actually used, the others can go.

lib/mayaUsd/ufe/UsdTransform3dRead.h
```diff
@@ -31,7 +32,9 @@ namespace ufe {
 // Note that all calls to specify time use the default time, but this
 // could be changed to use the current time, using getTime(path()).
 
-class MAYAUSD_CORE_PUBLIC UsdTransform3dRead : public Ufe::Transform3dRead
+class MAYAUSD_CORE_PUBLIC UsdTransform3dRead
+    : public UsdTransform3dReadImpl
```
ppt-adsk, line 36: Shouldn't we be inheriting privately from UsdTransform3dReadImpl?  It's just a code sharing technique, there is no class hierarchy involved here.  Also, I worry about ambiguity if there are identically-named virtual methods in Ufe::Transform3dRead and UsdTransform3dReadImpl.

lib/mayaUsd/ufe/UsdTransform3dReadImpl.h
```diff
@@ -0,0 +1,67 @@
...
+
+#ifdef UFE_V2_FEATURES_AVAILABLE
+    Ufe::Matrix4d matrix() const;
+#endif
+
+    Ufe::Matrix4d segmentInclusiveMatrix() const;
+    Ufe::Matrix4d segmentExclusiveMatrix() const;
+
+private:
+    UsdSceneItem::Ptr fItem;
```
ppt-adsk, line 61: Can this one be
const UsdSceneItem::Ptr
?  It's set on construction and never changed, right?

ppt-adsk (commented):
Looks good, thanks for making this change, just the private inheritance I'm wondering about.

test/lib/ufe/testParentCmd.py
```diff
@@ -1249,5 +1249,250 @@ def firstSubLayer(context, routingData):
...
+        cmds.redo()
+
+        checkParentDone()
+        checkCapsuleMatrix('/Xform1/Scope1/Capsule1')
+
+    @unittest.skipUnless(os.getenv('UFE_PREVIEW_VERSION_NUM', '0000') >= '4025', 'Require Transform3dRead interface in Ufe 0.4.25')
+    def testParentAbsoluteScope(self):
+        """
+        Test parent -absolute to move a scope with a prim under it under an xform.
+        Since tghe scoe is not transformable, the prim will move.
```
ppt-adsk, line 1377: "tghe" --> "the"

ppt-adsk (approved):

ppt-adsk (changes requested):
On second thought, I think we need an addition and a change to the tests:

1. Need to add a parent -relative test.  It's the trivial case, but it still needs a test :)
2. The new tests need to be protected by a ufeSupportFixLevel() decorator, so that they will only run if Maya has the proper support.  You can add the fixLevel in your Maya pull request.  See
https://github.com/Autodesk/maya-usd/blob/387a1fa6c11f8a7349d7ebe353e55382603e2347/test/lib/ufe/testGroupCmd.py#L926

ppt-adsk:
> > ```
> > 2. The new tests need to be protected by a ufeSupportFixLevel() decorator, so that they will only run if Maya has the proper support.  You can add the fixLevel in your Maya pull request.  See
> >    https://github.com/Autodesk/maya-usd/blob/387a1fa6c11f8a7349d7ebe353e55382603e2347/test/lib/ufe/testGroupCmd.py#L926
> > ```
> 
> I'm thinking a bit more about the UFE fix level: it seems that it does not work well with back-porting. We may not backport everything to an older version of Maya. so if we backport a fix at fix-level N, we may not have ported the fixes related to fix-level N-1, N-2, etc. It would be better if we the fix-level were instead a fix-features list. For now I'll add a fix level.

There is no fix level for backports.  Fix level only exists for master, to protect plugin code from execution using a Maya that is inappropriate.  For previous versions of Maya, you can assume the plugin code is already being executed using Maya master until you put in your fix into the previous version of Maya.  Once that's done, you can enable the test for that previous Maya version in the plugin.

The point of fix level is to protect against execution within the time scope of a Maya preview release, when no UFE version number is appropriate.  In such a case, you cannot rely on the Maya preview release number, or the Maya version, or the UFE version.

ppt-adsk (approved):

PR #2533, opened 2022-08-05
MAYA-113642 disallow switching variants in a weaker layer.

pierrebai-adsk:
Also fix a bug in the allow check when dealing with variant: it was refusing even when the allow-stronger-layer flag was on and in a stronger layer.


pierrebai-adsk:
The PF failure is only OSX master with clang crashing, unrelated, can ignore.

ppt-adsk (approved):

seando-adsk (approved):

PR #2534, opened 2022-08-05
Improve connecting to surface experience

JGamache-autodesk:
- If we can not find a shader source type, assume the user knows what he
  is doing.
- If trying to connect a node graph output to surface, dig into the node
  graph in case we find a shader source type by traversing the
  connections.

### DIFF
--- a/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
@@ -65,6 +65,43 @@ bool isConnected(const PXR_NS::UsdAttribute& srcUsdAttr, const PXR_NS::UsdAttrib
     return false;
 }
 
+PXR_NS::SdrShaderNodeConstPtr
+_GetShaderNodeDef(const PXR_NS::UsdPrim& prim, const PXR_NS::TfToken& attrName)
+{
+    UsdPrim           targetPrim = prim;
+    TfToken           targetName = attrName;
+    UsdShadeNodeGraph ngTarget(targetPrim);
+    while (ngTarget) {
+        // Dig inside, following the connection on targetName until we find a shader.
+        UsdShadeOutput graphOutput = ngTarget.GetOutput(targetName);
+        if (!graphOutput) {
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/UsdConnectionHandler.cpp
```diff
@@ -65,6 +65,43 @@ bool isConnected(const PXR_NS::UsdAttribute& srcUsdAttr, const PXR_NS::UsdAttrib
     return false;
 }
 
+PXR_NS::SdrShaderNodeConstPtr
+_GetShaderNodeDef(const PXR_NS::UsdPrim& prim, const PXR_NS::TfToken& attrName)
+{
+    UsdPrim           targetPrim = prim;
```
JGamache-autodesk, line 71: Traverse into layers of NodeGraphs until a shader is found. Similar to what we do when [traversing node graphs](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp#L669) in the material importer.

lib/mayaUsd/ufe/UsdConnectionHandler.cpp
```diff
@@ -140,20 +177,10 @@ bool UsdConnectionHandler::createConnection(
...
-                    TF_RUNTIME_ERROR(
-                        "Could not find node definition '%s' for node '%s'.",
-                        srcInfoId.GetText(),
-                        Ufe::PathString::string(srcAttr->sceneItem()->path()).c_str());
-                    return false;
-                }
-                TfToken renderContext = srcShaderNodeDef->GetSourceType() == "glslfx"
+                    = _GetShaderNodeDef(srcUsdAttr->usdPrim(), srcBaseName);
+                TfToken renderContext
+                    = !srcShaderNodeDef || srcShaderNodeDef->GetSourceType() == "glslfx"
```
JGamache-autodesk, line 183: If we do not find a shader node def, assume the user knows what he is doing and that connecting on the regular surface port is what he wants.

boudrey (approved):

PR #2538, opened 2022-08-05
Update callsite to use new Hydra API

dj-mcg:
Some Hydra API has changed recently. Updating callsites and version
guarding.


JGamache-autodesk (approved):
Makes sense. We are not using the info and USD no longer requires passing empty containers.

PR #2539, opened 2022-08-08
Disconnection code improvements

JGamache-autodesk:
- Cleanup empty connection array when last connection is undone
- Cleanup transient attributes that no longer have any value in the USD
  scene.
- Add notification handler to refresh AE when disconnecting ports

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
@@ -83,6 +83,25 @@ class UfeAttributesObserver(ufe.Observer):
         # Nothing needed here since we don't create any UI.
         pass
 
+class UfeConnectionChangedObserver(ufe.Observer):
+    def __init__(self, item):
+        super(UfeConnectionChangedObserver, self).__init__()
+        self._item = item
+
+    def __del__(self):
+        ufe.Attributes.removeObserver(self)
+
+    def __call__(self, notification):
+        if hasattr(ufe, "AttributeConnectionChanged") and isinstance(notification, ufe.AttributeConnectionChanged):
+            mel.eval("evalDeferred -low \"refreshEditorTemplates\";")
+
+    def onCreate(self, *args):
+        ufe.Attributes.addObserver(self._item, self)
+
+    def onReplace(self, *args):
+        # Nothing needed here since we don't create any UI.
+        pass
+
 class MetaDataCustomControl(object):
     # Custom control for all prim metadata we want to display.
     def __init__(self, item, prim, useNiceName):
@@ -513,6 +532,9 @@ class AETemplate(object):
         nodeDef = Sdr.Registry().GetShaderNodeByIdentifier(nodeId)
         if not nodeDef:
             return
+        # Add a custom control to monitor for connection changed.
+        cnxObs = UfeConnectionChangedObserver(self.item)
+        self.defineCustom(cnxObs)
         label = nodeDef.GetLabel()
         if not label:
             label = nodeDef.GetFamily()
--- a/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
@@ -186,8 +186,37 @@ bool UsdConnectionHandler::deleteConnection(
         return false;
     }
 
-    return UsdShadeConnectableAPI::DisconnectSource(
+    bool retVal = UsdShadeConnectableAPI::DisconnectSource(
         dstUsdAttr->usdAttribute(), srcUsdAttr->usdAttribute());
+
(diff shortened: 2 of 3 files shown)


lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -83,6 +83,25 @@ def onReplace(self, *args):
         # Nothing needed here since we don't create any UI.
         pass
 
+class UfeConnectionChangedObserver(ufe.Observer):
+    def __init__(self, item):
+        super(UfeConnectionChangedObserver, self).__init__()
+        self._item = item
+
+    def __del__(self):
```
ppt-adsk, line 91: Are you sure this is necessary?  Observers are kept by weak ptr by their Subject(s), so if the shared ptr goes away, they will be cleaned up by all Subject(s) they were added to.

lib/mayaUsd/ufe/UsdConnectionHandler.cpp
```diff
@@ -186,8 +186,37 @@ bool UsdConnectionHandler::deleteConnection(
         return false;
     }
 
-    return UsdShadeConnectableAPI::DisconnectSource(
+    bool retVal = UsdShadeConnectableAPI::DisconnectSource(
         dstUsdAttr->usdAttribute(), srcUsdAttr->usdAttribute());
+
+    // We need to make sure we cleanup on disconnection. Since having an empty connection array
```
ppt-adsk, line 192: If the DisconnectSource return value represents success or failure, is checking connectedAttrs.empty() at line 196 sufficient, or should we check retVal as well?

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
JGamache-autodesk: No idea. I simply replicated the structure of UfeAttributesObserver found directly above this class, including the __del__ found at line 63 of this file.
Either both are correct, or both are wrong. Want me to check?
I do have to agree that relying on the Python garbage collector to remove observers does not look like a good long term strategy.

ppt-adsk: Yeah, that would be a useful exercise.  We can document your findings in the code, wherever is most useful.

lib/mayaUsd/ufe/UsdConnectionHandler.cpp
JGamache-autodesk: That is a very good question. We mostly assume single source connections in our code, but recent discussions on the Pixar mailing list seem to suggest they plan to use multiple connections to support array-like connection (aka the ramp shader problem).
This code tries to be a bit future looking by not assuming we will always be disconnecting a single connection. The code infrastructure and APIs to support multiple connections are already there, and it even seems the Arnold exporter is a bit ahead of the game since it uses multiple connections when exporting ramps and layered textures to Arnold.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
JGamache-autodesk: After a bit of investigation, I would say the current code works very well and it is always a good idea to deregister a listener. Would be nice to be able to deregister as soon as the AE page gets replaced because I could find a very small window where the listener was still active even though the AE was showing another page. But that does not worry me since the garbage collector is very aggressive, the window is small, and the effect is only an unexpected AE refresh.

ppt-adsk (approved):

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
seando-adsk: The reason for this is because the AE actually has multiple tabs - one for each object type. So if you switch between a USD Cone and USD Sphere, the AE is actually hiding one tab and showing the other. All the controls on the hidden tab are left behind. This is because of the "replace" mode where Maya will reuse the existing controls for the next object of the type same.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -83,6 +83,25 @@ def onReplace(self, *args):
         # Nothing needed here since we don't create any UI.
         pass
 
+class UfeConnectionChangedObserver(ufe.Observer):
```
seando-adsk, line 86: Instead of creating a new one why can't you just add your "AttributeConnectionChanged" notif to the existing one and use that?

seando-adsk (approved):

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
JGamache-autodesk: That was actually the initial code I wrote, but the existing one seemed to be customized for transform, this is why I split into a second class that is more customized for shaders.

PR #2548, opened 2022-08-16
Always write color space to usd texture prims

dgovil:
This changes the behaviour of the texture prim writer to **always** write the colorSpace attribute, whereas currently they are elided if they match the default.
This is important for us, and other software vendors, who have software that needs to ingest data from a multitude of pipelines with different defaults. This means we currently have to apply heuristics to figure out the default for a given texture, but it would be most ideal to be able to just always write out the intended color space into the file.

I think this would be a beneficial default for Maya USD in general, because even in a film studio, we might share assets between shows that have different OCIO defaults, so it's beneficial to know that information up front as much as possible.

### DIFF
--- a/lib/usd/translators/shading/usdFileTextureWriter.cpp
+++ b/lib/usd/translators/shading/usdFileTextureWriter.cpp
@@ -51,6 +51,13 @@
 
 PXR_NAMESPACE_OPEN_SCOPE
 
+// Maya USD used to only write the colorspace if the colorspace was not the default.
+// This env var allows users to go back to the sparse write method when desired
+TF_DEFINE_ENV_SETTING(
+    MAYAUSD_EXPORT_VERBOSE_COLORSPACE_METADATA,
+    true,
+    "This env flag controls whether colorspace values are written even if they're the default.")
+
 class PxrUsdTranslators_FileTextureWriter : public UsdMayaShaderWriter
 {
 public:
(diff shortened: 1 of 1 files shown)


JGamache-autodesk (approved):
Goes against having sparse USD files, but is indeed necessary to prevent any confusion at this point in time.

dgovil:
@JGamache-autodesk yeah I agree. The only other options I could think of are:

* stage level color space metadata. It would be out of spec though. 
* use an env var to let people decide to  write it only if it's not the default. 

But this felt like the least complex solution in the interim

dgovil:
Any preference for what it should be called and what the default behaviour should be?

IMHO the default should be "always write colorspace" because it's least surprising, even if it results in more data, with the environment variable toggling it back to its old behaviour. 

That would also help us out when we're reading files that come from multiple locations/users etc to pick the right color settings.

JGamache-autodesk:
Hi @dgovil, after consulting with some of our pipeline/color experts, it looks like USD would benefit from adding `ACEScg` in the list of valid values for the `sourceColorSpace` attribute, or simply not define a set of valid values. That will have to be discussed with Pixar so we can stop using metadata for color space, since this is not transported by Hydra. 

In the meantime, can you please add a `MAYAUSD_EXPORT_VERBOSE_COLORSPACE_METADATA` Tf env var, with `ON` as the default to control the new behavior you implemented. Thanks!

dgovil:
@JGamache-autodesk , added the environment variable to control the behaviour

JGamache-autodesk (approved):
Thanks!

JGamache-autodesk:
Preflight issue is known and on our side. Ready for merge.

Issue #2420, opened 2022-06-14
Configuration fails specifying GULRAK_SOURCE_DIR on Windows


hinton-mgi:
**Describe the issue**
When specifying `GULRAK_SOURCE_DIR` to cmake, configuration fails. As far as I can tell, this is because Windows file paths are not valid URLs (because '\\' is the separator). Modifying gulrak.cmake worked for me:

```
file(TO_CMAKE_PATH ${GULRAK_SOURCE_DIR} GULRAK_SOURCE_URL)
set(GULRAK_SOURCE_URL "file://${GULRAK_SOURCE_URL}")
FetchContent_Declare(
    ${CONTENT_NAME}
    URL ${GULRAK_SOURCE_URL}
)
```

I know next to nothing about cmake, perhaps this doesn't always work and/or there is a better way to convert a path to a URL.

seando-adsk:
How are you specifying the path? You can always use / (even on Windows as this is the cmake default). 
```-DGULRAK_SOURCE_DIR=C:/gulrak```

Sean

**Comment by hinton-mgi** [comment]
The path is specified in a batch file, as "%~dp0\depends\gulrak-filesystem". Obviously I can do some processing on the path to use forward slashes, but it would be much more natural to support this feature on Windows without gotchas.

**Comment by seando-adsk** [comment]
Yes okay I agree with that. But I think this should also work. The FetchContent_Declare URL flag accepts a path and/or URL. So you don't need to add the file:
```
if(DEFINED GULRAK_SOURCE_DIR)
    file(TO_CMAKE_PATH "${GULRAK_SOURCE_DIR}" GULRAK_SOURCE_DIR)
    FetchContent_Declare(
        ${CONTENT_NAME}
        URL ${GULRAK_SOURCE_DIR}
    )
    ...
```

Would you be able to submit a pull-request with this fix? You'll need to fill out and sign the CLA if you haven't already done that.

Sean

hinton-mgi:
Great. This also works on my machine. I will look into getting the CLA signed. No promises.

santosg87:
@hinton-mgi please let us know if you are able to get the CLA signed and submit that bit. we can test it and close when things are all ready . :)

thank you!

hinton-mgi:
Sorry. I have asked my employer to sign the corporate CLA and it's somewhat out of my hands now. Unfortunately it's just not very high priority at the moment. I will post again when it's been handled, but it may take some time.

[...]

PR #2551, opened 2022-08-18
Fix configuration error when GULRAK_SOURCE_DIR is specified on Windows

jamestiotio:
This PR fixes issue #2420.

Perhaps @hinton-mgi can test this fix?

Minor note: I am submitting this PR as an individual. I have signed and sent in my Individual CLA form accordingly.


hinton-mgi:
I just tested with my build setup. I can confirm it works, and I can do a full rebuild without an internet connection. Thanks.

seando-adsk (approved):
Thank you for submitting the fix. I've started our internal preflight system on your change.

FYI, we are finalizing the v0.19.0 release right now so merges are temporarily blocked.

PR #2557, opened 2022-08-22
Add API writers

dj-mcg:
Some Maya nodes may not correspond directly to Usd prims. One example of
this is Renderman-For-Maya's PxrMeshLight, which should be exported as
an applied MeshLightAPI on a Mesh prim.

To support this, this change adds an apiWriterRegistry that allows us to
register apiWriters that get run for each exported prim. We also want to
prevent the export from creating a node for these Maya nodes, so we now
allow a type to be registered as a type that should not result in a Usd
prim.


dj-mcg:
Thanks for carefully thinking through all of this, @seando-adsk - Hold for now, let me follow up with the team to discuss and will follow up.

mrawde:
Hi!  Thanks for the feedback.  I took a look at the schemaApiAdaptor and we can get this working for the most part.  I filed some related issues:

https://github.com/Autodesk/maya-usd/issues/2605
https://github.com/Autodesk/maya-usd/issues/2606
https://github.com/Autodesk/maya-usd/issues/2607

PR #2562, opened 2022-08-23
MAYA-123179 - As a user, I'd like my prims to be in unshaded mode when in a Display Layer

vlasovi:


lib/mayaUsd/render/vp2RenderDelegate/mayaPrimCommon.cpp
```diff
@@ -706,8 +769,6 @@ void MayaUsdRPrim::_SyncSharedData(
         if (!usdVisibility)
             _MakeOtherReprRenderItemsInvisible(reprToken, reprs);
 
-        _SyncDisplayLayerModes(refThis);
```
vlasovi, line 709: This call is not needed here since the function _SyncDisplayLayerModes is already called from _InitReprCommon

PR #2564, opened 2022-08-23
Mark primitives as dirty if the shader has been rebuilt

JGamache-autodesk:
Happens when the shader topology changes, and requires a mesh update to push the latest MShaderInstance.

Thanks to William Krick and Huidong Chen for the code. I only needed to enable it for TopoChanged material updates.


vlasovi (approved):
Looks good. I may need this change for my "Untextured Mode" PR. Although I will need to verify if dirtying just the materialID is enough or I will also need to dirty primvars as well

PR #2565, opened 2022-08-24
MAYA-124820: When adding an Xform to a USD prim two entries are shown…

seando-adsk:
#### MAYA-124820: When adding an Xform to a USD prim two entries are shown in the channel box
* Restore Attribute cache for USD attributes which don't change.

### DIFF
--- a/lib/mayaUsd/ufe/UsdAttributes.cpp
+++ b/lib/mayaUsd/ufe/UsdAttributes.cpp
@@ -120,6 +120,11 @@ static PXR_NS::UsdAttribute _GetAttributeType(const PXR_NS::UsdPrim& prim, const
 
 Ufe::Attribute::Type UsdAttributes::attributeType(const std::string& name)
 {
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -249,6 +259,11 @@ Ufe::Attribute::Ptr UsdAttributes::attribute(const std::string& name)
         newAttr = ctorIt->second(fItem, usdAttr);
 #endif
 
+    // If this is a Usd attribute (cannot change) then we cache it for future access.
+    if (!canRemoveAttribute(fItem, name)) {
+        fUsdAttributes[name] = newAttr;
+    }
```
seando-adsk, line 265: Restored the USD attribute cache, but only for attributes which do not change.

JGamache-autodesk (approved):

ppt-adsk (approved):

ppt-adsk (approved):

PR #2569, opened 2022-08-26
Workaround for USD Issue 2013

JGamache-autodesk:
See https://github.com/PixarAnimationStudios/USD/issues/2013

I do hope to get a better workaround either in the issue comments or on
this USD-interests thread:

https://groups.google.com/g/usd-interest/c/5LZT-IOaOTU/m/XwLzFTq7AgAJ

But, in the meantime, creating a dummy prim under the material is enough to
cause the resync to happen... Feel free to revoke my coding license.
This is ugly.

### DIFF
--- a/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdConnectionHandler.cpp
@@ -102,7 +102,7 @@ _GetShaderNodeDef(const PXR_NS::UsdPrim& prim, const PXR_NS::TfToken& attrName)
     return registry.GetShaderNodeByIdentifier(srcInfoId);
 }
 
-void _SendIssue2013NotifWorkaround(const UsdPrim& usdPrim)
+void _SendStrongConnectionChangeNotification(const UsdPrim& usdPrim)
 {
     // See https://github.com/PixarAnimationStudios/USD/issues/2013 for details.
     //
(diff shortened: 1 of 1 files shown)


JGamache-autodesk:
No hurry on reviewing/merging this. I do still have hope a better solution will be proposed.

lib/mayaUsd/ufe/UsdConnectionHandler.cpp
```diff
@@ -102,6 +102,20 @@ _GetShaderNodeDef(const PXR_NS::UsdPrim& prim, const PXR_NS::TfToken& attrName)
     return registry.GetShaderNodeByIdentifier(srcInfoId);
 }
 
+void _SendIssue2013NotifWorkaround(const UsdPrim& usdPrim)
```
vlasovi, line 105: I would rename this function name into _SendStrongConnectionChangeNotification

vlasovi (dismissed):

hodoulp (approved):
Seems to be an acceptable & safe hack.

seando-adsk:
@JGamache-autodesk Should be this be merged, or do you still hope to have a better solution?

JGamache-autodesk:
Yes, please merge. I will keep an eye on the USD fix and add include guards as soon as we have a version with the fix.

PR #2576, opened 2022-08-31
Hide orphaned pulled nodes on ancestor structure change.

ppt-adsk:


lib/mayaUsd/fileio/primUpdaterManager.h
```diff
@@ -78,6 +80,53 @@ class PrimUpdaterManager : public PXR_NS::TfWeakBase
     bool hasPulledPrims() const { return _hasPulledPrims; }
 
 private:
+    struct VariantSelection
```
ppt-adsk, line 83: To detect when a change in variant should hide or show a pulled hierarchy, we save the complete list of variant sets and their variant selection along the ancestor chain.  If this variant configuration matches, we show the pulled hierarchy, if it doesn't, we hide the pulled hierarchy.  We cannot rely only on the path, because the path may be identical in two different variants, e.g. all variants of a cup might have a top-level prim called "Cup", so that the path to the selected variant is the same.

lib/mayaUsd/fileio/primUpdaterManager.h
```diff
@@ -103,14 +153,64 @@ class PrimUpdaterManager : public PXR_NS::TfWeakBase
     //! Create the pull parent and set it into the prim updater context.
     MDagPath setupPullParent(const Ufe::Path& pulledPath, VtDictionary& args);
 
+    //! Record pull information for the pulled path, for inspection on
+    //! scene changes.
+    void recordPullVariantInfo(const Ufe::Path& pulledPath, const MDagPath& pullParentPath);
```
ppt-adsk, line 158: When we pull, we record variant configuration information in a Ufe::Trie, for fast retrieval later on.

lib/mayaUsd/fileio/primUpdaterManager.h
```diff
@@ -103,14 +153,64 @@ class PrimUpdaterManager : public PXR_NS::TfWeakBase
...
+
+    // Maya file new or open callback.  Member function to access other private
+    // member functions.
+    static void beforeNewOrOpenCallback(void* clientData);
```
ppt-adsk, line 162: Flush the variant configuration information in the Trie on file new or open.

lib/mayaUsd/fileio/primUpdaterManager.h
```diff
@@ -103,14 +153,64 @@ class PrimUpdaterManager : public PXR_NS::TfWeakBase
...
+
+    void beginManagePulledPrims();
```
ppt-adsk, line 164: Start listening to scene structure changes when we pull, to know when to hide or show a pulled hierarchy.

lib/mayaUsd/fileio/primUpdaterManager.h
```diff
@@ -103,14 +153,64 @@ class PrimUpdaterManager : public PXR_NS::TfWeakBase
...
+    void endManagePulledPrims();
+
+    // Member function to access private nested classes.
+    static std::list<VariantSetDescriptor> variantSetDescriptors(const Ufe::Path& path);
+
     friend class TfSingleton<PrimUpdaterManager>;
 
+    class OrphanedNodesManager : public Ufe::Observer
```
ppt-adsk, line 172: This is the object that manages pulled hierarchy visibility.  It observes the scene and looks for scene structure changes, upon which it will hide orphaned nodes, or show them when they are no longer orphaned.

lib/mayaUsd/fileio/primUpdaterManager.h
```diff
@@ -103,14 +153,64 @@ class PrimUpdaterManager : public PXR_NS::TfWeakBase
...
 
     // Becomes true when there is at least one pulled prim.
     // The goal is to let code that can be optimized when there is no pull prim
     // to check rapidly.
     bool _hasPulledPrims { false };
+
+    // Trie for fast lookup of descendant pulled prims.  The Trie key is the
+    // UFE pulled path, and the Trie value is the corresponding Dag pull parent
+    // and all ancestor variant set selections.
+    Ufe::Trie<PullVariantInfo> _pulledPrims {};
```
ppt-adsk, line 205: This is where the variant configuration is stored for pulled hierarchies.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -69,17 +73,6 @@ template <> struct hash<MObjectHandle>
 };
 } // namespace std
 
-namespace MAYAUSD_NS_DEF {
```
ppt-adsk, line 72: Unused, removed.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1140,7 +1133,7 @@ bool PrimUpdaterManager::canEditAsMaya(const Ufe::Path& path) const
 bool PrimUpdaterManager::discardEdits(const MDagPath& dagPath)
 {
     Ufe::Path primPath;
-    if (!PXR_NS::PrimUpdaterManager::readPullInformation(dagPath, primPath))
```
ppt-adsk, line 1143: We're in the PrimUpdaterManager class...

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1595,6 +1592,10 @@ bool PrimUpdaterManager::removePullParent(const MDagPath& parentDagPath)
         return false;
     }
 
+    Ufe::Path pulledPath;
+    TF_VERIFY(readPullInformation(parentDagPath, pulledPath));
+    TF_VERIFY(_pulledPrims.remove(pulledPath) != nullptr);
```
ppt-adsk, line 1597: When we remove a pull parent, we remove the corresponding variant set configuration in the Trie.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1627,6 +1628,7 @@ bool PrimUpdaterManager::removePullParent(const MDagPath& parentDagPath)
                 TF_WARN("Cannot removed pulled prim from the pulled prim cache.");
                 return false;
             }
+            endManagePulledPrims();
```
ppt-adsk, line 1631: When we remove the pull root, empty out the Trie.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1659,6 +1661,8 @@ MDagPath PrimUpdaterManager::setupPullParent(const Ufe::Path& pulledPath, VtDict
     if (MDagPath::getAPathTo(pullParent, pullParentPath) != MStatus::kSuccess) {
         return MDagPath();
     }
+
+    recordPullVariantInfo(pulledPath, pullParentPath);
```
ppt-adsk, line 1665: When we pull, record the variant set configuration in the Trie.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1724,4 +1747,270 @@ bool PrimUpdaterManager::readPullInformation(const MDagPath& dagPath, Ufe::Path&
...
+            if (_pulledPrims.containsDescendant(op.path)) {
+                handleOp(op);
+            }
+        }
+    } else if (_pulledPrims.containsDescendant(changedPath)) {
+        handleOp(sceneNotification);
+    }
+}
+
+void PrimUpdaterManager::OrphanedNodesManager::handleOp(
```
ppt-adsk, line 1836: Handle scene structure changes.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1724,4 +1747,270 @@ bool PrimUpdaterManager::readPullInformation(const MDagPath& dagPath, Ufe::Path&
...
+    auto        changedPath = sceneNotification.changedPath();
+
+    // No changedPath means composite.  Use containsDescendant(), as
+    // containsDescendantInclusive() would mean a structure change on the
+    // pulled node itself, which is not possible (pulled objects are locked).
+    if (changedPath.empty()) {
+        const auto& sceneCompositeNotification
+            = static_cast<const Ufe::SceneCompositeNotification&>(n);
+        for (const auto& op : sceneCompositeNotification.opsList()) {
+            if (_pulledPrims.containsDescendant(op.path)) {
```
ppt-adsk, line 1827: When the scene structure changes, check in the Trie if we have a pulled descendant.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1724,4 +1747,270 @@ bool PrimUpdaterManager::readPullInformation(const MDagPath& dagPath, Ufe::Path&
...
+        // In the latter case, call recursiveHide(), because there is nothing
+        // below the invalidated node.
+
+        auto parentItem = Ufe::Hierarchy::createItem(op.path);
+        if (!TF_VERIFY(parentItem)) {
+            return;
+        }
+
+        auto parentHier = Ufe::Hierarchy::hierarchy(parentItem);
+        if (!parentHier->hasChildren()) {
```
ppt-adsk, line 1884: No children?  We're unloading a payload, so hide.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1724,4 +1747,270 @@ bool PrimUpdaterManager::readPullInformation(const MDagPath& dagPath, Ufe::Path&
...
+            auto ancestorNode = _pulledPrims.node(op.path);
+            if (ancestorNode) {
+                recursiveSetVisibility(ancestorNode, false);
+            }
+            return;
+        } else {
+            // On variant switch, given a pulled prim, the session layer will
```
ppt-adsk, line 1891: Has children?  Variant switch, check if we need to show or hide orphaned hierarchies.

lib/mayaUsd/fileio/primUpdaterManager.h
pierrebai-adsk: I think I'd rather have the function be name on what processing it does than what triggers it, but this is minor.

lib/mayaUsd/fileio/primUpdaterManager.h
pierrebai-adsk, line 83: I think there are/were other stories related to doing things when pulled objects become stale?

What I''m getting at is that we may have to record more data related to pulled objects in the future and maybe these data types and their management would be better off in a separate class that the prim updater manager would hold an instance of instead of having all the types and code directly in it. If we do end-up adding more things to manage, putting it all directly in the prim updater manager will become unwieldy. Only those variants already add quite a bit of data that is not really related to prim updates.

We may wait until we get to these other stories/issues, but it would add a burden of refactoring to these stories.

lib/mayaUsd/fileio/primUpdaterManager.h
pierrebai-adsk: Does this class need to live within the other class? It might even be able to live entirely in an anoymous namespace in the cpp?

pierrebai-adsk: (And if we have to add other thing to do when orphaned, it might even better live off as a separate class in its own header/cpp with a way to regsiter work to be done when a node is orphaned.)

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2021 Autodesk
```
pierrebai-adsk, line 2: Q: is it normal practice to update this? The copyright began in 2021. Just curious.

lib/mayaUsd/fileio/primUpdaterManager.cpp
pierrebai-adsk: IDK if it would be best to remove it at the end when successful or not. OTOH, half-made push will always be problematic.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1667,6 +1671,25 @@ MDagPath PrimUpdaterManager::setupPullParent(const Ufe::Path& pulledPath, VtDict
...
 
+void PrimUpdaterManager::recordPullVariantInfo(
+    const Ufe::Path& pulledPath,
+    const MDagPath&  pullParentPath)
+{
+    // Add the pull parent to our pulled prims prefix tree.  Also add the full
+    // configuration of variant set selections for each ancestor, up to the USD
+    // pseudo-root.  Variants on the pulled path itself are ignored, as once
+    // pulled into Maya they cannot be changed.
+    TF_VERIFY(!_pulledPrims.containsDescendantInclusive(pulledPath));
```
pierrebai-adsk, line 1682: Don't we already prevent editing a child or parent of an already pull prim? I guess rechecking here is harmless, I just want to understand if this protects against somethings I'm not seeing.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1667,6 +1671,25 @@ MDagPath PrimUpdaterManager::setupPullParent(const Ufe::Path& pulledPath, VtDict
...
+    TF_VERIFY(pulledPath.runTimeId() == MayaUsd::ufe::getUsdRunTimeId());
```
pierrebai-adsk, line 1683: Again, curious: how could we possibly be pulling a node that is not a USD node? Just paranoia? :)

lib/mayaUsd/fileio/primUpdaterManager.h
```diff
@@ -103,14 +153,64 @@ class PrimUpdaterManager : public PXR_NS::TfWeakBase
...
+
+    // Maya file new or open callback.  Member function to access other private
+    // member functions.
+    static void beforeNewOrOpenCallback(void* clientData);
+
+    void beginManagePulledPrims();
+    void endManagePulledPrims();
+
+    // Member function to access private nested classes.
+    static std::list<VariantSetDescriptor> variantSetDescriptors(const Ufe::Path& path);
```
pierrebai-adsk, line 168: Minor: having a class named VariantSetDescriptors and a function named variantSetDescriptors is a bit confusing. Maybe a "get"or "compute" prefix on the function?

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1724,4 +1747,270 @@ bool PrimUpdaterManager::readPullInformation(const MDagPath& dagPath, Ufe::Path&
     return false;
 }
 
+/* static */
+std::list<PrimUpdaterManager::VariantSetDescriptor>
+PrimUpdaterManager::variantSetDescriptors(const Ufe::Path& p)
+{
+    std::list<VariantSetDescriptor> vsd;
+    auto                            path = p;
+    while (path.nbSegments() > 1) {
```
pierrebai-adsk, line 1756: Maybe the loop condition could be that the run-time ID is still mayaUsd? Could be more future-proof.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1724,4 +1747,270 @@ bool PrimUpdaterManager::readPullInformation(const MDagPath& dagPath, Ufe::Path&
...
+        auto ancestor = Ufe::Hierarchy::createItem(path);
+        auto usdAncestor = std::static_pointer_cast<MayaUsd::ufe::UsdSceneItem>(ancestor);
+        auto variantSets = usdAncestor->prim().GetVariantSets();
+        std::list<VariantSelection> vs;
+        for (const auto& vsn : variantSets.GetNames()) {
+            vs.emplace_back(vsn, variantSets.GetVariantSelection(vsn));
```
pierrebai-adsk, line 1762: IDK if we could optimize to store nothing if the path has no variant set?

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1724,4 +1747,270 @@ bool PrimUpdaterManager::readPullInformation(const MDagPath& dagPath, Ufe::Path&
...
+        }
+        vsd.emplace_back(path, vs);
+        path = path.pop();
+    }
+    return vsd;
+}
+
+void PrimUpdaterManager::beginManagePulledPrims()
+{
+    TF_VERIFY(_pulledPrims.root()->empty());
```
pierrebai-adsk, line 1772: Isn't beginManagedPulledPrims() called unconditionally and this the observer might already exist if we are editing a second USD prim? I would have expected this function to start with a check that _orphanedNodesManager is null or not and do nothing if it already exists. Would be also more foolproof if for some reason it gets called twice in any case.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1724,4 +1747,270 @@ bool PrimUpdaterManager::readPullInformation(const MDagPath& dagPath, Ufe::Path&
...
+{
+}
+
+void PrimUpdaterManager::OrphanedNodesManager::operator()(const Ufe::Notification& n)
+{
+    const auto& sceneNotification = static_cast<const Ufe::SceneChanged&>(n);
+    auto        changedPath = sceneNotification.changedPath();
+
+    // No changedPath means composite.  Use containsDescendant(), as
+    // containsDescendantInclusive() would mean a structure change on the
```
pierrebai-adsk, line 1821: Is this micro-optimizatin worth it? What if studio build tooling to change variant in-bulk (which could affect a pulled node) for their pipeline? We're only saving a single comparison of the root of a whole prim sub-tree.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1724,4 +1747,270 @@ bool PrimUpdaterManager::readPullInformation(const MDagPath& dagPath, Ufe::Path&
...
+    const Ufe::SceneCompositeNotification::Op& op)
+{
+    switch (op.opType) {
+    case Ufe::SceneChanged::ObjectAdd: {
+        // Restoring a previously-deleted scene item may restore an orphaned
+        // node.  Traverse the trie, and show hidden pull parents that are
+        // descendants of the argument path.  The trie node that corresponds to
+        // the added path is the starting point.  It may be an internal node,
+        // without data.
+        auto ancestorNode = _pulledPrims.node(op.path);
```
pierrebai-adsk, line 1846: I"m not going to try to figure it out now, but a given prim could be hidden both because one of its variant change *and* it got deleted afterwards. So I think restoring visibility (here and for variant switch below in case multiple variants were switched...) need to check that all original causes why it was hidden are now gone.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1724,4 +1747,270 @@ bool PrimUpdaterManager::readPullInformation(const MDagPath& dagPath, Ufe::Path&
...
+        // the argument path.  First, get the trie node that corresponds to
+        // the path.  It may be an internal node, without data.
+        auto ancestorNode = _pulledPrims.node(op.path);
+        TF_VERIFY(ancestorNode);
+        recursiveSetVisibility(ancestorNode, false);
+    } break;
+    case Ufe::SceneChanged::SubtreeInvalidate: {
+        // On subtree invalidate, the scene item itself has not had a structure
+        // change, but its children have changed.  There are two cases:
+        // - the node has children: from a variant switch, or from a payload
```
pierrebai-adsk, line 1872: Ugh, unfortunate that we need to have such intimate knowledge of cause-and-effect of these notifications.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1724,4 +1747,270 @@ bool PrimUpdaterManager::readPullInformation(const MDagPath& dagPath, Ufe::Path&
...
+                // the new hierarchy is completely different from the one when
+                // the pull occurred, which means that the pulled object must
+                // stay hidden.
+                if (ancestorNode) {
+                    foundChild = true;
+                    recursiveSwitch(ancestorNode, cPath);
+                }
+            }
+            if (!foundChild) {
+                // Following a subtree invalidate, if none of the now-valid
```
pierrebai-adsk, line 1925: I wonder if a new variant could have some ancestor in the trie, just by happenstance because another pulled object happens to have the same ancestor? I'll be honest and I'm a bit foggy on what scenarios are possible or not.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1724,4 +1747,270 @@ bool PrimUpdaterManager::readPullInformation(const MDagPath& dagPath, Ufe::Path&
...
+        }
+
+        // If the variant set configuration of the pulled node and the current
+        // tree state don't match, the pulled node must be made invisible.
+        // Inactivation must not be considered, as the USD pulled node is made
+        // inactive on pull, to avoid rendering it.
+        const bool variantSetsMatch
+            = (trieNode->data().variantSetDescriptors == variantSetDescriptors(ufePath.pop()));
+        const bool visibility = (pulledNode && variantSetsMatch);
+        TF_VERIFY(setVisibilityPlug(trieNode, visibility));
```
pierrebai-adsk, line 1993: Again, there maybe multiple causes of invisiblity, so we should check that all have been restored before making a node visible.

pierrebai-adsk (changes requested):
Great work, but I'm afraid making things visible again will require more work still.

lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+    // We store a list of (path, list of (variant set, variant set selection)),
+    // for all ancestors, starting at closest ancestor.
+    auto ancestorPath = pulledPath.pop();
+    auto vsd = variantSetDescriptors(ancestorPath);
+
+    pulledPrims().add(pulledPath, PullVariantInfo(pullParentPath, vsd));
+}
+
+OrphanedNodesManager::Memento OrphanedNodesManager::remove(const Ufe::Path& pulledPath)
```
ppt-adsk, line 89: Now returning the previous Trie of pull info on remove, for undo.

lib/mayaUsd/fileio/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ if(CMAKE_UFE_V3_FEATURES_AVAILABLE)
     target_sources(${PROJECT_NAME}
         PRIVATE
             fallbackPrimUpdater.cpp
+            orphanedNodesManager.cpp
```
ppt-adsk, line 36: Factored out from primUpdaterManager.

lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -110,14 +112,31 @@ void OrphanedNodesManager::operator()(const Ufe::Notification& n)
             }
         }
     } else if (pulledPrims().containsDescendant(changedPath)) {
+#ifdef UFE_V4_FEATURES_AVAILABLE
+        // Use UFE v4 notification to op conversion.
         handleOp(sceneNotification);
+#else
+        // UFE v3: convert to op ourselves.  Only convert supported
```
ppt-adsk, line 119: Maya 2023 UFE v3 did not have nice notification to notification op conversion.

lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -110,14 +112,31 @@ void OrphanedNodesManager::operator()(const Ufe::Notification& n)
...
+        }
+#endif
     }
 }
 
 void OrphanedNodesManager::handleOp(const Ufe::SceneCompositeNotification::Op& op)
 {
     switch (op.opType) {
-    case Ufe::SceneChanged::ObjectAdd: {
+    case Ufe::SceneCompositeNotification::ObjectAdd: {
```
ppt-adsk, line 139: In UFE v3 the enumerant was in the SceneCompositeNotification derived class, in UFE v4 it's move to the base class, but the derived class scope still works, so using the derived class is UFE v3/v4 compatible, a requirement for 2023 compatibility.

lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -217,35 +236,28 @@ void OrphanedNodesManager::handleOp(const Ufe::SceneCompositeNotification::Op& o
             }
         }
     } break;
+    default: {
```
ppt-adsk, line 239: UNIX (Linux, macOS) warning as error fix.

ppt-adsk:
Still not 2023 compatible: missing crucial Ufe::TrieNode::childrenComponents() method, which will have to be added to UFE v3 for 2023 Update 3 (and it can be, as it's a non-virtual).

pierrebai-adsk (dismissed):
Great work!

pierrebai-adsk (approved):
I see you added detection for the fature instead of UFE version for the backport to Maya 2023 (or was it 2022?).

PR #2578, opened 2022-09-01
LOOKDEVX-571 | Initial implementation of AETemplate control creators.

feldstj:


lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -402,6 +402,26 @@ def __OnPrimsChanged(self, notice, sender):
                 if hasattr(ctrl, 'refresh'):
                     ctrl.refresh()
 
+def connectionsCustomControlCreator(self, c):
```
ppt-adsk, line 405: By Python convention, self should be reserved for the object argument on class methods.  Should call it aeTemplate, or similar.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -402,6 +402,26 @@ def __OnPrimsChanged(self, notice, sender):
...
+    if self.attributeHasConnections(c):
+        connectionsCustomControl = ConnectionsCustomControl(self.item, self.prim, c, self.useNiceName)
+        self.defineCustom(connectionsCustomControl, c)
```
ppt-adsk, line 408: Shouldn't this be done by AETemplate.addControls() code?  Otherwise, it will be repeated in every custom control creator.  This means you will have to remove defaultControlCreator() as the end of the chain of responsibility, since it doesn't call defineCustom(), but I think that's a good tradeoff.

ppt-adsk (commented):
Looks pretty good to me!  Just the defineCustom() call to move back to AETemplate, I think.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
feldstj: I wasn't sure if this was definitely something every control needed to do. If you're saying it is, then your suggestion sounds reasonable.

feldstj: Oh, and I didn't need to remove the defaultControlCreator() because it returns None, I'm only calling defineCustom(...) if the returned result is not None.

ppt-adsk (approved):
Nice work!

seando-adsk (dismissed):
Looks good thanks.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -447,18 +463,20 @@ def __init__(self, ufeSceneItem):
             except:
                 pass
 
+    _controlCreators[connectionsCustomControlCreator, arrayCustomControlCreator, defaultControlCreator]
```
ppt-adsk, line 466:
```suggestion
    _controlCreators = [connectionsCustomControlCreator, arrayCustomControlCreator, defaultControlCreator]
```

ppt-adsk (approved):

seando-adsk (approved):

PR #2581, opened 2022-09-06
Change how the active 3D view is retrieved in the test

pierrebai-adsk:
Otherwise, the test failed when run locally. The new method to retrieve the 3D view is more robust.

### DIFF
--- a/test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeLiveSurface.py
+++ b/test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeLiveSurface.py
@@ -69,8 +69,7 @@ class testProxyShapeLiveSurface(unittest.TestCase):
         cmds.refresh()
 
         # Get the viewport widget.
-        self._view = OMUI.M3dView()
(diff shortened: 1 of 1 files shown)


test/lib/mayaUsd/render/pxrUsdMayaGL/testProxyShapeLiveSurface.py
```diff
@@ -69,8 +69,7 @@ def setUp(self):
         cmds.refresh()
 
         # Get the viewport widget.
-        self._view = OMUI.M3dView()
-        OMUI.M3dView.getM3dViewFromModelPanel(self._panel, self._view)
+        self._view = OMUI.M3dView().active3dView()
```
seando-adsk, line 72: What does this change? The code seems to be setting the panel to a persp view and then getting the viewport from it. Now you are using the active viewport. Is that still the one with persp camera?

pierrebai-adsk: Yes, there is only on viewport open at that point. The old code would not return the correct panel when I ran it locally.

seando-adsk (approved):

PR #2583, opened 2022-09-06
MAYA-124106 preserve timeline during edit-as-Maya

pierrebai-adsk:
- Preserve the timeline start/end and min/max after editing a Maya reference.
- Add a unit test that checks the timeline preservation.
- Corrected two tests that had their named swapped.


JGamache-autodesk (approved):

JGamache-autodesk:
Although having it as an option feels strange. I would have expected all file I/O operations to preserve the timeline.
Should it also restore the current frame? Exporting animation will do a playback and end on the last frame, which might be outside of the preserved range.

pierrebai-adsk:
Existing import tests were explicitly checking that the timeline would be set to the animation start/stop values, so I had to add this option to preserve the timeline when doing edit-as-Maya and not preserve it when doing normal imports.

PR #2588, opened 2022-09-09
LOOKDEVX-571 | Adds support for converting strings with options into

feldstj:
enum strings.


test/lib/ufe/testAttribute.py
```diff
@@ -419,6 +420,43 @@ def testAttributeEnumString(self):
         # Run test using Maya's getAttr command.
         self.runMayaGetAttrTest(ufeAttr)
 
+    @unittest.skipIf(os.getenv('USD_HAS_MX_METADATA_SUPPORT', 'NOT-FOUND') not in ('1', "TRUE"), 'Test only available if USD can read MaterialX metadata')
+    @unittest.skipIf(os.getenv('UFE_PREVIEW_VERSION_NUM', '0000') < '4001', 'nodeDefHandler is only available in UFE preview version 0.4.1 and greater')
```
seando-adsk, line 424: @ppt-adsk Does using two conditions like this work? I've never seen this before.

JGamache-autodesk: Decorators can be chained like this. Was used for `testCreateUsdPreviewSurfaceAttribute` in `test\lib\ufe\testAttribute.py`.

JGamache-autodesk (approved):

ppt-adsk (approved):

PR #2598, opened 2022-09-14
Broken AL_USDMaya and AL_USDTransaction builds (AL)

fabal:
Sorry I didn't catch it earlier but one of the changes in https://github.com/Autodesk/maya-usd/pull/2456 , i.e., linking the boost python's library to the test executables, breaks our builds.
As I'm not across the reason for the removal in the first place, I reintroduce it here so we can discuss the best approach to get builds up and running.

### DIFF
--- a/plugin/al/schemas/AL/usd/schemas/maya/tests/CMakeLists.txt
+++ b/plugin/al/schemas/AL/usd/schemas/maya/tests/CMakeLists.txt
@@ -29,6 +29,7 @@ target_link_libraries(${TARGET_NAME}
     tf
     vt
     plug
+    ${Boost_PYTHON_LIBRARY}
 )
 
 # install
--- a/plugin/al/schemas/AL/usd/schemas/maya/tests/CMakeLists.txt
+++ b/plugin/al/schemas/AL/usd/schemas/maya/tests/CMakeLists.txt
@@ -29,7 +29,7 @@ target_link_libraries(${TARGET_NAME}
     tf
     vt
     plug
-    ${Boost_PYTHON_LIBRARY}
+    $<$<BOOL:${IS_LINUX}>:${Boost_PYTHON_LIBRARY}>
 )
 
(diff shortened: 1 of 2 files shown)


seando-adsk (approved):
I'm okay with putting this back as long it is doesn't start breaking the test again. We had test failures during our nightly test runs. Yves made some changes to fix that.

seando-adsk:
@boudrey  What do you think of this change?

boudrey:
I did a change to exactly remove those to fix an issue on a new OSX build machine. Commit [1659c82](https://git.autodesk.com/maya3d/maya-usd/commit/1659c826a964ba42dd50723bac00a1286e34a47a). So it will probably fail again if reverted. Could be related to the use of different CMake versions.

seando-adsk:
@boudrey The preflight was run and passed on this change. The MayaUsd builds for master were done on the new OSX machines. Was the test one of those random failures, meaning it might fail in the future if we merge this change?

fabal:
Thanks for looking into this. I can add a linux based conditional to the change if that helps.

nxkb:
Hi @fabal, what was the symptom that you saw?  I ran into a similar issue building today as well (we also build on linux).
This is basically what I was seeing: https://stackoverflow.com/questions/34143265/undefined-reference-to-symbol-pthread-createglibc-2-2-5?
I was able to solve it by using the `-pthread` solution... which would also be in a linux only block so the two solutions may be roughly equivalent at the end of the day. I can push my fix branch if its useful.

fabal:
Hey @nxkb , in our case it is a very straightforward build error, reporting missing boost python symbols. As far as I can tell, no issue on our end with `pthread` though.

seando-adsk:
@fabal A linux conditional would be great as the problem we saw on our side was OSX only.

seando-adsk (approved):
Great thanks.

PR #2599, opened 2022-09-14
MAYA-123181 - As a user, I'd like my prims to be in untextured mode when in a Display Layer

vlasovi:
- Introducing support for two simultaneous compiled material networks in HdVP2Material: Full and Untextured. Before, only one one network might be compiled at a time.
- Fixing mixed multi-viewport styles, which were broken lately, all the while keeping the optimal performance.
- Adding a separate representation for Untextured draw style with support from meshes, curves and points
- Supporting display layer's "texturing" attribute

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/material.cpp
@@ -1845,7 +1845,7 @@ void HdVP2Material::Sync(
 
             // full network is synced only if required by display style 
             auto* const param = static_cast<HdVP2RenderParam*>(_renderDelegate->GetRenderParam());
-            if (param->GetDrawScene().GetDisplayStyle() & MHWRender::MFrameContext::kTextured) {
+            if (param->GetDrawScene().NeedTexturedMaterials()) {
                 _compiledNetworks[kFull].Sync(sceneDelegate, fullNetworkMap);
             }
         } else {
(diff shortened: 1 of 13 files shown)


vlasovi:
@JGamache-autodesk This is a big PR with a lot of changes. It might be easier for you to go through individual commits

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1835,202 +1836,204 @@ void HdVP2Material::Sync(
...
-                : vtMatResource.UncheckedGet<HdMaterialNetworkMap>();
+            // untextured network is always synced
+            HdMaterialNetworkMap untexturedNetworkMap = fullNetworkMap;
+            ConvertNetworkMapToUntextured(untexturedNetworkMap);
+            _compiledNetworks[kUntextured].Sync(sceneDelegate, untexturedNetworkMap);
 
-            if (inUntexturedMode) {
-                untexturedNetworkMap = vtMatResource.UncheckedGet<HdMaterialNetworkMap>();
-                ConvertNetworkMapToUntextured(untexturedNetworkMap);
+            // full network is synced only if required by display style
```
JGamache-autodesk, line 1847: Unclear why the untextured network is always synced. Especially when we want the full network for the textured Repr.

JGamache-autodesk (approved):
Looks good to me. Telling Github to ignore whitespace really helped with the material changes.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
vlasovi: The default viewport style in Maya is untextured, so I thought of untextured material network as default as well, especially since it's lightweight compared to the other one. However, you are right and technically we shouldn't be forced to always sync it.

PR #2600, opened 2022-09-14
MAYA-124971 Do not change suffix if renamed to the same name

pierrebai-adsk:


lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -69,12 +69,16 @@ UsdUndoRenameCommand::UsdUndoRenameCommand(
...
 
     ufe::applyCommandRestriction(prim, "rename");
 
     // handle unique name for _newName
-    _newName = uniqueChildName(prim.GetParent(), TfMakeValidIdentifier(newName.string()));
+    const std::string validNewName = TfMakeValidIdentifier(newName.string());
+    if (validNewName != prim.GetName())
+        _newName = uniqueChildName(prim.GetParent(), validNewName);
+    else
+        _unchanged = true;
```
ppt-adsk, line 81: Minor: instead of adding a flag, we could rely on _newName being empty.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -89,6 +93,13 @@ UsdSceneItem::Ptr UsdUndoRenameCommand::renamedItem() const { return _ufeDstItem
 
 bool UsdUndoRenameCommand::renameRedo()
 {
+    // If the new name is the same as the current name, do nothing.
+    // This is the same behavior as the Maya rename command for Maya nodes.
+    if (_unchanged) {
+        _ufeDstItem = _ufeSrcItem;
```
ppt-adsk, line 99: Why can't we make this assignment in the constructor?  It never changes, right?  Also, why would we only be doing it in redo()?

pierrebai-adsk: Ah, yes, it could be done in the constructor. I was mimicking the behavior of the other cases that only set it in  redo and 
undo.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
pierrebai-adsk: True, but it would make the code a little more obscure, but not by much. Since i'll move the dst assignment in the constructor, I'll also get rid of the flag.

ppt-adsk (approved):

PR #2601, opened 2022-09-14
Update exporters for string varnames

JGamache-autodesk:
The varname input of all UsdPrimVarReader nodes went from a Token to a String in USD 20.11. This updates the exporter to follow the convention currently in use, and allows re-using varnames with MaterialX.

We also do not hardcode the output name of MaterialX nodes. We query the Sdr registry for the correct information.

There are still too many hardcoded names and types in the exporter. To be fixed later.


seando-adsk (changes requested):
I'm not familiar with this mtlx code. I just one minor comment about variable names.

seando-adsk (approved):

PR #2604, opened 2022-09-16
MAYA-124973 handle trailing # in rename

pierrebai-adsk:
- Transform the trailing # into a unique number.
- Add a unit test for it.


ppt-adsk (approved):
Nice and tidy, and thanks for the test!

seando-adsk (approved):

PR #2612, opened 2022-09-19
MAYA-122927 prevent crash when discarding deactivated maya ref.

pierrebai-adsk:


seando-adsk (approved):
The change looks good to me.

ppt-adsk (commented):
Looks good!  Can you add a test?

seando-adsk (approved):

test/lib/mayaUsd/fileio/testAddMayaReference.py
```diff
@@ -206,6 +208,68 @@ def testEditAndDiscardMayaRef(self):
...
+        #
+        # Note: we have to retrieved the Maya ref prim again because when its parent
+        #       was deactivated, the UsdPrim object became invalid and cannot be used
+        #       anymore.
+        mayaRefUsdItem = ufeUtils.createItem(mayaRefPrimPathStr)
+        mayaRefPrim = usdUtils.getPrimFromSceneItem(mayaRefUsdItem)
+        attr = mayaRefPrim.GetAttribute('mayaAutoEdit')
+        self.assertTrue(attr.IsValid())
+        self.assertEqual(attr.Get(), True)
+        self.assertFalse(mayaRefPrim.IsActive())
```
ppt-adsk, line 270: Why is the mayaRefPrim inactive?

test/lib/mayaUsd/fileio/testDiscardEdits.py
```diff
@@ -118,6 +119,72 @@ def testDiscardEdits(self):
...
+        # Recreate the UFE item since the USD Prim was temporarily deactive.
+        aUsdItem = ufeUtils.createItem(aUsdUfePath)
+        aUsdPrim = usdUtils.getPrimFromSceneItem(aUsdItem)
+        aUsdPrim.SetActive(True)
+
+        # The B item is still inactive.
+        # Recreate the UFE item since the USD Prim was temporarily deactive
+        bUsdItem = ufeUtils.createItem(bUsdUfePath)
+        bUsdPrim = usdUtils.getPrimFromSceneItem(bUsdItem)
+        self.assertFalse(bUsdPrim.IsActive())
```
ppt-adsk, line 169: Same question as test above, why is bUsdPrim inactive.

test/lib/mayaUsd/fileio/testDiscardEdits.py
```diff
@@ -118,6 +119,72 @@ def testDiscardEdits(self):
...
+
+        # Activate the B item. Need to target the session layer since edit-as-Maya
+        # had deactivated the prim in the session layer.
```
ppt-adsk, line 172: Hmm, feels like a bug in discard edits: upon discard edits, prim should no longer be inactive, right?  If that's the case, it's out of scope for your pull request.

test/lib/mayaUsd/fileio/testAddMayaReference.py
pierrebai-adsk: Because an ancestor (here, the parent) is inactive and prevent all editing of the child prim. So we cannot reset the inactive flag during discard edit, so it stays inactive. Note that this is the original bug reported as part of MAYA-122927, but we decided to push solving it back since we have higher priorities.

ppt-adsk (approved):

PR #2613, opened 2022-09-19
Use Adapter to author MeshLightAPI for RfM's PxrMeshLight

dj-mcg:
Also apply gamma correction to color values on RfM lights


dj-mcg:
Shoot it looks like the meshLightAPI is just slightly too new for this to go in without version guards.

dj-mcg:
Hi @seando-adsk - this is an attempt to use the existing adapters to author out a MeshLightAPI on export. Note that we didn't add import steps as they were not needed at Pixar, and unfortunately we are a bit overloaded on production-facing tasks. Despite this I'd love to integrate this into the codebase since the code spans several files and will likely cause merge conflicts going forward if we keep patching it on top.

Thanks!
Dan

JGamache-autodesk (approved):
Code looks good. Thanks!

PR #2615, opened 2022-09-20
MAYA-124951: Duplicate as USD Data timeline options don't show what is currently applied

seando-adsk:
MAYA-124951: Duplicate as USD Data timeline options don't show what is currently applied

* Do not auto-update the animation start/end to the playback range as it doesn't actually reflect the values used by the options other than export (edit as maya, merge to maya).

### DIFF
--- a/plugin/adsk/scripts/mayaUsdTranslatorExport.mel
+++ b/plugin/adsk/scripts/mayaUsdTranslatorExport.mel
@@ -796,9 +796,9 @@ global proc int mayaUsdTranslatorExport (string $parent,
         mayaUsdTranslatorExport_EnableAllControls();
         mayaUsdTranslatorExport_SetFromOptions($currentOptions, 1, 1);
 
-        // Set visibility for anim widgets
+        // Set visibility for anim widgets (but do not update the start/end time to playback range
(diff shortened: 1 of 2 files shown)


plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -796,9 +796,9 @@ global proc int mayaUsdTranslatorExport (string $parent,
         mayaUsdTranslatorExport_EnableAllControls();
         mayaUsdTranslatorExport_SetFromOptions($currentOptions, 1, 1);
 
-        // Set visibility for anim widgets
+        // Set visibility for anim widgets (but do not update the start/end time to playback range
+        // as they will be set by the caller).
         mayaUsdTranslatorExport_AnimationCB();
-        mayaUsdTranslatorExport_AnimationRangeCB();
```
seando-adsk, line 801: As per discussion with design, they chose to remove this auto-update of the animation range (to the playback range) as it does not reflect the true values used by the operations other than export (such as Duplicate as Usd).

pierrebai-adsk (approved):

PR #2617, opened 2022-09-21
Show correct render backend name

dgovil:
This is a small PR to show the correct render name for MtoH in the UI. This doesn't help mtoh work better on mac yet, but it does help my OCD slightly.

This API exists on USD 20.02 and higher. Hopefully that's fine for the current USD versioning being used for Maya USD

image


seando-adsk (approved):
Change looks good. The min USD version we support is 21.02.

seando-adsk:
@dgovil There is a clang format problem that needs to be fixed.

dgovil:
@seando-adsk ah apologies, that should be fixed now

seando-adsk (approved):

PR #2618, opened 2022-09-22
MAYA-123182 - As a user, I'd like my prims' wireframes to have a specific color and opacity when in a Display Layer

vlasovi:


seando-adsk (approved):
I'm not very familiar with this area of the code, but the changes look good to me.

PR #2622, opened 2022-09-23
Degiroa/lookdevx 125331/handle automatic creation of attribute unique name

alicedegirolamo:
# Description

When creating a new attribute, it is automatically checked it has a unique name, if it is not, a new and unique name is automatically provided.
The name is unique between all the existing item attributes.
Examples:
- new attribute name` outputs:out` -> but already exists an attribute with that name. So new name will be `outputs:out1`
- new attribute name `outputs:out` -> but already exists an attribute with that name AND an attribute with name `outputs:out1`. So new name will be `outputs:out2`

## Type of change

- [ ] New feature (non-breaking change which adds functionality) 

# Checklist:

- [ ] My code follows the style guidelines of [this ](https://github.com/Autodesk/maya-usd/blob/dev/doc/codingGuidelines.md)project


lib/mayaUsd/ufe/UsdUndoAttributesCommands.cpp
```diff
@@ -40,8 +40,9 @@ UsdAddAttributeCommand::Ptr UsdAddAttributeCommand::create(
     const std::string&          name,
     const Ufe::Attribute::Type& type)
 {
-    if (UsdAttributes::canAddAttribute(sceneItem, name, type)) {
-        return std::make_shared<UsdAddAttributeCommand>(sceneItem, name, type);
+    const std::string kUniqueName = UsdAttributes::getUniqueAttrName(sceneItem, name);
+    if (UsdAttributes::canAddAttribute(sceneItem, kUniqueName, type)) {
+        return std::make_shared<UsdAddAttributeCommand>(sceneItem, kUniqueName, type);
```
JGamache-autodesk, line 45: That covers only the command, and not the direct call API. Can you make the name unique in `UsdAttributes::doAddAttribute()` instead?

JGamache-autodesk (changes requested):

lib/mayaUsd/ufe/UsdUndoAttributesCommands.cpp
alicedegirolamo: You are right, at first, I tried making the unique name directly in` UsdAttributes::doAddAttribute()`. But I preferred to create the `UsdAddAttributeCommand` with the already unique name, in fact from the constructor the name is saved in the field `_name`. When the command has finished, to get the created attribute, we call `UsdAddAttributeCommand::attribute()` which uses `_name`. In `UsdAttributes::doAddAttribute` I cannot update directly the command `_name`. But I can make the unique name in `UsdAttributes::doAddAttribute` and outside that function add the update for the command field `_name `. I thought that creating the command immediately with the unique `_name` would have been less error-prone for a possible mismatch between what it is created in `doAddAttribute` and what we get from `UsdAddAttributeCommand::attribute()` if we forget in the future to update `_name` after calling `doAddAttribute `.

JGamache-autodesk: Not very fond of that workflow split between the command and the direct call. Both should provide the same services.

Can you add a `_uniqueName` non-const member to the command that gets set from the uniquely named Attribute returned by `doAddAttribute()`? Then use that in `UsdAddAttributeCommand::attribute()`.

You could also make `_name` non-const since we do not get much value from keeping it intact once the command has been executed.

alicedegirolamo: Got it. If it is not a problem I directly updated `_name` which now is non-const. I thought that continuing to have one member for the name would be more clear.

JGamache-autodesk (approved):

hodoulp:
@seando-adsk can someone from maya-usd team review and merge if ready?

seando-adsk:
@ppt-adsk Can you have a look at the changes. Once that is done we'll need to re-run the preflight as new commits were added after the last run.

lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -332,21 +332,40 @@ bool UsdAttributes::canAddAttribute(
 {
     // See if we can edit this attribute, and that it is not already part of the schema or node
     // definition
-    if (!item || !item->prim().IsActive() || UsdAttributes(item).hasAttribute(name)) {
```
ppt-adsk, line 335: This feels very tricky to me.  The previous meaning of the function was: "Can I add an attribute with this exact name?"

And previously, the answer was: if an attribute of that name already exists, no you can't.  You've changed that answer to yes you can, but it won't be an attribute of that name.  In fact, the name argument is completely unused now.

Either we change the interface and the documentation of the function, or we revert this change and create a separate function that ignores the name.

lib/mayaUsd/ufe/UsdUndoAttributesCommands.cpp
```diff
@@ -53,12 +53,19 @@ Ufe::Attribute::Ptr UsdAddAttributeCommand::attribute() const
...
 {
     // Validation has already been done. Just create the attribute.
     auto sceneItem
         = std::dynamic_pointer_cast<UsdSceneItem>(Ufe::Hierarchy::createItem(_sceneItemPath));
-    UsdAttributes::doAddAttribute(sceneItem, _name, _type);
+    auto attrPtr = UsdAttributes::doAddAttribute(sceneItem, _name, _type);
+
+    // Set the name, since it could have been changed in order to be unique.
+    if (attrPtr) {
+        setName(attrPtr->name());
```
ppt-adsk, line 67: Why didn't doAddAttribute() set the name on the return attrPtr?  That way, you wouldn't need the setName(), and you could leave the _name const.

lib/mayaUsd/ufe/UsdUndoAttributesCommands.h
```diff
@@ -60,12 +60,13 @@ class UsdAddAttributeCommand : public UsdUndoableCommand<Ufe::AddAttributeComman
         const Ufe::Attribute::Type& type);
 
     Ufe::Attribute::Ptr attribute() const override;
+    void                setName(const std::string& newName);
```
ppt-adsk, line 63: I'm hopeful you can remove this method.

ppt-adsk (changes requested):
Good work, I think we can simplify things a bit, and this definitely needs a test.

lib/mayaUsd/ufe/UsdAttributes.cpp
alicedegirolamo: I think it's better to change the interface and update the documentation since this function is only used once by the `UsdAddAttributeCommand::create` function.

lib/mayaUsd/ufe/UsdUndoAttributesCommands.cpp
alicedegirolamo: Sorry but maybe I didn't understand the comment. 
`doAddAttribute()` actually sets the new name for `attrPtr`. 
The `setName` function sets the name for the `UsdAddAttributeCommand`.

ppt-adsk: Right you are, sorry for the confusion.  Still a bit uncomfortable as the rename command determines the name for a prim in its constructor
https://github.com/Autodesk/maya-usd/blob/158c1d3ea5efa671bbbd6d1a217b60ba02519823/lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp#L92
not at execute time, whereas you are doing the reverse.  There are arguments for both, so I think what you have is fine.

ppt-adsk (approved):
Thanks for the test!

lib/mayaUsd/ufe/UsdUndoAttributesCommands.h
JGamache-autodesk: Or make it private. Feels strange to give programmers the option of changing something that is managed internally by the command.

JGamache-autodesk (approved):

lib/mayaUsd/ufe/UsdUndoAttributesCommands.cpp
alicedegirolamo: That's a really interesting class, I didn't know it, thanks for the info.

lib/mayaUsd/ufe/UsdUndoAttributesCommands.h
alicedegirolamo: Yes, you are right, not putting it as private is a mistake in this case. I correct it.

JGamache-autodesk (approved):

ppt-adsk (approved):

PR #2625, opened 2022-09-26
MAYA-124365 test prim undo delete vs display layer

pierrebai-adsk:


test/lib/ufe/testDisplayLayer.py
```diff
@@ -304,8 +304,55 @@ def testDisplayLayerItemDelete(self):
...
+
+        self._testLayerFromPath(self.CUBE1, self.LAYER1)
+        self._testLayerFromPath(self.SPHERE1, self.LAYER1)
+
+        # Delete the Sphere and make sure it is removed from the layer.
+        # To allow undo to work, we need to have an undo block open when
+        # executing the command. In Maya, using the UI, commands always
+        # have an undo chunk. In scripting, we must create an explicit one.
+        cmds.undoInfo(state=1)
+        cmds.undoInfo(openChunk=1)
```
pierrebai-adsk, line 342: When issuing commands in scripts, we need to create undo check by hand if we want them... display layer restoration only work when we have a undo chunk.

test/lib/ufe/testDisplayLayer.py
```diff
@@ -304,8 +304,51 @@ def testDisplayLayerItemDelete(self):
...
+        # Delete the Sphere and make sure it is removed from the layer.
+        cmds.delete(self.SPHERE1)
+
+        verifyInLayer({ self.CUBE1 }, { self.NEW_SPHERE1, self.SPHERE1 })
+
+        # Must issue two undo because delete spawed another command
+        # That updated the display layer. The alternative would be to
+        # open an explicit undo chunk to combine both commands into one undo.
+        cmds.undo()
+        cmds.undo()
```
ppt-adsk, line 346: This feels very awkward.  I know you've mentioned that in the interactive application all commands are chunked, so to test what the interactive application is doing I would follow your suggestion and open an explicit undo chunk before line 338.

ppt-adsk (changes requested):

test/lib/ufe/testDisplayLayer.py
```diff
@@ -304,8 +304,51 @@ def testDisplayLayerItemDelete(self):
             self.assertFalse(defaultLayer.contains(self.SPHERE1))
         self.assertFalse(layer1.contains(self.SPHERE1))
 
-        # TODO - test the undo/redo for delete of Ufe item in Layer.
-        # This currently doesn't work.
+    @unittest.skipUnless(mayaUtils.ufeSupportFixLevel() >= 6, "Requires Display Layer Ufe item delete undo fix.")
```
ppt-adsk, line 307: Is the fix level still O.K.?

pierrebai-adsk: Yes, it is still 6 for that fix.

pierrebai-adsk:
Another PF where the interactive Linux builds failed to start. The interactive Linux instances must be down. Unrelated to my changes, all other builds passed.

ppt-adsk (approved):

PR #2626, opened 2022-09-26
MAYA-125364 - MayaUsd: Maya master blessing failing (Linux only)

seando-adsk:
#### MAYA-125364 - MayaUsd: Maya master blessing failing (Linux only)
* Minor changes to settings used when compiling googletest required for building with gcc 11.

### DIFF
--- a/cmake/googletest.cmake
+++ b/cmake/googletest.cmake
@@ -25,6 +25,18 @@ macro(fetch_googletest)
         # sequence parsing errors.  PPT, 22-Nov-2018.
         file(TO_CMAKE_PATH ${CMAKE_MAKE_PROGRAM} CMAKE_MAKE_PROGRAM)
 
+        # Set some options used when compiling googletest.
+        set(CMAKE_CXX_STANDARD 11)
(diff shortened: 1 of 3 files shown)


cmake/googletest.cmake
```diff
@@ -25,6 +25,14 @@ macro(fetch_googletest)
         # sequence parsing errors.  PPT, 22-Nov-2018.
         file(TO_CMAKE_PATH ${CMAKE_MAKE_PROGRAM} CMAKE_MAKE_PROGRAM)
 
+        # Set some options used when compiling googletest.
+        set(CMAKE_CXX_STANDARD 11)
+        set(CMAKE_CXX_EXTENSIONS OFF)
+        set(CMAKE_CXX_STANDARD_REQUIRED ON)
+        if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+            set(disable_all_warnings_flag -w)
```
ppt-adsk, line 33: Would be good to add a comment to explain why we're doing this.

PR #2628, opened 2022-09-27
MAYA-123720 - As a user, I'd like Display Layer membership to survive the 'Edit As Maya Data' workflow

vlasovi:


lib/mayaUsd/ufe/Utils.h
```diff
@@ -231,5 +231,25 @@ Ufe::Selection recreateDescendants(const Ufe::Selection& src, const Ufe::Path& f
...
 std::vector<std::string> splitString(const std::string& str, const std::string& separators);
 
+class PullExtras
+{
+public:
+    void initRecursive(Ufe::SceneItem::Ptr);
+    void processItem(const Ufe::Path& path, const MObject& mayaObject);
+
+private:
+    std::unordered_map<Ufe::Path, MObject> _displayLayerMap;
```
seando-adsk, line 241: Why not store the MFnDisplayLayer so you don't need the extra check/object create in the processItem()?

lib/mayaUsd/ufe/Utils.h
```diff
@@ -231,5 +231,25 @@ Ufe::Selection recreateDescendants(const Ufe::Selection& src, const Ufe::Path& f
 MAYAUSD_CORE_PUBLIC
 std::vector<std::string> splitString(const std::string& str, const std::string& separators);
 
+class PullExtras
```
seando-adsk, line 234: Do we need two different classes? How about a single class that can handle both cases (Maya->USD and USD->Maya). Just have two sections of methods for each case. And I guess you were thinking that in the future we could add other types of extras (now we have DisplayLayers)? The names "Pull/Push" are specific to your operation, but we'll add this for duplicate usd prim. Can we have a generic name?

lib/mayaUsd/ufe/Utils.h
```diff
@@ -231,5 +231,25 @@ Ufe::Selection recreateDescendants(const Ufe::Selection& src, const Ufe::Path& f
...
+{
+public:
+    void initRecursive(Ufe::SceneItem::Ptr);
+    void processItem(const Ufe::Path& path, const MObject& mayaObject);
```
seando-adsk, line 238: These methods need comments as to what they are doing.

lib/mayaUsd/ufe/Utils.h
vlasovi: @seando-adsk I cannot put MFnDisplayLayer in a container because its operator= as well as its copy constructor are declared private and not accessible. Putting a pointer to MFnDisplayLayer is basically the same as using MObject

lib/mayaUsd/ufe/Utils.h
vlasovi: @seando-adsk The two classes share neither data members nor methods, so I don't see a reason to merge them. As for renaming, would ReplicateExtrasToUSD/ReplicateExtrasFromUSD wouldwork for you?

seando-adsk: Yes those names are better.

seando-adsk (changes requested):
As a general comment you need some method documentation and unit tests.

seando-adsk (dismissed):
Looks good thanks.

seando-adsk:
I'll wait until @ppt-adsk has had a chance to review this before merging.

lib/mayaUsd/fileio/primUpdaterContext.h
```diff
@@ -67,6 +68,9 @@ class UsdMayaPrimUpdaterContext
     MAYAUSD_CORE_PUBLIC
     MDagPath MapSdfPathToDagPath(const SdfPath& sdfPath) const;
 
+    mutable MayaUsd::ufe::ReplicateExtrasFromUSD _pullExtras;
```
ppt-adsk, line 71: These mutable data members feel uncomfortable.  Does the following make any sense:
- initRecursive() writes into the map.  Can we transform it into a static creator method?
- processItem() and finalize() read from the map.  They can be made const methods.
That way, you can pass in the ReplicateExtras objects as const into the context, make them const data members, and provide const accessors to them, keeping the PrimUpdaterContext read-only.

Hope I didn't skip over some crucial detail...

ppt-adsk (commented):
Great work, see my suggestion and let me know if it makes sense.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -895,5 +900,96 @@ std::vector<std::string> splitString(const std::string& str, const std::string&
...
+    if (!node) {
+        return;
+    }
+
+    // Go through the entire hierarchy
+    for (auto child : node->children()) {
+        initRecursive(child);
+    }
+
+#ifdef MAYA_HAS_DISPLAY_LAYER_API
```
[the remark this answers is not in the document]
vlasovi: Yeah, that was my thinking as well.

lib/mayaUsd/fileio/primUpdaterContext.h
vlasovi: Your plan will not fully work because ReplicateExtrasToUSD::processItem writes into the map.

ppt-adsk: O.K., FromUSD works, ToUSD does not :)  Let me think about it some more.

ppt-adsk: Here's a tweak to my suggestion to get it to work: everything as per above, but ReplicateExtrasToUSD::processItem() does a const cast if it needs to write into the map.  Alternately, the map is mutable.  I think this is acceptable, as the fact that processItem() is writing into the map is an invisible implementation detail.  In that way, the prim updater context can appear to be conceptually read-only.  What do you think?

ppt-adsk (approved):
Thanks for the changes, looks good!

PR #2630, opened 2022-09-27
MAYA-121636 fix fixlevel check of parent test

pierrebai-adsk:
Maya fixes were commited becore the fix that the test was expected. The fix level thus needs to be updated to one higher.


pierrebai-adsk:
A Maya branch that updated the fix level to 6 for a different fix was merged, so the test now needs to check for fix level 7 instead. The mayaUsd blessing will be broken until this is merged.

seando-adsk (approved):

pierrebai-adsk:
Two unrelated failure: OSX master CI failed to start. Linux test found a file in temp directory it did not like.

PR #2631, opened 2022-09-27
Disable edit as Maya for materials.

ppt-adsk:


lib/usd/translators/shading/materialUpdater.cpp
```diff
@@ -0,0 +1,46 @@
...
+MaterialUpdater::MaterialUpdater(
+    const UsdMayaPrimUpdaterContext& context,
+    const MFnDependencyNode&         depNodeFn,
+    const Ufe::Path&                 path)
+    : UsdMayaPrimUpdater(context, depNodeFn, path)
+{
+}
+
+/* virtual */
+bool MaterialUpdater::canEditAsMaya() const { return false; }
```
ppt-adsk, line 44: This is the single important line of this pull request: disabling edit as Maya for materials, as per design's request.  Perhaps we will find a workflow that will allow this in the future, but for now edit as Maya on a material does nothing and produces a warning.

test/lib/mayaUsd/fileio/testEditAsMaya.py
```diff
@@ -65,7 +64,6 @@ def tearDownClass(cls):
     def setUp(self):
         cmds.file(new=True, force=True)
 
-    @unittest.skipUnless(ufeFeatureSetVersion() >= 3, 'Test only available in UFE v3 or greater.')
```
ppt-adsk, line 68: The CMakeLists.txt file explicitly adds this test file only for UFE v3, so further checking in the test is redundant.

JGamache-autodesk (approved):

PR #2633, opened 2022-09-28
Fixes for building with gcc 11 (still using old ABI).

seando-adsk:


cmake/utils.cmake
```diff
@@ -274,12 +274,6 @@ function(get_external_project_default_values out_var)
     # Therefore avoid useless warnings in the log.
     list(APPEND setting_list --no-warn-unused-cli)
 
-    # Force the use of ABI version 0 on Linux.
-    # This is what Maya has been using for 2019...2023
-    if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
-        list(APPEND setting_list -D_GLIBCXX_USE_CXX11_ABI=0)
-    endif()
```
seando-adsk, line 281: I had previously added this here, where it gets used as a cmake arg. That was incorrect. It is really a "CMAKE_CXX_FLAGS" value.

cmake/googletest.cmake
```diff
@@ -35,6 +35,10 @@ macro(fetch_googletest)
             # Simply disable all warnings in googletest since we won't fix them anyways.
             # We will just update to newer version, if required.
             set(disable_all_warnings_flag -w)
+
+            # Force the use of ABI version 0 on Linux.
```
ppt-adsk, line 39: What about master / 2024?

ppt-adsk (approved):

PR #2634, opened 2022-09-28
 MAYA-125101 Correct handling of display list vs variants

pierrebai-adsk:
When undoing the deletion of an ancestor of a USD prim with variant which had some of its children edited as Maya, we need to validate that the correct set of variant is active before showing the edited Maya data.

Also, when going down the items of the trie that contains the UFE path component, we need to properly handle the points where we cross run-times (currenly there is only Maya and USD). When crossing runtimes, we must create a new segment.

### DIFF
--- a/lib/mayaUsd/fileio/orphanedNodesManager.cpp
+++ b/lib/mayaUsd/fileio/orphanedNodesManager.cpp
@@ -25,6 +25,7 @@
 #include <maya/MFnDagNode.h>
 #include <maya/MPlug.h>
 #include <ufe/hierarchy.h>
+#include <ufe/sceneSegmentHandler.h>
 #include <ufe/trie.imp.h>
 
 // For Tf diagnostics macros.
@@ -139,12 +140,12 @@ void OrphanedNodesManager::handleOp(const Ufe::SceneCompositeNotification::Op& o
     case Ufe::SceneCompositeNotification::OpType::ObjectAdd: {
         // Restoring a previously-deleted scene item may restore an orphaned
         // node.  Traverse the trie, and show hidden pull parents that are
-        // descendants of the argument path.  The trie node that corresponds to
-        // the added path is the starting point.  It may be an internal node,
-        // without data.
+        // descendants of the argument path that have all the proper variants.
+        // The trie node that corresponds to the added path is the starting
+        // point.  It may be an internal node, without data.
         auto ancestorNode = pulledPrims().node(op.path);
         TF_VERIFY(ancestorNode);
-        recursiveSetVisibility(ancestorNode, true);
+        recursiveSwitch(ancestorNode, op.path);
     } break;
     case Ufe::SceneCompositeNotification::OpType::ObjectDelete: {
         // The following cases will generate object delete:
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -207,22 +208,24 @@ void OrphanedNodesManager::handleOp(const Ufe::SceneCompositeNotification::Op& o
...
-                            c.GetPath().GetAsString(), MayaUsd::ufe::getUsdRunTimeId(), '/');
+                          child.GetPath().GetAsString(), MayaUsd::ufe::getUsdRunTimeId(), '/');
 
-                auto ancestorNode = pulledPrims().node(cPath);
+                auto ancestorNode = pulledPrims().node(childPath);
                 // If there is no ancestor node in the trie, this means that
                 // the new hierarchy is completely different from the one when
                 // the pull occurred, which means that the pulled object must
                 // stay hidden.
-                if (ancestorNode) {
```
pierrebai-adsk, line 222: Since the comment was describing the case where the if was not taken, I've inverted the condition so that the command now describes the if taken. Otherwise, it comment was confusing since it stated the opposite of the what was going on with the code.

lib/mayaUsd/ufe/UsdStageMap.cpp
```diff
@@ -182,7 +182,8 @@ MObject UsdStageMap::proxyShape(const Ufe::Path& path)
                 // Key is stale.  Remove it from our cache, and add the new entry.
                 auto count = fPathToObject.erase(cachedPath);
                 TF_AXIOM(count);
-                fPathToObject[newPath] = cachedObject;
+                if (!newPath.empty())
```
pierrebai-adsk, line 185: When undoing delete of stages, it sometimes occurred that the newPath was empty and caused a crash when used later on. Avoid filling the stage map with invalid paths.

test/lib/mayaUsd/fileio/testHideOrphanedNodes.py
```diff
@@ -245,5 +245,80 @@ def testHideOnNestingVariantSwitch(self):
...
+        self.assertFalse(pullParentVisibilityPlug[self.cPathStr].asBool())
+
+        # Revert back to variant selection a, pulled node is shown.
+        abVariant.SetVariantSelection('a')
+
+        self.assertTrue(pullParentVisibilityPlug[self.cPathStr].asBool())
+
+    def testShowOnCorrectVariantAndDeleteUndo(self):
+        '''
+        Edit-as-Maya an object, stay ton the correct variant in its ancestor,
```
ppt-adsk, line 293: "ton" --> "on", I believe.

lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -327,10 +330,38 @@ void OrphanedNodesManager::recursiveSwitch(
...
+                    Ufe::Hierarchy::ChildFilter filter
+                        = { Ufe::ChildFilterFlag("InactivePrims", "InactivePrims", 1) };
+                    for (auto child : itemHier->filteredChildren(filter)) {
+                        childRuntimeId = child->runTimeId();
+                        childSep = child->path().getSegments().back().separator();
+                        break;
+                    }
+                }
+            }
+        }
```
ppt-adsk, line 352: Unless I'm missing something, I don't think we need to do this.  We are the maya-usd plugin, and we have said that we would not support 3-segment paths --- pulled nodes are always single-segment Maya nodes, for compatibility with existing Maya code.

Therefore, the second segment will always be USD in this plugin, so we can simply get the USD run-time ID from getUsdRunTimeId() in Global.h, and use the USD path component separator '/'.  We can do this `if (Ufe::SceneSegmentHandler::isGateway(ufePath))`

ppt-adsk (changes requested):
Looks good, thanks for the fix!  I think the code in the orphaned node manager can be simplified.

PR #2638, opened 2022-10-03
Fix orphaned node discard edits with unloaded ancestor.

ppt-adsk:
As per MAYA-123378, discarding edits on an orphaned Maya node whose USD ancestor is unloaded causes an error.

Added a unit test for this case.  Without the fix, the test fails, with the fix, the test succeeds.

### DIFF
--- a/lib/mayaUsd/fileio/orphanedNodesManager.cpp
+++ b/lib/mayaUsd/fileio/orphanedNodesManager.cpp
@@ -259,6 +259,18 @@ bool OrphanedNodesManager::empty() const { return pulledPrims().root()->empty();
 
 void OrphanedNodesManager::restore(Memento&& previous) { _pulledPrims = previous.release(); }
 
+bool OrphanedNodesManager::isOrphaned(const Ufe::Path& pulledPath) const
+{
+    auto trieNode = pulledPrims().node(pulledPath);
+    if (!trieNode) {
+        // If the argument path has not been pulled, it can't be orphaned.
+        return false;
+    }
+    TF_VERIFY(trieNode->hasData());
+    // If the pull parent is visible, the pulled path is not orphaned.
+    return !getVisibilityPlug(trieNode);
+}
+
 /* static */
 bool OrphanedNodesManager::setVisibilityPlug(
     const Ufe::TrieNode<PullVariantInfo>::Ptr& trieNode,
@@ -269,6 +281,16 @@ bool OrphanedNodesManager::setVisibilityPlug(
     MFnDagNode  fn(pullParentPath);
     auto        visibilityPlug = fn.findPlug("visibility", /* tryNetworked */ true);
     return (visibilityPlug.setBool(visibility) == MS::kSuccess);
+}
+
+/* static */
+bool OrphanedNodesManager::getVisibilityPlug(const Ufe::TrieNode<PullVariantInfo>::Ptr& trieNode)
+{
+    TF_VERIFY(trieNode->hasData());
+    const auto& pullParentPath = trieNode->data().dagPath;
+    MFnDagNode  fn(pullParentPath);
+    auto        visibilityPlug = fn.findPlug("visibility", /* tryNetworked */ true);
+    return visibilityPlug.asBool();
 }
 
 /* static */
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1198,7 +1198,19 @@ bool PrimUpdaterManager::discardEdits(const MDagPath& dagPath)
...
+    auto ret = _orphanedNodesManager->isOrphaned(pulledPath)
+        ? discardOrphanedEdits(dagPath, pulledPath)
+        : discardPrimEdits(pulledPath);
+#else
+    // The following is incorrect: because of pull information in the session
+    // layer stored as overs, the usdPrim will never be invalid: a prim that
+    // exists only because of over opinions is valid, but is typeless.
+    // Therefore, the conditional will always succeed, and
+    // discardOrphanedEdits() is never called.  PPT, 30-Sep-2022.
     auto ret = usdPrim ? discardPrimEdits(pulledPath) : discardOrphanedEdits(dagPath, pulledPath);
```
ppt-adsk, line 1212: This is very uncomfortable: before the fix, discardOrphanedEdits() was never called.  However, the fix requires UFE support.

test/lib/mayaUsd/fileio/CMakeLists.txt
```diff
@@ -85,4 +84,25 @@ if(CMAKE_UFE_V3_FEATURES_AVAILABLE)
     )
 endif()
 
+if(CMAKE_UFE_V3_FEATURES_AVAILABLE)
+    set(TEST_DISCARD_EDITS testDiscardEdits.py)
+    mayaUsd_get_unittest_target(target ${TEST_DISCARD_EDITS})
+    if(UFE_TRIE_NODE_HAS_CHILDREN_COMPONENTS_ACCESSOR)
```
ppt-adsk, line 90: Enable test for orphaned node discard edits on ancestor unload only if UFE support is present.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1198,7 +1198,19 @@ bool PrimUpdaterManager::discardEdits(const MDagPath& dagPath)
...
+#ifdef HAS_ORPHANED_NODES_MANAGER
+    auto ret = _orphanedNodesManager->isOrphaned(pulledPath)
+        ? discardOrphanedEdits(dagPath, pulledPath)
+        : discardPrimEdits(pulledPath);
+#else
+    // The following is incorrect: because of pull information in the session
+    // layer stored as overs, the usdPrim will never be invalid: a prim that
+    // exists only because of over opinions is valid, but is typeless.
+    // Therefore, the conditional will always succeed, and
+    // discardOrphanedEdits() is never called.  PPT, 30-Sep-2022.
```
pierrebai-adsk, line 1211: Could we check if it is typeless? Would that help?

lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -259,6 +259,18 @@ bool OrphanedNodesManager::empty() const { return pulledPrims().root()->empty();
...
+bool OrphanedNodesManager::isOrphaned(const Ufe::Path& pulledPath) const
+{
+    auto trieNode = pulledPrims().node(pulledPath);
+    if (!trieNode) {
+        // If the argument path has not been pulled, it can't be orphaned.
+        return false;
+    }
+    TF_VERIFY(trieNode->hasData());
+    // If the pull parent is visible, the pulled path is not orphaned.
+    return !getVisibilityPlug(trieNode);
```
pierrebai-adsk, line 271: The only caveat I see with this approach, which we may not want to support, is: what if the user has set the node invisible by hand for whatever reason? Here we take for granted that invisible means orphaned.

pierrebai-adsk (approved):
Aside for a few interrogations, this looks fine.

lib/mayaUsd/fileio/orphanedNodesManager.cpp
ppt-adsk: If you touch the pull parent, which by default is not visible in the Outliner, you're voiding your warranty...  So yes, invisible pull parent means orphaned.

lib/mayaUsd/fileio/primUpdaterManager.cpp
ppt-adsk: That feels weak...  And yet another code path to support and test.  My hope is that since the required UFE support in Maya is coming very shortly on all Maya versions where this matters (UFE v3 and up), that we simply remove this broken code path soon.  Thanks for the thought, though.

PR #2641, opened 2022-10-05
Assign new material to multiple objects

stefanminning-autodesk:
This PR adds functionality to the existing `UsdUndoAssignNewMaterialCommand`, allowing for a new material to be assigned to multiple objects.

The command's original functionality remains untouched, as I wanted to avoid breaking API changes.


seando-adsk:
@stefanminning-autodesk Before we start the code review and preflight process can you please clean this up by squashing the 11 commits?

stefanminning-autodesk:
Sure, I've pushed with squashed history.

seando-adsk:
@womanyoyoyo-adsk Natalia can you please have a look at this from a design perspective? 
@stefanminning-autodesk From the code changes it looks to me as if you are always applying the material the selection, regardless of the item the context menu was opened on. Normally the behavior would be:
Q: "is the item on which the context menu was opened in the selection or not"?
Yes - apply operation to selection
No - apply operation only to context item

stefanminning-autodesk:
That's right, I've coded it to match the current behaviour, where a new material would be applied to an object when the right-click operation is applied upon it, regardless of its selection state.

Personally, I also find it somewhat confusing, but I'll leave it to the UX team to decide whether this is a good workflow or not.

Should be an easy change, either way.

seando-adsk:
@stefanminning-autodesk Actually sorry I don't think that is correct. With your changes it applies the material to **all** the selected objects. If the object you right click on is **not** selected, then the material is **not** applied to it. The normal context menu behavior would be "if the context object is selected?" then apply to selection, otherwise apply only to context object.

stefanminning-autodesk:
Ah yes, you're right: Looking at the code again, it should not be ignoring `fItem`. Could have sworn I'd accounted for that at some point... must have slipped through the cracks.

Anyway, guess I'll have to have another look at this when the designers are back from holidays and they've clarified the exact behaviour they'd like to see.

womanyoyoyo-adsk:
Hi, myself and Lookdevx designer are looking into this. Keep you posted.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.h
```diff
@@ -93,6 +96,9 @@ class MAYAUSD_CORE_PUBLIC UsdUndoAssignNewMaterialCommand : public Ufe::InsertCh
     UsdUndoAssignNewMaterialCommand(
         const UsdSceneItem::Ptr& parentItem,
         const std::string&       sdrShaderIdentifier);
+    UsdUndoAssignNewMaterialCommand(
+        const std::vector<const UsdSceneItem::Ptr>& parentItems,
```
hodoulp, line 100: There is already a predefined type `std::list<std::shared_ptr<SceneItem> > SceneItemList;` that the code can perhaps use.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.h
```diff
@@ -105,6 +111,11 @@ class MAYAUSD_CORE_PUBLIC UsdUndoAssignNewMaterialCommand : public Ufe::InsertCh
     //! sdrShaderIdentifier and assigns it to \p parentItem
     static UsdUndoAssignNewMaterialCommand::Ptr
     create(const UsdSceneItem::Ptr& parentItem, const std::string& sdrShaderIdentifier);
+    //! Create a UsdUndoAssignNewMaterialCommand that creates a new material based on \p
```
hodoulp, line 114: Not clear about Doxygen but I would have move `\p` close to the parameter name `sdrShaderIdentifier`

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -157,10 +157,32 @@ UsdUndoAssignNewMaterialCommand::UsdUndoAssignNewMaterialCommand(
...
 {
+    if (!parentItem || !parentItem->prim().IsActive())
+        return;
+
+    Ufe::Path parentPath        = parentItem->path();
+    const UsdStageWeakPtr stage = getStage(parentPath);
+    _stagesAndPaths[stage].push_back(parentPath);
+}
+
+UsdUndoAssignNewMaterialCommand::UsdUndoAssignNewMaterialCommand(
```
hodoulp, line 171: Does that code need `UFE_PREVIEW_XX` defines protection?

lib/mayaUsd/ufe/UsdUndoMaterialCommands.h
```diff
@@ -105,6 +111,11 @@ class MAYAUSD_CORE_PUBLIC UsdUndoAssignNewMaterialCommand : public Ufe::InsertCh
...
+    //! sdrShaderIdentifier and assigns it to multiple \p parentItems
+    static UsdUndoAssignNewMaterialCommand::Ptr 
+    create(const std::vector<const UsdSceneItem::Ptr>& parentItems,
+           const std::string&                          sdrShaderIdentifier);
```
hodoulp, line 118: Should the new method be protected by `UFE_PREVIEW_XX` defines?

lib/mayaUsd/ufe/UsdUndoMaterialCommands.h
```diff
@@ -93,6 +96,9 @@ class MAYAUSD_CORE_PUBLIC UsdUndoAssignNewMaterialCommand : public Ufe::InsertCh
     UsdUndoAssignNewMaterialCommand(
         const UsdSceneItem::Ptr& parentItem,
         const std::string&       sdrShaderIdentifier);
+    UsdUndoAssignNewMaterialCommand(
+        const std::vector<const UsdSceneItem::Ptr>& parentItems,
+        const std::string&                          sdrShaderIdentifier);
```
hodoulp, line 101: Should the new method be protected by UFE_PREVIEW_XX defines?

hodoulp (commented):
LGTM

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
stefanminning-autodesk: Good question, honestly not sure. 

These changes are entirely contained in MayaUSD, with no UFE update required. What UFE version would I be looking for?

lib/mayaUsd/ufe/UsdUndoMaterialCommands.h
stefanminning-autodesk: Good one, thank you. Much more readable.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -1249,10 +1249,22 @@ Ufe::UndoableCommand::Ptr UsdContextOps::doOpCmd(const ItemPath& itemPath)
...
+            sceneItems.push_back(fItem);
         }
+        const Ufe::GlobalSelection::Ptr selection = Ufe::GlobalSelection::get();
+        for (const auto& item : *selection) {
+            const UsdSceneItem::Ptr usdItem = std::dynamic_pointer_cast<UsdSceneItem>(item);
+            if (!usdItem || usdItem == fItem) {
+                continue;
+            }
+            sceneItems.push_back(usdItem);
+        }
```
seando-adsk, line 1265: This could be simplified by just using a Ufe Selection (which internally uses a SceneItemList).
```suggestion
        auto sceneItems = Ufe::GlobalSelection::get();
        sceneItems.append(fItem);
```

The Ufe Selection append will only add the item if its not already there. If you really need the item at the front then you can use insert instead. Then just pass this selection to the UsdUndoAssignNewMaterialCommand.

seando-adsk: No you cannot get the SceneItemList out of it. There are iterators for looping thru. But I would use Ufe::Selection not Ufe::GlobalSelection.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.h
```diff
@@ -116,8 +126,8 @@ class MAYAUSD_CORE_PUBLIC UsdUndoAssignNewMaterialCommand : public Ufe::InsertCh
     void connectShaderToMaterial(Ufe::SceneItem::Ptr shaderItem, PXR_NS::UsdPrim materialPrim);
     void markAsFailed();
 
-    const Ufe::Path   _parentPath;
-    const std::string _nodeId;
+    std::map<pxr::UsdStageWeakPtr, std::vector<const Ufe::Path>> _stagesAndPaths;
```
seando-adsk, line 129: Sorry I didn't catch this earlier. We shouldn't use the namespace name directly, but instead the define. So "pxr::" should be "PXR_NS::" (see line 126).

seando-adsk: Once this and the merge conflict is fixed, you can run another preflight and if it passes add the label "ready for merge" and I'll merge it. I confirmed with our designer that the multiple selection workflow is okay.

seando-adsk (approved):

seando-adsk:
@stefanminning-autodesk There are clang-format errors that need to be fixed. Please see [codingGuidelines](https://github.com/Autodesk/maya-usd/blob/dev/doc/codingGuidelines.md) for info on clang-format.

seando-adsk (approved):

seando-adsk (dismissed):

lib/mayaUsd/ufe/UsdContextOps.cpp
seando-adsk: Were you by chance getting the Ufe Selection into a reference pointer and them modifying it? You would just need to make sure to get a copy of it (using the copy constructor).

seando-adsk (changes requested):
I think you should also update the test in testContextOps.py to test the multiple selection scenario

test/lib/ufe/testContextOps.py
```diff
@@ -586,6 +586,160 @@ def checkItem(self, item, type, path):
...
+        cmds.redo()
+
+        checkMaterial(self, rootHier, 4, 1, 1, 0, "UsdPreviewSurface", "", "surface")
+
+        os.putenv("MAYAUSD_MATERIALS_SCOPE_NAME", "test_scope")
+        ufeCmd.execute(cmdSS)
+        checkMaterial(self, rootHier, 5, 1, 1, 0, "standard_surface", "mtlx", "out", "/test_scope")
+
+        # Clear the envvar, otherwise we're affecting the results of following tests.
+        os.putenv("MAYAUSD_MATERIALS_SCOPE_NAME", "")
```
seando-adsk, line 741: I think we should have a way to make sure this is automatically restored using a context manager object. Look at `TemporaryDirectory`. You can create one that will restore an env var. The input would be the env var and the new value. The __init__ will get the current value, store it and set the env var to input and the __exit__ will restore to stored value. Then you can wrap this section of code in a `with ...` statement.

seando-adsk (changes requested):
Thanks for the test. If you could just create a new context object and use it the two places that env var is set/restored that would be great. The new context object can be in testUtils.py

seando-adsk (approved):
Everything looks good now. Thanks.

seando-adsk:
@stefanminning-autodesk FYI, any new commit (such as your latest one merging in dev) will invalidate the preflight check and it must be run again. So I've removed the ready for merge label and re-started the preflight for you. When it passes, you may add back the label and I'll merge.

seando-adsk (approved):

PR #2645, opened 2022-10-12
Added test for getAllStages() versus delete.

ppt-adsk:
MAYA-111420 was entered to ensure that getAllStages() would not return stages whose proxy shape was deleted.  This bug was fixed by previous code, so simply added a test to demonstrate the fix, with undo and redo.

### DIFF
--- a/lib/mayaUsd/ufe/wrapUtils.cpp
+++ b/lib/mayaUsd/ufe/wrapUtils.cpp
@@ -17,8 +17,8 @@
 #include <mayaUsd/ufe/UsdSceneItem.h>
 #include <mayaUsd/ufe/Utils.h>
 
-#include <pxr/base/tf/stringUtils.h>
 #include <pxr/base/tf/pyResultConversions.h>
+#include <pxr/base/tf/stringUtils.h>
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/wrapUtils.cpp
```diff
@@ -236,6 +249,7 @@ void wrapUtils()
     // representation of Ufe::Path as comma-separated segments.  We know that
     // the USD path separator is '/'.  PPT, 8-Dec-2019.
     def("getStage", getStage);
+    def("getAllStages", getAllStages, return_value_policy<PXR_NS::TfPySequenceToList>());
```
ppt-adsk, line 252: Return value policy from example Pixar USD code, e.g.
https://github.com/PixarAnimationStudios/USD/blob/dc710925675f7f58f7a37f6c18d046a687b09271/pxr/usd/usd/wrapStageCache.cpp#L63
otherwise Boost Python complains at run-time that there is no converter for a vector of stages.

pierrebai-adsk (approved):

PR #2646, opened 2022-10-13
MAYA-124664 Update USD render when color prefs change

pierrebai-adsk:


vlasovi:
Looks great! One thing we can improve is reduce the bits we are dirtying inside _DirtyUsdSubtree. Option 1: add the dirtyBits parameter to the function _DirtyUsdSubtree and set it to MayaUsdRPrim::DirtySelectionHighlight. Option 2 : At the end of ProxyRenderDelegate::_Execute, inside the branch "if (reprSelector != HdReprSelector())" add a code like follows: "if (_colorPrefsChanged) dirtyBits |= MayaUsdRPrim::DirtySelectionHighlight;"

pierrebai-adsk:
The templated flag on the display layer also affects non-selected items, so I don't think we can only set the selection highlight dirty. What bits should be set for the general rendering of the items? Or does DirtySelectionHighlight really means "wireframe" style rendering ?

vlasovi:
Yes. SelectionHighlight really means all types of colors that could be applied to wireframe. In fact, your PR will fix much more than just template colors on display layers. It will also fix selection colors, active colors, dormant colors, etc.

vlasovi (dismissed):

PR #2647, opened 2022-10-13
MAYA-124644 - Create an auto-test for display layer attributes 'hide-on-playback', 'template' and 'reference'

vlasovi:


test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateDisplayLayers.py
```diff
@@ -70,6 +70,34 @@ def _GetMayaNode(self, nodeName):
...
+
+        panel = mayaUtils.activeModelPanel()
+        cmds.modelEditor(panel, edit=True, displayTextures=True)
+
+        testFile = testUtils.getTestScene("multipleMaterialsAssignment",
+                                          "MultipleMaterialsAssignment.usda")
+        proxyDagPath, usdStage = mayaUtils.createProxyFromFile(testFile)
+        cmds.select(cl=True)
+
+        # add the stage to a new display layer
```
pierrebai-adsk, line 88: Technically you are adding the root prim of the stage, not the stage proxy node, but the code reader will probably understand. (Beside, it is only a test.)

pierrebai-adsk (approved):

PR #2658, opened 2022-10-17
LOOKDEVX-784 | Makes use of USD_FORCE_DEFAULT_MATERIALS_SCOPE_NAME an…

feldstj:
…d MAYAUSD_MATERIALS_SCOPE_NAME environment variables for generating the scope name for assigned materials.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
@@ -17,11 +17,11 @@
 
 #include <mayaUsd/ufe/Utils.h>
 
+#include <pxr/base/tf/envSetting.h>
 #include <pxr/usd/sdr/registry.h>
 #include <pxr/usd/sdr/shaderProperty.h>
 #include <pxr/usd/usd/prim.h>
 #include <pxr/usd/usdUtils/pipeline.h>
-#include <pxr/base/tf/envSetting.h>
 
 #include <ufe/sceneItemOps.h>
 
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -194,6 +196,16 @@ static const std::string kDefaultMaterialScopeName("mtl");
...
 void UsdUndoAssignNewMaterialCommand::execute()
 {
+    std::string materialsScopeName = kDefaultMaterialScopeName;
+    if (TfGetEnvSetting(USD_FORCE_DEFAULT_MATERIALS_SCOPE_NAME)) {
+        materialsScopeName = UsdUtilsGetMaterialsScopeName().GetString();
+    } else {
+        const char* mayaUsdDefaultMaterialsScopeName = std::getenv("MAYAUSD_MATERIALS_SCOPE_NAME");
+        if (mayaUsdDefaultMaterialsScopeName) {
+            materialsScopeName = std::string(mayaUsdDefaultMaterialsScopeName);
+        }
```
seando-adsk, line 206: Why not use "TfGetenv" which returns you a std::string and you can provide a default value if the env var is not set.

seando-adsk (changes requested):

JGamache-autodesk (approved):

neilh-adsk (commented):
Un related, but is the Environment variable documented ?

feldstj:
> Un related, but is the Environment variable documented ?

I'm assuming we'll have documentation somewhere for at least the MAYAUSD variable.

neilh-adsk (approved):

PR #2663, opened 2022-10-18
Properly track usage of shared render items

JGamache-autodesk:
In #1339 code was added to share the selection render item between reprs. The code did not track usage, so as soon as one of the reprs is removed, the render item was removed from the subscene as well, leaving dangling pointers in the remaining reprs.

In #1597 code was added that seem to indicate there was suspicion about render item issues, but sharing was not identified as the cause.

This PR adds use count tracking to the render item so they only get removed from the subscene once all reprs have stopped using the shared render item.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/draw_item.cpp
@@ -27,7 +27,14 @@ namespace {
 class HdVP2DrawItemUserData : public MUserData
 {
 public:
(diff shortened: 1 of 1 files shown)


vlasovi (dismissed):
The change looks good. There exists another issue with shared render items, related to caching of their states. Every representation has its own copy of cached state, and so when we switch representations, the cache may become invalid when another representation change the state. We may need to address this isue later

vlasovi (approved):

PR #2664, opened 2022-10-18
MAYA-125816 fix nurbs curves UI when using export job context

pierrebai-adsk:
Only disable the Ui if there is actually a forced option for filter types.

### DIFF
--- a/plugin/adsk/scripts/mayaUsdTranslatorExport.mel
+++ b/plugin/adsk/scripts/mayaUsdTranslatorExport.mel
@@ -392,6 +392,7 @@ global proc mayaUsdTranslatorExport_SetFromOptions(string $currentOptions, int $
         tokenize($currentOptions, ";", $optionList);
         int $supportsMultiExport = 1;
         int $exportNurbsCurves = 1;
+        int $contextForcesFilterTypes = 0;
         for ($index = 0; $index < size($optionList); $index++) {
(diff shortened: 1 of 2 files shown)


plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -455,8 +457,10 @@ global proc mayaUsdTranslatorExport_SetFromOptions(string $currentOptions, int $
             }
         }
 
-        // We now know if we export nurbs curves or not:
-        checkBoxGrp -e -v1 $exportNurbsCurves -en $enable exportCurvesCheckBox;
+        // We now know if we export nurbs curves or not and if forced by the job context:
+        if ($contextForcesFilterTypes == 1) {
+            checkBoxGrp -e -v1 $exportNurbsCurves -en $enable exportCurvesCheckBox;
```
pierrebai-adsk, line 462: My understanding is the the NURBS curve Ui shuld *only* be disabled if the job context actually forced the filter types. Otherwise the previously-existing code would always disable that check-box as soon as the user selected a job context in the export dialog.

JGamache-autodesk (approved):

pierrebai-adsk:
The only PF failure is the known Windows interactive that refuses to even starts. All others, including interactive Linux, passed.

PR #2666, opened 2022-10-20
MAYA-125747 Implement filtered version of findGatewayItems().

frohnej-adsk:
Implement filtered version of `Ufe::SceneSegmentHandler::findGatewayItems()`.

Allows `findGatewayItems()` to filter the returned Selection by the specified nested runtime ID. For example, `findGatewayItems(path, usdRtid)` will only search for MayaUSD proxy shapes.

The according changes to UFE can be found in [this PR](https://git.autodesk.com/media-and-entertainment/ufe/pull/310).

### DIFF
--- a/lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
@@ -45,21 +45,29 @@ ProxyShapeSceneSegmentHandler::create(const Ufe::SceneSegmentHandler::Ptr& mayaS
 // Ufe::SceneSegmentHandler overrides
 //------------------------------------------------------------------------------
 
-Ufe::Selection ProxyShapeSceneSegmentHandler::findGatewayItems_(const Ufe::Path& path) const
+Ufe::Selection
+ProxyShapeSceneSegmentHandler::findGatewayItems_(const Ufe::Path& path, Ufe::Rtid nestedRtid) const
 {
-    // Handle other gateway node types that MayaUSD is not aware of
-    Ufe::Selection result = fMayaSceneSegmentHandler
-        ? fMayaSceneSegmentHandler->findGatewayItems_(path)
-        : Ufe::Selection();
+    Ufe::Selection result = Ufe::Selection();
 
-    // Find the MayaUSD proxyShapes
-    for (const auto& stage : getAllStages()) {
-        Ufe::Path proxyShapePath = stagePath(stage);
-        // recall that findGatewayItems searches for descendents of path that are
-        // gateway nodes. If path itself is a gateway node it should not be included
-        // in the results.
-        if (proxyShapePath.startsWith(path) && proxyShapePath != path) {
-            result.append(Ufe::Hierarchy::createItem(proxyShapePath));
+    // Handle other gateway node types that MayaUSD is not aware of.
+    // `nestedRtid` is used as a filter. If it matches the MayaUSD runtime ID, MayaUSD is aware of
+    // the requested gateway items so this step can be skipped.
+    if (nestedRtid != g_USDRtid && fMayaSceneSegmentHandler) {
+        result = fMayaSceneSegmentHandler->findGatewayItems_(path, nestedRtid);
+    }
+
+    // Find the MayaUSD proxyShapes.
+    // `nestedRtid` is used as a filter. Only add MayaUSD proxyShapes to the result if the argument
+    // matches the MayaUSD runtime ID or if the argument is 0 which means return all.
+    if (nestedRtid == 0 || nestedRtid == g_USDRtid) {
+        for (const auto& stage : getAllStages()) {
+            Ufe::Path proxyShapePath = stagePath(stage);
+            // recall that findGatewayItems searches for descendants of path that are gateway nodes.
+            // If path itself is a gateway node it should not be included in the results.
+            if (proxyShapePath.startsWith(path) && proxyShapePath != path) {
+                result.append(Ufe::Hierarchy::createItem(proxyShapePath));
+            }
         }
     }
 
--- a/lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
@@ -54,7 +54,7 @@ Ufe::Selection ProxyShapeSceneSegmentHandler::findGatewayItems_(const Ufe::Path&
 
     // Find the MayaUSD proxyShapes
     for (const auto& stage : getAllStages()) {
-        Ufe::Path proxyShapePath = stagePath(stage);
+        const Ufe::Path proxyShapePath = stagePath(stage);
         // recall that findGatewayItems searches for descendants of path that are
         // gateway nodes. If path itself is a gateway node it should not be included
         // in the results.
@@ -63,7 +63,7 @@ Ufe::Selection ProxyShapeSceneSegmentHandler::findGatewayItems_(const Ufe::Path&
         }
     }
 
-    // If there were Usd prims that are gateway items then we'd have an implementation
+    // TODO: If there were Usd prims that are gateway items then we'd have an implementation
     // of UsdSceneSegmentHandler that could find the gateway items and extra code
     // here to handle the case where isAGatewayType() for path is true. But right now
     // there are no gateway items in Usd, so I don't have to handle that.
@@ -87,7 +87,7 @@ ProxyShapeSceneSegmentHandler::findGatewayItems_(const Ufe::Path& path, Ufe::Rti
     // `nestedRtid` matches the MayaUSD runtime ID. Find the MayaUSD proxyShapes.
     Ufe::Selection result = Ufe::Selection();
     for (const auto& stage : getAllStages()) {
-        Ufe::Path proxyShapePath = stagePath(stage);
+        const Ufe::Path proxyShapePath = stagePath(stage);
         // recall that findGatewayItems searches for descendants of path that are
         // gateway nodes. If path itself is a gateway node it should not be included
         // in the results.
@@ -96,7 +96,7 @@ ProxyShapeSceneSegmentHandler::findGatewayItems_(const Ufe::Path& path, Ufe::Rti
         }
     }
 
-    // If there were Usd prims that are gateway items then we'd have an implementation
+    // TODO: If there were Usd prims that are gateway items then we'd have an implementation
     // of UsdSceneSegmentHandler that could find the gateway items and extra code
     // here to handle the case where isAGatewayType() for path is true. But right now
     // there are no gateway items in Usd, so I don't have to handle that.
--- a/lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
@@ -45,6 +45,7 @@ ProxyShapeSceneSegmentHandler::create(const Ufe::SceneSegmentHandler::Ptr& mayaS
 // Ufe::SceneSegmentHandler overrides
 //------------------------------------------------------------------------------
 
+#if (UFE_PREVIEW_VERSION_NUM >= 4033)
 Ufe::Selection
 ProxyShapeSceneSegmentHandler::findGatewayItems_(const Ufe::Path& path, Ufe::Rtid nestedRtid) const
 {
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
```diff
@@ -45,6 +45,41 @@ ProxyShapeSceneSegmentHandler::create(const Ufe::SceneSegmentHandler::Ptr& mayaS
...
+    // `nestedRtid` is used as a filter. If it matches the MayaUSD runtime ID, MayaUSD is aware of
+    // the requested gateway items so this step can be skipped.
+    if (nestedRtid != g_USDRtid && fMayaSceneSegmentHandler) {
+        result = fMayaSceneSegmentHandler->findGatewayItems_(path, nestedRtid);
+    }
+
+    // Find the MayaUSD proxyShapes.
+    // `nestedRtid` is used as a filter. Only add MayaUSD proxyShapes to the result if the argument
+    // matches the MayaUSD runtime ID or if the argument is 0 which means return all.
+    if (nestedRtid == 0 || nestedRtid == g_USDRtid) {
```
JGamache-autodesk, line 64: Zero does not feel magic enough as it is the RtID of the first registered handler as seen in  `Ufe::RunTimeMgr::register_()`. Can you explictly declare a `Ufe::RuntimeMgr::AllRtID` constant with value `std::numeric_limits<RtID>::max` and use that instead?

JGamache-autodesk (changes requested):

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
frohnej-adsk: As far as I understood it, `0` is an invalid runtime ID. This is also indicated by [this comment](https://git.autodesk.com/media-and-entertainment/ufe/blob/3624d2f225abd3044fcba31cfa6c4760deda92fe/src/runTimeMgr.cpp#L76). 

I think this is why `0` was used in the [proposed interface](https://git.autodesk.com/media-and-entertainment/ufe/pull/297). It may still make sense to add a constant like `Ufe::RuntimeMgr::AllRtid` or `Ufe::RuntimeMgr::InvlaidRtid` but I feel like the value itself should be fine.

JGamache-autodesk: Agree, zero is the correct value, but please still introduce it as a constant. Thanks.

seando-adsk: The Ufe rtid.h does indeed have a comment stating that "ID 0 is used to represent an invalid runtime ID." - its too bad it didn't also include a constant for that. 

BTW, this is the open source github repo so we should refrain from using links to our internal repos.

seando-adsk: It looks like the upper section of this method (in the new #if) is new and the bottom part is common to both if/else. If so can we just have the if inside this method for new code?

seando-adsk: Who is calling the new version with the extra Rtid param?

frohnej-adsk: The new version of the method will be used by LookdevX to get a list of gateways into USD. 

Having the `#if`/`#else` inside of the method would be great but I don't think we can, can we? The new version changes the signature of the method, so the signature has to be part of the `#if`/`#else`. Most of the code inside of the method is touched as well (i.e., enclosed in some comparison with `nestedRtid`). The only thing that remains completely unchanged is the comment about `If there were Usd prims that are gateway items` and the return statement.

We could have something like this, but I felt like it would be less readable:
```C++
#if (UFE_PREVIEW_VERSION_NUM >= 4033)
Ufe::Selection ProxyShapeSceneSegmentHandler::findGatewayItems_(const Ufe::Path& path, Ufe::Rtid nestedRtid) const
{
    // new code
#else
Ufe::Selection ProxyShapeSceneSegmentHandler::findGatewayItems_(const Ufe::Path& path)
{
    // old code
#endif
    // common code
}
```
Let me know what you think. Maybe there is another way to do it more elegantly that I'm not aware of.

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
```diff
@@ -45,6 +45,41 @@ ProxyShapeSceneSegmentHandler::create(const Ufe::SceneSegmentHandler::Ptr& mayaS
...
+    if (nestedRtid != g_USDRtid && fMayaSceneSegmentHandler) {
+        result = fMayaSceneSegmentHandler->findGatewayItems_(path, nestedRtid);
+    }
+
+    // Find the MayaUSD proxyShapes.
+    // `nestedRtid` is used as a filter. Only add MayaUSD proxyShapes to the result if the argument
+    // matches the MayaUSD runtime ID or `Ufe::kAllRtid` which is used to refer to all runtimes.
+    if (nestedRtid == g_USDRtid || nestedRtid == Ufe::kAllRtid) {
+        for (const auto& stage : getAllStages()) {
+            Ufe::Path proxyShapePath = stagePath(stage);
```
hodoulp, line 66: [Minor] The local variable must be a const.

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
```diff
@@ -45,6 +45,41 @@ ProxyShapeSceneSegmentHandler::create(const Ufe::SceneSegmentHandler::Ptr& mayaS
...
+Ufe::Selection
+ProxyShapeSceneSegmentHandler::findGatewayItems_(const Ufe::Path& path, Ufe::Rtid nestedRtid) const
+{
+    Ufe::Selection result = Ufe::Selection();
+
+    // Handle other gateway node types that MayaUSD is not aware of.
+    // `nestedRtid` is used as a filter. If it matches the MayaUSD runtime ID, MayaUSD is aware of
+    // the requested gateway items so this step can be skipped.
+    if (nestedRtid != g_USDRtid && fMayaSceneSegmentHandler) {
+        result = fMayaSceneSegmentHandler->findGatewayItems_(path, nestedRtid);
```
hodoulp, line 58: Based on the two conditions used in this method's implementation, it seems that the code can directly return.

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
```diff
@@ -45,6 +45,41 @@ ProxyShapeSceneSegmentHandler::create(const Ufe::SceneSegmentHandler::Ptr& mayaS
...
+            if (proxyShapePath.startsWith(path) && proxyShapePath != path) {
+                result.append(Ufe::Hierarchy::createItem(proxyShapePath));
+            }
+        }
+    }
+
+    // If there were Usd prims that are gateway items then we'd have an implementation
+    // of UsdSceneSegmentHandler that could find the gateway items and extra code
+    // here to handle the case where isAGatewayType() for path is true. But right now
+    // there are no gateway items in Usd, so I don't have to handle that.
```
hodoulp, line 78: [Question] Is it permanent (i.e., it will never happen) -- or -- transient (e.g., at some point the feature will be implemented) ?

If that's permanent the explanation must only mention `USD` prims cannot be gateways.
If that's transient, I suggest to append a `TODO: ` tag to highlight that it could change at some point.

hodoulp (changes requested):

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
frohnej-adsk: No, I don't think so. If `nestedRtid` is `Ufe::kAllRtid`, the method has to do both: Handle gateways that MayaUSD is not aware of (this if block) **and** find the MayaUSD proxyShapes (next if block).

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
frohnej-adsk: I'm not sure. I simply copied the comment from the old implementation. Just from reading the comment I would lean towards transient as it says "But *right now* there are no gateway items in Usd".

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
seando-adsk: I think based on the latest changes request to ufe, this will become an overloaded method with the new Rtid param.

frohnej-adsk:
I think I have addressed all the requested changes. @hodoulp @JGamache-autodesk 

@JGamache-autodesk We decided to overload `findGatewayItems_()`:
```C++
// Unfiltered version that returns gateways into all runtimes.
virtual Selection findGatewayItems_(const Path& path) const = 0;

// Filtered version that only returns gateways into the specified runtime.
virtual Selection findGatewayItems_(const Path& path, Rtid nestedRtid) const = 0;
```
This gets rid of the default parameter and the need for a constant.

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
```diff
@@ -54,23 +54,57 @@ Ufe::Selection ProxyShapeSceneSegmentHandler::findGatewayItems_(const Ufe::Path&
 
     // Find the MayaUSD proxyShapes
     for (const auto& stage : getAllStages()) {
-        Ufe::Path proxyShapePath = stagePath(stage);
-        // recall that findGatewayItems searches for descendents of path that are
+        const Ufe::Path proxyShapePath = stagePath(stage);
+        // recall that findGatewayItems searches for descendants of path that are
```
JGamache-autodesk, line 58: Original spelling, with an "e", was correct.

JGamache-autodesk (approved):

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
ppt-adsk: Both are correct, but according to Merriam-Webster spelling with an 'a' is more common:
https://www.merriam-webster.com/dictionary/descendant

JGamache-autodesk: I learned something today. Thanks!

neilh-adsk: this is why I love code reviews... and just a heads up, I am terrible at the English language :-)

seando-adsk (dismissed):
This is cleaner and easier to read now with an overloaded function.

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
```diff
@@ -54,23 +54,57 @@ Ufe::Selection ProxyShapeSceneSegmentHandler::findGatewayItems_(const Ufe::Path&
...
+    // Handle other gateway node types that MayaUSD is not aware of.
+    if (nestedRtid != g_USDRtid) {
+        // `nestedRtid` is used as a filter. If it doesn't match the MayaUSD runtime ID, the method
+        // can return early.
+        return fMayaSceneSegmentHandler
+            ? fMayaSceneSegmentHandler->findGatewayItems_(path, nestedRtid)
+            : Ufe::Selection();
+    }
+
+    // `nestedRtid` matches the MayaUSD runtime ID. Find the MayaUSD proxyShapes.
```
hodoulp, line 87: The lines 89 to 102 seems a copy & paste from lines 55 to 69. The code must be factorized in a protected (or private) method with `Ufe::Selection & result` as in & out parameter.

test/lib/ufe/testSceneSegment.py
```diff
@@ -83,5 +84,46 @@ def testProxyShapeSceneSegmentHandler(self):
...
+
+        usdRunTimeId = ufe.RunTimeMgr.instance().getId('USD')
+        result = handler.findGatewayItems(proxyShapePath, usdRunTimeId)
+        self.assertTrue(result.empty())
+
+        otherRunTimeId = 6174
+        result = handler.findGatewayItems(proxyShapePath, otherRunTimeId)
+        self.assertTrue(result.empty())
+
+        # Searching the the parent of a gateway item searches the Maya
```
hodoulp, line 110: typo `Searching the the parent` -> `Searching from the parent`

ppt-adsk (approved):

test/lib/ufe/testSceneSegment.py
```diff
@@ -83,5 +84,46 @@ def testProxyShapeSceneSegmentHandler(self):
...
+        # scene segment for gateway nodes without recursing into USD.
+        # If no filter is specified or if the USD runtime ID is used as
+        # a filter, this should return the proxy shape. If a different
+        # runtime ID is used as a filter, the result should be empty.
+        handler = ufe.RunTimeMgr.instance().sceneSegmentHandler(proxyShapeParentPath.runTimeId())
+        
+        result = handler.findGatewayItems(proxyShapeParentPath)
+        self.assertTrue(result.contains(proxyShapePath))
+        self.assertEqual(len(result), 1)
```
hodoulp, line 119: Add a check to validate the expected path like self.assert(ufe.Path.string(result[0]), 'xvcxvcx')

test/lib/ufe/testSceneSegment.py
```diff
@@ -83,5 +84,46 @@ def testProxyShapeSceneSegmentHandler(self):
...
+
+        result = handler.findGatewayItems(proxyShapeParentPath, usdRunTimeId)
+        self.assertTrue(result.contains(proxyShapePath))
+        self.assertTrue(len(result), 1)
```
hodoulp, line 123: Add a check to validate the expected path like self.assert(ufe.Path.string(result[0]), 'xvcxvcx')

hodoulp (changes requested):

lib/mayaUsd/ufe/ProxyShapeSceneSegmentHandler.cpp
frohnej-adsk: I factorized the actual code to find the MayaUSD proxyShapes but left the `TODO` block as is. I feel like the `TODO` will have to be handled differently in the filtered and unfiltered method.

test/lib/ufe/testSceneSegment.py
frohnej-adsk: This should already be covered by the current asserts. If `result.contains(proxyShapePath)` and `len(result) == 1` is true, then `result[0] == proxyShapePath` has to be true as well.

hodoulp (approved):

ppt-adsk (approved):

seando-adsk (approved):

PR #2667, opened 2022-10-20
MAYA-125835 - Temporarily disable support for display layers in EditAsMaya workflows to avoid the crash

vlasovi:


seando-adsk (approved):
Okay. I thought we talked about reverting this, but disabling is okay as well.

PR #2670, opened 2022-10-21
MAYA-125867 fix edit-as-Maya on fresh Maya session

pierrebai-adsk:
When a USD object is edited as Maya and the Maya scene is saved, if Maya is restarted and teh scene reloaded, the edited USD object cannot be manipulated again. That was due to internal non-saved state not being set properly when restarting Maya.

Avoid the problem by not using non-persistent state.

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.h
+++ b/lib/mayaUsd/fileio/primUpdaterManager.h
@@ -79,7 +79,7 @@ public:
     MAYAUSD_CORE_PUBLIC
     static bool readPullInformation(const MDagPath& dagpath, Ufe::Path& ufePath);
 
-    bool hasPulledPrims() const { return _hasPulledPrims; }
+    bool hasPulledPrims() const;
 
 private:
     PrimUpdaterManager();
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/fileio/primUpdaterManager.h
```diff
@@ -128,11 +128,6 @@ class PrimUpdaterManager : public PXR_NS::TfWeakBase
 
     bool _inPushPull { false };
 
-    // Becomes true when there is at least one pulled prim.
-    // The goal is to let code that can be optimized when there is no pull prim
-    // to check rapidly.
```
seando-adsk, line 133: What about this optimization? Now you are using findPullRoot() which searches thru the entire dag hierarchy looking for one.

pierrebai-adsk: It does search the entire DAG.

Only the direct children of the root are iterated. Note that we do that find operation whenever we actually edit-as-Maya or any other edit-related op, so it surely is not slow.

If we do find it became a bottleneck, we can even optimize it further by using a MSelection with the pull-set name, which would be even faster.

pierrebai-adsk: I changed findPullRoot() to use MSelectionList, so now there is no iteration at all, just direct access to the object, so there should be no performance risk. Bonus: the code is also faster for the actual operations.

seando-adsk (approved):

lib/mayaUsd/fileio/primUpdaterManager.h
seando-adsk: Great that is much better.

PR #2673, opened 2022-10-24
MAYA-123719 - Bringing back the display layers in EditAsMaya support by fixing the crash

vlasovi:


lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -971,16 +969,23 @@ void ReplicateExtrasToUSD::processItem(const MDagPath& dagPath, const SdfPath& u
...
         MFnDisplayLayerManager displayLayerManager(
             MFnDisplayLayerManager::currentDisplayLayerManager());
-        _primToLayerMap[usdPath] = displayLayerManager.getLayer(dagPath);
+
+        MObject displayLayerObj = displayLayerManager.getLayer(dagPath);
+        if (displayLayerObj.hasFn(MFn::kDisplayLayer)) {
+            MFnDisplayLayer displayLayer(displayLayerObj);
+            // UFE display layers coming from referenced files are not yet supported in Maya
+            // and their usage leads to a crash, so skip those for the time being
+            if (!displayLayer.isFromReferencedFile()) {
```
seando-adsk, line 978: I see this test here (and above) as well as the code to test hasFn and create MFnDisplayLayer from the MObject. How about creating a helper function like so:

MFnDisplayLayer supportedDisplayLayer(MObject)

This function would do all that (with the ref file test).

seando-adsk: In my changes built ontop of yours I'll also use this helper function.

seando-adsk (approved):

Issue #1773, opened 2021-10-19
[MAYA-114501] Allow users to add site specific items to context menus and radial menus


dj-mcg:
**Is your feature request related to a problem? Please describe.**
Users have requested items in the Outliner context menu and in various radial menus in the viewport, many of which are specific to studio needs.

**Describe the solution you'd like**
It would be nice if a plugin mechanism was provided to allow studios or users to add custom commands to context menus and radial menus.

**Describe alternatives you've considered**
We could hack the code itself and have it live in an internal patch branch, but a plugin mechanism would be cleaner and would allow us to avoid any future merge conflicts.

oumad:
This is definitely something we need, I've actually been looking how to hack this in, but if there was a way to add such options in a much simpler/cleaner way, ideally with python scripting. That would help us to greatly improve our artists worfkflow.

seando-adsk:
Are you asking for a MayaUsd plugin mechanism to or something from Maya?

From Maya, I would suggest looking at the ```callbacks``` command in Maya. There are various menu item hooks, like ```addRMBBakingMenuItems``` which you can use to add items to the context menu when RMB over an object in the viewport. We are using this hook in the mayaUsd plugin to add some menus.

There is also a new hook added available in the latest Maya PR ```addItemsToOutlinerNodePopupMenu``` which you can use to add menu items to the context menu in the Outliner. We are also using this hook in the mayaUsdPlugin.

We don't have anything from MayaUsd, but it does use Ufe and the new Ufe ```ContextOps``` interface available in Ufe v2 (Maya 2022) to add menu items. https://help.autodesk.com/cloudhelp/2022/ENU/Maya-SDK/ufe_ref/class_ufe_1_1_context_ops.html

Does that help?
Sean

seando-adsk:
There is a new menu hook (available in Maya 2023) that will be called when you RMB on a **non-Maya Ufe item** in the viewport. It looks for a MEL menu proc: <ufe runtime name> + "MenuProc" 
We are using this in the MayaUsd plugin: https://github.com/Autodesk/maya-usd/blob/dev/plugin/adsk/scripts/USDMenuProc.mel#L62

oumad:
[...]

PR #2674, opened 2022-10-25
MAYA-125745 - Implementation of UFE Material Interface

stefanminning-autodesk:
This PR implements the newly added UFE Material Interface, mainly intended for internal use in the LookdevX team.

The implemented functionality provides access to UsdSceneItems' materials, which was previously unavailable to us.


feldstj (dismissed):
LGTM! Just one super minor picky point.

lib/mayaUsd/ufe/UsdMaterial.h
```diff
@@ -0,0 +1,50 @@
...
+
+    UsdMaterial(const UsdSceneItem::Ptr& item);
+    ~UsdMaterial() override;
+
+    // Delete the copy/move constructors assignment operators.
+    UsdMaterial(const UsdMaterial&) = delete;
+    UsdMaterial& operator=(const UsdMaterial&) = delete;
+    UsdMaterial(UsdMaterial&&) = delete;
+    UsdMaterial& operator=(UsdMaterial&&) = delete;
+
```
hodoulp, line 36: For the Doxygen documentation of the previous group of methods, the syntax is:
```
//@{
///  Delete the copy/move constructors assignment operators.
<...>
//@}
```

lib/mayaUsd/ufe/UsdMaterialHandler.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+//------------------------------------------------------------------------------
+// UsdMaterialHandler overrides
+//------------------------------------------------------------------------------
+
+Ufe::Material::Ptr UsdMaterialHandler::material(const Ufe::SceneItem::Ptr& item) const
+{
+    UsdSceneItem::Ptr usdItem = std::dynamic_pointer_cast<UsdSceneItem>(item);
+#if !defined(NDEBUG)
```
hodoulp, line 38: The directive is useless as the `assert` is gone in release mode. Anyway the code must return if `usdItem` is null.

lib/mayaUsd/ufe/UsdMaterial.cpp
```diff
@@ -0,0 +1,125 @@
...
+std::vector<PXR_NS::UsdPrim> UsdMaterial::getConnectedShaders(const PXR_NS::UsdShadeOutput& port) const
+{
+    // Dig down across NodeGraph boundaries until a surface shader is found.
+    std::vector<PXR_NS::UsdPrim> shaders;
+    for (const auto& sourceInfo : port.GetConnectedSources()) {
+        if (sourceInfo.source.GetPrim().GetTypeName() != "Shader") {
+            if (sourceInfo.sourceType == UsdShadeAttributeType::Output) {
+                const std::vector<PXR_NS::UsdPrim> connectedShaders
+                    = getConnectedShaders(sourceInfo.source.GetOutput(sourceInfo.sourceName));
+                shaders.insert(end(shaders), begin(connectedShaders), end(connectedShaders));
```
hodoulp, line 52: For `end` and `begin` the code should be `std::end` and `std::begin`. Why the compilation succeeds?

lib/mayaUsd/ufe/UsdMaterialHandler.h
```diff
@@ -0,0 +1,45 @@
...
+    typedef std::shared_ptr<UsdMaterialHandler> Ptr;
+
+    UsdMaterialHandler();
+    ~UsdMaterialHandler() override;
+
+    // Delete the copy/move constructors assignment operators.
+    UsdMaterialHandler(const UsdMaterialHandler&) = delete;
+    UsdMaterialHandler& operator=(const UsdMaterialHandler&) = delete;
+    UsdMaterialHandler(UsdMaterialHandler&&) = delete;
+    UsdMaterialHandler& operator=(UsdMaterialHandler&&) = delete;
```
hodoulp, line 34: Same comment here concerning the Doxygen syntax.

hodoulp (commented):
LGTM. Only few minor comments.

lib/mayaUsd/ufe/CMakeLists.txt
```diff
@@ -96,6 +96,20 @@ if (UFE_LIGHTS_SUPPORT)
     )
 endif()
 
+message(STATUS "UFE_MATERIALS_SUPPORT is ${UFE_MATERIALS_SUPPORT}")
```
JGamache-autodesk, line 99: This printf might not be necessary anymore.

lib/mayaUsd/ufe/UsdMaterial.cpp
JGamache-autodesk: Looks like a case of [Argument-dependent Lookup](https://en.cppreference.com/w/cpp/language/adl). The namespace of std::map::insert ca be searched for begin/end functions.

lib/mayaUsd/ufe/UsdMaterial.cpp
```diff
@@ -0,0 +1,125 @@
...
+    std::vector<PXR_NS::UsdPrim> materialPrims;
+
+    const PXR_NS::UsdPrim&                            prim = fItem->prim();
+    PXR_NS::UsdShadeMaterialBindingAPI                bindingApi(prim);
+    PXR_NS::UsdShadeMaterialBindingAPI::DirectBinding directBinding = bindingApi.GetDirectBinding();
+
+    // 1. Simple case: A material is directly attached to our object.
+    const PXR_NS::UsdShadeMaterial material = directBinding.GetMaterial();
+    if (material) {
+        materialPrims.push_back(material.GetPrim());
```
JGamache-autodesk, line 82: That is a UsdShadeMaterial. You must iterate GetSurfaceOuputs and getConnectedShaders to get the same type as will be found on geom subsets.

lib/mayaUsd/ufe/UsdMaterial.h
```diff
@@ -0,0 +1,50 @@
...
+    //! Create a UsdMaterial.
+    static UsdMaterial::Ptr create(const UsdSceneItem::Ptr& item);
+
+    std::vector<Ufe::SceneItem::Ptr> getMaterials() const override;
+
+private:
+    std::vector<PXR_NS::UsdPrim> getConnectedShaders(const PXR_NS::UsdShadeOutput& port) const;
```
JGamache-autodesk, line 43: That function does not use fItem, so it could possibly be moved to an unnamed namespace in the .cpp file.

lib/mayaUsd/ufe/UsdMaterialHandler.cpp
JGamache-autodesk: The equivalent in USD land is TF_AXIOM. Grep a bit to see how it is used. Requires the directive to use the PXR namespace.

plugin/adsk/scripts/USDMenuProc.mel
```diff
@@ -76,6 +76,14 @@ global proc USDMenuProc(string $parent, string $obj)
         }
         menuItem -label "Duplicate As Maya Data" -command ("{waitCursor -state 1; mayaUsdMenu_duplicate \"" + $obj + "\"; waitCursor -state 0;}");
     }
+
+    // Allow LookdevX to define a proc with additional menu items
```
JGamache-autodesk, line 80: Not clean. Requires secret functions shared with a third party.
Please create a publicly known function to register additional menu procs, then use the list of registered procs here. The third party will then be able to register a proc without MayaUSD having to know about it.

JGamache-autodesk (changes requested):

lib/mayaUsd/ufe/CMakeLists.txt
seando-adsk: Correct, the message is output by FindUFE.cmake

lib/mayaUsd/ufe/UsdMaterial.h
```diff
@@ -0,0 +1,50 @@
...
+
+private:
+    UsdSceneItem::Ptr fItem;
```
seando-adsk, line 46: The [coding standard](https://github.com/Autodesk/maya-usd/blob/dev/doc/codingGuidelines.md) for MayaUsd is to use _item

lib/mayaUsd/ufe/UsdMaterial.cpp
```diff
@@ -0,0 +1,125 @@
...
+        } else {
+            shaders.push_back(sourceInfo.source.GetPrim());
+        }
+    }
+    return shaders;
+}
+
+std::vector<Ufe::SceneItem::Ptr> UsdMaterial::getMaterials() const
+{
+    PXR_NAMESPACE_USING_DIRECTIVE
```
seando-adsk, line 63: Why is this here? I see you are correctly using PXR_NS

lib/mayaUsd/ufe/UsdMaterial.h
```diff
@@ -0,0 +1,50 @@
+// ===========================================================================
+// Copyright 2022 Autodesk, Inc. All rights reserved.
+//
+// Use of this software is subject to the terms of the Autodesk license
+// agreement provided at the time of installation or download, or which
+// otherwise accompanies this software in either electronic or hard copy form.
+// ===========================================================================
```
seando-adsk, line 7: Every file needs a license notice and copyright. Check any of the newer files for an example.

plugin/adsk/scripts/USDMenuProc.mel
seando-adsk: This was code added to follow an existing style already used in Maya to call a user defined menu proc. What Jerry is suggesting is better for what you are trying to add which is a 3rd party (not user).

seando-adsk (changes requested):

lib/mayaUsd/ufe/UsdMaterialHandler.cpp
```diff
@@ -0,0 +1,52 @@
...
+    assert(usdItem);
```
ppt-adsk, line 39: Better to use TF_VERIFY and return nullptr.  See
https://github.com/Autodesk/maya-usd/blob/42854b32c9c1cd706917e088b477119ca1d791ab/lib/mayaUsd/ufe/UsdShaderNodeDefHandler.cpp#L50

ppt-adsk (changes requested):
Needs a test.

ppt-adsk:
Hi @stefanminning-autodesk, can you take me off this code review?  Thanks!

seando-adsk (dismissed):
Thanks for the test. In the future once a review has started and has comments, please don't squash. We decided this was not ideal since we lose the code review history in the pull-request (comments are no longer attached to commits since the commits don't exist anymore).

lib/mayaUsd/ufe/CMakeLists.txt
JGamache-autodesk: Might be a good idea if you have the time.

plugin/adsk/scripts/USDMenuProc.mel
JGamache-autodesk: Actually `USDUserMenuProc` was added specifically for the needs of LookdevX (and also other third parties, as seen in #1773), so my suggestion is that LookdevX should use `USDUserMenuProc` until we get news about another plugin hijacking it.

MAYA-125745 should be about finding a way to make that extensible enough that two plugins can register their menu hooks simultaneously.

JGamache-autodesk (changes requested):

JGamache-autodesk (approved):

seando-adsk (approved):

PR #2684, opened 2022-10-26
MAYA-125036 wipe and restore orphaned session data

pierrebai-adsk:
When an edited-as-Maya object is under a variant and that variant selection changes, we need to remove the information that was stored in the session layer about that edition. Note that if a prim with the same name is under the new variant, currently it cannot be edited. If we allow edition of variants with the same name, it would make the session cleanup harder since we would need to also restore the info from one edition at the same we remove the info from the other edition.

Note that the implementation would be broken by rename or reparent of ancestors, but there is another story to handle that.

Update the data held by the orphaned nodes manager:
- Add the DAG path to the proxy shape and edited Maya root to the pull variant info in order to be able to set or reset session info.
- Pass the edited Maya root to the add() function of the orphan manager.
- Add helper function to convert a trie node of pull variant info to a valid UFE path.
- Implement the clearing and restoring of session info when orphaning.
- Remove the explicit setting of prim activation since the clearing and restoring of session info now does it.
- Add the new pull variant info to the JSON I/O used to save the orphan data.

Renamed functions referencing visibility to instead reference orphans to make the intention clearer.
- So, recursiveSetVisibility becomes recursiveSetOrphaned.
- So, setVisibilityPlug becomes setOrphaned.
- getVisibilityPlug is no longer needed.

Refactored the prim updater manager to make some session info functions shared with the orphan manager. This centralizes pull information code and simplifies the PrimUpdaterManager class.
- Renamed generic "path" variable names to more specific "editedAsMayaRoot" to make the code clearer.
- Renamed the AddPullPathUndoItem class to RecordPullVariantInfoUndoItem to make the code clearer.
- Especially since it now records more than just the pull path...
- Renamed RemovePullPathUndoItem to RemovePullVariantInfoUndoItem for the same reason.
- Moved all pull information reading and writing into its own header and C++ file.
- Refactor all callers to use these new functions.

Modified an existing unit test that had two variants with prims of the same name to verify that the session layer data gets removed and restored properly.


pierrebai-adsk:
Note: the changes are a sub-set of what is in https://github.com/Autodesk/maya-usd/pull/2690. I applied the requested changes from that PR, you can verify them here.

ppt-adsk (approved):

pierrebai-adsk:
The 2020 and 2019 PF failed in a single common unit test: testUsdExportMesh. The error is:

   Error: Missing OpenColorIO configuration file: <MAYA_RESOURCES>/OCIO-configs/Maya-legacy/config.ocio
   # RuntimeError: Plug-in, "px_render", was not found on MAYA_PLUG_IN_PATH.

Seems unrelated to my changes.

pierrebai-adsk:
I think a new version of UFE broke MaysUSD... it fails on Linux and OSX because a newvalue was added to an enum.

pierrebai-adsk:
@ppt-adsk could you take a final look at this PR? I had to do fixes for Linux/MacOS.

ppt-adsk (approved):

PR #2687, opened 2022-10-26
MAYA-125378 Implementation batched duplication

JGamache-autodesk:
This allows Maya to duplicate multiple items and to properly fixup connections and relationships between all these individual duplicates.

Also implements the "no connections" mode that is the default Maya behavior. Use "Duplicate Special" with the "Duplicate input connections" active when duplicating shaders.


lib/mayaUsd/ufe/UsdUndoDuplicateFixupsCommand.h
```diff
@@ -0,0 +1,77 @@
...
+
+    void execute() override;
+    void undo() override;
+    void redo() override;
+
+private:
+    UsdUndoableItem _undoableItem;
+
+    typedef std::map<SdfPath, SdfPath>                        TDuplicatePathsMap;
+    typedef std::unordered_map<Ufe::Path, TDuplicatePathsMap> TDuplicatesMap;
```
seando-adsk, line 64: The coding standard says to prefer using over typedef. But in either case no "T".

seando-adsk (changes requested):

test/lib/ufe/testBatchOpsHandler.py
```diff
@@ -0,0 +1,263 @@
...
+        sel.append(matItem)
+
+        cmd = batchOpsHandler.duplicateSelection(sel, {"inputConnections": True})
+        cmd.execute()
+
+        dGeomPrim = usdUtils.getPrimFromSceneItem(cmd.targetItem(geomItem.path()))
+        dGeomBindAPI = UsdShade.MaterialBindingAPI(dGeomPrim)
+        # Now seeing and using ss3SG1
+        self.assertEqual(dGeomBindAPI.GetDirectBinding().GetMaterialPath(), Sdf.Path("/mtl/ss3SG1"))
+
```
ppt-adsk, line 85: Would be nice to test undo / redo of the UFE command.

test/lib/ufe/testBatchOpsHandler.py
```diff
@@ -0,0 +1,263 @@
...
+        ufe.GlobalSelection.get().clear()
+        ufe.GlobalSelection.get().append(geomItem)
+        ufe.GlobalSelection.get().append(matItem)
+        cmds.duplicate(inputConnections = True)
+
+        dGeomPrim = usdUtils.getPrimFromSceneItem(ufe.GlobalSelection.get().front())
+        dGeomBindAPI = UsdShade.MaterialBindingAPI(dGeomPrim)
+        # Now seeing and using ss3SG1
+        self.assertEqual(dGeomBindAPI.GetDirectBinding().GetMaterialPath(), Sdf.Path("/mtl/ss3SG1"))
+
```
ppt-adsk, line 199: Would be nice to test undo / redo.

test/lib/ufe/testBatchOpsHandler.py
```diff
@@ -0,0 +1,263 @@
...
+        connections = connectionHandler.sourceConnections(f3tDupItem)
+        self.assertIsNotNone(connections)
+        conn = set(["{}.{} -> {}.{}".format(i.src.path, i.src.name, i.dst.path, i.dst.name) for i in connections.allConnections()])
+        self.assertEqual(conn, expectedF3t)
+
+        # The connection between the new place2dTexture and the original NodeGraph will be gone.
+        connections = connectionHandler.sourceConnections(f3pDupItem)
+        self.assertIsNotNone(connections)
+        self.assertEqual(len(connections.allConnections()), 0)
+
```
ppt-adsk, line 260: Undo / redo.

lib/mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.h
```diff
@@ -0,0 +1,88 @@
...
+//
+#ifndef MAYAUSD_UFE_USDUNDODUPLICATESELECTIONCOMMAND_H
+#define MAYAUSD_UFE_USDUNDODUPLICATESELECTIONCOMMAND_H
+
+#include <mayaUsd/base/api.h>
+#include <mayaUsd/ufe/UsdUndoDuplicateCommand.h>
+#include <mayaUsd/undo/UsdUndoableItem.h>
+
+#include <pxr/usd/sdf/path.h>
+#include <pxr/usd/usd/prim.h>
```
ppt-adsk, line 24: Do we use this include?

lib/mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.cpp
```diff
@@ -0,0 +1,207 @@
...
+} // namespace
+
+UsdUndoDuplicateSelectionCommand::UsdUndoDuplicateSelectionCommand(
+    const Ufe::Selection&       selection,
+    const Ufe::ValueDictionary& duplicateOptions)
+    : Ufe::SelectionUndoableCommand()
+    , _copyExternalInputs(shouldConnectExternalInputs(duplicateOptions))
+{
+    // TODO: MAYA-125854. If duplicating /a/b and /a/b/c, it would make sense to order the
+    // operations by SdfPath, and always check if the previoulsy processed path is a prefix of the
```
ppt-adsk, line 57: Typo: "previoulsy"

ppt-adsk (commented):
Great work, just needs a few additional undo / redo tests IMO.

ppt-adsk (dismissed):

ppt-adsk (approved):

cmake/modules/FindUFE.cmake
```diff
@@ -82,6 +83,13 @@ find_library(UFE_LIBRARY
     NO_DEFAULT_PATH
 )
 
+# Can not be FALSE or 0 as this implies "not found" and fails.
+set(UFE_PREVIEW_BATCHOPS_SUPPORT "absent")
```
JGamache-autodesk, line 87: I now see why CACHED variables are used for bolean CMake variables. Here is an alternative that works,.

@seando-adsk : If you find that style of CMake unacceptable, I have no issue reverting to a CACHED boolean CMake variable.

cmake/modules/FindUFE.cmake
```diff
@@ -11,6 +11,7 @@
 # UFE_VERSION         UFE version (major.minor.patch) from ufe.h
 # UFE_LIGHTS_SUPPORT  Presence of UFE lights support
 # UFE_SCENE_SEGMENT_SUPPORT Presence of UFE scene segment support
+# UFE_PREVIEW_BATCHOPS_SUPPORT Presence of UFE batched operations support
```
JGamache-autodesk, line 14: Should be UFE_PREVIEW_FEATURES ...

cmake/modules/FindUFE.cmake
```diff
@@ -82,6 +83,13 @@ find_library(UFE_LIBRARY
     NO_DEFAULT_PATH
 )
 
+# Gather all preview features that might be there or not into a single list:
+list(APPEND UFE_PREVIEW_FEATURES v4)
+
+if (UFE_INCLUDE_DIR AND EXISTS "${UFE_INCLUDE_DIR}/ufe/batchOpsHandler.h")
+    list(APPEND UFE_PREVIEW_FEATURES BatchOps)
+endif()
```
JGamache-autodesk, line 91: Using a list to store keys looks like a viable solution to handle progressively added features.

ppt-adsk: I like it, but I'm not a CMake expert.

ppt-adsk (dismissed):

cmake/modules/FindUFE.cmake
```diff
@@ -82,6 +83,13 @@ find_library(UFE_LIBRARY
     NO_DEFAULT_PATH
 )
 
+# Gather all preview features that might be there or not into a single list:
+list(APPEND UFE_PREVIEW_FEATURES v4)
```
seando-adsk, line 87: We can't just say that it includes v4, because when compiling with older versions of Maya we aren't using Ufe v4.

cmake/modules/FindUFE.cmake
seando-adsk: I think each feature added to UFE_PREVIEW_FEATURES should be tagged with the version first. So "v4_BatchOps". This would make them easier to find and remove/modify to normal ufe ifdef once Ufe v4 is released.

cmake/modules/FindUFE.cmake
ppt-adsk: Isn't this rather saying that UFE_PREVIEW_FEATURES is a list that pertains to UFE v4?  Not sure what we're getting by placing this "v4" string into the list, because it will always be the case that preview features will pertain to the up-coming UFE version.

JGamache-autodesk: In fact, I would push one step further and say that maybe even "preview" is too specific. If we imagine this being only a list of Ufe features, then it enable them as soon as Ufe supports them in Maya, which becomes very interesteing from a feature backporting scenario. I do remember the MAYA_LIGHT_API_VERSION kicking in when I backported the light API fixes for 2020.3 without having to update MayaUSD.

Might be potential discussion subject for the UFE interest meeting.

JGamache-autodesk: Better initializer, since list can not be empty, would be "ufe".

ppt-adsk (dismissed):

cmake/modules/FindUFE.cmake
seando-adsk: Once Ufe v4 is released we remove all these "preview" checks and replace them with just "ufe v4" check.

lib/mayaUsd/ufe/UsdBatchOpsHandler.cpp
```diff
@@ -0,0 +1,50 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "UsdBatchOpsHandler.h"
+
+#include <mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.h>
+
+PXR_NAMESPACE_USING_DIRECTIVE
```
seando-adsk, line 20: Do you need this? All new code should use the fully qualified namespace name. But I don't see any USD code here.

lib/mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include <pxr/usd/usd/attribute.h>
+#include <pxr/usd/usd/prim.h>
+#include <pxr/usd/usd/primRange.h>
+#include <pxr/usd/usd/property.h>
+#include <pxr/usd/usd/relationship.h>
+#include <pxr/usd/usd/stage.h>
+
+#include <ufe/path.h>
+
+PXR_NAMESPACE_USING_DIRECTIVE
```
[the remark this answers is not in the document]
seando-adsk: If its for the TF_VERIFY below, then put the macro inside that method.

seando-adsk (commented):
Just a couple of minor comments.

ppt-adsk (approved):

seando-adsk (approved):
Excellent thanks.

PR #2691, opened 2022-10-31
MAYA-125895 transfer state from root layer when toggling shareable stage

pierrebai-adsk:
Currently, we reproduce the FPS (frames-per-second) of the shared root layer to the
newly-created unshared root layer so that the animation playback speed stays consistent.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -627,7 +627,7 @@ void reproduceSharedStageState(
         return;
     if (!TF_VERIFY(unsharedRootLayer))
         return;
-        
+
     // Transfer the FPS (frames-per-second) of the original root layer to the new unshared
     // root layer, so that the animation timeline does not change. We copy both the metadata
     // on the layer and on the stage object itself.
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -616,6 +616,24 @@ void remapSublayerRecursive(
...
+    const SdfLayerRefPtr& inRootLayer,
+    const SdfLayerRefPtr& unsharedStageRootLayer)
+{
+    // Transfer the FPS (frames-per-second) of the original root layer to the new unshared
+    // root layer, so that the animation timeline does not change. We copy both the metadata
+    // on the layer and on the stage object itself.
+    if (SdfDataRefPtr metadata = inRootLayer->GetMetadata()) {
+        VtValue fpsValue;
+        if (metadata->Has(SdfPath("/"), TfToken("framesPerSecond"), &fpsValue)) {
+            unsharedStageRootLayer->GetMetadata()->Set(
```
vlasovi, line 630: A few lines above you check that GetMetadata() is not null, and here you don't. Shouldn't we be more consistent?

pierrebai-adsk: Good catch! Although I will actually check the USD docs to see if it is possible for the metadata to be null. So I'll either add a check or remove the other one.

vlasovi (approved):

PR #2698, opened 2022-11-02
MAYA-125991 - Unable to switch variants when Session Layer is targeted

vlasovi:


pierrebai-adsk (dismissed):
I tried to think if your simple solution would not work in some case but could not imagine any. (famous last words, I know.) Good work.

pierrebai-adsk:
Oh, just one thing: could a unit test be created to test this?

pierrebai-adsk (approved):

vlasovi:
@seando-adsk The preflight failure is due to an internal build system issue and not to the PR. This can be merged.

PR #2700, opened 2022-11-02
MAYA-123716 - Duplicated prim to be on the Display Layer that the source prim was on

seando-adsk:
#### MAYA-123716 - Duplicated prim to be on the Display Layer that the source prim was on

* Use new MayaUsd::ufe::ReplicateExtrasToUSD helper class to perform dupication of Display Layer (USD -> USD).
* Added unit test.

### DIFF
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -932,14 +932,12 @@ void handleDisplayLayer(
 
 void ReplicateExtrasFromUSD::initRecursive(Ufe::SceneItem::Ptr ufeItem) const
 {
-    auto node = Ufe::Hierarchy::hierarchy(ufeItem);
-    if (!node) {
-        return;
-    }
-
-    // Go through the entire hierarchy
-    for (auto child : node->children()) {
-        initRecursive(child);
+    auto hier = Ufe::Hierarchy::hierarchy(ufeItem);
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -932,14 +932,12 @@ void handleDisplayLayer(
...
-    // Go through the entire hierarchy
-    for (auto child : node->children()) {
-        initRecursive(child);
+    auto hier = Ufe::Hierarchy::hierarchy(ufeItem);
+    if (hier) {
+        // Go through the entire hierarchy
+        for (auto child : hier->children()) {
+            initRecursive(child);
+        }
     }
```
seando-adsk, line 941: I think that it should process the input scene item even if there is no hierarchy (I did the same for my new method below).

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -994,6 +992,30 @@ void ReplicateExtrasToUSD::processItem(const MDagPath& dagPath, const SdfPath& u
 #endif
 }
 
+void ReplicateExtrasToUSD::initRecursive(const Ufe::SceneItem::Ptr& item) const
```
seando-adsk, line 995: New method similar to other one except for "To USD"

PR #2701, opened 2022-11-03
MAYA-125895 Handle timecodes-per-second when flipping shared state

pierrebai-adsk:
Tramsfer the timecodes-per-second when flipping shared flag on a stage.


vlasovi (approved):

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -636,6 +636,13 @@ void reproduceSharedStageState(
         unsharedRootLayer->SetFramesPerSecond(fps);
         stage->SetFramesPerSecond(fps);
     }
+
+    // Transfer the TCPS (timecodes-per-second) for the same reason as above.
+    if (sharedRootLayer->HasTimeCodesPerSecond()) {
```
tanant, line 641: Thanks! :D

PR #2704, opened 2022-11-04
MAYA-126109 | Triggers attribute metadata changed notification.

feldstj:
Note: A lot of this work is necessary as there is no native USD notification for attribute metadata changes.

### DIFF
--- a/lib/mayaUsd/ufe/Global.cpp
+++ b/lib/mayaUsd/ufe/Global.cpp
@@ -145,7 +145,7 @@ StagesSubject::Ptr g_StagesSubject;
 
 bool InPathChange::inGuard = false;
 bool InAddOrDeleteOperation::inGuard = false;
-bool InAttributeMetadataChange::inGuard = false;
+unsigned int InAttributeMetadataChange::inGuard = 0;
 
 //------------------------------------------------------------------------------
 // Functions
--- a/lib/mayaUsd/ufe/Global.cpp
+++ b/lib/mayaUsd/ufe/Global.cpp
@@ -143,8 +143,8 @@ Ufe::UIInfoHandler::Ptr g_MayaUIInfoHandler;
 // Subject singleton for observation of all USD stages.
 StagesSubject::Ptr g_StagesSubject;
 
-bool InPathChange::inGuard = false;
-bool InAddOrDeleteOperation::inGuard = false;
+bool         InPathChange::inGuard = false;
+bool         InAddOrDeleteOperation::inGuard = false;
 unsigned int InAttributeMetadataChange::inGuard = 0;
 
 //------------------------------------------------------------------------------
--- a/lib/mayaUsd/ufe/Global.cpp
+++ b/lib/mayaUsd/ufe/Global.cpp
@@ -143,9 +143,8 @@ Ufe::UIInfoHandler::Ptr g_MayaUIInfoHandler;
 // Subject singleton for observation of all USD stages.
 StagesSubject::Ptr g_StagesSubject;
 
-bool         InPathChange::inGuard = false;
-bool         InAddOrDeleteOperation::inGuard = false;
-unsigned int InAttributeMetadataChange::inGuard = 0;
+bool InPathChange::inGuard = false;
+bool InAddOrDeleteOperation::inGuard = false;
 
 //------------------------------------------------------------------------------
 // Functions
--- a/lib/mayaUsd/ufe/Global.cpp
+++ b/lib/mayaUsd/ufe/Global.cpp
@@ -145,6 +145,7 @@ StagesSubject::Ptr g_StagesSubject;
 
 bool InPathChange::inGuard = false;
 bool InAddOrDeleteOperation::inGuard = false;
+bool InAttributeMetadataChange::inGuard = false;
 
 //------------------------------------------------------------------------------
 // Functions
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -188,9 +188,9 @@ void sendAttributeChanged(
         notifyWithoutExceptions<Ufe::Attributes>(
             Ufe::AttributeConnectionChanged(ufePath, changedToken.GetString()));
     } break;
-    default: {
+    case AttributeChangeType::kMetadataChanged: {
         // Do nothing (needed to fix build error)
-    }
+    } break;
     }
 #else
     notifyWithoutExceptions<Ufe::Attributes>(
--- a/lib/mayaUsd/ufe/StagesSubject.cpp
+++ b/lib/mayaUsd/ufe/StagesSubject.cpp
@@ -874,13 +874,15 @@ AttributeChangedNotificationGuard::~AttributeChangedNotificationGuard()
     }
 
     for (const auto& notificationInfo : pendingAttributeChangedNotifications) {
-        if (const auto metadataNotificationInfo
-            = dynamic_cast<const AttributeMetadataNotification*>(&notificationInfo)) {
-            sendAttributeMetadataChanged(
-                metadataNotificationInfo->_path,
-                metadataNotificationInfo->_token,
-                metadataNotificationInfo->_type,
-                metadataNotificationInfo->_metadataKeys);
+        if (notificationInfo._type == AttributeChangeType::kMetadataChanged) {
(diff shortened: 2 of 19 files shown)


lib/mayaUsd/ufe/private/UfeNotifGuard.h
```diff
@@ -89,6 +89,29 @@ class InTransform3dChange
...
 };
 
+//! \brief Helper class to scope when we are in an attribute metadata change operation.
+//
+// This simple guard class can be used within a single scope, but does not have
+// recursive scope capability.
+class InAttributeMetadataChange
+{
+public:
+    InAttributeMetadataChange() { inGuard = true; }
```
JGamache-autodesk, line 99: Does not support two levels of guard active at the same time. Please use a counter instead.

test/lib/ufe/testAttribute.py
```diff
@@ -1929,5 +1929,30 @@ def testNamePrettification(self):
...
+
+        self.assertTrue(attrs.hasAttribute("inputs:in"))
+        attr = attrs.attribute("inputs:in")
+        value = ufe.Value("-1")
+        cmd = attr.setMetadataCmd("uisoftmin", value)
+        cmd.execute()
+
+        self.assertEqual(1, obs.notifications)
+
+        self.assertEqual(attr.getMetadata("uisoftmin").__str__(), value.__str__())
```
JGamache-autodesk, line 1955: Please use `str(X)` instead of `X.__str__()`. These are exactly the same and it prevents using double underscored members directly, which is considered bad Python style.

JGamache-autodesk (changes requested):

JGamache-autodesk (dismissed):

lib/mayaUsd/ufe/UsdUndoAttributesCommands.h
```diff
@@ -110,6 +110,35 @@ class UsdRemoveAttributeCommand : public UsdUndoableCommand<Ufe::UndoableCommand
...
+    UsdSetMetadataCommand& operator=(UsdSetMetadataCommand&&) = delete;
+
+    //! Create a UsdSetMetadataCommand
+    static UsdSetMetadataCommand::Ptr
+    create(MayaUsd::ufe::UsdAttribute& attr, const std::string& key, const Ufe::Value& newValue);
+
+    void executeUndoBlock() override;
+
+private:
+    MayaUsd::ufe::UsdAttribute& _attr;
```
ppt-adsk, line 136: It feels very awkward to keep a reference to the MayaUsd::ufe::UsdAttribute in the command, as this is usually unsafe, because of questions about the lifescope of the reference.  I realize that it isn't used after executeUndoBlock(), but can you add a comment about this here in the header?  I would even consider storing a plain C pointer to the MayaUsd::ufe::UsdAttribute, and nulling it out after executeUndoBlock() is done, for complete clarity.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -243,6 +253,10 @@ void processAttributeChanges(
     if (sendRemoved) {
         attributeChanged(ufePath, changedPath.GetNameToken(), AttributeChangeType::kRemoved);
     }
+    if (MayaUsd::ufe::InAttributeMetadataChange::inAttributeMetadataChange()) {
```
ppt-adsk, line 256: I am quite concerned about this approach, as it is very weak.  It depends on going through the maya-usd code path you've defined, and will not work when using e.g. the Pixar USD API alone.  We should move away from using this style of coding, and instead rely on USD notifications themselves, which will be bullet-proof.  USD does provide significant detail about what fields have changed in its notification system; e.g. see
https://github.com/Autodesk/maya-usd/blob/42854b32c9c1cd706917e088b477119ca1d791ab/lib/mayaUsd/ufe/StagesSubject.cpp#L517
I'm very, very hopeful we can use that.

ppt-adsk (commented):
As per comments, I think using a notification system that depends on going through our maya-usd code path is too weak.  We should really be observing the USD data model to provide robust notifications on attribute metadata changes.

lib/mayaUsd/ufe/StagesSubject.cpp
JGamache-autodesk: Indeed, this forces me to recheck the USD API to see if there is a better way:
On a Metadata change, the function SdfChangeList::DidChangeInfo() will be called, generating a very specifically shaped USD notification that should be intercepted instead of introducing a Metadata guard.
This means that looking for contents in the `.infoChanged` field of the entry should be enough to detect that metadata was modified.
Also means that the UFE API for AttributeMetadataChanged requires a list of metadata keys to copy all those found there.

JGamache-autodesk (changes requested):

lib/mayaUsd/ufe/StagesSubject.cpp
feldstj: I hadn't wanted to implement things this way, but I was under the impression that there was no way to find out that USD had made a metadata change. But now Jerry bellow is suggesting it is possible so I'll look into it.

ppt-adsk (dismissed):

lib/mayaUsd/ufe/Global.cpp
```diff
@@ -143,8 +143,9 @@ Ufe::UIInfoHandler::Ptr g_MayaUIInfoHandler;
 // Subject singleton for observation of all USD stages.
 StagesSubject::Ptr g_StagesSubject;
 
-bool InPathChange::inGuard = false;
-bool InAddOrDeleteOperation::inGuard = false;
+bool         InPathChange::inGuard = false;
+bool         InAddOrDeleteOperation::inGuard = false;
+unsigned int InAttributeMetadataChange::inGuard = 0;
```
JGamache-autodesk, line 148: This is now unused code. Please remove.

lib/mayaUsd/ufe/private/UfeNotifGuard.h
```diff
@@ -89,6 +89,28 @@ class InTransform3dChange
...
+    InAttributeMetadataChange& operator=(const InAttributeMetadataChange&) = delete;
+    InAttributeMetadataChange(InAttributeMetadataChange&&) = delete;
+    InAttributeMetadataChange& operator=(InAttributeMetadataChange&&) = delete;
+
+    static bool inAttributeMetadataChange() { return inGuard > 0; }
+
+private:
+    static unsigned int inGuard;
+};
+
```
JGamache-autodesk, line 113: Unused class.

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -768,8 +874,17 @@ AttributeChangedNotificationGuard::~AttributeChangedNotificationGuard()
     }
 
     for (const auto& notificationInfo : pendingAttributeChangedNotifications) {
-        sendAttributeChanged(
-            notificationInfo._path, notificationInfo._token, notificationInfo._type);
+        if (const auto metadataNotificationInfo
```
JGamache-autodesk, line 877: Check `_type` and only do dynamic_cast if it is a metadata notification.

JGamache-autodesk (changes requested):

feldstj:
Looks like I need one of Alice's MayaUSD changes to go in in order to fix my local MayaUSD build. I suppose I could merge it locally, but it showcases that her change needs to go in first.

JGamache-autodesk (dismissed):

ppt-adsk (dismissed):

ppt-adsk (dismissed):

ppt-adsk (approved):

lib/mayaUsd/ufe/StagesSubject.cpp
```diff
@@ -188,6 +188,9 @@ void sendAttributeChanged(
         notifyWithoutExceptions<Ufe::Attributes>(
             Ufe::AttributeConnectionChanged(ufePath, changedToken.GetString()));
     } break;
+    default: {
```
ppt-adsk, line 191: The only enumerant not covered is kMetadataChanged.  I think it would be better for maintainability to remove the catch-all default and explicitly add kMetadataChanged.

ppt-adsk (changes requested):

ppt-adsk (approved):

seando-adsk (approved):

JGamache-autodesk (approved):

PR #2708, opened 2022-11-08
MAYA125953 worldspace flag for Maya USD export

pierrebai-adsk:
- Added -worldspace flag to the mayaUSDExport command.
- Added worldspace flag to the UsdMayaJobExportArgs structure.
- Added worldspace to the UsdMayaJobExportArgsTokens.
- Added docs about worldspace flag to the commands read-me.
- Added default value for the worldspace flag in UsdMayaJobExportArgs.
- Added worldspace parsing to UsdMayaJobExportArgs.
- Added worldspace flag to the Python wrapper.
- Added worldspace flag default to the Python default dict.
- Fixed the extractVector template function to not be exported since it is inlined.
- Fixed the UsdMayaExportTranslator to use the root prim options the same way the export command does.
- Added worldspace flag to the export UI.
- Don't assume that all new channels have unique names when making their names unique.
- Walk up the DAG stack to generate the channels for all ancestors when using worldspace flag.
- Separate channel generation from writing the xform op, to support recursion to generate ancestor.
- Added tests for the worldspace flag in the unit tests.


pierrebai-adsk:
Jerry, if you are too busy, I could ask someone else from our team, but you are the export expert, so I chose you by default.

JGamache-autodesk (approved):

PR #2713, opened 2022-11-09
Fix default export logic if subdivScheme is none

dj-mcg:
Skip authoring normals by default when exporting a mesh with its subdivSceme set to NONE. Require the emitNormals tag to be set to TRUE for normals to be written for these meshes.


pierrebai-adsk (approved):

seando-adsk:
@dj-mcg The preflight failed. Can you have a look.

pierrebai-adsk:
The mesh render test failed due to "px_render" plugin not being found. I had the same error yesterday. Re-running the PF fixed it. I wonder if some PF computer are setup wrong?

dj-mcg:
Ah - I think this dependency snuck into the test maya file. I'll fix it up and kick off another preflight

dj-mcg:
Oops, I guess I'll need to use Maya 2020 (or Maya 2019?) to update the test file... Or maybe I can just hand edit the .ma?

dgovil:
Hi I'm curious why this change was necessary? The PR doesn't provide reasoning  and I think the logic is incorrect.
If subdivision schema is set to None, is when one should expect normals to be exported?
Per https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usd/usdGeom/schema.usda#L1018 the `none` type is when one should be exporting normals.

This is breaking our exports as a result because we suddenly don't have normals as of this PR.

dj-mcg:
You can still emit normals for the `none` type if you explicitly tag the mesh to do so. If the tag isn't explicitly authored, we shouldn't be writing normals as documented here:
http://grok/xref/mainline/ext/mayapkg/maya_open_source/maya-usd/lib/mayaUsd/commands/Readme.md?r=2255052#282

dgovil:
I think this is a breaking change though and a silent one at that. I acknowledge that y'all documented this years ago but the behaviour did not match the docs however the docs don't match expectation.

You'd suddenly be causing people's scenes to be exporting without normals as ours suddenly did when we pulled from top of tree. It would require everyone to go back and modify all their scenes to conform to a doc spec that wasn't implemented properly. 

I also don't understand who this benefits? In what scenario would you want to be opting IN to normals vs opting OUT? I can understand if the case was that the tag opted people out of them, but the default should be to export normals without users having to do anything.

So I posit that the fix should be:

* Default to normals on.
* If the `USD_EmitNormals` is set, then change `emitNormals` to that value. 
* If it's not set, then `emitNormals` should be `true`

This is IMHO the only way to both conform to that document and not break existing files for export.

dgovil:
I made a PR here to resolve this https://github.com/Autodesk/maya-usd/pull/2768
I'd really appreciate it if the PR was accepted.

dgovil:
Ah that makes sense why you'd want it off then. 
No worries, I just think the most intuitive default is to export normals when no subdiv is present. I think maybe if it needed to be modulated on a global level, an env var or compile time flag would have been better so Pixar could set it at a studio level?

dj-mcg:
Yeah for now we'll just patch in our current behavior and I'll figure out what to do to handle this long term. But you're right - it was presumptive of me to drop internally requested behavior changes on the wider community here, so I think preserving the behavior that was inadvertently established is probably the right way forward.

PR #2715, opened 2022-11-10
Remove reference to vaccine python code

### LINKED ISSUE #2711: Test File Vaccine
image

**Describe the bug**
Antivirus software says it's a virus

**Specs (if applicable):**
 - OS & version Windows 10

tanant:
Originally introduced in 9ee7aef with the modification of a test file - resolves #2711


tanant:
@dj-mcg - looks like it was a PR from a bit back, just adding a ping here incase you hadn't localised the issue on your machine

tanant:
@seando-adsk  / @neilh-adsk / @santosg87  I believe this is a correct fix for this, but I've done it as an eyeball-only fix for speed so you know, with speed comes risk and so would appreciate sanity check here.

seando-adsk (approved):

dj-mcg:
Ugh thanks - this has been blowing up at the studio. Any suggestions on how to combat it would be great! Sorry that it slipped into the open source code

tanant:
@dj-mcg all good, stuff happens - we do have the security tools - https://apps.autodesk.com/MAYA/en/Detail/Index?id=8637238041954239715 which has a cleanup inside it (but you probably already have a good idea how you'd do it with custom script).

If you can track it back to a point in time, then files post that date could be converted/checked (e.g. you could set a job to open scripts + patch/clean on one machine maybe?)

Other more global solutions could be to set (I believe it's a userpref) your to execute script nodes on file open - but you may actually want that behaviour - or just installing a callback to pick it up on script open time.

PR #2718, opened 2022-11-11
MAYA-125275 preserve the layer manager expanded layers

pierrebai-adsk:
When the layer manager model is rebuilt, preserve the state of the layer view (currently, only the expanded state of each layer) so that it can be restored once the model rebuilt is done.

- Add a LayerViewMemento to preserve the view state.
- Provide a preserve() and restore() method to save and restorethe state.
- Provide an empty method to avoid losing state when multiple calls to About-to-reset are made.
- Preserve befor ethe model reset, restore after.
- Note that, yes Qt emits the signal multiple times when the model is reset, so we have to take that into accouunt.

### DIFF
--- a/lib/usd/ui/layerEditor/layerTreeView.cpp
+++ b/lib/usd/ui/layerEditor/layerTreeView.cpp
@@ -166,6 +166,8 @@ void LayerViewMemento::restore(LayerTreeView& view, LayerTreeModel& model)
 {
     const LayerItemVector items = model.getAllItems();
 
+    QtDisableRepaintUpdates disableUpdates(view);
+
     for (const LayerTreeItem* item : items) {
         if (!item)
             continue;
@@ -177,7 +179,7 @@ void LayerViewMemento::restore(LayerTreeView& view, LayerTreeModel& model)
             const ItemId id = layer->GetIdentifier();
             const auto   state = itemsState.find(id);
             if (state != itemsState.end()) {
-                expanded = state->second.expanded;
+                expanded = state->second._expanded;
             }
         }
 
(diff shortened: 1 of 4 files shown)


lib/usd/ui/layerEditor/layerTreeView.cpp
```diff
@@ -131,7 +136,77 @@ void LayerTreeView::onItemDoubleClicked(const QModelIndex& index)
...
+        PXR_NS::SdfLayerRefPtr layer = item->layer();
+        if (layer) {
+            const ItemId id = layer->GetIdentifier();
+            const auto   state = itemsState.find(id);
+            if (state != itemsState.end()) {
+                expanded = state->second.expanded;
+            }
+        }
+
+        view.setExpanded(item->index(), expanded);
```
seando-adsk, line 184: When you call things like this you trigger a redraw. So normally when doing this in a loop, we'll first disable drawing updates before the loop, and then enable them (and redraw) after the loop.

seando-adsk (changes requested):
One minor variable renaming and a question about drawing.

lib/usd/ui/layerEditor/layerTreeView.cpp
pierrebai-adsk: Do we have a utility class that disable Qt redraws and trigger and repaint already existing?

lib/usd/ui/layerEditor/qtUtils.h
```diff
@@ -53,6 +53,20 @@ class QtUtils
     }
 };
 
+/**
+ * @brief Disable repaint updates for the given widget until the disabler is destoryed.
```
seando-adsk, line 57: Minor typo: destoryed -> destroyed

lib/usd/ui/layerEditor/layerTreeView.h
```diff
@@ -31,11 +31,37 @@ namespace UsdLayerEditor {
...
+    bool empty() const { return itemsState.empty(); }
+
+private:
+    using ItemId = std::string;
+    struct ItemState
+    {
+        bool _expanded = false;
+    };
+
+    std::map<ItemId, ItemState> itemsState;
```
seando-adsk, line 62: Sorry this one also needs an _

seando-adsk (changes requested):
Still one member variable to rename and one typo.

seando-adsk (approved):

Issue #2549, opened 2022-08-16
Setting visibility does not consistently route edits


gracekumagai:
**Describe the bug**
When working on https://github.com/Autodesk/maya-usd/pull/2546, @jallex and I noticed that after we had registered an edit router to route visibility edits to the session layer, setting visibility was not consistently set in the session layer.

**Expected behavior**
We would expect to have visibility routed to the session layer (our routing layer) after we've registered the edit router.

From chatting with our artists, their main workflow for setting visibility is to use Ctrl+h and Shift+h hotkeys. On our side, the Display menu actions and corresponding hotkeys would be the top priority to fix.

Fwiw - they also heavily rely on using Display Layers to toggle visibility. 

**Steps to reproduce**
We tested with this PR https://github.com/Autodesk/maya-usd/pull/2546. Then, we created a stage and using the context menu in the outliner, added prims to test with.

Ctrl + h hotkey | Display -> Hide -> Hide Selection menu item | `cmds.hide()`
> If we select a new prim + use action above, it successfully routes visibility edits to the session layer!
> 
> If we multi-select an invisible prim (where visibility is authored on the session layer) + a visible prim (where the prim does not have visibility explicitly authored yet), it will hide both and route invisibility to the session layer. However, it also prints the following error:
> 
> ` // Error: Cannot edit [visibility] attribute because there is a stronger opinion in [anonymousLayer1-session.usda].
> `

Shift + H hotkey | Ctrl + Shift + H hotkey | Display -> Show -> Show Selection or Show Last Hidden menu items | `cmds.showHidden()`
> If we select an invis prim (where visibility is set on the session layer) + use action above, it does not make the prim invisible and prints the same "stronger opinion" error as above. We would expect visibility to be updated in the session layer and for the prims to be visible.
>
> If we select an invis prim (where visibility is set on the target layer (incorrectly)), it sets visibility=inherited on the session layer and visibility=invisible remains on the target layer.

H hotkey | set visibility using pxr.Usd.Attribute
> To test setting the visibility using an pxr.Usd.Attribute, we used the following code:
```
prim = mayaUsd.ufe.ufePathToPrim(ufe.PathString.string(sceneItem.path()))
visibilityAttr = prim.GetAttribute('visibility')
visibilityAttr.Set(UsdGeom.Tokens.visible) # or UsdGeom.Tokens.invisible
```
[...]

PR #2725, opened 2022-11-15
Test if visibility edit is valid before returning command.

ppt-adsk:


lib/mayaUsd/ufe/UsdUndoVisibleCommand.cpp
```diff
@@ -39,19 +42,26 @@ UsdUndoVisibleCommand::Ptr UsdUndoVisibleCommand::create(const UsdPrim& prim, bo
...
-{
-    UsdGeomImageable primImageable(_prim);
+    auto layer = getEditRouterLayer(PXR_NS::TfToken("visibility"), prim);
+
+    UsdGeomImageable primImageable(prim);
+
+    EditTargetGuard guard(prim, layer);
 
     std::string errMsg;
     if (!MayaUsd::ufe::isAttributeEditAllowed(primImageable.GetVisibilityAttr(), &errMsg)) {
```
ppt-adsk, line 53: Perform the edit check on command creation, rather than execution.  Failure on execution should be reported through an exception, but here the check really belongs to command creation: we communicate to the caller that the operation is not possible by returning a nullptr.

Furthermore, the edit guard must be set before performing the check, otherwise the check will be done on the current target, then the edit guard changes that target.  In the case of an existing visibility edit in the session layer, this would mean that the check would be done on the current edit target layer, which typically is lower priority than the session layer, and the edit would fail.  If the check is done when the edit target guard is set, the check properly occurs on the session layer.  This accounts for many of the failures described in

https://github.com/Autodesk/maya-usd/issues/2549

seando-adsk (approved):

PR #2726, opened 2022-11-15
Adapt testConnections for USD 22.11

JGamache-autodesk:


seando-adsk (approved):
Excellent thanks for the quick turnaround.

PR #2728, opened 2022-11-16
MAYA-126117 - Point instancer: performance drop on point instanced selection.

vlasovi:
1. Introducing display layer cache in vp2RenderDelegate to drastically optimize Syncing of massive amounts of render items/instances
2. Explicitly disable display layers for point instancers until we figure out exactly what to do with them

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/mayaPrimCommon.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mayaPrimCommon.cpp
@@ -47,6 +47,7 @@ constexpr auto sDrawModeAllButBBox = (MHWRender::MGeometry::DrawMode)(
 #ifdef MAYA_NEW_POINT_SNAPPING_SUPPORT
 
 namespace {
+std::mutex        sMayaMutex;
 MayaUsdCustomData sMayaUsdCustomData;
 } // namespace
 
@@ -704,6 +705,9 @@ void MayaUsdRPrim::_ProcessDisplayLayerModes(
     const MObject&     displayLayerObj,
     DisplayLayerModes& displayLayerModes)
 {
+    // Maya's MPlug API is not multithreadable, so we need the mutex here
+    std::lock_guard<std::mutex> mutexGuard(sMayaMutex);
+
     MFnDependencyNode displayLayerNodeFn(displayLayerObj);
     MPlug             layerEnabled = displayLayerNodeFn.findPlug("enabled");
     if (!layerEnabled.asBool()) {
@@ -754,7 +758,7 @@ void MayaUsdRPrim::_PopulateDisplayLayerModes(
     auto ancestorsRange = usdPath.GetAncestorsRange();
     for (auto it = ancestorsRange.begin(); it != ancestorsRange.end(); ++it) {
         auto displayLayerObj = drawScene.GetDisplayLayer(*it);
-        if (displayLayerObj.hasFn(MFn::kDisplayLayer)) {
+        if (!displayLayerObj.isNull()) {
             _ProcessDisplayLayerModes(displayLayerObj, displayLayerModes);
         }
     }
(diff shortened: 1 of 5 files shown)


test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateDisplayLayers.py
```diff
@@ -147,6 +147,7 @@ def testDisplayLayers(self):
         cmds.setAttr('layer1.drawInfo.hideOnPlayback', False)
 
         # templated
+        cmds.displayRGBColor("templateActive", 1.0, 0.69, 0.69)
```
pierrebai-adsk, line 150: Unclear what this additional line tests.

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -174,7 +174,7 @@ class ProxyRenderDelegate
 #endif
 
     MAYAUSD_CORE_PUBLIC
-    MString GetUfePathPrefix() const;
+    bool SupportPerInstanceDisplayLayers(const SdfPath& rprimId) const;
```
pierrebai-adsk, line 177: I'd rather have it named as a question, like "HasPerInstanceDisplayLayers"

lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h
```diff
@@ -191,6 +191,21 @@ class ProxyRenderDelegate
...
+    void AddDisplayLayerToCache(MObject& displayLayerObj);
+
+    MAYAUSD_CORE_PUBLIC
+    void UpdateProxyShapeDisplayLayers();
+
+    MAYAUSD_CORE_PUBLIC
+    MObject GetDisplayLayer(const SdfPath& path);
+
+    MAYAUSD_CORE_PUBLIC
+    const MObjectArray& GetProxyShapeDisplayLayers() const { return _usdStageDisplayLayers; }
```
pierrebai-adsk, line 208: This class in general really lacks comments about its design and functions... but maybe for another day.

lib/mayaUsd/render/vp2RenderDelegate/mayaPrimCommon.cpp
```diff
@@ -702,72 +701,89 @@ void MayaUsdRPrim::_UpdatePrimvarSourcesGeneric(
 }
 
 #ifdef MAYA_HAS_DISPLAY_LAYER_API
-void MayaUsdRPrim::_PopulateDisplayLayerModes(
-    const MString&     pathString,
+void MayaUsdRPrim::_ProcessDisplayLayerModes(
```
pierrebai-adsk, line 704: The function actually fills the DisplayLayerModes, so a better name might be _FillDisplayLayerModes or _InitializeDisplayLayerModes. Again, not a deal breaker, really nit picking.

test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateDisplayLayers.py
vlasovi: I added this line to explicitly specify the color of the templated object in this test. For some reason, when I did some other changes in this file, this test started failing because the templated color was different, equal to the vaule set in a different place of this unit test, specifically from line 265: cmds.displayRGBColor("templateActive", 0.9, 0.3, 0.7)

lib/mayaUsd/render/vp2RenderDelegate/mayaPrimCommon.cpp
pierrebai-adsk: Maybe it could even be a member function on DisplayLayerModes, along with the clearing you do below: fill() and clear()?

pierrebai-adsk (approved):
My only comments are minor and not essential to fix. Address at your own discretion.

test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateDisplayLayers.py
pierrebai-adsk: Oh, that mean tests are not entirely independent of each other? That could cause flakiness. OTOH, not the focus of this change...

vlasovi: Yes, and this is not the first time I hit this behaviour that something that should have a default value may not necessarily have it in the end

PR #2729, opened 2022-11-17
Fix lost connection for output port with parent

alicedegirolamo:
Simple fix.
When renaming output ports on the compound, connections with the parent are lost. 
This was caused by the fact that the update of connections with the new path was done on the children of the parent and not also on the parent itself.


JGamache-autodesk (dismissed):
Very good. Can you add a unit test that checks this bug stays fixed?

alicedegirolamo:
> Very good. Can you add a unit test that checks this bug stays fixed?

Sure! I'm going to do that.

alicedegirolamo:
> Very good. Can you add a unit test that checks this bug stays fixed?

Ok, I pushed and tested the new test. I had to add a new simple compound to the .usd since I didn't find in the sample folder a compound suitable for my test case.

JGamache-autodesk:
Preflight was successful (after a review of the logs). The two failure to upload artifacts happened after the tests succeeded.

JGamache-autodesk (approved):

lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -504,28 +504,37 @@ bool UsdAttributes::canRenameAttribute(
...
             }
         }
+        // Update the connections with the new property path.
+        if (hasChanged) {
+            attr.SetConnections(sources);
+        }
+    }
+}
+
+static void setConnectionsWithChildren(
```
hodoulp, line 530: I find the name ambiguous. It could be:
- `setConnectionsOfAllChildren()`
- `updateChildrenConnections()` +1

lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -504,28 +504,37 @@ bool UsdAttributes::canRenameAttribute(
...
-            for (size_t i = 0; i < sources.size(); ++i) {
-                if (sources[i] == OldPropertyPath) {
-                    sources[i] = newPropertyPath;
-                    hasChanged = true;
-                }
-            }
-            // Update the connections with the new property path.
-            if (hasChanged) {
-                attr.SetConnections(sources);
+    for (auto& attribute : prim.GetAttributes()) {
```
hodoulp, line 510: It could/should be a const variable like `const auto& attribute`.

lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -504,28 +504,37 @@ bool UsdAttributes::canRenameAttribute(
 
 static void setConnections(
     const PXR_NS::UsdPrim& prim,
-    const SdfPath&         OldPropertyPath,
+    const SdfPath&         oldPropertyPath,
```
hodoulp, line 507: The method does not set the connection it only updates existing connections:
`updateConnections()`

lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -504,28 +504,37 @@ bool UsdAttributes::canRenameAttribute(
...
+            if (sources[i] == oldPropertyPath) {
+                sources[i] = newPropertyPath;
+                hasChanged = true;
             }
         }
+        // Update the connections with the new property path.
+        if (hasChanged) {
+            attr.SetConnections(sources);
+        }
+    }
```
hodoulp, line 527: [Question] Should that method be recursive to process children of children, etc. ?

test/lib/ufe/testAttributes.py
```diff
@@ -300,6 +300,48 @@ def testRenamingAttribute(self):
...
+        ufeCompoundItem = ufeUtils.createUfeSceneItem(shapeNode,
+            '/pCube2/Looks/standardSurface2SG/NodeGraph1')
+        self.assertIsNotNone(ufeCompoundItem)
+
+        # Then create the attributes interface for that item.
+        compoundAttrs = ufe.Attributes.attributes(ufeCompoundItem)
+        self.assertIsNotNone(compoundAttrs)
+
+        # Rename the attribute.
+        cmd = compoundAttrs.renameAttributeCmd("outputs:out","outputs:out1")
```
hodoulp, line 314: typo `"outputs:out","outputs:out1"` -> `"outputs:out", "outputs:out1"`

test/lib/ufe/testAttributes.py
```diff
@@ -300,6 +300,48 @@ def testRenamingAttribute(self):
...
+        srcAttr = conns[1].src
+        dstAttr = conns[1].dst
+
+        self.assertEqual(ufe.PathString.string(srcAttr.path),
+            '|stage|stageShape,/pCube2/Looks/standardSurface2SG/NodeGraph1')
+        self.assertEqual(srcAttr.name, 'outputs:out1')
+
+        self.assertEqual(ufe.PathString.string(dstAttr.path),
+            '|stage|stageShape,/pCube2/Looks/standardSurface2SG')
+        self.assertEqual(dstAttr.name, 'outputs:out')
```
hodoulp, line 344: Could enhance the test with `cmds.undo()` ?

hodoulp (changes requested):

JGamache-autodesk (approved):

hodoulp:
I think that there are two main pending points:
- The recursivity question [here](https://github.com/Autodesk/maya-usd/pull/2729/files/218a234bccfb647f2ea3b5bf7e1b05ff75846762#diff-db3eb61f02eef0d4aa689c03f427b680616a013e850e3ea7870664d3e009a191R527)
- the undo() / redo() improvement in the unit test

But I will @JGamache-autodesk and Orn have the last call on that.

hodoulp (approved):

JGamache-autodesk:
- Going recursive is unnecessary on well constructed graphs. Connections on outer ports can only come from nodes in the englobing compound (or inner ports of the englobing compound). Connections on inner ports can only come from nodes directly inside the compound.
- Agreed that testing undo redo would be nice.

alicedegirolamo:
> I think that there are two main pending points:
> 
>     * The recursivity question [here](https://github.com/Autodesk/maya-usd/pull/2729/files/218a234bccfb647f2ea3b5bf7e1b05ff75846762#diff-db3eb61f02eef0d4aa689c03f427b680616a013e850e3ea7870664d3e009a191R527)
> 
>     * the undo() / redo() improvement in the unit test
> 
> 
> But I will @JGamache-autodesk and Orn have the last call on that.

Yes, I have to add the undo/redo test but I'm having problems compiling the updated Maya. As soon as I fixed my build I'm going to add it. Sorry for the wait.

JGamache-autodesk (approved):

hodoulp (approved):

neilh-adsk (approved):

PR #2730, opened 2022-11-17
MAYA-124097 anim channel code clarification

pierrebai-adsk:
The code and structure were using "opName" to mean different things. In some cases it was the name of the Maya transform attribute, in other cases it was the suffix of the USD transform attribute. They sometimes corresponded, but sometimes not, which made the code intent harder to understand.

This change is made in preparation to supporting parent-offset-matrix, to avoid polluting that PR with these pure nomenclature changes.

- Renamed the cases where "opName" is a a Maya attribute to "mayaAttrName".
- Renamed "parentName" to "mayaAttrName".
- Renamed the cases where "opName" is a USD suffix to "suffix".
- Renamed the "setOpName" flag to "useSuffix".


seando-adsk (approved):
Looks good.

PR #2733, opened 2022-11-18
Allow connecting filename attributes.

JGamache-autodesk:
The connection API requires being able to cast to UfeAttribute, the UsdAttributeFilename class was the only one not allowing that. This makes the code uniform.

### DIFF
--- a/lib/mayaUsd/ufe/UsdAttribute.h
+++ b/lib/mayaUsd/ufe/UsdAttribute.h
@@ -188,7 +188,11 @@ public:
 //! \brief Interface for USD token attributes.
 class UsdAttributeFilename
     : public Ufe::AttributeFilename
+#ifdef UFE_V4_FEATURES_AVAILABLE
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/UsdAttribute.h
```diff
@@ -188,7 +188,11 @@ class UsdAttributeGeneric
 //! \brief Interface for USD token attributes.
 class UsdAttributeFilename
     : public Ufe::AttributeFilename
+#ifdef UFE_V4_FEATURES_AVAILABLE
```
JGamache-autodesk, line 191: Copied verbatim from UsdAttributeGeneric at line 165.

hodoulp: Right that was the only one not transitioned!!!

hodoulp (approved):

PR #2734, opened 2022-11-18
MAYA-126368 make import *not* use the shared stage cache

pierrebai-adsk:
Importing a USD file should not be able to affect staged USD files. Avoid such problem by not using any cache when importing.


vlasovi (approved):
Looks good

PR #2736, opened 2022-11-22
MAYA-126129 more precise edit restrictions for metadata and variants

pierrebai-adsk:
- Moved all edit restrictions functions into the same header and C++ file to help find them.
- Added new functions to verify editability of metadata, including variant selections.
- Use the new metadat edit restriction functions for the variant switching command.
- Add unit test


lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -408,11 +408,11 @@ class SetVariantSelectionUndoableCommand : public Ufe::UndoableCommand
...
-        MayaUsd::ufe::applyCommandRestriction(
-            prim,
-            "set variant set " + _varSet.GetName() + " to variant " + _newSelection + " on",
-            allowStronger);
+        std::string errMsg;
+        if (!MayaUsd::ufe::isPrimMetadataEditAllowed(
+                prim, SdfFieldKeys->VariantSelection, TfToken(_varSet.GetName()), &errMsg)) {
+            throw std::runtime_error(errMsg.c_str());
+        }
     }
```
pierrebai-adsk, line 416: In the command we call the new check for metadata editing instead of general prim-level command restriction.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -444,6 +447,267 @@ TfTokenVector getProxyShapePurposes(const Ufe::Path& path)
     return purposes;
 }
 
+namespace {
```
pierrebai-adsk, line 450: A large part of the code below was moved from ufe/private/Utils.cpp to centralize all restriction code in the same place.

lib/mayaUsd/ufe/Utils.h
```diff
@@ -150,14 +150,6 @@ PXR_NS::UsdTimeCode getTime(const Ufe::Path& path);
 MAYAUSD_CORE_PUBLIC
 PXR_NS::TfTokenVector getProxyShapePurposes(const Ufe::Path& path);
 
-//! Check if an attribute value is allowed to be changed.
-//! \return True, if the attribute value is allowed to be edited in the stage's local Layer Stack.
-MAYAUSD_CORE_PUBLIC
-bool isAttributeEditAllowed(const PXR_NS::UsdAttribute& attr, std::string* errMsg = nullptr);
-
```
pierrebai-adsk, line 157: Moved these declaration lower to regroup all restriction-related functions together in the header.

test/lib/ufe/testContextOps.py
```diff
@@ -157,6 +158,14 @@ def testSwitchVariantInLayer(self):
...
         self.assertIn('Variant Sets', contextItemStrings)
 
+        # Add an attribute in the session layer to see if it affects
+        # switching variant.
+        stage = self.ball35Prim.GetStage()
+        stage.SetEditTarget(stage.GetSessionLayer())
+        self.ball35Prim.GetAttribute("xformOpOrder").Set(Vt.TokenArray("translate"))
+
+        stage.SetEditTarget(stage.GetRootLayer())
+
```
pierrebai-adsk, line 168: This change to the existing test for switching in different layers reproduced the reported problem by authoring an unrelated attribute in the session layer. It prevented authoring when we were applying too strict restrictions. Now we are only checking that the desired metadata has no no stronger opinion.

vlasovi (approved):
Looks good

PR #2742, opened 2022-11-24
MAYA-122481 fix rendering crash with MtoH

pierrebai-adsk:
When rendering, with some options, the renderer adds atemporary light to the scene, then removes it as soon as the render ends. The Maya scene delegate receives scene addition notifications, but in this case by the time the consequences if the notifications are processed, the light has already been deleted and the delegate access freed memory.

Make the scene delegate also listen to scene item removal notifications to remove items that have been deleted before they can be processed.

### DIFF
--- a/lib/usd/hdMaya/delegates/sceneDelegate.cpp
+++ b/lib/usd/hdMaya/delegates/sceneDelegate.cpp
@@ -55,6 +55,12 @@ void _nodeAdded(MObject& obj, void* clientData)
     delegate->NodeAdded(obj);
 }
 
+void _nodeRemoved(MObject& obj, void* clientData)
+{
+    auto* delegate = reinterpret_cast<HdMayaSceneDelegate*>(clientData);
+    delegate->NodeRemoved(obj);
+}
+
 const MString defaultLightSet("defaultLightSet");
 
 void _connectionChanged(MPlug& srcPlug, MPlug& destPlug, bool made, void* clientData)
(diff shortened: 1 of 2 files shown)


lib/usd/hdMaya/delegates/sceneDelegate.cpp
```diff
@@ -595,6 +605,16 @@ void HdMayaSceneDelegate::InsertDag(const MDagPath& dag)
 
 void HdMayaSceneDelegate::NodeAdded(const MObject& obj) { _addedNodes.push_back(obj); }
 
+void HdMayaSceneDelegate::NodeRemoved(const MObject& obj)
+{
+    // Note: remove in backward order to avoid processing items twice.
+    for (size_t i = _addedNodes.size() - 1; i < _addedNodes.size(); ++i) {
```
vlasovi, line 611: This for loop looks like a bug

vlasovi (changes requested):
I think that the for loop inside ::NodeRemoved is broken

lib/usd/hdMaya/delegates/sceneDelegate.cpp
vlasovi: Also, in terms of algorythmic complexity, it,s better to do something like: std::removeif(...); _addedNodes.erase(last, end);

pierrebai-adsk: The loop looks broken, but actually works because size_t is unsigned. I can use remove_if + erase though, it is slightly more efficient when there are a lot of removal. I don't expect the add + remove notifications without any update in between to happen frequently, only in rare corner cases like when the renderer adds a temp light, but the proposed change is simple.

vlasovi: Hmm, I would believe that the loop worked if there was --i and not ++i. With ++i, you always only check the last element

vlasovi (approved):

lib/usd/hdMaya/delegates/sceneDelegate.cpp
pierrebai-adsk: Oh you are right! I misinterpreted you comment, I thought you were commenting on the comparison, not the increment that should have been a decrement.  Well, remove_if fixed that, just goes to show to always use the std algorithms....

PR #2743, opened 2022-11-24
MAYA-126367 issue error message for invalid variant selection

pierrebai-adsk:
- If the prim name, variant name or variant selection is invalid, issue an error.
- The import still proceeds.


pierrebai-adsk:
Only PF failure is in the Linux interactive one, a render delegate test crashed. Unrelated to import.

vlasovi (approved):
Looks good

PR #2746, opened 2022-11-25
Fix duplication of homonyms

JGamache-autodesk:
Fixes: Two items that would resolve to the same path after name conflict resolution actually end up being duplicated into the same name.

So duplicating "bob1" and "bob2" in the same selection would result in a single "bob3" and a missing "bob4".

The code now creates and executes sub-duplication tasks in the main loop to make sure the name conflict resolution sees the complete picture.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.cpp
@@ -177,7 +177,7 @@ Ufe::SceneItem::Ptr UsdUndoDuplicateSelectionCommand::targetItem(const Ufe::Path
     }
 
     while (numSegments == path.getSegments().size()) {
-        CommandMap::const_iterator it = _perItemCommands.find(path);
+        it = _perItemCommands.find(path);
         if (it != _perItemCommands.cend() && it->second->duplicatedItem()) {
             Ufe::Path duplicatedChildPath
                 = sourcePath.reparent(path, it->second->duplicatedItem()->path());
--- a/lib/mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.cpp
@@ -50,18 +50,8 @@ UsdUndoDuplicateSelectionCommand::UsdUndoDuplicateSelectionCommand(
     const Ufe::ValueDictionary& duplicateOptions)
     : Ufe::SelectionUndoableCommand()
     , _copyExternalInputs(shouldConnectExternalInputs(duplicateOptions))
+    , _sourceSelection(selection)
 {
-    // TODO: MAYA-125854. If duplicating /a/b and /a/b/c, it would make sense to order the
-    // operations by SdfPath, and always check if the previously processed path is a prefix of the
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.cpp
```diff
@@ -50,18 +50,8 @@ UsdUndoDuplicateSelectionCommand::UsdUndoDuplicateSelectionCommand(
...
 {
-    // TODO: MAYA-125854. If duplicating /a/b and /a/b/c, it would make sense to order the
-    // operations by SdfPath, and always check if the previously processed path is a prefix of the
-    // one currently being processed. In that case, a duplicate task is not necessary because the
-    // resulting SceneItem should be built by using SdfPath::ReplacePrefix on the current item to
-    // get its location in the previously duplicated parent item.
-    for (auto&& item : selection) {
-        if (UsdSceneItem::Ptr usdItem = std::dynamic_pointer_cast<UsdSceneItem>(item)) {
-            // Currently unordered_map since we need to streamline the targetItem override.
-            _perItemCommands[item->path()] = UsdUndoDuplicateCommand::create(usdItem);
```
JGamache-autodesk, line 62: We can't pre-create all the tasks because the name conflict resolution code will not incrementally see the new paths.

JGamache-autodesk:
Pre-flight passed. Only one random test failure on the usual `testUsdExportPackage`.

lib/mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.cpp
```diff
@@ -70,24 +60,46 @@ UsdUndoDuplicateSelectionCommand::Ptr UsdUndoDuplicateSelectionCommand::create(
...
-    auto retVal = std::make_shared<UsdUndoDuplicateSelectionCommand>(selection, duplicateOptions);
-    if (retVal->_perItemCommands.empty()) {
-        // Could not find any item from this runtime in the selection.
-        return {};
+    bool canDuplicate = false;
+    for (auto&& item : selection) {
+        UsdSceneItem::Ptr usdItem = std::dynamic_pointer_cast<UsdSceneItem>(item);
+        if (usdItem) {
+            canDuplicate = true;
+            break;
```
hodoulp, line 68: [Comment] Would it be interesting to continue the loop to remove null `usdItem` from the list?

hodoulp: .. allowing to remove lines 88 to 90 & perhaps right away create a list/set of `UsdSceneItem::Ptr`

hodoulp (dismissed):

lib/mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.cpp
JGamache-autodesk: Maybe. Anyway this will be revisited at some point to fix MAYA-125854 and we will also remove all usdItems that have a parent in the list.

JGamache-autodesk: OK, that was a great idea and it also provided a very efficient way to fix MAYA-125854. Done.

hodoulp (dismissed):

PR #2747, opened 2022-11-28
MAYA-123716 - As a user, I'd like my duplicated prim to be on the Display Layer that the source prim was on

vlasovi:
This PR fixes a bug in the original code by introducing a code to properly replace USD prefixes when copying to USD.

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -1483,8 +1483,10 @@ bool PrimUpdaterManager::duplicate(
         }
         progressBar.advance();
 
-        std::string* rootRename = (dstChildName != srcRootPath.GetName()) ? &dstChildName : nullptr;
-        context._pushExtras.finalize(MayaUsd::ufe::stagePath(context.GetUsdStage()), rootRename);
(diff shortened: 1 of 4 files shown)


pierrebai-adsk (changes requested):
It is unclear to me why the change make the new prim be on the display layer, but I trust that it works. OTOH, a unit test that demonstrates the old problem and how the change fixes the issue would be welcome.

neilh-adsk:
@vlasovi , can you confirm Pierre's question "It is unclear to me why the change make the new prim be on the display layer," before we merge this change ?

pierrebai-adsk (approved):

vlasovi:
> @vlasovi , can you confirm Pierre's question "It is unclear to me why the change make the new prim be on the display layer," before we merge this change ?

The old code calculated the wrong UFE path by applying the rename string in the wrong place. The new code fixes this issue by using the full USD prefixes

seando-adsk:
@vlasovi As PierreB suggested could you add a unit test that covers this case?

vlasovi:
> @vlasovi As PierreB suggested could you add a unit test that covers this case?

Yes. A separate Jira ticket was added for that.

PR #2752, opened 2022-11-29
Update deprecated Python 3.7 functions

dj-mcg:
Update Py3.7 functions that have been deprecated in Py3.9

### DIFF
--- a/plugin/pxr/maya/lib/usdMaya/testenv/testUsdExportPackage.py
+++ b/plugin/pxr/maya/lib/usdMaya/testenv/testUsdExportPackage.py
@@ -33,7 +33,7 @@ class testUsdExportPackage(unittest.TestCase):
 
         # Deprecated since version 3.3: assertNotRegexpMatches has been renamed
         # to assertNotRegex()
-        if sys.version_info.major < 3 or sys.version_info.minor < 3:
+        if not hasattr(cls, "assertNotRegex"):
             cls.assertNotRegex = cls.assertNotRegexpMatches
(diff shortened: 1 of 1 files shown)


dj-mcg:
Oops, forgot about the py2 support requirement - I suppose I need version checking

dj-mcg:
<sigh> copy/paste error - fix incoming

plugin/pxr/maya/lib/usdMaya/testenv/testUsdExportPackage.py
```diff
@@ -31,6 +31,11 @@ def setUpClass(cls):
         standalone.initialize('usd')
         cmds.loadPlugin('pxrUsd')
 
+        # Deprecated since version 3.3: assertNotRegexpMatches has been renamed
+        # to assertNotRegex()
+        if sys.version_info.major < 3 or sys.version_info.minor < 3:
+            cls.assertNotRegex = cls.assertNotRegexpMatches
```
seando-adsk, line 37: The "or" doesn't seem correct to me. Instead of a python version check can you instead use the hasattr() method?

dj-mcg:
Sure thing! I was just pattern matching from similar checks in the open source such as [this one](https://github.com/Autodesk/maya-usd/blob/a3c602228c44d925fab72b111b161df30350e0c5/test/lib/mayaUsd/utils/testDiagnosticDelegate.py#L40), but I like how explicit hasattr is - thanks!

PR #2754, opened 2022-11-29
Add Custom Layer Data export flag

dgovil:
This PR adds a new `customLayerData` flag to the `MayaUSDExport` command.
The command takes a multi-array of 3 strings of key, value and type.

This had previously been submitted and was rejected because there was a concern about adding new flags to the exporter.

However I would really like to make an appeal to get this in again at this point because since then we've had multiple third party teams ask us for how we've been adding custom layer data to our exports, and sharing patches + recommending they build their own plugin isn't very ideal, especially when we feel this would be a valuable addition for several studio pipeline teams.

We'd also considered using chasers, however that becomes a little unwieldy because you really do need the three values of `key,value,type`, and with chasers, you would need to provide multiple separate arguments `(chaserName, <key>, value) , (chaserName, <key>Type, value) ` or pass it in as a json `(chaserName, json)`, both of which have slightly high overhead, especially when considering MEL users.

Anyway just putting this up again in the hopes that I can make a good case that the cost of an extra export flag is worth the value it gives.

(Note: I also reformatted the command markdown table , because it was previously ill formatted)


neilh-adsk:
Hi, @dgovil - we will look at this again.  I am also curious whether chasers can be extended to help elevate this problem, but as it stands yes, I see why you take the approach.  I'll get back to you about this

dgovil:
Thanks! I look forward to your thoughts on this

seando-adsk:
The original PR was https://github.com/Autodesk/maya-usd/pull/1081

pierrebai-adsk (approved):

PR #2755, opened 2022-11-29
Allow retrieving Arnold "uimin" metadata.

JGamache-autodesk:
Arnold stores these in the "hints" data, which is as valid as storing them in "metadata", so we update Ufe getMetadata to also look at hints on shader port attributes.


feldstj (approved):
LGTM!

PR #2762, opened 2022-12-01
MAYA-125918 auto re-edit when merging a Maya Ref

pierrebai-adsk:
This required many fixes and improvements.

The main challenges were:
- The fact that a given prim is a Maya reference is the knowledge of custom prim update manager.
- Some things were not available as UFE commands.
- It is possible to edit multiple Maya Ref at the same time in a single edit session (for example two ref as child of an edited prim)
- Many small bugs or limitations that had to be fixed.

The first item was addressed by having the possibility for prim updater to add extra work to be done once an operation (edit-as-Maya, merge-to-USD, etc) is done successfully:
- Add a composite UFE command to the prim updater context.
- This command contains sub-commands that are executed at the end of a prim updater operation.
- The prim updater customization classes can thus add final work that need to be done after the edit-as-Maya or merge-to-USD is done.

Augment the updater undo items to support UFE commands:
- Added execute to OpUndoItem because UFE and Maya commands sometimes do things differently in execute vs redo.
- Added a UfeCommandUndoItem to execute UFE commands in the prim updater manager undo item system.

Add commands for edit and switching variants:
- Added Edit-as-Maya UFE command.
- Added Set-Variant-Selection UFE command.
- Change context op to use the new commands.

Modify the Maya ref prim updater to use this:
- Made the Maya Ref prim updater automatically switch to the Maya Ref variant when merge-to-USD ends.
- This uses the new prim updater context additional command system.
- This trigger an edit due to the existing code to edit when switching variant on auto-edit Maya Ref.
- (That code is in the prim updater manager, listening to scene notifications.)

Small fixes to make things work:
- Made the USD path to DAG path map settable on the context to avoid recreating the context multiple times.
- (Recreating the context  caused the additional commands to be lost...)
- Fixed the code that check if commands and property changes, etc are allowed when one layer is invalid (null).

Testing:
- Add a unit test.

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -760,6 +760,14 @@ public:
         }
     }
 
+    void end()
+    {
+        if (_controllingFlag) {
+            *_controllingFlag = false;
+            _controllingFlag = nullptr;
+        }
+    }
+
 private:
     bool* _controllingFlag { nullptr };
 };
@@ -873,6 +881,12 @@ private:
 };
 #endif
 
+void executeAdditionalCommands(const UsdMayaPrimUpdaterContext& context)
+{
+    std::shared_ptr<Ufe::CompositeUndoableCommand> cmds = context.GetAdditionalFinalCommands();
+    UfeCommandUndoItem::execute("Additional final commands", cmds);
+}
+
 } // namespace
 
 PXR_NAMESPACE_OPEN_SCOPE
@@ -922,7 +936,7 @@ bool PrimUpdaterManager::mergeToUsd(
     MString progStr(
         VtDictionaryIsHolding<std::string>(userArgs, "rn_primName") ? "Caching to USD"
                                                                     : "Merging to USD");
-    MayaUsd::ProgressBarScope progressBar(10, progStr);
+    MayaUsd::ProgressBarScope progressBar(11, progStr);
     PushPullScope             scopeIt(_inPushPull);
 
     auto ctxArgs = VtDictionaryOver(userArgs, UsdMayaJobExportArgs::GetDefaultDictionary());
@@ -996,11 +1010,7 @@ bool PrimUpdaterManager::mergeToUsd(
     // 2) Traverse the in-memory layer, creating a prim updater for each prim,
     // and call Push for each updater.  Build a new context with the USD path
     // to Maya path mapping information.
-    UsdMayaPrimUpdaterContext customizeContext(
-        proxyShape->getTime(),
-        proxyStage,
-        ctxArgs,
-        std::get<UsdPathToDagPathMapPtr>(pushCustomizeSrc));
+    context.SetUsdPathToDagPathMap(std::get<UsdPathToDagPathMapPtr>(pushCustomizeSrc));
 
     if (!isCopy) {
         if (!FunctionUndoItem::execute(
@@ -1016,7 +1026,7 @@ bool PrimUpdaterManager::mergeToUsd(
     }
     progressBar.advance();
 
-    if (!pushCustomize(pulledPath, pushCustomizeSrc, customizeContext)) {
+    if (!pushCustomize(pulledPath, pushCustomizeSrc, context)) {
         return false;
     }
     progressBar.advance();
@@ -1073,6 +1083,10 @@ bool PrimUpdaterManager::mergeToUsd(
     }
     progressBar.advance();
 
+    scopeIt.end();
+    executeAdditionalCommands(context);
+    progressBar.advance();
+
     return true;
 }
 
@@ -1093,7 +1107,7 @@ bool PrimUpdaterManager::editAsMaya(const Ufe::Path& path, const VtDictionary& u
         return false;
     }
 
-    MayaUsd::ProgressBarScope progressBar(6, "Converting to Maya Data");
+    MayaUsd::ProgressBarScope progressBar(7, "Converting to Maya Data");
 
     PushPullScope scopeIt(_inPushPull);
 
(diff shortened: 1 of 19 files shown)


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -760,6 +760,14 @@ class PushPullScope
...
     }
 
+    void end()
+    {
+        if (_controllingFlag) {
+            *_controllingFlag = false;
+            _controllingFlag = nullptr;
+        }
+    }
+
```
pierrebai-adsk, line 770: This allows resetting the "in-push-pull" flag early because otherwise it would interfere with the auto-edit mechanism that is triggered by the final additional command when merging to USD a Maya Ref with auto-edit on.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -996,11 +1010,7 @@ bool PrimUpdaterManager::mergeToUsd(
     // 2) Traverse the in-memory layer, creating a prim updater for each prim,
     // and call Push for each updater.  Build a new context with the USD path
     // to Maya path mapping information.
-    UsdMayaPrimUpdaterContext customizeContext(
-        proxyShape->getTime(),
-        proxyStage,
-        ctxArgs,
-        std::get<UsdPathToDagPathMapPtr>(pushCustomizeSrc));
+    context.SetUsdPathToDagPathMap(std::get<UsdPathToDagPathMapPtr>(pushCustomizeSrc));
```
pierrebai-adsk, line 1014: Avoid re-creating the context since it would interfere with having additional commands. The command would be added in the context, but a new one would be used afterward,,,,

lib/mayaUsd/python/wrapPrimUpdater.cpp
```diff
@@ -246,6 +246,10 @@ void wrapPrimUpdaterContext()
             "GetArgs",
             &UsdMayaPrimUpdaterContext::GetArgs,
             boost::python::return_internal_reference<>())
+        .def(
+            "GetAdditionalFinalCommands",
+            &UsdMayaPrimUpdaterContext::GetAdditionalFinalCommands,
+            boost::python::return_internal_reference<>())
         .def("MapSdfPathToDagPath", &UsdMayaPrimUpdaterContext::MapSdfPathToDagPath);
```
pierrebai-adsk, line 253: Unfortunately, calling this currently fails in Python because UFE does not expose the UndoableCommand and CompositeUndoableCommand to Python. Not a blocker until someone wants to add additional commands from aprim updater in Python.

lib/mayaUsd/undo/OpUndoItemList.h
```diff
@@ -52,6 +52,10 @@ class OpUndoItem
     MAYAUSD_CORE_PUBLIC
     virtual ~OpUndoItem() = default;
 
+    /// \brief execute a single sub-operation. By default calls redo.
+    MAYAUSD_CORE_PUBLIC
+    virtual bool execute() { return redo(); }
+
```
pierrebai-adsk, line 58: As explained, the old design assumed redo == execute, but this not the case for some UFE commands.

lib/mayaUsd/ufe/EditAsMayaCommand.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+EditAsMayaUfeCommand::EditAsMayaUfeCommand(const Ufe::Path& path)
+    : _path(path)
+{
+}
+
+EditAsMayaUfeCommand::~EditAsMayaUfeCommand() { }
+
+void EditAsMayaUfeCommand::execute()
+{
```
pierrebai-adsk, line 39: In case you're wondering: there was already a EditAsMayaCommand class... but it is a Maya Command (MPxCommand) not a UFE command (Ufe::UndoableCommand). So I had to use a more specific name to disambiguate.

lib/usd/translators/mayaReferenceUpdater.cpp
```diff
@@ -291,8 +292,46 @@ bool PxrUsdTranslators_MayaReferenceUpdater::pushEnd()
...
+                if (!mayaAutoEditAttr.IsValid())
+                    continue;
+
+                bool autoEdited = false;
+                if (!mayaAutoEditAttr.Get<bool>(&autoEdited))
+                    continue;
+
+                auto cmd = MayaUsd::ufe::SetVariantSelectionCommand::create(
+                    getUfePath().pop(), primWithVariants, variantSetName, variantSelection);
+                getContext()->GetAdditionalFinalCommands()->append(cmd);
```
pierrebai-adsk, line 331: Switching the variant to the Maya ref one is all it takes to re-edit because there is a notification observer in PrimUpdaterManager that detect the switch and triggers the re-edit. That was written a long time ago, but was off because the old code was turning off auto-edit on merge.

vlasovi (dismissed):
Overall looks good. However, it's a big change and I would propose to wait and not merge it for the next release so that QA had enough time to identify any regressions

pierrebai-adsk:
> it's a big change and I would propose to wait and not merge it for the next release so that QA had enough time to identify any regressions

We can ask QA what they prefer next week.

lib/mayaUsd/ufe/EditAsMayaCommand.cpp
```diff
@@ -0,0 +1,60 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "EditAsMayaCommand.h"
+
+#include <mayaUsd/fileio/primUpdaterManager.h>
+#include <mayaUsd/undo/OpUndoItemRecorder.h>
+
+PXR_NAMESPACE_USING_DIRECTIVE
```
seando-adsk, line 21: You should not need this and its use is discouraged unless absolutely needed.

vlasovi (approved):

seando-adsk (approved):

PR #2764, opened 2022-12-02
MAYA-126179 merge-to-USD into variants

pierrebai-adsk:
Code refactor:
- Add a get method on the orphaned node manager to access the saved variant sets.
- Expose more helping functions to convert to text the orphaned node manager data, to help debugging.

Fix mid-merge orphaning:
- Move where the orphaned node manager data for a pulled prim is removed when merging to USD.
- This is necessary because while merging, USD sends out notifications which would trigger the orphaned node manage mid-merge and cause havok.

Support merging into variants:
- Add support to merge into variants in mergePrims().
- This involves augmenting the target path with variant selections for intermediary prims and to create a edit target for any final variant selection.
- We need two different ways to handle variant selections because SdfCopySpec() has a bug that prevents it from copying into a target path that ends with a variant selection.
- Fix how we determine if merge to USD should be enabled.
- Fix prim retrieval during merge when calling GetPrimAtPath.

The function GetPrimAtPath in USD does not support variant selection. (Surprisingly, IMO.) So we need to explicitly strip variant selection when retrieving a prim.  This API limitation is both non-intuitive and error-prone: given that the function does not support variant selection, why does it not ignore them? This fix makes merging into sub-variants (variant within variant) work.

### DIFF
--- a/lib/mayaUsd/fileio/orphanedNodesManager.cpp
+++ b/lib/mayaUsd/fileio/orphanedNodesManager.cpp
@@ -180,6 +180,17 @@ OrphanedNodesManager::Memento OrphanedNodesManager::remove(const Ufe::Path& pull
     return oldPulledPrims;
 }
 
+const PullVariantInfo& OrphanedNodesManager::get(const Ufe::Path& pulledPath) const
+{
+    const auto infoNode = pulledPrims().find(pulledPath);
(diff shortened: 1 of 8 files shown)


vlasovi (approved):
Looks good

PR #2775, opened 2022-12-12
Outliner rework of Material items:

JGamache-autodesk:
Right-click menu on prim when material is assigned.
- Unbind Material should say Unassign Material

Right-click menu on any scope that matches the default scope name
- Bind Material menu should be removed
- Assign New Material should be changed to Add New Material

Right-click menu on material node when nothing else is selected
- Assign Material to Selection should be hidden, not just disabled, if there's no selection
- Bind Material menu should be removed
- Assign New Material: should you be able to assign a material to a material? This should be removed

Right-click menu on material node when there is a selection
- Assign New Material: should be removed

Letter-casing in Assign New Material sub-menus is incorrect
- USD instead of Usd
- glTF PBR instead of Gltf Pbr

Added unit test for the new "Add New Material" command.


lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -34,6 +35,45 @@ PXR_NAMESPACE_USING_DIRECTIVE
 namespace MAYAUSD_NS_DEF {
 namespace ufe {
 
+namespace {
+// We will not use the value of UsdUtilsGetMaterialsScopeName() for the material scope.
+static const std::string kDefaultMaterialScopeName("mtl");
```
pierrebai-adsk, line 40: Is the reason for that eplained somwehere? half for my own understanding, half maybe the reason behind this comment should be added here.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -34,6 +35,45 @@ PXR_NAMESPACE_USING_DIRECTIVE
...
+
+#if (UFE_PREVIEW_VERSION_NUM >= 4010)
+bool connectShaderToMaterial(
+    Ufe::SceneItem::Ptr shaderItem,
+    UsdPrim             materialPrim,
+    const std::string&  nodeId)
+{
+    auto                 shaderUsdItem = std::dynamic_pointer_cast<UsdSceneItem>(shaderItem);
```
pierrebai-adsk, line 48: Are we always guaranteed that these casts will work? This being in a local function, I suppose we are guaranteed that the casts will succeed by the caller?

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -34,6 +35,45 @@ PXR_NAMESPACE_USING_DIRECTIVE
...
+    auto                 shaderPrim = UsdShadeShader(shaderUsdItem->prim());
+    UsdShadeOutput       materialOutput;
+    PXR_NS::SdrRegistry& registry = PXR_NS::SdrRegistry::GetInstance();
+    PXR_NS::SdrShaderNodeConstPtr shaderNodeDef
+        = registry.GetShaderNodeByIdentifier(TfToken(nodeId));
+    SdrShaderPropertyConstPtr shaderOutputDef;
+    if (shaderNodeDef->GetSourceType() == "glslfx") {
```
pierrebai-adsk, line 55: Again, pointer used without checking for validity: are we guaranteed the node will always exists?

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -34,6 +35,45 @@ PXR_NAMESPACE_USING_DIRECTIVE
...
+                "Cannot resolve which output of shader %s should be connected to surface",
+                nodeId.c_str());
+            return false;
+        }
+        materialOutput
+            = UsdShadeMaterial(materialPrim).CreateSurfaceOutput(shaderNodeDef->GetSourceType());
+        shaderOutputDef = shaderNodeDef->GetShaderOutput(shaderNodeDef->GetOutputNames()[0]);
+    }
+    UsdShadeOutput shaderOutput = shaderPrim.CreateOutput(
+        shaderOutputDef->GetName(), shaderOutputDef->GetTypeAsSdfType().first);
```
pierrebai-adsk, line 70: Same question: could GetShaderOutput ever return a null pointer? This one seems safer, I assume the API guarantees returned output node are not null and are merely returned by pointer to allow sharing.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -375,43 +418,162 @@ void UsdUndoAssignNewMaterialCommand::redo()
...
-        connectShaderToMaterial(addShaderCmd->insertedChild(), addMaterialCmd->newPrim());
+        connectShaderToMaterial(addShaderCmd->insertedChild(), addMaterialCmd->newPrim(), _nodeId);
     }
 }
 
-void UsdUndoAssignNewMaterialCommand::connectShaderToMaterial(
-    Ufe::SceneItem::Ptr shaderItem,
-    UsdPrim             materialPrim)
+void UsdUndoAssignNewMaterialCommand::markAsFailed()
 {
```
pierrebai-adsk, line 426: Ideas for future UFE: have a "Transactional" command compositor helper class that automtically undo + reset if not commited, to avoid having to sprinkle the "markAsFailed" everywhere. The reason is that if something throws an exception, markAsFailed won't be called and as far as I can tell, the already-executed commands won't be undone.

It is something I've been mulling over recently due to the bad interaction between composite command and create + execute.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -375,43 +418,162 @@ void UsdUndoAssignNewMaterialCommand::redo()
...
+
+UsdUndoAddNewMaterialCommand::~UsdUndoAddNewMaterialCommand() { }
+
+UsdUndoAddNewMaterialCommand::Ptr
+UsdUndoAddNewMaterialCommand::create(const UsdSceneItem::Ptr& parentItem, const std::string& nodeId)
+{
+    // Changing the hierarchy of invalid items is not allowed.
+    if (!parentItem || !parentItem->prim().IsActive())
+        return nullptr;
+
```
pierrebai-adsk, line 449: The other thing I'd like discussed in UFE: checking condition at creation time instead of execution time. The state can change between creation and execution, so verifying early can be detrimental because

- The condition can pass at creation but be invalid at execution, leading to an invalid state.
- The condition can fail at creation but would be valid at execution (happens in composite commands), which leads to denial of service when it should have worked.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -375,43 +418,162 @@ void UsdUndoAssignNewMaterialCommand::redo()
...
+    return std::make_shared<UsdUndoAddNewMaterialCommand>(parentItem, nodeId);
+}
+
+Ufe::SceneItem::Ptr UsdUndoAddNewMaterialCommand::insertedChild() const
+{
+    if (_cmds) {
```
pierrebai-adsk, line 455: There is no doc about insertChild: why it just inserts in the 2nd position of the list of commands is quite obscure to me.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -375,43 +418,162 @@ void UsdUndoAssignNewMaterialCommand::redo()
...
+        auto cmdsIt = _cmds->cmdsList().begin();
+        std::advance(cmdsIt, _createMaterialCmdIdx + 1);
+        auto addShaderCmd = std::dynamic_pointer_cast<UsdUndoCreateFromNodeDefCommand>(*cmdsIt);
+        return addShaderCmd->insertedChild();
```
pierrebai-adsk, line 459: And again, it assumes the 2nd child can successfully be cast and the pointer will be valid.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -375,43 +418,162 @@ void UsdUndoAssignNewMaterialCommand::redo()
...
+    }
+    if (shaderNodeDef->GetOutputNames().empty()) {
+        TF_RUNTIME_ERROR("Surface shader %s does not have any outputs", _nodeId.c_str());
+        markAsFailed();
+        return;
+    }
+    auto scopeItem
+        = std::dynamic_pointer_cast<UsdSceneItem>(Ufe::Hierarchy::createItem(_parentPath));
+    auto createMaterialCmd = UsdUndoAddNewPrimCommand::create(
+        scopeItem, shaderNodeDef->GetFamily().GetString(), "Material");
```
pierrebai-adsk, line 520: create() function can return nullptr, so Cmd can be null... are all conditions always guaranteed to be met and never be null?

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -375,43 +418,162 @@ void UsdUndoAssignNewMaterialCommand::redo()
...
+    }
+
+    //
+    // Create the Shader:
+    //
+    auto materialItem = std::dynamic_pointer_cast<UsdSceneItem>(
+        Ufe::Hierarchy::createItem(createMaterialCmd->newUfePath()));
+    auto createShaderCmd = UsdUndoCreateFromNodeDefCommand::create(
+        shaderNodeDef, materialItem, shaderNodeDef->GetFamily().GetString());
+    createShaderCmd->execute();
```
pierrebai-adsk, line 537: Same comment: is the cmd guaranteed to never be null?

pierrebai-adsk (changes requested):
I think checking for pointers would make the code more resiliant to future changes and subtle error in logic or potential unexpected invalid program state.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
JGamache-autodesk: It is a good thing that we do not do coverage testing, as this will have me add three unreachable lines of code.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
JGamache-autodesk: Yes, same issue in the current implementation of `Ufe::UndoableCommandGuard` which requires a setSuccess call to signify everything completed correctly, and fails to undo subtasks that were completed prior to the failure point. Designing a complete solution is hard and takes a significant time investment. Also, the use of exceptions adds a level of complexity since some clients like to swallow those and not propagate them upwards.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
JGamache-autodesk: Looks like you are doing the review of PR #2641 which indeed completely breaks any assumption about command ordering in UsdUndoAssignNewMaterialCommand. You need to select prims in two separate stages to trigger the bad outcome, which is a broken `redo()` that fails to connect items in the second stage, and the final selection that will only contain the first created shader and not the ones in the other stages.

This will have to be fixed (along with quite a few other commands and workflows that would be baffled by a global selection containing items from multiple stages, especially if there are some well placed name clashes).

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
pierrebai-adsk: It is possible to work around these kinds of issues by moving the function into a utility file and be able to test the function independently from its caller. We'll cross that bridge when we've actually built it. :)

frohnej-adsk (approved):
LGTM

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -752,6 +752,21 @@ bool selectionSupportsShading()
     return false;
 }
 
+#ifdef UFE_V3_FEATURES_AVAILABLE
```
pierrebai-adsk, line 755: I think this function is unused and came back during the conflict resolution.

pierrebai-adsk (approved):

lib/mayaUsd/ufe/UsdContextOps.cpp
JGamache-autodesk: I see it used at line 1211. The Linux and OS/X builds do not allow an unused function to exist. I agree it should not have suddenly appeared from conflict resolution since it was added a month ago in PR #2762 and should have been in the branch I started with. This is strange.

JGamache-autodesk: Not strange finally. The code was added last month in a branch that was merged yesterday. This is why it appeared in as a conflict to resolve in my branch.

PR #2283, opened 2022-04-12
Added mesh export options to preserve and remap UV set names

J-Mo63:
Currently, the ADSK plugin always authors UV sets as `st`, `st1`, `st2`, etc. While AL's pipeline ultimately assumes that `map1` should be renamed to `st` on export and additional named sets should retain their naming as specified in Maya.

A new export option, `-preserveUVSetNames` (`-puv`), has been added to allow for the preservation of set names:
```
def Mesh "my_geo"
{
    texCoord2f[] primvars:rigging_utility = [...] (
        interpolation = "faceVarying"
    )
    int[] primvars:rigging_utility:indices = [...]

    texCoord2f[] primvars:map1 = [...] (
        interpolation = "faceVarying"
    )
    int[] primvars:map1:indices = [...]
}
```

A second export option, `-remapUVSetsTo` (`-ruv`), has been added on top of this to provide a generalised solution to our specific use-case, where the options `remapUVSetsTo=[[map1, st], [rigging_utility, my_uv_set]]` would produce:
```
def Mesh "my_geo"
{
    texCoord2f[] primvars:my_uv_set = [...] (
        interpolation = "faceVarying"
    )
    int[] primvars:my_uv_set:indices = [...]

    texCoord2f[] primvars:st = [...] (
        interpolation = "faceVarying"
    )
    int[] primvars:st:indices = [...]
}
```

PR #2776, opened 2022-12-12
Fix material export when preserving UVset names

JGamache-autodesk:
The export options -remapUVSetsTo and -preserveUVSetNames are now correctly handled when exporting materials.


JGamache-autodesk:
@J-Mo63 This fixes issues introduced in #2283 where the exported materials would not map to the correct UV set names.

vlasovi (approved):
Nice

neilh-adsk:
Thanks @JGamache-autodesk for finding and fixing this up.  Adding a utility function is also a nice touch

PR #2780, opened 2022-12-13
MAYA-125728 - Implement display layer's hide-on-playback flag for instanced geometry

vlasovi:


vlasovi:
@pierrebai-adsk I split the code into three commits for simplification. The first two are just refactoring in preparation for the main change. It may be easier to review the PR like that.

pierrebai-adsk (approved):

PR #2784, opened 2022-12-16
Fix empty values for remapUVSetsTo

dj-mcg:
This fixes this test for environments with remapUVSetsTo defaults configured


JGamache-autodesk (approved):

seando-adsk:
@dj-mcg Hi Dan, our devs have some comments/questions about this change.

Doesn't this change hide any problem that could happen with our default values by never letting them be tested?

This change is not problematic for us, but raises the question about accommodating users with custom environments that override default values.

Where is "remapUVSetsTo" default getting modified? Is it a custom plugin you are using internally?

PR #2788, opened 2022-12-16
Correctly expose Material and NodeGraph attributes in Maya AE

JGamache-autodesk:


lib/mayaUsd/utils/util.cpp
```diff
@@ -740,7 +740,7 @@ std::string UsdMayaUtil::prettifyName(const std::string& name)
                 prettyName += ' ';
             }
             prettyName += nextLetter;
-        } else if (name[i] == '_') {
+        } else if (name[i] == '_' || name[i] == ':') {
```
seando-adsk, line 743: Did you run this change by Natalia?

seando-adsk (dismissed):

seando-adsk:
@JGamache-autodesk It all looks good, just one minor question about UI change. 

Also did you verify the ae_template.py changes work in the older versions of USD that we support? At the very least it needs to work in USD 21.11 (used by Maya 2020/2022/2023).

lib/mayaUsd/utils/util.cpp
JGamache-autodesk: Just did. This change is approved. I need to make sure it is called for all UFE attributes, which requires a small change in UsdAttributeHolder.cpp to handle "uiname" metadata for all attributes instead of only NodeGraph and Materials.

JGamache-autodesk:
I am not worried about USD versions.
Was worried enough about UFE version to add a check for `hasattr(ufe, "NodeDef")`
Slightly worried about Python version since that code needs to be Python 2.7 compatible.
So I will try this under Maya 2020 Python 2.7 and see it it correctly reverts to having all attributes under "Extra Attributes" without emitting any error message.

JGamache-autodesk:
Tested with:
Maya 2020.4, USD 21.11, Python 2.7: AE code not exercised since parameters are not shown for USD data.
Maya 2022.3, USD 21.11, Python 2.7: AE code exercised. Check on UFE version hit, so we only see authored attributes in the Extra Attributes section. Fixed issues with Python 2.7.

seando-adsk (approved):

seando-adsk:
@alicedegirolamo Can you please do a code review before I merge this?

lib/mayaUsd/ufe/UsdAttributeHolder.cpp
```diff
@@ -188,14 +189,34 @@ Ufe::Value UsdAttributeHolder::getMetadata(const std::string& key) const
         if (PXR_NS::UsdShadeNodeGraph(usdPrim())) {
             if (PXR_NS::UsdShadeInput::IsInput(_usdAttr)) {
                 const PXR_NS::UsdShadeInput input(_usdAttr);
-                const std::string           metadata = input.GetSdrMetadataByKey(tok);
+                std::string                 metadata = input.GetSdrMetadataByKey(tok);
+                if (metadata.empty() && key == "uiname") {
```
alicedegirolamo, line 193: Maybe we can store this `"uiname"` as a constant somewhere, I see this string used in multiple parts.
Everything else seems ok to me.

alicedegirolamo:
@JGamache-autodesk I see we get the the `"uiname"` attribute metadata. So for this task I suppose we don't need to get also `"uiorder"` and `"uifolder"` metadata, correct?

seando-adsk (dismissed):

alicedegirolamo (approved):

JGamache-autodesk:
@seando-adsk can you please merge? Thanks!

PR #2789, opened 2022-12-19
MAYA-109455 UFE USD camera commands

pierrebai-adsk:
Implemeted the UFE commands necessary to set values on a camera via UFE.

Extended the available base classes to implement UFE commands that modify USD data. Make each base class do one thing only, to clarify the purpose of each.

Add UsdUndoCapture class:
- It captures the changes to USD data within a UFE command.
- Its sub-classes only need to implement executeImplementation() and optionally setImplementation().
- It provides executeWithUndoCapture to do the command execution implementation while capture USD changes.
- Same for setWithUndoCapture.
- Provides undoUsdChanges() to undo all changes that were capture.
- Same with redoUsdChanges, to redo the changes.

Refactor UsdUndoableCommand class:
- Make UsdUndoableCommand use UsdUndoCapture class.
- Add templated variable-arguments constructor to support any command base classs that can take arguments in its constructor.
- This includes taking zero argument.

Add UsdFunctionUndoableCommand class:
- This is a UFE command implementation where the actual command execution is done in a free function.
- Avoids having to declare entire classes only to implement a single execute function.

Added UsdUndoableSetCommand class:
- This is for undoable commands that sets value and must implement a set function.
- Also has a Function sub-class to impelment it with a single free function.

Refactor existing command to use the new base class:
- Refactor UsdAttribute commands to use the new classes.
- Refactor UsdLight commands to use the new classes.
- Refactor UsdSetMatridx4Undoable command the same way,
- This avoids code duplication.
- There are more commands that could be refactored, but this is a starting set.

Implement USDCamera commands.
- Implement various commands related to UFE/USD cameras.
- These use the new funciton-base command helper classes.
- Added helper function to convert units.
- Re-verified the units used by the camera and documented them.

### DIFF
--- a/lib/mayaUsd/ufe/CMakeLists.txt
+++ b/lib/mayaUsd/ufe/CMakeLists.txt
@@ -60,6 +60,7 @@ if(CMAKE_UFE_V2_FEATURES_AVAILABLE)
             UsdTransform3dUndoableCommands.cpp
             UsdUIInfoHandler.cpp
             UsdUIUfeObserver.cpp
+            UsdUndoableCommand.cpp
             UsdUndoAddNewPrimCommand.cpp
             UsdUndoCreateGroupCommand.cpp
             UsdUndoInsertChildCommand.cpp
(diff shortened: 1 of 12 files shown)


vlasovi (approved):
Looks good

PR #2790, opened 2022-12-20
MAYA-124105 Fix opacity computations for MaterialX and USD materials

JGamache-autodesk:


seando-adsk (approved):
Not my area of expertise, but the code changes look okay and there is a unit test. All good.

seando-adsk:
@vlasovi Ivan can you please code review this when you have time. Thanks.

lib/mayaUsd/render/MaterialXGenOgsXml/OgsFragment.cpp
```diff
@@ -145,7 +145,8 @@ class GlslGeneratorWrapperBase
 
         // Maya images require a texture coordinates to be flipped in V.
         genOptions.fileTextureVerticalFlip = true;
-        genOptions.hwTransparency = mx::isTransparentSurface(_element, generator.getTarget());
+        // Enabling hwTransparency to ensure a vec4 output with correct alpha value.
+        genOptions.hwTransparency = true;
```
vlasovi, line 149: @JGamache-autodesk What effect does this have? I just want to make sure we don't force opaque objects to be treated as transparent objects here.

JGamache-autodesk: Setting it to false causes the fragment to return a color3 and then there is no alpha we can use if the shader is transparent.

Setting it to true forces it to always return a color4. The alpha channel will be set correctly and this will allow to turn transparency at the effect instance level.

vlasovi (approved):
Looks good

JGamache-autodesk:
@seando-adsk Ready for merge.

PR #2794, opened 2022-12-20
MAYA-122927 interaction between deactivation and cancel edit

pierrebai-adsk:
When a Maya reference is edited, its USD prim is deactivated to hide it in the outliner. If its parent prim is then deactivated, the edit is orphaned. At that point the Maya Ref prim and its parent are both deactivated. If the edit is cancelled while the parent is deactivated, the code could not activate the Maya Ref USD prim because children of deactivated prims cannot be modified. We needed to fix this.

The fix is to add code to temporarily reactivate ancestors prims when trying to activate a prim. The design actually provides a class to temporarily activate all ancestors of a prim, for any possible edits.

- Add PrimActivation class to temporarily make a prim accessible even when its ancestors are deactivated.
- It works by activating all inactive ancestors by authoring opinions in the session layer.
- Use this class in the addExcludedFromRendering and removeExcludedFromRendering functions.
- Add a flag in the orphaned nodes manager to avoid responding to notifications caused by its own responses to notifications.
- This avoids infinite recursion when the work done to orphan an edited node caused notifications to be sent.
- In our case, the temporary reactivation of prims would cause an infinite flip-flop or orphaning and de-orphaning.

A unit test already checked the case where an edited prim was under a dectivated prim, but had not checked for the proper result due to the known bug. Simply fix the test for the fix.

Also, fix that the auto-edit flag was not being properly updated due to the same ancestor-is-deactivated problem.

### DIFF
--- a/lib/mayaUsd/fileio/orphanedNodesManager.cpp
+++ b/lib/mayaUsd/fileio/orphanedNodesManager.cpp
@@ -149,6 +149,22 @@ void handlePathChange(
     }
 }
 
+// Control the orphaned nodes manager in-orphaning flag.
+class Orphaning
+{
+public:
+    Orphaning(int& orphaning)
+        : _orphaning(orphaning)
+    {
+        ++_orphaning;
+    }
+
+    ~Orphaning() { --_orphaning; }
+
+private:
+    int& _orphaning;
+};
+
 } // namespace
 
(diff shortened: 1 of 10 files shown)


vlasovi (approved):
Looks good

PR #2800, opened 2022-12-29
Maya-126871 Delete compound attributes does not remove their connections.

alicedegirolamo:
# Description

With this PR, when deleting a compound attribute, the associated connections are also deleted. 

Fixes [issue 126871](https://jira.autodesk.com/browse/MAYA-126871).

# How Has This Been Tested?

I updated the existing `testAttributes.py` adding the test `testRemoveCompoundAttribute`.

### DIFF
--- a/lib/mayaUsd/ufe/UsdAttributes.cpp
+++ b/lib/mayaUsd/ufe/UsdAttributes.cpp
@@ -455,7 +455,7 @@ bool UsdAttributes::canRemoveAttribute(const UsdSceneItem::Ptr& item, const std:
     }
     return false;
 }
-static void removeConnections(const PXR_NS::UsdPrim& prim, const SdfPath& srcPropertyPath)
+static void removeConnections(const PXR_NS::UsdPrim& prim, const PXR_NS::SdfPath& srcPropertyPath)
 {
     // Remove the connections with source srcPropertyPath.
     for (const auto& node : prim.GetChildren()) {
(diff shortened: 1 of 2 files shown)


JGamache-autodesk (dismissed):
You will most probably have to run clang-format on this one. I can see lines 491-493 are not to spec.

alicedegirolamo:
> You will most probably have to run clang-format on this one. I can see lines 491-493 are not to spec.

I re-run the clang-format but it seems correct to the clang-format.

lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -455,6 +455,24 @@ bool UsdAttributes::canRemoveAttribute(const UsdSceneItem::Ptr& item, const std:
     }
     return false;
 }
+static void removeConnections(const PXR_NS::UsdPrim& prim, const SdfPath& srcPropertyPath)
```
seando-adsk, line 458: PXR_NS::SdfPath

test/lib/ufe/testAttributes.py
```diff
@@ -183,6 +183,64 @@ def testAddRemoveAttribute(self):
         with self.assertRaisesRegex(KeyError, "Attribute 'MyAttribute' does not exist") as cm:
             attr = ball35Attrs.attribute("MyAttribute")
 
+    @unittest.skipIf(os.getenv('UFE_PREVIEW_VERSION_NUM', '0000') < '4024', 'Test for UFE preview version 0.4.24 and later')
```
seando-adsk, line 186: Should really say something about why the test needs 0.4.24. Such as "Test requires xxx feature only available on Ufe 0.4.24 and later"

seando-adsk (changes requested):

test/lib/ufe/testAttributes.py
alicedegirolamo: I update it.

JGamache-autodesk (approved):

seando-adsk (approved):

PR #2803, opened 2023-01-05
MAYA-125036 correctly identify orphaned edited prims

pierrebai-adsk:
Also fixes MAYA-126891

- Provide a function to explicitly determine if an edited prim is orphaned.
- Provide the function in Python too.
- Use it both to build the context menu and do the outliner item rendering.

### DIFF
--- a/lib/mayaUsd/python/wrapPrimUpdaterManager.cpp
+++ b/lib/mayaUsd/python/wrapPrimUpdaterManager.cpp
@@ -117,6 +117,11 @@ std::string readPullInformationString(const PXR_NS::UsdPrim& prim)
     return dagPathStr;
 }
 
+bool isEditedPrimOrphaned(const PXR_NS::UsdPrim& prim)
+{
+    return MAYAUSD_NS_DEF::isEditedAsMayaOrphaned(prim);
+}
+
 } // namespace
 
 void wrapPrimUpdaterManager()
@@ -128,5 +133,6 @@ void wrapPrimUpdaterManager()
         .def("canEditAsMaya", canEditAsMaya)
         .def("discardEdits", discardEdits)
         .def("duplicate", duplicate, duplicate_overloads())
+        .def("isEditedAsMayaOrphaned", isEditedPrimOrphaned)
         .def("readPullInformation", readPullInformationString);
 }
--- a/lib/mayaUsd/python/wrapPrimUpdaterManager.cpp
+++ b/lib/mayaUsd/python/wrapPrimUpdaterManager.cpp
@@ -67,7 +67,7 @@ bool mergeToUsd(const std::string& nodeName, const VtDictionary& userArgs = VtDi
         return false;
 
     Ufe::Path path;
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/python/wrapPrimUpdaterManager.cpp
```diff
@@ -117,6 +117,11 @@ std::string readPullInformationString(const PXR_NS::UsdPrim& prim)
     return dagPathStr;
 }
 
+bool isEditedPrimOrphaned(const PXR_NS::UsdPrim& prim)
+{
+    return MAYAUSD_NS_DEF::isEditedAsMayaOrphaned(prim);
```
seando-adsk, line 122: Just use the namespace name "MayaUsd::"

lib/mayaUsd/python/wrapPrimUpdaterManager.cpp
```diff
@@ -128,5 +133,6 @@ void wrapPrimUpdaterManager()
         .def("canEditAsMaya", canEditAsMaya)
         .def("discardEdits", discardEdits)
         .def("duplicate", duplicate, duplicate_overloads())
+        .def("isEditedAsMayaOrphaned", isEditedPrimOrphaned)
```
seando-adsk, line 136: Why does the python wrapper have a different name than the internal helper function?

lib/mayaUsd/ufe/MayaUIInfoHandler.cpp
```diff
@@ -65,6 +67,13 @@ MayaUsd::ufe::UsdSceneItem::Ptr pulledUsdAncestorItem(const Ufe::SceneItem::Ptr&
         }
     }
 
+    // Validate that the prim pointed to by the UFE path really is the
+    // pulled node and not some node that just happens to have the same
+    // name. This can happen, for example, when two variants each contain
+    // a child with the same name, one pulled, one not.
+    if (MAYAUSD_NS_DEF::isEditedAsMayaOrphaned(usdItemPath))
```
seando-adsk, line 74: MayaUsd::

seando-adsk (changes requested):

lib/mayaUsd/python/wrapPrimUpdaterManager.cpp
pierrebai-adsk: The Python function has the same name as the real C++ function. The gave the internal function a different name because otherwise it would confuse the compiler (and code reader) which would see the same name as overload. The real C++ function is named isEditedAsMayaOrphaned like the Python function.

lib/mayaUsd/python/wrapPrimUpdaterManager.cpp
pierrebai-adsk: I was using the same as existing code above it. I will change both.

seando-adsk (approved):

seando-adsk:
@vlasovi Ivan can you please do a quick code review before I merge this. Thanks.

vlasovi (approved):
Looks good

PR #2807, opened 2023-01-06
Update mayaUsd_layerEditorFileDialogs.mel

santosg87:
this is a simple fix to enable the use of the native file dialog when loading layers on the layer Editor, the original code has an argument forcing use of the maya native browser. this change simply removes the that flag.

I have tested that this works as expected and it doesn't affect any other issues.


seando-adsk (approved):
Changes are okay since this will use the dialog pref from Maya.

seando-adsk:
No need for the preflight since this change contains just a MEL file change that won't be tested by Preflight.

PR #2809, opened 2023-01-10
Ability to put a proxy shape's translation to "sleep"

fabal:
This PR adds the ability to pause an AL proxy shape's translation.
This is a feature we use internally, here at Animal Logic, when users interact with a stage, via an in-house application, and don't want to see the changes dynamically translated, as they can take a while to load and render. Once happy with the stage's configuration, users would then "wake up" (resync) the proxy shape.


nxkb:
@fabal Do you think we might be able to reconcile this "sleepTranslation" with the "pauseUpdates" attribute? I think their intention is very similar. The user story above is pretty exact for why we added pauseUpdates originally (although i think the actual sleep broke in some of the refactors, so maybe we need your new lines). Is there a difference in how you'd want them to act?

fabal:
Thanks for the heads up @nxkb . Just had a chat with the team and it appears no one knew, and every one is a bit taken aback! I'm going to close this PR off for now. The team will look into how to reconcile these 2. If there's any details you can share about the original implementation, "pause updates", particularly about the caveats you alluded to, that would be greatly appreciated.

PR #2817, opened 2023-01-11
MAYA-127095 errors instead of crashes

pierrebai-adsk:
We should strive to reduce the number of scenarios under which Maya crashes. Instead, when possible, we should  report an error to user and abort the operation. MayaUSD has extensive undo/redo capabilities now and code is able to undo partially done operations. Given this, errors should be reported and propagated back.

The macros TF_AXIOM and TF_FATAL_ERROR cause crashes when their conditions are not met, even in release builds. Remove TF_AXIOM and TF_FATAL_ERROR, replacing them with:
- TF_WARN when the situation was not an error but an unusual situation.
- TF_VERIFY when it still warrant an error message to the user, for example during error handling.
- Simple if/else, with an error return value when not a hard error.
- Simple if when the condition was trivial, like an empty node when iterating over USD nodes.
- In a few cases, TF_DEV_AXIOM when the condition was really strictly enforced by a class constructor.
- Entirely removed, when the code was already dealing with the error.
- C++ exception when the error needs to propagate to caller, in particular in constructors.

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -238,7 +238,7 @@ bool allowTopologyModifications(MDagPath& root)
 // prim as the root of the USD hierarchy to be pulled.  The UFE path and
 // the prim refer to the same object: the prim is passed in as an
 // optimization to avoid an additional call to ufePathToPrim().
-using PullImportPaths = std::pair<std::vector<MDagPath>, std::vector<Ufe::Path>>;
+using PullImportPaths = std::vector<std::pair<MDagPath, Ufe::Path>>;
 PullImportPaths pullImport(
     const Ufe::Path&                 ufePulledPath,
     const UsdPrim&                   pulledPrim,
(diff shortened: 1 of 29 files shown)


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -238,7 +238,7 @@ bool allowTopologyModifications(MDagPath& root)
 // prim as the root of the USD hierarchy to be pulled.  The UFE path and
 // the prim refer to the same object: the prim is passed in as an
 // optimization to avoid an additional call to ufePathToPrim().
-using PullImportPaths = std::pair<std::vector<MDagPath>, std::vector<Ufe::Path>>;
+using PullImportPaths = std::vector<std::pair<MDagPath, Ufe::Path>>;
 PullImportPaths pullImport(
```
pierrebai-adsk, line 242: By changing from pair of vectors to vector of pair, we avoid all problems where the pair count do not match.

vlasovi (dismissed):
Looks good

Eiken:
Loving this initiative. I would call this our main problem with USD in Maya and it makes a lot of artists sad when it just crashes the whole scene because you looked at it wrong or tries to read something that is not supported yet.

pierrebai-adsk:
Rebasing on latest dev branch. Will repoen a new PR

PR #2823, opened 2023-01-17
LOOKDEVX-1091 - Add Ufe log on console to notify shader attribute without a default value

selvatm-autodesk:
- The goal is to notify the user when a property of any shader lacks a default value.
- The properties that lack a default value will start to log a warning when the user interacts with them in the editor (that involves the `get()` of the property value).
- Known attributes are maintained in a list (`std::unordered_set`) to avoid being too verbose for properties we have already printed on Console.

### DIFF
--- a/lib/mayaUsd/ufe/UsdShaderAttributeHolder.cpp
+++ b/lib/mayaUsd/ufe/UsdShaderAttributeHolder.cpp
@@ -29,6 +29,8 @@
 #include <pxr/base/vt/value.h>
 #include <pxr/pxr.h>
 
+#include <ufe/log.h>
+
 namespace MAYAUSD_NS_DEF {
 namespace ufe {
 
@@ -89,6 +91,10 @@ std::string UsdShaderAttributeHolder::defaultValue() const
 
 std::string UsdShaderAttributeHolder::nativeType() const { return _sdrProp->GetType(); }
 
+//! A List of {info:id|property_name} of the properties that lack the default value, to avoid
+//! repeating the warnings on console.
+static std::unordered_set<std::string> s_ListPropertiesWithoutDefaultValue {};
+
 bool UsdShaderAttributeHolder::get(PXR_NS::VtValue& value, PXR_NS::UsdTimeCode time) const
 {
     if (isAuthored()) {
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/UsdShaderAttributeHolder.cpp
```diff
@@ -89,13 +91,34 @@ std::string UsdShaderAttributeHolder::defaultValue() const
 
 std::string UsdShaderAttributeHolder::nativeType() const { return _sdrProp->GetType(); }
 
+//! A List of the properties that are known to lack the default value.
+//! Insert the name of the property to avoid logging the warning.
+static const std::unordered_set<std::string> s_ListPropertiesWithoutDefaultValue {
```
selvatm-autodesk, line 96: A list of properties that we have already noticed to lack the default value, to prevent them to produce warnings. 

It will be maintained as need. Is it the correct place to put it? (just above the method that make use of it). @JGamache-autodesk

JGamache-autodesk: Please initialize this set as empty. It will be used to track the shaders that previously printed a warning. See next comment.

lib/mayaUsd/ufe/UsdShaderAttributeHolder.cpp
```diff
@@ -89,13 +91,34 @@ std::string UsdShaderAttributeHolder::defaultValue() const
...
     value = vtValueFromString(usdAttributeType(), defaultValue());
+
+    if (defaultValue().empty()) {
+        const auto propertyName = _sdrProp->GetName().GetString();
+        bool       isKnownProperty = s_ListPropertiesWithoutDefaultValue.find(propertyName)
+            != s_ListPropertiesWithoutDefaultValue.end();
+        if (!isKnownProperty) {
+            const std::string msg
+                = "Warning: Shader property '" + propertyName + "' does not have a default value.";
+            UFE_LOG(msg.c_str());
```
JGamache-autodesk, line 118: This is where you add a new entry to s_ListPropertiesWithoutDefaultValue. Please include the shader info:id in both the message and in the entry because we would like to know the exact identity of the problematic shader.

JGamache-autodesk (changes requested):

JGamache-autodesk (approved):

seando-adsk (approved):

PR #2834, opened 2023-01-23
Skip new test if UFE attribute metadata not available

dj-mcg:
UFE attribute metadata API was introduced in UFE 3. The new testAttributeNiceNames should be disabled if the API is now available (UFE 2 and earlier)


seando-adsk (approved):
Thanks for finding and fixing this. It shows we have a missing build/test scenario in our preflight.

PR #2835, opened 2023-01-23
MAYA-124789 improve NURBS export

pierrebai-adsk:
- No longer require that the first two knots be equal. I don't know where that requirement came from... I think it was a mis-reading of the USD spec.
- Added validation of the knot domain: USD bezier curve do not support knot domains, so if a partial domain is needed, we have to use a NURBS curve, not a bezier curve.
- This solves the problem that exported curve were longer than expected.
- Fixed the import of USD NURBS curve when the degree is not 3.
- Add unit test for high-degree NURBS import.
- Fix the names of other import tests. They used to be called export, which was confusing.


pierrebai-adsk:
Sorry to re-open the review: I found bugs in the importer that I thought would be practical to fix in the PR.

lib/mayaUsd/fileio/translators/translatorCurves.cpp
```diff
@@ -99,7 +99,7 @@ bool UsdMayaTranslatorCurves::Create(
     }
 
     VtArray<GfVec3f> points;
-    VtArray<int>     curveOrder;
+    VtArray<int>     curveOrders;
     VtArray<int>     curveVertexCounts;
```
pierrebai-adsk, line 103: Changed the name of the variable to be plural like the other to represent this is a vector of values.

lib/mayaUsd/fileio/translators/translatorCurves.cpp
```diff
@@ -156,39 +156,49 @@ bool UsdMayaTranslatorCurves::Create(
...
+                nurbsSchema.GetKnotsAttr().Get(&curveKnots);  // not animatable
                 nurbsSchema.GetRangesAttr().Get(
                     &curveRanges); // not animatable or actually used....
             }
 
+            const int curveDegree = curveOrders[curveIndex] - 1;
+            const int pointCount = curveVertexCounts[curveIndex];
+
+            // The USD NURBS curve schema (UsdGeomNurbsCurves) defines the number
+            // of knots as: # points + degree + 1.
```
pierrebai-adsk, line 169: Note: the *official* USD docs are unfortunately wrong and claim the number of knots is the # segment + 2 * degree  + 1, The degree must not be multiplied by 2. I'll see if I can submit a documentation fix or at least report it in the USD interest group.

vlasovi (approved):
Looks good

PR #2836, opened 2023-01-23
Preserve NodeGraph boundaries on duplicate.

JGamache-autodesk:
The post-duplicate cleanup would delete an unconnected port knowing that it would still survive via the shader node definition.

However NodeGraphs do not have an underlying definition, so we need to preserve those unconnected ports even though they might not have a value at this point in time.


frohnej-adsk (approved):
LGTM

JGamache-autodesk:
@seando-adsk ready for merge.

seando-adsk (approved):

PR #2837, opened 2023-01-24
MAYA-127355 :  Deleting a node does not delete its connections.

alicedegirolamo:
# Description

This PR solves the issue described in [Jira ticket MAYA-127355](https://jira.autodesk.com/browse/MAYA-127355) : deleting a node does not delete its connections. 

## Type of change

# How Has This Been Tested?

- Added automatic test.


lib/mayaUsd/ufe/UsdAttributes.cpp
```diff
@@ -471,6 +474,39 @@ static void removeConnections(const PXR_NS::UsdPrim& prim, const PXR_NS::SdfPath
...
+    for (const auto& attr : kPrimAttrs) {
+        const PXR_NS::SdfPath kPropertyPath = kPrimPath.AppendProperty(attr.GetName());
+        const auto            kBaseNameAndType
+            = PXR_NS::UsdShadeUtils::GetBaseNameAndType(PXR_NS::TfToken(attr.GetName()));
+
+        if (kBaseNameAndType.second == PXR_NS::UsdShadeAttributeType::Output) {
+            removeConnections(primParent, kPropertyPath);
+        }
+        if (kBaseNameAndType.second == PXR_NS::UsdShadeAttributeType::Input) {
+            removeConnections(prim, kPropertyPath);
```
alicedegirolamo, line 508: Note: this is the case for the compound, we want to delete the connections to its children. For the delete feature, this case is redundant, since we are deleting also the compound children and so all the children connections. But I think we should include this case in this function since for generic cases, not only deleting a prim, is a necessary step.

test/lib/ufe/testDeleteCmd.py
```diff
@@ -361,5 +361,90 @@ def testDeleteRestrictionHierarchyWithChildrenOnDifferentLayer(self):
         self.assertTrue(stage.GetPrimAtPath('/TreeBase/trunk'))
         self.assertTrue(stage.GetPrimAtPath('/TreeBase/newChild'))
 
+    def testDeleteAndRemoveConnections(self):
+        '''Test deleting a prim and its connections'''
+        
+        # Load a scene with a compound.
+      
+        mayaUtils.openTestScene("MaterialX", "multiple_connections.ma" )
```
JGamache-autodesk, line 369: You are using a Maya scene that was created with Maya 2024. It will not load back in Maya 2022 or 2023. You have the option to rebuild the scene using Maya 2022 so the test can run across Maya versions, or to make the test run only on Maya 2024 and later using a `@SkipIf` decorator.

lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
```diff
@@ -67,6 +68,86 @@ UsdUndoDeleteCommand::Ptr UsdUndoDeleteCommand::create(const PXR_NS::UsdPrim& pr
...
+                            PXR_NS::UsdShadeAttributeType::Output);
+
+                    const std::string sourceName = prefix + sourcesInfo[0].sourceName.GetString();
+
+                    auto srcAttr = connectedPrim.GetAttribute(PXR_NS::TfToken(sourceName));
+
+                    if (srcAttr) {
+                        UsdShadeConnectableAPI::DisconnectSource(attr, srcAttr);
+                        // Check if we can remove also the property.
+                        // Remove the property.
```
JGamache-autodesk, line 116: Would be nice if you could implement that.

JGamache-autodesk (changes requested):

JGamache-autodesk (approved):

seando-adsk (approved):

PR #2839, opened 2023-01-24
MAYA-127119 - As a user, I'd like to save my USD root file as relative to the scene file

vlasovi:


lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -117,6 +117,14 @@ std::string UsdMayaUtilFileSystem::getMayaSceneFileDir()
     return std::string();
 }
 
+std::string UsdMayaUtilFileSystem::getPathRelativeToMayaSceneFile(const std::string& fileName)
+{
+    ghc::filesystem::path absolutePath(fileName);
+    ghc::filesystem::path basePath(getMayaSceneFileDir());
+    ghc::filesystem::path relativePath = absolutePath.lexically_relative(basePath);
+    return relativePath.generic_string();
```
pierrebai-adsk, line 125: Note that `lexically_relative` can return an empty path is multiple corner cases, for example if both path are entirely separate, for example in Windows if each path is on a different drive (example: C:, D:).

IDK how we want to handle the case where the user requested relative paths and they cannot be made relative: do we show a warning, an error?

At the very least, we probably want to return the original fileName in that case.

lib/mayaUsd/utils/utilSerialization.cpp
```diff
@@ -97,6 +97,10 @@ bool saveRootLayer(SdfLayerRefPtr layer, const std::string& proxy)
     fp = TfStringReplace(fp, "\\", "/");
 #endif
 
+    if (savePathAsRelative) {
+        fp = UsdMayaUtilFileSystem::getPathRelativeToMayaSceneFile(fp);
+    }
```
pierrebai-adsk, line 102: As mentioned before, getPathRelativeToMayaSceneFile currently can return an empty path, so we should protect against it here or within that function.

lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -370,7 +375,14 @@ void LayerTreeItem::saveAnonymousLayer()
 
             // now replace the layer in the parent
             if (isRootLayer()) {
-                sessionState->rootLayerPathChanged(fileName);
+                const bool makePathRelative
+                    = MGlobal::optionVarExists(MAKE_PATH_RELATIVE_TO_SCENE_FILE)
+                    && MGlobal::optionVarIntValue(MAKE_PATH_RELATIVE_TO_SCENE_FILE);
+
```
pierrebai-adsk, line 381: I think we may need to verify this in more places in the future so it may be good to provide a helper function to retrieve the option. (It would also make the code  support the possibility that we have a prefs or a UI for the option in the future.)

lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -370,7 +375,14 @@ void LayerTreeItem::saveAnonymousLayer()
...
+                sessionState->rootLayerPathChanged(
+                    makePathRelative
+                        ? UsdMayaUtilFileSystem::getPathRelativeToMayaSceneFile(fileName)
+                        : fileName);
```
pierrebai-adsk, line 385: Same comment as before, we need to handle the cases where we can't make the path relative even when requested.

lib/usd/ui/layerEditor/saveLayersDialog.cpp
```diff
@@ -165,26 +167,29 @@ SaveLayerPathRow::SaveLayerPathRow(
...
 }
 
 void SaveLayerPathRow::onOpenBrowser()
 {
     std::string fileName;
     if (SaveLayersDialog::saveLayerFilePathUI(fileName)) {
-        setAbsolutePath(fileName);
+        if (MGlobal::optionVarExists(MAKE_PATH_RELATIVE_TO_SCENE_FILE)
+            && MGlobal::optionVarIntValue(MAKE_PATH_RELATIVE_TO_SCENE_FILE)) {
+            fileName = UsdMayaUtilFileSystem::getPathRelativeToMayaSceneFile(fileName);
```
pierrebai-adsk, line 185: As suggested before, moving this into a common helper function would be preferable.

plugin/adsk/scripts/mayaUsd_USDRootFileRelative.py
```diff
@@ -0,0 +1,57 @@
...
+        haveSceneFile = cmds.file(q=True, exists=True)
+        cmds.checkBox(cls.kMakePathRelativeCheckBox, edit=True, enable=haveSceneFile)
+
+    @classmethod
+    def uiCommit(cls, parentLayout, selectedFile):
+        cmds.setParent(parentLayout)
+
+        # Get the current checkbox state and save to optionVar.
+        relative = cmds.checkBox(cls.kMakePathRelativeCheckBox, query=True, value=True)
+        cmds.optionVar(iv=('mayaUsd_MakePathRelativeToSceneFile', relative))
```
pierrebai-adsk, line 57: I'm not the most knowledgeable about MEL / Pythoin /Maya UI, so I'll assume all this is correct :)

pierrebai-adsk (changes requested):
As pointed out, we need to handle the cases where the path failed to be made relative.

lib/mayaUsd/utils/utilFileSystem.cpp
pierrebai-adsk: See: https://en.cppreference.com/w/cpp/filesystem/path/lexically_normal

lib/mayaUsd/utils/utilFileSystem.h
```diff
@@ -49,6 +49,17 @@ std::string getMayaSceneFileDir();
 MAYAUSD_CORE_PUBLIC
 std::string getMayaWorkspaceScenesDir();
 
+/*! \brief takes in an absolute file path and returns the path relative to maya scene file
+ */
+MAYAUSD_CORE_PUBLIC
+std::string getPathRelativeToMayaSceneFile(const std::string& fileName);
+
+/*! \brief returns the flag specifying whether Usd file paths should be saevd as relative to Maya
+ * scene file
```
pierrebai-adsk, line 58: Small typo: saevd -> saved. Not necessary to fix and redo a round-trip of PF + review just for this.

pierrebai-adsk (approved):
Looks good to me!

lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -117,6 +117,30 @@ std::string UsdMayaUtilFileSystem::getMayaSceneFileDir()
...
+{
+    ghc::filesystem::path absolutePath(fileName);
+    ghc::filesystem::path basePath(getMayaSceneFileDir());
+    ghc::filesystem::path relativePath = absolutePath.lexically_relative(basePath);
+
+    if (relativePath.empty()) {
+        TF_RUNTIME_ERROR(
+            "File name (%s) cannot be resolved as relative to the Maya scene file, using the "
+            "absolute path.",
+            fileName.c_str());
```
seando-adsk, line 130: runtime_error seems pretty strong in this case. You've detected the problem and are returning a good value. An error would make it appear as if the user did something wrong, but there is nothing they can do about it here. So I think a warning is sufficient.

lib/mayaUsd/utils/utilFileSystem.h
```diff
@@ -49,6 +49,17 @@ std::string getMayaSceneFileDir();
 MAYAUSD_CORE_PUBLIC
 std::string getMayaWorkspaceScenesDir();
 
+/*! \brief takes in an absolute file path and returns the path relative to maya scene file
```
seando-adsk, line 52: Should this comment mention that when there is no scene file, the absolute (input) path will be returned.

seando-adsk (changes requested):
Just a couple of small minor points to look at.

vlasovi:
@seando-adsk would it work for you if we merge this one as is, and I implement your suggestions in the next PR that I already have waiting in line

seando-adsk (approved):
Please fix these issues in your next PR for the relative loading.

PR #2840, opened 2023-01-24
MAYA-126515 clean options of materials scope name

pierrebai-adsk:
- The materials scope name is only modifiable with the environment variable.
- OTOH, the options used for duplicate, merge and cache to USD are all saved in options vars.
- This resulted in that the environment variable value was saved in options vars and would persists even when the env var was no longer set.
- This changes cleans up the option vars of the value that were set by the env var.
- It is a bit unfortunate that the materials scope choice was designed this way but now we are too close to release to re-design it.

### DIFF
--- a/lib/mayaUsd/resources/scripts/mayaUsdOptions.py
+++ b/lib/mayaUsd/resources/scripts/mayaUsdOptions.py
@@ -36,6 +36,24 @@ def setAnimateOption(nodeName, textOptions):
     else:
         return textOptions + ';' + animateOption
 
+def _cleanupOptionsText(text):
+    """
+    The saved options variable can have been polluted by forced value from
+    environment variables saved in previous run of Maya. If these forced
+    values are no longer forced, becaue they were saved, they would still
+    be forced... so, since all code paths that use MayaUSD export-like options,
+    like duplicate-to-USD, merge-to-USD, cache-to-USD, go through here,
+    we do the cleanup here.
+
+    Note that this kind of cleanup is only correct to do for options that have
+    no UI at all and are intended to only be controlled by such an environment
+    variable. Currently, only "materialsScopeName" is set like this, through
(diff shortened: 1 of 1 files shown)


vlasovi (dismissed):

lib/mayaUsd/resources/scripts/mayaUsdOptions.py
```diff
@@ -36,18 +36,36 @@ def setAnimateOption(nodeName, textOptions):
...
+    environment variables saved in previous run of Maya. If these forced
+    values are no longer forced, becaue they were saved, they would still
+    be forced... so, since all code paths that use MayaUSD export-like options,
+    like duplicate-to-USD, merge-to-USD, cache-to-USD, go through here,
+    we do the cleanup here.
+
+    Note that this kind of cleanup is only correct to do for options that have
+    no UI at all and are intended to only be controlled by such an environment
+    variable. Currently, only "materialsScopeName" is set like this, through
+    the env var "MAYAUSD_MATERIALS_SCOPE_NAME".
```
seando-adsk, line 51:
```suggestion
    The saved option variables can have been polluted by forced values from
    environment variables saved in a previous run of Maya. If these forced
    values are no longer forced, because they were saved, they would still
    be forced. Since all code paths that use MayaUSD export-like options like
    duplicate-to-USD, merge-to-USD and cache-to-USD go through here
    we do the cleanup here.

    Note: this kind of cleanup is only correct to do for options that have
    no UI at all and are intended to only be controlled by such an environment
    variable. Currently, only "materialsScopeName" is set like this, through
    the env var "MAYAUSD_MATERIALS_SCOPE_NAME".
```

seando-adsk (dismissed):
If you cleanup the comment text, there is no need to re-run the preflight.

pierrebai-adsk:
OK, accepted your suggested changes, thanks!

seando-adsk (approved):

PR #2845, opened 2023-01-25
Fixes an issue on layer editor save dialog, where an invalid extension could cause a crash. The fallback is .usd.

AramAzhari-adsk:
When an invalid extension string is provided, USD fails to relocate the file when trying update the cached stage layers and will return a TfNullPtr which results in crash. The fix is to check and update the file path prior to passing to layer export and load operations.


pierrebai-adsk:
Given your description of the problem, the fact that it caused crashes, while the code you fixed did not use the invalid null pointer, makes me think that it initially crashed in the caller? I see that it was called in layerManager.cpp, which passed the returned layer pointer to SetEditTarget(), which is where I assume it eventually crashed.

So I'd also add a check for null in layerManager.cpp (possibly with an error log). While you fixed one cause why the returned layer an be null, there might be others we don't know about. I think it would be nice to make the caller handle the null case.

pierrebai-adsk (changes requested):
As I wrote, I believe we should fix the caller too.

pierrebai-adsk (approved):

seando-adsk (approved):

PR #2847, opened 2023-01-26
Fixes an issue on layer editor when trying to save a root layer directly (by right clicking on the layer and Save As.), where an invalid extension could cause a crash. The fallback is .usd.

AramAzhari-adsk:
Fixes an issue on layer editor when trying to save a root layer directly (by right clicking on the layer and Save As.), where an invalid extension could cause a crash. The fallback is .usd.

It also blocks the path that will result in the crash for any other scenario where the file cannot be opened after it was exported. The user is then displayed a message dialog indicating the layer saving was not successful.

Similar PR: https://github.com/Autodesk/maya-usd/pull/2845

### DIFF
--- a/lib/usd/ui/layerEditor/layerTreeItem.cpp
+++ b/lib/usd/ui/layerEditor/layerTreeItem.cpp
@@ -365,33 +365,24 @@ void LayerTreeItem::saveAnonymousLayer()
     std::string fileName;
     if (sessionState->saveLayerUI(nullptr, &fileName)) {
 
-        std::string        filePath(fileName);
-        const std::string& extension = SdfFileFormat::GetFileExtension(filePath);
-        const std::string  defaultExt(UsdMayaTranslatorTokens->UsdFileExtensionDefault.GetText());
-        const std::string  usdCrateExt(UsdMayaTranslatorTokens->UsdFileExtensionCrate.GetText());
-        const std::string  usdASCIIExt(UsdMayaTranslatorTokens->UsdFileExtensionASCII.GetText());
-        if (extension != defaultExt && extension != usdCrateExt && extension != usdASCIIExt) {
-            filePath.append(".");
-            filePath.append(defaultExt.c_str());
-        }
+        MayaUsd::utils::ensureUSDFileExtension(fileName);
 
-        // the path we has is an absolute path
+        // the path we have is an absolute path
         const QString dialogTitle = StringResources::getAsQString(StringResources::kSaveLayer);
         std::string   formatTag = MayaUsd::utils::usdFormatArgOption();
-        if (saveSubLayer(dialogTitle, parentLayerItem(), layer(), filePath, formatTag)) {
-            printf("USD Layer written to %s\n", filePath.c_str());
+        if (saveSubLayer(dialogTitle, parentLayerItem(), layer(), fileName, formatTag)) {
+            printf("USD Layer written to %s\n", fileName.c_str());
 
             // now replace the layer in the parent
             if (isRootLayer()) {
-                sessionState->rootLayerPathChanged(filePath);
                 sessionState->rootLayerPathChanged(
                     UsdMayaUtilFileSystem::requireUsdPathsRelativeToMayaSceneFile()
-                        ? UsdMayaUtilFileSystem::getPathRelativeToMayaSceneFile(filePath)
-                        : filePath);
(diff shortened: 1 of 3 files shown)


lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -363,22 +364,34 @@ void LayerTreeItem::saveAnonymousLayer()
...
+
+        std::string        filePath(fileName);
+        const std::string& extension = SdfFileFormat::GetFileExtension(filePath);
+        const std::string  defaultExt(UsdMayaTranslatorTokens->UsdFileExtensionDefault.GetText());
+        const std::string  usdCrateExt(UsdMayaTranslatorTokens->UsdFileExtensionCrate.GetText());
+        const std::string  usdASCIIExt(UsdMayaTranslatorTokens->UsdFileExtensionASCII.GetText());
+        if (extension != defaultExt && extension != usdCrateExt && extension != usdASCIIExt) {
+            filePath.append(".");
+            filePath.append(defaultExt.c_str());
+        }
```
pierrebai-adsk, line 376: If this is the same code as in the other review, maybe create a function, something like ensureUSDFileExtension() or similar.

lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -363,22 +364,34 @@ void LayerTreeItem::saveAnonymousLayer()
 
     std::string fileName;
     if (sessionState->saveLayerUI(nullptr, &fileName)) {
+
+        std::string        filePath(fileName);
```
pierrebai-adsk, line 368: Not sure why we need to create a new variable here. I think we could process fileName directly.

pierrebai-adsk (changes requested):
Almost there, just a few improvements that I think would be worthwhile.

lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -363,22 +364,34 @@ void LayerTreeItem::saveAnonymousLayer()
...
         std::string   formatTag = MayaUsd::utils::usdFormatArgOption();
-        if (saveSubLayer(dialogTitle, parentLayerItem(), layer(), fileName, formatTag)) {
-            printf("USD Layer written to %s\n", fileName.c_str());
+        if (saveSubLayer(dialogTitle, parentLayerItem(), layer(), filePath, formatTag)) {
+            printf("USD Layer written to %s\n", filePath.c_str());
 
             // now replace the layer in the parent
             if (isRootLayer()) {
+                sessionState->rootLayerPathChanged(filePath);
                 sessionState->rootLayerPathChanged(
```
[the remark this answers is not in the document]
AramAzhari-adsk: Good catch! Result of a poor merge.

lib/usd/ui/layerEditor/layerTreeItem.cpp
AramAzhari-adsk: done

lib/usd/ui/layerEditor/layerTreeItem.cpp
AramAzhari-adsk: Removed the extra variable.

pierrebai-adsk (dismissed):

lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -10,6 +10,7 @@
 #include "warningDialogs.h"
 
 #include <mayaUsd/base/tokens.h>
+#include <mayaUsd/fileio/jobs/jobArgs.h>
```
vlasovi, line 13: Is this include necessary?

vlasovi (dismissed):
I think including "mayaUsd/fileio/jobs/jobArgs.h" is not necessary. Otherwise looks good

lib/usd/ui/layerEditor/layerTreeItem.cpp
AramAzhari-adsk: This is no longer necessary as the function now lives in utilSerialization. I'll remove it.

vlasovi (approved):

vlasovi (approved):

seando-adsk (approved):

PR #2848, opened 2023-01-26
MAYA-127353 - As a user, I'd like to load my USD root file as relative to the scene file

vlasovi:


pierrebai-adsk (approved):
Look good to me.

seando-adsk (approved):

PR #2852, opened 2023-01-30
MAYA-127483 - Deleting a connection does not delete source and destination properties.

alicedegirolamo:
# Description

This PR solves the bug reported in Jira ticket [MAYA-127483](https://jira.autodesk.com/browse/MAYA-127483): when we delete a connection, we should check if it is possible also to delete the source and destination properties.

We checked only the destination and not the source. 

We can delete a property when:
- it has no value,
- there are no connections

This last condition should be explored in all its complexity. 
If the source/destination attribute is on a Compound, we should check also the connection to/from the inner children and also consider, in this case, that a destination is also a source and vice versa. 

## Type of change

# How Has This Been Tested?

I've added extensive tests in order to cover all the cases of property removal:
- connections between nodes (not compound) and parent,
- connections between compound and parent,
- connections between compound and inner child.


lib/mayaUsd/ufe/UsdUndoConnectionCommands.cpp
```diff
@@ -284,6 +284,139 @@ UsdUndoDeleteConnectionCommand::Ptr UsdUndoDeleteConnectionCommand::create(
...
+    const auto prim = srcAttr.GetPrim();
+
+    if (!prim) {
+        return false;
+    }
+
+    PXR_NS::UsdShadeNodeGraph ngPrim(prim);
+
+    if (!ngPrim) {
+        return true;
```
alicedegirolamo, line 378: the functions `canRemoveSrcProperty `and `canRemoveDstProperty `are slightly different: if the prim is not a `PXR_NS::UsdShadeNodeGraph` and the attribute is a destination, we can return true immediately, instead if the attribute were a source, we have to do more checks.

lib/mayaUsd/ufe/UsdUndoConnectionCommands.cpp
```diff
@@ -284,6 +284,139 @@ UsdUndoDeleteConnectionCommand::Ptr UsdUndoDeleteConnectionCommand::create(
...
+        }
+    }
+
+    PXR_NS::UsdShadeNodeGraph ngPrim(prim);
+
+    if (!ngPrim) {
+        return true;
+    }
+
+    // Do not remove if there is a connection with a child prim.
```
JGamache-autodesk, line 341: There is a very useful workflow where you drag the first port of the type you want on the boundary, and connect it to the input node as many times as desired, then delete the connections.
In this scenario, the user expects the ports on the input node to remain so he can do further work on them. He can always delete the ones he does not want anymore.
That would mean returning false here even if there are no connections left. Same thing on the reverse function.

JGamache-autodesk (changes requested):

JGamache-autodesk:
The test file `test/testSamples/properties/properties.usd` should be saved as ASCII. Makes it more friendly for source control and reviewers.

alicedegirolamo:
> The test file `test/testSamples/properties/properties.usd` was converted to ASCII. Makes it more friendly for source control and reviewers.

Thank you for having corrected it.

lib/mayaUsd/ufe/UsdUndoConnectionCommands.cpp
alicedegirolamo: Thank you. I didn't know about this workflow. 
I've updated the solution also with the test.
 
I've also reintroduced the condition for the `UsdShadeMaterial` which we should keep ( it is also required by some tests).

lib/mayaUsd/ufe/UsdUndoConnectionCommands.cpp
```diff
@@ -284,6 +284,104 @@ UsdUndoDeleteConnectionCommand::Ptr UsdUndoDeleteConnectionCommand::create(
     return std::make_shared<UsdUndoDeleteConnectionCommand>(srcAttr, dstAttr);
 }
 
+bool canRemoveSrcProperty(const PXR_NS::UsdAttribute& srcAttr)
```
seando-adsk, line 287: I think this helper function (and the one below) should be in the unnamed namespace near the top of this file (with the other helper functions).

seando-adsk (changes requested):
Looks good from me, except one minor change requested to move helper functions.

lib/mayaUsd/ufe/UsdUndoConnectionCommands.cpp
alicedegirolamo: Could I move them to `mayaUsd/ufe/Utils.h`? I need them also for other issues I'm working on.

seando-adsk: Yes that is a good idea.

JGamache-autodesk (approved):

lib/mayaUsd/ufe/UsdUndoConnectionCommands.cpp
alicedegirolamo: I moved also the `isConnected` function since it is used by `canRemoveSrcProperty` and `canRemoveDstProperty`, I also need it for the issues I'm working on.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -447,6 +447,121 @@ TfTokenVector getProxyShapePurposes(const Ufe::Path& path)
...
+
+    PXR_NS::UsdShadeNodeGraph ngPrim(prim);
+
+    if (!ngPrim) {
+        return true;
+    }
+
+    UsdShadeMaterial asMaterial(prim);
+    if (asMaterial) {
+        const auto baseNameAndType = PXR_NS::UsdShadeUtils::GetBaseNameAndType(dstAttr.GetName());
```
alicedegirolamo, line 552: Here we should use `PXR_NS::UsdShadeUtils::GetBaseNameAndType` instead of UsdGeomPrimvar::GetBaseName(), in fact, this last function returns the last component of the return value of `GetName()`. 
The bug in [LOOKDEVX-1144](https://jira.autodesk.com/browse/LOOKDEVX-1144) was caused by this: the attribute `outputs:mtlx:surface` should have base name `mtlx:surface`, however `UsdGeomPrimvar::GetBaseName()` returns `surface` and hence the property was incorrectly removed.

JGamache-autodesk (dismissed):

seando-adsk (dismissed):

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -447,6 +447,120 @@ TfTokenVector getProxyShapePurposes(const Ufe::Path& path)
...
+
+    PXR_NS::UsdShadeNodeGraph ngPrim(prim);
+
+    if (!ngPrim) {
+        return true;
+    }
+
+    UsdShadeMaterial asMaterial(prim);
+    if (asMaterial) {
+        const TfToken baseName = dstAttr.GetBaseName();
```
alicedegirolamo, line 552: I reverted the change made on `GetBaseName` after your explanation @JGamache-autodesk . 
e.g. if the `outputs:mtlx:surface` is not authored, it is removed with the connection.

seando-adsk (approved):

JGamache-autodesk (approved):

PR #2858, opened 2023-01-31
Adding a MaterialX v1.38.5 update for closures

JGamache-autodesk:
Shading graphs containing closures would fail to work with MaterialX 1.38.5.

### DIFF
--- a/lib/mayaUsd/render/MaterialXGenOgsXml/GlslFragmentGenerator.cpp
+++ b/lib/mayaUsd/render/MaterialXGenOgsXml/GlslFragmentGenerator.cpp
@@ -386,9 +386,7 @@ ShaderPtr GlslFragmentGenerator::generate(
         }
 
         // Add all function calls (varies greatly by MaterialX version)
-#if MX_COMBINED_VERSION < 13803
-        emitFunctionCalls(graph, context, pixelStage);
-#else
+#if MX_COMBINED_VERSION >= 13805
         // Surface shaders need special handling.
         if (graph.hasClassification(
                 ShaderNode::Classification::SHADER | ShaderNode::Classification::SURFACE)) {
@@ -396,15 +394,6 @@ ShaderPtr GlslFragmentGenerator::generate(
             // closure/shader nodes and need to be emitted first.
             emitFunctionCalls(graph, context, pixelStage, ShaderNode::Classification::TEXTURE);
 
-#if MX_COMBINED_VERSION < 13805
-            // Emit function calls for all surface shader nodes.
-            // These will internally emit their closure function calls.
-            emitFunctionCalls(
-                graph,
-                context,
-                pixelStage,
-                ShaderNode::Classification::SHADER | ShaderNode::Classification::SURFACE);
-#else
             // Emit function calls for "root" closure/shader nodes.
             // These will internally emit function calls for any dependent closure nodes upstream.
             for (ShaderGraphOutputSocket* socket : graph.getOutputSockets()) {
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/render/MaterialXGenOgsXml/GlslFragmentGenerator.cpp
```diff
@@ -385,8 +385,45 @@ ShaderPtr GlslFragmentGenerator::generate(
                 true);
         }
 
-        // Add all function calls
+        // Add all function calls (varies greatly by MaterialX version)
```
JGamache-autodesk, line 388: Nothing complex here. We have a local version of the GlslGenerator that closely follows the original one. There were some missing updates that are visible by comparing the same lines for MaterialX [v1.38.2](https://github.com/AcademySoftwareFoundation/MaterialX/blob/v1.38.2/source/MaterialXGenGlsl/GlslShaderGenerator.cpp#L594), [1.38.3](https://github.com/AcademySoftwareFoundation/MaterialX/blob/v1.38.3/source/MaterialXGenGlsl/GlslShaderGenerator.cpp#L603), and [1.38.5](https://github.com/AcademySoftwareFoundation/MaterialX/blob/v1.38.5/source/MaterialXGenGlsl/GlslShaderGenerator.cpp#L639).

These versions were copied over and bracketed with the right conditional MX version compiler checks.

lib/mayaUsd/render/MaterialXGenOgsXml/GlslFragmentGenerator.cpp
```diff
@@ -385,8 +385,45 @@ ShaderPtr GlslFragmentGenerator::generate(
...
         emitFunctionCalls(graph, context, pixelStage);
+#else
+        // Surface shaders need special handling.
+        if (graph.hasClassification(
+                ShaderNode::Classification::SHADER | ShaderNode::Classification::SURFACE)) {
+            // Emit all texturing nodes. These are inputs to any
+            // closure/shader nodes and need to be emitted first.
+            emitFunctionCalls(graph, context, pixelStage, ShaderNode::Classification::TEXTURE);
+
+#if MX_COMBINED_VERSION < 13805
```
frohnej-adsk, line 399: Having the nested `#if` and `#else` block makes it a bit difficult to figure out, what code will be run for what version. Maybe it would be more readable to have completely separate blocks for each version:
```C++
#if MX_COMBINED_VERSION < 13803
// v1.38.2 code
// ...

#elif MX_COMBINED_VERSION < 13805
// 1.38.3 code
// ...

#else
// 1.38.5 code
// ...

#endif
```

This would introduce some duplicated code, but I don't think it's a big problem because the older versions will probably remain unchanged. Feel free to leave it as is if you think it is worth it to avoid the code duplication.

frohnej-adsk (approved):

seando-adsk (approved):

lib/mayaUsd/render/MaterialXGenOgsXml/GlslFragmentGenerator.cpp
JGamache-autodesk: Agreed. I reformatted the code by blocks, and put the most current at the top to improve readability.

JGamache-autodesk:
@seando Ready for merge when they reopen.

seando-adsk (approved):

lib/mayaUsd/render/MaterialXGenOgsXml/GlslFragmentGenerator.cpp
frohnej-adsk: Thanks!

Issue #2756, opened 2022-11-30
[MAYA-127103] Importer Errors when importing animated meshes that are skinned


dgovil:
**Describe the bug**
If you import a USD file that has a mesh that is both skinned and vertex animated, the importer will error with:

`# Error: Connection not made: 'skinCluster_pCylinder2.outputGeometry[0]' -> 'pCylinder1Shape.inMesh'.  Destination attribute must be writable.`

Unfortunately, this is an issue when a Mesh has even just a single static time sample as in the example below

**NOTE**: This error only shows up if you import using the `mayaUSDImport` command. If you use the file import command, it will not show up for some reason. Not sure if it's just swallowing the error or if it's somehow taking a different path.

I think both vertex animation and skinning should be able to coexist. However the issue is that the Maya USD importer imports vertex animation as blendshapes which then takes up the inMesh connection. Perhaps if it's skinned, it should add another blendshape for the skinning to apply to, and then have the blendshapes combine into the final inMesh?

**Steps to reproduce**
1. Import the file with `readAnimData` set to True

`cmds.mayaUSDImport(f="/Users/dhruvgovil/Documents/maya/projects/default/scenes/foo.usda", readAnimData=True)`

**Expected behavior**
I'm honestly not sure what the right behaviour here should be:
* Warn the user and take just the skinning (ignoring the vertex animation)
* Error when it detects both with a clear error?
* Try and import both?
* At the very least I think the vertex animation importer should detect when something is a single keyframe, and treat it as static if it can.

**Attachments**
[foo.usda.zip](https://github.com/Autodesk/maya-usd/files/10119130/foo.usda.zip)

**Specs (if applicable):**
 - OS & version : macOS 13
 - Compiler & version : clang 14
 - Maya version 2023.2
 - Maya USD commit SHA 8b7b598b3234918feb4cf3fa5055a66422caea85
 - Pixar USD commit SHA release at 0c7b9a95f155c221ff7df9270a39a52e3b23af8b

neilh-adsk:
I ran into a few other smaller inconsistencies with vertex animation and skinning locally.  I didn't manage to isolate those problems, but just to say it an area that is growing in attention. I'll bring this into the backlog and track it with any other issues we are seeing.

santosd:
@neilh-adsk should we log an issue for this internally?

neilh-adsk:
Yes, please log this and I'll flag it followup

santosd:
[...]

PR #2860, opened 2023-02-02
Import meshes with single samples as static geometry

dgovil:
MayaUSD defaults to importing any animated mesh as blendshape data. Unfortunately, without staticSingleSimples enabled on export, meshes may end up with single time samples on their primvars. This leads to exported files not being able to be roundtripped, and will break skinning as a result

This PR changes the logic so that a single time sample is treated as static instead.

Addresses https://github.com/Autodesk/maya-usd/issues/2756


pierrebai-adsk (approved):
LGTM

PR #2861, opened 2023-02-02
MAYA-127459 - Create an autotest for USD root layer having a relative path

vlasovi:


vlasovi:
@seando-adsk the preflight failed on autotest testUsdExportPackage, which is a known random failure. We can consider the preflight passed.

test/lib/testMayaUsdCreateStageCommands.py
```diff
@@ -72,3 +77,14 @@ def testCreateStageFromFile(self):
...
+        # Now switch to having USD stage paths as relative to Maya scene file
+        cmds.optionVar(iv=('mayaUsd_MakePathRelativeToSceneFile', 1))
+        
+        # Create the same stage and verify that now it's open as relative
+        shapeNodeRel = mel.eval('mayaUsd_createStageFromFilePath(\"'+ballFilePath+'\")')
+        filePathAttrRel = cmds.getAttr(shapeNodeRel+'.filePath')
+        self.assertEqual(filePathAttrRel, 'top_layer.usda')
+        
+        # Restore mayaUsd_MakePathRelativeToSceneFile
+        cmds.optionVar(iv=('mayaUsd_MakePathRelativeToSceneFile', 0))
```
seando-adsk, line 90: FYI, you shouldn't have to do this because every test will run with its own set of newly created prefs.

seando-adsk (approved):
Thanks for adding the test.

test/lib/testMayaUsdCreateStageCommands.py
vlasovi: Couple times in the past I stumbled upon problems with prefs that I thought would be reset to default but they werent't. Things like default background color and display layer color. Prefs set during one run of the test would persist at the next run. So now I prefer to be sure

PR #2864, opened 2023-02-03
MAYA-127478 fix bulk save of root layer filename

pierrebai-adsk:
When doing a bulk-save of layers when saving a Maya scene for the first time, the multiple-parts of the saving process could end-up using stale data. That is because there is a first batch-save delegate (customizable callback) that can do a partial save, then a fallback save is done for any layer that did not get saved.

When a stage has just been created, it contains an anonymous root layer which gets replaced by a non-anonymous by the batch save delegate. When the root layer is swapped, the stage proxy shape is updated by setting its stage name attribute.  Unfortunately, the USD stage object was being cached and the attribute change was thus not use immediately. Thus the 2nd part fallback-save thought that the root layer was still anonymous.

The fix is to refresh the cached USD stages by pulling again on the proxy stage node so that it updates itself when its attributes have been changed.

### DIFF
--- a/lib/mayaUsd/nodes/layerManager.cpp
+++ b/lib/mayaUsd/nodes/layerManager.cpp
@@ -248,6 +248,7 @@ private:
     void saveUsdLayerToMayaFile(SdfLayerRefPtr layer, bool asAnonymous);
     void clearProxies();
     bool hasDirtyLayer() const;
+    void refreshProxiesToSave();
 
     std::map<std::string, SdfLayerRefPtr> _idToLayer;
     TfNotice::Key                         _onStageSetKey;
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -555,6 +577,12 @@ bool LayerDatabase::saveUsd(bool isExport)
             return true;
         }
 
+        // After the potentially partial save, we need to refresh the stages
+        // to be saved because the saving might have modified the proxy shape
+        // attributes and we need to re-evaluate these nodes so that the stages
+        // are re-created with the new attribute values if needed.
+        refreshProxiesToSave();
```
seando-adsk, line 584: By doing this pull here does it cause any notifications to be sent and could that potentially mess up the clients since we've only done a partial save?

seando-adsk (approved):

lib/mayaUsd/nodes/layerManager.cpp
pierrebai-adsk: At the point this is called, we are in a pre-save callback AFAIK, so there should be no issue.

PR #2866, opened 2023-02-06
MAYA-126541 - Create a more extensive unit test for display layer support during USD to USD duplication 

vlasovi:
Additional testing for USD to USD duplication with display layers. This case covers the bug found by QA during the original feature testing


seando-adsk (changes requested):
Could you please add a unit test as requested by PierreB in the PR #2747

vlasovi:
@seando-adsk The test that I added verfies the fix introduced in the PR https://github.com/Autodesk/maya-usd/pull/2747 , which is "duplicating from USD to USD a node that is not in the stage root", and I think that's what @pierrebai-adsk wanted me to verify. I'm not sure what else I need to do here.

seando-adsk (commented):
Ah sorry about that. I missed that from the new test. Maybe add a comment stating that. If you do, don't worry about re-running the preflight.

seando-adsk (approved):

Issue #2859, opened 2023-02-01
Windows Python 3.8+ os.add_dll_directory search order is defined as unspecified


tbttfox:
**Describe the bug**
Random users across our studio would get this dll error when running `from pxr import Tf`. Unfortunately, this was machine specific as far as we could tell. Identical setups on different machines would give different results.

```
# #     from pxr import Tf
# #   File "c:\Program Files\Autodesk\MayaUSD\Maya2023\0.16.0\mayausd\USD\lib\python\pxr\Tf\__init__.py", line 163, in <module>
# #     PreparePythonModule()
# #   File "c:\Program Files\Autodesk\MayaUSD\Maya2023\0.16.0\mayausd\USD\lib\python\pxr\Tf\__init__.py", line 88, in PreparePythonModule
# #     module = importlib.import_module(
# #   File "C:\Program Files\Autodesk\Maya2023\Python\lib\importlib\__init__.py", line 127, in import_module
# #     return _bootstrap._gcd_import(name[level:], package, level)
# # ImportError: DLL load failed while importing _tf: The specified procedure could not be found.
```

**Steps to reproduce**
I don't think it's possible to reproduce. See Additional Context

**Expected behavior**
No DLL Error

**Specs (if applicable):**
 - Windows 10
 - Maya 2023.3
 - Mayausd 0.20.0 and 0.16.0

**Additional context**

We found that USD looks at the system's `PATH` variable to load its dll's on python 3.8+ here:
https://github.com/PixarAnimationStudios/USD/blob/release/pxr/base/tf/__init__.py#L47

So we started chasing the DLL's, and the breaking machines were loading Bifrost's usd dll's before maya-usd's. This seemed very strange because maya-usd came before Bifrost in the `PATH`, and it wasn't happening on other machines.

On a hunch, we checked the cpython source for `os.add_dll_directory`, which ends up calling this function:
https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-adddlldirectory

And, oh boy, is there a doozy in those docs:
> If AddDllDirectory is used to add more than one directory to the process DLL search path, the order in which those directories are searched is unspecified.

That says to me, that the paths added to the dll search are not necessarily searched in order, and without knowing the internal Windows DLL search mechanism, we apparently *can't* know that order.

**Proposed solution**
[...]

PR #2867, opened 2023-02-06
Add PXR_USD_WINDOWS_DLL_PATH to all of the windows module templates.

tbttfox:
Discussed in https://github.com/Autodesk/maya-usd/issues/2859

Paths added to the dll search list are not necessarily searched in order, so this change makes sure that PXR_USD_WINDOWS_DLL_PATH is defined in any .mod files for Windows so we don't get .dll clashes.

I changed all 3 windows mod files, because (I'm guessing) that Pixar and Animal Logic aren't using Windows, and those are there only for completeness. Plus anybody that looks to those files as an example of an alternative setup should probably have that variable set.

### DIFF
--- a/modules/alUSD_Win.mod.template
+++ b/modules/alUSD_Win.mod.template
@@ -2,5 +2,6 @@
 plug-ins: plugin
 PYTHONPATH+:=lib/python
 PATH+:=lib
+PXR_USD_WINDOWS_DLL_PATH+:=lib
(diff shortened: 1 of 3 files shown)


seando-adsk (approved):

seando-adsk:
No preflight needed as the .mod files are not used by the tests. See #2931 for corresponding cmake test changes.

PR #2873, opened 2023-02-10
MAYA-127699 - Avoid diving into material when user selects "Show in LookdevX"

stefanminning-autodesk:
## Description
This PR fixes a bug where `getMaterials` would in certain cases return a shader instead of a material, which lead to inconsistent behaviour in LookdevX.

### DIFF
--- a/lib/mayaUsd/ufe/UsdMaterial.cpp
+++ b/lib/mayaUsd/ufe/UsdMaterial.cpp
@@ -23,51 +23,6 @@
 
 PXR_NAMESPACE_USING_DIRECTIVE
 
-namespace {
-
-PXR_NS::UsdPrim getAncestorMaterial(const PXR_NS::UsdPrim& prim)
-{
-    // Make sure we're returning a Material, which is the (grand)parent of a Shader or NodeGraph.
-
-    if (prim.GetTypeName() == "Material")
-    {
-        return prim;
-    }
-
-    auto iterPrim = prim;
-    while (iterPrim.GetTypeName() == "Shader" || iterPrim.GetTypeName() == "NodeGraph")
-    {
-        iterPrim = iterPrim.GetParent();
-    }
-    if (iterPrim)
-    {
-        return iterPrim;
-    }
-
-    return PXR_NS::UsdPrim();
-}
-
-std::vector<PXR_NS::UsdPrim> getConnectedShaders(const PXR_NS::UsdShadeOutput& port)
-{
-    // Dig down across NodeGraph boundaries until a surface shader is found.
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/UsdMaterial.cpp
```diff
@@ -26,6 +25,28 @@ PXR_NAMESPACE_USING_DIRECTIVE
 
 namespace {
 
+PXR_NS::UsdPrim getAncestorMaterial(const PXR_NS::UsdPrim& prim)
```
JGamache-autodesk, line 28: This is a bug that can be fixed exclusively by removing code. You should not be adding this function.

lib/mayaUsd/ufe/UsdMaterial.cpp
```diff
@@ -26,6 +25,28 @@ PXR_NAMESPACE_USING_DIRECTIVE
...
+    }
+    if (iterPrim)
+    {
+        return iterPrim;
+    }
+
+    return PXR_NS::UsdPrim();
+}
+
 std::vector<PXR_NS::UsdPrim> getConnectedShaders(const PXR_NS::UsdShadeOutput& port)
```
JGamache-autodesk, line 50: This function can be removed as well. It is not longer necessary.

lib/mayaUsd/ufe/UsdMaterial.cpp
```diff
@@ -85,7 +106,7 @@ std::vector<Ufe::SceneItem::Ptr> UsdMaterial::getMaterials() const
     if (material) {
```
JGamache-autodesk, line 106: You have the material right here. Add it to the results, and remove all the traversal of the `GetSurfaceOutputs()` since we no longer want to dig inside the material.

lib/mayaUsd/ufe/UsdMaterial.cpp
```diff
@@ -98,14 +119,20 @@ std::vector<Ufe::SceneItem::Ptr> UsdMaterial::getMaterials() const
         if (material) {
             for (const auto& output : material.GetSurfaceOutputs()) {
```
JGamache-autodesk, line 120: Same thing here. You already have the material an do not need to traverse any surface output.

JGamache-autodesk (changes requested):

JGamache-autodesk (approved):

seando-adsk:
@stefanminning-autodesk You need to run a preflight on the changes first before it can be merged. Once the preflight passes you can add back the "ready-for-merge" label. In order to run a preflight set the "Assignees" to "ecp-maya-devops-adsk"

PR #2875, opened 2023-02-13
MAYA-125727 - Implement display layer's template and reference modes for instanced geometry

vlasovi:


lib/mayaUsd/render/vp2RenderDelegate/draw_item.h
```diff
@@ -125,6 +125,16 @@ class HdVP2DrawItem final : public HdDrawItem
         kSelectionHighlight = kRegular << 1 //!< Selection highlight.
     };
 
+    //! Flags specifying mod properties
+    enum ModFlags
```
pierrebai-adsk, line 129: Just curious: mod stands for modifier?

pierrebai-adsk (approved):

lib/mayaUsd/render/vp2RenderDelegate/draw_item.h
vlasovi: Mod is a draw item which serves as a modification of the original/basic draw item.

PR #2876, opened 2023-02-14
MAYA-125340 improve error messages

pierrebai-adsk:
When an operation cannot proceed because it would need to modify multiple layers at the same time, use a different error message. Do not tell the user to switch layer.


vlasovi (dismissed):
Looks good

pierrebai-adsk:
Only failure is a single test on Maya 2022 where the test checks the temp directory contains no USD file, but the temp dir actually contains one... this is a known random failure.

vlasovi (approved):

PR #2882, opened 2023-02-16
Releases the memory a data handle is holding onto when exporting meshes.

AramAzhari-adsk:
When exporting the entire scene, a data handle is used to create objects for writing meshes. When the user tries to export all again, the previous handle is not cleaned up and this results in a memory leak. The fix is to destruct the handle after use so that the memory does not increase after repetition.

### DIFF
--- a/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
@@ -1443,6 +1443,7 @@ MStatus UsdMayaMeshWriteUtils::exportComponentTags(UsdGeomMesh& primSchema, MObj
             }
         }
     }
+    outShp.destructHandle(geomDataHandle);
 
     return status;
 }


lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -1443,6 +1443,7 @@ MStatus UsdMayaMeshWriteUtils::exportComponentTags(UsdGeomMesh& primSchema, MObj
             }
         }
     }
+    outShp.destructHandle(geomDataHandle);
```
seando-adsk, line 1446: I did a little searching and found there is a helper class and function for automatically dealing with this to avoid the leak (when not calling destructHandle). Please see ```UsdMayaUtil::GetPlugDataHandle()```

lib/mayaUsd/fileio/utils/meshWriteUtils.cpp
```diff
@@ -1408,7 +1408,8 @@ MStatus UsdMayaMeshWriteUtils::exportComponentTags(UsdGeomMesh& primSchema, MObj
     MPlug outShp = depNodeFn.findPlug("outMesh", &status);
     CHECK_MSTATUS_AND_RETURN_IT(status);
 
-    MDataHandle geomDataHandle = outShp.asMDataHandle();
+    auto        outShpHolder = UsdMayaUtil::GetPlugDataHandle(outShp);
+    MDataHandle geomDataHandle = outShpHolder->GetDataHandle();
     MObject     geomObj = geomDataHandle.data();
```
seando-adsk, line 1413: Is this making a needless copy to store in geomDataHandle?
```suggestion
    MObject     geomObj = outShpHolder->GetDataHandle().data();
```

seando-adsk (dismissed):

seando-adsk (approved):

PR #2884, opened 2023-02-16
MAYA-114188 save the target layer

pierrebai-adsk:
- When saving a Maya scene, the current target layer of USD stage get saved in an attribute on the proxy shape.
- When loading a scene, when creating the USD stage, set the target layer to the one that was saved.
- The target layer is persisted as a text attribute containing the USD layer identifier.
- Listen to edit target change to save the attribute when the target is set to avoid losing it if the proxy shape gets recomputed.
- Add a unit test,


pierrebai-adsk:
Only failure is testUsdExportPackage detecting undeletable files in temp. Why are we even checking the content of temp in that test? It keeps randomly failing due to that.

vlasovi (approved):
Looks good

PR #2888, opened 2023-02-17
MAYA-000000 Reduce testUsdExportPackage failures

pierrebai-adsk:
The test expects that the temp folder be devoid of certain files, but it sometimes happens that files matching the pattern existed before the test started and are undeletable. (Locked by an other process or owned by another user?)

Record the list of pre-existing files and ignore them.


vlasovi (approved):
Nice. Thanks Pierre for doing that.

seando-adsk:
@pierrebai-adsk Thanks for this fix. Hopefully we've see the last of the random failure of this test.

PR #2890, opened 2023-02-17
MAYA-127808 update edit target when saving an anonymous layer

pierrebai-adsk:
- Record from which stage the saved layer are from.
- Update the edit target to the new layer of the stage if the old layer was the edit target.
- Adjust save dialog callers to track and pass the necessary info.


vlasovi (approved):
Looks good to me

PR #2891, opened 2023-02-20
Correctly handle MaterialX filename inputs

JGamache-autodesk:
Replaced hardcoded lookup of `image` and `tiledimage` with code that can handle any node with a `filename` input. Tested with triplanar projection node.


vlasovi (approved):
Looks good

PR #2892, opened 2023-02-20
MAYA-106352 keep hierarchy window on top of import window

pierrebai-adsk:
- Add a flag to the UsdImportDialog command to specify the parent window name.
- Use that flag when invoking the command from the import panel.
- Use the flag in the command to find the parent window.

### DIFF
--- a/lib/usd/ui/importDialog/USDImportDialogCmd.cpp
+++ b/lib/usd/ui/importDialog/USDImportDialogCmd.cpp
@@ -219,8 +219,7 @@ MStatus USDImportDialogCmd::doIt(const MArgList& args)
             QApplication::setOverrideCursor(QCursor(Qt::WaitCursor));
 
             QWidget*      parentWindow = MQtUtil::mainWindow();
-            const MString parentWindowName
-                = parseTextArg(argData, kParentWindowFlag, "");
+            const MString parentWindowName = parseTextArg(argData, kParentWindowFlag, "");
             if (parentWindowName.length() > 0) {
                 QWidget* potentialParent = MQtUtil::findWindow(parentWindowName);
                 if (potentialParent) {
--- a/lib/usd/ui/importDialog/USDImportDialogCmd.cpp
+++ b/lib/usd/ui/importDialog/USDImportDialogCmd.cpp
@@ -65,7 +65,7 @@ constexpr auto kSwitchedVariantCountFlagLong = "-switchedVariantCount";
 constexpr auto kParentWindowFlag = "-pw";
 constexpr auto kParentWindowFlagLong = "-parentWindow";
 
-MString parseTextArg(const MArgDatabase& argData, const char* flag, const MString& defaultValue)
+MString parseTextArg(const MArgParser& argData, const char* flag, const MString& defaultValue)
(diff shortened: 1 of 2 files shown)


pierrebai-adsk:
Unit test testUsdExportPackage randomly failed again on a single Windows platform, all other platforms and tests passed.

vlasovi (approved):
Looks good

lib/usd/ui/importDialog/USDImportDialogCmd.cpp
```diff
@@ -206,8 +218,24 @@ MStatus USDImportDialogCmd::doIt(const MArgList& args)
             // toggle the wait cursor to show that it's working.
             QApplication::setOverrideCursor(QCursor(Qt::WaitCursor));
 
-            std::unique_ptr<USDImportDialog> usdImportDialog(new USDImportDialog(
-                assetPath.asChar(), &importData, usdQtUtil, MQtUtil::mainWindow()));
+            QWidget*      parentWindow = MQtUtil::mainWindow();
+            const MString parentWindowName
+                = parseTextArg(MArgDatabase(syntax(), args), kParentWindowFlag, "");
```
seando-adsk, line 223: Do you really need to create a MArgDatabase here? Can't you just use the MArgParser variable (argData) from above?

pierrebai-adsk: Oh I reused code from somewhere else without analyzing if it could be optimized.

plugin/adsk/scripts/mayaUsdTranslatorImport.mel
```diff
@@ -451,10 +451,21 @@ global proc string mayaUsdTranslatorImport_AppendFromDialog(string $currentOptio
     return $currentOptions;
 }
 
-global proc mayaUsdTranslatorImport_ViewHierCB()
+global proc mayaUsdTranslatorImport_ViewHierCB(string $parent)
 {
+    string $parentWindow = "";
+    {
+        string $result[];
+        string $windows = `window -parent $parent`;
```
seando-adsk, line 459: I'm unclear as to why you need to create a new window here to use as the parentWindow. Also is this window being destroyed?

pierrebai-adsk: I misunderstood what the `window -parent` command does! I thought it queried the parent of the given window. I will investigate and fix it. (If you know the crrect way to query the parent of an existing window, it would be welcome.)

seando-adsk: I searched and don't see any Maya command that will return the parent of a window. One way I can think of would be to call python and use MQtUtil::findWindow() method to get the QWidget and then call parent() on that.

pierrebai-adsk: I'll look around if there is a way to get the widget hierarchy from the panel name. Maybe something with `lsUI`.

seando-adsk (approved):
Looks good now.

vlasovi (approved):

PR #2894, opened 2023-02-22
MAYA-127484 : Deleting a connected compound attribute does not delete its connection.

alicedegirolamo:
# Description

This PR solves [Jira MAYA-127484](https://jira.autodesk.com/browse/MAYA-127484): when we delete a connected compound attribute we also have to delete the connections to/from its children and its parent, we should also check and delete (if necessary) the source and destination properties.

## Type of change

## How Has This Been Tested?

We have updated the `testAttributes.py` in order to test more cases for deleted connections and properties.


JGamache-autodesk (approved):

seando-adsk (approved):
Looks okay with me. I'm not very familiar with this code so I'll defer to Jerry for the review.

PR #2895, opened 2023-02-22
MAYA-124737 improve duplicate command

pierrebai-adsk:
The concept of layer from within the USD API is confusing relative to the high-level concept of layers. When talking about layers at a high-level, what is meant are the root and session layers and their sub-layers. OTOH, in the API, the layers are also used for references.  A reference is kept in a layer object (SdfLayer), but is not part of the stage layer stack!

In the case of duplication, we use SdfCopySpec to do the heavy lifting, and that takes a source path and source layer. But these are low-level layers, not high-level layers. in particular, if the source prim is in a reference, we must use the reference layer as the layer and the path must be relative to reference, not the stage root!

So, change the duplicate code to handle references correctly.

- Add a helper function to retrieve the defining layer and path of a prim.
- The defining path may not be the prim path if the prim is in a reference.
- In that case, the path is relative to the reference target, in the referenced layer.
- Also, refactor the defPrimSpecLayer function to make each check alone and separate to clarify and help when debugging.
- Remove the restriction on duplicate: a new prim is about to be created, it does not care about the structure of the source prim, so it should not be restricted.
- Use the new helper to find the correct source layer and path.
- Make sure to create missing parent 'overs'.
- Make the error message clearer by making it about USD prim, not low-level prim spec that the user might not know about.
- Make duplicate work with multi-layers prims by iterating over all layers with opinions and merge all opinions into the destination.
- Add helper function to retrieve all layer and path that have opinions.
- Add a unit test for the duplication multi-layers prim.
- Add a unit test for the duplication of a prim in a reference.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
@@ -119,7 +119,7 @@ void UsdUndoDuplicateCommand::execute()
             result,
             "Failed to copy the USD prim at '%s' in layer '%s' to '%s'",
             path.GetText(),
-            layer->GetDisplayName(),
+            layer->GetDisplayName().c_str(),
             _usdDstPath.GetText());
 
         isFirst = false;
--- a/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
@@ -23,6 +23,7 @@
 #include <mayaUsd/utils/loadRules.h>
 #ifdef UFE_V2_FEATURES_AVAILABLE
 #include <mayaUsd/undo/UsdUndoBlock.h>
+#include <mayaUsdUtils/MergePrims.h>
 #endif
 #include <mayaUsdUtils/util.h>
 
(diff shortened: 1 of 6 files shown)


lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
```diff
@@ -92,12 +93,37 @@ void UsdUndoDuplicateCommand::execute()
...
+    // Make sure all necessary parent exists in the target layer, at least as over,
+    // otherwise SdfCopySepc will fail.
+    SdfJustCreatePrimInLayer(_dstLayer, _usdDstPath.GetParentPath());
+
+    // Retrieve the layers where there are opinion and order them from weak
+    // to strong. We will copy the weakest opinions first, so that they will
+    // get over-written by the stronger opinions.
+    using namespace MayaUsdUtils;
+    std::vector<LayerAndPath> authLayerAndPaths = getAuthoredLayerAndPaths(prim);
+    std::reverse(authLayerAndPaths.begin(), authLayerAndPaths.end());
```
vlasovi, line 105: I don't think this has much importance here, but it would be more performance-efficient to not do std::reverse here and instead do a reverse cycle in the 'for' loop below.

vlasovi (approved):
Looks good

PR #2896, opened 2023-02-23
MAYA-127700 - Extend Viewport and Outliner menus to allowing assigning new or existing materials

stefanminning-autodesk:
# Description
This PR extends the Outliner and Viewport menus with new entries that allow the user to assign new or existing materials to objects. 
The difficulty here is the jumping between the different languages: Going from MEL via Python into C++ doesn't strike me as the most elegant approach, but I wasn't sure how else to handle it, since the viewport menu itself lives in MEL.

Simple unit tests are included for the newly added Commands.

# TODOs
Currently, the renderers and their materials (MaterialX, USD, Arnold) are hard-coded; this was largely ported from existing code in `UsdContextOps.cpp`. A future iteration should feature a more robust solution that queries the loaded renderers (including third-party!) for their supported materials.


seando-adsk:
@stefanminning-autodesk pr-build doesn't work for github. To run a preflight here set the assignee to ecp-maya-devops-adsk

plugin/adsk/plugin/adskMaterialCommands.cpp
```diff
@@ -0,0 +1,180 @@
...
+        // TODO: Replace hard-coded materials with dynamically generated list.
+        const MString label = "AI Standard Surface";
+        const MString identifier = "arnold:standard_surface";
+        appendToResult("Arnold/" + label + "|" + identifier);
+    }
+}
+
+void ADSKMayaUSDGetMaterialsForRenderersCommand::appendUsdMaterials() const
+{
+    // TODO: Replace hard-coded materials with dynamically generated list.
```
JGamache-autodesk, line 92: No need for a TODO here. There will always be only one preview material.

plugin/adsk/plugin/adskMaterialCommands.cpp
```diff
@@ -0,0 +1,180 @@
...
+    if (ufePathString.isEmpty()) {
+        MGlobal::displayError("Missing argument 'sceneItem'.");
+        return MS::kFailure;
+    }
+
+    const auto  ufePath = Ufe::PathString::path(ufePathString.asChar());
+    UsdStagePtr stage = ufe::getStage(ufePath);
+    if (stage) {
+        for (auto prim : stage->Traverse()) {
+            if (prim.GetTypeName() == materialType) {
```
JGamache-autodesk, line 164: I would prefer you use the stronger `if (UsdShadeMaterial(prim))` as filter.

plugin/adsk/plugin/adskMaterialCommands.h
```diff
@@ -0,0 +1,92 @@
...
+#include <ufe/path.h>
+#include <ufe/undoableCommand.h>
+
+namespace MAYAUSD_NS_DEF {
+
+//------------------------------------------------------------------------------
+// GetMaterialXMaterialsCommand
+//------------------------------------------------------------------------------
+
+struct MxShaderMenuEntry
```
JGamache-autodesk, line 35: I don't think this struct is used. Can you recheck and see it it can be removed?

plugin/adsk/scripts/USDMenuProc.mel
```diff
@@ -55,6 +55,117 @@ global proc mayaUsdMenu_duplicate(string $ufePath)
...
+{
+    if (!hasPrimUpdater())
+        return;
+
+    if (size($ufePath) != 0 && size($material) != 0) {
+        string $temp = `python("import ufe;\
+                                ufePath = ufe.PathString.path('" + $ufePath + "');\
+                                item = ufe.Hierarchy.createItem(ufePath);\
+                                contextOps = ufe.ContextOps.contextOps(item);\
+                                cmd = contextOps.doOpCmd(['Assign New Material', 'USD', '" + $material + "']);\
```
JGamache-autodesk, line 68: The `'USD'` part could be misleading. Looks like the context ops are only interested in items 0 and 2 in that list. Maybe you should leave that middle string blank?

test/lib/mayaUsd/utils/testMaterialCommands.py
```diff
@@ -0,0 +1,92 @@
...
+        Checks that the list of materials found in the stage matches the expected values.
+        """
+        self._StartTest('multipleMaterials')
+        
+        expectedMaterials = ['/mtl/UsdPreviewSurface1', '/mtl/UsdPreviewSurface2']
+        
+        materialsInStage = cmds.mayaUsdGetMaterialsInScene(si="|stage|stageShape,/cube")
+        self.assertEqual(materialsInStage, expectedMaterials)
+
+
```
JGamache-autodesk, line 90: You could add a test for `ContextOp::getItems()` to check if the subtrees of add new/existing material are correctly populated.

JGamache-autodesk (changes requested):

plugin/adsk/plugin/adskMaterialCommands.cpp
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+
+    return MS::kSuccess;
+}
+
+MSyntax ADSKMayaUSDGetMaterialsInStageCommand::createSyntax()
+{
+    MSyntax syntax;
+    syntax.addFlag("-si", "-sceneItem", MSyntax::kString);
```
seando-adsk, line 176: I don't think you need or should use a flag like this. Its not really a "sceneitem". It is actually a string which is interpreted as a Ufe PathString. 

Instead look at calling ```addArg(MSyntax::kString)``` then simply get that string arg. There are examples of this in our code already. Plus you should probably set the ```syntax.enableQuery/enableEdit``` as well.

seando-adsk (changes requested):

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -1157,30 +1044,17 @@ Ufe::ContextOps::Items UsdContextOps::getItems(const Ufe::ContextOps::ItemPath&
...
-            items.emplace_back(
-                kAssignNewUsdPreviewSurfaceMaterialItem, kAssignNewUsdPreviewSurfaceMaterialLabel);
-        } else if (itemPath.size() == 2u && itemPath[1] == kAssignNewMaterialXMaterialItem) {
-            for (auto&& menuEntry : getMaterialXSurfaceShaders()) {
-                items.emplace_back(menuEntry._identifier, menuEntry._label);
-            }
-        } else if (itemPath.size() == 2u && itemPath[1] == kAssignNewArnoldMaterialItem) {
-            items.emplace_back(
-                kAssignNewAIStandardSurfaceMaterialItem, kAssignNewAIStandardSurfaceMaterialLabel);
+            MString script;
```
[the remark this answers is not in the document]
JGamache-autodesk: Yes.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -1154,33 +1042,78 @@ Ufe::ContextOps::Items UsdContextOps::getItems(const Ufe::ContextOps::ItemPath&
...
+            MStringArray                        materials;
+            MString                             script;
+            script.format(
+                "mayaUsdGetMaterialsInStage \"^1s\"",
+                Ufe::PathString::string(fItem->path()).c_str());
+            MGlobal::executeCommand(script, materials);
+
+            for (const auto& material : materials) {
+                MStringArray pathAndMaterial;
+                MStatus      status = material.split('/', pathAndMaterial);
```
JGamache-autodesk, line 1091: Why not a reverse find for the last "/" that will allow you to split the string in two at the right location?

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -1154,33 +1042,78 @@ Ufe::ContextOps::Items UsdContextOps::getItems(const Ufe::ContextOps::ItemPath&
...
+                // Expects a string in the format "/path1/path2|Material".
```
JGamache-autodesk, line 1092: Is that comment correct? I don't see any splitting on the "|" in the below code.

JGamache-autodesk (approved):

plugin/adsk/scripts/USDMenuProc.mel
```diff
@@ -55,6 +55,117 @@ global proc mayaUsdMenu_duplicate(string $ufePath)
...
+    if (!hasPrimUpdater())
+        return;
+
+    if (size($ufePath) != 0 && size($material) != 0) {
+        string $temp = `python("import ufe;\
+                                ufePath = ufe.PathString.path('" + $ufePath + "');\
+                                item = ufe.Hierarchy.createItem(ufePath);\
+                                contextOps = ufe.ContextOps.contextOps(item);\
+                                cmd = contextOps.doOpCmd(['Assign New Material', '', '" + $material + "']);\
+                                cmd.execute();")`;
```
JGamache-autodesk, line 69: This is where you need to use `Ufe::UndoableCommandMgr::instance().executeCmd(cmd);`

JGamache-autodesk (changes requested):

seando-adsk:
@stefanminning-autodesk According to our [coding guidelines](https://github.com/Autodesk/maya-usd/blob/dev/doc/codingGuidelines.md) you need to use clang-format v10.0.0

test/lib/mayaUsd/utils/testMaterialCommands.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        Checks that the list of new materials for different renderers matches the expected values.
+        Arnold is excluded here as it cannot be assumed to be installed in the testing environment.
+        """
+
+        self._StartTest()
+
+        expectedMaterials = ['USD/USD Preview Surface|UsdPreviewSurface',
+                             'MaterialX/Standard Surface|ND_standard_surface_surfaceshader',
+                             'MaterialX/glTF PBR|ND_gltf_pbr_surfaceshader',
```
JGamache-autodesk, line 71: Earlier versions of MaterialX in earlier versions of USD do not have this `glTf` shader. Removing this line should help the unit test succeed on legacy configurations.

plugin/adsk/plugin/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ target_sources(${TARGET_NAME}
         adskExportCommand.cpp
         adskListJobContextsCommand.cpp
         adskListShadingModesCommand.cpp
+        adskMaterialCommands.cpp
```
JGamache-autodesk, line 30: You need to add this to the list TARGET_NAME only is UFE_FOUND is true. This will fix the `ecg-mayausd-branch-preflight-2019-no-ufe-windows` build.

JGamache-autodesk (approved):

test/lib/mayaUsd/utils/testMaterialCommands.py
```diff
@@ -75,6 +75,10 @@ def testMayaUsdGetNewMaterials(self):
 
         materials = cmds.mayaUsdGetMaterialsFromRenderers()
 
+        # TODO: Remove after debugging
+        print(materials)
+        print(expectedMaterials)
```
JGamache-autodesk, line 80: Might need something like:
```
    if(CMAKE_WANT_MATERIALX_BUILD)
        list(APPEND TEST_SCRIPT_FILES
            testMaterialCommands.py
        )
    endif()
```
In the CMakeLists.txt.

seando-adsk (approved):
Looks good now. Thanks.

JGamache-autodesk (dismissed):

JGamache-autodesk (approved):

seando-adsk (approved):

PR #2899, opened 2023-02-24
MAYA-124922 use Load/Unload for payload

pierrebai-adsk:
Unfortunately, USD stage's SetLoadRules function does not optimize the change notifications it sends out. Its Load and Unload functions do optimize the notifications they send out. Normally, this is an implementation details inside USD that we should not have to rely on. Pragmatically though, using the functions which send out the most narrow notification is better. So even though calling SetLoadRules is simpler and more precise, it is less efficient.

### DIFF
--- a/lib/mayaUsd/ufe/UsdContextOps.cpp
+++ b/lib/mayaUsd/ufe/UsdContextOps.cpp
@@ -335,24 +335,20 @@ protected:
     LoadUnloadBaseUndoableCommand(const UsdPrim& prim)
         : _stage(prim.GetStage())
         , _primPath(prim.GetPath())
-        , _policy(getCurrentPolicy())
-    {
-    }
-
-    UsdLoadPolicy getCurrentPolicy() const
+        , _policy(UsdLoadPolicy::UsdLoadWithoutDescendants)
     {
         if (!_stage)
-            return UsdLoadPolicy::UsdLoadWithoutDescendants;
+            return;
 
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -324,59 +324,75 @@ class InsertChildAndSelectCommand : public Ufe::CompositeUndoableCommand
...
+        , _primPath(prim.GetPath())
+        , _policy(policy)
+    {
+    }
+
     LoadUnloadBaseUndoableCommand(const UsdPrim& prim)
         : _stage(prim.GetStage())
         , _primPath(prim.GetPath())
-        , _oldLoadRules(prim.GetStage()->GetLoadRules())
+        , _policy(getCurrentPolicy())
```
vlasovi, line 338: Technically, this looks pretty dangerous because you call the function getCurrentPolicy from inside the constructor, and inside it uses _stage which may not be initialized at this point, depending on the order of initialization of data members.

vlasovi (changes requested):
See comments for "_policy(getCurrentPolicy())"

lib/mayaUsd/ufe/UsdContextOps.cpp
pierrebai-adsk: Yes, I knew it and ordered the members for it to work. I can easily fix this by putting the code in the body.

PR #2900, opened 2023-02-24
Removed an artifical crash in transform operations and replaced it with a asserts.

AramAzhari-adsk:
Fixes the crash when multiple instances of the same stage file are being rotated / scaled simultaneously.

### DIFF
--- a/lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
+++ b/lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
@@ -473,7 +473,7 @@ UsdTransform3dMayaXformStack::rotateCmd(double x, double y, double z)
                 UsdGeomXformable xformable(usdSceneItem->prim());
 
                 auto r = xformable.AddRotateXYZOp(UsdGeomXformOp::PrecisionFloat, opSuffix);
-                TF_AXIOM(r);
+                TF_VERIFY(r);
                 r.Set(v);
                 auto result = setXformOpOrderFn(xformable);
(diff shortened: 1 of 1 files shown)


AramAzhari-adsk:
For any other TF_Axiom that isn't directly resolving the issues in this particular jira, we'll have a separate task to refactor.

pierrebai-adsk (approved):

lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
```diff
@@ -520,7 +520,7 @@ Ufe::ScaleUndoableCommand::Ptr UsdTransform3dMayaXformStack::scaleCmd(double x,
                 UsdGeomXformable xformable(usdSceneItem->prim());
 
                 auto s = xformable.AddScaleOp(UsdGeomXformOp::PrecisionFloat, opSuffix);
-                TF_AXIOM(s);
+                TF_VERIFY(s);
```
pierrebai-adsk, line 523: There are similar case for scale etc, for example line 687

AramAzhari-adsk: Those are actually pivot commands. I've noticed any translation cmd such as translateCmd, translateRotatePivotCmd etc do not have any crash issues.

seando-adsk: I hate that github won't let you comment on unmodified lines!

seando-adsk (approved):

PR #2902, opened 2023-02-24
MAYA-127737 allow loading relative sub layers

pierrebai-adsk:
- Add UI to load-layer file dialog to make the file path relative, in MEL.
- Modify the overall load-layer dialog to support relative sub-layers.


seando-adsk (dismissed):
Everything looks good. Just one minor thing about possibly adding a code comment. If you do, just add a PR saying that all you added was code comment and no need to run preflight again.

seando-adsk (approved):

PR #2904, opened 2023-02-27
MAYA-127427 fix rename when staged multiple times

pierrebai-adsk:
- Protect UsdSceneItem against invalid prims.
- Protect USD UI info handler against invalid prims.
- These can happen for example when we're in the middle of a rename and the outliner gets updated or redrawn.
- Fix the rename command to send rename notifications to all proxy shape that share the same stage.
- At the same time, refactor the redo/undo code that was 99% identical.


AramAzhari-adsk (approved):

vlasovi (approved):
Looks good

PR #2905, opened 2023-02-27
MAYA-127370 - Preview of my relative path (resolved and unresolved)

seando-adsk:
#### MAYA-127370 - Preview of my relative path (resolved and unresolved)
* Added two path preview fields with logic to update them on any change to path/file from save/load dialog where the relative path is used.
* Added python wrapper for util ensureUSDFileExtension()

### DIFF
--- a/lib/mayaUsd/python/wrapUtil.cpp
+++ b/lib/mayaUsd/python/wrapUtil.cpp
@@ -16,6 +16,7 @@
 #include <mayaUsd/fileio/jobs/jobArgs.h>
 #include <mayaUsd/utils/util.h>
 #include <mayaUsd/utils/utilFileSystem.h>
+#include <mayaUsd/utils/utilSerialization.h>
 
 #include <pxr/base/tf/pyResultConversions.h>
 #include <pxr/pxr.h>
@@ -44,6 +45,13 @@ VtDictionary getDictionaryFromEncodedOptions(const std::string& textOptions)
     return dictOptions;
 }
 
+std::string ensureUSDFileExtension(const std::string& fileToCheck)
+{
+    std::string ret(fileToCheck);
+    MayaUsd::utils::ensureUSDFileExtension(ret);
+    return ret;
+}
+
 } // namespace
 
 void wrapUtil()
@@ -56,5 +64,6 @@ void wrapUtil()
                   .def(
                       "getPathRelativeToMayaSceneFile",
                       UsdMayaUtilFileSystem::getPathRelativeToMayaSceneFile)
+                  .def("ensureUSDFileExtension", ensureUSDFileExtension)
                   .staticmethod("getPathRelativeToMayaSceneFile");
 }
--- a/plugin/adsk/scripts/mayaUsd_USDRootFileRelative.py
+++ b/plugin/adsk/scripts/mayaUsd_USDRootFileRelative.py
@@ -1,7 +1,21 @@
 import maya.cmds as cmds
 import maya.mel as mel
+import maya.OpenMayaUI as omui
+import mayaUsd.lib as mayaUsdLib
 from mayaUSDRegisterStrings import getMayaUsdString
 from mayaUsdMayaReferenceUtils import pushOptionsUITemplate
+
+try:
+    from PySide2.QtCore import QObject, QFileInfo
+    from PySide2.QtWidgets import QFileDialog, QLayout, QWidget, QLineEdit, QDialogButtonBox, QComboBox
+    from shiboken2 import wrapInstance
+except:
+    from PySide6.QtCore import QObject, QFileInfo
+    from PySide6.QtWidgets import QFileDialog, QLayout, QWidget, QLineEdit, QDialogButtonBox, QComboBox
+    from shiboken6 import wrapInstance
+
+# Global variables
+_relativeToFilePath = None
 
 class usdFileRelative(object):
     '''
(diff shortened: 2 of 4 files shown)


lib/mayaUsd/python/wrapUtil.cpp
```diff
@@ -56,5 +64,6 @@ void wrapUtil()
                   .def(
                       "getPathRelativeToMayaSceneFile",
                       UsdMayaUtilFileSystem::getPathRelativeToMayaSceneFile)
+                  .def("ensureUSDFileExtension", ensureUSDFileExtension)
```
seando-adsk, line 67: Expose a new python helper. Needed helper function with return value (as opposed to output variable) to call here.

plugin/adsk/scripts/mayaUsd_USDRootFileRelative.py
```diff
@@ -1,8 +1,22 @@
...
 from mayaUsdMayaReferenceUtils import pushOptionsUITemplate
 
+try:
+    from PySide2.QtCore import QObject, QFileInfo
+    from PySide2.QtWidgets import QFileDialog, QLayout, QWidget, QLineEdit, QDialogButtonBox, QComboBox
+    from shiboken2 import wrapInstance
+except:
+    from PySide6.QtCore import QObject, QFileInfo
+    from PySide6.QtWidgets import QFileDialog, QLayout, QWidget, QLineEdit, QDialogButtonBox, QComboBox
+    from shiboken6 import wrapInstance
```
seando-adsk, line 15: Need Qt to find and connect to controls in dialog.

plugin/adsk/scripts/mayaUsd_USDRootFileRelative.py
```diff
@@ -18,26 +32,38 @@ class usdFileRelative(object):
...
-    def getRelativeFilePathRoot(cls):
+    @staticmethod
+    def getRelativeFilePathRoot():
         '''
         Gets to which file the relative file will be anchored.
         Empty or None to have the file be absolute, not relative.
         '''
-        return cls._relativeToFilePath
+        global _relativeToFilePath
+        return _relativeToFilePath
```
seando-adsk, line 66: Made these static methods using global variable, to ensure we are always setting/getting the right value. There can only ever be a single instance of this dialog/class(es) at a time.

plugin/adsk/scripts/mayaUsd_USDRootFileRelative.py
```diff
@@ -67,12 +93,21 @@ def uiCreate(cls, parentLayout, relativeToWhat):
...
+        kUnresolvedPathAnnStr = getMayaUsdString("kUnresolvedPathAnn")
+        kResolvedPathStr = getMayaUsdString("kResolvedPath")
+        kResolvedPathAnnStr = getMayaUsdString("kResolvedPathAnn")
  
         optBoxMarginWidth = mel.eval('global int $gOptionBoxTemplateDescriptionMarginWidth; $gOptionBoxTemplateDescriptionMarginWidth += 0')
         cmds.setParent(topForm)
         cmds.frameLayout(label=kFileOptionsStr, collapsable=False)
         widgetColumn = cmds.columnLayout()
-        cmds.checkBox(cls.kMakePathRelativeCheckBox + relativeToWhat, label=kMakePathRelativeStr, ann=kMakePathRelativeAnnStr)
+        cmds.checkBox(cls.kMakePathRelativeCheckBox, label=kMakePathRelativeStr, ann=kMakePathRelativeAnnStr)
```
seando-adsk, line 105: Since there can only be a single dialog no need to have differently named check boxes.

plugin/adsk/scripts/mayaUsd_USDRootFileRelative.py
```diff
@@ -67,12 +93,21 @@ def uiCreate(cls, parentLayout, relativeToWhat):
...
+        if relativeToWhat == 'SceneFile':
+            cmds.checkBox(cls.kMakePathRelativeCheckBox, edit=True, changeCommand=cls.onMakePathRelativeChanged)
+            cmds.textFieldGrp(cls.kUnresolvedPathTextField, label=kUnresolvedPathStr, ann=kUnresolvedPathAnnStr, editable=False)
+            cmds.textFieldGrp(cls.kResolvedPathTextField, label=kResolvedPathStr, ann=kResolvedPathAnnStr , editable=False)
+            cls._haveRelativePathFields = True
```
seando-adsk, line 110: For now only add the preview fields for the SceneFile. Design is still working on the preview for the edit layer target case.

plugin/adsk/scripts/mayaUsd_USDRootFileRelative.py
```diff
@@ -110,9 +161,100 @@ def uiCommit(cls, parentLayout, relativeToWhat):
...
+            # Find the file name edit field and connect to it to be notified when text changes.
+            fileNameEditField = maya_window.findChild(QLineEdit, 'fileNameEdit')
+            if fileNameEditField:
+                fileNameEditField.textChanged.connect(cls.onfileNameEditFieldChanged)
+
+            # Find the accept button, so we can know when there is valid input in the file name edit field.
+            buttonBox = maya_window.findChild(QDialogButtonBox, 'buttonBox')
+            if buttonBox and cls._fileDialog:
+                btn = QDialogButtonBox.Open if (cls._fileDialog.acceptMode() == QFileDialog.AcceptOpen) else QDialogButtonBox.Save
+                cls._acceptButton = buttonBox.button(btn)
```
seando-adsk, line 198: The dialog will enable/disable the OK button according to whether or not there is valid filename. So I use the state of that button to know if there is a valid filename to query.

plugin/adsk/scripts/mayaUsd_USDRootFileRelative.py
```diff
@@ -110,9 +161,100 @@ def uiCommit(cls, parentLayout, relativeToWhat):
...
+
+            # Find the lookin combobox and connect to it so we know when user changes directories.
+            lookinComboBox = maya_window.findChild(QComboBox, 'lookInCombo')
+            if lookinComboBox:
+                lookinComboBox.currentTextChanged.connect(cls.onLookinTextChanged)
```
seando-adsk, line 203: Connect to this signal to know when user changes the folder.

pierrebai-adsk (approved):
LGTM

PR #2906, opened 2023-02-28
MAYA-127959 avoid crash when world node has an unexpected name

pierrebai-adsk:
Verify that the UFE item has been successfully created before using it.


seando-adsk (approved):
As a general fix this is great to have.

PR #2908, opened 2023-02-28
Add env var MAYAUSD_VP2_USE_ONLY_PREVIEWSURFACE

### LINKED ISSUE #2637: Option to disable MaterialX
I need to use MaterialX shaders for rendering in solaris, but as soon as I include them in a USD file, the geometry doesn't render in the Maya viewport. Basically making it impossible to work with assets with MaterialX shaders on in Maya.

I would like an option to fall back to the USD surface preview shader in a material if it exists in a USD file, rather than prioritising the materialX shader. This could be an ENV var set before launching Maya or an option on the mayaUsdProxyShape.

This could also be useful when materialX shaders are overly complex for the viewport and a simple diffuse would work.

Hopefully this isn't particularly difficult to implement as it's really blocking our Usd maya adoption.

JGamache-autodesk:
This env var forces using UsdPreviewSurface shading for VP2.

Fixes #2637.


vlasovi (approved):
I think a better name for this env var would have been MAYAUSD_VP2_ALLOW_MATERIAlX, but otherwise looks good

JGamache-autodesk:
@seando-adsk Ready for merge. 

MaterialX is allowed by default, so negative env var works better to express the requirement.

seando-adsk (approved):

Lucas3Dspain:
is there a doc where we have all these kinds of env vars? I see some of them in `lib/mayaUsd/commands/Readme.md`
I'm worried about people not knowing they are available.

JGamache-autodesk:
The most interesting part is that they can be [found directly](https://github.com/Autodesk/maya-usd/search?q=TF_DEFINE_ENV_SETTING) in this public repo and the TF_DEFINE_ENV_SETTING syntax requires a description string. We are indeed a bit behind since there is no official documentation on the dozen of env settings this query finds in the Adsk plugin.

PR #2915, opened 2023-03-02
Add a getter for usdimaging delegate in proxy adapter

Vsevolod1983:
Hello !
Could we add a getter for usd imaging delegate in proxy adapter?

We are making production render feature in our plugin with our hydra render delegate (RadeonProRender hdRPR) and we try reuse as much code as possible from MtoH.
Thank you!

### DIFF
--- a/lib/usd/hdMaya/adapters/proxyAdapter.h
+++ b/lib/usd/hdMaya/adapters/proxyAdapter.h
@@ -90,6 +90,7 @@ public:
     }
 
     const SdfPath& GetUsdDelegateID() const { return _usdDelegate->GetDelegateID(); }
+    HdMayaProxyUsdImagingDelegate* GetUsdImagingDelegate() { return _usdDelegate.get(); }
 
 private:
     /// Notice listener method for proxy stage set


lib/usd/hdMaya/adapters/proxyAdapter.h
```diff
@@ -90,6 +90,7 @@ class HdMayaProxyAdapter
     }
 
     const SdfPath& GetUsdDelegateID() const { return _usdDelegate->GetDelegateID(); }
+    HdMayaProxyUsdImagingDelegate* GetUsdImagingDelegate() { return _usdDelegate.get(); }
```
ppt-adsk, line 93: I would have commented that a const version might have been useful as well, but I see that HdMayaProxyUsdImagingDelegate appears to not be const-correct :(

ppt-adsk (approved):

ppt-adsk:
Hi @Vsevolod1983 we are looking in broader terms as to how Hydra is supported in Maya.  Feel free to contact us for more information.

Vsevolod1983:
Ok, thank you!

PR #2917, opened 2023-03-03
Check if node is referenced - use default value to detect change

tryptik:
When exporting file nodes from Maya to Usd and the file node is referenced, the export code does not recognize that the Maya attributes for scale/bias are not set properly despite these values being correct in the referenced Maya scene. This seems to be because the exporting code is using `MPlug::getSetAttrCmds` and checking the result to see if it is empty to determine if the values need to be written.

This P/R checks the Dependency node to see if it is from a referenced file, then uses `MPlug::isDefaultValue` to determine if the plug values need to be authored on the exported Usd data.

Note:
In my past testing, `Plug::isDefaultValue` seemed unreliable, and I suspect it remains so based on the code using `MPlug::getSetAttrCmds`, so if there is a better approach, please advise.

`MPlug::isFromReferencedFile` does not seem to work, which is why `MFnDependencyNode::isFromReferencedFile` is used.

### DIFF
--- a/lib/mayaUsd/utils/util.cpp
+++ b/lib/mayaUsd/utils/util.cpp
@@ -1236,9 +1236,18 @@ bool UsdMayaUtil::IsAuthored(const MPlug& plug)
     // make a copy of plug here.
     MPlug plugCopy(plug);
 
+    // MPlug::isFromReferencedFile did not seem to be accurate, so this checks the node state.
+    // If the node is referenced, use MPlug::isDefaultValue - it is not clear if isDefaultValue works
+    // in all contexts as the code below uses MPlug::getSetAttrCmds. 
+    //
+    MFnDependencyNode nodeFn(plug.node());
+    if(nodeFn.isFromReferencedFile() && !plug.isDefaultValue())
+        return true;
+
     MStringArray setAttrCmds;
     status = plugCopy.getSetAttrCmds(setAttrCmds, MPlug::kChanged);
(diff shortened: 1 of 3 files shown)


pierrebai-adsk (approved):

tryptik:
> @tryptik Any update on this PR? You'll need to fix the clang-format issue.

Will take a look today.

tryptik:
Pushed update.

tryptik:
Did this go through? I can't tell if the clang_format_linter has been re-run - the date on it looks like May 9.

pierrebai-adsk:
Maybe you need to merge the latest dev into your branch. Weirdly, I cannot find your branch in the list of branches. It does have a non-standard name with a column in it, which would wreak havoc on Windows. I suggest:
- Create a new branch off your branch. Name it dgovil/referenced-shader-attrs-issue
- Merge dev into it
- Push it and assign ecp-maya-devops-adsk so that the preflight runs

Sorry for the troubles...

tryptik:
I created a new [P/R](https://github.com/Autodesk/maya-usd/pull/3214) however I am unable to assign `ecp-maya-devops-adsk` - should I be able to?

pierrebai-adsk:
Fixed format, pushed onto your branch in your fork (hope you don't mind :) ) and started the build in the new PR

tryptik:
Haha I do not mind - thank you for fixing it! I've properly clanged before, so I'm not sure why I struggled so much with it this time.

PR #2918, opened 2023-03-03
MAYA-128040: isDisconnected local variable treated as error if pxr > v0.23

perrauo-adsk:
* I have found "isDisconnected  not used" treated as error that if PXR > v0.23 is used


seando-adsk (approved):

seando-adsk:
@perrauo-adsk There was no preflight run, so I removed the ready for merge label and started the preflight for you. Once it passed you can add back the label.

PR #2922, opened 2023-03-07
MAYA-126128 sync session layer for various commands

pierrebai-adsk:
Keep opinions in the session layer in sync when a prim is renamed while targeting another layer. Do the same for insert child, group delete and reorder commands.

Provide helper functions to deal with the session layer and enhance the various commands to allow them when the session layer contains opinions, by using these helper functions.

Allow authoring in session layer: detect that the target is in the session layer and allow edits there if it is the strongest session opinion.

- Move many layer-related helpers into their own file.
- Refactor code that used the old functions to use the new functions.
- Add applyToAllPrimSpecs and applyToAllLayersWithOpinions helper functions.
- Add applyToSomeLayersWithOpinions helper function to apply a change to a subset of layers.
- Use applyToAllPrimSpecs helper function in the rename command.
- Use the helper functions in the rename, delete, insert child and reorder commands..
- Properly delete all prim spec in session layer when deleting a prim.
- Properly throw an exception when failing.
- This is necessary so that composite commands know a sub-command has failed.
- Add rename unit test that demonstrate that prims can be renamed when they have opinions in the session layer.
- Add unit tests for group (insert child) and reorder.
- Fix the delete command unit test.
- Fix unit tests to use non-session layer when testing restrictions.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -25,6 +25,7 @@
 #include <mayaUsd/utils/customLayerData.h>
 #include <mayaUsd/utils/diagnosticDelegate.h>
 #include <mayaUsd/utils/layerMuting.h>
+#include <mayaUsd/utils/layers.h>
 #include <mayaUsd/utils/loadRules.h>
 #include <mayaUsd/utils/query.h>
 #include <mayaUsd/utils/stageCache.h>
@@ -929,7 +930,7 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
 
     if (isIncomingStage) {
         std::vector<std::string> incomingLayers { sharedUsdStage->GetRootLayer()->GetIdentifier() };
-        _incomingLayers = UsdMayaUtil::getAllSublayers(incomingLayers, true);
+        _incomingLayers = getAllSublayers(incomingLayers, true);
     } else {
         _incomingLayers.clear();
     }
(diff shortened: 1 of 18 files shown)


lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -940,14 +941,10 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
...
             _unsharedStageRootSublayers.clear();
-            for (const auto& identifier : subLayerIds) {
-                SdfLayerRefPtr sublayer = SdfLayer::Find(identifier);
-                if (sublayer) {
-                    _unsharedStageRootSublayers.push_back(sublayer);
-                }
-            }
+            auto subLayers = getAllSublayerRefs(_unsharedStageRootLayer);
+            _unsharedStageRootSublayers.insert(
+                _unsharedStageRootSublayers.begin(), subLayers.begin(), subLayers.end());
```
pierrebai-adsk, line 947: The new function already returns layer refs instead of layer names, simplifying the code.

lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
```diff
@@ -85,8 +87,10 @@ void UsdUndoDeleteCommand::execute()
     auto        targetPrimSpec = stage->GetEditTarget().GetPrimSpecForScenePath(_prim.GetPath());
 
     if (hasLayersMuted(_prim)) {
-        TF_WARN("Cannot remove prim because there are muted layers.");
-        return;
+        const std::string error = TfStringPrintf(
+            "Cannot remove prim \"%s\" because there are muted layers.", _prim.GetPath().GetText());
+        TF_WARN("%s", error.c_str());
+        throw std::runtime_error(error);
```
pierrebai-adsk, line 93: Maya catches exceptions from UFE command, so this change is fine. It is required so that UFE composite command can be properly failed when a sub-command fails. For example, the group and reparent UFE commands are actually composite commands and used to be able to half-fail without even knowing they had failed.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -164,7 +165,81 @@ UsdUndoInsertChildCommand::Ptr UsdUndoInsertChildCommand::create(
     return std::make_shared<MakeSharedEnabler<UsdUndoInsertChildCommand>>(parent, child, pos);
 }
 
-bool UsdUndoInsertChildCommand::insertChildRedo()
+static void doUsdInsertion(
+    const UsdStagePtr&    stage,
```
pierrebai-adsk, line 169: This functions refactor the common code between undo and redo. The same refactor pattern is also used in other commands, like rename. So most of the following code is actually moved from below, with the enhancement required to deal with the session layers added..

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.h
```diff
@@ -66,8 +66,8 @@ class MAYAUSD_CORE_PUBLIC UsdUndoInsertChildCommand : public Ufe::InsertChildCom
     void undo() override;
     void redo() override;
 
-    bool insertChildRedo();
-    bool insertChildUndo();
+    void insertChildRedo();
+    void insertChildUndo();
```
pierrebai-adsk, line 70: Now return void since errors are reported through exceptions.

vlasovi (approved):
Looks good

PR #2924, opened 2023-03-08
MAYA-126128 transform command error reporting

pierrebai-adsk:
Make the transform commands propagate the errors they encounters instead of just logging them. This is necessary to make UFE composite commands work properly: they need to know that a sub-command failed and stop execution. Otherwise they just keep going and pretend they worked when they didn't, corrupting the state of USD or Maya.

- Add helper functions to enforce blocked attribute edits by throwing exception with a clear message when edits are not allowed.
- Make all transform commands enforce the blocked edits.
- This fixes the group, ungroup and parenting UFE composite commands when a subset of items cannot be modified.


pierrebai-adsk:
I will rebase this branch off dev becasue it contains changes that were part of a previous branch that got merged.

PR #2927, opened 2023-03-09
[al]Fixing multiple colour sets export from Maya to USD

NickWu:
We found a logical error in multiple color sets exporting from Maya to USD in AL_USDMaya where vertex colours were written wrong. This PR addresses the said issue.


seando-adsk:
@JGamache-autodesk Jerry would you mind code reviewing this change.

JGamache-autodesk (approved):

PR #2929, opened 2023-03-09
MAYA-124715: Error: Failed verification: on File > New

seando-adsk:
#### MAYA-124715: Error: Failed verification: on File > New
* Removed TF_WARN - it is not appropriate for low-level utility function.


pierrebai-adsk (approved):
LGTM

PR #2930, opened 2023-03-09
Removed DisplayColor from ShadingMode export command as the flag has its own command exportDisplayColor

AramAzhari-adsk:
Removed DisplayColor from ShadingMode export command as the flag has its own command exportDisplayColor.


AramAzhari-adsk:
There is still a shadingMode with DisplayColor on the **import** section. Is that also removed?
https://github.com/Autodesk/maya-usd/blob/a87c16de07444e0d6dda064ed151827538916c52/lib/mayaUsd/commands/Readme.md?plain=1#L45

If so, then both the above and 
https://github.com/Autodesk/maya-usd/blob/a87c16de07444e0d6dda064ed151827538916c52/lib/mayaUsd/commands/Readme.md?plain=1#L43

need to be updated.

seando-adsk (approved):

seando-adsk:
@AramAzhari-adsk Sorry saw your questions here after approving. I don't know the answers. Maybe @pierrebai-adsk does.

AramAzhari-adsk:
> @AramAzhari-adsk Sorry saw your questions here after approving. I don't know the answers. Maybe @pierrebai-adsk does.

@seando-adsk it is not required to complete this PR.

PR #2933, opened 2023-03-10
MAYA-124702: retrieve ufe runtime id by data source integer reference

perrauo-adsk:
Summary:
* In order to enable UFE support with custom scene indices, the UFE runtime id must be retrieved.
* This is because Maya is unaware of what runtime a scene index needs.

Changes:
* In the case of USD, uppon call to _AppendSceneIndex retrieve the USD Ufe runtime, and return using the provide Ufe::Rtid& data source.


seando-adsk (changes requested):
I think all this new Ufe code should be ifdef protected: `#if WANT_UFE_BUILD`
I think the reason your preflight passed is because this "sceneIndex" folder is only include when PXR_VERSION >= 2211. In this case we are building with Ufe, but we should still support using that Usd version and no Ufe.

seando-adsk (approved):

PR #2934, opened 2023-03-10
MAYA-128211 block command when layers are muted

pierrebai-adsk:
When there is at least one muted layers, block the delete, rename, group and reparent commands. These are commands that re-organize the hierarchy and must thus affect all layers with opinions. When some of these layers are muted, they would not be able to be updated with the new prim path, thus we prevent the operation.

Add unit tests for the commands. Note that reparent and group are composite commands using the same underlying command (insert child) and thus require only one test.


pierrebai-adsk:
Note PF only failed in one OSX instance which timeout, but that particular platform passed in a previous attempt, so it was only a glitch.

vlasovi (approved):
Looks good

PR #2945, opened 2023-03-15
Detect if USD was built with extra security patches

seando-adsk:
Detect if USD was built with extra security patches and set mod file variable MAYA_USD_VERSION accordingly.


cmake/modules/FindUSD.cmake
```diff
@@ -66,6 +66,21 @@ elseif(USD_INCLUDE_DIR AND EXISTS "${USD_INCLUDE_DIR}/pxr/pxr.h")
...
+    # In the security fix there was a new inline method "max_size" added so we
+    # can use that to determine if this USD build has the security fix.
+    if (USD_INCLUDE_DIR AND EXISTS "${USD_INCLUDE_DIR}/pxr/base/vt/array.h")
+        file(STRINGS ${USD_INCLUDE_DIR}/pxr/base/vt/array.h USD_HAS_SECURITY_PATCH REGEX "constexpr size_t max_size\(\)")
+        if (USD_HAS_SECURITY_PATCH)
+            set(ADSK_USD_VERSION "${ADSK_USD_VERSION}-ad1")
+            message(STATUS "USD v${USD_VERSION} has security patches.")
+        endif()
+    endif()
+endif()
```
seando-adsk, line 82: Added new cmake variable "ADSK_USD_VERSION" which starts out equal to normal "USD_VERSION". Detect in an automatic way if the USD build (any version <= 0.23.02) has the extra security patches. Do this by looking at the vt/array.h header file for a new method which was added in one of the patches. If so then add "-ad1" to new variable ADSK_USD_VERSION.

neilh-adsk (approved):

PR #2948, opened 2023-03-16
Fix SceneIndex library path (linux/mac need lib prefix)

seando-adsk:


lib/mayaUsd/sceneIndex/CMakeLists.txt
```diff
@@ -27,14 +27,7 @@ install(FILES ${HEADERS}
...
 
-if(IS_WINDOWS)
-    set(PLUG_INFO_LIBRARY_PATH "mayaUsd.dll")
-elseif(IS_MACOSX)
-    set(PLUG_INFO_LIBRARY_PATH "mayaUsd.dylib")
-elseif(IS_LINUX)
-    set(PLUG_INFO_LIBRARY_PATH "mayaUsd.so")
-endif()
-
+set(PLUG_INFO_LIBRARY_PATH ${CMAKE_SHARED_LIBRARY_PREFIX}${PROJECT_NAME}${CMAKE_SHARED_LIBRARY_SUFFIX})
```
seando-adsk, line 30: On Linux/Mac the library is prefixed with "lib" - this comes from CMAKE_SHARED_LIBRARY_PREFIX.

PROJECT_NAME = mayaUsd, as set here https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/CMakeLists.txt#L18

CMAKE_SHARED_LIBRARY_SUFFIX is set to dll/so/dylib by cmake.

This same pattern is used by the other .json files.

perrauo-adsk:
This won't give us the .mll extension?
CMAKE_SHARED_LIBRARY_SUFFIX

ppt-adsk (approved):

perrauo-adsk (approved):

PR #2951, opened 2023-03-17
MAYA-127758 add edit routing for attributes

pierrebai-adsk:
- Add a helper funtion to restore all edit routers to their default.
- Add a getAttrEditRouterLayer helper function to route attribute editions.
- The routing operation for attribute is named... 'attribute'
- It receives the attribute name in the routing context under the 'attribute' key.
- Add the routing operation name in the routing context under the 'operation' key.
- Declare tokens used in edit routing globally.
- This avoid typos and avoid creating them on each call.
- Use attribute routing in UFE Object3d setVisibility implementation.
- Use attribute routing in UFE attribute holder inplementation for setMetadata, set, clearMetadata.
- Use routed layer when checking editability.
- Allow cancelling routed operations by throwing an exception.
- Add unit tests for attribute routing.
- Add a unit test for canceling a command from within an edit router


vlasovi (approved):
Looks good

lib/mayaUsd/ufe/UsdAttributeHolder.cpp
```diff
@@ -26,6 +27,11 @@
 #include <pxr/base/tf/token.h>
 #include <pxr/base/vt/value.h>
 #include <pxr/pxr.h>
+#include <pxr/usd/usd/editContext.h>
+#include <pxr/usd/usd/stage.h>
+
+namespace MAYAUSD_NS_DEF {
+namespace ufe {
```
seando-adsk, line 34: I'm a bit confused why you needed to move the function setUsdAttrMetaData (which existed inside an unnamed namespace) into the ```MayaUsd::ufe``` namespace.

seando-adsk (dismissed):
Good work. Thanks for putting all those hard-coded values into tokens.

lib/mayaUsd/ufe/UsdAttributeHolder.cpp
pierrebai-adsk: It made calling function in the namespace simpler as it avoided having to put the namespaces at each call site.

seando-adsk: Our coding standard recommends to use the fully qualified namespace name. I think this is a helper function which is not exported and therefore should not be in the MayaUsd::ufe namespace.

vlasovi (approved):

lib/mayaUsd/ufe/UsdAttributeHolder.cpp
```diff
@@ -54,10 +51,10 @@ bool setUsdAttrMetadata(
     }
 
     // If attribute is locked don't allow setting Metadata.
-    enforceAttributeEditAllowed(attr);
+    MAYAUSD_NS_DEF::ufe::enforceAttributeEditAllowed(attr);
```
seando-adsk, line 54: In actual code we don't use the macro. This should just be "MayaUsd".

seando-adsk: There are some other places in the code that are also wrong, so I'm sorry if you copy-pasted one of them. These need to be cleaned up (in a separate PR).

seando-adsk (approved):

PR #2954, opened 2023-03-20
Add metersPerUnit scaling

dgovil:
This commit adds a new metersPerUnit flag to the exporter that applies distance scaling to some attributes that are known to be of a specific linear distance unit.

To start, only the following data is scaled:

1. Point positions
2. Translate xformOps

The hope is that this can be expanded over time to cover more of the exporter, but just these two alone would cover a large section of use cases. 

The default remains to write it out using the internal units, but users can opt into the behaviour by specifying any positive, non-zero value. If they do so, the exporter emits a warning notifying them that this feature is experimental.

metersPerUnit takes a float value. Any positive, non-zero value will be treated as an explicit request for that metersPerUnit. This allows for bespoke values not reflected in Maya's MDistance keys, like decimeters which some studios choose to work in.

A value of 0 represents using the default, internal value.
A value of -1 or below will have the exporter use the current set display units of the scene.

This is of value to us because our engine (and other engines/DCCs) use meters as the base linear unit, whereas Maya is centimeters. While we handle unit conversion on ingest, it'd be best if we could start moving towards keeping units consistent in the pipeline.


neilh-adsk:
Hi, I just wanted to check the intent of this work.  I see that you are scaling various elements by a factor.  I am just wondering whether this is something that needs to happen on export or be dealt with on import via the importing application.  This assumes that MayaUSD writes out the correct scale unit, which currently it ignores and always reports the default Maya units.  So although I see your solution as a workable option, I am left wondering whether we should just in fact deal with the actual units being used.  Do you have a work case where you don't want to use the units in Maya and explicitly target the units of the resulting engine ?

dgovil:
@neilh-adsk take the example of USD composition. If your studio or engine is in a different unit size than Maya, whenever you reference/sublayer in an asset created in Maya it'll lose any unit information (since layer metadata isn't transitive), thus coming in at the wrong size. 

Having Maya be able to export into the correct studio units simplifies this, because you don't need to add in-between nodes to handle scaling things to the right units. 

It also enables better swappability of the USD files. Say someone made an asset in Blender (meters), but now I need to use some tool in Maya (centimetres) to enhance it, I'd need to add some scaling transform to my composition to compensate for this difference. Every time I go back and forth, I need to do this. 

Of course, I could work in relative units by working at 100x the size in Blender, or 1/100th the size in Maya to match their scale, but that's also cumbersome. 

Having the ability to export with a scalar solves these scenarios, which unfortunately come up as a result of composition, and aren't something that can be solved by scaling on import since composition loses that information.

neilh-adsk:
@dgovil  - yeah, MayaUSD is blank ignoring all those directives - on export or import.  I see what you are saying and how this provides a more flexible solution - I just wonder what other areas you see needing to be covered for your workflows...  Thanks for this, it's a great topic and happy to see the contributions.

neilh-adsk:
Also, are you resolving an issue where units are just ignored across engines ?  I see what you say about composition but if all transactions took into account the units, wouldn't this resolve your immediate problem ?

dgovil:
@neilh-adsk could you elaborate on what you mean by transaction? I'm not sure I use the same terminology but I assume you mean on ingest to the engine?

We do apply scaling from the source stage to the target unit size. It's an alright workflow if the stage is a single unit type throughout. Though it is  confusing to customers why their Maya scene with display units set to meters exports as centimetres and then comes in with a 0.01 scaling. 

And of course, that doesn't work when they invariably mix and match DCCs. 


For our needs it's for the following two reasons :

- static mesh assets which are just a mesh with materials. No camera, lights or skeletons. 

- while we can handle that fine in an internal pipeline, we try and make it so everything we do is possible for customers (especially those without their own TDs), hence why we contribute back to this repo. 

Eventually we'd like to expand it to other asset components like cameras, lights and skeletons but I wanted to cover the most common use case first which is export of mesh assets.

neilh-adsk:
@dgovil Yes, totally.  I have been thinking more about this recently, and I'm liking the suggestion, purely for its flexibility to target different engines from the same source.  I'll bring it back to to the team, but in the meantime I will go through the review process.

dgovil:
Thanks, that's much appreciated 😃

dgovil:
I'm just checking in again and seeing if there was any updates on this?

We used this PR in our pipeline to generate all the assets that ship for Reality Composer Pro ( https://developer.apple.com/wwdc23/10083 ) to stay native to the engine scale. 

We do get a lot of Maya users asking about why their own content comes in at 0.01 scale due to Maya's forced centimetre scale. Since the majority of other common DCCs and engines use meters, the majority of these questions come from Maya users. 

It would be really great to at least point a subset of them who are making static mesh assets to a flag they could use to export as meters instead.

neilh-adsk (dismissed):

lib/mayaUsd/fileio/jobs/writeJob.cpp
neilh-adsk, line 1: Perhaps remove the word experimental and use something like "Evolving " .. :-)

lib/mayaUsd/fileio/transformWriter.cpp
```diff
@@ -156,6 +157,11 @@ void UsdMayaTransformWriter::_ComputeXformOps(
                 for (unsigned int i = 0; i < 3; i++) {
```
neilh-adsk, line 157: remove the hardcoded values ?

neilh-adsk (commented):
Looks good, glad it's been working out for you in production environments, I was just trying to think of the usage of Experimental in the code.  We need to look to try to evolve this across other code paths as well...

lib/mayaUsd/fileio/transformWriter.cpp
dgovil: I assume you mean the `3` in the for loop? That line isn't from this PR, its the pre-existing transform writer code

lib/mayaUsd/fileio/jobs/writeJob.cpp
dgovil: Yep that makes sense. I'll do a rebase to bring this up to ToT and make that wording change

lib/mayaUsd/fileio/transformWriter.cpp
neilh-adsk: My bad - must be going cloud blind in my old age...

lib/mayaUsd/fileio/jobs/writeJob.cpp
dgovil: Done with the latest push

lib/mayaUsd/fileio/transformWriter.cpp
dgovil: All good :-)

neilh-adsk (approved):
All looks good - just a comment for the documentation, but other than that, it looks good from the code side

neilh-adsk (approved):

dgovil:
Thanks very much, @neilh-adsk !

PR #2958, opened 2023-03-21
Allow material-related functions only for specific node types

stefanminning-autodesk:
# Description

This PR implements several material-related quality of life features in the Outliner and Viewport right-click menus:
- The "Assign Existing Material" option will only be shown if there are existing materials in the object's stage
- The "Unassign Material" option will only be shown if a material is bound to the object
- Material assignment options are only shown on node types where it makes sense (in accordance with UX' documentation)

Unit tests are provided for the newly added Command.

### DIFF
--- a/lib/mayaUsd/commands/Readme.md
+++ b/lib/mayaUsd/commands/Readme.md
@@ -494,6 +494,53 @@ The purpose of this command is to find the names and annotations for registered 
 | `-importArguments`     | `-ig`      | string         | Retrieve the import arguments affected by the import job context nice name passed as parameter |
 | `-jobContext`          | `-jc`      | string         | Retrieve the job context name associated with a nice name. |
 
+## `mayaUsdGetMaterialsFromRenderers`
+
+The purpose of this command is to return the names of USD-compatible materials associated with the currently loaded renderers.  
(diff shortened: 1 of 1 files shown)


plugin/adsk/plugin/adskMaterialCommands.cpp
```diff
@@ -16,7 +16,9 @@
 #include "adskMaterialCommands.h"
 
 #include <mayaUsd/ufe/Global.h>
+#include <mayaUsd/ufe/UsdSceneItem.h>
```
JGamache-autodesk, line 19: Please add documentation for the material commands [here](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/commands/Readme.md).

JGamache-autodesk (dismissed):
LGTM. Would be even better with documentation.

vlasovi (approved):
Looks good

seando-adsk:
@stefanminning-autodesk This PR has the do-not-merge label - is that correct? The code review and preflight are done, so it appears ready to go in.

vlasovi (approved):

JGamache-autodesk (approved):

PR #2959, opened 2023-03-21
MAYA-128463 transform command return null.

pierrebai-adsk:
The creation of transform command cannot throw exception because Maya is not ready to handle them. Since we cannot change Maya at this point revert to the old behaviour of returning a null command.


pierrebai-adsk:
Note: this revert changes I had made in a previous PR.

vlasovi (approved):

PR #2960, opened 2023-03-22
Preparation for Ufe v4.1

seando-adsk:

PR #2962, opened 2023-03-22
Added size methods to UINodeGraphNode

hewettc-adsk:
Implemented `size` methods in `UINodeGraphNode` and added tests.


seando-adsk (changes requested):
Carl, please wait for my PR #2960 to be merged in (my preflight is running now). Once that is done you should merge in latest dev and then I would suggest you squash your changes down to a single commit to clean it up. That should remove the extra commits from Dan and I (which should not be part of your PR). You also might want to consider closing this PR and starting with a fresh one once the commits are cleaned up.

PR #2963, opened 2023-03-22
Assign to component tags

JGamache-autodesk:
Allows assigning materials to component tags
Will set family to "materialBind" if required
That is part 1 of 2. The second part requires being able to roundtrip these assignments thru import/export without gaining or losing any component tags.


lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -735,7 +735,7 @@ static const std::vector<MayaUsd::ufe::SchemaTypeGroup> getConcretePrimTypes(boo
 bool sceneItemSupportsShading(const Ufe::SceneItem::Ptr& sceneItem)
 {
 #if PXR_VERSION >= 2108
-    if (MayaUsd::ufe::BindMaterialUndoableCommand::CompatiblePrim(sceneItem).IsValid()) {
+    if (MayaUsd::ufe::BindMaterialUndoableCommand::CompatiblePrim(sceneItem)) {
```
JGamache-autodesk, line 738: Some signatures were changed. This one now only returns a boolean.

lib/mayaUsd/ufe/UsdContextOps.cpp
```diff
@@ -1239,24 +1239,23 @@ Ufe::UndoableCommand::Ptr UsdContextOps::doOpCmd(const ItemPath& itemPath)
 #endif
 #if PXR_VERSION >= 2108
     else if (itemPath[0] == BindMaterialUndoableCommand::commandName) {
-        return std::make_shared<BindMaterialUndoableCommand>(fItem->prim(), SdfPath(itemPath[1]));
+        return std::make_shared<BindMaterialUndoableCommand>(fItem->path(), SdfPath(itemPath[1]));
```
JGamache-autodesk, line 1242: And the constructor now takes a Ufe path.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -87,126 +88,100 @@ bool connectShaderToMaterial(
...
-        return usdPrim;
+        return true;
     }
-    return {};
+    return false;
 }
 
 BindMaterialUndoableCommand::BindMaterialUndoableCommand(
-    const UsdPrim& prim,
+    Ufe::Path      primPath,
```
JGamache-autodesk, line 120: I was about to say that keeping a Ufe::Path was better than keeping a Stage/SdfPath for undo/redo, but then noticed that since I modernized to use a UsdUndoBlock, this is not relevant anymore.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -87,126 +88,100 @@ bool connectShaderToMaterial(
...
-        if (prim.HasAPI<UsdShadeMaterialBindingAPI>()) {
-            bindingAPI = UsdShadeMaterialBindingAPI(prim);
-            _previousMaterialPath = bindingAPI.GetDirectBinding().GetMaterialPath();
-        } else {
-            bindingAPI = UsdShadeMaterialBindingAPI::Apply(prim);
-            _appliedBindingAPI = true;
-        }
-        bindingAPI.Bind(material);
+    if (auto subset = UsdGeomSubset(prim)) {
+        subset.GetFamilyNameAttr().Set(UsdShadeTokens->materialBind);
```
JGamache-autodesk, line 152: That was the only required change to make assignment work and have the material display in the viewport.

test/lib/ufe/testContextOps.py
```diff
@@ -312,7 +313,6 @@ def testAddNewPrim(self):
         cmds.file(new=True, force=True)
 
         # Create a proxy shape with empty stage to start with.
-        import mayaUsd_createStageWithNewLayer
         proxyShape = mayaUsd_createStageWithNewLayer.createStageWithNewLayer()
```
pierrebai-adsk, line 316: TY, it annoyed me too in past ;)

test/lib/ufe/testContextOps.py
```diff
@@ -1476,5 +1466,31 @@ def testAssignExistingMaterialToSingleObject(self):
...
+        self.assertEqual(topSubset.GetFamilyNameAttr().Get(), "componentTag")
+        self.assertFalse(topSubset.GetPrim().HasAPI(UsdShade.MaterialBindingAPI))
+
+        contextOps = ufe.ContextOps.contextOps(topItem)
+        cmd = contextOps.doOpCmd(['Assign New Material', 'USD', 'UsdPreviewSurface'])
+        self.assertIsNotNone(cmd)
+        ufeCmd.execute(cmd)
+
+        self.assertEqual(topSubset.GetFamilyNameAttr().Get(), "materialBind")
+        self.assertTrue(topSubset.GetPrim().HasAPI(UsdShade.MaterialBindingAPI))
```
pierrebai-adsk, line 1492: I guess it was discussed internally, but i'm not sure what is better: change the name of the set or create a new set? The family name, according to USD, is used to group tags that are related, so changing it changes the group-of-subset composed of with members with the same family name.

In this case we're the one generating the componentTag, but it feels bad to now have 5 side with componentTag family name and one with the material binding one.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -87,126 +88,100 @@ bool connectShaderToMaterial(
...
+void BindMaterialUndoableCommand::undo() { _undoableItem.undo(); }
+
+void BindMaterialUndoableCommand::redo() { _undoableItem.redo(); }
+
+void BindMaterialUndoableCommand::execute()
 {
-    if (!_stage || _primPath.IsEmpty() || _materialPath.IsEmpty()) {
+    if (_primPath.empty() || _materialPath.IsEmpty()) {
         return;
     }
```
pierrebai-adsk, line 137: IDk if it warrants a warning or some other user feedback when the arguments are not valid?

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -87,126 +88,100 @@ bool connectShaderToMaterial(
...
-            }
-        }
-        if (_appliedBindingAPI) {
-            prim.RemoveAPI<UsdShadeMaterialBindingAPI>();
-        }
+    UsdUndoBlock undoBlock(&_undoableItem);
+
+    auto prim = ufePathToPrim(_primPath);
+    if (!prim.IsValid()) {
+        return;
```
pierrebai-adsk, line 143: Same comment about all these early returns vs giving user feedback.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
pierrebai-adsk: OK, but as I asked, IDk what are the implication of changing the family name on an existing subset.

pierrebai-adsk (commented):
I would only want to know the rationale behind the decision to modify family names before approving.

pierrebai-adsk:
My question was not about the name itself, but the decision to modify the subset family name instead of creating a new subset with a different family name.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
JGamache-autodesk: If the error handling scheme is still `CTOR throws if command cannot execute` then I agree this test should be moved to the constructor and throw accordingly. The execute() function should assume these are valid. Will fix that this afternoon.

pierrebai-adsk (dismissed):

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -417,6 +434,15 @@ void UsdUndoAssignNewMaterialCommand::execute()
                 markAsFailed();
                 return;
             }
+            // There might be some unassignable items in the selection list. Skip and warn.
+            // We know there is at least one assignable item found in the ContextOps resolver.
+            if (!BindMaterialUndoableCommand::CompatiblePrim(parentItem)) {
+                const std::string error = TfStringPrintf(
+                    "Assign new material: Skipping incompatible prim [%s] found in selection.",
+                    Ufe::PathString::string(parentItem->path()).c_str());
```
JGamache-autodesk, line 442: Found in the unit tests where the first new material ended up in the selection list for the second assign new material call.

pierrebai-adsk (approved):

PR #2964, opened 2023-03-23
MAYA-114041 - As a user, I'd like my pulled Maya object to inherit visibility

vlasovi:
Add a custom attribute "combinedVisibility" to proxy accessors to output combined visibility to maya pulled prims

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -85,6 +85,7 @@
 #include <maya/MFnUnitAttribute.h>
 #include <maya/MGlobal.h>
 #include <maya/MItDependencyNodes.h>
+#include <maya/MNodeMessage.h>
 #include <maya/MObject.h>
 #include <maya/MPlug.h>
 #include <maya/MPlugArray.h>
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -2024,6 +2029,68 @@ bool MayaUsdProxyShapeBase::closestPoint(
...
+
+    // Add our own callback
+    MObject obj = thisMObject();
+    _ancestorCallbacks.push_back(
+        MNodeMessage::addNodeDirtyCallback(obj, _proxyShapeAncestorDirty, this));
+
+    // Remember the path for which we are accumulating the callbacks
+    MDagPath ancestorPath;
+    MDagPath::getAPathTo(obj, ancestorPath);
+    _ancestorCallbacksPath = ancestorPath.fullPathName();
```
pierrebai-adsk, line 2061: What will happen if the proxy shape node is renamed? Or any of its ancestors are renamed?

vlasovi: I just tested: renaming proxy shape node or any of its DAG ancestors works fine, as in this case we call updateAncestorCallbacks on the next dirty callback. However, after renaming a USD ancestor of a pulled node we loose the connection inside ProxyAcessors and this is a general problem that exists for all proxy accessors, including relaying the 3d transform. I think we may need to create a Jira ticket for that one.

pierrebai-adsk (approved):

pierrebai-adsk (approved):

PR #2965, opened 2023-03-23
MAYA-128460 - undoing rename of prim causes prim's icon to change to def

seando-adsk:
#### MAYA-128460 - undoing the rename of a prim causes prim's icon to change to a def

* For undo send the correct (new) prim in the Ufe notification.
* Refactor undo/redo into common helper method.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
@@ -222,7 +222,7 @@ void renameHelper(
     sendNotificationToAllStageProxies(stage, ufeDstItem->prim(), srcPath, dstPath);
 }
 
-} // namespace 
+} // namespace
 
 void UsdUndoRenameCommand::renameRedo()
 {


lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -113,7 +113,9 @@ UsdUndoRenameCommand::create(const UsdSceneItem::Ptr& srcItem, const Ufe::PathCo
 
 UsdSceneItem::Ptr UsdUndoRenameCommand::renamedItem() const { return _ufeDstItem; }
 
-static void sendNotificationToAllStageProxies(
+namespace {
```
seando-adsk, line 116: Put all these helpers into unnamed namespace.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -190,67 +192,63 @@ static void doUsdRename(
...
-    auto defaultPrimPath = _stage->GetDefaultPrim().GetPath();
-
-    const Ufe::Path srcPath = _ufeDstItem->path();
-    const Ufe::Path dstPath = _ufeSrcItem->path();
-    // Note: must fetch prim again from its path because undo/redo of composite commands
-    //       (or doing multiple undo and then multiple redo) can make the cached prim
-    //       stale.
-    const UsdPrim     prim = _stage->GetPrimAtPath(_ufeDstItem->prim().GetPath());
-    const std::string newName = _ufeSrcItem->prim().GetName();
+    const Ufe::Path srcPath = _ufeSrcItem->path();
```
pierrebai-adsk, line 234: I think scene item return teh path by ref, so this could be a &, to verify.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -190,67 +192,63 @@ static void doUsdRename(
...
+    // If the new name is the same as the current name, do nothing.
+    // This is the same behavior as the Maya rename command for Maya nodes.
+    if (_newName.empty())
+        return;
 
-    // update stage's default prim
-    if (_ufeDstItem->prim().GetPath() == defaultPrimPath) {
-        _stage->SetDefaultPrim(_ufeSrcItem->prim());
-    }
+    const Ufe::Path   srcPath = _ufeDstItem->path();
```
pierrebai-adsk, line 247: Same comment as above, could be const&

pierrebai-adsk (approved):
Minor tweaks possible, otherwise LGTM.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
```diff
@@ -190,67 +192,63 @@ static void doUsdRename(
     applyToAllPrimSpecs(prim, renameFunc);
 }
 
-void UsdUndoRenameCommand::renameRedo()
+void renameHelper(
+    PXR_NS::UsdStageWeakPtr  stage,
```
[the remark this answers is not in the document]
seando-adsk: Right good point.

lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp
seando-adsk: Good point, changed to auto so we don't need to worry about return type.

pierrebai-adsk (dismissed):
LGTM

PR #2971, opened 2023-03-24
Check skin cluster weights

tryptik:
If an artist uses polyReduce (or other node that changes the vertex count) after a skinCluster has been applied, the Usd exporter tries to compare the weights of the original vertex count with the assigned weights of the reduced vertex count. This results in an out-of-bounds access to the MDoubleArray that holds the weights.

This P/R checks for that condition and throws an exception if encountered (this occurs in a part of the plugin that is not using `MStatus` to signal success/failure/error).

### DIFF
--- a/lib/mayaUsd/fileio/utils/jointWriteUtils.cpp
+++ b/lib/mayaUsd/fileio/utils/jointWriteUtils.cpp
@@ -237,6 +237,16 @@ int UsdMayaJointUtil::getCompressedSkinWeights(
     unsigned int numInfluences;
     skinCluster.getWeights(outputDagPath, components.object(), weights, numInfluences);
 
+    if(weights.length() < numVertices * numInfluences) {
+        MString msg("Number of weights (");
+        msg += weights.length();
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/fileio/utils/jointWriteUtils.cpp
```diff
@@ -237,6 +237,16 @@ int UsdMayaJointUtil::getCompressedSkinWeights(
     unsigned int numInfluences;
     skinCluster.getWeights(outputDagPath, components.object(), weights, numInfluences);
 
+    if (weights.length() < numVertices * numInfluences) {
+        MString msg("Number of weights (");
+        msg += weights.length();
+        msg += ") exceeds number of influences * number of vertices (";
+        msg += numInfluences * numVertices;
+        msg += ") - skin binding will not be written";
+        MGlobal::displayError(msg);
```
seando-adsk, line 246: Would it be possible to add to this message a way to fix the problem or a workaround for problem?

seando-adsk (approved):

PR #2972, opened 2023-03-24
Add smart signaling for renderers.

JGamache-autodesk:
A renderer now only has to track the "updateid" and "resyncid" attributes of the stage to know if something significant happened.

All UINodeGraphNodeAPI notifications were suppressed.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -2092,8 +2092,9 @@ void MayaUsdProxyShapeBase::_OnStageObjectsChanged(const UsdNotice::ObjectsChang
 
     switch (_ClassifyUsdObjectsChanged(notice)) {
     case _UsdChangeType::kIgnored: return;
+    case _UsdChangeType::kResync: ++_UsdStageResyncCounter;
+    // [[fallthrough]]; // We want that fallthrough to have the update always triggered.
     case _UsdChangeType::kUpdate: ++_UsdStageUpdateCounter; break;
-    case _UsdChangeType::kResync: ++_UsdStageResyncCounter; break;
     }
 
     // Recompute the extents of any UsdGeomBoundable that has authored extents
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -50,17 +50,20 @@
 #include <pxr/usd/ar/resolver.h>
 #include <pxr/usd/sdf/attributeSpec.h>
 #include <pxr/usd/sdf/layer.h>
+#include <pxr/usd/sdf/listOp.h>
 #include <pxr/usd/sdf/path.h>
 #include <pxr/usd/usd/editContext.h>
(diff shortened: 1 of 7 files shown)


lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -388,6 +394,26 @@ MStatus MayaUsdProxyShapeBase::initialize()
     retValue = addAttribute(outStageDataAttr);
     CHECK_MSTATUS_AND_RETURN_IT(retValue);
 
+    updateCounterAttr
+        = numericAttrFn.create("updateid", "upid", MFnNumericData::kInt64, -1, &retValue);
```
pierrebai-adsk, line 398: Very minor: other attribute use camelCase. To follow that style this could be updateId and the other resyncId. I really don't think this alone is worth respinning a new version.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -1923,6 +1965,118 @@ void MayaUsdProxyShapeBase::_OnStageEditTargetChanged(
...
+    for (PathRange::const_iterator it = pathsToUpdate.begin(); it != pathsToUpdate.end(); ++it) {
+        if (it->IsAbsoluteRootOrPrimPath()) {
+            const TfTokenVector changedFields = it.GetChangedFields();
+            if (!changedFields.empty()) {
+                retVal = _UsdChangeType::kUpdate;
+            }
+        } else if (it->IsPropertyPath()) {
+            // We have a bunch of UI properties to ignore. Especially anything that comes from UI
+            // schemas.
+            if (it->GetName().rfind("ui:", 0) == 0) {
```
pierrebai-adsk, line 2062: If you refactor into another file, this code seems similar to code above. A helper function might be nice to abstract detecting UI fields.

pierrebai-adsk (approved):
I'll let you decide if you want to follow my suggestion or not. Eventually, we will need to refactor proxy shape into multiple files.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -1923,6 +1965,118 @@ void MayaUsdProxyShapeBase::_OnStageEditTargetChanged(
...
+{
+    kIgnored, // Change does not require redraw: UI change, metadata change.
+    kUpdate,  // Change requires redraw after refreshing parameter values
+    kResync   // Change requires refreshing cached buffers
+};
+
+// If the notification is about prepending a UI schema, we don't want a refresh. These structures
+// are quite large to inspect, but they hash easily, so let's compare known hashes.
+bool _IsUiSchemaPrepend(const VtValue& v)
+{
```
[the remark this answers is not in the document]
JGamache-autodesk: Yes, we are very specific at this moment and well aware of how MayaUSD proceeds, so we know in this case we do not need to support append or assign. Things might (and will) change so we will have to adapt.

JGamache-autodesk:
Preflight error is a traditional `testUsdExportPackage` random failure.

JGamache-autodesk:
@seando-adsk Ready for merge. Thanks!

PR #2974, opened 2023-03-28
gcc11 warnings

seando-adsk:


lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -186,7 +186,6 @@ VtValue _BuildCubicIndexArray(const HdBasisCurvesTopology& topology)
     }
 
     int vertexIndex = 0;
-    int curveIndex = 0;
```
seando-adsk, line 189: Removed unused variables (same as below).

AramAzhari-adsk (approved):

Issue #2953, opened 2023-03-20
Rename operation spans references and inherits


dj-mcg:
**Describe the bug**
Rename operations fail when applied to a prim that references a show asset because the rename operation is attempted on the reference and classes it inherits.

**Steps to reproduce**
1. Create a test asset that inherits from one or more classes
2. Create a prim within a Maya Usd scene that references this asset
3. Attempt to rename the prim
4. Observe the failure due to Sdf policy

**Expected behavior**
The rename operation only applies to the defined prim in the authoring layer and any overs in session layers.

**Specs (if applicable):**
 - Linux RHEL 7.9
 - Maya 2022
 - Maya USD commit SHA [dev at 5c01103]
 - Pixar USD commit SHA [dev at 84bd5a9]

**Additional context**
It looks like [applyToAllPrimSpecs](https://github.com/Autodesk/maya-usd/blob/e87e10dab08af4364f38e87f74bc41c2e6f0c9a1/lib/mayaUsd/ufe/UsdUndoRenameCommand.cpp#L190) may be a little greedy for this operation.

dj-mcg:
I have a potential fix for this - PR should land later today

santosg87:
@dj-mcg is it ok to close this issue for now?

dj-mcg:
Yeah this fixes our use cases, thanks!

PR #2977, opened 2023-03-28
Restrict namespace edits to current stage

dj-mcg:
This will restrict namespace edits to your current usd stage. This avoids the issue of spanning references and inherits, etc, which will often be illegal operations. This addresses the issue I posted here:

https://github.com/Autodesk/maya-usd/issues/2953

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeUpdateManager.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeUpdateManager.cpp
@@ -20,6 +20,8 @@
 #include <pxr/usd/usd/tokens.h>
 #include <pxr/usd/usdUI/tokens.h>
 
+#include <mutex>
+
 using namespace MAYAUSD_NS_DEF;
 
 PXR_NAMESPACE_OPEN_SCOPE
--- a/lib/mayaUsd/utils/layers.cpp
+++ b/lib/mayaUsd/utils/layers.cpp
@@ -124,6 +124,17 @@ static SdfPrimSpecHandleVector _GetLocalPrimStack(const UsdPrim& prim)
     if (!stage)
         return primSpecs;
 
+    // The goal is to avoid editing non-local layers. This issue is,
+    // for example, that a rename operation would fail when applied
+    // to a prim that references a show asset because the rename operation
+    // would be attempted on the reference and classes it inherits.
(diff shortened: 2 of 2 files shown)


lib/mayaUsd/nodes/proxyShapeUpdateManager.cpp
```diff
@@ -20,6 +20,8 @@
 #include <pxr/usd/usd/tokens.h>
 #include <pxr/usd/usdUI/tokens.h>
 
+#include <mutex>
```
pierrebai-adsk, line 23: No sure why this file got modified, was this to fix a build issue?

pierrebai-adsk (dismissed):
It's always hard to imagine all possible scenarios that the change could affect, but the principle seems correct, so this is good to go. If we find cases that need to modify non-local layer stacks in the future, we'll revisit the design and API.

lib/mayaUsd/utils/layers.cpp
```diff
@@ -116,16 +116,33 @@ void enforceMutedLayer(const PXR_NS::UsdPrim& prim, const char* command)
     }
 }
 
+static SdfPrimSpecHandleVector _GetLocalPrimStack(const UsdPrim& prim)
```
pierrebai-adsk, line 119: ON second thought, maybe adding a comments on what this code is trying to avoid, like the description you gave in teh bug would be helpful for future coders.

pierrebai-adsk (approved):

lib/mayaUsd/utils/layers.cpp
pierrebai-adsk: I too the liberty of adding the comment myself to your branch.

lib/mayaUsd/nodes/proxyShapeUpdateManager.cpp
dj-mcg: Yeah, I couldn't build locally without that change. It's PR'd here:
https://github.com/Autodesk/maya-usd/pull/2975

lib/mayaUsd/utils/layers.cpp
dj-mcg: Thank you!!

PR #2982, opened 2023-03-28
Update imaging baselines

dj-mcg:
Baseline update due to subtle changes in usd imaging


JGamache-autodesk (approved):
LGTM, looks like a better antialiasing setting.

PR #2983, opened 2023-03-28
Remove explicit call to sceneIndex::Populate

dj-mcg:
In most recent USD, Populate is called from within SetStage, so there is no need to callsites to call it explicitly


perrauo-adsk (changes requested):
Why the header inclusion?

dj-mcg:
> Why the header inclusion? Seems unrelated no?

Because I couldn't build without it. See:
https://github.com/Autodesk/maya-usd/pull/2975

perrauo-adsk (approved):

perrauo-adsk:
@dj-mcg Could you point out to the Pixar change that fixes this. This removal is now causing issues on my end.

dj-mcg:
My apologies! I had assumed that change would land with 2305 - thanks for the fix

PR #2986, opened 2023-03-29
MAYA-126040 improve preventing edits

pierrebai-adsk:
- Add documentations about edit routing.
- Make attribute editing not throw exception for now to avoid possibly destabilizing Maya.
- Instead, attribute commands return null command when edits are prevented.
- Make the commands use the pre-delcraed USD tokens (code cleanup)
- Make the visibility command always apply routing by moving routing code in its constructor.
- Add more unit test for edit routing, testing more commands.
- Add more unit tests for command prevention


vlasovi (approved):
LGTM

PR #2990, opened 2023-03-29
Updated Sub layer dialog to display the correct relative path when enabled.

AramAzhari-adsk:


plugin/adsk/scripts/mayaUsd_USDRootFileRelative.py
```diff
@@ -240,7 +241,12 @@ def updateFilePathPreviewFields(cls, makePathRelative=None):
             # the user is free to enter any extension they want. The layer editor code will then verify
             # (and fix if needed) the file path before saving. We do the same here for preview.
             unresolvedPath = mayaUsdLib.Util.ensureUSDFileExtension(selectedFile) if selectedFile else ''
-            relativePath = mayaUsdLib.Util.getPathRelativeToMayaSceneFile(unresolvedPath) if unresolvedPath else ''
+            """relativePath = mayaUsdLib.Util.getPathRelativeToDirectory(unresolvedPath, cls._parentLayerPath) if unresolvedPath and cls._parentLayerPath != "" else ''"""
```
seando-adsk, line 244: Was this commented code replaced by below? If so it should be removed.

seando-adsk (changes requested):

seando-adsk:
@pierrebai-adsk Can you have a look at these changes as well? You have done most of the relative path work.

plugin/adsk/scripts/mayaUsd_USDRootFileRelative.py
AramAzhari-adsk: removed the comment

seando-adsk (approved):

lib/mayaUsd/python/wrapUtil.cpp
```diff
@@ -56,14 +56,16 @@ std::string ensureUSDFileExtension(const std::string& fileToCheck)
...
-                  .staticmethod("prettifyName")
-                  .def("getDictionaryFromEncodedOptions", getDictionaryFromEncodedOptions)
-                  .def(
-                      "getPathRelativeToMayaSceneFile",
-                      UsdMayaUtilFileSystem::getPathRelativeToMayaSceneFile)
-                  .def("ensureUSDFileExtension", ensureUSDFileExtension)
-                  .staticmethod("getPathRelativeToMayaSceneFile");
+    scope s
+        = class_<UsdMayaUtilScope>("Util", no_init)
+              .def("IsAuthored", UsdMayaUtil::IsAuthored)
```
pierrebai-adsk, line 61: For the life of me, I cannot figure out why clang-format decided all this needed to be reformatted. The rules are so obscure!

lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -174,6 +174,23 @@ std::pair<std::string, bool> UsdMayaUtilFileSystem::makePathRelativeTo(
...
 
+std::string UsdMayaUtilFileSystem::getPathRelativeToDirectory(
+    const std::string& fileName,
+    const std::string& relativeToDir)
+{
+    auto relativePathAndSuccess = makePathRelativeTo(fileName, relativeToDir);
+
+    if (!relativePathAndSuccess.second) {
+        TF_WARN(
+            "File name (%s) cannot be resolved as relative to its parent layer directory (%s), "
```
pierrebai-adsk, line 185: This function is general, so I would think we should not mention "parent layer" in the error.

AramAzhari-adsk:
@pierrebai-adsk I took notes and will keep a commit to be checked in in the future alongside refactors / relative path work. 
Thanks.

PR #2994, opened 2023-03-30
Recognize string properties with IsAssetIdentifier as being filename

JGamache-autodesk:


lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -477,18 +477,26 @@ std::string UsdAttributeFilename::get() const
...
     }
 
     return std::string();
 }
 
 void UsdAttributeFilename::set(const std::string& value)
 {
-    SdfAssetPath path(value);
-    setUsdAttr<PXR_NS::SdfAssetPath>(*this, path);
+    if (UsdAttribute::nativeType() == "string") {
```
seando-adsk, line 494: Ideally this should be the constant Ufe::Attribute::kString

JGamache-autodesk: It would work, however nativeType is supposed to return a USD type as a string, so I have to use this longer constant that ends up having exactly the same value as Ufe::Attribute::kString.

pierrebai-adsk (approved):

pierrebai-adsk (approved):

PR #2997, opened 2023-03-31
MAYA128513 edit routing of composite commands

pierrebai-adsk:
Add a context classes that reuse previous edit routing before doing a new edit routing.

- One is OperationEditRouterContext, used to route commands.
- The other is AttributeEditRouterContext, used to route attribute modifications.
- This is done by keeping track of edit routing contexts higher in the call stack.
- Remove the EditTargetGuard class since it was doing the same work as USD UsdEditContext class.
- Add documentation about how to route a custom composite command.

Modified the commands that were routing edits to use the new EditRoutingContext. This makes the code using them clearer.
- Fetch the current edit target via the stage, not the context, since the context may not have a target layer if the operation was not routed.

Correctly support absence of edit router. This happens for custom commands that do not have built-in default edit router. It also makes the code more general, consistent and safe.

- Make the edit router lookup functions always return a null layer handler when there is no edit router.
- Make these same functions also return null if the router does not route to a layer.
- This allows the caller to know if the editing is not routed.
- This is necessary to make nested routing work.
- Make the edit router context use the fact that the layer is null to know it is not being routed.
- Allow removing an edit router of a custom command in the restoreDefaultEditRouter function.
- Make the registered edit routers container be initialized dynamically by wrapping it in a function.

Add unit tests:
- Added a unit test for composite command edit routing.
- The unit test serves as an example of how to route composite commands.
- The unit test validates the Python binding.

### DIFF
--- a/doc/EditRouting.md
+++ b/doc/EditRouting.md
@@ -160,6 +160,71 @@ def preventCommandEditRouter(context, routingData):
     raise Exception('Sorry, %s is not permitted' % opName)
 ```
 
+## Routing custom composite commands
+
+When writing composite commands, that is a command made-up of sub-commands,
+it is possible to route the whole command. This will route all sub-commands
+to the same destination layer and ignore individual routing of each individual
+sub-command.
+
+To do this, one must choose a new and unique operation name for the composite
+command. We suggest that you use some unique prefix for your operation names
+to avoid conflict with future operation the Maya team might add in the future.
+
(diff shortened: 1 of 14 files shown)


vlasovi (dismissed):
Looks good

pierrebai-adsk:
I did not notice that the last ecp assignment toggle had not worked, starting the PF now.

vlasovi (approved):

PR #3002, opened 2023-04-03
MAYA-127954 - As a user, I'd like to drag and drop a sublayer that has a relative path

vlasovi:
When a sublayer that has a relative path is reparented, we reparent its path as well.


pierrebai-adsk (approved):
LGTM

pierrebai-adsk (approved):

PR #3004, opened 2023-04-03
MAYA-128199 fix error on save

pierrebai-adsk:
The problem was that when saving an anonynous layer, the stage changes since the root layer changes identity. We need to set the target layer, but not on the current stage, but on the new computed stage.

Also, I added error messages when a save operation fails.

- Add a helper function to retrieve the proxy shape node from its node name.
- Pass the new layer to the helper function that sets the new proxy path of the proxy shape.
- Make sure that all layer info contain the proxy shape path since it is needed to update the target layer when a target anonymous layer is saved.
- Renamed the saveRootLayer function to updateRootLayer since it is not saving the layer, it is updating the proxy shape.


vlasovi (approved):
Looks good

PR #3006, opened 2023-04-04
HYDRA-300: Update proxyShapeSceneIndexPlugin.cpp

perrauo-adsk:
Came as a result of this
https://github.com/Autodesk/maya-usd/pull/2983

Implicit Populate was only added as part of patch version 2305, the code was added here
https://github.com/PixarAnimationStudios/USD/commit/3091ceaecd6d6dfc574199c2500594bade84939c

The PXR_VERSION used is incorrect it will break versions predating the patch version 2305.
https://github.com/PixarAnimationStudios/USD/blob/3091ceaecd6d6dfc574199c2500594bade84939c/cmake/defaults/Version.cmake

Should be fixed like this
```
#if PXR_VERSION <= 2305
                // In most recent USD, Populate is called from within SetStage, so there is no need to callsites to call it explicitly
                _usdImagingStageSceneIndex->Populate();
#endif
```

### DIFF
--- a/lib/mayaUsd/sceneIndex/proxyShapeSceneIndexPlugin.cpp
+++ b/lib/mayaUsd/sceneIndex/proxyShapeSceneIndexPlugin.cpp
@@ -133,7 +133,7 @@ void MayaUsdProxyShapeSceneIndex::Populate()
             // care of inside MayaUsdProxyShapeSceneIndex::_StageSet callback.
             {
                 _usdImagingStageSceneIndex->SetStage(stage);
-#if PXR_VERSION <= 2302
+#if PXR_VERSION <= 2305
                 _usdImagingStageSceneIndex->Populate();
 #endif
                 _populated = true;


seando-adsk (approved):

dj-mcg:
Hey folks - just circling back on this. It looks like the public `UsdImagingStageSceneIndex::Populate` function will indeed be going away with the Usd 23.05 release (scheduled to land soon). I'm not sure the details of what failed with my change that referenced 2302, but perhaps the correct check is:

`#if PXR_VERSION < 2305`

I needed to make this change locally in order to get things to build with latest USD, but I'll hold off pushing this change until we decide what the right check is. Be warned that the current code will fail to compile with the upcoming release of Usd 23.05

perrauo-adsk:
@dj-mcg Yes, I realized this too. This is the reason for this change.

PR #3015, opened 2023-04-06
Smart signaling must emit MNodeMessages

JGamache-autodesk:
In order for clients to react to changes, the smart signalling should send either NodeDirty or NodeAttributeChanged to signal clients that they need to update.


test/lib/ufe/testContextOps.py
```diff
@@ -62,6 +64,26 @@ def reset(self):
...
+        self.attributeChangedCbId = om.MNodeMessage.addAttributeChangedCallback(node, self.attributeChanged, None)
+        self.attributeChangedReceived = {}
+
+    def terminate(self):
+        om.MMessage.removeCallback(self.nodeDirtyCbId)
+        om.MMessage.removeCallback(self.attributeChangedCbId)
+
+    def nodeDirty(self, node, plug, listeners):
+        name = plug.partialName(False, False, False, False, False, True)
+        self.nodeDirtyReceived.setdefault(name, [0,])[0] += 1
```
pierrebai-adsk, line 80: That some bit-obscure Python trick. defaultdict would be clearer and avoid those [0] all over, of good old if/else :)

https://docs.python.org/3/library/collections.html#collections.defaultdict

pierrebai-adsk (dismissed):

lib/mayaUsd/nodes/proxyShapeListenerBase.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+        MDataHandle inHandle = dataBlock.inputValue(stageCacheIdAttr, &retValue);
+        CHECK_MSTATUS_AND_RETURN_IT(retValue);
+
+        int cacheIdNum = inHandle.asInt();
+
+        if (cacheIdNum != _lastKnownStageCacheId) {
+            const auto cacheId = UsdStageCache::Id::FromLongInt(cacheIdNum);
+            const auto stageCached
+                = cacheId.IsValid() && UsdUtilsStageCache::Get().Contains(cacheId);
```
pierrebai-adsk, line 158: Don't we have multiple stage cache in MayaUSD, for shared/unshared stages, and payload? See stageCache.h.

test/lib/ufe/testUINodeGraphNode.py
```diff
@@ -53,14 +53,18 @@ def setUp(self):
...
+
         # Clear selection to start off
         cmds.select(clear=True)
 
+    def tearDown(self):
+        self.messageHandler.terminate()
+
     # Helper to avoid copy-pasting the entire test
     def doPosAndSizeTests(self, hasFunc, setFunc, getFunc, cmdFunc):
-        initialUpdateCount = cmds.getAttr('|transform1|proxyShape1.updateId')
```
pierrebai-adsk, line 61: Note; a fix was already merged in dev. You might want to merge dev in your branch.

pierrebai-adsk (changes requested):
I'm not sure about the stage caches and I think you will need to merge dev in to your branch, so I will wait for both of these before approving.

lib/mayaUsd/nodes/proxyShapeListenerBase.cpp
JGamache-autodesk: I am aware of these caches now. But it looks like our clients prefer a cache accessible directly via USD APIs. This is also the cache used by the proxyShape for its `outStageCacheId` output, so I am not introducing anything new here.

test/lib/ufe/testUINodeGraphNode.py
JGamache-autodesk: Yes, I have merged with latest dev as I need the recent leak fix.

pierrebai-adsk (approved):
Great!

lib/mayaUsd/nodes/proxyShapeListenerBase.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+#include <pxr/usd/usdUtils/stageCache.h>
+
+#include <maya/MDataBlock.h>
+#include <maya/MDataHandle.h>
+#include <maya/MFnNumericAttribute.h>
+
+#include <mutex>
+
+using namespace MAYAUSD_NS_DEF;
```
seando-adsk, line 26: The macro is only for declaring the namespace so that it gets versioned.. When using the namespace, just use the name "MayaUsd" which will never change.

seando-adsk (approved):

PR #3025, opened 2023-04-17
MAYA-128070 - Update for USD v23.02

seando-adsk:
#### MAYA-128070 - Update for USD v23.02
* Add Maya location to LD_LIBRARY_PATH for executable tests.

### DIFF
--- a/cmake/test.cmake
+++ b/cmake/test.cmake
@@ -218,6 +218,13 @@ finally:
     if(IS_WINDOWS)
         list(APPEND MAYAUSD_VARNAME_PATH "${CMAKE_INSTALL_PREFIX}/lib/gtest")
         list(APPEND MAYAUSD_VARNAME_PATH "${MAYA_LOCATION}/bin")
+    elseif(IS_LINUX)
+        if(PREFIX_COMMAND)
+            # Necessary for tests like DiffCore to find python lib since the test
+            # is an exectuable and running on its own (not with maya/mayapy).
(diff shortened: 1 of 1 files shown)


cmake/test.cmake
```diff
@@ -218,6 +218,13 @@ finally:
     if(IS_WINDOWS)
         list(APPEND MAYAUSD_VARNAME_PATH "${CMAKE_INSTALL_PREFIX}/lib/gtest")
         list(APPEND MAYAUSD_VARNAME_PATH "${MAYA_LOCATION}/bin")
+    elseif(IS_LINUX)
+        if(PREFIX_COMMAND)
+            # Necessary for tests like DiffCore to find python lib since the test
+            # is an exectuable and running on its own (not with maya/mayapy).
+            # See DYLD_LIBRARY_PATH case below.
+            list(APPEND MAYAUSD_VARNAME_LD_LIBRARY_PATH "${MAYA_LOCATION}/lib")
```
seando-adsk, line 226: I needed to add this for the bump of Maya PR to USD v23.02. I'm not sure why but a few of the tests (such as DiffCore which is a standalone executable test) cannot find python3.10.so (which comes from Maya). So I added the Maya lib folder to LD_LIBRARY_PATH for these types of test only (have COMMAND set). The same thing was done for Mac a while ago.

boudrey: Should not be necessary since my last fix in Maya, but it might be another situation so if it is a simple fix like that, it is ok.

PR #3027, opened 2023-04-18
MAYA-128513 implement composite command handler

pierrebai-adsk:
We implement the composite command handler and make the group (insert-child) command work within composite commands. The real, true Maya group command is actually composed of several UFE command and has been modified in a Maya branch to use the new UFE composite-command handler.

Implement the composite command handler:
- Make the batch-ops handler derive from the correct class depending on which feature is supported in UFE.
- Implement a composite command handler derived from the batch ops handler.
- The composite handler creates edit-routing wrappers for composite commands.
- Cache the stage and layer used when first executing a composite command.
- This allows using the same values when undoing.
- This is necessary because the original prim that was used when executing the command might not exist anymore once the command executed.
- In particular, this is true for the group command, the original prim no longer exists.

Make the group command work with edit routing:)
- Use the generic undo/redo support to record the changes, undo and redo them.
- Don't assume that the current edit target is where the source prim has opinion.
- Instead, apply the child-insertion to the layer that have opinion about the moving prim.
- The command restrictions already ensures that we only affect prims that have a single opinion.
- (But the code will be ready if we support multi-layers command later on.)
- The problem was that the command restriction was correctly looking at where the prim was authored, but the command execution was assuming that the source opinion was on the target layer.
- When routing commands, that can obviously be false: the target might not be where the original opinions were.
- We do the right thing using similar code as when affecting the session layer, except the target layer is the layer where the command was routed.
- By default, when no routing is registered, that is the current edit target.

### DIFF
--- a/lib/mayaUsd/ufe/UsdBatchOpsHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdBatchOpsHandler.cpp
@@ -21,7 +21,6 @@ namespace MAYAUSD_NS_DEF {
 namespace ufe {
 
 UsdBatchOpsHandler::UsdBatchOpsHandler()
-    : Ufe::BatchOpsHandler()
 {
 }
 
--- a/lib/mayaUsd/ufe/UsdBatchOpsHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdBatchOpsHandler.cpp
@@ -20,9 +20,7 @@
 namespace MAYAUSD_NS_DEF {
 namespace ufe {
 
-UsdBatchOpsHandler::UsdBatchOpsHandler()
-{
-}
+UsdBatchOpsHandler::UsdBatchOpsHandler() { }
 
 UsdBatchOpsHandler::~UsdBatchOpsHandler() { }
 
--- a/lib/mayaUsd/ufe/UsdCompositeCommandHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdCompositeCommandHandler.cpp
@@ -36,7 +36,15 @@ public:
 
     void preCall() override
     {
-        _editRouterContext = std::make_unique<OperationEditRouterContext>(_operationName, _prim);
+        if (_alreadyRouted) {
+            _editRouterContext = std::make_unique<OperationEditRouterContext>(_stage, _layer);
+        } else {
+            _editRouterContext
+                = std::make_unique<OperationEditRouterContext>(_operationName, _prim);
+            _stage = _editRouterContext->getStage();
+            _layer = _editRouterContext->getLayer();
+            _alreadyRouted = true;
+        }
     }
 
     void postCall() override { _editRouterContext.reset(); }
@@ -56,6 +64,9 @@ private:
 
     PXR_NS::UsdPrim                             _prim;
     PXR_NS::TfToken                             _operationName;
+    bool                                        _alreadyRouted = false;
+    PXR_NS::UsdStagePtr                         _stage;
+    PXR_NS::SdfLayerHandle                      _layer;
     std::unique_ptr<OperationEditRouterContext> _editRouterContext;
 };
 
--- a/lib/mayaUsd/ufe/UsdCompositeCommandHandler.cpp
+++ b/lib/mayaUsd/ufe/UsdCompositeCommandHandler.cpp
@@ -1,0 +1,91 @@
+//
+// Copyright 2023 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "UsdCompositeCommandHandler.h"
+
+#include <mayaUsd/ufe/UsdSceneItem.h>
+#include <mayaUsd/utils/editRouterContext.h>
+
+#include <ufe/sceneItem.h>
+#include <ufe/selection.h>
+
+namespace MAYAUSD_NS_DEF {
+namespace ufe {
+
+namespace {
+class UsdCodeWrapper : public Ufe::CodeWrapper
+{
+public:
+    UsdCodeWrapper(const Ufe::Selection& selection, const std::string& operationName)
+        : _prim(findPrimInSelection(selection))
+        , _operationName(PXR_NS::TfToken(operationName))
+    {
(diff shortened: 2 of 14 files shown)


lib/mayaUsd/ufe/UsdBatchOpsHandler.cpp
```diff
@@ -20,10 +20,7 @@
 namespace MAYAUSD_NS_DEF {
 namespace ufe {
 
-UsdBatchOpsHandler::UsdBatchOpsHandler()
-    : Ufe::BatchOpsHandler()
-{
```
pierrebai-adsk, line 25: The base clas vary depending on the version of UFE, so dont call it explicitly. It's not needed anyway.

lib/mayaUsd/ufe/UsdCompositeCommandHandler.cpp
```diff
@@ -0,0 +1,102 @@
...
+    void preCall() override
+    {
+        if (_alreadyRouted) {
+            _editRouterContext = std::make_unique<OperationEditRouterContext>(_stage, _layer);
+        } else {
+            _editRouterContext
+                = std::make_unique<OperationEditRouterContext>(_operationName, _prim);
+            _stage = _editRouterContext->getStage();
+            _layer = _editRouterContext->getLayer();
+            _alreadyRouted = true;
```
pierrebai-adsk, line 46: We must cache the stage and layer that were used on first execution so that undo and redo use the same targets. Especially important since on undo, the original prim will not exist anymore.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -132,8 +133,6 @@ UsdUndoInsertChildCommand::UsdUndoInsertChildCommand(
     // Apply restriction rules
     ufe::applyCommandRestriction(childPrim, "reparent");
     ufe::applyCommandRestriction(parentPrim, "reparent");
-
-    _childLayer = childPrim.GetStage()->GetEditTarget().GetLayer();
 }
```
pierrebai-adsk, line 136: The old code assumed the original opinions about the prim that will be grouped were all in the current edit target. When the commands are being routed to other layers, that is not true. As a matter of fact, it was never true but just happened to work because we tend to use USD with a single layer.

The new code below correctly finds the source layer where the opinions actually are.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -184,41 +183,52 @@ static void doUsdInsertion(
...
+    // the target layer. The edit router context sets the edit target of the
+    // stage of the given prim, if it gets routed.
+    OperationEditRouterContext ctx(MayaUsdEditRoutingTokens->RouteParent, srcPrim);
+    const SdfLayerHandle&      dstLayer = srcPrim.GetStage()->GetEditTarget().GetLayer();
 
-    // Make sure the load state of the reparented prim will be preserved.
-    // We copy all rules that applied to it specifically and remove the rules
-    // that applied to it specifically.
-    duplicateLoadRules(*stage, srcUsdPath, dstUsdPath);
-    removeRulesForPath(*stage, srcUsdPath);
```
pierrebai-adsk, line 206: The load rule duplication is moved elsewhere, into a function shared with undo and redo.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -184,41 +183,52 @@ static void doUsdInsertion(
...
+    {
+        PrimLayerFunc insertionFunc
+            = [stage, srcUsdPath, dstLayer, dstUsdPath](
+                  const UsdPrim& prim, const PXR_NS::SdfLayerRefPtr& layer) {
+                  doUsdInsertion(stage, layer, srcUsdPath, dstLayer, dstUsdPath);
+              };
+
+        const bool includeTopLayer = true;
+        const auto rootLayers = getAllSublayerRefs(stage->GetRootLayer(), includeTopLayer);
+        applyToSomeLayersWithOpinions(srcPrim, rootLayers, insertionFunc);
```
pierrebai-adsk, line 215: This is how we find the layers with the original opinions. They will be grouped and authored in the current edit target, possibly selected by the edit router.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -269,29 +295,46 @@ void UsdUndoInsertChildCommand::insertChildRedo()
...
+    // to access the existing rules.
+    preserveLoadRules(_ufeSrcPath, _usdSrcPath, _usdDstPath);
+
     // We need to keep the generated item to be able to return it to the caller
     // via the insertedChild() member function.
-    _ufeDstItem = doInsertion(
-        _childLayer, _usdSrcPath, _ufeSrcPath, _parentLayer, _usdDstPath, _ufeDstPath);
-}
+    {
+        UsdUndoBlock undoBlock(&_undoableItem);
```
pierrebai-adsk, line 305: I now use the UsdUndoBlock to capture all USD modifications and use it in undo/redo instead of re-executing complicated code.

lib/mayaUsd/ufe/UsdUndoInsertChildCommand.cpp
```diff
@@ -269,29 +295,46 @@ void UsdUndoInsertChildCommand::insertChildRedo()
...
 {
     InPathChange pc;
-    insertChildUndo();
+
+    // Load rules must be duplicated before the prim is moved to be able
+    // to access the existing rules.
+    // Note: the arguments passed are the opposite of those in execute and redo().
+    preserveLoadRules(_ufeDstPath, _usdDstPath, _usdSrcPath);
+
+    _undoableItem.undo();
```
pierrebai-adsk, line 321: Undo (and redo below) can now use the capture undoable items.

lib/mayaUsd/utils/editRouterContext.h
```diff
@@ -77,6 +83,13 @@ class MAYAUSD_CORE_PUBLIC OperationEditRouterContext : public StackedEditRouterC
      */
     OperationEditRouterContext(const PXR_NS::TfToken& operationName, const PXR_NS::UsdPrim& prim);
 
+    /*! \brief Route to the given stage and layer.
+     *  Should be used in undo to ensure the same target is used as in the initial execution.
+     */
+    OperationEditRouterContext(
+        const PXR_NS::UsdStagePtr&    stage,
+        const PXR_NS::SdfLayerHandle& layer);
```
pierrebai-adsk, line 91: These new constructors allow creating a router context directly with the desired layer instead of calling the edit router. As explained in the comment above, this is to support undo and redo correctly, which both want to use the same target as was decided when executed the first time. We don't want to make different decisions on undo and redo.

vlasovi (approved):
Looks good

AramAzhari-adsk (approved):

Issue #906, opened 2020-11-09
USD Importer doesn't adjust for metersPerUnit metadata


dgovil:
**Describe the bug**
The USD Importer doesn't adjust imported objects scales to match the metersPerUnit metadata.
This is a hard problem to solve, as has been discussed in the USD group before I believe, but I did not see a related PR, so am filing for documentations sake at the very least, and with some hope that it can be fixed in the future.

**Steps to reproduce**
1. Import both attached usda files in the zip

**Expected behavior**
The cubes should be the same size, but are not

**Attachments**
The two assets in here should come in at the same scale if the metersPerUnit metadata was respected
[Archive.zip](https://github.com/Autodesk/maya-usd/files/5513126/Archive.zip)

![image](https://user-images.githubusercontent.com/4443090/98592685-ece00b00-2286-11eb-9c65-70555bd60e77.png)

**Specs (if applicable):**

- OS & version: macOS
- Compiler & version : Apple Clang
- Maya version: Maya 2018, 2020, 2020.3
- Maya USD commit SHA 8b3da3c
- Pixar USD commit SHA [e.g. dev at b85ddac] d8a405a1344480f859f025c4f97085143efacb53

**Additional context**

Using the mayaUSDplugin.bundle

PR #3031, opened 2023-04-19
MAYA-128598 fix crash when deleting a stage

pierrebai-adsk:
Dont' use an invalid pointer when checking if a layer is muted.


vlasovi (approved):

samuelliu-adsk (approved):

pierrebai-adsk:
PF was run internally, #906

PR #3032, opened 2023-04-19
UsdLight*.cpp: fix building with non-pxr namespace

nxkb:
Just a small find/replace build fix for those of us using a custom pxr namespace.

This brings these usages into line with the rest of the code base.


seando-adsk (dismissed):
Thanks for finding and fixing these. Did you run clang-format on the changes? I think it will fail in on the lines next to other variables since the names don't line up anymore.

nxkb:
> Thanks for finding and fixing these. Did you run clang-format on the changes? I think it will fail in on the lines next to other variables since the names don't line up anymore.

ah right! ok that's done now.

seando-adsk (approved):

PR #3034, opened 2023-04-21
MAYA-128764 fix AE refresh

pierrebai-adsk:
Setting the value of a USD property that previously had the default value for that property creates a USD property-spec, which trigger a USD notification about a property being added. In reality, it is not that a new custom property is added, just that an opinion about a property that is being set.

The attribute editor handled the addition of a property by recreating the UI entirely.

To differentiate between a real property being added and a property getting a non-default value, we now add a guard in the various places where we are setting a value to allow the notification handler to know the difference.


seando-adsk (dismissed):

samuelliu-adsk (approved):

pierrebai-adsk:
The only PF that failed are due to the known problem with artifactory.

seando-adsk (approved):

PR #3038, opened 2023-04-26
Separate mtoh .mod file.

ppt-adsk:
mtoh part of .mod file split into a separate mtoh.mod, which is installed only if mtoh is enabled in the build.


seando-adsk (dismissed):
Looks good, although technically we didn't need a separate windows mod file since MtoH doesn't change PATH.

ppt-adsk:
Should I remove the Windows-specific .mod file then?  Don't want to add dead code.

seando-adsk (approved):

PR #3046, opened 2023-04-28
Use Hd_VertexAdjacency API directly

dj-mcg:
Updated to use the vertex adjacency API directly to build the vertex adjacency table instead of using Storm specific shared builder computations.

This simplifies the code sightly while avoiding a dependency on API that is being moved from imaging/hd to imaging/hdSt.

Not using a version guard here since the interface we're switching to has been available for a long time.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
@@ -426,18 +426,15 @@ void HdVP2Mesh::_PrepareSharedVertexBuffers(
                 // note: normals gets dirty when points are marked as dirty,
                 // at change tracker.
                 if (!_meshSharedData->_adjacency) {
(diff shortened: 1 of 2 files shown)


dj-mcg:
Hold the review for now, it looks like I stomped some very recent changes with my own here. Update incoming...

vlasovi (approved):

JGamache-autodesk (approved):

dj-mcg:
Hey, just checking in to make sure that there's nothing pending on my side for getting this integrated - thanks!

vlasovi:
@dj-mcg  Our preflight system is back up again, so you can try to rerun your preflight.

dj-mcg:
I now seem to be failing the clang formatting tests, and it looks like it's from a different change?

https://github.com/Autodesk/maya-usd/commit/45610180f9e726de71ead885e2c180f9e7278506

Am I misreading the preflight results?

vlasovi:
@dj-mcg I restarted the preflight

PR #3052, opened 2023-05-02
MAYA-128363 pin stage in layer editor

pierrebai-adsk:
Add a pin to the stage selector widget, to the right of the selector box.
- Refactor code dealing with bitmaps from the layer editor into QtUtils.
- The pin switches between following the selection or not.
- When unpinned it follows the selection.
- Move UI creation into its own function.

Trigger auto stage-selection when the outliner selection changes:
- Add a UFE selection observer.
- The observer triggers the stage selector widget.
- Register the stage selector widgets with the selection observer.
- Make the selector widget switch stage based on the selection.


pierrebai-adsk:
FYI, PF build 942 passed, only the known failure of the isolate select image comparison test on Maya 2022 failed.

AramAzhari-adsk (approved):

AramAzhari-adsk (approved):

AramAzhari-adsk (approved):

PR #3056, opened 2023-05-03
MAYA-129061 - Refactor material commands.

frohnej-adsk:
In UsdUndoMaterialCommands, factor out the code related to materials scopes into separate commands and functions. This will allow reusing them and is required by future work (MAYA-128151).

Also, make `UsdUndoAddNewPrimCommand` derive from `SceneItemResultUndoableCommand`. This allows UFE clients to access the newly created prim by calling `SceneItemResultUndoableCommand::sceneItem()`.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoAddNewPrimCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoAddNewPrimCommand.cpp
@@ -22,6 +22,7 @@
 #include <mayaUsd/ufe/Utils.h>
 #include <mayaUsd/undo/UsdUndoBlock.h>
 
+#include <ufe/hierarchy.h>
 #include <ufe/pathString.h>
 
 namespace {
@@ -47,7 +48,11 @@ UsdUndoAddNewPrimCommand::UsdUndoAddNewPrimCommand(
     const UsdSceneItem::Ptr& usdSceneItem,
     const std::string&       name,
     const std::string&       type)
+#ifdef UFE_V4_FEATURES_AVAILABLE
+    : Ufe::SceneItemResultUndoableCommand()
+#else
     : Ufe::UndoableCommand()
+#endif
 {
     // First get the stage from the proxy shape.
     auto ufePath = usdSceneItem->path();
@@ -114,6 +119,11 @@ std::string UsdUndoAddNewPrimCommand::commandString() const
 {
     return std::string("CreatePrim ") + _primToken.GetText() + " "
         + Ufe::PathString::string(_newUfePath);
+}
+
+Ufe::SceneItem::Ptr UsdUndoAddNewPrimCommand::sceneItem() const
+{
+    return Ufe::Hierarchy::createItem(newUfePath());
 }
 #endif
 
--- a/lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
@@ -86,31 +86,6 @@ bool connectShaderToMaterial(
     UsdShadeConnectableAPI::ConnectToSource(materialOutput, shaderOutput);
     return true;
 }
-#endif
-
-bool _BindMaterialCompatiblePrim(const UsdPrim& usdPrim)
-{
-    if (UsdShadeNodeGraph(usdPrim) || UsdShadeShader(usdPrim)) {
-        // The binding schema can be applied anywhere, but it makes no sense on a
-        // material or a shader.
-        return false;
-    }
-    if (UsdGeomScope(usdPrim)
-        && usdPrim.GetName() == UsdMayaJobExportArgs::GetDefaultMaterialsScopeName()) {
-        return false;
-    }
-    if (auto subset = UsdGeomSubset(usdPrim)) {
-        TfToken elementType;
-        subset.GetElementTypeAttr().Get(&elementType);
-        if (elementType != UsdGeomTokens->face) {
-            return false;
-        }
-    }
-    if (PXR_NS::UsdShadeMaterialBindingAPI::CanApply(usdPrim)) {
-        return true;
-    }
-    return false;
-}
 
 //! Returns true if \p item is a materials scope.
 bool isMaterialsScope(const Ufe::SceneItem::Ptr& item)
@@ -178,6 +153,31 @@ Ufe::SceneItem::Ptr getMaterialsScope(const Ufe::Path& parentPath)
         // Name is already used by something that is not a scope. Try the next name.
         scopeName = scopeNamePrefix + std::to_string(i);
     }
+}
+#endif
+
+bool _BindMaterialCompatiblePrim(const UsdPrim& usdPrim)
+{
+    if (UsdShadeNodeGraph(usdPrim) || UsdShadeShader(usdPrim)) {
+        // The binding schema can be applied anywhere, but it makes no sense on a
+        // material or a shader.
+        return false;
+    }
+    if (UsdGeomScope(usdPrim)
+        && usdPrim.GetName() == UsdMayaJobExportArgs::GetDefaultMaterialsScopeName()) {
+        return false;
+    }
+    if (auto subset = UsdGeomSubset(usdPrim)) {
+        TfToken elementType;
+        subset.GetElementTypeAttr().Get(&elementType);
+        if (elementType != UsdGeomTokens->face) {
+            return false;
+        }
+    }
+    if (PXR_NS::UsdShadeMaterialBindingAPI::CanApply(usdPrim)) {
+        return true;
+    }
+    return false;
 }
 
 } // namespace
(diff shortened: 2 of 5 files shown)


lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -96,7 +95,8 @@ bool _BindMaterialCompatiblePrim(const UsdPrim& usdPrim)
         // material or a shader.
         return false;
     }
-    if (UsdGeomScope(usdPrim) && usdPrim.GetName() == kDefaultMaterialScopeName) {
+    if (UsdGeomScope(usdPrim)
+        && usdPrim.GetName() == UsdMayaJobExportArgs::GetDefaultMaterialsScopeName()) {
```
frohnej-adsk, line 99: Use the resolved materials scope name instead of the hardcoded value. This was probably an oversight, when   moving the default materials scope name into UsdMayaJobExportArgs.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -112,6 +112,74 @@ bool _BindMaterialCompatiblePrim(const UsdPrim& usdPrim)
     return false;
 }
 
+//! Returns true if \p item is a materials scope.
+bool isMaterialsScope(const Ufe::SceneItem::Ptr& item)
```
frohnej-adsk, line 116: Factored out from `UsdUndoAddNewMaterialCommand::CompatiblePrim` so that it can be reused by other commands in the future.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -112,6 +112,74 @@ bool _BindMaterialCompatiblePrim(const UsdPrim& usdPrim)
...
+            }
+        }
+    }
+
+    return true;
+}
+
+//! Searches the children of \p parentPath for a materials scope. Returns a null pointer if no
+//! materials scope is found.
+Ufe::SceneItem::Ptr getMaterialsScope(const Ufe::Path& parentPath)
```
frohnej-adsk, line 148: Factored out from `UsdUndoAssignNewMaterialCommand`. It is now being called by the new `UsdUndoCreateMaterialsScopeCommand`. This function is currently only called once but I still made it a separate function, as it could be useful for others at some point. This way it is easy to move it to a shared location if necessary.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -276,15 +344,8 @@ Ufe::SceneItem::Ptr UsdUndoAssignNewMaterialCommand::insertedChild() const
     return {};
 }
 
-std::string UsdUndoAssignNewMaterialCommand::resolvedMaterialScopeName()
-{
-    return UsdMayaJobExportArgs::GetDefaultMaterialsScopeName();
-}
```
frohnej-adsk, line 282: Not really needed anymore, since `UsdMayaJobExportArgs::GetDefaultMaterialsScopeName()` can simply be called directly.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -302,63 +363,17 @@ void UsdUndoAssignNewMaterialCommand::execute()
...
-            }
-            auto rename = itemOps->renameItemCmd(Ufe::PathComponent(materialsScopeName));
-            if (!rename.undoableCommand) {
-                markAsFailed();
-                return;
-            }
-            _cmds->append(rename.undoableCommand);
-            materialsScope = rename.item;
-        }
+        auto materialsScope = createMaterialsScopeCmd->sceneItem();
```
frohnej-adsk, line 376: Factored out the creation of the materials scope into a new command, so that it can be reused by other commands in future work.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -525,32 +540,8 @@ Ufe::SceneItem::Ptr UsdUndoAddNewMaterialCommand::insertedChild() const
...
-            if (child->nodeType() != "Material") {
-                // At least one non material
-                return false;
-            }
-        }
-    }
-
-    return true;
+    // Must be a materials scope.
+    return isMaterialsScope(target);
```
frohnej-adsk, line 544: Factored out into a new function, so that it can be reused by other commands in the future.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.h
```diff
@@ -178,6 +175,45 @@ class MAYAUSD_CORE_PUBLIC UsdUndoAddNewMaterialCommand : public Ufe::InsertChild
     UsdUndoCreateFromNodeDefCommand::Ptr _createShaderCmd;
 
 }; // UsdUndoAddNewMaterialCommand
+
+//! \brief This command is used to create a materials scope under a specified parent item. A
+//! materials scope is a USD Scope with a special name (usually "mtl"), which holds materials. By
+//! convention, all materials should reside within such a scope.
+class MAYAUSD_CORE_PUBLIC UsdUndoCreateMaterialsScopeCommand
+    : public Ufe::SceneItemResultUndoableCommand
```
frohnej-adsk, line 183: New command that takes care of creating a materials scope. Factored out from `UsdUndoAssignNewMaterialCommand`. This command will be reused in future work.

lib/mayaUsd/ufe/UsdUndoMaterialCommands.cpp
```diff
@@ -96,7 +163,8 @@ bool _BindMaterialCompatiblePrim(const UsdPrim& usdPrim)
         // material or a shader.
         return false;
     }
-    if (UsdGeomScope(usdPrim) && usdPrim.GetName() == kDefaultMaterialScopeName) {
+    if (UsdGeomScope(usdPrim)
+        && usdPrim.GetName() == UsdMayaJobExportArgs::GetDefaultMaterialsScopeName()) {
```
frohnej-adsk, line 167: Use the resolved materials scope name instead of the hardcoded value. This was probably an oversight when moving the default materials scope name into UsdMayaJobExportArgs.

JGamache-autodesk (approved):
Well done! This will help.

seando-adsk (approved):
Since this was just refactoring existing code I assume there is no new unit test(s) that are needed?

lib/mayaUsd/ufe/UsdUndoAddNewPrimCommand.cpp
```diff
@@ -115,6 +120,11 @@ std::string UsdUndoAddNewPrimCommand::commandString() const
     return std::string("CreatePrim ") + _primToken.GetText() + " "
         + Ufe::PathString::string(_newUfePath);
 }
+
+Ufe::SceneItem::Ptr UsdUndoAddNewPrimCommand::sceneItem() const
+{
+    return Ufe::Hierarchy::createItem(newUfePath());
```
frohnej-adsk, line 126: This method is currently not used or tested. However, I'm not sure if it can easily be tested right now, since `UsdUndoAddNewPrimCommand` is tested though `Ufe::ContextOps::doOpCmd()` (see [here](https://github.com/Autodesk/maya-usd/blob/60d70d6e7ccb844d920e051fd98639d5ba9f95fd/test/lib/ufe/testContextOps.py#L314)), which returns a `UndoableCommand::Ptr`, not a `Ufe::SceneItemResultUndoableCommand()`.

frohnej-adsk (commented):
Yes, the refactored code is covered by the existing tests.

`UsdUndoCreateMaterialsScopeCommand` is now separated out, so if we want that, we could also test it independently by creating a Python binding for it. However, it is used by `UsdUndoAssignNewMaterialCommand` and, thus, already covered by e.g. [this test](https://github.com/Autodesk/maya-usd/blob/60d70d6e7ccb844d920e051fd98639d5ba9f95fd/test/lib/ufe/testContextOps.py#L843).

One addition that is not used or tested yet is `UsdUndoAddNewPrimCommand::sceneItem()`.

lib/mayaUsd/ufe/UsdUndoAddNewPrimCommand.cpp
frohnej-adsk: I made the material commands call `UsdUndoAddNewPrimCommand::sceneItem()` instead of manually calling `Ufe::Hierarchy::createItem()` on `UsdUndoAddNewPrimCommand::newUfePath()`. This makes the code a little more tidy and ensures that `UsdUndoAddNewPrimCommand::sceneItem()` is covered by existing tests. See 088f54ab81ccfb4ddc941d5c120888d34dee9fea.

frohnej-adsk:
Rebase on dev as an attempt of fixing clang-format issues on code I have not touched.

frohnej-adsk:
Rebase on dev again to fix formating.

seando-adsk (approved):

JGamache-autodesk (approved):

PR #3057, opened 2023-05-03
MAYA-129155 fix shiboken crash

pierrebai-adsk:
The crash was due to the cleanup done by Maya with Shiboken. I believe the problem is that by the time the cleanup is called, the USD plugin is already unloaded or already destroyed in Qt UI.

- Don't use Shiboken to convert a layout raw pointer into a QWidget.
- Instead, provide a function to retrieve the window name that contains a layout.
- Then find the window by name.

### DIFF
--- a/lib/mayaUsd/python/CMakeLists.txt
+++ b/lib/mayaUsd/python/CMakeLists.txt
@@ -78,17 +78,6 @@ target_compile_definitions(${PYTHON_TARGET_NAME}
         MFB_PACKAGE_MODULE=${PROJECT_NAME}
 )
 
-# QT_NO_KEYWORDS prevents Qt from defining the foreach, signals, slots and emit macros.
-# this avoids overlap between Qt macros and boost, and enforces using Q_ macros.
-set_target_properties(Qt5::Core PROPERTIES INTERFACE_COMPILE_DEFINITIONS QT_NO_KEYWORDS)
-
-if(IS_LINUX)
-    target_compile_definitions(${PYTHON_TARGET_NAME}
-        PRIVATE
-            LINUX=1
-    )
-endif()
-
 mayaUsd_compile_config(${PYTHON_TARGET_NAME})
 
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/python/CMakeLists.txt
```diff
@@ -86,6 +91,10 @@ mayaUsd_compile_config(${PYTHON_TARGET_NAME})
 target_link_libraries(${PYTHON_TARGET_NAME}
     PRIVATE
         ${PROJECT_NAME}
+    PRIVATE
+        Qt5::Core
+        Qt5::Gui
+        Qt5::Widgets
```
seando-adsk, line 97: Sorry I don't like this. We should not make this lib link with Qt. The only lib in our repo that should link with Qt is the mayaUsdUI lib.

seando-adsk (changes requested):

pierrebai-adsk:
Related PF is 955.

plugin/adsk/scripts/mayaUsd_USDRootFileRelative.py
```diff
@@ -181,18 +197,13 @@ def connectToDialogControls(cls, parentLayout):
...
-                pp = pp.parent()
-                if pp:
-                    maya_window = pp
+        # Get the Qt Window containing the input parent layout from mel.
+        maya_window_name = usdFileRelative.findWindowNameFromLayout(parentLayout)
+        maya_window = usdFileRelative.findQtWindowFromTitle(maya_window_name)
+        if maya_window:
+            cls._fileDialog = maya_window
+        else:
+            print('Could not find dialog named %s' % maya_window_name)
```
seando-adsk, line 206: What is the purpose of the print? The user cannot really do anything to fix this error right? If you want the user to know about it then maybe use cmds.warning().

seando-adsk (approved):

seando-adsk (approved):

PR #3061, opened 2023-05-05
MAYA-126014 - Clipboard cut, copy and paste.

alicedegirolamo:
## Description

This PR solves Jira ticket [MAYA-126014](https://jira.autodesk.com/browse/MAYA-126014) .

The correspondent UFE PR is [here](https://git.autodesk.com/media-and-entertainment/ufe/pull/340). 
We are introducing the clipboard concept so we can cut, copy and paste.

The proposed approach is similar to what we do for  [cutCopyPaste.mel ](https://git.autodesk.com/maya3d/maya/blob/master/Maya/src/Maya/Maya/scripts/cutCopyPaste.mel) and [defaultRunTimeCommands.mel
](https://git.autodesk.com/maya3d/maya/blob/master/Maya/src/Maya/Maya/scripts/defaultRunTimeCommands.mel)

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.h
+++ b/lib/mayaUsd/ufe/UsdUndoDuplicateCommand.h
@@ -42,6 +42,10 @@ public:
     typedef std::shared_ptr<UsdUndoDuplicateCommand> Ptr;
 
     UsdUndoDuplicateCommand(const UsdSceneItem::Ptr& srcItem);
+    UsdUndoDuplicateCommand(
+        const UsdSceneItem::Ptr& srcItem,
+        const UsdSceneItem::Ptr& dstParentItem);
+
     ~UsdUndoDuplicateCommand() override;
 
     // Delete the copy/move constructors assignment operators.
@@ -52,6 +56,10 @@ public:
 
     //! Create a UsdUndoDuplicateCommand from a USD prim and UFE path.
     static UsdUndoDuplicateCommand::Ptr create(const UsdSceneItem::Ptr& srcItem);
+
+    //! Create a UsdUndoDuplicateCommand from a USD prim and its parent destination.
+    static UsdUndoDuplicateCommand::Ptr
+    create(const UsdSceneItem::Ptr& srcItem, const UsdSceneItem::Ptr& dstParentItem);
 
     UsdSceneItem::Ptr duplicatedItem() const;
     UFE_V4(Ufe::SceneItem::Ptr sceneItem() const override { return duplicatedItem(); })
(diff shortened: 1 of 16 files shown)


lib/mayaUsd/ufe/UsdUndoDuplicateCommand.h
```diff
@@ -42,6 +42,10 @@ class MAYAUSD_CORE_PUBLIC UsdUndoDuplicateCommand : public Ufe::UndoableCommand
     typedef std::shared_ptr<UsdUndoDuplicateCommand> Ptr;
 
     UsdUndoDuplicateCommand(const UsdSceneItem::Ptr& srcItem);
+    UsdUndoDuplicateCommand(
+        const UsdSceneItem::Ptr& srcItem,
+        const UsdSceneItem::Ptr& dstParentItem);
```
alicedegirolamo, line 47: I've modified this existing so we can have a different parent item as destination and so also having a different destination layer.

lib/mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.cpp
```diff
@@ -155,6 +181,16 @@ void UsdUndoDuplicateSelectionCommand::execute()
...
                     }
                 }
+            };
+
+            // Cleanup relationships and connections on the duplicate.
+            if (_dstParentItem) {
+                auto p = stage->GetPrimAtPath(duplicatePair.second);
+                updatePropertiesAndRelationship(p);
+            } else {
+                for (auto p : UsdPrimRange(stage->GetPrimAtPath(duplicatePair.second))) {
```
alicedegirolamo, line 191: I did some tests here since when I tried to update the paths using the `updatePropertiesAndRelationship `function, it updates also the paths obtained by the subtree in `UsdPrimRange`.

In the clipboard case, we are copying to a different destination layer and I got incorrect paths for the subtree (tried with nested compounds). 

I tried avoiding updating them using `updatePropertiesAndRelationship` and I got correct results, so I've investigated and I think the `SdfCopySpec` function we call in the `UsdUndoDuplicateCommand::execute` takes care of the subtree updates. From the [doc](https://openusd.org/dev/api/copy_utils_8h.html#ad20defdb6654ee0937942116177fd4cc):
" Attribute connections, relationship targets, inherit and specializes paths, and internal sub-root references that target an object beneath srcPath will be remapped to target objects beneath dstPath. "
 
I did other tests for the `updatePropertiesAndRelationship` function for the `UsdUndoDuplicateSelectionCommand `
( calling it in "standard" not clipboard case), and I saw the subtrees had always updated paths, so I'm not sure if we need to call it on the `UsdPrimRange`.

alicedegirolamo:
Edit: put PR status as Draft following the corresponding discussion in UFE.

PR #3065, opened 2023-05-05
MAYA-129169 improve the behavior of duplication vs payload

pierrebai-adsk:
Handling all possible scenarios is potentially impossible given that our Maya-level view are of fully composed prims. So when Maya requests to duplicate a prim, we get a simple request to duplicate whatever is at a given UFE path. In USD, that could be in the stage, in a payload, in a reference or a combination of these. Plus, given composed opinions from multiple layers, it could be a complex combination.

This change slightly improve the behaviour to handle some simple the cases.

- Add heuristics to try to do the right thing when payloads are involved.
- Make sure to always at least keep the defining layer, in case the heuristic thinks it should be rejected.
- Added a unit test for duplicate with payload.


AramAzhari-adsk (approved):

pierrebai-adsk:
PF #959, only failure is the known image comparison test failure in Windows 2020

PR #3068, opened 2023-05-08
Feldstj/lookdevx 861/ufe master enums

feldstj:


test/lib/ufe/testAttribute.py
```diff
@@ -532,6 +532,35 @@ def testAttributeInt(self):
         # Run test using Maya's getAttr command.
         self.runMayaGetAttrTest(ufeAttr)
 
+    @unittest.skipIf(os.getenv('UFE_PREVIEW_VERSION_NUM', '0000') < '4200',
+                     'getEnums interface only available in Ufe preview version greater equal to 4.1.200, or 0.5.4.')
```
seando-adsk, line 536: This doesn't test for the 0.5.4 (or greater). Instead of using ufe preview num you can have your UFE_ATTRIBUTES_GET_ENUMS cmake variable set as an env var for the tests. Then the test can check that, ex: https://github.com/Autodesk/maya-usd/blob/dev/test/lib/ufe/CMakeLists.txt#L134

seando-adsk (changes requested):
Everything looks great. Thanks. Just one minor comment about the new test.

seando-adsk (approved):

PR #3069, opened 2023-05-08
MAYA-129047 allow editing prims directly under the stage

pierrebai-adsk:
When handling the combined visibility of the edited prim, the code did not work when the prim was directly under the stage.

More precisely, the ProxyAccessor design did not support accessing properties of the absolute root node of a stage. Normally, this would not be a problem since that prim does not have attributes, but for Edit-as-Maya, the combined visibility assumed it could create a SdfPath to the associated property, which is impossible for the root prim.

The solution is to modify the ProxyAccessor code to avoid creating the SdfPath to the property and instead keep the prim path and property name separate in the code. Unfortunately, the way the ProxyAccessor works is by encoding the full path of the property ("prim.property") in the Maya attribute name. So we "cheat" by creating impossible names for the root node. For example "/.combinedVisibility". This requires never treating the Maya attribute name as a SdfPath, but splitting it ourselves into a SdfPath and a property name. We do this to be backward compatible.

Before the introduction of combinedVisibility, the only attribute that was accessed on the stage root prim was its transformation matrix, at the code cheated by *not* providing a property name and having the code magically knowing that when there is no property name, it meant the transform matrix was intended. Again, this magic is kept for backward compatibility.

While at it, some code cleanup was done to use an explicit structure instead of a std::tuple to make the code more readable and easier to maintain. (std::get<3> make is unclear what is being accessed...)

### DIFF
--- a/lib/mayaUsd/nodes/proxyAccessor.py
+++ b/lib/mayaUsd/nodes/proxyAccessor.py
@@ -249,7 +249,6 @@ def parentItems(ufeChildren, ufeParent, connect=True):
         print("This method implements parenting under USD prim. Please provide UFE-USD item for ufeParent")
         return
     
-    print('parentItems: %s' % str(ufeParent.path()))
     parentDagPath, parentUsdPrimPath = getDagAndPrimFromUfe(ufeParent)
     parentValueAttr = getOrCreateAccessPlug(ufeParent, '', Sdf.ValueTypeNames.Matrix4d )
     parentConnectionAttr = parentDagPath+'.'+parentValueAttr+'[0]' if parentValueAttr else None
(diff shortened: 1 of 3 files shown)


pierrebai-adsk:
Successful PF is #966 (only the known isolate select test fails in 2022)

lib/mayaUsd/nodes/proxyAccessor.py
```diff
@@ -219,17 +229,33 @@ def keyframeAccessPlug(ufeObject, usdAttrName):
...
+        cmds.connectAttr(parentAttr, childAttr)
+    else:
+        cmds.disconnectAttr(parentAttr, childAttr)
+
 def parentItems(ufeChildren, ufeParent, connect=True):
     if not isUfeUsdPath(ufeParent):
         print("This method implements parenting under USD prim. Please provide UFE-USD item for ufeParent")
         return
     
+    print('parentItems: %s' % str(ufeParent.path()))
```
vlasovi, line 252: Is this a debug print?

vlasovi (approved):
Looks good.

vlasovi (approved):

PR #3071, opened 2023-05-09
MAYA-128594 - Imported displayOpacity are not interpreted correctly

samuelliu-adsk:


lib/usd/translators/meshWriter.cpp
```diff
@@ -735,6 +742,12 @@ bool PxrUsdTranslators_MeshWriter::writeMeshAttrs(
                 clamped,
                 true,
                 _GetSparseValueWriter());
+
+            // We can only export display opacity separately if display color doesn't contain alpha
+            // data
+            if (colorSetRep == MFnMesh::kRGB) {
+                exportDisplayOpacity = true;
+            }
```
pierrebai-adsk, line 750: Should there be a pre-pass to detect this? The code as written seems to depends on the order of the color set?

samuelliu-adsk: Yes. If we can export DisplayOpacity separately actually depends on the type of DisplayColor. If the display color is in RGBA format (which means it contains alpha, aka opacity, already), then we can't export the DisplayOpacity. However if the color is in RGB format (no alpha data), we can safely export the opacity, and this should be the only scenario we export DisplayOpacity

pierrebai-adsk: Yes, but as I understand the code, the flag exportDisplayOpacity is set to true within the same loop where it is checked. So the behaviour depends on the order the color set names are encountered in this loop, which seems to me to be arbitrary?

In short, if the flags gets sets *after* DisplayOpacityColorSetName has already been processed, then it won't use the opacity even though the flag is now true.

That is why I suggested that the flag be set in a separate loop that would run before, so that it is *certain* to be set to the correct value when the opacity color set is encountered.

samuelliu-adsk: Oh I see. Let me look into that

pierrebai-adsk (approved):

seando-adsk (dismissed):

pierrebai-adsk (approved):

PR #3079, opened 2023-05-10
MAYA-129169 correctly duplicate prim

pierrebai-adsk:
Correctly handle all these case:
- Simple prim with a single opinion: simple copy of source layer into destination layer.
- Prim with opinion in multiple local layers: marge all opinion in destination layer.
- Prim containing a USD reference: keep the reference but do not copy the reference contents into the destination layer.
- Prim containing a USD payload: keep the payload but do not copy the payload contents into the destination layer.
- Prim inside a reference or payload: copy the reference or payload contents in the destination layer.

The code changes are:
- Move layer-related function from the geneic util.h header and C++ file to the more specific utils/layers.h header and C++ file.
- Make the getLocalPrimStack function available publicly.
- Add getPrimStackForLayers function to find where a prim has opinions in a setof layers.
- Add a description for the USD implementation of the UFE duplicate command.
- It explains that the command copies the local opinions of the prim where it is defined.
- Add a unit test that verify that duplicating a prim containing a USD reference does not copy the contents of the reference in the copy.
- Remove the LayerAndPath type and replace it with PrimSpecHandle.
- Prim spec contain the necessary data, and more, so it is more versatile.
- Change some function names to be clearer.


lib/mayaUsd/ufe/UsdUndoDuplicateCommand.h
```diff
@@ -32,6 +32,17 @@ namespace MAYAUSD_NS_DEF {
 namespace ufe {
 
 //! \brief UsdUndoDuplicateCommand
+//!
+//! \details The USD duplicate command copies all opinions related the the USD prim
```
seando-adsk, line 36: small typo -> double "the"

seando-adsk: I'm okay if you fix this in a follow-up PR.

PR #3080, opened 2023-05-11
MAYA-129332 fix batch saving of layers

pierrebai-adsk:
The change in the order the layers are presented required to change the order the loop to do the saving should be executed.


vlasovi (changes requested):
The function getLayersToSaveFromProxy, which behaviour had been changed, is called from several places in MayaUSD, so theoretically they all may be broken now. I would recommend to keep its old behaviour, which is good for saving, as suggests its name. And instead make a change to revert the order only of UI elements, as requested by designers.

vlasovi (approved):

PR #3085, opened 2023-05-12
MAYA-129295 prevent primInfo errors

pierrebai-adsk:
Do not call an instances functions when not an instance.


vlasovi (changes requested):
Thanks @pierrebai-adsk for fixing this. Somehow I was certain that using &= will not evaluate the right side if the left side is false. I guess it doesn't work like that since it's a bitwise &. Can you please also fix the same expression on line 895 in this file.

vlasovi (approved):

PR #3087, opened 2023-05-15
MAYA-129345 - Fixes broken prim creation caused by enum change.

feldstj:


lib/mayaUsd/resources/ae/usdschemabase/custom_enum_control.py
```diff
@@ -19,6 +19,9 @@
 import maya.cmds as cmds
 import maya.internal.ufeSupport.attributes as attributes
 
+def _getEnums(attrs, attrName):
+    return [] if not hasattr(attrs, "getEnums") else attrs.getEnums(attrName)
```
seando-adsk, line 23: This entire customEnumControl relies on having the getEnums method right? So in the __init__.py could it not check for `hasattr(ufe.Attributes.attributes.getEnums)` and only call the prepend when it does? That way we do the check once instead of for every single attribute.

seando-adsk (approved):

PR #3098, opened 2023-05-17
Use new UsdPrimDefintionAPI with latest USD

dj-mcg:
This API change will take effect with the next USD release - 23.08.

### DIFF
--- a/lib/mayaUsd/fileio/utils/adaptor.cpp
+++ b/lib/mayaUsd/fileio/utils/adaptor.cpp
@@ -858,9 +858,9 @@ UsdMayaSchemaAdaptor::CreateAttribute(const TfToken& attrName, MDGModifier& modi
         UsdMayaReadUtil::SetMayaAttr(plug, attrDef->GetDefaultValue(), modifier);
     }
 #else
-    std::string       mayaAttrName = _GetMayaAttrNameOrAlias(attrName);
-    std::string       mayaNiceAttrName = attrDef.GetName();
-    MFnDependencyNode node(_handle.object());
+    std::string                  mayaAttrName = _GetMayaAttrNameOrAlias(attrName);
+    std::string                  mayaNiceAttrName = attrDef.GetName();
+    MFnDependencyNode            node(_handle.object());
 
     bool    newAttr = !node.hasAttribute(mayaAttrName.c_str());
     MObject attrObj = UsdMayaReadUtil::FindOrCreateMayaAttr(
@@ -874,7 +874,7 @@ UsdMayaSchemaAdaptor::CreateAttribute(const TfToken& attrName, MDGModifier& modi
         return UsdMayaAttributeAdaptor();
     }
 
-    MPlug plug = node.findPlug(attrObj);
+    MPlug   plug = node.findPlug(attrObj);
     VtValue fallbackValue;
     if (newAttr && attrDef.GetFallbackValue(&fallbackValue)) {
         // Set the fallback value as the initial value of the attribute, if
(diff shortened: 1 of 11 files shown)


lib/mayaUsd/fileio/utils/adaptor.cpp
```diff
@@ -838,6 +857,32 @@ UsdMayaSchemaAdaptor::CreateAttribute(const TfToken& attrName, MDGModifier& modi
...
+        attrDef.GetVariability(),
+        node,
+        mayaAttrName,
+        mayaNiceAttrName,
+        modifier);
+    if (attrObj.isNull()) {
+        return UsdMayaAttributeAdaptor();
+    }
+
+    MPlug   plug = node.findPlug(attrObj);
```
pierrebai-adsk, line 877: AM I wrong that the difference actually only start from here? I think the ifdef/else could be made shorter. Maybe a helper function could make the code identical.

lib/mayaUsd/fileio/utils/adaptor.h
```diff
@@ -136,10 +145,15 @@ class UsdMayaAttributeAdaptor
...
 
+#if PXR_VERSION < 2308
     /// Gets the defining spec for this attribute from the schema registry.
     /// Returns a null handle if this attribute adaptor is invalid.
     MAYAUSD_CORE_PUBLIC
     const SdfAttributeSpecHandle GetAttributeDefinition() const;
+#else
+    MAYAUSD_CORE_PUBLIC
+    UsdPrimDefinition::Attribute GetAttributeDefinition() const;
+#endif
```
pierrebai-adsk, line 156: The old version could return a null, invalid attr spec. Is the new API similar except it got a IsValid() instead of being pointer-like?

lib/mayaUsd/python/wrapAdaptor.cpp
```diff
@@ -250,19 +250,8 @@ static bool _AttributeAdaptor_Set(UsdMayaAttributeAdaptor& self, const VtValue&
 
 static std::string _AttributeAdaptor__repr__(const UsdMayaAttributeAdaptor& self)
 {
-    std::string                  schemaName;
```
pierrebai-adsk, line 253: Any reason to no longer print the schema name? Not that I care that much...

lib/mayaUsd/ufe/UsdAttributeHolder.cpp
```diff
@@ -509,10 +509,10 @@ Ufe::AttributeEnumString::EnumValues UsdAttributeHolder::getEnumValues() const
...
     Ufe::AttributeEnumString::EnumValues retVal;
     if (_usdAttr.IsValid()) {
-        auto attrDefn
-            = _usdAttr.GetPrim().GetPrimDefinition().GetSchemaAttributeSpec(_usdAttr.GetName());
-        if (attrDefn && attrDefn->HasAllowedTokens()) {
-            for (auto const& token : attrDefn->GetAllowedTokens()) {
+        VtTokenArray allowedTokens;
+        if (_usdAttr.GetPrim().GetPrimDefinition().GetPropertyMetadata(
+                _usdAttr.GetName(), SdfFieldKeys->AllowedTokens, &allowedTokens)) {
+            for (auto const& token : allowedTokens) {
```
pierrebai-adsk, line 515: IDK enough to judge, but I'm going to rightly assume you know better than us about the right way to fetch enum tokens :)

pierrebai-adsk (commented):
Just a few questions, otherwise LGTM.

lib/mayaUsd/fileio/utils/adaptor.cpp
dj-mcg: It's subtly different... The old `attrDef` was a handle so we had to using the de-referencing operator when calling into its functions. The new one is a structure so we no longer dereference

lib/mayaUsd/fileio/utils/adaptor.h
dj-mcg: It has a `bool()` operator that is used to check validity

lib/mayaUsd/python/wrapAdaptor.cpp
dj-mcg: It's because there's no way to access it from the new UsdMayaAttributeAdaptor since that only stores the attribute definition and not which schema/prim definition the attribute came from. 

Since the "old" UsdMayaAttributeAdaptor was holding an SdfAttributeSpec as the attribute definition, the code was sussing out the schema name for the spec's path, but that was making use of an implementation detail that we make no guarantee will be correct. There's no requirement that the property spec is a child of a prim with the schema's name, and we've recently added some property definition composition behaviors could cause the schema name returned by this method to be wrong.

pierrebai-adsk (approved):

PR #3104, opened 2023-05-18
MAYA-129141 - As a user, I'd like to set my anonymous layers as relative in Bulk Save's 'Save As' dialog

vlasovi:
Ability to bulk-save layers with relative paths

### DIFF
--- a/lib/usd/ui/layerEditor/saveLayersDialog.cpp
+++ b/lib/usd/ui/layerEditor/saveLayersDialog.cpp
@@ -112,9 +112,12 @@ public:
 
     void setPathToSaveAs(const std::string& absolutePath, const std::string& relativeAnchor);
 
+    std::string calculateParentLayerDir() const;
+
 protected:
     void onOpenBrowser();
     void onTextChanged(const QString& text);
+    void postUpdate();
 
 public:
     ghc::filesystem::path _absolutePath;
@@ -124,6 +127,7 @@ public:
     LayerInfo         _layerInfo;
     QLabel*           _label { nullptr };
     QLineEdit*        _pathEdit { nullptr };
+    bool              _suppressTextCallback { false };
     QAbstractButton*  _openBrowser { nullptr };
 };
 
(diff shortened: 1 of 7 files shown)


pierrebai-adsk (commented):
I'll wait for the answer before approving.

lib/usd/ui/layerEditor/saveLayersDialog.cpp
```diff
@@ -227,6 +262,32 @@ void SaveLayerPathRow::onTextChanged(const QString& text)
...
+        tooltip += _relativeAnchor.generic_string().c_str();
+    }
+    _pathEdit->setToolTip(tooltip);
+
+    // Update relative anchors of child layers
+    _parent->forEachEntry([this](QWidget* w) {
+        auto entry = dynamic_cast<SaveLayerPathRow*>(w);
+        if (entry && entry->needToSaveAsRelative()
+            && (entry->_layerInfo.parent._layerParent == _layerInfo.layer)) {
+            ghc::filesystem::path relativeAnchor
```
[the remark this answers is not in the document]
vlasovi: @pierrebai-adsk not sure I understand what you mean exactly, but if you are talking about multiple-level inheritance, this will work because: Here we process our immediate children and call "entry->setPathToSaveAs" and inside there, there is a call to postUpdate which will process immediate children of our child and so on recursion will go, if needed.

vlasovi:
@seando-adsk This PR can be merged now. The failing preflight is due to a known issue in testUsdExportPackage, which is not related to my changes.

plugin/adsk/scripts/mayaUsd_layerEditorFileDialogs.mel
```diff
@@ -117,8 +117,20 @@ global proc string UsdLayerEditor_SaveLayerFileDialog(int $isRootLayer, string $
...
     }
 
+    string $titleCaption;
+    string $actionCaption;
+    if ($captionMode == 0) {
+        $titleCaption = "Save As";
+        $actionCaption = "Save";
+    } else {
+        $titleCaption = "Set As";
+        $actionCaption = "Set";
```
seando-adsk, line 127: These strings should really be in resource dictionary for future translation. But actually these strings are probably already in one (or more) of the Maya dictionary files. Search for uses of `uiRes` and you can see how we can directly use a string from Maya.

vlasovi: @seando-adsk If it's ok with you, I would have this PR merged as is and apply this change with my next PR, as there will be more changes going into Batch Save UI.

PR #3110, opened 2023-05-19
MAYA-128599 fixing pivot command

pierrebai-adsk:
In order to be able to apply some Maya more sophisticated transformations on USD prim that already use the USD common xform stack, we need to be able to use the Maya xform stack UFE Transformer3d. But that only accepted to be use if the existing USD form stack matched the Maya one. Unfortunately, the USD common API pivot ws not found in teh Maya stack.

Adding the common pivot to the Maya stack allows using the Maya transferm 3d on the USD common stack, allowing more complex operations on the pivot.

The real problem is that the chain of responsibility design pattern locks which xform-editor is used and thus what can be applied on the xform based on what opinions are already authored. Once chosen, the xform editor cannot be changed. This means some Maya operations cannot ever be applied on that prim.

The real solution would be to allow conversions between differnt xform representations. These conversions would be triggered when an operation not supported by the current xform representation need to be applied.

Concretely, instead of having multiple classes dereived from UFE Transform3d and which is used chosen at the time the Transform3d is created, we would have a single Transform3d class which would contain multiple implementations of the API and which implementation is used could change dynamically, with the USD xform representation converted from one to the other. It would be more of a hierarchy of more and more complex and complete representation of xform, with conversion from simpler forms to more complex forms.

Also:
- Remove spurious error generated by MayaSessionState.
- Add a unit test for center pivot command
- Fix unit test that expect that Maya stack cannot handle a custom xform stack: make the xform names really unique. The new change was able to handle the stack unit by the test.


lib/mayaUsd/fileio/utils/xformStack.cpp
```diff
@@ -437,10 +437,11 @@ UsdMayaXformStack::MatchingSubstack(const std::vector<UsdGeomXformOp>& xformops)
 
         if (GetNameMatters()) {
             // First try the fast attrName lookup...
-            const auto foundTokenIdxPairIter
-                = _sharedData->m_attrNamesToIdxs.find(xformOp.GetName());
+            const TfToken& opName = xformOp.GetName();
+            const auto     foundTokenIdxPairIter = _sharedData->m_attrNamesToIdxs.find(opName);
             if (foundTokenIdxPairIter == _sharedData->m_attrNamesToIdxs.end()) {
                 // Couldn't find the xformop in our stack, abort
+                // TF_WARN("Cannot find xform op %s", opName.GetText());
```
pierrebai-adsk, line 444: Note: I left these debug print in comments behind because I found them useful to understand what was going on, so They will be there for teh next dev that need to investigate problem with this complex code.

lib/usd/ui/layerEditor/mayaSessionState.cpp
```diff
@@ -82,9 +82,11 @@ bool MayaSessionState::getStageEntry(StageEntry* out_stageEntry, const MString&
 
     MObject shapeObj;
     MStatus status = UsdMayaUtil::GetMObjectByName(shapePath, shapeObj);
-    CHECK_MSTATUS_AND_RETURN(status, false);
+    if (!status)
```
pierrebai-adsk, line 85: These were printing error message when there were no stage loaded, which was annoying.

vlasovi (dismissed):
Looks good

vlasovi (dismissed):

pierrebai-adsk:
Sorry, rebased on latest dev after UFE refactor just to make sure.

vlasovi (approved):

PR #3114, opened 2023-05-24
MAYA-128420 Import Chaser doesn't recieve information on how to match USD prim to DAG paths

samuelliu-adsk:


vlasovi (changes requested):
Overall looks good, but I think we may need to do some improvements:
1. I noticed that you have two different code paths needed for C++ case and for Python. One stores the map in UsdMayaImportChaser and the other in FactoryContext. I think we need to choose only one and go with it.
2. Also you use two different maps: sdfToDagMap and newRegistry. They seem pretty similar to me. Can one map cover all our needs?
3. Also, I would rename WriteToNodeRegistry to SetNodeRegisrty, as this function doesn't really write anything

samuelliu-adsk:
@vlasovi 
Thanks for the comments. I forgot to remove the nodeRegistry, that was a previous design. Thanks!

samuelliu-adsk:
> 1. I noticed that you have two different code paths needed for C++ case and for Python. One stores the map in UsdMayaImportChaser and the other in FactoryContext. I think we need to choose only one and go with it.

Hi @vlasovi ,

It's a really good point but I don't see a way to just keep one.. 
The "sdfToDagMap" in ImportChaser is for cpp chaser, which will inherit the map from UsdMayaImportChaser class. However, the one in factory context is for python only. To pass reference to python using boost library, we need a member variable in the class, so we have to keep one inside our factoryContext class.

Correct me if I'm wrong or there's a potential work around I didn't think of!

Thanks!
Samuel

lib/mayaUsd/fileio/chaser/importChaser.h
```diff
@@ -18,6 +18,7 @@
 
 #include <mayaUsd/base/api.h>
 #include <mayaUsd/fileio/jobs/jobArgs.h>
+#include <mayaUsd/fileio/primReaderContext.h>
```
vlasovi, line 21: I think this include is not needed

vlasovi:
@samuelliu-adsk One way to go about accessing the map is to have function GetSdfToDagMap exported to python not from FactoryContext class but form ImportChaser class. This way you don't need to store the map in two places.

lib/mayaUsd/fileio/jobs/readJob.cpp
```diff
@@ -428,6 +428,23 @@ bool UsdMaya_ReadJob::Read(std::vector<MDagPath>* addedDagPaths)
...
     }
 
+    SdfPathSet   sdfImportedPaths;
+    MSdfToDagMap sdfToDagMap;
+    for (const UsdPrim prim : stage->TraverseAll()) {
+        sdfImportedPaths.insert(prim.GetPath());
+    }
+
+    TF_FOR_ALL(pathsIter, sdfImportedPaths)
+    {
```
vlasovi, line 438: I don't understand why you need two cycles here. You don't need to accumulate sdfImportedPaths first and then loop over it. You can what you need inside the first  loop.

lib/mayaUsd/fileio/jobs/readJob.cpp
```diff
@@ -437,16 +454,20 @@ bool UsdMaya_ReadJob::Read(std::vector<MDagPath>* addedDagPaths)
                 addedDagPaths->push_back(MDagPath::getAPathTo(obj));
                 currentAddedDagPaths.append(MDagPath::getAPathTo(obj));
                 fromSdfPaths.push_back(pathsIter->GetPrimPath());
+                sdfToDagMap[pathsIter->GetPrimPath()] = MDagPath::getAPathTo(obj);
```
vlasovi, line 457: Why do you need to do this here? Wasn't everything already collected in the loop above?

test/lib/mayaUsd/fileio/testImportChaser.py
```diff
@@ -67,6 +67,24 @@ def PostImport(self, returnPredicate, stage, dagPaths, sdfPaths, jobArgs):
...
+        for sdfToDag in sdfToDagMap:
+            allSdfPaths.append(sdfToDag.key())
+            allDagPaths.append(sdfToDag.data())
+
+        assert(len(allSdfPaths) == len(allDagPaths))
+
+        for i in range(len(allSdfPaths)):
+            print(("----------primitive ") + str(i))
+            print(str(allSdfPaths[i]))
+            print(str(allDagPaths[i]))
```
vlasovi, line 86: I don't think you are actually testing anything here. You need to have the code that asserts, and not just printing. You can actually remove printing as this isn't very useful.

lib/mayaUsd/fileio/jobs/readJob.cpp
samuelliu-adsk: I forgot to delete it...Thanks for pointing out

samuelliu-adsk:
> @samuelliu-adsk One way to go about accessing the map is to have function GetSdfToDagMap exported to python not from FactoryContext class but form ImportChaser class. This way you don't need to store the map in two places.

Hi @vlasovi ,

I tried some approaches but I don't think we can do that without changing the whole code structure. To call that "GetSdfToDagMap" function, we need to pass the reference of the class to python. Currently we only pass factoryContext to python, and factory context does not share any hierarchy relationship with MayaUsdImportChaser, so we can't use get the variables from ImportChaser in FactoryContext scope (i.e: In python script we need to call FactoryContext.GetJobArgs() ). That's why other variables (sdfPrimPaths and etc.) are passed into factory context as well.

Thanks,
Samuel

vlasovi:
@samuelliu-adsk importChaserTest python class inherits from mayaUsdLib.ImportChaser python which in it's turn originates from ImportChaserWrapper c++ class, which inherits from UsdMayaImportChaser c++ class, so I think you should be able to access any methods of UsdMayaImportChaser inside importChaserTest

vlasovi (changes requested):
Looks very good now. Only couple more things to finish:
1. In testImportChaser.py you need to add asserts to autotest the new feature. I would also remove prints as they may be polluting too much the autotest log. 
2. Rename WriteToSdfToDagMap to SetSdfToDagMap and also use const& for the parameter to make sure there is no extra copying of the potentially large map.

test/lib/mayaUsd/fileio/testImportChaser.py
```diff
@@ -67,6 +67,24 @@ def PostImport(self, returnPredicate, stage, dagPaths, sdfPaths, jobArgs):
...
```
seando-adsk, line 86: Why is the test printing stuff - is that just for you to debug?

seando-adsk (changes requested):

test/lib/mayaUsd/fileio/testImportChaser.py
samuelliu-adsk: No actually I want to do a demo here to show the customers how to use the feature..

samuelliu-adsk: If it's not needed I can remove it

test/lib/mayaUsd/fileio/testImportChaser.py
```diff
@@ -55,6 +58,18 @@ def PostImport(self, returnPredicate, stage, dagPaths, sdfPaths, jobArgs):
...
+        # Get the Sdf path to Dag path matching of all imported prims
+        # Call GetSdfToDagMap() to get the sdfToDagMap, 
+        # which contains SDF path to Dag path mapping
+        sdfToDagMap = self.GetSdfToDagMap()
+        assert(primCount == len(sdfToDagMap))
+        sdfToDagMappingStr = "Sdf to Dag mapping: \n"
+        for sdfToDag in sdfToDagMap:
+            sdfToDagMappingStr += (str(sdfToDag.key()) + "\n")
+            sdfToDagMappingStr += (str(sdfToDag.data()) + "\n")
+
```
vlasovi, line 70: One more thing you should do here is to verify the content of sdfToDagMap like: 
self.assertEqual(sdfToDagMap[0].key(), "WhateverKeyExpect");
self.assertEqual(sdfToDagMap[0].data(), "WhateverDataExpect");
....

vlasovi (changes requested):
Please verify the content of sdfToDagMap in the autotest

vlasovi (approved):

test/lib/mayaUsd/fileio/testImportChaser.py
samuelliu-adsk: Thanks for the suggestion!

vlasovi (approved):

seando-adsk (approved):
Looks good Samuel. Great work.

PR #3115, opened 2023-05-25
MAYA-129168: Update README.md

santosg87:
updating import/Export docs to not mention pxrUSD functions, as we don't build pxr usd on the installer.


seando-adsk:
@dj-mcg Since this change is in the Pixar plugin can you please review.

santosg87:
@seando-adsk good catch, I had the wrong doc file to update!

PR #3117, opened 2023-05-25
MAYA-128622 - As a user, I'd like to save all my USD files as relative on the Bulk Save dialog

vlasovi:
Improving Batch Save UI with new checkboxes and such

### DIFF
--- a/lib/usd/ui/layerEditor/stringResources.h
+++ b/lib/usd/ui/layerEditor/stringResources.h
@@ -106,6 +106,10 @@ const auto kSaveAnonymousIdenticalFiles { create("kSaveAnonymousIdenticalFiles",
 const auto kSaveLayerUsdFileFormatAnn    { create("kSaveLayerUsdFileFormatAnn", "Select whether the .usd file is written out in binary or ASCII. You can save a file in .usdc (binary) or .usda (ASCII) format. Manually entering a file name with an extension overrides the selection in this drop-down menu.") };
 const auto kSaveLayerUsdFileFormatSbm    { create("kSaveLayerUsdFileFormatSbm", "Select whether the .usd file is written out in binary or ASCII") };
 
+const auto kBatchSaveAllRelative { create("kBatchSaveAllRelative", "Convert All to Relative Paths.")};
+const auto kBatchSaveRelativeToScene { create("kBatchSaveRelativeToScene", "Relative to Scene File.")};
+const auto kBatchSaveRelativeToParent { create("kBatchSaveRelativeToParent", "Relative to Parent Layer.")};
+const auto kBatchSaveRelativeTooltip { create("kBatchSaveRelativeTooltip", "This path will be relative to ")};
 
 // -------------------------------------------------------------
 // Errors
--- a/plugin/adsk/scripts/mayaUSDRegisterStrings.mel
+++ b/plugin/adsk/scripts/mayaUSDRegisterStrings.mel
@@ -26,6 +26,7 @@ global proc string getMayaUsdString(string $key)
 global proc mayaUSDRegisterStrings()
 {
     register("kButtonSave", "Save");
+    register("kButtonSet", "Set");
     register("kButtonCancel", "Cancel");
     register("kCreateUsdStageFromFile", "Create USD Stage from File");
     register("kCreateUsdStageFromFileOptions", "Create USD Stage from File Options");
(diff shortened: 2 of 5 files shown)


plugin/adsk/scripts/mayaUSDRegisterStrings.mel
```diff
@@ -26,6 +26,7 @@ global proc string getMayaUsdString(string $key)
 global proc mayaUSDRegisterStrings()
 {
     register("kButtonSave", "Save");
+    register("kButtonSet", "Set");
     register("kButtonCancel", "Cancel");
     register("kCreateUsdStageFromFile", "Create USD Stage from File");
```
pierrebai-adsk, line 31: I think this file can be reverted

pierrebai-adsk (dismissed):
LGTM, might want to revert that one file

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
vlasovi: Why to revert it? kButtonSet and kMenuSetAs are used in mayaUsd_layerEditorFileDialogs.mel

pierrebai-adsk: OH for some reason I thought you were using the Maya resources in MEL.

lib/usd/ui/layerEditor/stringResources.h
```diff
@@ -106,6 +106,10 @@ const auto kSaveAnonymousIdenticalFiles { create("kSaveAnonymousIdenticalFiles",
 const auto kSaveLayerUsdFileFormatAnn    { create("kSaveLayerUsdFileFormatAnn", "Select whether the .usd file is written out in binary or ASCII. You can save a file in .usdc (binary) or .usda (ASCII) format. Manually entering a file name with an extension overrides the selection in this drop-down menu.") };
 const auto kSaveLayerUsdFileFormatSbm    { create("kSaveLayerUsdFileFormatSbm", "Select whether the .usd file is written out in binary or ASCII") };
 
+const auto kBatchSaveAllRelative { create("kBatchSaveAllRelative", "Convert All to Relative Paths.")};
+const auto kBatchSaveRelativeToScene { create("kBatchSaveRelativeToScene", "Relative to Scene File.")};
+const auto kBatchSaveRelativeToParent { create("kBatchSaveRelativeToParent", "Relative to Parent Layer.")};
+const auto kBatchSaveRelativeTooltip { create("kBatchSaveRelativeTooltip", "This path will be relative to ")};
```
seando-adsk, line 112: You cannot assume that in another language the path can just be appended to the end of this string. That may not be the case. So instead you need to use a marker in the string (see examples in this file using "^1s") and use format to compose the final string.

seando-adsk (changes requested):
Please fix the new "kBatchSaveRelativeTooltip" string.

pierrebai-adsk (dismissed):

lib/usd/ui/layerEditor/stringResources.h
```diff
@@ -109,7 +109,8 @@ const auto kSaveLayerUsdFileFormatSbm    { create("kSaveLayerUsdFileFormatSbm",
 const auto kBatchSaveAllRelative { create("kBatchSaveAllRelative", "Convert All to Relative Paths.")};
 const auto kBatchSaveRelativeToScene { create("kBatchSaveRelativeToScene", "Relative to Scene File.")};
 const auto kBatchSaveRelativeToParent { create("kBatchSaveRelativeToParent", "Relative to Parent Layer.")};
-const auto kBatchSaveRelativeTooltip { create("kBatchSaveRelativeTooltip", "This path will be relative to ")};
+const auto kBatchSaveRelativeTooltip { create("kBatchSaveRelativeTooltip", "This path will be relative to ^1s.")};
+const auto kBatchSaveRelativeTooltipEnding0 { create("kBatchSaveRelativeTooltipEnding0", "the saved scene file")};
```
seando-adsk, line 113: Sorry but I still think this should be changed to two individual whole strings. This will provide the proper context for the translation team and also allow for proper translation. You cannot just append one string to another as the order of words can change between languages. I think it should be:

```suggestion
const auto kBatchSaveRelativeLayerTooltip { create("kBatchSaveRelativeLayerTooltip", "This path will be relative to the layer: ^1s.")};
const auto kBatchSaveRelativeSceneTooltip { create("kBatchSaveRelativeSceneTooltip", "This path will be relative to the saved scene file.")};
```

seando-adsk (approved):

PR #3119, opened 2023-05-25
LOOKDEVX-1429 - Mark USD and MayaUSD Python bindings as safe

JGamache-autodesk:
This allows importing USD and MayaUSD modules in script nodes. Either directly, or from surprisingly deep in the C++ callstack as USD can call Python to read shader definitions via code found in the pxr.SdrGlslfx module.

### DIFF
--- a/test/lib/testMayaUsdPythonImport.py
+++ b/test/lib/testMayaUsdPythonImport.py
@@ -17,6 +17,8 @@
 #
 
 import unittest
+import os
+from maya import cmds
 
 
 class MayaUsdPythonImportTestCase(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


JGamache-autodesk:
Reviewed by Wayne Arnold (2023-05-25 15:16):
> yes, it looks ok.

JGamache-autodesk:
Preflight failed due to pipeline flakiness. Should be considered OK. Will retry to see if pipeline issue should be investigated.

seando-adsk (changes requested):
As discussed internally the `scripts:` line with two paths doesn't behave like you think and parse into two full paths.

JGamache-autodesk:
Results in the following paths in the final MAYA_SCRIPT_PATH and in the secure locations:
```
${USD_INSTALL_LOCATION}/lib/python
${MAYAUSD_INSTALL_LOCATION}/lib/scripts
${MAYAUSD_INSTALL_LOCATION}/lib/python
${MAYAUSD_INSTALL_LOCATION}/plugin/al/scripts
${MAYAUSD_INSTALL_LOCATION}/plugin/al/lib/python
${MAYAUSD_INSTALL_LOCATION}/plugin/adsk/scripts
${MAYAUSD_INSTALL_LOCATION}/plugin/pxr/maya/lib/usd/usdMaya/resources
${MAYAUSD_INSTALL_LOCATION}/plugin/pxr/lib/python
```

seando-adsk (approved):

test/lib/testMayaUsdPythonImport.py
```diff
@@ -34,3 +36,30 @@ def testImportModule(self):
...
         self.assertEqual(type(stageCache).__name__, 'StageCache')
+
+    def testPixarModulesAreSafeForScripting(self):
+        """Test that a script node can import the
+           Pixar Python modules in a secure scripting
+           context"""
+
+        node = cmds.createNode("script")
+
+        # This Pixar import will fail and stop the script unless it was added to the secure import paths.
```
seando-adsk, line 47: The unit tests do not use the .mod files. All the variables are setup in the [ctest.cmake](https://github.com/Autodesk/maya-usd/blob/dev/cmake/test.cmake#L233) file. I think you need to modify that for the tests.

JGamache-autodesk: Indeed, this is why I set MAYA_MODULE_PATH for the testMayaUsdPythonImport unit test [here](https://github.com/Autodesk/maya-usd/pull/3119/files#diff-9df95854a7077fa3a82f09331c026f5af41da0ca094280db79dece13749a3707R54). This is also where MAYA_SECURE_BATCH_ENABLE is added to stop the script from running if an unsecure module import is detected.

PR #3120, opened 2023-05-26
HYDRA-178 : Fix geom instances being at the wrong place when using ma…

lanierd-adsk:
…yaHydra in the viewport.


vlasovi (approved):
I think this is good. Theoretically this might have some unknown side effects. But I still think we need to go ahead with it

vlasovi (approved):

lanierd-adsk:
@vlasovi : preflight is good now. Sorry about the long time it took.

vlasovi (approved):
Looks good

PR #3123, opened 2023-05-26
MAYA-114043 Frame both USD and Maya data when edited as Maya

pierrebai-adsk:
- Added TrieVisitor helper to viit all UFE trie nodes.
- Added helper to combine UFE bounding-box.
- Added helper to get the bounding-box of all edited prims under a UFE path.
- Implemenetd adding the bounding box of edited prims in Usd Object3D and proxy shape.
- Added a function to retrieve the list of edited as Maya paths to the prim updater manager and orphaned nodes manager.
- Added a unit test.


lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -41,7 +41,7 @@ namespace MAYAUSD_NS_DEF {
 // Class OrphanedNodesManager::Memento
 //------------------------------------------------------------------------------
 
-OrphanedNodesManager::Memento::Memento(Ufe::Trie<PullVariantInfo>&& pulledPrims)
+OrphanedNodesManager::Memento::Memento(PulledPrims&& pulledPrims)
```
pierrebai-adsk, line 44: I added typedef and removed useless functions, so there is some noise in this file, sorry.

lib/mayaUsd/fileio/orphanedNodesManager.h
```diff
@@ -160,31 +160,28 @@ class OrphanedNodesManager : public Ufe::Observer
     // orphaned.
     bool isOrphaned(const Ufe::Path& pulledPath) const;
 
+    using PulledPrims = Ufe::Trie<PullVariantInfo>;
+    using PulledPrimNode = Ufe::TrieNode<PullVariantInfo>;
+    const PulledPrims& getPulledPrims() const { return _pulledPrims; }
+
```
pierrebai-adsk, line 166: The only relevant change outside of typedef is making this function public so that the PrimUpdaterManager can call it.

lib/mayaUsd/fileio/primUpdaterManager.h
```diff
@@ -70,8 +74,16 @@ class PrimUpdaterManager : public PXR_NS::TfWeakBase
...
+    /// \brief Verify if there are any data edited as Maya data.
+    MAYAUSD_CORE_PUBLIC
     bool hasPulledPrims() const;
 
+    using PulledPrimPaths = std::vector<std::pair<Ufe::Path, MDagPath>>;
+
+    /// \brief Retrieve the UFE path of the edited USD data and the corresponding path of Maya data.
+    MAYAUSD_CORE_PUBLIC
+    PulledPrimPaths getPulledPrimPaths() const;
+
```
pierrebai-adsk, line 86: Again, the only relevant change is this new function to retrieve the list of known edited-as-Maya root nodes.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -1139,5 +1144,71 @@ void ReplicateExtrasToUSD::finalize(
 #endif
 }
 
+static Ufe::BBox3d transformBBox(Ufe::SceneItem::Ptr& item, const Ufe::BBox3d& bbox)
+{
+    Ufe::BBox3d transformed(bbox);
```
pierrebai-adsk, line 1149: The true implementation of the story is here: calculating the pulled prim bbox.

lib/mayaUsd/utils/trieVisitor.h
```diff
@@ -0,0 +1,112 @@
...
+namespace MAYAUSD_NS_DEF {
+
+/// TrieVisitor allows visiting all nodes of a UFE Trie and receiving the full,
+/// correctly built UFE path of each node.
+///
+/// Note: this cannot be moved to UsdUfe since it needs to know about Maya run-time
+///       to build the UFE path segments with teh correct run-time ID.
+
+template <class T> struct TrieVisitor
+{
```
pierrebai-adsk, line 35: Add this utility because composing the UFE paths from the UFE Trie nodes is surprisingly difficult and non-obvious, because they only contain UFE path component, not UFE path segments and thus do not contain the run-time ID. This is extremely awkward and unfortunate.

AramAzhari-adsk (approved):

AramAzhari-adsk (dismissed):

lib/mayaUsd/ufe/UsdObject3d.cpp
```diff
@@ -84,8 +84,12 @@ Ufe::BBox3d UsdObject3d::boundingBox() const
...
     UsdMayaUtil::AddMayaExtents(bbox, fPrim, time);
 
-    auto range = bbox.ComputeAlignedRange();
-    return Ufe::BBox3d(toVector3d(range.GetMin()), toVector3d(range.GetMax()));
+    auto        range = bbox.ComputeAlignedRange();
+    Ufe::BBox3d ufeBBox(toVector3d(range.GetMin()), toVector3d(range.GetMax()));
+
+    Ufe::BBox3d pulledBBox = getPulledPrimsBoundingBox(path);
+
+    return combineUfeBBox(ufeBBox, pulledBBox);
```
seando-adsk, line 92: Thanks for putting the combineUfeBBox in the new UsdUfe library. Just for your information I'm currently working on moving Object3d to Usd Ufe. So I will have to adapt to these new changes, especially the calling of the new helper function `getPulledPrimsBoundingBox` (which is in MayaUsd).

lib/mayaUsd/utils/trieVisitor.h
```diff
@@ -0,0 +1,114 @@
...
+
+#include <functional>
+
+namespace MAYAUSD_NS_DEF {
+
+/// TrieVisitor allows visiting all nodes of a UFE Trie and receiving the full,
+/// correctly built UFE path of each node.
+///
+/// Note: this cannot be moved to UsdUfe since it needs to know about Maya run-time
+///       to build the UFE path segments with teh correct run-time ID.
```
seando-adsk, line 32: Minor typo - "teh" -> "the"

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -15,6 +15,10 @@
 //
 #include "Utils.h"
 
+#ifdef UFE_V3_FEATURES_AVAILABLE
+#include <mayaUsd/fileio/primUpdaterManager.h>
+#endif
```
seando-adsk, line 20: According to the coding standard conditional includes should come last.

lib/mayaUsd/utils/trieVisitor.h
seando-adsk: This looks very USD-specific, which I guess that is okay, but maybe it would be better to rename this `UsdTrieVisitor`.  Because by construction our Ufe::Path's are 1 or 2 segments, as soon as you've hit a gateway node, you don't need to check for gateway nodes again: you know there won't be one.  Therefore, when traversing down the hierarchy, you could pass the path component separator and the run-time ID, with a `checkGateway = true` flag, then as soon as you find the gateway, change the path component separator to be the USD separator, switch the run-time ID to be USD, and change `checkGateway = false`.

test/lib/mayaUsd/fileio/testEditAsMayaBBox.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python
+
+#
+# Copyright 2021 Autodesk
```
seando-adsk, line 4: Copyright year should be 2023

seando-adsk (changes requested):

pierrebai-adsk:
Only PF failure is that known test that assert when a stray file is found in the temp directory. Can ignore the failure.

lib/mayaUsd/utils/trieVisitor.h
seando-adsk: Okay for the naming.  The only concern I have is that repeatedly checking for gateway nodes could be a performance issue. We can always revisit that later if it turns out to be a perf problem.

seando-adsk (approved):

test/lib/mayaUsd/fileio/testEditAsMayaBBox.py
```diff
@@ -66,29 +66,32 @@ def testGroupBoundingBox(self):
             self.assertTrue(mayaUsd.lib.PrimUpdaterManager.canEditAsMaya(sphere2UfePathStr))
             self.assertTrue(mayaUsd.lib.PrimUpdaterManager.editAsMaya(sphere2UfePathStr))
 
+        cmds.select('Sphere2', r=True)
+        cmds.move(10., 10., 10., relative=True)
```
seando-adsk, line 70: Please add a comment here about why you are moving the Maya object (to make sure that after the edit as Maya, the Maya object is not the same as the Usd prim. Then test the BBox - it should be different.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -1139,5 +1147,107 @@ void ReplicateExtrasToUSD::finalize(
...
+            continue;
+
+        if (!pulledPath.startsWith(path))
+            continue;
+
+            // Note: Maya implementation of the Object3d UFE interface does not
+            //       implement the boundingBox() function. So we ask the DAG instead.
+#ifdef MAYA_OBJECT3D_HAS_BOUNDING_BOX
+        Ufe::BBox3d pulledBBox = getTransformedBBox(pulledPath);
+#else
```
seando-adsk, line 1232: Why did you leave this code (and above)? It is not used and never will be. We won't ever implement the boundingBox in Maya Object3d interface. Just leave the comment.

seando-adsk (changes requested):
I don't think we need the previous Maya (thru Ufe) bbox code as its not supported.

seando-adsk (approved):

PR #3127, opened 2023-05-29
Fix tests to write output in testing temp dir (not source tree).

seando-adsk:


plugin/al/plugin/AL_USDMayaTestPlugin/py/testProxyShape.py
```diff
@@ -48,8 +48,7 @@ def tearDownClass(cls):
 
     def setUp(self):
         unittest.TestCase.setUp(self)
-        self.workingDir = os.path.abspath(type(self).__name__)
-        os.mkdir(self.workingDir)
+        self.workingDir = os.environ.get('MAYA_APP_DIR')
```
seando-adsk, line 51: Don't write temporary testing files to the source tree. Instead use the `MAYA_APP_DIR` which is a different temp folder created for each unit test.

csyshing (approved):
LGTM! Thanks!

seando-adsk:
Preflight failure was random image diff in test `testVP2RenderDelegateDisplayLayers` which I didn't touch. I modified two other unit tests (no code changes).

PR #3130, opened 2023-05-31
MAYA-129309 fixing saving the pinned stage

pierrebai-adsk:
- Make the Maya session state resgister and unreguster its Maya notification on creation and deletion.
- No longer de-regiuster the notifictions whne the layer manager window is created or closed.
- That was causing the session state to not save the pinned stage because the layer manager would get destroyed before the pinned stage was saved and the notifications would no longer be received.
- When switching pinned stage, mark the Maya scene as dirty.


vlasovi (approved):
Looks good

PR #3131, opened 2023-05-31
MAYA-128511 fix export roots behaviour

pierrebai-adsk:
Only use the export roots if no node were specified on the export command-line. Otherwise, it is impossible to export a subset of nodes when usinfg the export roots flag.


AramAzhari-adsk (approved):

pierrebai-adsk:
Only PF failures is due to USD file being left behind in /tmp folder from another run.

PR #3134, opened 2023-06-02
Update FindUFE.cmake for Ufe v4.2

seando-adsk:


cmake/modules/FindUFE.cmake
```diff
@@ -51,13 +51,19 @@ if(UFE_INCLUDE_DIR AND EXISTS "${UFE_INCLUDE_DIR}/ufe/ufe.h")
...
         # be removed (along with all the UFE_PREVIEW_VERSION_NUM checks).
         set(UFE_PREVIEW_VERSION_NUM 4100)
-    elseif((UFE_VERSION VERSION_GREATER_EQUAL "4.0.100") AND (UFE_VERSION VERSION_LESS "5.0.0"))
-        # Temporary. Once next Maya PR is released with UFE v4.X.0 this should
-        # be changed into UFE_PREVIEW_VERSION_NUM of 4XYY, where YY is the last value from "4.0.YY"
-        math(EXPR UFE_PREVIEW_VERSION_NUM "4 * 1000 + ${UFE_MINOR_VERSION} * 100 + ${UFE_PATCH_LEVEL}")
+    elseif(UFE_VERSION VERSION_EQUAL "4.2.0")
+        # Temporary. Once next Maya is released with UFE v4.2.0 this should
+        # be removed (along with all the UFE_PREVIEW_VERSION_NUM checks).
+        set(UFE_PREVIEW_VERSION_NUM 4202)
```
seando-adsk, line 60: Set at 4202 since that is the last value (4.1.202) it was before being made final (4.2.0).

cmake/modules/FindUFE.cmake
```diff
@@ -51,13 +51,19 @@ if(UFE_INCLUDE_DIR AND EXISTS "${UFE_INCLUDE_DIR}/ufe/ufe.h")
...
+    elseif((UFE_VERSION VERSION_GREATER_EQUAL "4.2.300") AND (UFE_VERSION VERSION_LESS "5.0.0"))
+        # Temporary - will only be used if there is development for a Ufe v4.3.0 to create
+        # the UFE_PREVIEW_VERSION_NUM for that version while it is in development.
+        # Note: the UFE_PATCH_LEVEL will start at 300 and will thus encode the minor version
+        #       of 3 (so we don't use UFE_MINOR_VERSION in this formula).
+        math(EXPR UFE_PREVIEW_VERSION_NUM "4 * 1000 + ${UFE_PATCH_LEVEL}")
```
seando-adsk, line 66: Here we only want to catch anything for a possible Ufe v4.3 (thus we start looking at 4.2.300). 
We don't want to catch any minor updates to Ufe, such as v4.2.1 - this version should **not** have preview num for it since it would not be a preview release. It would just be a patch fix.

ppt-adsk (approved):

pierrebai-adsk (approved):

PR #3135, opened 2023-06-02
MAYA-128503 - Improve roundtripping of GeomSubsets

JGamache-autodesk:
We have two sources of GeomSubsets in Maya:
 - Shaded faces
 - Explicitly create component tags

In USD land GeomSubsets can be modified in ways that have no Maya equivalent:
 - We can change the family name
 - We can assign a material

So when importing/exporting USD data, we need to add just enough info that all Maya and USD sepcificities are accounted for and correctly handled.

The unit test will be quite informative.


lib/mayaUsd/fileio/jobs/meshDataReadJob.cpp
```diff
@@ -66,7 +66,8 @@ void getComponentTags(MFnMeshData& dataCreator, const UsdGeomMesh& mesh)
 #if MAYA_API_VERSION >= 20220000
 
     std::vector<UsdMayaMeshReadUtils::ComponentTagData> componentTags;
-    UsdMayaMeshReadUtils::getComponentTags(mesh, componentTags);
+    JsValue                                             meshRoundtrippingData;
+    UsdMayaMeshReadUtils::getComponentTags(mesh, componentTags, meshRoundtrippingData);
```
pierrebai-adsk, line 70: The variable is not used afterward? Or is this merely an APi change where you need to provide the parameter as empty?

JGamache-autodesk: I am unsure about this class. It looks like it is extracting a mesh from a USD stage, so I don't see a need to save the roundtripping data (especially since I have no idea how to save this one).

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -713,6 +718,86 @@ class _UVMappingManager
...
+                if (!bindMaterial) {
+                    continue;
+                }
+
+                if (!ctx->GetExportArgs().exportCollectionBasedBindings) {
+                    UsdShadeMaterialBindingAPI subsetBindingAPI
+                        = UsdMayaTranslatorUtil::GetAPISchemaForAuthoring<
+                            UsdShadeMaterialBindingAPI>(geomSubset.GetPrim());
+                    subsetBindingAPI.Bind(materialToBind);
+                }
```
pierrebai-adsk, line 781: I expected an else... if the flag is not set, nothing is done but the set is still removed as being handled?

JGamache-autodesk: Same weird code without an else clause as [found in the main function](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp#L806). What is missing is probably the 
```
            if (boundPrimPaths) {
                boundPrimPaths->insert(faceSubset.GetPath());
            }
```
That is used to manage the material collections. Looking at the `exportCollectionBasedBindings` in the unit tests, I do not find much testing of geom subsets.
I can add that, and write a second roundtrip test with that flag lit to make sure it works as I think it should.

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
```diff
@@ -425,6 +428,38 @@ bool UsdMayaTranslatorMaterial::AssignMaterial(
...
+                    if (!materialObj.isNull()) {
+                        MFnDependencyNode depNodeFn(materialObj);
+                        // Names are not yet final at this stage of import. The only reliable
+                        // identifier is the UUID.
+                        subsetRoundtrippingInfo[UsdMayaGeomSubsetTokens->MaterialUuidKey]
+                            = JsValue(depNodeFn.uuid().asString().asChar());
+                    }
+                    subsetRoundtrippingInfo.erase(materialIt);
+                    updatedInfo[ssInfoIt.first] = subsetRoundtrippingInfo;
+                }
```
pierrebai-adsk, line 455: We don't report not finding material in JSON or Maya? The user might want to know that some binding were lost? They might notice it visually, but they may not in a large scene.

lib/mayaUsd/fileio/utils/meshReadUtils.cpp
```diff
@@ -888,17 +894,66 @@ MStatus UsdMayaMeshReadUtils::assignSubDivTagsToMesh(
...
     // Find all the prims defining componentTags
-    TfToken                    componentTagFamilyName("componentTag");
-    std::vector<UsdGeomSubset> subsets
-        = UsdGeomSubset::GetGeomSubsets(mesh, UsdGeomTokens->face, componentTagFamilyName);
+    std::vector<UsdGeomSubset> subsets = UsdGeomSubset::GetGeomSubsets(mesh, UsdGeomTokens->face);
+
+    JsObject allRoundtripData;
 
     for (auto& ss : subsets) {
+        if (UsdMayaRoundTripUtil::IsPrimMayaGenerated(ss.GetPrim())) {
```
pierrebai-adsk, line 953: Wish there were a comment explaining why we're skipping these. Only add if you need another code version due to other things.

lib/mayaUsd/fileio/utils/meshReadUtils.cpp
```diff
@@ -919,8 +974,35 @@ UsdMayaMeshReadUtils::getComponentTags(const UsdGeomMesh& mesh, std::vector<Comp
...
+        JsObject subsetRoundtripData;
+
+        TfToken familyName;
+        ss.GetFamilyNameAttr().Get(&familyName);
+        if (familyName != UsdMayaGeomSubsetTokens->ComponentTagFamilyName) {
+            subsetRoundtripData[ss.GetFamilyNameAttr().GetBaseName()]
+                = JsValue(familyName.GetString());
+        }
+
+        // Could be any applicable API, but we mostly care about materials:
```
pierrebai-adsk, line 988: Not sure what that comment implies: is it that we could extend the work to other things?

pierrebai-adsk:
I'll wait for the code update and answers about the small details, otherwise LGTM.

lib/mayaUsd/fileio/utils/meshReadUtils.cpp
JGamache-autodesk: Changed to a less ambiguous comment, but you can imagine some very technical users adding new Schema APIs, custom data, attributes, to GeomSubsets. Not sure we want to blindly roundtrip everything, so let's concentrate on materials since they have a representation in Maya.

test/lib/usd/translators/testUsdExportImportRoundtripPreviewSurface.py
```diff
@@ -630,17 +641,35 @@ def ValidateUSDStage(self, stage, allSubsets):
...
 
+    def testGeomSubsetRoundtrip(self):
+        """Test that GeomSubsets numbers stay under control when rountripping
+           with direct material assignment"""
+        self.CommonGeomSubsetRoundtrip(False)
+
+    def testGeomSubsetRoundtripCollectionBased(self):
+        """Test that GeomSubsets numbers stay under control when rountripping
+           with collection based material assignemnt"""
+        self.CommonGeomSubsetRoundtrip(True)
```
JGamache-autodesk, line 671: The way the material checks are done with `ComputeBoundMaterial` work with both assignment methods. The contents of the USDA files will be slightly different since one will use collections based assignment.

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
```diff
@@ -449,6 +449,12 @@ bool UsdMayaTranslatorMaterial::AssignMaterial(
                         // identifier is the UUID.
                         subsetRoundtrippingInfo[UsdMayaGeomSubsetTokens->MaterialUuidKey]
                             = JsValue(depNodeFn.uuid().asString().asChar());
+                    } else {
+                        TF_RUNTIME_ERROR(
```
pierrebai-adsk, line 453: Would it be necessary to also have an error for the iterator check above? IDK what are the semantics for the find to fail.

pierrebai-adsk (dismissed):

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
JGamache-autodesk: Yes, I will revisit the error handling today as this seem a bit fragile.

pierrebai-adsk (dismissed):

PR #3137, opened 2023-06-05
MAYA-128462 relative path error fix

pierrebai-adsk:
Fix unnecessary error when saving a relative path: the path was empty, an empty path cannot be made relative.


pierrebai-adsk:
PF failure is only the interactive Windows one due to a stray file left in the temp directory from a previous run.

PR #3138, opened 2023-06-05
MAYA-129682 - Move `isMaterialsScope()` into Utils.

frohnej-adsk:
Move `isMaterialsScope()` to a shared location, so that it can be reused by LookdevX.

### DIFF
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -15,6 +15,7 @@
 //
 #include "Utils.h"
 
+#include <mayaUsd/fileio/jobs/jobArgs.h>
 #include <mayaUsd/nodes/proxyShapeBase.h>
 #include <mayaUsd/ufe/Global.h>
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -215,6 +216,36 @@ bool isAGatewayType(const std::string& mayaNodeType)
     return isInherited;
 }
 
+bool isMaterialsScope(const Ufe::SceneItem::Ptr& item)
```
frohnej-adsk, line 219: I feel like this should eventually go into `usdUfe/ufe/Utils.cpp`, but it currently still has a Maya dependency: `UsdMayaJobExportArgs::GetDefaultMaterialsScopeName()`. So, I think for now, `mayaUsd/ufe/Utils.cpp` makes more sense.

JGamache-autodesk (approved):

seando-adsk:
@frohnej-adsk The preflight build failed, so can you tell me why you added the "ready-for-merge" label?

PR #3139, opened 2023-06-05
Eliminate spaces from doc filenames

dj-mcg:
Filenames that include spaces will give some of our internal build and asset management systems problems.


dj-mcg:
Hm, it looks like testEditRouting is failing for Maya 2024 after this doc file rename, but I'm confused how my change is causing this failure. Just checking to see if this might be an issue on the test framework end?

seando-adsk:
> Hm, it looks like testEditRouting is failing for Maya 2024 after this doc file rename, but I'm confused how my change is causing this failure. Just checking to see if this might be an issue on the test framework end?

The problem with `testEditRouting` has been fixed. No issue on your end. Sorry about that.
Sean

pierrebai-adsk (approved):

PR #3141, opened 2023-06-06
MAYA-128907 fix layer parenting of relative layers

pierrebai-adsk:
When a layers is saved as relative to its parent just before tryign to reparenting it to anotehr layer, the parenting would fail. That was because the layer identifier use inside the layer itself was relative, which prevented finding the layer afterward.

The problem goes away when the stage is reloaded because when loading relative layers, it correctly set the lauer identifier to its absolute path.

Fix the problem for recently-saved layers by making sure its identifier uses the absolute path, not the relative path.


vlasovi (approved):
Looks good. When I was working around that code, I had a thought that doing this trick with temporary current directory wasn't right. Glad that we fixed that.

Issue #1125, opened 2021-01-28
[MAYA-107000] Plugins do not respect Maya selection highlighting toggle


radon199:
Plugins do not respect the Maya selection highlighting toggle in viewport-2.0. When selection highlighting is disabled wireframe selection is still drawn when selecting Prims inside the USD stage, in contrast with maya geometry which does not draw the selection.

1. Disable Show >Selection Highlighting
2. Select any Prim inside a stage to see the selection highlighting.
3. Select maya geometry like a sphere to not see the highlighting

maya-2020.3
maya_usd-0.6.0

santosg87:
Thank you for logging this! we have this logged internally as well.  updating the title to refelct the internal ID. :)

oumad:
I noticed this issue is still present, I'm using maya 2023 and maya-usd 0.18.0
Is there any other workaround to hide selection highlight for USD prims ? Some objects cause a lot of lag in maya for us with selection highlights, especially during remote work, so it's almost impossible to properly manipulate these objects with highlight on.

santosg87:
marking this on e as closed. the fix for this issue should be available on MayaUSD 0.24.0. :)

dj-mcg:
Just FYI @santosd - I appear to be on MayaUsd 0.25.0 and I'm still seeing this issue. Are there any build flags or environment variables I need to use to pick up this fix?

dj-mcg:
Hi @santosd  - we've integrated MayaUsd 0.27 and I'm scrubbing through our blocker list. This still seems to be an issue for us in Maya 2023 despite the ticket being closed. Does this require Maya 2024?

PR #3142, opened 2023-06-06
MAYA-107000 - [GitHub #1125]Show > Selection Highlighting in the Viewport doesn't work with USD Proxyshape

vlasovi:
Using MHWRender::MGeometry::kSelectionHighlighting flag to hide selection when required


pierrebai-adsk (approved):
LGTM

PR #3149, opened 2023-06-08
MAYA-124335 - Test framework support: Add test in maya-usd for animated cameras and lights

vlasovi:
Adding autotests for animated cameras and lights


pierrebai-adsk (approved):
LGTM

PR #3150, opened 2023-06-08
MAYA-129521 MacOS: Updates Maya USD to use MacOS Frameworks for Qt 6 instead of using loose libraries

AramAzhari-adsk:
MAYA-129521 MacOS: Updates Maya USD to use MacOS Frameworks for Qt 6 instead of using loose libraries


AramAzhari-adsk:
Note: This needs to be checked in after https://github.com/Autodesk/maya-usd/pull/3116
The delta between https://github.com/Autodesk/maya-usd/pull/3116 and this PR is:
lib/usd/ui/CMakeLists.txt
and cmake/utils.cmake

seando-adsk (approved):

PR #3153, opened 2023-06-09
MAYA-126029 - Implement experimental OCIO support

JGamache-autodesk:
Uses information gathered from experimental Maya API to generate OCIO v2 color managed shader fragments. The implementation and textures are provided by the host. Some assembly required...


JGamache-autodesk:
I used OCIO to create files with interesting color spaces and will use them in a unit test that will look like this:
image
Coming soon with the MaterialX implementation of OCIO.

JGamache-autodesk:
@seando-adsk Ready for merge. MaterialX support will be in a future PR.

PR #3159, opened 2023-06-14
MAYA-128925 merge default values

pierrebai-adsk:
When USD prim edited as Maya are merged back to USD, a temporary layer with the data from Maya is created. This layer contains the Maya data exported as USD. When an attribute is authored using a USD sparse-value-writer, if the value is the default value, nothing gets authored. When the temporary layer is merged to USD, it thus ignore that the value was set in Maya and fails to merge that value. So the value that existed in the original USD is not updated.

Change the code to use our own sparse-value-writer which ash an option to always write even default values. When turned on, this ensures that values authored in Maya that happen to be the default USD value will be correctly merged over the existing USD value. This option is turned on during merge-to-USD.

Unfortunately, the USD sparse value writer was passed around everywhere in the export code, so the change look more complicated than it really is.

The real changes are:
- Add a EnhancedSparseValueWriter class to wrap the USD sparse value writer.
- Add a new flag to the export command to control if default values are written.
- Add a new flag to the UsdMayaJobExportArgs to control if default values are written.
- Use the flag to initialize the EnhancedSparseValueWriter.
- Add the flag to the Python wrapper.
- Set the flag in the merge-to-USD defaults.
- Add a unit test.


lib/mayaUsd/fileio/enhancedSparseValueWriter.h
```diff
@@ -0,0 +1,86 @@
...
+
+#include <pxr/base/vt/value.h>
+#include <pxr/pxr.h>
+#include <pxr/usd/usd/attribute.h>
+#include <pxr/usd/usd/timeCode.h>
+#include <pxr/usd/usdUtils/sparseValueWriter.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+/// Enhanced spare value writer.
```
JGamache-autodesk, line 29: Enhanced does not feel right. We are removing a feature of the sparse value writer. Can you consider "FlexibleValueWriter", or the more traditional "MayaUsdSparseValueWriter"?

JGamache-autodesk (approved):

JGamache-autodesk (approved):

pierrebai-adsk:
Only preflight failures are VP2 render delegate on Linux that crashed, which are entirely unrelated to the changes.

PR #3161, opened 2023-06-14
MAYA-128473 - Move UFE to its own Project - Part 9

seando-adsk:
#### MAYA-128473 - Move UFE to its own Project - Part 9

**Commit 1:**
    * Moved following to UsdUfe:
    ** UsdCamera / UsdCameraHandler
    ** Functions: usdPathToUfePathSegment, ConvertUnit

**Commit 2:**
    * Moved StagesSubject to UsdUfe.
    * Create new DCC functions for set/get of stage and stagePath.

**Commit 3:**
    * Make UsdUfe buildable on its own.
    * Remove double-checking of pointer for DCC mandatory functions.


CMakeLists.txt
```diff
@@ -85,7 +85,9 @@ endif()
 # modules and definitions
 #------------------------------------------------------------------------------
 include(cmake/utils.cmake)
-find_package(Maya 2018 REQUIRED)
+if (BUILD_MAYAUSD_LIBRARY)
+    find_package(Maya 2018 REQUIRED)
+endif()
```
seando-adsk, line 90: I added bunch of cmake changes to make UsdUfe buildable on its own. This library only needs Usd & Ufe (no Maya, etc).

CMakeLists.txt
```diff
@@ -129,11 +138,13 @@ if (CMAKE_WANT_MATERIALX_BUILD)
 endif()
 include(cmake/usd.cmake)
 
-if(${MAYA_APP_VERSION} STRLESS "2019" AND CMAKE_WANT_UFE_BUILD)
-    set(CMAKE_WANT_UFE_BUILD OFF)
-    message(AUTHOR_WARNING "========================================================================================= \
-                            UFE is not available in Maya${MAYA_MAJOR_VERSION}. Maya2019 and later are supported only. \
-                            ========================================================================================= ")
+if(DEFINED MAYA_APP_VERSION)
```
seando-adsk, line 141: If we don't call `find_package(Maya)` above then `MAYA_APP_VERSION` is not defined.

lib/usdUfe/ufe/Utils.cpp
```diff
@@ -116,10 +182,9 @@ void setUfePathToPrimFn(UfePathToPrimFn fn)
 
 UsdPrim ufePathToPrim(const Ufe::Path& path)
 {
-    if (nullptr == gUfePathToPrimFn) {
-        throw std::runtime_error("Path to prim function cannot be empty.");
-    }
-
+#if !defined(NDEBUG)
+    assert(gUfePathToPrimFn != nullptr);
+#endif
```
seando-adsk, line 187: As discussed I've removed the double nullptr check since for all these required DCC functions the check is done when the set is called (from initialize). So it should never be null here.

PR #3166, opened 2023-06-16
MAYA-102168 enable color attribute test

pierrebai-adsk:


seando-adsk (commented):
I don't understand how this can be enabled now? This particular test is skipped if the Ufe version >= 4. That means it will run on Maya 2022/2023 which are still using CentOS 7.

PR #3173, opened 2023-06-19
EMSUSD-85 - As a user, I'd like my 'Merge as USD' edits to be specified as overs when authoring to a stronger layer

vlasovi:
Behaviour of the function mergePrims in MergePrims.cpp was changed: 
Before, the function SdfJustCreatePrimInLayer was called for the prim's parent. This way the required "overs" were created up to the prim's parent and therefore the prim itself was created as "def"
Now, we call SdfJustCreatePrimInLayer for the prim itself. This way we create all required "overs" including the prim itself, which is the right behaviour.

### DIFF
--- a/lib/usd/utils/MergePrims.cpp
+++ b/lib/usd/utils/MergePrims.cpp
@@ -935,15 +935,6 @@ bool mergeDiffPrims(
 }
 
 //----------------------------------------------------------------------------------------------------------------------
-/// Create any missing parents as "over". Parents may be missing because we are targeting a
-/// different layer than where the destination prim is authored. The SdfCopySpec function does not
-/// automatically create the missing parent, unlike other functions like UsdStage::CreatePrim.
-void createMissingParents(const SdfLayerRefPtr& dstLayer, const SdfPath& dstPath)
-{
-    SdfJustCreatePrimInLayer(dstLayer, dstPath.GetParentPath());
-}
-
-//----------------------------------------------------------------------------------------------------------------------
 // Augment a USD SdfPath with the variants selections currently active at all levels.
 std::pair<SdfPath, UsdEditTarget> augmentPathWithVariants(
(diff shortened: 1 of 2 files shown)


pierrebai-adsk:
That's a clever solution, but I do wonder what happens when we merge to the layer where the prim is defined. Do the "over" get merged and now the prim is pure-over and maybe does not exists anymore? I would like you to test it and see what happens. For example:

- Create a new stage
- Create a Maya sphere
- Duplicate the Maya sphere to the stage
- Delete the Maya sphere
- Edit-as-Maya the USD sphere
- Move the sphere and enlarge it
- Merge to USD
- Print the root layer to the script log

What happens to the sphere definition in the root layer?

pierrebai-adsk:
Oh, I just realized that the function will only create the over if the prim does not exist already. Forget it, your solution works.

pierrebai-adsk (approved):
LGTM, I thought the fix would be much more complicated. Good thing it was assigned to you! I'm afraid I would have gone down the wrong path.

pierrebai-adsk:
I do have a follow-up question: since the data in the temporary layer is defined as "def", don't we lose the "over" in the destination layer, getting over-written by the "def" during the merge? I think I'd like to have a unit-test to verify it.

pierrebai-adsk (changes requested):
Please add a unit test that merges into an empty, higher layer.
Something like:

- Create two layers
- Target the lower layer (least priority)
- Create a prim
- Target higer layer
- Edit as maya
- Move the prim as a Maya object
- Merge to USD
- Verify the lower layer in unchanged
- Verify that the higher layer has a "over"

vlasovi:
Yeah, you are right. It would be good to have this unit test.

pierrebai-adsk (approved):

PR #3174, opened 2023-06-19
Fix call to c_str on an MString object

dj-mcg:


seando-adsk (approved):
Wow I don't know how that crept in. I guess it was a change made after we turned off building the MtoH code.

PR #3183, opened 2023-06-22
EMSUSD-109 allow textures to be relative

pierrebai-adsk:
- Added an Attribute Editor control template for image assets.
- Added a function to detect that a UFE attribute is an asset.
- Added MEL callback trampolines to Python for the image file dialog.
- Added a usdImageRelativeToEditTargetLayer class to handle the file dialog UI, reusing code for other relative path handling.
- Simplified the layout of the relative file options.
- Don't auto-add the USD extension when browsing for image files.
- Auto-show the image preview in the dialog.
- Select the "All Files" filter by default.


vlasovi (approved):
Looks good

PR #3189, opened 2023-06-29
EMSUSD-101 Override primWriter

samuelliu-adsk:
- Add `CanExport` function to primWriter for overriding default primWriter
- Overload the Register function for backward support
- Add `Poke `function for cases when only wants to check for plugin register but not really find a primWriter

### DIFF
--- a/lib/mayaUsd/fileio/primWriterRegistry.cpp
+++ b/lib/mayaUsd/fileio/primWriterRegistry.cpp
@@ -53,7 +53,7 @@ struct _RegistryEntry
 // typedef std::map<std::string, UsdMayaPrimWriterRegistry::WriterFactoryFn> _Registry;
 typedef std::unordered_multimap<std::string, _RegistryEntry> _Registry;
 static _Registry                                             _reg;
-static std::set<std::string> _mayaTypesThatDoNotCreatePrims;
+static std::set<std::string>                                 _mayaTypesThatDoNotCreatePrims;
 static int                                                   _indexCounter = 0;
 
 _Registry::const_iterator
@@ -189,6 +189,21 @@ UsdMayaPrimWriterRegistry::WriterFactoryFn UsdMayaPrimWriterRegistry::Find(
     }
 
     return nullptr;
+}
+
+/* static */
+void UsdMayaPrimWriterRegistry::Poke(const std::string& mayaTypeName)
+{
+    TfRegistryManager::GetInstance().SubscribeTo<UsdMayaPrimWriterRegistry>();
+
+    _Registry::const_iterator first, last;
+    std::tie(first, last) = _reg.equal_range(mayaTypeName);
+
+    if (first == last) {
+        // If the type name is not currently in our registry, check for plugin registry
+        static const TfTokenVector SCOPE = { _tokens->UsdMaya, _tokens->PrimWriter };
+        UsdMaya_RegistryHelper::FindAndLoadMayaPlug(SCOPE, mayaTypeName);
+    }
 }
 
 /* static */
--- a/lib/mayaUsd/fileio/primWriterRegistry.cpp
+++ b/lib/mayaUsd/fileio/primWriterRegistry.cpp
@@ -59,7 +59,7 @@ static int                                                   _indexCounter = 0;
 _Registry::const_iterator _Find(
     const std::string&          mayaTypeName,
     const UsdMayaJobExportArgs& exportArgs,
-    const MFnDependencyNode&    exportNode)
+    const MObject&    exportObj)
 {
     using ContextSupport = UsdMayaPrimWriter::ContextSupport;
 
@@ -67,7 +67,7 @@ _Registry::const_iterator _Find(
     _Registry::const_iterator first, last;
     std::tie(first, last) = _reg.equal_range(mayaTypeName);
     while (first != last) {
-        ContextSupport support = first->second._pred(exportArgs, exportNode);
+        ContextSupport support = first->second._pred(exportArgs, exportObj);
         // Look for a "Supported" reader. If no "Supported" reader is found, use a "Fallback" reader
         if (support == ContextSupport::Supported) {
             ret = first;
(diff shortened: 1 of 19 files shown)


calligrapheric (approved):

AramAzhari-adsk (approved):

seando-adsk (approved):
Same comment as with the Reader PR. I'm not familiar with this code so I'll let the other code reviewers provided the feedback.

lib/mayaUsd/fileio/primWriterRegistry.cpp
```diff
@@ -42,9 +42,75 @@ TF_DEFINE_PRIVATE_TOKENS(
...
-static std::set<std::string> _mayaTypesThatDoNotCreatePrims;
+namespace {
+struct _RegistryEntry
+{
+    UsdMayaPrimWriterRegistry::ContextPredicateFn _pred;
+    UsdMayaPrimWriterRegistry::WriterFactoryFn    _writer;
+    int                                           _index;
+};
+
+// typedef std::map<std::string, UsdMayaPrimWriterRegistry::WriterFactoryFn> _Registry;
```
pierrebai-adsk, line 53: I think this typedef in comment was left-over and probably should be removed.

lib/mayaUsd/fileio/primWriterRegistry.h
```diff
@@ -109,7 +142,14 @@ struct UsdMayaPrimWriterRegistry
...
     MAYAUSD_CORE_PUBLIC
-    static WriterFactoryFn Find(const std::string& mayaTypeName);
+    static WriterFactoryFn Find(
+        const std::string&          mayaTypeName,
+        const UsdMayaJobExportArgs& exportArgs,
+        const MObject&              exportObj);
+
+    /// \brief Check for external primWriter for \p mayaTypeName.
+    MAYAUSD_CORE_PUBLIC
+    static void Poke(const std::string& mayaTypeName);
```
pierrebai-adsk, line 153: IDK if a better description and function name could be found. I read both the docs and function name and could not figure what the function did. The fact it returns nothing is adding to the confusion as it is unclear what the result of calling this is.  Since it seems to trigger scanning plugins for support for that Maya type, the description and function name could hint at that.

lib/mayaUsd/fileio/primWriterRegistry.cpp
```diff
@@ -74,32 +157,57 @@ void UsdMayaPrimWriterRegistry::RegisterRaw(
...
+UsdMayaPrimWriterRegistry::WriterFactoryFn UsdMayaPrimWriterRegistry::Find(
+    const std::string&          mayaTypeName,
+    const UsdMayaJobExportArgs& exportArgs,
+    const MObject&              exportObj)
 {
     TfRegistryManager::GetInstance().SubscribeTo<UsdMayaPrimWriterRegistry>();
 
-    // unfortunately, usdTypeName is diff from the tfTypeName which we use to
-    // register.  do the conversion here.
-    WriterFactoryFn ret = nullptr;
```
pierrebai-adsk, line 84: IDK what that erased comment about usdTypeName vs tfTypeName was about... I guess it is now irrelevant?

lib/mayaUsd/fileio/primWriterRegistry.cpp
```diff
@@ -74,32 +157,57 @@ void UsdMayaPrimWriterRegistry::RegisterRaw(
...
     UsdMaya_RegistryHelper::FindAndLoadMayaPlug(SCOPE, mayaTypeName);
 
+    it = _Find(mayaTypeName, exportArgs, exportObj);
+
+    if (it != _reg.end()) {
+        return it->second._writer;
+    }
+
     // ideally something just registered itself.  if not, we at least put it in
     // the registry in case we encounter it again.
```
pierrebai-adsk, line 183: This comment should be moved a few lines up before the 2nd _Find.

lib/mayaUsd/python/wrapPrimWriter.cpp
```diff
@@ -206,20 +206,36 @@ class PrimWriterWrapper
...
         {
             size_t classIndex = RegisterPythonObject(cl, GetKey(cl, mayaTypeName));
-            if (classIndex != UsdMayaPythonObjectRegistry::UPDATED) {
-                // Return a new factory function:
-                return FactoryFnWrapper { classIndex };
-            } else {
-                // We already registered a factory function for this purpose:
-                return nullptr;
-            }
+            updated = classIndex == UsdMayaPythonObjectRegistry::UPDATED;
```
pierrebai-adsk, line 236: I'm never sure about = vs == priority, so as a dev reading the code, parenthesis would make me feel safer.

lib/mayaUsd/python/wrapPrimWriter.cpp
```diff
@@ -246,10 +262,10 @@ class PrimWriterWrapper
...
     {
-        UsdMayaPrimWriterRegistry::WriterFactoryFn fn
-            = FactoryFnWrapper::Register(cl, mayaTypeName);
-        if (fn) {
-            UsdMayaPrimWriterRegistry::Register(mayaTypeName, fn, true);
+        bool             updated = false;
+        FactoryFnWrapper fn = FactoryFnWrapper::Register(cl, mayaTypeName, updated);
+        if (!updated) {
+            UsdMayaPrimWriterRegistry::Register(mayaTypeName, fn, fn, true);
         }
```
pierrebai-adsk, line 269: I guess the fact the class has an operator() allows it to be both the predicate and factory function, but a comment explaining why fn gets passed twice would make the code clearer

test/lib/mayaUsd/fileio/testPrimWriter.py
```diff
@@ -49,6 +49,10 @@ def __init__(self, *args, **kwargs):
         self._SetUsdPrim(usdPrim)
         primWriterTest.InitCalled = True
 
+    @classmethod
+    def CanExport(cls, exportArgs, exportObj=None):
+        return mayaUsdLib.PrimWriter.ContextSupport.Supported
```
pierrebai-adsk, line 54: I think adding a flag "CanExportCalled" like in the other test would help verify the functionality.

pierrebai-adsk (changes requested):
Just a few questions and suggestions, otherwise looks great.

lib/mayaUsd/fileio/primWriterRegistry.cpp
samuelliu-adsk: This comment was trying to explain what the code below does so I'll reword it to make it clearer

pierrebai-adsk (approved):

PR #3192, opened 2023-07-03
LOOKDEVX-1605 - Fix inactive sourceColorSpace in UsdUVTexture

JGamache-autodesk:
Found and fixed bug where changing the souceColorSpace value of a UsdUVTexture node did not change the rendering.

Also fixed memory corruption issues found while adding unit test for this fix.

Added env var to allow stamping an OCIO color space value in the sourceColorSpace attribute that is outside the current allowed triad.

Cleaned up the OCIO node code. It no longer hols on XML documents. It now also manages the NodeDef library.

Also fix incorrect color space name that resulted in color variations in the unit test comparison images. They now have all the same RGB values across all source color spaces.


lib/mayaUsd/render/MaterialXGenOgsXml/GlslOcioNodeImpl.cpp
```diff
@@ -62,8 +62,20 @@ constexpr const char ATTR_NAME[] = "name";
...
+    std::string              colorOutput;
+    std::vector<std::string> extraParams;
+    std::vector<std::string> samplerNames;
+    std::string              functionName;
+    std::string              sourceCode;
+};
+
+std::map<std::string, OcioData> knownOCIOFragments;
+std::vector<std::string>        knownOCIOImplementations;
+DocumentPtr                     knownLibrary;
```
JGamache-autodesk, line 78: This node now manages the list of known OCIO NodeDefs. Moved from material.cpp to keep the info centralized.

lib/mayaUsd/render/MaterialXGenOgsXml/GlslOcioNodeImpl.cpp
```diff
@@ -129,19 +140,9 @@ class GlslOcioNodeData : public GenUserData
 };
 } // namespace
 
-std::string
-GlslOcioNodeImpl::registerOCIOFragment(const std::string& fragName, DocumentPtr mtlxLibrary)
+std::string GlslOcioNodeImpl::registerOCIOFragment(const std::string& fragName)
 {
     if (knownOCIOFragments.count(fragName)) {
-        // Make sure the NodeDef are there:
-        if (!mtlxLibrary->getNodeDef(getUntypedNodeDefName(fragName) + OCIO_COLOR3)) {
```
JGamache-autodesk, line 137: No longer needed because the info is centralized.

lib/mayaUsd/render/MaterialXGenOgsXml/GlslOcioNodeImpl.cpp
```diff
@@ -174,53 +175,81 @@ GlslOcioNodeImpl::registerOCIOFragment(const std::string& fragName, DocumentPtr
...
+
     // Validate that the fragment structure is 100% as expected. We need the properties, values,
     // outputs, and GLSL source to proceed:
     auto fragment = doc.child(TAG_FRAGMENT);
     if (!fragment) {
         return {};
     }
 
+    ocioData.fragName = fragment.attribute(ATTR_NAME).as_string();
+    if (ocioData.fragName.empty()) {
```
JGamache-autodesk, line 188: Parsing everything in this function so we can release the XML document at the end.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -306,6 +312,8 @@ void CMPrefs::Refresh()
         // The color management settings are quietly reset on file new:
         _mayaColorManagementCallbackIds.push_back(
             MSceneMessage::addCallback(MSceneMessage::kBeforeNew, colorManagementRefreshCB, this));
+        _mayaColorManagementCallbackIds.push_back(
+            MSceneMessage::addCallback(MSceneMessage::kBeforeOpen, colorManagementRefreshCB, this));
```
JGamache-autodesk, line 316: My interactive tests showed that alternating between scenes with differing OCIO settings were giving incorrect results. So added a kBeforeOpen since OCIO will not send notifs if opening a scene results in a new config.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -575,7 +583,13 @@ size_t _GenerateNetwork2TopoHash(const HdMaterialNetwork2& materialNetwork)
                 auto cmIt = node.parameters.find(cmName);
                 if (cmIt != node.parameters.end()) {
                     MayaUsd::hash_combine(topoHash, hash_value(cmIt->first));
-                    MayaUsd::hash_combine(topoHash, hash_value(cmIt->second));
+                    if (cmIt->second.IsHolding<TfToken>()) {
+                        auto const& colorSpace = cmIt->second.UncheckedGet<TfToken>();
+                        MayaUsd::hash_combine(topoHash, hash_value(colorSpace));
+                    } else if (cmIt->second.IsHolding<std::string>()) {
```
JGamache-autodesk, line 589: That was the root of the problem we are trying to solve. The attribute can be a std::string or a TfToken, so the code needs to handle both.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1271,17 +1289,12 @@ void _AddColorManagementFragments(HdMaterialNetwork& net)
...
+            auto const& filenameVal = fileIt->second.Get<SdfAssetPath>();
+            auto const& resolvedPath = filenameVal.GetResolvedPath();
             if (resolvedPath.empty()) {
                 continue;
             }
-            const std::string& assetPath = filenameVal.GetAssetPath();
-            MString            colorRuleCmd;
-            colorRuleCmd.format(
-                "colorManagementFileRules -evaluate \"^1s\";",
-                (!resolvedPath.empty() ? resolvedPath : assetPath).c_str());
```
JGamache-autodesk, line 1283: resolvedPath can not be empty here (as per check on line 1294), so removed unnecessary code checking for assetPath.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -2780,8 +2793,7 @@ TfToken _RequiresColorManagement(
     }
 
     // Look for explicit color spaces first:
-    std::string       sourceColorSpace;
-    static const auto _mxFindColorSpace = [&sourceColorSpace](const auto& n) {
+    static const auto _mxFindColorSpace = [](const auto& n, auto& sourceColorSpace) {
```
JGamache-autodesk, line 2796: The lambda capture by reference was causing memory corruption. Passing the color space as a function parameter is safer in this case.

test/lib/mayaUsd/render/vp2RenderDelegate/VP2RenderDelegateMaterialXTest/baseline/OCIO_Integration.png
JGamache-autodesk, line 1: Now all color grids are identical. Issue was with "ARRI C4" that was incorrectly defined as "ARRI C4 - Curve" in the file rules.

vlasovi (approved):
Looks good

JGamache-autodesk:
@seando-adsk ready for merge. Thanks!

PR #3198, opened 2023-07-05
EMSUSD-190 fix for lights after edit-as-Maya

pierrebai-adsk:
Adding missing UFE notification after re-adding the edited object

The new code mirror what is done in edit-as-Maya when the object transtions from being USD to being Maya. When it did, the UFE light proxy in Maya was deleted,  and the missing notification in merge-to-USD (and discard-edits) was preventing the proxy from being recreated.


vlasovi (approved):

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1137,6 +1137,11 @@ bool PrimUpdaterManager::mergeToUsd(
     if (TF_VERIFY(hier)) {
         scene.notify(Ufe::SubtreeInvalidate(hier->defaultParent()));
     }
+
+    // We must recreate the UFE item because it has changed data models (Maya -> USD).
+    if (TF_VERIFY(ufeUsdItem))
+        scene.notify(Ufe::ObjectAdd(ufeUsdItem));
```
seando-adsk, line 1143: I don't understand why this extra notif is required. The one above sends a subtree invalidate on the hier default parent which includes this item.

pierrebai-adsk: I traced into why the sub-tree invalidation fails and it turns out it seems like a subtle bug in the code of the proxy-shape hierarchy implementation which result in UFE getting confused and failing to create a Hierarchy for a scene item. I'll try to see if it can be fixed instead.

lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
```diff
@@ -332,10 +332,23 @@ ProxyShapeHierarchy::reorderCmd(const Ufe::SceneItemList& orderedList) const
...
-    // Since the proxy shape corresponds to the USD root prim,
-    // therefore we implement the default parent as the proxy shape itself.
-    return UsdSceneItem::create(sceneItem()->path(), getUsdRootPrim());
+    // The documentation for defaultParent() stipulate that it should return
+    // where this node should be inserted to be added back. The proxy shape need
+    // to be inserted under its Maya shape node, which is its default parent,
+    // so we return that.
+    //
+    // It used to return the USD virtual root prim, but that caused problem since
+    // the UFE path pointed to a Maya node (the proxy shape) and teh default parent
```
seando-adsk, line 341: Minor typo = "teh" -> "the" - don't bother re-running a preflight just for this.

seando-adsk (dismissed):
Great thanks for investigating and the explanation code comment.

pierrebai-adsk:
The only change sin cethe last approval is I only merged dev branch into this branch.

seando-adsk (dismissed):

seando-adsk (approved):

PR #3202, opened 2023-07-06
EMSUSD-248 fix lost notification listeners

pierrebai-adsk:
The order of operations when loading a Maya scene depends on the relative order of unrelated events. Some things are triggered by Maya callbacks, some are done when a stage is set on a proxy node, which gets set when the proxy node is computed, which depends on when it gets dirtied and when something accesses it. Furthermore, some other things are done in the on-idle callback of Maya, which depends on when Qt events are processed. On top of all this, some things are triggered by timers.

For example:
- notification listeners are cleared in Maya callbacks.
- notification listeners are registered when the proxy node is computed
- The layer manager initialize itself with a timer.
- The layer manager is cleared by a Qt signal being emitted, which itself is triggered by a Maya callback.
- The Maya session state stage is reset in an on-idle callback.

The problem in our case is that the notification listeners were getting wiped out because they were cleared in a Maya after-new / after-open callback, but now that is getting called after the proxy shape has been computed, which is where the notification listeners for the new stage were registered.

IOW, the new listeners were cleared by a function that was meant to clear the old listeners. The fix is to clear the listeners in a before-new / before-open callback instead.

- Rename beforeNewCallback() to isInNewScene().
- Rename beforeNewCallback(bool) to setInNewScene(bool)
- Those were confusing because they had the same name as the real Maya callback, but with different parameters.
- Move the code that cleared listeners from afterOpen to beforeOpen.
- Added setupListener() and clearListener() to make the code clearer.
- Fixed a bug that left the in-new-scene flag to true when opening a Maya scene.


pierrebai-adsk:
Only the known flaky testVP2RenderDelegateDisplayLayers test failed.

AramAzhari-adsk (approved):

pierrebai-adsk:
The only PF failure is in a single test in a single Windows build due to a temporary file not being deleted before the test starts, so this can get merged.

seando-adsk (approved):
Thanks for the detailed description and investigation of the problem/solution.

PR #3205, opened 2023-07-06
Fixed a compile error when UFE_SCENE_SEGMENT_SUPPORT isn't available.

jjzanin-SMS:
I'm not sure if this is the correct fix, but there is a compile error when the header file "ufe/sceneSegmentHandler.h" isn't available. I'm assuming it's because our studio is using Maya 2023 and this feature is available in the next version.

### DIFF
--- a/lib/mayaUsd/utils/trieVisitor.h
+++ b/lib/mayaUsd/utils/trieVisitor.h
@@ -18,7 +18,9 @@
 
 #include <mayaUsd/base/api.h>
 
+#if UFE_SCENE_SEGMENT_SUPPORT
 #include <ufe/sceneSegmentHandler.h>
+#endif
 #include <ufe/trie.h>
 
 #include <functional>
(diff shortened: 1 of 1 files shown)


seando-adsk:
@jjzanin-SMS Thank you for submitting a pull-request to fix this. I believe the problem here is that that sceneSegmentHandler.h file was added to Ufe v3.2 which was integrated into Maya 2023.2. I guess you are building with an earlier version of Maya 2023. Sorry we should have caught this earlier. 

However in order to accept this pull-request we will need to have a signed CLA. Please see https://github.com/Autodesk/maya-usd/blob/dev/doc/CONTRIBUTING.md.

Sean

jjzanin-SMS:
Interesting, we are on 2023.3.0-14.0.PFIX. I assume that should have the new header file. I'll ask around to see why our include files aren't being updated.

I'll also talk to my lead about the CLA. Thanks!

seando-adsk:
Ufe is part of the Maya devkit. So you probably just need to download the devkit for Maya 2023.3

seando-adsk:
Closing old pull-request with no updates.

PR #3207, opened 2023-07-07
Proxy shape listener nodes should not be persisted

JGamache-autodesk:
These nodes are for runtime purpose only and should not appear in save files.


pierrebai-adsk (approved):

JGamache-autodesk:
@seando-adsk ready for merge.

seando-adsk (approved):

PR #3210, opened 2023-07-07
Reduce the reference warning

pierrebai-adsk:
- Only warn when referencing a mesh with a non mesh.
- The other cases are actually fine.
- Don't prevent referencing, onyl warn.


pierrebai-adsk:
Note: this was requested by Natalia.

PR #3214, opened 2023-07-10
Referenced shader attributes issue

tryptik:
When exporting file nodes from Maya to Usd and the file node is referenced, the export code does not recognize that the Maya attributes for scale/bias are not set properly despite these values being correct in the referenced Maya scene. This seems to be because the exporting code is using MPlug::getSetAttrCmds and checking the result to see if it is empty to determine if the values need to be written.

This P/R checks the Dependency node to see if it is from a referenced file, then uses MPlug::isDefaultValue to determine if the plug values need to be authored on the exported Usd data.

Note:
In my past testing, Plug::isDefaultValue seemed unreliable, and I suspect it remains so based on the code using MPlug::getSetAttrCmds, so if there is a better approach, please advise.

MPlug::isFromReferencedFile does not seem to work, which is why MFnDependencyNode::isFromReferencedFile is used.

*Note*
This was migrated from a previous P/R that had to be removed - [Check if node is referenced - use default value to detect change #2917](https://github.com/Autodesk/maya-usd/pull/2917).

### DIFF
--- a/lib/mayaUsd/utils/util.cpp
+++ b/lib/mayaUsd/utils/util.cpp
@@ -1210,9 +1210,18 @@ bool UsdMayaUtil::IsAuthored(const MPlug& plug)
     // make a copy of plug here.
     MPlug plugCopy(plug);
 
+    // MPlug::isFromReferencedFile did not seem to be accurate, so this checks the node state.
+    // If the node is referenced, use MPlug::isDefaultValue - it is not clear if isDefaultValue works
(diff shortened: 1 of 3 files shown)


tryptik:
This change was approved in P/R https://github.com/Autodesk/maya-usd/pull/2917 and it was recommended that I create a new P/R and will delete the other one.

pierrebai-adsk (approved):

PR #3220, opened 2023-07-11
Use Pixar's directive for namespacing

dj-mcg:

### DIFF
--- a/lib/usdUfe/ufe/UsdUndoAddRefOrPayloadCommand.cpp
+++ b/lib/usdUfe/ufe/UsdUndoAddRefOrPayloadCommand.cpp
@@ -23,7 +23,7 @@
 
 namespace USDUFE_NS_DEF {
 
-using namespace PXR_NS;
+PXR_NAMESPACE_USING_DIRECTIVE
 
 static std::string validatePrimSpec(const UsdPrim& prim, const SdfPrimSpecHandle& primSpec)
 {


seando-adsk (approved):

seando-adsk:
@dj-mcg What about this one: https://github.com/Autodesk/maya-usd/blob/dev/lib/usd/ui/layerEditor/loadLayersDialog.cpp#L291

dj-mcg:
> @dj-mcg What about this one: https://github.com/Autodesk/maya-usd/blob/dev/lib/usd/ui/layerEditor/loadLayersDialog.cpp#L291

Oh hm, the compiler doesn't complain about that one, but maybe we should instead do something like:
https://github.com/Autodesk/maya-usd/blob/26a15c356758484e9bffc69029704ea05ce2b06f/test/lib/mayaUsd/utils/testUtilsFileSystem.cpp#L7

seando-adsk:
> > @dj-mcg What about this one: https://github.com/Autodesk/maya-usd/blob/dev/lib/usd/ui/layerEditor/loadLayersDialog.cpp#L291
> 
> Oh hm, the compiler doesn't complain about that one, but maybe we should instead do something like:
> 
> https://github.com/Autodesk/maya-usd/blob/26a15c356758484e9bffc69029704ea05ce2b06f/test/lib/mayaUsd/utils/testUtilsFileSystem.cpp#L7

I think the one in the loadLayersDialog needs PXR_NS because its not in the pixar namespace (the namespace is UsdLayerEditor). The one in testUtilsFileSystem is in Pixar namespace. If the layer editor one isn't causing you compiler problems I guess there is no need to change it.

PR #3222, opened 2023-07-12
EMSUSD-255 Fixes an issue in Layer Editor where Save All icon was not functioning properly.

AramAzhari-adsk:
EMSUSD-255 Fixes an issue in Layer Editor where Save All icon was not functioning properly.

The issue occurred due to Qt UI layout changes. The solution is to keep track of the SaveLayerPathRow data separately so that it can be iterated over when saving all layers.

### DIFF
--- a/lib/usd/ui/layerEditor/saveLayersDialog.cpp
+++ b/lib/usd/ui/layerEditor/saveLayersDialog.cpp
@@ -504,7 +504,10 @@ void SaveLayersDialog::buildDialog(const QString& msg1, const QString& msg2)
         for (auto iter = _anonLayerInfos.crbegin(); iter != _anonLayerInfos.crend(); ++iter) {
             // Note: the row adds itself as a children of the dialog, so it will be deleted
             //       when the dialog closes.
-            new SaveLayerPathRow(this, anonLayout, gridRow++, (*iter));
+            auto saveLayerPathRow = new SaveLayerPathRow(this, anonLayout, gridRow++, (*iter));
+            // Note: We keep track of the row data so that it can be used when saving layers without
+            // having to iterate over SaveLayersDialog's entire layout.
+            _saveLayerPathRows.push_back(saveLayerPathRow);
         }
 
         _anonLayersWidget = new QWidget();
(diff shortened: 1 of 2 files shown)


lib/usd/ui/layerEditor/saveLayersDialog.cpp
```diff
@@ -632,13 +635,8 @@ QString SaveLayersDialog::buildTooltipForLayer(SdfLayerRefPtr layer)
...
 QWidget* SaveLayersDialog::findEntry(SdfLayerRefPtr key)
 {
-    if (!_anonLayersWidget || !_anonLayersWidget->layout()) {
-        return nullptr;
-    }
-
-    QLayout* anonLayout = _anonLayersWidget->layout();
-    for (int i = 0, count = anonLayout->count(); i < count; ++i) {
-        auto row = dynamic_cast<SaveLayerPathRow*>(anonLayout->itemAt(i)->widget());
+    for (int i = 0, count = _saveLayerPathRows.size(); i < count; ++i) {
```
AramAzhari-adsk, line 638: I kept the order of iteration the same as before, Although in this case it shouldn't matter as it is only picking one item.

lib/usd/ui/layerEditor/saveLayersDialog.cpp
```diff
@@ -649,13 +647,8 @@ QWidget* SaveLayersDialog::findEntry(SdfLayerRefPtr key)
...
 void SaveLayersDialog::forEachEntry(const std::function<void(QWidget*)>& func)
 {
-    if (!_anonLayersWidget || !_anonLayersWidget->layout()) {
-        return;
-    }
-
-    QLayout* anonLayout = _anonLayersWidget->layout();
-    for (int i = 0, count = anonLayout->count(); i < count; ++i) {
-        func(anonLayout->itemAt(i)->widget());
+    for (int i = 0, count = _saveLayerPathRows.size(); i < count; ++i) {
```
AramAzhari-adsk, line 650: Same here, the order of iteration is from top layer to bottom

AramAzhari-adsk:
The previous preflight builds fully passed, I only added a comment line in the last commit.
Please refer to Build# 1197 for verification.

pierrebai-adsk (approved):
LGTM

PR #3223, opened 2023-07-12
EMSUSD-92 - Ability to exclude primvars with specific Namespace

samuelliu-adsk:
Add a new import flag "excludePrimvarNamespace" which allows users to exclude the whole namespaces of primvars.
For example:
We can exclude this primvar `float[] primvars:ExcludeMeNamespace:ExcludeMeToo:testExclude = [.5]`  if we import with flag `-excludePrimvarNamespace "ExcludeMeNamespace:ExcludeMeToo"` or ` -excludePrimvarNamespace "ExcludeMeNamespace"`


lib/mayaUsd/fileio/utils/meshReadUtils.cpp
```diff
@@ -615,6 +616,23 @@ void UsdMayaMeshReadUtils::assignPrimvarsToMesh(
...
 
+        // Exclude primvars if they match with the exclude namespace
+        std::string primVarString = fullName.GetString();
+
+        bool skipPrimvar = false;
+        for (const TfToken& primVarNameSpace : excludePrivarNamespaceSet) {
+            std::string primVarNameSpaceString = primVarNameSpace.GetString();
+
+            // The primVarString is a substring of excludePrimVarNamespace string from the beginning
+            if (primVarString.find(primVarNameSpaceString) == 0) {
```
pierrebai-adsk, line 627: I guess the probability that a prim var has no namespace but has a name that matches a namespace is low? IDK if there is a way to extract just the namespace from the primVar (maybe [rimVar.GetPrimVarNamsapce or some such?)

pierrebai-adsk (commented):
I would just like to know if you think it would possible to target just the namespace of the prim var for the string match before approving.

lib/mayaUsd/fileio/utils/meshReadUtils.cpp
samuelliu-adsk: Yes I think I can do that

samuelliu-adsk: Instead of using GetNamepsace, I used `primvar.NameContainsNamespaces()` to check if the primvar we are importing does have a namespace. We only proceed the checking if this primvar has a namespace. In this way we can avoid the corner case you mentioned and be more efficient at the mean time.

pierrebai-adsk (approved):
LGTM

PR #3227, opened 2023-07-12
EMSUSD-189 detect invalid use of undo item

pierrebai-adsk:
The OpUndoItem must eventually be kept in a OpUndoList or deleted. This is usually done by using either a OpUndoItemRecorder (to keep them) or OpUndoItemMuting (to delete them). Failure to do so would leak undo items and is a symptom of a programming error where no recorder was created when they should.

Add code to detect such errors in non-release builds. Fixed problems found by the validator.

Adjust the CMake project:
- Add a CMake flag to validate undo item called CMAKE_WANT_VALIDATE_UNDO_ITEM.
- It is on by default in non-release builds.
- It is off by default in release builds to avoid emitting errors that users can't do anything about.
- Add a WANT_VALIDATE_UNDO_ITEM C++ symbol when enabled.

Add the validator:
- Add OpUndoItemValidator class to do the validation.
- Add details about orphaned undo items when the validator fails.
- Add return value to the validator in case calling code want to know the result.
- Don't trigger validation errors when an explicit undo item list is used instead of the global one.
- Print a stack trace of where the validation error are found to more easily find the root cause.

Update related undo item classes:
- Make the recorder and muting classes derive from the validator.
- Make muting code non-inlined.
- Add utility functions to the undo item list class to enable validation and testing.
- Add Python wrapper for the new functionality.

Unit tests and fixes for problems found during testing:
- Add a unit test to validate that the OpUndoItemList works.
- The unit test also indirectly exercises the validator in non-release builds.
- Fix the Python wrapper to not lose the undo items and allow undo and redo.
- Fix the prim updater manager proxy-changed callback to not leave orphaned undo items.
- Add undo item muting in multiple callbacks.
- Correctly mute undo item during the undo and redo of function-based items.
- Fix all tests that did not properly use a OpUndoItemList to capture the undo item of functions. In particular, the edit-as-Maya, merge-to-USD and duplicate functions.


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1693,8 +1693,6 @@ void PrimUpdaterManager::onProxyContentChanged(
             }
         }
     }
-
-    executeAdditionalCommands(context);
 }
```
pierrebai-adsk, line 1696: The auto-edit function above calls editAsMaya which already takes care of calling executeAdditionalCommands. Furthermore, the removed call was executed outside the undo/redo muting, and was "leaking" undo items.

lib/mayaUsd/python/wrapOpUndoItem.cpp
```diff
@@ -41,38 +41,20 @@ class PythonOpUndoItemList
...
+    void enter() { _recorder = std::make_unique<MayaUsd::OpUndoItemRecorder>(_undoItemList); }
 
-    void exit(object, object, object)
-    {
-        if (!TF_VERIFY(_undoItemList != nullptr)) {
-            return;
-        }
-        _recorder.reset();
-        _undoItemList.reset();
-    }
```
pierrebai-adsk, line 60: Getting rid of the undo item list was wrong: it prevented undo/redo from working at all! That is why the class keeps a permanent undo list now. Makes the code simpler, too.

lib/mayaUsd/undo/OpUndoItemMuting.cpp
```diff
@@ -0,0 +1,33 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include <mayaUsd/undo/OpUndoItemMuting.h>
+
```
pierrebai-adsk, line 18: All code in the C++ file is not new but moved from the header file, now that there is more code, including a base class, inlining it makes less sense.

lib/mayaUsd/undo/OpUndoItems.cpp
```diff
@@ -60,6 +62,14 @@ MString formatCommand(const MString& commandName, const MObject& commandArg)
 
 bool executeAndAdd(std::unique_ptr<OpUndoItem>&& item, OpUndoItemList& undoInfo)
 {
+#ifdef WANT_VALIDATE_UNDO_ITEM
+    // Note: validate only if using teh global list, which requires the use of
```
AramAzhari-adsk, line 66: minor typo.

AramAzhari-adsk (approved):

AramAzhari-adsk (dismissed):

lib/mayaUsd/undo/OpUndoItems.cpp
```diff
@@ -15,6 +15,8 @@
 //
 #include "OpUndoItems.h"
 
+#include "OpUndoItemMuting.h"
```
seando-adsk, line 18: According to our coding standard only the associated header file should be included with "
All other public headers should use <

CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ endif()
 set(BUILD_WITH_PYTHON_3_VERSION 3.7 CACHE STRING "The version of Python 3 to build with")
 option(CMAKE_WANT_UFE_BUILD "Enable building with UFE (if found)." ON)
 option(CMAKE_WANT_MATERIALX_BUILD "Enable building with MaterialX (experimental)." OFF)
+option(CMAKE_WANT_VALIDATE_UNDO_ITEM "Enable validating undo items list." $<IF:$<STREQUAL:${CMAKE_BUILD_TYPE},Release>.OFF,ON>)
```
seando-adsk, line 39: Three things
1. Are you sure using a generator expression works in an option?
2. I think the period . should be a comma ,
3. In your comments you say, "It is off by default in release builds" - so doesn't that mean it should be OFF for RelWithDebInfo as well? That is a release build and the one we ship.

seando-adsk (changes requested):
Please look at your option() in CMakeLists.txt

CMakeLists.txt
pierrebai-adsk: Oh we ship with RelWithDebInfo? :( that is specifically the one I wanted enabled because it is the one I use locally. I also think many dev are using it and I wanted this turned on for devs by default so it catches the errors. If off by default for dev then it is much less useful.

Is there another build flag we turn on only for dev and off for release I could detect instead?

lib/mayaUsd/undo/OpUndoItems.cpp
pierrebai-adsk: Ah, OK, I thought it was all the ones in the same folder. I'll fix that.

CMakeLists.txt
pierrebai-adsk: As for the dot-vs-comma, you're right, I wonder how it worked... must fall on some valid CMake syntax by accident with the dot!

pierrebai-adsk:
The only PF failure is one known random export failure we've been having recently.

AramAzhari-adsk (approved):

PR #3230, opened 2023-07-14
Include pxr header

dj-mcg:
This was getting picked up implicitly via the include in pxr/base/arch/defines.h but that is getting removed in usd 23.08


dj-mcg:
I'm a little confused by the preflight failure. It looks like `testVP2RenderDelegateDisplayLayers` is failing in its image compare but I'm not sure how my change would cause that

seando-adsk (approved):

seando-adsk:
@dj-mcg That test has been known to randomly fail. I'll merge this one.

PR #3233, opened 2023-07-17
EMSUSD-77 - Newly created USD Lights can't cast shadows

samuelliu-adsk:
Create USD light from USD will cause the bug. The root cause is that those shadow attributes are never added to the USD created lights. So create Shadow attributes at first query if the attribute is not created

### DIFF
--- a/lib/mayaUsd/ufe/UsdLight.cpp
+++ b/lib/mayaUsd/ufe/UsdLight.cpp
@@ -173,11 +173,12 @@ Ufe::Color3f UsdLight::color() const { return getLightColor(prim()); }
 bool getLightShadowEnable(const UsdPrim& prim)
 {
     const UsdLuxShadowAPI      shadowAPI(prim);
-    const PXR_NS::UsdAttribute lightAttribute = shadowAPI.GetShadowEnableAttr();
+    PXR_NS::UsdAttribute lightAttribute = shadowAPI.GetShadowEnableAttr();
 
     if (!lightAttribute) {
         // If the shadow enable attribute is not created yet, create one here
-        shadowAPI.CreateShadowEnableAttr(VtValue(true));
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/ufe/UsdLight.cpp
```diff
@@ -175,6 +175,11 @@ bool getLightShadowEnable(const UsdPrim& prim)
     const UsdLuxShadowAPI      shadowAPI(prim);
     const PXR_NS::UsdAttribute lightAttribute = shadowAPI.GetShadowEnableAttr();
 
+    if (!lightAttribute) {
+        // If the shadow enable attribute is not created yet, create one here
+        shadowAPI.CreateShadowEnableAttr(VtValue(true));
```
vlasovi, line 180: Should we default to true here? Does HdStorm render shadows when the attribute doesn't exist?

vlasovi: Also, you don't seem to set lightAttribute inside this if, so I'm not sure how the further code works in this case

samuelliu-adsk: Oh I forgot to assign lightAttribute lol...
So for the default value, I set it as true because I think its natural for a light to cast shadows? I see that's the behavior in Maya (if user created a maya light and duplicated as USD light, it casts shadows as default) so I assume we should do the same for USD light as well... Just dont want the user to be confused.

vlasovi: Ok, sounds good. So just make sure to set the lightAttribute, and we'll be good to go.

vlasovi: And please do it here and below for the shadow color attribute

vlasovi (approved):
Looks good

PR #3236, opened 2023-07-18
EMSUSD-256 allow Maya ref to be relative to the layer.

pierrebai-adsk:
- Prepare the current edit target before showing the file dialog.
- Make the function used for this preparatio clearer by naming the flag passed to disable the option for using the scene file as the relative root. When making things relative to the edit target, we don't want to accidentlly use the scene as the root when no saved layer exist.
- Make the code creating the dialog for Maya ref use the code to make the path relative.
- Change the label of the seperator in the file dialog.
- Refactor the common Python code to avoid passing around some options and at teh same time make the function override have the same parameters. Having the same paarmeters is necessary since differnt classes will end-up calling different functions with the same parameters.
- Split the code that create the relative path UI in two: one to setup the layout, one for the UI elements.
- This was necessary because the dialog for the Maya ref already setup the layout.
- Create new class for Maya ref relative UI.

### DIFF
--- a/lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
+++ b/lib/mayaUsd/resources/scripts/mayaUsdAddMayaReference.mel
@@ -92,7 +92,7 @@ proc fileOptionsTabPage(string $tabLayout)
     string $topForm = `columnLayout -rowSpacing 5 actionOptionsForm`;
 
     setParent $topForm;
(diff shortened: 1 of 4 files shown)


pierrebai-adsk:
Only PF failures are the known flaky tests.

seando-adsk (dismissed):
I have an PR #3248 which is moving ContextOps to UsdUfe. The change you made will be in the new MayaUsdContextOps.cpp (in the mayausd/ufe folder). If you don't mind I would like to merge mine first and then you will probably have to merge in latest dev.

seando-adsk (approved):

AramAzhari-adsk (approved):

PR #3239, opened 2023-07-19
 EMSUSD-259 relative Maya ref cache

pierrebai-adsk:
- Move the Python function getCurrentTargetLayerDir to the mayaUsdUtils file for reuse.
- Change section label to "Author Cache File to USD".
- Combine description text to improve the layout spacing.
- Add UI to make the cache file relative to the targeted layer.
- Add a new cache edit router parameter to control if the cache file is relative.
- Document the new edit router parameter in the user documentation.
- Modify the Python createCacheCreationOptions to allow specifying if we want relative paths.
- Pass the parameter when caching from the caching file dialog.
- Refactor C++ functions to get the edit target file path and folder to be reusable.
- Use the new C++ functions in the context menu and caching code.
- In the Maya reference edit router, add the cache payload or reference to the containing prim using the relative path if requested.
- Add unit tests to verify that relative cache work.
- Modify the existing unit tests to first save the root layer since we can only make the cache relative to the root layer if it the root layer has a file name.

### DIFF
--- a/lib/mayaUsd/base/tokens.h
+++ b/lib/mayaUsd/base/tokens.h
@@ -134,8 +134,10 @@ TF_DECLARE_PUBLIC_TOKENS(MayaUsdTokens, MAYAUSD_CORE_PUBLIC, MAYA_USD_GENERIC_TO
     ((Xform, "Xform"))                                  \
     /* Default USD file format for cache             */ \
     ((DefaultUSDFormat, "defaultUSDFormat"))            \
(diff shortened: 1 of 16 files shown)


pierrebai-adsk:
This PR is bigger (affects more files) than some previous ones related to relative file paths because:
- It required changes to the Maya Ref edit router to add the necessary options.
- It required some code refactoring to put C++ code in common.

AramAzhari-adsk (approved):

PR #3242, opened 2023-07-21
Use updated name for Hd task param

dj-mcg:
enableSelection has been replaced in Hdx API v10


dj-mcg:
It looks like preflight is failing to launch remote builds? Please let me know if there's anything I need to do on my side, thanks!

seando-adsk:
@dj-mcg We had an internal glitch with our Maya 2022 and 2024 builds. It should be fixed now.

vlasovi (approved):

PR #3247, opened 2023-07-27
Add excludePrimvarNamespace flag to readme

samuelliu-adsk:


seando-adsk (approved):

seando-adsk:
@samuelliu-adsk Since the only change is an md file, there is no need for a preflight. Just mark as ready-for-merge and I'll merge it.

PR #3249, opened 2023-07-27
EMSUSD-278 - Embed import options in subheadings 

samuelliu-adsk:
- Embed import options in subheading
- Change the style from progressive disclosure to Maya style (grey out children options when parent unselected)

### DIFF
--- a/plugin/adsk/scripts/mayaUsdTranslatorImport.mel
+++ b/plugin/adsk/scripts/mayaUsdTranslatorImport.mel
@@ -231,7 +231,7 @@ global proc int mayaUsdTranslatorImport (string $parent,
 //        optionMenuGrp -e -sl 1 mayaUsdTranslator_CoordSystemOptionMenu;
         frameLayout -label `getMayaUsdString("kImportMaterialsLbl")` materialsFrameLayout;
             separator -style "none";
-            checkBoxGrp  -cw 1 $cw1 -numberOfCheckBoxes 1 -label "" -label1 `getMayaUsdString("kImportMaterialsLbl")` -value1 1 -ann `getMayaUsdString("kImportMaterialsAnn")` -cc ("mayaUsdTranslatorImport_MaterialsCB") mayaUsdTranslator_MaterialsCheckBox;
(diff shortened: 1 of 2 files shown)


samuelliu-adsk:
Preview of the change:
image

plugin/adsk/scripts/mayaUsdTranslatorImport.mel
```diff
@@ -229,24 +229,26 @@ global proc int mayaUsdTranslatorImport (string $parent,
 //            menuItem -label "Object space";
 //            menuItem -label "World space";
 //        optionMenuGrp -e -sl 1 mayaUsdTranslator_CoordSystemOptionMenu;
+        frameLayout -label `getMayaUsdString("kImportMaterialsLbl")` materialsFrameLayout;
+            separator -style "none";
+            checkBoxGrp -label "" -label1 `getMayaUsdString("kImportMaterialsLbl")` -cw 1 $cw1 -value1 1 -ann `getMayaUsdString("kImportMaterialsAnn")` -cc ("mayaUsdTranslatorImport_MaterialsCB") mayaUsdTranslator_MaterialsCheckBox;
```
seando-adsk, line 234: I'm not sure what the default is, but I think you should pass the number of checkboxes with `-ncb 1` and I don't think you need `-label ""` as that flag is optional. Same with all the ones below.

seando-adsk (commented):
Just one minor comment.

plugin/adsk/scripts/mayaUsdTranslatorImport.mel
samuelliu-adsk: Hi @seando-adsk , `-label ""` is actually needed to achieve the alignment. Without `-label` the option will be aligned to the left as shown below: 
image

For `-ncb` I'm not sure it's necessary.. We didn't use that option before so I think that's more for the case that we have multiple checkboxes?

seando-adsk: Okay for the empty label - I didn't realize that.
I checked the code for the checkBoxGrp command and it does indeed default to 1.

seando-adsk (approved):

PR #3250, opened 2023-07-27
EMSUSD-263 export relative USD textures

pierrebai-adsk:
Modify the USD export command (mayaUSDExport and related AL or Pixar commands) to support exporting a Maya scene to USD with relative paths to the textures, when possible. Also support explicitly requested absolute paths, or a mix.

- Add a "exportRelativeTextures" (short form: "rtx") flag to the export command.
- The flag takes one of three options: absolute, relative or automatic.
- The absolute and relative options do what their name implies.
- The relative option will still export textures as absolute paths if the USD file and texture files are on different drives and cannot be made relative.
- The automatic mode exports textures in the Maya scene as relative if they were relative in the scene.
- This usually mean the textures were in the Maya project folder, which is how Maya organizes related files in a project.
- Updated the export command documentation.
- Updated the export job arguments and exposed Python API to add the new options.
- Modified the ResolveUsdTextureFileName function to support these three modes.
- Modified the export code to use the options and modified resolve function.
- Added a getPathRelativeToProject helper function to convert an absolute path to a relative path if the file is under the project.
- Added some unit tests around exporting relative textures.


lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -174,6 +174,37 @@ std::pair<std::string, bool> UsdMayaUtilFileSystem::makePathRelativeTo(
...
+    //       are on the same drive. We really only want to know if the
+    //       project path is the prefix of the file path. Maya will
+    //       preserve paths entered manually with relative folder ("..")
+    //       by keping an absolute path with ".." embedded in them,
+    //       so this works even in this situation.
+    const auto pos = fileName.find(projectPath.asChar());
+    if (pos != 0)
+        return {};
+
+    std::string relativePath = fileName.substr(projectPath.length(), fileName.length());
```
samuelliu-adsk, line 198: I'm not sure about this but can we use `makePathRelativeTo` helper function? I see it's being used in `getPathRelativeToDirectory`

pierrebai-adsk: That is what the comment above says: we cannot use these functions because they do not have the desired behaviour: they would succeed in making the path relative when we don't want them to be relative.

pierrebai-adsk: You're right. we can use it after we have made the checks.

samuelliu-adsk (approved):

seando-adsk:
@dj-mcg Just wanted you to be aware of a new option added to the `usdExport` command.

PR #3255, opened 2023-07-31
EMSUSD-291 Refactored MayaUSD help content ids

AramAzhari-adsk:
MayaUSD help content ids now live in its own help table (helpTableMayaUSD) in MayaUSD repo.
Please refer to mayaUsd_help.py for usage.


lib/mayaUsd/resources/scripts/mayaUsdDuplicateAsUsdDataOptions.py
```diff
@@ -150,22 +149,11 @@ def _fillDuplicateAsUsdDataOptionsDialog(subLayout, optionsText, action):
         mayaUsdTranslatorExport("{subLayout}", "{action}=all;!output", "{optionsText}", "")
         '''.format(optionsText=optionsText, subLayout=subLayout, action=action))
 
-
-def _hasDuplicateAsUsdDataOptionsHelp():
```
AramAzhari-adsk, line 154: I removed this method as this doesn't affect the localization. All the content ids will exist. Moreover, this type of check was only used in two places.

plugin/adsk/scripts/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ list(APPEND scripts_src
     mayaUsd_createStageFromFile.mel
     mayaUsd_createStageWithNewLayer.py
     mayaUsd_createStageFromAsset.mel
+	mayaUsd_help.py
```
AramAzhari-adsk, line 13: it seems that this is a tab vs spaces issue. I'll update this file to match the other lines.

plugin/adsk/scripts/mayaUsd_help.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    import os
+    import maya.mel as mel
+    # Finding the path to helpTableMayaUSD file.
+    dirName = mel.eval('pluginInfo -q -p mayaUsdPlugin')
+    dirName = os.path.dirname(dirName) + "/../helpTable/" + "helpTableMayaUSD"
+    # Setting the default helpTable to helpTableMayaUSD
+    mel.eval('showHelp -helpTable "%s"; showHelp "%s";'% (dirName,contentId))
+    # Restoring Maya's default helpTable
+    mel.eval('showHelp -helpTable "helpTable"')
```
seando-adsk, line 47: You do not need to use `mel` here, simply use the Maya python commands.

```
import maya.cmds as cmds
cmds.pluginInfo('mayaUsdPlugin', q=True, p=True)
cmds.showHelp(...)
```

I would also suggest wrapping the two showHelp calls in a try/finally  block , where in the finally you reset the help table back to default. This will ensure that it is always reset.

plugin/adsk/scripts/mayaUsd_help.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2023 Autodesk
```
seando-adsk, line 1: Same comment as the helpTable, this function should be in mayaUsd instead of the plugin. I would have just added your function to the existing "lib/mayaUsd/resources/scripts/mayaUsdUtils.py" file. But if you want to have it in a separate file in that folder that is also okay.

seando-adsk (changes requested):
I think the helpTable and python helper should be moved from adsk to mayaUsd.

lib/mayaUsd/resources/scripts/mayaUsdUtils.py
```diff
@@ -193,4 +193,39 @@ def saveWantPayloadLoaded(want):
...
+
+    Input contentId refers to the contentId that is registered in helpTableMayaUSD
+    file which is used to open help pages.
+    """
+    import os
     
+    try:
+        # Finding the path to helpTableMayaUSD file.
+        dirName = cmds.pluginInfo('mayaUsdPlugin', q=True, p=True)
+        dirName = os.path.dirname(dirName) + "/../../../lib/helpTable/" + "helpTableMayaUSD"
```
seando-adsk, line 224: Sorry this is my fault now. Since this is now in mayausd lib I don't think we should use the path of the mayaUsdPlugin. Instead I would suggest adding the line
`MAYAUSD_LIB_LOCATION:=lib` 
in the `MayaUSD_LIB` section of the mayaUSD.mod file (don't forget to add to the windows one too). I'm actually surprised we don't already have a variable for this path. You'll see we already have one for `USD_LOCATION`.

Then here you can get the path to the helpTable with:
`helpTablePath = os.path.join(os.environ['MAYAUSD_LIB_LOCATION'], 'helpTableMayaUSD')`

PR #3256, opened 2023-08-02
EMSUSD-263 export relative textures UI

pierrebai-adsk:
- Add the option menu for relative textures in the export UI-builder.
- Add the code to fill and extract the selected option to and from the UI.
- Add the required labels and tool-tips.
- Fix the problem that the export code was never able to export relative paths when goign through the UI becuase the output file argument was never set in merge-to-USD, cache-to-USD and duplicate.

### DIFF
--- a/plugin/adsk/scripts/mayaUsdTranslatorExport.mel
+++ b/plugin/adsk/scripts/mayaUsdTranslatorExport.mel
@@ -679,9 +679,13 @@ global proc int mayaUsdTranslatorExport (string $parent,
                 }
 
                 optionMenuGrp -l `getMayaUsdString("kExportRelativeTexturesLbl")` -annotation `getMayaUsdString("kExportRelativeTexturesAnn")` exportRelativeTexturesPopup;
-                    menuItem -l `getMayaUsdString("kExportRelativeTexturesAutomaticLbl")` -ann "automatic";
-                    menuItem -l `getMayaUsdString("kExportRelativeTexturesAbsoluteLbl")` -ann "absolute";
-                    menuItem -l `getMayaUsdString("kExportRelativeTexturesRelativeLbl")` -ann "relative";
+                    string $lbl;
+                    $lbl = getMayaUsdString("kExportRelativeTexturesAutomaticLbl");
+                    menuItem -l $lbl -ann $lbl;
+                    $lbl = getMayaUsdString("kExportRelativeTexturesAbsoluteLbl");
+                    menuItem -l $lbl -ann $lbl;
+                    $lbl = getMayaUsdString("kExportRelativeTexturesRelativeLbl");
+                    menuItem -l $lbl -ann $lbl;
 
                 separator -style "none";
             setParent ..;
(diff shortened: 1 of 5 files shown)


samuelliu-adsk (approved):

samuelliu-adsk:
LGTM

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -674,7 +677,12 @@ global proc int mayaUsdTranslatorExport (string $parent,
                         -label "" -label1 $conversion -annotation $ann
                         $widgetName;
                 }
-                
+
+                optionMenuGrp -l `getMayaUsdString("kExportRelativeTexturesLbl")` -annotation `getMayaUsdString("kExportRelativeTexturesAnn")` exportRelativeTexturesPopup;
+                    menuItem -l `getMayaUsdString("kExportRelativeTexturesAutomaticLbl")` -ann "automatic";
+                    menuItem -l `getMayaUsdString("kExportRelativeTexturesAbsoluteLbl")` -ann "absolute";
+                    menuItem -l `getMayaUsdString("kExportRelativeTexturesRelativeLbl")` -ann "relative";
```
seando-adsk, line 684: According to the JIRA item it says, "The tooltip on the dropdown is the same as on the label." With this code they aren't identical (case is different) and once translated the tooltip will only be in English. I think this should be changed to something like:

```
string $lbl;
$lbl = getMayaUsdString("kExportRelativeTexturesAutomaticLbl");
menuItem -l $lbl -ann $lbl;
<etc, for other two>
```

seando-adsk (changes requested):
@samuelliu-adsk Since Pierre is away, could checkout his branch and make the requested change and then push it? Thanks, Sean

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
samuelliu-adsk: Hi @seando-adsk , Can we do this?

seando-adsk: You can but you have to surround the getMayaUsdString call with ``
But that has extra overhead of calling getMayaUsdString twice for the same string. Why not store it for use twice?

samuelliu-adsk: Yea I just think this will make the code shorter. I can go with a variable as well

samuelliu-adsk: Updated UI:
image

seando-adsk (approved):

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
pierrebai-adsk: The story meant that the tooltip be the same on the label of the drop down ("Texture File Paths") and the drop-down itself. This is similar to other UI elements and other stories. So the tooltip should have been left as it was, with the long text explaining what each choices do.

samuelliu-adsk: Hi @pierrebai-adsk , yes I have changed that since the annotation is actually used for token string. Check out this PR: https://github.com/Autodesk/maya-usd/pull/3294

PR #3259, opened 2023-08-03
EMSUSD-280 import relative textures

pierrebai-adsk:
Refactor out the texture resolution code into a common reusable function:
- Add a file system function to make a path absolute but relative to the project folder.
- Add code to make the texture path relative to the project.
- Use the common texture import code for materialX.
- Split the texture handling into multiple helper functions to make each one clearer.
- Warn if a USDZ archive is imported without the required corresponding flag to extract textures.
- Properly detect absolute or relative path in automatic mode.

Add import command flag:
- Add the "importRelativeTextures" flag to the mayaUSDImport command.
- Added a "none" relative textures mode for backward compatibility that leave all file names alone.
- Made it the default for the command.
- Add handling of the flag.
- Add a "importRelativeTextures" option to the import job arguments class.
- Add handling and validation of the option.
- Wrap the option for Python.
- Update the import command documentation.
- Use the job args option in the import code.

Add some unit tests.


lib/usd/translators/shading/shadingAsset.cpp
```diff
@@ -0,0 +1,383 @@
...
+    if (!usdInput.Get(&val))
+        return {};
+
+    if (!val.IsHolding<SdfAssetPath>())
+        return {};
+
+    return val.UncheckedGet<SdfAssetPath>();
+}
+
+static bool handleMissingResolvedPath(SdfAssetPath* resolvedAssetPath)
```
pierrebai-adsk, line 65: This all code taken from usdUVTextureReader,cpp and refactored into multiple functions. You would need to diff with that file by hand to see what changed.

JGamache-autodesk (changes requested):

lib/usd/translators/shading/shadingAsset.cpp
```diff
@@ -0,0 +1,383 @@
...
+        return true;
+
+    const SdfAssetPath originalAssetPath = handleShaderInput(usdInput);
+    if (originalAssetPath == SdfAssetPath())
+        return true;
+
+    SdfAssetPath resolvedAssetPath = originalAssetPath;
+
+    if (!handleMissingResolvedPath(&resolvedAssetPath))
+        return false;
```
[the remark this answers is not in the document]
pierrebai-adsk: It is already set to the original asset path just above the if.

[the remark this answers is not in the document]
pierrebai-adsk: Do you merely means to not return false?

lib/usd/translators/shading/mtlxImageReader.cpp
```diff
@@ -118,34 +119,8 @@ bool MtlxUsd_ImageReader::Read(UsdMayaPrimReaderContext& context)
...
-        // colorSpace:
-        if (usdInput.GetAttr().HasColorSpace()) {
-            MString colorSpace = usdInput.GetAttr().GetColorSpace().GetText();
-            mayaAttr = depFn.findPlug(TrMayaTokens->colorSpace.GetText(), true, &status);
-            if (status == MS::kSuccess) {
-                mayaAttr.setString(colorSpace);
-            }
-        }
+    if (!ResolveTextureAssetPath(prim, shaderSchema, depFn, _GetArgs().GetJobArguments())) {
+        return false;
```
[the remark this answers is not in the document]
pierrebai-adsk: IDK what is best here and the other place that uses that function. I'll do what you suggest, we can revisit the best practice in the future. Note that in the original usdUVTextureReader, many error *were* considered fatal, so I'm just keeping the existing behavior.

lib/usd/translators/shading/shadingAsset.cpp
```diff
@@ -0,0 +1,383 @@
...
+    if (status != MS::kSuccess)
+        return true;
+
+    tilingAttr.setInt(3);
+
+    // USD did not resolve the path to absolute because the file name was not an
+    // actual file on disk. We need to find the first tile to help Maya find the
+    // other ones.
+    std::string udimPath(unresolvedFilePath.substr(0, udimPos));
+    udimPath += "1001";
```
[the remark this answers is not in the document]
pierrebai-adsk: Yes, that was existing code. Improvement for UDIM is out of scope.

lib/usd/translators/shading/shadingAsset.cpp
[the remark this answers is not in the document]
pierrebai-adsk: Oh, that function always returns true, so it could never fail. I'll just remove the return value to make it clear.

test/lib/usd/translators/testUsdImportRelativeTextures.py
```diff
@@ -0,0 +1,131 @@
...
+        """
+        Tests that the textures file paths are absolute.
+        """
+        self.runImportRelativeTexturesTest('absolute')
+
+    def testImportAutomaticTextures(self):
+        """
+        Tests that the textures file paths mode is automatically detected.
+        """
+        self.runImportRelativeTexturesTest('automatic')
```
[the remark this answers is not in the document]
pierrebai-adsk: There are no new failure mode, the only existing failures are failing to extract from a USDZ archive, which I guess can only happen when the archive is corrupted or the disk is full. This makes it hard to test.

JGamache-autodesk (approved):

lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -191,18 +197,35 @@ std::string UsdMayaUtilFileSystem::getPathRelativeToProject(const std::string& f
     //       preserve paths entered manually with relative folder ("..")
     //       by keping an absolute path with ".." embedded in them,
     //       so this works even in this situation.
-    const auto pos = fileName.find(projectPath.asChar());
+    const auto pos = fileName.rfind(projectPath, 0);
     if (pos != 0)
         return {};
 
-    auto relativePathAndSuccess = makePathRelativeTo(fileName, projectPath.asChar());
+    auto relativePathAndSuccess = makePathRelativeTo(fileName, projectPath);
```
samuelliu-adsk, line 204: Can we use the `makeProjectRelatedPath` method? Seems like `makeProjectRelatedPath` handles all the checkings and returns the related path

pierrebai-adsk: makeProjectRelatedPath cerates an *absolute* path, not a relative path. (I know the name of the function is awkward, because the function makes an absolute path, but relative to the absolute project path.)

pierrebai-adsk: No because this function wants to make a true *relative* path, but makeProjectRelativePath makes an *absolute* path, but relative to the absolute path of the project.

samuelliu-adsk: Oh I see

samuelliu-adsk:
LGTM

samuelliu-adsk (approved):

lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -174,14 +174,20 @@ std::pair<std::string, bool> UsdMayaUtilFileSystem::makePathRelativeTo(
     return std::make_pair(relativePath.generic_string(), true);
 }
 
+std::string UsdMayaUtilFileSystem::getProjectPath()
+{
+    MString projectPath = MGlobal::executeCommandStringResult("workspace -q -rd");
+    return projectPath.asChar();
+}
```
seando-adsk, line 181: This function already exists as `UsdMayaUtil::GetCurrentMayaWorkspacePath()`

seando-adsk (changes requested):
Pierre added a new helper function that is identical (except return value std::string vs MString) to an existing helper. For code maintenance I don't think we should have two helper functions that do the same thing.

@samuelliu-adsk Since PierreB is on vacation can you checkout his branch and make the fix?

lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -179,9 +179,9 @@ std::string UsdMayaUtilFileSystem::getPathRelativeToProject(const std::string& f
     if (fileName.empty())
         return {};
 
-    MString projectPath = MGlobal::executeCommandStringResult("workspace -q -rd");
     // Note: don't use isEmpty() because it is not available in Maya 2022 and earlier.
-    if (projectPath.length() == 0)
+    const MString projectPath = UsdMayaUtil::GetCurrentMayaWorkspacePath();
+    if (projectPath.isEmpty())
```
seando-adsk, line 184: The comment right above says you cannot use MString::isEmpty().

seando-adsk (changes requested):
Thanks for the changes, but needs a couple of extra fixes.

lib/mayaUsd/utils/utilFileSystem.cpp
samuelliu-adsk: Oh my bad. I missed that comment..

lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -180,8 +180,8 @@ std::string UsdMayaUtilFileSystem::getPathRelativeToProject(const std::string& f
         return {};
 
     // Note: don't use isEmpty() because it is not available in Maya 2022 and earlier.
```
seando-adsk, line 182: Would have been nice if you removed this comment as its not relevant anymore since we are using std::string instead of MString.

seando-adsk (approved):

seando-adsk (approved):

PR #3266, opened 2023-08-09
EMS-310 Usd Context Ops, fixed toggle visibility to used computed visibile + expose isRootChild as a DCC function.

deboisj:
1. Fixed usdUfe context ops to toggle the visibility. The visibility we want to toggle is the _computed_ visibility, not the authored visibility.

2. Exposed isRootChild as a an optional DCC function. Previous behavior kept as default (assumes more than one UFE runtime active, i.e. the maya case).


seando-adsk (approved):
Looks good thanks. When you are ready to have this merged add the label "ready for merge"

PR #3268, opened 2023-08-10
EMSUSD-100: Update docs with -primVariant flag info and example

santosg87:


lib/mayaUsd/commands/Readme.md
```diff
@@ -58,6 +59,17 @@ of the highest prim(s) imported. This is generally the fullDagPath
 that corresponds to the imported primPath but could be multiple
 paths if primPath="/".
 
+### Repeatable flags
+The `-primVar` flag can be used multiple times when used to import specific variants. on MEL, this simply means using the flag repeatedly for each prim that contains the chosen variants. On Python, the flag requires a string array to import. for example. importing multiple variant choices using python would for example use:
```
seando-adsk, line 63:
```suggestion
The `-primVar` flag can be used multiple times when used to import specific variants. In MEL, this simply means using the flag repeatedly for each prim that contains the chosen variants. In Python, the flag requires a string array to import. Importing multiple variant choices using python would for example use:
```

seando-adsk (approved):

PR #3272, opened 2023-08-11
EMSUSD-121 - Use display color when texture mode is disabled

samuelliu-adsk:
- Added optionVar "mayaUsd_ShowDisplayColorTextureOff" to toggle display mode
- The optionVar will be add to user Pref once the feature is ready. 
- To view the change in real time, run `ogs -reset`. This command will be added to the user pref call as well, so the user can see the difference when toggle the optionVar

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
@@ -893,6 +893,7 @@ void HdVP2Mesh::Sync(
             TfTokenVector requiredPrimvars;
             if (!_GetMaterialPrimvars(renderIndex, materialId, requiredPrimvars)
                 || !(reprToken == HdReprTokens->smoothHull)) {
+                // debug
                 requiredPrimvars = sFallbackShaderPrimvars;
             }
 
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -891,7 +891,9 @@ void HdVP2Mesh::Sync(
 
         auto addRequiredPrimvars = [&](const SdfPath& materialId) {
             TfTokenVector requiredPrimvars;
-            if (!_GetMaterialPrimvars(renderIndex, materialId, requiredPrimvars)) {
+            // Only load textures when texture mode is selected
+            if (!_GetMaterialPrimvars(renderIndex, materialId, requiredPrimvars)
+                || !(reprToken == HdReprTokens->smoothHull)) {
```
vlasovi, line 896: This looks wrong. If your intention here is to choose sFallbackShaderPrimvars for untextured mode then you need to have something like: || (reprToken == HdReprTokens->smoothHullUntextured)

vlasovi: Also, in case where hasColor is true below, we don't use fallback shader. So I guess this logic should be consistent

lib/mayaUsd/render/vp2RenderDelegate/mesh.h
```diff
@@ -21,6 +21,7 @@
 #include "meshViewportCompute.h"
 #include "primvarInfo.h"
 
+#include <mayaUsd/base/tokens.h>
```
vlasovi, line 24: I think this include can be in cpp file

vlasovi (changes requested):
Looks very good now. One tiny thing left: please move the "tokens.h" include into cpp. In general, the less includes we have in header files, the faster the compilation.

JGamache-autodesk (approved):

vlasovi (approved):

PR #3278, opened 2023-08-15
EMSUSD-266 Fixed bad USD/UFE notifications

deboisj:
Fixed bad notification translation. See comments in PR. 
(also see https://forum.aousd.org/t/what-are-inert-prims/330)

Also fixed a thrown exception when object add fails for whatever reason.

### DIFF
--- a/lib/usdUfe/ufe/StagesSubject.cpp
+++ b/lib/usdUfe/ufe/StagesSubject.cpp
@@ -448,12 +448,11 @@ void StagesSubject::stageChanged(
                 const std::vector<const SdfChangeList::Entry*>& entries = it.base()->second;
                 bool                                            sentNotif { false };
                 for (const auto& entry : entries) {
-                    if (entry->flags.didAddInertPrim || entry->flags.didAddNonInertPrim) {
+                    if (entry->flags.didAddNonInertPrim) {
                         sendObjectAdd(sceneItem);
                         sentNotif = true;
                         break;
-                    } else if (
-                        entry->flags.didRemoveInertPrim || entry->flags.didRemoveNonInertPrim) {
+                    } else if (entry->flags.didRemoveNonInertPrim) {
                         sendObjectPostDelete(sceneItem);
                         sentNotif = true;
                         break;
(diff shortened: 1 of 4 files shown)


lib/usdUfe/ufe/StagesSubject.cpp
```diff
@@ -452,11 +452,19 @@ void StagesSubject::stageChanged(
...
+                    }
+                    // Note : Do nothing here with didRemoveInertPrim and didRemoveNonInertPrim.
+                    // Indeed, we can get these if prim specs are removed from some layers, but it
+                    // does not mean that the prim is not longer in the composed stage. If the prim
+                    // was actually gone, we would either get an invalid prim (in which case we
+                    // would not even get here, and would send a object destroyed" notif in the else
+                    // bellow), or we would fall into the "HasInfoChange : Active" case bellow. If
+                    // nothing else sends a notif in the loop (typically via the info change :
+                    // active) we do not want to send the fallback notif, so act as if a notif was
+                    // sent.
```
seando-adsk, line 464:
```suggestion
                    // Note : Do nothing here with didRemoveInertPrim and didRemoveNonInertPrim.
                    // Indeed, we can get these if prim specs are removed from some layers, but it
                    // does not mean that the prim is no longer in the composed stage. If the prim
                    // was actually gone, we would either get an invalid prim (in which case we
                    // would not even get here, and would send a object destroyed" notif in the else
                    // below), or we would fall into the "HasInfoChange : Active" case below. If
                    // nothing else sends a notif in the loop (typically via the info change :
                    // active) we do not want to send the fallback notif, so act as if a notif was
                    // sent.
```

Minor typos. If you add only this commit (changing just the comment) then there is no need to re-run the preflight.

lib/usdUfe/ufe/StagesSubject.cpp
```diff
@@ -452,11 +452,19 @@ void StagesSubject::stageChanged(
...
+                    else if (
                         entry->flags.didRemoveInertPrim || entry->flags.didRemoveNonInertPrim) {
-                        sendObjectPostDelete(sceneItem);
                         sentNotif = true;
-                        break;
```
seando-adsk, line 459: Did you mean to take our the break?

seando-adsk (approved):

seando-adsk (approved):

PR #3285, opened 2023-08-18
EMSUSD-291 - Moved MAYAUSD_LIB_LOCATION before gtest path

AramAzhari-adsk:
This is necessary for the installer version as the gtest line is cleared but never removed. This results in a gap that will prevent MAYAUSD_LIB_LOCATION to bet set.


AramAzhari-adsk:
This PR fixes a regression that was caused by https://github.com/Autodesk/maya-usd/pull/3255

seando-adsk (approved):

PR #3297, opened 2023-08-24
Adding unit test for AETemplate

seando-adsk:
New unit test that attempts to find when an error exists in the AE template which prevents it from being loaded correctly. When this happens the AE for USD prims will contain no frame sections, no metadata and all attributes are listed one after another.

### DIFF
--- a/test/lib/testAttributeEditorTemplate.py
+++ b/test/lib/testAttributeEditorTemplate.py
@@ -1,0 +1,255 @@
+#!/usr/bin/env python
+
+#
+# Copyright 2023 Autodesk
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import unittest
+
+import mayaUtils
+import ufeUtils
+import fixturesUtils
+import testUtils
+
+import mayaUsd.lib
+
+from maya import cmds
+from maya import standalone
+
+from mayaUsdLibRegisterStrings import getMayaUsdLibString
+
+from pxr import Usd
+
+import ufe
+
+class AttributeEditorTemplateTestCase(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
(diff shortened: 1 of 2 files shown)


test/lib/testAttributeEditorTemplate.py
```diff
@@ -0,0 +1,255 @@
...
+                for child in childrenOfLayout:
+                    child = controlOrLayout + '|' + child
+                    if cmds.layout(child, exists=True):
+                        foundControl = self.searchForMayaControl(child, mayaControlCmd, labelToFind)
+                        if foundControl:
+                            return foundControl
+                    elif mayaControlCmd(child, exists=True):
+                        if mayaControlCmd(child, q=True, label=True) == labelToFind:
+                            return child
+        return None
```
seando-adsk, line 68: This function searched thru the layouts in the AE for either a layout (in this case we are searching for frame layout which is used for each section) with a specific name or a control (of a certain type) with a label equal to input name.

test/lib/testAttributeEditorTemplate.py
```diff
@@ -0,0 +1,255 @@
...
+        # by making sure we can find the one for capsule.
+        capsuleFormLayout = 'AttrEdUSDCapsuleFormLayout'
+        self.assertTrue(cmds.formLayout(capsuleFormLayout, exists=True))
+        startLayout = cmds.formLayout(capsuleFormLayout, query=True, fullPathName=True)
+        self.assertIsNotNone(startLayout, 'Could not get full path for Capsule formLayout')
+
+        # We should have a frameLayout called 'Capsule' in the template.
+        # If there is a scripting error in the template, this layout will be missing.
+        frameLayout = self.searchForMayaControl(startLayout, cmds.frameLayout, 'Capsule')
+        self.assertIsNotNone(frameLayout, 'Could not find Capsule frameLayout')
```
seando-adsk, line 109: Example: there should exist a section called "Capsule"

test/lib/testAttributeEditorTemplate.py
```diff
@@ -0,0 +1,255 @@
...
+
+        # We should also have float slider controls for 'Height' & 'Radius'.
+        heightControl = self.searchForMayaControl(frameLayout, cmds.floatSliderGrp, 'Height')
+        self.assertIsNotNone(heightControl, 'Could not find Capsule Height control')
+        radiusControl = self.searchForMayaControl(frameLayout, cmds.floatSliderGrp, 'Radius')
+        self.assertIsNotNone(radiusControl, 'Could not find Capsule Radius control')
```
seando-adsk, line 115: And that Capsule section should have attributes for height/radius.

AramAzhari-adsk (approved):

seando-adsk:
@AramAzhari-adsk Sorry had to make a specific fix for our Maya 2022 gold build (where the AE form layouts are not named the same way).

AramAzhari-adsk (approved):

PR #3305, opened 2023-08-31
EMSUSD-311 fix load sub-layer file preview.

pierrebai-adsk:
- Fix helper functions when the layer folder is empty.
- Don't use the current directory when there is no file path.
- Register the necessary callbacks.
- Simplified code logic in Python relative file code to be more linear.


pierrebai-adsk:
image

pierrebai-adsk:
Note: the UI is disabled if the parent layer is not saved since it cannot be made relative to an anonymous layer. This is not a change of behaviour, I'm only mentioning for completeness.

samuelliu-adsk (approved):

PR #3309, opened 2023-09-05
EMSUSD-533 - Incremental naming doesn't take zero into consideration

seando-adsk:
#### EMSUSD-533 - Incremental naming doesn't take zero into consideration
* Added optional DCC function for uniqueChildName() and override the UsdUfe one with Maya version: uniqueChildNameMayaStandard().
* Split out numerical suffix handling in UsdUfe::uniqueName() into separate helper function.
* When creating new unique name take into account length of numerical suffix (when padded with 0's).
* Added new unit tests to testDuplicateCmd to test all these naming rules.

### DIFF
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -162,13 +162,13 @@ std::string uniqueChildNameMayaStandard(const PXR_NS::UsdPrim& usdParent, const 
 
         std::string             childBaseName;
         std::pair<TfToken, int> largestMatching("", -1);
-        for (const auto child : allChildrenNames) {
+        for (const auto& child : allChildrenNames) {
             // While iterating thru all the children look for ones that match
             // the base name of the input. When we find one check its numerical
             // suffix and store the greatest one.
             splitNumericalSuffix(child.GetString(), childBaseName, suffix);
             if (baseName == childBaseName) {
-                int suffixValue = std::stoi(suffix);
+                int suffixValue = !suffix.empty() ? std::stoi(suffix) : 0;
                 if (suffixValue > largestMatching.second) {
                     largestMatching = std::make_pair(child, suffixValue);
                 }
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -159,9 +159,10 @@ std::string uniqueChildNameMayaStandard(const PXR_NS::UsdPrim& usdParent, const 
         // that to all the sibling names.
         std::string baseName, suffix;
         splitNumericalSuffix(childName, baseName, suffix);
+        int suffixValue = !suffix.empty() ? std::stoi(suffix) : 0;
 
-        std::string             childBaseName;
-        std::pair<TfToken, int> largestMatching("", -1);
+        std::string                 childBaseName;
+        std::pair<std::string, int> largestMatching(childName, suffixValue);
         for (const auto& child : allChildrenNames) {
             // While iterating thru all the children look for ones that match
             // the base name of the input. When we find one check its numerical
@@ -170,7 +171,7 @@ std::string uniqueChildNameMayaStandard(const PXR_NS::UsdPrim& usdParent, const 
             if (baseName == childBaseName) {
                 int suffixValue = !suffix.empty() ? std::stoi(suffix) : 0;
                 if (suffixValue > largestMatching.second) {
-                    largestMatching = std::make_pair(child, suffixValue);
+                    largestMatching = std::make_pair(child.GetString(), suffixValue);
                 }
             }
         }
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -154,8 +154,7 @@ std::string uniqueChildNameMayaStandard(const PXR_NS::UsdPrim& usdParent, const 
     // Example: with siblings Capsule001 & Capsule006, duplicating Capsule001
     //          will set new unique name to Capsule007.
     std::string childName { name };
-    if (allChildrenNames.find(TfToken(childName)) != allChildrenNames.end())
-    {
+    if (allChildrenNames.find(TfToken(childName)) != allChildrenNames.end()) {
         // Get the base name (removing the numerical suffix) so that we can compare
         // that to all the sibling names.
         std::string baseName, suffix;
@@ -163,8 +162,7 @@ std::string uniqueChildNameMayaStandard(const PXR_NS::UsdPrim& usdParent, const 
 
         std::string             childBaseName;
         std::pair<TfToken, int> largestMatching("", -1);
-        for (const auto child : allChildrenNames)
-        {
+        for (const auto child : allChildrenNames) {
             // While iterating thru all the children look for ones that match
             // the base name of the input. When we find one check its numerical
             // suffix and store the greatest one.
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -137,6 +137,54 @@ createSiblingSceneItem(const Ufe::Path& ufeSrcPath, const std::string& siblingNa
     return UsdSceneItem::create(ufeSiblingPath, ufePathToPrim(ufeSiblingPath));
 }
 
+std::string uniqueChildNameMayaStandard(const PXR_NS::UsdPrim& usdParent, const std::string& name)
+{
+    if (!usdParent.IsValid())
+        return std::string();
+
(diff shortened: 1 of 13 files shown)


lib/mayaUsd/ufe/Utils.h
```diff
@@ -73,15 +73,9 @@ MAYAUSD_CORE_PUBLIC
...
-    return UsdUfe::uniqueName(existingNames, srcName);
-}
-
-inline std::string uniqueChildName(const PXR_NS::UsdPrim& parent, const std::string& name)
-{
-    return UsdUfe::uniqueChildName(parent, name);
-}
+//! Returns a unique child name following the Maya standard naming rules.
+MAYAUSD_CORE_PUBLIC
+std::string uniqueChildNameMayaStandard(const PXR_NS::UsdPrim& usdParent, const std::string& name);
```
seando-adsk, line 78: Removed the two inline mayaUsd versions of these functions (modified calling locations to use them from UsdUfe).

Added new mayaUsd verision of uniqueChildName that follows the Maya standard of naming.

lib/usdUfe/ufe/Utils.cpp
```diff
@@ -278,27 +279,57 @@ int ufePathToInstanceIndex(const Ufe::Path& path, UsdPrim* prim)
     return instanceIndex;
 }
 
-std::string uniqueName(const TfToken::HashSet& existingNames, std::string srcName)
+bool splitNumericalSuffix(const std::string srcName, std::string& base, std::string& suffix)
```
seando-adsk, line 282: Split out the first part of the original unqiueName function which separated the input name into base and numerical suffix. Made into helper so I can call it elsewhere.

lib/usdUfe/ufe/Utils.cpp
```diff
@@ -278,27 +279,57 @@ int ufePathToInstanceIndex(const Ufe::Path& path, UsdPrim* prim)
...
+    if (splitNumericalSuffix(srcName, base, suffixStr)) {
+        lenSuffix = suffixStr.length();
+        suffix = std::stoi(suffixStr) + 1;
     }
-    std::string dstName = base + std::to_string(suffix);
+
+    // Create a suffix string from the number keeping the same number of digits as
+    // numerical suffix from input srcName (padding with 0's if needed).
+    suffixStr = std::to_string(suffix);
+    suffixStr = std::string(lenSuffix - std::min(lenSuffix, suffixStr.length()), '0') + suffixStr;
```
seando-adsk, line 311: Modified naming code here to maintain the 0 padded numerical suffix. So duplicating `Capsule001` will give you `Capsule002`

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -137,6 +137,54 @@ createSiblingSceneItem(const Ufe::Path& ufeSrcPath, const std::string& siblingNa
...
+                if (suffixValue > largestMatching.second) {
+                    largestMatching = std::make_pair(child, suffixValue);
+                }
+            }
+        }
+
+        // By sending in the largest matching name (instead of the input name)
+        // the unique name function will increment its numerical suffix by 1
+        // and thus it will be unique and follow Maya naming standard.
+        childName = uniqueName(allChildrenNames, largestMatching.first);
```
seando-adsk, line 183: In this mayaUsd version I find only the greatest matching child (in the list of all children) and send that to the uniqueName function. In that way it will follow the Maya naming. So if you have:
```
Capsule001
Capsule006
```
and you duplicate `Capsule001`, the new child will be `Capsule007`

test/lib/ufe/testDuplicateCmd.py
```diff
@@ -646,7 +646,85 @@ def testPrimWithPayload(self):
         rootLayerText = rootLayer.ExportToString()
         self.assertNotIn('"Geom"', rootLayerText)
         self.assertNotIn('Mesh', rootLayerText)
-    
+
+    def testDuplicateUniqueName(self):
+        '''Test the duplicate of prims and ensure the new name follows Maya
+        unique new name standard.'''
```
seando-adsk, line 652: Add a bunch of new test cases to verify my code changes for Usd prims, follow how Maya naming works.

lib/usdUfe/ufe/Utils.cpp
```diff
@@ -278,27 +279,57 @@ int ufePathToInstanceIndex(const Ufe::Path& path, UsdPrim* prim)
...
+    std::string dstName = base + suffixStr;
     while (existingNames.count(TfToken(dstName)) > 0) {
         dstName = base + std::to_string(++suffix);
     }
```
pierrebai-adsk, line 315: With the new code, this if should not be triggered, but if it would be, there would not be leading zeroes.

pierrebai-adsk (changes requested):
The while loop I flagged need to also add leading zeroes.

lib/usdUfe/ufe/Utils.cpp
seando-adsk: Ah yes thanks for catching that. I've added the zero padding and new unit test to cover this.

lib/usdUfe/python/wrapUtils.cpp
```diff
@@ -104,11 +101,12 @@ void wrapUtils()
     def("usdPathToUfePathSegment",
         _usdPathToUfePathSegment,
         (arg("usdPath"), arg("instanceIndex") = PXR_NS::UsdImagingDelegate::ALL_INSTANCES));
-    def("uniqueChildName", _uniqueChildName);
+    def("uniqueName", _uniqueName);
```
seando-adsk, line 104: Added missing python wrapper for this function.

lib/usdUfe/python/wrapUtils.cpp
```diff
@@ -104,11 +101,12 @@ void wrapUtils()
...
+    def("uniqueChildName", UsdUfe::uniqueChildName);
```
seando-adsk, line 105: For some of these we don't need the extra _ helper defined here. Can just call the UsdUfe function directly.

test/lib/ufe/testPythonWrappers.py
```diff
@@ -82,28 +83,71 @@ def testWrappers(self):
...
+        # by testPointInstances.
+
+        # isEditTargetLayerModifiable wrapper is tested by testBlockedLayerEdit.
+
+        # Test the getTime wrapper.
+        cmds.currentTime(10)
+        t = mayaUsd.ufe.getTime(stagePath)
+        self.assertEqual(t, Usd.TimeCode(10))
+
+        # isAttributeEditAllowed wrapper is tested by testAttribute.
```
seando-adsk, line 134: Added missing test for python wrapper.

test/lib/ufe/testPythonWrappers.py
```diff
@@ -82,28 +83,71 @@ def testWrappers(self):
...
+        self.assertIsNotNone(capsulePrim2)
+        self.assertEqual(capsulePrim, capsulePrim2)
+
+        # Test the maya-usd getNodeTypeFromRawItem() wrapper.
+        nodeType = mayaUsd.ufe.getNodeTypeFromRawItem(rawItem)
+        self.assertIsNotNone(nodeType)
+
+        # Test the maya-usd getNodeNameFromRawItem() wrapper.
+        nodeName = mayaUsd.ufe.getNodeNameFromRawItem(rawItem)
+        self.assertIsNotNone(nodeName)
```
seando-adsk, line 150: This is code from before - I just removed the if ufe >=2

test/lib/ufe/testPythonWrappers.py
```diff
@@ -82,28 +83,71 @@ def testWrappers(self):
...
+
+        # Test uniqueChildName wrapper.
+        newName = mayaUsd.ufe.uniqueChildName(capsulePrim, 'Cone1')
+        self.assertEqual(newName, 'Cone1')
+        mayaUsdStage.DefinePrim("/Capsule1/Cone1", "Cone")
+        newName = mayaUsd.ufe.uniqueChildName(capsulePrim, 'Cone1')
+        self.assertEqual(newName, 'Cone2')
+        mayaUsdStage.DefinePrim("/Capsule1/Sphere001", "Sphere")
+        newName = mayaUsd.ufe.uniqueChildName(capsulePrim, 'Sphere001')
+        self.assertEqual(newName, 'Sphere002')
```
seando-adsk, line 122: New tests to cover uniqueName/uniqueChildName, especially the missing numerical padding case you caught with my original code.

pierrebai-adsk (dismissed):

pierrebai-adsk (approved):

seando-adsk:
@pierrebai-adsk Sorry had some trouble with Linux/Mac builds and the unit test. Had to make some extra fixes.

pierrebai-adsk (approved):

PR #3315, opened 2023-09-07
EMSUSD-80 avoid creating an extra transform when in a prototype

pierrebai-adsk:
- When in a prototype, do not create an extra transform node.
- Adjust the related import tests, which were already mentioning those extra nodes.

### DIFF
--- a/test/lib/usd/translators/testUsdImportInstances.py
+++ b/test/lib/usd/translators/testUsdImportInstances.py
@@ -43,8 +43,11 @@ class testUsdImportInstances(unittest.TestCase):
 
     def sortedPathsTo(self, node):
         """Get a sorted list of paths going to the specified node"""
-        sel = om.MGlobal.getSelectionListByName(node)
-        self.assertEqual(sel.length(), 1)
+        try:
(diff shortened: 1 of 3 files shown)


vlasovi (approved):
Looks good

test/lib/usd/translators/testUsdImportInstances.py
```diff
@@ -43,8 +43,11 @@ def setUp(self):
 
     def sortedPathsTo(self, node):
         """Get a sorted list of paths going to the specified node"""
-        sel = om.MGlobal.getSelectionListByName(node)
-        self.assertEqual(sel.length(), 1)
+        try:
+            sel = om.MGlobal.getSelectionListByName(node)
+        except Exception:
+            self.assertTrue(False, 'Could not find node %s' % node)
+        self.assertEqual(sel.length(), 1, 'Could not find node %s' % node)
```
pierrebai-adsk, line 50: The goal of this change is to get a better error message on test failure.

pierrebai-adsk:
Only failure is one PF that lost connection after all tests ran and succeeded, so I'm marking this as ready for merge.

PR #3318, opened 2023-09-07
Add versioning to tests that use UsdSkelPackedJointAnimation 

dj-mcg:
UsdSkelPackedJointAnimation is deprecated and can safely be removed from tests

### DIFF
--- a/test/lib/ufe/testUIIcons.py
+++ b/test/lib/ufe/testUIIcons.py
@@ -115,7 +115,6 @@ class UIIconsTestCase(unittest.TestCase):
             ('NurbsCurves',             'out_USD_UsdGeomCurves.png'),
             ('NurbsPatch',              'out_USD_NurbsPatch.png'),
             ('OpenVDBAsset',            'out_USD_UsdGeomXformable.png'),
-            ('PackedJointAnimation',    'out_USD_SkelAnimation.png'),
             ('PluginLight',             'out_USD_PluginLight.png'),
             ('PluginLightFilter',       'out_USD_LightFilter.png'),
             ('PointInstancer',          'out_USD_PointInstancer.png'),
--- a/test/lib/ufe/testUIIcons.py
+++ b/test/lib/ufe/testUIIcons.py
@@ -146,6 +146,10 @@ class UIIconsTestCase(unittest.TestCase):
                 ('PhysicsScene',            'out_USD_UsdTyped.png'),
                 ('PhysicsSphericalJoint',   'out_USD_UsdTyped.png')
             ])
+        if usdVer <= (0, 23, 8):
+            primTypes.extend([
+                ('PackedJointAnimation',    'out_USD_SkelAnimation.png')
+            ])
 
         # Special case for node types which are in an AL schema.
         # They aren't available when compiling without the AL plugin.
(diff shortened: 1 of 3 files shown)


seando-adsk:
@dj-mcg We are still building MayaUsd for Maya 2022 & 2023 using USD v21.11 and Maya 2024 using USD v22.11. So don't we need to keep this for those versions and have usd version checks?

dj-mcg:
Hey @seando-adsk - according to the USD folks, this prim type has been deprecated for a long time despite the code still being present. Another sort of unexpected thing is that these tests pass for us locally even after using latest Usd, which no longer defines this type, so we were trying to figure out what these tests were actually intended to exercise.

If this prim type is still important to include in these tests, I guess we'll have to version them. I'm wondering how we would do that - specifically with regards to the usda input file. Would we need 2 different files that we'd use depending on the PXR_VERSION?

Thanks
Dan

dj-mcg:
Ah - Spiff clarified that this was retired in 2018 - it was actually renamed to SkelAnimation. The old code was allowed to linger deprecated by inheriting from the SkelAnimation. I'd be surprised if this was being used anywhere outside of these tests, which appear just to iterate through a list of prim types found in Usd perhaps?

test/lib/ufe/testUIIcons.py
```diff
@@ -115,7 +115,6 @@ def testUIIcons(self):
             ('NurbsCurves',             'out_USD_UsdGeomCurves.png'),
             ('NurbsPatch',              'out_USD_NurbsPatch.png'),
             ('OpenVDBAsset',            'out_USD_UsdGeomXformable.png'),
-            ('PackedJointAnimation',    'out_USD_SkelAnimation.png'),
```
seando-adsk, line 118: @dj-mcg What do we do about this prim type still appearing in the "Add New Prim" menu? This UI test is ensuring that each prim gets the proper icon from Ufe.
PackedJointAnimation

dj-mcg: This test was still passing for me locally even with the prim removed from USD. I didn't dig too deeply into why. I added versioning with this last commit to make sure everything is above board - please review and let me know if this addresses concerns, thanks!

seando-adsk: Sorry just to confirm. With USD v23.11 there will no longer be a PackedJointAnimation prim type? So that item won't appear in our "Add new prim" menu when using 23.11.

dj-mcg: That is correct. What I find a little surprising is that both tests pass as is (ie. still referencing the PackedJointAnimation prim) even when built/run against 23.11, so they are either both tolerant of missing prim types or not catching the errors if they are supposed to

seando-adsk: I still see the PackedJointAnimation in the USD dev branch https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdSkel/schema.usda#L144 - it has only been deprecated, but not removed yet.

dj-mcg: That will be gone with the next push to USD dev, likely at the end of this week

seando-adsk (approved):

PR #3320, opened 2023-09-08
EMSUSD-277 fix lost muted anonymous layers

pierrebai-adsk:
- Add a global functionality to hold onto muted layers.
- This is to work around the fact that OpenUSD will not keep muted layers in memory.
- Make sure to clean up the list of muted layers when a new Maya scene is created.
- Use the muted layer holder in the MuteLayer command.
- The base class of the command was trying to hold onto layer, but when the undo stack is cleared, so would be the anonymous layers.
- In particular, when unloading a Maya Reference, Maya always clears the undo stack.
- Add a test that check if muted anonymous layers are lost when editing a Maya Reference.
- The test fails without my fix, works with it.
- Improved test helper functions to enable creating anonymous sub-layers.

### DIFF
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -618,9 +618,9 @@ public:
             restoreSelection();
         }
 
-        // we perfer not holding to pointers needlessly, but we need to hold on to the layer if we
-        // mute it otherwise usd will let go of it and its modifications, and any dirty children
-        // will also be lost
+        // We prefer not holding to pointers needlessly, but we need to hold on
+        // to the muted layer. OpenUSD let go of muted layers, so anonymous
+        // layers and any dirty children would be lost if not explicitly held on.
         addMutedLayer(layer);
         return true;
     }
@@ -640,7 +640,7 @@ public:
             stage->MuteLayer(layer->GetIdentifier());
         }
 
-        // we can release the pointer
+        // We can release the now unmuted layers.
         removeMutedLayer(layer);
         return true;
     }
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -620,7 +621,7 @@ class MuteLayer : public BaseCmd
         // we perfer not holding to pointers needlessly, but we need to hold on to the layer if we
         // mute it otherwise usd will let go of it and its modifications, and any dirty children
         // will also be lost
```
samuelliu-adsk, line 623: I dont think we need this comment anymore

lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -638,8 +639,9 @@ class MuteLayer : public BaseCmd
             saveSelection();
             stage->MuteLayer(layer->GetIdentifier());
         }
+
         // we can release the pointer
-        _mutedLayer = nullptr;
+        removeMutedLayer(layer);
```
samuelliu-adsk, line 644: Question here: In AddMutedLayer, we recursively added all dirty sublayers to the mutedLayer list, so when remvoeMutedLayer should we do the same? For now it just removes one layer.

pierrebai-adsk: You're right, I will look into it.

lib/mayaUsd/commands/layerEditorCommand.cpp
pierrebai-adsk: Actually, the comment says exactly why we are calling addMutedLayer, so I'll keep the comment.

pierrebai-adsk:
Only failure is a crash on Linux in an interactive test about display layers, in malloc. Seems entirely unrelated.

samuelliu-adsk (approved):

PR #3324, opened 2023-09-12
EMSUSD-498 support ignoring variants when merging to USD

pierrebai-adsk:
- Add a "-ignoreVariants" flag to the mayaUsdMergeToUsd command.
- Add a "ignoreVariants" user argument to the UsdMayaPrimUpdaterArgs class.
- Use the ignoreVariants from UsdMayaPrimUpdaterArgs to set the ignoreVariants flag in MergePrimsOptions.
- Correctly strip the variant selection from the SdfPath in a few place where it was missing in the merge code.
- Add unit tests to verify that merging inside and outside variants work.


vlasovi (approved):
Looks good

PR #3325, opened 2023-09-13
HYDRA-496: Fix for crash on new scene with a USD Stage inside the Maya scene

roopavr-adsk:
Maya crashes when the MayaUsdProxyShapeSceneIndex tries to access MayaUsdProxyShape which is no longer valid.  This event happens for example in the case seen here in the reivew code to set time on the USD Stage. In this situation a Maya callback registered for "timechanged" event was also triggered on "New Scene" event but the USD and Hydra resources were deleted by then on which the callback tried to set the time and hence the crash.

The fix involves adding new API to MayaUsdProxyShapeStageExtraData which is a container of sorts for MayaUsdProxyShapeBase. The API allows to check if the proxy shape requested is valid.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeStageExtraData.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeStageExtraData.cpp
@@ -94,7 +94,7 @@ MStatus MayaUsdProxyShapeStageExtraData::finalize()
     return status;
 }
 
-bool MayaUsdProxyShapeStageExtraData::containsProxyShapeData(MayaUsdProxyShapeBase& proxyShape)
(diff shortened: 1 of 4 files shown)


lilike-adsk (approved):

seando-adsk (approved):

seando-adsk:
@roopavr-adsk You need to have a successful preflight run before marking as ready-for-merge. In order to run the preflight you need to assign the ecp-maya-devops-adsk user

roopavr-adsk:
FYI @lanierd-adsk @ppt-adsk @deb

PR #3338, opened 2023-09-20
EMSUSD-550 - Control the Types of Objects to Export

samuelliu-adsk:


test/lib/usd/translators/testUsdExportTypes.py
```diff
@@ -0,0 +1,123 @@
...
+        for m in meshPrims:
+            meshPrim = stage.GetPrimAtPath(m)
+            self.assertTrue(meshPrim)
+    
+    def testExportTypes(self):
+        self._ValidateMesh()
+        self._ValidateLight()
+        self._ValidateCamera()
+        self._ValidateMeshCamera()
+        self._ValidateMeshLight()
```
samuelliu-adsk, line 120: No need to test the case of export all three, that is already covered by other unit tests

lib/mayaUsd/fileio/writeJobContext.cpp
```diff
@@ -385,6 +385,39 @@ bool UsdMayaWriteJobContext::_NeedToTraverse(const MDagPath& curDag) const
         // If we're not going to create a prim at curDag, then we do not need to
         // traverse.
         return false;
+    } else {
```
pierrebai-adsk, line 388: Given the if returns, I'd not put the extra check in an else to avoid nesting. Other previous checks are not nested inside elses.

lib/mayaUsd/fileio/writeJobContext.cpp
```diff
@@ -385,6 +385,39 @@ bool UsdMayaWriteJobContext::_NeedToTraverse(const MDagPath& curDag) const
...
+        MDagPath shapeDagPath = curDag;
+        if (mArgs.mergeTransformAndShape) {
+            // if we're merging transforms, then we need to look at the shape.
+            shapeDagPath.extendToShape();
+        }
+
+        MStatus                 status;
+        MObject                 obj = shapeDagPath.node();
+        const MFnDependencyNode depFn(obj, &status);
+        if (!status) {
```
pierrebai-adsk, line 403: IDK if failure to cerate a MFnDependencyNode should be treated as failure. A node not being a MFnDependencyNode is not an error, it just that the node does not support the API.

It should just skip the type check.

lib/mayaUsd/fileio/writeJobContext.cpp
```diff
@@ -385,6 +385,39 @@ bool UsdMayaWriteJobContext::_NeedToTraverse(const MDagPath& curDag) const
...
+            return false;
+        }
+
+        const std::string mayaTypeName(depFn.typeName().asChar());
+
+        if (mArgs.excludeExportTypes.count(TfToken("Meshes")) != 0) {
```
pierrebai-adsk, line 409: Were  the name of types part of the design? I ask because I find that using "Meshes" instead of "mesh" for a type name feels wrong. The type should not be plural.

In any cases, the string comparison should be case-insensitive.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -749,8 +806,11 @@ global proc int mayaUsdTranslatorExport (string $parent,
         string $expandedSections[];
         string $collapsedSections[];
         string $sectionNames[] = parseActionSectionNames($sections, $expandedSections, $collapsedSections);
+        // collect all exclude types before append to option list
+        string $excludeExportTypes = "";
             
         if (stringArrayContains("geometry", $sectionNames)) {
+            $excludeExportTypes = mayaUsdTranslatorExport_SetSuboptionsFromCheckBox($excludeExportTypes, "", "Meshes", "exportMeshesCheckBox");
```
pierrebai-adsk, line 813: The flag is passed explicitly here, so changing from "Meshes" to "mesh" seem feasible without modifying UI shown to teh user.

pierrebai-adsk (changes requested):
Looks good aside from the few comments I left.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
samuelliu-adsk: I don't know if that will confuse the user.. Like they see "Meshes" in the UI but when use it in command it changes to "mesh". They won't know it unless they read extra documents to get that information...

samuelliu-adsk: But yea its definitely feasible

lib/mayaUsd/fileio/writeJobContext.cpp
```diff
@@ -406,15 +404,18 @@ bool UsdMayaWriteJobContext::_NeedToTraverse(const MDagPath& curDag) const
 
         const std::string mayaTypeName(depFn.typeName().asChar());
 
-        if (mArgs.excludeExportTypes.count(TfToken("Meshes")) != 0) {
+        if ((mArgs.excludeExportTypes.count(TfToken("Meshes")) != 0)
+            || (mArgs.excludeExportTypes.count(TfToken("meshes")) != 0)) {
```
pierrebai-adsk, line 408: I think it would be cleaner and more general to lowercase both strings and compare those.

pierrebai-adsk (approved):

pierrebai-adsk (dismissed):

test/lib/usd/translators/testUsdExportTypes.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env mayapy
+#
+# Copyright 2016 Pixar
```
seando-adsk, line 3: New file by you should be Copyright by Autodesk and have current year.

seando-adsk (changes requested):
Please fix the copyright in your new test file. Once you do, just mark as ready for merge (no re-run of the preflight needed).

seando-adsk (approved):

PR #3345, opened 2023-09-25
EMSUSD-490 Fixes outerliner not refreshing when performing a UsdUndoVisibleCommand

AramAzhari-adsk:
When undoing a visibility command, UsdNotice::ObjectsChanged does not produce a path for Info Change but still generates a path for Re-sync Change which has the Visibility token. The special case is added to refresh when in the latter path.

### DIFF
--- a/lib/usdUfe/ufe/StagesSubject.cpp
+++ b/lib/usdUfe/ufe/StagesSubject.cpp
@@ -162,6 +162,12 @@ void sendAttributeChanged(
         }
     } break;
     case AttributeChangeType::kRemoved: {
+        // Special case when Undoing a visibility change, the notice.GetChangedInfoOnlyPaths()
+        // does not contain the change, hence handling visibility notification in re-synch path.
+        if (changedToken == UsdGeomTokens->visibility) {
+            Ufe::VisibilityChanged vis(ufePath);
+            notifyWithoutExceptions<Ufe::Object3d>(vis);
+        }
         notifyWithoutExceptions<Ufe::Attributes>(
             Ufe::AttributeRemoved(ufePath, changedToken.GetString()));
     } break;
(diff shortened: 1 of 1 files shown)


AramAzhari-adsk:
It seems that when:

1- Making an item invisible the following notifications are sent:
    ValueAdded
    ValueChanged

2- When setting the visibility back to "Inherited" / "Visible", only the following is sent which makes sense as the attribute already exists:
    ValueChanged

3- When Undoing after manually making an item invisible, only the following notification is sent:
    ValueRemoved

This is checked here: 
https://github.com/Autodesk/maya-usd/blob/ce7fc176a7e907d16bf94aa3388e669846fd0d49/lib/usdUfe/ufe/StagesSubject.cpp#L142-L167

The bug is for #3.

lib/usdUfe/ufe/StagesSubject.cpp
```diff
@@ -395,7 +395,14 @@ void StagesSubject::stageChanged(
                     notifyWithoutExceptions<Ufe::Transform3d>(ufePath);
                 }
             }
+            // Special case when Undoing a visibility change, the notice.GetChangedInfoOnlyPaths()
+            // does not contain the change, hence handling visibility notification in re-synch path.
+            if (changedPath.GetNameToken() == UsdGeomTokens->visibility) {
+                Ufe::VisibilityChanged vis(ufePath);
+                notifyWithoutExceptions<Ufe::Object3d>(vis);
+            }
```
seando-adsk, line 403: I'm not sure the order matters, but I wonder if sending this extra visibilityChanged should come after the processAttributeChanges() just below. This would then match the order in the ChangedInfoOnlyPaths() loop.

AramAzhari-adsk: @seando-adsk  I moved it as per [this finding](https://github.com/Autodesk/maya-usd/pull/3345#issuecomment-1734439926) since having it in at line 398 caused duplicate notifications to be sent which failed the visibility test.

seando-adsk (approved):

AramAzhari-adsk:
One test on one platform did not pass which is a known random occurrence:
 testVP2RenderDelegateUsdSkel

PR #3350, opened 2023-09-26
EMSUSD-623 edit as Maya multiple variants

pierrebai-adsk:
This change adds support to edit-as-Maya a prim that is under a variant, with each variation having a different Maya reference. That is, a single prim can now be edited-as-Maya multiple times in parallel. This required some changes on how those edited prim are tracked.

Edit-as-Maya authors data in the session layer related to the edited prim. Previously, when edited-as-Maya were orphaned or ceased to be, the order in which they were processed was random. This caused two prims with the same name but in different variants to potentially clobber each other's data. Now we process new orphans first, then new non-orphans second. This allows new non-orphans to write their data without it being wiped out by an orphan.

Additional variant-related helper functions:
- Add the getEditTargetForVariants function helper function to create the USD edit target targeting all the variants that affect a given prim.
- Ensure getEditTargetForVariants works in more complex scenarios spanning payloads and references.
- Add the getVariantPath helper function to create a USD path containing a variant selection.

Additional pull information helper functions:
- Warn when we fail to remove pull information.

Change to the orphan manager:
- Support having multiple entries for a given UFE path.
- To differentiate between entries, the corresponding root Maya DAG path must be given in all functions to figure out which version of the prim we are dealing with.
- Fortunately, all callers always have on hand the DAG path of the root Maya node.
- Change all implementation functions to iterate over these variations.
- Change the orphaned nodes serialization to support the extra data.
- Make the serialization format backward compatible by special-casing the usual case of having a single variation at a given edited UFE path.
- Add a log to tell the user when an edited prim is orphaned.
- Add a has helper function in the orphaned node manager to check if a prim is already edited.
- Use it to avoid editing a prim variant twice.
- Fix how the orphaned manager handles notifications to determine orphans.
- In particular, handle deactivated prims correctly: when switching variants, the order of the consequences is not deterministic, so the handling must not assume that edited prims are active.
- Make the orphan manager handle orphans in two passes: first the new orphans, second the ones that are no longer orphans.
- This way, non-orphans can write their metadata and it won't be deleted by orphans of the same name.

Changes to the prim updater manager:
- Ignore edited prim that are orphaned when determining if a prim has edited descendants.
- Adapt to changes to the orphaned manager by passing the Maya DAG path to the root of the edited nodes.
- When merging to USD, when removing the pull information, use the helper function.

Changes to Maya reference translator (importer from USD to Maya):
- Make many functions be purely in the implementation instead of declared private in the class.
- This allows better hiding of the implementation.
- This was also necessary to make the function callable from internal implementation functions.
- Renamed LoadMayaReference to CreateMayaReference to better reflect what the function does.
- Split the long update function into multiple function for clarity.
- When trying to reuse Maya reference nodes, verify that the referenced file is the one that was expected.
- This allows two prim with the same USD path but different references to work alongside each other.
- This happens when one prim with two variants each contain a Maya reference.
- Add more comments in the code to explain what is going on.
- Add a log to tell the user when a reference does not have the expected file path.


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -198,22 +214,9 @@ bool writeAllPullInformation(const Ufe::Path& ufePulledPath, const MDagPath& edi
...
 {
-    UsdPrim     pulledPrim = MayaUsd::ufe::ufePathToPrim(ufePulledPath);
-    UsdStagePtr stage = pulledPrim.GetStage();
-    if (!stage)
-        return;
-
     MayaUsd::ProgressBarScope progressBar(1);
-    removePulledPrimMetadata(stage, pulledPrim);
+    removePulledPrimMetadata(ufePulledPath);
     progressBar.advance();
```
pierrebai-adsk, line 219: Note: all removed code has moved into removePulledPrimMetadata()

lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -141,8 +144,10 @@ void renamePullInformation(
         }
     }
 
-    const MDagPath& mayaPath = trieNode->data().editedAsMayaRoot;
-    TF_VERIFY(writePullInformation(pulledPath, mayaPath));
+    for (const PullVariantInfo& info : trieNode->data()) {
+        const MDagPath& mayaPath = info.editedAsMayaRoot;
+        TF_VERIFY(writePullInformation(pulledPath, mayaPath));
+    }
```
pierrebai-adsk, line 150: There are a lot of changes like this in the orphan manager where what use to be a single value is now a vector because we support a single prim path being under different variants.

lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -310,7 +354,8 @@ void OrphanedNodesManager::handleOp(const Ufe::SceneCompositeNotification::Op& o
         // point.  It may be an internal node, without data.
         auto ancestorNode = _pulledPrims.node(op.path);
         TF_VERIFY(ancestorNode);
-        recursiveSwitch(ancestorNode, op.path);
+        recursiveSwitch(ancestorNode, op.path, true);
+        recursiveSwitch(ancestorNode, op.path, false);
```
pierrebai-adsk, line 358: We need to process newly orphaned nodes first because that deletes information in various places while the ones that are no longer orphans add information.  If we let the change be done in any order, we would sometimes lose information that got deleted by orphans.

lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -345,63 +390,54 @@ void OrphanedNodesManager::handleOp(const Ufe::SceneCompositeNotification::Op& o
             return;
         }
 
-        auto parentHier = Ufe::Hierarchy::hierarchy(parentItem);
-        if (!parentHier->hasChildren()) {
```
pierrebai-adsk, line 349: This line was wrong because it ignored deactivated prims. Prims get deactivated when edited-as-Maya, so we would not process notifications correctly in all cases.

lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -626,8 +699,10 @@ OrphanedNodesManager::variantSetDescriptors(const Ufe::Path& p)
         auto ancestor = Ufe::Hierarchy::createItem(path);
         auto usdAncestor = std::static_pointer_cast<UsdUfe::UsdSceneItem>(ancestor);
         auto variantSets = usdAncestor->prim().GetVariantSets();
+        auto setNames = variantSets.GetNames();
+        std::sort(setNames.begin(), setNames.end());
```
pierrebai-adsk, line 703: Sort the variant set names to get a guaranteed consistent order for comparison purpose.

test/lib/mayaUsd/fileio/testCacheToUsd.py
```diff
@@ -564,10 +564,8 @@ def runTestMayaRefPrimTransform(self, createMayaRefPrimFn, checkCacheParentFn):
         checkCacheParentFn(self, cacheParentChildren, variantSet, cacheVariantName)
 
         # Maya reference prim should now have the updated transformation.
-        editTarget = self.stage.GetEditTarget()
-        if variantSet:
-            variantSet.SetVariantSelection('Rig')
-            editTarget = variantSet.GetVariantEditTarget(editTarget.GetLayer())
+        if variantSetName:
+            cacheParent.GetVariantSet(variantSetName).SetVariantSelection('Rig')
```
pierrebai-adsk, line 568: The code was doing unnecessary things. (I think this block was copy/paste from earlier in the same test.) I had to go through this unit test in early prototypes and fixed it back then, so I left the change in, even though it is not strictly necessary for this PR.

AramAzhari-adsk (approved):

test/lib/mayaUsd/fileio/testSwitchMayaReference.py
```diff
@@ -0,0 +1,167 @@
...
+        return 'testSwitchMayaRefCache.usda'
+
+    def removeCacheFile(self):
+        '''
+        Remove the cache file if it exists. Ignore error if it does not exists.
+        '''
+        try:
+            os.remove(self.getCacheFileName())
+        except:
+            pass
```
seando-adsk, line 68: Can you fix this so there is no Chorus warning? I think you just need to do `except Exception:` instead. I also found 2-3 other places identical to this (old code) that would be great if you could clean those up as well.

AramAzhari-adsk (approved):

seando-adsk (approved):
Thanks for doing this.

PR #3353, opened 2023-09-28
EMSUSD-622 - As a user, I'd like to add a relative sublayer on a anonymous layer

vlasovi:


lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -48,6 +48,9 @@ std::string generateUniqueName()
     }
     return uniqueName;
 }
+
+static std::map<PXR_NS::SdfLayerHandle, std::set<ghc::filesystem::path>> sPostponedRelativePaths;
```
pierrebai-adsk, line 52: I know that it is not multi-thread safe anyway, but just for order-of-initilization  of C++ globals, it is a better practice to put the static variable inside a static function that returns it because the C++ language standard guarantee hat the static will be initialized on first call (which avoid the order of init of globals problem) and ensure that this init is thread-safe.

So something like:
```
using RelativePaths = std::map<PXR_NS::SdfLayerHandle, std::set<ghc::filesystem::path>>;
static RelativePaths& getRelativePaths()
{
    static RelativePaths sPaths;
    return sPaths;
}
```

pierrebai-adsk (changes requested):
Small change requested to make it safer vs C++ link order.

Otherwise, looks good to me.

pierrebai-adsk (approved):

PR #3361, opened 2023-10-03
EMSUSD-579 - Add Untextured mode selections to user pref

samuelliu-adsk:
Preview in user pref window:
image

### DIFF
--- a/plugin/adsk/scripts/mayaUsd_preferenceTab.mel
+++ b/plugin/adsk/scripts/mayaUsd_preferenceTab.mel
@@ -133,7 +133,7 @@ global proc mayaUsd_PrefInitOptionVars(int $forceFactoryDefault)
 			optionVar -intValue mayaUsd_MakePathRelativeToImageEditTargetLayer false;
 		}
 		if ($forceFactoryDefault || !`optionVar -exists mayaUsd_ShowDisplayColorTextureOff`) {
-			optionVar -intValue mayaUsd_ShowDisplayColorTextureOff 0;
+			optionVar -intValue mayaUsd_ShowDisplayColorTextureOff false;
 		}
 
 	} else {
@@ -152,7 +152,7 @@ global proc mayaUsd_PrefInitOptionVars(int $forceFactoryDefault)
 			-iv mayaUsd_MakePathRelativeToParentLayer false
 			-iv mayaUsd_MakePathRelativeToEditTargetLayer false
 			-iv mayaUsd_MakePathRelativeToImageEditTargetLayer false
-			-iv mayaUsd_ShowDisplayColorTextureOff 0
+			-iv mayaUsd_ShowDisplayColorTextureOff false
 		;
 	}
 }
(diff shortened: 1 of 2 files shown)


plugin/adsk/scripts/mayaUsd_preferenceTab.mel
```diff
@@ -406,6 +429,11 @@ global proc mayaUsd_PrefHoldState(string $mode)
 				`optionVar -q mayaUsd_MakePathRelativeToImageEditTargetLayer`
 				usdPrefFileDependenciesRelativeToEditTargetLayer;
 		}
+		if (`checkBoxGrp -exists usdPrefUntexturedMode`) {
+			checkBoxGrp -e -v1
+				`optionVar -q mayaUsd_ShowDisplayColorTextureOff`
+				usdPrefUntexturedMode;
+		}
```
seando-adsk, line 436: You should also add your new pref to the prefOptionVars list above.

seando-adsk (changes requested):
You also need to add your new pref case to:
- mayaUsd_PrefInitOptionVars() - both the 2022 (and earlier case) and else (using new -init flag).
- mayaUsd_PrefUpdate()

seando-adsk (changes requested):
You still need to add your new option var to mayaUsd_PrefUpdate()

plugin/adsk/scripts/mayaUsd_preferenceTab.mel
```diff
@@ -434,8 +439,8 @@ global proc mayaUsd_PrefHoldState(string $mode)
 				`optionVar -q mayaUsd_MakePathRelativeToImageEditTargetLayer`
 				usdPrefFileDependenciesRelativeToEditTargetLayer;
 		}
-		if (`checkBoxGrp -exists usdPrefUntexturedMode`) {
-			checkBoxGrp -e -v1
+		if (`radioButtonGrp -exists usdPrefUntexturedMode`) {
+			radioButtonGrp -e -v1
```
seando-adsk, line 443: radioButtonGrp doesn't have -v1 flag, you need to use -sl flag (like above) and convert the boolean option: false (0), true (1) into the select flag which 1-based.

plugin/adsk/scripts/mayaUsd_preferenceTab.mel
```diff
@@ -333,6 +333,11 @@ global proc mayaUsd_PrefUpdate()
 			usdPrefConfirmExistFileSave;
 	}
 
+	if (`optionVar -exists "mayaUsd_ShowDisplayColorTextureOff"`) {
+		int $v = `optionVar -q "mayaUsd_ShowDisplayColorTextureOff"`;
+		radioButtonGrp -e -select $v usdPrefUntexturedMode;
```
seando-adsk, line 338: See my comment below about boolean optionVar vs 1-based select flag.

seando-adsk (changes requested):
Sorry I didn't catch the radioButton vs checkBox before. Some more changes are needed.

plugin/adsk/scripts/mayaUsd_preferenceTab.mel
```diff
@@ -139,8 +152,11 @@ global proc mayaUsd_PrefInitOptionVars(int $forceFactoryDefault)
 			-iv mayaUsd_MakePathRelativeToParentLayer false
 			-iv mayaUsd_MakePathRelativeToEditTargetLayer false
 			-iv mayaUsd_MakePathRelativeToImageEditTargetLayer false
+			-iv mayaUsd_ShowDisplayColorTextureOff false
 		;
 	}
+	// Need to refresh the viewport for display mayaUsd_ShowDisplayColorTextureOff
+	ogs -reset;
```
samuelliu-adsk, line 159: I added another viewport refresh here, because if we reset to factoryDefault we need to refresh to show the change

seando-adsk (approved):

PR #3364, opened 2023-10-03
EMSUSD-489 fix reparenting crash

pierrebai-adsk:
There were multiple problems that triggered each others to cause a crash when using relative sub-layers. The problems were:
- The relative sub-layer was not found.
- This cause the code that applies the reparenting to all layers to do no work.
- This would cause the prim to vanish.
- This would cause the transform command to crash because it did not check that the scene item was valid.

The corrections done are:
- Fix the getAllSubLayers helper function to use FindRelativeToLayer to find sub-layers, in case they are relative to their parent. Otherwise it would fail to find sub-layers.
- Modify the applyToAllPrimSpecs, applyToAllLayersWithOpinions and applyToSomeLayersWithOpinions helper functions to return the number of layers that were affected.
- This allows the caller to detect if no action was taken at all, indicating that an expected action failed.
- Make the UsdSetMatrix4dUndoableCmd UFE command not crash when the scene item is not found.
- Make the insert-child UFE command detect if the move to the new parent did no work at all and report an error instead of marching on and deleting the old location.
- Otherwise, the prim would not be at the new location and would be deleted from its old location, so it owuld vanish.

Add a unit test that try to reparent a prim in a relative sub-layer. It failed without the fixes, now it works.

### DIFF
--- a/lib/mayaUsd/ufe/UsdTransform3dMatrixOp.cpp
+++ b/lib/mayaUsd/ufe/UsdTransform3dMatrixOp.cpp
@@ -119,6 +119,9 @@ public:
         // transform3d() returns a whole-object interface, which may include
         // other transform ops.
         auto t3d = Ufe::Transform3d::editTransform3d(sceneItem());
+        if (!TF_VERIFY(t3d)) {
+            return;
+        }
         t3d->setMatrix(_newM);
     }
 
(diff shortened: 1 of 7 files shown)


seando-adsk (approved):
Great work. Thanks for adding the error checking and unit test.

PR #3370, opened 2023-10-06
EMSUSD-644 - As a user, I'd like to add a relative payload/ref on a anonymous layer

vlasovi:
This also fixes [EMSUSD-634] - Error when attempting to add a Maya Reference.

### DIFF
--- a/lib/mayaUsd/utils/utilFileSystem.cpp
+++ b/lib/mayaUsd/utils/utilFileSystem.cpp
@@ -313,6 +313,54 @@ void UsdMayaUtilFileSystem::unmarkPathAsPostponedRelative(
     }
 }
 
+template <typename TypePolicy>
+void updatePathList(
+    SdfListProxy<TypePolicy>                list,
+    const PostponedRelativePaths::iterator& layerEntry,
+    const std::string&                      anchorDirStr)
+{
+    for (auto proxy : list) {
+        typename TypePolicy::value_type item = proxy;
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -313,6 +313,54 @@ void UsdMayaUtilFileSystem::unmarkPathAsPostponedRelative(
...
+        proxy = item;
+    }
+}
+
+void updatePostponedRelativePathsForPrim(
+    const SdfPrimSpecHandle&                primSpec,
+    const PostponedRelativePaths::iterator& layerEntry,
+    const std::string&                      anchorDirStr)
+{
+    for (const SdfPrimSpecHandle& child : primSpec->GetNameChildren()) {
```
pierrebai-adsk, line 343: Question: does this only return child prim spec that are in the same layer or can it return prim spec from other layers? What about references? Can it traverse references?

IDK, for example, what happens if the user adds a payload to a prim that only lives in a reference? Will it be found?

pierrebai-adsk: Mmm, we're only trying to update things that are inside the layer that just got saved. So I guess my question is irrelevant.

pierrebai-adsk (approved):
LGTM

I think my question is not relevant but I'll leave it there.

PR #3372, opened 2023-10-06
EMSUSD-254 Allow selecting the referenced prim

pierrebai-adsk:
To allow selecting the referenced prim when adding a new USD reference or payload, multiple sub-systems had to be modified:
- The usdImportDialog command, to accept new parameters
- Qt dialog layout.
- The Qt tree model and tree item implementing the USD hierarchy view.
- The functions managing the USD ref/payload file dialog.
- The Python code that ties everything together.

Modify the usdImportDialog command:
- Add a new flag to the usdImprotDialog command to control if variants are shown.
- Add a new flag to the usdImprotDialog command to control if the root prim is shown.
- Add a new flag to the usdImprotDialog command to set the window title.
- Add a new flag to the usdImprotDialog command to set the help label.
- Add a new flag to the usdImprotDialog command to set the help target.
- Allow setting the prim path with the edit flag.
- Use the new prim path flag in the Python code that builds the UI.
- Added helper function to manage the transfer of the prim path from the UI to the command.

Modify the Qt dialog to enable all those flags:
- Add a USDImportDialogOptions structure to hold all the various options.
- Pass that around to all the relevant sub-functions.
- Modify the dialog to hide variants, header message, etc when relevant.
- Change the add ref/payload dialog accept button label to be "Reference".
- Implement most of the referenced prim logic.
- Fix the import dialog to only use the prim path if not empty.

Modify the USD hierarchy tree model and item:
- Change TreeModel to handle not showing the root and not showing variants.
- Made the tree item add a 'default prim' icon next to the prim name when it is the default prim.
- Added the default prim image used by the tree item.
- Make the tree model factory preselect the top prim.
- Modify the factory to create the correct number of column depending on if variants are shown.
- Make the tree model factory use the first root prim if no default prim are present.
- Added a helper method on the tree model to retrieve the first item.

Modify the relative-file utility classes:
- Allow all the relative-UI classes to hook into the filename selection.
- For the add payload/ref one, update the referenced prim UI when the filename changes.
- Still missing pre-selecting the correct prim in the selection dialog and some corner cases.
- Save the selected prim path in an option var to communicate it to the calling code.

Modify the context menu:
- Use all of the above to create the add-ref or add-payload command with the selected prim path.


pierrebai-adsk:
Another big PR, I know...

lib/mayaUsd/resources/scripts/mayaUsdMayaReferenceUtils.py
```diff
@@ -136,27 +216,48 @@ def createUsdRefOrPayloadUI(showLoadPayload=False):
         for label in listEditedAsLabels:
             cmds.menuItem(label=label)
 
-    if showLoadPayload:
+    if uiForLoad:
```
pierrebai-adsk, line 219: I change showLoadPayload to uiForLoad since the UI elements are for teh case we load vs save (example of saving: creating the USD cache when merging a Maya ref)

lib/usd/ui/importDialog/TreeItem.h
```diff
@@ -39,12 +39,13 @@ class MAYAUSD_UI_PUBLIC TreeItem : public QStandardItem
...
-        kLoad,
-        kName,
-        kType,
-        kVariants
+        kColumnLoad,
+        kColumnName,
+        kColumnType,
+        kColumnVariants,
+        kColumnLast
     };
```
pierrebai-adsk, line 49: I change this because TreeItem and TreeModel had both an enumwith the same values which were magically assumed to be in the same order. Now there is a single enum, and that enum must be convertible to an int to pass to Qt.

vlasovi (approved):
Looks good to me

AramAzhari-adsk (approved):

AramAzhari-adsk:
Looks good to me. Great job!

PR #3380, opened 2023-10-12
EMSUSD-636 always use the same folder to load layers

pierrebai-adsk:
No matter what the parent layer is, always use the last folder where a layer was loaded. (Unless no layer was ever loaded by the user, of course.)


vlasovi (approved):
Looks good

PR #3383, opened 2023-10-12
EMSUSD-686 better UI when selecting the proxy shape path

pierrebai-adsk:
Re-use the same UI functions as when initially creating the stage. Also, start the dialog in the path where the current file already is.

### DIFF
--- a/plugin/adsk/scripts/AETemplateHelpers.py
+++ b/plugin/adsk/scripts/AETemplateHelpers.py
@@ -131,15 +131,28 @@ def ProxyShapeFilePathChanged(filePathAttr, newFilePath=None):
             okCaption = getMayaUsdString("kLoad")
             fileFilter = getUSDDialogFileFilters()
 
+            # Note: empty or missing attribute return None, but we want an empty string
+            #       in that case.
+            primPath = cmds.getAttr(stageName+'.primPath') or ''
+            excludedPrimPaths = cmds.getAttr(stageName+'.excludePrimPaths') or ''
+            loadPayloads = cmds.getAttr(stageName+'.loadPayloads') or 0
+
+            cmds.optionVar(stringValue=('stageFromFile_primPath', primPath))
+            cmds.optionVar(stringValue=('stageFromFile_excludePrimPath', excludedPrimPaths))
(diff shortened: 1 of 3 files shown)


plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -207,5 +229,4 @@ def ProxyShapeFilePathRefresh(filePathAttr):
                     debugMessage('  User confirmed reload action, calling UsdStage.Reload()')
                     proxyStage.Reload()
     except Exception as e:
-        debugMessage('ProxyShapeFilePathRefresh() - Error: %s' % str(e))
-        pass
+        print('ProxyShapeFilePathRefresh() - Error: %s' % str(e))
```
seando-adsk, line 232: Why is this a plain print now? The user cannot do anything about the error.

plugin/adsk/scripts/mayaUSDRegisterStrings.mel
```diff
@@ -120,7 +120,7 @@ global proc mayaUSDRegisterStrings()
     register("kTimeAnn", "Edits the current time value of a stage, which corresponds to the animation frame drawn in the viewport. By default, this value connects to Maya's global time node.");
     register("kTipYouCanChooseMultipleFiles", "<b>Tip:</b>  You can choose multiple files.");
     register("kUniversalSceneDescription", "Universal Scene Description");
-    register("kUsdFileOptions", "USD File Options");
+    register("kUsdFileOptions", "Stage");
```
seando-adsk, line 123: In these cases I also like to rename the kUsdFileOptions variable (make it more descriptive of the string) 
 since it forces us to look at every place this string is used and ensure we aren't changing a place we don't intend to.

seando-adsk (changes requested):
Overall looks good. Just two minor comments.

plugin/adsk/scripts/AETemplateHelpers.py
pierrebai-adsk: The alternative is that if the button fails for any reason, it does so silently. I can remove the print, but I treat differently random errors printed without user action vs error triggered by user action. This error is rare enough that I don't really mind either way, so I'll just revert it.

seando-adsk (approved):

PR #3385, opened 2023-10-13
Determine how to read UsdLux prims with an envvar

dj-mcg:
This eliminates the need for the build var and allows this to be set at runtime


seando-adsk:
@dj-mcg I saw your preflight failed on a few tests related to lights.

dj-mcg:
I see the issue, but I need to ask the Usd team a few questions about how to properly approach translators of types with codeless schemas, so it may be a day or two before I update this PR

dj-mcg:
> @dj-mcg I saw your preflight failed on a few tests related to lights.

The problematic types are defined in renderman. We're currently rethinking how we go about registering these readers as it would be great to get rid of a build flag requirement if possible. I'm thinking we might want to define these readers in the renderman-for-maya plugin but I'm still unsure whether there is precedent for introducing dependencies on mayaUsd in that plugin. Let me know if you have thoughts re: registering mayaUsd readers for externally defined types

dj-mcg:
> @dj-mcg I saw your preflight failed on a few tests related to lights.

Sorry for the delay, @seando-adsk - I removed the offending light types and moved them to an internal plugin

test/lib/usd/translators/testUsdImportRfMLight.py
```diff
@@ -105,19 +105,9 @@ def _ValidateMayaLight(self, lightTypeName):
         elif lightTypeName == 'SphereLight':
             self.assertEqual(depNodeFn.typeName(), 'PxrSphereLight')
             testNumber = 7
-        elif lightTypeName == 'AovLight':
```
pierrebai-adsk, line 108: So those renderman lights are no longer tested, as I understand it?

pierrebai-adsk (approved):

PR #3392, opened 2023-10-16
EMSUSD-645 - As a user, I'd like to add a relative file dependency on a anonymous layer

vlasovi:
Postponed processing of relative asset paths assigned to anonymous layers

### DIFF
--- a/lib/mayaUsd/utils/utilFileSystem.h
+++ b/lib/mayaUsd/utils/utilFileSystem.h
@@ -144,19 +144,19 @@ void unmarkPathAsPostponedRelative(
     const std::string&            contentPath);
 
 /*! \brief Performs handling of a USD asset path attribute that maybe relative to a layer.
-Whether the file path is required to be relative is defined by the primvar which name is supplied
-through 'primVarName' variable. If the relative file path is required and the layer is saved on disk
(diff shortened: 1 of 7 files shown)


lib/mayaUsd/utils/utilFileSystem.h
```diff
@@ -143,6 +143,21 @@ void unmarkPathAsPostponedRelative(
     const PXR_NS::SdfLayerHandle& layer,
     const std::string&            contentPath);
 
+/*! \brief Performs handling of a USD asset path attribute that maybe relative to a layer.
+Whether the file path is required to be relative is defined by the primvar which name is supplied
+through 'primVarName' variable. If the relative file path is required and the layer is saved on disk
```
pierrebai-adsk, line 148: Is it not an 'optionVar' name and not a 'primVar' name?

pierrebai-adsk (changes requested):
Ingeniously done, just a small requested change.

lib/mayaUsd/utils/utilFileSystem.h
vlasovi: Yeah, you are right. I'll fix it

pierrebai-adsk (approved):
LGTM, good job

PR #3393, opened 2023-10-16
Fix errors when a USD object has an !invert! xform attribute.

pierrebai-adsk:
The code to handle it was off by one.


pierrebai-adsk:
The only test failure is a single test on a single platform because a usdz file was left-over in the temp folder from a previous preflight.

samuelliu-adsk (approved):

PR #3394, opened 2023-10-16
EMSUSD-105 - Set a prim as defaultPrim in the stage

samuelliu-adsk:


plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -11,6 +12,46 @@ def debugMessage(msg):
     if DEBUG:
         print(msg)
 
+def GetAllRootPrimNamesNaturalOrder(proxyShape):
+    # Custom comparator
```
samuelliu-adsk, line 16: The method below is a customized comparator for sorting in natural order. For example, proxyShape10 will come after proxyShape 2.

plugin/adsk/scripts/AEmayaUsdProxyShapeBaseTemplate.mel
```diff
@@ -21,13 +21,38 @@ global proc string AEmayaUsdProxyShapeInfoChanged(string $input)
...
     string $defaultPrim = `python("import AETemplateHelpers; AETemplateHelpers.GetDefaultPrimName('" + $fullNodeName + "')")`;
+    string $allRootPrims[] = `python("import AETemplateHelpers; AETemplateHelpers.GetAllRootPrimNamesNaturalOrder('" + $fullNodeName + "')")`;
 
     textFieldGrp -edit -text $rootLayer mayaUsdProxyShapeRootLayer;
-    textFieldGrp -edit -text $defaultPrim mayaUsdProxyShapeDefaultPrim;
+
+    optionMenuGrp -edit -deleteAllItems -changeCommand ("mayaUsd_DefaultPrimChanged(\"" + $fullNodeName + "\"\)") mayaUsdProxyShapeDefaultPrim;
+    string $menuName = `optionMenuGrp -q -fullPathName mayaUsdProxyShapeDefaultPrim`;
+    int $index = 2; // starts at 2 because the first one is always empty
+        menuItem -parent ($menuName + "|OptionMenu") -l "" -ann "Clear default prim";
```
samuelliu-adsk, line 31: The first one is always empty for clearing default prim

plugin/adsk/scripts/AEmayaUsdProxyShapeBaseTemplate.mel
```diff
@@ -21,13 +21,38 @@ global proc string AEmayaUsdProxyShapeInfoChanged(string $input)
...
+                $found = 1;
+            }
+            if($found == 0){
+                $index ++;
+            }
+        }
+    if ($defaultPrim == ""){
+        $index = 1;
+    }
+    optionMenuGrp -e -select $index mayaUsdProxyShapeDefaultPrim;
```
samuelliu-adsk, line 45: Highlight the current default prim

lib/usdUfe/ufe/UsdContextOps.cpp
```diff
@@ -306,6 +310,14 @@ Ufe::ContextOps::Items UsdContextOps::getItems(const Ufe::ContextOps::ItemPath&
                     items.emplace_back(kUSDToggleVisibilityItem, label);
                 }
             }
+
+            // Set as Default Prim:
+            // If the prim is a root prim, add set default prim
+            if (prim().GetPath().IsRootPrimPath()) {
+                items.emplace_back(kUSDSetAsDefaultPrim, kUSDSetAsDefaultPrim);
+                items.emplace_back(kUSDClearDefaultPrim, kUSDClearDefaultPrim);
```
pierrebai-adsk, line 318: The spec in EMSUSD-105 says that clear should only be provided if the prim is the current default one.

lib/usdUfe/ufe/UsdUndoClearDefaultPrimCommand.cpp
```diff
@@ -0,0 +1,60 @@
...
+}
+
+UsdUndoClearDefaultPrimCommand::~UsdUndoClearDefaultPrimCommand() { }
+
+void UsdUndoClearDefaultPrimCommand::execute()
+{
+    UsdUndoBlock undoBlock(&_undoableItem);
+
+    // Check if the layer selected is the root layer.
+    if (_stage->GetRootLayer() != _stage->GetEditTarget().GetLayer()) {
```
pierrebai-adsk, line 45: For now this is the only case, so it's fine, but I'd rather have a utility function that checks if metadata is allowed to be edited. So if we need to check it in other case, the code is the same and if we change the restriction rules in the future, there would be only one place to change it.

lib/usdUfe/ufe/UsdUndoSetDefaultPrimCommand.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+UsdUndoSetDefaultPrimCommand::~UsdUndoSetDefaultPrimCommand() { }
+
+void UsdUndoSetDefaultPrimCommand::execute()
+{
+    UsdUndoBlock undoBlock(&_undoableItem);
+
+    // Check if the layer selected is the root layer.
+    if (_stage->GetRootLayer() != _stage->GetEditTarget().GetLayer()) {
+        TF_WARN(
```
pierrebai-adsk, line 46: Seeing there are already two places with that code, please move it into either lib/mayaUsd/utils/restrictions.h (would be a new file) or lib/usdUfe/ufe/Utils.h (I really wish Utils.h were split into more specific files, but oh well... we got half a dozens of headers named utils.h all over the place.)

lib/usdUfe/ufe/UsdUndoSetDefaultPrimCommand.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace USDUFE_NS_DEF {
+
+UsdUndoSetDefaultPrimCommand::UsdUndoSetDefaultPrimCommand(const UsdPrim& prim)
+    : Ufe::UndoableCommand()
+    , _prim(prim)
+{
+    _stage = _prim.GetStage();
+    if (!_stage) {
+        return;
```
pierrebai-adsk, line 34: Returning here does nothing. Did you meant to throw an exception? Otherwise, the check would need to move the execute() function.

lib/usdUfe/ufe/UsdUndoClearDefaultPrimCommand.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace USDUFE_NS_DEF {
+
+UsdUndoClearDefaultPrimCommand::UsdUndoClearDefaultPrimCommand(const UsdPrim& prim)
+    : Ufe::UndoableCommand()
+    , _prim(prim)
+{
+    _stage = _prim.GetStage();
+    if (!_stage) {
+        return;
```
pierrebai-adsk, line 34: Same comments: the check and return does nothing in the constructor. Need to throw or be checked in execute()

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -11,6 +12,46 @@ def debugMessage(msg):
...
+    def naturalOrderComparator(a, b):
```
pierrebai-adsk, line 17: I would move this function out. I would also rewrite it like this (You can run `python -m doctest` on this to verify it is working). My version handle case where the numbers are 1 and 01 which are equal but not the same.

```
def _splitInParts(a):
    '''
    Convert a string into a tuple of parts, where each part is either a string
    without digits or a tuple containing a number and the length of its representation.

    >>> _splitInParts('a')
    ('a',)
    >>> _splitInParts('a1')
    ('a', (1, 1))
    >>> _splitInParts('a01b')
    ('a', (1, 2), 'b')
    >>> _splitInParts('120')
    ((120, 3),)
    '''
    parts = []
    part = None
    for c in a:
        if c.isdigit():
            if type(part) is str:
                parts.append(part)
                part = None
            if part is None:
                part = (int(c), 1)
            else:
                part = (part[0] * 10 + int(c), part[1] + 1)
        else:
            if type(part) is tuple:
                parts.append(part)
                part = None
            if part is None:
                part = c
            else:
                part = part + c
    if part is not None:
        parts.append(part)
    return tuple(parts)


def stringNumCompare(a, b):
    '''
    Compare two strings, but compare embedded numbers as numbers.
    >>> stringNumCompare('a', 'b')
    -1
    >>> stringNumCompare('b', 'a')
    1
    >>> stringNumCompare('abcd', 'abcd')
    0
    >>> stringNumCompare('10', '9')
    1
    >>> stringNumCompare('7', '12')
    -1
    >>> stringNumCompare('10a', '9b')
    1
    >>> stringNumCompare('7f', '12cde')
    -1
    >>> stringNumCompare('abcd10', 'abcd1')
    1
    >>> stringNumCompare('abcd01', 'abcd01')
    0
    >>> stringNumCompare('abcd1', 'abcd01')
    -1
    >>> stringNumCompare('abcd3', 'abcd1')
    1
    >>> stringNumCompare('abcd1', 'abcd5')
    -1
    '''
    a_parts = _splitInParts(a)
    b_parts = _splitInParts(b)
    if a_parts < b_parts:
        return -1
    elif a_parts > b_parts:
        return 1
    else:
        return 0
```

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -11,6 +12,46 @@ def debugMessage(msg):
...
+                while((i + 1 < lenA) and a[i+1].isdigit()):
+                    i += 1
+                    numA += a[i]
+                while((j + 1 < lenB) and b[j+1].isdigit()):
+                    j += 1
+                    numB += b[j]
+
+                if(numA == numB):
+                    continue
+                return int(numA) - int(numB)
```
pierrebai-adsk, line 33: If you keep your version, you wil lhave to fix that you are returning and int here but returning boolean (True, False) saying if a is greater or not at line 38, but return (True, False) saying if they are equal or not at line 39!

pierrebai-adsk: (I just tested your version, and it goes into an infinite loop if both strings are equal but end with digits. I guess root prim names never are equal, so that is why it happens to work).

pierrebai-adsk (approved):

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -23,6 +64,23 @@ def GetDefaultPrimName(proxyShape):
...
+        proxyStage = mayaUsd.ufe.getStage(proxyShape)
+        if(not primName):
+           proxyStage.ClearDefaultPrim()
+        defautlPrim = None
+        if proxyStage:
+            for prim in proxyStage.TraverseAll():
+                if(primName == prim.GetName()):
+                    defautlPrim = prim
+        if defautlPrim:
+            proxyStage.SetDefaultPrim(defautlPrim)
```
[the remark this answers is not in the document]
seando-adsk: Yes I think it should be undoable and use the command. @pierrebai-adsk Can you fix this in your follow-up PR?

seando-adsk (commented):
This new code (command and contextops) requires unit test(s). Since @pierrebai-adsk is blocked by these changes he has agreed to add the missing unit tests. I'll merge this PR now then.

PR #3398, opened 2023-10-18
EMSUSD-632 prevent root metadata changes when not targeting root

pierrebai-adsk:
- Add a applyRootLayerMetadataRestriction helper function to enforce not modifying the root layer metadata when not targeting the root layer.
- The only restriction applied is about the default prim.
- Call it from within the applyCommandRestriction helper function.
- Call it from within the default prim commands.
- Add unit tests for the rename, parent and group default prim restrictions.
- Add tests for the default prim commands.

### DIFF
--- a/plugin/adsk/scripts/AETemplateHelpers.py
+++ b/plugin/adsk/scripts/AETemplateHelpers.py
@@ -68,7 +68,7 @@ def SetDefaultPrim(proxyShape, primName):
         return True
     except Exception as e:
         # Note: we do want to tell the user why the set or clear failed.
-        OpenMaya.MGlobal.displayError((str(e)))
+        OpenMaya.MGlobal.displayError(str(e))
         return False
     
 def GetRootLayerName(proxyShape):
--- a/plugin/adsk/scripts/AETemplateHelpers.py
+++ b/plugin/adsk/scripts/AETemplateHelpers.py
@@ -1,7 +1,8 @@
-import functools
 import os.path
 import maya.cmds as cmds
-import ufe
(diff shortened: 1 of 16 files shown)


pierrebai-adsk:
The Attribute Editor now uses the command and can be undone.

pierrebai-adsk:
Once again, the only PF failure is one interactive test in a single PF where a USD file was left behind in the temp folder from e previous PF, that failure can be ignored.

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -47,19 +50,26 @@ def GetDefaultPrimName(proxyShape):
...
+            return False
+        
+        ufeCmd.execute(cmd)
+        return True
     except Exception as e:
-        debugMessage('SetDefaultPrim() - Error: %s' % str(e))
-        pass
-    return ''
+        # Note: we do want to tell the user why the set or clear failed.
+        OpenMaya.MGlobal.displayError((str(e)))
```
seando-adsk, line 71: Not a big deal, but isn't there an extra set of () here?

test/lib/ufe/testDefaultPrimCmds.py
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env python
+
+#
+# Copyright 2020 Autodesk
```
seando-adsk, line 4: Should be current year.

seando-adsk (changes requested):
Great work. Just a couple of very minor comments.

seando-adsk (approved):

PR #3401, opened 2023-10-20
EMSUSD-636 use the same saved folder location as when loading a stage.

pierrebai-adsk:
This allows remembering the last loaded layer folder between Maya session.


vlasovi (approved):
Looks good

Issue #3243, opened 2023-07-22
[EMSUSD-499] Dirty layers don't get saved in Maya files


NickWu:
**Describe the bug**
Hi, I found an issue where dirty layers don't get saved in Maya files when the `mayaUsdProxyShape` is created via `stageCacheId`. 

**Steps to reproduce**

1. Run the following code in a new scene:
```python
from pxr import Usd, UsdUtils
from maya import cmds 

filePath = './proxyPrimExample.usda'
stageCache = UsdUtils.StageCache.Get()
with Usd.StageCacheContext(stageCache):
    cachedStage = Usd.Stage.Open(filePath)
    
stageId = stageCache.GetId(cachedStage).ToLongInt()
shapeNode = cmds.createNode('mayaUsdProxyShape')
cmds.setAttr('{}.stageCacheId'.format(shapeNode), stageId)
cmds.setAttr('{}.filePath'.format(shapeNode), 
    cachedStage.GetRootLayer().identifier, type='string')
```
At this point the scene looks like this:
<img width="1169" alt="Screenshot 2023-04-18 at 15 02 43" src="https://user-images.githubusercontent.com/935278/232920571-f8e38088-5e68-4e44-b25e-5aa6027f131d.png">

2. Right click on the group prim and go **Variant Sets -> geo_vis -> default** to switch its `geo_vis` variant set value from `preview` to `default`. You can see the cube now has more polygons:
<img width="1006" alt="Screenshot 2023-04-18 at 15 52 57 (1)" src="https://user-images.githubusercontent.com/935278/232922162-3ab0a335-6fc1-4b7f-8c47-2ad89720df70.png">

3. Save the Maya file with the default USD setting which is to save the layer edits (dirty layers) to the maya file:
<img width="600" alt="Screenshot 2023-07-21 at 13 59 29" src="https://github.com/Autodesk/maya-usd/assets/935278/bd4fcddc-cb3e-467f-8818-8aa079e3e624">

4. Now, close Maya and relaunch Maya. Open the saved scene, and you will see the group's `geo_vis` variant is back to its default value which is `preview`, which means the variant change we did was not saved.

<img width="600" alt="Screenshot 2023-07-21 at 14 33 31" src="https://github.com/Autodesk/maya-usd/assets/935278/a7e0fe17-db33-4408-a24c-83aaad9d4d2a">

**Expected behavior**
[...]

PR #3402, opened 2023-10-20
EMSUSD-499 - Dirty layers don't get saved in Maya files

samuelliu-adsk:
For details and discussions please see this github issue: https://github.com/Autodesk/maya-usd/issues/3243

The idea is to check if the cache is actually connected (for example bifrost graph), if not this is not actually an incoming stage

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -1586,7 +1586,13 @@ bool MayaUsdProxyShapeBase::isStageIncoming() const
         CHECK_MSTATUS_AND_RETURN(localStatus, false);
         const auto cacheId = UsdStageCache::Id::FromLongInt(cacheIdNum);
         const auto stageCached = cacheId.IsValid() && UsdUtilsStageCache::Get().Contains(cacheId);
-        if (stageCached) {
+
+        // Check what is the cache connected to
+        MStringArray result;
+        MGlobal::executeCommand(("listConnections -shapes on " + this->name()), result);
+
+        // The stage is only incoming if the cache is connected to a shape
+        if (stageCached && result.length()) {
             isIncomingStage = true;
         }
     }


pierrebai-adsk (approved):

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -1586,7 +1586,13 @@ bool MayaUsdProxyShapeBase::isStageIncoming() const
         CHECK_MSTATUS_AND_RETURN(localStatus, false);
         const auto cacheId = UsdStageCache::Id::FromLongInt(cacheIdNum);
         const auto stageCached = cacheId.IsValid() && UsdUtilsStageCache::Get().Contains(cacheId);
-        if (stageCached) {
+
+        // Check what is the cache connected to
+        MStringArray result;
+        MGlobal::executeCommand(("listConnections -shapes on " + this->name()), result);
```
NickWu, line 1592: Hi @samuelliu-adsk , thanks for working on this for the issue https://github.com/Autodesk/maya-usd/issues/3243. I tested it but unfortunately it doesn't resolve the issue due to the following code: 

`MGlobal::executeCommand(("listConnections -shapes on " + this->name()), result);`

This is because a `mayaUsdProxyShape` usually has a `time1` node's connection to its `time` attribute. The maya-usd code base does this as well. Please see https://github.com/Autodesk/maya-usd/blob/dev/plugin/adsk/scripts/mayaUsd_createStageWithNewLayer.py#L39 and https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/ufe/UsdUndoCreateStageWithNewLayerCommand.cpp#L150-L151. Hence, `MGlobal::executeCommand(("listConnections -shapes on " + this->name()), result)` would return a `time1` node and the `isIncomingStage` would be set to `true` which is not correct. I guess you might have been using the example code on the issue page https://github.com/Autodesk/maya-usd/issues/3243 to test the solution but the code example is missing `time1` node's connection. Sorry for that.

According to the logic we discussed on the issue page: 

> If "mayaUsdProxyShape.stageCacheId" is connected, don't save, else if the stageCacheId refer an existing stage save its content.

I would suggest the following solution (I tested it and it resolves the issue): 

```C++
bool isIncomingStage = false;
if (!inDataHandle.data().isNull()) {
    isIncomingStage = true;
} 
```
This is because the `!inDataHandle.data().isNull() `check already takes care of checking whether the proxy shape has an incoming connection to its `inStageData` attribute which might be from a bifrost graph or another proxy shape node. If there is no connection to it then we can consider `isIncomingStage` is false. I don't think we need to check the `stageCacheId` value because if it has value and is valid and in USD stage cache then `isIncomingStage` should be false. And if `stageCacheId` value is set but it's not valid then `isIncomingStage` should be false too. 

I did a test for our internal patch which you could use to test as well: 

```python
from pxr import Usd, UsdUtils, UsdGeom
from maya import cmds 
import mayaUsd.lib
import tempfile

# you can find this file in maya-usd repo: maya-usd/test/lib/mayaUsd/nodes/ProxyShapeBaseTest/CubeModel.usda
filePath = '/home/users/nickw/Test/CubeModel.usda'
stageCache = UsdUtils.StageCache.Get()
with Usd.StageCacheContext(stageCache):
    cachedStage = Usd.Stage.Open(filePath)
    
stageId = stageCache.GetId(cachedStage).ToLongInt()
shapeNode = cmds.createNode('mayaUsdProxyShape')
cmds.setAttr('{}.stageCacheId'.format(shapeNode), stageId)
cachedStage.SetEditTarget(cachedStage.GetSessionLayer())

fullPath = cmds.ls(shapeNode, long=True)[0]
cmds.select("{},/CubeModel".format(fullPath))
# Move the cube prim
cmds.move(0, 0, 2, r=True)

# Make sure shareStage is ON
assert cmds.getAttr('{}.{}'.format(shapeNode,"shareStage"))

# Make sure there is no connection to inStageData
assert not cmds.listConnections('{}.inStageData'.format(shapeNode), s=True, d=False)

cmds.select(cmds.listRelatives(fullPath, p=True)[0], r=True)

with tempfile.TemporaryDirectory(prefix='ProxyShapeBase') as testDir:
    pathToSave = "{}/CubeModel.ma".format(testDir)

    cmds.file(rename=pathToSave)
    cmds.file(save=True, force=True, type='mayaAscii')
    
    cmds.file(pathToSave, force=True, open=True)

    stage = mayaUsd.lib.GetPrim(fullPath).GetStage()
    prim = stage.GetPrimAtPath('/CubeModel')
    assert prim.IsValid()
    
    # Verify we get the saved changes we did 
    xform = UsdGeom.Xformable(prim)
    translate = xform.GetOrderedXformOps()[0].Get()
    assert translate[2] == 2

```

samuelliu-adsk: Thanks @NickWu , I'll make a new PR with the fixes.

NickWu: Hi @samuelliu-adsk , thanks for working one this. I forgot to add the code to connect `time1` again. I have updated the test code with this line `cmds.connectAttr('time1.outTime', '{}.time'.format(shapeNode))`

PR #3403, opened 2023-10-20
Add test cases for material fanned out file nodes

AramAzhari-adsk:
This is a follow up to https://github.com/Autodesk/maya-usd/pull/3366 to add test cases. 
There is also a minor change from MObjectArray to MObjectHandleUnorderedSet.

The test scene contains 5 cubes that cover various materials:

File Node fanning out to two other File Nodes to the same Material (Lambert):
image

File Node fanning out to two other File Nodes to the same Material (Standard Surface):
image

File Node fanning out to two other File Nodes to the same Material (USD Preview Surface):
image

File Node fanning out to two other File Nodes to different Materials (Lambert):
image


lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -283,8 +285,8 @@ class UseRegistryShadingModeExporter : public UsdMayaShadingModeExporter
                 }
 
                 bool allowedNode = false;
```
JGamache-autodesk, line 287: Using a MObjectHandleUnorderedSet has powerful benefits:

```suggestion
                if (!allowedNodes.count(dstPlug.node()) {
                    continue
                }
```

This replaces the whole `for` loop, and all `allowedNode` boolean logic.

test/lib/usd/translators/CMakeLists.txt
```diff
@@ -133,6 +133,7 @@ endforeach()
 # it takes precedence over the provided material scope name.
 set(TEST_SCRIPT_FILES_WITH_MATERIAL_SCOPE
     testUsdExportDisplacement.py
+	testUsdExportFannedOutFileNodesMaterial.py
```
JGamache-autodesk, line 136: Minor: Tab vs spaces. Can you clean that up please.

JGamache-autodesk (changes requested):
This is not actually testing the fix you implemented, which was about stray file nodes.

Please add checks that:

1 - For UsdPreviewSurface exports:
    file13 and file15 are present for lambert4, file14 is not
    file16 and file18 are present for standardSurface2, file17 is not
      ....
    file10 is present for lambert3, file11 and file12 are not
    file12 is present for lambert2, file10 and file11 are not

2- For MaterialX exports:
    file13, file14 and file15 are present for lambert4
    file16, file17 and file18 are present for standardSurface2
      ....
    file10 and file11 are present for lambert3, file12 is not
    file11 and file12 are present for lambert2, file10 is not

AramAzhari-adsk:
Updated the test logic to correctly check for nodes that must be included and the nodes that must not be included. I also locally changed the list to make sure it is functioning properly.

Note: The purpose of the fix in https://github.com/Autodesk/maya-usd/pull/3366 was to prevent adding nodes that do not exist in Maya's dependency graph for each material. The fix or its test does **not** cover cases where an unsupported connection is not translated onto USD, which may end up containing zombie nodes that are not connected to anything. That is a separate issue and I will create a ticket to have it fixed based on priority.

JGamache-autodesk (dismissed):

JGamache-autodesk (approved):

seando-adsk (approved):

PR #3404, opened 2023-10-20
EMSUSD-715 fix context menu in viewport

pierrebai-adsk:
We must create the pop-up menu explicitly, otherwise the menu fails to show in some circumstances.

### DIFF
--- a/plugin/adsk/scripts/USDMenuProc.mel
+++ b/plugin/adsk/scripts/USDMenuProc.mel
@@ -105,6 +105,7 @@ global proc USDMenuProc(string $parent, string $obj)
 {
     $obj = expandToSelection($obj);
 
+    popupMenu -e -dai $parent;
     setParent -menu $parent;
     
     python("import maya.app.ufe.outlinerSupport; maya.app.ufe.outlinerSupport.buildNonMayaContextMenu('''" + $obj + "''')");
--- a/plugin/adsk/scripts/USDMenuProc.mel
+++ b/plugin/adsk/scripts/USDMenuProc.mel
@@ -105,7 +105,6 @@ global proc USDMenuProc(string $parent, string $obj)
 {
     $obj = expandToSelection($obj);
 
-    popupMenu -e -dai $parent;
     setParent -menu $parent;
     
     python("import maya.app.ufe.outlinerSupport; maya.app.ufe.outlinerSupport.buildNonMayaContextMenu('''" + $obj + "''')");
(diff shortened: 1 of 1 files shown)


plugin/adsk/scripts/USDMenuProc.mel
```diff
@@ -105,6 +105,9 @@ global proc USDMenuProc(string $parent, string $obj)
 {
     $obj = expandToSelection($obj);
 
+    popupMenu -e -dai $parent;
+    setParent -menu $parent;
+
```
seando-adsk, line 110: The previous code also had `setParent -menu ..;`

Note: this doesn't actually create the menu - it is created by Maya. What this is doing is clearing out the menu and then setting it for subsequent `menuItem` calls.

pierrebai-adsk: A! Well, the bug was in Maya, buildNonMayaContextMenu calls MEL command that fail because there are no menu active. So something in those two lines does the trick. I wrote those because they were in here before my previous change to use the UFE context menu, so I thought they were needed.

pierrebai-adsk: If you got the right line, we can commit it with "suggested code changes" in this review.

seando-adsk: I'm actually confused why the previous code also had `setParent -menu ..` - the line above set the menu parent and then this one would make it move up one level.

Can you do a test and print out what the original parent is and what the calls to the two different `setParent` would be?

pierrebai-adsk: The problem being fixed here is that as reported sometimes the parent is not a menu and does not have a popup menu. The version without the popuMenu call would only work by a happy accident most of the time. I'll check what were the parent when it work and did not work.

pierrebai-adsk: he parent being passed in is for example "MainPane|viewPanes|modelPanel4|modelPanel4|modelPanel4|modelPanel4ObjectPop". What is needed is to set it as the parent to be used by the menu being created.

seando-adsk: Thanks for checking. I wonder if we should leave in the `popupMenu -e -dai $parent;` which would clear the menu of all items (in case it has any) as we are going to populate it with MayaUsd menu items.

pierrebai-adsk: OK, I'll put it back then.

seando-adsk (approved):

PR #3418, opened 2023-10-26
EMSUSD-499 - Dirty layers don't get saved in Maya files

samuelliu-adsk:
A fix for previous PR: https://github.com/Autodesk/maya-usd/pull/3402
Github issue: https://github.com/Autodesk/maya-usd/issues/3243

Adding `-t shape` to the command will exclude the "time" nodes but leave only Bifrost nodes.
Deleting the `else` part will not work because for Bifrost to save properly we will need to check for cache

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -1589,7 +1589,7 @@ bool MayaUsdProxyShapeBase::isStageIncoming() const
 
         // Check what is the cache connected to
         MStringArray result;
-        MGlobal::executeCommand(("listConnections -shapes on " + this->name()), result);
+        MGlobal::executeCommand(("listConnections -t shape -shapes on " + this->name()), result);
 
         // The stage is only incoming if the cache is connected to a shape
         if (stageCached && result.length()) {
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -1589,7 +1589,7 @@ bool MayaUsdProxyShapeBase::isStageIncoming() const
 
         // Check what is the cache connected to
         MStringArray result;
-        MGlobal::executeCommand(("listConnections -shapes on " + this->name()), result);
+        MGlobal::executeCommand(("listConnections -t shape -shapes on " + this->name()), result);
```
pierrebai-adsk, line 1593: Is there any reason why you did not choose to go with the solution that was suggested on github? I would like to understand why that solution was not correct? The suggestion was:
```
bool isIncomingStage = false;
if (!inDataHandle.data().isNull()) {
    isIncomingStage = true;
} 
```

samuelliu-adsk: Hi @pierrebai-adsk , the reason is stated in the description. So if we remove the `else` part and leave it like this it will not work properly for Bifrost. For some reason (I get the conclusion from debugging),  inDataHandle can still be Null even if we have Bifrost shape connected to it. In that case, we need to check the stage Cache and find the Bifrost shape in Cache. Hence the `else` part.

pierrebai-adsk (approved):

PR #3421, opened 2023-10-30
EMSUSD-550 - Disable Material section when Mesh type is excluded from export

samuelliu-adsk:
Disable the material section when user excludes Meshes from export. Currently the material is attached to the mesh so when the mesh is not exported, the material can not be exported either.

### DIFF
--- a/plugin/adsk/scripts/mayaUsdTranslatorExport.mel
+++ b/plugin/adsk/scripts/mayaUsdTranslatorExport.mel
@@ -254,6 +254,7 @@ global proc mayaUsdTranslatorExport_MeshCB() {
         return;
 
     columnLayout -e -enable (`checkBoxGrp -q -v1 exportMeshesCheckBox` == 1) meshExportOptsCol;
+    columnLayout -e -enable (`checkBoxGrp -q -v1 exportMeshesCheckBox` == 1) materialOptsCol;
 }
 
 // Callback called when the animation frame layout is expanded.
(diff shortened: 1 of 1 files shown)


seando-adsk (approved):

samuelliu-adsk:
Hi @seando-adsk , just realize I have some debug code forgot to remove. Added another commit

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -488,10 +489,6 @@ global proc mayaUsdTranslatorExport_SetFromOptions(string $currentOptions, int $
                 mayaUsdTranslatorExport_SetCheckbox($optionBreakDown[1], $enable, "exportDisplayColorCheckBox");
             } else if ($optionBreakDown[0] == "exportInstances") {
                 mayaUsdTranslatorExport_SetOptionMenuByBool($optionBreakDown[1], $enable, "exportInstancesPopup");
-            } else if($optionBreakDown[0] == "exportCameras"){
-                mayaUsdTranslatorExport_SetOptionMenuByBool($optionBreakDown[1], $enable, "exportInstancesPopup");
-            } else if($optionBreakDown[0] == "exportLights") {
-                mayaUsdTranslatorExport_SetOptionMenuByBool($optionBreakDown[1], $enable, "exportInstancesPopup");
```
seando-adsk, line 494: Why were these lines removed?

samuelliu-adsk: Those lines are some debug code I used before and forgot to remove. They are totally useless now...

seando-adsk: Why would they be useless. Someone can call this function adding those strings to the input options to control the starting value of these two export options.

samuelliu-adsk: Oh no those flags are not longer available. Now we use `excludeExportTypes` flag, so it will be `excludeExportTypes=[Cameras, Lights]` like this. These two flags are abandoned. It's my fault I forgot to delete them here...

seando-adsk (approved):

PR #3427, opened 2023-10-31
EMSUSD-254 disable the default prim auto-filling.

pierrebai-adsk:
- Disable to code that auto-fills the prim path when creating a ref or payload.
- We are keeping the code because we think we likely will put it back in the next release under an advance option.
- Update the tooltip to be clearer and remove text about auto-filling.


pierrebai-adsk:
Sorry, we need to fix the refresh button.

seando-adsk (approved):

PR #3428, opened 2023-11-01
EMSUSD-490 Fixes Redo Visibility Visual Update in Outliner

AramAzhari-adsk:
Previously we fixed the Undo command that wasn't triggering a visibility update on the Outliner.
This change is about fixing the Re-do which is a different attribute event.

### DIFF
--- a/lib/usdUfe/ufe/StagesSubject.cpp
+++ b/lib/usdUfe/ufe/StagesSubject.cpp
@@ -157,13 +157,19 @@ void sendAttributeChanged(
                 notifyWithoutExceptions<Ufe::Camera>(ufePath);
             }
         } else {
+            // Special case when Redo-ing visibility change, the notice.GetChangedInfoOnlyPaths()
+            // does not contain the change, hence handling visibility notification in re-sync path.
+            if (changedToken == UsdGeomTokens->visibility) {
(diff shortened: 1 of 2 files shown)


lib/usdUfe/ufe/StagesSubject.cpp
```diff
@@ -157,13 +157,19 @@ void sendAttributeChanged(
                 notifyWithoutExceptions<Ufe::Camera>(ufePath);
             }
         } else {
+            // Special case when Redo-ing visibility change, the notice.GetChangedInfoOnlyPaths()
+            // does not contain the change, hence handling visibility notification in re-sync path.
+            if (changedToken == UsdGeomTokens->visibility) {
+                Ufe::VisibilityChanged vis(ufePath);
+                notifyWithoutExceptions<Ufe::Object3d>(vis);
+            }
```
seando-adsk, line 165: This notif is only within the ifdef V4 block, which means it won't be sent for Maya 2022/2023. That is why the test is failing.

AramAzhari-adsk: Thanks for catching that.

seando-adsk (approved):

PR #3432, opened 2023-11-02
EMSUSD-749 fix how stages are updated when saved

pierrebai-adsk:
The interaction between how the stage were initially open or created and how they were updated did not match. This caused the code that attempted to update the stage cache when saving anonymous layers to fail to achieve its goal. This led to the proxy shape next compute to fail to find the original session layer. This caused the information about edited-as-Maya prim to be lost. This resulted in the stage being in a weird state where a Maya reference prim was edited but the stage had forgotten about it.

Fix the code so that the way the stage os open in the proxy shape and in the function that update the stage cache match.

- Changed how the proxy shape open the USD stage by always initially not loading payloads.
- Also, do not pass an explicit resolver as it would affect which stage could be found in the cache.
- Document in the helper function that updates stage after saving anonymous layer to open the stages the same way.

We also needed to fix the layer manager UI update after save.

The previous code worked by happenstance of the order of indirect operations. In particular, it only worked if the proxy shape was recomputed at just the right time. Recomputing the proxy shape emitted a notice that refreshed the layer manager. For complex reasons, when there is an edited Maya reference, the order changes, which prevented the layer manager UI from updating with the latest data.

To avoid this reliance on the exact timing of when the proxy shape might get recomputed, an explicit call to refresh the underlying model of the layer manager is done when saving. Unfortunately, due to the complicated design of the layer manager, the underlying tree model and where the data is kept, multiple classes had to be modified to provide a path from the save code the the data model.

- Add functions on the MayaSessionState to refresh the current displayed stage layers.
- (The MayaSessionState is where the layer manager UI keeps its list of loaded USD stages... it really ough to be called LoadedStages instead... MayaSessionState is a combination of three meaningless words.)
- Add a updateLayerModel() abstract function to the AbstractLayerEditorWindow interface.
- Implement updateLayerModel() in MayaLayerEditorWindow by calling the refreshStageEntry() on the MayaSessionStage.
- Add a updateLayerManagers(0 helper function to the LayerDatabase used by the LayerManager (that is, the data holder, not the UI), to refresh all layer manager UI.
- Call it at the end of LayerDatabase::saveUsd(), which is where layers get saved.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -821,7 +821,7 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
     bool isIncomingStage = false;
 
     // Note: we explicitly always load no payload initially and will load them
-    //       later on if requested. The reason is that there are otehr code elsewhere
+    //       later on if requested. The reason is that there are other code elsewhere
     //       that updates the stages. Those functions must find the same existing stages.
     //       Somewhat unfortunately, the OpenUSD API to open a stage only finds an existing
     //       stage if *all* parameters passed to open the stage match, including the initial
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -820,19 +820,25 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
 
     bool isIncomingStage = false;
 
-    // Compute the load set for the stage.
-    MDataHandle loadPayloadsHandle = dataBlock.inputValue(loadPayloadsAttr, &retValue);
-    CHECK_MSTATUS_AND_RETURN_IT(retValue);
-
-    UsdStage::InitialLoadSet loadSet = loadPayloadsHandle.asBool()
-        ? UsdStage::InitialLoadSet::LoadAll
-        : UsdStage::InitialLoadSet::LoadNone;
-
-    // If there is a dynamic attribute containing the exact load rules
-    // for payload, start by loading nothing. The correct payload will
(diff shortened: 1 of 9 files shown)


lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -820,19 +820,25 @@ MStatus MayaUsdProxyShapeBase::computeInStageDataCached(MDataBlock& dataBlock)
...
-        ? UsdStage::InitialLoadSet::LoadAll
-        : UsdStage::InitialLoadSet::LoadNone;
-
-    // If there is a dynamic attribute containing the exact load rules
-    // for payload, start by loading nothing. The correct payload will
-    // be loaded by the load rules.
-    if (hasLoadRulesAttribute(*this))
-        loadSet = UsdStage::InitialLoadSet::LoadNone;
+    // Note: we explicitly always load no payload initially and will load them
+    //       later on if requested. The reason is that there are otehr code elsewhere
```
AramAzhari-adsk, line 824: minor typo, other

lib/mayaUsd/utils/utilSerialization.cpp
```diff
@@ -137,6 +137,9 @@ void updateAllCachedStageWithLayer(SdfLayerRefPtr originalLayer, const std::stri
         std::vector<UsdStageRefPtr> stages = cache.FindAllMatching(originalLayer);
         for (const auto& stage : stages) {
             auto sessionLayer = stage->GetSessionLayer();
+            // Note: See comments in lib\mayaUsd\nodes\proxyShapeBase.cpp, line 823
```
AramAzhari-adsk, line 140: The line number is most likely to go out of sync. Perhaps we could refer to the comment by function name ?

AramAzhari-adsk (approved):
Just some minor comment updates but looks good to me.

AramAzhari-adsk (approved):

PR #3441, opened 2023-11-03
EMS-764 properly reset the select prim dialog.

pierrebai-adsk:
Also, remove the parent window warning that serves no purpose for the user.


lib/usd/ui/importDialog/TreeModel.h
```diff
@@ -23,6 +23,7 @@
 #include <mayaUsdUI/ui/api.h>
 
 #include <pxr/usd/usd/prim.h>
+#include <pxr/usd/usd/stage.h>
```
AramAzhari-adsk, line 26: Is this needed in this header?

AramAzhari-adsk (approved):
Looks good to me.

PR #3443, opened 2023-11-06
EMSUSD-614 - Fixes Edit As Maya Options Convert Instances Not Retaining From Session to Session

AramAzhari-adsk:
EMSUSD-614 - Fixes Edit As Maya Options Convert Instances Not Retaining
Also fixed the height issue on Edit As Maya Option Dialog


vlasovi (approved):
Looks good

PR #3451, opened 2023-11-09
Update Readme.md

macevhicz:
I appears that the Frame Sample example for frameRange  = [1, 3], frameStride = 2.0, frameSample = [-3, 3] is incorrect. The resulting frames would be [-2.0, 0.0, 4.0, 6.0], if I'm understanding this correctly.


seando-adsk:
@santosg87 Can you verify this.

santosg87 (approved):

PR #3459, opened 2023-11-13
EMSUSD-675 - Build USD v23.11 with Python 3.11/OSD 3.6 and update ecg-maya-usd

seando-adsk:
#### EMSUSD-675 - Build USD v23.11 with Python 3.11/OSD 3.6 and update ecg-maya-usd

* Disabled test that is failing with USD 23.11 because of problem in Usd. https://github.com/PixarAnimationStudios/OpenUSD/issues/2800


JGamache-autodesk (dismissed):
You can possibly add the URL of the issue on the Pixar tracker.

PR #3474, opened 2023-11-20
EMSUSD-603 - Errors when trying to cache a rig with merge transforms ON and namespaces OFF

samuelliu-adsk:
- The issue is caused by transforms merged to parent node, which will cause a mismatch between SDF and DAG paths and ultimately cause clashes. We should only write to the node's parent node when it a shape node.
- Removed the `channel1` part of the unity test because that attribute is created because of the bug above. I compared the cache to usd result to a normal export and the files matches (there's no `channel1 ` in normal export usd files)

### DIFF
--- a/lib/mayaUsd/fileio/writeJobContext.cpp
+++ b/lib/mayaUsd/fileio/writeJobContext.cpp
@@ -195,7 +195,8 @@ SdfPath UsdMayaWriteJobContext::ConvertDagToUsdPath(const MDagPath& dagPath) con
     // write to the parent (transform) path instead.
     MDagPath parentDag(dagPath);
     parentDag.pop();
-    if (IsMergedTransform(parentDag)) {
+
+    if (IsMergedTransform(parentDag) && UsdMayaUtil::isShape(dagPath)) {
         path = path.GetParentPath();
     }
 
--- a/lib/mayaUsd/utils/util.h
+++ b/lib/mayaUsd/utils/util.h
@@ -690,6 +690,8 @@ void AddMayaExtents(
 MAYAUSD_CORE_PUBLIC
 SdrShaderNodePtrVec GetSurfaceShaderNodeDefs();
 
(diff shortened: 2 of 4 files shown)


test/lib/mayaUsd/fileio/testCacheToUsd.py
```diff
@@ -471,8 +471,6 @@ def testEditAndMergeRigMayaRef(self):
         value = xformOp.Get()
         self.assertIn("xformOp:translate", value)
         self.assertIn("xformOp:rotateXYZ", value)
-        self.assertIn("xformOp:translate:channel1", value)
-        self.assertIn("xformOp:rotateXYZ:channel1", value)
```
samuelliu-adsk, line 475: Removed these two lines because those are not valid any more, but I don't know if there's any other ways to test it...

lib/mayaUsd/fileio/writeJobContext.cpp
```diff
@@ -195,7 +195,8 @@ SdfPath UsdMayaWriteJobContext::ConvertDagToUsdPath(const MDagPath& dagPath) con
     // write to the parent (transform) path instead.
```
samuelliu-adsk, line 195: As the comment stated, we should only write to parent if the current node is a shape node

lib/mayaUsd/utils/util.h
```diff
@@ -690,6 +690,8 @@ void AddMayaExtents(
 MAYAUSD_CORE_PUBLIC
 SdrShaderNodePtrVec GetSurfaceShaderNodeDefs();
 
+bool isShape(const MDagPath& dagPath);
```
pierrebai-adsk, line 693: I think we shoudl export it by adding MAYAUSD_CORE_PUBLIC, like the other functions.

pierrebai-adsk (changes requested):
Only need that export and it will be all good

pierrebai-adsk (approved):

Issue #1337, opened 2021-04-14
[EMSUSD-599] USD files not registered in the filePath editor


yolov:
**Describe the bug**
usd files are not registered in the maya file path editor. This makes maya unaware that there are usd file dependencies to the maya scene and also archiving a maya scene will not pack the usd files.

**Steps to reproduce**
1. load any usd file in maya
2. go to window > general editors > filepath editor // it will be empty
or alternatively
2. save the maya scene
3. go to file > archive and inspect the resulting archive

**Expected behavior**
USD files to show up in the filepath editor and to be packed with maya scene archiving.

**Specs (if applicable):**
 - Windows 10
 - Maya 2022
 - Maya USD 0.9.0

swillisart:
Adding this here as the method in which both _File Path Editor_ and _File>Archive Scene_ use to get files paths is from the command **listAttr -usedAsFilename** 
The mayaUsdProxyShape does not currently yield it's path(s) to it.

santosg87:
@yolov we have made a fix for this issue on #3482. it is worth noting that registering the usd file paths still doesn't solve the archiving issue - but you should be able to see the files on the file path editor with that fix.

the fix is not currently on any of the released versions of mayaUSD, you would need to compile the plugin with a commit that includes #3482 in order to see the fix!

closing this issue for the time being. if this is still a problem feel free to reopen :)

yolov:
Thanks, @santosg87.

PR #3482, opened 2023-11-24
EMSUSD-599 - USD data is not registered on the Maya File Path Editor

samuelliu-adsk:
https://github.com/Autodesk/maya-usd/issues/1337
Register usd files to file path editor

### DIFF
--- a/plugin/adsk/plugin/plugin.cpp
+++ b/plugin/adsk/plugin/plugin.cpp
@@ -367,6 +367,12 @@ MStatus uninitializePlugin(MObject obj)
     status = plugin.deregisterCommand(MayaUsd::MayaUsdUndoBlockCmd::commandName);
     CHECK_MSTATUS(status);
 
+    // Deregister from file path editor
+    status
+        = MGlobal::executeCommand("filePathEditor -deregisterType \"mayaUsdProxyShape.filePath\" "
+                                  "-typeLabel \"mayaUsdProxyShape.filePath\" -temporary");
+    CHECK_MSTATUS(status);
+
     status = MayaUsd::ufe::finalize();
     CHECK_MSTATUS(status);
 
--- a/plugin/adsk/plugin/plugin.cpp
+++ b/plugin/adsk/plugin/plugin.cpp
@@ -51,6 +51,7 @@
 #include <maya/MFnPlugin.h>
 #include <maya/MStatus.h>
 #include <ufe/runTimeMgr.h>
+#include <maya/MGlobal.h>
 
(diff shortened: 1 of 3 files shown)


AramAzhari-adsk (approved):

plugin/adsk/plugin/plugin.cpp
```diff
@@ -243,6 +244,11 @@ MStatus initializePlugin(MObject obj)
         "mayaUsd_pluginBatchLoad",
         "mayaUsd_pluginBatchUnload");
 
+    // Register to file path editor
+    status = MGlobal::executeCommand("filePathEditor -registerType \"mayaUsdProxyShape.filePath\" "
+                                     "-typeLabel \"mayaUsdProxyShape.filePath\" -temporary");
+    CHECK_MSTATUS(status);
```
seando-adsk, line 250: Was there a reason this wasn't done in the [mayaUsd_pluginUICreation.mel](https://github.com/Autodesk/maya-usd/blob/dev/plugin/adsk/scripts/mayaUsd_pluginUICreation.mel)

And then the also using `deregisterType` in the mayaUsd_pluginUIDeletion (when our plugin is unloaded).

samuelliu-adsk: I see that all the registrations happening here (registerShape etc.) so I assume this is the place we register stuff? 
Yes I will add  `deregisterType`

seando-adsk: All the registration stuff here is done using the Maya C++ API. What you are calling is a MEL command. My thinking is that registering the filepatheditor type is a UI bit and thus belongs in our plugin UI creation file?

AramAzhari-adsk: My thinking is that the menu item resulting from this Maya API is similar to the drop down item resulting from Maya import / export api. This is a data-driven api that registers a type pertaining to file path editor, similar to how we register translator types to Maya import / export. 

In other words, we're not creating any UI.

seando-adsk: Okay far point. We still need the deregister though

samuelliu-adsk: Hi @seando-adsk , I just realized we don't actually need to deregister it, because I added the `-temporary` flag. With `-temporary` flag the type is only register for the current session. Correct me if I'm wrong

seando-adsk: Yes that `-temporary` flag means that it will only be for the current session. But if the user unloads the mayausd plugin then we should deregister the filepath.

seando-adsk (approved):
Looks good thanks.

PR #3487, opened 2023-11-29
EMSUSD-215 support custom display name for USD attributes

pierrebai-adsk:
Add support to have custom names for USD attributes to have nicer names for UI display purpose. We already had an algorithm to automatically make the names prettier, this adds support for user-defined attribute names. We now read the USD 'displayName' metadata on each attribute. If the metadata exists then we use its value as the display name.

The built-in function to make the name prettier has also been enhanced to remove the "xformOp:" prefix.

Modify the USD attribute-related classes:
- Add a displayName funtion to the UsdAttrbute class.
- Add a displayName funtion to the UsdAttributeHolder class.
- Add a displayName funtion to the UsdShaderAttributeHolder class.
- The displayName function in UsdAttributeHolder prettify the name.

Add support for display name in the Attribute Editor template:
- Refactor custom-control for various types into their own files.
- Add a AttributeCustomControl base class to handle display names.
- Make the default control creation use the display names.

Added a simple unit test.


lib/mayaUsd/resources/ae/usdschemabase/enum_custom_control.py
```diff
@@ -13,26 +13,25 @@
...
+from .attribute_custom_control import AttributeCustomControl
+
 import ufe
 import mayaUsd.ufe as mayaUsdUfe
 import mayaUsd.lib as mayaUsdLib
 import maya.cmds as cmds
 import maya.internal.ufeSupport.attributes as attributes
 
-class CustomEnumControl(object):
+class EnumCustomControl(AttributeCustomControl):
```
pierrebai-adsk, line 24: Changed the class name to follow the same pattern as the others.

pierrebai-adsk:
Will removed all refactoring and cleanups and pen another PR.

PR #3488, opened 2023-11-29
EMSUSD-835 - Provide access to the Export Selected options

samuelliu-adsk:
I think we still need to add them into default dictionary because ultimately we want to access this variable in export chasers


JGamache-autodesk (changes requested):
Good, but incomplete. You need to also add the export selected option in  `UsdMayaExportTranslator::writer()` in `plugin\adsk\plugin\exportTranslator.cpp`.

JGamache-autodesk (approved):

PR #3492, opened 2023-12-01
Pipe UsdStage data and Complexity to proxy nodes

dj-mcg:
This addresses an issue that was encountered with a recent Autodesk fix regarding sharing stages.


pierrebai-adsk (approved):
I have no real knowledge of pxr plugin details or assemblies to judge.... so looks good to me!

dj-mcg:
This should also allow you to enable testRefAssemblyDrawRepresentations

pierrebai-adsk:
Shall we put the read-for-merge label on this?

PR #3494, opened 2023-12-01
LOOKDEVX-2260 - Add a third party naming convention for shader outliner icons

JGamache-autodesk:
Allows any renderer like Arnold, VRay, or RenderMan to provide outliner icons for their custom nodes.

### DIFF
--- a/lib/mayaUsd/ufe/MayaUsdUIInfoHandler.cpp
+++ b/lib/mayaUsd/ufe/MayaUsdUIInfoHandler.cpp
@@ -23,9 +23,13 @@
 
 #include <maya/MDoubleArray.h>
 #include <maya/MGlobal.h>
+
+#ifdef UFE_V4_FEATURES_AVAILABLE
+#include <maya/MSceneMessage.h>
 #include <ufe/nodeDef.h>
 #include <ufe/nodeDefHandler.h>
 #include <ufe/runTimeMgr.h>
+#endif
 
 #include <algorithm>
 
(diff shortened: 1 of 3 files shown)


lib/usdUfe/ufe/UsdUIInfoHandler.cpp
```diff
@@ -184,6 +190,38 @@ Ufe::UIInfoHandler::Icon UsdUIInfoHandler::treeViewIcon(const Ufe::SceneItem::Pt
...
+    //       MyRenderer:nifty_surface
+    //   Will have this code search the full Maya icon path for a file named:
+    //       MyRenderer_nifty_surface.png
+    //   And will use it if found. At resolution 200%, the file:
+    //       MyRenderer_nifty_surface_200.png
+    //   Will alternatively be used if found.
+    //
+    if (auto shaderSchema = PXR_NS::UsdShadeShader(usdItem->prim())) {
+        static const auto iconContext = PXR_NS::ArDefaultResolverContext(
+            PXR_NS::TfStringSplit(PXR_NS::TfGetenv("XBMLANGPATH", ""), ARCH_PATH_LIST_SEP));
```
JGamache-autodesk, line 209: We might want to cache results so we do not hit the filesystem too hard every time we expand a shader node graph in the outliner.

feldstj (dismissed):
LGTM

feldstj (dismissed):

lib/mayaUsd/ufe/MayaUsdUIInfoHandler.cpp
```diff
@@ -88,5 +99,68 @@ UsdUfe::UsdUIInfoHandler::SupportedTypesMap MayaUsdUIInfoHandler::getSupportedIc
...
+            return icon;
+        }
+
+        std::string iconName = nodeDef->type();
+        std::replace(iconName.begin(), iconName.end(), ':', '_');
+        iconName += ".png";
+
+        // Since this will be hitting the filesystem hard, mostly to find nothing, let's cache
+        // search results. Note that "XBMLANGPATH" is Maya specific, which is why the code is here
+        // and not in the base class.
```
seando-adsk, line 139: I understand the desire for the cache. My only comment here is that this means a user is not able to change the "info:id" of the prim and see the new icon in the Outliner. They would have to quit/restart Maya. Is that a concern?

lib/mayaUsd/ufe/MayaUsdUIInfoHandler.cpp
```diff
@@ -88,5 +99,68 @@ UsdUfe::UsdUIInfoHandler::SupportedTypesMap MayaUsdUIInfoHandler::getSupportedIc
...
+        //
+        //  We take the info:id of the shader and make it safe by replacing : with _.
+        //  Then we search the Maya icon paths for a PNG file with that name. If found we will use
+        //  it. Please note that files with _150 and _200 can also be provided for high DPI
+        //  displays.
+        //
+        //   For example an info:id of:
+        //       MyRenderer:nifty_surface
+        //   Will have this code search the full Maya icon path for a file named:
+        //       MyRenderer_nifty_surface.png
```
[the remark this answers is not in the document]
JGamache-autodesk: That convention was the result of a discussion with our renderer team for a universal convention that also included:
- Native MaterialX nodes in the outliner
- Scalable SVG icons for LookdevX in both USD and MaterialX runtimes

I can see the usefulness of the `out_USD_` prefix for shorter names which might exist somewhere else on the Maya icon path. In the case of MaterialX nodes, we will see the characteristic `ND_` typical of MaterialX nodes, and in the case of Arnold nodes they will always be prefixed with the renderer name. I hope this will be sufficient to prevent name clashes.

We can always re-open the naming convention to add a single universal prefix if you think the risk of collisions is too great.

lib/mayaUsd/ufe/MayaUsdUIInfoHandler.cpp
JGamache-autodesk: This cache is keyed on the file name that was searched for. It will react to a change of `info:id` because that will change the file name to lookup. Only weakness is that it is not reset when the icon path changes so there might be issues if the outliner opens before the plugin gets loaded. Do you think I should reset the cache whenever a plugin gets loaded/unloaded?

lib/mayaUsd/ufe/MayaUsdUIInfoHandler.cpp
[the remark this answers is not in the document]
seando-adsk: Its not just about name clashes. The icons for Outliner are prefixed with "out_", and since these are coming via Ufe, we also add the ufe runtime name, so "out_USD_". Can you ask the renderer team if they are okay with not following the usual icon naming scheme in favor of a universal convention.

lib/mayaUsd/ufe/MayaUsdUIInfoHandler.cpp
seando-adsk: Ah right I got a little mixed up with the cache. The icon will change if the user changes `info:id` so I think that is good enough. I wouldn't worry about the icon path changes.

lib/mayaUsd/ufe/MayaUsdUIInfoHandler.cpp
```diff
@@ -88,5 +103,140 @@ UsdUfe::UsdUIInfoHandler::SupportedTypesMap MayaUsdUIInfoHandler::getSupportedIc
...
+        m_pluginUnloadCB = MSceneMessage::addStringArrayCallback(
+            MSceneMessage::kAfterPluginUnload, OnPluginStateChange);
+        m_beforeExitCB = MSceneMessage::addCallback(MSceneMessage::kMayaExiting, OnTerminateCache);
+    }
+
+    PXR_NS::ArDefaultResolverContext      m_iconContext;
+    std::unordered_map<std::string, bool> m_searchCache;
+    MCallbackId                           m_pluginLoadCB = 0;
+    MCallbackId                           m_pluginUnloadCB = 0;
+    MCallbackId                           m_beforeExitCB = 0;
```
seando-adsk, line 188: MayaUsd coding standard is just _ for member vars.

seando-adsk (changes requested):
Sorry one minor change related to variable naming.

seando-adsk (approved):

feldstj (approved):

JGamache-autodesk:
@seando-adsk ready for merge

PR #3495, opened 2023-12-04
EMSUSD-788 support relative cache to USD in anon layers

pierrebai-adsk:
If the layer does not have a file path when caching to USD, then remember the absolute path and the fact it must be made relative when the layer gets saved. Saving the layer automatically updates the path.

- Add a cache-to-USD test for relative cache in an anonymous layer.
- Exposed to Python the function to update the postponed relatve paths in a layer.


lib/mayaUsd/utils/utilFileSystem.h
```diff
@@ -165,6 +165,11 @@ void updatePostponedRelativePaths(
     const PXR_NS::SdfLayerHandle& layer,
     const std::string&            layerFileName);
 
+/*! \brief Turns the file paths marked through the call 'markPathAsPostponedRelative' to relative.
+ */
+MAYAUSD_CORE_PUBLIC
+void updatePostponedRelativePaths(const PXR_NS::SdfLayerHandle& layer);
```
samuelliu-adsk, line 171: I'm not sure about this but would it be better if we give the layerFileName a default value so we can just keep one method?

pierrebai-adsk: The default value that would get passed is taken from the layer handle, and we have to account for the possibility that the handle is null. It is simpler to have an overload.

samuelliu-adsk (approved):

PR #3499, opened 2023-12-06
EMSUSD-215 support custom display name for USD attributes

pierrebai-adsk:
Add support to have custom names for USD attributes to have nicer names for UI display purpose. We already had an algorithm to automatically make the names prettier, this adds support for user-defined attribute names. We now read the USD 'displayName' metadata on each attribute. If the metadata exists then we use its value as the display name.

The built-in function to make the name prettier has also been enhanced to remove the "xformOp:" prefix. Also, the prettyfying avoids generating two spaces in a row (if, for example, : was followed by _).

Modify the USD attribute-related classes:
- Add a displayName funtion to the UsdAttrbute class.
- Add a displayName funtion to the UsdAttributeHolder class.
- Add a displayName funtion to the UsdShaderAttributeHolder class.
- The displayName function in UsdAttributeHolder prettify the name.

Add support for display name in the Attribute Editor template:
- Add a AttributeCustomControl base class to handle display names.
- Make the default control creation use the display names.

Added a simple unit test and fix some attribute unit tests.


lib/mayaUsd/ufe/UsdAttribute.cpp
```diff
@@ -371,6 +371,15 @@ std::string UsdAttribute::name() const
     return _attrHolder->name();
 }
 
+#ifdef UFE_V4_FEATURES_AVAILABLE
```
AramAzhari-adsk, line 374: Shouldn't this be 

`#ifdef UFE_HAS_DISPLAY_NAME`

instead?

lib/mayaUsd/ufe/UsdAttribute.h
```diff
@@ -29,10 +29,18 @@
...
+    std::string displayName() const override { return UsdAttribute::_displayName(); }
+#else
+#define UFE_V5_ATTRIBUTE_OVERRIDES
+#endif
+
 #ifdef UFE_V4_FEATURES_AVAILABLE
 #define UFE_ATTRIBUTE_OVERRIDES                                                               \
     bool        hasValue() const override { return UsdAttribute::_hasValue(); }               \
     std::string name() const override { return UsdAttribute::_name(); }                       \
+    UFE_V5_ATTRIBUTE_OVERRIDES                                                                \
```
AramAzhari-adsk, line 43: This and the lines 32-37 could just be three lines, given that this isn't used anywhere else.

```
#ifdef UFE_HAS_DISPLAY_NAME
    std::string displayName() const override { return UsdAttribute::_displayName(); }
#endif
```

lib/mayaUsd/resources/ae/usdschemabase/attribute_custom_control.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def getNiceAttributeName(ufeAttr, attrName):
+    '''
+    Convert the attribute name into  nice name.
+    '''
+    # Note: the uiname metadata comes from LookdevX and was used for connections.
+    if hasattr(ufeAttr, 'displayName'):
+        attrName = ufeAttr.displayName
+    elif hasattr(ufeAttr, "hasMetadata") and ufeAttr.hasMetadata("uiname"):
```
AramAzhari-adsk, line 31: Can this not be addressed on LookdevX end, by setting the displayName instead?

It seems to me that this is client's business logic.

lib/usdUfe/utils/Utils.cpp
```diff
@@ -31,24 +31,28 @@ std::string sanitizeName(const std::string& name)
 
 std::string prettifyName(const std::string& name)
 {
-    std::string prettyName(1, std::toupper(name[0]));
-    size_t      nbChars = name.size();
-    bool        capitalizeNext = false;
-    for (size_t i = 1; i < nbChars; ++i) {
+    std::string prettyName;
+    prettyName.reserve(name.size());
```
AramAzhari-adsk, line 35: If the name is something like "NiceName" the string size will most certainly increase due to the added space in line 43.
Perhaps we could make this a name.size() + 5 to cover us for even a string ThatHasFiveCapitalizedWords.

Not a huge deal though, since there will only be a couple of reallocations at the end.

lib/mayaUsd/resources/ae/usdschemabase/attribute_custom_control.py
pierrebai-adsk: I believe the uiname is part of MaterialX specs. Either way, I did not want to break backward compatibility.

test/lib/testAttributeEditorTemplate.py
```diff
@@ -209,8 +209,8 @@ def testAECustomImageControl(self):
             self.assertIsNotNone(frameLayout, 'Could not find "Shader: Dot" frameLayout')
 
             # We should also have custom image control for 'Inputs In'.
```
AramAzhari-adsk, line 211: Should the comment still be 'Inputs In'?

lib/mayaUsd/ufe/UsdAttribute.h
pierrebai-adsk: NO, I need to declare a macro because I don;t think you can have a #ifdef in the middle of a #define.

AramAzhari-adsk (commented):
Overall looks good. Only a couple of minor questions regarding ifdefs

lib/mayaUsd/ufe/UsdAttribute.h
pierrebai-adsk: Honestly, I think this whole slew of ifdef made by LookdevX is terrible. I wish they'd done ifdef on THEIR codebase, by declaring a façade class or something instead of ugliying our code.

lib/usdUfe/utils/Utils.cpp
pierrebai-adsk: Yeah, I could over-reserve a bit. The old code did not reserve at all, bu yeah, adding a few extra to acount for typical numbers of spaces is easy.

lib/mayaUsd/ufe/UsdAttribute.cpp
pierrebai-adsk: The function exists based on V4/not V4 but has different name (leading underscore or not). It also needs to align with what is done in the header, which cannot have ifdef in the middle of a define.

lib/mayaUsd/ufe/UsdAttribute.h
AramAzhari-adsk: I see. I missed the define. Yes it's impossible to ifdef inside a define.

lib/mayaUsd/ufe/UsdAttribute.cpp
pierrebai-adsk: Basically, in one case there is a function with leading _ which contain the real call,, in the other, the one without leading _ contains the real call. It is all confusing.

AramAzhari-adsk (dismissed):

AramAzhari-adsk (approved):

lib/usdUfe/utils/Utils.cpp
samuelliu-adsk: Just dropping in a dumb question. Why do we have to reserve the size first? Couldn't we just declare an empty string and appending to it?

AramAzhari-adsk: @samuelliu-adsk since we're adding characters to prettyName over an iteration, C++ doesn't know ahead of time how many we need so it keeps allocating for every new character. We reserve an amount more than needed to do a single allocation for the entire operation. 

In short, allocating one large memory is more efficient than allocation many small memories.

pierrebai-adsk:
Only PF failure is the testVP2RenderDelegateDisplayLayers on Linux that is known to be flaky

PR #3504, opened 2023-12-08
EMSUSD-852 allow saving locked layers

pierrebai-adsk:
Relevant change is the removal of the call to isReadOnly(). I refactored the function to avoid nested if and complex conditional that are hard to understand. Also commented on what each condition aims to do.

### DIFF
--- a/lib/usd/ui/layerEditor/layerTreeItem.cpp
+++ b/lib/usd/ui/layerEditor/layerTreeItem.cpp
@@ -267,13 +267,21 @@ bool LayerTreeItem::isIncoming() const { return _isIncomingLayer; }
 
 bool LayerTreeItem::needsSaving() const
 {
-    if (_layer) {
-        if (!isSessionLayer() && !isReadOnly()
-            && (_isSharedStage || parentLayerItem() != nullptr)) {
-            return isDirty() || isAnonymous();
-        }
-    }
-    return false;
+    // If for any reason we don't hold a layer, then we cannot save it.
+    if (!_layer)
+        return false;
+
+    // Session layers are managed by Maya, not the Layer Editor,
+    // so their dirty state does not count.
+    if (isSessionLayer())
+        return false;
+
+    // The top layer of unshared layer are assumed to be managed
+    // somewhere else and do not get saved here. (Is this correct?)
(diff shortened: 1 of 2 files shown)


lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -267,13 +267,21 @@ bool LayerTreeItem::isIncoming() const { return _isIncomingLayer; }
 
 bool LayerTreeItem::needsSaving() const
 {
-    if (_layer) {
-        if (!isSessionLayer() && !isReadOnly()
```
pierrebai-adsk, line 271: The removal of the call to isReady() is the only real change, all the other is just avoiding the complex nesting of conditionals and multi-line conditions, allowing to document what each condition does.

lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -267,13 +267,21 @@ bool LayerTreeItem::isIncoming() const { return _isIncomingLayer; }
...
+
+    // Session layers are managed by Maya, not the Layer Editor,
+    // so their dirty state does not count.
+    if (isSessionLayer())
+        return false;
+
+    // The top layer of unshared layer are assumed to be managed
+    // somewhere else and do not get saved here. (Is this correct?)
+    if (!_isSharedStage && parentLayerItem() == nullptr)
+        return false;
```
pierrebai-adsk, line 282: Any remembrance if my comment about this particular check is correct?

lib/usd/ui/layerEditor/layerTreeItem.cpp
samuelliu-adsk: So if I understand it correctly, the needsSaving() function will return true even if `permissionToSave` is false, and that will be checked later when actually writing the layer

pierrebai-adsk: NO, isReadOnly checks permission to edit, not save. We currently do not have check for permission to save, but then again, USD itself will refuse to save, so we don't need to check it to prevent it. When we do the future work to have a UI for permission to save, we might had new verifications.

samuelliu-adsk (approved):

PR #3506, opened 2023-12-08
EMSUSD-807 - From LookdevX bring over UINodeGraphNode new virtual functions

seando-adsk:
#### EMSUSD-807 - From LookdevX bring over UINodeGraphNode new virtual functions
* Implemented new UINodeGraphNode methods and added unit tests.


lib/mayaUsd/ufe/UsdUINodeGraphNode.cpp
```diff
@@ -48,6 +48,38 @@ void UsdUINodeGraphNode::SetPosOrSizeCommand::executeImplementation()
     }
 }
 
+#ifdef UFE_UINODEGRAPHNODE_HAS_DISPLAYCOLOR
```
seando-adsk, line 51: Pretty much all the new code here came from Lookdex. I only modified the methods to take/return a Ufe::Color3f instead of std::string

JGamache-autodesk (approved):

PR #3508, opened 2023-12-11
EMSUSD-848 fix edit target after layer clear

pierrebai-adsk:
When clearing a layer that had sub-layers and one of those sub-layer was the edit target, we would no longer have a valid edit target. Fix this by putting the target on a valid layer, for now the root (since it cannot be muted and thus is valid). Also, remember the edit target so that undo can restore it.

A the same time, split the functions in the command class to have each function do only one thing.

### DIFF
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -19,6 +19,7 @@
 #include <mayaUsd/ufe/Global.h>
 #include <mayaUsd/utils/layerMuting.h>
 #include <mayaUsd/utils/query.h>
+#include <mayaUsd/utils/stageCache.h>
 #include <mayaUsd/utils/utilFileSystem.h>
 
 #include <usdUfe/ufe/Utils.h>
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -547,32 +548,116 @@ class BackupLayerBase : public BaseCmd
...
+                continue;
+
+            PXR_NS::UsdEditTarget target = weakStageAndTarget.second;
+            stage->SetEditTarget(target);
+        }
+    }
+
+    // Edit targets that were made invalid after the layer was cleared.
+    // The stages are kept with weak pointer to avoid forcing to stay valid.
+    using EditTargetBackups = std::map<PXR_NS::UsdStagePtr, PXR_NS::UsdEditTarget>;
```
samuelliu-adsk, line 658: Putting here as a reference: I talked to Pierre and the map is for the scenario that multiple stages share the same layer and we need to make sure when we restore edit target, we restore to the correct stage

samuelliu-adsk (approved):

PR #3513, opened 2023-12-12
EMSUSD-65 use wait cursor for payload commands

pierrebai-adsk:
- Add a USD UFE long-duration command to wrap other commands that may take a long time to run.
- Add two DCC-specific functions to start and stop the wait cursor.
- Add start/stop helper functions.
- Add a wait cursor class to automatically start and top the wait cursor.
- Wrap the load and unload payload commands in a long-duration command.
- Adjust existing commands that used a wait cursor to use the global helper class instead.

### DIFF
--- a/lib/mayaUsd/ufe/Global.cpp
+++ b/lib/mayaUsd/ufe/Global.cpp
@@ -78,6 +78,7 @@
 #include <usdUfe/ufe/UfeVersionCompat.h>
 #include <usdUfe/utils/editRouter.h>
 
+#include <maya/MGlobal.h>
 #include <maya/MSceneMessage.h>
 #include <ufe/hierarchyHandler.h>
 #include <ufe/pathString.h>
@@ -106,6 +107,10 @@ MCallbackId gExitingCbId = 0;
 
 // Subject singleton for observation of all USD stages.
 MayaUsd::ufe::MayaStagesSubject::RefPtr g_StagesSubject;
+
+void mayaStartWaitCursor() { MGlobal::executeCommand("waitCursor -state 1"); }
+
+void mayaStopWaitCursor() { MGlobal::executeCommand("waitCursor -state 0"); }
 
 } // namespace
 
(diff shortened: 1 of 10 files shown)


lib/mayaUsd/ufe/Global.cpp
```diff
@@ -107,6 +108,10 @@ MCallbackId gExitingCbId = 0;
 // Subject singleton for observation of all USD stages.
 MayaUsd::ufe::MayaStagesSubject::RefPtr g_StagesSubject;
 
+void mayaStartWaitCursor() { MGlobal::executeCommand("waitCursor -state 1"); }
```
AramAzhari-adsk, line 111: This seems like something that should be in mayaUsd/Utils and not Ufe, as this isn't manipulating or using anything Ufe and I think it is going to be useful for the rest of MayaUsd.

lib/mayaUsd/ufe/MayaUsdContextOps.cpp
```diff
@@ -102,12 +102,6 @@ const constexpr char  kClearAllRefsOrPayloadsLabel[] = "Clear All USD References
 const constexpr char  kClearAllRefsOrPayloadsItem[] = "ClearAllReferencesOrPayloads";
 
 //! \brief Change the cursor to wait state on construction and restore it on destruction.
```
AramAzhari-adsk, line 104: The code is removed but its comment is still remaining.

lib/mayaUsd/ufe/Global.cpp
pierrebai-adsk: Other code can use the usdUfe API, not this implementation. Particularly since the code is usdUfe manages the recursivity count and provide a utility WaitCursor class.

AramAzhari-adsk: mayaUsd has access to usdUfe but probably not the other way around. 

I also noticed that you've removed the struct in MayaUsdContextOps.

AramAzhari-adsk (approved):

pierrebai-adsk:
Only PF failure is a singel test on a single platform due to a USDC file remaining in the tmp folder.

lib/mayaUsd/ufe/Global.cpp
seando-adsk: This particular code cannot be in UsdUfe since it uses Maya. UsdUfe does not (and cannot) depend on Maya.

PR #3518, opened 2023-12-14
EMSUSD-671 fix reloading scene with layers saved in the Maya scene

pierrebai-adsk:
When reloading a scene that contained layers serialized directly in the scene, sub-layers would be lost because they were anonymous layers but on reload that status was lost and the sub-layer would be lost whenever the stage got recomputed.

Fix this by remembering that the sub-layer are anonymous on reload.

Add unit test to verify lost layer fix

- Add comments to the layer manager serialization test and fix the layer being targeted.
- Refactor common code in proxy shape saving tests.
- Add more checks in the proxy shape saving test to verify that the contents are correct after reload and after recompute.

### DIFF
--- a/test/lib/mayaUsd/nodes/testLayerManagerSerialization.py
+++ b/test/lib/mayaUsd/nodes/testLayerManagerSerialization.py
@@ -56,6 +56,10 @@ class testLayerManagerSerialization(unittest.TestCase):
         cmds.file(rename=self._tempMayaFile)
 
     def copyTestFilesAndMakeEdits(self):
+        '''
+        Copy an existing Maya scene that contains a stage and a few layer,
+        creates a few USD prim in the root, session and 1_1 layers.
+        '''
         self._currentTestDir = tempfile.mkdtemp(prefix='LayerManagerTest')
         fromDirectory = os.path.join(
             self._inputPath, 'LayerManagerSerializationTest')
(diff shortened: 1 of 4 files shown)


test/lib/mayaUsd/nodes/testLayerManagerSerialization.py
```diff
@@ -79,13 +83,20 @@ def copyTestFilesAndMakeEdits(self):
...
+        #            4: 2
+        #            5: 2_1
         self.assertEqual(6, len(stack))
 
         stage.SetEditTarget(stage.GetRootLayer())
         newPrimPath = "/ChangeInRoot"
         stage.DefinePrim(newPrimPath, "xform")
 
-        stage.SetEditTarget(stack[2])
+        stage.SetEditTarget(stack[3])
```
pierrebai-adsk, line 99: As noted above, the 1_1 layer is layer #3, not #2.

test/lib/mayaUsd/nodes/testProxyShapeBase.py
```diff
@@ -396,6 +396,46 @@ def testShareStagePreserveTCPS(self):
         self.assertEqual(stage.GetTimeCodesPerSecond(), tcps)
         self.assertEqual(stage.GetRootLayer().timeCodesPerSecond, tcps)
```
pierrebai-adsk, line 398: The functions below were declared repeatedly, inline in multiple tests. Declare them here and document them instead.

test/lib/mayaUsd/nodes/testProxyShapeBase.py
```diff
@@ -496,8 +511,24 @@ def verifyPrim():
...
 
-            verifyPrim()
+            self._verifyPrim(False)
+            self._verifySubLayer()
+
+            # Change shared status and verify again. Changing the shared flag recomputes
+            # the stage and its layer, which is what we really want to test here. So we
+            # change an input attribute that we know will recompute the layers.
+            cmds.setAttr('{}.{}'.format(proxyShapePath,"shareStage"), False)
+
```
pierrebai-adsk, line 528: These extra checks, after reload + modify stage are what were triggering the layer lost bug.

pierrebai-adsk:
@AramAzhari-adsk bump

AramAzhari-adsk (approved):

PR #3521, opened 2023-12-14
Update usage of the Hydra API

dj-mcg:
Deprecate UsdGeomSubset::ValidateSubsets in favor of ValidateFamily

Update UsdGeomSubset::GetUnassignedIndices to accept a prim and family name instead of a list of subsets

### DIFF
--- a/lib/mayaUsd/fileio/translators/translatorMaterial.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorMaterial.cpp
@@ -379,16 +379,28 @@ bool UsdMayaTranslatorMaterial::AssignMaterial(
 
         std::string reasonWhyNotPartition;
 
-        const bool validPartition = UsdGeomSubset::ValidateSubsets(
-            faceSubsets, faceCount, UsdGeomTokens->partition, &reasonWhyNotPartition);
+        UsdGeomImageable geom(primSchema.GetPrim());
+        if (UsdGeomSubset::GetFamilyType(geom, UsdShadeTokens->materialBind)
+            != UsdGeomTokens->partition) {
+            TF_WARN(
+                "The family type of family material bind "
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/fileio/translators/translatorMaterial.cpp
```diff
@@ -379,16 +379,28 @@ bool UsdMayaTranslatorMaterial::AssignMaterial(
...
+        UsdGeomImageable geom(primSchema.GetPrim());
+        if (UsdGeomSubset::GetFamilyType(geom, UsdShadeTokens->materialBind)
+            != UsdGeomTokens->partition) {
+            TF_WARN(
+                "The family type of family material bind "
+                "on <%s> is not set as a partition.",
+                primSchema.GetPath().GetText());
+        }
+        const bool validPartition = UsdGeomSubset::ValidateFamily(
+            geom, UsdGeomTokens->face, UsdShadeTokens->materialBind, &reasonWhyNotPartition);
```
pierrebai-adsk, line 391: Is the first check above not redundant with the reasonWhyNotPartition? Or are the two checks necessary?

pierrebai-adsk (approved):
Looks good, but IDK if we are not double-checking the validity of the partition. Probably not a problem, at worst we give more error messages to the user than necessary.

lib/mayaUsd/fileio/translators/translatorMaterial.cpp
dj-mcg: It looks like the first check - verifying the FamilyType is partition - is a case that is specific to this function and one which UsdGeomSubset::ValidateFamily doesn't check. I can check with the USD team whether it makes sense to move this to a more general purpose area, but for now this should not be causing redundant warnings.

PR #3522, opened 2023-12-14
Pr/deprecate boost optional for sdf copy spec

dj-mcg:
The USD team is phasing out usage of boost::optional. This updates the callsites for the upcoming switch in SdfCopySpec


pierrebai-adsk (approved):
LGTM

PR #3523, opened 2023-12-15
EMSUSD-671 Remove the load payloads proxy shape attribute

pierrebai-adsk:
Instead, we will always use the existing payload rules dynamic attribute.

- Remove the "loadPayloads" attribute from the proxy shape.
- Add Python binding to new functions to set and retrieve if all payloads are loaded or not.
- Split-up existing load-rules helper functions to give more fined-grained access to the saved load rules in order to support the Python API.
- Use the new Python helper functions when creating or loading a stage.
- Adjust one unit test.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -474,18 +474,18 @@ MStatus MayaUsdProxyShapeBase::initialize()
     retValue = attributeAffects(primPathAttr, outStageCacheIdAttr);
     CHECK_MSTATUS_AND_RETURN_IT(retValue);
 
-    retValue = attributeAffects(loadPayloadsAttr, inStageDataCachedAttr);
-    CHECK_MSTATUS_AND_RETURN_IT(retValue);
-    retValue = attributeAffects(loadPayloadsAttr, outStageDataAttr);
-    CHECK_MSTATUS_AND_RETURN_IT(retValue);
-    retValue = attributeAffects(loadPayloadsAttr, outStageCacheIdAttr);
-    CHECK_MSTATUS_AND_RETURN_IT(retValue);
-
     retValue = attributeAffects(shareStageAttr, inStageDataCachedAttr);
     CHECK_MSTATUS_AND_RETURN_IT(retValue);
     retValue = attributeAffects(shareStageAttr, outStageDataAttr);
     CHECK_MSTATUS_AND_RETURN_IT(retValue);
     retValue = attributeAffects(shareStageAttr, outStageCacheIdAttr);
+    CHECK_MSTATUS_AND_RETURN_IT(retValue);
+
+    retValue = attributeAffects(loadPayloadsAttr, inStageDataCachedAttr);
+    CHECK_MSTATUS_AND_RETURN_IT(retValue);
+    retValue = attributeAffects(loadPayloadsAttr, outStageDataAttr);
+    CHECK_MSTATUS_AND_RETURN_IT(retValue);
+    retValue = attributeAffects(loadPayloadsAttr, outStageCacheIdAttr);
     CHECK_MSTATUS_AND_RETURN_IT(retValue);
 
     retValue = attributeAffects(inStageDataAttr, inStageDataCachedAttr);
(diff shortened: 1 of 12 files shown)


lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -279,15 +278,6 @@ MStatus MayaUsdProxyShapeBase::initialize()
...
 
-    loadPayloadsAttr
-        = numericAttrFn.create("loadPayloads", "lpl", MFnNumericData::kBoolean, 1.0, &retValue);
-    CHECK_MSTATUS_AND_RETURN_IT(retValue);
-    numericAttrFn.setKeyable(true);
-    numericAttrFn.setReadable(false);
-    numericAttrFn.setAffectsAppearance(true);
-    retValue = addAttribute(loadPayloadsAttr);
-    CHECK_MSTATUS_AND_RETURN_IT(retValue);
-
```
pierrebai-adsk, line 290: In this file, we're removing the load-payloads proxy shape attribute.

lib/mayaUsd/python/wrapLoadRules.cpp
```diff
@@ -0,0 +1,55 @@
+//
```
pierrebai-adsk, line 1: Giving access to some load rules helpers in Python.

lib/mayaUsd/utils/loadRules.h
```diff
@@ -43,6 +44,21 @@ MStatus copyLoadRulesFromAttribute(
     const PXR_NS::MayaUsdProxyShapeBase& proxyShape,
     PXR_NS::UsdStage&                    stage);
```
pierrebai-adsk, line 46: Refactoring some load-rules helpers to be usable from Python.

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -183,7 +183,7 @@ def ProxyShapeFilePathChanged(filePathAttr, newFilePath=None):
             #       in that case.
             primPath = cmds.getAttr(stageName+'.primPath') or ''
             excludedPrimPaths = cmds.getAttr(stageName+'.excludePrimPaths') or ''
-            loadPayloads = cmds.getAttr(stageName+'.loadPayloads') or 0
```
pierrebai-adsk, line 186: Using the new Python load rules helpers to handle the load-payload checkbox in the attribute editor.

plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
```diff
@@ -23,8 +23,8 @@ proc setOptionVars(int $forceFactorySettings)
         optionVar -stringValue stageFromFile_excludePrimPath "";
     }
```
pierrebai-adsk, line 25: Using the new Python load-rules helpers in the UI used to create a new stage.

test/lib/mayaUsd/nodes/testProxyShapeBase.py
```diff
@@ -574,10 +574,10 @@ def testUnsavedStagePreserveRootLayerWhenUpdated(self):
         # verify that the prim exists.
         self._verifyPrim()
 
-        # Set an attribute on the proxy shape. Here we set the loadPayloads.
+        # Set an attribute on the proxy shape. Here we set the shareStage.
         # It was already set, this only triggers a Maya node recompute.
         _, proxyShapePath = self._getStage()
-        cmds.setAttr('{}.{}'.format(proxyShapePath,"loadPayloads"), True)
```
pierrebai-adsk, line 580: This attribute no longer exists, so this test needs to be changed.

AramAzhari-adsk (dismissed):

lib/mayaUsd/nodes/proxyShapeBase.cpp
seando-adsk: What happens with a saved Maya scene where this attribute was set and saved in the Maya file?

pierrebai-adsk: Maya does not allow removing attribute? I can put it back in the node definition but make it do nothing instead. Is that how legacy attribute must be handled?

pierrebai-adsk:
@AramAzhari-adsk @seando-adsk for a refresher: the thing I changed after approval was put back the attribute to avoid warnings when loading old scenes.

AramAzhari-adsk (approved):

seando-adsk (approved):

PR #3531, opened 2023-12-18
LOOKDEVX-2333 - Use uimin if uisoftmin is missing

JGamache-autodesk:
As noticed by Jannik Frohne the slider for subsurface_anisotropy was not going from -1 to +1 as suggested by its declaration of uimin="-1.0" uimax="1.0".

If we have hard min and max value and no soft min max, use the hard values.


seando-adsk (approved):

JGamache-autodesk:
@seando-adsk Ready for merge.

frohnej-adsk (approved):

PR #3534, opened 2023-12-21
HYDRA-764 : Fix instanceable prims not rendered in Hydra

debloip-adsk:
This PR fixes instanceable prims not being rendered in MayaHydra. The issue was due to the `MayaUsdProxyShapeSceneIndex::GetChildPrimPaths()` method not using its input scene index and instead skipping over to directly using the stage scene index, which made it so that the work done by the UsdImagingNiPrototypePropagatingSceneIndex was not carried over to the following scene indices. For reference, the scene index hierarchy looks something like this : 
```
MayaUsdProxyShapeSceneIndex
|__...
   |__UsdImagingNiPrototypePropagatingSceneIndex
      |__...
         |__UsdImagingStageSceneIndex
```

### DIFF
--- a/lib/mayaUsd/sceneIndex/proxyShapeSceneIndexPlugin.cpp
+++ b/lib/mayaUsd/sceneIndex/proxyShapeSceneIndexPlugin.cpp
@@ -397,7 +397,7 @@ HdSceneIndexPrim MayaUsdProxyShapeSceneIndex::GetPrim(const SdfPath& primPath) c
 
 SdfPathVector MayaUsdProxyShapeSceneIndex::GetChildPrimPaths(const SdfPath& primPath) const
 {
-    return _usdImagingStageSceneIndex->GetChildPrimPaths(primPath);
+    return _GetInputSceneIndex()->GetChildPrimPaths(primPath);
 }
 
 void MayaUsdProxyShapeSceneIndex::_PrimsAdded(


lib/mayaUsd/sceneIndex/proxyShapeSceneIndexPlugin.cpp
```diff
@@ -397,7 +397,7 @@ HdSceneIndexPrim MayaUsdProxyShapeSceneIndex::GetPrim(const SdfPath& primPath) c
 
 SdfPathVector MayaUsdProxyShapeSceneIndex::GetChildPrimPaths(const SdfPath& primPath) const
 {
-    return _usdImagingStageSceneIndex->GetChildPrimPaths(primPath);
+    return _GetInputSceneIndex()->GetChildPrimPaths(primPath);
```
ppt-adsk, line 400: Great catch!  Can we provide an automated test that will prevent regressions?

debloip-adsk: Yep, working on it, though it'll be on the MayaHydra side. I've been stripping down the ALab scene to get a minimal repro scene.

ppt-adsk: This is why we need to pull all the rendering-related USD code into maya-hydra...

ppt-adsk (approved):

seando-adsk:
@debloip-adsk We still use the ecp user to start our preflights. I've started it for you. Please add back the ready-for-merge label once you have a successful preflight.

debloip-adsk:
FYI @lanierd-adsk @roopavr-adsk @lilike-adsk

PR #3536, opened 2023-12-22
EMSUSD-181: Tooltips are not formatted correctly

seando-adsk:
#### EMSUSD-181: Tooltips are not formatted correctly
* Use tooltip cleaning method similiar to UsdView to remove leading whitespace from multi-line doc strings.
* Added tooltips to all attributes.

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
@@ -15,6 +15,7 @@
 
 from .custom_image_control import customImageControlCreator
 from .attribute_custom_control import getNiceAttributeName
+from .attribute_custom_control import cleanAndFormatTooltip
 from .attribute_custom_control import AttributeCustomControl
 
 import collections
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -433,7 +434,7 @@ def arrayCustomControlCreator(aeTemplate, c):
 def defaultControlCreator(aeTemplate, c):
     ufeAttr = aeTemplate.attrS.attribute(c)
     uiLabel = getNiceAttributeName(ufeAttr, c) if aeTemplate.useNiceName else c
-    cmds.editorTemplate(addControl=[c], label=uiLabel)
+    cmds.editorTemplate(addControl=[c], label=uiLabel, annotation=cleanAndFormatTooltip(ufeAttr.getDocumentation()))
```
seando-adsk, line 437: When debugging this I realized that we were only adding tooltips to the array and connected attributes. After discussion with Nat I added them to all the attributes.

lib/mayaUsd/resources/ae/usdschemabase/attribute_custom_control.py
```diff
@@ -32,6 +32,14 @@ def getNiceAttributeName(ufeAttr, attrName):
...
     return mayaUsd.lib.Util.prettifyName(attrName)
 
+def cleanAndFormatTooltip(s):
+    # Remove leading/trailing whitespace and replace newlines.
+    lines = s.splitlines()
+    stripped = [line.strip() for line in lines]
+    cleaned = '<br>'.join(stripped)
+
+    # Don't allow the tooltip to word-wrap.
+    return "<p style='white-space:pre'>" + cleaned + '</p>'
```
seando-adsk, line 42: splitlines() returns list splitting string at line breaks. Then create new list by removing spaces at beginning and end of strings. Then create one string using <br> instead of newline. Finally don't allow tooltip to wrap, as it contains breaks.

I think this is better than what UsdView is doing since it also removes the single leading space that subsequent lines have.

pierrebai-adsk (approved):
LGTM

seando-adsk:
The PF failure was a testing glitch and not related to my changes.

PR #3537, opened 2023-12-22
EMSUSD-804 duplicate-to-USD support relationship targets

pierrebai-adsk:
The goal is to solve problems with duplicating Maya node to USD when using the Arnold configuration. The main problem is that Arnold authors the materials used by the prim corresponding to the duplicated Maya node outside of the USD prim. The duplicate code was expecting to copy a single prim and its children.

To solve this, we add code to follow relationships and connections inside the new prim and also copy the targets of these relationships and connections.

Improve the traverseLayer function:
- Avoid infinite recursion by tracking which path have already been traversed.
- Correctly traverse relationships and not stop at first non-prim item.
- It worked for the original use case (prim updater) because that use case always pruned the traversal under each prim, avoiding traversal of non-prim "prim spec". That was hiding the fact the function did not work as expected.

Improve some helper code:
- Add uniqueChildPath helper function to make a SdfPath unique in its parent.
- Make the finalize function of ReplicateExtrasToUSD handle more than one renamed paths.

Support copying relationship and connection targets:
- Make the duplicate-to-USD function also copy the prims that are targeted by the copied prim in its relationships and connections.
- Do this recursively, so that if the targeted prim also targets other prims, those will be included.
- Adjust the targets to take into account renamed prims.
- Keep track of which prims have already been copied to avoid copying them again.

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -1145,7 +1145,7 @@ bool PrimUpdaterManager::mergeToUsd(
     }
     progressBar.advance();
 
-    context._pushExtras.finalize(MayaUsd::ufe::stagePath(context.GetUsdStage()));
+    context._pushExtras.finalize(MayaUsd::ufe::stagePath(context.GetUsdStage()), {});
     progressBar.advance();
 
     discardPullSetIfEmpty();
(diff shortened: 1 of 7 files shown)


pierrebai-adsk:
Note: the problem can only be reproduced with the Arnold plugin, so I was not able to write a unit test for it.

AramAzhari-adsk (approved):

PR #3540, opened 2024-01-02
Add explicit boost/optional include

dj-mcg:
This was implicitly included via sdf/allowed.h, but has been removed from that header.


dj-mcg:
Hey folks - looks like there are some tests segfaulting on testing early versions on Windows. Is this a preflight issue or is my new include causing this? I'm not sure how best to proceed/debug since I won't be able to reproduce this problem locally...

Thanks!

seando-adsk (approved):

seando-adsk:
Preflight failure was a test glitch not related to these changes. I'll merge.

PR #3544, opened 2024-01-09
EMSUSD-840 expand and collapse all layer items

pierrebai-adsk:
When holding shift while clicking the expand or collapse button, expand or collapse all items in the layer editor.

### DIFF
--- a/lib/usd/ui/layerEditor/layerTreeView.cpp
+++ b/lib/usd/ui/layerEditor/layerTreeView.cpp
@@ -101,6 +101,8 @@ LayerTreeView::LayerTreeView(SessionState* in_sessionState, QWidget* in_parent)
 
     // signals
     connect(this, &QAbstractItemView::doubleClicked, this, &LayerTreeView::onItemDoubleClicked);
+    connect(this, &QTreeView::expanded, this, &LayerTreeView::onExpanded);
+    connect(this, &QTreeView::collapsed, this, &LayerTreeView::onCollapsed);
 
     // renderSetuplike API
     auto actionButtons = LayerTreeItem::actionButtonsDefinition();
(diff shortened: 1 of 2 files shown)


lib/usd/ui/layerEditor/layerTreeView.cpp
```diff
@@ -136,6 +138,61 @@ void LayerTreeView::onItemDoubleClicked(const QModelIndex& index)
     }
 }
 
+bool LayerTreeView::shouldExpandOrCollapseAll() const
+{
+    return ((QGuiApplication::keyboardModifiers() & Qt::ShiftModifier) != 0);
```
AramAzhari-adsk, line 143: if I understand correctly this is a Shift + Click ?

pierrebai-adsk: Technically this only checks shift but it is a private function called while handling clicks on the expansion button.

AramAzhari-adsk: I think a documentation on Collapse All / Expand All would help to clarify this. Also a plus if the code mentions it.

AramAzhari-adsk (approved):

lib/usd/ui/layerEditor/layerTreeView.cpp
seando-adsk, line 143: I'm not sure we should be calling Qt directly here to get the keyboard modifiers. Instead I would suggest using the maya command `getModifiers`

lib/usd/ui/layerEditor/layerTreeView.cpp
```diff
@@ -136,6 +138,61 @@ void LayerTreeView::onItemDoubleClicked(const QModelIndex& index)
...
+
+    // Recursively expand each child node.
+    const int count = index.model()->rowCount(index);
+    for (int i = 0; i < count; i++) {
+        const QModelIndex& child = index.model()->index(i, 0, index);
+        expandChildren(child);
+    }
+
+    expand(index);
+#endif
```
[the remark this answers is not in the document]
pierrebai-adsk: Let me recheck, I thought I had seen 2022 using 5.12.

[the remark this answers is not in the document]
pierrebai-adsk: I was wrong, it uses 5.15.2, i'll remove the check

lib/usd/ui/layerEditor/layerTreeView.cpp
pierrebai-adsk: Is it a C++ command? going through MEL would seem inefficient. I guess this is a mere mouse click, so efficiency is not really a concern.

pierrebai-adsk: Do we still plan to share the Layer Editor with 3dsMax? Going through a Maya command would make porting harder.

AramAzhari-adsk: I think the keyboard modifiers are passed on in a mouse click event (ex Line 448 ). Perhaps it can be captured then ?

AramAzhari-adsk (approved):

seando-adsk (approved):

lib/usd/ui/layerEditor/layerTreeView.cpp
seando-adsk: I didn't think about the porting, but the LayerEditor already uses other stuff from MGlobal, so we'll deal with it all later.

PR #3550, opened 2024-01-11
EMSUSD-934 - Fixes a bug with UsdAttributeEnumString that have token

feldstj:
Fixes a bug with UsdAttributeEnumString that have token values. We discovered a case where this was possible and needed to handle it.


JGamache-autodesk (approved):
Similar to what UsdAttributeFilename does at line 505.

seando-adsk (approved):

PR #3552, opened 2024-01-12
LOOKDEVX-2354 - Fix the transform nodes crash

JGamache-autodesk:
A transform node with empty space values will cause the shadergen to crash. This has been fixed in future MaterialX 1.38.9 by introducing ShaderPort::getValueString() in https://github.com/AcademySoftwareFoundation/MaterialX/pull/1560

We have backported that fix in local copies of the transform node C++ implementations in order to prevent crashing with the current versions of MaterialX.

### DIFF
--- a/lib/mayaUsd/render/MaterialXGenOgsXml/Nodes/MayaTransformVectorNodeGlsl.cpp
+++ b/lib/mayaUsd/render/MaterialXGenOgsXml/Nodes/MayaTransformVectorNodeGlsl.cpp
@@ -1,0 +1,98 @@
+//
+// Copyright Contributors to the MaterialX Project
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "MayaTransformVectorNodeGlsl.h"
+
+#include <MaterialXGenShader/Shader.h>
+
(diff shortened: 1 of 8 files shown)


lib/mayaUsd/render/MaterialXGenOgsXml/Nodes/MayaTransformVectorNodeGlsl.cpp
```diff
@@ -0,0 +1,98 @@
...
+    const ShaderInput* in,
+    GenContext&        context) const
+{
+    const ShaderGenerator& shadergen = context.getShaderGenerator();
+    return "vec4(" + shadergen.getUpstreamResult(in, context) + ", 0.0)";
+}
+
+string MayaTransformVectorNodeGlsl::getPortValueString(const ShaderInput* input) const
+{
+    if (!input || !input->getValue()) {
```
JGamache-autodesk, line 92: This is where the crash gets fixed. The previous code was not checking for `!input->getValue()`.

vlasovi (approved):
Looks good

JGamache-autodesk:
@seando-adsk Ready for merge.

PR #3553, opened 2024-01-12
EMSUSD-717: USD Prefs: "Use Display Color" preferences are maintained when user cancels change

seando-adsk:
#### EMSUSD-717: USD Preferences: "Use Display Color" preferences are maintained when user cancels change

* Fix cancel case and optimize when to call 'ogs -reset'

### DIFF
--- a/plugin/adsk/scripts/mayaUsd_preferenceTab.mel
+++ b/plugin/adsk/scripts/mayaUsd_preferenceTab.mel
@@ -94,6 +94,10 @@ global proc mayaUsd_PrefUntexturedModeChanged()
 	if (`radioButtonGrp -exists usdPrefUntexturedMode`) {
 		int $sel = `radioButtonGrp -q -select usdPrefUntexturedMode`;
 		optionVar -iv mayaUsd_ShowDisplayColorTextureOff ($sel == 2);
+        
+        // Set a temporary optionVar saying that we have changed the untextured mode
+        // so in case user cancels, we can do reset again.
+        optionVar -transient -iv mayaUsd_PrefUntexturedModeChanged 1;
 		ogs -reset;
 	}
 }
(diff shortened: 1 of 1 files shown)


plugin/adsk/scripts/mayaUsd_preferenceTab.mel
```diff
@@ -94,6 +94,10 @@ global proc mayaUsd_PrefUntexturedModeChanged()
 	if (`radioButtonGrp -exists usdPrefUntexturedMode`) {
 		int $sel = `radioButtonGrp -q -select usdPrefUntexturedMode`;
 		optionVar -iv mayaUsd_ShowDisplayColorTextureOff ($sel == 2);
+        
+        // Set a temporary optionVar saying that we have changed the untextured mode
+        // so in case user cancels, we can do reset again.
+        optionVar -transient -iv mayaUsd_PrefUntexturedModeChanged 1;
```
seando-adsk, line 100: transient optionvars are not saved when exiting Maya. ogs -reset is not a cheap operation so we should not just call it unless needed (such as when opening the pref dialog we don't need to call it). Use this optionvar to keep track of when the untextured mode pref changes and only call ogs -reset when needed.

plugin/adsk/scripts/mayaUsd_preferenceTab.mel
```diff
@@ -392,10 +408,22 @@ global proc mayaUsd_PrefHoldState(string $mode)
...
 	prefsHoldOptionVars($prefOptionVars, $mode); 
 
-	if ($mode == "remove") {
-		// Finalize preference changes into MayaUsd.
+    if ($mode == "save") {
+        // Preferences dialog was opened, make sure we don't have this transient optionVar.
+        optionVar -remove mayaUsd_PrefUntexturedModeChanged;
+    }
+
+	if ($mode == "restore") {
```
samuelliu-adsk, line 416: Indentation seems off here..

plugin/adsk/scripts/mayaUsd_preferenceTab.mel
```diff
@@ -392,10 +408,22 @@ global proc mayaUsd_PrefHoldState(string $mode)
...
+        if (`optionVar -q mayaUsd_PrefUntexturedModeChanged`) {
+            // Need to refresh the viewport for display mayaUsd_ShowDisplayColorTextureOff
+            ogs -reset;
```
samuelliu-adsk, line 419: Should we remove the `mayaUsd_PrefUntexturedModeChanged` optionVar once the viewport is reset?

plugin/adsk/scripts/mayaUsd_preferenceTab.mel
seando-adsk: Sorry it is a mix of tabs/spaces. I'll fix it.

samuelliu-adsk (approved):

PR #3554, opened 2024-01-13
Apply root-to-prim transforms

dj-mcg:
When computing bounding boxes and transforms for ProxyShape nodes, apply the root-to-prim transforms defined within the referenced Usd stage for the specified prim for imaging and selection

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -1547,8 +1547,10 @@ MBoundingBox MayaUsdProxyShapeBase::boundingBox() const
     const TfToken purpose3 = drawProxyPurpose ? UsdGeomTokens->proxy : TfToken();
     const TfToken purpose4 = drawGuidePurpose ? UsdGeomTokens->guide : TfToken();
 
+    // Compute the bound in "Usd World" space. This will apply the transform the
+    // referenced prim may have relative to the root of its Usd scene
     GfBBox3d allBox
-        = imageablePrim.ComputeUntransformedBound(currTime, purpose1, purpose2, purpose3, purpose4);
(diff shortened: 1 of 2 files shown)


JGamache-autodesk:
@dj-mcg  Code looks fine to me, but the questions are: what bug are you fixing? Can this bug be expressed as a unit test that would fail without this fix? Is there a possibility that some other client codebase is relying on the current behavior?

dj-mcg:
> @dj-mcg Code looks fine to me, but the questions are: what bug are you fixing? Can this bug be expressed as a unit test that would fail without this fix? Is there a possibility that some other client codebase is relying on the current behavior?

This fixes a bug with our Scene Assembly based workflows. Users have the option to bring in "reference" from full shots by pointing the Scene Assembly at a full shot's USD file, then specifying a specific Prim Path (and frame number) to image in their scene. This is primarily used for "camera dressing" in case set changes have to be made to suit the context of the shots.

The issue was that both the Bounding Boxes and the Transforms of these scene assemblies were not honoring the transform for the prim that was authored within the shot USD file, so viewport selection wouldn't work (fixed via the boundingBox change) and manipulation of the object would cause it to snap to the Usd scene origin (fixed via the transform change).

This broke years ago so clearly there isn't currently sufficient test coverage. I'll think about how best to test this code short of actually having to simulate user input in an interactive test to do viewport selection

JGamache-autodesk (approved):

ppt-adsk (approved):

dj-mcg:
Hi folks, just checking in to see if we're still waiting on something to merge this. Please let me know if there's anything I need to do on my side, thanks!

seando-adsk:
@dj-mcg Sorry Dan I forgot about this one. I've just relaunched the preflight since it was a bit old. Once it passes, I'll merge this one.
Sean

PR #3555, opened 2024-01-15
EMSUSD-917 : Add a stable library as a subset of MayaUSD LIbrary call…

lanierd-adsk:
…ed mayaUsdAPI Library for maya-hydra, Bifrost etc.


lanierd-adsk:
FYI : @ppt-adsk , @lilike-adsk , @roopavr-adsk , @debloip-adsk

seando-adsk (approved):
Great work David!

lib/mayaUsdAPI/proxyStage.cpp
```diff
@@ -0,0 +1,80 @@
...
+ProxyStage::ProxyStage(const ProxyStage& other)
+{
+    _imp = UniqueProxyStageImp(new ProxyStageImp(other._imp->_proxyStageProvider));
+}
+
+ProxyStage::~ProxyStage() = default;
+
+void ProxyStageImpDeleter::operator()(ProxyStageImp* ptr) const { delete ptr; }
+
+bool ProxyStage::isValid() const { return (nullptr != _imp->_proxyStageProvider); }
```
ppt-adsk, line 67: Again, if you throw in the constructor, you cannot have a zombie object, and isValid() is no longer needed.

lib/mayaUsdAPI/proxyStage.h
```diff
@@ -0,0 +1,94 @@
...
+            auto stage = proxyStage.getUsdStage();
+            auto time = proxyStage.getTime();
+        }
+    }
+    \endcode
+*/
+class MAYAUSD_API_PUBLIC ProxyStage
+{
+public:
+    /*! Constructor from a MObject which is a the MObject of a subclass of mayaUsd
```
ppt-adsk, line 57:
```suggestion
    /*! Constructor from an MObject whose MPxNode must be a subclass of mayaUsd
```

lib/mayaUsdAPI/python/wrapProxyShapeNotice.cpp
```diff
@@ -0,0 +1,122 @@
...
+//
+
+#include <mayaUsdAPI/proxyShapeNotice.h>
+
+#include <pxr/base/tf/instantiateSingleton.h>
+#include <pxr/base/tf/instantiateType.h>
+
+#include <boost/python/def.hpp>
+
+PXR_NAMESPACE_OPEN_SCOPE
```
ppt-adsk, line 24: I know this is an implementation-only file, but it feels awkward that this is in the Pixar namespace.

lib/mayaUsdAPI/python/wrapProxyShapeNotice.cpp
```diff
@@ -0,0 +1,122 @@
...
+            = TfNotice::Register(ptr, &StageNoticesListener::_StageObjectsChanged);
+        TF_VERIFY(
+            _stageSetKey.IsValid() && _stageInvalidateKey.IsValid()
+            && _stageObjectsChangedKey.IsValid());
+    }
+
+    ~StageNoticesListener() { _Revoke(); }
+
+    // StageSet notice
+    bool StageSetHasBeenCalled() const { return _stageSetCalled; }
```
ppt-adsk, line 46: I know it's an implementation file, but StageSetHasBeenCalled() is not much clearer to me than just StageSet(), and is longer.  It's a minor point, admittedly.

lib/mayaUsdAPI/python/wrapProxyShapeNotice.cpp
```diff
@@ -0,0 +1,122 @@
...
+    void ResetStageSet() { _stageSetCalled = false; }
+
+    // StageInvalidate notice
+    bool StageInvalidateHasBeenCalled() const { return _stageInvalidateCalled; }
```
ppt-adsk, line 50: Similar to above, I would just go with StageInvalidated().

lib/mayaUsdAPI/python/wrapProxyShapeNotice.cpp
```diff
@@ -0,0 +1,122 @@
...
+    void ResetStageInvalidate() { _stageInvalidateCalled = false; }
+
+    // StageObjectsChanged notice
+    bool StageObjectsChangedHasBeenCalled() const { return _stageObjectsChangedCalled; }
```
ppt-adsk, line 54:
```suggestion
    bool StageObjectsChanged() const { return _stageObjectsChangedCalled; }
```

ppt-adsk (changes requested):
Great work!  Minor suggestions here and there.  One change to proxyStage.cpp is a requirement (no explicit version in namespace).

I would encourage you to think about throwing an exception in the ProxyStage constructor and avoid zombie object complexities.  I will let you decide on that one.

lib/mayaUsdAPI/proxyStage.cpp
```diff
@@ -0,0 +1,80 @@
...
+    ProxyStageImp(PXR_NS::ProxyStageProvider* proxyStageProvider)
+        : _proxyStageProvider(proxyStageProvider)
+    {
+    }
+
+    PXR_NS::ProxyStageProvider* const _proxyStageProvider {};
+};
+
+// Use unique_ptr with custom deleter
+using UniqueProxyStageImp = std::unique_ptr<MayaUsdAPI_v0::ProxyStageImp, ProxyStageImpDeleter>;
```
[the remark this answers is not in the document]
seando-adsk: @ppt-adsk Shouldn't it be the actual value of the namespace define "MayaUsdAPI"? That is what we do for "MayaUsd".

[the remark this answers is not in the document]
ppt-adsk: You're right, this line is a user of ProxyStageImp, it doesn't define it, so it should either be MAYAUSDAPI_NS (the macro), or as you prefer, the namespace itself, MayaUsdAPI.  Therefore, I'll change my suggestion to
```suggestion
using UniqueProxyStageImp = std::unique_ptr<MayaUsdAPI::ProxyStageImp, ProxyStageImpDeleter>;
```

lib/mayaUsdAPI/proxyStage.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+ProxyStage::ProxyStage(const ProxyStage& other)
+{
+    _imp = UniqueProxyStageImp(new ProxyStageImp(other._imp->_proxyStageProvider));
+}
+
+ProxyStage::~ProxyStage() = default;
+
+void ProxyStageImpDeleter::operator()(ProxyStageImp* ptr) const { delete ptr; }
```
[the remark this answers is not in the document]
lanierd-adsk: Actually, on Linux, even with the destructor in the cpp file, it doesn't build and complains about the unknown size of the ProxyStageImp, so I need this custom deleter to avoid this Linux build error.

lib/mayaUsdAPI/proxyStage.cpp
```diff
@@ -0,0 +1,80 @@
...
+        : _proxyStageProvider(dynamic_cast<PXR_NS::ProxyStageProvider*>(node))
+    {
+        PXR_NAMESPACE_USING_DIRECTIVE
+        TF_VERIFY(
+            _proxyStageProvider,
+            "The node passed to the constructor of ProxyStage is not a MayaUsdProxyShapeBase "
+            "subclass node while it should ! "
+            "Its type is : %s",
+            node->typeName().asChar());
+        TF_AXIOM(_proxyStageProvider);
```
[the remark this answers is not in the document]
lanierd-adsk: ok, I will change that. Thanks !

lib/mayaUsdAPI/proxyStage.cpp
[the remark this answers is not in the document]
ppt-adsk: I'd love to help you with this, because it is an example of
https://stackoverflow.com/questions/9954518/stdunique-ptr-with-an-incomplete-type-wont-compile
and should be fixed with an out of line destructor.

ppt-adsk (commented):
Good that we got to the bottom of the unique_ptr issue, thanks for that.

lib/mayaUsdAPI/proxyStage.cpp
```diff
@@ -45,36 +53,40 @@ struct ProxyStageImp
...
 ProxyStage::ProxyStage(const MObject& obj)
 {
-    PXR_NAMESPACE_USING_DIRECTIVE
-    TF_AXIOM(!obj.isNull());
+    if (obj.isNull()) {
+        const std::string errMsg
+            = PXR_NS::TfStringPrintf(
+                  "The MObject passed to the constructor of ProxyStage is null !")
+                  .c_str();
+        throw std::runtime_error(errMsg);
```
seando-adsk, line 65: Why do we need to use TfStringPrintf (there are no args in this string).
```suggestion
        throw std::runtime_error("The MObject passed to the constructor of ProxyStage is null !");
```

lib/mayaUsdAPI/proxyStage.cpp
```diff
@@ -45,36 +53,40 @@ struct ProxyStageImp
...
 ProxyStage::ProxyStage(const ProxyStage& other)
+    : _imp(new ProxyStageImp(other._imp->_proxyStageProvider))
 {
-    _imp = UniqueProxyStageImp(new ProxyStageImp(other._imp->_proxyStageProvider));
+    if (!other._imp->_proxyStageProvider) {
+        const std::string errMsg
+            = PXR_NS::TfStringPrintf(
+                  "The ProxyStage passed to the constructor of ProxyStage is null !")
+                  .c_str();
+        throw std::runtime_error(errMsg);
```
seando-adsk, line 79: Same thing here - just put the string directly in the throw.

lib/mayaUsdAPI/python/wrapProxyShapeNotice.cpp
```diff
@@ -21,102 +21,106 @@
...
             && _stageObjectsChangedKey.IsValid());
     }
 
     ~StageNoticesListener() { _Revoke(); }
 
     // StageSet notice
-    bool StageSetHasBeenCalled() const { return _stageSetCalled; }
-    void ResetStageSet() { _stageSetCalled = false; }
+    bool StageSet() const { return _bStageSet; }
+    void ResetStageSet() { _bStageSet = false; }
```
seando-adsk, line 50: According to our coding standard:
- Function/Method Names All functions should be lowerCamelCase.
- Variable Names Variables names, including function parameters and data members should use lowerCamelCase

In Ufe we used a trailing _ in the method name, so maybe you could have:

```
    bool stageSet const { return _stageSet; }
private:
    bool _stageSet;
    bool stageSet_(const MayaUsdAPI::ProxyStageSetNotice& notice) { _stageSet = true; }    
```

seando-adsk (changes requested):
Some of the variable/method naming doesn't follow our coding stanard.

ppt-adsk (changes requested):
Minor point with an out of date comment, but otherwise looks good!  Thanks for the changes!

ppt-adsk (approved):

lib/mayaUsdAPI/proxyStage.cpp
```diff
@@ -34,7 +34,15 @@ struct ProxyStageImp
...
             "Its type is : %s",
             node->typeName().asChar());
-        TF_AXIOM(_proxyStageProvider);
+        if (!_proxyStageProvider) {
+            const std::string errMsg
+                = TfStringPrintf(
+                      "The node passed to the constructor of ProxyStage is not a "
+                      "MayaUsdProxyShapeBase subclass node while it should ! Its type is : %s",
+                      node->typeName().asChar())
+                      .c_str();
```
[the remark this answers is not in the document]
seando-adsk: Sorry if it wasn't clear enough. I meant the entire string since its the same in both.

seando-adsk (approved):
Everything looks good. There is just the one comment about the duplicated string(s) in proxyStage.cpp if you want to fix that.

seando-adsk (approved):
Everything looks good. Thanks for all the hard work.

ppt-adsk (approved):

PR #3561, opened 2024-01-16
EMSUSD-956: Resource Identifier warning on first launch of MayaUSD

seando-adsk:
#### EMSUSD-956: Resource Identifier warning on first launch of MayaUSD
* Removed duplicate string resource.

### DIFF
--- a/plugin/adsk/scripts/mayaUSDRegisterStrings.mel
+++ b/plugin/adsk/scripts/mayaUSDRegisterStrings.mel
@@ -120,7 +120,6 @@ global proc mayaUSDRegisterStrings()
     register("kTimeAnn", "Edits the current time value of a stage, which corresponds to the animation frame drawn in the viewport. By default, this value connects to Maya's global time node.");
     register("kTipYouCanChooseMultipleFiles", "<b>Tip:</b>  You can choose multiple files.");
     register("kUniversalSceneDescription", "Universal Scene Description");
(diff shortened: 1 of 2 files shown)


plugin/adsk/scripts/mayaUSDRegisterStrings.mel
```diff
@@ -120,7 +120,6 @@ global proc mayaUSDRegisterStrings()
     register("kTimeAnn", "Edits the current time value of a stage, which corresponds to the animation frame drawn in the viewport. By default, this value connects to Maya's global time node.");
     register("kTipYouCanChooseMultipleFiles", "<b>Tip:</b>  You can choose multiple files.");
     register("kUniversalSceneDescription", "Universal Scene Description");
-    register("kUsdStage", "Stage");
```
seando-adsk, line 123: In PR #3383 you changed this from `register("kUsdFileOptions", "USD File Options");` to `register("kUsdStage", "Stage");`. Unfortunately that caused a duplicate string ID with the one from the Layer Editor [here](https://github.com/Autodesk/maya-usd/blob/dev/lib/usd/ui/layerEditor/stringResources.h#L91) which is `create("kUsdStage", "USD Stage:")`. Notice the strings are different. So I removed this renamed one and use the existing `register("kLabelStage", "Stage");` from line 46 of this file.

pierrebai-adsk (approved):

PR #3562, opened 2024-01-16
EMSUSD-958 : [github 3560] CXX ABI detection is broken

seando-adsk:
#### EMSUSD-958 : [github 3560] CXX ABI detection is broken
* When using Linux nm utility need "-D" flag for dynamic symbols


AramAzhari-adsk (approved):
looks good.

PR #3563, opened 2024-01-17
EMSUSD-65 wait cursor during long VP2 updates

pierrebai-adsk:
When a command may trigger a long viewport update, set a flag so that the proxy render delegate shows the wait cursor while it is updating the viewport data. We make the assumption that long-duration commands will generally cause long-duration viewport updates. This is done via the already-existing LongDurationCommand base class and wait cursor functions.

- Modify the WaitCursor helper class to take a flag to control if the cursor is shown or not.
- This allows conditionally showing the wait cursor based on a condition.
- Add a function to tell the render delegate that it should expect its update to take a long time.
- Make the wait cursor function used by long-duration commands tell the render delegate to expect a long update.

### DIFF
--- a/lib/mayaUsd/ufe/Global.cpp
+++ b/lib/mayaUsd/ufe/Global.cpp
@@ -18,6 +18,7 @@
 #include "private/UfeNotifGuard.h"
 
 #include <mayaUsd/nodes/proxyShapeStageExtraData.h>
+#include <mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.h>
 #include <mayaUsd/ufe/MayaStagesSubject.h>
 #include <mayaUsd/ufe/MayaUsdContextOpsHandler.h>
 #include <mayaUsd/ufe/MayaUsdObject3dHandler.h>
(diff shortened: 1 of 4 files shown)


pierrebai-adsk:
The only PF failure is on Linux, Maya 2022, a temporary file was not deleted. Unrelated.

AramAzhari-adsk (approved):

lib/mayaUsd/ufe/Global.cpp
```diff
@@ -108,7 +109,11 @@ MCallbackId gExitingCbId = 0;
 // Subject singleton for observation of all USD stages.
 MayaUsd::ufe::MayaStagesSubject::RefPtr g_StagesSubject;
 
-void mayaStartWaitCursor() { MGlobal::executeCommand("waitCursor -state 1"); }
+void mayaStartWaitCursor()
+{
+    ProxyRenderDelegate::setLongDurationRendering();
```
seando-adsk, line 114: When code somewhere creates a WaitCursor, it always sets the long duration. What happens if the action that used the wait cursor doesn't up in ProxyRenderDelegate::update()? The long duration flag will still be set.

pierrebai-adsk: Any non-trivial modification to USD will trigger an update of the render delegate. If somehow we get into that situation, at worst the next update will show the wait cursor, which is benign. I just wanted to avoid toggling the wait cursor in ***all*** render delegate updates since showing the cursor requires calling commands.

seando-adsk (approved):

PR #3564, opened 2024-01-18
LOOKDEVX-2403 - Allow creating custom types at NodeGraph boundaries

JGamache-autodesk:
When creating NodeGraph/Material inputs and outputs, allow passing the native type as a string instead of "Generic" when creating a port that can only be wrapped in UFE as an UfeAttributeGeneric.

This allows LookdevX to display the expected port type when promoting ports on inner nodes.


feldstj (dismissed):

pierrebai-adsk (dismissed):

JGamache-autodesk:
@seando-adsk Ready for merge. Build fix commit was only a missing UFE_V3 check and does not change the logic.

pierrebai-adsk (approved):
I was going to comment that the test would be affected, but it only runs for UFE v4, so all is good.

feldstj (approved):

PR #3566, opened 2024-01-19
EMSUSD-804 fix copying node with Arnold material.

pierrebai-adsk:
The code was not creating the necessary missing parent prims and was not detecting relationships correctly, among other smaller things.

- Add a helper to create the missing parent prims, trying to replicate the same prim type found in the temporary source stage, without actually copying those prims.
- Add a lot of debug logging to help investigate issues with duplicate-to-USD.
- Keep track of each prim that gets brought in by each prim that gets copied, recursively, so that we copy all necessary prims.
- Keep recursing into prims that have already been copied indirectly via SdfCopySpec so that their relationships can be detected too.
- This happens for example if one copies a Maya sphere than contained a Maya cube.
- Correctly detect relationships. The old way was detecting relational attributes...

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -1517,6 +1517,60 @@ void PrimUpdaterManager::discardPullSetIfEmpty()
     }
 }
 
+static void replicateMissingAncestors(
+    const UsdStageRefPtr& srcStage,
+    SdfPath               srcPath,
+    const UsdStageRefPtr& dstStage,
+    SdfPath               dstPath)
+{
+    // List f prims to be created. The last prim should be created
+    // first, (it will be the highest in the hierarchy)
+    std::vector<std::pair<SdfPath, SdfPath>> toBeCreated;
+
+    while (true) {
+        // If we reach the top of the hierarchy, stop.
+        srcPath = srcPath.GetParentPath();
+        if (srcPath.IsEmpty() || srcPath.IsAbsoluteRootPath())
+            break;
+
+        // If we reach the top of the hierarchy, stop.
+        dstPath = dstPath.GetParentPath();
+        if (dstPath.IsEmpty() || dstPath.IsAbsoluteRootPath())
+            break;
+
+        // If the destination prim already exists, stop.
+        UsdPrim dstPrim = dstStage->GetPrimAtPath(dstPath);
+        if (dstPrim) {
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/utils/traverseLayer.cpp
```diff
@@ -55,14 +55,14 @@ using AlreadyTraversedPaths = std::set<SdfPath>;
 
 void _traverseLayer(
     const SdfLayerHandle&           layer,
-    const SdfPath&                  path,
+    const SdfPath                   path,
```
samuelliu-adsk, line 58: Why can't we pass by reference? Can pathsToCopy be Null?

pierrebai-adsk: There was a problem with it becoming invalid during recursion for some reason. I did not investigate the fundamental reason.

pierrebai-adsk: I just tried again to pass by reference and again it leads to failure. It seems related to temporary variables getting mangled during recursion. Given the obscure nature of USD SdfSpec and the templated code, I currently do not have a clear explanation of why this is so.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1517,6 +1517,60 @@ void PrimUpdaterManager::discardPullSetIfEmpty()
...
+        if (srcPath.IsEmpty() || srcPath.IsAbsoluteRootPath())
+            break;
+
+        // If we reach the top of the hierarchy, stop.
+        dstPath = dstPath.GetParentPath();
+        if (dstPath.IsEmpty() || dstPath.IsAbsoluteRootPath())
+            break;
+
+        // If the destination prim already exists, stop.
+        UsdPrim dstPrim = dstStage->GetPrimAtPath(dstPath);
```
samuelliu-adsk, line 1542: Is it possible that the parent of this `dstPrim` needed to be created? I see that if `dstPrim` exists we just break the loop here.

pierrebai-adsk: No, if the prim exist, its parent necessarily exists.

lib/mayaUsd/utils/traverseLayer.cpp
pierrebai-adsk: Found the root cause, fixed it.

samuelliu-adsk: Nice!

samuelliu-adsk (approved):

PR #3572, opened 2024-01-24
EMSUSD-767 - Set the default prim on export

samuelliu-adsk:


plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -490,7 +544,7 @@ global proc mayaUsdTranslatorExport_SetFromOptions(string $currentOptions, int $
             } else if ($optionBreakDown[0] == "frameSample") {
                 mayaUsdTranslatorExport_SetTextField($optionBreakDown[1], $enable, "frameSampleField");
             } else if ($optionBreakDown[0] == "parentScope") {
-                mayaUsdTranslatorExport_SetTextField($optionBreakDown[1], $enable, "parentScopeField");
+                //mayaUsdTranslatorExport_SetTextField($optionBreakDown[1], $enable, "parentScopeField");
```
samuelliu-adsk, line 547: The reason I commented it out is that I dont think it make sense to fill the ParentScope section with the name of last export. This will also confuse the user since this field will be used as the default defaultPrim.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -625,6 +679,26 @@ global proc int mayaUsdTranslatorExport (string $parent,
 //        0 otherwise.
 //
 {
+    // check for Export All or Export Selection
+    // check for export option box window first, if it doesn't exist means user opened the main export window
```
samuelliu-adsk, line 683: Check the global variable from Maya to tell if we are in ExportAll or ExportSelection

lib/mayaUsd/fileio/jobs/writeJob.cpp
```diff
@@ -556,7 +556,10 @@ bool UsdMaya_WriteJob::_FinishWriting()
         UsdGeomSetStageMetersPerUnit(mJobCtx.mStage, mJobCtx.mArgs.metersPerUnit);
     }
 
-    if (usdRootPrim) {
+    if (!mJobCtx.mArgs.defaultPrim.empty()) {
+        defaultPrim = TfToken(mJobCtx.mArgs.defaultPrim);
+        mJobCtx.mStage->GetRootLayer()->SetDefaultPrim(defaultPrim);
```
samuelliu-adsk, line 561: This will handle the case of user selecting "None" in export, because writing `defaultPrim = "None"` means no defaultPrim

plugin/adsk/scripts/mayaUsd_exportHelpers.py
```diff
@@ -0,0 +1,66 @@
...
+    for i in allSelectedItems:
+        path = cmds.listRelatives(i, fullPath=True)[0]
+        root = path.split("|")[1]
+        allItems.add(root)
+
+    excludeSet = set()
+    if excludeMesh == "1":
+        meshes = cmds.ls(type=('mesh'), l=True)
+        for m in meshes:
+            excludeSet.add(cmds.listRelatives(m, parent=True)[0])
```
samuelliu-adsk, line 52: As stated in the story: _When using Export Selection and the selected objects are nested children, currently Maya will export its parents as well, we will continue to use this logic_

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
samuelliu-adsk: I'll remove this `else if` and put a comment here?

lib/mayaUsd/fileio/jobs/jobArgs.h
```diff
@@ -250,6 +251,7 @@ struct UsdMayaJobExportArgs
     const TfToken::Set includeAPINames;
     const TfToken::Set jobContextNames;
     const TfToken::Set excludeExportTypes;
+    const std::string  defaultPrim;
```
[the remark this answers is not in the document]
samuelliu-adsk: The default prim is parsed by the string provided by the user on export. We don't have a default token for it, so string is used

seando-adsk (dismissed):

seando-adsk (approved):

PR #3573, opened 2024-01-24
EMSUSD-972 refactor duplication code

pierrebai-adsk:
The goal is to make the duplication more testable.

Some scenarios would require the Arnold plugin to be present and we want to avoid depending on it. So instead, the code refactor split the duplication into finer-grained functions that can be tested by using custom-made temporary layers to reproduce the behavior of Arnold.

The refactor also makes the code clearer and more correct by having smaller functions. Multiple subtle problems were discovered and fixed with this strategy, related to complex cases of renaming and inter-prim relationships.

- Move code to copy temporary prims to a layer out of PrimUpdaterManager.
- This makes the code testable with pre-fabricated temp layers to test specific scenarios.
- This also lighten a bit the prim updater manager source code, which is very large.
- (More code should be refactored out of it in the future to make it simpler and more testable.)
- Fix copyLayerPrims to report all copied prims, even b=nested ones.
- This is important for display layers and renaming.
- Add Python wrapper for copyLayerPrims to allow testing.
- Add explanation to the Python wrapper why stage and layer are not passed by pointers.
- Write multiple unit tests of increasing complexities.
- Add unit tests for optionally copying relationships targets.
- Improve the renaming handling to avoid renaming chains.
- Improve renaming handling to avoid modifying the prim hierarchy while we are traversing it, otherwise the traversal becomes unpredictable and elements could be processed twice.
- Clearly explain in comments the pitfalls of rename chains.
- Add relationship targets verification to the unit tests.

### DIFF
--- a/lib/mayaUsd/fileio/primUpdaterManager.cpp
+++ b/lib/mayaUsd/fileio/primUpdaterManager.cpp
@@ -30,6 +30,7 @@
 #include <mayaUsd/ufe/Global.h>
 #include <mayaUsd/undo/OpUndoItemMuting.h>
 #include <mayaUsd/undo/OpUndoItems.h>
+#include <mayaUsd/utils/copyLayerPrims.h>
 #include <mayaUsd/utils/dynamicAttribute.h>
 #include <mayaUsd/utils/progressBarScope.h>
 #include <mayaUsd/utils/traverseLayer.h>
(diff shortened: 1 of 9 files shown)


pierrebai-adsk:
I pre-emptively fixed many typos in comments and a wrong copyright year,

AramAzhari-adsk (approved):

PR #3574, opened 2024-01-25
EMSUSD-972 add unit test for circular relationships

pierrebai-adsk:
The test revealed a flaw in the code that follows relations and connections. Fixed the code to not go into an infinite loop when there are circular relationships or connections.

- Fix infinite recursion when relations or connections form a circle.
- Added many test for connections, similar to those for relations.
- Refactored the test code to allow testing relations or connections.
- Added tests for circular targets.
- Added test for mix or relationships and connections.


pierrebai-adsk:
Seems I had to rebase to avoid github think I had a huge PR

AramAzhari-adsk (approved):

PR #3576, opened 2024-01-26
MAYA-132438 fix slow scene load performance

pierrebai-adsk:
Maya scene load without any USD data is slower when the MayaUSD plugin is loaded.

The reason behind the bad performance is that the UsdStageMap class constantly rebuilds itself because it thinks it is out of date. The reason is that its UsdStageMap::proxyShape() function assumes that it is always called to retrieve an existing proxy shape. So when it fails to find the requested proxy shape, it assumes its mapping between UFE paths and DAG paths must be stale. Those assumptions are false.

The culprit is the following sequence:
- Adding Maya nodes cause a UFE AddObject notification to be sent.
- The Maya outliner UI wants to know if this object might be a UFE gateway node.
- It asks UFE if is a gateway node.
- This ends-up being routed to MayaUSD.
- ProxyShapeSceneSegmentHandler::isGateway_() calls getStage(path) with the UFE path.
- This ends-up calling UsdStageMap::proxyShape().
- Which triggers the cache rebuilds, as explained above.

While we need to revisit how the UsdStageMap works, this would be a somewhat big and risky endeavour at this point in time. The simple fix is to not rebuild the cache when we know there are no proxy shape in existence. This fixes the performance problem for all Maya scene without USD data.

### DIFF
--- a/lib/mayaUsd/nodes/proxyShapeBase.cpp
+++ b/lib/mayaUsd/nodes/proxyShapeBase.cpp
@@ -136,6 +136,8 @@ const MString MayaUsdProxyShapeBase::displayFilterName(
     TfStringPrintf("%sDisplayFilter", MayaUsdProxyShapeBaseTokens->MayaTypeName.GetText()).c_str());
 const MString MayaUsdProxyShapeBase::displayFilterLabel("USD Proxies");
 
+std::atomic<int> g_proxyShapeInstancesCount = 0;
+
 // Attributes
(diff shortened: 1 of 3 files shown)


seando-adsk (dismissed):
Looks good, but couldn't you have used `unsigned int`?

seando-adsk (approved):

pierrebai-adsk:
The only PF failure is interactive Windows, with a temp file that was left behind, unrelated to these changes.

PR #3579, opened 2024-01-29
EMSUSD-967 fix AE flashing when adding many attributes

pierrebai-adsk:
When using the universal manipulator, especially the scaling, many attributes would get added to the prim and one refresh per attribute were being queued. Fix the code to avoid queuing many AE refreshes.


seando-adsk (approved):

pierrebai-adsk:
The testVP2RenderDelegateDisplayLayers test failed on Linux 2024. Since it passed in master, I assume it was a glitch, but given that it is an image comparison, I cannot be 100% sure.

PR #3590, opened 2024-02-02
EMSUSD-988 Maya node observer classes

pierrebai-adsk:
Add new classes to observe Maya nodes of a given type or a given Maya node. The observer forward changes to registered listeners. The goal is to standardize how a node is observed and to share the notification between multiple listeners.

This will be used by multiple classes in a subsequent pull request: MayaUsdProxyShapeBase, UsdStageMap and MayaSessionState.

The MayaNodeTypeObserver allows listening to new nodes being created or removed. The MayaNodeObserver allows listening to the node being renamed or reparented. Many use cases need to track the DAG path, UFE path or USD path of objects, mostly the USD stage itself.

### DIFF
--- a/lib/mayaUsd/utils/mayaNodeObserver.h
+++ b/lib/mayaUsd/utils/mayaNodeObserver.h
@@ -1,0 +1,121 @@
+//
+// Copyright 2024 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef MAYAUSD_MAYA_NODE_OBSERVER_H
+#define MAYAUSD_MAYA_NODE_OBSERVER_H
+
+#include <mayaUsd/base/api.h>
+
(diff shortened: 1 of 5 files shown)


pierrebai-adsk:
PF failure is that all Maya 2023 builds failed to launch at all, unrelated to my changes.

AramAzhari-adsk (approved):

pierrebai-adsk:
The only failures are three 2023 PF failed to launch, but otherwise every passed. Since this is new code not even used, it should not black merging.

The new code is used in the next PR (https://github.com/Autodesk/maya-usd/pull/3591_  which depends on this one and is working.

lib/mayaUsd/utils/mayaNodeObserver.h
```diff
@@ -0,0 +1,121 @@
...
+        //! Called when a plug is dirtied anywhere in the hierarchy of a given particular node.
+        virtual void processPlugDirty(
+            MObject& observedNode,
+            MObject& dirtiedNode,
+            MPlug&   plug,
+            bool     pathChanged);
+    };
+
+    //! Create a Maya node observer. To start observing the node, call startObserving().
+    MAYAUSD_CORE_PUBLIC
```
seando-adsk, line 55: Is there a reason why you are exporting each of the member functions rather than just exporting the whole class (like you did with the struct Listener)?

pierrebai-adsk: Yes, exporting the whole classes export the std::unordered_map below, which forces the compiler to generate all functions in the template, but some of these functions are not template-generatable because it uses types that are only moveable, so the compilations fails if we export the whole class.

Most other classes in MayaUSD also only export their public interface, not the private functions.

seando-adsk: Okay thanks for the explanation. Makes sense.

PR #3594, opened 2024-02-05
EMSUSD-1001 lost muted layer

pierrebai-adsk:
Fix that muted anonymous layers are lost when saved within the Maya scene file. The culprit was that UsdStage::GetLayerStack() did not return muted layers. The new code use an alternative method to retrieve the list of layers.

### DIFF
--- a/lib/mayaUsd/nodes/layerManager.cpp
+++ b/lib/mayaUsd/nodes/layerManager.cpp
@@ -26,6 +26,8 @@
 #include <mayaUsd/utils/utilFileSystem.h>
 #include <mayaUsd/utils/utilSerialization.h>
 
+#include <usdUfe/utils/layers.h>
+
 #include <pxr/base/arch/env.h>
 #include <pxr/base/tf/instantiateType.h>
 #include <pxr/base/tf/weakBase.h>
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -742,20 +767,11 @@ SaveStageToMayaResult saveStageToMayaFile(
 
     const MFnDependencyNode depNodeFn(proxyNode);
     MayaUsdProxyShapeBase*  pShape = static_cast<MayaUsdProxyShapeBase*>(depNodeFn.userNode());
+    if (!pShape)
+        return result;
 
-    SdfLayerHandleVector allLayers = stage->GetLayerStack(true);
```
samuelliu-adsk, line 746: Can we just replace this with `auto allLayers = UsdUfe::getAllSublayerRefs(layer, includeTopLayer);` ?

test/lib/mayaUsd/fileio/testSaveMutedAnonLayer.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python
+
+#
+# Copyright 2022 Autodesk
```
seando-adsk, line 4: Its minor, but date is not current

pierrebai-adsk:
Only change a comment, no need to run PF

PR #3597, opened 2024-02-06
EMSUSD-1000 - Implement new Hierarchy Cmd.

alicedegirolamo:
**Notes:** Implementing the new cmd added in [UFE PR](https://git.autodesk.com/media-and-entertainment/ufe/pull/375).
Marking this as a Draft PR as we will need it for next Maya release since it is breaking the Hierarchy interface by adding a new command. Pushing it to not lose progress and be ready to jump on it for the next release.
[related JIRA ticket
](https://jira.autodesk.com/browse/EMSUSD-1000)
The new command will allow us to not drop connected items in the Outliner. Please refer to the related [story ticket](https://jira.autodesk.com/browse/LOOKDEVX-1545) for more details.

### DIFF
--- a/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
+++ b/lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
@@ -251,6 +251,18 @@ ProxyShapeHierarchy::insertChild(const Ufe::SceneItem::Ptr& child, const Ufe::Sc
     return insertChildCommand->insertedChild();
 }
 
+Ufe::InsertChildCommand::Ptr
+ProxyShapeHierarchy::appendChildVerifyRestrictionsCmd(const Ufe::SceneItem::Ptr& child)
+{
+    const auto childCmd = appendChildCmd(child);
+
+    if (childCmd && isConnected(downcast(child))) {
+        throw std::runtime_error("The node you're trying to move has connections.");
+    }
+
+    return childCmd;
+}
+
 #ifdef UFE_V3_FEATURES_AVAILABLE
 Ufe::SceneItem::Ptr ProxyShapeHierarchy::createGroup(const Ufe::PathComponent& name) const
 {
--- a/lib/mayaUsd/ufe/Utils.h
+++ b/lib/mayaUsd/ufe/Utils.h
@@ -116,11 +116,6 @@ PXR_NS::UsdTimeCode getTime(const Ufe::Path& path);
 MAYAUSD_CORE_PUBLIC
 PXR_NS::TfTokenVector getProxyShapePurposes(const Ufe::Path& path);
 
-//! Check if the src and dst attributes are connected.
-//! \return True, if they are connected.
-MAYAUSD_CORE_PUBLIC
-bool isConnected(const PXR_NS::UsdAttribute& srcUsdAttr, const PXR_NS::UsdAttribute& dstUsdAttr);
-
 //! Check if a source connection property is allowed to be removed.
 //! \return True, if the property can be removed.
 MAYAUSD_CORE_PUBLIC
--- a/lib/usdUfe/ufe/UsdHierarchy.cpp
+++ b/lib/usdUfe/ufe/UsdHierarchy.cpp
@@ -257,6 +257,19 @@ UsdHierarchy::insertChildCmd(const Ufe::SceneItem::Ptr& child, const Ufe::SceneI
     return UsdUndoInsertChildCommand::create(fItem, downcast(child), downcast(pos));
 }
 
+Ufe::InsertChildCommand::Ptr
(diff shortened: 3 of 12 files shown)


lib/usdUfe/ufe/UsdHierarchy.cpp
```diff
@@ -257,6 +257,19 @@ UsdHierarchy::insertChildCmd(const Ufe::SceneItem::Ptr& child, const Ufe::SceneI
     return UsdUndoInsertChildCommand::create(fItem, downcast(child), downcast(pos));
 }
 
+Ufe::InsertChildCommand::Ptr
+UsdHierarchy::appendChildVerifyRestrictionsCmd(const Ufe::SceneItem::Ptr& child)
+{
+    const auto childCmd = appendChildCmd(child);
```
alicedegirolamo, line 263: A better and cleaner solution would be to create a new cmd that inherits from `UsdUndoInsertChildCommand` and in the new constructor check the restrictions and throw the exception.

lib/mayaUsd/ufe/Utils.h
```diff
@@ -116,11 +116,6 @@ PXR_NS::UsdTimeCode getTime(const Ufe::Path& path);
 MAYAUSD_CORE_PUBLIC
 PXR_NS::TfTokenVector getProxyShapePurposes(const Ufe::Path& path);
 
-//! Check if the src and dst attributes are connected.
-//! \return True, if they are connected.
-MAYAUSD_CORE_PUBLIC
-bool isConnected(const PXR_NS::UsdAttribute& srcUsdAttr, const PXR_NS::UsdAttribute& dstUsdAttr);
```
alicedegirolamo, line 122: Moving it to `usdUfe/ufe` as I see this function doesn't need the maya and mayaUsd .h included in `lib/mayaUsd/ufe/Utils.h`

lib/usdUfe/ufe/Utils.h
```diff
@@ -304,6 +304,16 @@ bool isEditTargetLayerModifiable(
...
 
+//! Check if the src and dst attributes are connected.
+//! \return True, if they are connected.
+USDUFE_PUBLIC
+bool isConnected(const PXR_NS::UsdAttribute& srcUsdAttr, const PXR_NS::UsdAttribute& dstUsdAttr);
+
+//! Check if the usdItem is connected (i.e. if there are in or out connections).
+//! \return True, if it is connected.
+USDUFE_PUBLIC
+bool isConnected(const UsdSceneItem::Ptr& usdItem);
```
alicedegirolamo, line 315: Moving out this function from [LookdevXUsd ](https://git.autodesk.com/media-and-entertainment/lookdevx/blob/1adbbe744fb7e64372acb9b858f75718bbc14250/source/LookdevXUsd/UsdClipboardHandler.cpp#L217)

lib/usdUfe/ufe/UsdHierarchy.cpp
```diff
@@ -257,6 +257,19 @@ UsdHierarchy::insertChildCmd(const Ufe::SceneItem::Ptr& child, const Ufe::SceneI
...
+
+    if (childCmd && isConnected(downcast(child))) {
+        throw std::runtime_error("The node you're trying to move has connections.");
```
alicedegirolamo, line 266: We should update this message according to the updated and approved UI linked in the related story

lib/mayaUsd/ufe/ProxyShapeHierarchy.cpp
```diff
@@ -251,6 +251,18 @@ ProxyShapeHierarchy::insertChild(const Ufe::SceneItem::Ptr& child, const Ufe::Sc
...
 }
 
+Ufe::InsertChildCommand::Ptr
+ProxyShapeHierarchy::appendChildVerifyRestrictionsCmd(const Ufe::SceneItem::Ptr& child)
+{
+    const auto childCmd = appendChildCmd(child);
+
+    if (childCmd && isConnected(downcast(child))) {
+        throw std::runtime_error("The node you're trying to move has connections.");
+    }
```
seando-adsk, line 261: In your Ufe PR it states "Returns a null pointer if the restrictions are not observed." - but here you are throwing. It should be one of the. Either return nullptr here or in Ufe change the doc to state that it will throw runtime_error if the restrictions are not observed.

seando-adsk (commented):
I have a general comment about the need for this new method. The JIRA item mentions "modify Outliner code to not allow dropped items if they are connected". Where is this new method called from? Wouldn't users be able to override the restriction by simply calling `appendChildCmd`? Therefore was any thought given to adding the restriction to that existing method?

seando-adsk:
After discussion with Alice this PR is no longer valid. There are no current plans to go back to work on it. Closing.

PR #3599, opened 2024-02-07
EMSUSD-988 Avoid rebuilding the stage map when a path is not found

pierrebai-adsk:
The UsdStageMap would always rebuild the mapping whenever a given UFE path was not found in its mapping. The code assumed that when it was called, it was always with a valid path to a valid stage. In reality, it gets called for various paths that are not stages. In particular, the isGateway() function is called for any Maya node, to verify if it would be a proxy shape node. Finding all proxy shape was done using an expensive call to list all node of the proxy shape type.

Change the logic to instead rely on proper notifications about proxy shape being created, renamed or reparented and do not rebuild the map on each call.

At the same time, share a common Maya node observer between multiple classes that want to observe the proxy shapes.

UsdStageMap changes:
- Add debug code for the UsdStageMap called MAYAUSD_STAGEMAP to help when debugging.
- Make the UsdStageMap singleton not be a global variable.
- Simplify the proxyShape() function of UsdStageMap to not rebuild on misses.
- Make the UsdStageMap use the MayaUsdProxyShapeBase node observer to receives notification that a proxy shape has changed name, or path, etc.

MayaUsdProxyShapeBase changes:
- Make the MayaUsdProxyShape use the new node-type observer, MayaNodeTypeObserver .
- Update the parent callbacks when ancestor nodes are modified.

MayaSessionState changes:
- Use the node observer instead of having similar code.


lib/mayaUsd/ufe/Global.cpp
```diff
@@ -117,6 +117,9 @@ void mayaStartWaitCursor()
 
 void mayaStopWaitCursor() { MGlobal::executeCommand("waitCursor -state 0"); }
 
+// Note: MayaUsd::ufe::getStage takes two parameters, so wrap it in a function taking only one.
+PXR_NS::UsdStageWeakPtr mayaGetStage(const Ufe::Path& path) { return MayaUsd::ufe::getStage(path); }
```
AramAzhari-adsk, line 121: Is the issue with the default parameters that requires this wrap?

pierrebai-adsk: yes

AramAzhari-adsk (approved):

PR #3600, opened 2024-02-07
EMSUSD-1007 fix layers not muted on reload

pierrebai-adsk:
- Allow retrieving a mapping of renamed layers from the LayerManager Maya node that keeps track of USD layers saved inside a Maya scene.
- Pass that map of layer names to the function that restore layer muting.
- Make sure muted layers are not lost by adding them to the set of retained muted layers.
- This is especially important for anonymous layers, as nothing else refer to them.


samuelliu-adsk:
LGTM

samuelliu-adsk (approved):

PR #3605, opened 2024-02-09
EMSUSD-708 instanceable prim performance fix

pierrebai-adsk:
Using pre-calculated cached data that was produced for isolate select.


lib/mayaUsd/render/vp2RenderDelegate/mayaPrimCommon.cpp
```diff
@@ -806,10 +806,15 @@ void MayaUsdRPrim::_SyncDisplayLayerModesInstanced(SdfPath const& id, unsigned i
 
     _forcedReprFlags = 0;
     _requiredModFlagsBitset.reset();
-    if (_useInstancedDisplayLayerModes) {
+    if (_useInstancedDisplayLayerModes && instanceCount > 0) {
         _displayLayerModesInstanced.resize(instanceCount);
+        InstancePrimPaths& instancePrimPaths = MayaUsdCustomData::GetInstancePrimPaths(id);
```
pierrebai-adsk, line 811: I'm not sure if using the cached MayaUsdCustomData here is valid and correct. That is why I opened this code review to validate with someone more familiar with the code than I.

vlasovi: From what I see in the code, this cache is updated inside mesh.cpp, in the cycle below line 1995. And you use this cache in the function _SyncDisplayLayerModesInstanced just above this line. So it seems like you use the cache before it's updated. Apart from that, using this cache seems like a good idea. Although, to calculate the cache, we still need to call GetScenePrimPath functions

pierrebai-adsk: Ah, that would explain the failed tests! Thank you.

PR #3606, opened 2024-02-12
EMSUSD-1016 - Link against mayaUsdAPI to preload it for other plugins

gz-adsk:
The mayaUsdAPI library is not directly needed by the mayaUsd plugin. However, there is no way currently for other plugins that need it to know where to locate it on mac/linux through rpaths. By linking it it will be preloaded like the other dynamic libs.


seando-adsk (approved):
Thanks for figuring this out and submitting the fix.

PR #3607, opened 2024-02-12
EMSUSD-708 fix performance of instanceable prims

pierrebai-adsk:
Call the Usdimaging function to retrieve all instance prim path at once instead of one-by-one, which optimizes many internal details.

- Use the same faster code in a few other similar places.
- Add a convenience function to retrieve all the prim paths when the instance ID are consecutive.
- Use it in a few places.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
@@ -910,12 +910,12 @@ void HdVP2BasisCurves::_UpdateDrawItem(
         if (0 == instanceCount) {
             instancerWithNoInstances = true;
         } else {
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/render/vp2RenderDelegate/basisCurves.cpp
```diff
@@ -910,12 +910,12 @@ void HdVP2BasisCurves::_UpdateDrawItem(
         if (0 == instanceCount) {
             instancerWithNoInstances = true;
         } else {
+            const SdfPathVector usdPaths = drawScene.GetScenePrimPaths(id, instanceCount);
```
AramAzhari-adsk, line 913: I believe this uses vectorization to save on calculations. Good call!

AramAzhari-adsk (approved):

PR #3608, opened 2024-02-13
EMSUSD-1066 - Minimal API needed by LookdevX to decouple from mayaUsd

gz-adsk:
This will remove the need for LookdevX to link directly to `mayaUsd` and `mayaUsdUtils` with a couple of exceptions that will be sorted out  at a later time.


seando-adsk (approved):
Ideally this new utils should have a python wrapper and have some kind of minimal unit tests. I'm okay with this PR going in now as long as we come back later and add the wrapper and tests.

PR #3619, opened 2024-02-15
EMSUSD-872 - Adds the ability to System-lock a layer

AramAzhari-adsk:
The system lock is a severity of locking a layer where it cannot be undone using UI. It is meant to be controlled through script or programmatically to prevent a user from editing or saving a layer.

The locking can be performed using the mayaUsdLayerEditor command with the lockLayer parameter:
- `lockLayer 0` is unlocked 
- `lockLayer 1` is locked: Sets the PermissionToEdit to false
- `lockLayer 2` is system locked: Sets both the PermissionToEdit and PermissionToSave to false

Example of System Locking:

In the above example:
- (1) asdf.usda is system locked, No Save, No Edit
- As a result of the system lock, the anonymous layers (2), (3) cannot be saved since saving them will cause a change in asdf.usda. The anonymous layers can still be edited.
- qwerty.usda changes can be saved since it doesn't affect the system locked parent.

Any Layer Editor operations that affect the locked parent are disabled on sublayers. This includes context menu operations.

### DIFF
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -17,6 +17,7 @@
 #include "layerEditorCommand.h"
 
 #include <mayaUsd/ufe/Global.h>
+#include <mayaUsd/utils/layerLocking.h>
 #include <mayaUsd/utils/layerMuting.h>
 #include <mayaUsd/utils/query.h>
 #include <mayaUsd/utils/stageCache.h>
@@ -59,6 +60,8 @@ const char kMuteLayerFlag[] = "mt";
 const char kMuteLayerFlagL[] = "muteLayer";
 const char kLockLayerFlag[] = "lk";
 const char kLockLayerFlagL[] = "lockLayer";
+const char kSystemLockLayerFlag[] = "sk";
+const char kSystemLockLayerFlagL[] = "SystemLockLayer";
 
 } // namespace
 
@@ -563,7 +566,7 @@ public:
         }
 
         // Note: backup the edit targets after the layer is cleared because we use
-        //       the fact that a stage edit target is now invalid to decice to backup
+        //       the fact that a stage edit target is now invalid to decide to backup
         //       that edit target.
         backupEditTargets(layer);
 
@@ -789,12 +792,22 @@ public:
         auto stage = getStage();
         if (!stage)
             return false;
-        if (_lockIt) {
+        if (_systemLock) {
+            saveSelection();
+            layer->SetPermissionToSave(false);
+            layer->SetPermissionToEdit(false);
+            MayaUsd::addSystemLockedLayer(layer);
+        } else if (_lockIt) {
             saveSelection();
             layer->SetPermissionToEdit(false);
+            layer->SetPermissionToSave(true);
+            MayaUsd::removeSystemLockedLayer(layer);
+
         } else {
             saveSelection();
             layer->SetPermissionToEdit(true);
+            layer->SetPermissionToSave(true);
+            MayaUsd::removeSystemLockedLayer(layer);
         }
 
         return true;
@@ -805,11 +818,15 @@ public:
         auto stage = getStage();
         if (!stage)
             return false;
-        if (_lockIt) {
+        if (_systemLock) {
             layer->SetPermissionToEdit(true);
+            layer->SetPermissionToSave(true);
+            MayaUsd::removeSystemLockedLayer(layer);
             restoreSelection();
-        } else {
-            layer->SetPermissionToEdit(false);
+        } else if (_lockIt) {
+            layer->SetPermissionToEdit(true);
+            layer->SetPermissionToSave(true);
+            MayaUsd::removeSystemLockedLayer(layer);
             restoreSelection();
         }
 
@@ -818,6 +835,7 @@ public:
 
     std::string _proxyShapePath;
     bool        _lockIt = true;
+    bool        _systemLock = false;
 
 private:
     UsdStageWeakPtr getStage()
@@ -949,7 +967,7 @@ MSyntax LayerEditorCommand::createSyntax()
     syntax.makeFlagMultiUse(kAddAnonSublayerFlag);
     // paramter: proxy shape name
     syntax.addFlag(kMuteLayerFlag, kMuteLayerFlagL, MSyntax::kBoolean, MSyntax::kString);
-    syntax.addFlag(kLockLayerFlag, kLockLayerFlagL, MSyntax::kBoolean, MSyntax::kString);
+    syntax.addFlag(kLockLayerFlag, kLockLayerFlagL, MSyntax::kLong, MSyntax::kString);
 
     return syntax;
 }
@@ -1091,8 +1109,11 @@ MStatus LayerEditorCommand::parseArgs(const MArgList& argList)
             _subCommands.push_back(std::move(cmd));
         }
         if (argParser.isFlagSet(kLockLayerFlag)) {
-            bool lockIt = true;
-            argParser.getFlagArgument(kLockLayerFlag, 0, lockIt);
+            int lockValue = 0;
+            // 0 = Unlocked
+            // 1 = Locked
+            // 2 = SystemLocked
+            argParser.getFlagArgument(kLockLayerFlag, 0, lockValue);
 
             MString proxyShapeName;
             argParser.getFlagArgument(kLockLayerFlag, 1, proxyShapeName);
@@ -1105,7 +1126,8 @@ MStatus LayerEditorCommand::parseArgs(const MArgList& argList)
             }
 
             auto cmd = std::make_shared<Impl::LockLayer>();
-            cmd->_lockIt = lockIt;
+            cmd->_lockIt = lockValue == 1 ? true : false;
+            cmd->_systemLock = lockValue == 2 ? true : false;
             cmd->_proxyShapePath = proxyShapeName.asChar();
             _subCommands.push_back(std::move(cmd));
         }
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -60,8 +60,6 @@ const char kMuteLayerFlag[] = "mt";
 const char kMuteLayerFlagL[] = "muteLayer";
 const char kLockLayerFlag[] = "lk";
 const char kLockLayerFlagL[] = "lockLayer";
-const char kSystemLockLayerFlag[] = "sk";
-const char kSystemLockLayerFlagL[] = "SystemLockLayer";
 
 } // namespace
 
--- a/lib/mayaUsd/utils/CMakeLists.txt
+++ b/lib/mayaUsd/utils/CMakeLists.txt
@@ -12,6 +12,7 @@ target_sources(${PROJECT_NAME}
         dynamicAttribute.cpp
         editability.cpp
         json.cpp
+		layerLocking.cpp
         layerMuting.cpp
         layers.cpp
         loadRulesAttribute.cpp
@@ -45,6 +46,7 @@ set(HEADERS
     editability.h
     hash.h
     json.h
+	layerLocking.h
     layerMuting.h
     layers.h
     loadRules.h
--- a/lib/mayaUsd/utils/CMakeLists.txt
+++ b/lib/mayaUsd/utils/CMakeLists.txt
@@ -12,7 +12,7 @@ target_sources(${PROJECT_NAME}
         dynamicAttribute.cpp
         editability.cpp
         json.cpp
-		layerLocking.cpp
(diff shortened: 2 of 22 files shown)


lib/mayaUsd/utils/layerLocking.h
```diff
@@ -0,0 +1,89 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef MAYAUSD_LAYERLOCKING_H
+#define MAYAUSD_LAYERLOCKING_H
+
+#include <mayaUsd/base/api.h>
+#include <mayaUsd/nodes/proxyShapeBase.h>
```
AramAzhari-adsk, line 20: The proxy shape will be updated on future iterations to add new functionality to layer locking.

lib/mayaUsd/utils/layerLocking.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+    // Note: C++ guarantees correct multi-thread protection for static
+    //       variables initialization in functions.
+    static SceneResetListener onSceneResetListener;
+    static LockedLayers       layers;
+    return layers;
+}
+} // namespace
+
+void addSystemLockedLayer(const PXR_NS::SdfLayerRefPtr& layer)
```
AramAzhari-adsk, line 59: There are similar functions in layer muting. They will be extended to in the future iterations.

lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -265,15 +266,13 @@ bool LayerTreeItem::sublayerOfShared() const
     return false;
 }
 
-bool LayerTreeItem::isReadOnly() const
-{
-    return (_isSharedLayer || (_layer && !_layer->PermissionToEdit()));
-}
+bool LayerTreeItem::isReadOnly() const { return (_isSharedLayer); }
```
AramAzhari-adsk, line 269: Note: By design some of the functionalities should still be available if a layer has no permission to edit.

lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -282,15 +281,21 @@ bool LayerTreeItem::isLocked() const { return _layer && _layer->PermissionToEdit
 
 bool LayerTreeItem::appearsLocked() const
```
AramAzhari-adsk, line 282: Named similar to appearsMuted(). It has to do with the parent / children's state of lock.
By design, some of the sublayer's actions cannot be performed if the parent is locked.

lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -282,15 +281,21 @@ bool LayerTreeItem::isLocked() const { return _layer && _layer->PermissionToEdit
...
+    if (item != nullptr) {
+        return item->isLocked();
     }
+
+    return false;
+}
+
+bool LayerTreeItem::isSystemLocked() const { return MayaUsd::isLayerSystemLocked(_layer); }
+
+bool LayerTreeItem::appearsSystemLocked() const
```
AramAzhari-adsk, line 294: Named similar to appearsLocked(). It has to do with the parent / children's state of system lock.
By design, some of the sublayer's actions cannot be performed if the parent is locked.

lib/usd/ui/layerEditor/layerTreeItemDelegate.cpp
```diff
@@ -283,7 +289,11 @@ void LayerTreeItemDelegate::paint_ActionIcon(
 
     QString tooltip;
     if (_lastHitAction == action._name) {
-        tooltip = action._tooltip;
+        if (item->isSystemLocked()) {
```
AramAzhari-adsk, line 292: When a layer is system locked, a different message is shown to the user to indicate that a layer cannot be unlocked or edited.

lib/usd/ui/layerEditor/layerTreeModel.cpp
```diff
@@ -151,7 +151,7 @@ bool LayerTreeModel::canDropMimeData(
     }
 
     auto parentItem = layerItemFromIndex(parentIndex);
-    if (!parentItem || parentItem->isReadOnly()) {
+    if (!parentItem || parentItem->isReadOnly() || parentItem->isLocked()) {
```
AramAzhari-adsk, line 154: By design, immediate children of a locked parent layer cannot be reorganised (also covers the system lock layers by the shared constraint of PermissionToEdit = false).

lib/usd/ui/layerEditor/layerTreeModel.cpp
```diff
@@ -235,7 +235,8 @@ bool LayerTreeModel::dropMimeData(
 
 void LayerTreeModel::setEditTarget(LayerTreeItem* item)
 {
-    if (!item->appearsMuted() && !item->isReadOnly()) {
+    if (!item->appearsMuted() && !item->isReadOnly() && !item->isLocked()
```
AramAzhari-adsk, line 238: A locked or system locked layer cannot be targeted.

lib/usd/ui/layerEditor/layerTreeModel.cpp
```diff
@@ -481,8 +482,11 @@ void LayerTreeModel::saveStage(QWidget* in_parent)
     auto saveAllLayers = [this]() {
         const auto layers = getAllNeedsSavingLayers();
         for (auto layer : layers) {
-            if (!layer->isAnonymous())
-                layer->saveEditsNoPrompt();
+            if (!layer->isSystemLocked()) {
```
AramAzhari-adsk, line 485: A System-locked layer cannot be saved by the extension that PermissionToSave = False.

lib/usd/ui/layerEditor/layerTreeModel.cpp
```diff
@@ -570,15 +574,15 @@ void LayerTreeModel::toggleMuteLayer(LayerTreeItem* item, bool* forcedState)
 
 void LayerTreeModel::toggleLockLayer(LayerTreeItem* item, bool* forcedState /*= nullptr*/)
 {
-    if (item->isInvalidLayer() || item->isSessionLayer())
+    if (item->isInvalidLayer() || item->isSessionLayer() || item->isSystemLocked())
```
AramAzhari-adsk, line 577: No system locking is allowed through toggling.

lib/usd/ui/layerEditor/saveLayersDialog.cpp
```diff
@@ -486,18 +487,41 @@ void SaveLayersDialog::getLayersToSave(
     for (const auto& layerInfo : StageLayersToSave._anonLayers) {
         _stageLayerMap.emplace(std::make_pair(layerInfo.layer, stageName));
     }
+
     for (const auto& dirtyLayer : StageLayersToSave._dirtyFileBackedLayers) {
         _stageLayerMap.emplace(std::make_pair(dirtyLayer, stageName));
     }
 
+    // We do not allow saving layers in any of the following conditions:
```
AramAzhari-adsk, line 495: SaveLayerDialog will no longer list anonymous or file backed layers that are system locked, whether directly or by the extension of the parent layers.

lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -805,11 +818,15 @@ class LockLayer : public BaseCmd
         auto stage = getStage();
         if (!stage)
             return false;
-        if (_lockIt) {
+        if (_systemLock) {
```
AramAzhari-adsk, line 821: Note: For the time being, the undo state of a system lock goes into fully unlock. In a future design iterations a system-lock may get changed to never be undoable to prevent accidental unlocks.

pierrebai-adsk: Ideally, it should return to the previous lock state (for example: locked -> system locked -> undo -> locked).

lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -1091,8 +1109,11 @@ MStatus LayerEditorCommand::parseArgs(const MArgList& argList)
             _subCommands.push_back(std::move(cmd));
         }
         if (argParser.isFlagSet(kLockLayerFlag)) {
-            bool lockIt = true;
-            argParser.getFlagArgument(kLockLayerFlag, 0, lockIt);
+            int lockValue = 0;
+            // 0 = Unlocked
```
pierrebai-adsk, line 1113: I wished there was an enum so that we would not have these magic values in the code.

lib/mayaUsd/utils/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ target_sources(${PROJECT_NAME}
         dynamicAttribute.cpp
         editability.cpp
         json.cpp
+		layerLocking.cpp
```
pierrebai-adsk, line 15: Not very important: indentation. If you make other adjustements in other files, you might as well fix it, otherwise it is really not critical.

lib/mayaUsd/utils/layerLocking.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+    if (!layer)
+        return;
+
+    LockedLayers& layers = getLockedLayers();
+    auto          iter = layers.find(layer);
+    if (iter != layers.end()) {
```
pierrebai-adsk, line 66: Given that it is a set, I'm not sure why you bother checking if it is already in the set. Adding again would not do any harm.

lib/usd/ui/layerEditor/abstractCommandHook.h
```diff
@@ -80,7 +80,7 @@ class AbstractCommandHook
     virtual void muteSubLayer(UsdLayer usdLayer, bool muteIt) = 0;
 
     // lock or unlock the given layer
-    virtual void lockSubLayer(UsdLayer usdLayer, bool lockIt) = 0;
+    virtual void lockSubLayer(UsdLayer usdLayer, bool lockIt, bool systemLock) = 0;
```
pierrebai-adsk, line 83: Could pass 0, 1, 2 like the MEL command and even reuse the suggested enum to be used instead of 0, 1, 2.

lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -282,15 +281,21 @@ bool LayerTreeItem::isLocked() const { return _layer && _layer->PermissionToEdit
 
 bool LayerTreeItem::appearsLocked() const
 {
-    if (isLocked()) {
-        return true;
```
pierrebai-adsk, line 286: Should it not *also* check if it is locked itself?

lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -282,15 +281,21 @@ bool LayerTreeItem::isLocked() const { return _layer && _layer->PermissionToEdit
...
+
+    return false;
+}
+
+bool LayerTreeItem::isSystemLocked() const { return MayaUsd::isLayerSystemLocked(_layer); }
+
+bool LayerTreeItem::appearsSystemLocked() const
+{
     auto item = parentLayerItem();
-    while (item != nullptr) {
```
pierrebai-adsk, line 289: Same question: it seems weird that appearsLocked does not check itself.

lib/usd/ui/layerEditor/layerTreeItem.cpp
pierrebai-adsk: Oh, so you want to be able to distinguish between something being locked and something being locked by its parent? So that is why appearsLocked does not check itself? OK, but the function name is a bit misleading. You can keep them like that, just a trap for the unwary :)

lib/usd/ui/layerEditor/layerTreeItem.h
```diff
@@ -138,6 +138,10 @@ class LayerTreeItem : public QStandardItem
     bool isLocked() const;
     // check if this layer is locked
     bool appearsLocked() const;
+    // is the layer system locked?
+    bool isSystemLocked() const;
+    // check if this layer is system locked
```
pierrebai-adsk, line 143: I'd change the comment to reflect that this check if parents are locked and not itself. I'd even add a similar comment in the cpp to help future readers understand that not checking itself is volontary and not an oversight.

lib/usd/ui/layerEditor/layerTreeItem.cpp
AramAzhari-adsk: This is similarly named to appearsMuted. The notion is that something "appears" locked but it's not really locked itself. Hence the word appear.

pierrebai-adsk (approved):
LGTM, only minor possible improvements.

PR #3624, opened 2024-02-19
LOOKDEVX-2547 - Hide component connection nodes from AE

JGamache-autodesk:


test/testSamples/MaterialX/multiple_connections.usda
```diff
@@ -101,5 +101,65 @@ def Material "Material1" (
             uniform float2 ui:nodegraph:node:pos = (-1.1293055, -3.50525)
         }
     }
+
```
feldstj, line 104: Should you also have a combine with no separate case?

feldstj: And a separate with no combine case?

feldstj (approved):
LGTM. Just wondering if the test file should include more cases.

seando-adsk (approved):
Looks okay for me. Thanks for the unit test.

test/testSamples/MaterialX/multiple_connections.usda
JGamache-autodesk: The code does not use any knowledge except the id strings and is very generic in how it traverses inputs. It will work across exotic scenarios (like a combine with one connection from a separate and another from a float node) without having to test every single permutation.

PR #3625, opened 2024-02-19
EMSUSD-1017 Fix reverting multiple files

pierrebai-adsk:
When reverting multiple USD layers, reverting the first layer would trigger the UI to update, thus making the the element holding the other layers invalid before they were processed.

To avoid this, add a mechanism to delay the execution of commands in the command-hook so that all item can be processed before the first command is executed.

In the Maya implementation of the command hook, the MEL and Python commands are accumulated when the commands are delayed and executed once the delayed commands are triggered.

### DIFF
--- a/lib/usd/ui/layerEditor/mayaCommandHook.cpp
+++ b/lib/usd/ui/layerEditor/mayaCommandHook.cpp
@@ -39,21 +39,6 @@ namespace {
 
 std::string quote(const std::string& string) { return STR(" \"") + string + STR("\""); }
 
-MString executeMel(const std::string& commandString)
-{
-    // executes maya command with display and undo set to true so that it logs
-    MStringArray result;
-    MGlobal::executeCommand(
-        MString(commandString.c_str()),
-        result,
-        /*display*/ true,
-        /*undo*/ true);
-    if (result.length() > 0)
-        return result[0];
-    else
-        return "";
-}
-
 // maya doesn't support spaces in undo chunk names...
 MString cleanChunkName(QString name) { return quote(name.replace(" ", "_").toStdString()).c_str(); }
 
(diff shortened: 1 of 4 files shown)


lib/usd/ui/layerEditor/mayaCommandHook.cpp
```diff
@@ -295,4 +283,50 @@ bool MayaCommandHook::isProxyShapeSharedStage(const std::string& proxyShapePath)
...
+void MayaCommandHook::executePython(const std::string& commandString)
+{
+    if (areCommandsDelayed()) {
+        _delayedCommands.push_back({ commandString, true });
+    } else {
+        MGlobal::executePythonCommand(commandString.c_str());
+    }
+}
+
+void MayaCommandHook::executeDelayedCommands()
```
AramAzhari-adsk, line 313: Would you mind adding a comment that this function should not be called directly?
Also, do we always want to have display on?

lib/usd/ui/layerEditor/mayaCommandHook.cpp
```diff
@@ -295,4 +283,50 @@ bool MayaCommandHook::isProxyShapeSharedStage(const std::string& proxyShapePath)
...
+{
+    if (areCommandsDelayed()) {
+        _delayedCommands.push_back({ commandString, false });
+    } else {
+        // executes maya command with display and undo set to true so that it logs
+        MStringArray result;
+        MGlobal::executeCommand(
+            MString(commandString.c_str()),
+            result,
+            /*display*/ true,
```
AramAzhari-adsk, line 296: Do we always want to have display on?

AramAzhari-adsk (approved):
LGTM

PR #3626, opened 2024-02-21
EMSUSD-873 - Adds write permission checks to layer system-locking

AramAzhari-adsk:
Some of the layer operations will refresh the system-lock based on the write permissions of those layers. The operations are as follows:

- When the user initiates a layer reload (through the context menu in the usd layer editor)
- When the user initiates a stage reload (through attribute editor)
- When the user loads a stage from a file.
- Anything that may cause a session stage change on the LayerTreeModel (including the case of stage reload).
- Using script (refreshSystemLock)

Note that:
- Anonymous layers are not checked for write permissions.
- The check is executed regardless of changes on disk. Only the write permissions are checked (as per design).

### DIFF
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -19,8 +19,10 @@
 #include <mayaUsd/ufe/Global.h>
 #include <mayaUsd/utils/layerLocking.h>
 #include <mayaUsd/utils/layerMuting.h>
+#include <mayaUsd/utils/layers.h>
 #include <mayaUsd/utils/query.h>
 #include <mayaUsd/utils/stageCache.h>
+#include <mayaUsd/utils/util.h>
 #include <mayaUsd/utils/utilFileSystem.h>
 
 #include <usdUfe/ufe/Utils.h>
@@ -29,6 +31,8 @@
 
 #include <maya/MArgList.h>
 #include <maya/MArgParser.h>
+#include <maya/MGlobal.h>
+#include <maya/MString.h>
 #include <maya/MStringArray.h>
 #include <maya/MSyntax.h>
 #include <ufe/globalSelection.h>
@@ -60,6 +64,8 @@ const char kMuteLayerFlag[] = "mt";
 const char kMuteLayerFlagL[] = "muteLayer";
 const char kLockLayerFlag[] = "lk";
 const char kLockLayerFlagL[] = "lockLayer";
+const char kRefreshSystemLockFlag[] = "rl";
+const char kRefreshSystemLockFlagL[] = "refreshSystemLock";
 
 } // namespace
 
@@ -77,7 +83,8 @@ enum class CmdId
     kClearLayer,
     kAddAnonLayer,
     kMuteLayer,
-    kLockLayer
+    kLockLayer,
+    kRefreshSystemLock
 };
 
 class BaseCmd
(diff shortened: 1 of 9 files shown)


lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -875,6 +882,110 @@ class LockLayer : public BaseCmd
...
+    // that are undoable.
+    bool undoIt(SdfLayerHandle layer) override { return true; }
+
+    std::string _proxyShapePath;
+    bool        _refreshSubLayers = false;
+
+private:
+    std::string _quote(const std::string& string)
+    {
+        return std::string(" \"") + string + std::string("\"");
```
pierrebai-adsk, line 925: std:: in C++has the [std::quoted function](https://en.cppreference.com/w/cpp/io/manip/quoted) which deals with nested quotes. No need to update just for this.

lib/usd/ui/layerEditor/mayaCommandHook.cpp
```diff
@@ -215,30 +219,57 @@ void MayaCommandHook::lockSubLayer(UsdLayer usdLayer, MayaUsd::LayerLockType loc
...
-    case MayaUsd::LayerLockType::LayerLock_Locked: {
-        cmd += "1";
-        break;
-    }
-    case MayaUsd::LayerLockType::LayerLock_SystemLocked: {
-        cmd += "2";
-        break;
-    }
-    }
+    cmd += std::to_string(lockState);
```
pierrebai-adsk, line 222: Neat, I always forget about std::to_string

lib/usd/ui/layerEditor/mayaCommandHook.cpp
```diff
@@ -215,30 +219,57 @@ void MayaCommandHook::lockSubLayer(UsdLayer usdLayer, MayaUsd::LayerLockType loc
...
+            _refreshLayerSystemLock(layer);
+        }
+    } else {
+        _refreshLayerSystemLock(usdLayer);
+    }
+}
+
+void MayaCommandHook::_refreshLayerSystemLock(UsdLayer usdLayer)
+{
+    // Anonymous layers do not need to be checked.
```
pierrebai-adsk, line 244: This code is pretty similar to the one in the command above. IDK if it would be possible to put it in a shared function?

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -278,5 +289,11 @@ def ProxyShapeFilePathRefresh(filePathAttr):
                 if res == kYes:
                     debugMessage('  User confirmed reload action, calling UsdStage.Reload()')
                     proxyStage.Reload()
+        
+        # Refresh the system lock status of the stage and its sublayers
+        stageFilePath = cmds.getAttr(filePathAttr)
+        fullStageName = GetFullStageNameFromProxyShapeAttr(filePathAttr)
+        cmds.mayaUsdLayerEditor(stageFilePath, edit=True, refreshSystemLock=(fullStageName, True))
```
pierrebai-adsk, line 296: I guess the stageFilePath is really the root layer name we want to refresh?

pierrebai-adsk (approved):
Minor comments, LGTM

plugin/adsk/scripts/AETemplateHelpers.py
AramAzhari-adsk: Yes, it's the root layer path/

lib/usd/ui/layerEditor/mayaCommandHook.cpp
AramAzhari-adsk: They are slightly different because they are leveraging different function calls.

lib/mayaUsd/commands/layerEditorCommand.cpp
AramAzhari-adsk: Good advice! std::quoted looks a lot cleaner. I'll be using that in the upcoming PR.

pierrebai-adsk (approved):
LGTM

plugin/adsk/scripts/AETemplateHelpers.py
```diff
@@ -109,6 +110,16 @@ def GetStageFromProxyShapeAttr(attr):
...
 
+def GetFullStageNameFromProxyShapeAttr(attr):
+    # Helper method which returns the full stage name
+    # First get the stage name from the input attribute.
+    stageName = attr.split('.')[0]
+    # Convert that into a long Maya path so we can get the USD stage.
+    res = cmds.ls(stageName, l=True)
+    fullStageName = res[0]
+    
+    return(fullStageName)
```
seando-adsk, line 121: Why was this function needed? Couldn't you just have called the one above and ignore the stage?
`fullStageName,_ = GetStageFromProxyShapeAttr(filePathAttr)`

AramAzhari-adsk: Fair point. I have an upcoming PR and I am taking note to address this.

seando-adsk: Okay thanks. I'll merge this one then.

AramAzhari-adsk: Thank you

PR #3628, opened 2024-02-22
Use the Using Directive macro

dj-mcg:

### DIFF
--- a/lib/mayaUsd/utils/copyLayerPrims.cpp
+++ b/lib/mayaUsd/utils/copyLayerPrims.cpp
@@ -22,7 +22,7 @@
 
 #include <maya/MGlobal.h>
 
-using namespace PXR_NS;
+PXR_NAMESPACE_USING_DIRECTIVE
 
 namespace {
 


dj-mcg:
Hi folks, I'm used to adding an Assignee (either the ecp bot or more recently myself) in order to kick off the preflight tests, but I can't seem to edit this field for this pull request. Has the process changed again?

seando-adsk:
> Hi folks, I'm used to adding an Assignee (either the ecp bot or more recently myself) in order to kick off the preflight tests, but I can't seem to edit this field for this pull request. Has the process changed again?

Hi Dan, there have been new security measures put in place that limit outside contributors in the Autodesk Org. We are no longer able to add new outside contributors to the repo. But you are supposed to be in the org and are still in the repo so you should be able to edit the assignee field. Maybe something else changed which I'm not aware of. 

The new procedure for outside contributors is to have the code review done first and then an Autodesk member will start the preflight for you. So I guess we'll have to start doing that for your PRs now as well.

Sean

seando-adsk (approved):

dj-mcg:
Thanks @seando-adsk!

PR #3640, opened 2024-02-28
Restrict UFE nodes according to proxyNode's primPath

dj-mcg:
Only create UFE nodes that are ancestors or descendants of the proxy node's primPath. This addresses a performance issue we encountered in our Reference Assembly workflows. For these workflows, there were reference assemblies that were referencing very large Usd scenes, but drawing a very small subset of it. While Maya would only image the subset, UFE was populating every node from the Usd stage.

This caused several performance problems, including:
* viewport -> panel -> perspective slowness due to it traversing all the UFE nodes looking for cameras
* selection with the transform tool

This fix limits creation of UFE nodes to those "relevant" to the prim path

This fix also changes behavior of imaging a proxy node with an invalid prim path. Previously, a non-empty invalid path would cause the entire stage to be imaged. With this change, we image nothing.


dj-mcg:
I fixed the typo in the comment in this branch but it looks like that part of the change is still being identified as a conflict. I would just accept my change in the conflict but I appear not to have permission to resolve it.

dj-mcg:
Also, note that we added a test with the change. This is enabled internally, but I did not enable it in the open source build. I made a naive attempt, but I'm woefully undereducated when it comes to CMake so I gave up after I ran into some import errors with ufe and mayaUsd. It would be a good idea to enable it if you don't mind updating the makefiles to include it. Thank yoU!

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -304,6 +304,15 @@ MayaUsdProxyShapeBase* getProxyShape(const Ufe::Path& path)
     return UsdStageMap::getInstance().proxyShapeNode(path);
 }
 
+SdfPath getPrimPath(const Ufe::Path& path)
```
pierrebai-adsk, line 307: Not sure I like the function name, I think it is misleading. One could think it return the USD prim path corresponding to the given UFE path, but it actually return the root prim path of the corresponding proxy shape. Maybe the function should be named getProxyShapePrimPath?

pierrebai-adsk (changes requested):
I would like the function to be renamed for clarity.

pierrebai-adsk:
> I fixed the typo in the comment in this branch but it looks like that part of the change is still being identified as a conflict. I would just accept my change in the conflict but I appear not to have permission to resolve it.

You need to pull the dev branch and resolve conflicts locally on your comptuer and push your branch again.

lib/mayaUsd/ufe/Utils.cpp
dj-mcg: > Not sure I like the function name, I think it is misleading. One could think it return the USD prim path corresponding to the given UFE path, but it actually return the root prim path of the corresponding proxy shape. Maybe the function should be named getProxyShapePrimPath?

Makes sense

dj-mcg:
> > I fixed the typo in the comment in this branch but it looks like that part of the change is still being identified as a conflict. I would just accept my change in the conflict but I appear not to have permission to resolve it.
> 
> You need to pull the dev branch, merge it in your branch and resolve conflicts locally on your computer and push your branch again.

Will do, thanks - stand by for the commit

lib/mayaUsd/ufe/.nfs806efa1600d06f8c00002121
dj-mcg, line 1: Ugh looks like some local files snuck into this comment. Sorry about that - I'll clean this up in a bit

lib/mayaUsd/ufe/Utils.cpp
pierrebai-adsk: ahah, I saw it too but chose not to mention it since you did not write it. You can leave it as is, as there is no time associated with prims anyway, so it is less confusing, even though I was tempted to suggest the change too.

dj-mcg:
OK the review notes and the conflict have been addressed. Thanks!

pierrebai-adsk (dismissed):

seando-adsk:
@dj-mcg I ran the preflight for you and it failed. Currently because of a security issue the build logs are not being uploaded to github. So I've looked at the logs and included the error here. It is on all the Maya 2022 builds:
```
S:\jenkins\workspace\ECP\ecg-maya-usd\ecg-mayausd-branch-preflight-2022-python3-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\nodes\proxyShapeBase.cpp(1931): error C2039: 'isEmpty': is not a member of 'Autodesk::Maya::OpenMaya20220000::MString'
P:\shared\Artifactory\Windows\rundev-2022\202303271415\runtime\795869a\runTime\include\maya/MString.h(49): note: see declaration of 'Autodesk::Maya::OpenMaya20220000::MString'
```

pierrebai-adsk:
> @dj-mcg I ran the preflight for you and it failed. Currently because of a security issue the build logs are not being uploaded to github. So I've looked at the logs and included the error here. It is on all the Maya 2022 builds:
> 
> ```
> S:\jenkins\workspace\ECP\ecg-maya-usd\ecg-mayausd-branch-preflight-2022-python3-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\nodes\proxyShapeBase.cpp(1931): error C2039: 'isEmpty': is not a member of 'Autodesk::Maya::OpenMaya20220000::MString'
> P:\shared\Artifactory\Windows\rundev-2022\202303271415\runtime\795869a\runTime\include\maya/MString.h(49): note: see declaration of 'Autodesk::Maya::OpenMaya20220000::MString'
> ```

You need to replace isEmpty() by retrieving the length and comparing to zero. Unfortunate, but required for 2022 compat.

dj-mcg:
> > @dj-mcg I ran the preflight for you and it failed. Currently because of a security issue the build logs are not being uploaded to github. So I've looked at the logs and included the error here. It is on all the Maya 2022 builds:
> > ```
> > S:\jenkins\workspace\ECP\ecg-maya-usd\ecg-mayausd-branch-preflight-2022-python3-windows\ecg-maya-usd\maya-usd\lib\mayaUsd\nodes\proxyShapeBase.cpp(1931): error C2039: 'isEmpty': is not a member of 'Autodesk::Maya::OpenMaya20220000::MString'
> > P:\shared\Artifactory\Windows\rundev-2022\202303271415\runtime\795869a\runTime\include\maya/MString.h(49): note: see declaration of 'Autodesk::Maya::OpenMaya20220000::MString'
> > ```
> 
> You need to replace isEmpty() by retrieving the length and comparing to zero. Unfortunate, but required for 2022 compat.

OK, thanks! I should've tried building it against 2022 internally before pushing -- sorry about that!

pierrebai-adsk (approved):

seando-adsk:
Added do not merge label, while we test this internally.

seando-adsk:
Internal Autodesk QA tested these changes and confirmed it with design. Okay for merging now.

PR #3646, opened 2024-03-04
Adds the ability to lock layers and sublayers at once

AramAzhari-adsk:
Adds the ability to lock layers and sublayers at once

- Updated the lockLayer command with the ability to lock sublayers.
- Added the LayerEditorWindow query command layerHasSubLayers to track if a layer contains sublayers.
- Added a context menu item for the layer editor to Locking a layer and its sublayers
- Updated the layerLocking test sample to include a sublayer.
- Updated RefreshSystemLockLayer and LockLayer to handle Undo correctly
    - RefreshSystemLockLayer now handles changing multiple layer permissions with a single Undo command.
    - Fixed Layer Lock undo when Undoing from Unlocked -> Locked
- Added a test unit to check lock status of sublayers.
- Updated the command documentation


lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -901,24 +930,42 @@ class RefreshSystemLockLayer : public BaseCmd
             _refreshLayerSystemLock(layer);
         }
 
+        // Execute lock commands
```
pierrebai-adsk, line 933: Do and undo stop at the first error and execute in the same order. For composite commands, the usual way to handle it is to record which were done so undo only undoes those that were actually successful. Also, undo usually execute in reverse order as do. Might be worthwhile to either execute as much as possible (execute all, return if any failed) or to undo on error.

For locking, these things are less critical, I don't think there will be anything bad so you can keep the simpler code that you currently have, but it is something to keep in mind.

lib/usd/ui/layerEditor/layerTreeView.cpp
```diff
@@ -356,7 +362,7 @@ void LayerTreeView::onLockLayer(const QString& undoName) const
 
     UndoContext context(params.commandHook, params.name);
     for (auto item : *params.selection) {
```
pierrebai-adsk, line 364: Any reason we go through the selection for lock layer and sub-layers, but not for lock-layer above?

pierrebai-adsk (approved):
LGTM

lib/mayaUsd/commands/layerEditorCommand.cpp
AramAzhari-adsk: Great point!

In this implementation, the layer hierarchy is flattened and since locking only affects the local variables of each layer, the lock / unlock doesn't have to be in an order. I'll keep the error checking in mind going forward.

lib/usd/ui/layerEditor/mayaLayerEditorWindow.cpp
```diff
@@ -207,6 +208,16 @@ void MayaLayerEditorWindow::lockLayer()
     }
 }
 
+void MayaLayerEditorWindow::lockLayerAndSubLayers()
+{
+    auto item = treeView()->currentLayerItem();
+    if (item != nullptr) {
+        QString name = item->isLocked() ? "Unlock Layer and Sublayers" : "Lock Layer and Sublayers";
```
seando-adsk, line 215: I had a quick look and it seems like this string is used for the undo name. Is that right? Is there a reason then you didn't also put these strings in the stringResources.h?

AramAzhari-adsk: This seems to be the pattern for the menu items in this file. I've made a EMSUSD-1096 to address all the strings in this file.

seando-adsk: Okay thanks for checking.

PR #3648, opened 2024-03-05
EMSUSD-1057 - Add support for getting native type via metadata

stefanminning-autodesk:
This PR adds support for getting an AttributeDef's native type via the metadata. Requires UFE 0.6.1.

### DIFF
--- a/test/lib/ufe/testShaderNodeDef.py
+++ b/test/lib/ufe/testShaderNodeDef.py
@@ -242,11 +242,8 @@ class ShaderNodeDefTestCase(unittest.TestCase):
         self.assertIsNotNone(connections)
         self.assertEqual(len(connections.allConnections()), 0)
 
+    @unittest.skipIf(os.getenv('UFE_HAS_NATIVE_TYPE_METADATA', 'NOT-FOUND') not in ('1', "TRUE"), 'Test only available if UFE AttributeDef \'NativeType\' metadata exists.')
     def testNativeTypeMetadata(self):
-        # Make sure the feature exists before attempting to test.
-        if not os.environ.get('UFE_HAS_NATIVE_TYPE_METADATA'):
-            return
-        
         type = "ND_constant_color3"
         nodeDefHandler = self.getNodeDefHandler()
         nodeDef = nodeDefHandler.definition(type)


pierrebai-adsk (approved):

lib/mayaUsd/ufe/UsdShaderAttributeDef.cpp
```diff
@@ -170,6 +170,13 @@ static const MetadataMap _metaMap = {
 Ufe::Value UsdShaderAttributeDef::getMetadata(const std::string& key) const
 {
     TF_DEV_AXIOM(fShaderAttributeDef);
+
+#if (UFE_PREVIEW_VERSION_NUM >= 6000)
+    if (key == Ufe::AttributeDef::kNativeType) {
+        return Ufe::Value(fShaderAttributeDef->GetType().GetString());
+    }
```
JGamache-autodesk, line 177: This is the Sdr type, which is OK but not perfect.
To compare, run this in the Maya script editor:
```python
from pxr import Sdr
r = Sdr.Registry()
n = r.GetNodeByIdentifier("ND_constant_color3")
i = n.GetShaderOutput("out")
print("Color3 Sdr type:", i.GetType())
print("Color3 Sdf type:", (str(i.GetTypeAsSdfType()[0]), i.GetTypeAsSdfType()[1]))
n = r.GetNodeByIdentifier("ND_constant_filename")
i = n.GetShaderOutput("out")
print("filename Sdr type:", i.GetType())
print("filename Sdf type:", (str(i.GetTypeAsSdfType()[0]), i.GetTypeAsSdfType()[1]))
n = r.GetNodeByIdentifier("ND_lama_add_edf")
i = n.GetShaderOutput("out")
print("EDF Sdr type:", i.GetType())
print("EDF Sdf type:", (str(i.GetTypeAsSdfType()[0]), i.GetTypeAsSdfType()[1]))
```
Will return:
```
Color3 Sdr type: color
Color3 Sdf type: ('color3f', '')
filename Sdr type: string
filename Sdf type: ('asset', '')
EDF Sdr type: EDF
EDF Sdf type: ('token', 'EDF')
```
So, here, you should do this instead:
```C++
    const auto sdfTypeTuple = fShaderAttributeDef->GetTypeAsSdfType();
    if (sdfTypeTuple.second.empty()) {
        return sdfTypeTuple.first.GetAsToken().GetString();
    } else {
        return sdfTypeTuple.second();
    }
```

JGamache-autodesk (changes requested):

lib/mayaUsd/ufe/UsdShaderAttributeDef.cpp
```diff
@@ -173,7 +173,13 @@ Ufe::Value UsdShaderAttributeDef::getMetadata(const std::string& key) const
 
 #if (UFE_PREVIEW_VERSION_NUM >= 6000)
     if (key == Ufe::AttributeDef::kNativeType) {
-        return Ufe::Value(fShaderAttributeDef->GetType().GetString());
+        // We return the Sdf type as that is more meaningful than the Sdr type.
+        const auto sdfTypeTuple = fShaderAttributeDef->GetTypeAsSdfType();
+        if (!sdfTypeTuple.second.size()) {
```
JGamache-autodesk, line 178: Oh! This is a TfToken, so you want to use `IsEmpty()` instead of the `empty()` I suggested. Using `size()` works, but is frowned upon by clang-tidy as not being as efficient.

JGamache-autodesk (approved):

pierrebai-adsk (approved):

JGamache-autodesk (dismissed):

lib/mayaUsd/ufe/UsdShaderAttributeDef.cpp
seando-adsk: @stefanminning-autodesk I heard that this metadata might get backported into Ufe v5. If that is true, then this ifdef check (and the others) should be modified to not be version dependent.  See https://github.com/Autodesk/maya-usd/blob/dev/cmake/modules/FindUFE.cmake for other examples of this where we look inside of header files (from Ufe) for a feature and set a flag if we find the new feature. That way if you backport to Ufe v5 no other MayaUsd changes will be required.

test/lib/ufe/testShaderNodeDef.py
```diff
@@ -242,8 +242,11 @@ def testNodeCreation(self):
         self.assertIsNotNone(connections)
         self.assertEqual(len(connections.allConnections()), 0)
 
-    @unittest.skipUnless(ufeUtils.ufeFeatureSetVersion() >= 6, 'NativeType metadata is only available in UFE v6 or greater')
     def testNativeTypeMetadata(self):
+        # Make sure the feature exists before attempting to test.
+        if not os.environ.get('UFE_HAS_NATIVE_TYPE_METADATA'):
+            return
+
```
seando-adsk, line 249: This is okay, but it would be better as:

`@unittest.skipIf(os.getenv('UFE_HAS_NATIVE_TYPE_METADATA', 'NOT-FOUND') not in ('1', "TRUE"), 'Test only available if UFE AttributeDef 'NativeType' metadata exists.')`

seando-adsk (approved):

test/lib/ufe/testShaderNodeDef.py
JGamache-autodesk, line 249: You can still use skipUnless:
```
    @unittest.skipUnless(os.getenv('UFE_HAS_NATIVE_TYPE_METADATA', 'FALSE') == 'TRUE', 'Test requires NativeType metadata.')
```

JGamache-autodesk (dismissed):

seando-adsk (approved):

JGamache-autodesk (approved):

JGamache-autodesk:
Ah, yes, preflight... Sorry! Ready for merge now.

PR #3649, opened 2024-03-06
EMSUSD-877 UI tweaks

pierrebai-adsk:
- Move add-layer button left to the border of the panel.
- Don't show context-menu mouse cursor over action buttons.
- Make hover highlight even over muted rows.
- Make hover highlight less dark.
- Rename some paint context flag to be clearer.
- Move "USD Stage" label to the left border.
- Don't draw text under icons with hovering.
- Make text lighter when selected.
- Simplify and centralize the action icon rectangle calculations.
- Improve arrow appearance: smaller, more square, in 6x6 square.
- Adjustment of spacing.
- Remove unnecessary error log when a layer cannot be found. The UI already shows a warning icon.

image


lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -140,37 +140,15 @@ void LayerTreeItem::populateChildren(RecursionDetector* recursionDetector)
     for (auto const path : subPaths) {
         std::string actualPath = SdfComputeAssetPathRelativeToLayer(_layer, path);
         auto        subLayer = SdfLayer::FindOrOpen(actualPath);
-        if (subLayer) {
```
pierrebai-adsk, line 143: The difference in the if/else was merely to print error message in stdout, which are not useful and we already show error icon in the UI. Simplified the code.

lib/usd/ui/layerEditor/layerTreeItemDelegate.cpp
```diff
@@ -23,34 +23,63 @@
 
 namespace {
 
+using ItemPaintContext = UsdLayerEditor::LayerTreeItemDelegate::ItemPaintContext;
+
+bool isInaccessible(const ItemPaintContext& ctx)
```
pierrebai-adsk, line 28: These function are used to control the opacity in various scenarios.

PR #3651, opened 2024-03-06
EMSUSD-79 - Use UsdPreviewSurface as the fallback shader

samuelliu-adsk:
Since the default light intensity change is made to Maya2024 and later, I added a check for maya version and decide which surface shader to use. 
Maya added an user pref for default light intensity before and decided to remove it so I'll respect that change. User can still toggle the light intensity in Hardware Renderer settings


lib/mayaUsd/render/vp2RenderDelegate/render_delegate.cpp
```diff
@@ -385,8 +385,12 @@ class MShaderCache final
             const MHWRender::MShaderManager* shaderMgr
                 = renderer ? renderer->getShaderManager() : nullptr;
             if (TF_VERIFY(shaderMgr)) {
+#if MAYA_API_VERSION >= 20240000
+                shader = shaderMgr->getFragmentShader("UsdPreviewSurface", _structOutputName, true);
```
JGamache-autodesk, line 389: Unsure. Having visual results in a unit test would help compare the results. I suspect you want to
- crank the roughness to 1.0
- enable "use specular workflow"
- keep specular color untouched (default is black)

To make sure you end up with a shading that is as flat as the default shader.

JGamache-autodesk (changes requested):

PR #3654, opened 2024-03-08
Un-shared stage and its sublayers will appear as system lock

AramAzhari-adsk:
Unshared stages and sublayers will appear as system-locked.

This is used for third party plugins that prefer to manipulate the content of the owned layers through their api while still present those layers in Maya USD's layer editor.

Note that by design the un-shared layers may visually appear as system-locked but the lock state of the layers are managed independent of the un-shared state. Therefore, if any layers are actually locked prior to getting unshared, they will still retain that state when they are shared again.


pierrebai-adsk (approved):

AramAzhari-adsk:
Just merged remote branch to resolve the conflict.

pierrebai-adsk (approved):

AramAzhari-adsk:
Fixed a merge regression.

pierrebai-adsk (approved):

PR #3655, opened 2024-03-12
EMSUSD-961 first pass at updating the export plugin configuration UI

pierrebai-adsk:
- Each export plugin now has their own UI with a checkbox and options button.
- Each can apply their options.
- The enabled plugins are remembered.
- The UI logic is updated.
- Add hasExportUI and showExportUI flags to the MayaUSDListJobContexts command.
- Add a exportUICallback to the export plugin context info.
- Add option icons.
- Make the export plugin UI work.
- Add the SetExportOptionsUI function to register a UI callback for the export plugins in the UsdMayaJobContextRegistry.
- Add a macro named REGISTER_EXPORT_JOB_CONTEXT_UI_FCT to define the UI callback in C++.
- Add Python wrapper to allow registering a Python function as the UI callback.
- Use the registered callback in the export dialog to show the UI and update the export settings from the return value of the callback.
- Pass the name of the MEL UI element that spawned the export options UI.
- Make sure export plugin are always listed in alphabetical order.
- Don't accidentally enable all options buttons when there are no plugin enabled.
- Don't update UI from the options dialog when the plugin is disabled.
- Don't show options button when plugin does not have options.
- Add section tooltip.
- Make each row thinner by making icon smaller.
- Add UI options example in C++ plugin.
- Fix the dictionary value parsing code to support double-precision floating point values.
- Change default dictionary to contain values instead of strings.
- Both of these fixes the mysterious bug that frame stride could not be edited as a flaoting-point value.
- Button to enable export plugin is now a toggle with nicer images.
- Add an example of an export plugin in Python.
- The example supports the new UI options.
- It it necessary to set the PYTHONPATH and MAYA_PXR_PLUGINPATH_NAME to the example folder for it to load.
- Make each plugin listed in the export UI appear as a separate rectangle item.
- Save and load the settings controlled by export plugin in the example.
- Document how the export plugin works and can be written in Python.


lib/mayaUsd/commands/baseListJobContextsCommand.cpp
```diff
@@ -48,9 +49,30 @@ _GetInfo(const MArgDatabase& argData, const char* optionName)
     return emptyInfo;
 }
 
+MString convertDictionaryToText(const VtDictionary& settings)
```
pierrebai-adsk, line 52: This code was simply moved from within the functions below, which had multiple copies of it.

lib/mayaUsd/resources/scripts/mayaUsdMergeToUSDOptions.py
```diff
@@ -255,28 +255,28 @@ def getDefaultMergeToUSDOptionsDict():
     Retrieves the default merge-to-USD options.
     """
     return {
```
pierrebai-adsk, line 257: Making the default have real value provides better validation of input values.

lib/mayaUsd/utils/util.cpp
```diff
@@ -1918,6 +1918,8 @@ VtValue _ParseArgumentValue(const std::string& value, const VtValue& guideValue)
     // The export UI only has boolean and string parameters.
     if (guideValue.IsHolding<bool>()) {
         return VtValue(TfUnstringify<bool>(value));
+    } else if (guideValue.IsHolding<float>()) {
```
pierrebai-adsk, line 1921: Missing float handling was making it impossible to set the frameStride value from Python.

lib/mayaUsd/fileio/jobContextRegistry.h
```diff
@@ -75,6 +75,21 @@ class UsdMayaJobContextRegistry : public TfWeakBase
     /// Enabler function, returns a dictionary containing all the options for the context.
     typedef std::function<VtDictionary()> EnablerFn;
 
+    // Show a UI, a modal dialog, for the plugin options.
```
AramAzhari-adsk, line 78: Is this the option dialog for each entry that needs it?
If so, does it have to be modal?

pierrebai-adsk: Yes, the designer would prefer that it be modal.

lib/mayaUsd/utils/util.cpp
AramAzhari-adsk: Does this hide the double in the line 1923

pierrebai-adsk: No, the type must natch exactly

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -318,6 +274,319 @@ global proc mayaUsdTranslatorExport_MeshCB() {
...
+        // layout within which each plugin is a separate layout separated by
+        // a thin 1-pixel spacing.
+        //
+        // Note: setting the background color of a panel changes the color of
+        //       text in the panel. So rather than set the background color of
+        //       the columnLayout, we insert thin layouts around the plugin
+        //       row. So we insert one on top and bottom and on each side, and
+        //       these have a darker color to simulate a border.
+
+        int $width = 400;
```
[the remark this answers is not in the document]
pierrebai-adsk: All other MEL code use absolute value, I assume the command would scale them internally. QA will tell us if it looks right on MacOSX in highDPI.

AramAzhari-adsk (approved):

lib/mayaUsd/resources/scripts/mayaUsdMergeToUSDOptions.py
seando-adsk: I just have one question before merging? Does that have any impact on users, such as are any of these values corresponding to optionVar's and now by changing their value the users saved values won't work? Or are these values used in any MEL scripting command and if a user has a MEL script using that command it will fail now since these values are no longer strings?

pierrebai-adsk: No, the values are saved all together in one long option var string. The purpose of giving them now their right type is for the validation of input and conversion from string to dictionary (either Python or USD VtDictionary).

pierrebai-adsk: The conversion using type is done in C++ and in mayaUsdOptions.py in the function _convertTextToType()

AramAzhari-adsk (dismissed):

seando-adsk (dismissed):

lib/mayaUsd/resources/scripts/mayaUsdMergeToUSDOptions.py
pierrebai-adsk: Ahah! Of course, the options in array form did not want to be converted to an array... These things are not obvious...

AramAzhari-adsk (approved):

PR #3656, opened 2024-03-12
EMSUSD-1045 : Add support for doubleSided attribute on USD prims in VP2 delegate

debloip-adsk:
This PR adds support for the "Double Sided" attribute on USD prims in the VP2 delegate. Both UsdPreviewSurface and MaterialX materials are supported.


lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -3005,6 +3005,7 @@ MHWRender::MShaderInstance* HdVP2Material::CompiledNetwork::_CreateMaterialXShad
         }
 
         shaderInstance = shaderMgr->getFragmentShader(fragmentName, "outColor", true);
+        shaderInstance->addInputFragment("NwFaceCameraIfNAN", "output", "Nw");
```
debloip-adsk, line 3008: This line is MaterialX-specific, UsdPreviewSurface fragments already reference the `NwFaceCameraIfNAN` fragment in their XML.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -3192,7 +3210,7 @@ void HdVP2Material::CompiledNetwork::_UpdateShaderInstance(
 
     const bool matIsTransparent = _IsTransparent(mat);
     if (matIsTransparent != _surfaceShader->isTransparent()) {
-        _surfaceShader->setIsTransparent(matIsTransparent);
+        SetShaderIsTransparent(matIsTransparent);
```
debloip-adsk, line 3213: Transparency was previously only being set for the surface shader (whereas all other properties were set for all shader variants), which caused issues when using the front face shader.

test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateUSDPreviewSurface.py
```diff
@@ -247,6 +247,29 @@ def testImportDisplayColor(self):
...
+        mayaUtils.setBasicCamera(3)
+        self.assertSnapshotClose('doubleSided_enabled_back.png')
+
+        cubePrim = stage.GetPrimAtPath("/Cube1")
+        cubePrim.GetAttribute('doubleSided').Set(False)
+
+        mayaUtils.setBasicCamera(20)
+        self.assertSnapshotClose('doubleSided_disabled_front.png')
+        mayaUtils.setBasicCamera(3)
+        self.assertSnapshotClose('doubleSided_disabled_back.png')
```
debloip-adsk, line 272: The test code for UsdPreviewSurface and MaterialX is 99% the same, however I don't really see where it could be sensibly extracted. It's not really a utility, they just happen to have the same steps. I'd have put both in a dedicated test file for double-sidedness, but MaterialX is optional and its tests separate, so I split it in two.

test/testUtils/mayaUtils.py
```diff
@@ -276,6 +277,14 @@ def activeModelPanel():
         if cmds.modelEditor(panel, q=1, av=1):
             return panel
 
+def setBasicCamera(offset=BASIC_CAMERA_DEFAULT_OFFSET):
+    """
+    Sets the camera in a similar position and rotation to the default ones.
+    An offset can be specified to change the distance of the camera from the origin.
+    """
+    cmds.setAttr('persp.translate', offset, .75 * offset, offset, type='float3')
+    cmds.setAttr('persp.rotate', -30, 45, 0, type='float3')
```
debloip-adsk, line 286: This method is essentially the same as `setBasicCam` from `mtohUtils`, but AFAIK Mtoh is essentially outdated in this repo, so I ported it here. Note that I did change the default value from 24 to 28 though, as that seems like the default value when creating a new scene.

test/lib/mayaUsd/render/vp2RenderDelegate/VP2RenderDelegateMaterialXTest/baseline/transparencyScene.png
debloip-adsk, line 1: This file needed to be updated as by default, prims are not double-sided. Previously this was ignored and backfaces were always rendered, but they are now culled.

debloip-adsk, line 1: This file needed to be updated as by default, prims are not double-sided. Previously this was ignored and backfaces were always rendered, but they are now culled.

debloip-adsk:
FYI @lilike-adsk @ppt-adsk @lanierd-adsk @roopavr-adsk

PR #3663, opened 2024-03-14
EMS1056 vertex processing robustness

pierrebai-adsk:
Fix reading past the end of the buffer causing incorrect rendering and crash later on.

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
@@ -1010,6 +1010,11 @@ void HdVP2Mesh::Sync(
             for (size_t i = 0; i < numFaceVertexIndices; i++) {
                 const int sceneFaceVtxId = faceVertexIndices[i];
 
+                // Scene index is actually greater than anticipated, increase the buffer size.
+                if (sceneFaceVtxId >= _meshSharedData->_sceneToRenderingFaceVtxIds.size()) {
+                    _meshSharedData->_sceneToRenderingFaceVtxIds.resize(sceneFaceVtxId + 1, -1);
+                }
+
                 int renderFaceVtxId = _meshSharedData->_sceneToRenderingFaceVtxIds[sceneFaceVtxId];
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1010,6 +1010,11 @@ void HdVP2Mesh::Sync(
             for (size_t i = 0; i < numFaceVertexIndices; i++) {
                 const int sceneFaceVtxId = faceVertexIndices[i];
```
pierrebai-adsk, line 1012: See line 1005, which created the buffer filled with -1, assuming the number of vertices would reflect the highest possible index. Sometimes, those indices can be greater, probably when geometry is processed and some vertices are dropped?

Whatever is the cause, we need to validate the input data when creating the rapid lookup tables.

lib/mayaUsd/render/vp2RenderDelegate/mesh.cpp
```diff
@@ -1010,6 +1010,11 @@ void HdVP2Mesh::Sync(
...
 
+                // Scene index is actually greater than anticipated, increase the buffer size.
+                if (sceneFaceVtxId >= _meshSharedData->_sceneToRenderingFaceVtxIds.size()) {
+                    _meshSharedData->_sceneToRenderingFaceVtxIds.resize(sceneFaceVtxId + 1, -1);
```
AramAzhari-adsk, line 1015: Does this trigger a resize only on the first index i ?

pierrebai-adsk: I can trigger a resize whenever the scene face vertex ID is too large.

AramAzhari-adsk (approved):

AramAzhari-adsk (approved):

pierrebai-adsk:
The PF passed. One MacOSX passed, all tests passed, but for some reason iyt aborted after all tests were run without reporting its results.

PR #3667, opened 2024-03-19
EMSUSD-1120 - Stage loading problems using MEL proc

seando-adsk:
EMSUSD-1120 - Stage loading problems using MEL proc
* Make sure to init the optionvars used by the MEL command.

### DIFF
--- a/plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
+++ b/plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
@@ -273,5 +273,9 @@ global proc mayaUsd_createStageFromFile()
 
 global proc string mayaUsd_createStageFromFilePath(string $usdFile)
 {
+    // This proc is used by the unit tests (which run in standalone mode).
+    // In that case we need to make sure the optionVars are initialized as
+    // they are queried in doCreateStage().
+    setOptionVars(false);
     return doCreateStage($usdFile);
 }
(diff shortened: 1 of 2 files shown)


plugin/adsk/scripts/mayaUsd_createStageFromFile.mel
```diff
@@ -273,5 +273,9 @@ global proc mayaUsd_createStageFromFile()
 
 global proc string mayaUsd_createStageFromFilePath(string $usdFile)
 {
+    // This proc is used by the unit tests (which run in standalone mode).
+    // In that case we need to make sure the optionVars are initialized as
+    // they are queried in doCreateStage().
+    setOptionVars(false);
```
seando-adsk, line 279: So what happens is on line 141 it would do:
`string $ppath = `optionVar -q stageFromFile_primPath`;`
If we are running in standalone (say from a unit test) the options vars would not have been init, and in Maya when you query an option var that doesn't exist it returns 0. The MEL code above converts that to a string and you get "0". That would then we set as the primPath on the stage. In Dan's new code this is not a valid primPath so the compute would bail early. With this fix the option vars are init and it will set "" as the primpath which converts to the root prim "/" and the compute works.

test/lib/testMayaUsdCreateStageCommands.py
```diff
@@ -77,6 +77,18 @@ def testCreateStageFromFile(self):
...
+        stage = mayaUsd.ufe.getStage(shapeNode)
+        self.assertTrue(stage)
+
+        # This stage has five layers.
+        layerStack = stage.GetLayerStack()
+        self.assertEqual(5, len(layerStack))
+
+        # Finally, verify that we can get at least one of the ball prims.
+        ballPrim = stage.GetPrimAtPath('/Room_set/Props/Ball_1')
+        self.assertTrue(ballPrim.IsValid())
```
seando-adsk, line 90: Added some test cases to ensure the stage was correctly loaded.

JGamache-autodesk (approved):

pierrebai-adsk (approved):

PR #3669, opened 2024-03-20
EMSUSD-1098 Fixed saving vs locked layers

pierrebai-adsk:
Also EMSUSD-1097, EMSUSD-1099 and EMSUSD-1100

EMSUSD-1097 Refactor anonymous layer saving. There was code that was nearly identical between the Layer Editor item saveAnonymousLayer function, the saveSubLayer function and the generic saveAnonymousLayer function in utilSerialization.h. Refactor the layer editor item to use the generic one instead, avoiding having to fix problems in two places.

- Remove the saveSubLayer function that is now entirely redundant.
- Refactor the LayerTreeItem function to use the generic saveAnonymousLayer function.
- All the code left in the LayerTreeItem is now specific the the layer editor or its session state.
- Add a FileBackup helper class in the utilFileSystem to backup a file and possibly restore it if the new file is not commited as valid.
- Add a parameter to the generic saveAnonymousLayer function to return an optional error message.
- Improve error messages when saving
- Prevent saving system-locked layers.
- Add helper functions to transfer the muted and locked status of anonymous layers to their new layers.

EMSUSD-1100 preserve the locked and muted status of layer when saved. When transitioning from anonymous to non-anonymous, update the muted and locked status of the layers.

EMSUSD-1098 don't allow saving system-locked layer Prevent saving when double-clicking on a system-locked layer.

EMSUSD-1099 fix parent layer when converting anonynous layers. The parent layer that was passed to saveAnonymousLayer when converting anonymous layers into saved layers was incorrect. It was passing the parent of the parent instead. This lead to the stage root layer getting replaced by its oen sub-layer!

Add unit tests and Python wrappers

- Add Python wrappers to lock, unlock and verify lock on layers.
- Write unit tests to verify the behavior of locked layers.
- Improve the tests for muted layers.


lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -159,27 +159,19 @@ void convertAnonymousLayersRecursive(
...
-        parentPtr._layerParent = nullptr;
-        parentPtr._proxyPath = basename;
-    } else if (stage->GetSessionLayer() == layer) {
-        parentPtr._layerParent = nullptr;
-        parentPtr._proxyPath = basename;
-    } else {
-        parentPtr._layerParent = layer;
-        parentPtr._proxyPath = basename;
-    }
-
```
pierrebai-adsk, line 173: The code above was creating the parent layer description of the given layer, but this was used to describe the parent of the sub-layer! So when processing the sub-layer, it was using its grand-parent instead. This led to the confusing behaviour that an anonymous sub-layer of the root would become the root.

lib/mayaUsd/utils/utilFileSystem.cpp
```diff
@@ -750,3 +750,56 @@ std::string UsdMayaUtilFileSystem::pathFindExtension(std::string& filePath)
     ghc::filesystem::path ext = p.extension();
     return ext.string();
 }
+
+UsdMayaUtilFileSystem::FileBackup::FileBackup(const std::string& filename)
```
pierrebai-adsk, line 754: Automatic backup/restore of a file. Taken from code that was in the layer editor and generalized to be reusable.

lib/mayaUsd/utils/utilSerialization.cpp
```diff
@@ -92,6 +96,41 @@ void populateChildren(
     recursionDetector->pop();
 }
 
+void updateMutedLayers(
```
pierrebai-adsk, line 99: Update muted and locked layers when converted from anonymous to disk-based.

lib/mayaUsd/utils/utilSerialization.cpp
```diff
@@ -321,52 +360,87 @@ SdfLayerRefPtr saveAnonymousLayer(
...
     SdfLayerRefPtr  anonLayer,
     const PathInfo& pathInfo,
     LayerParent     parent,
-    std::string     formatArg)
+    std::string     formatArg,
+    std::string*    errorMsg)
 {
-    // TODO: the code below is very similar to LayerTreeItem::saveAnonymousLayer().
-    //       When fixing bug here or there, we need to fix it in the other. Refactor to have a
-    //       single copy.
```
pierrebai-adsk, line 340: Now both code paths are merged here, which will simplify future maintenance.

AramAzhari-adsk (approved):

PR #3671, opened 2024-03-20
EMSUSD-1114 - Adds a callback for RefreshSystemLock when a layer's lock status is changed

AramAzhari-adsk:
- Adds a callback system that allow registering for changes coming from either commands or UI interactions. Each callback can receive a context (PXR_NS::VtDictionary) as well as callback data (PXR_NS::VtDictionary).
- Added onRefreshSystemLock callback to know when a layer's lock status is changed due to disk write permission checeks.
- Added a test case for the callback
- Added a new LayerLocking.md document encompassing Layer Locking how-tos to be used through C++ or scripting with the included examples.


lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -1051,6 +1060,28 @@ class RefreshSystemLockLayer : public BaseCmd
         }
     }
 
+    void _notifySystemLockIsRefreshed()
+    {
+        PXR_NS::VtDictionary callbackContext;
+        callbackContext["proxyShapePath"] = PXR_NS::VtValue(_proxyShapePath.c_str());
+        PXR_NS::VtDictionary callbackData;
+
+        UsdUfe::UICallback::Ptr dstCallback = UsdUfe::getUICallback(TfToken("onRefreshSystemLock"));
```
pierrebai-adsk, line 1069: Maybe check this before creating the first dictionary. I expect that there won't be callbacks for many users.

test/lib/testMayaUsdLayerEditorCommands.py
```diff
@@ -604,6 +604,31 @@ def testLayerLockWritePermission(self):
...
+        # 3- Attach callbacks to capture any system-lock changes due to refreshSystemLock
+        def refreshSystemLockCallback(context, callbackData):
+            layerIds = callbackData.get('affectedLayerIds')
+            for layerId in layerIds:
+                mel.eval(f'mayaUsdLayerEditor -edit -lockLayer 2 0 "{proxyShapePath}" "{layerId}"')
+        
+        from usdUfe import registerUICallback
+        mayaUsd.lib.registerUICallback('onRefreshSystemLock', refreshSystemLockCallback)        
+        # 4- Refreshing the system lock should remove the lock but also get re-locked by refreshSystemLockCallback
+        cmds.mayaUsdLayerEditor(topLayer.identifier, edit=True, refreshSystemLock=(proxyShapePath, 1))
```
pierrebai-adsk, line 630: One issue I can see is that the layer are system-locked, so if this call fails, they won't be refreshed, but the test will pass because the lack of refresh means they will still be locked.

Maybe instead of re-locking, the callback could capture the list of layer IDs it receive and assert that you have receive the expected layer ID instead of their lock state? (Might also then assert that they did get unlocked?)

pierrebai-adsk (dismissed):
LGTM, the test could be modified a bit if you want.

test/lib/testMayaUsdLayerEditorCommands.py
AramAzhari-adsk: Good call, I'll update the test to check for the ids.

pierrebai-adsk (approved):

AramAzhari-adsk:
Only one platform failed due to server errors. Marking it ready-for-merge

PR #3677, opened 2024-03-25
MAYA-132000: implement Ufe::Camera::renderable() for Maya Master (2026)

compso:
In UFE 6 a new method is added to return if the given camera is renderable. Here I have implemented it so it will return True if the purpose of the `UsdCamera` object is either "default" or "render".

This in turn allows USD cameras to be rendered in May Batch renders if the current renderer supports UFE cameras

### DIFF
--- a/lib/mayaUsd/nodes/layerManager.cpp
+++ b/lib/mayaUsd/nodes/layerManager.cpp
@@ -600,7 +600,14 @@ bool LayerDatabase::saveLayerManagerSelectedStage()
     if (!status)
         return false;
 
-    selectedStageHandle.setString(getSelectedStage().c_str());
+    // Note: when empty, we clear the the selected stage attribute so that the
+    //       attribute does not get written to the scene, which improve backward
+    //       compatibility.
+    const std::string stageName = getSelectedStage();
+    if (stageName.size() > 0)
+        selectedStageHandle.setString(stageName.c_str());
+    else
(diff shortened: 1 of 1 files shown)


lib/usdUfe/ufe/UsdCamera.h
```diff
@@ -90,6 +90,9 @@ class USDUFE_PUBLIC UsdCamera : public Ufe::Camera
     Ufe::ProjectionUndoableCommand::Ptr projectionCmd(Ufe::Camera::Projection projection) override;
     Ufe::Camera::Projection             projection() const override;
 
+#ifdef UFE_V6_FEATURES_AVAILABLE
+    bool              renderable() const override;
+#endif
```
seando-adsk, line 95: Having the UFE_V6 is not correct. This method was only added in a specific v0.6.x version. Have a look at https://github.com/Autodesk/maya-usd/blob/dev/cmake/modules/FindUFE.cmake#L202 which is an example of where we look for the new method in the header file and then set a cmake var. Search for that cmake var and you'll see where it gets turned into a define for the h/cpp and an env var for the tests. A test is something you need to add as well. You can add something to the existing testCamera.py

seando-adsk (changes requested):
Ufe version check needs adjustment and a unit test added.

seando-adsk (changes requested):
@compso You still need to add a unit test.

PR #3678, opened 2024-03-25
Fix conflict in shaders with MaterialX 1.38.9

dgovil:
MaterialX 1.38.9 has incorporated this commit (https://github.com/AcademySoftwareFoundation/MaterialX/commit/eae062a038bcd8f8172b1725e93e35641dc73f61) that causes a conflict with a similar change that is inside Maya USD resulting in Pixel shader compilation errors;

```
ERROR: 0:1092: Function 'mx_latlong_compute_lod' redefines an existing function
```

This PR resolves that by renaming the Maya one to a non-conflicting name. This seemed easiest rather than trying to detect the MaterialX version and handling it accordingly.

Without this PR, Maya fails to generate any MaterialX based shaders.


JGamache-autodesk (approved):

JGamache-autodesk:
Preflight failed on `testUsdExportPackage` in 2 configs. @seando-adsk I think this fix is ready for merge.

PR #3688, opened 2024-03-29
EMSUSD-1039 Adding a C++ export plugin example

pierrebai-adsk:
C++ example plugins:
- Wrote working example of C++ import and export plugin with UI.
- Add C++ plugin example code and CMake files.
- Add way to find the C++ plugin via MAYA_PLUG_IN_PATH, finding it as a Maya plugin.
- Add way to find the C++ plugin via MAYA_PXR_PLUGINPATH_NAME, finding ist as a USD plugin.

Python example plugins:
- Change Python plugin JSON to load both plugin. Seems USD does not like having more than one Python plugin in a single JSON.
- Overlay the forced settings over the given settings in the Python example.
- Catch exception when accessing a settings that may not exists in Python.
- The Python import plugin example use MEL-style UI creation.

Documentation:
- Write documentation about the C++ import and export plugins.
- Add link to C++ doc from main doc.
- Document that we recommend not forcing existing settings unless specific values are required.

Code fixes:
- Add a macro to register the import plugin UI.
- Fix registry code to not lose the export info when adding the import info and vis--versa.
- Fix import job dictionary decoding to use the correct guiding dictionary.
- Fix typos in the example Python plugins.
- Allow not reporting errors when converting text to VtDictionary when a value is not known to the guiding dictionary.

Helper functions for the C++ example plugins:
- Add a helper function to convert a VtDictionary to JsObject.
- Add helper function to create UI to simplify and clarify the C++ example plugins.
- Put all helper functions into a Helpers namespace to make clear where the function come from in the C++ example plugins.


lib/mayaUsd/fileio/jobContextRegistry.cpp
```diff
@@ -82,10 +82,9 @@ void UsdMayaJobContextRegistry::RegisterExportJobContext(
                     itFound->niceName.GetText());
             }
             // Fill the export part:
-            ContextInfo updatedInfo {
-                key,        itFound->niceName,          TfToken(description),
-                enablerFct, itFound->importDescription, itFound->importEnablerCallback
-            };
+            ContextInfo updatedInfo(*itFound);
```
pierrebai-adsk, line 85: Old code was losing the import context UI callback. New code preserve all already registered info automatically.

lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -909,8 +909,15 @@ MStatus UsdMayaJobExportArgs::GetDictionaryFromEncodedOptions(
                         return MS::kFailure;
                     }
                 }
+
+                // Note: when round-tripping settings, some extra settings are not part
+                //       of the guiding dictionary. Ignore them.
+                const static bool reportError = false;
```
pierrebai-adsk, line 915: When import/export plugin add their custom settings, this would report errors about "unknown" settings.

lib/mayaUsd/utils/jsonDict.h
```diff
@@ -0,0 +1,216 @@
...
+#include <vector>
+
+namespace MAYAUSD_NS_DEF {
+
+////////////////////////////////////////////////////////////////////////////
+//
+// Convert a VtDictionary to JsValue.
+// Modeled after PXR_NS::JsValueTypeConverter
+
+class VtDictionaryToJsValueConverter
```
pierrebai-adsk, line 36: USD library provides a JSON -> VtDictionary converter, but not one going the other way. This is the other way. Not everything is supported, for example array-of-arrays is not supported and other such things.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -560,11 +560,7 @@ global proc mayaUsdTranslatorExport_contextOptionsShowExportUI(string $optionsSt
         return;
 
     // Fill the settings UI with the options from the export options UI.
-    mayaUsdTranslatorExport(
-        $_mayaUsdTranslatorExport_parent,
-        "fill=" + $_mayaUsdTranslatorExport_sections,
-        $newSettings,
-        "mayaUsdTranslatorExport_contextOptionsShowExportUI");
+    mayaUsdTranslatorExport_fillOptionsUIForcedByContextOptions();
```
pierrebai-adsk, line 563: Previous code was accidentally re-enabling the UI of settings controlled by import/export plugins.

pierrebai-adsk:
I will be on vacations all week, please set as ready-to-merge once reviewed. We can fix minor issues if any in another PR. This is purely demonstration code, so it can go in even if small improvement are suggested.

lib/mayaUsd/utils/jsonDict.h
AramAzhari-adsk: if this is more like a helper class may be we could rename the file to reflect the class itself.

AramAzhari-adsk (approved):

pierrebai-adsk:
The only failure is a renderer test on Linux that crashed, unrelated to my changes... but the stack trace indicate that this crash may be fixed by my other PR.

PR #3693, opened 2024-04-03
LOOKDEVX-2358 - Fix image file attribute detection

JGamache-autodesk:
An image UsdShade node that never had its filename attribute set would fail the isImageAttribute test. Allow the image code since a filename in a shader is always an image.

### DIFF
--- a/lib/mayaUsdAPI/utils.h
+++ b/lib/mayaUsdAPI/utils.h
@@ -111,6 +111,13 @@ makePathRelativeTo(const std::string& fileName, const std::string& relativeToDir
 MAYAUSD_API_PUBLIC
 bool requireUsdPathsRelativeToEditTargetLayer();
 
+MAYAUSD_API_PUBLIC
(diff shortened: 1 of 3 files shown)


lib/mayaUsdAPI/utils.h
```diff
@@ -111,6 +111,13 @@ makePathRelativeTo(const std::string& fileName, const std::string& relativeToDir
 MAYAUSD_API_PUBLIC
 bool requireUsdPathsRelativeToEditTargetLayer();
 
+MAYAUSD_API_PUBLIC
+std::string handleAssetPathThatMaybeRelativeToLayer(
```
JGamache-autodesk, line 115: Allow using @vlasovi 's relative path improvement in MayaUSD clients.

JGamache-autodesk:
Preflight passed (with one known ramdom failure of testUsdExportPackage).

seando-adsk (approved):

vlasovi (approved):

PR #3695, opened 2024-04-04
Support EditRouter in delete command

NickWu:
Hi, in our pipeline, we have a tool in Maya to delete prims from a specified layer using USD's native API, and we are using the `delete` key as the hotkey to run the tool. In order the make the prim deletion behaviour consistent between our tool and in Maya viewport (so that users could hit the `delete` key either with the tool or viewport in focus and get the same result), we'd like to map the `delete` key in Maya to our prim deletion tool, but we learned that the `delete` key executed in viewport cannot be remapped and it has been reserved to Maya's `cmds.delete()` command which has its own logic to delete the selected prims from the layers retrieved based on its own logic. `cmds.delete()` calls `UsdUndoDeleteCommand` internally and it doesn't support `EditRouter` to get an edit target layer from users. 

We made these changes in our pipeline to be able to send an `UsdEditTarget` to `UsdUndoDeleteCommand`  via `EditRouter` so that a prim can be deleted in a layer we specify. By sending a `UsdEditTarget` object instead of a `SdfLayer`, clients have full control on how to construct a `UsdEditTarget`;  if it's constructed with a pcp node, it can delete prims defined in non-local layers, which is one of the main intentions of this update. 

The changes are:

- add `getEditRouterEditTarget` function in `EditRouter` to retrieve `UsdEditTarget`. 
- turn on `allowStronger` in `UsdUfe::applyCommandRestrictionNoThrow` to pass the check.
- add new `EditRouter` tokens for registering new `EditRouter` callback for deletion and retrieving `UsdEditTarget` from callbacks.
- update `wrapEditRouter` to be able to retrieve `UsdEditTarget` object from Python.

### DIFF
--- a/lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
+++ b/lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
@@ -17,8 +17,10 @@
 
 #include "private/UfeNotifGuard.h"
 
+#include <usdUfe/base/tokens.h>
 #include <usdUfe/ufe/Utils.h>
 #include <usdUfe/undo/UsdUndoBlock.h>
+#include <usdUfe/utils/editRouter.h>
 #include <usdUfe/utils/layers.h>
 #include <usdUfe/utils/usdUtils.h>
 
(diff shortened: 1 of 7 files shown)


lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
```diff
@@ -60,20 +62,38 @@ void UsdUndoDeleteCommand::execute()
...
+    // attributes that were authored only to be the destination of a connection.
+    if (!UsdUfe::cleanReferencedPath(_prim)) {
+        const std::string error = TfStringPrintf(
+            "Failed to cleanup references to prim \"%s\".", _prim.GetPath().GetText());
+        TF_WARN("%s", error.c_str());
+        throw std::runtime_error(error);
+    }
+
+    if (!routingEditTarget.IsNull()) {
+        PXR_NS::UsdEditContext ctx(stage, routingEditTarget);
```
pierrebai-adsk, line 84: The edit context to change the edit target should be set before calling `applyCommandRestrictionNoThrow()` above, since that function verifies that the edit can be made for the current target. Otherwise the check and the action are not in sync.

lib/usdUfe/python/wrapEditRouter.cpp
```diff
@@ -77,9 +77,34 @@ class PyEditRouter : public UsdUfe::EditRouter
             TF_WARN("%s", ex.what());
             throw;
         }
-        boost::python::extract<PXR_NS::VtDictionary> extractedDict(dictObject);
-        if (extractedDict.check()) {
-            routingData = extractedDict;
+
+        const boost::python::object items = dictObject.items();
```
pierrebai-adsk, line 81: Could you add a comment in the code explaining what the new code is doing better than the old one? I'm sure you added all this code to fix an issue, I'd just like to understand what was going wrong and future reader will appreciate knowing too.

pierrebai-adsk: Oh, it is the bit about allowing to pass an edit target object instead of just a layer! Good work, but do document it in the code.

lib/usdUfe/ufe/Utils.cpp
```diff
@@ -400,7 +400,12 @@ bool allowedInStrongerLayer(
         ? stage->GetSessionLayer()
         : stage->GetRootLayer();
 
-    return getStrongerLayer(searchRoot, targetLayer, topLayer) == targetLayer;
```
pierrebai-adsk, line 403: I'd like a comment explaining why we should allow when stronger is null. I know that historically, getting the logic and handling weird cases has been delicate, so I'd like to know when the stronger layer is null and why it is okay.

pierrebai-adsk: Reading your test, I think it is because you target a layer within the reference?

We are presently considering allowing targeting layer in references, but we do not have a design yet for the UI nor how it would work, so if I'm correct that you're doing this because you want to target a layer in a reference, this is forward-looking code, but I think it may evolve and change behavior in the future.

In particular, I think we may enforce the same edit restrictions with the layer stack of the reference, instead of what would be a blanket "allow everything" that the null check implies. Let me know if my understanding is correct. It is not a blocker for your PR, but in the future it may affect your workflow in some cases if you were deleting prims with opinions in multiple layers in the reference and we no longer allow it in the future. (I suppose we could implement a special mode to allow it. Hard to know in advance.)

pierrebai-adsk (changes requested):
Need to fix teh edit context and add a few comments in the code to help us underatdn what is going on in the future.

lib/usdUfe/ufe/Utils.cpp
NickWu: Yes, it's mainly for targeting a layer in references. It's also the main intention of this PR. Alternatively, we could skip running `UsdUfe::applyCommandRestrictionNoThrow()` check altogether (which was the reason I modified `getStrongerLayer()`) if we receive an `UsdEditTarget` from the `EditRouter`. This way, we don't need to change the logic of `getStrongerLayer()` for now as it's meant for local layers. What do you think?

lib/mayaUsd/ufe/UsdUndoDeleteCommand.cpp
NickWu: Sure thing.

lib/usdUfe/ufe/Utils.cpp
NickWu: Hey @pierrebai-adsk , any thoughts on this?

pierrebai-adsk: I'd rather we keep calling it.

NickWu: Ok. I'll add comments to convey the intention of the changes.

NickWu:
Hi @pierrebai-adsk , I have added some comments as requested.

pierrebai-adsk (approved):

pierrebai-adsk:
Ran internal preflight #1847 and it passed.

PR #3697, opened 2024-04-04
EMSUSD-1161 - Fix the crash when attempting to save a USD file from USD Layer Editor

samuelliu-adsk:


seando-adsk (dismissed):
Seems a bit strange that it returns true if the input parent item is null. How is that a safe path to add?

samuelliu-adsk:
> Seems a bit strange that it returns true if the input parent item is null. How is that a safe path to add?

The reason I return `true` there is that this function checks if "sublayer path that is the same as the item or one of its parent", so if there's no parent layer then there's not gonna be any conflicts...

pierrebai-adsk:
While a corner case, there is no correct answer to the question "is is safe to add something to nothing". One could argue both ways. If it solves a crash, then I think there is no harm.

pierrebai-adsk (dismissed):
The only improvement I could think of is documenting the behaviour with null parent layer in the header file.

pierrebai-adsk (approved):

seando-adsk (approved):

PR #3702, opened 2024-04-08
EMSUSD-1056 fix crash due to invalid shader

pierrebai-adsk:
The multi-threaded viewport update code would sometimes end-up using a shader that was no longer valid. Seems related to reference counting and multi-threading and reusing the same material on multiple (possibly instanced) meshes. Seems to be due to a race condition between multiple threads, the guarantee about multi-threaded in VP2 / OGS are not entirely clear.

Fix it by delaying releasing shaders until the entire update is done. We keep these unused shaders in a set. Each unused shader gets released at the end of the update.


samuelliu-adsk:
LGTM. I'm not sure if releasing dead shader at the end of the update will slow down the loading process, might need some more QA work on that.

samuelliu-adsk (approved):

PR #3703, opened 2024-04-08
EMSUSD-1180 Fix blendshape names

barbalt:
When exporting maya blendshapes to usd, it's used the name of the mesh shape. However, the Maya's shape editor lists the name of the parent node to drive the blendshape weights. 

This PR it's an improvement so that the default USD blendshapes are created with proper names.

### DIFF
--- a/lib/usd/translators/meshWriter_BlendShapes.cpp
+++ b/lib/usd/translators/meshWriter_BlendShapes.cpp
@@ -677,10 +677,10 @@ MObject PxrUsdTranslators_MeshWriter::writeBlendShapeData(UsdGeomMesh& primSchem
                             MFnDagNode parentDagNode(dagNode.parent(0));
                             curTargetNameMStr
                                 = UsdMayaUtil::GetUniqueNameOfDagNode(parentDagNode.object());
-                            curTargetLongNameMStr = MString(curTargetNameMStr);
+                            curTargetLongNameMStr = curTargetNameMStr;
                         } else {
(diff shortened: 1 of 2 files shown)


lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -672,6 +671,17 @@ MObject PxrUsdTranslators_MeshWriter::writeBlendShapeData(UsdGeomMesh& primSchem
                     MString                   curTargetNameMStr;
                     MString                   curTargetLongNameMStr;
                     if (!targetMesh.isNull()) {
+                        MFnDagNode dagNode(targetMesh);
+                        MString    nodeName;
+                        if (dagNode.parentCount() > 0) {
+                            MFnDagNode parentDagNode(dagNode.parent(0));
+                            curTargetNameMStr
+                                = UsdMayaUtil::GetUniqueNameOfDagNode(parentDagNode.object());
+                            curTargetLongNameMStr = MString(curTargetNameMStr);
```
pierrebai-adsk, line 680: IDK why there is an explicit construction/cast of a MString here, but the old code was doing it too, so there is little harm.

pierrebai-adsk (approved):

lib/usd/translators/meshWriter_BlendShapes.cpp
barbalt: I used them because they were there before, but I can remove them as well, no problem

pierrebai-adsk (approved):

PR #3705, opened 2024-04-09
EMSUSD-1030 - Persist the default prim selection if its available

samuelliu-adsk:


plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -207,6 +207,26 @@ proc string mayaUsdTranslatorExport_AppendFromTextField(string $currentOptions,
...
+    }
+    if ((`checkBoxGrp -exists exportMeshesCheckBox` == 0)||(`checkBoxGrp -exists exportLightsCheckBox` == 0)||(`checkBoxGrp -exists exportCamerasCheckBox` == 0)){
+        return;
+    }
+    // check if the previous option is still available before set
+    string $allItems[] = `optionMenuGrp -q -itemListLong exportDefaultPrim`;
+    int $count = 1;
+    for($item in $allItems){
+        string $labelText = `menuItem -query -label $item`;
+        if($labelText == $defaultPrimName){
```
samuelliu-adsk, line 222: Select the prim as default prim if it's in the candidates' list.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -207,6 +207,26 @@ proc string mayaUsdTranslatorExport_AppendFromTextField(string $currentOptions,
...
+            optionMenuGrp -e -select $count exportDefaultPrim;
```
seando-adsk, line 223: Here you could use `optionMenuGrp -e -value $defaultPrimName exportDefaultPrim;`

to set the menu item by value. Then you don't need to keep track of the count.

You could even just wrap that command in a `catchQuiet` then you don't even need to check if its in the menus. It will just silently fail if its not present anymore.

samuelliu-adsk: Oh thanks! That should be a much better way to do it

seando-adsk (approved):

PR #3706, opened 2024-04-09
EMSUSD-1149 prevent unlocking system-locked layer when unlocking parents

pierrebai-adsk:
- When unlocking recursively, do not unlock sub-layers that are system-locked.
- Correctly check that the *request* was for system lock when undoing a lock command.
- Add more lock tests to cover more cases, like having sub-layers and system-locked layers.

### DIFF
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -1202,8 +1202,7 @@ MSyntax LayerEditorCommand::createSyntax()
     // parameter 2: refresh sub layers
     syntax.addFlag(
         kRefreshSystemLockFlag, kRefreshSystemLockFlagL, MSyntax::kString, MSyntax::kBoolean);
-    syntax.addFlag(
-        kSkipSystemLockedFlag, kSkipSystemLockedFlagL);
+    syntax.addFlag(kSkipSystemLockedFlag, kSkipSystemLockedFlagL);
 
     return syntax;
 }
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -848,7 +848,19 @@ public:
 
         // Execute lock commands
         for (size_t layerIndex = 0; layerIndex < _layers.size(); layerIndex++) {
-            MayaUsd::lockLayer(_proxyShapePath, _layers[layerIndex], _lockType, true);
+            auto curLayer = _layers[layerIndex];
+            // Note: per design, we refuse to affect the lock status of system-locked
+            //       sub-layers. To change the lock status of a system-locked layer,
+            //       it must be targeted directly.
+            if (curLayer != layer) {
+                if (_lockType != MayaUsd::LayerLockType::LayerLock_SystemLocked) {
+                    if (MayaUsd::isLayerSystemLocked(curLayer)) {
+                        continue;
+                    }
+                }
+            }
+
+            MayaUsd::lockLayer(_proxyShapePath, curLayer, _lockType, true);
         }
 
         updateEditTarget(stage);
@@ -868,7 +880,7 @@ public:
         // Execute lock commands
         for (size_t layerIndex = 0; layerIndex < _layers.size(); layerIndex++) {
             // Note: the undo of system-locked is unlocked by design.
-            if (_previousStates[layerIndex] == MayaUsd::LayerLockType::LayerLock_SystemLocked) {
+            if (_lockType == MayaUsd::LayerLockType::LayerLock_SystemLocked) {
                 MayaUsd::lockLayer(
                     _proxyShapePath,
                     _layers[layerIndex],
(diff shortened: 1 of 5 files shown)


AramAzhari-adsk (changes requested):
The design for the locklayer command is to allow locking / systemlocking / unlocking any layer and its sublayers through script, but not through UI.

I believe this will prevent the scenario of going through the script.

AramAzhari-adsk:
> Well, if we all recursive unlocking of system-locked layers in the command, then the recursive option will not be useful because we will need to do the recursion in the UI code and skip the system-locked layer. The UI will not be able to use the recursive flag.

We may either have to add a new parameter to mark the command to be run as a super user or a way to pass the information through UI to know it's being run through there. Any other suggestions are also welcome!

pierrebai-adsk:
No, that starts to be byzantine. The Maya command hook will do the layer recursion.

AramAzhari-adsk:
> No, that starts to be byzantine. The Maya command hook will do the layer recursion. Maaaybe there could be a flag to skip system-locked layers? That seems reasonable.

Either that or a generic way to indicate that overrides are allowed, but that creates the concept of user roles.
Even something as simple as a `force` parameter.

pierrebai-adsk:
Added -skipSystemLocked flag. Seems explicit enough to make its behavior clear.

lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -868,7 +883,7 @@ class LockLayer : public BaseCmd
         // Execute lock commands
         for (size_t layerIndex = 0; layerIndex < _layers.size(); layerIndex++) {
             // Note: the undo of system-locked is unlocked by design.
-            if (_previousStates[layerIndex] == MayaUsd::LayerLockType::LayerLock_SystemLocked) {
+            if (_lockType == MayaUsd::LayerLockType::LayerLock_SystemLocked) {
```
AramAzhari-adsk, line 886: I believe this should also be conditioned?

lib/usd/ui/layerEditor/mayaCommandHook.cpp
```diff
@@ -208,8 +208,13 @@ void MayaCommandHook::lockLayer(
     MayaUsd::LayerLockType lockState,
     bool                   includeSubLayers)
 {
+    // Per design, we refuse to change the lock state of system-locked
+    // layers through the UI.
+    if (MayaUsd::isLayerSystemLocked(usdLayer))
```
AramAzhari-adsk, line 213: This also needs to be conditioned whether we skip system locked files or not.

lib/mayaUsd/commands/layerEditorCommand.cpp
pierrebai-adsk: I don't understand what you mean

lib/usd/ui/layerEditor/mayaCommandHook.cpp
pierrebai-adsk: Again, I don't understand what you mean. We never want to affect system-locked layer from thr UI.

lib/mayaUsd/commands/layerEditorCommand.cpp
pierrebai-adsk: The previous state will always have the correct value when undoing, so there is no need to protect, at worst it will set the same state it already was.

AramAzhari-adsk: As long as any consecutive number of system-locks results in Unlock, then it's all good.

lib/usd/ui/layerEditor/mayaCommandHook.cpp
AramAzhari-adsk: What I would suggest though, is to add the `skipSystemLocked`. Because unlike Muting, sub-layers of a system-locked layer can be unlocked.

AramAzhari-adsk (changes requested):
MayaCommandHook::lockLayer should allow unlocking sub-layers of a system-locked layer since the system-lock is not inheritance-based such as Muting layers.

AramAzhari-adsk (approved):
Actually, if the menu item is disabled, then there's no need to apply to sub-layers. 
The sub-layers can be unlocked through the selection of those layers.

All good.

lib/usd/ui/layerEditor/mayaCommandHook.cpp
pierrebai-adsk: Ah, so we should et it do its work if include sub-layers is true? OK.

AramAzhari-adsk: Actually the code behind the lock layer and sublayer menu is
https://github.com/Autodesk/maya-usd/blob/7466596d572c800c68e37c587e8687ab11b534a0/plugin/adsk/scripts/mayaUsdMenu.mel#L945-L953

which seems to be enabled only if the right-clicked layer isn't system locked.

PR #3709, opened 2024-04-11
EMSUSD-993: When opening a file dependency in the AE, the dialog doesn't open to that path

seando-adsk:
EMSUSD-993: When opening a file dependency in the AE, the dialog doesn't open to the location of the file
* If we can resolve path to absolute, open the file dialog on that path.
* Fix spelling mistakes

### DIFF
--- a/lib/mayaUsd/utils/utilFileSystem.h
+++ b/lib/mayaUsd/utils/utilFileSystem.h
@@ -250,7 +250,7 @@ std::string getNumberSuffix(const std::string& text);
 MAYAUSD_CORE_PUBLIC
 std::string increaseNumberSuffix(const std::string& text);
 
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/utils/utilFileSystem.h
```diff
@@ -250,7 +250,7 @@ std::string getNumberSuffix(const std::string& text);
 MAYAUSD_CORE_PUBLIC
 std::string increaseNumberSuffix(const std::string& text);
 
-/*! \brief returns the aboluste path relative to the maya file
+/*! \brief returns the absolute path relative to the Maya file
```
seando-adsk, line 253: Trivial - fixed a few spelling mistakes.

PR #3712, opened 2024-04-11
EMSUSD-1067 allow deleting prims carrying a loaded payload

pierrebai-adsk:
- Apply the same logic as for references.
- Add test for deletion of a prim containing a payload or a ref.

### DIFF
--- a/test/lib/ufe/testPayloadCommands.py
+++ b/test/lib/ufe/testPayloadCommands.py
@@ -144,7 +144,7 @@ class PayloadCommandsTestCase(unittest.TestCase):
         cmd.redo()
         self.assertTrue(prim.IsLoaded())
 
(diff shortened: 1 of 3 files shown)


test/lib/ufe/testPayloadCommands.py
```diff
@@ -144,5 +144,30 @@ def testLoadAndUnloadPayloadCommands(self):
         cmd.redo()
         self.assertTrue(prim.IsLoaded())
 
+    @unittest.skipUnless(mayaUtils.mayaMajorVersion() > 2022, 'Delete restriction on delete was only after Maya 2022.')
```
seando-adsk, line 147: Minor, but I think its clearer to use ">= 2023" and say "Delete restriction requires Maya 2023 or greater". I think most of the tests do that.

seando-adsk (approved):

seando-adsk (approved):

PR #3714, opened 2024-04-11
EMSUSD-895 - Payload command undo to restore payload rules

brossee-adsk:
Payload undo doesn't go back to previous descendants states. The fix keeps the state of the payload rules applied on the stage before executing the command to load/unload a payload. The kept state is reapplied on the stage when the undo command is requested.


lib/usdUfe/ufe/UsdUndoPayloadCommand.cpp
```diff
@@ -70,22 +56,39 @@ void UsdUndoLoadUnloadBaseCommand::saveModifiedLoadRules() const
     UsdUfe::saveStageLoadRules(_stage);
 }
 
+// Macros to handle the redo/undo actions on stage load/unload rules
+// The load rules are saved before applying new rules and are
+// applied back when the commands are undo
+#define STAGE_PAYLOAD_ACTION(action)     \
```
deboisj, line 62: Could use a function instead of a macro - with a lambda you can do the same thing. Personnally I find macros annoying to maintain and work with.

lib/usdUfe/ufe/UsdUndoPayloadCommand.h
```diff
@@ -39,10 +39,11 @@ class USDUFE_PUBLIC UsdUndoLoadUnloadBaseCommand : public Ufe::UndoableCommand
 
     void saveModifiedLoadRules() const;
 
-private:
+protected:
     const PXR_NS::UsdStageWeakPtr _stage;
     const PXR_NS::SdfPath         _primPath;
     PXR_NS::UsdLoadPolicy         _policy;
+    PXR_NS::UsdStageLoadRules     _undoRules;
```
deboisj, line 46: You probably want to update / add to https://git.autodesk.com/maya3d/maya-usd/blob/dev/test/lib/ufe/testPayloadCommands.py

pierrebai-adsk (approved):

AntonK-Adsk (dismissed):

pierrebai-adsk (approved):
Great, LGTM

deboisj (approved):
👍

seando-adsk:
@brossee-adsk I didn't see a preflight build. I think it might be because you assigned the PR to me which remove the check. So I unassigned your PR and then just assigned it to you. I see the preflight started. I removed the `ready-for-merge` label. Once the preflight is successful, please add the label back then someone will merge your PR.

PR #3715, opened 2024-04-12
EMSUSD-1142 layer editor icon scaling

pierrebai-adsk:
Correctly define the scaling icon and don't try to calculate the DPI in the plugin, let the pixmap creation do the maths. In that scheme, the default 100% scaling pixamp does not have a DPI suffix, so fix the resource names.

### DIFF
--- a/lib/usd/ui/layerEditor/layerTreeItem.cpp
+++ b/lib/usd/ui/layerEditor/layerTreeItem.cpp
@@ -32,8 +32,8 @@ LayerActionDefinitions LayerTreeItem::_actionButtons;
 
 static void createPixmapPair(const QString& name, QPixmap& normal, QPixmap& hover)
 {
-    normal = utils->createPNGResPixmap(utils->getDPIPixmapName(name));
-    hover = utils->createPNGResPixmap(utils->getDPIPixmapName(name + QString("_hover")));
+    normal = utils->createPNGResPixmap(name);
+    hover = utils->createPNGResPixmap(name + QString("_hover"));
 }
 
 const LayerActionDefinitions& LayerTreeItem::actionButtonsDefinition()
--- a/lib/usd/ui/layerEditor/qtUtils.cpp
+++ b/lib/usd/ui/layerEditor/qtUtils.cpp
@@ -35,6 +35,14 @@ QPixmap QtUtils::createPNGResPixmap(QString const& in_pixmapName, int width, int
     if (pixmapName.left(2) != resourcePrefix) {
         pixmapName = resourcePrefix + pixmapName;
     }
+
+    // Note: createPNGResPixmap calls QtUtils::createPNGResPixmap, but that is
+    //       *not* the function below but rather MayaQtUtils::createPixmap, since
+    //       these functions are virtual.
+    //
+    //       The MayaQtUtils version calls MQtUtil::createPixmap which calls
+    //       QmayaQtHelper::createPixmap which generates the scaled image name
+    //       by adding the _150 or _200 suffix as necessary.
     return createPixmap(pixmapName, width, height);
 }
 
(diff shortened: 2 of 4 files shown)


lib/usd/ui/layerEditor/layerTreeItem.cpp
```diff
@@ -32,8 +32,8 @@ LayerActionDefinitions LayerTreeItem::_actionButtons;
 
 static void createPixmapPair(const QString& name, QPixmap& normal, QPixmap& hover)
 {
-    normal = utils->createPNGResPixmap(utils->getDPIPixmapName(name));
-    hover = utils->createPNGResPixmap(utils->getDPIPixmapName(name + QString("_hover")));
+    normal = utils->createPNGResPixmap(name);
+    hover = utils->createPNGResPixmap(name + QString("_hover"));
```
seando-adsk, line 36: So when user is running at 200% how will it pick the _200.png file?
It seems like with these changes it will always use the 100% file and scale that?

pierrebai-adsk: The function createPNGResPixmap calls createPixmap which modifies the image name to add the scaling value to it (_150, _200) when necessary.

pierrebai-adsk: createPNGResPixmap calls QtUtils::createPNGResPixmap  which calls MayaQtUtils::createPixmap which calls MQtUtil::createPixmap which calls QmayaQtHelper::createPixmap which generates the scaled image name by adding the _150 or _200 suffix as necessary.

What is somewhat misleading is that there is a QtUtils::createPixmap function, but that funciton is virtual, so that is not the one that gets called, instead it calls the MayaQtUtils version which does the magic.

pierrebai-adsk: Maybe I could put a comment in createPNGResPixmap with these explanations?

lib/usd/ui/layerEditor/qtUtils.cpp
```diff
@@ -35,6 +35,14 @@ QPixmap QtUtils::createPNGResPixmap(QString const& in_pixmapName, int width, int
...
     }
+
+    // Note: createPNGResPixmap calls QtUtils::createPNGResPixmap, but that is
+    //       *not* the function below but rather MayaQtUtils::createPixmap, since
+    //       these functions are virtual.
+    //
+    //       The MayaQtUtils version calls MQtUtil::createPixmap which calls
+    //       QmayaQtHelper::createPixmap which generates the scaled image name
+    //       by adding the _150 or _200 suffix as necessary.
     return createPixmap(pixmapName, width, height);
```
pierrebai-adsk, line 46: Added the explanation how the scaled image names are generated.

lib/usd/ui/layerEditor/layerTreeItem.cpp
seando-adsk: Ah yes I missed that overridden function. Thanks for the explanation. Yeah maybe a comment in "createPNGResPixmap" would be nice, saying that when running in Maya the proper png (based on scaling) will be used.

lib/usd/ui/layerEditor/qtUtils.cpp
seando-adsk: Thank you.

lib/usd/ui/layerEditor/layerTreeItem.cpp
pierrebai-adsk: No worry, when you asked, I knew the answer because I had traced the execution, but when I went to read the code I was as confused as you and had to re-run the code to figure out what was going on...

AramAzhari-adsk (approved):

AramAzhari-adsk:
Looks good. Thanks for the fix!

PR #3719, opened 2024-04-15
EMSUSD-1173 filter items hidden in the outliner

pierrebai-adsk:
Filter items hidden in the outliner for the root prim drop-down when caching to USD.

### DIFF
--- a/plugin/adsk/scripts/mayaUsd_exportHelpers.py
+++ b/plugin/adsk/scripts/mayaUsd_exportHelpers.py
@@ -7,6 +7,14 @@ __naturalOrderRE = re.compile(r'([0-9]+)')
 def natural_key(item):
     return [int(s) if s.isdigit() else s.lower() for s in __naturalOrderRE.split(item)]
 
+def isHiddenInOutliner(item):
+    '''Verify if the Maya item is hidden in the outliner.'''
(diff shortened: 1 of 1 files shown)


samuelliu-adsk (approved):

plugin/adsk/scripts/mayaUsd_exportHelpers.py
```diff
@@ -33,7 +41,7 @@ def updateDefaultPrimCandidates(excludeMesh, excludeLight, excludeCamera):
 
         allItems = list(set(allItems) - set(startup_cameras))
     
-    allItems = list(set(allItems) - set(excludeList))
+    allItems = removeHiddenInOutliner(list(set(allItems) - set(excludeList)))
```
samuelliu-adsk, line 44: Feels like there's an easier way to just check if hidden in L19 when we query the item list. In python if you do set(a) - set(b), and b has items a doesn't have it will just ignore them

pierrebai-adsk: The reason I do it here is that there will be fewer items to check after we removed the ones we already don't want.

samuelliu-adsk: Good point!

PR #3720, opened 2024-04-15
Fix incorrect version check in test

JGamache-autodesk:


test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateMaterialX.py
```diff
@@ -97,7 +97,7 @@ def testMayaSurfaces(self):
             suffix = "_ocio"
 
         # MaterialX 1.38.8 has a new triplanar node with superior blending
-        if os.getenv('MATERIALX_VERSION', '1.38.0') >= '1.38.8':
+        if [int(i) for i in os.getenv('MATERIALX_VERSION', '1.38.0').split(".")] >= [1, 38, 8]:
```
JGamache-autodesk, line 100: String base comparison fails as `1` is not greater than `8`.

vlasovi (approved):

seando-adsk (approved):

PR #3722, opened 2024-04-15
EMSUSD-1092 fix layer editor refresh when reloading from AE

pierrebai-adsk:
- Support having multiple callbacks for a given UI callback.
- Support having multiple callbacks from Python.
- Support unregistering a callback.
- Register a UI callback for the onRefreshSystemLock"" notification in the layer view.
- Repaint when receiving that notification.

### DIFF
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -1078,8 +1078,8 @@ private:
 
     void _notifySystemLockIsRefreshed()
     {
-        UsdUfe::UICallback::Ptr dstCallback = UsdUfe::getUICallback(TfToken("onRefreshSystemLock"));
-        if (!dstCallback)
+        auto dstCallbacks = UsdUfe::getUICallbacks(TfToken("onRefreshSystemLock"));
+        if (dstCallbacks.size() == 0)
             return;
 
         PXR_NS::VtDictionary callbackContext;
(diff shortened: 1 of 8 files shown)


AramAzhari-adsk:
The command that handles mayaUsdLayerEditor should be able to call refresh system lock directly (see https://github.com/Autodesk/maya-usd/pull/3626/files#diff-4766f35a32ff713eeced3da2ad76371ee864c08640cc1451f4974e4931c6cfa7 ) .

I'm not sure if there's a need to refactor the callback or even need to use it in this instance.

Is this regarding the mouse cursor needing to be over USD Layer Editor to refresh its UI ?

pierrebai-adsk:
The problem is not calling or not calling the refresh lock command, the problem is that the UI does not updates itself when the locks are refreshed. Even if reload would call the refresh lock command, it would still not update the UI without the UI listening to the callback,

Furthermore, I think the enhancements are necessary:
- Supporting multiple callbacks, not just one.
- Allowing unregistering callbacks.
- Support both from Python.

AramAzhari-adsk:
> The problem is not calling or not calling the refresh lock command, the problem is that the UI does not updates itself when the locks are refreshed. Even if reload would call the refresh lock command, it would still not update the UI without the UI listening to the callback,
> 
> Furthermore, I think the enhancements are necessary:
> 
> * Supporting multiple callbacks, not just one.
> * Allowing unregistering callbacks.
> * Support both from Python.

Fair enough. The issue of UI not refreshing seem to extend to script commands as well. Perhaps in the future we'll find a way to listen to "LockStatusChanged" as a separate callback to refresh UI there as well.

Thanks

AramAzhari-adsk (approved):

PR #3725, opened 2024-04-16
EMSUSD-1002 repaint layer editor when layer muting changes

pierrebai-adsk:


pierrebai-adsk:
Similar to the other issue, except here we have a USD notification already available to listen to.

AramAzhari-adsk (approved):

AramAzhari-adsk:
> Similar to the other issue, except here we have a USD notification already available to listen to.

I think for anything that cause a change to the layer we should be able to call the UI refresh. For those that don't have a change callback in core USD, we could use our own callbacks.

Thanks for doing this.

seando-adsk:
@pierrebai-adsk Sorry I think what happened is I merged your other PR first. So now this one has merge conflicts.

AramAzhari-adsk (approved):

PR #3731, opened 2024-04-19
EMSUSD-208 Material custom control for AE

pierrebai-adsk:
- Added a new material custom control for the attribute editor (AE).
- It creates text fields for the material path, inherited material and ancesto prim path.
- If the material is directly on the prim, the inherited and ancestor fields are hidden.
- If the material is inherited, the material path is set to "inheriting".
- All fields are not editable.
- The direct or inherited material is found by using the ComputeBoundMaterial function from the USD material binding API.
- Added a button to go to the prim from which the material is inherited.
- Don't show the material section for materials themselves.
- Show the direct material when there is one even though an inherited material over-rides it.
- Show tooltip explaining how inherited material affect the prim in the assigned material field.
- Use italics font for the assigned material when inheriting a material.
- The material section is added after the section corresponding the prim type.

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
@@ -127,6 +127,119 @@ class UfeConnectionChangedObserver(ufe.Observer):
         # Nothing needed here since we don't create any UI.
         pass
 
+class MaterialCustomControl(object):
+    def __init__(self, item, prim, useNiceName):
+        super(MaterialCustomControl, self).__init__()
+        self.item = item
+        self.prim = prim
+        self.useNiceName = useNiceName
+
+    def onCreate(self, *args):
+        '''
+        Create the custom UI for the material.
+        '''
+        self.assignedMatLayout, self.assignedMatField, _ = self._createTextField(
+            'material',  'kLabelAssignedMaterial')
+        self.inheritedMatLayout, self.inheritedMatField, _ = self._createTextField(
+            'inherited', 'kLabelInheritedMaterial')
+        # Note: icon image taken from Maya resources.
+        self.fromPrimLayout, self.inheritedFromPrimField, self.gotoPrimButton = self._createTextField(
+            'from prim', 'kLabelInheritedFromPrim', 'inArrow.png')
+                
+        # Fill the UI.
+        self._refreshUI()
+
+    def _createTextField(self, longName, uiNameRes, image=None):
+        '''
+        Create a disabled text field group and an optional image button with the correct label.
+        '''
+        uiLabel = getMayaUsdLibString(uiNameRes) if self.useNiceName else longName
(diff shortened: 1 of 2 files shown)


lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -127,6 +127,119 @@ def onReplace(self, *args):
...
+        '''
+        direct = directMatPath.pathString if directMatPath else ''
+        inherited = mat.GetPath().pathString
+        fromPath = matRel.GetPrim().GetPath().pathString
+
+        ufePath = ufe.Path([
+            self.item.path().segments[0],
+            ufe.PathSegment(fromPath, mayaUsdUfe.getUsdRunTimeId(), '/')])
+        ufePathStr = ufe.PathString.string(ufePath)
+        command = lambda *_: cmds.select(ufePathStr)
```
JGamache-autodesk, line 212: The icon used usually means that the AE will update to the new object without selecting it. This introduces a new effect from an old and familiar icon. Please double check with UX team as selecting the object might require a new or different icon.

JGamache-autodesk (approved):

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
pierrebai-adsk: Ah, I did not know we could display something new in AE without selecting it. I'll look into what command could achieve that.

pierrebai-adsk: I have another PR based on this, so I'll do the research and change in that one.

pierrebai-adsk: Seems to be updateAE. I'll make the change in the next PR

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -127,6 +127,119 @@ def onReplace(self, *args):
         # Nothing needed here since we don't create any UI.
         pass
 
+class MaterialCustomControl(object):
```
seando-adsk, line 130: Is there a reason this class doesn't derive from `AttributeCustomControl`?

Also is there a reason for not placing this custom control in its own file (which we've done for others)?

pierrebai-adsk: I can fix both of these in the next PR. It is based on this one and awaiting his one to be open

pierrebai-adsk: (Not deriving from AttributeCustomControl did not seem to affect anything)

pierrebai-adsk: (The others also don't seem to derive from AttributeCustomControl?)

PR #3736, opened 2024-04-23
EMSUSD-204 Material binding strength in AE

pierrebai-adsk:
- Move the custom control class to its own file.
- Add an option menu to control the material binding strength in the attribute editor (AE).
- Change the binding strength when the user select a new value in the option menu.
- Hide the menu when the material in inherited and there is no direct material.
- Disable the menu when there is a stronger inherited material and a direct material.
- Show a tooltip that explain why it is disabled.
- Update children when material strength is changed.
- Don't change selection when going to the prim from which the material in inherited.
- Don't use italics.

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
@@ -17,10 +17,10 @@ from .custom_image_control import customImageControlCreator
 from .attribute_custom_control import getNiceAttributeName
 from .attribute_custom_control import cleanAndFormatTooltip
 from .attribute_custom_control import AttributeCustomControl
+from .material_custom_control import MaterialCustomControl
 
 import collections
 import fnmatch
-from functools import partial
 import re
 import ufe
 import usdUfe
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -127,119 +127,6 @@ def onReplace(self, *args):
         # Nothing needed here since we don't create any UI.
         pass
 
-class MaterialCustomControl(object):
```
pierrebai-adsk, line 130: Moved to its own file

AramAzhari-adsk (approved):

lib/mayaUsd/resources/ae/usdschemabase/material_custom_control.py
```diff
@@ -0,0 +1,202 @@
...
+        '''
+        direct = directMatPath.pathString if directMatPath else ''
+        inherited = mat.GetPath().pathString
+        fromPath = matRel.GetPrim().GetPath().pathString
+
+        ufePath = ufe.Path([
+            self.item.path().segments[0],
+            ufe.PathSegment(fromPath, mayaUsd.ufe.getUsdRunTimeId(), '/')])
+        ufePathStr = ufe.PathString.string(ufePath)
+        melCommand = 'updateAE "%s"' % ufePathStr
```
JGamache-autodesk, line 161: Another important change is using `updateAE` here instead of `cmds.select`. Thanks!

JGamache-autodesk (approved):

PR #3740, opened 2024-04-25
EMSUSD-209 graph the material in LookdevX

pierrebai-adsk:
- Detect if the LookdevX plugin is loaded. Everything else is turned off when not loaded.
- Add a button to graph the material in a LookdevX tab.
- Use a named tuple to hold text field info.
- Refactor the code to show buttons and update the associated commands.
- Use the lookdevXGraph command to open the tab.
- Add a utility function to build a UFE path from a USD path.
- Fixed that the AE would not update when assigning an existing material.
- Add a popup menu to the graph material buttons.
- Allow creating a new tab or re-using an existing tab.
- The material types is always USD since these are USD materials.

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/material_custom_control.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/material_custom_control.py
@@ -14,6 +14,7 @@
 #
 
 from functools import partial
+import collections
 
 import ufe
 import mayaUsd.ufe
@@ -36,6 +37,8 @@ class MaterialCustomControl(object):
         getMayaUsdLibString('kLabelStrongerMaterial')   : 'strongerThanDescendants',
     }
 
+    TextField = collections.namedtuple('TextField', ['layout', 'field', 'button'])
+
     def __init__(self, item, prim, useNiceName):
         super(MaterialCustomControl, self).__init__()
         self.item = item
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/resources/ae/usdschemabase/material_custom_control.py
```diff
@@ -139,35 +151,32 @@ def _refreshUIForDirect(self, mat, strength):
...
-        direct = directMatPath.pathString if directMatPath else ''
-        inherited = mat.GetPath().pathString
-        fromPath = matRel.GetPrim().GetPath().pathString
-
-        ufePath = ufe.Path([
-            self.item.path().segments[0],
-            ufe.PathSegment(fromPath, mayaUsd.ufe.getUsdRunTimeId(), '/')])
-        ufePathStr = ufe.PathString.string(ufePath)
-        melCommand = 'updateAE "%s"' % ufePathStr
-        command = lambda *_: mel.eval(melCommand)
```
pierrebai-adsk, line 162: Move to its own function for clarity

lib/mayaUsd/resources/ae/usdschemabase/material_custom_control.py
```diff
@@ -36,6 +37,8 @@ class MaterialCustomControl(object):
         getMayaUsdLibString('kLabelStrongerMaterial')   : 'strongerThanDescendants',
     }
 
+    TextField = collections.namedtuple('TextField', ['layout', 'field', 'button'])
```
pierrebai-adsk, line 40: Keep created button info into a named tuple for clarity and consistency.

lib/mayaUsd/resources/ae/usdschemabase/material_custom_control.py
```diff
@@ -190,13 +199,72 @@ def _fillUIValues(self, direct, inherited, fromPath, strength):
...
+            return
+
+        # Note: only show the graph button if LookdevX is currently loaded.
+        hasLookdevX = self._hasLookdevX()
+        canGraph = bool(matPathStr and hasLookdevX)
+        cmds.symbolButton(button, edit=True, enable=canGraph, visible=hasLookdevX)
+
+        if canGraph:
+            ufePathStr = self._createUFEPathFromUSDPath(matPathStr)
+            command = lambda *_: cmds.lookdevXGraph(graphNode=ufePathStr, tabName='USD')
```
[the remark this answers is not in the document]
JGamache-autodesk: Will work, but looks a bit arbitrary to either create or re-use a tab named "USD". I suspect we would like that pressing on that button opens a menu with options to graph in a new tab, or graph in any of the existing tabs currently opened in LookdevX. You might want to look at the `MaterialXSurfaceShader_createMenu` in `source\contrib\LookdevXMaya\scripts\AEMaterialXSurfaceShaderTemplate.mel` for inspiration on how to create that menu.

Sean showed me that `functools.partial` is very nice if you want to implement more complex callbacks. He used it in a few places in this file.

JGamache-autodesk (approved):

JGamache-autodesk (approved):

JGamache-autodesk (dismissed):
Ouf, a bit ugly. And also means that in Python 3.11 you should be using `functools.partialmethod` on those `@staticmethod` to prevent packaging `self` in the closures.

The other option is to make these plain functions outside of the class, which removes these `ignored` parameters. I would tend to think this would be prettier, but that is in the eye of the coder.

Maybe a comment about why these parameters exist?

JGamache-autodesk:
Another solution in this [stackoverflow thread](https://stackoverflow.com/questions/36922532/python-functools-partial-how-to-apply-it-to-a-class-method-with-the-static-dec) is to use the `__func__` attribute, but that would be using a private member, which is a bit ugly.

pierrebai-adsk:
The ignored parameters are not self, they are the parameter provided by the MEL callback for the menu and menu item selection, which we don't need. I'll add a comment about that.

JGamache-autodesk (approved):

test/lib/testAttributeEditorTemplate.py
```diff
@@ -201,11 +201,16 @@ def testAECustomMaterialControl(self):
         self.assertTrue(cmds.formLayout(meshFormLayout, exists=True))
         startLayout = cmds.formLayout(meshFormLayout, query=True, fullPathName=True)
         self.assertIsNotNone(startLayout, 'Could not get full path for Mesh formLayout')
-        
+
+        # In the AE there is a formLayout for each USD prim type. We start
+        # by making sure we can find the one for the mesh.
+        materialFormLayout = self.searchForMayaControl(startLayout, cmds.frameLayout, 'Material')
```
JGamache-autodesk, line 207: I suspect you want `getMayaUsdLibString('kLabelMaterial')` instead of `'Material'`

JGamache-autodesk (approved):

JGamache-autodesk (approved):

PR #3744, opened 2024-04-30
EMSUSD-1118 instance selection highlight

pierrebai-adsk:
Detect the need to redo the selection highlight by detecting if the instance count changed. Unfortunately, the instancer index version only changes when an instance was toggle off, not when togged on. Might be a bug or not in USD, but we work around it by tracking the instance count too.


AramAzhari-adsk:
> Detect the need to redo the selection highlight by detecting if the instance count changed. Unfortunately, the instancer index version only changes when an instance was toggle off, not when togged on. Might be a bug or not in USD, but we work around it by tracking the instance count too.

Does it depend on an attribute change?
If so, some attributes are not reported as "Changed" if they are simply removed or added. Listening to those events should generally fix it.

AramAzhari-adsk (dismissed):

AramAzhari-adsk (dismissed):

PR #3746, opened 2024-05-01
EMSUSD-1111 Edit routing for transform commands

pierrebai-adsk:
Add helpers to support permissions and routing:
- Provide a helper function to check permission-to-edit in the base class of the USD implementation of the UFE 3D Transform interface.
- The helper ensures that the correct edit context is used when checking permissions.

Use helpers in transforms and commands:
- Use those helper functions to check for permission to edit in command creation.
- Use those helper functions to do the edits, in command execution.
- Add edit routing context in all transform-related edits function.
- This context queries and sets where the edits should be routed.
- In particular, it is used in the execute, undo, redo and set functions of the command base class.
- Properly check permissions for matrix commands.
- The common API, 3d Matrix Op, point instances and Maya Xform stack are done.

Permissions for point instance transform
- Make the getAttribute function of the point instance modifier base class be public.
- Same for its derived classes.
- Add helper function to check permission for point instances.
- Use it to check permissions for point instances.

Add unit tests for transform routing.

### DIFF
--- a/lib/mayaUsd/ufe/UsdSetXformOpUndoableCommandBase.h
+++ b/lib/mayaUsd/ufe/UsdSetXformOpUndoableCommandBase.h
@@ -20,6 +20,7 @@
 #include <usdUfe/undo/UsdUndoableItem.h>
 
 #include <pxr/base/vt/value.h>
+#include <pxr/usd/usd/prim.h>
 #include <pxr/usd/usd/timeCode.h>
 
 #include <ufe/transform3dUndoableCommands.h>
@@ -72,6 +73,9 @@ public:
     void redo() override;
 
     PXR_NS::UsdTimeCode writeTime() const { return _writeTime; }
+
+    // Retrieve the USD prim affected by the command.
+    virtual PXR_NS::UsdPrim getPrim() const;
 
 protected:
     // Create the XformOp attributes if they do not exists.
--- a/lib/mayaUsd/ufe/UsdTransform3dBase.cpp
+++ b/lib/mayaUsd/ufe/UsdTransform3dBase.cpp
@@ -17,8 +17,13 @@
 
 #include <mayaUsd/ufe/Utils.h>
 
+#include <usdUfe/base/tokens.h>
+#include <usdUfe/utils/editRouterContext.h>
+
 #include <pxr/usd/usdGeom/xformCache.h>
 #include <pxr/usd/usdGeom/xformCommonAPI.h>
+
+#include <maya/MGlobal.h>
 
 PXR_NAMESPACE_USING_DIRECTIVE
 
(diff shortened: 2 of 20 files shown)


lib/mayaUsd/ufe/UsdSetXformOpUndoableCommandBase.h
```diff
@@ -73,6 +74,9 @@ class UsdSetXformOpUndoableCommandBase : public Ufe::SetVector3dUndoableCommand
 
     PXR_NS::UsdTimeCode writeTime() const { return _writeTime; }
 
+    // Retrieve the USD prim affected by the command.
+    virtual PXR_NS::UsdPrim getPrim() const;
```
pierrebai-adsk, line 78: The reason to make it virtual is that sub-classes can make getting the UsdPrim more efficient.

lib/mayaUsd/ufe/UsdTransform3dBase.cpp
```diff
@@ -121,5 +126,37 @@ Ufe::Matrix4d UsdTransform3dBase::segmentExclusiveMatrix() const
     return UsdTransform3dReadImpl::segmentExclusiveMatrix();
 }
 
+bool UsdTransform3dBase::isAttributeEditAllowed(const TfToken attrName) const
+{
+    return isAttributeEditAllowed(&attrName, 1);
+}
+
+bool UsdTransform3dBase::isAttributeEditAllowed(const TfToken* attrNames, int count) const
+{
```
pierrebai-adsk, line 135: This was moved from another class to share with multiple sub-classes.

lib/mayaUsd/ufe/UsdTransform3dMayaXformStack.cpp
```diff
@@ -751,24 +758,6 @@ UsdTransform3dMayaXformStack::getCvtRotXYZToAttrFn(const TfToken& opName) const
     return cvt.at(opName);
 }
 
-bool UsdTransform3dMayaXformStack::isAttributeEditAllowed(
```
pierrebai-adsk, line 754: This is where the now shared function was taken from.

lib/mayaUsd/ufe/UsdTransform3dPointInstance.cpp
```diff
@@ -137,6 +149,29 @@ Ufe::Matrix4d UsdTransform3dPointInstance::segmentExclusiveMatrix() const
     return UsdTransform3dBase::segmentInclusiveMatrix();
 }
 
+bool UsdTransform3dPointInstance::isAttributeEditAllowed(const PXR_NS::UsdAttribute& attr) const
+{
```
pierrebai-adsk, line 153: Point instancer Transform3d implementation does not derived from the shared base class, so it needed similar code.

test/lib/ufe/testEditRouting.py
```diff
@@ -125,67 +128,65 @@ def setUp(self):
...
         cmds.file(new=True, force=True)
-        import mayaUsd_createStageWithNewLayer
- 
-        # Create the following hierarchy:
+
+    def tearDown(self):
+        # Restore default edit routers.
+        mayaUsd.lib.restoreAllDefaultEditRouters()
+
+    def _prepareSimpleScene(self):
```
pierrebai-adsk, line 136: Not all tests use teh same scene anymore, so the scene setup was moved from setUp() to a helper function.

lib/mayaUsd/ufe/UsdTransform3dCommonAPI.cpp
```diff
@@ -64,6 +66,8 @@ class CommonAPITranslateUndoableCmd : public UsdSetXformOpUndoableCommandBase
 
     bool set(double x, double y, double z) override
     {
+        UsdUfe::OperationEditRouterContext editContext(
+            UsdUfe::EditRoutingTokens->RouteTransform, getPrim());
```
AramAzhari-adsk, line 70: Thanks for doing this. I think having a unified callback for all transform types has been a good decision. If the need to expose them separately rises, we could always add individual ones.

AramAzhari-adsk (approved):
Looks good to me.

PR #3747, opened 2024-05-02
EMSUSD-1258 extend the delete cmd edit routing

pierrebai-adsk:
The new code for doing edit routing using an edit target directly was not supporting the usual use-case of providing the target by the layer identifier or layer handle.

It also always allowed stronger opinions, even though we only need it if the target is within a reference or payload.

This fixes both issues: delete can now be routed like other commands by ID or layer handle and will only allow stronger opinions when targeting a non-local layer.

Extending using an edit target directly in all other commands will be done separately in the future.

### DIFF
--- a/lib/usdUfe/utils/editRouter.cpp
+++ b/lib/usdUfe/utils/editRouter.cpp
@@ -147,6 +147,17 @@ EditRouter::Ptr getEditRouter(const PXR_NS::TfToken& operation)
     return (foundRouter == editRouters.end()) ? nullptr : foundRouter->second;
 }
 
+static PXR_NS::SdfLayerHandle _extractLayer(const PXR_NS::VtValue& value)
+{
+    if (value.IsHolding<std::string>())
+        return PXR_NS::SdfLayer::Find(value.Get<std::string>());
+
+    if (value.IsHolding<PXR_NS::SdfLayerHandle>())
+        return value.Get<PXR_NS::SdfLayerHandle>();
+
+    return nullptr;
+}
+
 PXR_NS::SdfLayerHandle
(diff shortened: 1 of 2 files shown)


lib/usdUfe/utils/editRouter.cpp
```diff
@@ -229,14 +219,21 @@ getEditRouterEditTarget(const PXR_NS::TfToken& operation, const PXR_NS::UsdPrim&
...
+        auto&& editTarget = value.Get<PXR_NS::UsdEditTarget>();
+        return editTarget;
     }
-    return PXR_NS::UsdEditTarget();
+
+    auto layer = _extractLayer(found->second);
+    if (!layer)
+        PXR_NS::UsdEditTarget();
+
+    return PXR_NS::UsdEditTarget(layer);
```
NickWu, line 236: So essentially this allows users use pass in a SdfLayer and get an UsdEditTarget back, right? That's nice to have for local layer use cases.

NickWu:
> @NickWu This enhance the PR that you opened related to targeting non-local layers for the delete command. The unit test that you added still passes, so I believe this is a pure backward-compatible enhancement.

This looks good to me. Thanks.

lib/usdUfe/utils/editRouter.cpp
pierrebai-adsk: Yes, the other routable commands allow the routing function to return a layer or layer ID, so this makes delete support those too.

We would still need to update the other routable commands to use your trick to allow a non-local layers, but that would be done in a future pull request.

NickWu: Sounds good.

AramAzhari-adsk (approved):

PR #3749, opened 2024-05-03
EMSUSD-1111 move all transform-related UFE code to their own folder

pierrebai-adsk:
This make finding all transform-related code easier. Also, remove a few unused classes.


pierrebai-adsk:
This moves existing code together in a folder and deletes unused classes. Makes it easier to find all transform-related classes.

pierrebai-adsk:
We can either keep this or you could base your future work on it.

PR #3752, opened 2024-05-06
EMSUSD-1046 manip pivot position

pierrebai-adsk:
Fix the position of the Maya manipulators when USD pivot are used instead of (or in addition to) Maya pivots.


pierrebai-adsk:
Note: the only PF failure is a random failure due to a file being left behind in the tmp folder in a previous run.

samuelliu-adsk (approved):

PR #3758, opened 2024-05-08
EMSUSD-1272 - CopyPaste: MayaUSD clipboard error messages when opening up preferences window

seando-adsk:
#### EMSUSD-1272 - CopyPaste: MayaUSD clipboard error messages when opening up preferences window
* cmake mistake caused missing python binding

### DIFF
--- a/plugin/adsk/scripts/mayaUsd_preferenceTab.mel
+++ b/plugin/adsk/scripts/mayaUsd_preferenceTab.mel
@@ -46,8 +46,8 @@ global proc mayaUsd_PrefFileFormatChanged()
 	}
 
 	// Update the clipboard file format to match the format here.
-
-	if (`optionVar -exists $saveLayerFormatArgBinaryOptionPref`) {
+	int $haveClipboardSupport = `python("import mayaUsd.ufe as mufe; callable(getattr(mufe, 'setClipboardFileFormat', None))")`;
+	if ($haveClipboardSupport && `optionVar -exists $saveLayerFormatArgBinaryOptionPref`) {
 		int $isBinary = `optionVar -q $saveLayerFormatArgBinaryOptionPref`;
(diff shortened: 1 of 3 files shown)


plugin/adsk/scripts/mayaUsd_preferenceTab.mel
```diff
@@ -46,8 +46,8 @@ global proc mayaUsd_PrefFileFormatChanged()
 	}
 
 	// Update the clipboard file format to match the format here.
-
-	if (`optionVar -exists $saveLayerFormatArgBinaryOptionPref`) {
+	int $haveClipboardSupport = `python("import mayaUsd.ufe as mufe; callable(getattr(mufe, 'setClipboardFileFormat', None))")`;
+	if ($haveClipboardSupport && `optionVar -exists $saveLayerFormatArgBinaryOptionPref`) {
```
seando-adsk, line 50: This clipboard function won't be there for Maya 2024 (and earlier) so check if function exists before calling it.

test/lib/ufe/testClipboard.py
```diff
@@ -52,6 +52,9 @@ def setUp(self):
     def testClipboardCopyPaste(self):
         '''Basic test for the Clipboard copy/paste support.'''
 
+        # Verify that the method to set the clipboard format exists.
+        self.assertTrue(callable(getattr(mayaUsd.ufe, 'setClipboardFileFormat', None)))
```
AramAzhari-adsk, line 56: Does this have to run for Maya 2024 and earlier?

seando-adsk: This entire test file is only included when the Ufe version being built with contains the clipboard support. So this file won't be included for 2024.

AramAzhari-adsk (approved):

PR #3762, opened 2024-05-10
EMSUSD-1285 fix all test that relied on syncolor

pierrebai-adsk:
Update the baseline images. Change the name of one color transform.


pierrebai-adsk:
There were 3 tests that now fails on 2022 and one that failed on 2024 but passed when I ran it locally, so I disabled these on those platforms for now to get blessing to pass again.

I'll see what I can do for the 2022 ones in another PR. That will only leave testVP2RenderDelegateDrawModes, which passes for me locally on 2024 and 2022 but not during PF.

pierrebai-adsk:
I already fixed them, so I'll close this one and we can use the other PR

PR #3763, opened 2024-05-10
EMSUSD-1285 fix all test that relied on syncolor

pierrebai-adsk:
Update the baseline images. Change the name of one color transform.
Disable one test in 2022 and 2024 that I cannot repro locally.
Added images for Maya 2022.


seando-adsk (approved):
Looks good to me. I'll add @JGamache-autodesk to have a look.

JGamache-autodesk (approved):
LGTM.

PR #3768, opened 2024-05-15
EMSUSD-1301 export materials checkbox

pierrebai-adsk:
- Add a "-exportMaterials" (-mat) flag to the export command.
- Add it to the job arguments class. It defaults to true.
- Document the new flags..
- Use the flag during export to control material exports.
- Expose the flag state to Python.
- Add a checkbox in the export UI to control the flag.
- Disable material UI when material exports are off.
- Add unit test that turns off materials.


lib/mayaUsd/fileio/shading/shadingModeUseRegistry.cpp
```diff
@@ -377,6 +381,9 @@ class UseRegistryShadingModeExporter : public UsdMayaShadingModeExporter
             *mat = material;
         }
 
+        if (!context.GetExportArgs().exportMaterials)
```
samuelliu-adsk, line 384: Do we need this line? Isn't L349 already checked the flag?

samuelliu-adsk (approved):

PR #3769, opened 2024-05-15
Add support for env var expansions in file node writers

### LINKED ISSUE #3766: mayaUsdExport wrongly translates file texture names with envvar expansions.
**Describe the bug**

We use environment variables expansion in file texture paths for asset management needs. This is supported by maya file node but not by maya-usd export, which results in invalid usd asset paths.

**Steps to reproduce**

```python
import os
from maya import cmds
from pxr import Usd, UsdShade
cmds.loadPlugin("mayaUsdPlugin", quiet=True)
```

1. Prepare environment and open scene from [mayaUSDExport_file_envvar_expansion.zip](https://github.com/Autodesk/maya-usd/files/15320332/mayaUSDExport_file_envvar_expansion.zip)
```python
MY_TEST_DIR = "/path/to/issue_files/mayaUSDExport_file_envvar_expansion"
os.environ["MY_TEST_DIR"] = MY_TEST_DIR
cmds.file(os.path.join(MY_TEST_DIR, "scene.ma"), open=True, force=True)
```

2. It reads a texture file with  `MY_TEST_DIR` expansion, this is supported by file node

```python
maya.cmds.getAttr("file1.ftn")
# Result: $MY_TEST_DIR/images/grid.png
```

3. Export it
```python
outStage = Usd.Stage.CreateInMemory()
cmds.mayaUSDExport(file=outStage.GetRootLayer().identifier)
```

4. Notice that the file path was wrongly translated

```python
UsdShade.Shader.Get(outStage, "/pPlane1/mtl/lambert2SG/file1").GetInput("file").Get()
# Result: Sdf.AssetPath('/path/to/issue_files/mayaUSDExport_file_envvar_expansion/$MY_TEST_DIR/images/grid.png')
```

**Expected behavior**

mayaUSDExport should produce an asset path that is valid for USD's default resolver.

**Specs (if applicable):**

 - CentOS 7.8
 - Maya 2023.3
 - maya-usd 0.27.0
 - USD 23.11

jufrantz:
Fixes #3766 .

Adds support for environment variable expansions found in `file.fileTextureName` attributes at export, thanx to `MRenderUtil::exactFileTextureName`.

A more conservative surgical change would  be to use `MString::expandEnvironmentVariablesAndTilde`.  I opted for `exactFileTextureName` as it  aligns with file node logic, supporting other resolution options like `dirmap` and `MPxFileResolver` plugins, although it might introduce more side effects.

### DIFF
--- a/lib/usd/translators/shading/mtlxFileTextureWriter.cpp
+++ b/lib/usd/translators/shading/mtlxFileTextureWriter.cpp
@@ -298,7 +298,7 @@ void MtlxUsd_FileWriter::Write(const UsdTimeCode& usdTime)
         return;
     }
 
-    // Resolve texture path à la file node
+    // Resolve texture path like a file node
     const MString exactTextureName = MRenderUtil::exactFileTextureName(
         rawTextureName,
         /* useFrameExt = */ false,
--- a/lib/usd/translators/shading/mtlxFileTextureWriter.cpp
+++ b/lib/usd/translators/shading/mtlxFileTextureWriter.cpp
@@ -52,6 +52,7 @@
 #include <maya/MGlobal.h>
 #include <maya/MObject.h>
 #include <maya/MPlug.h>
+#include <maya/MRenderUtil.h>
 #include <maya/MStatus.h>
 #include <maya/MString.h>
 
@@ -292,9 +293,27 @@ void MtlxUsd_FileWriter::Write(const UsdTimeCode& usdTime)
         return;
     }
 
-    std::string fileTextureName(fileTextureNamePlug.asString(&status).asChar());
+    const MString rawTextureName = fileTextureNamePlug.asString(&status);
     if (status != MS::kSuccess) {
         return;
+    }
+
+    // Resolve texture path à la file node
+    const MString exactTextureName = MRenderUtil::exactFileTextureName(
+        rawTextureName,
+        /* useFrameExt = */ false,
+        /* currentFrameExt = */ MString(),
+        depNodeFn.absoluteName(),
+        &status);
+    if (status != MS::kSuccess) {
+        return;
+    }
+
+    std::string fileTextureName(exactTextureName.asChar(), exactTextureName.length());
+
+    // Fallback to raw name if maya resolution failed, eg file was not found
+    if (fileTextureName.empty()) {
+        fileTextureName.assign(rawTextureName.asChar(), rawTextureName.length());
     }
 
     const MPlug tilingAttr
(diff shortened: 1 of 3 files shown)


lib/usd/translators/shading/mtlxFileTextureWriter.cpp
```diff
@@ -292,11 +293,29 @@ void MtlxUsd_FileWriter::Write(const UsdTimeCode& usdTime)
         return;
     }
 
-    std::string fileTextureName(fileTextureNamePlug.asString(&status).asChar());
+    const MString rawTextureName = fileTextureNamePlug.asString(&status);
     if (status != MS::kSuccess) {
         return;
     }
 
+    // Resolve texture path à la file node
```
pierrebai-adsk, line 301: I know this is somewhat trivial and maybe unimportant, but that "à" requires that the file encoding be known. I suppose most OS and compilers are fine and assume UTF-8, but I'd rather we keep the contant ASCII :)

lib/usd/translators/shading/usdFileTextureWriter.cpp
```diff
@@ -265,11 +266,29 @@ void PxrUsdTranslators_FileTextureWriter::Write(const UsdTimeCode& usdTime)
         return;
     }
 
-    std::string fileTextureName(fileTextureNamePlug.asString(&status).asChar());
+    const MString rawTextureName = fileTextureNamePlug.asString(&status);
     if (status != MS::kSuccess) {
         return;
     }
 
+    // Resolve texture path à la file node
```
pierrebai-adsk, line 274: same

pierrebai-adsk (changes requested):
Just a small ASCII fix, otherwise LGTM

lib/usd/translators/shading/mtlxFileTextureWriter.cpp
jufrantz: Oops you're right, well spotted. Fixed in [bebbe7e](https://github.com/Autodesk/maya-usd/pull/3769/commits/bebbe7ea35aaf40edf26a562fce843b269afff84).

lib/usd/translators/shading/usdFileTextureWriter.cpp
jufrantz: Fixed

jufrantz:
All fixed.
Also added an assertion to strengthen the unit test, sorry for adding post review: [4863573](https://github.com/Autodesk/maya-usd/pull/3769/commits/4863573c37656b088a46fd7e8f330930da12561d).
Thank you @pierrebai-adsk

pierrebai-adsk (dismissed):

PR #3771, opened 2024-05-15
LOOKDEVX-2903 - Update for a change in MaterialX 1.38.10

JGamache-autodesk:
Handle the case where a shader input has no path, yet is a valid parameter of that node. This is a behavior that was changed in MaterialX 1.38.10.


JGamache-autodesk:
Does not affect MayaUSD, but that change is vital for LookdevX.

vlasovi (approved):
Looks good

JGamache-autodesk:
@seando-adsk ready for merge.

PR #3773, opened 2024-05-16
EMSUSD-1121 - Pasted prims should be selected

seando-adsk:
#### EMSUSD-1121 - Pasted prims should be selected
* Select only the normal pasted prims, ignoring shaders/nodegraph/materials for now.

### DIFF
--- a/lib/usdUfe/ufe/UsdClipboardCommands.cpp
+++ b/lib/usdUfe/ufe/UsdClipboardCommands.cpp
@@ -27,6 +27,8 @@
 #include <pxr/usd/usdShade/nodeGraph.h>
 #include <pxr/usd/usdShade/shader.h>
 
+#include <ufe/globalSelection.h>
+#include <ufe/observableSelection.h>
 #include <ufe/runTimeMgr.h>
 #include <ufe/sceneItemOps.h>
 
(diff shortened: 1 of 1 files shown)


lib/usdUfe/ufe/UsdClipboardCommands.cpp
```diff
@@ -484,12 +488,22 @@ void UsdPasteClipboardCommand::execute()
...
         _pasteInfos.push_back(std::move(pasteInfo));
     }
 
+    if (!itemsToSelect.empty()) {
+        // Select the newly pasted items. This matches native DCC (ex: Maya) behavior.
+        Ufe::Selection pasteSelect;
+        for (const auto& item : itemsToSelect) {
+            pasteSelect.append(item);
+        }
+        Ufe::GlobalSelection::get()->replaceWith(pasteSelect);
```
seando-adsk, line 504: Is it okay to select this way in a command? Normally I would have used [UfeSelectionUndoItem::select()](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/undo/OpUndoItems.cpp#L413), but that is in MayaUsd (not UsdUfe).

seando-adsk: Note: according to the JIRA we only select the normal prims (ignoring materials and shaders for now).

pierrebai-adsk (approved):

PR #3774, opened 2024-05-17
EMSUSD-1113 export material scope

pierrebai-adsk:
- Make it so that materials are always under a material scope.
- The material scope is either under the default prim if there is one or at the root.
- Simplify material prim name code to not use nested if and be linear instead.
- Move the material name generation into its own function.
- Fix that choosing to have no default prim does *not* instead set the default prim to be "None".
- Fix that the same localization key was used in two places.
- Add unit tests for material scopes.
- Fix other affected unit tests.


pierrebai-adsk:
While the PR seems large, most of the change are to fix existing unit tests affected by the change of where materials are exported.

lib/mayaUsd/commands/baseExportCommand.cpp
```diff
@@ -119,6 +119,10 @@ MSyntax MayaUSDExportCommand::createSyntax()
         kExportMaterialsFlag,
         UsdMayaJobExportArgsTokens->exportMaterials.GetText(),
         MSyntax::kBoolean);
+    syntax.addFlag(
+        kExportMaterialUnderPrimFlag,
+        UsdMayaJobExportArgsTokens->exportMaterialUnderPrim.GetText(),
+        MSyntax::kBoolean);
```
pierrebai-adsk, line 125: This new flag is necessary to make merge-to-USD and duplicateto-USD work. That is because they need all generated USD data to be under the merge or copied prim. The flag ensures its material are under that prim.

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -453,63 +456,77 @@ bool isSurfaceNodeType(const std::string& nodeType)
 
     return std::binary_search(sKnownSurfaces.cbegin(), sKnownSurfaces.cend(), nodeType);
 }
+
+std::string getMaterialName(
```
pierrebai-adsk, line 460: Factored out the code that generates the material name, to make the function below simpler and clearer.

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -357,9 +357,8 @@ UsdMayaShadingModeExportContext::GetAssignments() const
     return ret;
 }
 
-static UsdPrim _GetMaterialParent(
+static SdfPath _GetCommonAncestor(
```
pierrebai-adsk, line 360: Factored out the code to find the common ancestor into its own function for clarity.

plugin/adsk/scripts/mayaUsdTranslatorExport.mel
```diff
@@ -152,6 +152,11 @@ proc string mayaUsdTranslatorExport_AppendDefaultPrim(string $currentOptions, st
     if (`optionMenuGrp -exists exportDefaultPrim` == 0)
         return $currentOptions;
     
+    // If the first item, "None" is chosen, then we do *not* set the default prim.
+    int $index = `optionMenuGrp -q -select exportDefaultPrim`;
+    if ($index == 1)
+        return $currentOptions;
```
pierrebai-adsk, line 158: Fixed bug where chosing "None" for the default prim actually authored that the default prim was a prim named "None".

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
barbalt: Should we rename this function to follow the ones above? _GetMaterialName? Or is there a different logic being applied here?

barbalt (approved):
Looks good! Just a minor, not required comment about naming! Thank you!

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
AramAzhari-adsk: We usually use `_` for private / protected functions.

pierrebai-adsk: Our coding standard says to name functions with lowercase staring letters. Not all functions are following the standard, we inherited code.

AramAzhari-adsk (approved):
Looks good.

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -374,21 +373,25 @@ static UsdPrim _GetMaterialParent(
...
-    if (commonAncestor == SdfPath::AbsoluteRootPath()) {
-        return stage->GetPseudoRoot();
-    }
+    if (shaderExportLocation.IsEmpty())
+        shaderExportLocation = SdfPath::AbsoluteRootPath();
 
-    SdfPath shaderExportLocation = commonAncestor;
-    while (!shaderExportLocation.IsRootPrimPath()) {
-        shaderExportLocation = shaderExportLocation.GetParentPath();
-    }
```
dj-mcg, line 389: Hi @pierrebai-adsk - this change broke a few of our export tests since the logic to find the MaterialParent path is significantly different. Prior to this change, all materials used under prim `/Foo` would be found in `/Foo/Looks`, but now they will now either be included under the root prim ('/Looks', as a sibling to '/Foo') or will be located under the specific child prim in which the material binding is found ('/Foo/Geom/Child/Looks'), depending on whether or not we set the new export option.

This violates a few internal assumptions we have in our pipeline, and I'm wondering if it was an intentional decision to change existing and established export behavior. If so we may need to either patch this change out of our build, or introduce another export option that would preserve the behavior our pipeline expects.

thanks

dj-mcg: Also, just so I'm sure I understand what the intended behavior is:

- If I specify `exportMaterialUnderPrim = true` then the materialParent should be `/Path/to/materialbinding`
- If I specify only a `defaultPrim` in my export args, then the parent should be the default prim path
- If I specify both, should the parent be `/Path/to/materialbinding + /defaultPrimPath`?

dj-mcg: It seems like I might be able to recover the established behavior if I leave `exportMaterialUnderPrim` and I specify `defaultPrim` in my export args, but we don't typically specify defaultPrim in our export args - we allow the export process to determine it for us here (I think):
http://grok/xref/mainline/ext/mayapkg/maya_open_source/maya-usd/lib/mayaUsd/fileio/jobs/writeJob.cpp?r=2328671#565

Unfortunately we don't have access to that defaultPrim data at the time of this call

pierrebai-adsk: Hi @dj-mcg 

Yes, the changes were intentional, the designer wanted to make the location of materials more predictable. The materials now are generally under the default prim if there is one, under the root if there is none.

The exportMaterialUnderPrim was added so that our edit-as-Maya / merge-to-USD workflow work, since that assumed everything was under the edited prim.

As you can see in this PR, we also had to adjust multiple tests to take into account the new location. For your case, the export-under-prim flag does not do the job of restoring the behavior that you expect?

dj-mcg:
Part of this change was to remove the logic to find the RootPrim path if the Common Ancestor was neither empty or the AbsoluteRootPath, which our export process depended on to determine where the Materials scope should live. Now it appears that we either create the Materials scope under the prim which actually uses the binding if `exportMaterialUnderPrim` is true (which will typically be a nested child prim for our models), at the root scope, or under the Default Prim but _only if it is specified in the export args_.

As I mentioned, we aren't setting the `defaultPrim` export arg in our pipeline when building models, we compute it here:
https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/fileio/jobs/writeJob.cpp#L567

It feels wrong to now require this export arg to always be set. Can we restore the logic to create the materials scope under the root prim, perhaps governed under _another_ export arg?

pierrebai-adsk:
I personally don't have objection, but I'll let the designers chime in.

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
scottrenaud: Hey @dj-mcg, in the meantime while we look at this issue on our end, I'll invite you to take a look at the first part of github #3674 where I describe the existing behavior versus the new behavior in terms of where materials end up.   Lemme know if you have any thoughts/concerns, thanks!

dj-mcg: Hi @scottrenaud - thanks for the link. The proposal made complete sense to me, but my only concern with the implementation is that the Default Prim needs to be explicitly specified in exportArgs, and it does not use the same logic of falling back to the "UsdRootPrim" if this export arg is empty as the code in `UsdMaya_WriteJob::_FinishWriting`. Perhaps this was an intentional decision and I made an incorrect assumption when originally reading the proposal - If so I wonder if we can add this behavior via another exportArg?

pierrebai-adsk:
We went the route of removing `exportMaterialUnderPrim` and adding `legacyMaterialScope` which uses the exact code that was used before to ensure backward compatibility. The legacy flag is off by default and need to be explicitly used. See this PR: https://github.com/Autodesk/maya-usd/pull/3810

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
scottrenaud: Hey @dj-mcg, just to follow up here, as mentioned in https://github.com/Autodesk/maya-usd/pull/3810#discussion_r1631464040 you're shining light on the fact that this is a gap, we'll create a ticket for this internally and discuss it with the team, thanks!

Issue #3101, opened 2023-05-17
UsdSkel BlendShape and BlendShape Animation Import


gelshoffNV:
**Description**
Support importing blendShapes and additionally any animation assigned from SkelAnimation. 

**Additional context**
BlendShapes have been recently added to maya-usd export. It would be additionally good to support importing USD BlendShapes back into Maya. Goal would be to support full round-trip originating from Maya blendShape + animation -> USD Skel + BlendShape + SkelAnimation -> Maya blendShape + animation.

wallworm:
Thank you for your input. This is on our roadmap and we intend to work on it for both Maya and Max.

gelshoffNV:
I figured, but wanted to track it at least, is there a public roadmap I can review?

santosg87:
Unfortunately, we don't currently have a public roadmap - we are still investigating the possible work around this, and we can update this when we have more information available.

wallworm:
I wanted to give an update that we are currently working on blendshape import.

PR #3775, opened 2024-05-17
Add support to import Usd Blendshape as Maya blendshapes

barbalt:
This PR adds support to import UsdSkelBlendShapes into Maya BlendShapes

This PR solves #3101

### DIFF
--- a/lib/mayaUsd/fileio/translators/translatorBlendShape.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorBlendShape.cpp
@@ -26,6 +26,7 @@
 #include <pxr/usd/usdSkel/blendShape.h>
 #include <pxr/usd/usdSkel/root.h>
 
+#include <maya/MFloatVectorArray.h>
 #include <maya/MFnAnimCurve.h>
 #include <maya/MFnBlendShapeDeformer.h>
 #include <maya/MFnGeometryFilter.h>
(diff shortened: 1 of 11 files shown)


lib/mayaUsd/fileio/translators/translatorSkel.cpp
```diff
@@ -1220,17 +1210,6 @@ bool UsdMayaTranslatorSkel::CreateSkinCluster(
         CHECK_MSTATUS_AND_RETURN(status, false);
     }
 
-    // Connect restMesh.outMesh -> groupParts->inputGeometry
```
barbalt, line 1223: Those were moved as they are to L1311. I don't think we need a rest pose if we have blendShapes attached to it. We'll use the blendShapes instead

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -315,14 +311,12 @@ MStatus mayaGetBlendShapeInfosForMesh(
         // NOTE: (yliangsiew) Because this can end up grabbing the wrong plug (i.e.
         // blendShape.weight[1]), we double-check here and advance the iterator if necessary.
         if (!outputGeomElemPlug.isElement()) {
-            itDg.next();
```
barbalt, line 318: This is inside a for loop which already calls .next(). 
My understand is that there might be a hidden bug here where we can accidentally skip some elements.

lib/usd/translators/meshWriter_BlendShapes.cpp
```diff
@@ -315,14 +311,12 @@ MStatus mayaGetBlendShapeInfosForMesh(
...
             continue;
         }
         MPlug outputGeomPlug = outputGeomElemPlug.array(&stat);
         CHECK_MSTATUS_AND_RETURN_IT(stat);
         MString outputGeomPlugName = outputGeomPlug.partialName(0, 0, 0, 0, 0, 1);
         if (outputGeomPlugName != "outputGeometry") {
-            itDg.next();
```
barbalt, line 325: This is inside a for loop which already calls .next().
My understand is that there might be a hidden bug here where we can accidentally skip some elements.

lib/mayaUsd/fileio/translators/translatorBlendShape.cpp
```diff
@@ -0,0 +1,299 @@
...
+    const MObject&         originalShape,
+    const MObject&         parentTransform,
+    const MPointArray&     originalPoints,
+    const VtIntArray&      pointIndices,
+    const std::string      name,
+    const VtVec3fArray     offsetArray,
+    int                    blendShapeTargetIndex,
+    float                  weight)
+{
+    MFnMesh ibMeshFn;
```
pierrebai-adsk, line 53: just curious: what does ib stand for? input blend?

lib/mayaUsd/fileio/translators/translatorBlendShape.cpp
```diff
@@ -0,0 +1,299 @@
...
+            MPoint(original.x + offset[0], original.y + offset[1], original.z + offset[2]), index);
+    }
+    ibMeshFn.setPoints(ibPoints);
+
+    MFnDependencyNode ibDepNodeFn;
+    ibDepNodeFn.setObject(newShape);
+    ibDepNodeFn.setName(MString(name.c_str()));
+
+    blendShapeFn.addTarget(originalShape, blendShapeTargetIndex, newShape, weight);
+    ibMeshFn.setIntermediateObject(true);
```
pierrebai-adsk, line 73: I suppose the order of setting it as intermediate after instead of before adding it as a target does not matter.

lib/mayaUsd/fileio/translators/translatorBlendShape.cpp
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    std::vector<unsigned int> correctBlendShapeIndices;
+    correctBlendShapeIndices.reserve(skinnedBlendShapes.size());
+    // The animation prim can hold blendShapes from several mesh prims.
+    // So it's necessary to find the the indices only for those on the current mesh prim.
+    for (const TfToken& token : skinnedBlendShapes) {
+        const auto iter = std::find(animBlendShapes.begin(), animBlendShapes.end(), token);
+        const auto idx = static_cast<int>(std::distance(animBlendShapes.begin(), iter));
+        if (idx >= 0) {
```
pierrebai-adsk, line 226: I'll be honest and may not understand the code properly. I thought std::find would return end() if not found, but end() - begin() would be greater than zero? Why not compare iter with end()?

lib/mayaUsd/fileio/translators/translatorBlendShape.cpp
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    MObject  transformNode = context.GetMayaNode(skinnedPrim.GetPath(), false);
+    MDagPath shapeDagPath;
+    status = MDagPath::getAPathTo(transformNode, shapeDagPath);
+    shapeDagPath.extendToShape();
+    MObject shapeObject = shapeDagPath.node(&status);
+
+    if (shapeObject.isNull() || !shapeObject.hasFn(MFn::kMesh)) {
+        return false;
```
pierrebai-adsk, line 258: I'm always unsure why in the import and export code, sometimes things return true on errors (or when not finding things), sometimes false.

pierrebai-adsk (changes requested):
Just a few questions.

lib/mayaUsd/fileio/translators/translatorBlendShape.cpp
barbalt: Yea, not sure what would be the idea behavior here. I try to add errors/warnings when necessary

lib/mayaUsd/fileio/translators/translatorBlendShape.cpp
barbalt: I'll add comments to try to clarify this a bit further.
Basically, what we're looking for is the index where token was found. So after calling find, we check the distance from the start and cache that number.

lib/mayaUsd/fileio/translators/translatorBlendShape.cpp
barbalt: In this case, what really matter is how we setup the plugs connection. The order in which the nodes are created aren't essential.

lib/mayaUsd/fileio/translators/translatorBlendShape.cpp
pierrebai-adsk: Yes, but if find returns end(), we should not cache it?

pierrebai-adsk (approved):

gelshoffNV:
Thanks! Excited to try this out!

PR #3776, opened 2024-05-17
Add remapUVSetsTo option to UsdImport

dj-mcg:
Our users have requested that UV Sets that were renamed on export round trip cleanly on import. The Maya:name metadata accomplishes this, but only for assets that were exported after the introduction of this code. This allows us to specify a remapping for UV Set names on import so older assets work.

If the Maya:name metadata is found, this remapping is not used.


lib/mayaUsd/fileio/utils/readUtil.h
```diff
@@ -42,6 +42,16 @@ struct UsdMayaReadUtil
     MAYAUSD_CORE_PUBLIC
     static bool ReadFloat2AsUV();
 
+    /// Returns the imported name of a UV set at \p uvIndex according to the renaming rule and the
```
pierrebai-adsk, line 45: This function description does not seem accurate and refer to a wrong parameter name.

pierrebai-adsk (commented):
Just a function header comments that does not seem right, otherwise LGTM.

lib/mayaUsd/fileio/utils/readUtil.h
dj-mcg: Oh, excellent catch - I changed the function signature but forgot to change the comment - I'll fix@

dj-mcg:
I fixed the comment in place - thanks for catching!

PR #3786, opened 2024-05-27
EMSUSD-1134 export assigned materials

pierrebai-adsk:
Modify the export command:
- Add new export flag "exportAssignedMaterials" (-ama).
- Default to be turned on.
- Expose it in Python.
- Document it.
- Add a checkbox UI for it in the export options.

Modify how material assignments are found:
- Bundle all information about assignments for a material in a single struct.
- Add a flag to the info set if the material is used anywhere, even by meshes not being exported.
- Add a flag to the info set if the material is used by something in the selection.
- Also check any ancestor of the mesh for the selection inclusion since selection in the viewport selects the parent transform.
- Also, by default, the outliner shows the parent transforms, not the shape/mesh, so selection there would also select an ancestor.
- Modify the GetAssignments function to fill those flags.

Modify how materials are filtered:
- In the filter function for materials, add the logic for assigned-materials.
- Use the flags about assignments and assignments-in-selection to determine if a material should be exported.
- When exporting assigned materials, if the material is not used at all, then the material is not exported.
- When exporting the selection, if a material was used by something in the selection, export it even if the user was something not exported, like when the meshes are off.

Add unit tests for when assigned materials are off. (The case when it is on was already tested by existing tests.)


lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -300,12 +300,20 @@ UsdMayaShadingModeExportContext::GetAssignments() const
...
 
+        ret.hasAnyAssignment = true;
+
+        // Note: the connections will be to the mesh, but the selection often
+        //       contains the transform above the mesh or a higher-up ancestor,
+        //       so also check the ancestors of the mesh.
+        for (MDagPath walkUpPath(dagPath); walkUpPath.length() > 0; walkUpPath.pop())
+            if (GetExportArgs().fullObjectList.hasItem(walkUpPath))
+                ret.hasAnyAssignmentInSelection = true;
+
```
pierrebai-adsk, line 316: To support the logic of determining if a material should be exported, we need to know if it is used in the scene and if it used by something that was selected. This is what the new code above does.

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -300,12 +300,20 @@ UsdMayaShadingModeExportContext::GetAssignments() const
...
             continue;
         }
 
+        ret.hasAnyAssignment = true;
+
+        // Note: the connections will be to the mesh, but the selection often
+        //       contains the transform above the mesh or a higher-up ancestor,
+        //       so also check the ancestors of the mesh.
+        for (MDagPath walkUpPath(dagPath); walkUpPath.length() > 0; walkUpPath.pop())
+            if (GetExportArgs().fullObjectList.hasItem(walkUpPath))
```
barbalt, line 314: Is there a policy in code guidelines for one line if cases like if they should have {} or not? I couldn't find one. This is minor, anyways

seando-adsk: There is no policy for this (no {} is allowed), other than the formatting done by clang-format.

barbalt (approved):

PR #3787, opened 2024-05-27
LOOKDEVX-2653 - Fix UDIMs for custom textures

JGamache-autodesk:
Make sure the uv_scale and us_offset inputs of an image node inside a NodeDef/NodeGraph combo gets propagated to top level.

Fixes UDIMs on:
- MaterialX UsdUvTexture
- glTf image nodes

The added unit test uses these two image types.

The uv_scale and offset will also be propagated for these nodes:
- tiledimage
- triplanar projection But it will not affect them since these nodes never sample textures outside of the 0-1 UV range.


lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -1474,8 +1474,17 @@ _LoadUdimTexture(const std::string& path, bool& isColorSpaceSRGB, MFloatArray& u
         return nullptr;
     }
 
+    // The UDIM scale and offset cannot be retrieved from a MTexture*, so we need to remember
+    // what was the result the last time we created one. This is required because a pointer
+    // in the global and local cache can go stale, while Maya still remembers about that
+    // texture.
```
JGamache-autodesk, line 1480: Found while playing around with UDIMs. It was very easy to break them. Just open the regular "stage_with_udims.usda" in Maya without this fix and duplicate the grid. This will cause the scale and offset to be lost and break the render.

vlasovi: Should we clean kScaleOffsetCache somewhere? Should it really be static?

JGamache-autodesk: Can only be cleaned if we know for sure the texture manager forgot about this texture. I do not see any API for this.  Should not get very big, and can not get stale since we only fetch the cached value if the MTextureManager found it.

JGamache-autodesk: Would be nice if the information could be retrieved from the MTextureDescription returned by MTexture::getDescription().

JGamache-autodesk:
Official fix submitted to MaterialX 1.39 here: https://github.com/AcademySoftwareFoundation/MaterialX/pull/1840

vlasovi (dismissed):

lib/mayaUsd/render/MaterialXGenOgsXml/Nodes/MayaHwImageNode.cpp
```diff
@@ -0,0 +1,76 @@
...
+        ncGraphNode->addPropagatedInput(node, UV_SCALE);
+    }
+    ShaderInput* offsetInput = node.addInput(UV_OFFSET, Type::VECTOR2);
+    offsetInput->setValue(Value::createValue<Vector2>(Vector2(0.0f, 0.0f)));
+    if (ncGraphNode) {
+        ncGraphNode->addPropagatedInput(node, UV_OFFSET);
+    }
+}
+
+void MayaHwImageNode::setValues(const Node& node, ShaderNode& shaderNode, GenContext& context) const
```
frohnej-adsk, line 43: [Question] This method looks exactly like `HwImageNode::setValues()`. Would it make sense to have `MayaHwImageNode` extend `HwImageNode` and only override `addInputs()`? Or am I missing something?

vlasovi (approved):

frohnej-adsk (approved):
I'm not very familiar with the MaterialX shader gen yet, but it looks good to me. This will be very useful! I only have two minor comments.

vlasovi (approved):

JGamache-autodesk:
@seando-adsk Ready for merge.

frohnej-adsk (approved):

PR #3788, opened 2024-05-27
EMSUSD-1264 add tooltip for lookdevx button in material section in AE

pierrebai-adsk:


lib/mayaUsd/resources/ae/usdschemabase/material_custom_control.py
```diff
@@ -76,7 +76,7 @@ def _hasLookdevX():
         '''
```
AramAzhari-adsk, line 76: I didn't realize that we have soft dependency on LookdevX. Is this necessary or could we expose a callback for LookdevX to add their own ?

AramAzhari-adsk (approved):

Issue #3778, opened 2024-05-22
Ability to route prim metadata edits


jufrantz:
**Is your feature request related to a problem? Please describe.**

Currently, maya-usd supports routing various operations to a USD layer. For instance we can route attribute edits and target a layer based on the edited prim and attribute name. However, it does not yet support routing `prim metadata` which would be beneficial for some workflows.

Our first usecase is routing variant selections: some of our variants configure the stage just for the artist work and should be routed to the stage's session, while others need to be public and target identified layers. 

**Describe the solution you'd like**

We would like to be able to register a custom editRouter for "primMetadata" edits,  similar to the current support for attributes. the router could consider the metadata name and the edited key path for dictionary-valued metadata like `VariantSelection`.

A fist look of the UFE code suggests that it could be added to the following commands:
- SetVariantSelection
- Set kind
- Toggle instanceable
- Prim reorder
- Add/Clear payload/reference
- Set/Clear SceneItemMetadata

**Describe alternatives you've considered**

Overriding standart context ops to execute custom commands, eg a custom SetVariantSelectionCommand.

**Additional context**

We have implemented it on our end for a custom command. If we haven't overlooked anything, we should be able to file a PR straightforwardly.  

Would you consider accepting this enhancement as a PR?

PR #3716, opened 2024-04-12
EMSUSD-1091 Save and restore non local edit target layer and anonymous layer

csyshing:
This PR aims to fix [issue-3637](https://github.com/Autodesk/maya-usd/issues/3637), the changes include three different but still highly related problems all together, I have split them to cover their own problems and added unit test accordingly:

- 0c28b99ab78ffa99d694c9f6d2df3481d2d292d3:
  -  Fixes the non local edit target layer issue by extending `stage->GetLayerStack()` to be `stage->GetUsedLayers()` to include reference layers and their sub layers which are not visible in local layer stack;
  - The main code to locate a non local layer when restoring is in [`copyTargetLayerToAttribute()`](https://github.com/Autodesk/maya-usd/commit/0c28b99ab78ffa99d694c9f6d2df3481d2d292d3#diff-3003792a8ba0c46971021e5768a07b291f62eae545db8b0ffda15b215e1f89faR123) and [`findPrimNode()`](https://github.com/Autodesk/maya-usd/commit/0c28b99ab78ffa99d694c9f6d2df3481d2d292d3#diff-3003792a8ba0c46971021e5768a07b291f62eae545db8b0ffda15b215e1f89faR40) functions in file `utils/targetLayer.cpp`, it has assumption that: if a non local edit target layer can be constructed by a PrimNode, there must be at least one valid prim path contributes to the final composition, which must be able to find it at the time when reopening, this is the reason the code serializes one reference prim path in [`copyTargetLayerToAttribute()` function at  `utils/targetLayer.cpp#L141`](https://github.com/Autodesk/maya-usd/commit/0c28b99ab78ffa99d694c9f6d2df3481d2d292d3#diff-3003792a8ba0c46971021e5768a07b291f62eae545db8b0ffda15b215e1f89faR141), and will need it to reconstruct the edit target in [`getEditTargetFromAttribute()` function at line 233](https://github.com/Autodesk/maya-usd/commit/0c28b99ab78ffa99d694c9f6d2df3481d2d292d3#diff-3003792a8ba0c46971021e5768a07b291f62eae545db8b0ffda15b215e1f89faR233).
  - We use this logic internally in our pipeline and it's been working fine for us for a few years, we haven't had other use cases to verify if this logic is also useful for other studios, that said, it's arguable a generic solution to fix this 'non local edit target layer' issue.
  - A unit test `test/lib/mayaUsd/fileio/testNonLocalEditTargetLayer.py` has been provided to demo the typical use case we have.
[...]

PR #3790, opened 2024-05-28
Prim metadata editRouting

jufrantz:
Following #3778, this proposal introduces support for a new registrable `primMetadata` router, designed similarly to attribute routing. A unique router will receive the USD metadata name in its context. For `customData` and `variantSelection`, the router will also receive a dictionary key path, such as the `variantSet` name for `variantSelection`.

These commands support this routing:
- `UsdUndoToggleInstanceableCommand`
- `UsdUndoToggleActiveCommand`
- `UsdUndoSetKindCommand`
- `UsdUndoAddReferenceCommand`
- `UsdUndoAddPayloadCommand`
- `UsdUndoClearPayloadsCommand`
- `UsdUndoClearReferencesCommand`
- `SetVariantSelectionCommand`
- `ClearSceneItemMetadataCommand`
- `SetSceneItemMetadataCommand`

I still need to add unit tests and update EditRouting.md. But before going further, I wanted to push these initial changes to ensure they fit the project.


jufrantz:
Hello @seando-adsk, @pierrebai-adsk,
Is there anything I can improve before the code review? I'm happy to adjust the design or remove certain aspects if they don't align with your vision. Of course, it's also fine to close it if the timing isn't right or if this is already planned on your end. 
Julien

seando-adsk:
@jufrantz We have been discussing this PR internally. I'll collect all our notes and get back to you.

Sean

seando-adsk:
@jufrantz Sorry for the delay. We have discussed this internally and your changes do fit with ones we had upcoming, such as being able to edit route the custom data.

One question that came up internally was, `it only support local layer routing and another user added non-local layer routing, so we might want to support it also for metadata.`

And just for your information, `both visibility and locking are metadata but have their own specific edit router and would not use this metadata edit router. We aren't sure if we would want to also use the metadata edit router for them, either as a replacement (but that would break existing users) or in addition to the metadata edit router.`

And yes new unit tests will be required to go along with these changes.

Thank-you,
Sean

jufrantz:
Thank you @seando-adsk for the feedbacks. I'll free up some time next week to revisit this, I'll see how it may interact with non-local targets and also check other metadata routing that I may have missed.
Julien

jufrantz:
Thank you once again @seando-adsk, I did have a deeper look at your comments,

Regarding non-local layers routing,I could not find changes related to it. However, I found #3716 that addresses non-local layer persistence in the Maya scene. 3716 and this PR should be compatible: if a non-local layer is used as editTarget before metadata editRouting (to a local layer), the non-local target will be correctly restored. From my understanding, there is currently no editRouter that can target a non-local layer, as routing callbacks respond with an SdfLayer and not a whole UsdEditTarget. Correct me if I am wrong, this might be addressed more globally in another PR.

Regarding other metadata routing, I found `mayaLock` and `mayaSelectability`, these metadata only apply to properties. When set with `UsdAttributeHolder`, they will be routed with the `attribute` router. I believe it is fine that they are not routed by `primMetadata` since its intent was only for metadata applied to prim. While it could be more symmetric to have a more specific `attributeMetadata` route, IMHO it makes sense that attribute routing affects attribute metadata the same way it affects attribute values, and I don't see right now a use case for a specific attributeMetadata route. `visibility` is routed with `attribute` or `visibility` and should not collide neither, it looks all good to me.

Concerning backward compatibility, it appears possible to chain editRouters. eg if an `attributeMetadata` router is introduced in the future, the `attribute` router could still be executed first at least for a transition period. Client code with `attribute` routing but no `attributeMetadata` routing would not break.

If the current design of `primMetadata` is acceptable to you, I will complete the PR with unit tests. Otherwise, I propose to change it and address only our main concern with a specific `variantSelection` editRouter.

Best,
Julien

pierrebai-adsk (changes requested):
It would be nice to have at least one unit test to verify it all works.

jufrantz:
Yes. As mentioned in my first message, it was an initial drop. Since the PR is quite old, I wasn't sure if there was interest in this addition. I'll make some time next week to complete it.

lib/usdUfe/base/tokens.h
```diff
@@ -39,13 +39,16 @@ namespace USDUFE_NS_DEF {
     /* Stage received in the context of some router  */ \
     ((Stage, "stage"))                                  \
     ((EditTarget, "editTarget"))                        \
+    /* Metadata key path received in the context     */ \
+    ((KeyPath, "keyPath"))                              \
```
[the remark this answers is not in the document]
jufrantz: Yes. More generally it provides the key for all dictionary-valued metadata. Currently there are only 2 cases:
- For `variantSelection`, it will simply provide the single-level`variantSet` name. 
- For USD `customData` edited through UFE metadata, `keyPath` will be composed of the UFE metadata group and name, eg "myorg:mycategory:foo."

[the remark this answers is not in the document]
AramAzhari-adsk: @pierrebai-adsk regarding the keyPath convention, are there other instances that we format the use case this way?

Alternatively we could have grouped metadata as we do in some of the UFE interfaces (setGroupMetadata) where the storage of the group is not concatenated with the rest of the path.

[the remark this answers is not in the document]
pierrebai-adsk: It is the standard way to specify a composite metadata key path in USD.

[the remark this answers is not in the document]
jufrantz: As the context seems more in the USD world - for example, it gives a prim instead of a scene item - I believe it's better to express the path the USD way,

lib/usdUfe/ufe/UsdUndoSetSceneItemMetadataCommand.cpp
```diff
@@ -80,7 +84,8 @@ void SetSceneItemMetadataCommand::setGroupMetadata()
         PXR_NS::TfToken        fullKey(prefixlessGroupName + std::string(":") + _key);
```
jufrantz, line 84: @pierrebai-adsk, @AramAzhari-adsk,
Should I enable edit routing here? "SessionLayer-" customData is always expected to be authored in the session, or is this a default behavior that could be overriden ? 
I noticed this is used for `Autodesk:Text Color`, it would be nice if we could author it in an identified layer when changed from AE.

pierrebai-adsk: One option would be to call getLayer() on the edit router context and only set it to the session layer if it returns nullptr. OTOH, for composite operation or if sessions metadata is set in the middle of another operation (say, due to a notification), then it might end-up in a layer by accident. I think my apprehension are unwarranted though, and we could just do what I suggested without concerning ourselves with strange corner cases, which after all can affect all edit routing.

jufrantz: Thank you @pierrebai-adsk,
My question was about the expected behavior for "SessionLayer-" metadata. I was wondering if it's acceptable to allow them to be authored in layers other than the session layer.
As for the implementation, my initial thought was to apply edit routing after the existing `UsdEditContext` that targets the session layer. Would this approach work?:

```c++
    if (isSessionLayerGroupMetadata(_group, &prefixlessGroupName)) {
        PXR_NS::UsdEditContext editCtx(_stage, _stage->GetSessionLayer());
        {
            PXR_NS::TfToken               fullKey(prefixlessGroupName + std::string(":") + _key);
            PrimMetadataEditRouterContext editRouterCtx(prim, PXR_NS::SdfFieldKeys->CustomData, fullKey);
            prim.SetCustomDataByKey(fullKey, ufeValueToVtValue(_value));
        }
    } 
```

jufrantz: Re-thinking about it, I like your suggestion as it avoids unnecessary changes to the stage edit target.
I could do that efficiently by adding a `fallbackLayer` argument to `PrimMetadataEditRouterContext`:
```c++
    if (isSessionLayerGroupMetadata(_group, &prefixlessGroupName)) {
        PrimMetadataEditRouterContext ctx(
            prim,
            PXR_NS::SdfFieldKeys->CustomData,
            fullKey,
            /*fallbackLayer=*/_stage->GetSessionLayer());
        prim.SetCustomDataByKey(fullKey, ufeValueToVtValue(_value));
    }
```

seando-adsk: By default the color data is being written to the SessionLayer since we set the group here: https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py#L673

But we always intended it to be editRoutable (from your PR). So yes please make it edit routable.

Sean

jufrantz:
> It would be nice to have at least one unit test to verify it all works.

Done.

lib/usdUfe/ufe/UsdUndoSetSceneItemMetadataCommand.cpp
jufrantz: Done.

pierrebai-adsk (approved):
Very well done

pierrebai-adsk:
Ah, you will have to reformat the files that were rejected by the linter...

jufrantz:
Oops! It should be all good now. 
Thank you @pierrebai-adsk.

AramAzhari-adsk (approved):

pierrebai-adsk (dismissed):

jufrantz:
Hello,
I fixed unit-test for Maya<2025, it was not compatible with USD<23.11. The test now passes on Maya 2024:  c0f8a94c45573a3b76fb3c0d3df17705a1ef50fc.
I also made a last-minute update: while testing AE, I noticed that `SetKind` and `SetSceneItemMetadata` were not enforcing edit restrictions. I’m including it in the PR, aligning with other commands that set metadata values: b317f571a50690fd9d66f2f310aa887a111b4f87  ca1ebb730cdb0d5b4ad6baf2aa019825165e0103.
Julien

AramAzhari-adsk (approved):

pierrebai-adsk (approved):

seando-adsk (approved):

PR #3791, opened 2024-05-28
EMSUSD-1122 - As a user, I'd like to copy and then paste a prim as a sibling

seando-adsk:
#### EMSUSD-1122 - As a user, I'd like to copy and then paste a prim as a sibling
* Added selection observer to UsdClipboard to track when to reset the "paste as sibling".
* Fixed bug with cut command which wasn't removing the cut item from selection. Added unit test for this.
* New logic in paste command (for regular prims only) to paste as a sibling of the copy rather than into the paste target if the selection wasn't changed between copy and paste.
* Overrode the paste command execute to set a selection guard since the paste command selects newly pasted prims.
* New unit test for pasting multiple times to test paste as sibling.

### DIFF
--- a/lib/usdUfe/ufe/UsdClipboard.cpp
+++ b/lib/usdUfe/ufe/UsdClipboard.cpp
@@ -21,12 +21,42 @@
 #include <pxr/usd/usd/usdcFileFormat.h>
 #include <pxr/usd/usdShade/connectableAPI.h>
 
+#include <ufe/globalSelection.h>
+#include <ufe/observableSelection.h>
+#include <ufe/observer.h>
+#include <ufe/selectionNotification.h>
+
 #if (__cplusplus < 201703L)
 #error "Compiling ClipboardHandler requires C++17"
 #endif
 #include <filesystem> // Requires C++17
 
 namespace USDUFE_NS_DEF {
+
+// TufeSelectionObserver
+//
+class UsdClipboard::TufeSelectionObserver : public Ufe::Observer
+{
+public:
+    TufeSelectionObserver(UsdClipboard& cb)
+        : Ufe::Observer()
+        , _clipboard(cb)
+    {
+    }
+
+    void operator()(const Ufe::Notification& notif)
+    {
+        // EMSUSD-1122 - As a user, I'd like to copy and then paste a prim as a sibling
+        // If the selection hasn't changed between the copy and the paste, then we'll
+        // paste into the parent of each destination parent item.
+        if (dynamic_cast<const Ufe::SelectionChanged*>(&notif)) {
+            _clipboard.ufeSelectionChanged();
+        }
+    }
+
+private:
+    UsdClipboard& _clipboard;
+};
 
 // Ensure that UsdClipboard is properly setup.
 static_assert(!std::is_copy_constructible<UsdClipboard>::value);
@@ -56,6 +86,8 @@ void UsdClipboard::setClipboardData(const PXR_NS::UsdStageWeakPtr& clipboardData
         throw std::runtime_error(error);
     }
 
+    setPasteAsSibling();
+
     // Unload the stage, otherwise when we try to set and get the next clipboard data we end up with
     // the old stage.
     clipboardData->Unload();
@@ -80,6 +112,19 @@ PXR_NS::UsdStageWeakPtr UsdClipboard::getClipboardData()
     clipboardStage->Reload();
 
     return clipboardStage;
+}
+
+void UsdClipboard::setPasteAsSibling()
+{
+    // Create a Ufe selection observer. If the selection changes after
+    // the copy or paste (not from the paste command itself though) we'll
+    // paste as sibling, rather than into paste target.
+    // This mimics the duplicate behavior.
+    if (!_ufeSelObserver) {
+        _ufeSelObserver = std::make_shared<TufeSelectionObserver>(*this);
+        Ufe::GlobalSelection::get()->addObserver(_ufeSelObserver);
+        _pasteAsSibling = true;
+    }
 }
 
 void UsdClipboard::setClipboardPath(const std::string& clipboardPath)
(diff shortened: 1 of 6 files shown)


lib/usdUfe/ufe/UsdClipboard.cpp
```diff
@@ -21,13 +21,43 @@
...
+
+    void operator()(const Ufe::Notification& notif)
+    {
+        // EMSUSD-1122 - As a user, I'd like to copy and then paste a prim as a sibling
+        // If the selection hasn't changed between the copy and the paste, then we'll
+        // paste into the parent of each destination parent item.
+        if (dynamic_cast<const Ufe::SelectionChanged*>(&notif)) {
+            _clipboard.ufeSelectionChanged();
+        }
+    }
```
seando-adsk, line 55: Used to track if there is a selection change between the copy and paste commands.

lib/usdUfe/ufe/UsdClipboard.cpp
```diff
@@ -116,4 +161,15 @@ void UsdClipboard::cleanClipboardStageCache()
...
+void UsdClipboard::ufeSelectionChanged()
+{
+    if (!_inSelectionGuard) {
+        // Any ufe selection changed event that occurs between the copy and paste (that didn't
+        // come from the paste itself) means that the user has selected what they want to be
+        // the paste target.
+        _pasteAsSibling = false;
+        _ufeSelObserver.reset();
+    }
+}
```
seando-adsk, line 173: The selection guard is set from the paste command since it actually modifies the selection by selecting the newly pasted prims. We want to ignore that selection. But any other selection will remove the observer and reset the paste as sibling.

lib/usdUfe/ufe/UsdClipboardCommands.cpp
```diff
@@ -320,9 +321,12 @@ void UsdCutClipboardCommand::execute()
     const auto& sceneItemOpsHandler
         = Ufe::RunTimeMgr::instance().sceneItemOpsHandler(UsdUfe::getUsdRunTimeId());
 
-    // Step 2. Delete the selected items.
+    auto ufeSn = Ufe::GlobalSelection::get();
+
+    // Step 2. Deselect then delete the selected items.
     for (auto&& item : _selection) {
         if (auto usdItem = downcast(item)) {
+            Ufe::SelectionRemoveItem::createAndExecute(ufeSn, item);
```
seando-adsk, line 329: Fixed a bug with the cut command not actually removing the cut item from the selection. The Ufe global selection still contained the path of the cut prim. If you asked Maya for the Ufe selection it would return empty string because it could not create an item from the path (since the prim is actually deleted). But the Ufe selection still contained that one invalid path.

lib/usdUfe/ufe/UsdClipboardCommands.cpp
```diff
@@ -480,20 +497,62 @@ void UsdPasteClipboardCommand::execute()
...
+            if (pasteAsSibling) {
+                auto hier = Ufe::Hierarchy::hierarchy(dstParentItem);
+                auto parentItem = hier ? hier->parent() : nullptr;
+                auto newPasteTarget = downcast(parentItem);
+                if (newPasteTarget) {
+                    pasteTarget = newPasteTarget;
+                } else {
+                    pasteAsSibling = false;
+                }
+            }
```
seando-adsk, line 510: When pasting as sibling get the parent of the destination item so when pasting the new prim will be a sibling.

lib/usdUfe/ufe/UsdClipboardCommands.cpp
```diff
@@ -480,20 +497,62 @@ void UsdPasteClipboardCommand::execute()
...
+                    = UsdUndoDuplicateSelectionCommand::create(clipboardPrims, pasteTarget);
+                duplicateCmd->execute();
+                appendToVector(
+                    duplicateCmd->targetItems(),
+                    pasteAsSibling ? extraPasteInfo.successfulPastes : pasteInfo.successfulPastes);
+                auto tmpTargetItems = duplicateCmd->targetItems();
+                _targetItems.insert(
+                    _targetItems.end(), tmpTargetItems.begin(), tmpTargetItems.end());
+                _itemsToSelect.insert(
+                    _itemsToSelect.end(), tmpTargetItems.begin(), tmpTargetItems.end());
```
seando-adsk, line 540: Extra paste info to add since we have a different paste target.

lib/usdUfe/ufe/UsdClipboardCommands.cpp
seando-adsk: Only applies to regular prims and not materials or shaders.

lib/usdUfe/ufe/UsdClipboardCommands.cpp
```diff
@@ -480,20 +497,62 @@ void UsdPasteClipboardCommand::execute()
...
+            }
         }
 
         // Add the paste info.
         _pasteInfos.push_back(std::move(pasteInfo));
     }
 
+    if (!pasteAsSibling) {
+        _clipboard->setPasteAsSibling();
+    }
```
seando-adsk, line 554: This case is when the user selects an item, does copy, then selects a new item and does paste. We want to continually paste into this newly selected item.

lib/usdUfe/ufe/UsdClipboardHandler.cpp
```diff
@@ -66,25 +66,61 @@ UsdClipboardHandler::Ptr UsdClipboardHandler::create()
...
+        return std::make_shared<UsdPasteClipboardCommandWithSelection>(dstParentItems, clipboard);
+    }
+};
+
 Ufe::PasteClipboardCommand::Ptr
 UsdClipboardHandler::pasteCmd_(const Ufe::SceneItem::Ptr& parentItem)
 {
-    return UsdUfe::UsdUndoSelectAfterCommand<UsdUfe::UsdPasteClipboardCommand>::create(
-        parentItem, _clipboard);
+    return UsdPasteClipboardCommandWithSelection::create(parentItem, _clipboard);
```
seando-adsk, line 118: Use new class above which overrides the execute method.

lib/usdUfe/ufe/UsdClipboardHandler.cpp
```diff
@@ -66,25 +66,61 @@ UsdClipboardHandler::Ptr UsdClipboardHandler::create()
...
+    {
+        _clipboard->_inSelectionGuard = true;
+        try {
+            Parent::execute();
+        } catch (...) {
+            _clipboard->_inSelectionGuard = false;
+            throw;
+        }
+        _clipboard->_inSelectionGuard = false;
+    }
```
seando-adsk, line 102: Override the execute method of paste so that we can tell the clipboard to set a selection guard (to ignore the selection change that will come from the paste itself).

pierrebai-adsk (approved):

PR #3799, opened 2024-05-31
EMSUSD-1134 fix maya-hydra compilation.

pierrebai-adsk:


barbalt (approved):

pierrebai-adsk:
@seando-adsk This is a fix for compilation discovered by Pixar that is blocking them. Our PF do not compile Maya-Hydra, so this went undiscovered. Due to that, this PF will not be affected by the fix or its absence, but I did compile Maya-hydra locally and it failed without this fix and passed with it. That is why I think we shoudl skip the PF and merge this quickly.

Thanks.

PR #3801, opened 2024-05-31
EMSUSD-1196 material scope as default prim

pierrebai-adsk:
Add the material scope name to the default-prim drop-down when meshes are off and materials are on.

Bug fixes and improvements:
- Prune empty group at the start of the finish step of export to avoid using a prim that will be deleted as a root prim.
- Don't add the default prim name as a parent of the material scope when the material scope *is* the default prim.
- Add a Python helper to retrieve the default material scope name.
- Remove unused default prim data member from UsdMayaPrimWriterArgs.
- Refactor the parsing of encoded arguments to have the same behavior for import and export.
- We now allow for empty argument values.

Change how no-default-prim is expressed
- At the command level, passing "None" as the default prim flag now means no default prim.
- OTOH, not passing the default prim flag or an empty string means to use the first root prim as the default prim.
- When exporting, if the default prim is "None", then no default prim metadata is authored.
- Make sure to also handle "None" for the name of the material scope.

Add unit tests.


lib/mayaUsd/commands/Readme.md
```diff
@@ -32,7 +32,6 @@ Each base command class is documented in the following sections.
 | `-apiSchema`                  | `-api`     | string (multi) | none                              | Imports the given API schemas' attributes as Maya custom attributes. This only recognizes API schemas that have been applied to prims on the stage. The attributes will properly round-trip if you re-export back to USD. |
 | `-chaser`                     | `-chr`     | string(multi)  | none                              | Specify the import chasers to execute as part of the import. See "Import Chasers" below. |
 | `-chaserArgs`                 | `-cha`     | string[3] multi| none                              | Pass argument names and values to import chasers. Each argument to `-chaserArgs` should be a triple of the form: (`<chaser name>`, `<argument name>`, `<argument value>`). See "Import Chasers" below. |
-| `-defaultPrim`                | `-dp`      | string         | none                              | Set the default prim on export. |
```
pierrebai-adsk, line 35: The doc was in the import section instead of the export section.

lib/mayaUsd/fileio/primWriterArgs.h
```diff
@@ -41,8 +41,7 @@ class UsdMayaPrimWriterArgs
     UsdMayaPrimWriterArgs(
         const MDagPath&     dagPath,
         const bool          exportRefsAsInstanceable,
-        const TfToken::Set& excludeExportTypes,
-        const std::string&  defaultPrim);
```
pierrebai-adsk, line 45: The default prim was not used anywhere so I removed it.

AramAzhari-adsk: I also checked and the _defaultPrim is private and isn't being used from anywhere. Good catch!

AramAzhari-adsk (approved):

PR #3802, opened 2024-06-03
Fix tests that are modifying source folder.

seando-adsk:


pierrebai-adsk (approved):
Great! I encountered that a few times but never investigated.

PR #3806, opened 2024-06-06
Create SECURITY.md

neilh-adsk:
Based on security policy

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,27 @@
+# Security
+
+### Reporting Security Issues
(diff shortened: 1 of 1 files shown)


seando-adsk (approved):
Should there be a link from our [README.md](https://github.com/Autodesk/maya-usd/blob/dev/README.md)?

And should be in the doc folder?

neilh-adsk:
@seando-adsk Apparently it just needs to be in the root folder, its a standard file.  I verified this against other repos that were required to update as well

seando-adsk:
@neilh-adsk Ah yes I didn't know about this file. https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository
It does mention there that to help users find it you can link to it from your readme.

PR #3807, opened 2024-06-06
Fix display layer loss when merging/duplicating to USD

jufrantz:
This PR addresses an issue where displayLayer memberships were not correctly transferred to USD when merging or duplicating to USD non-root prims. 

`ReplicateExtrasFromUSD` was attempting to reapply memberships based on prim paths from the temporary layer of `pushExport`. Now, paths are correctly prefixed after `pushExport` by `mergeToUsd` and `duplicate`. 

Also updated unit-tests to verify non-root prims as well as root prims.


pierrebai-adsk (approved):

jufrantz:
Hi @seando-adsk,
I noticed that the build_preflight failed. From the log, it appears that the build job failed to start.

jufrantz:
No problem, thank you!

PR #3813, opened 2024-06-10
EMSUSD-1358 fix blessing failure

pierrebai-adsk:
Force the viewport to update by moving the camera.


test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateUsdCamera.py
```diff
@@ -107,6 +107,9 @@ def testStaticUsdCamera(self):
         clippingAttr = usdCamera.GetAttribute('clippingRange')
         clippingRange = clippingAttr.Get()
         clippingAttr.Set(Gf.Vec2f(10, 5000))
+        # Translate the camera to force the view to update
```
AramAzhari-adsk, line 110: This seems a bit hacked. But if it's unblocking us for the time being I think it should be fine.

AramAzhari-adsk (approved):

Issue #3765, opened 2024-05-15
Merge Maya Edits to USD, when Skeletons are involved, ends up in an error


daniele-niero:
## Issue Description

Maya is able to read and show (USD) Skeletons when loaded through `Create->Universal Scene Description (USD)->Stage From File`, but it doesn't show individual joints and therefore I can't change them.  
However it is possible to right click on the joint root, in the Outliner and choose `Edit As Maya Data` and the skeleton structure is now converted into Maya's joints.  
Now that I have Maya joints I can modify them and right clicking again on the joint root, now a Maya joint, I can see two extra options: `Merge Maya Edits to USD` and `Discard Maya Edits`  
Discarding edits works just fine. Maya joints disappear and the USD structure come back. Good.  
Unfortunately `Merge Maya Edits to USD` ends always with an error

## Steps to reproduce

1. Save this usd file anywhere you want: 
```
#usda 1.0
(
    defaultPrim = "SkeletonGroup"
    metersPerUnit = 0.01
    upAxis = "Z"
)

def SkelRoot "SkeletonGroup" (
    kind = "assembly"
)
{
    def Skeleton "joint1" (
        prepend apiSchemas = ["SkelBindingAPI"]
        customData = {
            dictionary Maya = {
                bool generated = 1
            }
        }
    )
    {
        uniform matrix4d[] bindTransforms = [( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ), ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ), ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) )]
        uniform token[] joints = ["joint1", "joint1/joint2", "joint1/joint2/joint3"]
        uniform matrix4d[] restTransforms = [( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ), ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ), ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) )]
        rel skel:animationSource = </SkeletonGroup/joint1/Animation>

        def SkelAnimation "Animation"
        {
[...]

PR #3820, opened 2024-06-12
EMSUSD-1359 Export of Maya skeleton as USD comes in as a stage in the wrong place.

### LINKED ISSUE #2857: [MAYA-127818] Incorrect rest transform for joints that are not skinned (when exporting scene to USD)
**Describe the bug**
Hello!
I stumbled across a bug regarding rest transform when a joint is not skinned. It happens when exporting the scene from maya to usd file. Since their bind pose is identity, the resulting rest transform (skel:resttransform usd attribute) is the inverse of their parent joint bind transform. I would expect the rest transform to be the local deformation, kind of like how non-skinned bones keep their local transform when we put the asset in "Go to Bind Pose".
The following example will make this issue clearer.

**Steps to reproduce**

1. Extract the zip file
2. Open the scene "cylinder-new-bones.mb" in maya. Notice that "new_bone_long" and "new_bone_short" don't have bind pose.
image

3. Export all to usd (example result in cylinder-new-bones.usda)
4. Open the asset in USD view, you can see that the bones are not properly displayed.
image
The bone "new_bone_short" is invisible. You can also see that the rest transform of "new_bone_long" is the inverse of the bind transform of its parent ("cylinder_joint3"):
image
And, the rest transform of "new_bone_short" is identity (because its parent, "new_bone_long", as an identity bind transform).
image

5. Clear your maya scene.
6. Import the usd file you created previously.
7. Notice that the bones are not displayed like they were on import.
image

**Expected behavior**
I would expect the rest transform to be the local transform of the joints, as seen in "cylinder-new-bones.mb". This way, the scene obtained by importing "cylinder-new-bone.usda" would be the same as "cylinder-new-bones.mb". The expected usda file is then "cylinder-new-bones-expected.usda".

This second point is debatable: I would also expect the bones to be properly displayed in usdview, but that would require defining a value other than identity for the bind pose of a non-skinned joint. I am mentioning it in case you agree and want to change it, but I don't necessarily consider it a bug.

**Attachments**
Here is the zip:
[cylinder-files.zip](https://github.com/Autodesk/maya-usd/files/10551152/cylinder-files.zip)
Only "cylinder-new-bones.*" are used in the steps to reproduce, but I linked the other files in case they are usefull for the investigation. What I did to obtain "cylinder-new-bones.usda" is: "cylinder.mb" -> "cylinder.usda" -> import and edit to get "cylinder-new-bones.mb" -> "cylinder-new-bones.usda".

**Specs (if applicable):**
 - Windows 10
 - Maya 2022.4
 - Maya USD 0.20.0

Thank you in advance!

barbalt:
We were exporting the rest transform as the inverse of the bind transform. 
This was causing issues for bones that weren't bound to any skin cluster.
With this pr, we will use the inverse of the joint's transform as the rest transform.
https://openusd.org/dev/api/_usd_skel__a_p_i__intro.html#UsdSkel_API_WritingSkels

This PR fixes #2857 and #3765

### DIFF
--- a/lib/usd/translators/jointWriter.cpp
+++ b/lib/usd/translators/jointWriter.cpp
@@ -91,21 +91,6 @@ PxrUsdTranslators_JointWriter::PxrUsdTranslators_JointWriter(
     _usdPrim = _skel.GetPrim();
 }
 
-/// Whether the transform plugs on a transform node are animated.
-static bool _IsTransformNodeAnimated(const MDagPath& dagPath)
-{
-    MFnDependencyNode node(dagPath.node());
-    return UsdMayaUtil::isPlugAnimated(node.findPlug("translateX"))
-        || UsdMayaUtil::isPlugAnimated(node.findPlug("translateY"))
-        || UsdMayaUtil::isPlugAnimated(node.findPlug("translateZ"))
-        || UsdMayaUtil::isPlugAnimated(node.findPlug("rotateX"))
-        || UsdMayaUtil::isPlugAnimated(node.findPlug("rotateY"))
-        || UsdMayaUtil::isPlugAnimated(node.findPlug("rotateZ"))
-        || UsdMayaUtil::isPlugAnimated(node.findPlug("scaleX"))
-        || UsdMayaUtil::isPlugAnimated(node.findPlug("scaleY"))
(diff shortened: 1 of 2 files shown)


pierrebai-adsk (approved):
I"m not expert enough in skel to judge, everything looks good to my untrained eyes!

PR #3827, opened 2024-06-17
EMSUSD-1351 - MayaUsd: Code cleanup - never use reserved identifiers

seando-adsk:
#### EMSUSD-1351 - MayaUsd: Code cleanup
* According to our coding standard we should not use reserved identifiers, like underscore followed immediately by an uppercase letter or double underscore.


AramAzhari-adsk (approved):

seando-adsk:
Preflight failures were one random test failure (testUsdExportPackage) and a network glitch on the VM. Since my changes were very minor, I'll set to merge.

PR #3833, opened 2024-06-19
LOOKDEVX-3065 - Fixing renamed parameters

JGamache-autodesk:
Issues when using ND_mix_surfaceshader as the root surface node. The MaterialX shadergen will rename the `mix` parameter to `mix1`.

Need to fix in two places:
1- Where we build the parameter map, as these parameters will most often appear on the surface node itself. Also fixes MaterialX native in LookdevX.
2- Where MayaUSD updates the parameter. It used to try to force back the name, but this causes shader compilation errors. It now remembers them and corrects at the place where the parameter name is built.

Added one unit test that would render green without the fix and yellow if things are done right.


vlasovi (approved):
Looks good

JGamache-autodesk:
@seando-adsk Ready for merge. Preflight failed on a random `testUsdExportPackage` failure.

PR #3835, opened 2024-06-25
EMSUSD-1379 write Autodesk metadata to the session layer

pierrebai-adsk:
- Add an Autodesk USD token.
- Make the scene item automatically route to the session layer for the Autodesk group.
- Add unit tests.


seando-adsk:
Added label "do-not-merge-yet" until we have our internal meeting with others teams to discuss this change.

pierrebai-adsk:
I'll need to change the metadata name. I also realized I need to fix the clear-group-metadata command. I'll close this PR and re-open a new one once everything is done.

PR #3842, opened 2024-07-04
MAYA-131348 - Adapt to Maya Ufe world node change

seando-adsk:
MAYA-131348 - Nothing prevents users from naming nodes "world", which breaks USD workflow

* Incorrect to create Ufe::Path for Maya (hidden) world node using Ufe::PathString::path().
* Let DuplicateCommand accept empty destination path which means Maya (hidden) world node.
* Let CreateStageWithNewLayer command accept null input parent which means Maya world node.
* Use ufe.PathString.string() instead of str()

### DIFF
--- a/lib/mayaUsd/commands/PullPushCommands.cpp
+++ b/lib/mayaUsd/commands/PullPushCommands.cpp
@@ -98,22 +98,42 @@ MSyntax createSyntaxWithUfeArgs(int paramCount)
 bool isPrimPath(const Ufe::Path& path) { return ufePathToPrim(path).IsValid(); }
 
 // Parse the indexed argument as text.
-MStatus parseTextArg(const MArgParser& argParser, int index, MString& outputText)
+MStatus parseTextArg(
+    const MArgParser& argParser,
+    int               index,
+    MString&          outputText,
+    const bool        allowEmpty = false)
 {
-    argParser.getCommandArgument(index, outputText);
-    if (outputText.length() <= 0)
+    MStatus st;
+    st = argParser.getCommandArgument(index, outputText);
+    CHECK_MSTATUS_AND_RETURN_IT(st);
+
+    // Note: when requested, we allow the outputText string to be empty
+    //       (input string to command was empty string). In that case
+    //       it will mean the Maya (hidden) world node.
+    if (!allowEmpty && outputText.length() <= 0)
         return MS::kNotFound;
 
     return MS::kSuccess;
 }
 
 // Parse the indexed argument as a UFE path.
-MStatus parseUfePathArg(const MArgParser& argParser, int index, Ufe::Path& outputPath)
+MStatus parseUfePathArg(
+    const MArgParser& argParser,
+    int               index,
+    Ufe::Path&        outputPath,
+    const bool        allowEmpty = false)
 {
     MString text;
-    MStatus status = parseTextArg(argParser, index, text);
-    if (MS::kSuccess != status)
-        return status;
+    MStatus status = parseTextArg(argParser, index, text, allowEmpty);
+    CHECK_MSTATUS_AND_RETURN_IT(status);
+
+    // Note: when requested we allow the Ufe outputPath to be empty
+    //       which will mean the Maya (hidden) world node.
+    if (allowEmpty && (text.length() == 0)) {
+        outputPath = Ufe::Path();
+        return MS::kSuccess;
+    }
 
     return parseArgAsUfePath(text, outputPath);
 }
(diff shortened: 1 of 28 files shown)


lib/mayaUsd/commands/PullPushCommands.cpp
```diff
@@ -416,14 +426,13 @@ MStatus DuplicateCommand::doIt(const MArgList& argList)
...
-    if (status != MS::kSuccess)
-        return status;
+    CHECK_MSTATUS_AND_RETURN_IT(status);
 
     status = parseUfePathArg(argParser, 0, _srcPath);
     if (status != MS::kSuccess)
         return reportError(status);
 
-    status = parseUfePathArg(argParser, 1, _dstPath);
+    status = parseUfePathArg(argParser, 1, _dstPath, true /*allowEmpty*/);
```
seando-adsk, line 435: I set new `allowEmpty` bool here to be able to accept as command input "" which means Maya "world" node.

lib/mayaUsd/commands/PullPushCommands.cpp
```diff
@@ -455,12 +463,19 @@ MStatus DuplicateCommand::doIt(const MArgList& argList)
...
+                childPath = _dstPath.empty() ? ps : _dstPath + ps;
+            } else {
                 // USD duplicate to Maya
-                : _dstPath + _srcPath.back();
+                if (_dstPath.empty()) {
+                    childPath = Ufe::PathString::path("|" + _srcPath.back().string());
+                } else {
+                    childPath = _dstPath + _srcPath.back();
+                }
+            }
```
seando-adsk, line 478: Had to rework logic to handle when _dstPath is empty.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -554,10 +554,7 @@ using UsdPathToDagPathMapPtr = std::shared_ptr<UsdPathToDagPathMap>;
 using PushCustomizeSrc
     = std::tuple<SdfPath, UsdStageRefPtr, SdfLayerRefPtr, UsdPathToDagPathMapPtr>;
 
-PushCustomizeSrc pushExport(
-    const Ufe::Path&                 ufePulledPath,
-    const MObject&                   mayaObject,
-    const UsdMayaPrimUpdaterContext& context)
+PushCustomizeSrc pushExport(const MObject& mayaObject, const UsdMayaPrimUpdaterContext& context)
```
seando-adsk, line 557: Removed unused input var `ufePullPath`

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -1560,8 +1557,10 @@ bool PrimUpdaterManager::duplicate(
     const Ufe::Path&    dstPath,
     const VtDictionary& userArgs)
 {
-    MayaUsdProxyShapeBase* srcProxyShape = MayaUsd::ufe::getProxyShape(srcPath);
-    MayaUsdProxyShapeBase* dstProxyShape = MayaUsd::ufe::getProxyShape(dstPath);
+    MayaUsdProxyShapeBase* srcProxyShape
+        = srcPath.empty() ? nullptr : MayaUsd::ufe::getProxyShape(srcPath);
+    MayaUsdProxyShapeBase* dstProxyShape
+        = dstPath.empty() ? nullptr : MayaUsd::ufe::getProxyShape(dstPath);
```
seando-adsk, line 1563: Either of these is now allowed to be null.

lib/mayaUsd/ufe/MayaUsdContextOps.cpp
```diff
@@ -683,9 +683,10 @@ Ufe::UndoableCommand::Ptr MayaUsdContextOps::doOpCmd(const ItemPath& itemPath)
     } else if (itemPath[0] == kEditAsMayaOptionsItem) {
         executeEditAsMayaOptions(path());
     } else if (itemPath[0] == kDuplicateAsMayaItem) {
+        // Note: empty string for target means Maya (hidden) world node.
         MString script;
         script.format(
-            "^1s \"^2s\" \"|world\"",
+            "^1s \"^2s\" \"\"",
```
seando-adsk, line 689: Replace "world" with empty string which is now handled in command.

seando-adsk (commented):
The code wasn't "incorrect" before its just that we were relying on knowledge of the Maya `hidden` world node (which is the root). Removed all that world knowledge except in very special cases. These changes are required now before I make a fix in Maya which will allow users to create their own top-level "world" node and use that with Ufe. Currently they cannot because of a bug in Maya.

pierrebai-adsk (approved):
Great!

Only downside is passing an empty string is not super-obvious when reading code, but nto much we can do about it.

seando-adsk:
Yeah I added a few comments in those places to explain the empty string.

PR #3851, opened 2024-07-16
EMSUSD-1160 improve Maya pivot support

pierrebai-adsk:
Multiple tools require the use of Maya-style pivots. The difference between Maya-style and USD-style is that Maya has two separate pivots for rotation and scaling while USD has a single pivot for both. In addition, Maya has pivot-compensation attributes.

When there is already a USD-style pivot authored, we need to convert it to the Maya-style pivots when these tools try to modify the object. We detect that such a conversion is needed and adjust the values when they are read and do the conversion the first time a tools tries to write the values.

Fix center pivot test to take into account the new behaviour. Add new unit test for pivot conversion.

### DIFF
--- a/lib/mayaUsd/ufe/trf/UsdTransform3dMayaXformStack.cpp
+++ b/lib/mayaUsd/ufe/trf/UsdTransform3dMayaXformStack.cpp
@@ -441,25 +441,25 @@ void UsdTransform3dMayaXformStack::convertToMayaPivotIfNeeded()
     }
 
     // Adjust possibly existing Maya rotate pivot by the common pivot.
+    // Note: must explicitly qualify the call to rotatePivot because
+    //       the overload on this class hides the other.
     {
         Ufe::Vector3d currentPivotValue = rotatePivot();
-        auto          cmd = rotatePivotCmd(
+        Ufe::Transform3d::rotatePivot(
             currentPivotValue.x() + commonPivotValue[0],
             currentPivotValue.y() + commonPivotValue[1],
             currentPivotValue.z() + commonPivotValue[2]);
-        if (cmd)
-            cmd->execute();
     }
 
     // Adjust possibly existing Maya scale pivot by the common pivot.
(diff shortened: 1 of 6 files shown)


lib/mayaUsd/ufe/trf/UsdTransform3dMayaXformStack.cpp
```diff
@@ -374,12 +374,95 @@ MAYAUSD_VERIFY_CLASS_SETUP(UsdUfe::UsdTransform3dBase, UsdTransform3dMayaXformSt
...
+        } else {
+            commonPivotValue.Set(0, 0, 0);
+        }
+        pivotAttr.Set(GfVec3f(0, 0, 0));
+    }
+
+    // Adjust possibly existing Maya rotate pivot by the common pivot.
+    {
+        Ufe::Vector3d currentPivotValue = rotatePivot();
+        auto          cmd = rotatePivotCmd(
```
samuelliu-adsk, line 446: If this is not an undoable command, you can just use rotePivt() which will automatically execute

lib/mayaUsd/ufe/trf/UsdTransform3dMayaXformStack.cpp
```diff
@@ -374,12 +374,95 @@ MAYAUSD_VERIFY_CLASS_SETUP(UsdUfe::UsdTransform3dBase, UsdTransform3dMayaXformSt
...
+            currentPivotValue.y() + commonPivotValue[1],
+            currentPivotValue.z() + commonPivotValue[2]);
+        if (cmd)
+            cmd->execute();
+    }
+
+    // Adjust possibly existing Maya scale pivot by the common pivot.
+    {
+        Ufe::Vector3d currentPivotValue = scalePivot();
+        auto          cmd = scalePivotCmd(
```
samuelliu-adsk, line 457: Same here with scalePivot()

samuelliu-adsk (approved):

PR #3852, opened 2024-07-16
MAYA-131348 - Nothing prevents users from naming nodes "world"

seando-adsk:
MAYA-131348 - Nothing prevents users from naming nodes "world", which breaks USD workflow
* Add unit test to verify Maya fix.

### DIFF
--- a/test/lib/testMayaUsdCreateStageCommands.py
+++ b/test/lib/testMayaUsdCreateStageCommands.py
@@ -150,7 +150,7 @@ class MayaUsdCreateStageCommandsTestCase(unittest.TestCase):
         # With fix    = '|world|world'
         worldPath = ufe.PathString.path('|world')
         if str(worldPath) == '|world':
-            self.skipTest('Requires Maya for hidden world node.')
+            self.skipTest('Requires Maya fix for hidden world node.')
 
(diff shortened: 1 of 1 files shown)


test/lib/testMayaUsdCreateStageCommands.py
```diff
@@ -143,3 +143,39 @@ def testCreateStageWithCommand(self):
...
         stageUfeSceneItem = ufe.Hierarchy.createItem(stageUfePath)
         self.assertIsNotNone(stageUfeSceneItem)
+
+    def testCreateStageNamedWorld(self):
+        # This test requires a Maya fix for the hidden world node.
+        # Without fix = '|world'
+        # With fix    = '|world|world'
+        worldPath = ufe.PathString.path('|world')
+        if str(worldPath) == '|world':
+            self.skipTest('Requires Maya for hidden world node.')
```
ppt-adsk, line 153: Don't quite understand the message here: should it be
```suggestion
            self.skipTest('Requires Maya fix for hidden world node.')
```

ppt-adsk (dismissed):

ppt-adsk (approved):

PR #3864, opened 2024-07-23
EMSUSD-1463 - MayaUsd: 2026 Maya blessing is broken because of TBB change in Maya

seando-adsk:
EMSUSD-1463 - MayaUsd: 2026 Maya blessing is broken because of TBB change in Maya

* Use PXR_INCLUDE_DIR from pxrConfig.cmake
* Because USD include path is set as system, make Maya include path a system one and put USD before Maya. Note: system paths always come after regular include paths.

### DIFF
--- a/cmake/modules/FindUSD.cmake
+++ b/cmake/modules/FindUSD.cmake
@@ -158,8 +158,7 @@ include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(USD
     REQUIRED_VARS
         PXR_USD_LOCATION
-        USD_INCLUDE_DIR
-        USD_LIBRARY_DIR
+        PXR_INCLUDE_DIRS
         USD_CONFIG_FILE
         USD_VERSION
         ADSK_USD_VERSION
@@ -170,8 +169,7 @@ find_package_handle_standard_args(USD
 
 if (USD_FOUND)
     # This will follow a message "-- Found USD: <path> ..."
-    message(STATUS "   USD include dir: ${USD_INCLUDE_DIR}")
-    message(STATUS "   USD library dir: ${USD_LIBRARY_DIR}")
+    message(STATUS "   USD include dirs: ${PXR_INCLUDE_DIRS}")
     if (USD_GENSCHEMA)
         message(STATUS "   usdGenSchema: ${USD_GENSCHEMA}")
     endif()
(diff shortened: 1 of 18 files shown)


cmake/modules/FindUSD.cmake
```diff
@@ -158,8 +158,7 @@ include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(USD
     REQUIRED_VARS
         PXR_USD_LOCATION
-        USD_INCLUDE_DIR
-        USD_LIBRARY_DIR
+        PXR_INCLUDE_DIRS
```
seando-adsk, line 161: Use this var which comes from pxrConfig.cmake (which we include above) rather than our own USD_INCLUDE_DIR which came from find a header.

lib/mayaUsd/CMakeLists.txt
```diff
@@ -219,10 +219,13 @@ mayaUsd_compile_config(${PROJECT_NAME})
...
-        ${MAYA_INCLUDE_DIRS}
+    SYSTEM PUBLIC
         ${PXR_INCLUDE_DIRS}
+        ${MAYA_INCLUDE_DIRS}
         ${UFE_INCLUDE_DIR}
+)
+target_include_directories(${PROJECT_NAME} 
+    PUBLIC
         ${CMAKE_BINARY_DIR}/include
 )
```
seando-adsk, line 230: USD include paths were already set as system (by pxrConfig.cmake). According to cmake docs the system paths **always** come after regular paths. So need to set Maya as a system path as well and put if after USD so that USD will come first. This will make our build find the TBB from the USD artifact rather than Maya which will be updating to OneTBB.

seando-adsk:
@roopavr-adsk Rakesh I added you as you might need to make similar changes to maya-hydra repo when updating your latest Maya blessing.

JGamache-autodesk (approved):

roopavr-adsk:
> @roopavr-adsk Rakesh I added you as you might need to make similar changes to maya-hydra repo when updating your latest Maya blessing.

Thanks @seando-adsk Is the new change pertaining to Maya/TBB expected in a new maya cut? If so, do you have a date/cutid? I ask so that I can keep an eye out when it happens. We have not seen any failures pertaining to TBB yet.

seando-adsk:
> > @roopavr-adsk Rakesh I added you as you might need to make similar changes to maya-hydra repo when updating your latest Maya blessing.
> 
> Thanks @seando-adsk Is the change pertaining to Maya/TBB expected in a new maya cut? If so, do you have a date/cutid? I ask so that I can keep an eye out when it happens. We have not seen any failures pertaining to TBB yet.

Yes its because of the Maya TBB change. The change happened yesterday.

PR #3867, opened 2024-07-25
EMSUSD-1357 material scope when no default prim given

pierrebai-adsk:
Change when the default prim is selected when the export job has no default prim explicitly given. It is now chosen at the beginning the export. This change allows the material scope to know that it must be placed under that default prim.

Take into account root prim and export roots. When the root prim or export roots are explicitly given, use them as the default prim if no explicit default prim was given.

Fix all unit tests affected by the change to where the material scope is placed. Some tests are fixed by explicitly requested that no default prim be used, some are fixed by changing where the material scope is checked to be. These two variations allow testing the new behaviour without having to write new unit tests.

### DIFF
--- a/lib/mayaUsd/fileio/jobs/jobArgs.cpp
+++ b/lib/mayaUsd/fileio/jobs/jobArgs.cpp
@@ -707,6 +707,7 @@ UsdMayaJobExportArgs::UsdMayaJobExportArgs(
     , dagPaths(dagPaths)
     , fullObjectList(fullList)
     , timeSamples(timeSamples)
+    , exportRoots(extractVector<std::string>(userArgs, UsdMayaJobExportArgsTokens->exportRoots))
     , rootMapFunction(PcpMapFunction::Create(
           _ExportRootsMap(
               userArgs,
--- a/lib/mayaUsd/fileio/jobs/jobArgs.h
+++ b/lib/mayaUsd/fileio/jobs/jobArgs.h
@@ -270,7 +270,7 @@ struct UsdMayaJobExportArgs
     const TfToken::Set includeAPINames;
     const TfToken::Set jobContextNames;
     const TfToken::Set excludeExportTypes;
-    const std::string  defaultPrim;
+    std::string        defaultPrim;
 
     using ChaserArgs = std::map<std::string, std::string>;
     const std::vector<std::string>          chaserNames;
@@ -307,7 +307,8 @@ struct UsdMayaJobExportArgs
     // When using export roots feature we will leverage map function to
     // override the sdfpath generated from source DAG path. Will be empty
     // if export roots is not used.
-    const PcpMapFunction rootMapFunction;
+    const std::vector<std::string> exportRoots;
+    const PcpMapFunction           rootMapFunction;
 
     // Maya type ids to avoid exporting; these are EXACT types, the constructor will also add all
     // inherited types (so if you exclude "constraint", it will also exclude "parentConstraint")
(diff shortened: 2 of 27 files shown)


pierrebai-adsk:
While the PR seems large, most of it are fixes to existing unit tests to account for the new behaviour of material scopes when no default prims were specified. Various test were fixed differently either:

- Specifying an explicit default prim (Usually so that the same default prim gets authored)
- Requesting that no default prim be authored (usually so that no default prim be defined as before)
- Fix the assertions to take into account the new location of the material scope.

These varied changes allow testing the different modes of the new behaviour introduce by this PR.

lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -707,6 +707,7 @@ UsdMayaJobExportArgs::UsdMayaJobExportArgs(
     , dagPaths(dagPaths)
     , fullObjectList(fullList)
     , timeSamples(timeSamples)
+    , exportRoots(extractVector<std::string>(userArgs, UsdMayaJobExportArgsTokens->exportRoots))
     , rootMapFunction(PcpMapFunction::Create(
```
pierrebai-adsk, line 711: Previously the export roots that were given were not kept explicitly but indirectly in the rootMapFunction and were thus not accessible for code to check later on, so they are now also kept around.

lib/mayaUsd/fileio/jobs/jobArgs.h
```diff
@@ -270,7 +270,7 @@ struct UsdMayaJobExportArgs
     const TfToken::Set includeAPINames;
     const TfToken::Set jobContextNames;
     const TfToken::Set excludeExportTypes;
-    const std::string  defaultPrim;
+    std::string        defaultPrim;
```
pierrebai-adsk, line 273: When staring to export, when no default prim were specified a default prim will be set, so the variable must no be const anymore.

lib/mayaUsd/fileio/jobs/writeJob.cpp
```diff
@@ -163,10 +163,74 @@ bool UsdMaya_WriteJob::Write(const std::string& fileName, bool append)
     return true;
 }
 
+static MStringArray GetExportDefaultPrimCandidates(const UsdMayaJobExportArgs& exportArgs)
```
pierrebai-adsk, line 166: This is the core of the change: select a default prim from the available Maya nodes. This reuse the existing code used to show the selection of possible default prims in the export UI. We need to reuse it in the case the export was done by a direct command and no default prim were given and no request for having no default prim was made (by using specifying that the default prim be "None").

lib/mayaUsd/fileio/jobs/writeJob.cpp
```diff
@@ -163,10 +163,74 @@ bool UsdMaya_WriteJob::Write(const std::string& fileName, bool append)
...
+{
+    MStringArray roots;
+
+    // If the user provided a root prim, used it as the default prim.
+    if (!exportArgs.rootPrim.IsEmpty()) {
+        roots.append(exportArgs.rootPrim.GetName().c_str());
```
barbalt, line 172: on the check a little bit below, we check if the string is not empty. We don't do that here, would that be necessary?

barbalt (approved):

lib/mayaUsd/fileio/jobs/writeJob.cpp
pierrebai-adsk: The IsEmpty(0 is equivalent. IsEmpty is on TfToken, empty is on std::string. A non-empty TfToken is a non-empty std::string.

barbalt: Ah, right.. I got confused with the array

barbalt:
👍 Looks good

PR #3869, opened 2024-07-30
EMSUSD-1410 - MayaUsdAPI - enhance with missing functions required for LookdevX

seando-adsk:
EMSUSD-1410 - MayaUsdAPI - enhance with missing functions required for LookdevX
* Update mayaUsdAPI to v1.0.0 From this point on no binary breaking changes are allowed.
* Add missing functionality required to move LookdevX off MayaUsd and UsdUfe.


cmake/mayausdapi_version.info
```diff
@@ -1,3 +1,3 @@
-set(MAYAUSDAPI_MAJOR_VERSION 0)
+set(MAYAUSDAPI_MAJOR_VERSION 1)
```
seando-adsk, line 1: Now set to major version 0 - binary breaking changes are not allowed.

lib/mayaUsdAPI/global.h
```diff
@@ -0,0 +1,72 @@
...
+    You must provide each of the mandatory functions in order for the plugin
+    to function correctly for your runtime.
+*/
+struct MAYAUSD_API_PUBLIC DCCMandatoryFunctions
+{
+    StageAccessorFn     stageAccessorFn = nullptr;
+    StagePathAccessorFn stagePathAccessorFn = nullptr;
+    UfePathToPrimFn     ufePathToPrimFn = nullptr;
+    TimeAccessorFn      timeAccessorFn = nullptr;
+};
```
seando-adsk, line 47: These are the mandatory dcc funcs required when registering UsdUfe. If a new mandatory function is added in the future, then we'll have to make a "2" version of this struct.

lib/mayaUsdAPI/undo.h
```diff
@@ -0,0 +1,75 @@
...
+
+    UsdUndoableItem();
+    ~UsdUndoableItem();
+
+    void undo();
+    void redo();
+
+private:
+    //! Private implementation member
+    std::unique_ptr<UsdUndoableItemImpl> _imp;
```
seando-adsk, line 53: There are a few of these added there. Using an impl to hide the real UsdUfe/MayaUsd class in the .cpp file and using thru this pointer to avoid any binary breaking.

lib/mayaUsdAPI/undo.cpp
```diff
@@ -0,0 +1,67 @@
...
+};
+
+UsdUndoableItem::UsdUndoableItem()
+    : _imp(new UsdUndoableItemImpl())
+{
+}
+
+// When using a pimpl you need to define the destructor here in the
+// .cpp so the compiler has access to the impl.
+UsdUndoableItem::~UsdUndoableItem() = default;
```
seando-adsk, line 52: This was previously discovered during the initial MayaUsdAPI implementation.

lib/mayaUsdAPI/render.h
```diff
@@ -0,0 +1,123 @@
...
+//
+#ifndef MAYAUSDAPI_RENDER_H
+#define MAYAUSDAPI_RENDER_H
+
+#include <mayaUsdAPI/api.h>
+
+#include <MaterialXCore/Document.h>
+#include <MaterialXCore/Element.h>
+#include <MaterialXCore/Node.h>
+#include <MaterialXFormat/File.h>
```
seando-adsk, line 24: Do you think its okay to include MaterialX headers here (in the .h) or should I try and forward declare everything? We don't do that for pxr includes, so I figured it was okay here. MaterialX comes from the USD artifact. Anytime we update USD there could be binary API changes coming from that update.

seando-adsk (commented):
Bascially added everything need to completely remove UsdUfe and MayaUsd from LookdevX. In an upcoming LookdevX PR you'll see it using only MayaUsdAPI.

cmake/mayausdapi_version.info
ppt-adsk: This means that current binaries that depend on mayaUsdAPI v0 must be recompiled.

lib/mayaUsdAPI/global.h
ppt-adsk: Or a version e.g. _1_1 perhaps, for mayaUsdAPI version 1.1

ppt-adsk (dismissed):
LGTM.

lib/mayaUsdAPI/render.cpp
```diff
@@ -0,0 +1,160 @@
...
+const MaterialX::StringMap& OgsFragment::getPathInputMap() const
+{
+    return _imp->_ogsFragment.getPathInputMap();
+}
+
+std::string OgsFragment::getMatrix4Name(const std::string& matrix3Name)
+{
+    return _imp->_ogsFragment.getMatrix4Name(matrix3Name);
+}
+
```
JGamache-autodesk, line 112: I added a new function very recently, you might want to also export `MaterialXMaya::OgsFragment::isTransparentSurface()`

JGamache-autodesk (dismissed):

cmake/mayausdapi_version.info
seando-adsk: Yes true. Is that going to be a problem? The only two that I know of that use the API are Hydra and LookdevX. I will update the MayaUsd cut used by LookdevX in my LDX PR (after this one is merged). Hydra will get the latest one the next time they update and we'll ship new 2025 binaries the next release.

lib/mayaUsdAPI/render.cpp
seando-adsk: Thanks. I've synced with latest LookdevX and added the missing function to this PR.

JGamache-autodesk (approved):

ppt-adsk (dismissed):

JGamache-autodesk (dismissed):

ppt-adsk (dismissed):

ppt-adsk (approved):

JGamache-autodesk (approved):

lib/mayaUsdAPI/mayaUsdAPI.h.src
```diff
@@ -47,4 +47,11 @@ using namespace MAYAUSDAPI_VERSIONED_NS;
 #define MAYAUSDAPI_NS_DEF MAYAUSDAPI_VERSIONED_NS
 #endif
 
+// Any MayaUsd cmake defines that are used in the MayaUsdAPI project
+// must be added here so that clients using the API will have the
+// needed defines.
+#ifndef WANT_MATERIALX_BUILD
+#cmakedefine WANT_MATERIALX_BUILD
+#endif
```
seando-adsk, line 55: This will give the following when it is or is not defined. I needed to add the ifndef as clients like MayaHydra are setting this variable in their own cmake files.
```
#ifndef WANT_MATERIALX_BUILD
#define WANT_MATERIALX_BUILD
#endif

#ifndef WANT_MATERIALX_BUILD
/* #undef WANT_MATERIALX_BUILD */
#endif
```

lib/mayaUsdAPI/render.h
```diff
@@ -18,13 +18,16 @@
 
 #include <mayaUsdAPI/api.h>
 
+#include <string>
+
+#ifdef WANT_MATERIALX_BUILD
```
seando-adsk, line 23: Because this variable is used in an API header I needed to add it to the mayaUsdAPI.h file, so clients like LookdevX will get the correct build (since they don't set this var in their cmake files).

lib/mayaUsdAPI/ufeHandlers.h
```diff
@@ -18,16 +18,21 @@
 
 #include <mayaUsdAPI/api.h>
 
-#include <ufe/connectionHandler.h>
 #include <ufe/hierarchyHandler.h>
-#include <ufe/nodeDefHandler.h>
 #include <ufe/sceneItemOpsHandler.h>
+
+#ifdef UFE_V4_FEATURES_AVAILABLE
```
seando-adsk, line 24: Using ufe defines is fine for any client since they are set correctly in ufe.h

JGamache-autodesk (approved):

seando-adsk:
Sorry for all the different commits. It took me a while to get a green preflight build.

JGamache-autodesk (approved):

ppt-adsk (dismissed):

PR #3870, opened 2024-07-31
EMSUSD-986 - Fix the issue that edit as maya on an instanced prim will move it away

samuelliu-adsk:


lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -602,6 +602,7 @@ void UsdMayaTranslatorXformable::Read(
                 "Unable to successfully decompose matrix at USD prim <%s>",
                 xformSchema.GetPath().GetText());
         }
+        resetsXformStack = true;
```
pierrebai-adsk, line 605: Could you put a comment on why we need this? It will help understand what is going on in the future.

PR #3871, opened 2024-07-31
LOOKDEVX-3214 - Improve UI name generation

JGamache-autodesk:
- Prettify UI names provided by USD that are identical to the family name (for MaterialX shader nodes)
- Do more corrections in the prettifyName function based on observations on the MaterialX library
- Store known USD metadata using Sdf keys instead of Sdr metadata

### DIFF
--- a/lib/usdUfe/ufe/StagesSubject.cpp
+++ b/lib/usdUfe/ufe/StagesSubject.cpp
@@ -15,6 +15,7 @@
 //
 #include "StagesSubject.h"
 
+#include <usdUfe/base/tokens.h>
 #include <usdUfe/ufe/Global.h>
 #include <usdUfe/ufe/UfeNotifGuard.h>
 #include <usdUfe/ufe/UfeVersionCompat.h>
@@ -338,6 +339,18 @@ void processAttributeChanges(
                         metadataKeys.insert(newMetadataKey);
                     }
                 }
+            } else if (infoChanged.first == SdfFieldKeys->AllowedTokens) {
+                sendMetadataChanged = true;
+                metadataKeys.insert(UsdUfe::MetadataTokens->UIEnumLabels);
+            } else if (infoChanged.first == SdfFieldKeys->Documentation) {
+                sendMetadataChanged = true;
+                metadataKeys.insert(UsdUfe::MetadataTokens->UIDoc);
+            } else if (infoChanged.first == SdfFieldKeys->DisplayGroup) {
+                sendMetadataChanged = true;
+                metadataKeys.insert(UsdUfe::MetadataTokens->UIFolder);
+            } else if (infoChanged.first == SdfFieldKeys->DisplayName) {
+                sendMetadataChanged = true;
+                metadataKeys.insert(UsdUfe::MetadataTokens->UIName);
             }
         }
     }
--- a/lib/usdUfe/ufe/UsdAttributeHolder.cpp
+++ b/lib/usdUfe/ufe/UsdAttributeHolder.cpp
@@ -50,8 +50,20 @@ bool setUsdNativeMetadata(
         // not recommended.
         return false;
     }
+
+    // It is possible we are dealing with a legacy scene where the data was stored in SdrMetadata.
+    const auto clearKnownSdrMetadata = [](auto const& a, auto const& k) {
+        if (UsdShadeInput::IsInput(a)) {
+            UsdShadeInput(a).ClearSdrMetadataByKey(TfToken(k));
+        }
+        if (UsdShadeOutput::IsOutput(a)) {
+            UsdShadeOutput(a).ClearSdrMetadataByKey(TfToken(k));
+        }
+    };
+
     if (key == UsdUfe::MetadataTokens->UIDoc) {
         attr.SetDocumentation(value.get<std::string>());
+        clearKnownSdrMetadata(attr, key);
         return true;
     } else if (key == UsdUfe::MetadataTokens->UIEnumLabels) {
         const auto   enumStrings = UsdUfe::splitString(value.get<std::string>(), ",");
@@ -61,6 +73,7 @@ bool setUsdNativeMetadata(
             allowedTokens.push_back(TfToken(TfStringTrim(tokenString, " ")));
         }
         attr.SetMetadata(SdfFieldKeys->AllowedTokens, allowedTokens);
+        clearKnownSdrMetadata(attr, key);
         return true;
     } else if (key == UsdUfe::MetadataTokens->UIFolder) {
         // Translate '|' to ':'.
@@ -72,9 +85,11 @@ bool setUsdNativeMetadata(
         std::string group = value.get<std::string>();
         std::replace(group.begin(), group.end(), '|', ':');
         attr.SetDisplayGroup(group);
+        clearKnownSdrMetadata(attr, key);
         return true;
     } else if (key == UsdUfe::MetadataTokens->UIName) {
         attr.SetDisplayName(value.get<std::string>());
+        clearKnownSdrMetadata(attr, key);
         return true;
     }
 
@@ -139,7 +154,10 @@ Ufe::Value getUsdNativeMetadata(const PXR_NS::UsdAttribute& attr, const std::str
     return retVal;
 }
 
-// Returns true if handled, and set cleared on success
+// When clearing metadata we have two boolean states to return:
+//   1- Did we handle the key: returned by value
+//   2- If we handled the key, did we actually clear the value. Returned by reference in the
+//   "cleared" variable.
 bool clearUsdNativeMetadata(const PXR_NS::UsdAttribute& attr, const std::string& key, bool& cleared)
 {
     PXR_NAMESPACE_USING_DIRECTIVE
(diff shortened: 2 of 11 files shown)


lib/usdUfe/ufe/StagesSubject.cpp
```diff
@@ -338,6 +339,18 @@ void processAttributeChanges(
                         metadataKeys.insert(newMetadataKey);
                     }
                 }
+            } else if (infoChanged.first == SdfFieldKeys->AllowedTokens) {
```
JGamache-autodesk, line 342: USD has SDF/SDR metadata for 4 of the regular UI metadata. The USD MaterialX reader does copy those [metadata as SDF attributes](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdMtlx/reader.cpp#L550) instead of Sdr metadata, so we need to add code in multiple locations to handle those SDF metadata and in the case here, send the right UFE notification.

lib/usdUfe/utils/Utils.cpp
```diff
@@ -42,13 +42,16 @@ std::string prettifyName(const std::string& name)
             if ((i > 0 && (i < (nbChars - 1))
                  && (!std::isupper(name[i + 1]) && !std::isdigit(name[i + 1])))
                 || std::islower(name[i - 1])) {
-                prettyName += ' ';
+                if (prettyName.size() > 0 && prettyName.back() != ' ') {
```
JGamache-autodesk, line 45: Fixes two observed cases where a double space was added because an explicit space character "_" was followed by a lower to upper case transition.

lib/usdUfe/ufe/UsdAttributeHolder.cpp
```diff
@@ -59,17 +61,58 @@ bool setUsdAttrMetadata(
...
+
     PXR_NS::TfToken tok(key);
     if (PXR_NS::UsdShadeNodeGraph(attr.GetPrim())) {
         if (PXR_NS::UsdShadeInput::IsInput(attr)) {
             PXR_NS::UsdShadeInput input(attr);
-            input.SetSdrMetadataByKey(tok, value.get<std::string>());
-            return true;
+            if (knownUSDMetadata.count(key)) {
+                // Cleanup known USD metadata previously set as SdrMetadata
+                input.ClearSdrMetadataByKey(tok);
```
pierrebai-adsk, line 77: I guess this is for the case where we put the metadata in older version of MayaUSD and we want to clean it up now.

lib/usdUfe/ufe/UsdAttributeHolder.cpp
```diff
@@ -59,17 +61,58 @@ bool setUsdAttrMetadata(
...
+            }
+        }
+    }
+
+    if (key == UsdUfe::MetadataTokens->UIDoc) {
+        attr.SetDocumentation(value.get<std::string>());
+        return true;
+    } else if (key == UsdUfe::MetadataTokens->UIEnumLabels) {
+        auto       sdfAttr = TfStatic_cast<SdfAttributeSpecHandle>(attr.GetPropertyStack().front());
+        const auto enumStrings = UsdUfe::splitString(value.get<std::string>(), ", ");
```
pierrebai-adsk, line 99: Is the text format a specification where we know it is a comma-separated string or we just happen to be in-the-know?

lib/usdUfe/ufe/UsdAttributeHolder.cpp
```diff
@@ -59,17 +61,58 @@ bool setUsdAttrMetadata(
...
+        VtTokenArray allowedTokens;
+        allowedTokens.reserve(enumStrings.size());
+        for (const auto& tokenString : enumStrings) {
+            allowedTokens.push_back(TfToken(tokenString));
         }
+        sdfAttr->SetAllowedTokens(allowedTokens);
+        return true;
+    } else if (key == UsdUfe::MetadataTokens->UIFolder) {
+        // Translate '|' to ':'.
```
pierrebai-adsk, line 108: Similar question I guess. I suppose we have no choice but "know" what the format is. Is there docs about this?

lib/usdUfe/ufe/UsdAttributeHolder.cpp
JGamache-autodesk: Yes, we want it as close to USD specs as possible, so we will move it from SdrMetadata to Sdf if there was old metadata from a previously saved file.

pierrebai-adsk (approved):

lib/usdUfe/ufe/UsdAttributeHolder.cpp
JGamache-autodesk: That is in the MaterialX specification. And since the second argument of UsdUfe::splitString has the "any of" semantics, it will work if there are no spaces in the list of tokens. Though I am unsure if it will behave correctly if there are spaces in the labels themselves: "Blinn, Oren Nayar, Lambert" should unpack as 3 tokens and the space in the middle of "Oren Nayar" needs to be preserved. I will add tests for that.

lib/usdUfe/ufe/UsdAttributeHolder.cpp
```diff
@@ -59,17 +61,58 @@ bool setUsdAttrMetadata(
...
+                // Cleanup known USD metadata previously set as SdrMetadata
+                output.ClearSdrMetadataByKey(tok);
+            } else {
+                output.SetSdrMetadataByKey(tok, value.get<std::string>());
+                return true;
+            }
+        }
+    }
+
+    if (key == UsdUfe::MetadataTokens->UIDoc) {
```
frohnej-adsk, line 94: [Readability] Would it make sense to factor out this code into a separate function? Something like:
```C++
namespace {
bool setUsdMetadata(UsdAttribute attribute, std::string key, Ufe::Value value)
{
    if (key == UsdUfe::MetadataTokens->UIDoc) {
        attr.SetDocumentation(value.get<std::string>());
        return true;
    } else if (key == UsdUfe::MetadataTokens->UIEnumLabels) {
    // ...
}
}
```

In the code above, we could then do something like this:
```C++
        } else if (PXR_NS::UsdShadeOutput::IsOutput(attr)) {
            if (knownUSDMetadata.count(key)) {
                // Cleanup known USD metadata previously set as SdrMetadata
                output.ClearSdrMetadataByKey(tok);
                // Set the USD metadata
                return setUsdMetadata(attr, key, value);
            }

            output.SetSdrMetadataByKey(tok, value.get<std::string>());
            return true;
        }
```

This way, we could get rid of the else block to reduce indentation by a bit and the clearing of the Sdr metadata and the setting of the USD metadata would be right next to each other, which might improve readability.

frohnej-adsk (dismissed):

lib/usdUfe/ufe/UsdAttributeHolder.cpp
JGamache-autodesk: Done. Also added missing hasMetadata and clearMetadata callbacks as well as a full featured unit test making sure UFE and USD are in sync.

lib/usdUfe/ufe/UsdAttributeHolder.cpp
```diff
@@ -59,19 +201,40 @@ bool setUsdAttrMetadata(
...
             PXR_NS::UsdShadeInput input(attr);
-            input.SetSdrMetadataByKey(tok, value.get<std::string>());
-            return true;
+            if (knownUSDMetadata.count(key)) {
+                // Cleanup known USD metadata previously set as SdrMetadata
+                input.ClearSdrMetadataByKey(tok);
+            } else {
+                input.SetSdrMetadataByKey(tok, value.get<std::string>());
+                return true;
+            }
```
frohnej-adsk, line 221: [Suggestion] We might not necessarily need `knownUSDMetadata` anymore, since `setUsdNativeMetadata()` returns a Boolean telling us if the key was a valid USD metadata key.

So, we could possibly do:
```C++
    if (setUsdNativeMetadata(attr, key, value)) {
        // Cleanup known USD metadata previously set as SdrMetadata
        input.ClearSdrMetadataByKey(tok);
        return true;
    }
    
    input.SetSdrMetadataByKey(tok, value.get<std::string>());
    return true;
```

Not sure if having the map improves performance, but if the set of know USD metadata changes, we'd currently need to adjust both `knownUSDMetadata` and `setUsdNativeMetadata()`.

frohnej-adsk (dismissed):
LGTM! One more suggestion but not mandatory.

lib/usdUfe/ufe/UsdAttributeHolder.cpp
```diff
@@ -38,6 +38,146 @@ namespace {
...
+        retVal = Ufe::Value(attr.GetDisplayName());
+    }
+    return retVal;
+}
+
+// Returns true if handled, and set cleared on success
+bool clearUsdNativeMetadata(const PXR_NS::UsdAttribute& attr, const std::string& key, bool& cleared)
+{
+    PXR_NAMESPACE_USING_DIRECTIVE
+    cleared = false;
```
pierrebai-adsk, line 146: Not sure why there is this cleared variable, it gets set but never used. Might be a remnant of older code? Or where all the "return true"meant to be "return cleared"?

pierrebai-adsk (dismissed):

lib/usdUfe/ufe/UsdAttributeHolder.cpp
JGamache-autodesk: Good idea. I put the cleanup code where it belongs so it is no longer split in two functions.

lib/usdUfe/ufe/UsdAttributeHolder.cpp
JGamache-autodesk: Sadly I have two states to return: handled and cleared. One is done by value, the other one by reference. Could also use a tri-state enum (unhandled, cleared, nothing to clear) if it makes things simpler.

frohnej-adsk (approved):
Nice!

pierrebai-adsk (approved):

PR #3874, opened 2024-08-02
EMSUSD-1467 fix multi-layer parenting

pierrebai-adsk:
The parenting code was not copying the data correctly. Changed to use the similar code as in the duplicate command. The duplicate command could not be used because the parenting command targets an arbitrary destination.

Added a unit test that exercises the case where a prim is in multiple layers and targeting the session layer. The old code would lose data.

Nested edit router failed to route to the correct layer and would result instead in an infinite loop.

Improve edit routing unit test to test deeper routing nesting.


samuelliu-adsk:
Note super familiar with how duplicate works in Edit Routing but the code looks good to me

samuelliu-adsk (approved):

PR #3876, opened 2024-08-05
EMSUSD-1432 duplicate materials without meshes

pierrebai-adsk:
- Keep track of the exported material paths in the write job and write job context.
- No longer disable material UI when not exporting meshes in duplicate mode.
- Add the exported material to the write job in the shading engine.
- Cleanup the result of the export in the prim updater manager.
- When copying to USD, make sure we don't use legacy mode and have no default prim.
- Make the duplicate command and function return the list of duplicated destinations.
- This fixes the problem that the destination of duplicating without meshes cannot be known in advance.
- Add unit test for duplication without meshes but with materials.

### DIFF
--- a/lib/mayaUsd/commands/PullPushCommands.cpp
+++ b/lib/mayaUsd/commands/PullPushCommands.cpp
@@ -452,9 +452,9 @@ MStatus DuplicateCommand::doIt(const MArgList& argList)
         OpUndoItemRecorder undoRecorder(_undoItemList);
 
         auto& manager = PXR_NS::PrimUpdaterManager::getInstance();
-        status = manager.duplicate(_srcPath, _dstPath, userArgs) ? MS::kSuccess : MS::kFailure;
+        auto  dstUfePaths = manager.duplicate(_srcPath, _dstPath, userArgs);
 
-        if (status == MS::kSuccess) {
+        if (dstUfePaths.size() > 0) {
             // Select the duplicate.
             //
             // If the duplicate src is Maya, the duplicate child of the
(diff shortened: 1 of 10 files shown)


lib/mayaUsd/commands/PullPushCommands.cpp
```diff
@@ -452,9 +452,9 @@ MStatus DuplicateCommand::doIt(const MArgList& argList)
         OpUndoItemRecorder undoRecorder(_undoItemList);
 
         auto& manager = PXR_NS::PrimUpdaterManager::getInstance();
-        status = manager.duplicate(_srcPath, _dstPath, userArgs) ? MS::kSuccess : MS::kFailure;
+        auto  dstUfePaths = manager.duplicate(_srcPath, _dstPath, userArgs);
```
pierrebai-adsk, line 456: Duplicate function now returns a list of destination paths. Allows getting rid of much code below.

lib/mayaUsd/fileio/jobs/writeJob.h
```diff
@@ -55,6 +55,10 @@ class UsdMaya_WriteJob
     MAYAUSD_CORE_PUBLIC
     const UsdMayaUtil::MDagPathMap<SdfPath>& GetDagPathToUsdPathMap() const;
 
+    // Retrieve all exported material paths.
+    MAYAUSD_CORE_PUBLIC
+    const std::vector<SdfPath>& GetMaterialPaths() { return mJobCtx.GetMaterialPaths(); }
```
pierrebai-adsk, line 60: We now track the materials that were generated.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -550,18 +550,27 @@ void fillUserArgsFileIfEmpty(VtDictionary& userArgs, const std::string& fileName
...
-    = std::tuple<SdfPath, UsdStageRefPtr, SdfLayerRefPtr, UsdPathToDagPathMapPtr>;
+struct PushExportResult
+{
+
+    SdfPath                              srcRootPath;
+    UsdStageRefPtr                       stage;
+    SdfLayerRefPtr                       layer;
+    std::shared_ptr<UsdPathToDagPathMap> usdToDag;
+    std::vector<SdfPath>                 materialPaths;
+};
```
pierrebai-adsk, line 561: Using a struct instead of a tuple make code much easier to read.

lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -582,38 +591,49 @@ PushCustomizeSrc pushExport(const MObject& mayaObject, const UsdMayaPrimUpdaterC
...
     UsdMaya_WriteJob writeJob(jobArgs);
     if (!writeJob.Write(fileName, false /* append */)) {
-        return pushCustomizeSrc;
+        return result;
     }
     progressBar.advance();
 
-    std::get<SdfPath>(pushCustomizeSrc) = writeJob.MapDagPathToSdfPath(dagPath);
+    result.srcRootPath = writeJob.MapDagPathToSdfPath(dagPath);
+    if (result.srcRootPath.IsEmpty()) {
```
pierrebai-adsk, line 621: When exporting only materials, there might be nothing in the DAG-to-USD map, because the materials are not in the DAG. instead we use the parent scope of the materials, which contains all material when in non-legacy mode. (Duplicate is always in non-legacy mode).

barbalt (approved):

PR #3884, opened 2024-08-07
EMSUSD-1527 support selection in duplicate to USD

pierrebai-adsk:
- Add an option to the copyLayerPrims function to merge the top USD scope instead of creating new ones.
- This allows copying multiple prims with a common root tree without adding new branch every time.
- This is useful when duplicating materials only because each one creates a "mtl" scope to contain itself. With this change, all material will be under a common "mtl" scope instead of having a different scope for each.
- Make use of that mode in the duplicate code. Also, make sure we set the export selected flag so that it only duplicate materials in the selected Maya nodes.
- Modify the duplicate menu handler to parse the whole selection if the clicked object is in the selection.
- Add a unit test.

### DIFF
--- a/plugin/adsk/scripts/mayaUsdMenu.mel
+++ b/plugin/adsk/scripts/mayaUsdMenu.mel
@@ -667,17 +667,49 @@ global proc mayaUsdMenu_duplicateToUSD( string $proxy, string $obj )
     if (!hasPrimUpdater())
         return;
 
+    // Convert proxy to long form.
+    {
+        string $proxyLong[] = `ls -l $proxy`;
+        if (size($proxyLong) == 0)
+            return;
+        $proxy = $proxyLong[0];
+    }
+
+    string $bulkObjects[];
+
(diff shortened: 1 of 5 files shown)


plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -667,17 +667,49 @@ global proc mayaUsdMenu_duplicateToUSD( string $proxy, string $obj )
     if (!hasPrimUpdater())
         return;
 
+    // Convert proxy to long form.
```
AramAzhari-adsk, line 670: What does a long form entail?

AramAzhari-adsk (approved):

plugin/adsk/scripts/mayaUsdMenu.mel
pierrebai-adsk: Retrieves the full absolute name of the Maya object.

PR #3888, opened 2024-08-13
Remove matrix multiply operator overloads that potentially lose

dj-mcg:
precision

The overloads of the methods that take and return a GfVec3f can cause inadvertent loss of precision. This change updates call sites where the result of a call to one of these overloads now needs to be converted to GfVec3f. The advantage of requiring these conversions is that the loss of precision is explicit.


seando-adsk (approved):

seando-adsk:
@dj-mcg We are currently having issues with our preflight system (internal Jenkins problems). That is why this preflight failed. I'll relaunch it once the issue is fixed.

PR #3889, opened 2024-08-13
Update AE interface for Pxr Reference Assembly nodes

dj-mcg:
This provides a way to customize how variantSets are presented in the Attribute Editor for reference assemblies.


seando-adsk:
@samuelliu-adsk Can you please do a quick code review on this change. Its in the Pixar plugin which Dan is responsible for. Once you approve, can you also start the preflight (by assigning to dj-mcg).

samuelliu-adsk (approved):

PR #3890, opened 2024-08-15
Updating callsites to use new NdrSdfTypeIndicator class

dj-mcg:
The class version of NdrSdfTypeIndicator will land in USD 2411. The callsites are version guarded


JGamache-autodesk (approved):

seando-adsk:
Preflight failed because of an internal network glitch. Will re-run.

PR #3899, opened 2024-09-03
EMSUSD-1477 merge job context arguments with user arguments

pierrebai-adsk:
The VRay plugin provides the chaser name by the user arguments instead of returning it in the plug-in job-context callback. (How they inject the chaser name in the arguments is still a mystery.)

The code used to overwrite such user arguments with the job-context arguments. (By calling VtDictionaryOver.) Now we instead merge the dictionaries with code refactored from the code used to merge job context arguments.

Do the same for the import context.

### DIFF
--- a/lib/mayaUsd/fileio/jobs/jobArgs.cpp
+++ b/lib/mayaUsd/fileio/jobs/jobArgs.cpp
@@ -430,6 +430,111 @@ std::set<unsigned int> _FilteredTypeIds(const VtDictionary& userArgs)
     return result;
 }
 
+std::vector<VtValue>
+_MergeVectors(const std::vector<VtValue>& existingValues, const std::vector<VtValue>& newValues)
+{
+    std::vector<VtValue> resultValues = existingValues;
+
+    for (const VtValue& element : newValues) {
+        if (element.IsHolding<std::vector<VtValue>>()) {
+            // vector<vector<string>> is common for chaserArgs and shadingModes
+            auto findElement = [&element](const VtValue& a) {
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -430,6 +430,111 @@ std::set<unsigned int> _FilteredTypeIds(const VtDictionary& userArgs)
     return result;
 }
 
+std::vector<VtValue>
+_MergeVectors(const std::vector<VtValue>& existingValues, const std::vector<VtValue>& newValues)
+{
```
pierrebai-adsk, line 435: These new function are refactored code from the previously existing function named _MergeJobContexts, fiund below.

lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -487,81 +597,55 @@ bool _MergeJobContexts(bool isExport, const VtDictionary& userArgs, VtDictionary
...
+        const bool allowConflicts = false;
+        if (!_MergeDictionaries(allContextArgs, argInitialSource, dict, sourceName, allowConflicts))
+            canMergeContexts = false;
     }
     return canMergeContexts;
 }
 
+VtDictionary
+_MergeAllArguments(bool isExport, const VtDictionary& userArgs, const VtDictionary& defaultsArgs)
+{
```
pierrebai-adsk, line 617: This function is refactored from the import/export job args constructors found below, but now we properly merge user arg and job option args instead of overwriting users args entirely.

barbalt (approved):

PR #3900, opened 2024-09-04
EMSUSD-1496 export stages as relative references

pierrebai-adsk:
The problem was that the fact that the stage root layer was a relative path was lost when saved mid-export. Change the code so that the knowledge of the user's choice is always correctly preserved. Then the exported stage can correctly detect that the file path is relative.

- Add a function to check if a proxy shape is holding the file path as relative.
- Modify the setNewProxyPath function to take a flag to control if the new path is relative or not, including optionally following the mode selected in a option var or in the proxy shape.
- Use that new parameter when saving anonymous layers based on the known information that is passed when saving that anonymous layer, in the saveAnonymousLayer function.
- Use that new parameter when converting anonymous layers to saved layers.
- Use that new parameter when handling notification that the root layer path changed.

### DIFF
--- a/lib/mayaUsd/utils/utilSerialization.h
+++ b/lib/mayaUsd/utils/utilSerialization.h
@@ -64,6 +64,33 @@ enum USDUnsavedEditsOption
 MAYAUSD_CORE_PUBLIC
 USDUnsavedEditsOption serializeUsdEditsLocationOption();
 
+/*! \brief Return if the relative-path plug is set to true on the proxy shape.
+ */
+MAYAUSD_CORE_PUBLIC
+bool isProxyShapePathRelative(MayaUsdProxyShapeBase& proxyShape);
+
+/*! \brief File path mode for the setNewProxyPath and isProxyPathModeRelative functions.
+ *         kProxyPathRelative makes the file path relative to the Maya scene.
+ *         kProxyPathAbsolute makes the file path absolute.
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/utils/utilSerialization.h
```diff
@@ -64,6 +64,33 @@ enum USDUnsavedEditsOption
...
+ */
+enum ProxyPathMode
+{
+    kProxyPathRelative,
+    kProxyPathAbsolute,
+    kProxyPathFollowProxyShape,
+    kkProxyPathFollowOptionVar
+};
+
+/*! \brief Convert the proxy path mode into a boolean telling if the path should relative.
```
AramAzhari-adsk, line 88: minor typo close to the end.

AramAzhari-adsk (approved):

lib/mayaUsd/utils/utilSerialization.h
pierrebai-adsk: I'll take note of it for a future PR, seems minor to respin everything.

seando-adsk: kkProxyPathFollowOptionVar also has two "k"

AramAzhari-adsk (approved):

seando-adsk (approved):

PR #3902, opened 2024-09-04
LOOKDEVX-3222 - Expose hardcoded texture paths

JGamache-autodesk:
This will expose the MaterialX path of a library definition containing hardcoded image paths, allowing clients to fetch the relevant data required to populate those shader inputs.


frohnej-adsk (approved):
Nice, thanks!

seando-adsk:
Preflight (from 2 weeks ago) failed and has been lost now. I'll re-run it.

PR #3904, opened 2024-09-05
Bailp/emsusd 1091/save and restore non local edit target layer

pierrebai-adsk:
See original PR here: https://github.com/Autodesk/maya-usd/pull/3716
This copy fixes the unit test.


csyshing (approved):
Thanks for fixing the unit tests, @pierrebai-adsk !

csyshing (approved):

seando-adsk (approved):

PR #3906, opened 2024-09-06
EMSUSD-1629 fix copying grouped meshes

pierrebai-adsk:
- Return the correct path of the copied object from the duplicate function.
- Don't try to process the virtual root prim path.
- Prepare the correct export options.
- Add a unit test for duplicate grouped meshes.

### DIFF
--- a/plugin/adsk/scripts/mayaUsdMenu.mel
+++ b/plugin/adsk/scripts/mayaUsdMenu.mel
@@ -708,7 +708,7 @@ global proc mayaUsdMenu_duplicateToUSD( string $proxy, string $obj )
     }
 
     for ($item in $bulkObjects) {
-        string $exportOptions = `python("import mayaUsdDuplicateAsUsdDataOptions; import mayaUsdOptions; mayaUsdOptions.setAnimateOption('''" + $obj + "''', mayaUsdDuplicateAsUsdDataOptions.getDuplicateAsUsdDataOptionsText())")`;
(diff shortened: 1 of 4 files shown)


plugin/adsk/scripts/mayaUsdMenu.mel
```diff
@@ -708,7 +708,7 @@ global proc mayaUsdMenu_duplicateToUSD( string $proxy, string $obj )
     }
 
     for ($item in $bulkObjects) {
-        string $exportOptions = `python("import mayaUsdDuplicateAsUsdDataOptions; import mayaUsdOptions; mayaUsdOptions.setAnimateOption('''" + $obj + "''', mayaUsdDuplicateAsUsdDataOptions.getDuplicateAsUsdDataOptionsText())")`;
+        string $exportOptions = `python("import mayaUsdDuplicateAsUsdDataOptions; import mayaUsdOptions; mayaUsdOptions.setAnimateOption('''" + $item + "''', mayaUsdDuplicateAsUsdDataOptions.getDuplicateAsUsdDataOptionsText())")`;
```
AramAzhari-adsk, line 711: Is obj the parent node? and if so, do we also need to setAnimateOption on that?
Was item ever used before this change?

pierrebai-adsk: $item is replacing $obj, but in the original change I forgot to change it here.

AramAzhari-adsk (approved):

PR #3907, opened 2024-09-10
EMSUSD-1612 allow duplicating material to USD

pierrebai-adsk:
Refactor the prim update manager:
- Refactor the duplicate function into two functions: duplicateToMaya and duplicateToUsd.
- The duplicateToUsd function takes a Maya node instead of a UFE path to support Maya materials, which are not DAG objects.
- Support non-DAG object in the pushExport function when creating the full object list.
- This is necessary so that the material-only export correctly identify the material as being in the selection.

Improve the duplicate command:
- Support parsing pure Maya object names as the first argument to the duplicate command.
- When the first argument cannot be parsed as UFE then directly call the USD duplicate function with the Maya object instead of the UFE path.

Add a unit test for duplicating to USD a material.


lib/mayaUsd/fileio/primUpdaterManager.cpp
```diff
@@ -580,14 +580,19 @@ PushExportResult pushExport(const MObject& mayaObject, const UsdMayaPrimUpdaterC
```
pierrebai-adsk, line 580: For the diff in this file, turn off whitespace changes, because a lot of line only changed indenttion.

barbalt (dismissed):

PR #3913, opened 2024-09-12
Use legacy DefaultPrim logic if using legacy material scopes

dj-mcg:
Since the default prim logic was needed to determine the parent scope of materials in the current material scope behavior, it isn't needed if using the legacy material scope strategy


dj-mcg:
The change that introduced the new default prim logic is here:
https://github.com/Autodesk/maya-usd/pull/3867

I left some comments in that change since we encountered several issues with this change. Ultimately we decided to opt completely out since it only seems necessary for the new material scope parenting logic using the same "legacy" export option used for the material scope behavior itself.

There are remaining issues here, though. The new default prim logic can pick a root maya node that isn't even part of the export. Also, the logic of choosing a default prim in the _FinishWriting and _WriteVariants functions has been discarded. I'm wondering if the new behavior of picking a seemingly random root node in the maya file is actually correct?

pierrebai-adsk (approved):

PR #3919, opened 2024-09-17
EMSUSD-1555 Duplicating a Physical Sky to USD using Arnold options is missing data

barbalt:
When "Duplicate as Usd data" is select from the context ops menu on a maya object, we only translate to Usd the selected maya object.
This can be a problem when there are extra prims created by a chaser, for example. Those would be discarded.

With those changes, we allow the users to register the path for those extra prims created so that it can be queried when merging the maya object into Usd.


pierrebai-adsk (dismissed):
Great!

pierrebai-adsk:
IDK if we would not be better off squashing all those commits into a single one? Maybe creating a new PR?

barbalt:
> IDK if we would not be better off squashing all those commits into a single one? Maybe creating a new PR?
@pierrebai-adsk I can do that. Out of curiosity, can't we squash during merge?

lib/mayaUsd/fileio/jobs/writeJob.cpp
```diff
@@ -640,12 +634,19 @@ bool UsdMaya_WriteJob::_FinishWriting()
...
+
     // Run post export function on the chasers.
     MayaUsd::ProgressBarLoopScope chasersLoop(mChasers.size());
     for (const UsdMayaExportChaserRefPtr& chaser : mChasers) {
         if (!chaser->PostExport()) {
             return false;
         }
+
+        // Collect extra prims paths from chasers
+        for (const SdfPath& path : chaser->GetExtraPrimsPaths()) {
```
pierrebai-adsk, line 647: I think there is a way to insert a range in C++ std::vector? Not super important.

pierrebai-adsk (approved):

PR #3920, opened 2024-09-18
EMSUSD-986 fix export of instances with transform on geometry

pierrebai-adsk:
The mesh reader (MayaUsdPrimReaderMesh) which handles the import of USD meshes, was *not* creating a Maya transform node for the USD geom prim when in a Maya prototype. Since Maya shape nodes do not have transforms, this prevented the reader from importing any transform that was authored directly on the USD geom prim.

This behavior of not creating the transform node was designed so that round-trip between USD import and export of instances would not create new Maya transform nodes on each round-trip. Unfortunately, this is incorrect when the USD geometry node has transforms. We cannot work around this, USD allows transform on geometry nodes, Maya does not.

The fix is to detect that the USD geom prim has transforms on it and create the Maya transform node in that case. Correctness trumps "nice" round-tripping.

- Add a HasXformOps helper function to detect if a USD prim has XformOps authored.
- Use the helper to decide of prototype nodes should have a transform node or not.
- Use the helper to decide how to name the Maya geometry shape node.
- Add a unit test to cover this situation.


lib/usd/translators/meshReader.cpp
```diff
@@ -79,10 +79,11 @@ bool MayaUsdPrimReaderMesh::Read(UsdMayaPrimReaderContext& context)
...
-    // Note: in prototype instances, we don't need an intermediary transform node.
+    // Note: in prototype instances, we don't need an intermediary transform node,
+    //       unless the prim has transforms that need to be preserved.
     //       See: EMSUSD-80
     const bool inPrototype = UsdPrim::IsPathInPrototype(prim.GetPath());
-    if (inPrototype) {
+    if (inPrototype && !UsdMayaTranslatorUtil::HasXformOps(prim)) {
         transformObj = parentNode;
     } else {
         bool retStatus = UsdMayaTranslatorUtil::CreateTransformNode(
```
pierrebai-adsk, line 89: This is the key fix: when the prototype has transform on the prim, a Maya transform node must be created to hold that transform.

barbalt (approved):

PR #3922, opened 2024-09-19
Copy a test file for use in overwrite testing

dj-mcg:
Modifying a test file that is used in multiple tests is causing issues in our test framework. This change splits the file that is overwritten during the testReferenceCommands test to avoid corrupting other tests that are trying to use the same test file


seando-adsk (approved):

dj-mcg:
I see that this causes the `2022 Windows python3` build to fail, which is surprising. Also, I don't seem to have access to the individual logs - I only see a `result` logfile and a `start` logfile

seando-adsk:
@dj-mcg The 2022 build failed because of an internal network glitch accessing git. Unrelated to your changes. Since the other 2022 builds passed, I'll mark merge this.

Sean

PR #3927, opened 2024-09-23
Copy input file for test in separate location

dj-mcg:
Rather than have an external dependency in a mayaUsd test, copy the test file locally


pierrebai-adsk (approved):

seando-adsk:
@dj-mcg There was one test failure on Maya 2022, python 2, Windows with the test `testUsdExportPackage`. I have seen this test randomly fail in the past, so I don' think its related to your changes. But just wanted to verify with you first before merging.

dj-mcg:
> @dj-mcg There was one test failure on Maya 2022, python 2, Windows with the test `testUsdExportPackage`. I have seen this test randomly fail in the past, so I don' think its related to your changes. But just wanted to verify with you first before merging.

I didn't do anything python specific with the change so it would be surprising if my change was causing the failure. Unfortunately I'm not able to easily set up Windows tests on my end so I can't confirm, but I would assume it's safe

PR #3930, opened 2024-09-25
EMSUSD-949 up axis in USD export

pierrebai-adsk:
- Add "-upAxis" (-upa) option to the export command.
- Document it.
- Add upAxis to the export job arguments structure.
- Add upAxis and its possible values to the export tokens.
- Expose upAxis job arguments to Python.
- Add up axis UI to the export dialog.
- Implement temporarily changing Maya up-axis. We rotate all root nodes 90 degrees before and after the export.
- Add unit test.


lib/mayaUsd/fileio/jobs/writeJob.cpp
```diff
@@ -112,6 +112,112 @@ static TfToken _GetFallbackExtension(const TfToken& compatibilityMode)
     return UsdMayaTranslatorTokens->UsdFileExtensionDefault;
 }
 
+class AutoUpAxisChanger
```
pierrebai-adsk, line 115: This is how change of axis during export is done: we temporarily rotate the Maya scene so that we can rely on the existing correct behavior already written in Maya to handle all possible nodes properly.

PR #3934, opened 2024-09-26
EMSUSD-1654 Fixed Test a typo in testLight's testUsdVolumeLights

AramAzhari-adsk:
EMSUSD-1654 Fixed Test a typo in testLight's testUsdVolumeLights that causes the test to fail when there's no `Light_v5_5` attribute defined.


seando-adsk (approved):

AramAzhari-adsk:
The preflights succeeded with one platform randomly failing on an unrelated test case which is known: 

`256/256 Test #233: testUsdExportAssemblyEdits ......................***Timeout 500.01 sec`

PR #3937, opened 2024-09-30
EMSUSD-948 scale export to USD according to units

pierrebai-adsk:
Add UI for export units:
- Add a new -unit (-unt) flag to the export command.
- Change the existing -exportDistanceUnit default to be false, to avoid clashing with the new -unit option.
- Document it.
- Add corresponding unit tokens and value in the job arguments.
- Expose it to Python.
- Add UI to control the flag in the export dialog.

Scale during export:
- Use the same method as for up-axis: modify root node before export and undo after export.
- That is group all root nodes and scale them, export to USD, then ungroup.
- Re-use the same group as for up-axis and same undo, for efficiency.
- Correctly handle the difference between Maya UI units and Maya internal data units.

Add unit tests.


lib/mayaUsd/fileio/jobs/jobArgs.cpp
```diff
@@ -1111,7 +1127,7 @@ const VtDictionary& UsdMayaJobExportArgs::GetDefaultDictionary()
         d[UsdMayaJobExportArgsTokens->exportAssignedMaterials] = true;
         d[UsdMayaJobExportArgsTokens->legacyMaterialScope] = false;
         d[UsdMayaJobExportArgsTokens->exportDisplayColor] = false;
-        d[UsdMayaJobExportArgsTokens->exportDistanceUnit] = true;
+        d[UsdMayaJobExportArgsTokens->exportDistanceUnit] = false;
```
pierrebai-adsk, line 1130: Since the default for the new unit option is "Maya Prefs", we are better off with exportDistanceUnit being off by deault, otherwise it would overwrite what the unit option would try to do.

lib/mayaUsd/fileio/jobs/writeJob.cpp
```diff
@@ -388,7 +490,10 @@ bool UsdMaya_WriteJob::_BeginWriting(const std::string& fileName, bool append)
     }
 
     // Temporarily change Maya's up-axis if needed.
-    _autoAxisChanger = std::make_unique<AutoUpAxisChanger>(mJobCtx.mStage, mJobCtx.mArgs.upAxis);
+    _autoAxisAndUnitsChanger = std::make_unique<AutoUpAxisAndUnitsChanger>(
+        mJobCtx.mStage, mJobCtx.mArgs.upAxis, mJobCtx.mArgs.unit);
+
+    // TODO: handle mJobCtx.mArgs.unit
```
barbalt, line 496: Doesn't the command above also handle the units?

barbalt (approved):

lib/mayaUsd/fileio/jobs/writeJob.cpp
pierrebai-adsk: Ah yes, I will remove this comment in the next PR

dgovil:
Hi, I know this PR is about a year old by now, but is there a reason you didn't just hook up the units to the existing metersPerUnit?

We're now seeing divergence between the two and I'm not sure why it was necessary to add another argument to the mix instead of unifying them, or why they weren't just forwarding into the same codepath.

pierrebai-adsk:
We did not change the behavior of the other flags because we did not want to break backward compatibility.

dgovil:
I guess I'm not sure why this would break backward compatibility unless
there's some undocumented result difference between the two?
I can understand keeping the old flags around to avoid breaking script
usage, but surely they should result in equivalent outputs?

If they do not, is the difference documented?

pierrebai-adsk:
That is what I meant.

barbalt:
@dgovil can you give an example of what you are seeing that is different than the expected?

dgovil:
We’re seeing a character rig that exports with different issues depending
on if we use the float attribute or the string one.

I’m trying to see if we can distill it down to something we can share in an
issue.

PR #3941, opened 2024-10-02
EMSUSD-947 units conversion for USD import

pierrebai-adsk:
- Add -unit (-unt) option to the import command.
- Document it.
- Add unit option to the job import argument structure.
- Parse the unit option.
- Expose it to Python.
- Add check-box UI to the import UI.
- Add private structure to the import read job class to hold all conversion info.
- Remove old warning about unit conversion not being implemented.
- Refactor the code for up-axis conversion to also handle the units conversion.
- Change prefs for units if requested and needed.
- Add unit tests.

### DIFF
--- a/lib/mayaUsd/utils/util.h
+++ b/lib/mayaUsd/utils/util.h
@@ -152,6 +152,11 @@ double ConvertMDistanceUnitToUsdGeomLinearUnit(const MDistance::Unit mdistanceUn
 MAYAUSD_CORE_PUBLIC
 MDistance::Unit ConvertUsdGeomLinearUnitToMDistanceUnit(const double linearUnit);
 
+/// Convert the given \p mdistanceYnit into its text representation suitable
+/// to be used with the currentUnit MEL command. Invalid units return "cm".
+MAYAUSD_CORE_PUBLIC
+MString ConvertMDistanceUnitToText(const MDistance::Unit mdistanceUnit);
+
 /// Returns a scaling value from Maya's internal units to the specified \p metersPerUnit
 MAYAUSD_CORE_PUBLIC
 double GetExportDistanceConversionScalar(const double metersPerUnit);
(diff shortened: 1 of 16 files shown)


lib/mayaUsd/utils/util.h
```diff
@@ -152,6 +152,11 @@ double ConvertMDistanceUnitToUsdGeomLinearUnit(const MDistance::Unit mdistanceUn
 MAYAUSD_CORE_PUBLIC
 MDistance::Unit ConvertUsdGeomLinearUnitToMDistanceUnit(const double linearUnit);
 
+/// Convert the given \p mdistanceYnit into its text representation suitable
```
barbalt, line 155: small typo

barbalt (approved):
Looks good. My only concern is with skinned data: Skel and morphers. Specially when importing then re-exporting it (concerned about bone scaling)

lib/mayaUsd/utils/util.h
pierrebai-adsk: Will be fixed in next PR about units and up-axis in USD stages.

pierrebai-adsk:
> Looks good. My only concern is with skinned data: Skel and morphers. Specially when importing then re-exporting it (concerned about bone scaling)

Anything specific? You can log issues in Jira if you think something cold be broken.

barbalt:
> > Looks good. My only concern is with skinned data: Skel and morphers. Specially when importing then re-exporting it (concerned about bone scaling)
> 
> Anything specific? You can log issues in Jira if you think something cold be broken.

I have some bad experiences when implementing the import/export when converting different units. Scaling was the biggest culprit of bad things happening.. Animations wouldn't translate properly due to mesh have different matrices and so on.. Hence why I mentioned scaling in particular.. 

I guess we'll see with tests..

PR #3945, opened 2024-10-04
EMSUSD-943 handle up-axis and units when staging a USD file

pierrebai-adsk:
Up-axis and units UI when staging a USD file:
- Add up-axis, unit and conversion method UI to the staging UI.
- Add that the conversion method menu is disabled in the import UI.
- Make import sections collapsable.
- Fix typo in a header file.

Up-axis and units conversion for stages:
- Add Python helpers to do the conversion since MEL cannot manipulate USD.
- Call the helpers from the MEL script that creates the stage.
- Change prefs or modify stage shape based on necessary conversion.

Add unit tests
- Move helper functions to verify transforms to their own file.
- Update some existing tests to use the new functions.
- Add unit tests when stage USD file for up-axis and units conversion.


lib/mayaUsd/resources/scripts/mayaUsdStageConversion.py
```diff
@@ -0,0 +1,135 @@
...
+            return _UsdGeom.LinearUnits.centimeters
+        return StageConversionInfo._mayaToMetersPerUnit[mayaUnits]
+    
+    _metersPerUnitToMayaUnitName = {
+        _UsdGeom.LinearUnits.inches      : "inch",
+        _UsdGeom.LinearUnits.feet        : "foot",
+        _UsdGeom.LinearUnits.yards       : "yard",
+        _UsdGeom.LinearUnits.miles       : "mile",
+        _UsdGeom.LinearUnits.millimeters : "mm",
+        _UsdGeom.LinearUnits.centimeters : "cn",
```
samuelliu-adsk, line 70: This one should be "cm"

samuelliu-adsk (approved):

PR #3948, opened 2024-10-07
EMSUSD-1700 make up-axis and units attribute visible

pierrebai-adsk:
- Change the dynamic attribute APi to allow controlling the attribute flags.
- Make the up-axis and units attributes be visible.

### DIFF
--- a/lib/mayaUsd/utils/dynamicAttribute.cpp
+++ b/lib/mayaUsd/utils/dynamicAttribute.cpp
@@ -29,19 +29,27 @@ bool hasDynamicAttribute(const MFnDependencyNode& depNode, const MString& attrNa
     return depNode.hasAttribute(attrName);
 }
 
-MStatus createDynamicAttribute(MFnDependencyNode& depNode, const MString& attrName)
+MStatus createDynamicAttribute(
+    MFnDependencyNode&     depNode,
+    const MString&         attrName,
+    const DynamicAttrFlags flags)
 {
     MStatus status;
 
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/utils/dynamicAttribute.cpp
```diff
@@ -29,19 +29,27 @@ bool hasDynamicAttribute(const MFnDependencyNode& depNode, const MString& attrNa
...
+    MFnTypedAttribute attrFn;
+    MObject attr = attrFn.create(attrName, "", MFnData::kString, MObject::kNullObj, &status);
     CHECK_MSTATUS_AND_RETURN_IT(status);
 
-    typedAttrFn.setReadable(true);
-    typedAttrFn.setWritable(true);
-    typedAttrFn.setHidden(true);
-    typedAttrFn.setKeyable(false);
-    typedAttrFn.setStorable(true);
+    attrFn.setAffectsAppearance(isFlagSet(flags, DynamicAttrFlags::kAppearance) ? true : false);
```
barbalt, line 43: minor personal preference. We could remove all the ternary operators here since isFlagSet already returns the expected true/false

barbalt (dismissed):

lib/mayaUsd/utils/dynamicAttribute.cpp
pierrebai-adsk: Oh you're right, the code was a bit different before and that is a remnant

barbalt (approved):

PR #3949, opened 2024-10-08
Use macros for the pxr namespace

dj-mcg:


dj-mcg:
> Thanks for fixing these. I noticed there is one more in UsdClipboardHandler.cpp, does that one also affect you?

Hi @seando-adsk - that one doesn't affect us yet... We're still on Maya2023/UFE3. Since these are easy to fix and only affect Pixar internally, I'm happy to continue fixing them up as we encounter them.

PR #3951, opened 2024-10-08
EMSUSD-1682 assign material to lock layer

pierrebai-adsk:
Fix crash when trying to assigned a material to a prim in a locked layer. The code was failing to check for null in a critical place.

- Check for null item or stage when renaming prim.
- Check for null after creating a scope.
- Properly sequence undo vs undo-block when the execution fails.
- Add a unit test to verify the crash fix.


frohnej-adsk (approved):
LGTM!

PR #3956, opened 2024-10-10
EMSUSD-1719 units prefs during import

pierrebai-adsk:
When using the overwrite-prefs method to change the units during import, we need to compare with teh Maya UI units, not the internal units. (When using the scaling method, we are scaling to internal units.)


pierrebai-adsk:
@AramAzhari-adsk ping for this review

PR #3959, opened 2024-10-11
EMSUSD-1504 prevent warnings when reloading the plugin

pierrebai-adsk:
Use our own version of the function to register string resources to avoid spamming warning about duplicate resources.


AramAzhari-adsk (dismissed):

seando-adsk (changes requested):
How do the existing MEL files that call `getMayaUsdLibString` work now? You've deleted the MEL file `mayaUsdLibRegisterStrings.mel` that contained this proc. I think those scripts will error now.

pierrebai-adsk:
Yes, that is a mistake. I must not have opened a window that depended on them when testing. I fill revert that part.

pierrebai-adsk:
The reason I missed it is that they are only used in a single dialog: Add Maya Reference. Added them back and tested that window. Now it works again.

AramAzhari-adsk (approved):

seando-adsk (approved):

test/lib/mayaUsd/nodes/testProxyShapeConversion.py
```diff
@@ -50,9 +50,11 @@ def setUp(self):
         """Clear the scene and setup the Maya preferences."""
         cmds.file(f=True, new=True)
         # Make sure up-axis is Z.
-        cmds.upAxis(axis='z')
+        if cmds.upAxis(query=True, axis=True).lower() != 'z':
```
AramAzhari-adsk, line 53: Do these occur when reloading the plugin?

AramAzhari-adsk (approved):

test/lib/mayaUsd/nodes/testProxyShapeConversion.py
pierrebai-adsk: The commands were issuing warnings that the up-axis was already Z, I'm just reducing the amount of noise when running the tests.

PR #3965, opened 2024-10-17
EMSUSD-1722 find the correct layer manager node

pierrebai-adsk:
When loading a Maya scene containing a stage, possibly through other Maya reference, find the correct layer manager node for the stage. Each layer manager node is specific to the scene or reference.

- Add an optional parameter to some layer manager function to specify the Maya reference that should contain the information.
- Pass the correct Maya reference, if any, when computing the proxy shape node layers.
- Add a unit test to load a stage in a Maya reference.

### DIFF
--- a/lib/mayaUsd/nodes/layerManager.cpp
+++ b/lib/mayaUsd/nodes/layerManager.cpp
@@ -114,7 +114,10 @@ MStatus disconnectCompoundArrayPlug(MPlug arrayPlug)
     return dgmod.doIt();
 }
 
-static bool isFromReference(const MFnReference* fromReference, const MFnDependencyNode& node)
+/// @brief Verify if the given node is either from the given reference, if that is not null,
+///        or from the main Maya scene if the reference is null.
+static bool
+isNodeFromDesiredOrigin(const MFnReference* fromReference, const MFnDependencyNode& node)
 {
     if (fromReference) {
         return fromReference->containsNodeExactly(node.object());
@@ -139,9 +142,11 @@ MayaUsd::LayerManager* findNode(const MFnReference* fromReference)
     for (; !iter.isDone(); iter.next()) {
         MObject mobj = iter.item();
         fn.setObject(mobj);
-        if (fn.typeId() == MayaUsd::LayerManager::typeId && isFromReference(fromReference, fn)) {
-            layerManagerHandle = mobj;
-            return static_cast<MayaUsd::LayerManager*>(fn.userNode());
+        if (fn.typeId() == MayaUsd::LayerManager::typeId) {
+            if (isNodeFromDesiredOrigin(fromReference, fn)) {
+                layerManagerHandle = mobj;
+                return static_cast<MayaUsd::LayerManager*>(fn.userNode());
+            }
         }
     }
     return nullptr;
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -130,17 +139,17 @@ MayaUsd::LayerManager* findNode()
...
+        if (fn.typeId() == MayaUsd::LayerManager::typeId && isFromReference(fromReference, fn)) {
             layerManagerHandle = mobj;
             return static_cast<MayaUsd::LayerManager*>(fn.userNode());
         }
     }
     return nullptr;
 }
 
-MayaUsd::LayerManager* findOrCreateNode()
+MayaUsd::LayerManager* findOrCreateNode(const MFnReference* fromReference)
```
pierrebai-adsk, line 150: Most of the changes are related to adding this new fromReference parameter to a few functions.

lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -114,7 +114,16 @@ MStatus disconnectCompoundArrayPlug(MPlug arrayPlug)
     return dgmod.doIt();
 }
 
-MayaUsd::LayerManager* findNode()
+static bool isFromReference(const MFnReference* fromReference, const MFnDependencyNode& node)
```
pierrebai-adsk, line 117: Re-reading the code, this function name and getMayaReferenceOrigin in ProxyShapeBase.cpp are not the best names.

Maybe this function should be named isNodeFromCorrectSceneFile or something like that.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -684,12 +684,35 @@ MStatus MayaUsdProxyShapeBase::compute(const MPlug& plug, MDataBlock& dataBlock)
     return MS::kUnknownParameter;
 }
 
+static std::unique_ptr<MFnReference> getMayaReferenceOrigin(const MayaUsdProxyShapeBase& proxyShape)
```
pierrebai-adsk, line 687: Maybe this function should be renamed getProxyNodeOrigin

test/lib/testMayaUsdCreateStageInMayaRef.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python
+
+#
+# Copyright 2020 Autodesk
```
barbalt, line 4: minor. This is a new file, should it be 2024 instead of 2020?

lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -600,7 +614,8 @@ std::string LayerDatabase::getSelectedStage() const { return _selectedStage; }
 
 bool LayerDatabase::saveLayerManagerSelectedStage()
 {
-    MayaUsd::LayerManager* lm = findOrCreateNode();
+    const MFnReference*    fromReference = nullptr;
+    MayaUsd::LayerManager* lm = findOrCreateNode(fromReference);
```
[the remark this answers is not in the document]
pierrebai-adsk: Depends on personal preferences, I prefer to be explicit, but both ways are fine. Initially, all functions (like saveLayerManagerSelectedStage) received fromReferences, but I realized that the save functions are never saving Maya references. I'll change it since I need to change the copyright anyway.

barbalt (approved):

PR #3969, opened 2024-10-22
EMSUSD-1521 refactor AE template

pierrebai-adsk:
- Move each custom control to its own file.
- Rename custom control and their file to follow the same pattern.
- Move the creator function to static function in the corresponding custom control.
- Move some helper functions to their corresponding custom control.
- Rename some parameters to have clearer name (attrNames instead of controls and attrName instead of c).
- Document how the AE template code works.


pierrebai-adsk:
While the change may seem big, most of it is simply code that was moved out of the ae_template.py file into their own file. There is no new code in the new files. (Other than making all the custom control class names follow the same naming pattern.)

lib/mayaUsd/resources/ae/usdschemabase/__init__.py
```diff
@@ -1,5 +1,5 @@
 from .ae_template import AETemplate
-from .custom_enum_control import customEnumControlCreator
+from .enumCustomControl import EnumCustomControl
```
AramAzhari-adsk, line 2: This could potentially impact external scripts that use the old naming.

pierrebai-adsk: I doubt user scripts were referring to internal classes used to create the AE UI. Assuming otherwise would prevent us from ever modifying code.

AramAzhari-adsk (approved):

PR #3970, opened 2024-10-22
Make unit test scene file names unique

pierrebai-adsk:
Two tests were using the same scene file names. Maybe they were randomly overwriting each other when run in parrallel.

### DIFF
--- a/test/lib/mayaUsd/fileio/testSaveLockedAnonLayer.py
+++ b/test/lib/mayaUsd/fileio/testSaveLockedAnonLayer.py
@@ -122,11 +122,11 @@ class SaveLockedLayerTest(unittest.TestCase):
 
     def testSaveSystemLockedAnonLayerInUSD(self):
         # Save the file. Make sure the USD edits will go to a USD file.
-        self._runTestSaveLockedAnonLayer(1)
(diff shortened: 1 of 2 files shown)


test/lib/mayaUsd/fileio/testSaveLockedAnonLayer.py
```diff
@@ -122,11 +122,11 @@ def verifyPrims(stage):
 
     def testSaveSystemLockedAnonLayerInUSD(self):
         # Save the file. Make sure the USD edits will go to a USD file.
-        self._runTestSaveLockedAnonLayer(1)
+        self._runTestSaveSystemLockedAnonLayer(1)
```
seando-adsk, line 125: Was this (and one below) using the wrong function before?

seando-adsk (approved):

test/lib/mayaUsd/fileio/testSaveLockedAnonLayer.py
pierrebai-adsk: Yes! Two of the tests were not even run... thankfully, they were passing.

PR #3972, opened 2024-10-23
EMSUSD-1630: Support custom callback for primitives AE template

seando-adsk:
#### EMSUSD-1630: Support custom callback for primitives AE template
* Added new `triggerUICallback` function and removed `getUICallbacks`
* Updated AE doc on how to use new AE custom callback.
* New custom callback section in AE template giving clients ability to add to AE template.
* New attribute nice name callback to give clients ability to inject function into attribute nice name.

### DIFF
--- a/lib/mayaUsd/commands/layerEditorCommand.cpp
+++ b/lib/mayaUsd/commands/layerEditorCommand.cpp
@@ -1078,8 +1078,7 @@ private:
 
     void _notifySystemLockIsRefreshed()
     {
-        auto dstCallbacks = UsdUfe::getUICallbacks(TfToken("onRefreshSystemLock"));
-        if (dstCallbacks.size() == 0)
+        if (!UsdUfe::isUICallbackRegistered(TfToken("onRefreshSystemLock")))
             return;
 
         PXR_NS::VtDictionary callbackContext;
@@ -1095,8 +1094,7 @@ private:
         VtStringArray lockedArray(affectedLayers.begin(), affectedLayers.end());
         callbackData["affectedLayerIds"] = lockedArray;
 
-        for (UsdUfe::UICallback::Ptr& dstCallback : dstCallbacks)
-            (*dstCallback)(callbackContext, callbackData);
+        UsdUfe::triggerUICallback(TfToken("onRefreshSystemLock"), callbackContext, callbackData);
     }
 
     UsdStageWeakPtr getStage()
--- a/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
@@ -29,13 +29,14 @@ import re
 import ufe
 import maya.mel as mel
 import maya.cmds as cmds
+import mayaUsd.lib as mayaUsdLib
 import mayaUsd.ufe as mayaUsdUfe
 import maya.internal.common.ufe_ae.template as ufeAeTemplate
 from mayaUsdLibRegisterStrings import getMayaUsdLibString
 
 # We manually import all the classes which have a 'GetSchemaAttributeNames'
 # method so we have access to it and the 'pythonClass' method.
-from pxr import Usd, UsdGeom, UsdShade, Tf, Sdr
+from pxr import Usd, UsdGeom, UsdShade, Tf, Sdr, Vt
 
 
 def defaultControlCreator(aeTemplate, attrName):
@@ -235,6 +236,7 @@ class AETemplate(object):
         cmds.editorTemplate(beginScrollLayout=True)
         self.buildUI()
         self.createAppliedSchemasSection()
+        self.createCustomCallbackSection()
         self.createCustomExtraAttrs()
         self.createMetadataSection()
         cmds.editorTemplate(endScrollLayout=True)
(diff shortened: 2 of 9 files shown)


lib/mayaUsd/commands/layerEditorCommand.cpp
```diff
@@ -1078,8 +1078,7 @@ class RefreshSystemLockLayer : public BaseCmd
 
     void _notifySystemLockIsRefreshed()
     {
-        auto dstCallbacks = UsdUfe::getUICallbacks(TfToken("onRefreshSystemLock"));
-        if (dstCallbacks.size() == 0)
+        if (!UsdUfe::isUICallbackRegistered(TfToken("onRefreshSystemLock")))
```
seando-adsk, line 1081: New trigger callback function and removed the get. I think this is safer than allowing code to access all the callbacks.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -492,6 +494,35 @@ def createAppliedSchemasSection(self):
...
+            'ufe_path_string' : ufe.PathString.string(self.item.path())
+        }
+        cbContextDict = Vt._ReturnDictionary(cbContext)
+        cbDataDict = Vt._ReturnDictionary({})
+
+        # Trigger the callback which will give other plugins the opportunity
+        # to add controls to our AE template.
+        global _aeTemplate
+        try:
+            _aeTemplate = self
```
seando-adsk, line 520: Set our template into the global variable so a callback can retrieve it using the new function `getAETemplateForCustomCallback` above. They can use this to add/suppress controls (which then follows all the rules like show array attributes, nice name, etc).

lib/mayaUsd/resources/ae/usdschemabase/attributeCustomControl.py
```diff
@@ -21,16 +21,51 @@
...
+        # If there are any nice attribute name functions registered call them now.
+        # this gives clients the ability to inject some nice naming.
+        global _niceAttributeNameFuncs
+        for fn in _niceAttributeNameFuncs:
+            tempName = fn(ufeAttr, attrName)
+            if tempName:
+                attrName = tempName
+    except:
+        # Do not let any of the callback failures affect our template.
+        pass
```
seando-adsk, line 66: Allow a client to inject a function into the attribute nice naming. This will give clients (such as Arnold) the opportunity to nice name their attributes.

lib/usdUfe/ufe/UsdContextOps.cpp
```diff
@@ -748,6 +748,7 @@ UsdContextOps::SchemaNameMap UsdContextOps::getSchemaPluginNiceNames() const
         { "usdSkel", "Skeleton" },
         { "usdUI", "UI" },
         { "usdVol", "Volumes" },
+        { "usdArnold", "Arnold" }
```
seando-adsk, line 751: Nicer name for Arnold prims in the context menu.

lib/usdUfe/utils/uiCallback.h
```diff
@@ -44,14 +42,6 @@ class USDUFE_PUBLIC UICallback
         = 0;
 };
 
-using UICallbacks = PXR_NS::
-    TfHashMap<PXR_NS::TfToken, std::vector<UICallback::Ptr>, PXR_NS::TfToken::HashFunctor>;
-
-// Retrieve the callbacks for the argument operation.
-// If no such callback exists, an empty vector is returned.
-USDUFE_PUBLIC
-const std::vector<UICallback::Ptr>& getUICallbacks(const PXR_NS::TfToken& operation);
```
seando-adsk, line 53: Hid all this in the .cpp file.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
pierrebai-adsk: There was not any way to pass it as a parameter to the callback instead? I find that having to go through a global not great, but if there are limitation onbacks, then I suppose we have no choice.

lib/mayaUsd/resources/ae/usdschemabase/attributeCustomControl.py
pierrebai-adsk, line 66: Maybe log the error like what was done in the other callback. (Would also silence the Chorus checker)

pierrebai-adsk (changes requested):
Great!

Just a few questions.

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
seando-adsk: No, not using the existing UICallback class which takes VtDictionary. I talked to PierreT about this and confirmed using the variable is okay. Its already done in our Ufe support plugin (in Maya) for the command.

lib/mayaUsd/resources/ae/usdschemabase/attributeCustomControl.py
seando-adsk: Okay will do.

pierrebai-adsk (dismissed):

pierrebai-adsk (approved):

PR #3974, opened 2024-10-24
EMSUSD-1758 author metadata only on anonymous root layers

pierrebai-adsk:
- Only author the up-axis and units metadata when saving a root layer on-disk.
- Update the unit test to verify this.


pierrebai-adsk:
@AramAzhari-adsk reminder about this review.

AramAzhari-adsk:
> @AramAzhari-adsk reminder about this review.

Thanks for the reminder. I had looked at it but didn't click on the button. Approved.

AramAzhari-adsk (approved):

PR #3861, opened 2024-07-22
Fix reparenting/renaming of pulled objects USD ancestor

jufrantz:
Hello,

This PR addresses an issue where pulled Maya objects transform and visibility were disconnected from the USD stages after renaming or reparenting a USD ancestor of the pulled prim. Our artists very frequently have multiple pulled objects in the same stage while re-structuring the hierarchy.

Before this fix:
beforeFixPullParentAncestor

Ater this fix:
afterFixPullParentAncestor

Changes included in this PR:

- Added the ability to cleanly reparent a DAG to a USD item using `mayaUsd.lib.proxyAccessor`, ensuring it does not leave disconnected proxyAccessor outputs, potentially accessing invalid prims: a25dcd60cb84981e69d2bb5f64a6625a3e45a59b and unit-test 415dcb653db407229cf07991a623c5379230326f.
- Small refactoring of `PrimUpdaterManager` to mutualize the `mayaUsd.lib.proxyAccessor.parent` python call: dc8956361e167de3ae292e36612665fab9d38ede.
- Added the proxyAccessor update in `OrphanedNodeManager`, following the `pullInformation` update: 2969283dc556eefe591182959870c97f0721018b.
- Modified CMake targets to define the `HAS_ORPHANED_NODES_MANAGER` env to add a unit test in testEditAsMaya.py. Disabled a failing test. It was failing without my changes with Maya2023 + USD-23.11 and Maya2024 + USD-23.11. The CMake/test changes: 243eecdec1bd768b5a800d2e4dff244e09cbad6b. The new test: 1a93630fc8942ab094a2d0181a613f0eb16b0fc6. 

--
Julien

PR #3979, opened 2024-10-28
Fix crash in Attribute Editor when renaming a prim with pulled descendants

jufrantz:
This PR addresses a crash that occurs when renaming a USD prim with descendant prims edited as Maya, while the Attribute Editor is visible and displaying the prim being renamed. I observed the issue on Maya 2023 and 2025 on Linux. It occurs since PR #3861.

The crash can be reproduced using this [maya scene](https://github.com/user-attachments/files/17546929/crash_ae_on_ufe_rename.zip). To trigger the issue, open the AE, select and rename a prim such as `/Root1`. This should result in a hard crash within `TPSdatabase::doRebuild`, typically producing the following stack trace:
```
  TPSdataNode::typeName(bool) const
  TPSdatabase::generateUI(QLayout*, TPSdataNode*, bool)
  TPSdatabase::generateUI(QLayout*, TdependNode*, bool)
  TPSdatabase::doRebuild(Tevent const&)
  /usr/autodesk/maya2025/lib/libSharedUI.so(+0x3cae17) [0x7f694dc14e17]
  TeventHandler::doIdles()
  /usr/autodesk/maya2025/lib/libExtensionLayer.so(+0x33ccf5) [0x7f6947411cf5]
  QObject::event(QEvent*)
  QApplicationPrivate::notify_helper(QObject*, QEvent*)
  /usr/autodesk/maya2025/lib/libExtensionLayer.so(+0x3271a3) [0x7f69473fc1a3]
  QCoreApplication::notifyInternal2(QObject*, QEvent*)
  QTimerInfoList::activateTimers()
  /usr/autodesk/maya2025/lib/libQt6Core.so.6(+0x3e36ac) [0x7f69440bb6ac]
  g_main_context_dispatch
  /lib64/libglib-2.0.so.0(+0x4dea8) [0x7f6926a24ea8]
  g_main_context_iteration
  QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>)
  QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>)
  QCoreApplication::exec()
  Tapplication::start()
  /usr/autodesk/maya2025/bin/maya.bin() [0x41ba3e]
  /usr/autodesk/maya2025/bin/maya.bin() [0x416b60]
  __libc_start_main
  /usr/autodesk/maya2025/bin/maya.bin() [0x41a1ce]
```

While investigating the cause of the crash, I simplified the context and was able to reproduce it using the following script. It triggers the crash by disconnecting a `proxyAccessor` output during `Ufe::ObjectRename`, which is a subset of the operations performed by OrphanedNodesManager on rename since PR #3861.

```python
import ufe
from maya import cmds, mel

# Open the test scene
cmds.file("crash_ae_on_ufe_rename.ma", open=True, force=True)

# Add observer
class CrashyObserver(ufe.Observer):
    def __call__(self, notif):
        if isinstance(notif, ufe.ObjectRename):
            for dest in cmds.connectionInfo("stageShape1.AP__Root1_combinedVisibility", dfs=True):
                cmds.disconnectAttr("stageShape1.AP__Root1_combinedVisibility", dest)

obs = CrashyObserver()
ufe.Scene.addObserver(obs)

# Select the prim to rename and ensure the AE is shown
cmds.select("|stage1|stageShape1,/Root1")
mel.eval('showEditor("|stage1|stageShape1,/Root1")')

# Wait and rename the prim. Maya shoud crash.
cmds.rename("NewRoot1")
```

I could not find the root cause of this crash within the maya-usd codebase. The issue may lie on the Maya or UFE side, related to the timing of AE updates and Ufe::ObjectRename events. The workaround proposed in this PR defers the proxyAccessor connection updates, having them occur at a more appropriate time for the AE, preventing the crash.

#### Included Changes:

- Modified `mayaUsd.lib.proxyAccessor` `parent` and `unparent` functions to accept an optional list of UFE children and parent paths: jufrantz/maya-usd@7bf2b98.
- Thanks to this, `OrphanedNodesManager` does not alter Maya's selection when updating `proxyAccessor` connections: jufrantz/maya-usd@008b0da.
- Updated `OrphanedNodesManager` to execute `ProxyAccessorUndoItem::parentPulledObject` on idle to avoid the crash. This change is enclosed within a `#ifdef` scope, making it easier to remove if a better fix becomes available: jufrantz/maya-usd@776e65d

### DIFF
--- a/lib/mayaUsd/fileio/orphanedNodesManager.cpp
+++ b/lib/mayaUsd/fileio/orphanedNodesManager.cpp
@@ -36,7 +36,7 @@
 #include <ufe/trie.imp.h>
 
 // Workaround to avoid a crash that occurs when the Attribute Editor is visible
-// and displaying a USD prim that is being renamed, while we are updating proxyAccessor 
+// and displaying a USD prim that is being renamed, while we are updating proxyAccessor
 // connections for its descendant pulled objects.
 #define MAYA_USD_AE_CRASH_ON_RENAME_WORKAROUND
 
--- a/lib/mayaUsd/fileio/orphanedNodesManager.cpp
+++ b/lib/mayaUsd/fileio/orphanedNodesManager.cpp
@@ -29,10 +29,16 @@
 #include <pxr/usd/usd/editContext.h>
 
 #include <maya/MFnDagNode.h>
+#include <maya/MGlobal.h>
 #include <maya/MPlug.h>
 #include <ufe/hierarchy.h>
 #include <ufe/sceneSegmentHandler.h>
 #include <ufe/trie.imp.h>
+
+// Workaround to avoid a crash that occurs when the Attribute Editor is visible
+// and displaying a USD prim that is being renamed, while we are updating proxyAccessor 
+// connections for its descendant pulled objects.
+#define MAYA_USD_AE_CRASH_ON_RENAME_WORKAROUND
 
 // For Tf diagnostics macros.
 PXR_NAMESPACE_USING_DIRECTIVE
@@ -163,8 +169,22 @@ void renamePulledObject(
     for (const PullVariantInfo& info : trieNode->data()) {
         const MDagPath& mayaPath = info.editedAsMayaRoot;
         TF_VERIFY(writePullInformation(pulledPath, mayaPath));
-        if (usdPathChanged) {
-            TF_VERIFY(reparentPulledObject(mayaPath, pulledPath.pop()));
+        if (usdPathChanged && mayaPath.isValid()) {
+#ifdef MAYA_USD_AE_CRASH_ON_RENAME_WORKAROUND
+            // Update the proxyAccessor connections on idle to avoid a crash when
+            // attribute editor is visible and showing the ancestor USD prim beeing renamed.
+            if (MGlobal::mayaState() == MGlobal::kInteractive) {
+                using ReparentArgs = std::pair<MDagPath, Ufe::Path>;
+                MGlobal::executeTaskOnIdle(
+                    [](void* data) {
+                        const auto* args = static_cast<ReparentArgs*>(data);
+                        TF_VERIFY(reparentPulledObject(args->first, args->second));
+                        delete args;
+                    },
+                    new ReparentArgs(mayaPath, pulledPath.pop()));
+            } else
+#endif
+                TF_VERIFY(reparentPulledObject(mayaPath, pulledPath.pop()));
         }
     }
 }
--- a/lib/mayaUsd/nodes/proxyAccessor.py
+++ b/lib/mayaUsd/nodes/proxyAccessor.py
@@ -325,7 +325,7 @@ def parentItems(ufeChildren, ufeParent, connect=True):
         # Cannot use 'visibility' here because it's already used by orphan manager
         connectParentChildAttr(parentVisibilityAttr, childDagPath, 'lodVisibility', connect)
 
-def __parent(*ufeItemPathStrings, doParenting=True, forceUnparenting=False):
+def __parent(doParenting, forceUnparenting, *ufeItemPathStrings):
    if ufeItemPathStrings:
       ufeSelectionList = [_createUfeSceneItem(pStr) for pStr in ufeItemPathStrings]
    else:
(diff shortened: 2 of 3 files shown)


seando-adsk:
@jufrantz Would it be possible to create a unit test out of the script you posted above? I guess it will have to an interactive test (run within Maya) in order to have the AE.

lib/mayaUsd/nodes/proxyAccessor.py
```diff
@@ -345,11 +347,11 @@ def __parent(doParenting, forceUnparenting=False):
    
    parentItems(ufeChildren, ufeParent, doParenting)
 
-def parent(force=False):
-    __parent(True, forceUnparenting=force)
+def parent(*ufeItemPathStrings, force=False):
```
pierrebai-adsk, line 350: This change is to provide more flexibility? I don't see anything calling this in this PR

lib/mayaUsd/fileio/orphanedNodesManager.cpp
```diff
@@ -163,8 +169,22 @@ void renamePulledObject(
     for (const PullVariantInfo& info : trieNode->data()) {
         const MDagPath& mayaPath = info.editedAsMayaRoot;
         TF_VERIFY(writePullInformation(pulledPath, mayaPath));
-        if (usdPathChanged) {
-            TF_VERIFY(reparentPulledObject(mayaPath, pulledPath.pop()));
+        if (usdPathChanged && mayaPath.isValid()) {
+#ifdef MAYA_USD_AE_CRASH_ON_RENAME_WORKAROUND
+            // Update the proxyAccessor connections on idle to avoid a crash when
+            // attribute editor is visible and showing the ancestor USD prim beeing renamed.
+            if (MGlobal::mayaState() == MGlobal::kInteractive) {
```
pierrebai-adsk, line 176: Is there a reason why we would only do this for interactive Maya? I know that sometimes when in non-interactive, on-idle are "too late" when scripting, is that the issue? It is somewhat unfortunate if it is, as it means that scripts run inside Maya will have the same problem. We do resort to this same kind of fix elsewhere, so it is not the first time, though.

pierrebai-adsk (commented):
Just waiting for the answers before approval.

lib/mayaUsd/fileio/orphanedNodesManager.cpp
jufrantz: Yes, that’s exactly it; I wasn’t sure if idle would even be triggered in a non-interactive Maya session. It's a similar logic to mel's `evalDeferred`, which performs an immediate `eval` when in batch mode. Since the issue is related to the Attribute Editor, I chose to apply this workaround in interactive sessions only. It's not an ideal solution even for interactive, but unfortunately I haven't been able to find a better fix.

lib/mayaUsd/nodes/proxyAccessor.py
jufrantz: `ProxyAccessorUndoItem::parentPulledObject` now passes the child and parent ufe paths as `proxyAccessor,parent` arguments. This is to avoid modifying maya's selection 

https://github.com/jufrantz/maya-usd/commit/008b0da#diff-f40afd2ca9921208caf4e38e91fa4ab00a2aa102b6fcf54fc82fdb4ff26427fdR46

jufrantz:
> @jufrantz Would it be possible to create a unit test out of the script you posted above? I guess it will have to an interactive test (run within Maya) in order to have the AE.

Alright, I can add a unit test that opens the AE and renames the selected prim to verify that the crash issue is resolved. I'll add this as soon as possible.

pierrebai-adsk (dismissed):

jufrantz:
@seando-adsk I attempted to add a test.

I made a simple script that crashes when run manually in a maya session, but only if I wait between showing the AE and renaming the prim. When executed in one go, it doesn’t crash. I attempted to add it as an interactive unit-test, using `cmds.flushIdleQueue` and `maya.utils.processIdleEvents` in place of the pause, as I suspect the crash might be related to queued AE refreshs, but this doesn’t crash.

It seems tricky to reproduce the crash in this context. Would it be acceptable if we skip adding this test?

seando-adsk:
@jufrantz I've asked our QA team to create an automation test for this, so yes you can skip adding the test. I'll start the preflight.

jufrantz:
> @jufrantz Unfortunately there were test failures on Maya 2022, python 2 builds:
> 
> ```
>   File "...\maya-usd\test\lib\testMayaUsdProxyAccessor.py", line 35, in <module>
>     from mayaUsd.lib import proxyAccessor as pa
>   File "W:\install\RelWithDebInfo\MayaUSD\lib\python\mayaUsd\lib\proxyAccessor.py", line 328
>     def __parent(*ufeItemPathStrings, doParenting=True, forceUnparenting=False):
>                                                 ^
> SyntaxError: invalid syntax
> ```
> 
> I think its because the keyword-only args are not allowed after the capture of positional args. This is a python3 feature.
> 
> Sean

@seando-adsk, 
Sorry for this. It is fixed by commit https://github.com/jufrantz/maya-usd/commit/a5a1178e797c7f2ae7c1db006002374e5dfe5668
Thank you,
Julien

pierrebai-adsk (approved):

seando-adsk:
@jufrantz As you can see I merged this PR. Thanks for submitting the crash fix and working thru all the test failures. 

Sean

PR #3982, opened 2024-10-29
EMSUSD-1455 no save during export

pierrebai-adsk:
- Add an helper function to check if a stage is dirty.
- Do not save dirty layers when exporting to USD.
- Instead, warn about anonymous or dirty stages.
- Add a unit test.

### DIFF
--- a/lib/usdUfe/utils/layers.h
+++ b/lib/usdUfe/utils/layers.h
@@ -152,9 +152,9 @@ bool isSessionLayer(
 // Verify if a stage has any dirty layer.
 enum class StageDirtyState : int
 {
-    Clean,
-    DirtyRootLayers,
-    DirtySessionLayers
+    kClean,
+    kDirtyRootLayers,
+    kDirtySessionLayers
 };
 
 /**
--- a/lib/usdUfe/utils/layers.h
+++ b/lib/usdUfe/utils/layers.h
@@ -149,6 +149,23 @@ bool isSessionLayer(
     const PXR_NS::SdfLayerHandle&           layer,
     const std::set<PXR_NS::SdfLayerRefPtr>& sessionLayers);
 
+// Verify if a stage has any dirty layer.
+enum class StageDirtyState : int
+{
+    Clean,
+    DirtyRootLayers,
+    DirtySessionLayers
+};
+
+/**
+ * Verify if a stage is dirty.
+ *
+ * @param stage the stage to verify.
+ */
+
+USDUFE_PUBLIC
+StageDirtyState isStageDirty(const PXR_NS::UsdStage& stage);
+
 /**
  * Get which of the two given layers is the strongest under the given root layer hierarchy.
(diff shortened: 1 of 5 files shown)


test/lib/usd/translators/UsdExportStagesAsRefsTest/sceneWithStage.ma
```diff
@@ -2,12 +2,13 @@
...
 //Last modified: Wed, Jun 26, 2024 04:41:34 PM
 //Codeset: 1252
+requires maya "2019ff01";
 requires -nodeType "mayaUsdLayerManager" -nodeType "mayaUsdProxyShape" -dataType "pxrUsdStageData"
 		 "mayaUsdPlugin" "0.30.0";
 currentUnit -l centimeter -a degree -t film;
 fileInfo "application" "maya";
-fileInfo "product" "Maya 2022";
-fileInfo "version" "Preview Release 123";
+fileInfo "product" "Maya 2019";
```
seando-adsk, line 10: Is there a reason for this change? We don't support Maya 2019 anymore.

lib/usdUfe/utils/layers.h
```diff
@@ -149,6 +149,23 @@ bool isSessionLayer(
     const PXR_NS::SdfLayerHandle&           layer,
     const std::set<PXR_NS::SdfLayerRefPtr>& sessionLayers);
 
+// Verify if a stage has any dirty layer.
+enum class StageDirtyState : int
+{
+    Clean,
+    DirtyRootLayers,
+    DirtySessionLayers
```
seando-adsk, line 157: According to our coding guidelines enum names should be named like constants and start with `k`

test/lib/usd/translators/UsdExportStagesAsRefsTest/sceneWithStage.ma
pierrebai-adsk: There was a "requires" missing for the Maya version, above this. I took the requires line from another file we had in our unit tests and I thought it was better if the "requires" matched the fileInfo data, just in case it matters.

lib/usdUfe/utils/layers.h
pierrebai-adsk: Dang! I looked at another enum for guidance and it lacked the k. I'll add them.

test/lib/usd/translators/UsdExportStagesAsRefsTest/sceneWithStage.ma
seando-adsk: I think it would be better to leave it at 2022 since the original file was. This should work:
```
requires maya "2022";
fileInfo "product" "Maya 2022";
fileInfo "version" "2022";
```

seando-adsk (approved):

samuelliu-adsk (approved):

PR #3984, opened 2024-10-31
EMSUSD-1722 fix node origin detection

pierrebai-adsk:
- Pass the proxy node to the layer manager so it can determine which layer manager corresponds to the proxy node.
- We keep track of the association by connecting a new proxy shape attribute to the layer manager so that we can traverse the connection to find the layer manager.
- Fix a crash in MayaSessionState when the proxy shape has been deleted before the on-idle callback is run.
- This can happen in scripts that creates scene, a stage and clear the scene in one go, before an on-idle callback is run.
- Make the test more similar to the script in the ticket.

### DIFF
--- a/lib/mayaUsd/nodes/layerManager.cpp
+++ b/lib/mayaUsd/nodes/layerManager.cpp
@@ -114,20 +114,23 @@ MStatus disconnectCompoundArrayPlug(MPlug arrayPlug)
     return dgmod.doIt();
 }
 
-/// @brief Verify if the given node is either from the given reference, if that is not null,
-///        or from the main Maya scene if the reference is null.
+/// @brief Verify if the given node is from a reference.
 static bool
-isNodeFromDesiredOrigin(const MFnReference* fromReference, const MFnDependencyNode& node)
+isNodeFromDesiredOrigin(const MFnDependencyNode& node, MayaUsdProxyShapeBase* forProxyShape)
 {
-    if (fromReference) {
-        return fromReference->containsNodeExactly(node.object());
-    } else {
-        return !node.isFromReferencedFile();
-    }
+    const bool proxyIsFromReference
+        = (forProxyShape && MFnDependencyNode(forProxyShape->thisMObject()).isFromReferencedFile());
+    return node.isFromReferencedFile() == proxyIsFromReference;
 }
 
-MayaUsd::LayerManager* findNode(const MFnReference* fromReference = nullptr)
+MayaUsd::LayerManager* findNode(MayaUsdProxyShapeBase* forProxyShape)
 {
+    if (forProxyShape) {
+        if (MayaUsd::LayerManager* layerManager = forProxyShape->getLayerManager()) {
+            return layerManager;
+        }
+    }
+
     // Check for cached layer manager before searching
     MFnDependencyNode fn;
     if (layerManagerHandle.isValid() && layerManagerHandle.isAlive()) {
@@ -143,7 +146,7 @@ MayaUsd::LayerManager* findNode(const MFnReference* fromReference = nullptr)
         MObject mobj = iter.item();
         fn.setObject(mobj);
         if (fn.typeId() == MayaUsd::LayerManager::typeId) {
-            if (isNodeFromDesiredOrigin(fromReference, fn)) {
+            if (isNodeFromDesiredOrigin(fn, forProxyShape)) {
                 layerManagerHandle = mobj;
                 return static_cast<MayaUsd::LayerManager*>(fn.userNode());
             }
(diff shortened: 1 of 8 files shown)


lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -114,20 +114,23 @@ MStatus disconnectCompoundArrayPlug(MPlug arrayPlug)
...
 
-MayaUsd::LayerManager* findNode(const MFnReference* fromReference = nullptr)
+MayaUsd::LayerManager* findNode(MayaUsdProxyShapeBase* forProxyShape)
 {
+    if (forProxyShape) {
+        if (MayaUsd::LayerManager* layerManager = forProxyShape->getLayerManager()) {
+            return layerManager;
+        }
+    }
+
```
pierrebai-adsk, line 133: This is the main change: the proxy shape node is now able to know which layer manager node to use. Except for existing Mays scene files until they're saved again, so we need the backup option of trying to find the node ourselves. We also need that backup plan in case the previous save had any problem setting the new proxy shape attribute.

lib/mayaUsd/nodes/layerManager.cpp
```diff
@@ -143,7 +146,7 @@ MayaUsd::LayerManager* findNode(const MFnReference* fromReference = nullptr)
         MObject mobj = iter.item();
         fn.setObject(mobj);
         if (fn.typeId() == MayaUsd::LayerManager::typeId) {
-            if (isNodeFromDesiredOrigin(fromReference, fn)) {
+            if (isNodeFromDesiredOrigin(fn, forProxyShape)) {
```
pierrebai-adsk, line 149: All other changes are just passing the proxy shape node around.

lib/mayaUsd/nodes/proxyShapeBase.cpp
```diff
@@ -479,6 +481,19 @@ MStatus MayaUsdProxyShapeBase::initialize()
...
+    typedAttrFn.setWritable(true);
+    typedAttrFn.setReadable(true);
+    typedAttrFn.setInternal(true);
+    typedAttrFn.setHidden(true);
+    typedAttrFn.setDisconnectBehavior(MFnNumericAttribute::kReset); // on disconnect, reset to Null
+    typedAttrFn.setAffectsAppearance(false);
+    CHECK_MSTATUS_AND_RETURN_IT(retValue);
+    retValue = addAttribute(layerManagerAttr);
+    CHECK_MSTATUS_AND_RETURN_IT(retValue);
+
```
pierrebai-adsk, line 496: The new proxy shape attribute will keep the layer manager node UUID as a string.

lib/usd/ui/layerEditor/mayaSessionState.cpp
```diff
@@ -223,10 +224,17 @@ void MayaSessionState::processNodeAdded(MObject& node)
...
+
     // doing it on idle give time to the Load Stage to set a file name
     // but we don't do a second idle because we could get a delete right after a Add
-    MDagPath dagPath;
-    MFnDagNode(obj).getPath(dagPath);
+    MDagPath   dagPath;
+    MFnDagNode dagNode;
+    if (!dagNode.setObject(obj))
+        return;
+
```
pierrebai-adsk, line 236: While trying to fix the bug in various ways, I encountered a crash in Maya when calling getPath on a MDagNode that contained a null MObject.

In this case, it can happen when a new Maya scene is create, a stage added to it, then the scene delete before any on-idle task are processed. This can happen when writing scripts.

test/lib/testMayaUsdCreateStageInMayaRef.py
```diff
@@ -52,52 +93,23 @@ def testCreateStageInMayaRef(self):
...
-        cmds.optionVar(iv=("mayaUsd_SerializedUsdEditsLocation", USD_EXPORT_SETTING))
-        om.MFileIO.saveAs(maya_with_usd_ref, "mayaAscii")
-
-        # Make sure we don't retain references to USD data.
-        stage = None
-        ref_prim = None
+        usd_cube_file = createCubeUsdFile(testDir)
+        proxy_shape, usd_cube_path, maya_with_usd_ref = createMayaSceneWithStageWithUsdRef(testDir, usd_cube_file)
+        namespace = "my_ref"
+        createMayaSceneWithMayaRef(testDir, maya_with_usd_ref, namespace)
```
pierrebai-adsk, line 99: Moved test code in functions, mostly so that I could be sure I was running the same code in the unit test as in the example script given in the bug ticket.

barbalt (approved):

PR #3986, opened 2024-11-04
EMSUSD-1686 Camera Performance Fix Backport

AramAzhari-adsk:
This change adds support for a find / findAll methods in Camera Handler that is compatible with an earlier version of Ufe.
This also addresses a performance issue with Maya manipulators (TRS).


lib/mayaUsd/ufe/CMakeLists.txt
```diff
@@ -235,12 +248,17 @@ if(CodeWrapperHandler IN_LIST UFE_PREVIEW_FEATURES)
     )
 endif()
 
+if(UFE_CAMERAHANDLER_HAS_FINDALL)
+    list(APPEND HEADERS
+        ProxyShapeCameraHandler.h
```
AramAzhari-adsk, line 253: This seems to be causing an issue for Maya 2023 platform when cameraHandler_v3_4::findAll isn't present.

`Global.cpp(74): fatal error C1083: Cannot open include file: 'mayaUsd/ufe/ProxyShapeCameraHandler.h': No such file or directory`

AramAzhari-adsk:
The change was compiled locally with Maya 2023 with and without the UFE update and successfully compiled.

PR #3987, opened 2024-11-06
EMSUSD-1570 add a class-prims filter to USD UFE

pierrebai-adsk:
- Add a getUsdPredicate helper function to UsdUfe to create the USD predicate corresponding to a list of UFE filters.
- Add the "Class Prims" filter to the list of USD filters in the hierarchy handler.
- Use the helper to create the USD predicate.
- Rely on the USD predicate having a correct predicate for inactive prim instead of using an explicit flag.
- Adjust the context menu for class prims.
- Add a unit test for the class prims filter.

Work around the problem that Maya does not correctly handle UFE filters. Maya only handles one UFE filter. If there are more than one filters, only one is handled.

To work around this issue, we manually ask the outliner what are the desired settings and use these settings as the default values for the filters. When the outliner fails to handle multiple filters, it will end-up using the default values, which will be the desired values.

### DIFF
--- a/lib/mayaUsd/ufe/MayaUsdHierarchyHandler.cpp
+++ b/lib/mayaUsd/ufe/MayaUsdHierarchyHandler.cpp
@@ -22,6 +22,9 @@
 #include <usdUfe/ufe/UsdSceneItem.h>
 #include <usdUfe/ufe/Utils.h>
 
+#include <maya/MGlobal.h>
+#include <maya/MString.h>
+
 namespace MAYAUSD_NS_DEF {
 namespace ufe {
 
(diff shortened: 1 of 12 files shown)


lib/usdUfe/ufe/Utils.h
```diff
@@ -477,6 +479,12 @@ void removeSessionLeftOvers(
     UsdUndoableItem*              undoableItem,
     bool                          extraEdits = true);
 
+//! Return the USD prims predicate for the given UFE child filter.
+//! Note: an empty filter or unknown filter will filter out everything.
+//!       Yes, that means no-filter actually means filter everything out.
+USDUFE_PUBLIC
+PXR_NS::Usd_PrimFlagsPredicate getUsdPredicate(const Ufe::Hierarchy::ChildFilter& childFilter);
```
AramAzhari-adsk, line 486: Looking at the comment, as long as UsdHierarchyHandler::childFilter() is accessible and will always include-all I think this should be okay. May be there's a way to guarantee include-all with a function without having to repeat the emplace_back() of filters in various code locations (if it comes to that).

test/lib/ufe/testChildFilter.py
```diff
@@ -116,6 +118,60 @@ def testFilteredChildren(self):
...
+        activeCubeItem = ufe.Hierarchy.createItem(ufe.PathString.path(topPathStr + '/active_cube'))
+        inactiveCylinderItem = ufe.Hierarchy.createItem(ufe.PathString.path(topPathStr + '/inactive_cylinder'))
+        activeClassItem = ufe.Hierarchy.createItem(ufe.PathString.path(topPathStr + '/active_class'))
+        inactiveClassItem = ufe.Hierarchy.createItem(ufe.PathString.path(topPathStr + '/inactive_class'))
+
+        # Declare different filter settings that will be tested.
+        FilterSettings = collections.namedtuple('FilterSettings', ['filters', 'expecteditems'])
+        filterSettings = [
+            FilterSettings(
+                filters={ 'InactivePrims': False, 'ClassPrims': False },
```
AramAzhari-adsk, line 150: This is my thinking, that a full list of possibilities may not be known in case there's a need to include all.
Perhaps we can document the supported filters somewhere to be used as a reference.

AramAzhari-adsk (approved):

lib/mayaUsd/ufe/MayaUsdHierarchyHandler.cpp
```diff
@@ -50,5 +53,90 @@ Ufe::Hierarchy::Ptr MayaUsdHierarchyHandler::hierarchy(const Ufe::SceneItem::Ptr
...
+    { "InactivePrims", true },
+    { "ClassPrims", false },
+};
+
+static std::atomic<bool> hasPendingFilterDefaultsUpdate = false;
+
+static void updateFilterDefaults(void*)
+{
+    // This script finds a valid outiner panel that has valid UFE filters
+    // settings.
```
seando-adsk, line 68: What happens if there are multiple Outliners open? How do you return the correct result for the Outliner you are being called from?

lib/usdUfe/ufe/UsdHierarchy.cpp
```diff
@@ -189,18 +189,8 @@ Ufe::SceneItemList UsdHierarchy::children() const
...
-        const bool             showInactive = childFilter.front().value;
-        Usd_PrimFlagsPredicate flags
-            = showInactive ? UsdPrimIsDefined && !UsdPrimIsAbstract : kUsdUfePrimDefaultPredicate;
-        return createUFEChildList(getUSDFilteredChildren(_item, flags), !showInactive);
-    }
-
-    UFE_LOG("Unknown child filter");
-    return Ufe::SceneItemList();
+    Usd_PrimFlagsPredicate flags = UsdUfe::getUsdPredicate(childFilter);
+    return createUFEChildList(getUSDFilteredChildren(_item, flags), false);
```
seando-adsk, line 193: You are now always passing `false` here when before the code used to have a boolean. How is that correctly filtering the inactive now?

seando-adsk:
@pierrebai-adsk Removed the ready-for-merge just until my questions are answered.

lib/usdUfe/ufe/UsdHierarchy.cpp
pierrebai-adsk: They are filtered by the USD predicate. We never needed to pass true here, unlike inthe chidlren() function, which having no USD predicate, need the flag to filter inactive prims.

lib/mayaUsd/ufe/MayaUsdHierarchyHandler.cpp
pierrebai-adsk: The children filter value kept by the outliner are in a static variable, shared by all outliners. There is only one set of value, so it does not matter which outliner we use. I only need to go through the list of them all because Maya creates a second hidden outliner on start-up, which is invalid until first shown but returned as part of the list of outliner panels.

pierrebai-adsk: After further testing, turns out I had tested with local Maya 2025 build with the fix for multiple-filter in it, so this script does not actually fully solve the problem. The problem is when loading a Maya scene, the filters are not restored correctly in Maya 2025. It might turn out that we can't support class prims filter in Maya 2025...

lib/usdUfe/ufe/UsdHierarchy.cpp
pierrebai-adsk: When I edit as Maya, I see the Maya node under the USD node, replacing the USD node, if that is what you meant.

lib/mayaUsd/ufe/MayaUsdHierarchyHandler.cpp
pierrebai-adsk: I fixed that problem by not calling the script on-idle but calling it right away.

pierrebai-adsk:
I believe I answered Sean's concerns, so I'll mark this a ready for merge.

lib/usdUfe/ufe/UsdHierarchy.cpp
seando-adsk: I don't recall exactly what that comment was referring to, but thanks for verifying by testing.

seando-adsk (approved):

PR #3992, opened 2024-11-08
EMSUSD-1770 - Initial support for USD v24.11

seando-adsk:
#### EMSUSD-1770 - Initial support for USD v24.11
* USD ver > v24.11: no longer need boost linking cmake flags.
* New support to promote a "top-level" header.
* Updated coding guidelines regarding Boost and Boost python.
* Pixar has removed Boost dependency from USD v24.11, but MayaUsd still uses boost (header only).
* Copied (from Pixar USD) new code to wrap static token.
* Fixed bug in "testShaderNodeDef" revealed by USD v24.11


.clang-format
```diff
@@ -62,7 +62,7 @@ IncludeCategories:
     Priority:        4
 
   # 5. Pixar + USD headers
-  - Regex:           '^<pxr/'
+  - Regex:           '(pxr_python\.h|^<pxr/)'
```
seando-adsk, line 65: Sort this new top-level header in with the other pxr headers. I would have preferred if it was at the top of the group, but clang-format puts it at the bottom.

CMakeLists.txt
```diff
@@ -226,6 +226,18 @@ if(BUILD_MAYAUSD_LIBRARY)
...
+if(USD_VERSION VERSION_GREATER_EQUAL "0.24.11")
+    # In USD v24.11 Pixar USD has completely removed Boost.
+    # However MayaUsd is still using a few of the Boost components, so
+    # when using USD v24.11 (or greater) we need to find Boost ourselves
+    # (from the USD build, since Boost is still required for OpenImageIO).
+    if(NOT DEFINED ENV{BOOST_ROOT})
+        set(ENV{BOOST_ROOT} ${PXR_USD_LOCATION})
+    endif()
+
+    find_package(Boost REQUIRED COMPONENTS headers)
```
seando-adsk, line 238: With USD v24.11 we only need boost headers which we must find ourselves, since USD no longer has Boost as a dependency (and thus not in the USD cmake files in our USD artifact).

cmake/utils.cmake
```diff
@@ -192,7 +194,9 @@ function(mayaUsd_promoteHeaderList)
 
     set(BASEDIR ${CMAKE_BINARY_DIR}/include)
     if (PREFIX_BASEDIR)
-        set(BASEDIR ${BASEDIR}/${PREFIX_BASEDIR})
+        if(NOT ${PREFIX_BASEDIR} MATCHES "NONE")
```
seando-adsk, line 197: Allow promoting at the very top-level (not in subdir).

lib/mayaUsd/CMakeLists.txt
```diff
@@ -17,25 +17,13 @@ project(mayaUsd)
...
 
-#------------------------------------------------------------------------------
-# modules and definitions
-#------------------------------------------------------------------------------
-if(NOT DEFINED ENV{BOOST_ROOT})
-    set(ENV{BOOST_ROOT} ${PXR_USD_LOCATION})
-endif()
-if(MAYAUSD_DEFINE_BOOST_DEBUG_PYTHON_FLAG)
-    set(Boost_USE_DEBUG_PYTHON ON)
-endif()
```
seando-adsk, line 28: These aren't need as we don't call "find_package(Boost)".

lib/mayaUsd/CMakeLists.txt
```diff
@@ -259,6 +255,7 @@ target_link_libraries(${PROJECT_NAME}
         ${UFE_LIBRARY}
         ${MAYA_LIBRARIES}
         usdUfe
+        $<$<BOOL:$<VERSION_GREATER_EQUAL:${USD_VERSION},0.24.11>>:Boost::headers>
```
seando-adsk, line 258: Since USD no longer has Boost, we need to add a Boost dependency so cmake will add the boost include path (for the Boost headers we still use).

lib/mayaUsd/fileio/registryHelper.cpp
```diff
@@ -337,7 +336,7 @@ void UsdMaya_RegistryHelper::AddUnloader(const std::function<void()>& func, bool
 {
     if (fromPython) {
         g_pythonUnloaders.emplace_back(func);
-        if (boost::python::import("atexit")
+        if (PXR_BOOST_PYTHON_NAMESPACE::import("atexit")
```
seando-adsk, line 339: You'll see a ton of changes like this. Removing any `<boost/python/xxx>` includes replacing with new `<pxr_python.h>` and using define for namespace. This handles pre 24.11 and >= 24.11

lib/usd/schemas/wrapTokens.cpp
```diff
@@ -13,52 +13,24 @@
...
-        boost::python::make_function(
-            _WrapStaticToken(&token),
-            boost::python::return_value_policy<boost::python::return_by_value>(),
-            boost::mpl::vector1<std::string>()));
-}
-
-} // namespace
+#define _ADD_TOKEN(cls, name) \
+    cls.add_static_property(  \
+        #name, +[]() { return MayaUsd_SchemasTokens->name.GetString(); });
```
seando-adsk, line 24: Copied new method (from Pixar) for wrapping static tokens. This removes the need for Boost::mpl

lib/usdUfe/ufe/UsdShaderAttributeDef.cpp
```diff
@@ -180,7 +180,8 @@ Ufe::Value UsdShaderAttributeDef::getMetadata(const std::string& key) const
         }
 #else
         const auto sdfTypeIndicator = _shaderAttributeDef->GetTypeAsSdfType();
-        if (sdfTypeIndicator.HasSdfType()) {
+        if (sdfTypeIndicator.HasSdfType()
+            && sdfTypeIndicator.GetSdfType() != SdfValueTypeNames->Token) {
```
seando-adsk, line 184: Fix from Jerry to fix test affected by update to 24.11

plugin/al/lib/AL_USDMaya/AL/usdmaya/nodes/wrapLayerManager.cpp
```diff
@@ -15,32 +15,17 @@
...
-#include <boost/python/def.hpp>
-
-// On Windows, against certain versions of Maya and with strict compiler
-// settings on, we are getting warning-as-error problems with a couple
-// boost includes.  Disabling those warnings for the specific includes
-// for now instead of disabling the strict settings at a higher level.
-#ifdef _MSC_VER
-#pragma warning(push)
-#pragma warning(disable : 4275)
-#endif
```
seando-adsk, line 28: I'm hoping this is no longer needed. Other files include this header without problems.

plugin/al/schemas/AL/usd/schemas/maya/CMakeLists.txt
```diff
@@ -129,6 +129,11 @@ target_compile_definitions(${PYTHON_LIBRARY_NAME}
         MFB_PACKAGE_MODULE=AL.usd.schemas.maya
 )
 
+target_include_directories(${PYTHON_LIBRARY_NAME} 
+  PRIVATE
+    ${CMAKE_BINARY_DIR}/include
+)
```
seando-adsk, line 135: For the top-level pxr_python.h

pxr_python.h
```diff
@@ -0,0 +1,57 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#ifndef USDUFE_PYTHON_H
+#define USDUFE_PYTHON_H
+
+// clang-format off
```
seando-adsk, line 20: Turned off, so the header includes stay in the order I've set below.

pxr_python.h
```diff
@@ -0,0 +1,57 @@
...
+
+//
+// Helper header to deal with USD removing boost python and replacing
+// it with their own version.
+//
+#include <pxr/pxr.h>
+
+#ifdef PXR_USE_INTERNAL_BOOST_PYTHON // PXR_VERSION >= 2411
+
+#include <pxr/external/boost/python.hpp>
```
seando-adsk, line 30: This includes most of the python files needed, except the two extra below.

pxr_python.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include <boost/python.hpp>
+
+#include <boost/noncopyable.hpp>
+
+#include <boost/python/suite/indexing/map_indexing_suite.hpp>
+#include <boost/python/wrapper.hpp>
+
+#define PXR_BOOST_NAMESPACE        boost
+#define PXR_BOOST_PYTHON_NAMESPACE boost::python
```
seando-adsk, line 45: Define the same namespace names as the >= 24.11 case so we can use this define in all USD versions.

pxr_python.h
```diff
@@ -0,0 +1,57 @@
...
+
+namespace PXR_BOOST_NAMESPACE { namespace python {
+    using boost::noncopyable;
+}} // namespace PXR_BOOST_NAMESPACE::python
```
seando-adsk, line 49: Put into boost::python namespace, so we can use it from single namespace. Pixar has done this in their external boost/python.

AramAzhari-adsk (approved):

deboisj (approved):

PR #4001, opened 2024-11-15
Fix issue with multiple filters

JGamache-autodesk:


pierrebai-adsk:
The issue is already fixed by this PR: https://github.com/Autodesk/maya-usd/pull/3987

(Which is the true change that was meant to go with the Maya change.)

JGamache-autodesk:
OK. Thanks! Withdrawing this PR then.

PR #4006, opened 2024-11-19
Shader parameter sections should be expanded

JGamache-autodesk:


JGamache-autodesk:
Logged internally as EMSUSD-1826.

pierrebai-adsk (approved):

PR #4007, opened 2024-11-20
EMSUSD-1625 - Certain USD light shapes disappear if user try to edit as maya data.

samuelliu-adsk:


lib/usd/translators/lightWriter.cpp
```diff
@@ -222,4 +223,114 @@ void PxrUsdTranslators_AreaLightWriter::Write(const UsdTimeCode& usdTime)
...
+    : UsdMayaPrimWriter(depNodeFn, usdPath, jobCtx)
+{
+    MPlug   plug = depNodeFn.findPlug("lightShape");
+    VtValue lightShape
+        = UsdMayaWriteUtil::GetVtValue(plug, MayaUsd::Converter::getUsdTypeName(plug), false);
+    if (lightShape.IsEmpty()) {
+        return;
+    }
+
+    if (lightShape == 1) {
```
samuelliu-adsk, line 242: Note sure if we use an enum than the number 1 is better. It is an arbitrary number either way...

lib/usd/translators/lightWriter.cpp
```diff
@@ -222,4 +223,114 @@ void PxrUsdTranslators_AreaLightWriter::Write(const UsdTimeCode& usdTime)
         usdTime, primSchema, lightFn, _GetSparseValueWriter());
 }
 
+#ifdef UFE_VOLUME_LIGHTS_SUPPORT
+PXRUSDMAYA_REGISTER_WRITER(volumeLight, PxrUsdTranslators_VolumeLightWriter);
```
samuelliu-adsk, line 227: One major problem I was struggling is here. As shown on L179, we should register a schema here but one Maya light type can only map to one usd type. Here there will be three possible schemas to register for volumeLight so I didn't register at the time. After playing around I dont see a major issue caused by this but I think we better address this.

AramAzhari-adsk (changes requested):
As discussed, please check that:
- Each of the recently added lights gizmos actually shade similarly when they are a USD light and when they are being edited as a Maya light.
- If any of them do not have a matching light shading, please block / remove the option of being able to Edit As Maya for them.

PR #4008, opened 2024-11-21
EMSUSD-1771 remove schema from prims

pierrebai-adsk:
Add functions to support schema removal in UsdUfe:
- Add removeSchemaFromPrim helper function to remove a single-apply schema from a prim.
- Add removeMultiSchemaFromPrim helper function to remove a single-apply schema from a prim.
- Expose those functions to Python.
- Add unit test for the functions.

Remove schema command:
- Add a -rem (-removeSchema) flag to the schema command.
- Implement schema removal in the command.
- Update the documentation about the command.
- Add unit tests for the removal of schemas via the command.

Remove schema menu items:
- Add a "Remove Schema" menu item in the "Attributes" menu of the AE.
- Add each applied schema of the selected prims to the menu.
- Echo to the user the schema commands invoked by the menu items.


plugin/adsk/scripts/mayaUsdMenu.py
```diff
@@ -167,4 +212,108 @@ def aeAttributesMenuCallback(aeAttributeMenu):
...
+#
+# Remove schema
+
+def _removeSchemaMenuItemCallback(prettyTypeName, schemaTypeName, instanceName, menuItem):
+    '''
+    Remove the given schema from the prims in the current selection.
+    '''
+    primPaths = _getUsdPathsInSelection()
+    if not primPaths:
+        print('Cannot remove schema "%s": no USD prim are currently selected.' % schemaTypeName)
```
seando-adsk, line 228: Two questions. (1) should this string be in resource and (2) should use `cmds.warning` instead of print?

seando-adsk (commented):
Looks good. Just one minor comment about string.

seando-adsk (approved):

PR #4011, opened 2024-11-25
EMSUSD-1625 - Disable EditAsMaya for cylinder, disk and dome light

samuelliu-adsk:
As discussed, disable the editAsMaya for now instead of writing light reader/writers

### DIFF
--- a/lib/usd/translators/lightReader.cpp
+++ b/lib/usd/translators/lightReader.cpp
@@ -26,6 +26,12 @@ TF_DEFINE_ENV_SETTING(
     false,
     "Whether to import UsdLux as Renderman-for-Maya lights.");
 
+// Unregister the light readers for cylinder, disk and dome lights
+// As for now those lights are not supported by import/export to Maya
+// This will also disable edit as maya
+// User can register a custom light reader for those lights if needed
+
+/*
 PXRUSDMAYA_DEFINE_READER(UsdLuxCylinderLight, args, context)
 {
     if (TfGetEnvSetting(MAYAUSD_IMPORT_RFM_LIGHTS)) {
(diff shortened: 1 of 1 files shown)


lib/usd/translators/lightReader.cpp
```diff
@@ -42,20 +48,21 @@ PXRUSDMAYA_DEFINE_READER(UsdLuxDiskLight, args, context)
     return false;
 }
 
-PXRUSDMAYA_DEFINE_READER(UsdLuxDistantLight, args, context)
+PXRUSDMAYA_DEFINE_READER(UsdLuxDomeLight, args, context)
```
AramAzhari-adsk, line 51: Are the definitions swapped? it seems that usdluxdome light was after distant light and now it is before it. The diff is getting confused.

lib/usd/translators/lightReader.cpp
```diff
@@ -42,20 +48,21 @@ PXRUSDMAYA_DEFINE_READER(UsdLuxDiskLight, args, context)
...
 {
     if (TfGetEnvSetting(MAYAUSD_IMPORT_RFM_LIGHTS)) {
         return UsdMayaTranslatorRfMLight::Read(args, context);
     }
-    return UsdMayaTranslatorLight::Read(args, context);
+    return false;
 }
+*/
```
AramAzhari-adsk, line 58: Is this `*/` supposed to be here?

lib/usd/translators/lightReader.cpp
samuelliu-adsk: It is just git diff getting confused.. I moved the dome light up so they can all be commented out together and that's it

AramAzhari-adsk (approved):

PR #4014, opened 2024-11-27
EMSUSD-1522 Light Linking member expression and expansion menu

barbalt:
- Added tabs to the Light Linking widget
- Added Expression tabs
- Added Expansion rules menu

image

image

image


lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/resources.py
```diff
@@ -0,0 +1,60 @@
+# Resource object code (Python 3)
```
barbalt, line 1: this file was generated by compiling the resources.qrc file using the command:
`pyside6-rcc .\resources.qrc -o resources.py`

PR #4015, opened 2024-11-27
EMSUSD-1522 I want the ability to see and set the expansion rule of the collection

barbalt:
- Added tabs to the Light Linking widget
- Added Expression tabs
- Added Expansion rules menu

image
image
image


lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/ExpressionWidget.py
```diff
@@ -0,0 +1,74 @@
...
+        usdExpressionAttr = self._collection.GetMembershipExpressionAttr().Get()
+        usdExpression = ""
+        if usdExpressionAttr != None:
+            usdExpression = usdExpressionAttr.GetText()
+
+        textExpression = self._expressionText.toPlainText()
+        if usdExpression != textExpression:
+            # assign default value if text is empty
+            if textExpression == "":
+                self._collection.CreateMembershipExpressionAttr()
```
pierrebai-adsk, line 59: I would have thought that maybe when the expression is empty we delete the expression? I guess I would need to go see how it was specified in the design.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/ExpressionWidget.py
```diff
@@ -0,0 +1,74 @@
...
+            else:
+                self._collection.CreateMembershipExpressionAttr(Sdf.PathExpression(textExpression))
+
+            if self._expressionCallback != None:
+                self._expressionCallback()
+
+    def eventFilter(self, obj, event):
```
pierrebai-adsk, line 66: Are there not Qt signal for these things (pressing enter, losing focus) that could be connected instead of using event filter?

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/IncludeExcludeWidget.py
```diff
@@ -0,0 +1,77 @@
+from ..common.list import StringList
+from ..common.resizable import Resizable
+from ..common.kebabMenu import KebabMenu
+from .ExpressionRulesMenu import ExpressionMenu
```
pierrebai-adsk, line 4: I guess it is too late because I know git is finnicky about these things, but I wish the casing of files was more consistent. Now we have some with initial uppercase and some with initial lowercase.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/IncludeExcludeWidget.py
```diff
@@ -0,0 +1,77 @@
...
+
+        headerWidget = QWidget(self)
+        headerLayout = QHBoxLayout(headerWidget)
+        headerLayout.addStretch(1)
+        # Will need a separator for future designs
+        #headerLayout.addWidget(separator)
+        headerLayout.addWidget(menuButton)
+        includeExcludeLayout.addWidget(headerWidget)
+
+        self._include = StringList(includes, "Include", "Include all", self)
```
pierrebai-adsk, line 49: I would have like the widget class to have names that reflect they are UI element. StringList and Resizable are very generic, at first I did not understand what was going on.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/list.py
```diff
@@ -13,7 +13,7 @@
     from PySide6.QtGui import QPainter
     from PySide6.QtWidgets import QStyleOptionViewItem, QStyledItemDelegate, QListView, QLabel, QVBoxLayout, QHBoxLayout, QWidget, QCheckBox
 except:
-    from PySide2.QtCore import (
+    from PySide2.QtCore import ( # type: ignore
```
pierrebai-adsk, line 16: What are these type: ignore comment?

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/IncludeExcludeWidget.py
barbalt: That's a fair point. I can try to change it

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/list.py
barbalt: It's so that the editor doesn't complain about the types it doesn't know about

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/IncludeExcludeWidget.py
barbalt: Those were there on the initial version of this widget that was moved to maya. So I kept them with the same name for ease of updating

pierrebai-adsk (dismissed):

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/ExpressionWidget.py
barbalt: There isn't an specific event for pressing enter. We could follow the route of implementing the keypress event and checking if the key is enter and doing the same thing as above. 
There's a focusOut event which could be used, but since I already had the eventFilter for enter, I figured I would use that. Is it bad to use event filter?

pierrebai-adsk (approved):

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/persistentStorage.py
```diff
@@ -1,5 +1,5 @@
 try:
-    from PySide6.QtCore import QSettings
+    from PySide6.QtCore import QSettings  # type: ignore
 except:
     from PySide2.QtCore import QSettings  # type: ignore
```
pierrebai-adsk, line 5: I discovered that this writes files in a random folder. I think we will need to have some DCC-specific callback to save the data in DCC-specific locations. (In another PR)

pierrebai-adsk (approved):

PR #4023, opened 2024-11-29
EMSUSD-1856 Add filtering search to the collection widget

pierrebai-adsk:
- Add a FilteredStringListModel class.
- Use that model in the collection string list widgets.
- Add a text edit widget to enter the search text
- Connect the search changed signal to the filtered model.
- Add a button to clear the search filter.
- Add a flag on the filtered model to remember if it was entirely filtered out.
- Properly re-filter the model when a new list of text is given.
- Moved the filtered view in its own file.
- Give the resizable lists a sensible default size.

### DIFF
--- a/lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/resizable.py
+++ b/lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/resizable.py
@@ -103,6 +103,7 @@ class Resizable(QWidget):
         persistentStorageGroup: str = None,
         persistentStorageKey: str = None,
         parent: QWidget = None,
+        defaultSize = -1,
     ):
         super(Resizable, self).__init__(parent)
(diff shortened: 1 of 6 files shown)


lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/resizable.py
```diff
@@ -127,6 +128,8 @@ def __init__(
         self._widget: QWidget = None
 
         self.loadPersistentStorage()
+        if self._contentSize < 0 and defaultSize > 0:
+            self._contentSize = defaultSize
```
pierrebai-adsk, line 132: This avoids the lists to be very large by default. With the 80 passed in by the include/exclude panel, we get about 3 lines by default. Otherwise the lists a re way too big.

barbalt (approved):

PR #4030, opened 2024-12-05
EMSUSD-1541 - Drag and drop one or more scene objects from the outliner into the include or exclude list

samuelliu-adsk:


lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/includeExcludeWidget.py
```diff
@@ -87,4 +107,73 @@ def setIncludeAll(self, value: bool):
...
+    def _validatePath(self, collection, path):
+       
+        if not Sdf.Path.IsValidPathString(path):
+            raise ValueError("Invalid sdf path: " + path)
+
+        stage = self.control._collection.GetPrim().GetStage()
+        prim = stage.GetPrimAtPath(Sdf.Path(path))
+
+        if not prim or not prim.IsValid():
+            raise ValueError("Value must be a float or an int")("Error: The dragged object is not in the same stage as the collection. Ensure that objects belong to the same stage before adding them")
```
samuelliu-adsk, line 177: Not sure if there's a better way than ValueError in this case, trying to avoid using Maya or UFE specific methods...

barbalt: In other points of the code there's a 
`raise RuntimeError("Needs to be implemented in derived class")`
So I think the ValueError should be fine here

barbalt (approved):

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/filteredStringListView.py
```diff
@@ -60,6 +61,11 @@ def __init__(self, items: Sequence[str] = None, parent=None):
 
         self.selectionModel().selectionChanged.connect(lambda: self.selectedItemsChanged.emit())
 
+        self.placeholder_label = QLabel("Drag objects here or click “+” to add", self)
+        self.placeholder_label.setAlignment(Qt.AlignmentFlag.AlignCenter)
+        self.placeholder_label.setStyleSheet("color: gray; font-size: 18px;")
```
barbalt, line 66: I'm a bit worried about this specific color and pixel size here to work on max as well with different dpi monitors

barbalt:
I'll test those internally first

PR #4031, opened 2024-12-05
MAYA-133671 - Add textured mode handling to topo grapher

JGamache-autodesk:
Allows quickly reducing a graph to only a surface for when the viewport is in "untextured" mode.

### DIFF
--- a/lib/mayaUsd/render/MaterialXGenOgsXml/ShaderGenUtil.cpp
+++ b/lib/mayaUsd/render/MaterialXGenOgsXml/ShaderGenUtil.cpp
@@ -56,17 +56,26 @@ const std::string& TopoNeutralGraph::getMaterialName()
     return kMaterialName;
 }
 
-TopoNeutralGraph::TopoNeutralGraph(const mx::ElementPtr& material) { computeGraph(material); }
+TopoNeutralGraph::TopoNeutralGraph(const mx::ElementPtr& material) { computeGraph(material, true); }
 
 TopoNeutralGraph::TopoNeutralGraph(
     const mx::ElementPtr&  material,
     const LobePruner::Ptr& lobePruner)
 {
     _lobePruner = lobePruner;
-    computeGraph(material);
+    computeGraph(material, true);
 }
 
-void TopoNeutralGraph::computeGraph(const mx::ElementPtr& material)
+TopoNeutralGraph::TopoNeutralGraph(
+    const mx::ElementPtr&  material,
+    const LobePruner::Ptr& lobePruner,
+    bool                   textured)
+{
+    _lobePruner = lobePruner;
+    computeGraph(material, textured);
+}
+
+void TopoNeutralGraph::computeGraph(const mx::ElementPtr& material, bool textured)
 {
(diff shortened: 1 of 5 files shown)


lib/mayaUsd/render/MaterialXGenOgsXml/ShaderGenUtil.cpp
```diff
@@ -138,7 +147,16 @@ void TopoNeutralGraph::computeGraph(const mx::ElementPtr& material)
             if (!sourceInput) {
                 continue;
             }
-            auto              connectedNode = sourceInput->getConnectedNode();
+            auto connectedNode = sourceInput->getConnectedNode();
+
+            // In textured mode we traverse everything, but in untextured mode we only traverse PBR
+            // level connections.
+            static const auto isPbrType
+                = [](const auto& t) { return t == "BSDF" || t == "EDF" || t == "surfaceshader"; };
```
frohnej-adsk, line 155: [Nitpick] Can we compare against the string constants from `MaterialXCore/Types.h` instead? E.g.: `t == mx::BSDF_TYPE_STRING`.

[Question] Do we need to worry about volume-, light- and displacement shaders at all?

frohnej-adsk (dismissed):
LGTM! Just one minor comment.

lib/mayaUsd/render/MaterialXGenOgsXml/ShaderGenUtil.cpp
JGamache-autodesk: This code already leans heavily towards processing surface materials. I suspect it will never hit a lightshader connection.

frohnej-adsk (approved):

feldstj (approved):
LGTM, just one suggestion you can take or leave.

lib/mayaUsd/render/MaterialXGenOgsXml/ShaderGenUtil.cpp
```diff
@@ -56,17 +57,26 @@ const std::string& TopoNeutralGraph::getMaterialName()
...
+TopoNeutralGraph::TopoNeutralGraph(
+    const mx::ElementPtr&  material,
+    const LobePruner::Ptr& lobePruner,
+    bool                   textured)
+{
+    _lobePruner = lobePruner;
+    computeGraph(material, textured);
+}
+
+void TopoNeutralGraph::computeGraph(const mx::ElementPtr& material, bool textured)
```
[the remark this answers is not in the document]
JGamache-autodesk: And then the code of both functions end up identical except for three lines of code. I know of other C++ gurus that have a strict policy of using fully specified enums in this case. Especially useful when you start having calls like:
```
doSomething(true, false, true, true, false, false, false);
```
Sadly not something I can rework in the mayaUsdApi library as it needs to stay binary compatible. I can add constructors, but that is about it.

feldstj (approved):

JGamache-autodesk:
@seando-adsk ready for merge. Remaining preflight error is a known random failure of the testUsdExportPackage unit test.

PR #4033, opened 2024-12-05
EMSUSD-1912 disable new light test on USD 24.11

pierrebai-adsk:
For some reasons, USD 24.11 reports the root layer as being modified even though dumping the layer as string shows the data is not modified. Will need to be investigated further.


seando-adsk (approved):
Okay. Just don't forget to remove the print statement when you fix the test and reenable it.

PR #4036, opened 2024-12-05
Update change processing code

dj-mcg:
This handles the case where we receive a prim deleted message outside of a maya command, which results in the removal of a selected prim. In this case, we trigger a subtree invalidated notice so that the global selection list will be updated.

### DIFF
--- a/lib/usdUfe/ufe/StagesSubject.cpp
+++ b/lib/usdUfe/ufe/StagesSubject.cpp
@@ -13,10 +13,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-#include "StagesSubject.h"
-
 #include <usdUfe/base/tokens.h>
 #include <usdUfe/ufe/Global.h>
+#include <usdUfe/ufe/StagesSubject.h>
 #include <usdUfe/ufe/UfeNotifGuard.h>
 #include <usdUfe/ufe/UfeVersionCompat.h>
 #include <usdUfe/ufe/UsdCamera.h>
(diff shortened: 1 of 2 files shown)


pierrebai-adsk (approved):

lib/usdUfe/ufe/StagesSubject.cpp
```diff
@@ -528,6 +528,24 @@ void StagesSubject::stageChanged(
             Ufe::SceneItem::Ptr sceneItem = Ufe::Hierarchy::createItem(ufePath);
             if (!sceneItem || InAddOrDeleteOperation::inAddOrDeleteOperation()) {
                 sendObjectDestroyed(ufePath);
+
+                // If we are not in an add or delete operation, and a prim is
+                // removed, we need to trigger a subtree invalidation.  This is
+                // necessary in order to prevent stale items from being kept in
+                // the global selection set.
```
seando-adsk, line 535: I did a little testing on this trying to just remove the item from the selection list (since subtree invalidate can be quite destructive). That doesn't work since the prim is invalid (at this time) which means we cannot create a scene item for it. Can we add this to the comment here, something like:

```suggestion
                // If we are not in an add or delete operation, and a prim is
                // removed, we need to trigger a subtree invalidation.  This is
                // necessary in order to prevent stale items from being kept in
                // the global selection set.
                // Note: at the point of this notif the prim is not valid anymore
                // and thus we cannot create a scene item to simply remove
                // the item from selection list.
```

seando-adsk (approved):

PR #4037, opened 2024-12-05
Fix UsdMayaPrimWriter and UsdMayaPrimReader Python Wrappings

dj-mcg:
The PrimReaderWrapper<T> and PrimWriterWrapper<T> helper classes ignored the template parameter when deriving from TfPyPolymorphic. This caused unexpected duplicate entries in boost::python's type conversion table. boost::python silently accepted this, but the new pxr_boost::python in OpenUSD 24.11 flagged this with an assert in debug mode.

The wrappings for ShaderReaderWrapper and ShaderWriterWrapper incorrectly specified PrimReaderWrapper<> and PrimWriterWrapper<> as their bases. For example, this meant that passing an instance of a ShaderReader to a a Python-wrapped C++ function that takes a UsdMayaPrimReader would fail. This doesn't show up in practice since clients aren't expected to create their own instances of these types but it seemed good to fix anyway.

### DIFF
--- a/lib/mayaUsd/python/wrapPrimReader.cpp
+++ b/lib/mayaUsd/python/wrapPrimReader.cpp
@@ -38,7 +38,7 @@ PXR_NAMESPACE_USING_DIRECTIVE
 template <typename T = UsdMayaPrimReader>
 class PrimReaderWrapper
     : public T
-    , public TfPyPolymorphic<UsdMayaPrimReader>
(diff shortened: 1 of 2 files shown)


pierrebai-adsk (approved):

seando-adsk:
Preflight failed - could be random failure that I've seen in the past. Re-running the preflight to confirm.

PR #4040, opened 2024-12-09
EMSUSD-1692 - Adds LookdevXUsd Extension

AramAzhari-adsk:
This change adds USD support for LookdevX to allow compiling with different versions of USD.


plugin/adsk/plugin/CMakeLists.txt
```diff
@@ -100,6 +100,12 @@ if(IS_MACOSX OR IS_LINUX)
     )
 endif()
 
+# For initializing the handlers of lookdevXUsd
+target_link_libraries(${TARGET_NAME}
+    PRIVATE
+        lookdevXUsd
+)
```
seando-adsk, line 107: This should be wrapped in the cmake flag. We only need to link with it when its built.

seando-adsk: And you'll need to add a `target_compile_definitions` so you can use ifdef in the plugin.cpp below.

plugin/adsk/plugin/plugin.cpp
```diff
@@ -56,6 +56,7 @@
 #include <ufe/runTimeMgr.h>
 
 #include <basePxrUsdPreviewSurface/usdPreviewSurfacePlugin.h>
+#include <lookdevXUsd/LookdevXUsd.h>
```
seando-adsk, line 59: This line and the two calls below need to be wrapped in `#ifdef` and only compiled in when we built LookdevXUsd.

seando-adsk (changes requested):
A couple of minor changes are required.

AramAzhari-adsk:
Currently waiting on https://github.com/Autodesk/maya-usd/pull/4042 to be integrated into the system that the clang formatting will be check against.

test/lib/ufe/CMakeLists.txt
```diff
@@ -173,6 +173,7 @@ foreach(script ${TEST_SCRIPT_FILES})
             "UFE_SCENE_SEGMENT_HANDLER_ROOT_PATH=${UFE_SCENE_SEGMENT_HANDLER_ROOT_PATH}"
             "UFE_VOLUME_LIGHTS_SUPPORT=${UFE_VOLUME_LIGHTS_SUPPORT}"
             "UFE_SCENEITEM_HAS_METADATA=${UFE_SCENEITEM_HAS_METADATA}"
+            "BUILD_LOOKDEVXUSD_LIBRARY=${BUILD_LOOKDEVXUSD_LIBRARY}"
```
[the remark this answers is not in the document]
seando-adsk: Ah yes, sorry got them reversed.

frohnej-adsk (dismissed):
LGTM

seando-adsk (approved):
Excellent work thanks

PR #4043, opened 2024-12-10
Fix fatal error exit

pierrebai-adsk:
When aborting the process due to a fatal error (for example TF_CODING_ERROR or setting the environment variable TF_FATAL_VERIFY to 1), MayaUSD would get in a deadlock. Fix this by avoiding trying to unregister the diagnostic delegate when logging a fatal error.


AramAzhari-adsk (approved):
LGTM.

PR #4053, opened 2024-12-17
EMSUSD-1519 refactor collection UI

pierrebai-adsk:
Separate data from UI:
- Add data and usdData modules.
- Add collectionData and stringData to the data module as abstract interface to read and modify collection data.
- Add usdCollectionData and usdCollectionStringListData to usdData, for USD implementation.
- This also avoids business logic referring to UI labels and such.
- This also automatically updates the data source for the UI since the model is the data source.
- Allow setting the collection on the usdCollectionData.
- Fix the default value for the include-all flag.
- Still need to derive a Maya implementation to host picking and undo/redo.
- Make all UI use the collectionData and stringListData classes to hold and modify data instead of having business logic in UI.

Refactor UI code:
- Rename list.py to stringListPanel.py to avoid clashing with Python's list type.
- Removed all update and updateUI functions and flags and listen to dataChanged signals instead.
- Moved drag-and-drop event filter to string list view.
- Don't stop adding dropped item just because one is invalid.
- Move dropped item conversion and validation to the data class.

Small UI fixes:
- Fix selection mode of list views.
- Fixed issues with the check box callback getting the wrong state.
- Add include-all checkbox tooltip.
- Made the color and font size for the filtered list view come from the Theme.
- Removed place holder label and use paint event instead to both paint "no object found" and "drop stuff here".

### DIFF
--- a/lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/expressionWidget.py
+++ b/lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/expressionWidget.py
@@ -1,6 +1,7 @@
 from .expressionRulesMenu import ExpressionMenu
 from ..common.menuButton import MenuButton
 from ..common.theme import Theme
+from ..data.collectionData import CollectionData
 
 try:
     from PySide6.QtCore import QEvent, Qt  # type: ignore
@@ -12,16 +13,16 @@ except ImportError:
 from pxr import Usd, Sdf
 
 class ExpressionWidget(QWidget):
-    def __init__(self, collection: Usd.CollectionAPI, parent: QWidget, expressionChangedCallback):
+    def __init__(self, data: CollectionData, parent: QWidget, expressionChangedCallback):
         super(ExpressionWidget, self).__init__(parent)
-        self._collection = collection
+        self._collData = data
         self._expressionCallback = expressionChangedCallback
 
         mainLayout = QVBoxLayout(self)
         margin: int = Theme.instance().uiScaled(2)
         mainLayout.setSpacing(margin)
 
-        self._expressionMenu = ExpressionMenu(collection, self)
+        self._expressionMenu = ExpressionMenu(data, self)
         menuButton = MenuButton(self._expressionMenu, self)
 
         menuWidget = QWidget(self)
@@ -45,31 +46,20 @@ class ExpressionWidget(QWidget):
         mainLayout.addWidget(self._expressionText, 1)
 
         self._expressionText.installEventFilter(self)
-        self.update()
+        self.setLayout(mainLayout)
 
-    def update(self):
-        usdExpressionAttr = self._collection.GetMembershipExpressionAttr().Get()
+        self._collData.dataChanged.connect(self._onDataChanged)
+        self._onDataChanged()
+
+    def _onDataChanged(self):
+        usdExpressionAttr = self._collData.getMembershipExpression()
         if usdExpressionAttr != None:
             self._expressionText.setPlainText(usdExpressionAttr.GetText())
 
-        self._expressionMenu.update()
-
     def submitExpression(self):
-        usdExpressionAttr = self._collection.GetMembershipExpressionAttr().Get()
-        usdExpression = ""
-        if usdExpressionAttr != None:
-            usdExpression = usdExpressionAttr.GetText()
-
-        textExpression = self._expressionText.toPlainText()
-        if usdExpression != textExpression:
-            # assign default value if text is empty
-            if textExpression == "":
-                self._collection.CreateMembershipExpressionAttr()
-            else:
-                self._collection.CreateMembershipExpressionAttr(Sdf.PathExpression(textExpression))
-
-            if self._expressionCallback != None:
-                self._expressionCallback()
+        self._collData.setMembershipExpression(self._expressionText.toPlainText())
+        if self._expressionCallback != None:
+            self._expressionCallback()
 
     def eventFilter(self, obj, event):
         if event.type() == QEvent.KeyPress and obj is self._expressionText:
--- a/lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/includeExcludeWidget.py
+++ b/lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/includeExcludeWidget.py
@@ -117,11 +117,17 @@ class IncludeExcludeWidget(QWidget):
             self._include.cbIncludeAll.setChecked(incAll)
 
     def onAddToIncludePrimClicked(self):
-        items = self._collData.pick("Add Include Objects")
+        stage = self._collData.getStage()
+        if not stage:
+            return
+        items = Host.instance().pick(stage, "Add Include Objects")
         self._collData.getIncludeData().addStrings(items)
 
     def onAddToExcludePrimClicked(self):
-        items = self._collData.pick("Add Exclude Objects")
+        stage = self._collData.getStage()
+        if not stage:
+            return
+        items = Host.instance().pick(stage, "Add Exclude Objects")
         self._collData.getExcludeData().addStrings(items)
 
     def onRemoveSelectionFromInclude(self):
--- a/lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/includeExcludeWidget.py
+++ b/lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/includeExcludeWidget.py
@@ -16,7 +16,19 @@ except ImportError:
     from PySide2.QtWidgets import QFrame, QHBoxLayout, QVBoxLayout, QLineEdit, QMenu, QSizePolicy, QToolButton, QWidget  # type: ignore
 
 # TODO: support I8N
-kSearchPlaceHolder = "Search..."
+SEARCH_PLACEHOLDER_LABEL = "Search..."
+ADD_OBJECTS_TOOLTIP = "Add Objects to the Include/Exclude list"
+REMOVE_OBJECTS_TOOLTIP = "Remove Selected Objects from Include/Exclude list"
+REMOVE_FROM_INCLUDE_TOOLTIP = "Remove Selected Objects from Include"
+REMOVE_FROM_EXCLUDE_TOOLTIP = "Remove Selected Objects from Exclude"
+INCLUDE_OBJECTS_LABEL = "Include Objects..."
+EXCLUDE_OBJECTS_LABEL ="Exclude Objects..."
+REMOVE_FROM_INCLUDES_LABEL = "Remove Selected Objects from Include"
+REMOVE_FROM_EXCLUDES_LABEL = "Remove Selected Objects from Exclude"
+INCLUDE_LABEL = "Include"
+EXCLUDE_LABEL = "Exclude"
+ADD_INCLUDE_OBJECTS_TITLE = "Add Include Objects"
+ADD_EXCLUDE_OBJECTS_TITLE = "Add Exclude Objects"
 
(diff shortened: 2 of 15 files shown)


lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/includeExcludeWidget.py
```diff
@@ -110,114 +106,40 @@ def __init__(
...
-        excludeSelected = self._exclude.list.hasSelectedItems
+        includesSelected = self._include.list.hasSelectedItems()
+        excludeSelected = self._exclude.list.hasSelectedItems()
         self._deleteBtn.setEnabled(includesSelected or excludeSelected)
 
-        self._deleteBtnActionFromIncludes.setEnabled(includesSelected)
-        self._deleteBtnActionFromExcludes.setEnabled(excludeSelected)
+        # TODO: these QAction are reported as already being deleted in C++
+        # self._deleteBtnActionFromIncludes.setEnabled(includesSelected)
+        # self._deleteBtnActionFromExcludes.setEnabled(excludeSelected)
```
pierrebai-adsk, line 142: Note: these line fails in Maya, even in the current dev branch without my changes. Qt complains that the C++ QAction object has already been deleted. I think there is an issue with the retention of the QAction.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/expressionWidget.py
```diff
@@ -45,31 +46,20 @@ def __init__(self, collection: Usd.CollectionAPI, parent: QWidget, expressionCha
         mainLayout.addWidget(self._expressionText, 1)
 
         self._expressionText.installEventFilter(self)
-        self.update()
+        self.setLayout(mainLayout)
```
haberlu, line 49: this is in general not needed - as creating the mainLayout in line 21 already did set it to the widget - as the layout is getting created with this widget as a parent. But cannot hurt, as well.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/includeExcludeWidget.py
haberlu: -- whild guess begin -- 

maybe creating the QMenu in line 64 with self as a parent object may help with the retention of the menu?

-- whild guess end --

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/filteredStringListView.py
```diff
@@ -2,42 +2,44 @@
...
-    from PySide2.QtGui import QPalette, QPainter, QPaintEvent  # type: ignore
-    from PySide2.QtWidgets import QFrame, QLabel, QListView, QStyledItemDelegate, QStyleOptionViewItem  # type: ignore
+    from PySide2.QtGui import QPainter, QPaintEvent, QFont  # type: ignore
+    from PySide2.QtWidgets import QListView, QStyledItemDelegate, QStyleOptionViewItem  # type: ignore
 
 
 kNoObjectFoundLabel = "No objects found"
+kDragObjectsHereLabel = "Drag objects here"
+kPickObjectsLabel = "Click '+' to add"
+kDragOrPickObjectsLabel = "Drag objects here or click '+' to add"
```
haberlu, line 42: doesn't python normally use UPPER_CASE for constants? 
https://peps.python.org/pep-0008/#constants

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/filteredStringListView.py
```diff
@@ -65,101 +67,79 @@ def paint(
...
-            self.placeholder_label.setGeometry(self.viewport().geometry())
+        return bool(self.selectionModel().hasSelection())
+
+    def _paintPlaceHolder(self, placeHolderText):
+        painter = QPainter(self.viewport())
+        theme = Theme.instance()
+        painter.setPen(theme.palette.colorPlaceHolderText)
+        font: QFont = painter.font()
+        font.setPixelSize(theme.uiScaled(18))
+        painter.setFont(font)
```
haberlu, line 123: this was removed as of Scott's advice

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/stringListPanel.py
```diff
@@ -19,21 +21,26 @@
     from PySide2.QtGui import QPainter  # type: ignore
     from PySide2.QtWidgets import QCheckBox, QHBoxLayout, QVBoxLayout, QStyle, QStyleOptionHeaderV2, QStylePainter, QWidget  # type: ignore
 
+# TODO: support I8N
+kIncludeAllLabel = "Include all"
+kIncludeAllTooltip = textwrap.dedent(
+    """
+    When enabled, all prims are illuminated. When disabled,
+    only the prims in the Include list are illuminated.
+    """).strip()
```
haberlu, line 30: ... a bit of overkill to introduce textwrap - just for one line break - imho ?

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/data/__init__.py
haberlu, line 1: maybe just add a comment in here - to avoid confusing some programs with empty files

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/data/collectionData.py
```diff
@@ -0,0 +1,80 @@
...
+        '''
+        Return if there is a UI to pick items.
+        '''
+        return False
+    
+    def pick(self) -> Sequence[AnyStr]:
+        '''
+        Pick items and return the sequnce of picked items.
+        '''
+        return []
```
haberlu, line 28: What's the semantics about data being able to pick??? Only some kind of host can pick - data can only change. 

And if this would make syntactically sense, then canDrop would be missing (what seems to be still in host).

haberlu (commented):
Adding some comments.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/filteredStringListView.py
pierrebai-adsk: In the MayaUSD repo, the coding standard is to have constant be in the form I used, that's why I followed it in this case, as other Maya dev would have chimed in to make the constant follow the coding standard of the project. I can change them, but whatever standard we follow will clash with either DCC.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/expressionWidget.py
pierrebai-adsk: I think it might have been due to the merge conflicts being resolve like this. I can remove the line.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/filteredStringListView.py
pierrebai-adsk: You mean he did not want a larger font? I can remove the font size

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/stringListPanel.py
pierrebai-adsk: It simplify writing the code by not having to worry about code layout, indentation or it getting messed up by auto code layout.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/data/__init__.py
pierrebai-adsk: All other __init__.py in MayaUSD are empty, just like this one, we never had problems. The file is only used to mark the folder as being a Python module. I can add a comment to not make it empty.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/data/collectionData.py
pierrebai-adsk: I was worrying that picking in the future might need to know about the data (collection, prim or any others) that would affect it. But for now it doesn't, I can remove it and leave it in Host for now.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/data/__init__.py
pierrebai-adsk: Ah, pick needs to know the stage... I guess I will expose the stage for now. This makes the UI code "know" that the data is about USD, but I guess it is only minor.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/data/collectionData.py
pierrebai-adsk: Looking at the code, the picking does need to know about the data to get the stage. I guess I will expose the stage in the data for now.

barbalt (dismissed):

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/includeExcludeWidget.py
pierrebai-adsk: I found a way around the problem: don't try to keep the action objects as variable, look them up when needed. I suppose Qt destroys and rebuilds them? I know Maya create menus dynamically, maybe it is due to that.

barbalt (approved):

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/filteredStringListView.py
haberlu: yes, no larger font - the same font size is requested as everywhere else.

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/filteredStringListView.py
```diff
@@ -65,101 +67,76 @@ def paint(
...
-        super().resizeEvent(event)
-        if self.placeholder_label is not None:
-            self.placeholder_label.setGeometry(self.viewport().geometry())
+        return bool(self.selectionModel().hasSelection())
+
+    def _paintPlaceHolder(self, placeHolderText):
+        painter = QPainter(self.viewport())
+        theme = Theme.instance()
+        painter.setPen(theme.palette.colorPlaceHolderText)
+        painter.drawText(self.rect(), Qt.AlignCenter, placeHolderText)
```
haberlu, line 121: could be moved inside the paintEvent() override - to avoid being accidentally called outside the paintEvent - what must not be done in Qt.

haberlu (approved):

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/filteredStringListView.py
pierrebai-adsk: True even if unlikely. I will move it in a subsequent PR that is based on this one. (Not another refactor one :) )

PR #4054, opened 2024-12-18
EMSUSD-1971 Make job context UI registeration be order-neutral

pierrebai-adsk:
- Change the job context registration macro to be order-independent.
- Change unit test to verify that the macro are order-neutral by registering UI first.

### DIFF
--- a/test/lib/usd/plugin/nullApiExporter.cpp
+++ b/test/lib/usd/plugin/nullApiExporter.cpp
@@ -102,17 +102,6 @@ REGISTER_EXPORT_JOB_CONTEXT_FCT(
     return extraArgs;
 }
 
-REGISTER_EXPORT_JOB_CONTEXT_FCT(
-    Curly,
-    "Curly's special",
-    "Test coverage of error handling part deux")
-{
-    VtDictionary extraArgs;
-    // Incorrect type:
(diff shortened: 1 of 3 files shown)


test/lib/usd/plugin/nullApiExporter.cpp
```diff
@@ -129,6 +118,17 @@ REGISTER_EXPORT_JOB_CONTEXT_UI_FCT(Curly)
...
+    Curly,
+    "Curly's special",
+    "Test coverage of error handling part deux")
+{
+    VtDictionary extraArgs;
+    // Incorrect type:
+    extraArgs[UsdMayaJobExportArgsTokens->apiSchema] = VtValue(std::string("testApi"));
+    return extraArgs;
+}
+
```
seando-adsk, line 131: I don't see any changes other than moving the macro. Is there a reason this macro moved below the other one?

seando-adsk (approved):

test/lib/usd/plugin/nullApiExporter.cpp
seando-adsk: Ah yes perfect. Thanks for thinking of that.

PR #4055, opened 2024-12-19
EMSUSD-1975 - Optimize material libraries

JGamache-autodesk:
Allow deep library optimization ff a user introduces a material library that has nodes implemented via NodeGraph that embed optimizable nodes.

Also fixed issues encountered while dealing with MaterialX namespaces.


lib/mayaUsd/render/MaterialXGenOgsXml/LobePruner.cpp
```diff
@@ -242,7 +331,7 @@ void LobePrunerImpl::addOptimizableValue(
     valueMap.find(value)->second.push_back(PXR_NS::TfToken(input->getParent()->getName()));
 }
 
-bool LobePrunerImpl::getOptimizedNodeCategory(const mx::Node& node, std::string& nodeCategory)
+bool LobePrunerImpl::getOptimizedNodeDef(const mx::Node& node, mx::NodeDefPtr& nodeDef)
```
frohnej-adsk, line 334: [Minor] If the output parameter `nodeDef` is still needed, can we rename it to `optimizedNodeDef`? It's very easy to confuse `nd` and `nodeDef` while reading this method.

lib/mayaUsd/render/MaterialXGenOgsXml/LobePruner.cpp
```diff
@@ -288,8 +377,7 @@ bool LobePrunerImpl::getOptimizedNodeCategory(const mx::Node& node, std::string&
     }
 
     if (canOptimize) {
-        ensureLibraryHasOptimizedShader(ndName, flags);
-        nodeCategory = node.getCategory() + "_" + flags;
+        nodeDef = ensureLibraryHasOptimizedShader(ndName, flags);
```
frohnej-adsk, line 380: [Question] Could `ensureLibraryHasOptimizedShader()` now take a `Material::NodeDefPtr` instead of a name? Internally, it calls 
```C++
const auto originalNodeDef = _library->getNodeDef(nodeDefName.GetString());
```
which should just give us `nd`, so we might be able to pass that directly.

frohnej-adsk: Ah, it's also called in `getOptimizedNodeId()` and passing a name might be simpler there.

lib/mayaUsd/render/MaterialXGenOgsXml/LobePruner.cpp
```diff
@@ -356,39 +444,53 @@ PXR_NS::TfToken LobePrunerImpl::getOptimizedNodeId(const PXR_NS::HdMaterialNode2
...
     return retVal;
 }
 
-void LobePrunerImpl::ensureLibraryHasOptimizedShader(
+bool LobePrunerImpl::isOptimizedNodeId(const PXR_NS::TfToken& nodeId)
+{
+    return _optimizedNodeIds.count(nodeId) != 0;
+}
+
+mx::NodeDefPtr LobePrunerImpl::ensureLibraryHasOptimizedShader(
```
frohnej-adsk, line 458: [Thought] Wondering if the name could be improved, now that the method actually returns the node definition. Something like `addOptimizedNodeDef()` would be intuitive but it doesn't capture that the method first checks if it already exists.

Maybe we could just overload `getOptimizedNodeDef()`?

frohnej-adsk (commented):
Very nice! I added some thoughts on improving readability, but I'm not sure if all of it is possible or makes sense.

lib/mayaUsd/render/MaterialXGenOgsXml/LobePruner.cpp
JGamache-autodesk: Default semantics of "get" is to return noting if not found. Maybe "getOrAdd" would express the intent better?

frohnej-adsk (dismissed):

frohnej-adsk (approved):

JGamache-autodesk:
@seando-adsk ready for merge.

PR #4058, opened 2024-12-20
EMSUSD-1519 Collections: remove all includes and excludes

pierrebai-adsk:
- Add to the data classes a function to remove all includes and excludes.
- Add to the UI a menu item to remove all includes and excludes.
- Make the UI call the data when triggered.
- Make menu item separators show up even as first item.
- Fix place holder text in list view.
- Fix non-light collections.

### DIFF
--- a/lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/expressionRulesMenu.py
+++ b/lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/expressionRulesMenu.py
@@ -9,15 +9,28 @@ except ImportError:
 
 from pxr import Usd
 
+# TODO: support I8N
 EXPAND_PRIMS_MENU_OPTION = "Expand Prims"
 EXPAND_PRIMS_PROPERTIES_MENU_OPTION = "Expand Prims and Properties"
 EXPLICIT_ONLY_MENU_OPTION = "Explicit Only"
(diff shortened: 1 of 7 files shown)


lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/expressionRulesMenu.py
```diff
@@ -9,15 +9,28 @@
 
 from pxr import Usd
 
+# TODO: support I8N
 EXPAND_PRIMS_MENU_OPTION = "Expand Prims"
 EXPAND_PRIMS_PROPERTIES_MENU_OPTION = "Expand Prims and Properties"
 EXPLICIT_ONLY_MENU_OPTION = "Explicit Only"
+kIncludeExcludeLabel = "Include/Exclude"
```
barbalt, line 16: minor: I know this is maya convention, but we probably would want to keep it capitalized for consistency in this case

barbalt (approved):

lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/expressionRulesMenu.py
pierrebai-adsk: Ah, right I wrote these a few days ago, before the other review. We can change them afterward

PR #4060, opened 2025-01-03
HYDRA-1347 : Free resources and do not update when not using the VP2 …

lanierd-adsk:
…render delegate

### DIFF
--- a/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
+++ b/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
@@ -548,7 +548,9 @@ bool _IsVP2RenderDelegate()
     MHWRender::MRenderer* theRenderer = MHWRender::MRenderer::theRenderer();
     if (theRenderer) {
         const MString overrideName = theRenderer->activeRenderOverride();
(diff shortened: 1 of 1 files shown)


vlasovi (approved):
Looks good

jufrantz:
Hello @lanierd-adsk,

This change is very disruptive for our workflows. We rely on a custom MRenderOverride using [MSceneRender operations](https://help.autodesk.com/view/MAYADEV/2025/ENU/?guid=MAYA_API_REF_cpp_ref_class_m_h_w_render_1_1_m_scene_render_html). It must render the entire VP2 scene, including mayaUsdProxyShape through vp2RenderDelegate. This is no longer possible.

Could this be limited to the mtohRenderOverrides context only?

Additionally, we anticipate that we may need multiple viewports simultaneously, one using mtoh and others not. It would be helpful to preserve usage of vp2RenderDelegate for the non-mtoh viewport(s) is this case.

Best,
Julien

lanierd-adsk:
Hello @jufrantz ,
Thank you for your feedback, ok, I will change this to be narrowed down when using a hydra render delegate as I need to update this again as it doesn't work in all expected cases for Hydra Storm.

lanierd-adsk:
This commit is being fully reverted by PR : https://github.com/Autodesk/maya-usd/pull/4069 to avoid breaking a customer's workflow and it also breaks maya hydra unit tests.

PR #4066, opened 2025-01-08
EMSUSD-2000 - Do not contaminate HdStorm libraries with lobe pruning

JGamache-autodesk:
MayaUSD should use a copy of the HdStorm MaterialX library in order to optimize it without affecting any Storm based render delegate.


lanierd-adsk (approved):
Thanks for the fix !

JGamache-autodesk:
@seando-adsk ready for merge

PR #4067, opened 2025-01-09
EMSUSD-1657 light linking undo and redo

pierrebai-adsk:
- Added methods on the Host class to create the data used by the UI.
- Host create USD collection and string list by default.
- Use the Host to create the data in the UI classes.
- UsdCollectionData uses Host to create its string lists.
- Make the expression widget update itself on USD data changed notification even when there is no expression attribute.
- Rename the light linking custom control to be collection custom control.
- Move the Maya host class to its own file.
- Implement Maya-specific collection data with all needed setters.
- Add a _UsdUndoBlockContext class to capture all USD changes into undo items and later transfer them to a Maya command.
- Add a _UsdUndoBlockCommand that transfer the captured undo items into itself for future undo and redo.
- Add a _UndoItemHolder class to hold the undo items to be transferred between the Python undo context and the Maya command.
- Add multiple command sub-classes so that the undo UI shows a nice and comprehensible undo name.

In theory, we could have just used the existing UsdUndoBlock and that was in the initial prototype, but this makes all entries in the undo UI all be "UsdUndoBlock" which is incomprehensible for the user and does not state what would be undone or redone. Using one command per action allows the user to know what is in the undo stack.

### DIFF
--- a/lib/mayaUsd/resources/ae/CMakeLists.txt
+++ b/lib/mayaUsd/resources/ae/CMakeLists.txt
@@ -43,7 +43,8 @@ install(FILES __init__.py DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/python/${PROJE
 if(MAYA_APP_VERSION VERSION_GREATER_EQUAL 2023)
     foreach(_SUBDIR ${MAYAUSD_AE_TEMPLATES})
         install(FILES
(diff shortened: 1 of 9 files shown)


pierrebai-adsk:
Note: the renamed files are purely Maya-only code, not shared code.

barbalt (approved):

PR #4069, opened 2025-01-13
HYDRA-1347 : Fully revert the changes from PR #4060.

lanierd-adsk:


lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp
```diff
@@ -542,20 +542,6 @@ bool _DrawItemFilterPredicate(const SdfPath& rprimID, const void* predicateParam
 
 bool _longDurationRendering = false;
 
-bool _IsVP2RenderDelegate()
```
lanierd-adsk, line 545: I am fully reverting this change as it breaks a customer's workflow and also breaks maya hydra unit tests.

lanierd-adsk:
This is a full revert of https://github.com/Autodesk/maya-usd/pull/4060 as it breaks a customer's workflow and also maya hydra unit tests.

vlasovi (approved):

PR #4075, opened 2025-01-15
EMSUSD-2012 implement theme scale in Maya Host

pierrebai-adsk:
Implement the uiScaleFactor function in the Maya host.

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/collectionMayaHost.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/collectionMayaHost.py
@@ -7,6 +7,7 @@ from usd_shared_components.usdData.usdCollectionStringListData import Collection
 from maya.api.OpenMaya import MPxCommand, MFnPlugin, MGlobal, MSyntax, MArgDatabase
 import mayaUsd.lib
 import maya.mel as mel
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/resources/ae/usdschemabase/collectionMayaHost.py
```diff
@@ -313,6 +314,10 @@ def __init__(self):
         self._palette = None
         self._icons = {}
 
+    @property
+    def uiScaleFactor(self) -> float:
+        return float(cmds.mayaDpiSetting(query=True, scaleValue=True))
```
seando-adsk, line 319: The cmd does not exist on OSX, so you should check it exists. And I think you should use the realScaleValue.

`1.0 if not hasattr(cmds, "mayaDpiSetting") else cmds.mayaDpiSetting(query=True, realScaleValue=True)`

seando-adsk (changes requested):
Verify if cmd is present (its not on OSX)

seando-adsk (approved):

pierrebai-adsk:
The only failure was a temp file being left in the TEMP folder on a Windows machine.

PR #4083, opened 2025-01-22
EMSUSD-2095 multiple small UI fixes

pierrebai-adsk:
- Don't draw rectangle in list paint event that cause error output.
- Don't let the delete and backspace key events out of the list view to avoid deleting the DCC selection. (What is selected in the Maya outliner.)
- Don't add items that are already present.
- Fix invalid expression error message.
- Make use of the Host reportMesage function to report errors and info.
- Default to the expression tab if only expressions are used.

### DIFF
--- a/lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/filteredStringListView.py
+++ b/lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/filteredStringListView.py
@@ -72,13 +72,11 @@ class FilteredStringListView(QListView):
 
         self.setCursor(Qt.ArrowCursor)
 
-        DragAndDropEventFilter(self, data)
+        DragAndDropAndDeleteEventFilter(self, data)
 
         self.selectionModel().selectionChanged.connect(self.itemSelectionChanged)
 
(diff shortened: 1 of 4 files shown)


lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/common/filteredStringListView.py
```diff
@@ -72,13 +72,11 @@ def __init__(self, data: StringListData, headerTitle: str = "", parent=None):
...
         self.setCursor(Qt.ArrowCursor)
 
-        DragAndDropEventFilter(self, data)
+        DragAndDropAndDeleteEventFilter(self, data)
 
         self.selectionModel().selectionChanged.connect(self.itemSelectionChanged)
 
     def paintEvent(self, event: QPaintEvent):
-        painter = QPainter(self)
-        painter.fillRect(event.rect(), Theme.instance().palette.colorListBorder)
```
pierrebai-adsk, line 81: Not sure why we were painting this rectangle, I could not see a visual change when I removed it and it cased Qt to issue errors on stdout

barbalt (approved):

PR #4094, opened 2025-01-29
HYDRA-1347 : Fix double texture loading when using Hydra

debloip-adsk:
This PR skips MayaUsd handling of USD data when Hydra is being used, to avoid unnecessary processing and data loading. The change essentially rehabilitates an existing fix, but also updates it for the new MayaHydra render override prefix.

The changes will only take effect in viewports using MayaHydra; this should address the concerns raised in the previous attempt at solving this issue : https://github.com/Autodesk/maya-usd/pull/4060#issuecomment-2579789457. We only block rendering and selection, as they are both handled by MayaHydra.

I ran MayaHydra unit tests and all passed (except `testMaterialXOnNative` as I did not have LookdevX installed).

### DIFF
--- a/lib/mayaUsd/render/mayaToHydra/utils.h
+++ b/lib/mayaUsd/render/mayaToHydra/utils.h
@@ -20,9 +20,6 @@
 #include <pxr/imaging/hd/renderDelegate.h>
 #include <pxr/pxr.h>
 
-#include <maya/MFrameContext.h>
-#include <maya/MString.h>
-
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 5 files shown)


lanierd-adsk (approved):

vlasovi (approved):
Looks good

debloip-adsk:
@jufrantz This PR should address the concerns you raised here : https://github.com/Autodesk/maya-usd/pull/4060#issuecomment-2579789457

This fix only applies to Hydra viewports, so you should still be able to use a custom render override, as well as have a mix of Hydra and non-Hydra viewports. Let us know if there are other concerns, thanks

jufrantz:
Thanks for the update @debloip-adsk. 
This new version will fit our needs perfectly.

PR #4097, opened 2025-01-30
EMSUSD-1516 Sync drag and drop from max

barbalt:
Sync drag and drop elements from shared component
Added a new undoable command to replace strings when editing an element in the collection component

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/collectionMayaHost.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/collectionMayaHost.py
@@ -222,6 +222,10 @@ class _RemoveItemsCommand(_UsdUndoBlockCommand):
     def __init__(self):
         super().__init__()
 
+class _ReplaceItemsCommand(_UsdUndoBlockCommand):
+    commandName = 'usdCollectionReplaceItems'
+    def __init__(self):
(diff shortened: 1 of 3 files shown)


pierrebai-adsk (dismissed):

lib/mayaUsd/resources/ae/usdschemabase/collectionMayaHost.py
```diff
@@ -307,19 +312,26 @@ class MayaStringListData(CollectionStringListData):
...
         '''
         with _UsdUndoBlockContext(_RemoveItemsCommand.commandName):
-            super().removeStrings(items)
+            return super().removeStrings(items)
+
+    def replaceStrings(self, oldString, newString) -> bool:
+        '''
+        Replace the old string with thew newString on the model.
+        '''
+        with _UsdUndoBlockContext(_ReplaceItemsCommand.commandName):
```
pierrebai-adsk, line 333: This will end-up calling teh Maya version of the addStrings and removeStrings. Not sure how commands-within-commands will end-up showing up in the undo queue in Maya. We can try this out and test it and improve it later if need be.

pierrebai-adsk (approved):

PR #4100, opened 2025-01-31
EMSUSD-2111 print collection membership

pierrebai-adsk:
- Add getCollectionName and getObjectName functions to the CollectionData class.
- Implement those functions in the UsdCollectionData class.
- Add a menu item to prim the collection.
- Print the collection by calling computeMembership and printing the result.


lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/expressionRulesMenu.py
```diff
@@ -15,6 +15,7 @@
 INCLUDE_EXCLUDE_LABEL = "Include/Exclude"
 REMOVE_ALL_LABEL = "Remove All"
 CLEAR_OPINIONS_LABEL = "Clear Opinions from Target Layer"
+PRINT_PRIMS_LABEL = "Print Prims to Script Editor"
```
barbalt, line 18: Need to expose a way to have this saying be different in max/maya

pierrebai-adsk: Yes -- but I have no idea how to best make label be dependent on DCC. Maybe have a method on Theme to pass labels to to "translate" them?

barbalt (approved):

PR #4106, opened 2025-02-06
ShadingModeExporter: Fix potential crash when exporting collection-based material bindings.

jufrantz:
This was caused by an out-of-range vector access.


lib/mayaUsd/fileio/shading/shadingModeExporter.cpp
```diff
@@ -130,8 +130,8 @@ void UsdMayaShadingModeExporter::DoExport(
 
         if (mat) {
             writeJobContext.AddMaterialPath(mat.GetPath());
-            exportedMaterials.push_back(mat);
             if (!boundPrimPaths.empty()) {
+                materialsToBind.push_back(mat);
```
jufrantz, line 134: Here is the real fix. It ensures that `materialsToBind` and `matAssignments` sizes match, as expected later in the function.

pierrebai-adsk (approved):

PR #4107, opened 2025-02-06
EMSUSD-1626 : Support MaterialX document in USD Exporter

derlerk:
ShaderWriter for the LookdevX MaterialXStack material.
Test that validate an export of MaterialXStack with surfaceshader and displacement


lib/usd/translators/shading/mtlxMaterialXSurfaceShaderWriter.cpp
```diff
@@ -0,0 +1,516 @@
...
+void _SetShaderInfoAttributes(
+    const MaterialX::NodePtr& node,
+    UsdPrim&                  shaderPrim,
+    const Ufe::Path&          ufePath)
+{
+    auto infoAttr = shaderPrim.CreateAttribute(
+        UsdShadeTokens->infoId, SdfValueTypeNames->Token, SdfVariabilityUniform);
+
+    auto nodeDefString = _GetNodeDefString(node, ufePath);
+    infoAttr.Set(TfToken(nodeDefString));
```
JGamache-autodesk, line 87: That works fine, but it would be preferable to use the UsdShadeShader API:
``` 
    auto shader = UsdShadeShader(shaderPrim);
    shader.CreateIdAttr(VtValue(TfToken(nodeDefString)));
```

lib/usd/translators/shading/mtlxMaterialXSurfaceShaderWriter.cpp
```diff
@@ -0,0 +1,516 @@
...
+    // Render Document is the MaterialX document
+    childPlug = depNodeFn.findPlug("renderDocument", true, &status);
+    MString renderDocumentString;
+    childPlug.getValue(renderDocumentString);
+
+    auto mtlxDoc = MaterialX::createDocument();
+    readFromXmlString(mtlxDoc, renderDocumentString.asChar());
+
+    // surfaceMaterialNode
+    auto MaterialNode = mtlxDoc->getNode(depNodeFn.name().asChar());
```
JGamache-autodesk, line 454: I try to keep that name in sync with the MaterialX surfacematerial node name, but it can diverge since it must be unique in Maya. The real material node name can be found at the end of the path: `ufePath.back()`.

For example, if you import the same `standard_surface_gold.mtlx` MaterialX document twice, and assign `Gold` from each document on separate spheres, then the first depNodeFn will have `Gold` as name, but the second one will be stuck with `Gold1` since these need to be unique. In both cases the last component of the Ufe path will be correct:
```
|materialXStack1|materialXStackShape1,%standard_surface_gold%Gold
|materialXStack1|materialXStackShape1,%standard_surface_gold2%Gold
```

test/lib/usd/translators/testUsdExportMaterialXSurfaceShader.py
```diff
@@ -0,0 +1,105 @@
...
+        cls._inputPath = fixturesUtils.setUpClass(__file__)
+
+    @classmethod
+    def tearDownClass(cls):
+        standalone.uninitialize()
+
+    def testExportMaterialXStack(self):
+
+        cmds.file(f=True, new=True)
+
```
JGamache-autodesk, line 48: A very brutal script to assign the brick shader to a sphere is provided for inspiration. It could be modified to import any MaterialX file of interest for testing:

```python
import maya.cmds as cmds
import maya.mel as mel
import ufe

stackName = mel.eval("createNode materialxStack")
stackPathString = mel.eval("ls -l " + stackName)[0]
stackItem = ufe.Hierarchy.createItem(ufe.PathString.path(stackPathString))
stackHierarchy = ufe.Hierarchy.hierarchy(stackItem)
stackContextOps = ufe.ContextOps.contextOps(stackItem)
stackContextOps.doOp(['MxImportDocument', r'MX_INSTALL_PATH\resources\Materials\Examples\StandardSurface\standard_surface_brick_procedural.mtlx'])
documentItem = stackHierarchy.children()[-1]
sphere, sphereShp = cmds.polySphere()
cmds.move(1, 1, 0, sphere)
surfPathString = ufe.PathString.string(documentItem.path()) + "%M_BrickPattern"
cmds.select(sphere)
materialContextOps = ufe.ContextOps.contextOps(ufe.Hierarchy.createItem(ufe.PathString.path(surfPathString)))
materialContextOps.doOp(['Assign Material to Selection'])                 
```

This would be a lot more flexible than loading a Maya file with embedded MaterialX data.

JGamache-autodesk (changes requested):
Very good. I would say 99% done. Sadly there are a few corner cases in MaterialX that need to be handled correctly. I have described the ones I encountered in the past.

JGamache-autodesk:
Sorry about that, but I remembered another corner case that will hit us at some point. Simplest form is:
image
You need to handle `shared_image1` correctly if exporting both surfacematerials. I suspect a first implementation will split it and create one image node per material. Maybe later we can consider using USD referencing to keep it shared.

JGamache-autodesk (approved):
Thanks for doing the work! Looks good to me now.

JGamache-autodesk (approved):

PR #4110, opened 2025-02-07
EMSUSD-1910 - Make USD camera invisible when the parents turn invisible

samuelliu-adsk:
- This change is the Maya-usd side of the story.
- Currently if the user hide the parent of a camera or the entire stage node, the camera gizmo will still show up in the viewport.
- Adding the visibility function to camera because the original visibility function in UsdObject3d will be used to determine the outliner row visibility as well. If we return false in that function, the outliner will hide the camera as well, which does not align with the behavior of other prim types.


lib/usdUfe/ufe/UsdObject3d.cpp
```diff
@@ -108,6 +110,23 @@ bool UsdObject3d::visibility() const
     auto            visAttr = PXR_NS::UsdGeomImageable(_prim).GetVisibilityAttr();
     visAttr.Get(&visibilityToken);
 
+    // Check if the prim is a camera
+    // if its a camera, then we need to check if the camera's parents are visible
+    if (_prim.IsA<PXR_NS::UsdGeomCamera>()
```
deboisj, line 115: This will break the UFE explorer we have in 3dsmax : https://git.autodesk.com/brossee/3dsmax-component-usd-public/blob/4718c8ae447cea4ed005d3642cd6dcfd37c54682/src/UFEUI/standardTreeColumns.cpp#L203 

We need visibility() to return the authored attribute

samuelliu-adsk: This issue should be addressed by my new changes. Basically I moved the change to MayaUsdObject3d instead of object3d.

lib/mayaUsd/ufe/MayaUsdObject3d.cpp
```diff
@@ -52,5 +54,33 @@ MayaUsdObject3d::adjustAlignedBBox(const Ufe::BBox3d& bbox, const PXR_NS::UsdTim
...
+#ifndef UFE_CAMERA_HAS_VISIBILITY
+bool MayaUsdObject3d::visibility() const
+{
+    PXR_NS::TfToken visibilityToken;
+    auto            visAttr = PXR_NS::UsdGeomImageable(prim()).GetVisibilityAttr();
+    visAttr.Get(&visibilityToken);
+
+    // Check if the prim is a camera
+    // if its a camera, then we need to check if the camera's parents are visible
+    if (prim().IsA<PXR_NS::UsdGeomCamera>()
```
deboisj, line 66: Will not break 3dsmax, but i still find it a bit iffy that now the behavior varies per object type. Makes the API inconsistant / contract unclear

deboisj: Also awkward because now setVisiblity and getVisibility will not be symmetrical, you will set the authored attr but return the resolved visibility

seando-adsk: @deboisj Do you have another suggestion on how we can accomplish this task to hide cameras in the 3d viewport when they are hidden because their parent is hidden?

deboisj: I am not sure - how do you handle this for lights? Isnt it the same problem? In max - we have a mechanism to get the resolved visibility and maintain / update a cache for that - but i guess in your case that would have to be done in maya which sounds impractical Could you "hide" it by giving it no gizmo / render item?

samuelliu-adsk: Hi @deboisj , we can't do it in Maya because Maya shouldn't know anything about USD and unlike max Maya doesnt cache for it. Having the computedVisibility calculated in UFE (implemented in mayaUSD) seems like the best approach to do hence the added `computedVisiblity()` function in camera.cpp. This change here is for older maya versions. Yes, ideally this should be the "computedVisibilty" as in UsdCamera.cpp but I dont see a way to do that without breaking backward compatibilty... If the user set it invisible/visible it won't change the behavior, this only changes the inherited behavior.

seando-adsk (approved):

lib/usdUfe/ufe/UsdCamera.cpp
```diff
@@ -478,4 +480,33 @@ bool UsdCamera::renderable() const
     return purpose == UsdGeomTokens->render || purpose == UsdGeomTokens->default_;
 }
 #endif
+
+#ifdef UFE_CAMERA_HAS_COMPUTEDVISIBILITY
```
deboisj, line 484: Is there a reason why "computedVisibility" is on camera and not object3d? Wouldnt that be useful / make sense for all objects that carry visibility?

lib/usdUfe/ufe/UsdCamera.cpp
```diff
@@ -478,4 +480,33 @@ bool UsdCamera::renderable() const
...
+bool UsdCamera::computedVisibility() const
+{
+    PXR_NS::TfToken visibilityToken;
+    auto            visAttr = PXR_NS::UsdGeomImageable(prim()).GetVisibilityAttr();
+    visAttr.Get(&visibilityToken);
+
+    // Check if the prim is a camera
+    // if its a camera, then we need to check if the camera's parents are visible
+    if (prim().IsA<PXR_NS::UsdGeomCamera>()
```
deboisj, line 493: This check might on type might be redundant with https://git.autodesk.com/maya3d/maya-usd/blob/1082fa25ecf241d870733be9f5249bfa74a68a85/lib/usdUfe/ufe/UsdCameraHandler.cpp#L53 ?

lib/usdUfe/ufe/UsdCamera.cpp
samuelliu-adsk: We put it in Camera because other prims (non-proxy) will handle compute visibility on USD side, but camera prims need to feed this computedVisibility to Maya due to the proxy nature.

deboisj (approved):

PR #4114, opened 2025-02-11
EMSUSD-1932 applied schemas in AE

pierrebai-adsk:
- Check both suppressed and already-added attributes when creating section to avoid adding attributes twice to the AE.
- Support schemas without Python support.


AramAzhari-adsk (dismissed):
Will there be a need for a unit test?
I suppose this is all visual changes.

AramAzhari-adsk (approved):

AramAzhari-adsk (approved):

PR #4130, opened 2025-02-21
EMSUSD-2171 Fix multiple collection UI issues

pierrebai-adsk:
Also addresses: EMSUSD-2168, EMSUSD-2169, EMSUSD-2171, EMSUSD-2167, EMSUSD-2151

Fix validation error:
- Make the convertToCollectionString return the cause of any error.
- Make the validator only handle the attribute error exception caused by the absence of expression support in old USD, not all exceptions.

Fix list editor:
- Replace ad-hoc code with QCompleter and QItemDelegate to edit the list items.
- This fixes multiple interaction bugs.
- Mark the filtered string list model items as being editable.
- Make sure the list view gets focus after edit.
- Make sure the new item can be added before deleting the old item in replaceStrings.
- Simplify how the stage prims are gathered for suggestions.


barbalt (approved):

barbalt:
So much simpler with qcompleter. Thank you

barbalt (approved):

Issue #4129, opened 2025-02-21
MaterialX 1.39.2 support


boberfly:
**Describe the issue**
When building OpenUSD 25.02a with MaterialX 1.39.x, I've found that due to the API changes it cannot build.

This is just slightly ahead of 1.39.2 release but I think it's due to the 1.38 to 1.39 changes that OpenUSD/usdMtlx needed to address.

**Build log**
[maya-usd.log](https://github.com/user-attachments/files/18903087/maya-usd.log)

**Specs:**
 - OS & version - Windows 10
 - Compiler & version - Visual Studio 17 2022 v143
 - Maya version - Maya 2025.3
 - Maya USD commit - release/v0.31.0 (also testing dev branch now `804686e`)
 - Pixar USD commit - release (v25.02a)
 - MaterialX 1.39.x (commit `95fc7c5`)

**Additional context**
Looks like I might be a bit ahead on things. I can take a look and see if I can add 1.39.2 support myself and make a PR.

boberfly:
Just to note https://github.com/Autodesk/maya-usd/pull/4160 addresses this, thank you.

PR #4134, opened 2025-02-23
MaterialX 1.39.x support

boberfly:
Hi,

This is a simple modification to work with MaterialX 1.39.x so that it can compile against the changes that were made from the 1.38.x version, all pre-processor modifications eg.
```
#if MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION <= 38
...
#else
...
#endif
```

I've not made extensive testing with this to see if everything is working, it merely lets you compile. The channels API was removed so this probably needs to most attention to see if the behaviour has changed or broke things.

Addresses https://github.com/Autodesk/maya-usd/issues/4129


lib/mayaUsd/render/MaterialXGenOgsXml/GlslFragmentGenerator.cpp
```diff
@@ -161,7 +161,11 @@ const string& HwSpecularEnvironmentSamples::name()
 
 string GlslFragmentSyntax::getVariableName(
     const string&   name,
+#if MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION <= 38
```
JGamache-autodesk, line 164: Please use the combined syntax:
```c++
#if MX_COMBINED_VERSION < 13900
```

lib/mayaUsd/render/MaterialXGenOgsXml/GlslFragmentGenerator.cpp
```diff
@@ -896,12 +924,21 @@ ShaderPtr GlslFragmentGenerator::generate(
...
 
+#if MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION <= 38
 void GlslFragmentGenerator::toVec3(const TypeDesc* type, string& variable)
 {
     if (type->isFloat2()) {
         variable = "vec3(" + variable + ", 0.0)";
     } else if (type->isFloat4()) {
         variable = variable + ".xyz";
+#else
+void GlslFragmentGenerator::toVec3(const TypeDesc type, string& variable)
```
JGamache-autodesk, line 935: Please use `const TypeDesc&` ?

lib/mayaUsd/render/MaterialXGenOgsXml/ShaderGenUtil.cpp
```diff
@@ -316,7 +316,11 @@ std::string TopoNeutralGraph::gatherChannels(const mx::Input& input)
     }
     const auto& ngInput = interfaceInput ? *interfaceInput : input;
 
+#if MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION <= 38
     std::string channelInfo = ngInput.getChannels();
+#else
```
JGamache-autodesk, line 321: Remove the `#else` part. Channels are gone, so we should not see any here. And if we see some, it will be because we forgot to upgrade the document from 1.38 to 1.39, which might be needed in a few places.

JGamache-autodesk (changes requested):
That will allow compiling. I do have doubts about the shadergen fully working with 1.39 out of the box. There are a few 1.39 backports that are no longer required.

lib/mayaUsd/render/MaterialXGenOgsXml/GlslFragmentGenerator.cpp
boberfly: Done.

lib/mayaUsd/render/MaterialXGenOgsXml/GlslFragmentGenerator.cpp
boberfly: Done.

lib/mayaUsd/render/MaterialXGenOgsXml/ShaderGenUtil.cpp
boberfly: For this one, do you think `gatherChannels` could just be bypassed entirely, or return an empty string?

JGamache-autodesk: Yes, please bypass it entirely on 1.39.

boberfly: No worries, I got some time to look at this - instead of passing an empty `channelInfo` around I needed to tweak the `cloneConnection`/`cloneNodeGraphConnection` functions to not have `channelInfo`.

boberfly:
Just to note when I rebased this change to the latest HEAD of dev, I ran into commit https://github.com/Autodesk/maya-usd/commit/1524d43650f14238ff1105376eebe186d49990ed where Maya 2025.3's devkit hasn't got some of the stuff that it needs to compile.

JGamache-autodesk (approved):
Yes, very good, thank you!

I suspect, though, that the topoChannels test, found in `test\lib\mayaUsd\utils\test_ShaderGenUtils.cpp`, will fail with MaterialX 1.39.X as it contains many channel related checks that are no longer valid. I don't think this test is easily salvageable though, so maybe you want to disable it for 1.39? The other tests in that unit should still work fine (fingers crossed). Can you try running them and see if we need to fix or disable more of them.

JGamache-autodesk:
Sorry to be the bearer of bad news, but have you noticed https://github.com/AcademySoftwareFoundation/MaterialX/pull/2176 ?

We are trying to evaluate the impact at this point and see if it would be preferable to skip directly to 1.39.3.

I was about to build MayaUSD using this PR and got caught by surprise by all the compilation errors from MaterialX 1.39.3.

PR #4136, opened 2025-02-25
EMSUSD-1896 avoid crash when shader has no uvLink

pierrebai-adsk:
The uvLink command can return nothing if it cannot determine the UV link of the shader node. This would break assumptions in the export code and lead to a crash when accessing missing data. Handle the case where the uvLink command returns nothing to prevent that crash.

Add a unit test that reproduced the crash and now no longer crashes.


pierrebai-adsk:
The changes are good? I was unsure how to best handle the missing uvLink. I was wondering if maybe the command did not support the connected node? It might be worth digging into this further in the future?

For now, all I wanted was to at least avid the crash.

pierrebai-adsk:
There were UV in the resulting file for that node. The UV were not replaced by that "largest" one.

PR #4141, opened 2025-02-28
Use new Sdr API

dj-mcg:
USD 2505 introduces new shader registry API, moving from Ndr to Sdr. This converts maya-usd callsites with API guards


pierrebai-adsk:
While the changes work now, since Maya is not using USD 25.05 in any branches, we're likely to break things from time to time in new code, unfortunately. I'm already regularly being tripped up on differences between Maya 2022 up to most recent versions due to differences in USD (and UFE).

I'm only saying this to expect we will probably get things wrong and beak USD 25.05 until we have an internal build exercising it.

pierrebai-adsk (dismissed):

lib/mayaUsd/fileio/utils/shadingUtil.cpp
```diff
@@ -153,7 +153,11 @@ UsdShadeOutput UsdMayaShadingUtil::CreateShaderOutputAndConnectMaterial(
     TfToken               outputName = terminalName;
     SdrShaderNodeConstPtr shaderNodeDef = registry.GetShaderNodeByIdentifier(nodeID);
     if (shaderNodeDef) {
+#if PXR_VERSION >= 2025
```
seando-adsk, line 156: Should be `2505`. This caused an error in the preflight since it compiled with your new code.

seando-adsk (changes requested):
version mistake in shadingUtil.cpp

lib/mayaUsd/fileio/utils/shadingUtil.cpp
dj-mcg: Thank you! Fixed

seando-adsk (dismissed):

lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -158,7 +163,12 @@ def parseShaderAttributes(self):
                 pageLabel = 'Extra Shader Attributes'
             group = AEShaderLayout.Group(pageLabel)
             for name in nodeDef.GetPropertyNamesForPage(page):
-                if nodeDef.GetInput(name):
+                if Usd.GetVersion() < (0, 25, 5):
+                    nodeInput = nodeDef.GetInput()
+                else:
+                    nodeInput = nodeDef.GetShaderInput()
```
pierrebai-adsk, line 169: The line above are in error: forgot to pass the `name` parameter. That is the cause of the unit test failure.

So much for doing code reviews eh? :)

pierrebai-adsk: I fixed it for you.

seando-adsk (approved):

pierrebai-adsk (approved):

seando-adsk:
Internal failure with one of the preflight jobs not related to your changes. Marking as ready-for-merge.

PR #4143, opened 2025-03-05
EMSUSD-2084 - adding or removing of items in collections should respect the edit restrictions as other attributes

haberlu:
Adding isRelationshipEditAllowed() to UsdUfe.

This function can be used to determine if a list of paths could be added to or removed from a collection within the current edit target.


lib/usdUfe/ufe/Utils.cpp
```diff
@@ -130,6 +130,18 @@ UsdUfe::Transform3dMatrixOpNameFn  gTransform3dMatrixOpNameFn = nullptr;
...
+{
+    std::string result;
+    for (const auto& item : items) {
+        if (!result.empty()) {
+            result += ", ";
+        }
+        result += item;
+    }
+    return result;
+}
```
seando-adsk, line 143: Could you just use [TfStringJoin](https://openusd.org/docs/api/group__group__tf___string.html#ga2693efcb60908bf5b07caa0786ec000a)?

seando-adsk (approved):
Looks good to me. Thanks for the unit test.

seando-adsk:
Added Pierre as reviewer as he knows a lot more about edit restrictions that I do.

lib/usdUfe/ufe/Utils.h
```diff
@@ -394,6 +394,25 @@ bool isEditTargetLayerModifiable(
...
+//! \param targetsToAdd The optional paths to be added to the relationship. If
+//!        not nullptr, the entries will be adjusted to only keep the paths that
+//!        are allowed to be added to the relationship.
+//! \param targetsToRemove The optional paths to be removed from the
+//!        relationship. If not nullptr, the entries will be adjusted to only
+//!        keep the paths that are allowed to be removed from the relationship.
+//! \param errMsg The optional error message to be filled in case of any target
+//!        not being able to be added/removed.
+//! \return True, if one ore more of the targets given are allowed to be
+//!        added/removed to the relationship in the stage's local Layer Stack.
```
pierrebai-adsk, line 408: Not sure if we should require that *all* desired changes be allowed? Was that specified in the story that we allow if a single one is correct and all the other incorrect?

lib/usdUfe/python/wrapUtils.cpp
```diff
@@ -101,6 +101,79 @@ bool _isAttributeEditAllowed(const PXR_NS::UsdAttribute& attr)
...
+    std::string errMsg;
+    bool        result = true;
+    {
+        PXR_NS::SdfPathVector targetsToAddVec = extract<PXR_NS::SdfPathVector>(targetsToAdd);
+        PXR_NS::SdfPathVector targetsToRemoveVec = extract<PXR_NS::SdfPathVector>(targetsToRemove);
+
+        result = UsdUfe::isRelationshipEditAllowed(
+            relationship, &targetsToAddVec, &targetsToRemoveVec, &errMsg);
+
+        // pass the removal of elements to the python lists
```
pierrebai-adsk, line 118: Would it not be much simpler to copy the vector in full instead of these complicated iterations when the length changed? Popping multiple times will probably be less efficient anyway...

pierrebai-adsk (dismissed):
I think the wrapping code list update is error-prone and could be made simpler, but otherwise this looks fine.

lib/usdUfe/python/wrapUtils.cpp
haberlu: I'm not sure if we can re-assign the list as the list object is a python object living in the calling context and getting passed from python into this function. In my humble and still tiny understanding of the python/c++ inter-op, I thought this needs to be adapted in place to have any affects on the outside scope, hence the popping.

haberlu: ...well, I tried to use your proposal - basically doing this ...
```c++
if (len(targetsToAdd) != static_cast<int>(targetsToAddVec.size())) {
    targetsToAdd = list(targetsToAddVec);
}
if (len(targetsToRemove) != static_cast<int>(targetsToRemoveVec.size())) {
    targetsToRemove = list(targetsToRemoveVec);
}
```
... and it stops working. 

The reason why it stops working is, as I wrote before (and what's the reason for this kinda complicated modification of the original list instead of reassigning a new one), that re-assigning the list in python does not **modify** the existing one.

But as we need to modify the lists **in place** to be able to see the changes within the calling scope, the only way to adapt an existing list in python is to add things to or to pop things from it. Hence the complicated code.

I am replacing the code with a loop that does the same but is easier to reason about.  Performance-wise, that shouldn't matter much, it does some more adding and deletion, but omits the comparisons... See the next commit on this PR.

pierrebai-adsk: Well, I suggested using "=", but maybe the correct code was .clear + .splice or something similar. Obviously, if you can call .pop on it, you should be able to assign to it. The new code you wrote, is basically what I'm saying. I think you should be able to call .clear or something similar instead of a loop of pops, but really, at this point, it does not matter much, this is the error path anyway, so performance is not that much an issue., al least now the code is much easier to understand!

pierrebai-adsk (approved):

PR #4146, opened 2025-03-06
Selectable imaging node caused issues with XRay shading

dj-mcg:
This also uses a more current DrawOverride API

### DIFF
--- a/lib/mayaUsd/render/pxrUsdMayaGL/hdImagingShapeDrawOverride.h
+++ b/lib/mayaUsd/render/pxrUsdMayaGL/hdImagingShapeDrawOverride.h
@@ -88,12 +88,12 @@ public:
     bool wantUserSelection() const override { return true; }
     MAYAUSD_CORE_PUBLIC
     bool userSelect(
-        const MSelectionInfo&           selectionInfo,
-        const MHWRender::MDrawContext&  context,
-        const MDagPath&                 objPath,
-        const MUserData*                data,
-        MSelectionList&                 selectionList,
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/render/pxrUsdMayaGL/hdImagingShapeDrawOverride.h
```diff
@@ -81,6 +81,23 @@ class PxrMayaHdImagingShapeDrawOverride : public MHWRender::MPxDrawOverride
         const MHWRender::MFrameContext& frameContext,
         MUserData*                      oldData) override;
 
+    // wantUserSelection() and userSelect() are overridden to make sure
+    // HdImagingShape is not selectable. It causes a deadlock in VP2 in x-ray
+    // mode, if this node is selectable. (MAYA-2941)
```
AramAzhari-adsk, line 86: The jira ticket may cause confusion since it is not an Autodesk's ticket.

AramAzhari-adsk: Is it worth updating the class docs at line 44 to indicate the updated selection behavior?

dj-mcg: I removed the ticket, and I don't think there is any user-facing selection behavior differences introduced by this change. This only makes the imaging node itself non-selectable, which is the expected behavior anyway

AramAzhari-adsk (approved):

PR #4149, opened 2025-03-10
Add details to the axis and units import flag doc.

pierrebai-adsk:


PallavAdsk (commented):
Select how units and axis are handled during import. You can choose only one of the following options:
**rotateScale**: Adjusts imported units and axis by rotating and scaling them to fit the scene.
**addTransform**: Adds a parent transform to the imported object, aligning it with the scene while preserving original settings.
**overwritePrefs**: Replaces the current scene's unit and axis preferences with those of the imported file.
Note: Only one option can be selected at a time.

seando-adsk (approved):

PR #4152, opened 2025-03-11
EMSUSD-2218 Light Gizmo workaround for Rect Light / Portal Light

AramAzhari-adsk:
Retrives the width / height parameters of the Rect light / Portal Light directly from the USD prim instead of the GizmoShape node's width/height plugs.


samuelliu-adsk (approved):
LGTM!!

AramAzhari-adsk:
There are internal scanning errors on some of the Linux builds, including Linux interactive which stops the build before the testing step.

PR #4153, opened 2025-03-12
PrimUpdaterManager: Avoid spurious `allowTopologyMod` reference edits

jufrantz:
When editing as Maya a `mayaReference` prim, we noticed that `allowTopologyMod` attributes were being force-enabled on all referenced meshes, cluttering the reference edits list and preventing these attributes from being disabled up-front if needed. 

This PR skips these forced edits unless allowTopologyMod was really disabled by the preceding LockNodeUndoItems execution.


pierrebai-adsk (approved):
I don't know anything about topology mod nodes, but I'll assume this is correct. The code looks fine.

jufrantz:
Thank you @pierrebai-adsk,
allowTopologyMod is a actually built-in attribute on mesh nodes. From my understanding of the code, maya's lockNode command automatically turns it off. This likely caused issues by preventing users from changing the topology of USD meshes (not maya-referenced).

PR #4154, opened 2025-03-14
EMSUSD-2168 Adjust some sizes of the light linking widget to better match the design

pierrebai-adsk:


lib/mayaUsd/resources/ae/usd-shared-components/src/python/usdSharedComponents/collection/includeExcludeWidget.py
```diff
@@ -151,7 +151,7 @@ def __init__(
             "USD_Light_Linking",
             "IncludeListHeight",
             self,
-            defaultSize=Theme.instance().uiScaled(80),
+            defaultSize=Theme.instance().uiScaled(124),
```
barbalt, line 154: Does this also affect the width or only the height?

pierrebai-adsk: Ah...I assumed it was only the height, but I did not check. The height was indeed taller when I tested.

pierrebai-adsk: Only height

barbalt: I'm asking because there's an issue in max where the width can cause problems on the command panel. 
If it's just the width, then there's no problem.

barbalt (approved):

Issue #4125, opened 2025-02-19
UFE4 find all cameras very slow


dj-mcg:
Pixar's current referenceAssembly approach to its set construction workflow results in a single Maya scene containing potentially thousands of ProxyShapes - essentially one per asset in the set. Users have noticed that selecting the Panels > Perspective menu was very slow to populate - on the order of 20 seconds or so - if there are a large number of ProxyShape nodes loaded.

I tracked the slowness down to an O(n^2) operation that UFE4's Find All Cameras logic triggers. It appears that for each node in the scene, this logic makes a call into the `find_` function of each camera handler - one call into the `ProxyShapeCameraHandler` and another call into the `UsdCameraHandler`. Then, for each node, it triggers a call to `findUsdGatewayItems`, which calls `getAllStages`, which traverses the entire scene getting each proxyShape, hence the O(n^2) growth rate.

I haven't been able to track down the call to `findUsdGatewayItems` in the open source mayaUsd code so I assume the call is within UFE itself. I'm curious if this call or the call to `getAllStages` is necessary to perform for each ProxyShape node in the scene, or if it may be cached internally to avoid all this redundant work?

This was also slow in UFE3, but we're just chasing it down now. I can generate an example test scene if needed.

seando-adsk:
The camera finding performance fixes were made in Ufe v4, so it should be much better Maya 2024 than in 2023. 

The `findUsdGatewayItems` is helper function within the `ProxyShapeSceneSegmentHandler.cpp` so you should have access to this source code. Anything with `Usd` in the name would only be found within MayaUsd and not in Ufe or Maya. Neither Ufe nor Maya know anything about Usd. 

@dj-mcg If you can provide a test scene with a repro step that would be great. We can have a look at the performance. 

Sean

santosg87:
I was able to reproduce this on my side, i also noticed it seems we have a similar performance degradation with Exporting cameras, for some reason, which is easier to reproduce from scratch on simple steps. I logged this issue internally as EMSUSD-2199.

dj-mcg:
Hi @seando-adsk - FWIW the performance findings in my original post were from Maya 2024. It may be faster than Maya 2023 but it is still very slow. It may be specific to our Reference Assembly workflow I mentioned, where each Reference Assembly node in a Maya scene (which could easily contain thousands) is actually it's own tiny Usd stage. Traversing through that many stages seems to be what really hamstrings the current discovery process

PR #4158, opened 2025-03-18
EMSUSD-2199 - Fixes a performance issue regarding findGatwayItems() / cameras

AramAzhari-adsk:
EMSUSD-2199 - Fixes a performance issue regarding findGatwayItems() / cameras

This change addresses some of the performance issues mentioned in https://github.com/Autodesk/maya-usd/issues/4125

- Created methods to check / get the list of stages without having to re-generate / update the stage map cache.
- Added an early out in findGatwayItems to avoid unnecessary looping.

### DIFF
--- a/lib/mayaUsd/ufe/Utils.h
+++ b/lib/mayaUsd/ufe/Utils.h
@@ -35,6 +35,7 @@
 #include <ufe/types.h>
 
 #include <cstring> // memcpy
+#include <vector>
 
 UFE_NS_DEF
 {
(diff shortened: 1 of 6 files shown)


lib/mayaUsd/ufe/Utils.h
```diff
@@ -65,6 +66,14 @@ Ufe::Path stagePath(PXR_NS::UsdStageWeakPtr stage);
 MAYAUSD_CORE_PUBLIC
 PXR_NS::TfHashSet<PXR_NS::UsdStageWeakPtr, PXR_NS::TfHash> getAllStages();
 
+//! Return all the USD stages' paths without rebuilding the stage list.
```
seando-adsk, line 69: This calls the new `allStagesPath()` which has a call to `rebuildIfDirty()` - so it does rebuild the list (if needed).

AramAzhari-adsk: I understand, I'll update the comment . Thanks

PR #4159, opened 2025-03-18
HYDRA-1519 - Fixes viewport filter for USD lights

AramAzhari-adsk:
updates the Object Type Exclusion Flag for gizmoGeometryOverride.


samuelliu-adsk (approved):

AramAzhari-adsk:
The preflight failure is due to an internal problem not related to the PR changes.

PR #4160, opened 2025-03-20
Update to MaterialX 1.39.3

JGamache-autodesk:
Starting point is https://github.com/Autodesk/maya-usd/pull/4134 by @boberfly. Updated for recent breaking changes in 1.39.3

Compiles and passes all unit tests on a local Windows build of USD at commit https://github.com/PixarAnimationStudios/OpenUSD/commit/f03754834e28356781de5df73d12b48548ed475f with https://github.com/PixarAnimationStudios/OpenUSD/pull/3568 using MaterialX 1.39.3 without any modification.


lib/mayaUsd/render/MaterialXGenOgsXml/LobePruner.cpp
```diff
@@ -486,13 +486,11 @@ mx::NodeDefPtr LobePrunerImpl::getOrAddOptimizedNodeDef(
     auto optimizedNodeDef
         = _library->addNodeDef(optimizedNodeDefName, "surfaceshader", optimizedNodeName);
     optimizedNodeDef->copyContentFrom(originalNodeDef);
-    optimizedNodeDef->setSourceUri("");
```
JGamache-autodesk, line 489: Not necessary, and was polluting the debug output.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -704,7 +720,12 @@ void _MaterialXData::_FixLibraryTangentInputs(mx::DocumentPtr& mtlxDoc)
         for (mx::NodePtr node : nodeGraph->getNodes()) {
             mx::NodeDefPtr nodeDef = node->getNodeDef();
             if (!nodeDef) {
+#if MX_COMBINED_VERSION >= 13900
+                // Can happen with legacy data. Just continue.
+                continue;
+#else
```
JGamache-autodesk, line 726: The UsdShade networks in USD can still be 1.38. They might have `node:id` that are no longer in the 1.39 library (`ND_swizzle_*`, `ND_normalmap`). So not being able to get a NodeDef for a USD `node:id` must be treated as "business as usual".

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -972,6 +993,78 @@ void _AddMissingTangents(mx::DocumentPtr& mtlxDoc)
     }
 }
 
+#if MX_COMBINED_VERSION >= 13900
```
JGamache-autodesk, line 996: 1.39 added a bitangent input on the normal map node. This allows right and left handed renderers to decide if the bitangent is cross(N, T) or cross(T, N). This means we must add the proper cross product node here to get normal maps to render.

lib/mayaUsd/render/vp2RenderDelegate/material.cpp
```diff
@@ -2717,10 +2810,9 @@ TfToken _RequiresColorManagement(
     const HdMaterialNetwork2& inNet)
 #endif
 {
-    const mx::NodeDefPtr nodeDef = _GetMaterialXData()._mtlxLibrary->getNodeDef(node.nodeTypeId);
```
JGamache-autodesk, line 2720: We were not using that nodedef, and it was pointing to a missing swizzle in two unit tests, make it business as usual here as well to not have all definitions.

Please note that this applies only when looking at id straight from HdMaterialNetwork data prior to sending them to be converted to a MaterialX document. The returned MaterialX document will have been updated to 1.39 and should not contain missing nodes anymore.

lib/usd/translators/shading/mtlxMaterialXSurfaceShaderWriter.cpp
```diff
@@ -26,6 +26,9 @@
 #include <pxr/usd/usdShade/material.h>
 #include <pxr/usd/usdShade/nodeGraph.h>
 #include <pxr/usd/usdUI/nodeGraphNodeAPI.h>
+#if PXR_VERSION >= 2505
+#include <pxr/usd/usdMtlx/tokens.h>
+#endif
```
JGamache-autodesk, line 31: That is on the USD side. In 25.05 we will need this include to get the usdMtlx tokens.

test/lib/mayaUsd/render/vp2RenderDelegate/VP2RenderDelegateMaterialXTest/baseline/DemoQuads_render.png
JGamache-autodesk, line 1: Only change is in 2nd row 1st quad. The normal map texture was incorrectly connected to the normal input of the normal map node. This prevented connecting the bitangent correctly. Fixed in the test file below.

test/lib/mayaUsd/utils/materialx_test_data/Channel1_topo_1_39.mtlx
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<materialx version="1.39">
+  <surfacematerial name="N0" type="material" nodedef="ND_surfacematerial">
+    <input name="surfaceshader" type="surfaceshader" nodename="N1" />
```
JGamache-autodesk, line 4: Need new files since 1.39 does not have the concept of "channels", thus simplifying the test suite a lot.

test/testSamples/MaterialX/DemoQuads.usda
```diff
@@ -581,7 +581,7 @@ def Scope "mtl"
         def Shader "normalmap1"
         {
             uniform token info:id = "ND_normalmap"
-            float3 inputs:normal.connect = </mtl/standard_surface1/image1.outputs:out>
+            float3 inputs:in.connect = </mtl/standard_surface1/image1.outputs:out>
```
JGamache-autodesk, line 584: Incorrect connection fixed.

frohnej-adsk: [Question] Does this .usda file store in some way which MaterialX version it usus? "normal" would be correct in 1.38, right?

I think you mentioned at some point that all USD files using MaterialX 1.39 will explicitly specify that. If nothing is specified, 1.38 is assumed. Is that correct?

If so, do we need an additional change in this file to specify that it's now using 1.39? Or should the name remain as "normal" and tests use the upgrade path to convert it so 1.39 before running?

test/lib/mayaUsd/utils/materialx_test_data/topology_tests_1_39.mtlx
```diff
@@ -0,0 +1,166 @@
+<?xml version="1.0"?>
+<materialx version="1.38" colorspace="lin_rec709">
```
frohnej-adsk, line 2: Should this say 1.39?
```suggestion
<materialx version="1.39" colorspace="lin_rec709">
```

frohnej-adsk (commented):
Very nice!

lib/mayaUsd/render/MaterialXGenOgsXml/Nodes/SurfaceNodeMaya.cpp
```diff
@@ -113,9 +115,22 @@ void SurfaceNodeMaya::emitLightLoop(
...
+        shadergen.emitLine(
+            "ClosureData closureData = ClosureData(CLOSURE_TYPE_REFLECTION, L, V, N, P, occlusion)",
+            stage);
+        shadergen.emitFunctionCall(*bsdf, context, stage);
+    } else {
+        shadergen.emitLineBegin(stage);
+        shadergen.emitOutput(bsdf->getOutput(), true, true, context, stage);
+        shadergen.emitLineEnd(stage);
+    }
+#endif
```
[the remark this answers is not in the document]
JGamache-autodesk: This specific change is for https://github.com/AcademySoftwareFoundation/MaterialX/pull/2205. Modus operandi is to open a [diff between 1.38.10 and 1.39.3](https://github.com/AcademySoftwareFoundation/MaterialX/compare/d43cd93551bb74c791352d96f603a1f9a5ef7ffa...3b79bd735036dc96942836a9d076b6af3bd8fbdf) and check for anything affecting us. Since this file is a copy of [MaterialXGenGlsl/Nodes/SurfaceNodeGlsl.cpp](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenGlsl/Nodes/SurfaceNodeGlsl.cpp) with Maya specific additions, it needed to be updated.

test/lib/mayaUsd/utils/materialx_test_data/topology_tests_1_39.mtlx
JGamache-autodesk: Yes, and no. It does not contain any node that changed between versions, so it is safe to stay at 1.38.

test/testSamples/MaterialX/DemoQuads.usda
JGamache-autodesk: Data without a version is to be assumed 1.38. There is a new [MaterialXConfigAPI](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdMtlx/schema.usda) to be used to identify MaterialX versions. You can see it in use in [this file](https://github.com/PixarAnimationStudios/OpenUSD/blob/d2f309cb30c7b811b34dd5c9b643791472092af6/pxr/usd/usdMtlx/testenv/testUsdMtlxFileFormat.testenv/baseline/usd_preview_surface_gold.usda#L8).

JGamache-autodesk: The version info is propagated thru Hydra, so the material received by the VP2 render delegate for this data is known to be 1.38. The code in hdMtlx will convert the HdMaterialNetwork by first [setting the version ](https://github.com/PixarAnimationStudios/OpenUSD/blob/d2f309cb30c7b811b34dd5c9b643791472092af6/pxr/imaging/hdMtlx/hdMtlx.cpp#L709)it received, than [upgrading it](https://github.com/PixarAnimationStudios/OpenUSD/blob/d2f309cb30c7b811b34dd5c9b643791472092af6/pxr/imaging/hdMtlx/hdMtlx.cpp#L742) to the current one before returning the MaterialX document.

lib/mayaUsd/render/MaterialXGenOgsXml/Nodes/SurfaceNodeMaya.cpp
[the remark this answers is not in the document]
frohnej-adsk: I see, thanks!

test/lib/mayaUsd/utils/materialx_test_data/topology_tests_1_39.mtlx
frohnej-adsk: I see! Seeing `1_39` in the filename and `version="1.38"` in the document looked like a typo, but that makes sense.

test/testSamples/MaterialX/DemoQuads.usda
frohnej-adsk: Got it! That's very good to know, thanks!

I actually got confused here. I thought the input got renamed in the new MaterialX version, but that connection was genuinely wrong independent of the MaterialX version.

frohnej-adsk (approved):

test/lib/mayaUsd/render/vp2RenderDelegate/VP2RenderDelegateMaterialXTest/baseline/DemoQuads_render_1.38.5.png
JGamache-autodesk, line 1: Change is very small in the last row first quad and was sufficiently small to pass under the radar. By fixing the first quad of second row, the difference became sufficient to trigger a comparison failure with Maya 2024 builds. Note the absence of a specular highlight. It looks as if only the diffuse was affected by the normal map it that older MaterialX version.

lanierd-adsk (commented):
Thanks for doing all this. I have questions, mainly why do you duplicate lines with a #if in the code, I guess this is on purpose but why ?
Thank you.

lib/mayaUsd/render/MaterialXGenOgsXml/GlslFragmentGenerator.cpp
```diff
@@ -798,20 +829,34 @@ ShaderPtr GlslFragmentGenerator::generate(
...
             }
         } else {
             const string outputValue = outputSocket->getValue()
                 ? _syntax->getValue(outputSocket->getType(), *outputSocket->getValue())
                 : _syntax->getDefaultValue(outputSocket->getType());
 
-            if (!context.getOptions().hwTransparency && !outputSocket->getType()->isFloat3()) {
+            if (!context.getOptions().hwTransparency &&
+#if MX_COMBINED_VERSION < 13900
+                !outputSocket->getType()->isFloat3()) {
```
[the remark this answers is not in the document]
JGamache-autodesk: Extremely subtle difference. One has `->` the other one has `.` because the returned type used to be by pointer, and is now by value. This leads to a very large number of almost identical lines.

lib/mayaUsd/render/MaterialXGenOgsXml/OgsXmlGenerator.cpp
```diff
@@ -34,6 +35,22 @@ const OGS_TYPE_MAP_T& getOgsTypeMap()
...
 }
+#else
+const OGS_TYPE_MAP_T& getOgsTypeMap()
+{
+    // Data types used by OGS
+    // Delayed initialization to survive C++ init order fiasco.
+    // Keyed by string to survive multiple redefinitions of global symbols.
+    static const OGS_TYPE_MAP_T OGS_TYPE_MAP
+        = { { Type::BOOLEAN.getName(), "bool" },     { Type::FLOAT.getName(), "float" },
+            { Type::INTEGER.getName(), "int" },      { Type::STRING.getName(), "int" },
```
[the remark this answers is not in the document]
JGamache-autodesk: Yes. No string type in GLSL, HLSL, or other hardware shading languages. Pure strings never reach the OpenGL shadergen. We do have a few enumerated types in MaterialX that have type string in MaterialX, but end up being mapped to an integer in the shader gen code. For example `uaddressmode` in `ND_image_*`  will map to an integer in the range 0-3 because there are four string values.

lib/mayaUsd/render/MaterialXGenOgsXml/GlslFragmentGenerator.cpp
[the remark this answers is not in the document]
lanierd-adsk: Thanks for your reply. I missed that.

lanierd-adsk (approved):

JGamache-autodesk:
@seando-adsk ready for merge.

PR #4164, opened 2025-03-21
EMSUSD-1668 Added export splines anim for camera attributes

barbalt:
# EMSUSD-1668 | As a user, I want the ability to export camera animation curves

The original spline proposal can be read here: https://github.com/PixarAnimationStudios/OpenUSD-proposals/blob/main/proposals/spline-animation/README.md#feature-reduction
The main thing to point out is:
```
Maya tangents are specified by height and length.

The units of height are the same as the units of values.

Height and length are both specified multiplied by 3; e.g. if a tangent vector is 1.5 units in the time dimension, its length is recorded as 4.5.

Heights are positive for upward-sloping post-tangents, and negative for upward-sloping pre-tangents.
```
As it is described here: https://github.com/PixarAnimationStudios/OpenUSD-proposals/blob/main/proposals/spline-animation/README.md#maya-tangents

Conversion helper functions are provided by USD here: https://github.com/PixarAnimationStudios/OpenUSD/blob/319925ecc74c72bd3d11c74052d64ce7bd66f44a/pxr/base/ts/tangentConversions.cpp#L200

# What attributes are being exported?

At the moment, we chose to export only the camera attributes as the initial test. The only attribute not being exported is the clipping range because it uses a tuple and that's not supported by usd splines yet.

# What USD version is supported?

OpenUSD 24.11 and above has the required spline support for this initial implementation

# What changed?
A new export option has been added: Animation Type (timesample, both and curves). 
- Timesample is the default and everything should export the same as it was before those changes
- Both: the supported attributes will be exported using both timesamples and spline animation. In those cases, Usd will prefer to use timesamples when evaluating the animation
- Curves: only splines will be written for the given attribute.

The new export option can be used as a parameter to the export command. For example: `animationType="both"`


pierrebai-adsk (approved):
Great work!

barbalt:
The one failing build is due to tests timing out.

PR #4172, opened 2025-03-31
LOOKDEVX-4039 - Add Outliner icons for NodeGraphs.

frohnej-adsk:
New Outliner icon matching the NodeGraph icon in LookdevX.
image


feldstj (approved):

frohnej-adsk:
All builds in the preflight were green except for one Linux build where two of the tests timed out:
```
    The following tests FAILED:
    	130 - testVP2RenderDelegateMaterialX (Timeout)
    	254 - testUsdImportXforms (Timeout)
```

That Linux build took extremely long in general so the default timeout for the overall preflight expired as well.

These issues shouldn't be related to the PR, so I'm marking it as ready-for-merge.

PR #4174, opened 2025-04-02
Fix layer editor path display and optimize rebuild model

adubois-spin:
There are two requests in this pull request:

    1. At our studio the layer editor sometimes displays the wrong path when putting the mouse on the layer. Adding the right resolver context around the call and having a switch that takes the real path instead of the identifier solve the problem.

    2. Rebuilding the model all the time takes time and only doing on a subset of cases improve the user experience.


lib/usd/ui/layerEditor/layerTreeModel.cpp
```diff
@@ -300,13 +302,7 @@ void LayerTreeModel::rebuildModel(bool refreshLockState /*= false*/)
...
 
     beginResetModel();
-
-    //  Note: do *not* call clear() here! Unfortunately, clear() itself calls,
-    //        beginResetModel() and endResetModel(). Qt does not detect the nested
-    //        begin/end/ So calling clear() would make the layer manager flicker
-    //        to be empty for a brief time.
-    if (rowCount() > 0)
-        removeRows(0, rowCount());
+    clear();
```
pierrebai-adsk, line 305: Is this change required? Like the comments said, clear() would cause multiple notifications of changes to be sent and cause flickers. Any reason why changing it to call clear?

lib/usd/ui/layerEditor/mayaCommandHook.cpp
```diff
@@ -114,7 +114,10 @@ void MayaCommandHook::insertSubLayerPath(UsdLayer usdLayer, Path path, int index
     cmd = "mayaUsdLayerEditor -edit -insertSubPath ";
     cmd += std::to_string(index);
     cmd += quote(path);
-    cmd += quote(usdLayer->GetIdentifier());
+    // Here we choose the identifier only if the layer is anonymous
+    std::string layerPath{usdLayer->IsAnonymous() ? usdLayer->GetIdentifier() :
+                                                    usdLayer->GetRealPath()};
+    cmd += quote(layerPath);
```
pierrebai-adsk, line 120: For cases like this where the same pattern is ued multiple times, it is better to put it in a function to avoid the repetition. Maybe a function named getLayerPathForEditorCommand()

adubois-spin: Yes I agree a function would be better.
We did not make one because we didn't know where in the package it would be best defined.

seando-adsk:
@adubois-spin Did you ever get a chance to address the code review feedback provided by Pierre?

PR #4176, opened 2025-04-02
Fix get stronger layer layers not found

adubois-spin:
This request is about corner cases regarding allowing the edit of a layer metadata.

getStrongerLayer has corner cases for which it returns a null layer instead of one of the two passed as argument.
The idea is to allow editing in those cases.

This was our first way of solving our issue where the fact that the targetLayer could not be found was making the problem arise.
We then realized that the reason why our layer could not be found when iterating through the sub layers is because this traversal was not done with the current resolver context.

Both fix are in this branch. We would be glad if the resolver context could be at least added to the call. We think this resolver context problem might occur in other parts of maya-usd.


seando-adsk:
@adubois-spin Thanks for your interest in contributing to our maya-usd repo, before we can accept PR from a new contributor, we need you to fill, sign and email back the proper CLA (individual or corporate). See [CONTRIBUTING.md](https://github.com/Autodesk/maya-usd/blob/dev/doc/CONTRIBUTING.md) for details about CLA.

adubois-spin:
> @adubois-spin Thanks for your interest in contributing to our maya-usd repo, before we can accept PR from a new contributor, we need you to fill, sign and email back the proper CLA (individual or corporate). See [CONTRIBUTING.md](https://github.com/Autodesk/maya-usd/blob/dev/doc/CONTRIBUTING.md) for details about CLA.

@seando-adsk Thank you for your quick answer. Okay let me see with my studio so we sign this.

lib/usdUfe/ufe/Utils.cpp
```diff
@@ -1087,7 +1087,15 @@ bool isPropertyMetadataEditAllowed(
...
-    bool allowed = (strongestLayer == targetLayer);
+    /*
+     * Here there are two cases in which we want to allow the metadata edit:
+     *     1. the target layer of the edit is stronger than the strongest layer having an opinion on the matter
+     *     2. the strongest layer having an opinion on the matter is not stronger than the target layer
+     *
+     * The 2nd case appears mostly if the getStrongerLayer function fails to find which layer is stronger,
+     * returning an empty layer reference.
+    */
+    bool allowed = (strongestLayer != topAuthoredLayer);
```
pierrebai-adsk, line 1098: I don't understand the check.It does not even use the targeted layer. Also, it is confusing because the inequality comparison to the topAuthoredLayer seems opposite of what we would want, but is done as a "hidden" check that strongestLayer is null.

The old check must stay but other checks could be added if needed.

The correct way to write your change would be something like:

```
bool allowed = (strongestLayer == targetLayer || !strongestLayer);
```

... but even that I'm not sure we want, IIRC, the reason we didn't allow editing when the strongest layer is null was that because this indicates the layer is a non-local layer and we did not want to allow modifying references. I think this could be changed but I'll confirm with designers.

PR #4177, opened 2025-04-02
EMSUSD-2278 - Creating or parenting a camera under a USD def will crash Maya

samuelliu-adsk:


seando-adsk (changes requested):
There should be a new unit test that does exactly what the JIRA item says. Test which creates  a camera under a def. Please add the test first without this change and verify that the test fails (it should crash). Then add your new fix and the test should pass.

test/lib/mayaUsd/render/vp2RenderDelegate/testVP2RenderDelegateCameras.py
```diff
@@ -117,12 +119,33 @@ def _RunTest(self):
...
+        stage.GetRootLayer().Save()
+        proxyShape = cmds.createNode('mayaUsdProxyShape')
+        cmds.setAttr('mayaUsdProxyShape1.filePath', usdFilePath, type='string')
+        ufeItem = ufeUtils.createItem('|mayaUsdProxy1|mayaUsdProxyShape1,/parent')
+        object3d = ufe.Object3d.object3d(ufeItem)
+        object3d.setVisibility(False)
+        self.assertSnapshotClose('%s_invisibile.png' % (self._testName))
+        object3d.setVisibility(True)
+        self.assertSnapshotClose('%s_visibile.png' % (self._testName))
+
```
samuelliu-adsk, line 148: Added camera creation under scope and def. Also added tests to test camera visibility if its parent is set to invisible

samuelliu-adsk:
My commit history is totally messed up.. Will create a new one

PR #4186, opened 2025-04-10
Properly handle deleted attributes

JGamache-autodesk:
We currently changed the initial AttributeAdded inside a UFE SetValue into a AttributeValueChanged.
We now do the same for the final ArreibuteRemoved.

### DIFF
--- a/lib/usdUfe/ufe/StagesSubject.cpp
+++ b/lib/usdUfe/ufe/StagesSubject.cpp
@@ -168,14 +168,23 @@ void sendAttributeChanged(
         }
     } break;
     case AttributeChangeType::kRemoved: {
-        // Special case when Undoing a visibility change, the notice.GetChangedInfoOnlyPaths()
-        // does not contain the change, hence handling visibility notification in re-sync path.
-        if (changedToken == UsdGeomTokens->visibility) {
-            Ufe::VisibilityChanged vis(ufePath);
-            notifyWithoutExceptions<Ufe::Object3d>(vis);
+        if (UsdUfe::InSetAttribute::inSetAttribute()) {
+            notifyWithoutExceptions<Ufe::Attributes>(
+                Ufe::AttributeValueChanged(ufePath, changedToken.GetString()));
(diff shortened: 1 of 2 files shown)


lib/usdUfe/ufe/StagesSubject.cpp
```diff
@@ -168,14 +168,23 @@ void sendAttributeChanged(
         }
     } break;
     case AttributeChangeType::kRemoved: {
-        // Special case when Undoing a visibility change, the notice.GetChangedInfoOnlyPaths()
-        // does not contain the change, hence handling visibility notification in re-sync path.
-        if (changedToken == UsdGeomTokens->visibility) {
-            Ufe::VisibilityChanged vis(ufePath);
-            notifyWithoutExceptions<Ufe::Object3d>(vis);
+        if (UsdUfe::InSetAttribute::inSetAttribute()) {
```
JGamache-autodesk, line 171: Identical to how kAdded is treated just above.

test/lib/ufe/testAttribute.py
```diff
@@ -1311,40 +1311,41 @@ def testObservationWithFineGrainedNotifications(self):
...
         globalObs.assertNotificationCount(self, numValue = 6)
 
-        # Undo-ing the modification which created the USD specs is a little
-        # different in USD, but from Ufe we should just still see one notification.
+        # The spec creation done on the first setValue was filtered so it
+        # appeared only as a single value change. Symetrically, the unto
+        # of the spec creation should be filtered as well.
         cmds.undo()
-        ball34Obs.assertNotificationCount(self, numRemoved = 1, numValue = 6)
+        ball34Obs.assertNotificationCount(self, numValue = 7)
```
JGamache-autodesk, line 1318: Makes more sense that if numAdded ended up being zero, then numRemoved should follow.

pierrebai-adsk (approved):

AramAzhari-adsk (approved):

PR #4188, opened 2025-04-14
Update diffCore.cpp

metaremi:
Fix dllimport applied to non-inline function error

### DIFF
--- a/lib/usdUfe/utils/diffCore.cpp
+++ b/lib/usdUfe/utils/diffCore.cpp
@@ -1253,7 +1253,7 @@ bool compareArray3Dto4D(
         return false;
     }
 #ifdef __AVX2__
-    const f128 eps4 = splatf4f(eps);
+    const f128 eps4 = splat4f(eps);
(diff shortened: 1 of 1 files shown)


metaremi:
splatf4f -> splat4f

metaremi:
@seando-adsk

I am Rémi ARNAUD from Meta.
The CLA was signed/sent last week.

lib/usdUfe/utils/diffCore.cpp
```diff
@@ -1253,7 +1253,7 @@ bool compareArray3Dto4D(
         return false;
     }
 #ifdef __AVX2__
-    const f128 eps4 = splatf4f(eps);
+    const f128 eps4 = splat4f(eps);
```
AramAzhari-adsk, line 1256: I presume the `splatf4f` was a typo? I couldn't find it in our code base.

metaremi: indeed, looks like a typo to me.

PR #4190, opened 2025-04-14
EMSUSD-2232 - Avoids rebuilding cache when trying to getProxyShape

AramAzhari-adsk:
This update the getProxyShape() util function to avoid rebuilding cache by default.

### DIFF
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -273,8 +273,8 @@ MayaUsdProxyShapeBase* getProxyShape(const Ufe::Path& path)
     if (!TF_VERIFY(!path.empty())) {
         return nullptr;
     }
-
-    return UsdStageMap::getInstance().proxyShapeNode(path);
+    const bool rebuildCacheIfNeeded = false;
+    return UsdStageMap::getInstance().proxyShapeNode(path, rebuildCacheIfNeeded);
 }
 
 SdfPath getProxyShapePrimPath(const Ufe::Path& path)
--- a/lib/mayaUsd/ufe/Utils.cpp
+++ b/lib/mayaUsd/ufe/Utils.cpp
@@ -273,8 +273,15 @@ MayaUsdProxyShapeBase* getProxyShape(const Ufe::Path& path)
     if (!TF_VERIFY(!path.empty())) {
         return nullptr;
     }
-    const bool rebuildCacheIfNeeded = false;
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -273,8 +273,8 @@ MayaUsdProxyShapeBase* getProxyShape(const Ufe::Path& path)
     if (!TF_VERIFY(!path.empty())) {
         return nullptr;
     }
-
-    return UsdStageMap::getInstance().proxyShapeNode(path);
+    const bool rebuildCacheIfNeeded = false;
+    return UsdStageMap::getInstance().proxyShapeNode(path, rebuildCacheIfNeeded);
```
seando-adsk, line 277: What happens if the stage map is dirty and we cannot find the proxy shape (I guess this would be the case if the proxy shape was just created).

AramAzhari-adsk: @seando-adsk  The current scenarios that this function is getting called are time not sensitive so it should already be ready by the time this is getting called. 

Here are the scenarios:

PrimUpdaterManager:
- mergeToUsd
- editAsMaya
- discardPrimEdits
- duplicate
- duplicateToMaya
- duplicateToUsd

LayerEditor:
- getChildProxyShape
- selectionChanged

My understanding is that the `rebuildCacheIfNeeded` was added to `proxyShapeNode` to turn off the force rebuild for one case without having to assess if all other cases would benefit from it or not. So far, it seems those are all working.

We have a choice of isolating this to only `StageSelectorWidget::selectionChanged()` but I think the change is minimal enough since all the unit tests are passing we can let it soak and if any regressions occur we can still do that.

AramAzhari-adsk: @seando-adsk Should we display a debug message if it the result is null ?

seando-adsk (approved):

lib/mayaUsd/ufe/Utils.cpp
seando-adsk: No, I don't think a message is needed.

AramAzhari-adsk:
Fails on Linux Interactive for these tests which are unrelated:

```
17:16:02      The following tests FAILED:
17:16:02      	127 - testProxyShapeDrawPerformance (Timeout)
17:16:02      	130 - testVP2RenderDelegateMaterialX (Timeout)
17:16:02      	254 - testUsdImportXforms (Timeout)
17:16:02      Errors while running CTest
```

Re-running just in case.

PR #4202, opened 2025-04-28
EMSUSD-2232 - Disables cache rebuild due to layer editor

AramAzhari-adsk:
This is an improvement on https://github.com/Autodesk/maya-usd/pull/4190 .
There are many cases where the proxy shape is null and will still be null even after a cache rebuild. This causes slowdowns.

To address the layer editor issue in particular, we are aware that the layer editor should not force a stage cache rebuild because by the time the layer editor is trying to respond to selection changes the proxy node list has already been updated.

This change isolates the cache disable to only layer editor and will assume all the other cases still need a cache rebuild. 

If other performance issues are noticed elsewhere, they can be investigated individually.


seando-adsk (approved):
Okay this is better. Only disable rebuild from the LE

PR #4211, opened 2025-05-06
LOOKDEVX-4368 - `UsdAttributeUInt`: unsigned int ufe type support

selvatm-autodesk:
`usdUfe` : Add support for Unsigned Integer Usd Attribute.

### DIFF
--- a/test/lib/ufe/testAttribute.py
+++ b/test/lib/ufe/testAttribute.py
@@ -565,12 +565,10 @@ class AttributeTestCase(unittest.TestCase):
         self.runUndoRedo(ufeAttr, ufeAttr.get()+1)
 
         # # Run test using Maya's setAttr command.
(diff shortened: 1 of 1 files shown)


seando-adsk (changes requested):
You also need to add a `testAttributeUInt` test in [testAttribute.py](https://github.com/Autodesk/maya-usd/blob/dev/test/lib/ufe/testAttribute.py)

JGamache-autodesk (changes requested):
Indeed, only thing missing is the unit test. Big issue however is that none of the USD schemas have a `uint` parameter. You will have to artificially inject one into an existing test scene, or create a mini test scene with a
```
    uint testUInt = 42
```
Somewhere in a prim.

selvatm-autodesk:
Thanks, looking around for creating the resource for the test!

JGamache-autodesk (approved):

seando-adsk (approved):

PR #4216, opened 2025-05-09
EMSUSD-2210	 As a user, I want the ability to import camera animation curves

barbalt:


barbalt:
Tests are failing with unrelated changes

lib/mayaUsd/fileio/translators/translatorCamera.cpp
```diff
@@ -26,6 +26,10 @@
 #include <pxr/base/gf/vec2f.h>
 #include <pxr/base/tf/token.h>
 #include <pxr/base/tf/type.h>
+#if PXR_VERSION >= 2411
```
pierrebai-adsk, line 29: I know it is more work, but in the past we use CMake to create symbols like USD_HAS_SPLINE or somethign similar, so that the check for USD version is done in CMake instead of C++ and C++ uses more meaningful #ifdef.

lib/mayaUsd/fileio/translators/translatorCamera.cpp
```diff
@@ -204,13 +208,173 @@ static bool _CreateAnimCurveForPlug(
     return true;
 }
 
+#if PXR_VERSION >= 2411
+static MFnAnimCurve::InfinityType _ConvertUsdExtrapolationTypeToMaya(TsExtrapMode usdExtrapolation)
+{
```
pierrebai-adsk, line 213: I won't review the details since IDK about USD splines and Maya anim curve much.

lib/mayaUsd/fileio/translators/translatorCamera.cpp
```diff
@@ -204,13 +208,173 @@ static bool _CreateAnimCurveForPlug(
...
+}
+
+static bool _CreatePlugSpline(
+    MPlug&                    plug,
+    TsSpline                  spline,
+    UsdMayaPrimReaderContext* context,
+    const MDistance::Unit     convertToUnit = MDistance::kMillimeters)
+{
+    auto valueType = spline.GetValueType();
+    if (valueType != Ts_GetType<float>()) {
```
pierrebai-adsk, line 240: I'm going to guess USD only supports floats by definition? Or is double support coming in the future or was decided not to be supported?

lib/mayaUsd/fileio/translators/translatorCamera.cpp
```diff
@@ -204,13 +208,173 @@ static bool _CreateAnimCurveForPlug(
...
+static MFnAnimCurve::TangentType _ConvertUsdTanTypeToMayaTanType(TsInterpMode usdTanType)
+{
+    switch (usdTanType) {
+    case TsInterpMode::TsInterpHeld: return MFnAnimCurve::TangentType::kTangentStep;
+    case TsInterpMode::TsInterpLinear: return MFnAnimCurve::TangentType::kTangentLinear;
+    default: return MFnAnimCurve::TangentType::kTangentAuto;
+    }
+}
+
+static bool _CreatePlugSpline(
```
pierrebai-adsk, line 233: I suppose we will move this elsewhere and generalize it when we support anim curve for other objects?

barbalt: Exactly. The plan will probably be take in the usd type to be usd on the spline as an argument

lib/mayaUsd/fileio/translators/translatorCamera.cpp
```diff
@@ -204,13 +208,173 @@ static bool _CreateAnimCurveForPlug(
...
+    MIntArray    weightsLockedArray(numKnots, 0);
+    MDoubleArray tangentInXArray(numKnots, 0.0);
+    MDoubleArray tangentInYArray(numKnots, 0.0);
+    MDoubleArray tangentOutXArray(numKnots, 0.0);
+    MDoubleArray tangentOutYArray(numKnots, 0.0);
+
+    unsigned int knotIdx = 0;
+    auto         preTanType = MFnAnimCurve::TangentType::kTangentFixed;
+    for (const TsKnot& knot : knots) {
+        float value;
```
pierrebai-adsk, line 273: I'm a bit wary of uninitialzed variables. Maybe default to zero? I know currently all code paths below potentially initialize it. (Although IDK if GetValue() can fail?)

lib/mayaUsd/fileio/translators/translatorCamera.cpp
barbalt: The export was merged as this, so I figured I would do the same for import. 
I can change this in my next pr where I move those functions to splineUtils.h file?

Question, the float implementation works for usd 24.11, but in 25.05 there will some other things that will only be supported with that. How should that be approached with the versioning that you are talking about?

pierrebai-adsk (dismissed):
Minor stuff, change them if you agree.

lib/mayaUsd/fileio/translators/translatorCamera.cpp
barbalt: 24.11 only supports floating point numbers, but 25.05 will have other types

pierrebai-adsk (approved):

PR #4217, opened 2025-05-12
EMSUSD-2265 : Mtlx-USD export should add ND_geompropvalue node to the image node during convertion

derlerk:
Add geomprop node when exporting MaterialX containing images not connected to a geomprop node.
Surface up the texcoord attribute to the material to allow the _UVMappingManager to connect the proper UV channel.

### DIFF
--- a/lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
+++ b/lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
@@ -715,7 +715,6 @@ public:
                     selList.getPlug(0, uvNamePlug);
 
                     MFnMesh meshFn(uvNamePlug.node());
-                    MGlobal::displayInfo(meshFn.name());
                     TfToken shapeName(meshFn.name().asChar());
                     if (!exportedShapes.count(shapeName)) {
                         continue;
--- a/lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
+++ b/lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
@@ -593,7 +593,8 @@ public:
     _UVMappingManager(
         const UsdShadeMaterial&                                  material,
         const UsdMayaShadingModeExportContext::AssignmentVector& assignmentsToBind,
-        const UsdMayaJobExportArgs&                              exportArgs)
+        const UsdMayaJobExportArgs&                              exportArgs,
+        const std::string&                                       materialType)
         : _material(material)
         , _exportArgs(exportArgs)
     {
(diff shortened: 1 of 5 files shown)


lib/usd/translators/shading/mtlxMaterialXSurfaceShaderWriter.cpp
```diff
@@ -377,6 +515,12 @@ void _AddNode(
     for (auto input : node->getInputs()) {
         _AddShaderInput(input, shader, parentPath, shaderUfePath, stage);
     }
+
+    // Special case for Image Nodes.
+    // A geompromvalue node might be needed.
+    if (node->getCategory() == "tiledimage" || node->getCategory() == "image") {
+        _AddGeompropValueNode(node, stage, parentPath, shader);
```
JGamache-autodesk, line 522: Good enough for a start, but definitely not generic enough and missing `UsdUVTexture`. The best fix would browse the MaterialX library for nodes containing an input that has a "defaultgeomprop" set to "UV0". Those would be the ramp and pattern nodes.

test/lib/usd/translators/CMakeLists.txt
```diff
@@ -102,6 +102,13 @@ set(TEST_SCRIPT_FILES
     testUsdMayaAdaptorUndoRedo.py
 )
 
+# This test is disabled by default since it requires the LookdevX plugin
+#if(MAYA_APP_VERSION VERSION_GREATER_EQUAL 2025)
+#    list(APPEND TEST_SCRIPT_FILES        
+#        testUsdExportMaterialXSurfaceShader.py
+#    )
+#endif()
```
JGamache-autodesk, line 110: ????

lib/usd/translators/shading/mtlxMaterialXSurfaceShaderWriter.cpp
derlerk: Good point. Will do that!

lib/mayaUsd/fileio/shading/shadingModeExporterContext.cpp
```diff
@@ -652,57 +653,92 @@ class _UVMappingManager
...
+                    //           |mesh|mesh.uvSet[0].uvSetName
+                    //       the best way to prevent confusion is to move to the object model
+                    //       immediately and process from there.
+                    MSelectionList selList;
+                    selList.add(uvSetRef.c_str());
+                    MPlug uvNamePlug;
+                    selList.getPlug(0, uvNamePlug);
+
+                    MFnMesh meshFn(uvNamePlug.node());
+                    MGlobal::displayInfo(meshFn.name());
```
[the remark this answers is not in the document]
derlerk: Indeed, removing

test/lib/usd/translators/CMakeLists.txt
derlerk: Locally, I can get it to run. But on the preflight system is just fails to find the LookdevX plugin. Am I missing something?

JGamache-autodesk: Preflight has only a Maya rundev and the freshly built MayaUSD. LookdevX will not be available. Needs to be tested somewhere else, probably in LookdevX itself.

JGamache-autodesk (approved):

PR #4218, opened 2025-05-13
Use new wireframe draw modes

dj-mcg:
Hydra was applying alpha values to the wireframe representation of meshes, which was causing issues with selection highlighting. A new draw mode was added and adopted in Pixar's batch renderer


vlasovi (approved):
Looks good

PR #4219, opened 2025-05-13
Fixes for building with QT_DISABLE_DEPRECATED_BEFORE

dj-mcg:
The maya_open_source Qt code already has version checks for Qt 6,
        but we need to follow the v6 branch when we are building with 5.15
        deprecation turned on. Adding this define to the checks as a minimal
        way to get this to compile internally with deprecation turned on,
        without changing the code in a way that makes it hard to merge
        updates from upstream.


dj-mcg:
Pushing in case this might be helpful elsewhere, but let me know if you'd prefer this to be an internal patch for Pixar

AramAzhari-adsk (approved):
LGTM

PR #4227, opened 2025-05-30
Added support for USD Domelight_1

AramAzhari-adsk:


lib/mayaUsd/fileio/translators/translatorRfMLight.cpp
```diff
@@ -67,6 +71,7 @@ TF_DEFINE_PRIVATE_TOKENS(
     ((DiskLightMayaTypeName, "PxrDiskLight"))
     ((DistantLightMayaTypeName, "PxrDistantLight"))
     ((DomeLightMayaTypeName, "PxrDomeLight"))
+    ((DomeLight_1MayaTypeName, "PxrDomeLight_1"))
```
AramAzhari-adsk, line 74: @dj-mcg I couldn't find sufficient information on domelight_1 tokens. The equivalent PxrDomeLight exist in [here](https://github.com/PixarAnimationStudios/OpenUSD/blob/b9282cb274d111878707baff97d4223a81ef23d8/pxr/imaging/hdx/taskController.cpp#L74) but no indication of `PxrDomeLight_1` 

Is this token string assumption correct ?

samuelliu-adsk (approved):

lib/mayaUsd/fileio/translators/translatorRfMLight.cpp
dj-mcg: Oh thanks for updating the RfM translator! This is fine for now, I can check with the USD folks to see if we should instead just deprecate the old dome light with _1

PR #4228, opened 2025-06-02
EMSUSD-2300 : Support Import for spline attribute of USD Lights.

derlerk:
Fix bugs in light spline export.
Add support for light attribute spline import.

### DIFF
--- a/lib/mayaUsd/fileio/translators/translatorCamera.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorCamera.cpp
@@ -212,6 +212,10 @@ static bool _TranslateAnimatedUsdAttributeToPlug(
     UsdMayaPrimReaderContext*    context,
     const MDistance::Unit        convertToUnit = MDistance::kMillimeters)
 {
+    if (args.GetTimeInterval().IsEmpty()) {
+        return false;
+    }
+
 #if PXR_VERSION >= 2411
     // If the attribute has a spline, we ignore time samples.
     if (usdAttr.HasSpline()) {
@@ -221,10 +225,6 @@ static bool _TranslateAnimatedUsdAttributeToPlug(
         }
     }
 #endif
-
-    if (args.GetTimeInterval().IsEmpty()) {
-        return false;
-    }
 
     MTimeArray   timeArray;
     MDoubleArray valueArray;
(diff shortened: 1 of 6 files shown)


lib/mayaUsd/fileio/translators/translatorCamera.cpp
```diff
@@ -212,6 +212,10 @@ static bool _TranslateAnimatedUsdAttributeToPlug(
     UsdMayaPrimReaderContext*    context,
     const MDistance::Unit        convertToUnit = MDistance::kMillimeters)
 {
+    if (args.GetTimeInterval().IsEmpty()) {
```
derlerk, line 215: Moved this up. we want to import animation only if this is not empty.

barbalt: Spline animation aren't tied to this interval

lib/mayaUsd/fileio/translators/translatorLight.cpp
```diff
@@ -559,32 +528,85 @@ static bool _ReadSpotLight(
...
 
-    float UsdConeAngle = 1.f;
-    shapingAPI.GetShapingConeAngleAttr().Get(&UsdConeAngle, timeCode);
-    float coneSoftness = 0.f;
-    shapingAPI.GetShapingConeSoftnessAttr().Get(&coneSoftness, timeCode);
-
-    float penumbraAngle = UsdConeAngle * coneSoftness;
-    float mayaConeAngle = 2.f * (UsdConeAngle - penumbraAngle);
+        MPlug penumbraAnglePlug = depFn.findPlug(_tokens->PenumbraAnglePlugName.GetText(), &status);
+        success &= (status == MS::kSuccess);
```
barbalt, line 563: why do we need this bitwise here?

derlerk: Just bool & operand, equivalent to "success = success && (status == MS::KSuccess)".
That's the logic that was used before, kept it going.

lib/mayaUsd/fileio/translators/translatorCamera.cpp
derlerk: Yes the time interval is useless for processing the splines, but this is how to retrieve from the options if the user wants animation

lib/mayaUsd/fileio/translators/translatorLight.cpp
barbalt: Those aren't equivalent, but I see what you meant that it was being used before. So I guess that's fine

PR #4229, opened 2025-06-04
EMSUSD-2491 Export Xform transformations with Animation Curves

barbalt:


lib/mayaUsd/fileio/transformWriter.cpp
```diff
@@ -164,18 +169,40 @@ void UsdMayaTransformWriter::_ComputeXformOps(
...
+                animChannel.setXformOp(value, matrix, usdTime, valueWriter);
+            }
+        }
+
+#if USD_SUPPORTS_SPLINES
+        if (animChannel.valueType == _ValueType::Value && usdTime == UsdTimeCode::Default()) {
+            if (UsdGeomXformOp::GetPrecisionFromValueTypeName(
+                    animChannel.op.GetAttr().GetTypeName())
+                == UsdGeomXformOp::PrecisionDouble) {
+                UsdMayaSplineUtils::WriteSplineAttribute<double, double>(
```
derlerk: Maybe just specify the type once since it's the same

lib/mayaUsd/fileio/transformWriter.cpp
```diff
@@ -164,18 +169,40 @@ void UsdMayaTransformWriter::_ComputeXformOps(
...
             }
 
-            animChannel.setXformOp(value, matrix, usdTime, valueWriter);
+            if (_GetExportArgs().animationType != UsdMayaJobExportArgsTokens->curves) {
+                animChannel.setXformOp(value, matrix, usdTime, valueWriter);
+            }
+        }
+
+#if USD_SUPPORTS_SPLINES
+        if (animChannel.valueType == _ValueType::Value && usdTime == UsdTimeCode::Default()) {
```
pierrebai-adsk, line 178: I see you check the export args for animationType curves above, don't you need to check it here too to *not* export curves if not requested in the args?

barbalt: I can change it, but it's expected that `animChannel.valueType` will only be `_ValueType::Value` when exporting usd splines

lib/mayaUsd/fileio/transformWriter.cpp
```diff
@@ -295,7 +323,78 @@ bool UsdMayaTransformWriter::_GatherAnimChannel(
         } else {
             return false;
         }
+#if USD_SUPPORTS_SPLINES
+        // when using usd spline animation, we need to break down the transform elements as it's
+        // smallest components. USD spline only supports floating point numbers and vec2.
+        if (animType != UsdMayaJobExportArgsTokens->timesamples) {
```
pierrebai-adsk, line 329: Minor: above you check for Tokens->Curve, here you check for not Tokens->timeSamples, which I found a bit confusing at first.

barbalt: I'm checking for timeSamples because the user can select `both` and `curves`. Both those options should allow the if to be true. In this case, we don't want to export splines only if they selected `timeSamples` only

lib/mayaUsd/fileio/utils/splineUtils.h
```diff
@@ -51,14 +53,18 @@ struct UsdMayaSplineUtils
      * processes its keyframes, and converts the tangent and value data into a USD knot
      * map.
      *
+     * Template parameters:
+     * T - The type of the value stored in the knot tangent (e.g., float, double).
+     * V - The type of the value stored in the knot value (e.g., float, double).
+     *
```
pierrebai-adsk, line 59: MInor: maybe call the template types TANGENT and VALUE then?

pierrebai-adsk: Oh T -> tangent, V -> value, got it.... still, beside that C++ likes so much their template types to be one letter, maybe use the full names...

pierrebai-adsk (approved):
OK, got your answer... great work!

test/lib/usd/translators/testUsdExportSplineXforms.py
```diff
@@ -0,0 +1,330 @@
...
+        cmds.usdExport(mergeTransformAndShape=True,
+            file=cls._usdFilePath,
+            shadingMode='none',
+            animationType='curves',
+            frameRange=(cls.START_TIMECODE, cls.END_TIMECODE))
+
+        cls._stage = Usd.Stage.Open(cls._usdFilePath)
+
+    @unittest.skipUnless(Usd.GetVersion() >= (0, 25, 5), 'Splines transforms are only supported in USD 0.25.05 and later')    
+    def testExportSphereTransformSplines(self):
```
derlerk: I would remove "Sphere" from the test name

pierrebai-adsk (approved):

derlerk (approved):

PR #4231, opened 2025-06-05
EMSUSD-2535 : Default Camera Attributes are not exported when using Animation Curve option

derlerk:
Make sure that camera write their attributes even if not animated while exporting splines.


barbalt (dismissed):
Nice job. Really like the cleanup

barbalt (dismissed):

barbalt (approved):

barbalt (approved):

PR #4232, opened 2025-06-05
EMSUSD-2492 Import Xform transformations with Animation Curves

barbalt:
When the USD Import comes across transform animations using animation curves, import those curves for better editing experience.


lib/mayaUsd/fileio/utils/splineUtils.h
```diff
@@ -194,14 +194,130 @@ struct UsdMayaSplineUtils
         return spline;
     }
 
-    template <typename TANGENT>
+    template <typename T>
     static bool WriteUsdSplineToPlug(
```
barbalt, line 198: This function was just moved from the .cpp file. 
The only difference is that, for the value type and tangent type, It's now using T, instead of a fixed float

lib/mayaUsd/fileio/utils/xformStack.cpp
```diff
@@ -529,6 +528,67 @@ UsdMayaXformStack::OpClassList UsdMayaXformStack::FirstMatchingSubstack(
...
+            UsdMayaXformOpClassification(
+                UsdMayaXformStackTokens->translateZ, UsdGeomXformOp::TypeTranslateZ),
+            UsdMayaXformOpClassification(
+                UsdMayaXformStackTokens->pivot, UsdGeomXformOp::TypeTranslate),
+            UsdMayaXformOpClassification(
+                UsdMayaXformStackTokens->rotatePivotTranslate, UsdGeomXformOp::TypeTranslate),
+            UsdMayaXformOpClassification(
+                UsdMayaXformStackTokens->rotatePivot, UsdGeomXformOp::TypeTranslate),
+            UsdMayaXformOpClassification(
+                UsdMayaXformStackTokens->rotate, UsdGeomXformOp::TypeRotateXYZ),
```
pierrebai-adsk: Just want to  point out that this will match XYZ rotation order but not any other order. That is "fine" as that is Maya default order, but it has always meant this does not match any other order.

lib/mayaUsd/fileio/utils/xformStack.cpp
```diff
@@ -542,6 +602,12 @@ const UsdMayaXformStack& UsdMayaXformStack::MayaStack()
                 UsdMayaXformStackTokens->rotatePivot, UsdGeomXformOp::TypeTranslate),
             UsdMayaXformOpClassification(
                 UsdMayaXformStackTokens->rotate, UsdGeomXformOp::TypeRotateXYZ),
+            UsdMayaXformOpClassification(
+                UsdMayaXformStackTokens->rotateX, UsdGeomXformOp::TypeRotateX),
+            UsdMayaXformOpClassification(
+                UsdMayaXformStackTokens->rotateY, UsdGeomXformOp::TypeRotateY),
+            UsdMayaXformOpClassification(
+                UsdMayaXformStackTokens->rotateZ, UsdGeomXformOp::TypeRotateZ),
```
pierrebai-adsk, line 613: We are adding rotateX, Y and Z even without spline?

lib/mayaUsd/fileio/utils/xformStack.cpp
```diff
@@ -585,6 +687,12 @@ const UsdMayaXformStack& UsdMayaXformStack::CommonStack()
               UsdMayaXformStackTokens->pivot, UsdGeomXformOp::TypeTranslate),
           UsdMayaXformOpClassification(
               UsdMayaXformStackTokens->rotate, UsdGeomXformOp::TypeRotateXYZ),
+          UsdMayaXformOpClassification(
+              UsdMayaXformStackTokens->rotateX, UsdGeomXformOp::TypeRotateX),
+          UsdMayaXformOpClassification(
+              UsdMayaXformStackTokens->rotateY, UsdGeomXformOp::TypeRotateY),
+          UsdMayaXformOpClassification(
+              UsdMayaXformStackTokens->rotateZ, UsdGeomXformOp::TypeRotateZ),
```
pierrebai-adsk: Same question: adding support for rotX, Y Z in the non-spline case?

lib/mayaUsd/fileio/utils/xformStack.cpp
```diff
@@ -567,16 +633,52 @@ const UsdMayaXformStack& UsdMayaXformStack::MayaStack()
...
+          UsdMayaXformOpClassification(UsdMayaXformStackTokens->scale, UsdGeomXformOp::TypeScale),
+          UsdMayaXformOpClassification(UsdMayaXformStackTokens->scaleX, UsdGeomXformOp::TypeScaleX),
+          UsdMayaXformOpClassification(UsdMayaXformStackTokens->scaleY, UsdGeomXformOp::TypeScaleY),
+          UsdMayaXformOpClassification(UsdMayaXformStackTokens->scaleZ, UsdGeomXformOp::TypeScaleZ),
```
pierrebai-adsk: Has OpenUSD updated their definition of the common stack?

lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -240,6 +238,36 @@ static bool _pushUSDXformOpToMayaXform(
     const UsdMayaPrimReaderArgs&    args,
     const UsdMayaPrimReaderContext* context)
 {
+#if USD_SUPPORTS_SPLINES
```
derlerk: Maybe it's not so much USD_SUPPORTS_SPLINES, but more USD_SUPPORT_INDIVIDUAL_TRANSFROMS (or something similar)?

lib/mayaUsd/fileio/utils/xformStack.cpp
barbalt: That's the case for the rotate -> `UsdGeomXformOp::TypeRotateXYZ`? 
That's how it was before. I added only 
```
UsdMayaXformOpClassification(
                UsdMayaXformStackTokens->rotateX, UsdGeomXformOp::TypeRotateX),
            UsdMayaXformOpClassification(
                UsdMayaXformStackTokens->rotateY, UsdGeomXformOp::TypeRotateY),
            UsdMayaXformOpClassification(
                UsdMayaXformStackTokens->rotateZ, UsdGeomXformOp::TypeRotateZ),
```
Will those also only match xyz order? I thought they would match the specific token and the order wouldn't matter

lib/mayaUsd/fileio/utils/xformStack.cpp
barbalt: rotate was there before. 2505 only added the translate and scale. I thought I could add that here and remove from the other function, but seeing the errors, I may be mistaken

lib/mayaUsd/fileio/utils/xformStack.cpp
pierrebai-adsk: As you said, this is how it has always been, I was just pointing it out.

pierrebai-adsk (dismissed):

lib/mayaUsd/fileio/utils/xformStack.cpp
```diff
@@ -105,11 +93,21 @@ UsdMayaXformStack::TokenIndexPairMap _buildAttrNamesToIdxs(
...
             UsdMayaXformStack::IndexPair indexPair = _makeInversionIndexPair(i, inversionMap);
-            // Insert, and check if it already existed
+#if USD_SUPPORT_INDIVIDUAL_TRANSFROMS
+            if (result.find(attrName) == result.end()) {
+                // Insert, and check if it already existed
+                TF_VERIFY(
+                    result.insert({ attrName, indexPair }).second,
+                    "AttrName %s already found in attrName lookup map",
+                    attrName.GetText());
+            }
```
pierrebai-adsk, line 103: IDK why this needs to be different when we have individual transforms?

pierrebai-adsk: Also, this is the MayaXformStack, which I thought you now not modify?

barbalt: Because those lines here will also add rotation xform to the list. This verify will be triggered prompting a message to the user saying that the operation already exists
https://github.com/Autodesk/maya-usd/pull/4232/files#diff-416277513d592c48d2c7edba7cc94543dc64ebaf0f0c0dccb70693b6b5eed71fR211-R212

barbalt: Not sure what you mean by that is the maya stack? That function is called in all stacks

pierrebai-adsk (changes requested):
I only have a single question concerning the last commit.

lib/mayaUsd/fileio/utils/xformStack.cpp
pierrebai-adsk: Yeah, I realized it afterward and deleted my comments.

pierrebai-adsk: Oh so you mean the old code used to add rotateX, Y and Z. Honestly I don't remember the detail of the code at lines 211+, but it was already possible in the past to have both rotateXYZ and rotateX in a uSD file. IDK how the old code was supporting that. I'm going to trust that testing will find any problem.

OTOH, would we not need to add the same code as line 211+ for translate and scale when the support for individual transforms is supported?

barbalt: The operation maya stack only defines `rotatexyz` as an operation, that's why it needed to be added rotatex,rotatey,rotatez individually. 
In the new matrix, I've added rotatex,rotatey,rotatez explicitly (as well as the translate and scale equivalents). Those new operations  are being taken care by this part of the code: https://github.com/Autodesk/maya-usd/pull/4232/files#diff-416277513d592c48d2c7edba7cc94543dc64ebaf0f0c0dccb70693b6b5eed71fR236-R241
Hence why they don't need to be specifically broken down.
I'm also not 100% sure how we dealt with the rotatex,rotatey,rotatez before. maybe ufe?

PR #4234, opened 2025-06-10
EMSUSD-2295 : Texture paths are not saved as relative when exporting an object with MaterialX Document

derlerk:
Add relative path support for texture file while exporting MaterialX stack.

### DIFF
--- a/lib/usd/translators/shading/mtlxMaterialXSurfaceShaderWriter.cpp
+++ b/lib/usd/translators/shading/mtlxMaterialXSurfaceShaderWriter.cpp
@@ -17,6 +17,7 @@
 
 #include <mayaUsd/fileio/shaderWriter.h>
 #include <mayaUsd/fileio/shaderWriterRegistry.h>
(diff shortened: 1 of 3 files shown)


lib/usd/translators/shading/mtlxMaterialXSurfaceShaderWriter.cpp
derlerk, line 1: This is mostly refactoring, moved all the function to be private method of the ShaderWriter, in order to be able to access the export options easily.

lib/usd/translators/shading/mtlxMaterialXSurfaceShaderWriter.cpp
```diff
@@ -275,13 +534,25 @@ void _ConnectToNodeGraph(
     }
 }
 
-// Sets the value of a USD input based on a MaterialX input.
-void _SetInputValue(
+void MtlxMaterialXSurfaceShaderWriter::_SetInputValue(
```
derlerk, line 537: The only actual code change happens in this method.

PR #4238, opened 2025-06-13
LOOKDEVX-4467 - Move ProxyShapeLookdevHandler to MayaUSD.

frohnej-adsk:
To decouple form USD, LookkdevX wants to move out any remaining USD dependencies. 

This PR moves the `ProxyShapeLookdevHandler` and its associated command `MayaUsdCreateLookdevEnvironmentCommand` to MayaUSD. This was missed when moving LookdevXUsd to MayaUSD: The `ProxyShapeLookdevHandler` is registered against the Maya runtime and, thus, resided in a different folder.

Unfortunately, there are no Python bindings for the `LookdevXUfe::LookdevHandler` yet, so I'm not sure if the test can be moved over right now. I logged another ticket to add bindings and move the test. 

This PR also makes a small change to `LookdevXUfe::UsdMaterialHandler`. See comment below.

### DIFF
--- a/lib/lookdevXUsd/ProxyShapeLookdevHandler.h
+++ b/lib/lookdevXUsd/ProxyShapeLookdevHandler.h
@@ -1,0 +1,88 @@
+//*****************************************************************************
+// Copyright (c) 2025 Autodesk, Inc.
+// All rights reserved.
+//
+// These coded instructions, statements, and computer programs contain
+// unpublished proprietary information written by Autodesk, Inc. and are
+// protected by Federal copyright law. They may not be disclosed to third
+// parties or copied or duplicated in any form, in whole or in part, without
+// the prior written consent of Autodesk, Inc.
+//*****************************************************************************
+#pragma once
+
+#include <LookdevXUfe/LookdevHandler.h>
+
+#include <ufe/undoableCommand.h>
+
+namespace LookdevXUsd
+{
+
+//! \brief Maya run-time Lookdev handler. Used to deal with MayaUsdProxyShape items.
+class ProxyShapeLookdevHandler : public LookdevXUfe::LookdevHandler
+{
+public:
+    using Ptr = std::shared_ptr<ProxyShapeLookdevHandler>;
+
+    explicit ProxyShapeLookdevHandler(LookdevXUfe::LookdevHandler::Ptr previousHandler);
+    ~ProxyShapeLookdevHandler() override = default;
+
+    //@{
+    //! Delete the copy/move constructors assignment operators.
+    ProxyShapeLookdevHandler(const ProxyShapeLookdevHandler&) = delete;
+    ProxyShapeLookdevHandler& operator=(const ProxyShapeLookdevHandler&) = delete;
+    ProxyShapeLookdevHandler(ProxyShapeLookdevHandler&&) = delete;
+    ProxyShapeLookdevHandler& operator=(ProxyShapeLookdevHandler&&) = delete;
+    //@}
+
+    //! Create a ProxyShapeLookdevHandler.
+    static ProxyShapeLookdevHandler::Ptr create(const LookdevXUfe::LookdevHandler::Ptr& previousHandler);
+
+    [[nodiscard]] Ufe::SceneItemResultUndoableCommand::Ptr createLookdevContainerCmdImpl(
+        const Ufe::SceneItem::Ptr& parent, const Ufe::PathComponent& name) const override;
+    [[nodiscard]] Ufe::SceneItemResultUndoableCommand::Ptr createLookdevContainerCmdImpl(
+        const Ufe::SceneItem::Ptr& parent, const Ufe::NodeDef::Ptr& nodeDef) const override;
+    [[nodiscard]] Ufe::SceneItemResultUndoableCommand::Ptr createLookdevEnvironmentCmdImpl(
+        const Ufe::SceneItem::Ptr& ancestor, Ufe::Rtid targetRunTimeId) const override;
+    [[nodiscard]] bool isLookdevContainerImpl(const Ufe::SceneItem::Ptr& item) const override;
+
+private:
+    LookdevXUfe::LookdevHandler::Ptr m_previousHandler;
+}; // ProxyShapeLookdevHandler
+
+class MayaUsdCreateLookdevEnvironmentCommand : public Ufe::SceneItemResultUndoableCommand
+{
+public:
+    using Ptr = std::shared_ptr<MayaUsdCreateLookdevEnvironmentCommand>;
+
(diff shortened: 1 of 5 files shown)


lib/lookdevXUsd/UsdMaterialHandler.cpp
```diff
@@ -43,14 +43,6 @@ LookdevXUfe::Material::Ptr UsdMaterialHandler::material(const Ufe::SceneItem::Pt
         return nullptr;
     }
 
-    // Test if this item is imageable. If not, then we cannot create a material
-    // interface for it, which is a valid case (such as for a material node type).
-    PXR_NS::UsdGeomImageable primSchema(MayaUsdAPI::getPrimForUsdSceneItem(item));
-    if (!primSchema)
-    {
-        return nullptr;
-    }
```
frohnej-adsk: In LookdevX, we add an action "Select objects with Material" to the context ops of USD materials. Previously, that action used the USD API directly to fetch the assigned Material of geos or geom-subsets.

To decouple from USD, I changed it to use the `LookdevXUfe::Material` interface instead (see LookdevX PR), but it didn't work for geom-subsets: Geom-subsets are apparently not Imageable.

I removed this check to allow fetching the materials assigned to geom-subsets. However, I'm not super familiar with USD, so if someone could doublecheck that [lookdevXUsd/UsdMaterial.cpp](https://github.com/Autodesk/maya-usd/blob/dev/lib/lookdevXUsd/UsdMaterial.cpp) doesn't contain any code that relies on the prim to be imageable, it would be greatly appreciated.

lib/lookdevXUsd/ProxyShapeLookdevHandler.h
```diff
@@ -0,0 +1,88 @@
...
+
+    void execute() override;
+    void undo() override;
+    void redo() override;
+
+private:
+    bool executeCommand();
+
+    Ufe::Path m_ancestor;
+    Ufe::SceneItem::Ptr m_materialParent;
```
seando-adsk: Can this scene item become stale or invalid in any way? Its usually better to store the Ufe::Path and create the item as needed.

lib/lookdevXUsd/UsdMaterialHandler.cpp
JGamache-autodesk: I would prefer creating an exception for geom subsets rather than allowing material interface creation on non-imageable geometries like lights, cameras, skeletons, etc
```c++
prim = MayaUsdAPI::getPrimForUsdSceneItem(item);
if (!PXR_NS::UsdGeomImageable(primSchema) && !item->IsA<UsdGeomSubset>())
{
    return nullptr;
}
```
You can check if any prim currently has a material using:
```c++
prim = MayaUsdAPI::getPrimForUsdSceneItem(item);
if (!prim.HasAPI<UsdShadeMaterialBindingAPI>()) 
{
    return nullprt;
}
```
You can check if the prim can be assigned a material even if it does not currently using:
```c++
prim = MayaUsdAPI::getPrimForUsdSceneItem(item);
if (!PXR_NS::UsdShadeMaterialBindingAPI::CanApply(prim))
{
    return nullprt;
}
```
But the CanApply rules are extremely lax, so might not be strong enough for our needs.

lib/lookdevXUsd/ProxyShapeLookdevHandler.h
frohnej-adsk: Good call! It hasn't caused any issues yet, since we only ever access the result immediately after running the command, but it's still better to store the path. See 84a8e9c.

lib/lookdevXUsd/UsdMaterialHandler.cpp
frohnej-adsk: Agreed! See 84a8e9c.

lib/lookdevXUsd/ProxyShapeLookdevHandler.h
JGamache-autodesk: Actually vital if you go `undo, undo, undo, undo, redo, redo, redo, redo` and somehow one of the extra undo has deleted the scene item you were holding, causing a crash with the stale item kept in the command which has been recreated somewhere else in memory.

JGamache-autodesk (approved):

frohnej-adsk:
One test `testDiagnosticDelegate` failed on `2026 (OSX)`, `2026 (OSX) [USD 25.05]` and `2027 (OSX Release)`. Is that expected?

Everything else passed.

seando-adsk:
> One test `testDiagnosticDelegate` failed on `2026 (OSX)`, `2026 (OSX) [USD 25.05]` and `2027 (OSX Release)`. Is that expected?
> 
> Everything else passed.

I've seen that test randomly fail in the past. I don't see how your changes would have affected this test, so I think we can ignore that failure.

lib/lookdevXUsd/UsdMaterialHandler.cpp
```diff
@@ -43,10 +44,10 @@ LookdevXUfe::Material::Ptr UsdMaterialHandler::material(const Ufe::SceneItem::Pt
         return nullptr;
     }
 
-    // Test if this item is imageable. If not, then we cannot create a material
+    // Test if this item is imageable or a geom subset. If not, then we cannot create a material
     // interface for it, which is a valid case (such as for a material node type).
-    PXR_NS::UsdGeomImageable primSchema(MayaUsdAPI::getPrimForUsdSceneItem(item));
-    if (!primSchema)
+    const auto prim = MayaUsdAPI::getPrimForUsdSceneItem(item);
+    if (!PXR_NS::UsdGeomImageable(prim) && !prim.IsA<UsdGeomSubset>())
```
stefanminning-autodesk: Nit: `PXR_NS` might be redundant as we're already declaring using namespace above.

seando-adsk: Actually its in the MayaUsd coding standards to always use the fully qualified namespace. So we should use `PXR_NS::`. 
The reason for the `using namespace PXR_NS` above is because of the `TF_VERIFY` macro.

frohnej-adsk: Oh, I saw this too late. Will revert.

frohnej-adsk: Added `PXR_NS::` in fort of `UsdGeomSubset` as well. See 9b04cdd.

frohnej-adsk:
Same CI result: `testDiagnosticDelegate` failed on 2026 (OSX), 2026 (OSX) [USD 25.05] and 2027 (OSX Release), but everything else passed, so we should be ready to merge.

PR #4240, opened 2025-06-16
EMSUSD-2576 : Rotation is incorrectly written on export to USD when using splines.

derlerk:
Fix Rotation op order.
USD rotation should be in degree.

### DIFF
--- a/lib/mayaUsd/fileio/translators/translatorXformable.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorXformable.cpp
@@ -657,8 +657,8 @@ void UsdMayaTranslatorXformable::Read(
                                     { "y", MTransformationMatrix::RotationOrder::kYXZ },
                                     { "z", MTransformationMatrix::RotationOrder::kZXY } };
 
-                auto  it = rotOrderMap.find(rotOrderStr);
-                auto  rotOrder = (it != rotOrderMap.end()) ? it->second : [&]() {
+                auto it = rotOrderMap.find(rotOrderStr);
+                auto rotOrder = it != rotOrderMap.end() ? it->second : [&]() {
                     TF_WARN(
                         "Unsupported rotation order '%s' for prim <%s>",
                         rotOrderStr.c_str(),
--- a/lib/mayaUsd/fileio/translators/translatorXformable.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorXformable.cpp
@@ -246,8 +246,9 @@ static bool _pushUSDXformOpToMayaXform(
         if (!plg.isNull()) {
             auto                       spline = opAttr.GetSpline();
             const UsdGeomXformOp::Type opType = xformop.GetOpType();
-            if (opType == UsdGeomXformOp::TypeRotateX || opType == UsdGeomXformOp::TypeRotateY
-                || opType == UsdGeomXformOp::TypeRotateZ) {
+            bool                       rotOp = opType == UsdGeomXformOp::TypeRotateX
+                || opType == UsdGeomXformOp::TypeRotateY || opType == UsdGeomXformOp::TypeRotateZ;
+            if (rotOp) {
                 MFnTransform trans;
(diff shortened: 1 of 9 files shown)


lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -246,25 +247,16 @@ static bool _pushUSDXformOpToMayaXform(
         if (!plg.isNull()) {
             auto                       spline = opAttr.GetSpline();
             const UsdGeomXformOp::Type opType = xformop.GetOpType();
-            if (opType == UsdGeomXformOp::TypeRotateX || opType == UsdGeomXformOp::TypeRotateY
```
derlerk, line 249: This code was always putting the rotation order to be XYZ on import. We need to analyze the stack to get the proper order.

lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -246,25 +247,16 @@ static bool _pushUSDXformOpToMayaXform(
...
+                || opType == UsdGeomXformOp::TypeRotateY || opType == UsdGeomXformOp::TypeRotateZ;
 
             if (UsdGeomXformOp::GetPrecisionFromValueTypeName(xformop.GetAttr().GetTypeName())
                 == UsdGeomXformOp::PrecisionDouble) {
                 return UsdMayaSplineUtils::WriteUsdSplineToPlug<double>(plg, spline, context);
             }
 
-            return UsdMayaSplineUtils::WriteUsdSplineToPlug<float>(plg, spline, context);
+            return UsdMayaSplineUtils::WriteUsdSplineToPlug<float>(
+                plg, spline, context, rotOp ? M_PI / 180.0 : 1.f);
```
derlerk, line 259: USD rotation angles are in degrees. -> convert to rad

lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -630,7 +623,55 @@ void UsdMayaTranslatorXformable::Read(
...
+                                    { "zyx", MTransformationMatrix::RotationOrder::kZYX },
+                                    { "xy", MTransformationMatrix::RotationOrder::kXYZ },
+                                    { "xz", MTransformationMatrix::RotationOrder::kXZY },
+                                    { "yx", MTransformationMatrix::RotationOrder::kYXZ },
+                                    { "yz", MTransformationMatrix::RotationOrder::kYZX },
+                                    { "zx", MTransformationMatrix::RotationOrder::kZXY },
+                                    { "zy", MTransformationMatrix::RotationOrder::kZYX },
+                                    { "x", MTransformationMatrix::RotationOrder::kXYZ },
+                                    { "y", MTransformationMatrix::RotationOrder::kYXZ },
+                                    { "z", MTransformationMatrix::RotationOrder::kZXY } };
```
pierrebai-adsk: Would it work if those last three all used XYZ order? That is the most usual order and for single-op, the order should not be important.

Maybe the same for the xy, xz and yz above that.

lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -630,7 +623,55 @@ void UsdMayaTranslatorXformable::Read(
...
+
+                auto it = rotOrderMap.find(rotOrderStr);
+                auto rotOrder = it != rotOrderMap.end() ? it->second : [&]() {
```
pierrebai-adsk: I'd really add some parentheses here to clarify evaluation order,

lib/mayaUsd/fileio/translators/translatorXformable.cpp
derlerk: Agreed for the single one. If two, I think as a first step let's keep them ordered as specified in the stack. Maybe we'll get feedback on that.

PR #4241, opened 2025-06-18
EMSUSD-2577 : Some tangent usage are not translated correctly in 25.05 on Stage Creation

derlerk:
Fix export tangents Y values. It was missing scaling.
Fix import of individual transforms.

### DIFF
--- a/lib/mayaUsd/fileio/translators/translatorXformable.cpp
+++ b/lib/mayaUsd/fileio/translators/translatorXformable.cpp
@@ -289,13 +289,13 @@ static bool _pushUSDXformOpToMayaXform(
 {
 #if USD_SUPPORT_INDIVIDUAL_TRANSFROMS
     // If the xformop has a spline, we write it to the plug directly
-    const auto& opAttr = xformop.GetAttr();
+    const auto&                opAttr = xformop.GetAttr();
+    const UsdGeomXformOp::Type opType = xformop.GetOpType();
     if (opAttr.HasSpline()) {
         MPlug plg = MdagNode.findPlug(MString(opName.GetString().c_str()), false);
         if (!plg.isNull()) {
-            auto                       spline = opAttr.GetSpline();
-            const UsdGeomXformOp::Type opType = xformop.GetOpType();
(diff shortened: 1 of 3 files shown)


lib/mayaUsd/fileio/utils/splineUtils.h
```diff
@@ -101,6 +101,9 @@ struct UsdMayaSplineUtils
             inTangentX *= convert.as(MTime::uiUnit());
             outTangentX *= convert.as(MTime::uiUnit());
 
+            inTangentY *= scaling;
```
derlerk, line 104: Tangents Y also need the scaling applied to them. For example for angles, it need to be converted from rad to degree.

lib/mayaUsd/fileio/translators/translatorXformable.cpp
derlerk, line 1: The goal here is to fix a bug on import (or missing feature...). The individual transform would not been supported if not animated with spline. Now it's supported for static values or time-samples.

derlerk: Which would be pretty bad if you splined only X translate and had static Y for example. Y would not be imported at all.

lib/mayaUsd/fileio/translators/translatorXformable.cpp
```diff
@@ -52,11 +52,21 @@ PXR_NAMESPACE_OPEN_SCOPE
...
+    const TfToken&        opName,
+    GfVec3d&              value,
+    const UsdTimeCode&    usdTime)
 {
     bool retValue = false;
 
+#ifdef USD_SUPPORT_INDIVIDUAL_TRANSFROMS
+    if ((xformOp.GetTypeName() == SdfValueTypeNames->Float
+         || xformOp.GetTypeName() == SdfValueTypeNames->Double)
+        && opName != UsdMayaXformStackTokens->rotateAxis) {
```
pierrebai-adsk, line 68: I would like a comment explaining why this is only for rotateAxis.

pierrebai-adsk (approved):
Very minor request, otherwise looks good.

pierrebai-adsk (approved):

PR #4243, opened 2025-06-18
EMSUSD-2578: Material prims break AE on Maya 2023 / 2022

seando-adsk:
#### EMSUSD-2578: Material prims break AE on Maya 2023 / 2022
* NodeDefHandler was only added in Ufe v4

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
@@ -422,6 +422,9 @@ class AETemplate(object):
                     if self.attrS.attribute(item):
                         self.addControls([item])
     def isRamp(self):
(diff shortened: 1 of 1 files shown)


lib/mayaUsd/resources/ae/usdschemabase/ae_template.py
```diff
@@ -422,6 +422,9 @@ def addShaderLayout(self, group):
                     if self.attrS.attribute(item):
                         self.addControls([item])
     def isRamp(self):
+        if not hasattr(ufe, "NodeDefHandler"):
+            return False
```
seando-adsk, line 426: NodeDefHandler (used on line 429 below) was only added in Ufe v4, so its not there in Maya 2023.

PR #4256, opened 2025-06-30
API doc is no longer included schema's "doc" field

dj-mcg:
The assertion that the doc metadata should exist is no longer valid

### DIFF
--- a/test/lib/ufe/testAttribute.py
+++ b/test/lib/ufe/testAttribute.py
@@ -1607,15 +1607,10 @@ class AttributeTestCase(unittest.TestCase):
         # Test for non-existing metdata.
         self.assertFalse(attr.hasMetadata('NotAMetadata'))
 
-        # Verify that this attribute has documentation metadata.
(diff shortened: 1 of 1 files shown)


test/lib/ufe/testAttribute.py
```diff
@@ -1607,15 +1607,10 @@ def testMetadata(self):
         # Test for non-existing metdata.
         self.assertFalse(attr.hasMetadata('NotAMetadata'))
 
-        # Verify that this attribute has documentation metadata.
-        self.assertTrue(attr.hasMetadata('documentation'))
-        md = attr.getMetadata('documentation')
-        self.assertIsNotNone(md)
-        self.assertIsNotNone(str(md))
```
seando-adsk, line 1614: Should we wrap this in a if based on the USD version and still do the test for the older USD versions?

dj-mcg: Sounds good!

PR #4260, opened 2025-07-11
EMSUSD-2150 & EMSUSD-2499: Outliner icons and add class prim contexts menu

seando-adsk:
#### EMSUSD-2150: We need to add Outliner icons for new prims
* New icons for: Physics, Procedural, Render and UI.
* Update icon unit test.

#### EMSUSD-2499: Create Class Prim
* Added context menu item and modifed command to create class prim.
* Added test to ensure proper icon.

#### EMSUSD-2499: Create Class Prim
* When creating a class prim, enable display of class prims in Outliner.


lib/mayaUsd/ufe/MayaUsdContextOps.cpp
```diff
@@ -599,8 +603,20 @@ Ufe::UndoableCommand::Ptr MayaUsdContextOps::doOpCmd(const ItemPath& itemPath)
...
+        // EMSUSD-2499: Create Class Prim
+        // Special case when adding a class prim via context menu make sure the Outliner
+        // is displaying class prims.
+        if (!itemPath.empty() && (itemPath[0] == kUSDAddNewPrimItem)) {
+            // At this point we know the last item in the itemPath is the prim type to create
+            auto primType = itemPath[itemPath.size() - 1];
+            if (primType == kUSDClassPrimItem) {
+                enableOutlinerClassPrims();
+            }
+        }
```
seando-adsk, line 616: Do this in the Maya decoration as it applies to the Maya Outliner.

lib/mayaUsd/ufe/Utils.cpp
```diff
@@ -148,9 +148,11 @@ std::string uniqueChildNameMayaStandard(const PXR_NS::UsdPrim& usdParent, const
         return std::string();
 
     // See uniqueChildNameDefault() in lib\usdUfe\ufe\Utils.cpp for details.
+    // Note: removed 'UsdPrimIsAbstract' from the predicate since the Maya
+    //       Outliner can show class prims now.
     TfToken::HashSet allChildrenNames;
     for (auto child : usdParent.GetFilteredChildren(
-             UsdTraverseInstanceProxies(UsdPrimIsDefined && !UsdPrimIsAbstract))) {
+             UsdTraverseInstanceProxies(UsdPrimIsDefined))) {
```
seando-adsk: @deboisj Julien, do you think I need to also remove `UsdPrimIsAbstract` from usdUfe/ufe/Util.cpp in the `uniqueChildNameDefault` function. This one here is Maya specific.

lib/usdUfe/ufe/UsdUIInfoHandler.cpp
```diff
@@ -202,18 +218,33 @@ Ufe::UIInfoHandler::Icon UsdUIInfoHandler::treeViewIcon(const Ufe::SceneItem::Pt
         return Ufe::UIInfoHandler::Icon("out_USD_UsdTyped.png"); // Default USD icon
     }
 
+    // If the node type ends with _ followed by a number, we strip the number.
+    // This is to handle the case where we have a node type like "Capsule_1".
+    auto nodeType = item->nodeType();
+    if (nodeType.size() > 2 && nodeType[nodeType.size() - 2] == '_'
+        && std::isdigit(nodeType[nodeType.size() - 1])) {
+        nodeType = nodeType.substr(0, nodeType.size() - 2);
+    }
```
seando-adsk, line 229: Special case for schema types like `Capsule_1` (and future proofed in case Pixar ever does a `_2`

lib/usdUfe/ufe/UsdUIInfoHandler.cpp
```diff
@@ -202,18 +218,33 @@ Ufe::UIInfoHandler::Icon UsdUIInfoHandler::treeViewIcon(const Ufe::SceneItem::Pt
...
 
-    // Check if we have any composition meta data - if yes we display a special badge.
     auto usdItem = downcast(item);
-    if (usdItem && usdItem->prim()) {
+    auto usdPrim = usdItem ? usdItem->prim() : PXR_NS::UsdPrim();
+
+    // Special case for Class prims.
+    if (usdPrim && usdPrim.GetSpecifier() == PXR_NS::SdfSpecifierClass) {
+        icon.baseIcon = "out_USD_Class.png";
+    }
```
seando-adsk, line 244: Like Def, Class prims are typeless and that is handled at line 170. So we need a special case to handle Class prims to give them their own icon.

lib/mayaUsd/ufe/Utils.cpp
deboisj: Probably makes sense to consider all prims yes. Abstract or not - seems to me even before this would have been an issue - even if hidden you still want to generate a "real" unique sibling no?

lib/usdUfe/ufe/UsdContextOps.cpp
```diff
@@ -421,6 +429,7 @@ Ufe::ContextOps::Items UsdContextOps::getItems(const Ufe::ContextOps::ItemPath&
         }     // Variant sets
         else if (itemPath[0] == kUSDAddNewPrimItem) {
             if (itemPath.size() == 1u) { // Root setup
+                items.emplace_back(kUSDClassPrimItem, kUSDClassPrimLabel, kUSDClassPrimImage);
```
deboisj, line 432: i find it a bit odd to mix specifiers with types - while generally not very useful to my knowledge, class prims can have a type. I see how design wise i guess it works but in principle type and specifier are different properties of a prim

lib/usdUfe/ufe/UsdUIInfoHandler.cpp
deboisj: What is Capsule_1?

lib/usdUfe/ufe/UsdUIInfoHandler.cpp
deboisj: only half true as mentionned above

deboisj (approved):
i think indeed maybe do the uniquename change in usdufe, otherwise looks good

lib/usdUfe/ufe/UsdUIInfoHandler.cpp
seando-adsk: They are new schema types introduced by Pixar.

lib/mayaUsd/ufe/Utils.cpp
seando-adsk: Yes agreed. I changed this location and a few below in another function to get a unique name.

deboisj (dismissed):

deboisj (approved):

PR #4262, opened 2025-07-14
Apply color management file rules on setValue

JGamache-autodesk:
Affects only shader filename attributes.


feldstj (approved):

JGamache-autodesk:
@seando-adsk ready for merge.

PR #4265, opened 2025-07-16
EMSUSD-2589 Add asset info section to the attribute editor

pierrebai-adsk:
- Added a `AssetInfoCustomControl` to display asset info in the AE.
- Removed asset info from the metadata section.
- Use the new asset info control in the AE template.
- Only show the asset info section if there is info to show.
- Added a button to open a window to show the payloads.
- Don't open a new payloads window if one is already shown for the prim.

### DIFF
--- a/lib/mayaUsd/resources/ae/usdschemabase/assetInfoCustomControl.py
+++ b/lib/mayaUsd/resources/ae/usdschemabase/assetInfoCustomControl.py
@@ -1,4 +1,4 @@
-# Copyright 2024 Autodesk
+# Copyright 2025 Autodesk
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.


lib/mayaUsd/resources/ae/usdschemabase/assetInfoCustomControl.py
```diff
@@ -0,0 +1,171 @@
...
+
+from mayaUsdLibRegisterStrings import getMayaUsdLibString
+
+from pxr import Usd
+
+class AssetInfoCustomControl(object):
+    '''Custom control for the prim asset info metadata.'''
+
+    def __init__(self, item, prim, useNiceName):
+        # In Maya 2022.1 we need to hold onto the Ufe SceneItem to make
```
barbalt, line 27: Is this still relevant since we don't support maya 2022 anymore?

pierrebai-adsk: We don't? I copied the lines from another UI we have that did the same. Does not cost much if a user want to have MayaUSD on Maya 2022.

barbalt: I'm just asking based on the comment

seando-adsk: We no longer provide 2022 builds (uploaded to Github) and we don't test 2022. But there is no harm in keeping this code since it was copied from another location and we haven't removed existing 2022 from our repo.

PR #4271, opened 2025-07-18
EMSUSD-2663 - MayaUsd: Maya blessing broken after Qt 6.8.3 update

seando-adsk:
#### EMSUSD-2663 - MayaUsd: Maya blessing broken after Qt 6.8.3 update
* Qt targets are ALIAS targets in newer Qt.


AramAzhari-adsk (approved):
LGTM

PR #4272, opened 2025-07-21
ProgressBar: Prevent premature completion and warnings

jufrantz:
When `nbLoopSteps` exceeded `maxStepsForLoops`, floor division could cause
`loopAdvance` to emit less steps than `maxStepsForLoops`, the progression could
stop too early and warn “did not advance progress bar correct number of steps".

Replaces the division with integer accumulator ensuring it ends as expected.

Also silence unnecessary warning messages in batch mode.


jufrantz:
> Changes seem generally okay to me. However when I tried to build this locally on Windows I got an error from `uint` which isn't a standard type. You probably want `uint32_t`

Thanks for catching that @seando-adsk,
I replaced the uints so it will be all good on Windows now. I guess that’s a bad practice I picked up from using gcc for too long.
I’ve updated the existing commit to keep a clean history.

AramAzhari-adsk (approved):

PR #4275, opened 2025-07-23
EMSUSD-2637 - Performance is slow with variants on certain Kitbash scene assets

seando-adsk:
#### EMSUSD-2637 - Performance is slow with variants on certain Kitbash scene assets
* Instead of sending out subtreeInvalidate to fix selection list just remove the items from selection list.

### DIFF
--- a/lib/usdUfe/ufe/StagesSubject.cpp
+++ b/lib/usdUfe/ufe/StagesSubject.cpp
@@ -29,9 +29,11 @@
 #include <pxr/usd/usdGeom/xformOp.h>
 
 #include <ufe/attributes.h>
+#include <ufe/globalSelection.h>
 #include <ufe/hierarchy.h>
 #include <ufe/object3d.h>
 #include <ufe/object3dNotification.h>
+#include <ufe/observableSelection.h>
 #include <ufe/path.h>
 #include <ufe/scene.h>
 #include <ufe/sceneNotification.h>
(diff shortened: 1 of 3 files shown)


lib/usdUfe/ufe/StagesSubject.cpp
```diff
@@ -553,9 +551,15 @@ void StagesSubject::stageChanged(
...
+                    // Filter the global selection, removing items below our parent prim.
+                    auto globalSn = Ufe::GlobalSelection::get();
+                    if (!globalSn->empty()) {
+                        bool itemRemoved = false;
+                        auto newSel
+                            = UsdUfe::removeDescendants(*globalSn, parentUfePath, &itemRemoved);
+                        if (itemRemoved) {
+                            globalSn->replaceWith(newSel);
+                        }
                     }
```
seando-adsk, line 563: The code I changed was added in PR #4036 to fix a problem with selection list when deleting prims directly using USD (and not going thru one of the MayaUsd commands). Its only purpose was to remove stale items (that were removed prims or a child of removed prim) from the selection list. Instead I used code that we already had in the set variant command to remove the descendant prims. For this particular bug the SetVariantSelectionCommand already fixes up the selection list, so this code will do nothing now. Before it would send out many, many Ufe SubtreeInvalidate notifs on the root prim of the scene (which was highly destructive and slow). 

I also ran all the tests locally, including the `testSelection` which has a test for this specific case.

lib/usdUfe/ufe/Utils.cpp
```diff
@@ -1457,6 +1460,8 @@ Ufe::Selection removeDescendants(const Ufe::Selection& src, const Ufe::Path& fil
         // The filterPath itself is still valid.
         if (!itemPath.startsWith(filterPath) || itemPath == filterPath) {
             dst.append(item);
+        } else if (nullptr != itemRemoved) {
+            *itemRemoved = true;
```
seando-adsk, line 1464: Added extra (optional) boolean flag to know if something was removed. Helps to optimize the replace of selection (line 560 in StagesSubject.cpp).

AramAzhari-adsk (approved):
Great work!

PR #4278, opened 2025-07-25
Keep session  data in the session layer

pierrebai-adsk:
- Modify the parent and duplicate commands to keep data that is in the session in the session layer.
- Added a unit test to verify this.


lib/mayaUsd/ufe/UsdUndoDuplicateCommand.cpp
```diff
@@ -134,7 +142,8 @@ void UsdUndoDuplicateCommand::execute()
             layer->GetDisplayName().c_str(),
             _usdDstPath.GetText());
 
-        isFirst = false;
+        if (!isInSession)
+            isFirst = false;
```
barbalt, line 150: As I'm not familiar with this code, why do we needed a flag for the first layer and now we don't anymore?

pierrebai-adsk: The flag is meant to say "first non session layer" So it is only set to false when we have processed a non-session layer.

PR #4279, opened 2025-07-25
Batched export and merge to USD

jufrantz:
With this PR `PrimUpdaterManager::mergeToUsd` can now merge multiple pulled DAG paths to separate prims in a *single* invocation.  

This is mostly interesting in the context of animated rigs, referenced with `MayaReference` prims.

The feature matches our previous Alembic "cache-and-swap" workflow that used `AbcExport` with multiple `jobArgs` flags, to export multiple rigs animation to separate files in one go.

--- 

#### Constraint‑safe

`mergeToUsd` removes the Maya data at end of the operation, swapping it to the USD data. For a `MayaRefeference` prim, the Maya reference gets unloaded.

In animation work, rigs are often connected to each others. Characters might be constrained to vehicles, props to characters, characters to characters, etc

Artists (and tools) must therefore cache the rig animations in an appropriate order (driven first then driver) so every constrained rig evaluates with their complete upstream graph, this can become a complex task. If the rig graph contains a cycle, which is possible at the rig-level, caching becomes impossible without another manual pre-baking.

With a "merge batch", all referenced rigs remain loaded until all USD data is authored, so the rigs remain connected to live Maya data during the whole evaluation. Artists can export the whole scene, or a group of related rigs, in one go. Order and cycles stop mattering.

#### Less waiting

Batching merge operations optimizes `“Cache to USD”` for animated rigs. In production, rig animation evaluation is the main cost of this caching. Currently, artists must export each referenced rig one‑by‑one; each export walks the same frame range again, making some rigs recompute identical results.

The new batched export walks the timeline **once** and shares evaluations across all rigs, eliminating this redundancy.

I tested performances with this scene [`four_animated_rhinos.ma`](https://github.com/user-attachments/files/21432802/four_animated_rhinos.ma.gz) and using this script [test_merge_four_animated_rhinos.py](https://github.com/user-attachments/files/21432785/test_merge_four_animated_rhinos.py.gz). The scene is featuring 4 animated references to `rhino_rig_001.ma`, the rig file was found in maya-usd unit-tests. 

|              | 4 sequential mergeToUsd | Batched mergeToUsd  | Speed‑up |
|--------------|-------------------------|----------------------------|----------|
| DG           | 56 s                    | 35 s                       | **1.8×** |
| EM parallel  | 42 s                    | 25 s                       | **1.7×** |

#### UsdMaya_WriteJobBatch
New class, a collection of `UsdMaya_WriteJob`, that will execute all jobs, each writing to a separate stage, and write the timeSamples in a single unioned Maya timeline pass. https://github.com/jufrantz/maya-usd/commit/aacaf4de009f865c97ffc8d4762bef969bb974da

#### UsdMaya_WriteJob
Destination file name and *append* flag are now supplied in the constructor, instead of the Write method. This is to ease implementation of `UsdMaya_WriteJobBatch`. If this is a concern, I could rework this change. https://github.com/jufrantz/maya-usd/commit/5e32d35c5ba3aed9015ce034d89574f89673c8d0

#### Internal refactors
- Shared write logic for both `UsdMaya_WriteJob` and `UsdMaya_WriteJobBatch` is centralized in a new private class `UsdMaya_WriteJobImpl`.  https://github.com/jufrantz/maya-usd/commit/aacaf4de009f865c97ffc8d4762bef969bb974da
- `UsdMaya_WriteJob::_BeginWriting` was adapting the maya scene for`renderLayerMode`, `upAxis`, and `unit` options.
- These tweaks need to remain for whole scene evaluation with multiple active (began) `UsdMaya_WriteJob`, so I refactored those aspects:
  - Maya scene tweaking is now done in `UsdMaya_WriteJobBatch::WriteJobs` and encloses multiple jobs evaluation.
  - USD authoring related to these options is still done by each `UsdMaya_WriteJob`.
  - cf https://github.com/jufrantz/maya-usd/commit/5a062137933a339d516ee7646ee71cfda4cc5629, https://github.com/jufrantz/maya-usd/commit/afb00e2d61585c9597ea033b1416ab0dad563855

#### PrimUpdaterManager  
- Added a new `PushToUsdArgs` struct that bundles all inputs common to push‑style operations (`mergeToUsd`, `duplicateToUsd`). https://github.com/jufrantz/maya-usd/commit/aaa06a74024ba606b95aef92f3a035ed0a56a0fa
- `PrimUpdaterManager::mergeToUsd` now takes a `std::vector<PushToUsdArgs>`, so multiple DAGs can be exported in one call.  https://github.com/jufrantz/maya-usd/commit/3f68054088bfd0ea84a69c5960afa22fe2aed517
- Added a mergeToUsd python overload taking a sequence of tuples giving the dagPath and associated options dict. https://github.com/jufrantz/maya-usd/commit/80d417d6b8444ea4985be04e766a1bfe4e25d0ff. 
```
mayaUsd.lib.PrimUpdaterManager.mergeToUsd([
    ("MayaReference1", {"rn_layer":"cache1.usd", "rn_primName":"Cache1"}), 
    ("MayaReference2", {"rn_layer":"cache2.usd", "rn_primName":"Cache2"})])
```
- The original single‑object python signature remains for backward compatibility.  
- The **mayaUsdMergeToUsd** command is also extended. It can export multiple DAGs with their exportOptions. https://github.com/jufrantz/maya-usd/commit/d857156ff0a2a83afb2976db98de9f5c9b1fad8e.
```
cmds.mayaUsdMergeToUsd(
    "Xform1", "Xform2",
    exportOptions=["animation=1;startTime=1;endTime=5", "animation=1;startTime=1;endTime=10"])
```

#### Tests
Added unit tests for batched merge, custom rig updaters, `MayaReference` caching, undo/redo integrity.
https://github.com/jufrantz/maya-usd/commit/0bbf95589c84ef1ae56fb593530a99b7d7a826bc, https://github.com/jufrantz/maya-usd/commit/a537f29c86852b480f3207817c90a9c2d74dbc7b 

### I did not

* Add batched export support to the **`mayaUsdExport`** command. I imagine this could be added with a flag similar to `AbcExport -jobArg` / `mayaUsdMergeToUsd -exportOptions`, but I am unsure of the best syntax.
* Implement a batched **`duplicateToUsd`**; though the same technique would work.
* Expose the feature in UFE context menus (e.g., a multi‑selection “Cache To USD”). For now it’s meant mainly for studio pipeline tools.

I'd be happy to add any of these in a later PR if you think they are useful.

### PS

This PR introduces a large set of changes, but I prefered to keep everything in one pull request to give you the full picture. However there are clearly 2 subsets of changes (UsdMaya_WriteJob + PrimUpdaterManager), I could split this PR in two if you prefer.

I've organized the changes into logical commits to make the review easier. Hope that helps.

Feedbacks are very welcome,
Julien

### DIFF
--- a/lib/mayaUsd/fileio/jobs/writeJob.cpp
+++ b/lib/mayaUsd/fileio/jobs/writeJob.cpp
@@ -177,7 +177,7 @@ static MObject _WantedRenderLayerNode(const TfToken& renderLayerMode)
 
 MStatus _ActivateRenderLayer(const MObject& renderLayerNode)
 {
-    static const MString cmdFmt("editRenderLayerGlobals( -currentRenderLayer ^1s");
+    static const MString cmdFmt("editRenderLayerGlobals -currentRenderLayer ^1s");
     if (renderLayerNode != MFnRenderLayer::currentLayer()) {
         MStatus        status;
         MFnRenderLayer renderLayerFn(renderLayerNode, &status);


lib/mayaUsd/commands/PullPushCommands.cpp
```diff
@@ -264,54 +296,67 @@ MStatus MergeToUsdCommand::doIt(const MArgList& argList)
...
-    if (exportOptions.length() > 0) {
-        status = PXR_NS::UsdMayaJobExportArgs::GetDictionaryFromEncodedOptions(
-            exportOptions, &userArgs);
-        CHECK_MSTATUS_AND_RETURN_IT(status);
-    }
+    const MStringArray exportOptions = parseTextArrayArg(argData, kExportOptionsFlag);
+    const bool         hasExportOptions = exportOptions.length() > 0;
+
+    if (hasExportOptions && (exportOptions.length() != dagPaths.length()))
+        reportError("You must provide the same number of dag objets and exportOptions");
```
pierrebai-adsk: Should it not return here? As I understand it, if there are fewer export options then it is invalid? Although I could see that one may want to use the same options for all.

pierrebai-adsk: Checked the code below, it must return here, since you access the array of export options below.

lib/mayaUsd/commands/PullPushCommands.cpp
```diff
@@ -264,54 +296,67 @@ MStatus MergeToUsdCommand::doIt(const MArgList& argList)
...
+    std::vector<PushToUsdArgs> mergeArgsVect;
+    mergeArgsVect.reserve(dagPaths.length());
+
+    for (uint i = 0; i < dagPaths.length(); i++) {
+        PXR_NS::VtDictionary dagUserArgs;
+
+        if (hasExportOptions) {
+            status = PXR_NS::UsdMayaJobExportArgs::GetDictionaryFromEncodedOptions(
+                exportOptions[i], &dagUserArgs);
+
```
pierrebai-adsk: As I said, maybe you could cap the index used so that if fewer options are given, you use the last one for all remaining DAG objects. OTOH, requiring 1:1 and matching count is fine by me.

lib/mayaUsd/commands/PullPushCommands.cpp
```diff
@@ -264,54 +296,67 @@ MStatus MergeToUsdCommand::doIt(const MArgList& argList)
...
     // Scope the undo item recording so we can undo on failure.
     {
         OpUndoItemRecorder undoRecorder(_undoItemList);
 
-        auto& manager = PXR_NS::PrimUpdaterManager::getInstance();
-        status = manager.mergeToUsd(dagNode, pulledPath, userArgs) ? MS::kSuccess : MS::kFailure;
+        auto&      manager = PXR_NS::PrimUpdaterManager::getInstance();
+        const auto mergedPaths = manager.mergeToUsd(mergeArgsVect);
+        status = mergedPaths.size() == mergeArgsVect.size() ? MS::kSuccess : MS::kFailure;
```
pierrebai-adsk, line 368: I usually prefer have parenthesis to make order of evaluation clear. So i'd put parentheses around the == check.

lib/mayaUsd/fileio/jobs/writeJob.cpp
```diff
@@ -107,106 +117,144 @@ static TfToken _GetFallbackExtension(const TfToken& compatibilityMode)
     return UsdMayaTranslatorTokens->UsdFileExtensionDefault;
 }
 
+/// Fallback USD unit (centimeters).
+static double _GetMetersPerUnitFallback() { return UsdGeomLinearUnits::centimeters; }
+
+/// Converts Maya units to metersPerUnit values used in USD metadata.
+static double _ConvertMayaUnitToMetersPerUnit(MDistance::Unit mayaUnit)
```
pierrebai-adsk: There was already UsdMayaUtil::ConvertMDistanceUnitToUsdGeomLinearUnit(), but that does raise a coding error when the units are not found. We could remove the coding error if that is why you did not use it.

lib/mayaUsd/fileio/jobs/writeJob.cpp
```diff
@@ -107,106 +117,144 @@ static TfToken _GetFallbackExtension(const TfToken& compatibilityMode)
...
-            "change up-axis and units",
-            _prepareCommands(stage, upAxisOption, unitsOption))
+    /// Constructs AutoUndoCommands that changes optionally maya upAxis or metersPerUnit
+    AutoUpAxisAndUnitsChanger(const TfToken* upAxis, const double* metersPerUnit)
+        : AutoUndoCommands("change up-axis and units", _prepareCommands(upAxis, metersPerUnit))
     {
     }
 
 private:
-    static double _convertOptionUnitsToUSDUnits(const TfToken& unitsOption)
```
pierrebai-adsk, line 125: Ah, it seems we already had the duplicated code...

lib/mayaUsd/fileio/jobs/writeJob.cpp
```diff
@@ -107,106 +117,144 @@ static TfToken _GetFallbackExtension(const TfToken& compatibilityMode)
...
+/// Fallback USD unit (centimeters).
+static double _GetMetersPerUnitFallback() { return UsdGeomLinearUnits::centimeters; }
+
+/// Converts Maya units to metersPerUnit values used in USD metadata.
+static double _ConvertMayaUnitToMetersPerUnit(MDistance::Unit mayaUnit)
+{
+    switch (mayaUnit) {
+    case MDistance::kMillimeters: return UsdGeomLinearUnits::millimeters;
+    case MDistance::kCentimeters: return UsdGeomLinearUnits::centimeters;
+    case MDistance::kMeters: return UsdGeomLinearUnits::meters;
```
pierrebai-adsk: I see you dropped the decimeter support... not sure if this is right... although the UsdMayaUtil also does not support decimeters.... but the old function below did.

lib/mayaUsd/fileio/jobs/writeJob.cpp
```diff
@@ -107,106 +117,144 @@ static TfToken _GetFallbackExtension(const TfToken& compatibilityMode)
...
 private:
-    static double _convertOptionUnitsToUSDUnits(const TfToken& unitsOption)
-    {
-        static std::map<TfToken, double> unitsConversionMap
-            = { { UsdMayaJobExportArgsTokens->nm, UsdGeomLinearUnits::nanometers },
-                { UsdMayaJobExportArgsTokens->um, UsdGeomLinearUnits::micrometers },
-                { UsdMayaJobExportArgsTokens->mm, UsdGeomLinearUnits::millimeters },
-                { UsdMayaJobExportArgsTokens->cm, UsdGeomLinearUnits::centimeters },
-                // Note: there is no official USD decimeter units, we have to roll our own.
-                { UsdMayaJobExportArgsTokens->dm, 0.1 },
```
pierrebai-adsk, line 133: Support for decimeter dropped... not sure if this is right, would have to ask product manager.

lib/mayaUsd/fileio/jobs/writeJob.cpp
```diff
@@ -261,49 +306,189 @@ class AutoUpAxisAndUnitsChanger : public MayaUsd::AutoUndoCommands
...
+        auto argsAreCompatible = [&](const UsdMayaJobExportArgs& args) -> bool {
+            return (_WantedUSDUpAxis(args.upAxis) == usdUpAxis)
+                && (_WantedUSDMetersPerUnit(args.unit) == usdMetersPerUnit)
+                && (_WantedRenderLayerNode(args.renderLayerMode) == renderLayer);
+        };
+
+        std::set<std::string> seenFileNames;
+        for (auto& job : jobs) {
+            // Verify that we are not writing twice to the same filename.
+            if (!seenFileNames.insert(job->_fileName).second) {
```
pierrebai-adsk, line 406: Maybe in the future we could support same file name in append mode?

I guess it is safer to have different files for now, due to possible collisions.

lib/mayaUsd/commands/PullPushCommands.cpp
```diff
@@ -264,54 +296,67 @@ MStatus MergeToUsdCommand::doIt(const MArgList& argList)
...
         const int index = 0;
-        userArgs[UsdMayaPrimUpdaterArgsTokens->ignoreVariants]
+        commandUserArgs[UsdMayaPrimUpdaterArgsTokens->ignoreVariants]
             = argData.flagArgumentBool(kIgnoreVariantsFlag, index);
     }
 
+    std::vector<PushToUsdArgs> mergeArgsVect;
+    mergeArgsVect.reserve(dagPaths.length());
+
+    for (uint i = 0; i < dagPaths.length(); i++) {
```
pierrebai-adsk: uint does not exist in all compilers. You need to use unsigned int

pierrebai-adsk (changes requested):
Needs a few fixes, but otherwise great work.

lib/mayaUsd/fileio/jobs/writeJob.cpp
pierrebai-adsk: Talked to people internally: we want decimeter support back. We had added it at the request of ours users,

jufrantz:
Thank you for the review @pierrebai-adsk
I’m currently out of office but will update the PR ASAP.

lib/mayaUsd/commands/PullPushCommands.cpp
jufrantz: I had indeed missed the return statement, this is now fixed. Thanks for catching this bug.
I also agree that it's convenient to use the same options across all merge operations. I updated the code to allow this.
https://github.com/Autodesk/maya-usd/pull/4279/commits/cdb6f814007ca081cf5c1fca3961cc9ccad34fc0

lib/mayaUsd/commands/PullPushCommands.cpp
jufrantz: Code was updated to allow a single set of options reused across all merges: https://github.com/Autodesk/maya-usd/pull/4279/commits/cdb6f814007ca081cf5c1fca3961cc9ccad34fc0

lib/mayaUsd/commands/PullPushCommands.cpp
jufrantz: Fixed. I updated the original commit so that all commits will compile. https://github.com/Autodesk/maya-usd/blob/f9910a6a4d754b83c5c4e740fc25f3dac0e46e5f/lib/mayaUsd/commands/PullPushCommands.cpp#L186

lib/mayaUsd/fileio/jobs/writeJob.cpp
jufrantz: I had the same thought when I added this check. For now I decided to keep it simple by enforcing unique filenames.
Also, batching to the same file in append-mode feels like a niche use case; this can be certainly revisit it later if there’s a concrete need.

lib/mayaUsd/fileio/jobs/writeJob.cpp
jufrantz: Thank you for pointing out `UsdMayaUtil::ConvertMDistanceUnitToUsdGeomLinearUnit()`. I’ve replaced my local impl, I simply missed the existing one. https://github.com/Autodesk/maya-usd/pull/4279/commits/c733a53d9fa7064f5d89f26d339599a9411e0822

The coding error seems appropriate here: we only call it with the values returned by `MDistance::uiUnit()` and `MDistance::internalUnit()`, I believe an error would surface only if a valid enum value is not handled yet by the function.

lib/mayaUsd/commands/PullPushCommands.cpp
jufrantz: Fixed in https://github.com/Autodesk/maya-usd/pull/4279/commits/cdb6f814007ca081cf5c1fca3961cc9ccad34fc0

lib/mayaUsd/fileio/jobs/writeJob.cpp
jufrantz: I might be missing something, but I don’t believe there is a behaviour change in this commit regarding `unit=dm` option.

`_ConvertMayaUnitToMetersPerUnit` kind of replaced `AutoUpAxisAndUnitsChanger::_convertMayaUnitsToOptionUnits`, it is only used when reading the current Maya preferences. Up to Maya 2026 SDK, MDistance::Unit does not define kDecimeter, so there would be no case to handle here.

Decimeter support is still handled when parsing export options by the new function [_WantedUSDMetersPerUnit()](https://github.com/Autodesk/maya-usd/blob/7f469f5687eb28b066dfb9edd1b01ee1f0c5b095/lib/mayaUsd/fileio/jobs/writeJob.cpp#L142). When `unit=dm` is given:
-  [AutoUpAxisAndUnitsChanger](https://github.com/Autodesk/maya-usd/blob/7f469f5687eb28b066dfb9edd1b01ee1f0c5b095/lib/mayaUsd/fileio/jobs/writeJob.cpp#L172C68-L172C81) is constructed with `metersPerUnit` set to 0.1, so the scene is still scaled to fit dm.
-  `metersPerUnit=0.1` is still authored in the USD stage, here [writeJob.cpp#L503](https://github.com/Autodesk/maya-usd/blob/7f469f5687eb28b066dfb9edd1b01ee1f0c5b095/lib/mayaUsd/fileio/jobs/writeJob.cpp#L503).

There is also a pre‑existing test `testExportUnitsDecimeters` that verifies those two aspects and it still passes on this branch.

lib/mayaUsd/fileio/jobs/writeJob.cpp
jufrantz: As replied previously, decimeter should still be supported, but the logic of the code changed.

jufrantz:
I’ve addressed the requested changes and fixes.
The decimeter support is unchanged from my initial implementation, but @pierrebai-adsk , if you spot anything I’ve missed, please let me know. Otherwise this PR is ready to merge from my side.

lib/mayaUsd/fileio/primUpdaterManager.h
```diff
@@ -38,19 +39,45 @@ class OrphanedNodesManager;
 
 PXR_NAMESPACE_OPEN_SCOPE
 
+/// \brief Arguments used by PrimUpdaterManager pushToUsd operations mergeToUsd and duplicateToUsd.
+struct PushToUsdArgs
```
pierrebai-adsk: There was a failure on Windows. I think this entire structore need to be MAYAUSD_CORE_PUBLIC.

pierrebai-adsk: Otherwise the Python wrapper DLL fails to link.

jufrantz: This should be fixed, hopefully the last Windows issue :crossed_fingers: Thanks!
https://github.com/Autodesk/maya-usd/blob/ff9c06b642cad02fe70fc815bc2a2a559f4407ea/lib/mayaUsd/fileio/primUpdaterManager.h#L43

pierrebai-adsk:
The test testCacheToUsd failed:
```
11:45:45          self.assertRegexpMatches(self.stage.GetRootLayer().ExportToString(), 'payload = @.*testCacheToUsd/cache.usda')
11:45:45      AssertionError: Regex didn't match: 'payload = @.*testCacheToUsd/cache.usda' not found in '#sdf 1.4.32\n\ndef Xform "CacheParent" (\n    variants = {\n        string animation = "Cache"\n    }\n    prepend variantSets = "animation"\n)\n{\n    variantSet "animation" = {\n        "Cache" {\n            def Xform "cachePrimName" (\n                prepend payload = @W:\\build\\RelWithDebInfo\\test\\lib\\mayaUsd\\fileio\\testCacheToUsdOutput\\testCacheToUsd\\cache.usda@</cachePrimName>\n            )\n            {\n            }\n\n        }\n        "Rig" {\n            def MayaReference "Reference1"\n            {\n                bool mayaAutoEdit = 0\n                string mayaNamespace = "simpleSphere"\n                asset mayaReference = @simpleSphere.ma@\n            }\n\n        }\n    }\n}\n\n'

```

Something changed that makes the payload have backslashes instead of forward slashes. I don't know if you can easily identify what could have affected this in your changes? I might take a look this afternoon, but you might have an idea.

jufrantz:
> The test testCacheToUsd failed:
> 
> ```
> 11:45:45          self.assertRegexpMatches(self.stage.GetRootLayer().ExportToString(), 'payload = @.*testCacheToUsd/cache.usda')
> 11:45:45      AssertionError: Regex didn't match: 'payload = @.*testCacheToUsd/cache.usda' not found in '#sdf 1.4.32\n\ndef Xform "CacheParent" (\n    variants = {\n        string animation = "Cache"\n    }\n    prepend variantSets = "animation"\n)\n{\n    variantSet "animation" = {\n        "Cache" {\n            def Xform "cachePrimName" (\n                prepend payload = @W:\\build\\RelWithDebInfo\\test\\lib\\mayaUsd\\fileio\\testCacheToUsdOutput\\testCacheToUsd\\cache.usda@</cachePrimName>\n            )\n            {\n            }\n\n        }\n        "Rig" {\n            def MayaReference "Reference1"\n            {\n                bool mayaAutoEdit = 0\n                string mayaNamespace = "simpleSphere"\n                asset mayaReference = @simpleSphere.ma@\n            }\n\n        }\n    }\n}\n\n'
> ```
> 
> Something changed that makes the payload have backslashes instead of forward slashes. I don't know if you can easily identify what could have affected this in your changes? I might take a look this afternoon, but you might have an idea.

Yes, I think I know why. In `testCacheToUsd` I added a temporary subfolder so we could export multiple cache files for the batched-export test cases. Previously the cache went to `testCacheToUsd.usda`; now it’s `testCacheToUsd/<name>.usda`. I wrongly use `os.path.join` to generate this payload path, it generates this wrong asset path on windows.
I look into fixing this now.

jufrantz:
> > The test testCacheToUsd failed:
> > ```
> > 11:45:45          self.assertRegexpMatches(self.stage.GetRootLayer().ExportToString(), 'payload = @.*testCacheToUsd/cache.usda')
> > 11:45:45      AssertionError: Regex didn't match: 'payload = @.*testCacheToUsd/cache.usda' not found in '#sdf 1.4.32\n\ndef Xform "CacheParent" (\n    variants = {\n        string animation = "Cache"\n    }\n    prepend variantSets = "animation"\n)\n{\n    variantSet "animation" = {\n        "Cache" {\n            def Xform "cachePrimName" (\n                prepend payload = @W:\\build\\RelWithDebInfo\\test\\lib\\mayaUsd\\fileio\\testCacheToUsdOutput\\testCacheToUsd\\cache.usda@</cachePrimName>\n            )\n            {\n            }\n\n        }\n        "Rig" {\n            def MayaReference "Reference1"\n            {\n                bool mayaAutoEdit = 0\n                string mayaNamespace = "simpleSphere"\n                asset mayaReference = @simpleSphere.ma@\n            }\n\n        }\n    }\n}\n\n'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Something changed that makes the payload have backslashes instead of forward slashes. I don't know if you can easily identify what could have affected this in your changes? I might take a look this afternoon, but you might have an idea.
> 
> Yes, I think I know why. In `testCacheToUsd` I added a temporary subfolder so we could export multiple cache files for the batched-export test cases. Previously the cache went to `testCacheToUsd.usda`; now it’s `testCacheToUsd/<name>.usda`. I wrongly use `os.path.join` to generate this payload path, it generates this wrong asset path on windows. I look into fixing this now.

I looked at it more closely. It seems payloads had backslashes on Windows even before these changes. What changed is that the regex only matched one component of the asset path, so the slash style didn’t matter. I would propose to change the test to assert on the actual `cacheFile`, like this:

```diff
@@ -317,7 +317,7 @@ class CacheToUsdTestCase(unittest.TestCase):
         if relativePath:
             if self.stage.GetRootLayer().anonymous:
                 self.assertNotIn('payload = @testCacheToUsd/cache.usda', self.stage.GetRootLayer().ExportToString())
-                self.assertRegexpMatches(self.stage.GetRootLayer().ExportToString(), 'payload = @.*testCacheToUsd/cache.usda')
+                self.assertIn('payload = @' + cacheFile, self.stage.GetRootLayer().ExportToString())
                 self.makeRootLayerNotAnonymous()
                 mayaUsd.lib.Util.updatePostponedRelativePaths(self.stage.GetRootLayer())
```
This passes on linux but  I dont have a windows box available.
What to you think @pierrebai-adsk ? I can push this and we see what it gives ?

jufrantz:
@pierrebai-adsk  It looks like the CI failed due to a timeout, it isn’t a code error?

I took the opportunity to push two last-minute commits: 
- Minor optimization in `mayaUsdMergeToUsd` when a single `exportOptions` is given for multiple objects (avoids redundant parsing).  https://github.com/Autodesk/maya-usd/pull/4279/commits/5a8e9219fef42b9b387c345b3d0429fd1ec41e80
- A unit test to verify `exportOptions` parsing with multiple objects (single vs per-object options). https://github.com/Autodesk/maya-usd/pull/4279/commits/f3c9cd04ee377b441c1f6c9b09b46e80dab69952

pierrebai-adsk:
We will do some QA on it internally before approving

jufrantz:
@pierrebai-adsk, I've fixed a typo introduced during the `-renderLayerMode` option refactor https://github.com/Autodesk/maya-usd/pull/4279/commits/ff9c06b642cad02fe70fc815bc2a2a559f4407ea.
I expected the unit tests to catch this kind of issue, but I’ve just realized the relevant tests were disabled by default.

santosd:
Hi @jufrantz, I have been running some tests on my end with the scene file and python script that was provided. One thing that I have noticed when using the script test_merge_four_animated_rhinos.py is that I lose the reference and the connection to the reference. In the screen shot below you will notice that the three cached files at the top are done in batch and the last one was done through the menu and you will notice that the reference node is kept there. I am wondering if I am possibly missing an option or if this is how batch is designed to work?
image

Another question I have, is in regards to the UI way of doing this, will these changes have any affect on how that works?

jufrantz:
> Hi @jufrantz, I have been running some tests on my end with the scene file and python script that was provided. One thing that I have noticed when using the script test_merge_four_animated_rhinos.py is that I lose the reference and the connection to the reference. In the screen shot below you will notice that the three cached files at the top are done in batch and the last one was done through the menu and you will notice that the reference node is kept there. I am wondering if I am possibly missing an option or if this is how batch is designed to work? image
> 
> Another question I have, is in regards to the UI way of doing this, will these changes have any affect on how that works?

Hi @santosd, thanks for looking at this,

I see the mismatch. In my test script, the cache destination prim is the same as the source `MayaReference`. After the merge, the `MayaReference` prim ends up with the `USD reference` to the cache which makes it an `Xform`. That’s why it no longer appears in the outliner as a maya reference. This setup was chosen for coding simplicity, not as a real-world case. I think the maya-usd `Cache To USD` dialog doesn’t even allow these settings.

To make the test closer to what happens via the menu, you can adjust the script so the USD cache is referenced on a new prim. For example, in `test_merge_four_animated_rhinos.py`, you can generate a new name by appending "_cache" like this:

```diff
@@ -12,7 +12,7 @@
     pathAndOptions = []
     for dagName in cmds.ls(sl=True):
         pathAndOptions.append((dagName, cacheOptions | {
-            "rn_primName": dagName, "rn_layer": f"{tmpDir}/{dagName}.usd"}))
+            "rn_primName": dagName + "_cache", "rn_layer": f"{tmpDir}/{dagName}.usd"}))
 
     with OpUndoItemList():
         start = perf_counter()
```

Regarding the GUI: this PR doesn’t add “batched merge” to the Maya menus. Existing actions still work on a single object and shouldn’t be affected by these changes. GUI exposure could be added later, I think it would fit well into the current maya-usd Bulk Actions UX:
- Bulk `Merge to USD...` could reuse the existing merge dialog, the same options would apply to all selected items.
- Bulk `Cache to USD...` would need some extra design work, at least for supporting multiple destination cache files selection.

santosd:
@jufrantz Thank you for the response. I didn't come across any other issues in my tests. I think we can go ahead and merge these changes @seando-adsk and @pierrebai-adsk .

seando-adsk:
Thanks for testing and the feedback. @pierrebai-adsk Can you finish the code review and then run the preflight.

jufrantz:
I’ve just rebased onto dev and resolved a conflict with `testCacheToUsd.py` (https://github.com/Autodesk/maya-usd/pull/4279/commits/c7e1d3a09b1ccc97a57243d6f263652435f3aad2).
@pierrebai-adsk let me know if there’s anything else needed from my side for approval.

seando-adsk:
@jufrantz Sorry the delay. QA has internally tested the changes and given approval. Pierre is currently away so we are waiting for him to come back to finish the code review. Then we'll get this one merged.

PR #4280, opened 2025-07-25
EMSUSD-2681 - Fix duplicate not duplicating connections to NodeGraph outputs.

frohnej-adsk:
The bug could be reproduced easily in LookdevX:
 - Create an add node inside of a compound
 - Connect the add node to an output of the compound
 - Copy/Paste the compound
 - Connection is missing in the copied compound

### DIFF
--- a/test/lib/ufe/testClipboard.py
+++ b/test/lib/ufe/testClipboard.py
@@ -546,22 +546,24 @@ class ClipboardHandlerTestCase(unittest.TestCase):
         
         # Create a NodeGraph and add an output attribute.
         stage.DefinePrim('/Material1', 'Material')
-        materialItem = ufeUtils.createItem(psPathStr + ',/Material1')
+        stage.DefinePrim('/Material2', 'Material')
+        materialItem1 = ufeUtils.createItem(psPathStr + ',/Material1')
+        materialItem2 = ufeUtils.createItem(psPathStr + ',/Material2')
 
         # Create two shaders within the NodeGraph.
-        nodeDef = ufe.NodeDef.definition(materialItem.runTimeId(), 'ND_add_color3')
-        srcItem = nodeDef.createNode(materialItem, ufe.PathComponent('src1'))
-        dstItem = nodeDef.createNode(materialItem, ufe.PathComponent('dst1'))
+        nodeDef = ufe.NodeDef.definition(materialItem1.runTimeId(), 'ND_add_color3')
+        srcItem = nodeDef.createNode(materialItem1, ufe.PathComponent('src1'))
+        dstItem = nodeDef.createNode(materialItem1, ufe.PathComponent('dst1'))
 
         # Connect the shaders.
-        connectionHandler = ufe.RunTimeMgr.instance().connectionHandler(materialItem.runTimeId())
+        connectionHandler = ufe.RunTimeMgr.instance().connectionHandler(materialItem1.runTimeId())
         srcOutput = ufe.Attributes.attributes(srcItem).attribute('outputs:out')
         dstInput = ufe.Attributes.attributes(dstItem).attribute('inputs:in1')
         connection = connectionHandler.connect(srcOutput, dstInput)
         self.assertIsNotNone(connection)
 
         # Copy both shaders.
-        ch = ufe.ClipboardHandler.clipboardHandler(materialItem.runTimeId())
+        ch = ufe.ClipboardHandler.clipboardHandler(materialItem1.runTimeId())
         ufe.ClipboardHandler.preCopy()
         copyCmd = ch.copyCmd_(ufe.Selection([srcItem, dstItem]))
         copyCmd.execute()
@@ -570,12 +572,12 @@ class ClipboardHandlerTestCase(unittest.TestCase):
         ufe.GlobalSelection.get().clear()
 
         # Paste both shaders.
-        pasteCmd = ch.pasteCmd_(materialItem)
+        pasteCmd = ch.pasteCmd_(materialItem2)
         pasteCmd.execute()
         self.assertIsNotNone(pasteCmd.targetItems())
         self.assertEqual(2, len(pasteCmd.targetItems()))
-        pastedSrcItem = pasteCmd.targetItems()[0]
-        pastedDstItem = pasteCmd.targetItems()[1]
+        pastedSrcItem = ufe.Hierarchy.createItem(materialItem2.path() + srcItem.nodeName())
+        pastedDstItem = ufe.Hierarchy.createItem(materialItem2.path() + dstItem.nodeName())
 
         # Verify that the connection was copied.
         self.assertEqual(1, len(connectionHandler.sourceConnections(pastedDstItem).allConnections()))


lib/usdUfe/ufe/UsdUndoDuplicateSelectionCommand.cpp
```diff
@@ -170,14 +170,9 @@ bool UsdUndoDuplicateSelectionCommand::updateSdfPathVector(
     for (size_t i = 0; i < pathVec.size(); ++i) {
         const PXR_NS::SdfPath& path = pathVec[i];
         PXR_NS::SdfPath        finalPath = path;
-        // Paths are lexicographically ordered, this means we can search quickly for bounds of
-        // candidate paths.
-        auto itPath = otherPairs.lower_bound(finalPath);
-        if (itPath != otherPairs.begin()) {
-            --itPath;
-        }
-        const auto endPath = otherPairs.upper_bound(finalPath);
```
frohnej-adsk: The bug was caused by these `lower_bound` and `upper_bound` optimizations. 

## Cause of the bug

Assume the following hierarchy:
```
mtl
\---standard_surface1
    \---compound1
        \---add1
```
Let `/mtl/standard_surface1/compound1.outputs:out` be connected to `/mtl/standard_surface1/compound1/add1.outputs:out`.

Upon copying `/mtl/standard_surface1/compound1` to the clipboard, `updateSdfPathVector()` will be called for the attribute `/compound1.outputs:out` within the clipboard with the following parameters:
- `pathVec       = ["/compound1/add1.outputs:out"]`
- `duplicatePair = ("/mtl/standard_surface1/compound1", "/compound1")`
- `otherPairs    = [("/mtl/standard_surface1/compound1", "/compound1")]`

Note that `pathVec` already contains the correct updated path. This case is supposed to be handled by the `if (*itPath == duplicatePair)` branch. However, due to the optimization, that line is never hit and we ended up erroneously deleting the connection:

SdfPaths are compared lexicographically. Thus, `finalPath == /compound1/add1.outputs:out` is less than `/mtl/standard_surface1/compound1`.

- `otherPairs.lower_bound(finalPath)` returns the first element `e` such that `finalPath <= e`. That's `/mtl/standard_surface1/compound1`.
- `otherPairs.upper_bound(finalPath)` returns the first element `e` such that `finalPath < e`. That's `/mtl/standard_surface1/compound1`.

I.e., `lower_bound(finalPath) == upper_bound(finalPath)`, which causes the loop to never be entered and the connection to be deleted because it's assumed to be external.

## Optimization and proposed fix

I'm still struggling to wrap my head around the optimization. I understand why it fails in this case but I don't fully understand how it works in regular cases. The STL `lower_bound()` and `upper_bound()` methods don't really do what their name suggests, which makes it kind of confusing (see https://stackoverflow.com/a/67551612) and makes me wonder if it actually works as intended.

Thus, I simply removed the optimization in this PR. That might be good enough and it makes the code easier to read and more similar to the other duplicate command ([lib\mayaUsd\ufe\UsdUndoDuplicateSelectionCommand.cpp](https://github.com/Autodesk/maya-usd/blob/2c746b5670bfebd91a40c803a678d4f69ad2a19d/lib/mayaUsd/ufe/UsdUndoDuplicateSelectionCommand.cpp#L201)). If you can think of a good way to fix the bug while keeping the optimization, let me know!

frohnej-adsk (commented):
All builds are green except for `2023 (Windows)`, where `305 - GTest:AL_USDTransactionTests` timed out.

lib/usdUfe/ufe/UsdUndoDuplicateSelectionCommand.cpp
JGamache-autodesk: The big question is "why is a remapping function being called on something that is already remapped". The updated code currently "stumbles" across it via the `duplicatePair` entry because it iterates everything, which is quite bad performance wise.

Let's try to explain the weird code you are seeing. The map contains prefixes that are the shortest possible, so even though you duplicated both `/mtl/standard_surface1/compound1` and `/mtl/standard_surface1/compound1/add1` the map will only contain `/mtl/standard_surface1/compound1`.

Now, let's say I want to know what happened to `/mtl/standard_surface1/compound1/add1` and search for it using `lower_bound()`. I then get an iterator that is 1 element past `/mtl/standard_surface1/compound1` so I need to back up one element when possible to include the prefix in the loop. I will end up on the `/mtl/standard_surface1/compound1` entry which will allow me to remap correctly using `ReplacePrefix`.

If, instead I want to know about `/mtl/standard_surface1/compound1`, then the iterator returned by `lower_bound()` will actually directly point to it. We still go back one item (for nothing), but will get it on the second iteration. On the second iteration we will hit the equal case and directly rewrite to the destination path in pathVec.

In all cases `upper_bound()` will either be equal to `lower_bound()` if `finalPath` was not directly in the map, or will point one element past it if it was found, which is exactly what we want.

It is an interesting dance, but it actually works. This obviously requires a bit more code documentation.



But let's get back to your problem. The case you are seeing is that you have a connection on the compound boundary that is pointing inside the compound and should be left alone. If you look at what happens in the caller on the exit of that function you either remap modified paths, or remove the attribute. There is no "leave it alone" option.

This means the caller should check for that and remove paths that are obviously correct:
```c++
void UsdUndoDuplicateSelectionCommand::execute()
{
...
                    attr.GetConnections(&sources);
                    // Sources that are already inside the duplicate should be ignored:
                    const auto new_end = std::remove_if(sources.begin(), sources.end(), [&duplicatePair](const auto& path){ return path.HasPrefix(duplicatePair.second); });
                    sources.erase(new_end, sources.end());
                    // Now we know we only have external sources left:
                    if (!sources.empty() && updateSdfPathVector(sources, duplicatePair, stageData.second)) {
...
                    // Need the exact same for properties
...
}
```

JGamache-autodesk (changes requested):

lib/usdUfe/ufe/UsdUndoDuplicateSelectionCommand.cpp
JGamache-autodesk: Not good enough. Too brutal. If there is a mix of internal and external connections we might end up calling `ClearConnections()`. I suspect we need to process connections one by one.
What we want is:
```
finalSources = []
for cnx in sources:
    if cnx.HasPrefix(duplicatePair.second):
        # Internal, keep it:
        finalSources.push_back(cnx)
        continue  # Internal

    remappedCnx = None
    it = stageData.second.lower_bound(cnx)
    if it.first == cnx:
        # Direct match to one of the duplicate sources: remap
        # Note. Probably won't happen.
        remappedCnx = it.second
    else:
        # Can we find the prefix by going back one item?
        if it != stageData.second.begin():
            --it
            if cnx.HasPrefix(it.first):
                remappedCnx = cnx.RemapPrefix(it.first, it.second)
                
    if remappedCnx:
        # Remapped from external to internal:
        finalSources.push_back(remappedCnx)
    
    # Else that external connection gets removed
    
# Identical as before:
if (sources.empty()) {
    attr.ClearConnections();
    if (!attr.HasValue() && !UsdShadeNodeGraph(attr.GetPrim())) {
        p.RemoveProperty(prop.GetName());
    }
} else {
    attr.SetConnections(sources);
}
```

Could probably be done by the existing algo, by ignoring the `changed` return value in the caller and by keeping items that are already remapped by adding a `(finalPath.HasPrefix(duplicatePair.second) || finalPath == duplicatePair.second)` check at the beginning of the loop to `continue` and leave the item unchanged in the final vector.

frohnej-adsk: Thank you, this was very helpful to understand the idea of the optimization!

This is my current understanding and what I implemented in my latest commit. Please double-check if it's correct.

The goal is to check if any path in the DuplicatePathsMap is a prefix of the path we want to update. Due to lexicographical ordering, the prefix of a path is always less than or equal to the path itself. We can utilize that to check only a single candidate: The last/greatest path that's less then or equal to the path we want to update. This candidate is right before the path returned by `upper_bound()`, which is the first path that's greater than the path we want to update.

seando-adsk:
@pierrebai-adsk Can you code review these changes?

lib/usdUfe/ufe/UsdUndoDuplicateSelectionCommand.cpp
```diff
@@ -161,47 +161,72 @@ Ufe::SceneItemList UsdUndoDuplicateSelectionCommand::targetItems() const
...
+            // All paths in the source set are greater, so none of them can be a prefix.
+            indicesToRemove.push_front(i);
+            continue;
         }
-        if (isExternalPath) {
-            hasChanged = true;
+
+        // Check if our candidate is a prefix of the referenced path.
+        const auto lastSmallerPath = std::prev(firstLargerPath);
+        const bool isInOriginalSet = referencedPath.HasPrefix(lastSmallerPath->first);
```
JGamache-autodesk, line 215: Need to check also for `referencedPath == lastSmallerPath` because the `HasPrefix` function uses the strictest definition and does not consider equal paths.

lib/usdUfe/ufe/UsdUndoDuplicateSelectionCommand.cpp
```diff
@@ -161,47 +161,72 @@ Ufe::SceneItemList UsdUndoDuplicateSelectionCommand::targetItems() const
...
         }
+
+        // Update the path to point to the respective path in the duplicate set.
+        referencedPath.ReplacePrefix(lastSmallerPath->first, lastSmallerPath->second);
+        referencedPathsChanged = true;
     }
+
     for (size_t toRemove : indicesToRemove) {
-        pathVec.erase(pathVec.cbegin() + toRemove);
+        referencedPaths.erase(referencedPaths.cbegin() + toRemove);
```
JGamache-autodesk, line 229: For clarity, would be good to mention that this works because the indices are in reverse order due to the use of `push_front`. This means no need to adjust because the vector has closed the gap of a deleted item.

lib/usdUfe/ufe/UsdUndoDuplicateSelectionCommand.cpp
pierrebai-adsk: I don't understand, your comment does not refer to any code, there is no equality check?

JGamache-autodesk: ```const bool isInOriginalSet = referencedPath.HasPrefix(lastSmallerPath->first) || referencedPath == lastSmallerPath->first;```

frohnej-adsk: @JGamache-autodesk I should have checked that before, thanks for being vigilant!

I just checked and `path.HasPrefix(path)` returns true. I also checked the [SdfPath::HasPrefix()](https://github.com/PixarAnimationStudios/OpenUSD/blob/d0d5c33d5655f206a02b58311efa841890fb3894/pxr/usd/sdf/path.cpp#L554) implementation, which does consider equality, so no explicit check should be needed.


I updated the code comments accordingly. See 37dc249. Does that resolve your question as well @pierrebai-adsk?

lib/usdUfe/ufe/UsdUndoDuplicateSelectionCommand.cpp
frohnej-adsk: See 37dc249.

test/lib/ufe/testClipboard.py
```diff
@@ -488,5 +488,99 @@ def testClipboardErrorConditions(self):
...
+        pasteCmd.execute()
+        self.assertIsNotNone(pasteCmd.targetItems())
+        self.assertEqual(1, len(pasteCmd.targetItems()))
+        pastedNodeGraph = pasteCmd.targetItems()[0]
+
+        # Verify that the connection was copied.
+        self.assertEqual(1, len(connectionHandler.sourceConnections(pastedNodeGraph).allConnections()))
+
+
+    def testClipboardCopyConnections(self):
```
frohnej-adsk, line 541: I assumed this was already covered by tests, but it wasn't. Added another regression test for the regular case.

pierrebai-adsk (approved):

lib/usdUfe/ufe/UsdUndoDuplicateSelectionCommand.cpp
JGamache-autodesk: Sorry, got confused with Ufe::Path which has this issue. All good now.

JGamache-autodesk (approved):
LGTM. Thanks!

test/lib/ufe/testClipboard.py
```diff
@@ -570,12 +572,12 @@ def testClipboardCopyConnections(self):
...
         # Paste both shaders.
-        pasteCmd = ch.pasteCmd_(materialItem)
+        pasteCmd = ch.pasteCmd_(materialItem2)
         pasteCmd.execute()
         self.assertIsNotNone(pasteCmd.targetItems())
         self.assertEqual(2, len(pasteCmd.targetItems()))
-        pastedSrcItem = pasteCmd.targetItems()[0]
-        pastedDstItem = pasteCmd.targetItems()[1]
+        pastedSrcItem = ufe.Hierarchy.createItem(materialItem2.path() + srcItem.nodeName())
+        pastedDstItem = ufe.Hierarchy.createItem(materialItem2.path() + dstItem.nodeName())
```
frohnej-adsk, line 580: The newly added test failed on Mac and Linux. I assume it was due to this assumption about the order of `targetItems()`.

JGamache-autodesk (approved):

PR #4281, opened 2025-07-25
Fix point instancer crash

pierrebai-adsk:
Apply a fix similar to other places in the same file. Not 100% it is entirely right, as my knowledge of the VP2 render delegate and instancer is not high.

Turns out this does not fix the crash... oh well


seando-adsk:
@pierrebai-adsk What should we do with this PR?

pierrebai-adsk:
I kept it around to not lose info on what I tried, but as I said in my comments, it did not fix the problem.

PR #4282, opened 2025-07-25
EMSUSD-2614 - EMSUSD-2642 - Register Maya Project Tokens to USD AssetResolver and Asset Resolver Presets

samuelliu-adsk:


samuelliu-adsk:
This is the mayaUsd side of the changes, do not merge untill the resolver changes are in

plugin/adsk/plugin/plugin.cpp
```diff
@@ -403,6 +403,26 @@ MStatus initializePlugin(MObject obj)
...
+            "    import maya.cmds as cmds\n"
+            "    import AdskAssetResolver as ar\n"
+            "    directory = cmds.workspace(q=True, directory=True)\n"
+            "    tokenList = cmds.workspace(fileRuleList=True)\n"
+            "    mayaUsdResolver = ar.AssetResolverContextDataManager.RegisterContextData('MayaUSDExtension')\n"
+            "    if mayaUsdResolver is not None:\n"
+            "        mayaUsdResolver.AddStaticToken('Project', directory)\n"
+            "        for token in tokenList:\n"
+            "            tokenValue = cmds.workspace(fileRuleEntry=token)\n"
+            "            mayaUsdResolver.AddStaticToken(token, tokenValue)\n"
```
seando-adsk: Is this identical to the function `include_maya_project_tokens` below? Can't you just call that instead?

plugin/adsk/scripts/mayaUsd_preferenceTab.mel
```diff
@@ -110,6 +110,26 @@ global proc mayaUsd_PrefUntexturedModeChanged()
...
+			// Include Maya project tokens in the USD Asset Resolver.
+			// This is done by calling the python function that does it.
+			// The python function will also register the context data for MayaUSD.
+			python("import mayaUsd_AdskAssetResolver; mayaUsd_AdskAssetResolver.include_maya_project_tokens()");
+		} else {
+			// Remove Maya project tokens from the USD Asset Resolver.
+			// This is done by calling the python function that does it.
+			python("import mayaUsd_AdskAssetResolver; mayaUsd_AdskAssetResolver.unload_maya_project_tokens()");
+		}
+		//python("import mayaUsd_AdskAssetResolver; mayaUsd_AdskAssetResolver.include_maya_project_tokens()");
```
seando-adsk: Why is this line commented?

plugin/adsk/plugin/plugin.cpp
```diff
@@ -403,6 +403,26 @@ MStatus initializePlugin(MObject obj)
     PrimUpdaterManager::getInstance();
 #endif
 
+    // TODO: Put a define here
```
seando-adsk: What define should be used here?

seando-adsk: Is that because of this comment `do not merge until the resolver changes are in`
Anything in this PR that is dependance on code in the AR should be protected in some way using cmake to detect whether the AR exists and then setting a define so this can be conditionally compiled.

seando-adsk (changes requested):

plugin/adsk/plugin/plugin.cpp
samuelliu-adsk: Yes. That sounds like a good idea

plugin/adsk/plugin/plugin.cpp
samuelliu-adsk: Then do I need to add plugin/script to the pythonpath in the module?

